PR #552, opened 2015-07-24
Incorporate some changes from PR#495.

chrisdembia:
These changes only affect python wrapping. @aymanhab could you review and merge?

### DIFF
--- a/OpenSim/Wrapping/Python/CMakeLists.txt
+++ b/OpenSim/Wrapping/Python/CMakeLists.txt
@@ -26,7 +26,8 @@ set(swig_interface_file_fullname
 # it's out of date. Previously, we used a custom target only. But, custom
 # targets are always out of date.
 add_custom_command(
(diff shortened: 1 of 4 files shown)


aymanhab:
@chrisdembia What's the change in CMakeLists for?  I'm also assuming the change to OpenSim::FrameSet::adoptAndAppend was done earlier since it shows as removed by github.

chrisdembia:
> I'm also assuming the change to OpenSim::FrameSet::adoptAndAppend was done earlier since it shows as removed by github.

No I explicitly removed it with this PR; it is redundant since we are taking care of Frames with the swig macro (SET_ADOPT_HELPER).

> What's the change in CMakeLists for?

Just for making the python target recognize that it produces the `opensim.py` file. That way, if the `opensim.py` file got deleted, the target would know to regenerate it.

chrisdembia:
Thank you.

Issue #528, opened 2015-07-13
OpenSim is misusing the SimTK::ReferencePtr class in many places


sherm1:
`SimTK::ReferencePtr` is a smart pointer that conveys *no* ownership. It is used to point at things owned by someone else. It is better than a raw pointer because it self-initializes to `nullptr` and clear itself back to `nullptr` on copy to avoid dangling references from the copy. The point of it is to allow us to avoid writing copy constructors and copy assignment, as we would have to if we used raw pointers.

But since there is no ownership it is *never* OK to invoke `delete` on a `ReferencePtr`! Nor should a `ReferencePtr` ever be set to `new Something()`! There are many instances of both of these errors and they will start producing warnings when PR simbody/simbody#410 is merged. They will have to be fixed very carefully.

sherm1:
Broken `ReferencePtr`: 
ControllerSet::_controlStore
CoordinateLimitForce::_upStep, _loStep
GeometryPath::_maSolver
Model::_assemblySolver,_modelViz,_contactSubsystem,_gravityForce,_forceSubsystem,_matter,_system

chrisdembia:
I see you didn't put Coordinate::_lockFunction in your list:

https://github.com/opensim-org/opensim-core/pull/510/files
https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/SimbodyEngine/Coordinate.h#L309

This is a case where a ReferencePtr is set to `new Something()`, but the pointer is passed to a class that will take ownership of the pointer. In this case, it's still nice to have the ReferencePtr so that the pointer is set to `nullptr` upon copy. What should be done here?

sherm1:
> I see you didn't put Coordinate::_lockFunction in your list

Thanks, that's an interesting case I didn't notice. It is still a misuse (temporarily) because the ReferencePtr is used to hold ownership until it can be transferred. After that it really is just a reference to someone else's object so is OK. This is likely harmless here but would cause a leak if an exception occurred before the ownership could be transferred, since a ReferencePtr will not destruct its contents.

A nicer way to do this I think would be to use an `std::unique_ptr` to hold the freshly-allocated `Function` object until its ownership can be transferred to the `PrescribedMotion` constraint. It would look something like this:
```c++
std::unique_ptr<Function> func(new Function()); // func has ownership
ReferencePtr<Function> funcref(func.get()); // keep a reference to Function object
Constraint::PrescribedMotion(func.release()); // take over ownership from func
```

[...]

PR #553, opened 2015-07-25
Fix misuse of ReferencePtr in ControllerSet::_controlStore

chrisdembia:
The member variable *was* supposed to own the memory of the `_controlStore`, and so I use a `unique_ptr` instead of a `ReferencePtr`.

Addresses part of #528.

### DIFF
--- a/OpenSim/Simulation/Model/ControllerSet.h
+++ b/OpenSim/Simulation/Model/ControllerSet.h
@@ -99,7 +99,7 @@ private:
     std::unique_ptr<Storage> _controlStore;
 
     // Set of actuators controlled by the set of controllers.
-    SimTK::ReferencePtr<Set<Actuator> > _actuatorSet;
+    SimTK::ReferencePtr<Set<Actuator>> _actuatorSet;
 //=============================================================================
 };  // END of class ControllerSet
 //=============================================================================
--- a/OpenSim/Simulation/Model/ControllerSet.h
+++ b/OpenSim/Simulation/Model/ControllerSet.h
@@ -51,41 +51,28 @@ OpenSim_DECLARE_CONCRETE_OBJECT(ControllerSet, ModelComponentSet<Controller>);
 //=============================================================================
 // METHODS
 //=============================================================================
+public:
     //--------------------------------------------------------------------------
     // CONSTRUCTION
     //--------------------------------------------------------------------------
-public:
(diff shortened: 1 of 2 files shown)


OpenSim/Simulation/Model/ControllerSet.h
```diff
@@ -102,6 +91,13 @@ OpenSim_DECLARE_CONCRETE_OBJECT(ControllerSet, ModelComponentSet<Controller>);
     virtual void computeControls(const SimTK::State& s, SimTK::Vector &controls) const; 
 
     virtual void printInfo() const;
+
+private:
+
+    std::unique_ptr<Storage> _controlStore;
+
+    // Set of actuators controlled by the set of controllers.
+    SimTK::ReferencePtr<Set<Actuator> > _actuatorSet;
```
sherm1: In C++11 the extra space isn't needed any more so you can write `SimTK::ReferencePtr<Set<Actuator>>`.

sherm1:
Changes look good. I wonder if a newer version of SWIG could handle those `= default`s? It will probably get to be a pain `#ifdef`ing them all.

sherm1:
I didn't see that you added `#include <memory>` to provide access to `std::unique_ptr`. Was that already included somewhere explicitly, or did it just come in with some other header by accident? If the latter it would be better to including it explicitly (somewhere) to make the dependence on that header clear.

chrisdembia:
> Changes look good. I wonder if a newer version of SWIG could handle those = defaults? It will probably get to be a pain #ifdefing them all.

Yes, 3.0.2 handles the `= default` (we use 3.0.2 on appveyor) but 2.0.10 does not (on travis). I'll create an issue for removing those `#ifndef`'s when we change SWIG versions officially.

chrisdembia:
@sherm1 this is ready for review again. I would have preferred to remove the space between the `> >`, but swig 2.0.10 does not recognize it. Upgrading to a newer version of swig is a different issue. 

I included `<memory>` explicitly.

chrisdembia:
@sherm1 I could also try to use the default constructor, assignment operator, etc. and use ClonePtr instead.

sherm1:
> I could also try to use the default constructor, assignment operator, etc. and use ClonePtr instead.

That would be nice but I think is a separate project. I think this fixes the ref ptr problem adequately -- merging.

Issue #206, opened 2014-10-09
Add test case for model conversion


aymanhab:
To make sure old models deserialize and display correctly, need a textual rep of Display and a suite of old models using old format.

jenhicks:
Additional comment from @aseth1 
Most components are verified by virtue of older model files being used by other test cases. But in many test cases (esp. Joints, Constraints and Forces) models are built programmatically (without loading a model from file) and the serialization loop is tested only in the latest version.

In order to ensure that the content of older models is preserved they must be verified to contain the same content in the latest version of the model.

aymanhab:
I tried to open the simple model bouncing_block from release 3.3 and it failed to open with 
PointToPointSpring::connect() failed to connect ....

aseth1:
@aymanhab In which build did you try to open the model? Is that model in the repository. Is there a test case for it? What was the complete error message? That will help track it down. Thanks!

aymanhab:
@aseth1 This is occuring in my build of the GUI based off master before your last commit, so it may have been fixed. The model is from earlier distribution (3.3) and the error message is attached.
![captureerrormsg](https://cloud.githubusercontent.com/assets/3972208/12830284/4b08bf74-cb43-11e5-94b0-94bb0ea8a56d.JPG)

aseth1:
Great. It's helpful to know t hat the component could not be found. It may be addressed by recent changes but it could also be due to a bug in the update from XML. Can you point me to the model file.
Thanks.

aymanhab:
Sure,  OpenSim3.3\Models\BouncingBlock\bouncing_block.osim Since we split the repositories, the model file doesn't live in the opensim-core repo.

aseth1:
PointToPointSpring does not override updateFromXMLNode(). We should not spot fix this. Instead we should write a test to find all the Components that should be updated and we know it should find this one.

tkuchida:
From #1407:
> Related to #206 (marked "Next Release"). We should have a folder of model files in all formats to rigorously exercise the `updateFromXMLNode()` methods in a dedicated test case…

chrisdembia:
> PointToPointSpring does not override updateFromXMLNode(). We should not spot fix this. Instead we should write a test to find all the Components that should be updated and we know it should find this one.

Do you have an idea for how to write such a test? We at least need to add `updateFromXMLNode()` for PointToPointSpring before the release.

chrisdembia:
[...]

Issue #161, opened 2014-08-09
Name of Connector::connected_to_name property


chrisdembia:
I think the language surrounding connections still needs some cleaning up. I think this should just be "connectee". I still vote for a word other than connectee, but whatever that noun is should be the name of this property, I think. I feel that `connected_to_name` reads verbosely in the XML.

During the hackathon, we brainstormed some alternative names, like plug, socket (using a circuit analogy) I know some of them weren't good, but I think we should still discuss this.

I'm used to the "ee" syntax with Employer, Employee, in which both objects are of the same type. Connector and the connectee are not of the same type; Connector is a middle man.

chrisdembia:
In #558 @sherm1 said

> I see I didn't like connector/connectee in #161 but I think I got used to it from Ajay using the term in discussions so I was surprised that it wan't used across the API. If it is used anywhere I'd like to see it used everywhere. "connectee" does have the advantages of (a) being a made-up word with no a priori meaning, (b) clearly being a noun, and (c) conciseness.



I am fine with `connectee` so long as we create a beautiful and clear diagram explaining all the new terms (connector, connectee, input, output), with a "plug" showing what a `connectee` is, etc., and we put this diagram where it will be found easily by component developers and users. I'm thinking in the description of the Component class, or preferably in our "Component developer guide", and perhaps in the user's guide as well.

aseth1:
I've been reluctant to document when the names are in flux - the names are in flux because the terms are not liked - part of not liking them is that there is no clear documentation ... I will add to the [definitions.md](https://gist.github.com/aseth1/0d3e16028d985d6ed95d)  to start to address the latter problem at least. 

Consistency with the code would also be helpful. Changing `connected_to_name` to `connectee_name`, for example, would help.  @sherm1 you suggested that and sounds like @chrisdembia agrees. @jenhicks @aymanhab  OK with that?

chrisdembia:
I'm closing this now since #572 is about to be merged.

PR #558, opened 2015-07-27
Make use of Frames with BushingForce as example

aseth1:
The main objective of this PR is to address how Frames should be used to attach Components like Joints, Constraints, and most Forces that need Frames to join/attach to. Frames remain grossly underutilized.

I targeted BushingForce to show how to use PhysicalOffsetFrames as subcomponents for the purpose of abstracting these frame connections. The BushingForce gets its dependencies (and connectors) by employing the PhysicalOffsetFrame. The offset is more than likely to be of interest to the BushingForce (only) and not necessarily the Body (or other PhysicalFrame) origin. It also does not preclude using the Body frame with an identity offset transform in that (no offset) case.

I hope to use this approach for Joints, Constraints and any Force that currently require a PhysicalFrame  (e.g. Body) and have transform properties (location and orientation) to locate the attachment.

The only "significant" change to the API was the introduction of `Component::extendAddToSystemAfterSubcomponents()` which is a necessity for any Component that needs its Components to be added first, like the BushingForce that needs the resources (system elements) of its subcomponents (e.g. the MobilizedBodies that the underlying `SimTK::LinearBushing` requires).

Extended testComponents (that now can walk through subcomponents for their connectors) and testForces (to include deserialization (updateFromXML) comparison from previous version of the BushingForce).

There is a need to have an explicit conversion/deserialization test from previous file versions (to  speak more broadly than #206) so we don't have to add specific coverage for every component that gets updated.

There are other minor changes like removing the use of Body (e.g. in reference to Ground) where we now mean a Frame in the code (including comments). Also updated `CoupledBushingForce` plugin example to use new property system.


OpenSim/Common/Component.h
```diff
@@ -1552,6 +1566,21 @@ template <class T> friend class ComponentMeasure;
...
+    void componentsConnect(Component& root) const;
+
+    /// Invoke addToSystem() on the (sub)components of this Component.*/
+    void componentsAddToSystem(SimTK::MultibodySystem& system) const;
+
+    /// Invoke initStateFromProperties() on (sub)components of this Component */
+    void componentsInitStateFromProperties(SimTK::State& state) const;
+
+    /// Invoke setPropertiesFromState() on (sub)components of this Component */
+    void componentsSetPropertiesFromState(const SimTK::State& state);
```
chrisdembia: All 5 of these methods mix doxygen comment styles. I presume `*/` will show up in the doxygen output.

OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.h
```diff
@@ -41,8 +40,8 @@ namespace OpenSim {
  * where components of the resulting BodyForce are coupled to all any or all 
  * deviations, such that the general stiffness and damping matrices are 6x6.
  *
- * Deviations of a frame1 on body1 and another (frame2) on body2 are represents in
- * termsof x-y-z Euler angle sequence and x,y,z position of frame2 in frame1.
+ * Deviations of a frame1 on frame1 and another (frame2) on body2 are represents
```
chrisdembia: represented.

This comment is confusing to me; it now seems inconsistent (frame1 on frame1, frame2 on body2).

aymanhab:
@aseth1 I assumed based on previous discussion that we'll abandon the use of the term subComponent and use hidden or internal instead. Are these names temporary? Thanks.

sherm1:
> The only "significant" change to the API was the introduction of Component::extendAddToSystemAfterSubcomponents() 

This addition is disappointing to me. It is an additional complexity that every Component writer now has to deal with, even if they don't use subcomponents (because they now have to decide which of two possible methods they should be using). Isn't there some convention we could impose instead that would make this unnecessary? I feel like the need for this method is trying to tell us that something is wrong but I can't put my finger on what it is yet.

OpenSim/Simulation/Model/BushingForce.h
```diff
@@ -80,15 +75,28 @@ OpenSim_DECLARE_CONCRETE_OBJECT(BushingForce, Force);
...
     to be at their body origins, and sets all bushing parameters to zero. **/
     BushingForce();
-    /** This constructor provides everything needed to create a real 
-    bushing. See property declarations for more information. **/
-    BushingForce(const std::string& body1Name, 
+
+    /** Construct the BushingForce given the offset frames that it tries to
+    align according to the physical properties of the bushing.
+    See property declarations for more information. **/
+    BushingForce(const PhysicalOffsetFrame& frame1,
```
sherm1: I don't understand why all these constructors are needed and why the BushingForce doesn't use `PhysicalFrame`. Isn't a `PhysicalOffsetFrame` a kind of `PhysicalFrame`? I would think the only need constructor would be
```c++
BushingForce(
    const PhysicalFrame& frame1,
    const PhysicalFrame& frame2,
    const SimTK::Vec3& transStiffness,
    const SimTK::Vec3& rotStiffness,
    const SimTK::Vec3& transDamping,
    const SimTK::Vec3& rotDamping);
```
Why isn't that good enough?

OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.cpp
```diff
@@ -136,192 +96,62 @@ void CoupledBushingForce::copyData(const CoupledBushingForce &aForce)
...
- *
- * @return Reference to this object.
- */
-CoupledBushingForce& CoupledBushingForce::operator=(const CoupledBushingForce &aForce)
-{
-    Force::operator=(aForce);
-    copyData(aForce);
-    return(*this);
+    //mark the two PhysicalOffsetFrames as subcomponents 
+    addComponent(&upd_offset_frame1());
```
sherm1: I'm confused about why these `addComponent()` calls are in finalizeFromProperties. Are these what we could eliminate if object properties were automatically made into subcomponents?

OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.cpp
```diff
@@ -477,13 +311,13 @@ OpenSim::Array<double> CoupledBushingForce::getRecordValues(const SimTK::State&
 
     //get the net force added to the system contributed by the bushing
     simtkSpring.calcForceContribution(state, bodyForces, particleForces, mobilityForces);
-    SimTK::Vec3 forces = bodyForces(_model->getBodySet().get(_body1Name).getMobilizedBodyIndex())[1];
-    SimTK::Vec3 torques = bodyForces(_model->getBodySet().get(_body1Name).getMobilizedBodyIndex())[0];
+    SimTK::Vec3 forces = bodyForces(get_offset_frame1().getMobilizedBodyIndex())[1];
```
sherm1: Should this be square-bracket indexing `bodyForces[]`?

OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.h
```diff
@@ -38,81 +37,64 @@ namespace OpenSim {
...
- * where components of the resulting BodyForce are coupled to all any or all 
+ * where components of the resulting BodyForce are coupled to any or all 
  * deviations, such that the general stiffness and damping matrices are 6x6.
  *
- * Deviations of a frame1 on body1 and another (frame2) on body2 are represents in
- * termsof x-y-z Euler angle sequence and x,y,z position of frame2 in frame1.
- * Damping is applied to the relative angular velocity and linear velocity of
- * frame2 in frame1.
+ * Deviations of an offset_frame (1) (on a PhysicalFrame e.g. a Body) and another 
+ * offset_frame2 (on another PhysicalFrame) are expressed in terms of and x-y-z
```
sherm1: and -> an

OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.h
```diff
@@ -38,81 +37,64 @@ namespace OpenSim {
...
-    SimTK::Vec3& _orientationInBody2;
+public:
+    //==============================================================================
+    // PROPERTIES
+    //==============================================================================
+    /** @name Property declarations
+    These are the serializable properties associated with the CoupleBushingForce. **/
+    /**@{**/
+    /// The frames that are connected by the bushing force
+    OpenSim_DECLARE_PROPERTY(offset_frame1, PhysicalOffsetFrame,
```
sherm1: Why not `PhysicalFrame`?

OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.h
```diff
@@ -38,81 +37,64 @@ namespace OpenSim {
...
+    OpenSim_DECLARE_PROPERTY(damping_row2, SimTK::Vec6,
+        "2nd row of damping matrix 3 rotation (Nm/(rad/s)) & 3 translation (N/(m/s)).");
+    OpenSim_DECLARE_PROPERTY(damping_row3, SimTK::Vec6,
+        "3rd row of damping matrix 3 rotation (Nm/(rad/s)) & 3 translation (N/(m/s)).");
+    OpenSim_DECLARE_PROPERTY(damping_row4, SimTK::Vec6,
+        "4th row of damping matrix 3 rotation (Nm/(rad/s)) & 3 translation (N/(m/s)).");
+    OpenSim_DECLARE_PROPERTY(damping_row5, SimTK::Vec6,
+        "5th row of damping matrix 3 rotation (Nm/(rad/s)) & 3 translation (N/(m/s)).");
+    OpenSim_DECLARE_PROPERTY(damping_row6, SimTK::Vec6,
+        "6th row of damping matrix 3 rotation (Nm/(rad/s)) & 3 translation (N/(m/s)).");
```
sherm1: So nice to see those old properties disappear!

OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.h
```diff
@@ -120,73 +102,77 @@ OpenSim_DECLARE_CONCRETE_OBJECT(CoupledBushingForce, Force);
...
+    See property declarations for more information. **/
+    CoupledBushingForce(const std::string& frame1Name,
+                            SimTK::Vec3 point1, SimTK::Vec3 orientation1,
+                        const std::string& body2Name,
+                            SimTK::Vec3 point2, SimTK::Vec3 orientation2,
+                         SimTK::Mat66 stiffnessMat, SimTK::Mat66 dampingMat);
 
-    CoupledBushingForce& operator=(const CoupledBushingForce &aForce);
-    void copyData(const CoupledBushingForce &aForce);
+    virtual ~CoupledBushingForce();
```
sherm1: Better to write `~CoupledBushingForce() override;` which will fail to compile if the base class doesn't have a virtual destructor (which it should).

OpenSim/Simulation/Model/BushingForce.cpp
```diff
@@ -47,27 +47,95 @@ BushingForce::BushingForce()
...
+
+    upd_offset_frame2().updConnector<PhysicalFrame>("parent")
+        .set_connected_to_name(frame2.getName());
+    Rotation rotation2(BodyRotationSequence,
+        orientation2[0], XAxis,
+        orientation2[1], YAxis,
+        orientation2[2], ZAxis);
+    upd_offset_frame2().setOffsetTransform(Transform(rotation2, point2));
+
+    set_rotational_stiffness(rotStiffness);
```
sherm1: Lots of code duplication here between constructors. Better just to have one constructor I think, but if there has to be more than one the duplicate code should be factored out into an init() method or even better (since C++11) you can invoke one constructor from the other ones.

OpenSim/Simulation/Model/BushingForce.cpp
```diff
@@ -85,64 +153,96 @@ void BushingForce::setNull()
...
+            SimTK::Xml::element_iterator orientBody1Elt = aNode.element_begin("orientation_body_1");
+            SimTK::Xml::element_iterator locBody2Elt = aNode.element_begin("location_body_2");
+            SimTK::Xml::element_iterator orientBody2Elt = aNode.element_begin("orientation_body_2");
+
+            // If default constructed then elements not serialized since they are default
+            // values. Check that we have associated elements, then extract their values.
+            if (body1Element != aNode.element_end()){
+                std::string frame1_name("");
+                body1Element->getValueAs<std::string>(frame1_name);
+                upd_offset_frame1().updConnector(0).set_connected_to_name(frame1_name);
```
sherm1: Random comment: would it be more consistent to call this `set_connectee_name()`?

chrisdembia: See #161 for discussion of this.

OpenSim/Tests/Components/testComponents.cpp
```diff
@@ -167,31 +171,45 @@ void testComponent(const Component& instanceToTest)
...
-                dependency = Object::newInstanceOfType("Body");
+    Component* sub = instance;
+    ComponentList<Component> comps = instance->getComponentList<Component>();
+    ComponentList<Component>::const_iterator it = comps.begin();
+
+    while(sub) {
+        int nc = sub->getNumConnectors();
+        for (int i = 0; i < nc; ++i){
+            AbstractConnector& connector = sub->updConnector(i);
+            string dependencyTypeName = connector.getConnectedToTypeName();
```
sherm1: `getConnecteeTypeName()` ?

OpenSim/Tests/Components/testComponents.cpp
```diff
@@ -352,4 +371,19 @@ void addObjectAsComponentToModel(Object* instance, Model& model)
         throw Exception(className + " is not a ModelComponent.",
             __FILE__, __LINE__);
     }
+
+    try {
+        // Current Component iterator requirement to build the component list (tree)
+        // forces us to connect the model so we can traverse subcomponents
+        model.connect(model);
```
sherm1: model connecting to itself??

OpenSim/Simulation/Model/BushingForce.cpp
sherm1: I see I didn't like connector/connectee in #161 but I think I got used to it from Ajay using the term in discussions so I was surprised that it wan't used across the API. If it is used anywhere I'd like to see it used everywhere. "connectee" does have the advantages of (a) being a made-up word with no a priori meaning, (b) clearly being a noun, and (c) conciseness.

aseth1:
> >  The only "significant" change to the API was the introduction of Component::extendAddToSystemAfterSubcomponents()

> This addition is disappointing to me. It is an additional complexity that every Component writer now has to deal with, even if they don't use subcomponents (because they now have to decide which of two possible methods they should be using). Isn't there some convention we could impose instead that would make this unnecessary? I feel like the need for this method is trying to tell us that something is wrong but I can't put my finger on what it is yet.

It was dsiappointing to me as well, but I could not see a way around it that was less convoluted. It was just fortunate that we didn't need it before. There are two fundamental ways that subcomponents can be added to the system. In one case the subcomponents parcel up responsibilities of the parent like Coordinates do for the Joint. The joint acquires the necessary system resources (when it adds itself to the system) and those are shared/provided to its coordinates when they add themselves so that individual q's and u's are accessible.  That was one major use case of subcomponents.

In BushingForce, the PhysicalOffsetFrames are its subcomponents (they do not belong to the PhyscialFrame they are offsetting) and I do not want to rely on the Model having a flat list of all the possible frames that will be used downstream. Without the attachment frames already a part of the system, we cannot add the underlying Force element in Simbody. We need the mobilized body indices of the PhysicalOffsetFrames to specify the underlying LinearBushing, which requires they be added to the System before the BushingForce can add itself.

It indicates that there are two legitimate ways of adding system resources of the parent component - before or after your subcomponents. In general, I could envision Components that did both: e.g. added mobilized bodies (before), then add subcomponents such as as frames/points for attaching forces, and then add force element(s) attached to the frames/points (after).

I am happy to consider alternatives. Perhaps worth discussing tomorrow.

sherm1:
>  There are two legitimate ways of adding system resources of the parent component - before or after your subcomponents. In general, I could envision Components that did both.

Maybe that suggests that parent components should take care of their own child (sub)components since in general only the parent can know the proper ordering?

OpenSim/Simulation/Model/BushingForce.h
aseth1: > I targeted BushingForce to show how to use PhysicalOffsetFrames as subcomponents for the purpose of abstracting these frame connections. The BushingForce gets its dependencies (and connectors) by employing the PhysicalOffsetFrame. The offset is more than likely to be of interest to the BushingForce (only) and not necessarily the Body (or other PhysicalFrame) origin. It also does not preclude using the Body frame with an identity offset transform in that (no offset) case.

The alternative is to assume the offset exists somewhere (e.g. in the Model), but this does not map well to the use cases/examples. When you create a bushing force you generally also specify the offset (location, orientation in frame 1 and 2) then it makes more sense to be an attribute of the BusingForce (which is what I have done).

You are right, the only convenience constructor that is necessary is the one that takes the offset frames, since you can easily construct the PhysicalOffsetFrame first.

I will remove the others.

aseth1:
> Maybe that suggests that parent components should take care of their own child (sub)components since in general only the parent can know the proper ordering?

In many cases it does not matter and the assumption to do the parent before the children works. Adding this one method gives the parent the option to do additional stuff (or all of it) afterwards if necessary. I can tell you don't like it. Should it be more flexible?

I haven't come across a case where the parent has to control the order of the children adding to the system, so I think it is still a valuable service to automatically delegate the calls to the children rather than relying on each parent component to do it manually. I am not sure what you are advocating.

chrisdembia:
I discussed the addition of the `Component::extendAddToSystemAfterSubcomponents()` complexity at length with Ajay when he was deciding what to do, and I support this decision. 

To reduce the perceived complexity, there is a "default": `Component::extendAddToSystem()` is what should be normally used.

> Maybe that suggests that parent components should take care of their own child (sub)components since in general only the parent can know the proper ordering?

This introduces more opportunity for confusing errors of the type "d'oh, I forgot to invoke `subcomponentsAddToSystem()`, and people (including me) are really good at making this kind of mistake. 

Ajay also considered a number of alternatives; this was not a quick choice. For example, he considered using a flag to ensure that `subcomponentsAddToSystem()` was indeed called by the component developer so that an error message could be generated. This flag presented some other challenges.

OpenSim/Simulation/Model/BushingForce.h
chrisdembia: It is starting to sound valuable to not make a component like BushingForce choose between "ownership" OR "reference". Ideally, if the user wanted to refer to a body for the PhysicalFrame, they could; but if they wanted to directly insert a new PhysicalOffsetFrame that would be managed by BushingForce, they could do that too.

This specific issue came up with the pseudocde session. I'd like to do either of the following:

```
<components>
    <data_to_track>
        <DataSource> <filename>mydata.csv</filename></DataFile>
    </data_to_track>
</components>
```
(DataSource owned by data_to_track)

-or-

```
<components>
    <data_to_track ref="mydata"/>
    <DataSource name="mydata"> <filename>mydata.csv</filename></DataFile>
<components>
```
(DataSource is simply referenced)

This resembles part of what @aymanhab once suggested for the xml related to connectors.

I think forcing to choose either ownership or reference may put us in a lot of undesirable design corners. Perhaps we could have the notion of a "ReferenceProperty".

I don't think this is a topic to consider within this PR though.

sherm1: > The alternative is to assume the offset exists somewhere (e.g. in the Model), but this does not map well to the use cases/examples. 

I would have expected to be able to put a bushing (or joint or constraint) between any two PhysicalFrames. I don't see why it has to be limited to a PhysicalOffsetFrame. (Making one with a zero offset seems like a hack!) In some cases the PhysicalFrame would be a specially-built frame just for use with the bushing (say), but in others it would be a pre-existing frame. For example, I may have reason to attach one end of a bushing to the same frame as I'm using for one end of a joint.

I'm wondering if this problem has to do with *ownership* of the PhysicalFrame -- in some cases we would like to use an existing frame, in others we might want to craft special frames that "go with" the bushing. I think a bushing should connect to two PhysicalFrames. As an orthogonal issue, some component will be the owner of those frames; that component might be the bushing or might be someone else.

sherm1:
>I think forcing to choose either ownership or reference may put us in a lot of undesirable design corners. I don't think this is a topic to consider within this PR though.

I think it is partly relevant here, though. Ajay started out this PR thus:
> The main objective of this PR is to address how Frames should be used to attach Components like Joints, Constraints, and most Forces that need Frames to join/attach to. 

I think the Bushing example shows that to use Frames cleanly as they were envisioned requires decoupling of ownership from reference. I really like the simple concept that Bushings, Joints, Constraints and so on connect PhysicalFrames. Then we have a separate scheme for creating PhysicalFrames with whatever methods are useful, and if we add new ways to do it later (DenavitHartenbergFrame or whatever) everything still works fine. I'm imagining the XML would look something like this (please forgive imaginary syntax):
```xml
<!-- Bushing referencing someone else's frames -->
<BushingForce name="myBFF">
    <frame1>name_of_some_physical_frame</frame1>
    <frame2>name_of_another_one</frame2>
    < ... stiffness & damping stuff ... >
</BushingForce>

<!-- Bushing that has its own frames -->
<BushingForce name="myBFF">
   <components>
        <PhysicalOffsetFrame name=first> <body>l_femur</body><offset>1 2 3</offset>
        </PhysicalOffsetFrame>
         <PhysicalOffsetFrame name=second> <body>r_femur</body><offset>1 2 3</offset>
        </PhysicalOffsetFrame>
   </components>
    <frame1>myBFF.first</frame1>
    <frame2>myBFF.second</frame2>
    < ... stiffness & damping stuff ... >
</BushingForce>
```

The ownership is conveyed by the XML tree structure but is orthogonal to how references are made, which would be the same in the two cases.

aymanhab:
@sherm1 Totally agree on the decoupling. BushingForces are not in any way different from Joint. If we now support only references (not ownership) of PhysicalFrames, we should stick with it until we come up with an alternative ownership scheme (e.g. using Properties) rather than invent the scheme as a part of this PR.

OpenSim/Simulation/Model/BushingForce.h
aseth1: The BushingForce connects two PhysicalFrames. The purpose of a PhysicalOffsetFrame is to offset a PhysicalFrame to which it connects. I am using the PhysicalOffsetFrame to make my task of connecting anywhere on any PhysicalFrame easier (in my opinion). If you ask a BushingForce what connectors it has they are PhysicalFrames, not the PhysicalOffsetFrames but their connectors. I was able to get a connector<PhysicalFrame> and generally useful offset properties for "free" by making the PhysicalOffsetFrames an attribute  (property) of the BushingForce. I thought it was a nice feature to use a Component as a property in this way.

OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.cpp
aseth1: Yes, this is what we could get rid of.

sherm1:
> In many cases it does not matter and the assumption to do the parent before the children works. ... I can tell you don't like it. ... I am not sure what you are advocating.

I don't have a plan, just whining about the extra complexity. Can we talk about it a little on Thu? I'd like to brainstorm to see if there is a way to design the problem out of existence. For example, maybe the problem could be fixed by minor changes to the classes that are used as subcomponents?

aseth1:
Thanks all for the feedback and discussion today. I will open a new PR with the structure we discussed (no implied OffsetFrame). @aymanhab I may need your help to properly update from XML. 

Until we have a general solution for a Components list (property) on Component (which requires a systematic way to process by resource requirement), I will use a more restrictive list of PhysicalFrames.

PR #560, opened 2015-07-29
ReferencePtr misuse GeometryPath

chrisdembia:
Related to #528 .

### DIFF
--- a/OpenSim/Simulation/Model/GeometryPath.cpp
+++ b/OpenSim/Simulation/Model/GeometryPath.cpp
@@ -57,23 +57,11 @@ static const Vec3 DefaultDefaultColor(.5,.5,.5); // boring gray
  */
 GeometryPath::GeometryPath() :
     ModelComponent(),
-    _preScaleLength(0.0),
-    _owner(nullptr)
-{
-    setNull();
-    constructProperties();
- }
-
-
-//_____________________________________________________________________________
-/*
- * Set the data members of this GeometryPath to their null values.
- */
-void GeometryPath::setNull()
+    _preScaleLength(0.0)
 {
     setAuthors("Peter Loan");
-    _maSolver.reset();
-}
+    constructProperties();
+ }
 
 //_____________________________________________________________________________
 /*
--- a/OpenSim/Simulation/Model/GeometryPath.cpp
+++ b/OpenSim/Simulation/Model/GeometryPath.cpp
@@ -64,15 +64,6 @@ GeometryPath::GeometryPath() :
     constructProperties();
  }
 
-//_____________________________________________________________________________
-/*
- * Destructor.
- */
-GeometryPath::~GeometryPath()
-{
-    delete _maSolver;
-}
-
 
 //_____________________________________________________________________________
 /*
@@ -81,7 +72,7 @@ GeometryPath::~GeometryPath()
 void GeometryPath::setNull()
 {
     setAuthors("Peter Loan");
-    _maSolver = NULL;
+    _maSolver.reset();
 }
 
 //_____________________________________________________________________________
@@ -232,7 +223,7 @@ void GeometryPath::namePathPoints(int aStartingIndex)
     {
         sprintf(indx,"%d",i+1);
         PathPoint& point = get_PathPointSet().get(i);
-        if(point.getName()=="" && _owner) {
+        if (point.getName()=="" && _owner) {
             point.setName(_owner->getName() + "-P" + indx);
         }
     }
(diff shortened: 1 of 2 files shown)


OpenSim/Simulation/Model/GeometryPath.cpp
```diff
@@ -1278,10 +1269,10 @@ calcLengthAfterPathComputation(const SimTK::State& s,
 double GeometryPath::
 computeMomentArm(const SimTK::State& s, const Coordinate& aCoord) const
 {
-    if(!_maSolver)
-        _maSolver = new MomentArmSolver(*_model);
+    if (!_maSolver)
+        const_cast<Self*>(this)->_maSolver.reset(new MomentArmSolver(*_model));
```
sherm1: That `const_cast` seems unfortunate! Isn't there a way to make sure that the MomentArmSolver gets created earlier, in one of the methods that has write access to the GeometryPath? It looks like it just needs to get done after `_model` first gets set.

OpenSim/Simulation/Model/GeometryPath.h
```diff
@@ -71,11 +71,13 @@ OpenSim_DECLARE_CONCRETE_OBJECT(GeometryPath, ModelComponent);
...
-    // object that owns this GeometryPath object
+    // Pointer to the Object that owns this GeometryPath object.
     SimTK::ReferencePtr<Object> _owner;
 
-    // solver used to compute moment-arms
-    mutable SimTK::ReferencePtr<MomentArmSolver> _maSolver;
+    // Solver used to compute moment-arms. The GeometryPath owns this object,
+    // but we cannot simply use a unique_ptr because we want the pointer to be
+    // cleared on copy.
+    SimTK::NullOnCopyUniquePtr<MomentArmSolver> _maSolver;
```
sherm1: Cool!

OpenSim/Simulation/Model/GeometryPath.h
```diff
@@ -85,7 +87,9 @@ OpenSim_DECLARE_CONCRETE_OBJECT(GeometryPath, ModelComponent);
     //--------------------------------------------------------------------------
 public:
     GeometryPath();
-    virtual ~GeometryPath();
+#ifndef SWIG
+    ~GeometryPath() override = default;
```
sherm1: Do you think it will be a problem if there is no destructor for SWIG to wrap? Or does that happen automatically somehow? You could avoid this by writing `~GeometryPath() override {}`.

OpenSim/Simulation/Model/GeometryPath.cpp
```diff
@@ -81,7 +72,7 @@ GeometryPath::~GeometryPath()
 void GeometryPath::setNull()
 {
     setAuthors("Peter Loan");
-    _maSolver = NULL;
+    _maSolver.reset();
```
sherm1: Shouldn't `_preScaleLength` and `_owner` get initialized here also?

OpenSim/Simulation/Model/GeometryPath.cpp
chrisdembia: I agree about modifying member variables in const methods; this is probably one place that makes the models non-threadsafe.

1. I tried moving the allocation of this object to `extendConnectToModel()`, but then more than half of the test cases failed. The `MomentArmSolver` was trying to do something with the `Model`, but the `Model` had not yet had `initSystem()` called on it. I decided to stay focused on the task at hand.
2. I prefer using `const_cast` instead of making a member variable mutable. Mutable member variables can be changed by any const method, whereas `const_cast` localizes the mutability of the member.

OpenSim/Simulation/Model/GeometryPath.h
chrisdembia: : ) thank you for writing the class.

OpenSim/Simulation/Model/GeometryPath.cpp
chrisdembia: Alternatively, I will get rid of `setNull()` and do everything in the constructor.

OpenSim/Simulation/Model/GeometryPath.h
chrisdembia: The [SWIG manual](http://binf.gmu.edu/software/SWIG/SWIGPlus.html#SWIGPlus_nn7) says:

> If a C++ class does not define any public constructors or destructors, SWIG will automatically create a default constructor or destructor.

So I think we're fine here.

chrisdembia:
Ready for review again.

sherm1:
Looks great -- thanks for fixing this Chris (and for taking a chance on the new smart pointer!).

chrisdembia:
Thank you for merging.

chrisdembia:
(and for writing the NullOnCopyUniquePtr)

PR #561, opened 2015-07-30
Fix CoordinateLimitForce::_upStep, _loStep.

chrisdembia:
Related to #528

### DIFF
--- a/OpenSim/Simulation/Model/CoordinateLimitForce.h
+++ b/OpenSim/Simulation/Model/CoordinateLimitForce.h
@@ -120,7 +120,9 @@ public:
     //use compiler default copy constructor and assignment operator
 
     /** Destructor */
-    ~CoordinateLimitForce();
+#ifndef SWIG
+    ~CoordinateLimitForce() override = default;
+#endif
 
 
     //--------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


OpenSim/Simulation/Model/CoordinateLimitForce.h
```diff
@@ -235,9 +237,10 @@ OpenSim_DECLARE_CONCRETE_OBJECT(CoordinateLimitForce, Force);
 
     // Smooth step functions for continuous transition from no stiffness and 
     // damping to constant values beyond the limits. These are heap allocated
-    // and owned here so must be deleted in destructor.
-    SimTK::ReferencePtr<SimTK::Function::Step> _upStep;
-    SimTK::ReferencePtr<SimTK::Function::Step> _loStep;
+    // and owned here, but we don't want these functions to be copied if a
+    // CoordinateLimitForce is copied.
```
sherm1: I can't see why these functions shouldn't be copied. It appears to me that they depend only on the values of properties that are copied, and they contain no pointers that might be left dangling. Why not just:
```c++
SimTK::Function::Step> _upStep, _loStep;
```
?

sherm1: Oh, is it because the weighting comes from the model?

chrisdembia: > Oh, is it because the weighting comes from the model?

Yes; it looks like the function parameters depend on whether the coordinate is rotational or translational.

chrisdembia: I feel bad now that you modified the Function classes, but there is no need to construct the Function objects during construction of the CoordinateLimitForce.

I'm pretty indifferent though. I'm happy to go back to what I had previously: I can *not* use pointers for the two objects, and initialize them to something silly in the CoordinateLimitForce constructor.

sherm1: > I feel bad now that you modified the Function classes

No worries! I like that the Function classes got improved regardless of whether they are used now.

chrisdembia:
Thank you.

PR #562, opened 2015-07-30
Geometry appearance

aymanhab:
Rearrange Appearance classes and move Properties to correct Appearance subclasses. Also introduces a couple intermediate Geometry subclasses to treat Geometry subclasses uniformly


OpenSim/Simulation/RegisterTypes_osimSimulation.cpp
```diff
@@ -273,7 +272,8 @@ OSIMSIMULATION_API void RegisterTypes_osimSimulation()
     Object::renameType("MusclePoint",       "PathPoint");
     Object::renameType("MuscleViaPoint",    "ConditionalPathPoint");
     Object::renameType("MovingMusclePoint", "MovingPathPoint");
-    Object::renameType("MusclePointSet",    "PathPointSet");
+    Object::renameType("MusclePointSet", "PathPointSet");
+    Object::renameType("Appearance", "SurfaceAppearance");
```
chrisdembia: Is this necessary? We didn't release a version of OpenSim with Appearance yet.

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -37,22 +37,21 @@ class Model;
 //=============================================================================
 //=============================================================================
 /**
- * A class that holds the Display Attributes (Appearance) of an object displayed 
- * in the OpenSim Visualizer.
+ * A class that holds the Appearance Attributes of an object displayed
```
chrisdembia: Is Attributes a class name? If not I think it should not be capitalized.

jenhicks:
@aseth1 Can you review?

OpenSim/Simulation/RegisterTypes_osimSimulation.cpp
aymanhab: True @chrisdembia but the code to convert old Geometry in Body.cpp uses "Appearance" rather then SurfaceAppearance, it seems cleaner to treat these as two separate XML changes but I don't feel strongly about it. If we remove the rename then we have to go back and fix the conversion code accordingly.

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -37,22 +37,21 @@ class Model;
...
- * utilize LineAppearance for Arrow, Line, and Frame (unused for now)
+ * TODO: Add Resolution or Quality to Appearance from DecorativeGeometry 
  *
  * @author Ayman Habib
  * @version 1.0
  */
-class OSIMSIMULATION_API BaseAppearance : public Object {
-    OpenSim_DECLARE_CONCRETE_OBJECT(BaseAppearance, Object);
+class OSIMSIMULATION_API Appearance : public Object {
+    OpenSim_DECLARE_CONCRETE_OBJECT(Appearance, Object);
```
chrisdembia: Should this be abstract?

OpenSim/Simulation/RegisterTypes_osimSimulation.cpp
chrisdembia: I am not really familiar with the geometry changes but we could have issues with the fact that there is still a valid Appearance class. People's valid Appearance classes could be silently converted to SurfaceAppearance. If the new Appearance class is in fact virtual, then the issue might go away. Not sure though. 

If this line of code is not changed, then we should at least add a comment so people know that this instance of "Appearance' is referring to what will be a historical intermediate version of the SurfaceAppearance class.

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -37,22 +37,21 @@ class Model;
...
 //=============================================================================
 /**
- * A class that holds the Display Attributes (Appearance) of an object displayed 
- * in the OpenSim Visualizer.
+ * A class that holds the Appearance Attributes of an object displayed 
+ * in the OpenSim Visualizer/GUI.
  * 
- * BaseAppearance contains Appearance properties that apply to all geometry.
+ * Appearance objects contain properties that are common to all geometry.
  * Geometry that have a surface so that it can be textured can use the subclass
```
aseth1: It is not clear from the class description why Appearance is needed rather than having properties (attributes) of Geometry. We need to state the difference between Geometry and Appearance. Geometry being a representation of physical shapes as a Component for computation and Appearance being a convenience class of knobs for Visualizers to use to control how that geometry is visualized if at all. Is that correct?

aymanhab:
@chrisdembia Indeed Appearance will be abstract class fater this change though we  usually use the macro OpenSim_DECLARE_ABSTRACT_OBJECT for this functionality rather than explicitly stating the class to be abstract and I agree it'd be good to do this change. If Appearance doesn't show in files as expected then the rename will be transparent to users.

chrisdembia:
> If Appearance doesn't show in files as expected then the rename will be transparent to users.

Sorry for not being on top of the geometry changes; I did not know that Appearance won't be in the XML.

aymanhab:
@aseth1 Yes, that's a correct interpretation of the distinction.

OpenSim/Simulation/Model/Geometry.h
```diff
@@ -188,13 +187,32 @@ class OSIMSIMULATION_API Geometry : public Component {
     }    //=====================================================================
 };  // END of class Geometry
 
+class OSIMSIMULATION_API DrawingGeometry : public Geometry
+{
```
aseth1: Why is DrawingGeometry and DrawingAppearance necessary. I see a potential explosion of Geometry and Appearance classes with no apparent gain in functionality. Why are we doing this?

aymanhab:
@aseth1 I don't see a potential for "explosion" :boom: since geometry is either 1D or 2D (the naming is what you suggested and we agreed on i.e. BaseAppearance -> Appearance, Appearance -> SurfaceAppearance). All 2D Geometries can have surface properties e.g. texture, shininess etc. while 1D Geometry doesn't, instead they may have thickness, line-style, etc. There're two kinds of 1D Geometry that are used (1. a rubber-band line or muscle connecting points and 2. drawings that represent icons e.g. Arrows, Frames, Axes, ..) I don't see any reason that list would grow. 
 This rather shallow class hierarchy on the OpenSim side is to make sure only relevant attributes are shown to users (e.g. no thickness for a Sphere or texture for a line or frame).

aseth1:
> while 1D Geometry doesn't, instead they may have thickness, line-style, etc. There're two kinds of 1D Geometry that are used (1. a rubber-band line or muscle connecting points and 2. drawings that represent icons e.g. Arrows, Frames, Axes, ..)

DrawingAppearance to represent icons is somewhat strange. I don't see a Drawing being conceptually related to the visuals of Axes.  

It is not clear to me how Geometry and Appearance are related. It looks like a Geometry has an Appearance (or at least can cough one up) but how does someone adding new Geometry associate an Appearance with their Geometry? 

For example, if I have SphereLike geometry should I create a SphereLikeAppearance for it or use SurfaceAppearance? What if you want it to appear bigger? For example a Marker. Would that be related to size of a DrawingAppearance?  

If you generally have to create a new Appearance for new Geometry types then I think there will be many more Appearances popping up.

aymanhab:
@aseth1 Geometry has a kind-of-Appearance Property but the specific class depends on the type of Geometry (Surface vs. Icon etc.). To clear up your concern, adding new blobGeometry that's kind of a surface will get and re-use SurfaceAppearance out of the box and so will introduce no new Appearance subclasses whatsoever. The separation between Geometry and Appearance is actually pretty much the standard in threejs, Collada and other modern formats (the term used is sometimes Material but we avoided the confusion with other physical material properties early on.)

aseth1:
I brought up the SphereLike geometry because it isn't clear that it would be sufficient to use SurfaceAppearance. How would we control the display radius (or) size of the Geometry? That seems to be the main visual knob for DrawingAppearance.

The use case is this. We have Marker which is a Point with Sphere geometry for the purpose of visualizing the marker. Physically the Marker has no size or radius - it is a point on a PhysicalFrame.  We want the user to be able to control the size, color and/or texture only for the appearance of the Marker. How does either SurfaceAppearance or IconAppearance achieve this? It seems like we need MarkerAppearance? How else do you relate the Sphere's radius to the Appearance's size?

If the user can change the size of the sphere through its appearance we would then want to hide the radius property of the Sphere. In which, the Sphere could be a "hidden" property (not exposed for editing) of the Marker but its visualization is not hidden (unless its representation in the Appearance is set to Hidden).

sherm1:
The visualization of a point that happens to be done with a sphere should not be confused with an actual Sphere! In that case it is a kind of icon that happens to be sphere-like. I was using spheres for points until a few years ago when I switched to little crosses -- but the real geometry remained a point either way.

chrisdembia:
I really like the cross for contact points, but markers *are* spheres, not points.

sherm1:
> I really like the cross for contact points, but markers are spheres, not points.

Aren't they actually little reflective circles?

aymanhab:
Let me recap here:
1. For objects that have intrinsic shape (e.g. ContactSphere, WrapCylinder or an Ellipsoid to visualize a body) there's no issue with either Geometry (primitive or mesh) or its appearance (SurfaceAppearance which allows for Texture as expected).
2. Objects that represent icons (e.g. Axes, Frames, ..) have associated appearance that have a knob for thickness, don't allow for texture etc. as expected too.
3. Some objects are in the gray area as they do not have intrinsic shape (e.g. a point is really a location and not an object. Examples are Marker, PathPoint?, COM?, COP?, ContactPoint if it can be visualized) and we can decide to use either option 1 (spheres) or option 2 (circles or crosses). 
My concern really is not the specific choices/preferences which may change over time but whether we're flexible enough to accommodate all reasonable choices.

aymanhab:
Closing due to changes proposed in Dev meeting 08/13/2015

PR #564, opened 2015-07-31
Make use of Frames with BushingForce as example r2

aseth1:
Based on review of previous PR, I updated the structure:
BushingForce dependency on PhysicalFrame is no longer made through the PhysicalOffsetFrame subcomponents/properties. BushingForce itself has connectors to PhysicalFrames.
It supports internal (owned) PhysicalFrame as subcomponents to satisfy the connectors if necessary. 

`BushingForce::updateFromXMLNode()` was revised to create PhysicalOffsetFrames from older version's properties and adds them to the BushingForce.

There is only one convenience constructor that takes two PhysicalFrames + stiffness and damping properties.


OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.cpp
```diff
@@ -50,37 +50,23 @@ CoupledBushingForce::~CoupledBushingForce()
...
-    _body1Name = body1Name;
-    _body2Name = body2Name;
-    _locationInBody1 = point1;
-    _orientationInBody1 = orientation1;
-    _locationInBody2 = point2;
-    _orientationInBody2 = orientation2;
+    constructInfrastructure();
+
+    updConnector<PhysicalFrame>("frame1").set_connected_to_name(frame1.getName());
+    updConnector<PhysicalFrame>("frame2").set_connected_to_name(frame2.getName());
```
sherm1: If the `CoupledBushingForce` only depends on the *names* of the `PhysicalFrame` arguments, should those arguments just be the names rather than the binary object? That would allow some out-of-order model building as can be done through XML, I think.

OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.cpp
```diff
@@ -89,239 +75,78 @@ CoupledBushingForce::CoupledBushingForce( std::string body1Name, SimTK::Vec3 poi
...
+    by the bushing force. Angular displacement expressed in Euler angles.
     The force and potential energy are determined by the deflection.  */
 SimTK::Vec6 CoupledBushingForce::computeDeflection(const SimTK::State& s) const
 {
-    const Transform& X_GB1 = _b1->getBodyTransform(s);
-
-    const Transform& X_GB2 = _b2->getBodyTransform(s);
+    // get connected frames
+    const PhysicalFrame& frame1 = getConnectee<PhysicalFrame>("frame1");
+    const PhysicalFrame& frame2 = getConnectee<PhysicalFrame>("frame2");
```
sherm1: Connectee! I like it.

OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.h
```diff
@@ -120,73 +100,77 @@ OpenSim_DECLARE_CONCRETE_OBJECT(CoupledBushingForce, Force);
...
-                  SimTK::Mat66 stiffnessMat, SimTK::Mat66 dampingMat);
-    CoupledBushingForce(const CoupledBushingForce &aForce);
-    virtual ~CoupledBushingForce();
+    /** Construct the CoupleBushingForce given the two offset frames to which
+    it attaches by name. Stiffness and damping properties are supplied by 6x6
+    matrices. Rotational stiffness (damping) in N/rad(/s) and translational in
+    N/m(/s). Off-diagonals represent the coupling terms.
+    See property declarations for more information. **/
+    CoupledBushingForce(const PhysicalFrame& frame1,
+                        const PhysicalFrame& frame2,
```
sherm1: names here rather than `PhysicalFrame`s?

OpenSim/Simulation/Model/BushingForce.cpp
```diff
@@ -44,30 +44,24 @@ using namespace OpenSim;
...
-                           const Vec3&      point2, 
-                           const Vec3&      orientation2,
-                           const Vec3&      transStiffness, 
-                           const Vec3&      rotStiffness, 
-                           const Vec3&      transDamping, 
-                           const Vec3&      rotDamping)
+/* Convenience construction that creates the Offsets on the Physical Frames (e.g.
+   Bodies that the BushingForce acts between. */
+BushingForce::BushingForce( const PhysicalFrame& frame1,
+                            const PhysicalFrame& frame2,
```
sherm1: names?

OpenSim/Simulation/Model/BushingForce.cpp
```diff
@@ -85,64 +79,126 @@ void BushingForce::setNull()
...
+            SimTK::Xml::element_iterator locBody2Elt =
+                aNode.element_begin("location_body_2");
+            SimTK::Xml::element_iterator orientBody2Elt =
+                aNode.element_begin("orientation_body_2");
+
+            //Create two new PhysicalOffsetFrames
+            PhysicalOffsetFrame frame1;
+            PhysicalOffsetFrame frame2;
+            frame1.setName("frame1_offset");
+            frame2.setName("frame2_offset");
```
sherm1: Would it make sense to *not* create these frames if the pose was (0,0,0,0,0,0) or not present? Could use the body frames directly in that case rather than locally-owned ones.

OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.cpp
chrisdembia: I agree. I know @aseth1 in other contexts has supported passing in actual objects rather than string names. I think that is a good idea where it makes sense (e.g., when the function uses the object for more than its name), but not here. Especially considering the fact that, as @aseth1 discussed with me, the object that `append_frames()` creates a copy of the frame and so the object passed to this constructor isn't the same as the one that ends up in the property.

sherm1:
Reviewed. Looks good to me. The only substantive question I have is why the constructor wants a `PhysicalFrame` object rather than a `PhysicalFrame` name which appears to be the only thing it depends on and would be more like XML model building.

Love to see more use of "connectee" too!

OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.cpp
aseth1: Good points. The change is on its way.

OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.cpp
aseth1: I will change `Connector::set_connected_to_name()` to `Connector::set_connectee_name()` in a separate PR.

aseth1:
@sherm1 + @chrisdembia convenience constructor updated to use frame name.

sherm1:
I think this is ready to go in, other than nuking a few underscores.

aseth1:
Thanks for pointing these out. It should now conform to our naming conventions.

OpenSim/Simulation/Model/BushingForce.cpp
```diff
@@ -200,39 +247,42 @@ double BushingForce::computePotentialEnergy(const SimTK::State& s) const
...
+    labels.append(getName()+"."+frame1Name+".force.Z");
+    labels.append(getName()+"."+frame1Name+".torque.X");
+    labels.append(getName()+"."+frame1Name+".torque.Y");
+    labels.append(getName()+"."+frame1Name+".torque.Z");
+    labels.append(getName()+"."+frame2Name+".force.X");
+    labels.append(getName()+"."+frame2Name+".force.Y");
+    labels.append(getName()+"."+frame2Name+".force.Z");
+    labels.append(getName()+"."+frame2Name+".torque.X");
+    labels.append(getName()+"."+frame2Name+".torque.Y");
+    labels.append(getName()+"."+frame2Name+".torque.Z");
```
chrisdembia: @aseth1 could you add some stuff to the CHANGELOG, including this backwards incompatible change here? Someone could have a postprocessing script that expects "mybushing.pelvis.torque.X"; now they'll get "mybushing.frame1_offset.torque.X",

OpenSim/Simulation/Test/BushingForceModel_30000.osim
```diff
@@ -0,0 +1,181 @@
...
+					<!--The other of the two bodies connected by the bushing.-->
+					<body_2>ball</body_2>
+					<!--Location of bushing frame on body 1.-->
+					<location_body_1>0 0 0</location_body_1>
+					<!--Orientation of bushing frame in body 1 as x-y-z, body fixed Euler rotations.-->
+					<orientation_body_1>0 0 0</orientation_body_1>
+					<!--Location of bushing frame on body 2.-->
+					<location_body_2>0 0 0</location_body_2>
+					<!--Orientation of bushing frame in body 2 as x-y-z, body fixed Euler rotations.-->
+					<orientation_body_2>0 0 0</orientation_body_2>
```
chrisdembia: Is there any particular reason for using zeros here?

OpenSim/Wrapping/Python/tests/test_swig_additional_interface.py
```diff
@@ -63,10 +63,9 @@ def test_markAdopted2():
     constr.setConstantDistance(1)
     a.addConstraint(constr)
 
-    # Force requires body names. If not provided, you get a segfault.
-    f = osim.BushingForce()
-    f.setBody1ByName("ground")
-    f.setBody2ByName("body")
+    # Force requires frames. If not provided, you get a segfault.
+    f = osim.BushingForce("ground", "body", osim.Vec3(2, 2, 2), osim.Vec3(1, 1, 1), osim.Vec3(0, 0, 0), osim.Vec3(0, 0, 0))
+
```
chrisdembia: Hmm..perhaps this test should be revisited. I wanted to test in particular using the default constructor. I wonder if with your changes the segfault goes away.

chrisdembia: I think the segfault will have gone away, since BushingForce now uses connectors thoroughly.

aseth1:
@sherm1 @chrisdembia merge at will.

OpenSim/Wrapping/Python/tests/test_swig_additional_interface.py
aseth1: Can you test it and revise the comment?

chrisdembia: Testing.

chrisdembia:
@aseth1 I made the changes I wanted to the python wrapping. We can merge after the tests pass.

sherm1:
All tests passed. Merging ...

PR #565, opened 2015-08-01
Use MathJax for doxygen equations

### LINKED ISSUE #374: Important windows dependencies missing from readme
The windows build -> Get Dependencies -> API Documentation lists Doxygen as the only dependency.  This is incorrect.  The doxygen build on windows also depends on:

1. Tex
2. LaTex
3. GhostScript

There are many distributions and options for obtaining any/all three, so some discussion about the best, recommended options is in order.

chrisdembia:
This addresses most of #374: building doxygen requires latex, but that dependency is not mentioned in the readme and on confluence it's mentioned as optional.

With this PR, doxygen uses MathJax instead of latex for equations by default. How does this work?

1. Users can set an option `OPENSIM_DOXYGEN_USE_MATHJAX` which is ON by default. 
2. If using mathjax, we try to download MathJax from the internet when you run CMake. You need internet when running doxygen but you don't need internet when viewing doxygen.
3. If we cannot download mathjax, we turn off the option above.

If the option is off:

1. We look for latex on your machine. If you have it, we use it for equations.
2. Otherwise we use mathjax CDN (the internet) to render equations *when viewing* doxygen (that is, you need internet when viewing the docs). If you don't have internet when viewing docs, then you see raw latex instead.

I considered updating the README but I decided it's currently correct and simple; we don't need to tell people we're going to download mathjax for them.

We cannot yet close #374 until we also update the confluence documentation, but we can't do that unless we make a private copy of the doxygen instructions (or decide the fate of such documentation for 4.0).

I tested extensively on osx and briefly on windows. I think this is ready to go in.

Could a subset of @msdemers @jimmyDunne @aymanhab review?


chrisdembia:
@aymanhab we discussed in the dev meeting that we'd like for you to take a look at this before it gets merged.

aymanhab:
Thanks @chrisdembia the changes look fine though will need to test if it works and produces well formatted documentation/equations cross-platform. Assuming this is the case (if not let me know what platform to test on), a couple quick questions:
+ Does it always download even if a version of MathJax is available locally?
+ Is MathJax changing frequently that we need to verify that new versions do not break our docs?

chrisdembia:
> Does it always download even if a version of MathJax is available locally?

Yes. On Ubuntu, you *can* get MathJax from `apt-get`. From my reading of the MathJax documentation, it doesn't seem like MathJax is something that you generally install system-wide. It's not like latex; there's no installer, etc.

> Is MathJax changing frequently that we need to verify that new versions do not break our docs?

We are downloading a specific version of MathJax, not the latest version, so I think this is fine. The issue is moreso that the URL may become invalid in the future; that's where the fallback to latex and the online CDN comes in handy....oh now I realize that the CDN will always have the latest version...this is more on doxygen to keep up to date with mathjax than for us.

chrisdembia:
Sorry..I should have answered your question. I don't know how frequently mathjax is changing, but since it is using the latex markup for input I'm not too concerned.

chrisdembia:
Also, our installed doxygen on the internet will always use an internal copy of mathjax, so we don't have to worry about how mathjax changes.

aymanhab:
Thanks much @chrisdembia the other part was which platforms we tested on, if enough testing is done I can merge this.

chrisdembia:
I fully tested this on OSX, so I don't think we need to test this on Linux. I did very minimal testing on Windows: I made sure that MathJax gets downloaded, but I didn't try building doxygen. I'm happy to test this fully on Windows if you'd like.

chrisdembia:
I just addressed the merge conflicts.

aymanhab:
@chrisdembia If it's easy for you to do in background then great, else let me know and I'll test in my windows environment. Thanks.

chrisdembia:
@aymanhab ready for review. I tested more completely on windows and fixed the issues I ran into.

aymanhab:
@chrisdembia Thanks, it all looks good. Can we document somewhere the changes to CMake configuration process if any for release builds and/or internal dev. builds?

chrisdembia:
Thanks @aymanhab. I don't think documentation is required.

tkuchida:
Will this PR fix the equations in the documentation posted online (e.g., [here](https://simtk.org/api_docs/opensim/api_docs/classOpenSim_1_1Millard2012EquilibriumMuscle.html#details))?

eqn

jenhicks:
It will fix the equations for future releases. @jimmyDunne is also giving Henry an updated build of doxygen with equations as images. This should be posted in the next couple days (or sooner).

tkuchida:
@jenhicks Thanks. Online documentation issue resolved [here](https://github.com/opensim-org/opensim32/issues/69).

PR #566, opened 2015-08-02
CMake cleanup for Mac, and Mac users don't need to set DYLD_LIBRARY_PATH

chrisdembia:
Mac users no longer need to set DYLD_LIBRARY_PATH to use OpenSim executables or python wrapping, which could save a lot of headache. They still might need to set DYLD_LIBRARY_PATH for java wrapping, MATLAB, or when building their own executables and libraries (though if they make use of RPATH themselves, they don't need to).

Basically, the path to the libraries is saved in the executables: e.g., the analyze tool contains a relative path to the opensim libs, and the opensim libs contain a full path to the simbody libs. The python _opensim.so library contains the full path to the opensim libs.

This all works great if you're building from source on a mac, which is currently the only option. If we ever distribute binaries for mac of OpenSim or Simbody, we'll have to revisit. In such a case, I think we should be using OSX frameworks. If we want to be on OSX for 4.0, then I can look into OSX frameworks (though I am not sure if netbeans works with OSX frameworks). I think they would be very straightforward to set up.

I may eventually get around to using RPATH for Linux as well.

@sherm1 would you like a similar PR for Simbody?

I also did some vacuuming of the root CMakeLists.txt.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,10 +12,6 @@ cmake_minimum_required(VERSION 2.8.8)
 if(COMMAND cmake_policy)
     cmake_policy(SET CMP0003 NEW)
     cmake_policy(SET CMP0005 NEW)
-    if(NOT (${CMAKE_VERSION} VERSION_LESS 3.0))
-        # MACOSX_RPATH enabled by default; policy introduced with cmake 3.0.
-        cmake_policy(SET CMP0042 NEW)
-    endif()
 endif(COMMAND cmake_policy)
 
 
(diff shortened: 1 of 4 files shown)


sherm1:
> RPATH on OSX: @sherm1 would you like a similar PR for Simbody?

Thanks, Chris. I'm not sure about this because (a) I don't know if there is a problem that needs solving here, and (b) I'm not sure how to test it adequately. @scpeters may have some thought on this -- Steve, do you make use of RPATH for shared library finding?

scpeters:
I checked with someone here at OSRF who says that RPATH is good for OSX, though he is neutral about its use on Linux.

+1

chrisdembia:
Thanks @scpeters . 

I understand the neutral feeling on Linux because libraries on linux are all in one place. We don't yet support a system-wide install, though.

I think having users set DYLD_LIBRARY_PATH on their own is a problem. People forget to do it, people shouldn't have to do it, and it makes OpenSim feel like it doesn't really belong on your computer (that you have to do something special to use it).

sherm1:
Chris, makes sense for OpenSim. For Simbody I'm not sure where this would help (example programs, sample CMake build scripts?).

chrisdembia:
> Chris, makes sense for OpenSim. For Simbody I'm not sure where this would help (example programs, sample CMake build scripts?).

Good point. I think for Simbody we should put `@rpath/` in the install name of the SimTK libraries, but but we don't necessarily need to define an RPATH for the executables that use the libraries. This might already be done; not sure though.

chrisdembia:
@msdemers would you like to review this? If not, I'll merge it since I'm not sure anyone else is currently interested.

chrisdembia:
I'm going to merge this. We can undo if we run into issues.

PR #571, opened 2015-08-05
Suppress VS2015 warnings.

chrisdembia:
This is the same fix that I think @sherm1 made for opensim32.


sherm1:
Looks good to me, Chris.

PR #572, opened 2015-08-06
Replaced "connected_to" language with "connectee" ...

aseth1:
such that a connector has a connectee and it can be identified by name that is the connectee_name.

### DIFF
--- a/Applications/Scale/test/std_toyLigamentModelScaled.osim
+++ b/Applications/Scale/test/std_toyLigamentModelScaled.osim
@@ -10,7 +10,7 @@
 					<connectors>
 						<Connector_Frame_ name="frame">
 							<!--Name of the component this Connector should be connected to.-->
-							<connected_to_name>ground</connected_to_name>
+							<connectee_name>ground</connectee_name>
 						</Connector_Frame_>
 					</connectors>
 					<!--Scale factors in X, Y, Z directions respectively.-->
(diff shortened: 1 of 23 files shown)


chrisdembia:
Could you update the doxygen for AbstractConnector to formally define the term "connectee" and also update it in general, since it is incorrect (joints connect to physical frames, and not to just 1; joints are not owned by a body). Here's a copy of the current doxygen:

```
 * A Connector formalizes the need for a connection between a Component and a
 * dependent object, without owning the object it is connected to. The purpose
 * of a Connector is to specify: 1) the object type that the Component is 
 * dependent on, 2) by when (what stage) the connector must be connected in
 * order for the component to function, 3) the name of object to connect to and
 * 4) whether it is connected or not. The specific instance that satisfies the
 * connector's requirement is maintained by the Component's list of Connections.
 *
 * For example, a Joint has one Connector for the parent Body that it joins
 * its owning Body to. The type for the connector is Body and any attempt to 
 * connect to a non-Body object will throw an exception.
 * The connectAt Stage is Topology. That is the Joint's connection to a Body 
 * must be performed at the Topology system stage, and any attempt to change 
 * the connection status will invalidate that Stage and above.
 *
 * Other Components like a Marker or a Probe that do not change the system
 * topology or add new states could potentially be connected at later stages
 * like Model or Instance.
```

After that this is good to merge.

chrisdembia:
Merging... However, @aseth1, the `@file`-level doxygen comments in `ComponentConnector.h`...where does it show up in doxygen? I couldn't find it. Could you address this eventually?

sherm1:
Looks great! :+1:

PR #573, opened 2015-08-08
Split targets into VS folders.

chrisdembia:
I added some changes to cmake so that the targets are split up into categories. I basically took the prefix in the project names and made them into folders. Here's what the solution explorer looks like with these changes:

image

The only downside I see is that unexperienced users may have trouble finding "INSTALL", which might be the only target they're interested in.


sherm1:
That looks good. Does it help in XCode also?

chrisdembia:
Seems like it does:

image


I got the idea from @tgeijten

sherm1:
Nice!

chrisdembia:
@aseth1, @aymanhab do you guys like the change or should we close the PR?

aymanhab:
@chrisdembia I like the fact that when working on a single test or library that I can collapse the lists and find my target easily. Would be best if INSTALL can be pulled out but if not that's still an improvement. I wouldn't worry about unexperienced developers building from source, they'll hit other hurdles before they get to install!

chrisdembia:
Thanks for looking @aymanhab . I investigated if INSTALL could be pulled out and I could not find a way :/

aymanhab:
No big deal, can we easily add a group/Folder for Scripting as it used to be hard to locate these Wrapping projects in the long list, or I can add it myself separately.

Thanks again @chrisdembia

chrisdembia:
Oh good idea. I'll do it.

aseth1:
I second Ayman's sentiment. I welcome the organization. If the default view is that the list is fully expanded (which very much resembles the current flat list) then finding INSTALL isn't any more difficult than it is now. I wouldn't let that hold back this change. Thanks @chrisdembia.

chrisdembia:
Great. I added wrapping folders. This is ready to merge.

aymanhab:
Great, thanks @chrisdembia

chrisdembia:
Thanks @aymanhab

PR #574, opened 2015-08-09
Use in-class member initializers for OpenSim_DECLARE_PROPERTY

chrisdembia:
We've all forgotten to call `constructProperty_my_prop()` in our constructors. Thankfully, we get a great error message when this happens.

With the introduction of inputs, outputs, and structural connectors, there's a lot more that must happen during construction and we need a way to manage all of these tasks. For this reason, @aseth1 introduced `constructInfrastructure()`; unfortunately there are some issues with that approach.

An alternative is to avoid adding any code into constructors by using C++11 in-class member initializers.

This PR is an attempt to allow one to initialize (and set the default value for) required properties (those declared with OpenSim_DECLARE_PROPERTY) from the macro call, removing the need for a call to `constructProperty_my_prop()`. For example:

```
OpenSim_DECLARE_PROPERTY(mass, double, "Subject mass", SimTK::NaN);
```

With this new macro, we automatically add the default value to the doxygen documentation for the property (should it be called initial value instead?). Automatically documenting the default value is a huge plus; since otherwise it's undocumented or it is documented in an unmaintainable way.

Here's the documentation for the macro:

image

Here's the brief description for 3 properties that use the new macro:
image

And here's a detailed description:
image

I want this change to be backwards-compatible, so I introduced an overloaded macro. This adds quite a bit of complexity. For example, compiler errors will be harder to interpret, since there are more intermediate macros.

If we decide that the complexity is unwarranted, I can create a separate macro `OpenSim_DECLARE_PROPERTY_INIT` that takes the additional argument.

If we like the change, I can go ahead and make the same change for the remaining property macros. Notice that this PR does not go into the master branch; it goes into a feature branch.


chrisdembia:
Here's the stackoverflow page I used for the overloaded macro: http://stackoverflow.com/questions/11761703/overloading-macro-on-number-of-arguments

sherm1:
This looks great, Chris. I would propose that in 4.0 we go further and eliminate the `constructProperty_xxx` stuff altogether rather than attempt to explain it, which in some sense make the whole system even *more* complicated. We could either require a default value for every property, or figure out how to provide a default initializer.

chrisdembia:
@sherm1 I agree about removing `constructProperty_xxx()` altogether eventually, and I agree that this change increases complexity. But what about backwards compatibility in the short term?

We could use the default constructor for each object to initialize the property if the user doesn't provide an initial value, but I prefer requiring a default value.

sherm1:
> We could use the default constructor for each object to initialize the property if the user doesn't provide an initial value, but I prefer requiring a default value.

Using a default constructor would probably be a reasonable choice since that's what C++ does anyway, so people will expect that behavior. That would allow for backwards compatibility at the macro. For compatibility with existing `constructProperty_xxx` calls we could still generate those methods and maybe deprecate them.

We would need to invent some defaults for the built-in types, like setting an int property to 0.

chrisdembia:
So as I understand: we would have the macro *always* call `addProperty<T>()`, regardless of whether or not the user provides an initial value via the macro. Then, we would change the implementation of `constructProperty_xxx()` so that instead of calling `addProperty<T>()`, it simply changes the value of the property.

chrisdembia:
I like that idea. If we take that route, we should remove all our `constructProperty_xxx()` calls (I guess that's obvious, since we'd deprecate that method).

chrisdembia:
Let me go back on what I said a bit: I don't think that the next release should require a default value in the macro; that would be backwards incompatible. In future releases, we can require it.

chrisdembia:
I just spent some time now trying to use the default constructor if the user does not provide a default value through the property. I ran into two issues:

1. Right now, you can have properties for incomplete types (that are forward-declared). We make use of this in OpenSim (`HuntCrossleyForce::ContactParameterSet`).
2. Some properties have an abstract type, and we can't default-construct an abstract type.

For both of these reasons, using a default constructor within the property macro is not backwards-compatible.

For these reasons, my preference is now to *not* default-construct within the property macro, and to permanently allow the option of using `constructProperty_prop_name()` because I think the flexibility of forward-declaring the type of a property is nice; we can still talk of this method as "advanced".

chrisdembia:
@sherm1 suggested creating a new `addProperty<T>` that does not take an initial value. I'll look into this.

chrisdembia:
3 cases: OpenSim_DECLARE_UNINITIALIZED_PROPERTY, OpenSim-DECLARE_INITIALIZED_TO_DEFAULT_CONSTRUCTOR_PROPERTY, DECLARE_INITIALIZED_PROPERTY

chrisdembia:
With the help of @sherm1's ResetOnCopy class, I figured a way to use just one macro, `OpenSim_DECLARE_PROPERTY`, to handle (1) user provides 3 arguments (no initial value), abstract classes (actually, non-default-constructible) in which case the property is not initialized; (2) user provides 3 arguments (no initial value), default-constructible classes in which case the property is initialized to a default-constructed instance; and (3) user provides 4 arguments, the 4th being an initial value. So this is backwards-compatible for abstract Object properties.

**However**, in the case of incomplete types, you can't use the single OpenSim_DECLARE_PROPERTY macro.

So the tradeoff is between (a) introducing a backwards-incompatibility, and requiring those who had incomplete types in their property use a different macro, say OpenSim_DECLARE_PROPERTY_NOINIT; and (b) staying backwards-compatible and requiring that people use a new macro if they want the property to be initialized via the in-class member initializer (OpenSim_DECLARE_PROPERTY_INIT).

There is still one more option I have to explore (that @sherm1 suggested a while ago): creating a new `addProperty(name, comment)` that does *not* take an initial value argument. However, then we can't default-construct if the user provides only 3 arguments. So this might not be different from (3) above. I have to think about this more.

PR #576, opened 2015-08-10
[cmake] Update required simbody version to 3.6.

chrisdembia:
I know simbody 3.6 is not released yet, but the master version of simbody is labeled as 3.6.


aymanhab:
:+1:  This prevents CMake from quietly finding an earlier install of Simbody and then failing to compile later with mysterious errors due to API changes (e.g. ClonePtr).

sherm1:
Good idea, Chris. Mergeable after CI.

PR #577, opened 2015-08-10
Move to VS2015.

chrisdembia:
I've created this PR in preparation for our move to VS2015. I think this PR can be used to mark our official move to VS2015 (so, not just yet).

I've updated the README instructions to say that Windows users need VS2015. When Simbody moves to VS2015, I'll also update the appveyor script so that it builds with VS2015.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,11 +112,15 @@ On Windows using Visual Studio
 * **operating system**: Windows 7 or 8.
 * **cross-platform build system**:
   [CMake](http://www.cmake.org/cmake/resources/software.html) >= 2.8.8
-* **compiler / IDE**: Visual Studio 2013. We recommended either:
-    * *Visual Studio Express 2013 for Windows Desktop*, which is free, or
-    * *Visual Studio Professional 2013* through
-      [Dreamspark](https://www.dreamspark.com) if you are at an academic
-      institution.
+* **compiler / IDE**: [Visual Studio 2015](https://www.visualstudio.com/).
+    * *Visual Studio Community 2015* is sufficient and is free for everyone.
+        If you want to use *Visual Studio Enterprise 2015*, you may be able 
+        to get it for free at [Dreamspark](https://www.dreamspark.com) if 
+        you are at an academic institution.
+    * Visual Studio 2015 does not install C++ 
+      support by default. During the installation you must select 
+      *Custom*, and check *Programming Languages > Visual C++ > Common Tools for Visual C++ 2015*.
+      You can uncheck all other boxes.
 * **physics engine**:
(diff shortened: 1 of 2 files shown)


sherm1:
Do you want to make a similar PR with the appveyor change for Simbody, Chris? I think we are just about ready to make the switch.

chrisdembia:
Gladly.

sherm1:
Simbody's AppVeyor build is now using VS2015 (see PR simbody/simbody#425). OpenSim should follow suit today to avoid AppVeyor failures.

/cc @aymanhab, @aseth1, @klshrinidhi

chrisdembia:
There is at least one issue to deal with before we can merge this:

```
CMake Warning at C:/Program Files (x86)/CMake/share/cmake-3.1/Modules/InstallRequiredSystemLibraries.cmake:424 (message):
  system runtime library file does not exist: 'C:/Program Files
  (x86)/Microsoft Visual Studio
  14.0/VC/redist/x64/Microsoft.VC140.CRT/msvcr140.dll'
Call Stack (most recent call first):
  CMakeLists.txt:277 (include)
```

sherm1:
I'm confused about the C4275 warnings in AppVeyor (non-dll base class `std::exception` used for dll class `OpenSim::Exception`) because that warning is disabled in Exception.h [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Exception.h#L37). I'm not seeing that in my local build but I'm not doing wrapping -- maybe this is SWIG-related?

sherm1:
> There is at least one issue to deal with before we can merge this (redist library not available for VS2015)

That's just a warning so we could merge this if it is needed to get the build working and address the warning separately.

chrisdembia:
> That's just a warning so we could merge this if it is needed to get the build working and address the warning separately.

True...

> I'm not seeing that in my local build but I'm not doing wrapping -- maybe this is SWIG-related?

Ah thanks...I was confused about this as well. I'll look into this eventually.

sherm1:
Builds worked -- ready to merge IMO.

sherm1:
I'm merging this change to use VS2015 on AppVeyor so the OpenSim 4.0 builds will continue to work. There are still some warnings to deal with.

chrisdembia:
I read up on the CMake warning here:

http://www.cmake.org/cmake/help/v3.0/module/InstallRequiredSystemLibraries.html

This makes me think that we just don't need to address the warning. I didn't see the warning myself on my machine, and I don't think others have (Ajay didn't say anything).

chrisdembia:
Also, I no longer see any of the Exception.h C4275 messages in appveyor, so I think there is nothing to do.

PR #581, opened 2015-08-16
Run debug tests on travis.

chrisdembia:
@aymanhab ready for review.


aymanhab:
@chrisdembia I don't know how you came up with the list to exclude and if it's cross platform. Also the clang Debug build fails even though all chosen tests seem to have passed, any clues?

chrisdembia:
I came up with the list based on which tests fail on linux and which tests were timing out on travis's machines (if a test takes more than 10 minutes, travis fails). Yeah the travis-pr test failed because travis just happened to time out that time. I'm trying to exclude more tests.

My motivation here was not to ensure that we test all tests that we know should pass...it was to do *some* checking of the tests in debug. In the future we can try including more tests.

aymanhab:
@chrisdembia I agree with the criteria, I'm just hoping that we don't cut it too close to timeout and end up with bogus failures on non-travis machines. I'd probably cut any test that exceeds 5 min if 10 min times out to be safe.

aymanhab:
This looks fine now, @chrisdembia, thanks.  Is there a way to make this change in our CMake files rather than only on travis so we don't have to wait for Travis to check if we broke Debug builds? That would be good to know but I'd be happy to merge this as is since it's more testing/confidence anyway.
On windows with VS2015 debug builds, I'm not getting any logs of failures in Debug builds. Is this expected?

aymanhab:
@chrisdembia I'll merge this and we'll deal with debug builds in dev. environment separately.

chrisdembia:
Thanks @aymanhab . It seems we can use the `WILL_FAIL` test property to solve some of these issues.

http://www.cmake.org/cmake/help/v3.0/manual/cmake-properties.7.html?highlight=test%20properties#properties-on-tests

Issue #533, opened 2015-07-15
Member variable name convention: `_name`, `m_name`, `name_`


chrisdembia:
In #530, @sherm1 said:

> Trailing underscore to indicate member variables does not match the OpenSim coding guidelines, which specify leading underscore. Personally I hate leading underscore since it is borderline non-compliant with the C++ standard. Much more common (and conformant) is the m_ convention which I have been moving to in Simbody. We should discuss to see whether the OpenSim team would be amenable to using m_, otherwise I think you have to stick with leading underscore. I don't think we should introduce a third convention in any case.

I'm opening this issue to start this discussion.

I personally really like `m_`. It looks professional to me. I am in favor of allowing `_name` and `m_name`, as long as only one convention is used in any given class.

sherm1:
> I am in favor of allowing _name and m_name, as long as only one convention is used in any given class.

I like that idea. I would propose that we formally make `m_` the "preferred" convention for new code, and encourage gradual switching to that convention whenever a class undergoes major hacking.

chrisdembia:
> I like that idea. I would propose that we formally make m_ the "preferred" convention for new code, and encourage gradual switching to that convention whenever a class undergoes major hacking.

+1

klshrinidhi:
I would like OpenSim to follow Google C++ Style Guide:
https://google-styleguide.googlecode.com/svn/trunk/cppguide.html

I know it is very important to stick to one convention for readability and maintainability. I don't like some things about the OpenSim coding conventions. I am suggesting the above style guide because it comprehensive, it is updated regularly, and a lot of projects follow it. C++ developers are used to seeing that convention in other open source projects. There will be things in that style guide that some of us don't like but I think it is a very good compromise.

sherm1:
Google's is a nice style guide, but not uniformly revered. For example, I think this would be awkward for us (from the section on Exceptions):
> We do not use C++ exceptions.

klshrinidhi:
I think exceptions are the only thing they don't cover because I heard that internally Google does not use exceptions in their code. But because exceptions are tiny classes, we can apply class conventions to exceptions too.

sherm1:
[...]

PR #582, opened 2015-08-17
Use of TimeSeriesTable in MarkersReference

klshrinidhi:
This is to test use of TimeSeriesTable in MarkersReference instead of MarkersData.

The files:
* DataAdapter.h & DataAdapter.cpp
* FileAdapter.h & FileAdapter.cpp

follow up on the files committed by Ajay @21807925adc2477e3a87536faf084105b794b29c by filling up the unimplemented functions and editing some implemented functions.

The files:
* TRCAdapter.h & TRCAdapter.cpp

implement the class TRCAdapter which derives from FileAdapter.

[Review on Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/582)


aseth1:
Review status: 0 of 14 files reviewed at latest revision, 9 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Common/DataAdapter.h, line 24 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/582#-JxhmBtcYQxUGI8XQNBm)** ([raw file](https://github.com/opensim-org/opensim-core/blob/33e148e01934aafbcbea61724e8981036ae11960/OpenSim/Common/DataAdapter.h#L24)):</sup>
Our convention is to put the header guard at the top of the file. We also don't include the dll name (e.g. Common) and underscore separated/terminated words. E.g. OPENSIM_DATA_ADAPTER_H_

---

<sup>**[OpenSim/Common/DataAdapter.h, line 53 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/582#-JxhnUlNdvlKPa2hNYBR)** ([raw file](https://github.com/opensim-org/opensim-core/blob/33e148e01934aafbcbea61724e8981036ae11960/OpenSim/Common/DataAdapter.h#L53)):</sup>
need DLL export (i.e. OSIMCOMMON_API)

---

<sup>**[OpenSim/Common/DataAdapter.h, line 56 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/582#-Jxho4iVelEh_ONAn9Pp)** ([raw file](https://github.com/opensim-org/opensim-core/blob/33e148e01934aafbcbea61724e8981036ae11960/OpenSim/Common/DataAdapter.h#L56)):</sup>
Seems unnecessary to alias a type here for private data member.

---

<sup>**[OpenSim/Common/FileAdapter.h, line 1 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/582#-JxhvkbkPiGY3A2vBh5i)** ([raw file](https://github.com/opensim-org/opensim-core/blob/33e148e01934aafbcbea61724e8981036ae11960/OpenSim/Common/FileAdapter.h#L1)):</sup>
why this change?

---

<sup>**[OpenSim/Common/FileAdapter.h, line 11 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/582#-JxhvTAs5rHK4hQvaLAd)** ([raw file](https://github.com/opensim-org/opensim-core/blob/33e148e01934aafbcbea61724e8981036ae11960/OpenSim/Common/FileAdapter.h#L11)):</sup>
Add yourself as an author!

---

<sup>**[OpenSim/Common/FileAdapter.h, line 56 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/582#-JxhwZ7B2jdI1G457LLd)** ([raw file](https://github.com/opensim-org/opensim-core/blob/33e148e01934aafbcbea61724e8981036ae11960/OpenSim/Common/FileAdapter.h#L56)):</sup>
for better or worse OpenSim uses preceding _ for class data members. There is some use of m_ which would be preferable to a new convention.

---

<sup>**[OpenSim/Common/TRCAdapter.cpp, line 39 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/582#-JxhzTOlNKwefjCtckyi)** ([raw file](https://github.com/opensim-org/opensim-core/blob/33e148e01934aafbcbea61724e8981036ae11960/OpenSim/Common/TRCAdapter.cpp#L39)):</sup>
Shouldn't prepareForReading/Writing do this check and open the file the required mode?

---

<sup>**[OpenSim/Examples/DataAdapter/example.cpp, line 34 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/582#-Jxi2nZ_UeGFpJifNzDz)** ([raw file](https://github.com/opensim-org/opensim-core/blob/33e148e01934aafbcbea61724e8981036ae11960/OpenSim/Examples/DataAdapter/example.cpp#L34)):</sup>
Why hide that this is a TimeSeriesTable? We want TimeSeriestTable to be generally useful.

---

<sup>**[OpenSim/Examples/DataAdapter/example.cpp, line 48 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/582#-Jxi2cRO5mJS6O5PAK29)** ([raw file](https://github.com/opensim-org/opensim-core/blob/33e148e01934aafbcbea61724e8981036ae11960/OpenSim/Examples/DataAdapter/example.cpp#L48)):</sup>
Need more examples use cases and test cases. 
1. At a minimum verify that the TimeSeriesTable for MarkerData still generates the same answers during IK.
2. A great addition would be to save the model marker location from IK in a TimeSeriesTable and write it out as a .trc file. Also store the errors into a TimeSeriesTable and write out to a .csv (or .sto) file.
3. Demonstrate access to rows of the TimeSeriesTable by time and column by Marker name.
4. How do you add to a TimeSeriesTable?
5. Once I have a TimeSeriesTable can I append rows and columns? Should columns be limited to the same type and must have the same number of rows?
6. Can you demonstrate a conversion from a .csv to a .trc or vice-versa?
7. Can you show how you would access units? You show in the MarkersReference that you access units as metadata but is that file metadata or column-based metadata? Should we be able to handle column-wise units? If"no" and a Table has to be a homogeneous type with one definition of units for all the columns,  then how should we handle hetergeneous units as in the case of external force (e.g. forces in N and torques in N*m or N*mm and/or point of application in cm or mm)
8. What is lost if TimeSeriesTable is treated as a DataTable? Do we lose the time column? Looks like DataAdpapters only require the interface of AbstractDataTable.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/582)

chrisdembia:
<sup>**[OpenSim/Common/FileAdapter.h, line 56 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/582#-JxhwZ7B2jdI1G457LLd)** ([raw file](https://github.com/opensim-org/opensim-core/blob/33e148e01934aafbcbea61724e8981036ae11960/OpenSim/Common/FileAdapter.h#L56)):</sup>
#533.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/582)

OpenSim/Common/CSVAdapter.h
```diff
@@ -0,0 +1,161 @@
...
+    static std::string getIdentifier() {
+        return "csv";
+    }
+
+private:
+    Table* table_;
+    
+    static const std::string delimiter_read_;
+    static const std::string delimiter_write_;
+    static const std::string time_column_label_;
```
sherm1: Naming convention for class members (data and methods) is `camelCaseWithInitialLowerCase`; we use underscore separation only in property names and when we're trying to match STL, like in a container that has a `size_type` member. (This is separate from the data member prefix discussed in #533.) So conforming names for these could be:
```c++
static const std::string m_delimiterRead;
static const std::string m_delimiterWrite;
static const std::string m_timeColumnLabel;
```

PR #585, opened 2015-08-20
Start using ResetOnCopy mixin class, and remove NullOnCopyUniquePtr from OpenSim.

chrisdembia:


OpenSim/Simulation/Model/CoordinateLimitForce.h
```diff
@@ -235,8 +235,8 @@ OpenSim_DECLARE_CONCRETE_OBJECT(CoordinateLimitForce, Force);
     // damping to constant values beyond the limits. These are heap allocated
     // and owned here, but we don't want these functions to be copied if a
     // CoordinateLimitForce is copied.
-    SimTK::NullOnCopyUniquePtr<SimTK::Function::Step> _upStep;
-    SimTK::NullOnCopyUniquePtr<SimTK::Function::Step> _loStep;
+    SimTK::ResetOnCopy<std::unique_ptr<SimTK::Function::Step> > _upStep;
+    SimTK::ResetOnCopy<std::unique_ptr<SimTK::Function::Step> > _loStep;
```
sherm1: Yuck -- why does swig get to vote on syntax for data members? Do we have to support such an old swig?

OpenSim/Simulation/Model/Model.h
```diff
@@ -1032,16 +1032,22 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Model, ModelComponent);
...
+    // us to manage the heap memory for the handles.
+    SimTK::ResetOnCopy<std::unique_ptr<SimTK::MultibodySystem> > _system;
+
+    SimTK::ResetOnCopy<std::unique_ptr<SimTK::SimbodyMatterSubsystem> >
+        _matter;     
+    SimTK::ResetOnCopy<std::unique_ptr<SimTK::Force::Gravity> >
+        _gravityForce;
+    SimTK::ResetOnCopy<std::unique_ptr<SimTK::GeneralForceSubsystem> >
+        _forceSubsystem;
+    SimTK::ResetOnCopy<std::unique_ptr<SimTK::GeneralContactSubsystem> >
```
sherm1: Do these handles actually have to be heap allocated?

OpenSim/Simulation/Model/CoordinateLimitForce.h
chrisdembia: Just for a little longer. @aymanhab and I are discussing the move; we've both tried newer versions of swig and I'm currently grappling with a backwards incompatibility. However, I want this PR merged before we upgrade swig.

OpenSim/Simulation/Model/Model.h
chrisdembia: I don't think so. My hope with this PR was to *not* heap-allocate them. I tried that and got some test segfaults. So I thought I might do two passes: just get rid of ReferencePtr for now, and get rid of the heap memory later.

sherm1:
Reviewed. All looks good. I agree further changes should be in a separate PR.

sherm1:
There was a suspicious-looking failure in the Debug test run:
```
terminate called after throwing an instance of 'SimTK::Exception::Assert'
  what():  SimTK Exception thrown at SubsystemGuts.h:443:
  Internal bug detected: Subsystem::getOwnerSubsystemHandle()
  (Assertion 'm_myHandle' failed).
```

Could be related to this change if it is a new failure.

sherm1:
Sorry, forgot to mention that was testActuators.

chrisdembia:
Ready for review.

sherm1:
Looks great, Chris -- ready to merge IMO. Nice to see all the deleted stuff. Could still be cleaned up further but I think that's a separate project. Is this the last of the NullOnCopyUniquePtr use?

chrisdembia:
> Is this the last of the NullOnCopyUniquePtr use?

No; I haven't removed all misused ReferencePtr's. I know of only one more (Model::_modelViz), and I'm working on it.

I tried using simbody handles directly instead of unique_ptr's but I ran into the issue that SimbodyMatterSubsystem can't be copied.

sherm1:
> I tried using simbody handles directly instead of unique_ptr's but I ran into the issue that SimbodyMatterSubsystem can't be copied.

That's true, but neither can unique_ptrs be copied. `ResetOnCopy<T>` doesn't require T to be copyable.

sherm1:
Do you want this PR merged or do you have plans to add more to it?

chrisdembia:
> That's true, but neither can unique_ptrs be copied. ResetOnCopy<T> doesn't require T to be copyable.

Oh right...sorry I should have given more detail. The obvious way for me to use the handles directly was with copy assignment:

```cpp
_matter = SimbodyMatterSubystem(*_system);
```

With the smart pointers we use:

```cpp
_matter.reset(new SimbodyMatterSubsystem());
```

and avoid the copy.

I don't know how to avoid the copy assignment if we use handles directly.

> Do you want this PR merged or do you have plans to add more to it?

I'd like this merged. I'm making the other changes in different branches.

chrisdembia:
Thank you.

PR #459, opened 2015-05-24
Handful of small changes to python wrapping.

chrisdembia:
1. On Linux, the python wrapper is no longer optimized. This decreases the
memory requirements and build time a little.
2. Added __str__() methods to Vec and Vector, allowing someone to `print myvec`
in python to see the contents of a vector.
3. Created two iterators for Set's, allowing both of the following use cases:
```python
for force in model.getForceSet():
    print force.getName()
```
```python
for name, musc in model.getMuscles().items():
    print name, musc.get_max_isometric_force()
```
4. Allow python indexing into Vec3's and Vector's: `v = Vec3(1,2,3); v[0]`.
5. Allow the python length function on Vec3's and Vector: `len(v) == v.size()`.
6. Better exception handling. Previously, if OpenSim threw an exception, the
python session would crash. Now, it gracefully continues. This wraps **every**
function call in a try-catch block though.
7. Got rid of the swig warnings, largely by just silencing them.
8. Removed OpenSimContext for python wrapping, since it's not used.
9. I added tests corresponding to the changes I made.

The wrapping is slightly slower now, but I think that's fine.

I wanted to also add in support for the ComponentList iterator, but I realize
that we'll need some design discussion about it.

It's very important that we document these changes for our users. I know we've mentioned it but I think it's time to break off a new copy of confluence for 4.0 so that we can start writing new documentation.

/cc @aymanhab

PR #587, opened 2015-08-24
[WIP] Upgrade swig 3.0.5

aymanhab:
Thanks @chrisdembia for the heads up, I made the change to AnalysisSet (as ->analysisset) and now the failure occurs in test_consistent_with_jython.test_addMetabolicProbes, can you take a log at the travis output? Thanks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ env:
   global:
     # The python tests look for OPENSIM_HOME.
     - OPENSIM_HOME=~/opensim-install
-    - SWIG_VER=2.0.10
+    - SWIG_VER=3.0.5
     - USE_CCACHE=1
     - CCACHE_COMPRESS=1
     # for Clang to work with ccache.
--- a/OpenSim/Simulation/Model/AnalysisSet.h
+++ b/OpenSim/Simulation/Model/AnalysisSet.h
@@ -113,7 +113,7 @@ public:
     //--------------------------------------------------------------------------
     // UTILITY
     //--------------------------------------------------------------------------
-    static void getAvailableAnalyses(AnalysisSet& as);
+    static void getAvailableAnalyses(AnalysisSet& analysisset);
 
 //=============================================================================
 };  // END of class AnalysisSet
--- a/OpenSim/Simulation/Model/Model.h
+++ b/OpenSim/Simulation/Model/Model.h
@@ -462,15 +462,15 @@ public:
     /**
      * Add ModelComponents to the Model. Model takes ownership of the objects.
      */
-    void addModelComponent(ModelComponent* aModelComponent);
-    void addBody(Body *body);
-    void addJoint(Joint *joint);
-    void addConstraint(Constraint *constraint);
-    void addForce(Force *force);
-    void addProbe(Probe *probe);
-    void addContactGeometry(ContactGeometry *contactGeometry);
-    void addFrame(Frame* frame);
-    void addMarker(Marker *marker);
+    void addModelComponent(ModelComponent* adoptee);
+    void addBody(Body *adoptee);
+    void addJoint(Joint *adoptee);
+    void addConstraint(Constraint *adoptee);
+    void addForce(Force *adoptee);
+    void addProbe(Probe *adoptee);
+    void addContactGeometry(ContactGeometry *adoptee);
+    void addFrame(Frame* adoptee);
+    void addMarker(Marker *adoptee);
     /** remove passed in Probe from model **/
     void removeProbe(Probe *probe);
     //--------------------------------------------------------------------------
--- a/OpenSim/Simulation/Model/Model.h
+++ b/OpenSim/Simulation/Model/Model.h
@@ -817,9 +817,9 @@ public:
      *
      * @param analysis  pointer to the Analysis to add
      */
-    void addAnalysis(Analysis *analysis);
+    void addAnalysis(Analysis *adoptee);
     /** Add a Controller to the %Model. **/
-    void addController(Controller *aController);
+    void addController(Controller *adoptee);
 
     /**
      * Remove an Analysis from the %Model.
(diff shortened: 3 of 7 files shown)


chrisdembia:
Ah right...to get the tests to pass I had to make this change: https://github.com/opensim-org/opensim-core/pull/459/files#diff-452dbbd4833b667aef6a6e74a88c9ba5L307

@aymanhab I'd like to talk at some point about reviewing #459 and getting it merged. We were waiting on looking at #459 until the geometry PR was merged.

aymanhab:
@chrisdembia I have no problem with #459 other than documenting the resulting API changes caused by the new/improved python wrapping so that our examples in C++, Java/Matlab and Python do not diverge too much. I'm not sure though how's that affecting the current test failures on travis since we're not including any code from #459 yet.

aymanhab:
@chrisdembia latest build may have failed due to size of file exaggerated by the exception handling clauses. Have you seen this?

chrisdembia:
Yes I have. You should be able to remove the changes to exception handling. They are unrelated to upgrading swig.

aymanhab:
Build with wrapping now passes on travis/clang. Windows build succeeds but test fails though using swig 3.0.2

chrisdembia:
Okay great. I'll try to fix the windows 3.0.2 issue.

chrisdembia:
@aymanhab could you also update the README? The swig version is listed in about 3 different places there.

aymanhab:
@chrisdembia I sure can update the README. For AppVeyor you need to make sure you use the right Python distribution (x86 vs x64) since linking the wrong version leads to mysterious unresolved symbols. Also do we have documentation for how to run the "nosetests" in a development environment?

chrisdembia:
> For AppVeyor you need to make sure you use the right Python distribution (x86 vs x64) since linking the wrong version leads to mysterious unresolved symbols.

Are you saying you think this is why appveyor is failing? Hmm...it worked before so I don't know why this would be the cause.



> Also do we have documentation for how to run the "nosetests" in a development environment?

No, but #459 removes the need for using nosetests; I created a ctest that runs the python tests instead, which will make it much easier to run the python tests (from visual studio for example). For now, I'd just try to follow what the appveyor or travis script does.

> Also what version of Python, Simbody to suggest? we support Python 2.7.x but not 3.4.x AFAIK, also Simbody 3.6 or 4.0?

Yes, we support 2.7.x and not 3.4.x (just because we haven't tested it; it might work). We use simbody 3.6 for now, just like the rest of OpenSim.

chrisdembia:
By the way are you planning on testing the matlab stuff locally? Since we don't test it on travis/appveyor.

aymanhab:
Sure, I plan to test locally before I declare the stack of VS2015, SWIG 3.0.X, Matlb/Java/Python working well enough to proceed.

aymanhab:
I don't think the x86 and x64 difference to cause the failure, I just think we need to document it somewhere since the errors that show are mysterious.

chrisdembia:
It's documented (by me, and not well-enough) [on confluence](http://simtk-confluence.stanford.edu:8080/display/OpenSim/Scripting+in+Python#ScriptinginPython-Windows).

It would be a good idea to document the x86 vs x64 issue on the README.

aymanhab:
README updated but the failure will not get fixed. I'll run locally to debug.

chrisdembia:
Changes to README look good.

aymanhab:
@chrisdembia I have a RelWithDebugInfo build of the Python wrapping but can't run the tests using nose since it fails to load a DLL with no name, path or any diagnostics. Same happens if I try to invoke any opensim API call. install finished cleanly. Any clues?

chrisdembia:
Can you share any exception message you get? Did you set the PATH? What do you mean, invoking an API call? You're sure you're using 64bit python and 64bit opensim (or the other way around)?

aymanhab:
@chrisdembia No exception just a failed to load DLL. All my env. is 64 bit nd I did set the path. Here's a screen capture. 
capture

chrisdembia:
(that's a python exception ). I don't think I've *ever* seen that particular exception! Though that is usually the line of the python wrapping (opensim.py) where we get an error.

You can also go into the opensim installation's `sdk/python/` folder and try running nosetests from there (or just `import opensim` from a python shell in that dir).

aymanhab:
@chrisdembia I fixed the problem by rebuilding the whole stack again. The crash happens because the change we made to MODEL_ADOPT_HELPER is broken per
http://swig.10945.n7.nabble.com/How-to-use-pythonprepend-and-pythonappend-to-access-arguments-and-return-values-td14377.html
 I can manually change the .py or change the .i file to fill the add* methods using actual argument name. Any other suggestions/idea?

chrisdembia:
It's great you found the problem, and thanks for that link. I'm not sure I understand completely though. The change to MODEL_ADOPT_HELPER is what *fixed* the build for me. What does swig generate as the signature?  The first argument isn't `self`? You're using swig 3.0.5?

aymanhab:
Chris, the problem is that the method markAdopted needs to be called on the
second argument e.g body or joint not self which is model. The naming
convention changed going to 3.0.5

chrisdembia:
Ohhh I see. Then `self._markAdopted()` is definitely the wrong thing to do!. A second argument to the macro with the name of the argument would work, I think.

aymanhab:
@chrisdembia I fixed the issue by unifying the name of the argument in the signature/headers and now all tests pass but I'm not satisfied with this as a final solution. Can you think of a better alternative? Or we can drop the macro and surgically append to the less than a dozen methods in the .i file. Please let me know what you think. Thanks

chrisdembia:
Your solution seems preferable to copying and pasting. I think what you have is fine. If someone changes the name away from `adoptee`, the python tests should fail. Unless you have anything else planned for this PR, this is good to merge.

aymanhab:
Thanks @chrisdembia We can now address #459 and other issues related to wrapping.

PR #588, opened 2015-08-25
Search listprop byname

aymanhab:
Extend Property interface to allow for search by name in List-of-Objects properties.


OpenSim/Common/Property.h
```diff
@@ -529,6 +529,12 @@ class Property : public AbstractProperty {
         return -1;
     }
 
+    /** Return index of passed in name if the Property contains objects that are
+    derived from OpenSim::Object, and -1 if no such Object is found. Throws an 
+    Exception if the List doesn't contain OpenSim Objects (e.g. Primitive types)
+    since these are not named **/
```
sherm1: This comment should mention the time complexity -- it is useful information in some cases.
Also, there is a missing period.

chrisdembia: Could we also make Primitive lower case?

chrisdembia:
@aymanhab could you add some brief tests somewhere? Otherwise this looks good to me. One step closer to moving away from sets!

OpenSim/Simulation/Model/PhysicalFrame.cpp
```diff
@@ -88,9 +88,10 @@ void PhysicalFrame::extendConnectToModel(Model& aModel)
 {
     Super::extendConnectToModel(aModel);
 
-    // TODO: Better use name search or more robust method
-    if (upd_geometry(0).getFrameName() == "")
-        upd_geometry(0).setFrameName(getName());
+    // 
+    int idx = getProperty_geometry().findIndexForName("frame_geometry");
+    if (idx != -1 && upd_geometry(idx).getFrameName() == "")
+        upd_geometry(idx).setFrameName(getName());
```
sherm1: This seems like a bad hack to me. Is there some well-defined convention that a Frame's geometry list will have a piece of geometry on it whose name is literally `frame_geometry` unquote and that if so, that piece of geometry should have its frame name automatically set to the name of the current Frame object, unless someone set it somewhere else? And why is this in PhysicalFrame rather than Frame? The `geometry` property itself belongs to ModelComponent. If this code is correct I think it needs a very long comment explaining why. 

@aseth1 does this seem right to you?

sherm1:
Done reviewing. I like the lookup by name feature but I really don't like how it is being used in PhysicalFrame. There must be a better way!

OpenSim/Simulation/Model/PhysicalFrame.cpp
chrisdembia: @sherm1 I wasn't so concerned because I think @aseth1 is interested in having frames visualize themselves via `generateDecorations()`, and that `frame_geometry` and related code is temporary until that happens.

aymanhab: @sherm1 As it is today, yes there's a piece of geometry with the name "frame_geometry" that is used to visualize the PhysicalFrame. Whether this will stay or be moved to another class (e.g. Frame) or removed completely seems to be separate from this PR, it's basically how I tested that the API to lookup by name works since we have no other list properties that contain named objects.

chrisdembia: > it's basically how I tested that the API to lookup by name works since we have no other list properties that contain named objects.

I still think we should have independent tests in a test case.

aymanhab:
@chrisdembia testSerialization beefed up to test the search by name functionality.

chrisdembia:
Great! Feel free to merge?

sherm1:
> Great! Feel free to merge?

I'm hoping we'll hear from @aseth1 about the `force_geometry` part before merging it.

aymanhab:
Sherm, since I prefer not to entangle this small self conained change with
another  possibly long discussion, I'd rather undo the change to physical
frame and push the rest. Please let me know if you're ok with that. Thanks

sherm1:
> I'd rather undo the change to physical frame and push the rest. 

Good idea! In that case I'd be happy to see it merged. Maybe we can discuss frame geometry more on Thu?

aymanhab:
Sure, will do. Thanks.

aymanhab:
Thanks @sherm1 and @chrisdembia the change was merged.

PR #589, opened 2015-08-28
Joints connect to PhysicalFrames

aseth1:
Update to the Joint and its derived classes, which includes:

1. Connecting to any PhysicalFrame and not just Body (which is a PhysicalFrame)
2. Removal of parent and child Joint frames maintained by location and orientation properties.
3. Joint automatically creates PhysicalOffsetFrames from old Joint definitions (with location and orientation properties) but these constructors are now deprecated.
4. All derived Joints use the base constructors (thus deprecated constructors are isolated to the base Joint)
5. Using a non static initializer (`Joint::constructCoordinate()`) for derived Joints to specify their Coordinates and their MotionType.  The initializer automatically adds the Coordinate to the Joint's CoordinateSet (property).  This was required, because derived joints use the base constructor and thus we cannot have each joint construct its Coordinates in their own constructor. This was temporarily delayed until `finalizeFromProperties()` but that proved unsatisfactory (requiring `finalizeFromProperties to be called everywhere a Joint's coordinate was being accessed). Now only CustomJoint (which requires its SpatialTransform to be specified) has its own constructor and its own constructCoordinates() since the number of dofs cannot be known apriori.
6. Address the fallout from the above modifications, which include updating from older model file versions, different number of components now present in a Model, different drawing order, etc ...
7. I rerouted `Component::getComponent<C>(name)` through the old Component::findComponent() so I could find Frames before calling `Model::connect()` which is when the Component traversal tree gets built. This is kind of a mess, but I would like to address that as a completely separate PR.

How to review this PR (in order of priority):

1. Look at Joint.h and the subsequent concrete Joint classes. Are you happy with the Joint(s) API?
2. Verify that the Joint::updateFromXML (and helper code in XMLDocument) is implemented correctly. I had introduced bugs earlier by calling upon the API to help with the update. @chrisdembia correctly noted that this could be dangerous! Updates should be restricted to XML editing only. If you see otherwise please flag it!
3. `Model::extendFinalizeFromProperties()`'s Multibody tree building is based on bodies and I do this by tunneling through the OffsetFrames to the base. I think this right, but a sanity check here would be appreciated.
4. Finally the smaller accommodating changes and test case updates. 

Happy reviewing. Thanks!

[Review on Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/589)


OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -60,16 +60,14 @@ A Seth, M Sherman, P Eastman, S Delp; Nonlinear dynamics 62 (1), 291-303
 \code{.cpp}
     // Define a Pin joint between ground and platform.
```
chrisdembia: Can you add a note saying something like `The "platform" body must be added to the Model by the time you call `initSystem()`?

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -78,28 +76,12 @@ class OSIMSIMULATION_API Joint : public ModelComponent {
...
-        "reference frame. Default is (0,0,0).");
-
-    OpenSim_DECLARE_PROPERTY(orientation_in_child, SimTK::Vec3,
-        "Orientation of the joint in the child body specified in the child body "
-        "reference frame. Euler XYZ body-fixed rotation angles (in radians) "
-        "are used to express the orientation. Default is (0,0,0)" );
-
+    /** @name Property declarations
+    These are the serializable properties associated with a Joint. **/
+    /**@{**/
```
chrisdembia: Remove this block (and the corresponding end `/**@}**/`); they are no longer necessary.

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -131,30 +113,65 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
...
-                                       of the joint frame orientation in the
-                                       parent body frame.
-        @param[in] child    the child PhysicalFrame that joint connects to
-        @param[in] locationInChild     Vec3 of the location of the joint in the
-                                       child body frame.
-        @param[in] orientationInChild  Vec3 of the XYZ body-fixed Euler angles
-                                       of the joint frame orientation in the
-                                       child body frame.
+        @param[in] parentName   the name of the parent PhysicalFrame of the joint
+        @param[in] childName    the name of the child PhysicalFrame of the joint
```
chrisdembia: super nitpicky (sorry!) but "of" implies ownership; perhaps "for the joint"?

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -131,30 +113,65 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
...
         @param[in] reverse  Advanced optional flag (bool) specifying the 
                             direction of the Joint in the multibody tree. 
-                            Default is false (that is, forward).
+                            Default is false (that is, forward. parent to child).
```
chrisdembia: You could remove the word "forward"

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -131,30 +113,65 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
...
-          const PhysicalFrame& child,
-          const SimTK::Vec3& locationInChild,
-          const SimTK::Vec3& orientationInChild,
-          bool reverse = false);
+    Joint( const std::string& name,
+           const std::string& parentName,
+           const std::string& childName,
+           bool reverse = false);
+
+    /** DEPRECATED Convenience Constructor
```
chrisdembia: We should have a consistent convention; @sherm1 uses `<b>(Deprecated)</b>`: https://github.com/simbody/simbody/blob/master/SimTKcommon/include/SimTKcommon/internal/ClonePtr.h#L293

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -131,30 +113,65 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
...
+    Joint(const std::string& name,
+        const PhysicalFrame& parent,
+        const SimTK::Vec3& locationInParent,
+        const SimTK::Vec3& orientationInParent,
+        const PhysicalFrame& child,
+        const SimTK::Vec3& locationInChild,
+        const SimTK::Vec3& orientationInChild,
+        bool reverse);
+
+    DEPRECATED_14("Use Joint(name, parentName, childName) instead.")
```
chrisdembia: Doxygen for this?

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -131,30 +113,65 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
...
+    */
+    DEPRECATED_14("Use Joint(name, parentName, childName) instead.")
+    Joint(const std::string& name,
+        const PhysicalFrame& parent,
+        const SimTK::Vec3& locationInParent,
+        const SimTK::Vec3& orientationInParent,
+        const PhysicalFrame& child,
+        const SimTK::Vec3& locationInChild,
+        const SimTK::Vec3& orientationInChild,
+        bool reverse);
```
chrisdembia: I think these deprecated constructors should take `Body`, not `PhysicalFrame`. That's what the old constructors took. It sounds messy for people to be specifying their own custom PhysicalOffsetFrames and then applying *additional* offsets via the Vec3's.

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -267,27 +253,36 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
...
     virtual void scale(const ScaleSet& aScaleSet);
 
-    /**
-    * ModelComponent display interface.
-    *
-    * This method appends the visuals for the Joint to the list appendToThis.
-    * Base class adds geometry for the two Frames making the joint, any extra 
-    * gometry needed to visualize the Joint would be added by the subclass.
+protected:
+    /** A Coordinate place-holder created by constructCoordinate(). E.g.:
```
chrisdembia: Clarify that you are talking about a new joint class? Otherwise I thought you were suggesting the use of a nested class.

chrisdembia: This could also be an opportunity to have separate user and dev doxygen. I think the alias should still be documented for users, but the description of how to use `CoordinateP` in a joint subclass can be for developers.

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -164,61 +181,30 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
...
-    void setOrientationInParent(const SimTK::Vec3& aOrientation);
-    const SimTK::Vec3& getOrientationInParent() const;
-
-    /** Get the Joint frames expressed as body transforms. Only available after
-        extendConnectToModel() has been called on the Joint. */
-    const SimTK::Transform& getParentTransform() const
-        { return _jointFrameInParent; }
-    const SimTK::Transform& getChildTransform() const
-        { return _jointFrameInChild; }
-
```
chrisdembia: What should people do if they were using these methods? Two suggestions:

1. still implement these methods, have them throw an exception if called (like OpenSim::DeprecatedException, or OpenSim::NoLongerExists) and also add a DEPRECATED_14.
2. Add comments to the CHANGELOG explaining what people should do instead to achieve the same functionality.

OpenSim/Simulation/SimbodyEngine/PinJoint.cpp
```diff
@@ -26,63 +26,19 @@
...
-    const SimTK::Vec3& locationInchild, const SimTK::Vec3& orientationInChild,
-    bool reverse) : 
-        Super(name, parent, locationInParent, orientationInParent,
-            child, locationInchild, orientationInChild, reverse)
-{
-    setAuthors("Ajay Seth");
-    constructCoordinates();
-
-    const CoordinateSet& coordinateSet = get_CoordinateSet();
-    coordinateSet[0].setMotionType(Coordinate::Rotational);
```
chrisdembia: nice to see this go away.

OpenSim/Simulation/SimbodyEngine/PinJoint.h
```diff
@@ -42,35 +42,16 @@ are in the desired direction.
...
-private:
-    static const int _numMobilities = 1;
-//=============================================================================
-// DATA
-//=============================================================================
-protected:
-
-    /** Pin has no additional properties*/
+    /** Specify the Coordinate of the PinJoint */
+    CoordinateP rz{ constructCoordinate(Coordinate::MotionType::Rotational) };
```
chrisdembia: It's not clear to me whether these are supposed to be in public, private, or protected.

chrisdembia: Also I'm not sure if the doxygen comment makes sense to users.

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -267,27 +253,36 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
...
+    Mobilizer Qs unless those Qs are intended to be hidden. */
+    Coordinate* constructCoordinate(Coordinate::MotionType mt) {
+        Coordinate* coord = new Coordinate();
+        coord->setMotionType(mt);
+        std::stringstream name;
+        name << getName() << "_coord_" << get_CoordinateSet().getSize();
+        coord->setName(name.str());
+        // CoordinateSet takes ownership
+        upd_CoordinateSet().adoptAndAppend(coord);
+        return coord;
```
chrisdembia: Have you thought through and tested what happens when you copy construct, copy assign, etc. a joint? Are you sure the coordinate set doesn't double in size or something silly like that? It seems like you might lose your custom names for the joints. That is, does the following test pass or fail?

```cpp
auto jointA = new PinJoint(...);
jointA->setName("jointA");
jointA->upd_CoordinateSet()[0].setName("flexion");
PinJoint jointB(*jointA);
assert(jointB.get_CoordinateSet()[0].getName() == "flexion");
```

OpenSim/Simulation/SimbodyEngine/PinJoint.h
```diff
@@ -42,35 +42,16 @@ are in the desired direction.
...
-    // Convenience constructor
-    PinJoint(const std::string &name,  const PhysicalFrame& parent,
-             const SimTK::Vec3& locationInParent, const SimTK::Vec3& orientationInParent,
-             const PhysicalFrame& child,
-             const SimTK::Vec3& locationInchild, const SimTK::Vec3& orientationInChild,
-             bool reverse = false);
-    virtual ~PinJoint();
-
-    int numCoordinates() const override {return _numMobilities;};
+    /** Use Joint's constructors. @see Joint */
```
chrisdembia: Does this look okay if you look at the generated Doxygen html? I think usually the `@see` goes on its own line.

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -267,27 +253,36 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
...
-    *
-    * This method appends the visuals for the Joint to the list appendToThis.
-    * Base class adds geometry for the two Frames making the joint, any extra 
-    * gometry needed to visualize the Joint would be added by the subclass.
+protected:
+    /** A Coordinate place-holder created by constructCoordinate(). E.g.:  
+    class My2DofJoint::Joint {
+        CoordinateP dof1{ constructCoordinate(Coordinate::MotionType::Rotational) };
+        CoordinateP dof2{ constructCoordinate(Coordinate::MotionType::Translatinal) };
+        ...
```
chrisdembia: Are these guaranteed to construct in the order you list? I think this is worth mentioning in the comment. This could be another reason to use an initializer list.

OpenSim/Simulation/SimbodyEngine/SliderJoint.h
```diff
@@ -34,46 +34,25 @@ namespace OpenSim {
 /**
 A class implementing a Slider joint.  A class implementing a Slider joint.  The
 underlying implementation in Simbody is a SimTK::MobilizedBody::Slider.
-Slider provides a single mobility along the common X-axis of the joint frames
+Slider provides a single corodinate along the common X-axis of the joint frames
```
chrisdembia: typo

sherm1:
Review status: 0 of 51 files reviewed at latest revision, 15 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 39 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3TRPft6umKxEWflr_)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L39)):</sup>
"(one on each of two bodies)"?

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 48 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3Tu0qBotE5g-vphEy)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L48)):</sup>
*has* a coordinate value ...

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 97 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3US_oRW4ZjkV0m4ku)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L97)):</sup>
I understand why this is here, but this might be a case for using a generic "components" property inherited from Component. Then it wouldn't have to appear in the Joint API.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 103 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3VG6C3pnawtI6p_U0)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L103)):</sup>
This comment is not helpful. Would be good to say what the default constructor does. Should it be protected and only allowed for concrete Joints?

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 108 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3Vk3o7LbJHw9mjmuF)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L108)):</sup>
Doesn't reversal get handled automatically now because of using MultibodyGraphMaker? It is only needed if users are constructing the multibody tree directly. They should simply be able to build the joint with a parent and child and always get parent->child sense for the coordinates, even if MGM decided to reverse the joint.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 127 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAk--r1-127)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L127)):</sup>
I did it that way after some experimentation determined that that particular syntax is the only way to get it to come out bold in doxygen both in the quick summary and detailed text.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 155 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3XWO7kVIZBG3oV1sN)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L155)):</sup>
If this is deprecated, what is the preferred way for a user to create a Joint component that owns its two frame? Is that what the default constructor is for? Something like:
```c++
Joint myJoint("joint1", "parent", "child");
myJoint.append_component(new PhysicalOffsetFrame("parent", ...));
myjoint.append_component(new PhysicalOffsetFrame("child", ...));
```
does that work?

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 163 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAk--r1-163)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L163)):</sup>
I think a Body is a PhysicalFrame so this constructor ought to work with bodies as it did before [is that right, Ajay?], and in addition would work with some offset frame. So it seems better this way to me.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

OpenSim/Simulation/SimbodyEngine/Joint.cpp
```diff
@@ -56,29 +57,69 @@ Joint::Joint() : Super()
...
+ 
+
+    // Append the offset frames to the Joints internal list of frames
+    append_frames(parentOffset);
+    append_frames(childOffset);
+
+    updConnector<PhysicalFrame>("parent_frame").set_connectee_name(
+        getName() + "/" + parentOffset.getName() );
+    updConnector<PhysicalFrame>("child_frame").set_connectee_name(
+        getName() + "/" + childOffset.getName() );
```
chrisdembia: Sorry to pollute this PR with an unrelated comment but I think absolute paths *must* begin with a `/`. Otherwise this ambiguous: do you mean `/jointA/flexion` or `/jointA/jointA/flexion`?

chrisdembia:
Oh darnit. I didn't realize @sherm1 was using reviewable. Hmm... I'll continue there later today.

chrisdembia:
<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 163 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAk--r1-163)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L163)):</sup>
I know the constructor will still work with a Body but I don't think we *want* it to work with PhysicalFrames. Such a constructor didn't exist in 3.2 and so it doesn't make sense to "deprecate" it.  I'd rather them get an error, especially since it leads to more complex behavior that we shouldn't have to anticipate. We're trying to get people *not* to use that constructor.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

chrisdembia:
<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 103 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3VG6C3pnawtI6p_U0)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L103)):</sup>
Oh interesting...aren't constructors in abstract classes already kinda reserved for subclasses since you can't construct an abstract type? Could `Joint()` be called anywhere other than a constructor of a derived type?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

chrisdembia:
<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 127 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAk--r1-127)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L127)):</sup>
I like the bold syntax and I think we should use it.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

sherm1:
Review status: 0 of 51 files reviewed at latest revision, 27 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 254 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3Zm2Zohgr_xPhzUJl)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L254)):</sup>
Does this apply only to physical frames that are owned by the Joint component? Should this kind of scaling be delegated to the physical frames rather than the joint?

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 258 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3_HTS3al1noNvfSWd)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L258)):</sup>
Missing `@code/@endcode` here for doxygen?

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 274 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3aGfKg-40yNwWwuTB)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L274)):</sup>
In C++11 you can replace these three lines with
```c++
coord->setName(getName() + "_coord_" + std::to_string(get_CoordinateSet().getSize()));
```

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 331 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3bMfhNcAaUqndDmnD)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L331)):</sup>
again I'm confused by the interaction between MultibodyGraphMaker and this flag

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 335 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3b3-AbYMqUNjZnuAC)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L335)):</sup>
`std::swap(inbX,outbX);` would be more clear

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

sherm1:
Review status: 0 of 51 files reviewed at latest revision, 28 unresolved discussions, some commit checks broke.

---

<sup>**[Applications/Scale/test/testScale.cpp, line 36 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3c6_2OG7vstf_x-uj)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/Applications/Scale/test/testScale.cpp#L36)):</sup>
Did you mean for ligament scaling to be included in this PR?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

sherm1:
Review status: 0 of 51 files reviewed at latest revision, 28 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 261 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAk--r1-261)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L261)):</sup>
> Are these guaranteed to construct in the order you list? 

If they are data members then yes.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

sherm1:
Review status: 0 of 51 files reviewed at latest revision, 29 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/BallJoint.h, line 57 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3ddnHA82qJaEAF4Yw)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/BallJoint.h#L57)):</sup>
I'm confused about this convention. I think I need a whiteboard explanation when you have a chance. Shouldn't this be something defined in the base Joint class that is filled in by the derived class? Here it looks like each defined Joint makes up its own data member names but then I presume that has to get registered with the base class eventually. 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

sherm1:
Review status: 0 of 51 files reviewed at latest revision, 30 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Common/Component.h, line 355 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3fToqVi2eDD75Z73C)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Common/Component.h#L355)):</sup>
It is an error if a reference is every returned null, so `compC` is null only because a component existed but was not of type `C`. Then the code below goes looking for another component of the same name of type `C`. Is that what you intended? If `getComponent()` every returns a null reference it should be redesigned not to do so since that is illegal in C++.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

sherm1:
Review status: 0 of 51 files reviewed at latest revision, 30 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/BallJoint.h, line 57 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3ddnHA82qJaEAF4Yw)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/BallJoint.h#L57)):</sup>
Like maybe the base class should have a constructor that takes an initializer list of coordinates, so the ball joint constructor would be
```c++
BallJoint : Joint{Coordinate::MotionType::Rotational,
                  Coordinate::MotionType::Rotational, 
                  Coordinate::MotionType::Rotational} {}
```
or something like that?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

sherm1:
> Oh darnit. I didn't realize @sherm1 was using reviewable. 

I tend to use it for gigantic PRs to avoid email flooding.

chrisdembia:
> I tend to use it for gigantic PRs to avoid email flooding.

Good idea.

sherm1:
Review status: 0 of 51 files reviewed at latest revision, 31 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.cpp, line 117 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3jlnvi3omceK40pz2)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.cpp#L117)):</sup>
Could use `append_component()` here if there is a generic property for that purpose. Not sure if that's better but it might be since people might be surprised to find that the Joint's `frames` property can be empty if the Joint Component doesn't own its frames.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.cpp, line 122 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAjz-r1-122)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.cpp#L122)):</sup>
I'm confused about the scope here too, although the physical frame component should be local, not absolute since it is relative to this Joint. Otherwise it wouldn't copy correctly. But, shouldn't it be just `set_connectee_name(parentOffset.getName()` to get it interpreted relative to the current component?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

chrisdembia:
<sup>**[OpenSim/Simulation/SimbodyEngine/BallJoint.h, line 0 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3ddnHA82qJaEAF4Yw)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/BallJoint.h#L0)):</sup>
Do you still get to inherit the constructors from Joint (`using Joint::Joint;`)?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

sherm1:
Review status: 0 of 51 files reviewed at latest revision, 33 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Simulation/Model/Model.cpp, line 521 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3nyelc_wKDrZ6rHSR)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/Model/Model.cpp#L521)):</sup>
Ajay, if you are using Euler angle coordinates for Ball joints you might not want MGM to replace them with loop joints. It makes sense to do that when Balls are using quaternions because the loop Ball does the same thing. But the loop joint doesn't have axes.

---

<sup>**[OpenSim/Simulation/Model/Model.cpp, line 581 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3pJmyBStSQe4CIilz)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/Model/Model.cpp#L581)):</sup>
The same condition is tested twice here. empty() methods length==0.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

sherm1:
Done reviewing. The PR is so big I had to do some random sampling -- sorry! I think I covered the specific items you listed except for the XML conversion. Please let me know if you have something specific you want me to look at.

The main comment I have about the Joint API is that the "add data members for coordinates" convention seems odd and I suspect this could be cleaned up using a base class constructor designed for the purpose.

sherm1:
Review status: 0 of 51 files reviewed at latest revision, 33 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/BallJoint.h, line 57 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3ddnHA82qJaEAF4Yw)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/BallJoint.h#L57)):</sup>
> Do you still get to inherit the constructors from Joint (using Joint::Joint;)?

You still can but I don't think you would want to because of the need to provide specific coordinate information. If that has to be done on construction (and the data member initializers are still part of construction) then this seems more direct.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

aseth1:
Thanks Sherm. You touched on the areas of main concern. The "add data members for coordinates" strategy allows us to not implement the deprecated constructors on all the derived classes. That is a big improvement for maintainability. I am open to any alternative that doesn't require implementing the deprecated interface for each Joint type.

One option that might be nicer, is to have constructCoordinate return the local index (in the CoordinateSet) and store that as the data member instead of a reference. As Chris pointed out, copying resets those references (to nullptr), so they are not valid after a copy. I did not intend them to be used, but I realize that they should be usable as data members, so if the data member was an index into the CoordinateSet it would be valid for the copy and accessing the local set with that index would also be correct/usable interface for faster index access to a Coordinate.  

---

Review status: 0 of 51 files reviewed at latest revision, 26 unresolved discussions, some commit checks broke.

---

<sup>**[Applications/Scale/test/testScale.cpp, line 36 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3c6_2OG7vstf_x-uj)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/Applications/Scale/test/testScale.cpp#L36)):</sup>
I wasn't sure why that test case was failing while the others were passing. Since it is a Ligament specific test I thought I would check the Ligaments individually. This helped me discover an issue. I could unwind the comparison of Ligaments it so that it just compares the models (like it does at the end).

---

<sup>**[OpenSim/Common/Component.h, line 0 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3fToqVi2eDD75Z73C)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Common/Component.h#L0)):</sup>
Yes, that is my intention. Get component throws an Exception if it does not find the component by name.

---

<sup>**[OpenSim/Simulation/Model/Model.cpp, line 0 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3nyelc_wKDrZ6rHSR)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/Model/Model.cpp#L0)):</sup>
Good point. I would like to make BallJoint and FreeJoint always use quaternions underneath but expose Euler angles to the user but that won't happen for awhile. I will not treat BallJoint as a possible loop joints for the time being. 

---

<sup>**[OpenSim/Simulation/SimbodyEngine/BallJoint.h, line 0 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3ddnHA82qJaEAF4Yw)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/BallJoint.h#L0)):</sup>
I like the idea of the constructor that takes an initializer list, but how would I support the Deprecated constructors? The main benefit was that I was able to isolate the deprecated method to the base Joint and thereby cleanup all the derived Joints. I agree it is a bit strange that the data member of the derived class is being used primarily for its initializer to add coordinates into the Joint's coordinate set.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.cpp, line 0 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAjz-r1-122)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.cpp#L0)):</sup>
You are correct. I did this to ensure we find the local offset frame. Right now the search starts with the root and therefore finds another frame with the same name.  I agree we should always try to satisfy connectors by looking at its Component's subcomponents and then traverse from the root. The component traversal issue needs to be fixed and that will be another PR.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.cpp, line 0 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3jlnvi3omceK40pz2)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.cpp#L0)):</sup>
Since the generic component property list does not yet exist, yet, I added frames to Joint. The idea to use the generic list and filter for meaningful sublists sounds very appealing to me, but we don't have the capability working yet. I think this will have to be done piecemeal. 

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 103 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3VG6C3pnawtI6p_U0)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L103)):</sup>
I agree the comment should be more helpful. Joint is abstract so it cannot be constructed directly.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 108 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3Vk3o7LbJHw9mjmuF)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L108)):</sup>
It behaves as you suggest Sherm and the flag is set appropriately by the MGM. The only reason to leave it exposed is to allow the user to specify their own tree preference, in which case it acts as an initial guess for MGM. Overall, I do think it is confusing and it could be easily eliminated. Unless there are cases where having the option to set the reversal yourself is useful. If you have a standing model (like Clay's first model), the MGM will arbitrarily chose the right or left leg to reverse (to go foot up) up to the pelvis and down to the other leg. But, you may want both legs to be from the ground up and use a hip joint or the pelvis itself as a constraint. In this case, user influence by be necessary.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 127 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAk--r1-127)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L127)):</sup>
Will do.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 155 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3XWO7kVIZBG3oV1sN)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L155)):</sup>
Yes, that should work. I will add a case.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 163 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAk--r1-163)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L163)):</sup>
You are probably right that not being able to use PhysicalFrames (restricted to Body only) might provide more impetus for users to change. Also some unplanned cases where you are using the deprecated method to add and offset to another PhysicalOffset. It will workm but you are making a case that it shouldn't. Otherwise, I don't think it matters if it is a new constructor (PhysicalFrame) or identically the old constructor (Body) as long as the users' code doesn't have to change.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 254 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3Zm2Zohgr_xPhzUJl)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L254)):</sup>
The joint used to own the offset and it scaled them according to the scale factors of the underlying Body. This behavior was preserved until we sort out how scaling should work. 

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 257 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAk--r1-257)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L257)):</sup>
I thought user doxygen was restricted to the public interface? I will test that. In that case this would be developer doxygen by virtue of being protected. Dhould the user know about the internal alias for creating coordinates? I will make it more clear that it is for creating new Joint types, but it should remain only for developers since it is protected.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 274 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3aGfKg-40yNwWwuTB)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L274)):</sup>
Nice. Thanks!

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 277 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAk--r1-277)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L277)):</sup>
The clone test in testComponents verifies that clone and assign are identical, but I will add a similar case to testJoints for the copy constructor.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 331 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3bMfhNcAaUqndDmnD)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L331)):</sup>
It is the MGM that has the last word on the reverse flag.  See testJoints::testAutomaticJointReversal() where I am testing that the model reversed the joints automatically. This is just how the Joint implements the reversal when creating the underlying MobilizedBody (performed in `extendAddToSystem()`).

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 335 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3b3-AbYMqUNjZnuAC)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L335)):</sup>
Nice. Will do.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/PinJoint.h, line 0 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAk1-r1-50)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/PinJoint.h#L0)):</sup>
They are private. The doxygen will not be available for users. I think the information about the number of dofs and their interpretation should be in the class description. I can eliminate doxygen style comments.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/PinJoint.h, line 0 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAk1-r1-53)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/PinJoint.h#L0)):</sup>
Does not look bad to me.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

sherm1:
Review status: 0 of 51 files reviewed at latest revision, 25 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 108 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3Vk3o7LbJHw9mjmuF)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L108)):</sup>
> The only reason to leave it exposed is to allow the user to specify their own tree preference, in which case it acts as an initial guess for MGM.

I think MGM supports only two preferences: "must be base body" (for bodies) and "must be loop joint" (for joints). I don't think it supports "reverse" as a preference, but I believe those other two flags are sufficient to cover all reasonable user preferences -- exposing them as advanced hints in the opensim model would be a good idea at some point. I think they would adequately cover the standing model example. "Reverse" is supposed to be invisible to users. (I'm not sure but I think that example might work as you described anyway if the user builds the system up from the ground rather than down from the pelvis -- that does affect MGM's strategy.)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

sherm1:
Review status: 0 of 51 files reviewed at latest revision, 25 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.cpp, line 122 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAjz-r1-122)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.cpp#L122)):</sup>
> I agree we should always try to satisfy connectors by looking at its Component's subcomponents and then traverse from the root.

Shouldn't names work like file system names do? I'm trying to imagine how I build a subassembly as a component. If I have local subscomponents that are connected internally, I would need to be able to specify names like "../../xyz". Otherwise the subassembly won't copy properly. Alternatively, it could contain root-relative names where the root is the subassembly root, but then those names must be modified whenever the subassembly is adopted by a higher-level component.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

chrisdembia:
<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.cpp, line 122 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAjz-r1-122)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.cpp#L122)):</sup>
 I like being able to use "../../xyz". Relative paths are usually more useful.

>  but then those names must be modified whenever the subassembly is adopted by a higher-level component.

Oh good point...the renaming can be managed by a "mount" routine.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

sherm1:
Review status: 0 of 51 files reviewed at latest revision, 25 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.cpp, line 122 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAjz-r1-122)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.cpp#L122)):</sup>
> I like being able to use "../../xyz". Relative paths are usually more useful.

Also something we should discuss -- the naming system should be built from the tree of Objects (if it isn't already), *not* from just Components. For example, from our last discussion we want Appearance to be an object (so it can serialize and be referenced by name), but we don't want it to be a Component. That's easy if the naming "file system" is based on Objects.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

aseth1:
Review status: 0 of 51 files reviewed at latest revision, 23 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.cpp, line 0 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAjz-r1-122)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.cpp#L0)):</sup>
I agree we should come up with a robust solution, but that is beyond the scope of this PR.  There a several issues including caching the pathName with the Component that I don't like as well.
My intent here was to just reproduce the current behavior with Frames for the offsets.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 63 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAk--r1-63)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L63)):</sup>
I should add it and make something like what Sherm suggested work.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 84 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAk--r1-84)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L84)):</sup>
Great. Will do.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 95 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAk--r1-95)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L95)):</sup>
I will remove and add more details to property description.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 97 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3US_oRW4ZjkV0m4ku)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L97)):</sup>
I agree. Do you mind if we do that as another PR?

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 108 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3Vk3o7LbJHw9mjmuF)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L108)):</sup>
I am happy to get rid of it now. It is a vestige of a time before we had MGM and I think that the `testAutomaticJointReversal()` provides the verification that the MGM reverses the joints appropriately.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

sherm1:
Review status: 0 of 51 files reviewed at latest revision, 23 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 97 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3US_oRW4ZjkV0m4ku)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L97)):</sup>
> [components property] Do you mind if we do that as another PR?

Not at all! This PR is plenty big already.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

aymanhab:
Review status: 0 of 51 files reviewed at latest revision, 25 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 39 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy8v_UK7dI7j5llP3FG)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L39)):</sup>
Since the term "Frame" is already used, can we qualify this as physical frames?

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 90 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy8wkIKVMJwYf9LWNBF)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L90)):</sup>
Can we use 'is' instead of == since this's not code

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 127 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAk--r1-127)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L127)):</sup>
If name is jointName, can we change it since there're three names already in the signature?

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 163 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAk--r1-163)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L163)):</sup>
I'd agree with Chris to make deprectaed methods as narrowly useful as possible.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

aseth1:
Review status: 0 of 51 files reviewed at latest revision, 25 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 0 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy8v_UK7dI7j5llP3FG)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L0)):</sup>
I thought in the general description I would use frames (lowercase) so it is easy to grasp what a Joint is and does as an overview. In the method descriptions describe that these are of type PhysicalFrame. 

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 0 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy8wkIKVMJwYf9LWNBF)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L0)):</sup>
Good point. But I think we will be removing the "reverse" property altogether. There will be an internal flag that the MGM can set to direct the Joint appropriately. 

---

<sup>**[OpenSim/Simulation/SimbodyEngine/Joint.h, line 0 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/589#-Jy3SxMcDvvcvyBdLAk--r1-163)** ([raw file](https://github.com/opensim-org/opensim-core/blob/bce3f379735a9191881e9a3243abefceee979f1a/OpenSim/Simulation/SimbodyEngine/Joint.h#L0)):</sup>
I tending to agree.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

aseth1:
I have a version where constructCoordinate returns the local index as the member variable initializer. The index is consistent across instances of the same Joint type so that default copying the indices is correct. It is then valid to access the Coordinate from the Joint's CoordinateSet (property, which should become a list) using that index. What do you think? We can discuss this afternoon if you have some time, if this is unclear. @sherm1 @chrisdembia @aymanhab

---

Review status: 0 of 51 files reviewed at latest revision, 25 unresolved discussions, some commit checks broke.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

aseth1:
@aymanhab can you take a look at the implementations of `Joint::updateFromXMLNode()` and the static helper method `XMLDocument::addPhysicalOffsetFrame()` that was included  (similar to your `addConnector`) for the purpose of reading older model versions. I don't think that has had close scrutiny. Thanks!

---

Review status: 0 of 51 files reviewed at latest revision, 25 unresolved discussions, some commit checks broke.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/589)

OpenSim/Common/XMLDocument.cpp
```diff
@@ -423,4 +425,37 @@ void XMLDocument::addConnector(SimTK::Xml::Element& element, const std::string&
...
+    const SimTK::Vec3& location, const SimTK::Vec3& orientation)
+{
+    SimTK::Xml::element_iterator  frames_node = element.element_begin("frames");
+    SimTK::String debug;
+    if (frames_node == element.element_end()) {
+        SimTK::Xml::Element framesElement("frames");
+        element.insertNodeBefore(element.element_begin(), framesElement);
+        frames_node = element.element_begin("frames");
+    }
+    // Here we're guaranteed connectors node exist, add individual connector
```
aymanhab: The comment should say frames for future development. Not doxygen-ated anyway.

OpenSim/Common/XMLDocument.cpp
```diff
@@ -423,4 +425,37 @@ void XMLDocument::addConnector(SimTK::Xml::Element& element, const std::string&
...
+    SimTK::Xml::Element translationElement("translation", transValue.str());
+    newFrameElement.insertNodeAfter(newFrameElement.element_end(), translationElement);
+
+    std::ostringstream orientValue; 
+    orientValue << orientation[0] << " " << orientation[1] << " " << orientation[2];
+    SimTK::Xml::Element orientationElement("orientation", orientValue.str());
+    newFrameElement.insertNodeAfter(newFrameElement.element_end(), orientationElement);
+
+    frames_node->insertNodeAfter(frames_node->element_end(), newFrameElement);
+    frames_node->writeToString(debug);
```
aymanhab: We should probably remove or comment out the debug lines from this and other places.

aymanhab:
@aseth1 Is there an example .osim file in this PR that shows the new XML for a builttin simple Joint and for a custom joint?, I tried searching the file changes and couldn't find any. Thanks.

sherm1:
> I have a version where constructCoordinate returns the local index as the member variable initializer. ... We can discuss this afternoon if you have some time, if this is unclear.

I'm fuzzy on this. Could meet this afternoon any time.

aymanhab:
@aseth A general high level question that I kept asking myself is whether Coordinate is really an independent ModelComponent vs. an internal object used for book-keeping by a ModelComponent (Joint) and as such should be only narrowly exposed. Am I completely off-track on this? Thanks!

aseth1:
> Coordinate is really an independent ModelComponent vs. an internal object used for book-keeping by a ModelComponent ...

It's a good question. You are right that Coordinate does do the bookeeping/management of the q's and u's that arise from the underlying MobilizedBody of the Joint. But it also serves other purposes: 1) It is as the main input/output mechanism for coordinate values and their speeds and 2) It implements prescribe motion and locking constraints (so requires further resources beyond the Joint) and will likely use joint stops just as soon as Sherm has it ready :grinning: . In that regard I think it stands as a useful ModelComponent. If Coordinates were not exposed how would you envision the GUI interacting with the Model? Would we rethink coordinate sliders?

chrisdembia:
The travis-ci/pr build fails because we turn warnings into errors and there's a deprecation warning related to Joint's constructor. Until recently, on travis we turned off the deprecating warnings (#595). If your intent is to test the deprecated constructor I'd use `#pragma`'s to turn off the warning in the specific file you're in, or set a compiler flag (maybe `-Wno-error=deprecated-declarations`) for that file in CMake.

chrisdembia:
As for the travis-ci/push swig failure  `/home/travis/build/opensim-org/opensim-core/OpenSim/Simulation/SimbodyEngine/Joint.h:290: Error: Macro 'SimTK_DEFINE_AND_EXPORT_UNIQUE_LOCAL_INDEX_TYPE' expects 4 arguments`, you should surround that macro call with a `#ifndef SWIG`.

aseth1:
@sherm1
 
1. I believe  concerns about the Coordinate initializers (with CoordinateIndex as the class member) were addressed in our meeting early last week.
2. a test case was added to verify that you can construct a joint with current constructor e.g. `Joint(name, parentName, childName)` and then append frames to the Joint with the PhysicalOffsetFrames constructed inline.
3. After some debate/discussion I think we decided not to deprecate the old style constructors that add offsets for you given locations and orientations of the parent and child frames. The comment still says it is deprecated although i removed DEPRECATED_14.
 
@chrisdembia  

1. If the convenience constructor is no longer deprecated then I think it should take `PhysicalFrame` and not `Body`. It still remains backward compatible.
2. Test to verify default copy constructor and assignment operator was added. Test component verifies clone().

@aymanhab 

1. your recommended changes to updateFromXML and helpers in XMLDocument were made.
2. helpers for SWIG wrapping inherited constructors was also added and python wrapping test is currently passing by using those constructors (thanks to pointers from @chrisdembia).

Related Issues (for a future PR):

1. connectors to local frames should not have to qualify with joint name. This is because we currently search from the root. We should only look where the connector indicates: that is the name should be a pathname. If no name is given then the dependent component must be a subcomponent. #610
2. explicit control of Joint reversal is no longer necessary. The MultibodyGraphMaker is capable of deciding which joints should be reversed mostly likely better than the user can. See #611
3. potential inconsistencies when finalizeFromProperties is being invoked by some Components (e.g. Model) that should be made consistent. Related to #606
4. If Components will have a property list of Components (for its subcomponents)  it will not be necessary for Joint to have a `frames` list. We could use Joint as an example for converting separate properties for Component sets and demonstrate iterator filtering on type to provide access to subsets of Components without maintaining different properties (e.g. Sets)  for each.

I would be grateful if you could confirm that that your issues have been adequately addressed either directly of in the list of issues (for an upcoming PR).

Many thanks!

OpenSim/Common/XMLDocument.cpp
```diff
@@ -404,23 +404,60 @@ bool XMLDocument::isElementEqual(SimTK::Xml::Element& elt1, SimTK::Xml::Element&
...
+    //SimTK::String debug; //Only used for debugging
+    
+    if (frames_node == element.element_end()) {
+        SimTK::Xml::Element framesElement("frames");
+        element.insertNodeBefore(element.element_begin(), framesElement);
+        frames_node = element.element_begin("frames");
+    }
+    // Here we're guaranteed frames node exists, add individual frame
+    SimTK::Xml::Element newFrameElement("PhysicalOffsetFrame");
+    newFrameElement.setAttributeValue("name", frameName);
```
sherm1: This looks like it would allow multiple frames of the same name in the "frames" element. Is that OK?

OpenSim/Simulation/Model/PhysicalOffsetFrame.h
```diff
@@ -53,28 +53,15 @@ class OSIMSIMULATION_API PhysicalOffsetFrame : public OffsetFrame<PhysicalFrame>
...
-    PhysicalOffsetFrame(const PhysicalFrame& parent,
-                        const SimTK::Transform& offset);
+    ~PhysicalOffsetFrame() final {}
 
 protected:
     /** Extend Component interface for adding the PhysicalOffsetFrame to the 
         underlying multibody system */
-    void extendAddToSystem(SimTK::MultibodySystem& system) const override;
-
+    void extendAddToSystem(SimTK::MultibodySystem& system) const override final;
```
sherm1: Why do you want `final` here and for the destructor? That would prevent anyone from deriving a class from PhysicalOffsetFrame, even one that just adds a convenience constructor of some sort.

OpenSim/Simulation/SimbodyEngine/BallJoint.h
```diff
@@ -49,38 +49,22 @@ of position (\f$\vec{u} \neq \dot{\vec{q}}\f$).
...
 
-private:
-    static const int _numMobilities = 3;
-//=============================================================================
-// DATA
-//=============================================================================
+    /** Specify the Coordinates of the BallJoint */
+    CoordinateIndex rx{ constructCoordinate(Coordinate::MotionType::Rotational) };
+    CoordinateIndex ry{ constructCoordinate(Coordinate::MotionType::Rotational) };
+    CoordinateIndex rz{ constructCoordinate(Coordinate::MotionType::Rotational) };
```
sherm1: Chris pointed out yesterday that side-effect initializers like this have an impact on backwards-compatibility of copy construction (not saying its a problem here; it just seems like a good time to mention it). A default copy constructor is fine since it won't invoke the initializers. A fully-implemented copy constructor will work since the author will have said what to do with each member. But a copy constructor that only mentions some of the members would have expected default construction of the unmentioned members. Now instead those members will execute the initialization code, possibly resulting in unwanted side effects.

OpenSim/Simulation/SimbodyEngine/Joint.cpp
```diff
@@ -56,34 +57,70 @@ Joint::Joint() : Super()
...
+    setName(name);
     set_reverse(reverse);
 
-    updConnector<PhysicalFrame>("parent_body").set_connectee_name(parent.getName());
-    updConnector<PhysicalFrame>("child_body").set_connectee_name(child.getName());
+    updConnector<PhysicalFrame>("parent_frame").set_connectee_name(parentName);
+    updConnector<PhysicalFrame>("child_frame").set_connectee_name(childName);
+}
+
+/* Deprecated Constructor*/
```
sherm1: This seems wimpy. If it isn't deprecated I'd take this out and just call this a convenience constructor for the common case where the joint will own the two Frames it uses to attach to its bodies.

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -131,30 +132,65 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
...
-          const PhysicalFrame& child,
-          const SimTK::Vec3& locationInChild,
-          const SimTK::Vec3& orientationInChild,
-          bool reverse = false);
+    Joint( const std::string& name,
+           const std::string& parentName,
+           const std::string& childName,
+           bool reverse = false);
+
+    /** <b>(Deprecated)</b> Use Joint(name, parentName, childName) instead.
```
sherm1: remove "deprecated"?

sherm1:
Done re-reviewing -- looks good.

OpenSim/Simulation/Model/PhysicalOffsetFrame.h
aseth1: `PhysicalOffsetFrame` is a mixin (i.e.. `OffsetFrame<PhysicalFrame>`) so I would want `OffsetFrame<>` to be extended (including more convenient constructors). It is for that particular reason (where i almost added a new convenience constructor to PhysicalOffsetFrame instead of OffsetFrame<>) that I wish to block extensions here. In hindsight that may be too harsh and even impractical, but that was my reasoning. I can add that reasoning to the class description and/or eliminate the restriction.

What is your preference here? @sherm1  @chrisdembia @aymanhab

OpenSim/Simulation/SimbodyEngine/Joint.h
aseth1: Is everyone happy to have this method as not being deprecated? I am leaning that way.
Of we vote to deprecate then I will convert PhysicalFrame to Body as well.

Do we have consensus to leave as convenience? @sherm1 @chrisdembia @aymanhab

OpenSim/Simulation/Model/PhysicalOffsetFrame.h
sherm1: > What is your preference here? (regarding `final` for PhysicalOffsetFrame)

Mine would be to allow extension for any class unless it would actually fail to function properly when extended.

OpenSim/Simulation/SimbodyEngine/Joint.h
sherm1: > Do we have consensus to leave as convenience? [Joint constructor that also takes frame information]

My vote would be to leave the constructor because 
- then the two constructors together represent the two most-common ways to structure a Joint subassembly -- either it is the Joint alone and requires connecting to two frames, or it is the Joint plus two offset frames and requires connecting to two bodies;
- it's backwards-compatible; and,
- I'm also imagining that having both of these could help people get a grip on the ownership vs. connection distinction.

aymanhab: I'm fine with leaving it as well. Thanks.

OpenSim/Simulation/Model/PhysicalOffsetFrame.h
aymanhab: I agree with @sherm1 particularly that the error message users get if they try to extend will likely be cryptic which is more harm than good in this case.

OpenSim/Wrapping/Java/swig/javaWrapOpenSim.i
```diff
@@ -565,6 +565,41 @@ SWIG_JAVABODY_PROXY(public, public, SWIGTYPE)
...
+
+EXPOSE_JOINT_CONSTRUCTORS_HELPER(FreeJoint);
+EXPOSE_JOINT_CONSTRUCTORS_HELPER(BallJoint);
+EXPOSE_JOINT_CONSTRUCTORS_HELPER(PinJoint);
+EXPOSE_JOINT_CONSTRUCTORS_HELPER(SliderJoint);
+EXPOSE_JOINT_CONSTRUCTORS_HELPER(WeldJoint);
+EXPOSE_JOINT_CONSTRUCTORS_HELPER(GimbalJoint);
+EXPOSE_JOINT_CONSTRUCTORS_HELPER(UniversalJoint);
+EXPOSE_JOINT_CONSTRUCTORS_HELPER(PlanarJoint);
+
```
aymanhab: I don't think we have examples already that use these custom constructors, do we want to promote them in C++ or in new examples? if so I'm ok with leaving this, otherwise I'd just remove this block.

OpenSim/Simulation/SimbodyEngine/Joint.h
aseth1: I spoke with @chrisdembia and he is OK with not deprecating.  I am removing all indications of deprecation.

OpenSim/Wrapping/Java/swig/javaWrapOpenSim.i
chrisdembia: This swig macro implements the previously-existing constructor and also provides the new Joint constructor. In python, we do have tests for the previously-existing constructor but not for new one.

Might be good to add comments in the swig .i file saying that these are only here b/c swig doesn't generated constructors for C++11 constructor inheritance (`using Derived::Derived;`).

aymanhab: Thanks much @chrisdembia I checked the SWIG documentation and found the snippet regarding constructor inheritance. Glad you found it and had a solution already. Will need a similar solution for Java/Matlab.

chrisdembia: The same thing should work for java, right?

aseth1:
Thanks @sherm1 @chrisdembia @ayman for the quick feedback.
Let me know is I should merge (upon CI passing) or feel free to do it!

sherm1:
I love doing that!

PR #595, opened 2015-09-01
Use ResetOnCopy<unique_ptr<ModelVisualizer>> in Model.

chrisdembia:
Since ModelVisualizer has a private destructor, I needed to create a custom deleter within Model, which is friended to ModelVisualizer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -117,7 +117,7 @@ before_install:
 install:
   - mkdir ~/opensim-core-build && cd ~/opensim-core-build
   # Configure OpenSim.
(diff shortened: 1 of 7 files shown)


sherm1:
That deleter is pretty cool, but I have a few questions:
- I couldn't spot where the deleter class was made a friend of ModelVisualizer. Did I miss it or did it somehow earn friendship by being a local class of Model, which is explicitly friended?
- Wouldn't it be better just to make ModelVisualizer's destructor public?

chrisdembia:
> Did I miss it or did it somehow earn friendship by being a local class of Model, which is explicitly friended?

This was kinda mysterious to me and I just assumed what you say.



> Wouldn't it be better just to make ModelVisualizer's destructor public?

Probably. I didn't want to spend too much time on this, and thought making the destructor public might require more extensive review. Though, I suppose the important thing is that the constructor is private; I don't know why the destructor would need to be private. I'll go this route instead.

chrisdembia:
I got rid of the deleter. Ready for review again.

sherm1:
Looks great. Merging.

chrisdembia:
Thanks

PR #599, opened 2015-09-03
[python] JOINT_ADOPT_HELPER no longer needed.

### LINKED ISSUE #597: Python wrapping may no longer need  JOINT_ADOPT_HELPER macro
This macro might have been needed because the memory for joint was managed by body? I'm not sure but maybe we should revisit.

chrisdembia:
Fixes #597. The python tests pass so I think it's okay to remove this part of the swig interface.

@aymanhab could you review and merge?


aymanhab:
I'm ok with removing if the tests pass, I think you're right about the history of how it got there.

chrisdembia:
@aymanhab please merge if you think this is good to go in.

PR #600, opened 2015-09-04
[doxygen] Remove unwarranted hyperlinks to Set

tkuchida:
There are probably more of these…


chrisdembia:
probably?

tkuchida:
> probably?

Yes, like [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Object.h#L264).

PR #604, opened 2015-09-08
[cmake] Some housekeeping for the CMake files.

chrisdembia:
I made a few of the cmake options non-advanced, e.g., `BUILD_(JAVA|PYTHON)_WRAPPING` are no longer advanced CMake variables.

I also deleted some stuff from various CMakeLists files.

I think the changes here are largely uncontroversial and I'd like someone to review and merge this quickly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,33 +52,27 @@ if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
 endif()
 
 
-# Build binary directory.
-# -----------------------
-set(BUILD_BINARY_DIR ${CMAKE_BINARY_DIR}
-    CACHE PATH
-    "The OpenSim build (not the install) puts all the libraries and executables together here (with /Release, etc. appended on some platforms).")
-
 # Make everything go in the same binary directory. (These are CMake-defined
 # variables.)
-set(EXECUTABLE_OUTPUT_PATH ${BUILD_BINARY_DIR})
-set(LIBRARY_OUTPUT_PATH ${BUILD_BINARY_DIR})
-link_directories(${BUILD_BINARY_DIR})
+set(EXECUTABLE_OUTPUT_PATH ${CMAKE_BINARY_DIR})
+set(LIBRARY_OUTPUT_PATH ${CMAKE_BINARY_DIR})
+link_directories(${CMAKE_BINARY_DIR})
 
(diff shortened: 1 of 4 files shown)


sherm1:
This all looks good to me, Chris. If the builds work I think it is fine to merge.

chrisdembia:
Okay thank you.

aymanhab:
Looks fine @chrisdembia The main reason the WRAPPING was marked advanced is that you need to know what you're doing to turn it on (e.g. have SWIG correct version, and either Python or Java installed) which may not be obvious to non advanced users. Any thoughts? 
All the other changes are good cleanup. Thanks,

chrisdembia:
I expect that the only people who would build the python or java wrapping would be those people we'd expect to see at an Advanced Users workshop; people who we would say are 'advanced opensim users'. However, I think that's different from advanced cmake configuration. My feeling about advanced cmake vars is that if we expect build-from-source users to change a cmake variable, it should not be advanced. Advanced cmake vars should be ones that only the dev team cares about (such as `OPENSIM_SIMBODY_DOXYGEN_LOCATION`).

I think it should be as easy as possible for people to build python/java wrapping from source. Making the variables non-advanced makes the process slightly easier. Of course, the need to have swig, etc still remains, but that is described in the README.md.

chrisdembia:
I actually think the cmake people chose the wrong word, 'advanced'. I think they meant something more like 'hidden'.

aymanhab:
Ok, we can revert the variable to ADVANCED after 10 questions along the line "I turned Wrapping On but now I have 10 different red lines in CMake that I don't understand" and I can't build anymore  ;)

chrisdembia:
I'd be happy to go back to marking it advanced if that happens. But I don't understand that scenario. Users should only alter the variable if they *want* to build python wrapping, in which case they must deal with those red lines anyway. Have you found that users twiddle cmake variables when they didn't mean to?

aymanhab:
My experience is that users would have good intentions so they'd try building everything without complete understanding of the consequences. Can't tell for this one specifically because it has been "hidden" all along, so let's go ahead and merge it.

chrisdembia:
Okay. Thanks for your willingness to try this. I could see how a user could be tempting to flick the option.

chrisdembia:
We can also provide more useful error messages in the case that swig/python/java can't be found.

PR #605, opened 2015-09-08
Remove dependency on SWIGSimTK files

chrisdembia:
@aymanhab the java and python wrapping _should_ work now, but I think using the SimTK headers may have made the files larger and that is causing the failure on travis. On appveyor, the wrapping works though.

So I suppose this is not ready to merge yet. Perhaps splitting the wrapping into multiple modules must happen first?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ env:
   global:
     # The python tests look for OPENSIM_HOME.
     - OPENSIM_HOME=~/opensim-install
-    - SWIG_VER=3.0.5
+    - SWIG_VER=3.0.6
     - USE_CCACHE=1
     - CCACHE_COMPRESS=1
     # for Clang to work with ccache.
--- a/OpenSim/Wrapping/CMakeLists.txt
+++ b/OpenSim/Wrapping/CMakeLists.txt
@@ -1,6 +1,6 @@
 
 if(BUILD_PYTHON_WRAPPING OR BUILD_JAVA_WRAPPING)
-    find_package(SWIG 3.0.5 REQUIRED)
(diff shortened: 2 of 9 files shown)


aymanhab:
@chrisdembia Thanks I'll try it. Since the GUI repository has the SWIG output before the change I can tell what changed with the move to the native SimTK files and then we can discuss.

chrisdembia:
Oh okay; that would be helpful. However, travis is failing on the python wrapping. I'm not sure if travis fails on the java wrapping. We might need to diff the generated files for the python wrapping.

aymanhab:
@chrisdembia Is there a reason to require SWIG 3.0.6?

chrisdembia:
Yes. With swig 3.0.5, swig thinks there's a syntax error on line 48 of Vec.h :/ Is 3.0.6 okay? We can wait to merge this until we can make the switch.

aymanhab:
@chrisdembia Building the wrapping with SWIG 3.0.7 succeeds but:
- There're more than 120 classes that SWIG didn't know how to resolve as template instantiations that ended up cluttering the wrapping result (they end up as SWIGTYPE_*)
- Key SimTK namespace classes were not mapped to instantiated templates and cannot be passed in through the interface, in particular: Vector, Matnn, UnitVec
- Many extraneous methods that return those SWIGTYPE_* objects
Basically any class that appears in the headers that we don't even care to expose ends up in the wrapping. Similar problem if a class appears in a header before it is defined to SWIG.

aymanhab:
@chrisdembia Considering the size of the changes that need to be made to produce essentially the hand tuned Bindings we have today, it'd be easier to try to restore the usage of the 14 files in this folder, one at a time.

PR #607, opened 2015-09-09
[README] Noted remedy if VS installed without C++

tkuchida:
It's surprising that Visual Studio 2015 does not support C++ by default. I selected Typical during the install, which is what I would expect most people to do. Analogous to [this Simbody PR](https://github.com/simbody/simbody/pull/437).

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,8 @@ On Windows using Visual Studio
     * Visual Studio 2015 does not install C++ 
       support by default. During the installation you must select 
       *Custom*, and check *Programming Languages > Visual C++ > Common Tools for Visual C++ 2015*.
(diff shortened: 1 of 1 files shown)


sherm1:
Thanks again, Tom!

sherm1:
P.S. For changes like this that don't need testing, you can put `[ci skip]` in your commit comment to prevent GitHub from kicking off unnecessary Travis and AppVeyor builds. (No harm if it builds though except for some delay before builds can get done for other things.)

tkuchida:
> you can put [ci skip] in your commit comment to prevent GitHub from kicking off unnecessary Travis and AppVeyor builds

Thanks for the reminder.

PR #612, opened 2015-09-10
[readme] Update required cmake version for windows.

chrisdembia:
Update required cmake version on windows to a version of cmake that supports vs 2015.

I did not update the minimum required CMake version listed in CMakeLists.txt since non-visual studio users can use earlier versions of cmake, and ubuntu users do not have easy access to cmake 3.


tkuchida:
:+1:

chrisdembia:
Thank you.

Issue #482, opened 2015-06-09
Current Visualization Design may lead to numerous Geometry and Appearance classes


aseth1:
Currently Components are offloading their visualization to Geometry and using Appearance to control how it is visualized. For example, Components can have lines drawn via a LineGeomtery component. A LineAppearance controls its thickness when drawn in a Visualizer. Similarly, a Frame (component) has a FrameGeometry but is has `display_radius` to control the thickness of the axes. Changing a property of Component is interpreted as a change in its behavior so changing axis thickness will force a recreation of the system. It seems we would then want a FrameAppearance. I think this becomes the case for Point (to exposed radius of the sphere) and AnalyticalGeomtery (to describe the number of facets used to visualize smooth analytical shapes). 

I really like the idea of having an Appearance that only changes the visualization and not the physics of the model. The current approach is coming at the expense of duplicate code and multiple (practically empty) classes (e.g. to expose a property).

Would it be easier/cleaner to then to have a new AppearanceProperty of an Object/Component? The purpose of Appearance is to expose properties that can affect the appearance of a Component and not its physics. The Appearance classes only contain properties (as far as I can tell) and that makes me question their existence (i.e. they have no actions). Unlike regular properties, AppearanceProperties could have their values changed and they would not mark the Object as not being up-to-date (thus would not trigger setting up from properties, connecting, etc.. to re-initSystem). That is my main worry for including things that only affect visualization as properties. We must have a clear separation between physics (properties) and visuals (appearance) and the latter should have absolutely no effect on the former. For example Line would have `thickness` as an AppearanceProperty.

Perhaps another way of controlling appearance without polluting the modeling/physics API with Appearances or AppearanceProperties is for the Visualizer to employ a Visitor to apply appearances based on the Component (and its Geometry) that are encountered. That is, the Visualizer maintains a mask to apply to individual Components or by types (e.g. draw WrapObjects as wireframe  or make PhysicalFrame axes green), 

A combination of the two may be best. Where Components have default AppearanceProperties defined by the modeler but the Visualizer/GUI is free to override based on direct user input. Neither of which affects the computations of the system.

sherm1:
I wonder if we should call these specified-by-the-model properties `DefaultAppearance` or something like that. I think that's what they are and if so they are just kind of a hint by the modeler as to how they think the user will want to view things. But the actual appearance might be quite different after the user takes control in the GUI.

[...]

Issue #480, opened 2015-06-08
Appearance class names and hierarchy need improvement


aymanhab:
In the past OpenSim GUI was capable of displaying surfaces only. The new Geometry interface allows for more general objects (e.g. made of line-segments, text, maybe others). Appearance specification should expose all-and-only-relevant properties for the corresponding type of Geometry.

aymanhab:
Discussion from geometry integration PR #417 
Besides the big picture above here is a summary of issues:

-  The BaseAppearance class should just be the Appearance.
- Appearance should define Representation (currently still DisplayPreference) not Geometry.
The current Appearance class should be SurfaceAppearance since textures can only be applied to surfaces.
- DisplayPreference Enum should be renamed Representation since properties and accessors refer to “representation”. All use of DisplayPrefence (including comments) should be eliminated.
Eliminate “Draw” from the enumerated types of Representation. The Draw prefix adds little. “Hidden’, “Vertices”, “WireFrame”, “Surface” seem clear enough. Why is an explicit Default required is that not implied if one is not specified?
- The Appearance::Representation name (string) should be serialized (e.g. Hidden, Vertices, etc…) (as in the case of Coordinate::MotionType) opposed to its integer value.

jenhicks:
@aymanhab Is this issue still relevant? Or is there another related issue based on the latest PR?

aymanhab:
@jenhicks The arguments have been revised and rediscussed since then in-person, in dev. meeting and ongoing with latest PR of Geometry/Appearance.

PR #616, opened 2015-09-11
Geometry Appearance refactoring

aymanhab:
Phase 1 of Geometry/Appearance refactoring that includes:
1. Implementing an Appearance class that serves as a container for generic attributes, CurveAppearance and SurfaceAppearance.
2. Removal of DisplayPreference enum and replacing with SimTK::DecorativeGeometry::Representation  counterpart.
3. Removal of BaseAppearance, LineAppearance and other classes now unused.

The remaining part of implementing the agreed upon design is replacing the Appearance Property with a reference to Appearance that can be located using name lookup.


OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,157 @@ namespace OpenSim {
...
-class OSIMSIMULATION_API LineAppearance : public BaseAppearance {
-    OpenSim_DECLARE_CONCRETE_OBJECT(LineAppearance, Object);
+//=============================================================================
+/**
+ * A class that holds the display attributes (Appearance) of an object displayed 
+ * in the OpenSim Visualizer.
+ * 
+ * Appearance contains properties that apply to all geometry.
+ * Geometry that have a surface so that it can be textured will utilize
+ * surface_ppearance, while schematic line drawings (e.g. Arrows, Frames)
```
tkuchida: typos: surface_appearance, curve_appearance

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,157 @@ namespace OpenSim {
...
+        constructProperties();
     }
-    virtual ~Appearance() {};
+    virtual ~CurveAppearance() {};
 
-    bool hasTexture() {
-        return getProperty_texture_file().size() > 0;
+private:
+    void constructProperties() {
+        constructProperty_thickness(.05);
```
tkuchida: Why is default 0.05? Perhaps scale this property internally so user-facing default is 1.0?

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,157 @@ namespace OpenSim {
...
     //==========================================================================
     // PROPERTIES
     //==========================================================================
-    OpenSim_DECLARE_PROPERTY(color, SimTK::Vec3,
-        "The color (rgb) used to display the object. ");
-    OpenSim_DECLARE_PROPERTY(opacity, double,
-        "The opacity (0-1) used to display the object. ");
     OpenSim_DECLARE_PROPERTY(representation, int,
-        "The representation (0:Hidden, 1:Points, 2:Wire 3:Shaded) used to display the object. ");
+        "The representation (1:Points, 2:Wire 3:Shaded) used to display the object. ");
```
tkuchida: Missing comma after 2:Wire

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,157 @@ namespace OpenSim {
...
+ * Appearance contains properties that apply to all geometry.
+ * Geometry that have a surface so that it can be textured will utilize
+ * surface_ppearance, while schematic line drawings (e.g. Arrows, Frames) 
+ * can utilize curve_ppearance which offers thickness. 
+ *
+ * TODO: Add Resolution to Appearance from DecorativeGeometry and 
+ * utilize  for Arrow, Line, and Frame (unused for now)
+ *
+ * @author Ayman Habib
+ * @version 1.0
```
tkuchida: Are these version numbers actually maintained?

tkuchida:
Oops, I forgot the `</review>` message. I just had minor suggestions; feel free to ignore. Someone else should review for philosophical content.

aymanhab:
Thanks for the non philosophical review @tkuchida Will fix.

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,157 @@ namespace OpenSim {
 class Body;
 class Model;
 
-
-//=============================================================================
-//=============================================================================
+using SimTK::DecorativeGeometry;
```
sherm1: It is generally a bad idea to put `using` statements in header files because they quietly introduce symbols into the namespaces of anyone who includes the header, even indirectly. You may be able to put the `using` statement within a class declaration; if not you should write it out explicitly with the `SimTK::` prefix. You can of course put `using` at the top of any `.cpp` file so that the implementation does not have to be cluttered the same way as the header.

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,157 @@ namespace OpenSim {
...
     //==========================================================================
-    OpenSim_DECLARE_PROPERTY(color, SimTK::Vec3,
-        "The color (rgb) used to display the object. ");
-    OpenSim_DECLARE_PROPERTY(opacity, double,
-        "The opacity (0-1) used to display the object. ");
     OpenSim_DECLARE_PROPERTY(representation, int,
-        "The representation (0:Hidden, 1:Points, 2:Wire 3:Shaded) used to display the object. ");
+        "The representation (1:Points, 2:Wire 3:Shaded) used to display the object. ");
+    OpenSim_DECLARE_OPTIONAL_PROPERTY(texture_file, std::string,
+        "Name of file containing texture. ");
```
sherm1: Is it always obvious how the texture is mapped to the surface? How is that dealt with?

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,157 @@ namespace OpenSim {
...
 
     //--------------------------------------------------------------------------
     // CONSTRUCTION
     //--------------------------------------------------------------------------
 public:
-    BaseAppearance() {
-        constructProperties();
+    SurfaceAppearance() {
+        constructProperty_representation(3); // Surface shaded by default
```
sherm1: Could this be an enum?

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,157 @@ namespace OpenSim {
...
-    virtual ~BaseAppearance() {};
+    virtual ~SurfaceAppearance() {};
 
-private:
-    void constructProperties() {
-        constructProperty_color(SimTK::Vec3(1.0));
-        constructProperty_opacity(1.0);
-        constructProperty_representation(3);
+    bool hasTexture() {
+        return getProperty_texture_file().size() > 0;
```
sherm1: `std::string` has a bool method `empty()` for this purpose: should say `return !getProperty_texture_file().empty();`.

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,157 @@ namespace OpenSim {
...
 public:
-    //==========================================================================
+    //==============================================================================
     // PROPERTIES
-    //==========================================================================
-    OpenSim_DECLARE_OPTIONAL_PROPERTY(texture_file, std::string,
-        "Name of file containing texture. ");
+    //==============================================================================
+    OpenSim_DECLARE_PROPERTY(thickness, double,
+        "The thickness of lines used to render a curve or a drawing. ");
```
sherm1: What are the units of thickness? Should say in the description.

OpenSim/Simulation/Model/Appearance.h
sherm1: The `@version`s should be removed.

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,157 @@ namespace OpenSim {
...
-    LineAppearance() {
+    Appearance() {
         constructProperties();
     }
-    virtual ~LineAppearance() {};
+    virtual ~Appearance() {};
+
+    DecorativeGeometry::Representation get_representation() const { return (DecorativeGeometry::Representation)get_surface_appearance().get_representation(); }
+
+    void set_representation(DecorativeGeometry::Representation& rep) { upd_surface_appearance().set_representation(rep); }
```
sherm1: There are line length problems here and above. Should not exceed 80 chars. I couldn't see the whole line in the GitHub viewer, but I believe the argument here should be const, or you can just pass it by value (probably better in this case since it is just an int).

OpenSim/Simulation/Model/Geometry.h
```diff
@@ -155,12 +144,12 @@ class OSIMSIMULATION_API Geometry : public Component {
     };
 
     /// Convenient access to set Appearance/representation
-    void setRepresentation(const DisplayPreference& rep) { 
+    void setRepresentation(DecorativeGeometry::Representation rep) {
```
sherm1: How do these work if a Geometry object is using a shared appearance?

sherm1:
Done reviewing. Looks good -- a few comments above.

OpenSim/Simulation/Model/PhysicalFrame.cpp
```diff
@@ -48,7 +48,7 @@ PhysicalFrame::PhysicalFrame() : Frame()
     FrameGeometry frm(0.2);
```
aseth1: FrameGeometry should be removed and Frames should be able to draw/generate its own geometry.

OpenSim/Simulation/Model/Appearance.h
aseth1: In the first review I asked that these be strings that get mapped to enums. Also, is '0' now an illegal option? Or is that a backdoor for getting the geometry to be hidden?

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,157 @@ namespace OpenSim {
...
-class OSIMSIMULATION_API Appearance : public BaseAppearance {
-    OpenSim_DECLARE_CONCRETE_OBJECT(Appearance, BaseAppearance);
+//=============================================================================
+/** CurveAppearance class holds the display attributes  of a piece of Geometry
+displayed in the OpenSim visualizer or GUI.The attributes in this
+class are specific to curves or line drawings so that their thickness, line-style
+etc. can be maintained. Attributes common to all possible geometries that can be 
+rendered (e.g.color, opacity, visibility) are not included
+here but provided by the Appearance class instead.
+*/
```
aseth1: The first line for the class description of SurfaceAppearance is identical to CurveAppearance. It would be better if it they said the they contain the Appearance attributes for a surface and a curve, respectively.

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,157 @@ namespace OpenSim {
...
+};  // END of class CurveAppearance
 
 //=============================================================================
-class OSIMSIMULATION_API LineAppearance : public BaseAppearance {
-    OpenSim_DECLARE_CONCRETE_OBJECT(LineAppearance, Object);
+//=============================================================================
+/**
+ * A class that holds the display attributes (Appearance) of an object displayed 
+ * in the OpenSim Visualizer.
+ *
```
aseth1: Appearance is a clear concept. It is unnecessary to introduce new terms like "display attributes". Something like "Appearance is a class used to control how Geometry appears in the Visualizer" seems more straightforward.

OpenSim/Simulation/Model/Geometry.h
aseth1: Comment also seems incorrect. Since it is the 'DecorativeGeometry::Representation' that is being set. Unless is is also the same thing as the `SurfaceAppearance` `representation` (property) , which is confusing.

OpenSim/Simulation/RegisterTypes_osimSimulation.cpp
```diff
@@ -162,9 +162,9 @@ OSIMSIMULATION_API void RegisterTypes_osimSimulation()
     Object::registerType( PathPointSet() );
     Object::registerType( ConditionalPathPoint() );
     Object::registerType( MovingPathPoint() );
-    Object::registerType( BaseAppearance());
+    Object::registerType( SurfaceAppearance());
+    Object::registerType( CurveAppearance());
     Object::registerType( Appearance());
```
aseth1: `SurfaceAppearance` and `CurveAppearance` probably should not be exposed as individual objects since Appearance (which contains both)  is the only way to access/modify.

aseth1:
Done reviewing:

Recap:

0. Class descriptions should be more concise and meaningful.

1. grouping curve and surface attributes of an Appearance should not elevate them to accessible individual classes that are now part of the API.  I am OK  with them being hidden inside if Appearance.

2. The public API should not expose `DecorativeGeometry`. The fact that `OpenSim::Geometry` is backed by `SimTK::DecorativeGeometry` will be confusing to most since one would expect `DecorativeGeometry` to derive from `Geometry`, right?


It would also be nice to see an example of a Model (.osim) file with Appearances in them. My concerns are: 

1. Should the corresponding property as it appears in XML be `default_appearance` to at least indicate that the Visualizer/GUI has final say?

2. How do appearances now appear in the Model file? Can we use sensible defaults so that very little real-estate is taken up by Appearance. Just its name and then only non-default values.

3. Related to 2. Even if the values are default will it not still take up quite a bit of space just to list the SurfaceAppearance and CurveAppearance with none of their contents exposed? e.g.

```XML
<Component name = "my_component">
    <Geometry name = "some_geometry">
        <Appearance name = "the_appearance">
            <SurfaceAppearance name = "a_surface_appearance">
            </SurfaceAppearance>
            <CurveAppearance name = "a_curve_appearance">
            </CurveAppearance >
        </Appearance>
    </Geometry>
</Component>
```

OpenSim/Simulation/Model/Appearance.h
aymanhab: @tkuchida Good point, in the past we used the model units (meters) for sizing, however we can change the convention since we're going to a new system. This will map to SimTK property under the covers and it'd be good to keep them consistent. I'm not sure what the SimTK visualizer use as units for thickness.

OpenSim/Simulation/Model/Appearance.h
aymanhab: @sherm1  Will use whatever provided by the visualization package. Normally built in primitive shapes have texture coordinates, things get more tricky with arbitrary meshes.

OpenSim/Simulation/Model/Appearance.h
aymanhab: Since this is an optional property, the check/test is not done on an actual string but on a Property that may or may not have been populated. Would be nicer if optional properties had a "isPopulated" method to do this test.

OpenSim/Simulation/Model/Appearance.h
chrisdembia: I think users might like specifying visualization dimensions in meters, etc. For example, I like saying markers have a radius of 0.01 m (or whatever it is).

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,159 @@ namespace OpenSim {
...
-};  // END of class BaseAppearance
-
+    //=============================================================================
+};  // END of class SurfaceAppearance
 
-class OSIMSIMULATION_API Appearance : public BaseAppearance {
-    OpenSim_DECLARE_CONCRETE_OBJECT(Appearance, BaseAppearance);
+//=============================================================================
+/** CurveAppearance class holds the display attributes  of a piece of Geometry
+displayed in the OpenSim visualizer or GUI.The attributes in this
```
chrisdembia: missing space after period.

OpenSim/Simulation/Model/Appearance.h
chrisdembia: fyi I started working on a FilesystemPath property that could be used here.

OpenSim/Simulation/Model/Appearance.h
chrisdembia: @aymanhab the Property also has the `empty()` method. @sherm1 referred to both in his comment; I'm not sure whether he meant `std::string::empty()` or `Property<std::string>::empty()`.

chrisdembia: (it's used in PrescribedController).

OpenSim/Simulation/Model/Appearance.h
sherm1: > I'm not sure what the SimTK visualizer use as units for thickness.

The Simbody visualizer passes the thickness value (a float) on to OpenGL's [`glLineWidth()`](https://www.opengl.org/sdk/docs/man3/xhtml/glLineWidth.xml) method. 1.0 is the default. Basically OpenGL makes 2.0 twice as fat, etc., but the actual meaning depends on whether antialiasing is in use.

OpenSim/Simulation/RegisterTypes_osimSimulation.cpp
aymanhab: Can't serialize/deserialize without registration.

OpenSim/Simulation/Model/Appearance.h
tkuchida: I didn't think these lines would scale with the model. If they do (e.g., if they increase in thickness when you zoom in), then I agree that physical units would make the most sense (though 0.05 m seems too thick for a default). I thought this property referred to the thickness of the lines regardless of scale, in which case perhaps pixels would be the most intuitive unit.

OpenSim/Simulation/Model/Appearance.h
sherm1: > Property also has the empty() method. @sherm1 referred to both in his comment; I'm not sure whether he meant std::string::empty() or Property<std::string>::empty().

I didn't notice that it was a Property rather than a string, but lucky for me Property has empty() also so I can look like I knew what I was saying!

>Would be nicer if optional properties had a "isPopulated" method to do this test.

Property objects are `(name, list-of-things)` pairs. They only differ in the allowable minimum and maximum lengths for the lists. "optional" is just a catchy name for a `(minLength=0,maxLength=1)` property list. So checking whether the list is `empty()` is the right thing to do.

OpenSim/Simulation/Model/Appearance.h
sherm1: > I think users might like specifying visualization dimensions in meters, etc. 

I agree for things that are actually drawn using geometric objects. Line thickness is best done unitless though because it is an abstract representation, depending for its details on the device, anti-aliasing treatment, screen resolution etc. To draw a line of a particular thickness given in meters, draw a cylinder or long rectangular shape rather than using an abstract line.

chrisdembia: Ah okay thanks @tkuchida and @sherm1; I agree.

OpenSim/Simulation/Model/Appearance.h
aymanhab: Cool, will do.

aymanhab:
I believe all comments that fall within the scope of this PR have been addressed. Thanks all. 
@aseth If properties within Appearance, CurveAppearance or SurfaceAppearance are not changed from default they don't show in XML altogether. Now using unnamed property because there's only one Appearance per Geometry. 
@sherm1 @chrisdembia Will document behavior of thickness when it gets wired/used.

aseth1:
I will review soon.

OpenSim/Examples/ExampleMain/OutputReference/TugOfWar_Complete.cpp
```diff
@@ -115,7 +115,7 @@ int main()
         
         Cylinder cylGeometry(0.2, .3);
         cylGeometry.setFrameName("CylAnchor");
-        cylGeometry.setRepresentation(Geometry::DrawWireframe);
+        cylGeometry.setRepresentation(DecorativeGeometry::DrawWireframe);
```
aseth1: Is this the same as `VisualRepresenation` below? Why does Geometry expose it? Isn't it an attribute of the Appearance?

If you have multiple Geometry share the same Appearance, what does it mean for the Geometry to set an attribute of the shared appearance? Does it get its own copy at that point or do the appearance of all the objects that share that Appearance suddenly change?

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,160 @@ namespace OpenSim {
 class Body;
 class Model;
 
+/**
+VisualRepresentation is the OpenSim name used across the OpenSim API, it maps to
+SimTK::DecorativeGeometry::Representation.
+*/
```
aseth1: This comment does not say what a `VisualRepresentation` is. Is it useful for the user of the API to know the existence of `SimTK::DecorativeGeometry::Representation`?

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,160 @@ namespace OpenSim {
...
     // PROPERTIES
     //==========================================================================
-    OpenSim_DECLARE_PROPERTY(color, SimTK::Vec3,
-        "The color (rgb) used to display the object. ");
-    OpenSim_DECLARE_PROPERTY(opacity, double,
-        "The opacity (0-1) used to display the object. ");
     OpenSim_DECLARE_PROPERTY(representation, int,
-        "The representation (0:Hidden, 1:Points, 2:Wire 3:Shaded) used to display the object. ");
+     "The representation (1:Points, 2:Wire, 3:Shaded) used to display the object.");
+    OpenSim_DECLARE_OPTIONAL_PROPERTY(texture_file, std::string,
```
aseth1: What is 0? Does it result in hiding? Is this the same as the `VisualRepresentation` above?

I have also stated before that I would prefer strings over ints since they are more informative. Any comments/response to that?

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,160 @@ namespace OpenSim {
...
+    //========================================================================
+};  // END of class SurfaceAppearance
 
-class OSIMSIMULATION_API Appearance : public BaseAppearance {
-    OpenSim_DECLARE_CONCRETE_OBJECT(Appearance, BaseAppearance);
+//=============================================================================
+/** CurveAppearance class holds the Appearance properties of Geometry
+displayed in the OpenSim visualizer or GUI.The attributes in this
+class are specific to curves or line drawings so that thickness, line-style
+etc. can be maintained.
```
aseth1: why isn't line_style a property?

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,160 @@ namespace OpenSim {
...
+//=============================================================================
+/** CurveAppearance class holds the Appearance properties of Geometry
+displayed in the OpenSim visualizer or GUI.The attributes in this
+class are specific to curves or line drawings so that thickness, line-style
+etc. can be maintained. 
+*/
+//=============================================================================
+class OSIMSIMULATION_API CurveAppearance : public Object {
+    OpenSim_DECLARE_CONCRETE_OBJECT(CurveAppearance, Object);
 public:
```
aseth1: As I mentioned before, I don't think this class should exist to hold one or couple properties. While I agree curves and surfaces are not the same thing, nowhere in the code or examples is there a distinction made between different sub appearances. An Appearance could easily have `line_thickness` and `line_color` in addition to `surface_color` and `surface_representation`.

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,160 @@ namespace OpenSim {
...
+};  // END of class CurveAppearance
 
 //=============================================================================
-class OSIMSIMULATION_API LineAppearance : public BaseAppearance {
-    OpenSim_DECLARE_CONCRETE_OBJECT(LineAppearance, Object);
+//=============================================================================
+/**
+ * A class that holds the Appearance properties of an object displayed 
+ * in the OpenSim Visualizer.
+ *
```
aseth1: should say something about `Appearance` affects <i>how</i> Geometry is displayed.
It does not apply to Objects.

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,160 @@ namespace OpenSim {
...
+    //==========================================================================
     // PROPERTIES
-    //==============================================================================
-    OpenSim_DECLARE_PROPERTY(thickness, double,
-        "The thickness used to visualize a LineGeometry. ");
+    //==========================================================================
+    OpenSim_DECLARE_PROPERTY(visible, bool, 
+        "Flag indicating whether the associated Geometry is visible or hidden.")
+    OpenSim_DECLARE_PROPERTY(opacity, double,
+            "The opacity (0-1) used to display the object. ");
```
aseth1: value between (0), which is completely transparent, and (1), which is completely opaque?

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,160 @@ namespace OpenSim {
...
+    OpenSim_DECLARE_PROPERTY(color, SimTK::Vec3,
+        "The color (rgb) used to display the object. ");
+
+    OpenSim_DECLARE_PROPERTY(surface_appearance, SurfaceAppearance,
+        "Visuals applied to surfaces associated with this Appearance. ");
+    OpenSim_DECLARE_PROPERTY(curve_appearance, CurveAppearance,
+        "Visuals applied to curves or line drawings associated with this Appearance. ");
```
aseth1: I don't believe there was a good case made for two separate classes to contain sub-lists of properties.

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,160 @@ namespace OpenSim {
...
 public:
-    LineAppearance() {
+    Appearance() {
         constructProperties();
     }
-    virtual ~LineAppearance() {};
+    virtual ~Appearance() {};
+
+    VisualRepresentation get_representation() const { 
+        return (VisualRepresentation)get_surface_appearance().get_representation(); }
```
aseth1: What is gained by a separate surface appearance if you just short circuit it anyway?

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -33,105 +33,160 @@ namespace OpenSim {
...
+    void set_representation(VisualRepresentation& rep) { 
+        upd_surface_appearance().set_representation(rep); }
+
+protected:
+    /** Updating XML formating to latest revision */
+    void updateFromXMLNode(SimTK::Xml::Element& aNode, int versionNumber) override {
+        int documentVersion = versionNumber;
+        if (documentVersion < XMLDocument::getLatestVersion()) {
+            if (documentVersion < 30505) {
+                // if representation = 0 visible = false, else true
```
aseth1: if A <i>then</i> B. Else .. 
Where is Appearance added to XML in old formats that do not have Appearance?

OpenSim/Simulation/Model/Geometry.h
```diff
@@ -132,9 +118,12 @@ class OSIMSIMULATION_API Geometry : public Component {
         SimTK::DecorativeGeometry& decoration) const {
```
aseth1: Who is supposed to call `setDecorativeGeometryTransform` and `setDecorativeGeometryAppearance`?
Are they intended to be part of the public interface? I also do not understand what is their intended purpose.

In `setDecorativeGeometryTransform` where does the Array of decorations come from? Is this supposed to be a helper to help add to the decorations Array that generateDecorations is supposed to fill in?

The comment for `setDecorativeGeometryAppearance` is confusing. Manage A by applying A?
Shouldn't the push down to the decorative geometry be completely hidden from the user API? It is not clear that the passed in decoration is being updated according to the Appearance of this Geometry.

Could the underlying SimTK::DecoratievGeometry array representing this OpenSim::Geometry be updated automatically without exposing these or similar methods?

OpenSim/Simulation/Model/Geometry.h
```diff
@@ -155,12 +144,12 @@ class OSIMSIMULATION_API Geometry : public Component {
...
     /// Convenient access to set Appearance/representation
-    void setRepresentation(const DisplayPreference& rep) { 
+    void setRepresentation(VisualRepresentation rep) { 
         upd_Appearance().set_representation(rep); 
     };
     /// Convenient access to get Appearance/representation
-    const DisplayPreference& getRepresentation() { return 
-        (const DisplayPreference&)get_Appearance().get_representation(); 
+    VisualRepresentation getRepresentation() { return
+        get_Appearance().get_representation();
```
aseth1: Having this list of conveniences that set directly on the members of Appearance further negates reasons to separate surface and curve appearances, in my view.

I vote against these convenience accessors in favor of getting and setting an Appearance. That makes handling the notion of a shared Appearance more straightforward, I think.

aseth1:
I am done reviewing. Issues:

0. Is the Appearance property of a Geometry its Appearance or default appearance? If it is its default, how does the actual Appearance get set by the user/visualizer/GUI at display time?

1. `CurveAppearance` and `SurfaceAppearance` classes only provide two separate bags of properties inside of Appearance. Why is this helpful or necessary? The distinction doesn't seem to be used and they are exposed as members of the Geometry via convenience accessors.

2. It is unclear what is the purpose and usage of `setDecorativeGeometryTransform` and `setDecorativeGeometryAppearance`? 

3. Public descriptions of the classes and their properties could be more clear.

4. There is only one  example (`TugOfWar_Complete.cpp`) where a single appearance attribute is being set. I would like to see more use cases (e.g. for shared Appearance and `setDecorativeGeometry` methods above) and the XML file for a typical model as part of the PR.

OpenSim/Simulation/Model/Geometry.h
sherm1: > (Ajay) I vote against these convenience accessors in favor of getting and setting an Appearance. That makes handling the notion of a shared Appearance more straightforward, I think.

I agree.

aymanhab:
Great feedback, thanks all. I'll address the high level issues brought by @aseth1 first:
1. Up to version 3.3 and as of today every piece of Geometry has a single associated Appearance (as such it's still unnamed property). When we're able to have a name search for Appearance, this will turn into a String property with name "default_appearance" that binds the qualified name specified in the property to an actual Appearance object. At this point will have to answer the important questions that will arise. I didn't want to entangle the two issues and also wanted to work with the abstractions we all agreed upon (rather than go back to square 0).
2. It is necessary only because it scales easily,and because it is the right abstraction (I'd envision different tabs/groups for different items in the "bag" and name filtering based on prefix is more error prone). The bag can  grow to include shininess, lighting model properties, text font, text color, and can become unusable adding overhead and custom code to the GUI for little benefit.
3. Great point, the functions are used internally but they don't need to be in the public interface. I moved them to the protected section.
4. Agreed, I'll incorporate the suggestions from the PR feedback.
5. As described in 1 there's no shared appearance yet, the XML from the provided example follows:
'''
<FrameGeometry name="frame_geometry">
..
  <Appearance>
      
     <visible>false</visible>
  </Appearance>
</FrameGeometry>
..
<Brick name="..">
   <Appearance>
	
	<color>0 1 0</color>
   </Appearance>
</Brick>
..
<Cylinder name-" ">
	<Appearance>
		
		<SurfaceAppearance name="surface_appearance">
			
			<representation>2</representation>
		</SurfaceAppearance>
	</Appearance>
</Cylinder>
...

OpenSim/Simulation/Model/PhysicalFrame.cpp
aymanhab: Agreed if we can provide customization since otherwise all frames show the same and if on top of each other are practically useless. Outside the scope of this PR I believe.

OpenSim/Simulation/Model/Appearance.h
aymanhab: Good point I'll elaborate so we don't need to refer to DecorativeGeometry::Representation explicitly.

OpenSim/Simulation/Model/Appearance.h
aymanhab: Yes on both counts: 0 to hide, the name of the class is VisualRepresentation, the name of the Property is representation. 

Strings add their own problems (case senstivity, spaces that cause problems in parsing, there're standard GUI tools to map enums to Strings but not vice versa). Mappnig Enums to ints is the standard in most all programming contexts, why invent a new convention that's more error prone?

OpenSim/Simulation/Model/Appearance.h
aymanhab: For now I'm keeping only the properties supported by the old visualizer (VTK). When we switch we'll see what we can expose.

OpenSim/Simulation/Model/Appearance.h
aymanhab: I went back and forth on this, the main reason these convenience methods are here is that they were part of the old DisplayGeometry interface and so are used a couple dozen times in the GUI code base, when the new API is plugged in the GUI will deal with the SurfaceAppearance, CurveAppearance etc.

OpenSim/Simulation/Model/Appearance.h
aymanhab: :+1:

sherm1:
Ajay: please add comments to explain the decorative geometry methods (point 3 above).

aymanhab:
Changes implemented per dev. meeting, ready for review.

jenhicks:
@aseth1 will review Ayman's latest response.

jenhicks:
Fixes #480 
Fixes #482  
(Or at least deprecates them)

sherm1:
@aseth1 can you review this?

OpenSim/Simulation/Model/Geometry.h
```diff
@@ -180,8 +150,34 @@ class OSIMSIMULATION_API Geometry : public Component {
     /// Methods to support frame as a connection, implement Component interface
     void constructConnectors() override;
 
+protected:
+    /// Map this Geometry into a list of primitives aka SimTK::DecorativeGeometry 
+    /// and return the result in the passed in Array.
+    virtual void createDecorativeGeometry(
+        SimTK::Array_<SimTK::DecorativeGeometry>&) const = 0;
```
aseth1: Does it make sense to close off all derived classes of Geometry from extending `generateDecorations()`? I like that the task is simplified for derived Geometry, but is it reasonable to close them off completely from having a say in how they draw themselves based on the whether *fixed* decorations is requested or making use of display *hints*? 

One can imagine a `LinePathGeometry` as a functional and drawing replacement for GeometryPath, in which case we would want the geometry to draw differently for many of the `ModelDisplayHints` options.

OpenSim/Simulation/Model/Geometry.h
```diff
@@ -180,8 +150,34 @@ class OSIMSIMULATION_API Geometry : public Component {
...
+
 private:
-    
+    /// Compute Transform of this geometry relative to its base frame, utilizing 
+    /// passed in state. Both transform and body_id are set in the passed-in 
+    /// decorations as a side effect.
+    void setDecorativeGeometryTransform(
```
aseth1: I still don't really know what this method does. The comment is confusing since a geometry is fixed to a frame and that frame's base frame can be accessed directly, so what is being computed here? As a private method you don't need the doxygen style comments, since i) we don't doc private members in general and ii) even if we did allow some, you probably would not want these methods to be known, correct?

Is this method setting the transforms of the drawing primitives used by the visualizer? As a private method, a user doesn't need to know about it, but it may be helpful for whoever has to maintain it to have some clear tips.

OpenSim/Simulation/Model/Geometry.h
```diff
@@ -180,8 +150,34 @@ class OSIMSIMULATION_API Geometry : public Component {
...
+    /// Manage Appearance (how the Geometry is rendered) by applying Appearance 
+    /// from Geometry to DecorativeGeometry.
+    void setDecorativeGeometryAppearance(
+        SimTK::DecorativeGeometry& decoration) const {
+        decoration.setColor(get_Appearance().get_color());
+        decoration.setOpacity(get_Appearance().get_opacity());
+        if (get_Appearance().get_visible())
+            decoration.setRepresentation(
+                (VisualRepresentation)
+                get_Appearance().get_representation());
```
aseth1: You are not setting the Appearance of DecorativeGeometry, right? This method translates the appearance of the OpenSim::Geometry to the attributes of the underlying SimTK::DecorativeGeometry. This distinction has to be more clear otherwise it is very confusing.

If the underlying `SimTK::DecorativeGeometry` only has a `representation` and you are are short-circuiting the `SurfaceAppearance` accessor with `Appearance::get_Appearance()`, why isn't `representation` just a member of Appearance? If the underlying `DecorativeGeometry` has not distinction, what is the utility of having it here? It seems above and below you just need a `representation`. 

How is (will) texture (be) handled by `DecorativeGeometry`? I ask because I don't see where there is any distinction made between the underlying `DecorativeGeometry` (which has to understand *representation* and *texture*?), which makes me wonder why is there is a separation of Surface and Curve properties inside Appearance if the `DecorativeGeometry` doesn't distinguish them?

OpenSim/Simulation/Model/Geometry.h
```diff
@@ -232,6 +228,7 @@ class OSIMSIMULATION_API LineGeometry : public Geometry
         upd_end_point() = aPoint2;
     }
 
+protected:
     /** Virtual method to map LineGeometry to SimTK::Array of
     SimTK::DecorativeGeometry.  Appearance, Transforms are handled by base
     Geometry class. */
```
aseth1: In the context of this method's comment, it sounds like `LineGeometry::createDecorativeGeometry()` must call `Super::createDecorativeGeometry()` to handle Appearance and Transforms. What do you mean by handle them?

aseth1:
Is the sole purpose of Surface and Curve Appearance primarily to group properties so they can be presented in the GUI in different groups/tabs? If so, it might be more generally useful to have "groups" for properties in general. I can imagine for complex components (e.g. Muscle) it might be handy to have groupings (like "fiber_properties", "tendon_properties", etc...

I am OK with being being overruled, but it bothers me to define new classes only to group data  without any behavior (e.g. methods).  Are you envisioning them to have methods down the road, if so what might they be? I would also be satisfied with that.

OpenSim/Simulation/Model/Geometry.h
aymanhab: @aseth1 We have off-the-shelf Geometry that work based on xml properties and fully customized Geometry by using generateDecorations mechanism. Having a hybrid maybe useful down the line at some point but it'd be following the custom route. For off-the-shelf Geometry it's clean that what you put in XML is what you get without any meddling. Also makes the interface simpler IMHO.

OpenSim/Simulation/Model/Geometry.h
aymanhab: Since Geometry is responsible for computing the transform itself using the Frames, and the computation is identical since it depends only on Frames and the state, this common functionality is implemented in the base class. Your interpretation in the second paragraph is accurate and it will not show in doxygen because the method is private but is supposed to help developers maintaining this code.

OpenSim/Simulation/Model/Geometry.h
aymanhab: Correct, SimTK::DecorativeGeometry has a different set of options that we map to. Representation belongs to SurfaceAppearance since Curves can't have shading/lighting-model, or drawn in wire-frame etc. Good point on texture not handled by the DecorativeGeometry altogether or the API visualizer. We'll have to decide, later, if it will stay on the GUI side (you apply textures once to create pretty picture or movies) or be pushed to API (if/when the API visualizer and GUI converge).

OpenSim/Simulation/Model/Geometry.h
aymanhab: @aseth1 Agreed, even though the base class is doing nothing now, it may do in the future.

aymanhab: @aseth1 Making the Super call requires some implementation in the base class which can be empty but has the negative side effect that when you add a new kind of Geometry you don't know what to implement. I'm leaning toward keeping it as clear as possible with a pure virtual in the base class. I think this a place to apply the Impl naming convention, so we can leave the method pure virtual and change the name into implementCreateDecorativeGeometry() per #648

aymanhab:
@aseth1 Indeed the reason for the classes is grouping, it's like a struct rather than a class, unfortunately we don't support these in serialization and an earlier attempt to use internal classes in "ContactParameters" led to invalid XML syntax with "::" qualified tags. 
I sure agree that a more general grouping mechanism would be useful (Group by function, by commonly-used/advanced, ....) but that's outside the scope of this PR. Please let me know if you'd prefer opening an issue to start this discussion.

aseth1:
> Indeed the reason for the classes is grouping, it's like a struct rather than a class ...

Do you think it is better to remove the distinctions between the internals of Appearance (Surface vs Curve)  in favor of adding groups in later? The grouping of properties seems like a good idea and it may also be necessary to eliminate the use of `Set`'s.

The alternative is to leave `CurveAppearance` and `SurfaceAppearance` as classes and object properties of Appearance and then replace with groupings down the road. I prefer the first approach, because it is closer to the desired behavior and structure and it does not introduce two more classes that are exposed both in the API and via XML, which seems more to difficult to unwind once it is in place and being used.

I would rather see a single bag of properties in Appearance (no `CurveAppearance` and `SurfaceAppearance` classes) that then get grouped down the road for purpose of editing these properties via the GUI.

It would be good to get other opinions on this (@sherm1, @chrisdembia, @tkuchida, @msdemers

chrisdembia:
I don't have complete knowledge of the issues, but I prefer a single Appearance class. I don't feel that I've heard a convincing reason for the additional classes.

sherm1:
I'd be happy with a single Appearance class. We can add nuance later if needed. I think there are lots of important design issues left for this and the curve/surface issue is confusing things. The big issues for me are how defaults work and get overridden, how multiple objects share the same Appearance, how do we say things like "all bones should use this Appearance", etc.

chrisdembia:
Yeah those sound like big issues. IIRC, @aymanhab has another round of changes to make in the GUI before he's ready to approach those.

aymanhab:
All, thanks for your feedback. 
 I think the comments that I made 28 days back on this trail still hold. Sorry it has been a long time!
1. There's exactly one Appearance class, the CurveAppearance  and SurfaceAppearance are basically internal groupings of properties within it to help the GUI. As the developer who interacts most with Geometry as of now and in the short term, this grouping is very helpful to me as the number of available properties grows. Having a big bag of Properties that gets looked up and grouped based on name/prefix is next to impossible to maintain. I'm ok with changing names of these "internal" classes or making them nested (there might be an issue with the XML serialization) but I'm against flattening the insides of the class which is basically an implementation detail. 
2. Assuming all other issues are resolved, we should resolve the naming/nesting and merge to move on.
3. @aseth If grouping gets implemented in another way, it'll be very easy to update the XML since it's still within the single Appearance class.
4. I mainly wanted to separate the issues we "agreed" upon after months of discussion/back-and-forth before the next round (which will address Sherm's questions/concerns with defaults and reuse) If however that's not satisfactory, I'll be happy to leave this as is and continue working on the technology development/communication side.

5. A unique ID scheme and traversal mechanism that includes non-components will be needed to address Sherm's .concerns, we should have it on the schedule. I'll be happy to discuss.

Thanks all for your feedback and thoughts..

aseth1:
To make the distinction clear that `SurfaceAppearance` and `CurveAppearance` are not Appearances can you rename these to SurfaceProperties, etc..  or something that isn't Appearance? It is confusing to have internal classes that have parent class name as its root name, which sounds like they derive from `Appearance` in the current case. 

> Having a big bag of Properties that gets looked up and grouped based on name/prefix is next to impossible to maintain.

I don't understand why that is an issue now or in the near future. Here is the current list:
0. visible           (base)
1. color             (base)
2. opacity          (base)
3. representation (Surface)
4. texture_file       (Surface)
5. thickness      (Curve)

I don't think we are anywhere close to reaching the *impossible to maintain* point  (with these 5-6 properties) and we could have a much better long-term solution in property groupings. I would err on a single bag and use the future problem (if it ever comes to that) to increase priority on an issue to get property groups in place.

On the flip side,  I also don't think these two classes are sufficient to contain future Appearance properties nor provide adequate context for new properties. For example, where would properties of *fill*, or *font* go? Font type in a new TextAppearance? How about a *shadow* effect?  My sense is that the two groups will not be adequate. They also require tunneling through the Appearance to get to the properties of the internal groupings. A flat list would allow us to simply add the new appearance attributes we want to support in Appearance. A light weight solution to then group them makes more sense for displaying Appearance properties to users (say in separate tabs) via property groups.

From the client (visualizer) side of an `Appearance`, it also seems easier to use a flat list of properties. I can ask an Appearance if it `hasProperty(name)` and get is its value, check if it is default, etc.... when I draw. In the current scenario I have to first access two (and potentially more?) other objects (and know their type) and then ask if they have the appearance property that I can control.

>I'm ok with changing names of these "internal" classes or making them nested (there might be an issue with the XML serialization) but I'm against flattening the insides of the class which is basically an implementation detail.

At a minimum they should be renamed, but as object properties these classes are exposed in the API by the property macro defined accessors: `get_()`, `set_()`, `upd_()`, `getProperty_()`, .... and I don't like the idea of adding more classes to handle more attributes and their grouping. The class names are also exposed in XML, which will then require `updateFromXML` methods to handle intermediate incarnations if we do decide to switch to groups. That doesn't seem like hidden implementation details to me and seems to add a lot of unnecessary weight, which will make it more difficult to change going forward.

jenhicks:
I think that we should merge this PR so Ayman can move on with the end-to-end initial test. We seem to be going in circles  Then we can review and update the implementation as needed. @aymanhab can you make sure appropriate issues are files for the traversal scheme you mentioned above, as well?

aseth1:
I am happy to merge it without the `CurveAppearance` and `SurfaceAppearance` classes. We had discussed and agreed on a flat list of properties for appearance over a month ago and I think we should stick to it. It is the least cumbersome to handle down the road.

aymanhab:
Thanks @jenhicks & @aseth1 for the feedback.
1. I renamed the "internal" classes to SurfaceProperties and CurveProperties respectively to avoid confusion that these are derived classes. Hope this gets us past this issue. Thanks @aseth1 , so we have no SurfaceAppearance or CurveAppearance anywhere!
2. I also renamed the pure virtual method to begin with "implement" to follow the latest guidelines.
3. As to the properies contained in the Surface/CurveProperties, these are not too many as of now but here's a list of potential items (I expect some subset will be made available)
Surface:
========
Lighting model: Phong, Lambert
Colors: Specular, Emmissive
UV parametrization
Cast Shdows
Receive Shadows
Sidedness
Skinning
show-edges
Edge-colors
Color vertices
....
Text
=====
Font
Size
Style/3D
Billboarding
Screen or World coordinates
....etc.

With a limited real state in the Properties window and on screen in general these have to be grouped somehow or shown in a dialog with different tabs and I'm trying to avoid writing custom code to handle this  grouping, that's the part that can be hard to maintain based on my experience with Tools which has the exact same bag of properties issues.

Hope this explains, and thanks to all reviewers for their  time.

I opened issue #668 for the grouping of Properties, and there're a few open issues already regarding naming/lookup but I'll open one specific to visualization needs.

aseth1:
@aymanhab  Are you now proposing there will `TextProperties` inside `Appearance`? What next?

I gave a detailed response as to why these are not simply internal issues.  Can you explain why these sub containers simplify handling of Appearance - I just don't get it?

Would it not be better to focus on property groupings as a general reusable capability instead of inventing new classes simply to group properties?

sherm1:
Could we agree that Appearance will need to be substantially revised once we've had a chance to experiment with it? We could make a 4.0 issue to re-engineer Appearance. I don't think we have yet come up with a satisfactory design -- so far we have plenty of complaints but no good solutions. This PR is probably not the right venue for redesign. I think if it needs to be merged just so we can make *some* progress that would be OK provide there is another issue.

aseth1:
@sherm1 Yes! Ayman and I were just discussing it.

aseth1:
I understand using the "internal" classes for organizing the properties and I can accept them as a temporary stand-in solution. My main concern is that they do lead to API and XML changes that will then have to be handled again when the mechanism is revamped. My preference is to avoid that, but Ayman assures me it would not be a big deal for him.

Talking with Ayman, it is also more apparent that the property grouping (although a good general idea) may not be a good fit for this problem. We discussed alternative data structures (like ValueDictionary) that could be far more flexible and expansive without impact the modeling API.
We agreed to work on these ideas through an issue as @sherm1 suggests.

I will merge it.

aseth1:
AppVeyor is not running the python wrapping tests, but I don't see a related error (although the build itself is not passing). @aymanhab Any ideas? I restarted the build to see if it was just an anomaly.

aymanhab:
Thanks Ajay I'll investigate if it fails again.

aymanhab:
All tests passed on rerun, please merge when you have a chance. Thanks.

aseth1:
Here you go- merged!

aseth1:
@aymanhab did you create an issue a re-engineer of Appearance as suggested by @sherm1?

aymanhab:
Thanks much @aseth1 

Issues that has been opened related to this PR
#675 #674 #668

PR #617, opened 2015-09-11
A home for pseudocode

chrisdembia:
As we've discussed earlier this year, it is important to have examples of pseudocode, etc that we view as targets for 4.0. The attribute of these example targets/goals don't compile yet.

I'd like to create a folder in the repo that houses .cpp, .py, etc files containing pseudocode that contains what we want to have working in the future.

The folder should contain a mix of files: some files will compile, and some will not. Files that are intended to compile (that are further along in development) should be named `future*.cpp`, while files with less mature pseudocode that does not compile should be named `pseudo*.cpp`. 

The executables are *not* test cases, and they are only built if you explicitly build them.

This allows us to create PR's containing simply pseudocode, and thus allows us to have PRs simply to discuss designs.

The directory should be somewhat organized; different projects/designs can have their own files.

I started us off with some brainstorming for how we might replace the metabolics probes (@tkuchida).


OpenSim/Tests/Future/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+# This folder holds example code that shows what we *want* to be able to do
+# with OpenSim in the future. There is an executable target for every
+# C++ file whose name begins with `future`, but these executables are not
+# built by default (not part of the "all" target).
+
+# These executables are *not* tests.
```
tkuchida: …then why do they live in the Tests directory? What about OpenSim/Future, or perhaps OpenSim/Examples/Future if you think most of the pseudocode will eventually become examples.

tkuchida:
Cool. I like this idea and think it could facilitate discussions on high-level design.

chrisdembia:
I'm fine with `OpenSim/Future`. I actually expect these to most likely become test cases that test expected usage, etc. The code here could become actual components too. Putting the directory in `OpenSim/Future` might clarify that the code could end up in examples, tests, or in one of the opensim libraries.

jenhicks:
Change name of directory to OpenSim/Sandbox. @aseth1 will review first example. Should add a readme as well.

chrisdembia:
Ready for review again; I moved OpenSim/Future to Sandbox.

OpenSim/Sandbox/futureMuscleMetabolicsResponse.cpp
```diff
@@ -0,0 +1,210 @@
...
+    ComplexResponse() {
+        constructInfrastructure();
+        constructProperty_strength(3.0);
+    }
+    double getTerm1(const State& s) const {
+        if (!isCacheVariableValid(s, "term_1")) {
+            const auto& coord = getConnectee<Coordinate>("coord");
+            const auto value = coord.getValue(s);
+            setCacheVariableValue(s, "term_1", get_strength() * value);
+            markCacheVariableValid(s, "term_1");
```
aseth1: `setCacheVariableValue` automatically marks as valid.

OpenSim/Sandbox/futureMuscleMetabolicsResponse.cpp
```diff
@@ -0,0 +1,210 @@
...
+    AggregateResponse() { constructInfrastructure(); }
+// TODO want to use list property, but that makes clones and inputs/outputs
+// don't get copied yet.
+//    OpenSim_DECLARE_LIST_PROPERTY(responses, ComplexResponse,
+//            "for individual coordinates.");
+    // Temporary solution:
+    std::vector<std::shared_ptr<ComplexResponse>> responses;
+
+    // TODO how do we propagate this scaling_factor to the ComplexResponses?
+    OpenSim_DECLARE_PROPERTY(scaling_factor, double, "Affects each coord.");
```
aseth1: Now (as of #634) that each component can have a parent, you can have the child component check if it has a parent, and if so to get its scaling_factor.

aseth1:
Done reviewing. This example is great @chrisdembia!
Some small changes suggested (like removing unnecessary `markCacheVariableValid` calls).

I will aim to resolve the issues of copying Inputs/Outputs with a new mechanism for defining Input and Outputs for a Component soon.

chrisdembia:
Thanks @aseth1; ready for review/merge again.

chrisdembia:
Thank you!

PR #619, opened 2015-09-14
Removed unwarranted hyperlinks to OpenSim::Set

### LINKED ISSUE #601: [doxygen] Remove unwarranted hyperlinks to Set class
[Here's an example](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Object.h#L264).

tkuchida:
Searched for `" Set "` in header files and prefixed with `%` where appropriate to remove hyperlinks.

Fixes #601
[ci skip]

### DIFF
--- a/OpenSim/Actuators/TorqueActuator.h
+++ b/OpenSim/Actuators/TorqueActuator.h
@@ -98,7 +98,7 @@ public:
     // Uses default (compiler-generated) destructor, copy constructor, copy 
     // assignment operator.
     
-    /** Set the 'axis' property to the supplied value; frame is interpreted
+    /** %Set the 'axis' property to the supplied value; frame is interpreted
     according to the 'torque_is_global' property. **/
     void setAxis(const SimTK::Vec3& axis) 
     {   set_axis(axis); }
@@ -106,7 +106,7 @@ public:
     const SimTK::Vec3& getAxis() const 
     {   return get_axis(); }
 
-    /** Set the 'torque_is_global' property that determines how to interpret
+    /** %Set the 'torque_is_global' property that determines how to interpret
     the 'axis' vector; if not global (Ground frame) it is in body A's frame. **/
     void setTorqueIsGlobal(bool isGlobal) 
     {   set_torque_is_global(isGlobal); }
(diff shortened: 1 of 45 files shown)


OpenSim/Actuators/TorqueActuator.h
```diff
@@ -114,18 +114,20 @@ class OSIMACTUATORS_API TorqueActuator : public ScalarActuator {
...
-    /** Set the 'optimal_force' property. **/
+    /** %Set the 'optimal_force' property. **/
     void setOptimalForce(double optimalForce)
     {   set_optimal_force(optimalForce); }
     /** Get the current value of the 'optimal_force' property. **/
     double getOptimalForce() const override // Part of Actuator interface.
     {   return get_optimal_force(); }
 
-    /* Set the bodies to which this actuator applies torque. */
+    /** %Set the first body to which this actuator applies torque. */
```
chrisdembia: Nice catch

chrisdembia: Hmm..is there really a "first" and "second" body?

OpenSim/Common/Set.h
```diff
@@ -251,7 +251,7 @@ friend std::ostream& operator<<(std::ostream &aOut,const Set<T> &aSet)
 //=============================================================================
 //_____________________________________________________________________________
 /**
- * Set whether or not this set owns the memory pointed to by the pointers
+ * %Set whether or not this Set owns the memory pointed to by the pointers
```
chrisdembia: haha nice

chrisdembia:
I made one comment that you can consider addressing; feel free to leave it as is. Merge at will.

OpenSim/Actuators/TorqueActuator.h
tkuchida: >Hmm..is there really a "first" and "second" body?

From the class description: "A TorqueActuatorr [sic… I'll fix the typo] applies equal and opposite torques on the two bodies (bodyA and B) that it connects."

tkuchida:
>I made one comment that you can consider addressing; feel free to leave it as is. Merge at will.

Thanks. I'll add `(bodyA)` and `(bodyB)` to avoid any confusion.

OpenSim/Actuators/TorqueActuator.h
chrisdembia: I guess I meant: are the bodies actually ordered in a meaningful way? "first" and "second" might imply that one body "comes first", like maybe it's closer to the root node of the tree.

I don't think it's a big deal. We use a mix of `body*.1` and `body*.A` throughout the code I think.

tkuchida: >I guess I meant: are the bodies actually ordered in a meaningful way?

It matters which body is bodyA for two reasons: the sign and the frame in which the torque is expressed. I don't think referring to bodyA as *the first of two bodies to which torque is applied* is confusing.

PR #620, opened 2015-09-14
Data table based on ValueArrayDictionary

klshrinidhi:
Files to review:
* Common/ValueArray.h
* Common/ValueArrayDictionary.h
* Common/DataTable.h
* Common/TimeSeriesTable.h
* Examples/DataTable/example.cpp


OpenSim/Common/ValueArray.h
```diff
@@ -0,0 +1,84 @@
...
+class ValueArray : public AbstractValueArray {
+public:
+    template<typename U>
+    using Value = SimTK::Value<U>;
+
+    ValueArray* clone() const override {
+        auto copy = new ValueArray{};
+        copy->_values = _values;
+        return copy;
+    }
```
sherm1: Better to use the copy constructor for implementing `clone()`:
```c++
ValueArray* clone() const override {
    return new ValueArray(*this);
}
```
That avoids the default initialization and assignments, meaning value type `T` need only be `CopyConstructible`.

OpenSim/Common/ValueArrayDictionary.h
```diff
@@ -0,0 +1,77 @@
...
+#include "OpenSim/Common/ValueArray.h"
+
+#include <memory>
+
+namespace OpenSim {
+
+class ValueArrayDictionary {
+private:
+    using Dictionary = std::map<std::string, 
+                                std::shared_ptr<AbstractValueArray>>;
```
sherm1: Why shared? That could cause unexpected behavior if the dictionary is copied, I think.

chrisdembia: That could be the *expected* behavior. Side note: swig can automatically wrap shared_ptr's, but doesn't know what to do with unique_ptr's.

sherm1:
Done reviewing. I did not scrutinize every line of code but the overall structure and style look beautiful to me. The code seems much more compact and expressive now which I take to be a good sign! Here are a few minor things I spotted:
- The ValueArray classes need doxygen comments; they are currently uncommented.
- Data member names need to be changed for consistency and to follow the OpenSim coding guidelines: they should be `camelCaseNames` with a *leading* underscore or leading `m_`. Trailing underscore is not an option. Currently in the code I see both leading and trailing underscores being used. We reserve the STL-style `names_with_underscores` for Property names. 
- You should put yourself as "Author" in the appropriate file header blocks. That's not just for credit; it helps people figure out who to ask and what style to expect. You can't achieve anonymity even if it were a good idea -- GitHub puts your name on every line.
- It is not clear to me that shared data is appropriate for the ValueArrayDictionary. That would mean that copying followed by subsequent modifications to the copy would affect the original; is that the intention? If space and performance are a worry you could use CloneOnWritePtr instead of shared_ptr and get deep copy semantics with shallow copy performance (unless someone writes to the copy).

A thought: If the ValueArray<T> class will get a lot of use in code and examples it might be worth making it a container by promoting the interators and useful methods from the internal `std::vector<T>`, e.g. 
```c++
    using iterator=std::vector<T>::iterator; 
    iterator begin() {return _values.begin();}
    // etc.
```
Could do something similar for AbstractValueArray but that would be a little more complicated. I'm not sure if that will get enough use and the `size()` and `operator[]()` methods are sufficient.

Issue #593, opened 2015-09-01
Reduce duplication of swig interface files


aymanhab:
As of now (09/15) the Python and Java bindings use two unrelated swig interface files, it would be good to have them share resources (e.g. common includes when possible) so addition/removal/instantiation of classescan be done in one place, and we can easily support more bindings (e.g. Matlab, Go, C# or whatever). The target language specific parts should stay in separate files to avoid confusion and limit scope of changes.

chrisdembia:
+1

PR #623, opened 2015-09-17
Wrapping cleanup 

aymanhab:
Gradually moving Bindings to a different folder and reusing swig interface files to make Python and Java/Matlab interfaces consistent.


aymanhab:
Should address #593

Bindings/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+
+if(BUILD_PYTHON_WRAPPING)
+    find_package(SWIG 3.0.5 REQUIRED)
+    add_subdirectory(Python)
+endif()
+if(BUILD_PYTHON_WRAPPING)
+    find_package(SWIG 3.0.5 REQUIRED)
+    add_subdirectory(Java)
+endif()
```
chrisdembia: The above needs to be modified and could be consolidated.

chrisdembia:
@aymanhab it'd be nice to make these changes in a series of distinct PR's: moving OpenSim/Wrapping to Binding could be separate from breaking up the interface files. Well, maybe we should just review what's already been done and merge it (perhaps into a feature branch) before splitting up more of the interface files.

Also, could we maybe think of a different name for the "common" folder? I think it can be confused with the OpenSim Common library and directory. One option is to just put all these files in the OpenSim/Bindings directory (and not create a separate dir).

aymanhab:
@chrisdembia I don't mind the split into multiple pull requests but in my mind the only actual change is the rename/move, the other changes are just no-op so that future changes are done in one place (e.g. python specific changes)

Bindings/CMakeLists.txt
aymanhab: That's possible though I can imagine different languages requiring different swig versions or Python wrapping requiring other packages (e.g. numpy, scipy, ..) that will be added separately. Is it worth the tradeoff to avoid duplicating the find_package line in CMake? I'm open.

chrisdembia: Oh good point; I hadn't thought of that.

Right now both Java and Python do require other packages, but those are handled in their own directories; should those be moved up here?

Also, if the `if() ... endif()` is here, we should remove them from the Java and Python subdirs so we can unindent all that cmake code; which would be nice :)

As for the duplication, I suggest we address that when we encounter requiring different swig versions for different wrapping.

Nonetheless there is still a bug above; the second `BUILD_PYTHON_WRAPPING` should be `BUILD_JAVA_WRAPPING`.

aymanhab:
@chrisdembia Looking fine, just had to rearrange headers in one case due to API changes, waiting for build. Also noticed all concrete joint types have no default constructor in Java, likely due to inherited constructors. Is that true for Python too? Thanks.

chrisdembia:
> Also noticed all concrete joint types have no default constructor in Java, likely due to inherited constructors. Is that true for Python too?

Ah good point. That is unrelated to this PR but I think you're spot-on; this would be true in Python as well (I did not test). @aseth1. How urgent are they? Do you use them in Java? We could add them as part of this PR (it would be fairly straightforward), though it'd be preferable to have a separate PR for that.

chrisdembia:
@aymanhab looks good to me. Also, good catch on renaming the VS folders. Feel free to merge when you're ready.

Let me know about timing for an udpated python PR.

aymanhab:
Thanks @chrisdembia

PR #633, opened 2015-09-21
Add test for ExpressionBasedBushingForce

### LINKED ISSUE #618: ExpressionBasedBushingForce has no test cases
AFAIK the main user of this Force is @msdemers 
It would be great if you could contribute a test case (or two) based on your work.

msdemers:
Added a test case for ExpressionBasedBushingForce withing testForces.  Also did a little bit of cleanup throughout testForces, specifically removing declared but unused variable. Confirmed that it works on linux, OS X, and Windows 10. Fixes #618

### DIFF
--- a/OpenSim/Simulation/Test/testForces.cpp
+++ b/OpenSim/Simulation/Test/testForces.cpp
@@ -791,9 +791,13 @@ void testExpressionBasedBushingForce()
     // Setup OpenSim model
     Model *osimModel = new Model;
     osimModel->setName("ExpressionBasedBushingTest");
-    //OpenSim bodies
+    osimModel->setGravity(gravity_vec);
+
+    // Create ball body and attach it to ground
+    // with a vertical slider
+
     const Ground& ground = osimModel->getGround();
-    
+
     OpenSim::Body ball("ball", mass, Vec3(0), mass*SimTK::Inertia::sphere(0.1));
     ball.addMeshGeometry("sphere.vtp");
     ball.scale(Vec3(ball_radius), false);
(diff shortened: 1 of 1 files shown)


OpenSim/Simulation/Test/testForces.cpp
```diff
@@ -782,6 +774,131 @@ void testFunctionBasedBushingForce()
...
+    ball.scale(Vec3(ball_radius), false);
+    
+    SliderJoint sliderY("", ground, Vec3(0), Vec3(0,0,Pi/2), ball, Vec3(0), Vec3(0,0,Pi/2));
+    
+    double positionRange[2] = {-10, 10};
+    // Rename coordinates for a slider joint
+    CoordinateSet &slider_coords = sliderY.upd_CoordinateSet();
+    slider_coords[0].setName("ball_h");
+    slider_coords[0].setRange(positionRange);
+    slider_coords[0].setMotionType(Coordinate::Translational);
```
tkuchida: Why is this line necessary? Aren't coordinates associated with `SliderJoint`s necessarily translational?

OpenSim/Simulation/Test/testForces.cpp
```diff
@@ -782,6 +774,131 @@ void testFunctionBasedBushingForce()
...
+    osimModel->addJoint(&weld);
+    
+    // create an ExpressionBasedBushingForce that represents an
+    // uncoupled, linear bushing between the ball body and welded base body
+    Vec3 rotStiffness(0);
+    Vec3 transStiffness(stiffness);
+    Vec3 rotDamping(0);
+    Vec3 transDamping(0);
+    
+    ExpressionBasedBushingForce spring("base_body", Vec3(0), Vec3(0), "ball", Vec3(0), Vec3(0), transStiffness, rotStiffness, transDamping, rotDamping);
```
tkuchida: Violation of [80 character line limit](https://github.com/opensim-org/opensim-core/blob/master/CONTRIBUTING.md#each-line-of-text-should-be-at-most-80-characters) :rabbit2:

OpenSim/Simulation/Test/testForces.cpp
```diff
@@ -782,6 +774,131 @@ void testFunctionBasedBushingForce()
...
+    RungeKuttaMersonIntegrator integrator(osimModel->getMultibodySystem() );
+    integrator.setAccuracy(1e-6);
+    Manager manager(*osimModel,  integrator);
+    manager.setInitialTime(0.0);
+    
+    double final_t = 2.0;
+    double nsteps = 10;
+    double dt = final_t/nsteps;
+    
+    for(int i = 1; i <=nsteps; i++){
```
tkuchida: Violation of [pre-increment operator rule](https://github.com/opensim-org/opensim-core/blob/master/CONTRIBUTING.md#always-use-pre-increment-and-pre-decrement-operators-when-you-have-a-choice) :turtle:

OpenSim/Simulation/Test/testForces.cpp
```diff
@@ -782,6 +774,131 @@ void testFunctionBasedBushingForce()
...
+        
+        // get the forces applied to the base and ball
+        Array<double> model_force = spring.getRecordValues(osim_state);
+        
+        // compute the expected force on the ball
+        double analytical_force = -stiffness*height;
+
+        // check analytical force corresponds to the force on the ball 
+        // in the Y direction, index = 7
+        ASSERT_EQUAL(analytical_force, model_force[7], 2e-4);
```
tkuchida: Why 2e-4 here but 1e-4 above? Where did these tolerances come from and how do they relate to the integration accuracy?

tkuchida:
Great to see a test case that compares to an analytic solution :+1:

msdemers:
@tkuchida  Thanks, but I can't take all the credit for that!  I just emulated most of the test cases in testForces, many of which use a similar system and analytic solution.  In fact, regarding your questions about error tolerances and comparison tests, I borrowed those from other tests in this suite!

msdemers:
@tkuchida On reviewing the test tolerances used throughout testForces, I think their values are reasonable for the integrator tolerance chosen (a useful exercise!).  RGM integrators have O(h^5) error per step and O(h^4) cumulatively. Hence err_q_step = 1e-6 yields err_q_cumulative = alpha*1e-5, and I don't know alpha. That means I'd expect the computed positions to be between 1 and 2 orders of magnitude larger than the stepwise integrator tolerance. The bushing force is linear, with K scaling q to force. Therefore, I'd expect err_force_cumulative ~= K*err_q_cumulative.  In this case, K = 10 so I suppose err_force = 1e-3 should be allowable.  This test, as of now, is more strict, with err_force_cumulative = 2e-4. I can be less strict and increase err_force to match err_q if you like, or I suppose this suggest we have wiggle room to tighten tolerance on the kinematic error. 

Thoughts?

msdemers:
ooops.  Had to edit the message to escape all the "\*" and such.  Ended up with a lot of italicized text instead.

msdemers:
Per @tkuchida recommendations, cleaned up issues to adhere to code style requirements.

tkuchida:
@msdemers Thanks for the explanation. I don't have strong feelings about the tolerances; it just wasn't obvious to me why those particular values were chosen. Looks great with the latest commit. :+1:

aseth1:
Looks good to me too. 

@msdemers It did remind me that we should have cases that are equivalently simple for the rotational deflection of the bushings. Maybe you'll add that (as another PR) since you are on roll :stuck_out_tongue_winking_eye:

aseth1:
Added #638 for the future. Merging now. Thanks!

Issue #610, opened 2015-09-10
Connector::findAndConnect() should be removed.


aseth1:
Currently connector dependencies are found by traversing from a root Component. This is incorrect and leads to potential name collisions (non-unique names)  for Components that have locally unique named subcomponents. If the connectee_name is not a qualified pathname, then it MUST be a subcomponent of this component (the one with the Connector).

chrisdembia:
@klshrinidhi suggested a single `connect()` method with overloads for `Object` and `std::string`. `connect(string)` will do what `findAndConnect()` had done.

Issue #473, opened 2015-06-05
testInitState fails on OSX, probably due to memory leaks


chrisdembia:
Related to #393.

Also, sometimes testComponents fails as well.

chrisdembia:
```
17: Loaded model PushUpToesOnGround from file PushUpToesOnGroundWithMuscles.osim
17: Model unable to assemble: AssemblySolver::assemble() Failed: SimTK Exception thrown at Assembler.cpp:871:
17:   Method Assembler::assemble() failed because:
17: Optimizer failed with message: SimTK Exception thrown at InteriorPointOptimizer.cpp:261:
17:   Optimizer failed: Ipopt: Restoration failed (status -2)
17: Assembly error tolerance achieved: 5.91402810057318e-09 required: 1e-10.
17: Model relaxing constraints and trying again.
17: Model unable to assemble: AssemblySolver::assemble() Failed: SimTK Exception thrown at Assembler.cpp:871:
17:   Method Assembler::assemble() failed because:
17: Optimizer failed with message: SimTK Exception thrown at InteriorPointOptimizer.cpp:261:
17:   Optimizer failed: Ipopt: Restoration failed (status -2)
17: Assembly error tolerance achieved: 1.36861858651693e-10 required: 1e-10.
17: Model relaxing constraints and trying again.
17: *********************** testMemoryUsage ***********************
17: MODEL: PushUpToesOnGroundWithMuscles.osim uses 11516KB
17: 17872KB change in memory use after 100 state initializations.
17: Approximate leak size: 178.72KB or 1.55193% of model size.
17: Average initialization time: 623.911ms
17: testInitState failed: 
17: Exception:
17:   testMemoryUsage: state initialization leak > 0.5% of model memory
17:   footprint.
17:   file= /Users/chris/repos/opensim/1/opensim-core/OpenSim/Simulation/Test/testInitState.cpp
17:   line= 195
17: 
17: 
1/1 Test #17: testInitState ....................***Failed   63.31 sec

0% tests passed, 1 tests failed out of 1

Total Test time (real) =  63.31 sec

The following tests FAILED:
         17 - testInitState (Failed)
```

sherm1:
It's odd that the output above says "relaxing constraints and trying again" yet on the second try it still has a required tolerance of 1e-10. And it came **so** close at 1.4e-10! If it had really relaxed the constraint tolerance, say to 1e-9 it would have worked. Thoughts, @aseth1?

[...]

PR #634, opened 2015-09-22
Connectors to local frame does not require component name qualifier

aseth1:
Addresses issue 1. remaining from #589 where the Joint names were prefixed to the connectee_names of the offset frames to avoid conflicts with other Joints that have similarly named offsets. For example, both the hip and knee from existing model files will add a femur_offset on as the child (of the hip) and the other as the parent (of the knee) of their respective Joints. Given that the offsets are subcomponents of the respective Joints there should be no conflict whatsoever, except that `Connector<T>::findAndConnect() always resolves beginning at the root Component. I was running into the same issues with the offset frames fro BushingForces and Constraints, and so decided to resolve it separately here.

1. The current solution is also temporary but is moving in the right direction. It explicitly tries to find the Component locally (which defined as either a child or a sibling of the current component)  before finding from the root. I agree with a longer term solution that gets rid of findAndConnect() (#610) but this isn't it.

2. Adds methods for accessing a Component's parent. I needed that so I could find siblings. I would prefer a solution without the back pointer, but could not come up with any.

3. The `Component::hasSystem()` method is part of my Component changes in my other branch (using frames in Forces and Constraints) but, I thought it was would be better here since I am already making changes to Component.

4.  Added a check that if multiple components satisfy a Connector then an exception is thrown and the first component is not simply used. At least, the find is more strict.


OpenSim/Common/Component.cpp
```diff
@@ -1011,7 +1046,7 @@ void Component::extendRealizeAcceleration(const SimTK::State& s) const
 const SimTK::MultibodySystem& Component::getSystem() const
 {
     if (_system.empty()){
```
sherm1: could use the new `hasSystem()` here

OpenSim/Common/Component.h
```diff
@@ -1531,23 +1554,30 @@ template <class T> friend class ComponentMeasure;
...
     const AbstractConnector* findConnector(const std::string& name) const;
 
     const StateVariable* findStateVariable(const std::string& name) const;
 
+    /** Access the parent of this Component.
+        An exception is thrown if the Component has no parent. */
+    const Component& getParent() const;
+
+    /** %Set this Component's reference to its parent Component */
+    void setParent(const Component& parent);
```
sherm1: will you need a `hasParent()`?

OpenSim/Common/Component.cpp
```diff
@@ -797,27 +830,29 @@ void Component::constructOutputForStateVariable(const std::string& name) {
 
 // Include another Component as a subcomponent of this one. If already a
 // subcomponent, it is not added to the list again.
-void Component::addComponent(Component *aComponent)
+void Component::addComponent(Component* component)
```
sherm1: ah ... so nice to see those name changes!

sherm1:
Done reviewing. Not sure if I'm grasping all the implications but it looks good to me.

OpenSim/Common/Component.cpp
```diff
@@ -178,14 +178,33 @@ void Component::connect(Component &root)
         AbstractConnector& connector = upd_connectors(ix);
         connector.disconnect();
         try{
-            connector.findAndConnect(root);
+            const std::string& compName = connector.get_connectee_name();
+            std::string::size_type front = compName.find("/");
+            if (front != 0) { // local name
+                // Local is considered (1) children 
+                const Component* comp = findComponent(compName);
+                if (!comp) {// and (2) siblings(same level)
```
tkuchida: I don't understand these comments ("Local is considered children and siblings"). Perhaps clarify for posterity? :camel:

OpenSim/Common/Component.cpp
aseth1: Yes!

OpenSim/Common/Component.h
aseth1: Very likely. Will add it.

OpenSim/Common/Component.cpp
aseth1: Will do.

aymanhab:
@aseth1 The low level implementation looks fine, the main high level issue I have is that we're building on/beefing up the tree of "Components" (by adding Parent, etc.) while there was a plan to move this tree to the "Object" level and make a tree based on Properties that have a builtin naming mechanism etc. as discussed with @sherm1 and would be a more general solution. Is that still on the table? If so I don't know how the Component level solution helps with the more general problem.

aseth1:
@aymanhab We can move the solution to the Object level if we agree on that.
To me the tree should be composed of Components (only) to represent the components of a physical model and its dynamics. From what I recall from our discussion in the dev meeting, was that the tree would still be composed of Components, but that Objects (specifically object property list) would be allowed to be leafs, so that we can find Objects that Components are dependent on (to satisfy connectors).
I have some reservations about doing that only to enable connectors to Appearances. 
Do we agree, it is beyond the scope of this PR?

aseth1:
@sherm1 @tkuchida updates based on the review were made.

aymanhab:
@aseth1 It boils down to this, If you envision this work will be completely redone at Object level then we should probably avoid duplicating the effort and do the more general solution once and be done with it. If you see this refactored to move up to Object then I agree it's a step in the right direction. In that spirit, it'd be great to explain what facilities were provided by Component to make this search possible (since looking at code/deltas doesn't provide the high level picture). Thanks.

sherm1:
Looks good, Ajay. I'm just waiting for builds to complete.

aseth1:
@aymanhab I am pretty sure this will have to be redone with a more holistic and efficient approach, in any case, but that is a bigger project and I need something working to get the integration of Frames done. 

There were very limited facilities added to enable the desired behavior, which I described in the summary at the beginning of the PR. Basically, I used existing capabilities of `Component::findComponent()` to find a dependency among the component's subcomponents (i.e. children), Then I added the ability to access the parent, which allowed us to find sibling Components in the same way. That's pretty much it.

sherm1:
@aseth1, there was one build failure [here](https://travis-ci.org/opensim-org/opensim-core/jobs/81637226) -- testInitState is complaining about a memory leak when built with gcc in the merged pr.

tkuchida:
> testInitState is complaining about a memory leak when built with gcc in the merged pr.

"Approximate leak size: 0KB or -nan% of model size."

chrisdembia:
I think it might be sporadic...it should be restarted. This is the same as issue #473  .

sherm1:
Still failing the same way:
```
Updating Model file from 10901 to latest format...
Loaded model arm26 from file arm26.osim
*********************** testMemoryUsage ***********************
MODEL: arm26.osim uses 0KB
0KB change in memory use after 100 state initializations.
Approximate leak size: 0KB or -nan% of model size.
Average initialization time: 0.7ms
testInitState failed: 
Exception:
  testMemoryUsage: state initialization leak > 0.5% of model memory
  footprint.
  file= /home/travis/build/opensim-org/opensim-core/OpenSim/Simulation/Test/testInitState.cpp
  line= 195
```

tkuchida:
> Still failing the same way

Could it be an issue with how `getCurrentRSS( )` reports memory usage on Linux? There's some relevant discussion [here](http://unix.stackexchange.com/questions/34795/correctly-determining-memory-usage-in-linux): "The RSS value doesnt include shared memory. Because shared memory isn't owned by any 1 process, top doesnt include it in RSS." …though I don't know how that could explain a *sporadic* failure on Travis.

chrisdembia:
Oh interesting...

aseth1:
@sherm1 @chrisdembia @tkuchida are we good to merge?

sherm1:
> are we good to merge?

Yes, all good now Ajay. Merge it!

PR #636, opened 2015-09-22
Add in missing overrides.

chrisdembia:
Also, did *not* add in overrides for constructProperties(), since we may
make it a non-virtual method.

Also removed 'virtual' from declarations that now contain 'override'.

We should wait until Thursday to discuss when this can be merged.


sherm1:
Here goes ...

PR #637, opened 2015-09-23
Partially address issue #473 when testInitState sporadically fails (seen in #634)

aseth1:
due to the memory footprint estimated by getCurrentRSS() on Linux being incorrect at 0KB. The model_size cannot be zero.  This in turn yields a perplexing error message that the test failed because of zero memory usage is not less than 0.5%, but it is the relative leak size (normalized by model_size) that is NaN, caused by dividing by the erroneous model_size of zero. Any comparison with a NaN fails.

The "fix" here is to ASSERT when the the erroneous model_size estimate of 0 occurs before the downstream comparison/test. It won't fix the sporadic failures, but it will provide a more meaningful reason for the failure.

### DIFF
--- a/OpenSim/Simulation/Test/testInitState.cpp
+++ b/OpenSim/Simulation/Test/testInitState.cpp
@@ -154,9 +154,23 @@ void testMemoryUsage(const string& modelFile)
     // Setup OpenSim model
     // base footprint
     size_t mem0 = getCurrentRSS( );
-    Model model(modelFile);
+    size_t model_size{ 0 };
 
-    size_t model_size = getCurrentRSS( )-mem0;
+    Model model;
+
+    int ntries = 0;
+    while ((model_size == 0) && (ntries++ < 10)) {
+        model = Model(modelFile);
+        model_size = getCurrentRSS() - mem0;
+    }
+
+    if (ntries > 1) {
+        cout << "testMemoryUsage: required " << to_string(ntries);
+        cout << " load attempts to record a nonzero model_size." << endl;
+    }
+
+    cout << "*********************** testMemoryUsage ***********************" << endl;
+    cout << "MODEL: " << modelFile << " uses " << model_size / 1024.0 << "KB" << endl;
 
     ASSERT(model_size > 0, __FILE__, __LINE__,
         "testMemoryUsage: model size was found to be zero.\n"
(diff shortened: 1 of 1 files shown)


chrisdembia:
Looks good to me.

tkuchida:
I agree that this change is an improvement, but the test will still fail on Travis in the event that the memory footprint estimated by `getCurrentRSS()` is incorrect. Can we try to avoid the nuisance of having to restart the Travis build whenever it decides to act up by, perhaps, loading the model again if `model_size==0`?

aseth1:
>  Can we try to avoid the nuisance of having to restart the Travis build whenever it decides to act up by, perhaps, loading the model again if model_size==0?

Great idea. The current commit attempts to do this. Will try reloading up to 10 times before finally failing the ASSERT.

tkuchida:
Looks great! :+1:

aseth1:
@chrisdembia can you take a quick look and then merge if you are happy.

chrisdembia:
Perhaps a comment in the test about why this was necessary? Or perhaps simply refer to relevant github issues in a comment. Then merge.

aymanhab:
Is getCurrentRSS is part of some standard library that is supposed to work cross platform and it doesn't? If not then we should enable the test only on supported platforms so that users who get the source eventually don't run into this issue, I think.

chrisdembia:
> Is getCurrentRSS is part of some standard library that is supposed to work cross platform and it doesn't?

This is my hypothesis but I cannot be sure that it isn't a problem with our code base. I've spent a lot of time investigating this but I am not ready to say this hypothesis is confirmed.

I agree we need a more long-term cross-platform solution. One possible solution is to get rid of most of the memory leaks so it's less of a concern overall; @klshrinidhi has expressed interest in leak-squashing I think.

tkuchida:
> Is getCurrentRSS is part of some standard library that is supposed to work cross platform and it doesn't?

[Here's the code for `getCurrentRSS()`](http://nadeausoftware.com/articles/2012/07/c_c_tip_how_get_process_resident_set_size_physical_memory_use#Code). The implementation is platform specific. As discussed [here](http://unix.stackexchange.com/questions/34795/correctly-determining-memory-usage-in-linux), the issue might be the exclusion of shared memory in the reported RSS.

aseth1:
@aymanhab getRSS.h is not a standard lib header. It is 3rd party (creative commons license) code for estimating process memory usage on multiple platforms (delegating to each platform's OS/utilities). It seems to work generally on our test platforms, and I have no explanation for why it fails on some occasions.

Interestingly, `testComponents` uses the same instrumentation and it has never run into this issue AFAIK. I am reluctant to exclude this test (even on some platforms), because I believe it provides more value and protection than harm.

> I agree we need a more long-term cross-platform solution. One possible solution is to get rid of most of the memory leaks so it's less of a concern overall;

That's all fine and good for the existing code base, but it would be great to know there are leaks during development of new components. We could instrument with valgrind on one of the CI platforms and rely on that. I like the idea of having both for new components (in test cases) and overall system (simulation) integrity (valgrind/memcheck). The current approach seems like a pretty cheap way of nipping bad components in  the bud, which I like.

aseth1:
I am eager to see if this will resolve what is happening in #634.

aymanhab:
@aseth1 Please feel free to merge this so you can move on with your other PR, and let's hope it resolves the problem there. Will need a robust solution long term but that's outside the scope of this one.

aseth1:
> Will need a robust solution long term but that's outside the scope of this one.
I agree, and If we wait long enough it might be in std for C++17 :grin:

PR #640, opened 2015-09-24
[WIP] [contrib] implementDoSomething()

chrisdembia:
This PR is to keep track of our discussion of `extend` vs `implement`. We may change the name `implement` to something else, like `attempt` :)


msdemers:
@chrisdembia "attempt" would make me very wary as a customer of an API.  Even if I was trusting, I'd assume it would recover from errors and failures gracefully.

PR #646, opened 2015-09-30
Fix bug with installing developer docs.

chrisdembia:
Previously, in the installation directory, both the user and dev doxygen
folders held the user doxygen. With this PR, the dev doxygen is properly
installed into the html_developer folder.

This fixes an issue with the myosin PR doxygen uploads, which incorrectly were showing the user docs.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -182,6 +182,9 @@ if(${DOXYGEN_FOUND})
         list(APPEND DOXYGEN_TARGET_DEPENDENCIES "${html_images_binary_dir}")
 
     endforeach()
+    # Unset loop variables to avoid bugs later on.
+    unset(html_binary_dir)
+    unset(audience)
 
     # Define the doxygen target (now that we know the dependencies).
     # --------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


doc/CMakeLists.txt
```diff
@@ -182,6 +182,9 @@ if(${DOXYGEN_FOUND})
         list(APPEND DOXYGEN_TARGET_DEPENDENCIES "${html_images_binary_dir}")
 
     endforeach()
+    # Unset loop variables to avoid bugs later on.
+    unset(html_binary_dir)
```
sherm1: is this variable still used somewhere?

chrisdembia: It's used above this line but it's no longer used below this line.

sherm1: OK. The comment and unset are a little, er ... unsettling though. We don't usually do that at the end of a variables useful lifetime so it makes me wonder why it's being done here. I'm fine with it like this if you think it should be there though.

sherm1:
/review. OK to merge.

chrisdembia:
I removed the `unset`s. Thanks for reviewing.

PR #649, opened 2015-09-30
WeldConstraint and BushingForces utilize Frames

aseth1:
Update to `WeldConstraint` and all `BushingForces` to take advantage of Frames since they clearly link two Frames together.

There was a ton of code duplication (and there is still some, especially in Bushings and `testForces`). `WeldConstraint`  and all (4) `BushingForces` require the same structural setup. In particular, they link two (Physical)Frames together. Therefore they share connectors to frames and associated properties as well as constructors. 

I added an intermediate utility (mixin) class (`LinkTwoFrames<ComponentType, FrameType>`) for the purpose of eliminating obvious code duplication, ease maintainability, and eliminate opportunities for inconsistent naming and documentation to arise (which there was a lot of). I did not invest much time in its naming so I am happy to revisit it as part of this review. I would not expect the casual API user to use it. (As an aside, Joints could potentially be included if we agree that `frame1` and `frame2`, is as informative as `parent` and `child` for Joints- then they would be identical to what `LinkTwoFrames<>` defines). I very much like having one place that standardizes how any Component should link two frames (including convenience constructors for including offsets) from the perspective of API consistency.

Visualization has been left to the individual `BushingForces` (WeldConstraint and Constraints in general do not visualize) but I think these could also be centralized, but I view that as secondary to getting the Constraint and Forces to compute correctly using Frames (main focus of this PR).

The updated Constraint and Forces have also been added to `testComponents` and the definition and tolerance for memory leaks has been tightened (so we may see failures on different platforms).

[Review on Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/649)


OpenSim/Simulation/Model/LinkTwoFrames.h
```diff
@@ -0,0 +1,631 @@
...
+/* -------------------------------------------------------------------------- *
+ *                        OpenSim:  LinkTwoFrames.h                           *
+ * -------------------------------------------------------------------------- *
+ * The OpenSim API is a toolkit for musculoskeletal modeling and simulation.  *
+ * See http://opensim.stanford.edu and the NOTICE file for more information.  *
+ * OpenSim is developed at Stanford University and supported by the US        *
+ * National Institutes of Health (U54 GM072970, R24 HD065690) and by DARPA    *
+ * through the Warrior Web program.                                           *
+ *                                                                            *
+ * Copyright (c) 2005-2015 Stanford University and the Authors                *
```
sherm1: If this is all new code I would just put `Copyright (c) 2015 ...`.

sherm1:
Review status: 0 of 18 files reviewed at latest revision, 15 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 57 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-WabVgImFlunaxU3Z2)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L57)):</sup>
I'm unclear (again) on why this property exists. Couldn't it just make use of a generic `subcomponents` property inherited from `Component`? If it is here it should have a max length of 2 (I think there is a property macro for setting min/max allowable lengths). Also, is there an implication that if some Frame components are provided in this Property then the parent component will automatically connect to them? 


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 102 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-W_kLZ-kZUnRcZgMUs)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L102)):</sup>
name doesn't match the parameter name `frame1Name` (and frame2 below)


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 113 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-WaLESbquVk2THpJaG)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L113)):</sup>
linelength > 80? (several more below too)


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 135 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-Wd19OR1-1B7IssDXk)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L135)):</sup>
Is it worth saying "XYZ Body-fixed Euler angles" or is that always understood in OpenSim?


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 137 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-WdRFS4aBdc0nktgw6)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L137)):</sup>
I don't understand this comment. Euler angles are capable of representing any orientation.


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 144 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-WdfZn65tWxroQ-sY7)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L144)):</sup>
Here it would make sense to say that the Euler angle derivatives are singular if the second Euler angle is close to 90 degrees.


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 156 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-Wdx59QRVjnJgR99XH)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L156)):</sup>
Typos above: Physcial->Physical. *please* make the parameter name`scaleSet` rather than `aScaleSet` (barf!).


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 173 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-WgFRP2oWZ0bF0jtcX)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L173)):</sup>
This comment could use some improvement I think but I'm not sure exactly how to do it. Something like: "Frame2's pose in frame 1 has been decomposed in to 6 generalized coordinates q. Their derivatives dq form a generalized "mobility" basis that can be used to express frame2's spatial velocity in frame1 or to apply generalized forces that apply equal and opposite spatial forces to the two frames. This method takes a generalized force in the mobility basis and calculates the equivalent spatial forces on each frame as though they were applied at the respective frame origins."
(there is also a parameter name mismatch)


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 182 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-WiH8jMT5oGmluhUgi)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L182)):</sup>
`const Vec6& internalForce;`


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 192 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-Wil59KkgN9MzP9H4_)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L192)):</sup>
The documentation for this is confusing. I can't tell what it does -- is it different than the above other than adding rather than assigning? Is that really a Vector of forces for all physical frames in the System? Not just the two belonging to this component? "physical forces" should be "spatial forces" I think.


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 201 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-WjZ8Rbn0_R4y78WhT)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L201)):</sup>
Smart pointers are self initializing so don't need the `{}`.


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 255 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-WjqJfP8QS00K_w5yk)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L255)):</sup>
commented out?


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 301 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-Wk5rgAsSFDT8ZYpS-)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L301)):</sup>
what is that `this->template` doing there???


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 524 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-WlAopar0MY8DKzj06)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L524)):</sup>
If I'm reading this right these aren't applied to PhysicalFrames in general but strictly to the body frames. I don't see any provision for PhysicalOffsetFrames, for example.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/649)

sherm1:
Reviewed 1 of 18 files at r1.
Review status: 1 of 18 files reviewed at latest revision, 15 unresolved discussions, some commit checks broke.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/649)

sherm1:
Ajay, I'll stop there for now since that seems like enough comments for a conversation.

OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.h
```diff
@@ -36,11 +37,11 @@ namespace OpenSim {
 //=================================================================================
 /**
  * A class implementing a Coupled Bushing Force.
- * A Couple Bushing Force is the force proportional to the deviation of two frames, 
+ * A Couple Bushing Force is the force proportional to the deflection of two frames, 
  * where components of the resulting BodyForce are coupled to any or all 
- * deviations, such that the general stiffness and damping matrices are 6x6.
+ * deflection, such that the general stiffness and damping matrices are 6x6.
```
msdemers: small typo. deflection -> deflections

OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.h
```diff
@@ -36,11 +37,11 @@ namespace OpenSim {
...
  *
- * Deviations of an offset_frame (1) (on a PhysicalFrame e.g. a Body) and another 
+ * Deflection of an offset_frame (1) (on a PhysicalFrame e.g. a Body) and another
```
msdemers: Deflection -> Deflections

OpenSim/Simulation/Model/ExpressionBasedBushingForce.cpp
```diff
@@ -177,91 +166,13 @@ void ExpressionBasedBushingForce::extendConnectToModel(Model& aModel)
...
-        // ?Keep a pointer to body2?
-    }
-    else {
-        errorMessage = "Invalid bushing body2 (" + body2Name 
-                        + ") specified in Force " + getName();
-        throw OpenSim::Exception(errorMessage);
+    // fill damping matrix with damping from vector property
+    for (int i = 0; i<3; i++) {
+        _dampingMatrix[i][i] = get_rotational_damping(0)[i];
+        _dampingMatrix[i + 3][i + 3] = get_translational_damping(0)[i];
```
msdemers: It's always bugged me that I didn't spend time designing deflection velocities into the force expressions.  Not in this PR, but soon we should discuss whether or not to treat damping separately as we do no, or incorporate it into the expressions.

aymanhab:
@aseth1 Before this goes much too long and into all fine details, do you know how that affects the exposed interfaces/inheritance from Scripting side? I already see the python test failing but I'm not sure if you or @chrisdembia looked into it.

msdemers:
@aseth1 It's awesome to see so much code disappear! The bushings went on a diet and are looking mighty lean.

chrisdembia:
Update CHANGELOG?

aseth1:
Review status: 1 of 18 files reviewed at latest revision, 9 unresolved discussions, some commit checks broke.

---

<sup>**[OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.h, line 42 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-WZ8gbvMAE8mem21ef-r1-42)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.h#L42)):</sup>
I think of it as a deflection (like displacement) with 3 rotational and 3 translation components. 


---

<sup>**[OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.h, line 44 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-WZ8gbvMAE8mem21ef-r1-44)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Examples/Plugins/CoupledBushingForceExample/CoupledBushingForce.h#L44)):</sup>
I see the confusion. It should say " ... is expressed ... " below.


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 57 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-WabVgImFlunaxU3Z2)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L57)):</sup>
I thought i would make the use of Frames by other components consistent with Joint first and then change it. That is another reason for isolating the Frame related services to this class so that the upgrade should be relatively painless. 


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 135 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-Wd19OR1-1B7IssDXk)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L135)):</sup>
I will add it.


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 137 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-WdRFS4aBdc0nktgw6)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L137)):</sup>
I will clarify. Using the angular deflection in Euler angles to compute forces (like {tau} = [K]{dq}) is not valid for large deflections. That is because the decomposition of the the relative orientation (rotation matrix) is not unique. When computing the relative transform, I don't think it is possible to determine if the Frame is rotated pi or -pi, in the first rotation and then distinguish between 0 and pi (or -pi/2 and pi/2) for the other two subsequent rotations. To be safe, I don't think deflection forces should be computed for deflections approaching 90degs. 


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 192 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-Wil59KkgN9MzP9H4_)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L192)):</sup>
It is for adding to the system's "bodyForces" except `Body` is no longer the only `Frame` that can accept forces, since any `PhysicalFrame` can be used.  It is intended use is inside a Force's `computeForce()`. For example, CoupledBushingForce:
```cpp
void CoupledBushingForce::computeForce(const SimTK::State& s, 
                              SimTK::Vector_<SimTK::SpatialVec>& bodyForces, 
                              SimTK::Vector& generalizedForces) const
{
    // Calculate stiffness generalized forces of bushing by first computing
    // the deviation of the two frames measured by dq
    Vec6 fk = -_stiffnessMatrix*computeDeflection(s);

    // velocity dependent force according to the speed of frame2
    // relative to frame1 
    Vec6 fv = -_dampingMatrix * computeDeflectionRate(s);

    // total bushing force in the internal basis of the deflection (dq) 
    Vec6 f = fk + fv; 

    // convert internal forces to spatial and add then add to system
    // physical (body) forces
    addInPhysicalForcesFromInternal(s, f, bodyForces);
}
```


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 255 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-WjqJfP8QS00K_w5yk)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L255)):</sup>
Oversight on my part. I used it during migration, but no longer needed. It will be removed.


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 301 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-Wk5rgAsSFDT8ZYpS-)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L301)):</sup>
It is required to call a  template method on `this`, which is also a template class, so that the template argument (in this case) can be defined.


---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 524 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-WlAopar0MY8DKzj06)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L524)):</sup>
This is how the internal bushing forces are mapped  (including the effects of applying to a PhysicalOffsetFrame) to the underlying multibody system  where we are only have MobilizedBodies and mobilityForces. It isn't a public method and in not intended as general mechanism for applying forces to Frames in OpenSim. It is a service it provides so derived Force classes can implement `computeForce()`.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/649)

sherm1:
Review status: 0 of 21 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 192 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-Wil59KkgN9MzP9H4_)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L192)):</sup>
OK, I see. But I still think the comment is wrong for the `spatialForces` argument. It says "on each PhysicalFrame in the System". But every PhysicalOffsetFrame is a kind of PhysicalFrame, so there are way too many. There is actually one force per  mobilized body in the System, which is a subset of the PhysicalFrames in the OpenSim model. Is there some concept of "base physical frame" that maps to the mobilized body frames?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/649)

aseth1:
<sup>**[OpenSim/Simulation/Model/LinkTwoFrames.h, line 192 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/649#-K-Wil59KkgN9MzP9H4_)** ([raw file](https://github.com/opensim-org/opensim-core/blob/70106b31ab8421e1e438035bab98425a09fd47bf/OpenSim/Simulation/Model/LinkTwoFrames.h#L192)):</sup>
Yes, there is - good point. I will update accordingly.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/649)

aseth1:
> @aseth1 Before this goes much too long and into all fine details, do you know how that affects the exposed interfaces/inheritance from Scripting side? I already see the python test failing but I'm not sure if you or @chrisdembia looked into it.

@aymanhab Except for having to know that the BushingForces are Forces and the WeldConstraint is a Constraint knowledge that they derive from the LinkTwoFrames mixin is unnecessary. The API for the derived classes in most cases are identical to what it was before except for some improvements in consistency (in naming, documentation, updating from XML).

The solution is to add more python (Java) test cases that make use of the BushingForces, WeldConstraint and PhysicalOffsetFrame. Should, I add those test cases as part of this PR?

sherm1:
From Dev Team meeting: we looked at the documentation and everyone felt that having the LinkTwoFrames mixin in the hierarchy is a problem (esp. the inheritance diagram). @msdemers points out that most of the functionality could be built in to Frames. This could be a time to use a helper class member rather than using inheritance.

sherm1:
Could the mixin be privately inherited to make it disappear?

sherm1:
Maybe multiple inheritance (bushing is a public force and private linktwoframes?)

klshrinidhi:
>Maybe multiple inheritance (bushing is a public force and private linktwoframes?)

Using multiple inheritance may introduce *diamond pattern* in the inheritance hierarchy. With this pattern, if the common base class (Force) is *not virtual*, then derived class (BushingForce) will contain two sub-objects for the same base class (Force). This can create ambiguity in *name lookup*. When a name contained in Force is not fully qualified, it can be ambiguous -- which of the two sub-objects the name should come from ?

sherm1:
> Using multiple inheritance may introduce diamond pattern in the inheritance hierarchy.

Hopefully that wouldn't occur with mixin classes -- they are supposed to be self-contained.

klshrinidhi:
>Hopefully that wouldn't occur with mixin classes -- they are supposed to be self-contained.

Okay. If that is the case, there may be no problem.
But since LinkTwoFrames<Force, PhysicalFrame> is deriving from Force and BushingForce derives from  both Force and LinkTwoFrames, Force becomes a common base class.

sherm1:
> But since LinkTwoFrames is deriving from Force and BushingForce derives from both Force and LinkTwoFrames, Force becomes a common base class.

Yes, good point. Maybe the mixin could be redesigned to avoid that?

aseth1:
I am not a fan of the mixin appearing in the hierarchy, but mainly because of the indirection it caused in the diagram that now hides the derived BushingForces. If you go to `BushingForce` itself, the inheritance diagram seems sensible to me, but that assumes you can find it in the first place. Looking at the `Force` tree one should be able to see/find all the Forces in OpenSim. 

I tried to "have a" `LinkTwoFrames` helper component instead, but it turned out not to be helpful because: 1) we cannot inherit the constructors which is a lot of the code in these classes and  2) hiding its existence as a private member forces you to create accessors in each class that has it. If not private you then have to tunnel through this now weird internal component.

I don't think a simple switch to multiple inheritance can work either, since the extension class isn't a pure interface- it is the one that introduces the frame properties and connectors - so it has to derive from Component.  @klshrinidhi is right that this will form a diamond pattern. 

I will think about it a bit more, but the easiest solution may be to hide this class from the hierarchy if it allows us to see the BushingForces as leafs of the `Force` inheritance tree. 

I am open to suggestions.

sherm1:
>  @klshrinidhi is right that this will form a diamond pattern.

I'm not suggesting it, but a diamond pattern isn't necessarily a problem -- C++ has "virtual inheritance" for exactly this situation. I have used diamond inheritance before (in a past life) and it did work OK. So it *might* be worth playing with before rejecting it; I'm not sure.

aseth1:
How about something like this for complete view of Forces in OpenSim?
classopensim_1_1force__inherit__graph


Which looks like this from the `BushingForce` perspective.
classopensim_1_1functionbasedbushingforce__inherit__graph

I realize it is a A LOT more detail, but at least you can see all the Forces. It requires having **dot** installed, which is part of the [*graphviz*](http://www.graphviz.org) utilities.

@chrisdembia also suggested having curated lists of components for *Joints*, *Constraints*, *Forces*, *Actuators*, ..., which makes a lot of sense for quickly seeing which concrete things are available without navigating the inheritance diagram. The diagrams could be an option for only the developer doxygen.

sherm1:
I like that first diagram! I don't understand why the second one makes LinkTwoFrames appear to use multiple inheritance though.

If "dot" is only required when we generate the doxygen and not for end user viewing then it seems very reasonable to use.

aseth1:
>  I don't understand why the second one makes LinkTwoFrames appear to use multiple inheritance though.

I found this explanation:
*A yellow dashed arrow denotes a relation between a template instance and the template class it was instantiated from. The arrow is labeled with the template parameters of the instance.* 

So it really should not show up in the inheritance diagram. Might be options to suppress OR we can publicize the explanation.

sherm1:
> So it really should not show up in the inheritance diagram. Might be options to suppress OR we can publicize the explanation.

Hmm. Might be useful once we got used to it. Can we do it that way for a while and see what it does to other parts of the documentation?

aseth1:
This is what it looks like for `PhysicalOffsetFrame` (another mixin product):

classopensim_1_1physicaloffsetframe__inherit__graph

>A box with a red border indicates a node that has more arrows than are shown! In other words: the graph is truncated with respect to this node. The reason why a graph is sometimes truncated is to prevent images from becoming too large. For the graphs generated with dot doxygen tries to limit the width of the resulting image to 1024 pixels. 

I was experimenting with the dot's MAX_DOT_GRAPH_DEPTH and set it to 2.

sherm1:
The dotted yellow part doesn't seem very useful in that one.

aseth1:
> The dotted yellow part doesn't seem very useful in that one.

We could ignore template relations in the *user* doxygen and keep for *developer*?

Looks like this with `TEMPLATE_RELATIONS = NO`:
classopensim_1_1physicaloffsetframe__inherit__graph_no_template_relations

sherm1:
That looks better. I couldn't tell if the template relations would be helpful for developers. Looks to me like the template arguments are already listed. Are they adding anything else?

aseth1:
You are right they are listed and having a bifurcation in the graph (even if it isn't inheritance) is confusing. Are you convinced enough that we could use *dot* to make sensible inheritance diagrams (e.g. ignoring template relations). If so, we could merge this PR and then integrate dot and refine its DOXYGEN options/settings as part of another. What do you think?

sherm1:
Sounds good -- but let's discuss w/team at 11am meeting.

chrisdembia:
TwoFrameLinker, TwoFrameComponent, PairOfFrames.

aseth1:
@sherm1 @chrisdembia 
The Appveyor tests are in fact passing (see details) but the *after_test:* section of the `appveyor.yml` is failing:

>----------------------------------------------------------------------
Ran 14 tests in 0.875s
> 
>OK
IF %APPVEYOR_REPO_BRANCH% EQU master IF NOT DEFINED APPVEYOR_PULL_REQUEST_NUMBER SET DISTR=TRUE
& was unexpected at this time.
Command exited with code 255

Do you know why? If the PR branch is simply behind in a change, can we go ahead and merge?

aseth1:
Appveyor failure is due to "&" in the branch name (sorry!).

Issue #673 files to address rendering Doygen diagrams with **dot** as a separate PR.

sherm1:
Is this ready to merge then Ajay?

aseth1:
> Is this ready to merge then Ajay?

@sherm1 @chrisdembia I think so!

chrisdembia:
I have nothing to say.

sherm1:
> I have nothing to say.

I do -- Yay!

PR #651, opened 2015-10-01
Fix "missing override" warnings

chrisdembia:
I chose to move `assign()` up to the `Object` class. I hid it from doxygen since it was previously hidden from doxygen.


sherm1:
/review. Looks good to me.

chrisdembia:
Thanks @sherm1 . @aymanhab could you take a quick look?

sherm1:
@aymanhab says assign() isn't being used in Java and can be removed. Then merge?

chrisdembia:
@aymanhab let me know when you've checked the GUI; then I'll remove the method. Low priority.

aymanhab:
@chrisdembia Found that assign is used only once in the GUI, however considering the lack of an operator with similar assignment semantics in scripting, I'd keep it around. So, the change in this PR is fine. Will merge.

chrisdembia:
Thanks @aymanhab

PR #652, opened 2015-10-01
Fix testStaticOptimization reruns

aymanhab:
Version number was missing from subject01_Setup_StaticOptimization.xml, this's a bug we fixed in version 2.0. The xml file was manually created. The missing version number caused the wrong calls in AnalyzeTool::updateFromXML

### DIFF
--- a/Applications/Analyze/test/subject01_Setup_StaticOptimization.xml
+++ b/Applications/Analyze/test/subject01_Setup_StaticOptimization.xml
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<OpenSimDocument Version="20201">
 <AnalyzeTool name="subject01_walk1">
   <!--Name of the .osim file used to construct a model.-->
   <model_file> subject01.osim </model_file>
@@ -59,3 +60,4 @@
 	    filtering. The default value is -1.0, so no filtering.-->
   <lowpass_cutoff_frequency_for_load_kinematics> -1 </lowpass_cutoff_frequency_for_load_kinematics>
 </AnalyzeTool>
+</OpenSimDocument>


aymanhab:
Ready for merge after tests pass, Please review @aseth1 @chrisdembia

chrisdembia:
+1

aseth1:
Glad you found the cause @aymanhab. Still a mystery why the file on disk was being modified. I understand that the incorrect update code would be executed by missing the version number, but how that translated to  the file being changed is still lingering, yes? 

This suggests that if you are a user still using unversioned setup files (pre 2.0) and you ran your setup files in 4.0 we would be changing those files on disk. Is that true?

aymanhab:
@aseth1 Here's the longer explanation. Overtime, we moved properties from Tools to ExternalLoads file which didn't exist in earlier versions. The missing version at the top implied OpenSim version 1.5 so the code tried to copy the filtering (which was supposed to exist in the AnalyzeTool) to the externalLoads file. 
In general updateFromXML works on DOM/XML and not on files but in this case of ExternalLoads the properties are moved between objects and the tools downstream use the file directly. Hope this explains.

aseth1:
> but in this case of ExternalLoads the properties are moved between objects and the tools downstream use the file directly.

That sounds scary. It does seem like it could creep up in other instances. The fix is beyond the scope of this PR, but can you open an issue and give some details for how/why the tools are modifying the file directly? I don't want this to fall through the cracks. From the perspective the tools input files should be `const.`

Thanks! I'll be happy to merge at this point.

aymanhab:
Issue #654 opened.

aseth1:
Thanks for opening #654. We can discuss the broader issue there. Merging now.

aymanhab:
Great, thanks @aseth1

PR #659, opened 2015-10-07
Python multiconfig generators

chrisdembia:
Previously, the python package in the build tree was not organized correctly
for multiconfiguration generators like visual studio. With this commit, now the
python package in the build tree is properly assembled for each
configuration.

Also, there is now a target that runs only the python tests; `RunPythonTests`. On MSVC, this appears as `Bindings -> Python - run tests`.

### DIFF
--- a/Bindings/Python/CMakeLists.txt
+++ b/Bindings/Python/CMakeLists.txt
@@ -210,6 +210,8 @@ add_custom_target(RunPythonTests
     COMMAND ${CMAKE_CTEST_COMMAND} --tests-regex python
                                    --build-config $<CONFIG>
                                    --extra-verbose
+    PROJECT_LABEL "Python - run tests"
+    FOLDER "Bindings")
     )
 
 
--- a/Bindings/Python/CMakeLists.txt
+++ b/Bindings/Python/CMakeLists.txt
@@ -182,14 +182,36 @@ set_target_properties(PythonBindings PROPERTIES
 
 # Test.
 # =====
-add_test(NAME python WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/$<CONFIG>"
-    # This command runs all the python tests in the tests directory from the
-    # source tree. It's important to run the tests in the source tree so that
-    # one can edit the tests and immediately re-run the tests without any
-    # intermediate file copying.
-    COMMAND "${PYTHON_EXECUTABLE}" -m unittest discover
-                "${CMAKE_CURRENT_SOURCE_DIR}/tests" --verbose
-    )
+# This test runs all the python tests in the tests directory from the
+# source tree. It's important to run the tests in the source tree so that
+# one can edit the tests and immediately re-run the tests without any
+# intermediate file copying.
+# TODO when minimum cmake is v3.0.0 and add_test WORKING_DIRECTORY supports
+# generator expressions, use the following instead:
+#   add_test(NAME python
+#       WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/$<CONFIG>"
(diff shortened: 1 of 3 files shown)


chrisdembia:
@aymanhab would you mind reviewing? Thank you.

aymanhab:
@chrisdembia This maybe unrelated by @aseth1 reported yesterday that he changed the .i file and the python wrapping didn't rerun. Can you test that in this branch as well? Thanks.

Bindings/Python/CMakeLists.txt
```diff
@@ -29,25 +40,17 @@ add_custom_command(
         -I${OpenSim_SOURCE_DIR}/Bindings/
         -I${Simbody_INCLUDE_DIR}
         -o ${swig_output_cxx_file_fullname}
-        -outdir "${PYTHON_PACKAGE_BINARY_DIR}/opensim"
+        -outdir "${CMAKE_CURRENT_BINARY_DIR}"
         ${swig_interface_file_fullname}
     DEPENDS ${swig_interface_file_fullname}
         "${OpenSim_SOURCE_DIR}/Bindings/opensim.i"
         "${OpenSim_SOURCE_DIR}/Bindings/OpenSimHeaders.h"
```
chrisdembia: @aymanhab do you know which *.i file he edited? @aseth1. Both *.i files are listed as dependencies of this command, so should invoke swig to rerun. I am near certain that @aseth1 was using an intermediate version of the code that did not have the above two lines.

aymanhab:
@chrisdembia Likely he was using outdated file(s) but since you were modifying CMakeLists I thought you'd be able to confirm if dependencies are correct. He was editing the common top opensim.i file as far as I can tell, if you touch it and the Bindings are rebuilt then we're golden.

aymanhab:
@chrisdembia For multi-configuration, I was never able to build debuggable _opensim python Binding since it requires debuggable Python environment, is that Windows specific? If not then what's the expected behavior  if you try generating debug build of Python Bindings?

chrisdembia:
> @aymanhab For multi-configuration, I was never able to build debuggable _opensim python Binding since it requires debuggable Python environment, is that Windows specific?

I'm aware of this issue on Windows and I don't know if a debuggable python environment is necessary on UNIX as well.

> If not then what's the expected behavior if you try generating debug build of Python Bindings?

I don't know.

This PR does not address having a successful debuggable opensim python package. I thought about that, but it seems difficult. I'm trying to solve a much smaller problem, more practically related to being able to have concurrent builds from different configurations.

For example, consider the following procedure in MSVC in the current version of master (without this PR):

1. BUILD_ALL in the Release configuration.
2. RUN_TESTS in the Release configuration: python tests pass.
3. BUILD_ALL in the Debug configuration.
4. RUN_TESTS in the *Release* configuration (not ever trying to run python bindings in debug): python tests don't work anymore.

That's because in the current master (the PR you merged for me yesterday; thank you), the _opensim library is written to the same location no matter the build configuration; the debug/release/relwithdebinfo library would get written to BUILD_DIR/Bindings/Python/_opensim.pyd. I took a shortcut temporarily :(.

Anyway, this PR would put the _opensim.pyd library in Bindings/Python/Debug/opensim/_opensim.pyd, etc.

chrisdembia:
> if you touch it and the Bindings are rebuilt then we're golden.

Just tested; we are golden :). Touching opensim.i causes PythonBindings to run again.

aymanhab:
Testing on travis/c-lang failed with unknown command python, is that expected? Otherwise the changes look fine.

chrisdembia:
> Testing on travis/c-lang failed with unknown command python, is that expected?

Not expected. Hopefully this is fixed in the latest commit.



> Otherwise the changes look fine.

Okay thank you.

PR #662, opened 2015-10-08
[travis] Suppress output from Simbody install

chrisdembia:


sherm1:
Looks like the title should be "Suppress *output* from Simbody install"; I think the install is still happening. :+1: to merge with title fix!

chrisdembia:
I edited the title; thank you. Merging.

Issue #148, opened 2014-08-06
Remove Model::disownAllComponents() method and fix test cases accordingly


aymanhab:
The method Model::disownAllComponents() was introduced to work around improper memory management when stack allocated objects are added to a Model that takes ownership of them, it potentially introduces leaks (since no components are deallocated on Model destruction) and promotes improper use of stack allocated objects (mostly seen in test cases). Per discussion with @sherm1 and @aseth1 we'll stick to heap allocated Models and ModelComponents (using new) so that we don't have to invoke the above method.

PR #663, opened 2015-10-09
Proposed rigid contact interface

tkuchida:
Analogous to existing HuntCrossleyForce interface for consistency. :basketball:

[ci skip]


OpenSim/Sandbox/futureRigidContact.cpp
```diff
@@ -0,0 +1,87 @@
...
+    Model *osimModel = new Model;
+    osimModel->setName("BallDrop");
+    osimModel->setGravity(Vec3(0, -9.81, 0));
+
+    // Create unconstrained ball.
+    OpenSim::Body ball("ball", mass, Vec3(0), mass*SimTK::Inertia::sphere(0.1));
+    ball.scale(Vec3(radius), false);
+    FreeJoint free("", osimModel->getGround().getName(), "ball");
+    osimModel->addBody(&ball);
+    osimModel->addJoint(&free);
```
chrisdembia: You will likely get a double-free error if these are not all dynamically allocated (though this comment is not relevant for pseudocode).

OpenSim/Sandbox/futureRigidContact.cpp
```diff
@@ -0,0 +1,87 @@
...
+    double currTime = 0.;
+    while (currTime < simDuration) {
+        currTime += simStepSize;
+        manager.setFinalTime(currTime);
+        manager.integrate(state);
+    }
+
+    // TODO: Check the results.
+
+    osimModel->disownAllComponents();
```
chrisdembia: Oh I see.

chrisdembia:
I like that the two are pretty much exactly the same. I wonder if it makes sense though to *not* have rigid contact and compliant contact be interchangeable (i.e. have the same interface). One is a force element while the other is a constraint. For example, I don't know how RigidContactForce would implement `Force::calcForce()`.  Or maybe they can have the same interface but not both be forces.

Would we also have a RigidCoordinateLimitForce to go along with CoordinateLimitForce? Or would we have JointStopConstraint? Should all unilateral constraints appear in the same inheritance hierarchy, or can we split up things like joint stops (put in constraints) and sphere-plane contact (put in forces)?

chrisdembia:
Anyway, looks good to me. 



> [ci skip]

:+1: saving the environment

OpenSim/Sandbox/futureRigidContact.cpp
sherm1: Arggg! Shouldn't do it that way in real life. That's separate from the rigid contact stuff though.

tkuchida: > Arggg! Shouldn't do it that way in real life.

Arggg! Just following the test cases. Already raised in #148.

sherm1:
Thanks for starting this thread, Tom!

That's a very close match for OpenSim's current HuntCrossley stuff, but that is obsolete and I hope we can change to a better model in 4.0. The problem is that it associates contact parameters with the force element rather than with the contact surfaces. That was replaced in Simbody a long time ago with a better approach where contact properties are associated with ContactSurface objects fixed to bodies.
Those are then combined pairwise when they are in contact to compute the effective properties for a particular impact or contact. Then forces get applied when pairs bump, but there is no separate force element. 

A ContactSurface is a pair: geometry+material properties. The material properties can be compliant or rigid or both. Then conceivably it is just a request to the solver whether to use compliant or rigid contact; probably we'll want to use both in the same simulations though.

That said, there are some simpler cases that it would be useful to try first. My first cut at rigid contact in Simbody doesn't have the fancy broad phase determination of which pairs of n things are in contact. At first I'll just have pairwise contact elements: joint stops, ropes, and point-ground contacts where each point has its own element. In those cases the contact element will be the keeper of the contact parameters. It would be cool if those could also support either compliant or rigid modeling for apples-to-apples comparisons; I don't know how that would look in OpenSim. There are advantages to these pairwise elements and I think we would want to keep them even once we have the nXn stuff working.

Another tricky issue is how we should get output from these things. They will generate forces when in contact but also very substantial impulses during impact that ought to be reported somehow. Would that be in scope for pseudocode?

sherm1:
> Would we also have a RigidCoordinateLimitForce to go along with CoordinateLimitForce? Or would we have JointStopConstraint? 

I don't think we should have any of these. Instead we should have a JointStop (or CoordinateLimit) whose job is to keep a coordinate in a given range. Then we should be able to control which model (compliant vs. rigid) at run time, just as we control accuracy now. That is, we don't have "AccurateCoordinateLimitForce" and "SloppyCoordinateLimitForce"; we just consider that something the solver controls. Similarly, switching between compliant and rigid models is an "outer loop" of accuracy control where we choose accurate vs. fast at the modeling level.

chrisdembia:
What's the criterion for merging this? I say we merge it now.

sherm1:
I think we should do it your way since this was your idea!

tkuchida:
I'm not averse to improving this before it goes in (e.g., adding a joint stop and reporter). I really just wanted to get the ball rolling (or bouncing) and thought an interface analogous to the existing API would be a good place to start the discussion. :mushroom:

Also, in deciding how much to change for 4.0, we should probably consider how many person-hours are being allocated for this (and the collateral damage to examples, Confluence, etc.). :boom:

sherm1:
I'm hoping we can leave the old H&CForce contact system in place for backwards compatibility in 4.0. I don't think it prevents us from adding a new contact scheme. But I don't want to model the rigid contact after the obsolete H&CForce stuff.

chrisdembia:
Okay, @tkuchida merge at will!

chrisdembia:
@tkuchida can I merge this?

tkuchida:
>can I merge this?

We haven't gone through the design process, and I don't think this interface is something we want to propose/advertise. If the problem is that this is an open PR, I can move Sherm's comments into an Issue and move the code into a gist.

chrisdembia:
> If the problem is that this is an open PR,

Yeah, but I'm also fine with leaving the PR open.



> I can move Sherm's comments into an Issue and move the code into a gist.

If closing the PR, then I'd prefer renaming the file to `pseudo...` and merging.

sherm1:
I want to discuss this more but I think it would be easier in person and with a whiteboard first. Hopefully we'll have some time at Thursday's meeting. 

For example: most existing models provide a range for joint coordinates, but as I understand it that range is only enforced during IK, which I'm sure is a surprise to users. Now we could by default enforce those as momentum-conserving, zero coefficient of restitution rigid joint limits. Should we?

chrisdembia:
I don't think we should do something backwards-incompatible to something so important. I think we could have an additional flag like "turn_range_into_motion_constraints" or something named much better. 

I definitely think we should exploit that range property though to avoid redundancy and confusion.

chrisdembia:
I suggest we merge this devprop.

tkuchida:
I think @sherm1 raised good points that we should discuss/address. We should at least add them as comments to the devprop so they don't get lost, with the plan to address them when the project is resumed.

chrisdembia:
Okay. Sorry; I just noticed we had a similar back-and-forth a while ago.

tkuchida:
I'll update the code to avoid the egregious use of `disownAllComponents()` and add Sherm's comments at the top.

tkuchida:
> I suggest we merge this devprop.

This is not a development proposal, it's just the beginning of a design discussion. I don't recommend proceeding with this design. I've updated the code to work with the new API, addressed the feedback about using `disownAllComponents()`, and added the comments from this PR to the top of the file. I don't recommend adopting this process for closing PRs in the future since it doesn't seem to add any value.

chrisdembia:
Okay; sorry.

PR #666, opened 2015-10-20
Add gitignore file.

klshrinidhi:
Would like to add .gitignore file to ignore certain directories (like build trees) and files (like backup files editors create).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,4 @@
 # Ignore tmp directory/file. We use it sometimes to write debug information.
 tmp
 # Ignore build trees.
-build-*
+build*
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
-# Ignore backup files created by editors like emacs.
+# Ignore backup files created by editors like emacs, vim.
 *~
+*.swp
 # Ignore cmake files.
 *CMakeCache.txt
 *CMakeFiles
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+*~
+*CMakeCache.txt
+*CMakeFiles
+tmp
+build-*
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,9 @@
+# Ignore backup files created by editors like emacs.
 *~
+# Ignore cmake files.
 *CMakeCache.txt
 *CMakeFiles
+# Ignore tmp directory/file. We use it sometimes to write debug information.
 tmp
+# Ignore build trees.
 build-*


.gitignore
```diff
@@ -0,0 +1,9 @@
+# Ignore backup files created by editors like emacs.
+*~
+# Ignore cmake files.
+*CMakeCache.txt
+*CMakeFiles
+# Ignore tmp directory/file. We use it sometimes to write debug information.
+tmp
+# Ignore build trees.
+build-*
```
sherm1: Does the hyphen here mean that directories called `build` won't get ignored? I think those are fairly common.

klshrinidhi: The hyphen in `build-` is part of the directory name. Any directory with name starting with `build-` will be ignored. This is to allow having build trees in the directory (which I do all the time). If we later decide to have a directory that starts with `build-` in the repository, we can remove/change this one line.

sherm1:
:+1: Good to have a `.gitignore`. But -- life generally works better if you don't mix build stuff inside your repo directory.

tkuchida:
A diabolical PR :smiling_imp:. Does everyone call their build directory `build-*`?

klshrinidhi:
Using a prefix `build` is just a convention. If you are using another name for your build directory, you can add that to `.gitignore` if you like. If/when we do intend to have directories in the repo whose name is currently ignored due to `.gitignore`, we remove/change those specific entries.

.gitignore
sherm1: Shrinidhi, you misunderstood my comment. I'm asking about the hyphen "-". I commonly use a build directory called just `build`, without the hyphen. I have also seen `build_`. So I'm asking whether it would be better to use the pattern `+build*` rather than `+build-*`.

chrisdembia:
Looks good to me. I agree with Shrinidhi that the main contribution here is having a .gitignore; we can add more things to ignore later. One more reviewer to approve/merge?

chrisdembia:
> A diabolical PR :smiling_imp:. Does everyone call their build directory build-*?

This is what I've settled on over time: My git repo dir is `~/repos/opensim-core/1/opensim-core` and my build dir is `~/repos/opensim-core/1/build`. My install dir is `~/repos/opensim-core/1/install`. Then, I have 5 or so copies of the repo (replace 1 with 2, 3, etc.) If I have multiple builds for a single copy of the repo, I do things like `~/repos/opensim-core/1/build-xcode`

sherm1:
Looks great -- thanks, Shrinidhi! Merging.

PR #670, opened 2015-10-22
[cmake] Different required cmake version for MSVC.

chrisdembia:
We don't want to change the required version for *all* platforms since we expect people to get cmake through package managers on linux, and so linux users will likely have an older version of cmake.


sherm1:
Good idea! Have you verified that 2.8.8 is sufficient for Linux and OSX? If so, feel free to merge.

chrisdembia:
Yes. Thank you.

PR #671, opened 2015-10-22
[cmake] Clean up how we specify installation layout

chrisdembia:
I moved all variables that define the layout of the installation to the root CMakeLists, so we can control the installation layout from one place. 

I also allow people to change to an alternative installation layout that attempts to conform to the linux filesystem hierarchy standard (FHS), bringing us closer to allowing unix users to install opensim into /usr or /usr/local.

I did *not* change the default installation layout on any platform.

Could @klshrinidhi and @aymanhab review this please?


Bindings/Java/CMakeLists.txt
```diff
@@ -36,6 +35,13 @@ if( JAVA_COMPILE )
...
+install(DIRECTORY Matlab DESTINATION "${OPENSIM_INSTALL_SCRIPTSDIR}"
+    PATTERN "Matlab/Dynamic_Walker_Example" EXCLUDE
+    PATTERN "Matlab/configureOpenSim.m.in" EXCLUDE)
 
+# This script contains paths into the OpenSim installation that may be
+# different on different platforms, so we configure it with CMake variables.
+configure_file(Matlab/configureOpenSim.m.in
+    "${CMAKE_CURRENT_BINARY_DIR}/configureOpenSim.m" @ONLY)
+install(FILES "${CMAKE_CURRENT_BINARY_DIR}/configureOpenSim.m"
+        DESTINATION "${OPENSIM_INSTALL_SCRIPTSDIR}/Matlab")
```
klshrinidhi: * Does the configuration leave the original `.in` file or does it overwrite it ? If it leaves the original, may be deleting it is a good idea ?
* Is it useful to create another variable (in the root CMakeLists.txt) holding the destination for Matlab (similarly for Python etc) stuff instead of hard-coding it here ?

CMakeLists.txt
```diff
@@ -103,6 +105,81 @@ if(UNIX AND
     CMAKE_INSTALL_PREFIX.")
 endif()
 
+# Configure installation directories across platforms.
+# ----------------------------------------------------
+
+## OPENSIM_INSTALL_UNIX_FHS option.
+set(OPENSIM_INSTALL_UNIX_FHS_DEFAULT OFF)
```
klshrinidhi: My understanding about this option is:
* When ON, the installation layout is what unix uses in its standard directories. This option *has to be ON* for a user to install OpenSim in standard directories.
* When OFF, the installation layout is not the one unix uses. This can be used while installing OpenSim anywhere but standard directories. 

I am suggesting to have this option *ON by default* resulting in OpenSim installation layout being the same all the time -- like unix standard directories. This way user can install anywhere in unix and the layout will look the same. User is allowed to turn this option OFF by going advanced on CMake but will see a fatal error if attempting to install in unix standard directories.

Bindings/Java/CMakeLists.txt
chrisdembia: > Does the configuration leave the original .in file or does it overwrite it ? If it leaves the original, may be deleting it is a good idea ?

The configuration leaves the original `.in` file; it's part of the git repo, so it doesn't make sense to delete it. There's no overwriting. 



> Is it useful to create another variable (in the root CMakeLists.txt) holding the destination for Matlab (similarly for Python etc) stuff instead of hard-coding it here ?

Yes; I added a new variable for this.

CMakeLists.txt
chrisdembia: I changed the default behavior as we discussed (ON by default on UNIX).

chrisdembia:
Ready for review again.

klshrinidhi:
:heavy_check_mark:

chrisdembia:
@aymanhab, when you get a chance could you review this PR?

aymanhab:
Looks fine @chrisdembia We'll see how this works when trying to make an actual distribution. I'll merge. Do we have old/new documentation for the installation layout? Thanks.

chrisdembia:
> Do we have old/new documentation for the installation layout?

I haven't created new documentation, though we could at least do something similar to what simbody's readme has (listing the installation layout on all platforms). Also, we probably use installation paths everywhere in the documentation. I'll open a new issue about accounting for the new installation layout in the documentation. Though as you mention, I expect it to change before the next release. This PR is mostly about making it easier to change.

chrisdembia:
Thanks for merging.

PR #680, opened 2015-10-27
[doxygen] Ground is the Newtonian frame.

chrisdembia:
I think it's important for the doxygen for Ground to mention that it is the Newtonian reference frame.

I removed "Points" b/c we don't yet have a Point class.

### DIFF
--- a/OpenSim/Simulation/Model/Ground.h
+++ b/OpenSim/Simulation/Model/Ground.h
@@ -30,7 +30,7 @@ namespace OpenSim {
 //=============================================================================
 //=============================================================================
 /**
-* Ground provides an inertial reference frame in which the
+* Ground is an inertial reference frame in which the
 * motion of all Frames and points may conveniently and efficiently 
 * be expressed. As a PhysicalFrame, Ground supports physical connections
 * (e.g. Joints, Constraints), and forces can be applied to it.
--- a/OpenSim/Simulation/Model/Ground.h
+++ b/OpenSim/Simulation/Model/Ground.h
@@ -30,9 +30,10 @@ namespace OpenSim {
 //=============================================================================
 //=============================================================================
 /**
-* Ground is a PhysicalFrame in which all Frames (transforms) and Points 
-* (locations) can be expressed. As a PhysicalFrame it supports physical 
-* connections (e.g. Joints, Constraints) and forces can be applied to it.
+* Ground is the Newtonian reference frame in which the equations of motion 
+* are solved; the orientation of all Frames and the location of points 
+* can be expressed in this frame. As a PhysicalFrame, Ground supports physical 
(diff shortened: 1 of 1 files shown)


sherm1:
Ground's description should include the phrase "inertial reference frame" also; I think that it more common than "Newtonian frame".

aseth1:
I believe it is also incorrect to say that it is the frame in which the equations of motion are solved, or at least I do not understand what you mean. The equations are expressed and solved in the basis of the internal coordinates (specifically the mobilities). Internal quantities can can be mapped to spatial and inertial reference frame quantities relatively easily. If they were truly solved in the inertial reference frame then we would have six equations for each body and joints would be equivalent to constraints and be captured by constraint forces acting on floating bodies. In fact, the solution of internal coordinate formulations require "ficticious" centrifugal an Coriolis forces to account for non-inertial frames, because the frames of reference in which the coordinates are expressed are rotating.

aseth1:
> Ground's description should include the phrase "inertial reference frame" also; I think that it more common than "Newtonian frame". 

Google scholar's "inertial reference frame" 20,600 results vs. 557 for "Newtonian frame" supports your claim! Newton's laws of motion describe the motion of a body in an inertial reference frame, right?

chrisdembia:
Thanks for that google search. I'll change it to "inertial reference frame". By "in which the equations of motion are solved", I meant that the vector time derivatives are taken with respect to the Ground frame. But I think this is sufficiently implied by "inertial reference frame", so I'll take out the phrase.

chrisdembia:
Updated.

sherm1:
There can be many inertial reference frames (any Frame fixed w.r.t. Ground). How about this for wording:

```Ground provides a distinguished inertial reference frame in which the orientation and motion of all Frames and locations and motion of all points may conveniently and efficiently be expressed.```

chrisdembia:
@sherm1 I like that sentence a lot, but I think "distinguished" is perhaps more confusing than helpful; people might think "what do they mean by distinguished?" What if we say "*an* inertial reference frame" instead of saying "*the* inertial reference frame"?

sherm1:
> What if we say "an inertial reference frame" instead of saying "the inertial reference frame"?

Yeah, that sounds better.

aseth1:
:+1:

chrisdembia:
I updated the docs again. I also removed "orientation" and "location" and just left "motion"; I think it's more concise this way and has nearly the same meaning.

sherm1:
I like it. Ready to merge if Ajay agrees.

aseth1:
Small change:

`Ground is an inertial ...` instead of `Ground provides an inertial ...` since it is a Frame.

chrisdembia:
I addressed @aseth1's comment. Please merge when ready.

PR #681, opened 2015-10-27
Sandbox inverse kinematics based on IMUs only

mitkof6:
Demonstrates how orientation sensors can be used for tracking the movements
of the arm and to perform the inverse kinematics. Two IMUs are attached, one 
on the humerus and the other on the radius. The recordings are stored in a 
.trc file. The model is an extended version of the arm26. Currently, OpenSim
does not provide direct tools for employing inverse kinematics based on the
orientation sensors, while Simbody provides these facilities. This code
uses the Simbody interface to execute the inverse kinematics for the specific 
problem and it is not a generic solution.

The resultant motion was validated visually.


sherm1:
Looks great, thanks Dimitar! It would be good to get this built into OpenSim. The builds failed because there are apparently tabs lurking in the code somewhere. Tabs are considered anti-social because they only look right in the author's editor. Please be sure you have "convert tabs to spaces" set in your editor so that everyone will see the code as you intended it. You'll need to fix the existing tabs by replacing them with spaces so that this can get by our automatic checks.

chrisdembia:
Coool!  @mitkof6 I think the example won't run in the build tree because CMake will not copy your resources over to the build tree. Could you add something like `file(COPY x y)` to the Sandbox CMakeLists.txt? You could either glob up trc, osim, etc. or manually list the files.

Also, if this code gets used in a future OpenSim tool and you'd like credit/authorship, you might want to add the copyright block (just copy and paste from another source file and add your name under Author(s)).

chrisdembia:
I suggest that after those small changes are made, we merge this PR and then we can continue any discussion on this PR (we can still add comments on a closed PR).

mitkof6:
> Tabs are considered anti-social because they only look right in the author's editor. Please be sure you have "convert tabs to spaces" set in your editor so that everyone will see the code as you intended it. You'll need to fix the existing tabs by replacing them with spaces so that this can get by our automatic checks.

Yes I had it set, but there where some tabs that where not removed. Hope it is ok now.

> Could you add something like file(COPY x y) to the Sandbox CMakeLists.txt? You could either glob up trc, osim, etc. or manually list the files.

I use the configure_file(${dataFile} ${CMAKE_CURRENT_BINARY_DIR}/${dataFile} COPYONLY) command.

sherm1:
@chrisdembia, one of the 4 Travis builds [failed](https://travis-ci.org/opensim-org/opensim-core/jobs/88085128). I couldn't tell why from looking at the log -- can you? Does that failure have anything to do with this PR or can we merge it?

chrisdembia:
> I couldn't tell why from looking at the log -- can you?

It failed on the line that uploads doxygen to myosin.sourceforge.net. My fault! The reason is that this PR is coming from a fork; my shell script stuff didn't anticipate that.

> Does that failure have anything to do with this PR or can we merge it?

Has nothing to do with this PR. I'll merge.

chrisdembia:
Thanks @mitkof6

PR #690, opened 2015-11-02
[java] Install jar file.

chrisdembia:
Also, set RPATH in osimJavaJNI to find other opensim libraries. Now, matlab
users do not need to set DYLD_LIBRARY_PATH.

In the install tree, the sdk folder now has:

```
sdk/Java/org/opensim/modeling/*.java
sdk/Java/org/opensim/modeling/*.class
sdk/Java/org-opensim-modeling.jar
```

Does this cause java classpath issues where the .class files and .jar file may conflict if `sdk/Java` is on the classpath?

@aymanhab could you review?

### DIFF
--- a/Bindings/Java/CMakeLists.txt
+++ b/Bindings/Java/CMakeLists.txt
@@ -19,11 +19,11 @@ if( JAVA_COMPILE )
   # message( "WD=" ${CMAKE_BINARY_DIR}"/Bindings/Java/OpenSimJNI/src/" )
 
   add_custom_target( JavaCompile
-     COMMAND ${JAVA_COMPILE} 
-     ${CMAKE_BINARY_DIR}/Bindings/Java/OpenSimJNI/src/org/opensim/modeling/*.java 
-     -source 1.6 -target 1.6
-     WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/Bindings/Java/OpenSimJNI/src
-     )
+      COMMAND ${JAVA_COMPILE} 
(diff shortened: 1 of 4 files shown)


aymanhab:
@chrisdembia The build with wrapping fails to install because JAVA_Compile is not baked into the BUILD_ALL (we manually invoke it as needed).  Not sure why "Quiet" find Simbody is included in this PR either. Once these are resolved I'm ok with the merge so that others can test the latest API from Matlab then we can refine the process when the GUI is integrated. Thanks

chrisdembia:
I removed the "Quiet" change.

Oops... I had already started refining the Java bindings cmake scripts before I saw your message. I rearranged some stuff so that the jar is built as part of BUILD_ALL.

If this is too much for one PR, I can split it up.

aymanhab:
No worries as long as the changes are in the same "area". Please let me
know when it's ready. Thanks.

chrisdembia:
This is ready for review.

aymanhab:
@chrisdembia The changes look fine but are impossible to test without building, if you did (or had travis/appveyor do) then we need to check a couple things:
1. The jar file was created.
2. The jar file could be loaded in some environment (e.g. Matlab or Java). This's because there's assumed directory structure inside the jar file (corresponding to java package layout) so we need to verify the jar file is actually good. 
I may use this opportunity to add a couple test cases to have this process baked into the build as well but that's probably better done with ant rather than cmake. 
Please let me know about 1 & 2. Thanks.

chrisdembia:
Just tested both on OSX: successfully created a Model in Java.

aymanhab:
Excellent, thanks much @chrisdembia

chrisdembia:
Thanks for merging.

PR #691, opened 2015-11-02
[readme] brew python issues with cmake.

chrisdembia:
There's a bug with cmake/homebrew when OSX users have installed python via homebrew. I've tried to help people who get stuck in this situation. I also changed "Mac" to "OSX" since I think "OSX" is more accurate ('mac' refers to the hardware and the OS, but I really want to refer to the OS).

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ Building from the source code
 We support a few ways of building OpenSim:
 
 1. [On Windows using Microsoft Visual Studio](#on-windows-using-visual-studio).
-2. [On Mac using Xcode](#on-mac-using-xcode).
+2. [On OSX using Xcode](#on-osx-using-xcode).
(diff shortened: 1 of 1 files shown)


aymanhab:
@chrisdembia This looks fine, but I'm wondering if this's consistent with CMake conventions (Apple vs. Mac OSX) or if there're other places that need to adjust. Any thoughts?

chrisdembia:
Good point. The cmake documentation uses a mix of APPLE, MAC and MACOSX. MACOSX is used most frequently in their documentation. I don't have a strong preference for if we use `APPLE` or `MACOSX` in cmake; I think we use `APPLE`.

aymanhab:
Thanks @chrisdembia I'd then suggest we change the word OSX to Mac OSX which seems like a reasonable less cryptic name to show in docs and then merge.

chrisdembia:
@aymanhab I addressed your comment; please re-review and merge if ready.

PR #695, opened 2015-11-04
[cmake] find_package(Simbody) is not QUIET.

chrisdembia:
The messages from find_package(Simbody) were unnecessarily suppressed; sometimes CMake can give pretty useful information about why it couldn't find Simbody.


klshrinidhi:
:ok:

chrisdembia:
Thank you.

PR #700, opened 2015-11-12
Fix typo in OpenSim/Common/Test/testEformatParsing.trc

klshrinidhi:
This was caught while testing the TRCFileAdapter (coming soon).


tkuchida:
:+1:
Should the full path ("C:/Documents and Settings/Khatib/My Documents/Ellen/experiments-jan2012/feb 12/chris_mri-prot-writing-trial-11-twosheet-right.trc") in row 1 be replaced with just the file name, as in [this file](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Test/TRCFileWithNANs.trc)?

chrisdembia:
+1 if it doesn't affect the other tests that were using this file.

klshrinidhi:
Changed the filename in the header (first row).

tkuchida:
Looks good :+1:

PR #705, opened 2015-11-14
[cmake] Enumerated list for CMAKE_BUILD_TYPE.

chrisdembia:


sherm1:
What does that do?

chrisdembia:
In cmake-gui and ccmake, this change (1) shows users all 4 possible settings for this variable, and (2) restricts the variable to one of these 4 values. Thus, this change can help avoid typos and show users what the options are.

sherm1:
Sounds good. Looks mergeable to me.

PR #706, opened 2015-11-14
[swig python] Separate into 6 modules.

chrisdembia:
I separated the interface and header files into simbody, and the 5 OpenSim libraries. For python, I build 6 separate swig modules. For Java, we still just build a single module. So this PR really only affects the python wrapping.

My initial plan was to only separate out the simbody module, but it proved easy enough that I just continued on.

These changes reduce the time required to build the python wrapping by a factor of 2 (on an 8 core machine). Furthermore, these changes make it much easier to *develop* the python wrappings. Previously, you'd have to wait over three minutes after making a change to the interface file to see the effect. Now, you'll only need to rebuild the modules that were affected by your change. So, developing/improving the python bindings should be a significantly more enjoyable experience.

@aymanhab if you're interested in having separate modules for the Java wrapping as well, I'm happy to start on it, but I will need your assistance when it comes to assembling the modules in a logical way.

@aymanhab could you please review? If you'd prefer, I can split up the PR into smaller ones.


aymanhab:
@chrisdembia very nice job indeed. A few questions come to mind that were not clear from looking into the code (rather than building it)
1. Namespace concern in python (are our packages accessible as common, tools, etc.?) Would be good to avoid polluting the global namespace with these commonly used names if possible.
2. I'm assuming we're creating multiple bindings libraries, this could be error prone but more importantly if we changed the one-line import to multiple imports then it's a bit of inconvenience especially if most end users will only use a top level module rather than the finer granularity provided. We don't have that in C++ either so we end up using tools library since it pulls everything else, maybe a module opensim-all that pulls everything would solve this problem.
3. I love the fact that building sped up as a side effect and the development of the python API is now more enjoyable, python users and developers will be very grateful. :+1:
4. I'll open an issue regarding the library structure because it's not clear what's the long term vision for these since tools are going to diminish, as well as analyses but that shouldn't block this PR.

I'd wait until we have clear answer to these issues before trying the breakdown in Java (it's enjoyable already ;)

Please let me know what you think.

chrisdembia:
Thanks for looking at this PR.

> 1. Namespace concern in python (are our packages accessible as common, tools, etc.?) Would be good to avoid polluting the global namespace with these commonly used names if possible.

You can access all our classes directly from the `opensim` namespace. You should also be able to access classes from Common via `opensim.common.Component`, etc. However, I did not test that and we are not advertising it. Are you suggesting removing the ability to access classes using `opensim.common.Component`, `opensim.simulation.Model`, etc?

> 2. I'm assuming we're creating multiple bindings libraries, this could be error prone but more importantly if we changed the one-line import to multiple imports then it's a bit of inconvenience especially if most end users will only use a top level module rather than the finer granularity provided. We don't have that in C++ either so we end up using tools library since it pulls everything else, maybe a module opensim-all that pulls everything would solve this problem.

I completely agree and I would not have considered this change if it would have required users to import multiple libraries. This PR does *not* change anything about how users use the python package; this is evidenced by the fact that the existing python tests still pass. Of course, our coverage isn't too great...

> this could be error prone

I tried to do a few simple checks to make sure the modules interact with each other properly, and they seem to interact properly.

> 3. I love the fact that building sped up as a side effect and the development of the python API is now more enjoyable, python users and developers will be very grateful. :+1:

I love it too. The purpose of this PR is to speed up building and reduce the memory required for building the python wrapping; this behavior is not a side effect.

> 4. I'll open an issue regarding the library structure because it's not clear what's the long term vision for these since tools are going to diminish, as well as analyses but that shouldn't block this PR.

Since these changes do not change the interface of the python package, we should not need a separate PR. Also, if we rearrange OpenSim libraries in the future, we can do that without affecting the python interface.

> (it's enjoyable already ;)

Happy to hear that :) It does seem that the java bindings are generated and compiled quicker. It's also nice to have a separate file for each class instead of one HUGE opensim.py (for debugging).

aymanhab:
1. No I'm not suggesting we remove the access, I'm just making sure our classes don't end up in the global namespace as common.Vec3 (rather than opensim.Vec3 or opensim.common.Vec3 or opensim.simbody.Vec3?) since 'common, simulation, tools' are very generic and are likely to collide with others.

The issue of multiple C++ libraries, and what functionality goes to what library is unrelated to tyhis specific PR except as much as the 1-1 mapping so if the C++ layout changes, this one can follow.

I can merge if you don't have anything to change/add to this PR so we can keep moving. Thanks.

chrisdembia:
> I'm just making sure our classes don't end up in the global namespace as common.Vec3

Ohhh! In python, this choice is always left to the user. The following scripts are all possible:

```py
import opensim
v = opensim.Vec3()
v = opensim.simbody.Vec3()
```
```py
import opensim as osm
v = osm.Vec3()
v = osm.simbody.Vec3()
```
```py
from opensim import Vec3
v = Vec3()
```
```py
from opensim import *
v = Vec3()
v = simbody.Vec3()
```
The python community discourages this last one, but is fine with the others.

The following script is not possible:
```py
import opensim
v = common.Vec3()
```

chrisdembia:
This is ready to merge IMO.

chrisdembia:
Thank you!

PR #709, opened 2015-11-16
Data adapter classes

klshrinidhi:
Files to review --
* DataAdapter.h / DataAdapter.cpp
* FileAdapter.h / FileAdapter.cpp
* TRCFileAdapter.h / TRCFileAdapter.cpp
* Exception.h / Exception.cpp
* exampleTRCFileAdapter.cpp
* testTRCFileAdapter.cpp

Rest of files are still work in progress.

The implementation follows the [design](
https://docs.google.com/drawings/d/1jtvCUukkL8zMwtR6ae06hMIIfAPkKAgOeikyKT7xqNY/edit?usp=sharing) me & Ajay discussed, with following changes:
* Read functions return a collection of tables depending on the file format. This is to account for formats like C3D which contain multiple tables. The original design had these functions return just one table.
* For the same reason, write functions take a collection of tables, depending on the file format.
* Read/Write functions in DataAdapter class are not implemented.
* AdapterInstantiator class in the design is just a bool variable in an un-named namespace in the implementation.

The CI builds failed because BTK is not found. Will fix it when I submit C3DFileAdapter for review.

[Review on Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/709)


sherm1:
Review status: 0 of 20 files reviewed at latest revision, 13 unresolved discussions, some commit checks broke.

---

_[OpenSim/Common/DataAdapter.h, line 110 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/709#-K3RXhCi8hsA-7Kpq880:-K3RXhCje_F7zzSLB1Pr:807241915) ([raw file](https://github.com/opensim-org/opensim-core/blob/5679c7064d926df7df444b0734db6efdb1cd2470/OpenSim/Common/DataAdapter.h#L110)):_
The registry will likely need a few more support methods, something like `hasDataAdapter(identifier)` and probably some introspection methods that would allow making a list of the available adapters.

---

_[OpenSim/Common/FileAdapter.cpp, line 10 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/709#-K3R_4n-uMCXWNA7DJiZ:-K3R_4n-uMCXWNA7DJi_:1221352689) ([raw file](https://github.com/opensim-org/opensim-core/blob/5679c7064d926df7df444b0734db6efdb1cd2470/OpenSim/Common/FileAdapter.cpp#L10)):_
This casting doesn't make sense to me. Wouldn't this code work identically without it? (That is, you could just `return data_adapter.extendRead(fileName);`.) Also, how do you know that the returned DataAdapter is actually a FileAdapter?  I'm confused about the relationship between DataAdapter and FileAdapter. We should discuss in person.

Also, local variables in OpenSim use camelCase not underscore_separated.

---

_[OpenSim/Common/FileAdapter.h, line 73 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/709#-K3RYHshMIZntx7g9u9K:-K3RYHshMIZntx7g9u9L:625545178) ([raw file](https://github.com/opensim-org/opensim-core/blob/5679c7064d926df7df444b0734db6efdb1cd2470/OpenSim/Common/FileAdapter.h#L73)):_
Why isn't this an IOError?

---

_[OpenSim/Common/FileAdapter.h, line 89 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/709#-K3RavDftTtnpToyRlGj:-K3RavDftTtnpToyRlGk:-1643839101) ([raw file](https://github.com/opensim-org/opensim-core/blob/5679c7064d926df7df444b0734db6efdb1cd2470/OpenSim/Common/FileAdapter.h#L89)):_
I'm confused about why FileAdapter derives from DataAdapter, yet has only static members and doesn't override anything. Then some of the static members get access to a DataAdapter and then cast it to a FileAdapter??? I am going to need an in-person explanation to review further.

---

_[OpenSim/Common/FileAdapter.h, line 125 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/709#-K3RYyo2RBqlHPAFWI0A:-K3RYyo2RBqlHPAFWI0B:579258985) ([raw file](https://github.com/opensim-org/opensim-core/blob/5679c7064d926df7df444b0734db6efdb1cd2470/OpenSim/Common/FileAdapter.h#L125)):_
Shouldn't the methods tokenize() and getNextLine() be static?

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/709)_

sherm1:
I got lost in the inheritance logic going from DataAdapter to FileAdapter so I'll stop reviewing at this point and come by for an in-person explanation.

klshrinidhi:
I have made the following changes as per our discussions:
* Change the return type of read functions from a std::vector to std::unordered_map.
* Change the parameter type of the write functions from std::vector to std::unoredered_map.
* Added MOTFileAdapter and CSVFileAdapter following the same guidelines.
* C3DFileAdapter is ready too except for its tests (for which James is going to give some files to check against).

chrisdembia:
I think changing the compilers for CI should be its own PR and that maybe we should discuss it. We should consider which compiler versions are readily available on Ubuntu.

klshrinidhi:
@chrisdembia 
I was getting errors with clang-3.4 that could not reproduce. Once I see that CI builds successfully with latest compilers, I will undo changes to `.travis.yml` and start another PR for it.

Clang-3.7 binaries are available for both Linux and Mac.
GCC-5 binaries are available for Linux.

chrisdembia:
Oh okay; thanks for clarifying.

aseth1:
*[OpenSim/Examples/DataAdapter/exampleTRCFileAdapter.cpp, line 45 \[r4\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/709#-K9z_cQN42tOnWAkgEAm:-K9z_cQOE1cBiu-lYIGV:-1762572446) ([raw file](https://github.com/opensim-org/opensim-core/blob/f92c941512af035a22b68a1cefad6e6970eb22d2/OpenSim/Examples/DataAdapter/exampleTRCFileAdapter.cpp#L45)):*
The static cast to an internal data type of the TRCFileAdapter is unreasonable.  The point of having adapters is to decouple data (in the form of data tables) from their source (files and their formats) so this seems not ideal. A cast to a TimeSeriesTable or SimpleDataTable seems more reasonable. We could also introduce more commonly used types of tables like MarkerTable, ForceTable which would be aliases for TimeSeriesTable_<Vec3> for example with different meta data specifications.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/709)*

aseth1:
Review status: 0 of 39 files reviewed at latest revision, 21 unresolved discussions, some commit checks failed.

---

*[OpenSim/Common/C3DFileAdapter.h, line 44 \[r4\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/709#-KAC0VaGJ4S9EDI6_rKt:-KAC0VaHsVIjpSug8OCw:-717400991) ([raw file](https://github.com/opensim-org/opensim-core/blob/f92c941512af035a22b68a1cefad6e6970eb22d2/OpenSim/Common/C3DFileAdapter.h#L44)):*
The definition of DataTable types should not be nested inside a specific data adapter. We should think about a central location, perhaps even DataTable.h to define the most commonly used Table types. For example, how does someone using a C3D adapter know that `MarkerTable` is just a `TimeSeriesTable_<Vec3>`? In reality Markers and Forces are different things so they must have other fundamental differences, for example, Units: markers are in m, and forces are in N and moments are N*m. Because the validation of the DataTable is virtual a MarkersTable could test these conditions upon conversion.

---

*[OpenSim/Common/C3DFileAdapter.h, line 45 \[r4\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/709#-KACIf90b3gH4HqhVl8Z:-KACIf91dhJjIQVyN37T:-893256367) ([raw file](https://github.com/opensim-org/opensim-core/blob/f92c941512af035a22b68a1cefad6e6970eb22d2/OpenSim/Common/C3DFileAdapter.h#L45)):*
How is an EventTable read out of a C3D?

---

*[OpenSim/Common/C3DFileAdapter.h, line 47 \[r4\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/709#-KACIvHqtw-QyDKU-faX:-KACIvHrkUF2-YGVnUM3:-1027540275) ([raw file](https://github.com/opensim-org/opensim-core/blob/f92c941512af035a22b68a1cefad6e6970eb22d2/OpenSim/Common/C3DFileAdapter.h#L47)):*
What about other analog data: EMG, or other sensors and switches?

---

*[OpenSim/Common/C3DFileAdapter.h, line 48 \[r4\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/709#-KACFe7Qfh8H1eK4B3gZ:-KACFe7RyjZppBv4Yxv4:-1802396544) ([raw file](https://github.com/opensim-org/opensim-core/blob/f92c941512af035a22b68a1cefad6e6970eb22d2/OpenSim/Common/C3DFileAdapter.h#L48)):*
Why two ForceTables?

---

*[OpenSim/Common/C3DFileAdapter.h, line 61 \[r4\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/709#-KAC2n8Oc3L4vnBZrkqi:-KAC2n8PUWO1SKgoylVf:2066912195) ([raw file](https://github.com/opensim-org/opensim-core/blob/f92c941512af035a22b68a1cefad6e6970eb22d2/OpenSim/Common/C3DFileAdapter.h#L61)):*
This should be called `allTables` not markerTable.

---

*[OpenSim/Common/DataAdapter.h, line 73 \[r4\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/709#-K9zfHbMaUGbPDeguJ4w:-K9zfHbMaUGbPDeguJ4x:-1225220577) ([raw file](https://github.com/opensim-org/opensim-core/blob/f92c941512af035a22b68a1cefad6e6970eb22d2/OpenSim/Common/DataAdapter.h#L73)):*
typo still here

---

*[OpenSim/Examples/DataAdapter/exampleTRCFileAdapter.cpp, line 54 \[r4\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/709#-K9zcfp30lRpjgRAIzmn:-K9zcfp30lRpjgRAIzmo:-207794149) ([raw file](https://github.com/opensim-org/opensim-core/blob/f92c941512af035a22b68a1cefad6e6970eb22d2/OpenSim/Examples/DataAdapter/exampleTRCFileAdapter.cpp#L54)):*
I think the implementation of the DataTable meta data should not be exposed. We should have something like `table1->getMetaData<string>("DataRate")` and `table1->getMetaDataForColumn(1, "Units")` or by column name `table1->getMetaDataForColumn("Moment", "Units")` which I think is straightforward to do given the meta data structures in place.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/opensim-org/opensim-core/709)*

klshrinidhi:
To Do:
* Remove type aliases in adapter classes and use underlying types. 
 * `using TimeSeriesTableVec3 = TimeSeriesTable_<SimTK::Vec3>`.
 * `using TimeSeriesTable = TimeSeriesTable_<SimTK::Real>`.
* Add constructor to accept file name. Implementation would use `FileAdapter::readFile()` and downcast the received table appropriately.
* Move `struct Event` out into its own header file. Make a type named `DataTable_<Event>` which is returned by C3DFileAdapter. The key for this table will be "events".

Future To Do [Not for this PR]:
* Reading analog & EMG data from C3D.

klshrinidhi:
* Removed type aliases inside adapter classes. Introduced type aliases directly into OpenSim as discussed.
* Adding a construction to TimeSeriesTable to accept a file name to read is bit more work that I initially thought. I can get back on it in another PR.
* Moved the struct Event out of C3DFileAdapter class into its own header file.
* Cannot make DataTable of Event because SimTK::Matrix is for arithmetic types only.
* Got rid of user-forces extracted from C3D files to avoid confusion as discussed. Now C3D adapter will return markers and raw forces only.

klshrinidhi:
Tests on Linux are failing because the available C++ standard library does not fully support regex.

klshrinidhi:
@aymanhab 
I have pushed commits to make C3D support through BTK optional. This addresses your issue #766.

klshrinidhi:
The CMake variable `WITH_BTK` controls building with/without BTK. It is `OFF` by default.

chrisdembia:
I would like to follow a convention for CMake variables that configure how OpenSim is built. The previous convention was `BUILD_*` and I'm trying to move to a convention of `OPENSIM_*`. This is particularly helpful when using the CMake GUI, which allows you to group variables with the same prefix.

chrisdembia:
Also, for boolean variables, it may be more natural to use the CMake `option()` command, which creates a cache variable, etc.

klshrinidhi:
I can rename the variable but `BUILD_BTK` may not sound right because OpenSim will not build BTK. It is common to use `--with-<dependency>` when configuring projects on Linux.

chrisdembia:
I agree. I'm thinking `OPENSIM_WITH_BTK`

klshrinidhi:
Okay. I will rename the variable to `OPENSIM_WITH_BTK`. 
Maybe I should rename the variables `SUPERBUILD_*` to `BUILD_*` in `dependencies/CMakeLists.txt`.

chrisdembia:
> Maybe I should rename the variables SUPERBUILD_* to BUILD_* in dependencies/CMakeLists.txt.

I have no preference for that.

chrisdembia:
I mean, I like both options equally.

aseth1:
Noticed The following tests FAILED:
	  6 - testMOTFileAdapter (OTHER_FAULT)

@klshrinidhi what is the status of this error and the solution?

klshrinidhi:
@aseth1 
I think I have mentioned it before. The standard library in use does not support `regex` completely.

aseth1:
Yes, I remember now. What is the solution? Update clang/gcc to version on Ubuntu to one that supports `regex`?  Is it instrumental for using the MotAdapter (as a concrete DelimFileAdapter)? 
I imagine people will want to load .mot and .sto files the most in the hackathon.

klshrinidhi:
Yes. I believe that is the solution. I even had PR for it --> #737.

chrisdembia:
My preference is still to use the minimum compiler versions necessary, and to edit the travis script in a way that still allows us to use ccache (if possible; this means `sudo: false`).

aseth1:
@klshrinidhi Unfortunately it seems that upgrading to the latest compilers was too large a leap that would alienate developers on older compilers that come with their Ubunto distro. How much of the `regex` is really necessary vs. just easier? If we can have a solution that works for what is a minimum expectation (i.e. the version of gcc that comes with the current LTS) that seems like a better solution atm.

chrisdembia:
I'm fine with upgrading to gcc 4.9 and clang 3.5, which I think are the requirements for simbody anyway (despite the documentation). This could be done with very few changes to .travis-ci.yml; see what I tried out for Simbody here: https://github.com/simbody/simbody/compare/master...chrisdembia:travis-gcc49

aseth1:
@klshrinidhi do the upgrades that @chrisdembia suggests (as required by Simbody) sufficient to support what you need for `regex`? If so let's do that.

chrisdembia:
(FYI Sherm upgraded the required compilers for regex also).

aseth1:
@aymanhab do you have any objections to these upgrades? I seems like it is already a requirement for Simbody and makes most sense for OpenSim to be in sync anyways.

klshrinidhi:
@aseth1 
I don't agree.
It looks like we are okay for people to take hours to download and install latest XCode and Visual Studio, which don't come with the OS. But we are not okay to take less than a minute to upgrade compilers on Ubuntu to latest versions, which can actually co-exist with their current compilers without problems.
I will make the change to upgrade compilers to whatever Simbody users.
Also, I think less than 4% of OpenSim users are Ubuntu users. This is really applicable to probably handful of people.

aseth1:
> @aseth1 I don't agree.

OK lets discuss in person. I will be there shortly.

aymanhab:
I have no objection since it is required for Simbody already. The sentiment is to keep at the lowest commonly supported version that we can get away with rather than chase the latest advances in the language and compilers for all practical reasons and considering our user community.

klshrinidhi:
#819 needs to get merged so this can compiler on travis.

klshrinidhi:
I got it all :white_check_mark: !!!!

PR #711, opened 2015-11-18
Working version of TaskSpace controller for OpenSim

mitkof6:
The computation of the task space was previously implemented by @chrisdembia and @sherm1 for Simbody as an example. I have ported it as an OpenSim ModelComponent, because of the way the cache is allocated by Simbody, is not consistent with OpenSim model building, which probably discards any allocation when the model is built. Correct me if I am wrong, but I believe it is because there is an extra mechanism that wraps the allocation through Component addCacheVariable. 

I have converted all computations so as not to have an internal state reference, but this being a part of the function call. As a consequence the convenient operators were converted to function calls. 

For this test the arm's end effector is specified to have a constant acceleration in the y direction and the gravity compensation is considered.

I would like to discuss with you how this can be improved and consequently integrated into OpenSim, because I find it very useful. Some suggestions that I have in mind:

* Find a way to keep the original operators of the task space quantities (e.g. JT * u), as in the original implementations, but without having an internal reference to the state or if there is a reference state, to ensure that it is correctly used for the computations.

* In the current implementation the TaskSpace class is a model component, it may be a Force so it can be used as a controller during integration. But I think the latter is restrictive because, depending on the Force controller someone may be interested to have direct access to the underlying mathematics (e.g. disable gravity compensation).

* In the TaskSpace.cpp method calcInverseDynamics() there is a TODO for the transformation of extra forces that may act on the model. Is there a generic way to have access to those forces and consider them in the computations.

* Solve the problem of finding the torques that satisfy multiple task with the same priority.


chrisdembia:
This is awesome, @mitkof6 . The travis-ci tests failed becaused you put tabs in OpenSim/Sandbox/CMakeLists.txt; convert them to spaces.

We are interested in having some sort of support for task space control in OpenSim. We have wanted it in OpenSim for a long time and we have some ideas for cool ways to use it. For example, OpenSim users should be able to reorient a model in the GUI by clicking a point on a body and moving it; this could be done easily with task space control. Task Space control could also be an alternative way of performing IK (and perhaps RRA).

Do you anticipate supporting hierarchies / null spaces? If so, @sherm1 and I discussed some ideas for how to improve the abstractions.

I'll look through the PR in the next few days. I think we should merge this fairly quickly, and use it as a starting point for discussions about the design/implementation of a proper OpenSim feature.

I think the project of integrating task space control in OpenSim will be more successful if you have a personal need for the feature (if it is related to your research, etc.).

OpenSim/Sandbox/CMakeLists.txt
```diff
@@ -18,6 +20,10 @@ endforeach()
 foreach(exec_file ${TO_COMPILE})
     get_filename_component(_target_name ${exec_file} NAME_WE)
     add_executable(${_target_name} EXCLUDE_FROM_ALL ${exec_file})
+	if(${exec_file} MATCHES "futureTaskSpace.cpp")
+		add_library(taskSpaceLib STATIC ${TASK_SPACE_FILES})
+		target_link_libraries(${_target_name} taskSpaceLib)
+	endif()
```
chrisdembia: It might be cleaner to move `futureTaskSpace.cpp` to the `TaskSpace` folder and build it there?

mitkof6:
> This is awesome, @mitkof6 .

Actually, the credit is yours, I just mess up a little bit with the implementation :).

> We are interested in having some sort of support for task space control in OpenSim. We have wanted it in OpenSim for a long time and we have some ideas for cool ways to use it. 

To me it seems as a cool way to preform forward dynamic simulations.

>  For example, OpenSim users should be able to reorient a model in the GUI by clicking a point on a body and moving it; this could be done easily with task space control.

Useful. Also, you can add a feature to report a desired position on a specific body. Currently not possible without the marker through the GUI.

> Do you anticipate supporting hierarchies / null spaces? If so, @sherm1 and I discussed some ideas for how to improve the abstractions.  
> I think the project of integrating task space control in OpenSim will be more successful if you have a personal need for the feature (if it is related to your research, etc.).

Currently, I don't have many tasks. But surely, it should be integrated in later versions.

mitkof6:
Also, there is this issue that I came up with, where if the coordinates are locked, still the algorithm computes the torques of the locked joints. I have tried locking some coordinates (e.g. the base) and the resultant motion is wrong, because the algorithm believes it can use the locked coordinate to achieve the task. On the other hand, when unlocked, the correct motion is produced by actuating every joint. This is a bit restrictive because I may be interested to lock some degrees of freedom and to see what will be the outcome. I am not sure if a locked coordinate must be able to produce torque.

mitkof6:
I have encountered some problems when some joint coordinates are locked. Should the algorithm compute torques for the locked coordinates, or should it try to ignore them in the computation of the task torques?

chrisdembia:
Task space control has a proper way to deal with unactuated coordinates; we should incorporate that. Might be called a "selection matrix".

sherm1:
Underactuated coordinates are harder to deal with, but I think locked coordinates could be considered actuated. The locking constraint can be viewed as a high-priority task operating directly in joint space (its Jacobian is identity). The actual tasks could be done in the null space of the joint-locking task. I'm not sure whether this would be easy to do in the current framework.

mitkof6:
This seems as an interesting problem. Does the locking of the joints impose some constraint on the multibody internally?

> Might be called a "selection matrix".

@chrisdembia would this work as a filter and map the full system of n DOFs to a smaller subset of free (unlocked) variables? Maybe all the properties can be multiplied by the selection matrix S so we can have a mapping.

> The locking constraint can be viewed as a high-priority task operating directly in joint space (its Jacobian is identity). 

How will be the task defined?

chrisdembia:
Dimitar, I think Luis Sentis' thesis should answer your questions. http://www.me.utexas.edu/~lsentis/files/Thesis-Sentis-2007.pdf @sherm1 likes Sentis' thesis, if I remember correctly.

sherm1:
>  @sherm1 likes Sentis' thesis, if I remember correctly.

Yes, Luis is awesome. But as I mentioned in the OSRF report, I think the attempts to incorporate everything into the operational space inertia goes too far. For example, it leaves no good way to deal with torque limits so they typically just truncate, which is suboptimal. Better would be simply to set up the problem using the op space approach, but then solve it directly as an optimization problem to obtain an optimal solution in the presence of underactuation and torque limits.

sherm1:
>>The locking constraint can be viewed as a high-priority task operating directly in joint space (its Jacobian is identity).

>How will be the task defined?

I don't know if it would fit directly in the existing TaskSpace framework, but it is easily accommodated by the theory I outlined in the OSRF report, as a "priority 1" task, meaning one of the system constraints.

OpenSim/Sandbox/TaskSpace/TaskBasedController.cpp
```diff
@@ -0,0 +1,66 @@
...
+void TaskBasedController::computeForce(const State& s,
+    Vector_<SpatialVec>& bodyForces, Vector& generalizedForces) const
+{
+    if (!m_isInitialized)
+    {
+        m_taskSpace->addStationTask(
+            _model->getBodySet().get(m_endEffectorBody).getMobilizedBodyIndex(),
+            m_endEffectorBodyOffset);
+        m_isInitialized = true;
+    }
```
chrisdembia: I think this should happen in `addToSystem()` or `connectToModel()`.

chrisdembia:
My main comment is that I think the TaskSpace class should be a subcomponent of the TaskSpaceController. Better yet, the TaskSpace class would remain in Simbody. I don't think there needs to be such a class in OpenSim.

I'm going to merge this. Thanks @mitkof6.

mitkof6:
Do you happen to know how I can account for the muscle forces? When the muscles are not presented the computed torques can actuate the skeletal system correctly, but when the muscles are presented they apply forces, which are not considered in the inverse computations. Should I use: ~J * R * fm to map into task space, or there is a better way?

chrisdembia:
To be clear, you're not trying to control the muscles but just account for the forces they apply? Then yeah I think you're right, you just need to subtract off their moments.

mitkof6:
Actually, I am implementing a muscle controller, following the task space command. It is currently partially working. Fortunately, I need to compensate for any acting force on the skeletal system. So, basiclly, correct me if I am wrong, should I add an extra term in the task space equation, which will project all the moments into the task? Or should the resultant moment of each force be projected by the null space matrix and then subtracted from the computed torque?

For the muscles I can compute the moment arm, but for the other forces, how can I compute the torques that they produce? Can I have a direct access to the Jacobian or is there some function that computes that for me, so I can iterate for each force and get the resultant moment?

I think if I get this correct it will be an interesting feature for OpenSim.

chrisdembia:
@mitkof6 you're delving into new research here, combining task space with muscles. Emel Demircan has done the most work combining task space with muscles, as far as I know. However, she only performed an inverse analysis: back-computing the muscle forces that would have been able to drive a motion.

I think I would need to see some equations and math to understand what you're talking about.

How are you dealing with the activation dynamics/delay? Are you ignoring activation dynamics?

mitkof6:
Basically, I have managed to implement a first version of a task oriented muscle driven simulation. The idea is similar to CMC, where the tracking tasks are in task space, and I have closed the loop with an optimal controller (could be any controller). I compute the torques following task space formulation and the torques are used with static optimization, which I reimplemented (can be an arbitrary objective function). 

Unfortunately, I have made many simplifications to the model (upper limb for reaching tasks), so it can work properly. I have encountered many problems that need to be solved so this approach can be generalized and used for any model. Namely, I have removed every constraint in the model, I have discarded every locked joint (also some joints can be made wield), I have weakened the muscle passive forces and I have ignored activation dynamics. The latter depends if the controller can provide a future estimate of the task. 

The equations in task space must be rewritten to account for: 1) constraints (not sure of it should be formulated as in the DARPA report), muscle and other forces, 2) we need to implement a priorization algorithm (as discussed previously) 3) to implement a selection matrix or to constraint the locked joints. On the other hand, for the resultant torques, we need to add extra compensation terms (similar to the gravity compensation) to cancel out each term of the equation of motion (included constraint forces) in the null space of the task.

I find this very useful and I have many ideas how this can be used in inverse and in a forward manner. Currently, I am going to use this to reproduce some recorded tasks in a forward manner so I can predict the impact of some parameters. I remembered that you mentioned that you are interested in predictive simulations and I think this framework can be the base for it. If you are interested in this and have time it would be a pleasure to work with you. I can try to write a small description with the equations so we can discuss it further.

chrisdembia:
The tasks you list (1-3) should all be possible, given existing literature on task space or operational space control. What sounds more difficult is including things like activation dynamics, etc. I'm most interested in this for interactive simulation (to use a GUI to tell the model where to move). However, I don't have time to work on it right now :/. I'd like to continue discussing it with you, though, and to see how it can fit in OpenSim.

mitkof6:
I have a question, if you could help me with. How can I compute the constraint term ~Gλ? I think there is a method in SimbodyMatterSubsystem, but how I get the multipliers also, which should be computed during forward dynamics?

sherm1:
> How can I compute the constraint term ~Gλ? I think there is a method in SimbodyMatterSubsystem, but how I get the multipliers also, which should be computed during forward dynamics?

λ=[`State::getMultipliers()`](https://simtk.org/api_docs/simbody/latest/classSimTK_1_1State.html#a393f6dc9c403f22f7209b5328820c356) is probably what you are looking for. That returns a reference into the State cache entry that contains the multipliers calculated in the last `state.realize(Acceleration)` call. You can then feed the result to [`SimbodyMatterSubsystem::multiplyByGTranspose()`](https://simtk.org/api_docs/simbody/latest/classSimTK_1_1SimbodyMatterSubsystem.html#a3d2451d87d7cae837062cd9a036bf698) to get `~Gλ`.

You can also use λ with [`SimbodyMatterSubsystem::calcConstraintForcesFromMultipliers()`](https://simtk.org/api_docs/simbody/latest/classSimTK_1_1SimbodyMatterSubsystem.html#a6cae3da23f051dc8116fbf0e054dd302) to get the constraint forces as applied to bodies and joints rather than as generalized forces.

mitkof6:
Do you happen to have a convenient method to work with spatial tasks for the Jacobian. Currently, I use calcStationJacobian (3xN), but in case of a task that is both a position and orientation descriptor I am unable to calculate the Jacobian (6xN).

chrisdembia:
I think the `calcFrameJacobian` variants are what you want.

chrisdembia:
https://simtk.org/api_docs/simbody/3.5/classSimTK_1_1SimbodyMatterSubsystem.html#a741150aaf33c1b667067b11f3cb59657

mitkof6:
> I think the calcFrameJacobian variants are what you want.

Thanks, I missed this one.

PR #712, opened 2015-11-20
[Don't merge this] Sketch of formatters and adapters for discussion.

sherm1:
@aseth1, @klshrinidhi: I took a crack at sketching the classes we discussed on the whiteboard. Came out a little differently than I thought. Basic structure looks like this, with two main class families:
```
DataStreamFormatter
      TRCDataStreamFormatter
      MotDataStreamFormatter
          ...

DataAdapter
      StreamDataAdapter
          StringDataAdapter
      FileDataAdapter
      (plus two helper classes)

There is also C3dFormatter which is just a standalone wrapper for BTK and is *not* a DataStreamFormatter. 
```
Take a look at the header file associated with this PR. It is not compilable but should work for discussion in this PR.

### DIFF
--- a/OpenSim/Sandbox/DataFormattersAndAdapters.h
+++ b/OpenSim/Sandbox/DataFormattersAndAdapters.h
@@ -170,7 +170,7 @@ the DataStreamFormatters. Users will use the FileDataAdapter instead. **/
 class FileStreamDataAdapter : public StreamDataAdapter {
 public:
     FileStreamDataAdapter(const std::string& pathName)
-    :   StreamDataAdapter(std::fstrea(pathName), 
+    :   StreamDataAdapter(std::fstream(pathName), 
                           findExtension(pathName)) {}
 };
 
--- a/OpenSim/Sandbox/DataFormattersAndAdapters.h
+++ b/OpenSim/Sandbox/DataFormattersAndAdapters.h
@@ -146,7 +146,7 @@ public:
     registry. **/
     StreamDataAdapter(iostream& stream, const std::string& formatKey) 
     :   m_stream(stream) {
-        m_formatter = createDataStreamFormatter(formatKey);
+        m_formatter = DataStreamFormatter::createFormatter(formatKey);
     }
 private:
     OutputTables implementRead() const override 
--- a/OpenSim/Sandbox/DataFormattersAndAdapters.h
+++ b/OpenSim/Sandbox/DataFormattersAndAdapters.h
@@ -1,0 +1,212 @@
+
+// Two main class families here:
+//      DataStreamFormatter
+//          TRCDataStreamFormatter
+//          MotDataStreamFormatter
+//          ...
+//
+// and  DataAdapter
+//          StreamDataAdapter
+//              StringDataAdapter
+//          FileDataAdapter
+//          (plus two helper classes)
+//
+// There is also C3dFormatter which is just a wrapper for BTK; it is not
+// a DataStreamFormatter or DataAdapter. Alternative: create an abstract
+// DataFormatter base class and derive DataStreamFormatter and C3dFormatter
+// from it; I don't think that's necessary.
+
+
+//==============================================================================
+//                    DATA STREAM FORMATTER classes
+//==============================================================================
+/** Read/write DataTables from/to a stream whose format is known by the 
+concrete DataStreamFormatter object. Also maintains a static registry of these
+objects indexed by a string identifying the expected format, and serves as
+a factory for concrete DataStreamFormatter objects. 
+
+Note that concrete objects of this type are associated with a *format*; they
+are not associated with any particular *stream*. The stream is supplied at the
+time it is to be read from or written to, and the same Formatter object can be
+used for different streams as long as they have the right format.
+**/
+class DataStreamFormatter {
+public:
+    // Use non-virtual interface design pattern.
+
+    /** Given an input stream expected to be in the format associated with this 
+    Formatter, read DataTables out of it. The particular set of DataTables 
+    returned will vary based on the format and the stream contents. **/
+    OutputTables read(std::istream& in) const;
+
+    /** Given an output stream, append the given tables to it in the format
+    associated with this Formatter. **/
+    void write(std::ostream& out, InputTables);
+
+    // Registry
+
+    /** Use the Formatter registry to create an appropriate stream formatter
+    for the given format key. **/ 
+    static unique_ptr<DataStreamFormatter> 
+        createFormatter(const std::string& formatKey);
+
+    /** Return true if there is a Formatter suitable for the given format. **/
+    static bool hasFormatter(const std::string& formatKey);
+
+    /** Transfer ownership of the formatter object to the registry and associate
+    it with the given key, which is typically a file extension. **/
+    static void registerDataStreamFormatter
+       (const std:string&                 formatKey,
+        unique_ptr<DataStreamFormatter>&& formatter);
+
+    // Bookkeeping stuff
+
+    virtual ~DataStreamFormatter() = default;
+
+    /** Produce an identical copy of this concrete DataStreamFormatter. Internal
+    state, if any, is reset in the copy to just-constructed condition. **/
+    unique_ptr<DataStreamFormatter> clone() const {return implementClone();}
+protected:
+    DataStreamFormatter() = default;
+
+    // Virtual interface is protected.
+    virtual unique_ptr<DataStreamFormatter> implementClone() const = 0;
+    virtual OutputTables implementRead(std::istream&) const = 0;
+    virtual void implementWrite(std::ostream&, InputTables) = 0;
+private:
+    // The static registry map.
+};
+
+/** This is a DataStreamFormatter that understands the trc format. **/
+class TRCDataStreamFormatter : public DataStreamFormatter {
+public:
+    TRCDataStreamFormatter();
+
+private:
+    unique_ptr<DataStreamFormatter> implementClone() const
+    {   return TRCDataStreamFormatter(*this); }
+
+    OutputTables implementRead(std::istream& trcIn) const override;
+    void implementWrite(std::ostream& trcOut, InputTables) override;
+};
+
+/** This is a DataStreamFormatter that understands the mot format. **/
+class MotDataStreamFormatter : public DataStreamFormatter {
+public:
+    MotDataStreamFormatter();
+
+private:
+    unique_ptr<DataStreamFormatter> implementClone() const
+    {   return MotDataStreamFormatter(*this); }
+
+    OutputTables implementRead(std::istream& motIn) const override;
+    void implementWrite(std::ostream& motOut, InputTables) override;
+};
+
+//==============================================================================
+//                         C3D FORMATTER
+//==============================================================================
+/** A standalone wrapper class for the BTK library; this is not a 
+DataStreamFormatter. **/
+class C3dFormatter {
+public:
+    // Useful stuff for dealing with BTK.
+};
+
+//==============================================================================
+//                       DATA ADAPTER classes
+//==============================================================================
+/** Associated with a data source/sink of a known type and can read and write
+DataTables from/to that source/sink. 
+
+Note that these objects are associated with both a format and a data source,
+which might be a file or cin or a string or whatever.
+**/
+class DataAdapter {
+public:
+    OutputTables read() const;
+    void write(InputTables);
+    virtual ~DataAdapter() = default;
+
+protected:
+    DataAdapter() = default;
+    virtual OutputTables implementRead() const = 0;
+    virtual void implementWrite(InputTables) = 0;
+};
+
+/** This is a DataAdapter for use with an already-existing read/write stream.
+For read-only streams (istream) like std::cin or write-only (ostream) like
+std::cout we'll need IStreamDataAdapter and OStreamDataAdapter also. Those
+should raise exceptions if disallowed read() or write() is called.
+**/
+class StreamDataAdapter : public DataAdapter {
+public:
+    /** Associate this adapter with a stream and a format for that stream. It
+    is an error if the formatKey cannot be found in the DataStreamFormatter
+    registry. **/
+    StreamDataAdapter(iostream& stream, const std::string& formatKey) 
+    :   m_stream(stream) {
+        m_formatter = createDataStreamFormatter(formatKey);
+    }
+private:
+    OutputTables implementRead() const override 
+    {   return m_formatter->read(m_stream); }
+    void implementWrite(InputTables input) override 
+    {   return m_formatter->write(m_stream, input); }
+
+    std::iostream&                  m_stream;
+    unique_ptr<DataStreamFormatter> m_formatter;
+};
+
+/** A string adapter is a kind of stream adapter. **/
+class StringDataAdapter : public StreamDataAdapter {
+public:
+    StringDataAdapter(const std::string& data, const std::string& formatKey)
+    :   StreamDataAdapter(std::stringstream(data), formatKey) {}
+};
+
+/** Helper class for when you know this file will be handled using one of
+the DataStreamFormatters. Users will use the FileDataAdapter instead. **/
+class FileStreamDataAdapter : public StreamDataAdapter {
+public:
+    FileStreamDataAdapter(const std::string& pathName)
+    :   StreamDataAdapter(std::fstrea(pathName), 
+                          findExtension(pathName)) {}
+};
+
+/** Helper class for when you know this file will be handled using the
+C3dFormatter. Users will use the FileDataAdapter instead. **/
+class C3dFileDataAdapter : public DataAdapter {
+public:
+    C3dFileDataAdapter(const std::string& c3dFileName) 
+    :   m_formatter(c3dFilesName) {}
+private:
+    OutputTables implementRead() const override 
+    {   return m_formatter->read(m_stream); }
+    void implementWrite(InputTables input) override 
+    {   return m_formatter->write(m_stream, input); }
+
+    C3dFormatter m_formatter;
+};
+
+
+/** Instantiate a DataAdapter for a particular file given by name. This class 
+figures out the appropriate format to use for the file. **/
+class FileDataAdapter : public DataAdapter {
+public:
+    FileDataAdapter(const std::string& pathName) {
+        const std::string ext = findExtension(pathName);
+        if (ext == "c3d") {
+            m_adapter.reset(new C3dFileDataAdapter(pathName));
+        } else
+            m_adapter.reset(new FileStreamDataAdapter(pathName));
+    }
+private:
+    // These just forward to the actual DataAdapter in use.
+    OutputTables implementRead() const override 
+    {   return m_adapter->read(); }
+    void implementWrite(InputTables input) override 
+    {   return m_adapter->write(input); }
+
+    unique_ptr<DataAdapter> m_adapter;
+};


sherm1:
One thing I found awkward about this design is the mixing of reading and writing. That's a problem because, for example, `std::cin` is an `istream`, so it cannot be passed to something that expects an `iostream`. Also it is very likely that sometimes we will only have read access to files in which case we can obtain an `istream` but not an `ostream`. Similarly, `std::cout` is an `ostream`.

I think the design needs to be extended either to include separate read and write classes, or to construct with access flags saying whether we want read/wrote/both access that can be managed at run time. Any ideas for how to do that?

OpenSim/Sandbox/DataFormattersAndAdapters.h
```diff
@@ -0,0 +1,212 @@
...
+/** Instantiate a DataAdapter for a particular file given by name. This class 
+figures out the appropriate format to use for the file. **/
+class FileDataAdapter : public DataAdapter {
+public:
+    FileDataAdapter(const std::string& pathName) {
+        const std::string ext = findExtension(pathName);
+        if (ext == "c3d") {
+            m_adapter.reset(new C3dFileDataAdapter(pathName));
+        } else
+            m_adapter.reset(new FileStreamDataAdapter(pathName));
```
chrisdembia: How do I register a data format for MAT or HDF5? I probably don't want to use `FileStreamDataAdapter`, and I think we should expect third parties will want to write support for formats that opensim itself does not support.

OpenSim/Sandbox/DataFormattersAndAdapters.h
```diff
@@ -0,0 +1,212 @@
...
+/** A string adapter is a kind of stream adapter. **/
+class StringDataAdapter : public StreamDataAdapter {
+public:
+    StringDataAdapter(const std::string& data, const std::string& formatKey)
+    :   StreamDataAdapter(std::stringstream(data), formatKey) {}
+};
+
+/** Helper class for when you know this file will be handled using one of
+the DataStreamFormatters. Users will use the FileDataAdapter instead. **/
+class FileStreamDataAdapter : public StreamDataAdapter {
```
chrisdembia: It seems like there is a distinction between plain text files and non-plain-text files. Is this true? If so, it'd be helpful if that were apparent from the class names.

chrisdembia:
Could you share some background, or perhaps some user code? My understanding is that the Adapters are user-interface and the Formatters are implementation; users never have to construct a formatter? And further, that people who want to add a new plain-text format need only write a Formatter, and then can automatically leverage the Adapter.

Is the decoupling of adapters and readers to accommodate C3D?

What if I wanted to write a DataAdapter that creates DataTables from Eigen matrices or some other in-memory database (not from a file)? Would I only need a DataAdapter and not a Formatter?

I find "Formatter" to be a clear name, but I find "Adapter" to be too abstract to mean anything specific to me (even it's technically the correct design pattern term).

OpenSim/Sandbox/DataFormattersAndAdapters.h
sherm1: > How do I register a data format for MAT or HDF5?

In general I hope everything can use a DataStreamFormatter; the only reason we didn't handle c3d that way is apparently BTK can't take a stream as input (although I suspect we could change it easily to do so). So you would write MatDataStreamFormatter and HDF5DataStreamFormatter. At least that's what I am thinking at the moment.

OpenSim/Sandbox/DataFormattersAndAdapters.h
sherm1: > It seems like there is a distinction between plain text files and non-plain-text files. Is this true? 

No, there should be no distinction. The particular format implies whether the expected stream will be treated as binary or text. But otherwise they are both just streams of bytes. In code, the distinction occurs only at the time a file is opened, and the only significance is that line ending characters are meaningful in text files but not in binary files.

sherm1:
> Could you share some background, or perhaps some user code? 

I think all of this code is internal. It should be seen only by people who are adding new formats.

> Is the decoupling of adapters and readers to accommodate C3D?

The treatment of c3d is a hack; it would be much cleaner if we could get BTK to work with a stream. Then we would just have a c3dDataStreamFormatter and the whole thing would make more sense! @klshrinidhi, how hard would it be to make BTK start with a stream or to get its author to allow that?

> What if I wanted to write a DataAdapter that creates DataTables from Eigen matrices or some other in-memory database (not from a file)? Would I only need a DataAdapter and not a Formatter?

Right! You would still have to implement read() and write() methods that would populate the DataTables from the internal database.

> I find "Formatter" to be a clear name, but I find "Adapter" to be too abstract to mean anything specific 

I agree. We're not using "Adapter" alone though. How about "DataTableAdapter" to make it more specific and less surprising that it takes and returns DataTables? That's still not great. It is really something like ExternalSource2DataTableConverter but even that is insufficient since it both reads and writes.

chrisdembia:
> How about "DataTableAdapter" to make it more specific and less surprising that it takes and returns DataTables?

It's the word *adapter* that I don't like, no matter what surrounds it. I like converter. `DataFormatConverter`? I guess adapter is okay if I think about electrical adapters instead of the OOP concept (even though they're the same?).

chrisdembia:
> The treatment of c3d is a hack;

Then I may not understand the need to separate formatter and adapter. Perhaps I should have instead asked "Is the decoupling of adapters and formatters primarily to accommodate formats that we cannot easily read from a stream?" Is this the picture: "To create a DataTable from your own format, write a DataAdapter. By the way, if your format can produce a stream, we have an infrastructure to make it even easier to support your format."?

I don't see the need for the separate `C3DFormatter` class; seems that all the C3D parsing could go in the `C3DFileDataAdapter`. Is the reason just so that all classes that do parsing have "formatter" in the name?

sherm1:
> Is the decoupling of adapters and formatters primarily to accommodate formats that we cannot easily read from a stream?

No. The idea was to split the two concepts of "data source" from "formatting". Lots of data sources (file, string, socket, std::cin) could use the same format. C++ has streams exactly for the purpose of abstracting this common serialization concept, so once you have a stream you can get bytes out of it without knowing the source. Then separately there is the question of how to interpret that stream of bytes; that's the format.

The stream concept is a good abstraction and can be used with *any* file, for example. It is only a quirk of BTK that the external interface apparently won't work with an already-open c3d stream, so we can't implement DataStreamFormatter's read(stream) and write(stream) methods. Any C++ I/O library should work with streams (and no doubt it does internally). So it could be worth messing around with BTK to get it in that format in which case we could remove the hacks.

>"To create a DataTable from your own format, write a DataAdapter. By the way, if your format can produce a stream, we have an infrastructure to make it even easier to support your format."

I would rather be able to say "write a DataStreamFormatter that understands your format; register the file extension; and we'll do the rest." For someone to do it as you described, it wouldn't be enough to write a DataAdapter, they would also have to somehow hack FileDataAdapter to be able to recognize the extension.

I think we should investigate whether BTK could be upgraded to work properly with streams, and if so try to insist that every Formatter has to work on a stream.

sherm1:
> I don't see the need for the separate C3DFormatter class; seems that all the C3D parsing could go in the C3DFileDataAdapter. Is the reason just so that all classes that do parsing have "formatter" in the name?

Yes. I just wrote it that way as an attempt to pretend that there was still a source/format separation. No doubt we can come up with a cleaner solution.

sherm1:
> It's the word adapter that I don't like. I like converter. 

`DataTableConverter` sounds good to me. It converts a file, string, stream, or other external representation to or from DataTables.

sherm1:
@aseth1, since the proposed read() and write() methods seem designed to read or write the whole file at a time, is there really any reason for them to exist? Would it make sense just to the read or write as part of construction? That would make it seem more like a "converter". Or is the idea that more than one read or write could be performed somehow?

chrisdembia:
> Or is the idea that more than one read or write could be performed somehow?

Do we have a place in the repo that contains a sufficient number of use cases that would answer this question for us?

sherm1:
> Do we have a place in the repo that contains a sufficient number of use cases that would answer this question for us?

Since these are very internal classes, the use cases would be some kind of internal code. As I understand them they are used to populate DataTables. DataTables are internal classes that I think are intended to make it easy to create DataSource and DataSink Components. If that's right, those are clearly unidirectional so I would think we could have something like DataTablesToExternal and DataTablesFromExternal classes, or maybe SerializeDataTables and DeserializeDataTables or something like that rather than combining the directions.

chrisdembia:
> the use cases would be some kind of internal code.

That would still help answer the question about the necessity for `read()` and `write()`.  Also I think we expect users to use DataTables directly (C++ users and scripters), though of course not if DataSink/Source suffice for their task.

aseth1:
> @aseth1, since the proposed read() and write() methods seem designed to read or write the whole file at a time, is there really any reason for them to exist?

That was not my initial intent. In the existing use cases of reading markers from a file or writing out joint angles that does seem to be all that is necessary.  In my initial design the rough idea was that adapter could persist and multiple read calls would continue to fill up the DataTable it was bound too.

``` cpp 
        CSVFileReader csvReader("dataTestFile.csv");
        ASSERT(csvReader.isWriteAccess() == false);
        ASSERT(csvReader.isReadAccess() == true);
        DataTable csvTable;
        csvReader.prepareForReading(csvTable);
        csvReader.read();

        // Try to read the table in again. Should throw an exception
        // about no more lines to read from file.
        ASSERT_THROW(std::exception, csvReader.read());
        // If we set the file to be loaded it must close and open the new.
```

The first `read()` would read to the end of the file, which the assert checks. But if the file is being written too by another process, a subsequent could be successful, in which case the `csvTable` would grow- not be reread from the beginning. 'DataAdpater::prepareForReading' was intended to i) provide the DataTable to be read/written and ii) throw an exception if the source (in this case) the file could not be accessed and iii) allow the adapter to take advantage of any file or column MetaData in the DataTable for helping in the read/write task. Of course, in the read case it would require the MetaData to be read in first, but that seems OK to me once the source is accessed successfully. 

I suggest we revisit some of 'features' in the initial prototype [here](https://github.com/opensim-org/opensim-core/commit/e9238fa1d6150d637daba14b18ccd22950dae879) which also included knowledge of the access mode.

sherm1:
> In my initial design the rough idea was that adapter could persist and multiple read calls would continue to fill up the DataTable it was bound too.

Makes sense. That model has changed somewhat since we now expect a single adapter to deal with a whole collection of DataTables. That doesn't seem to map as well onto an ongoing reading process. Maybe we could start thinking of these as "converters" as Chris suggested rather than "adapters". Then the idea would be that they simply take some external serialization of interesting stuff and convert it to an internal collection of DataTables containing the same stuff.

sherm1:
Here's a new question I think: if a file now maps to a collection of DataTables, where does the file metadata go? It seems strange to have it associated with any one of the resulting DataTables.

chrisdembia:
> Here's a new question I think: if a file now maps to a collection of DataTables, where does the file metadata go? It seems strange to have it associated with any one of the resulting DataTables.

It would be natural to have a data container object (maybe just called `Data` or `DataBook`) that contains (1) metadata, and (2) a collection of data tables (which don't have metadata), but maybe DataSource/Sink is just the container.

There seem to be two models out there. HDF5, MAT, and C3D files are collections of data tables. Plain text files (mot, trc, csv) are most reasonably just a single data table. Perhaps we should allow some readers to create just a DataTable (it's a pain for me to have to go through some data table container if I'm reading a csv and there would always be just one table), and others would create the Data container of DataTables. Perhaps we can treat these two models as distinct in the design.

chrisdembia:
Perhaps we can get rid of some big headaches by trying to pretend that C3D and trc, csv are all the same.

sherm1:
> There seem to be two models out there. HDF5, MAT, and C3D files are collections of data tables. Plain text files (mot, trc, csv) are most reasonably just a single data table. 

My inclination would be to use just one concept if it can cover all cases. If we say a file represents a collection of DataTables, then that already includes files that represent a singleton collection, without any further documentation or explanation required. No one would be confused about that. Mental space, I think, is a much more precious commodity than code space. And we could also throw in a few convenience methods that serve as abbreviations for the case where there is only one DataTable; IMO that occupies much less mental space than having two separate classifications for files, with different support classes.

aseth1:
>  That doesn't seem to map as well onto an ongoing reading process. Maybe we could start thinking of these as "converters" as Chris suggested rather than "adapters".

The converter model does not fit with the idea of being "plugged" into a streaming source, e.g. from a device. It may be easier to think of having both things, with FileConverters for now and DataAdapters for streaming sources. 

> Here's a new question I think: if a file now maps to a collection of DataTables, where does the file metadata go? It seems strange to have it associated with any one of the resulting DataTables.

I think the meta data for each Table from the C3D is completely different. A ForceTable is very different from a MarkerTable or an EMGTable in terms of units, range and associated parameters. For example, the ForceTable should include the frames (or corner coordinates) of the force-plate used to get the reading. Remember, it isn't FileMetaData it is TableMetaData and ColumnMetadata. 

As it stands now, the C3DFileAdapter is more a one shot converter of all its contents. I would rather more atomic behavior of being able to only read the table of interest.

We could go back to the single DataTable to read() relationship, with the a single DataAdapter populating a table. To read multiple table you have to bound each table to the adapter one at a time (or use multiple instances of the adapters to read in parallel). 

This would requires that  the adapter can tell from its inputs which table you want out of the C3D. My intention was to use the concrete DataTable type and/or its MetaData for that purpose. For example, if I have a DataTable<Vec3> with a meta data key of 'content' = ForcePlate,  then it only reads a DataTable containing force plate data from the C3D, otherwise I get an exception. Alternatively, we have ForceTable table, in which case you have to pass in a ForceTable to get it out. 

For a generic data reader like CSVFileReader you only can only get a SimpleDataTable but you can construct a ForceTable, MarkerTable, etc.. from SimpleDataTables. That builds on the capability we added to DataTables to validate their contents.

chrisdembia:
> The converter model does not fit with the idea of being "plugged" into a streaming source, e.g. from a device. It may be easier to think of having both things, with FileConverters for now and DataAdapters for streaming sources.

I thought the *DataSource* is what provides the idea of being plugged. I think I suggested this earlier, but we can have a different type of `DataSource` that supports streaming, and it can be different from all this file-reading stuff.

klshrinidhi:
I would suggest the following:
We get rid of the notion of *streams*. We only talk about *files*. So for OpenSim 4.0, we will be supporting reading/writing of different *file* formats (TRC, MOT, CSV, C3D and more if you like) and *not generic streams*. The class hierarchy can be same as what it is right now with DataAdapter removed and all its code moved to FileAdapter. The new hierarchy will be:
```
FileAdapter <-- TRCFileAdapter
            <-- MOTFileAdapter
            <-- CSVFileAdapter
            <-- C3DFileAdapter
            ..... more ........
```
My opinion is that the notion of generic streams, though a good idea for genericity, is not going to help much for our goals for OpenSim 4.0. The goal is to basically support different file formats and be able to read a C3D file into OpenSim. This goal can be achieved with the current implementation with the above mentioned modification.

sherm1:
> We get rid of the notion of streams. We only talk about files. 

That does sound appealing. I would combine that with some name changes. Also, there is still the issue of read and write access to sort out. Maybe
```
FileReader <- TRCFileReader, MotFileReader, ...
FileWriter <- TRCFileWriter, MotFileWriter, ...
```
Then we could have just a constructor like `FileReader("myFile.trc")` that would read in the whole file and turn it into DataTables, then a method for obtaining some or all of the resulting tables.

sherm1:
> we can have a different type of DataSource that supports streaming, and it can be different from all this file-reading stuff

This would also let us deal separately with real-time streaming which is *very* different than reading or writing a file.

aseth1:
@sherm1 Shrinidhi, Chris and I had a good discussion. We concluded to:

1. remove persistent access to the source. Although potentially necessary to support real-time streaming, streaming requires far more design and implementation work and won't be in 4.0
2. only support FileAdapters to read in DataTables and  write out files from DataTables.
3. do not close off the possibility for StreamAdapters and these could be split off from existing FileAdapters when necessary (post 4.0)
4.  keep the multiple table output from adapter read since they are valid, but return a map of DataTables keyed by their name (when named e.g C3D) if un-named then there can only be one 'table' (in the generic case) otherwise these are generally 'markers', 'forces', 'emg', ...
5. move quickly to integrate DataTables to replace existing Storages so we can use C3Ds natively in existing Tools., which would be a big win for 4.0 
6. start prototyping Components (i.e. DataSource, DataSink ) that allow us to complete the analysis wiring diagram and generate examples. 

We may prefer to have the Adapter with read() and write() instead of separate Reader and Writer because they share most of their implementation and we don't have to introduce yet another class. The file is only accessed with the appropriate read/write permission upon read()/write() there isn't an internal state since access is not persistent.

The usage:
```cpp
TRCFileAdapter trcAdapter;
table  = trc.read("myFile.trc");
```
seems comparable to:
```cpp
table = TRCFileReader("myFile.trc").getTable();
 ```

**Comment by sherm1** [comment]
Sounds great. Would there still be a generic FileAdapter class that would instantiate the appropriate format-specific adapter based on the extension or whatever? Then the read/write version would presumably allow:
```cpp
FileAdapter fileAdapter;
table  = fileAdapter.read("myFile.trc");
table  = fileAdapter.read("myFile.mot");
table  = fileAdapter.read("myFile.c3d");
fileAdapter.write("yourFile.trc", table);
// etc
```
Is that how you want it? It has the odd feature that the Adapter classes aren't really objects since they have no internal state, just namespaces for wrapping a few static methods. An alternative would be `FileAdapter("myFile.trc")` where the adapter would contain the in-memory representation of the file from which a user could obtain one or more of the DataTables.

aseth1:
> Would there still be a generic FileAdapter class that would instantiate the appropriate format-specific adapter based on the extension or whatever? Then the read/write version would presumably allow:

Yes, that is the intention.

> It has the odd feature that the Adapter classes aren't really objects since they have no internal state, just namespaces for wrapping a few static methods. An alternative would be FileAdapter("myFile.trc") where the adapter would contain the in-memory representation of the file from which a user could obtain one or more of the DataTables.

That's a good point and we were purposely avoiding having to manage the file access mode as an internal state. In that sense the FileAdapters are more like one shot converters. But in your case the FIleAdapter would have to be a Reader or a Writer otherwise how could it know to read out tables or be ready to append or overwrite upon construction? Otherwise, it would need to know its allowable access modes and then change modes in order to read()/write() from/to the file later. Being stateless is not a requirement, but it seems simpler. As for only needing static methods I don't think that would be sufficient or desirable -  we want the read()/write() to delegate to virtual methods of concrete FileAdapters. We may need to maintain state or data internally down the road. 

Do you think we are now oversimplifying?

I am certain we will have to revisit DataAdapter to account for more scenarios. We can address FileAdapters again when we have to generalize/expand DataAdapters to support more input types (e.g. streams), but that isn't an immediate need and we will learn a lot in the meantime by using the concrete FileAdapters.

sherm1:
Hmmm ... it looks to me like `FileAdapter` is now a singleton class; its methods can be static. Its job is to figure out which formatter to use. It could also be the place we keep the registry of formatters. The base class for the formatters would have virtual methods and a registry, but there would be no need to clone the formatters either; they could also be singletons. I'll come by to see if I'm confused.

sherm1:
> Do you think we are now oversimplifying?

No, but now I think you might be *under*-simplifying a little. There is very little abstraction left (just the particular format) so the class structure should be simplified to reflect that. FileAdapter can just be all static and not part of any hierarchy. Its read() and write() methods would make use of a FormattedFileAdapter (or whatever) that is an abstract base class for the concrete adapters that are stored in the registry. There can be just one object of each concrete formatter class, living in the registry and it could be used from there without copying.

aseth1:
> FileAdapter is now a singleton class; its methods can be static. Its job is to figure out which formatter to use. It could also be the place we keep the registry of formatters. The base class for the formatters would have virtual methods and a registry, but there would be no need to clone the formatters either; they could also be singleton.

That sounds right. FileAdapter is a factory for concrete adapters and you are also right there is no need for the concrete adapters to even derive from FileAdapter. I just didn't want to call `FileAdapter`, `FileAdapterFactory` (which would be accurate) and have that name exposed to users. I like the interface as you described with user talking to the singleton FileAdapter (factory) and it delegates to the appropriate concrete implementations via a keyed registry of adapters.

We discussed leaving DataAdapter as the base class only because the next step (post 4.0) and perhaps sooner if @msdemers sticks around to work on it, will be to enable streaming and then then it may warrant beefing up the base DataAdapter (or not). If not we could remove DataAdapter altogether at that time. @klshrinidhi @chrisdembia were for getting rid of it too, so maybe it is just easier to eliminate it now. 

What other unnecessary class/abstraction is left?

sherm1:
FileAdapter wouldn't even be a factory now since it doesn't create anything. It just turns arbitrary files into DataTables and vice versa. Internally it finds the right formatter and invokes it to do the actual work. Users wouldn't even have to know about the formatters; only developers who are adding new ones would see them. Very clean, I think!

klshrinidhi:
Closing this. 
Refer to adapters PR #709.

Issue #702, opened 2015-11-13
Handling OpenSim dependencies for building from sources.


klshrinidhi:
As OpenSim grows and uses more libraries (like BTK is being used now; catch/gtest, glog etc hopefully coming soon), handling dependencies is something we have to address. @chrisdembia and I have talked about this for a while.

For the purpose of this issue, OpenSim community is classified into two groups:

1. People who **do not edit** OpenSim sources. This set includes people who use OpenSim from Python/Matlab/C++ to write their own code but they don't intend to edit OpenSim code itself. This also includes people who intend to write plugins for OpenSim. [Estimate = 95% of OpenSim community.]
For this group, OpenSim will release binaries for Windows, Mac OS X and Linux.

2. People who **do edit** OpenSim sources. This set is mainly *OpenSim developers* -- people who intend to edit (and maybe contribute to) OpenSim-core. This set also includes some unfortunate people with platforms we don't have binaries for. We can encourage these people to either:
 * host binaries if/when they have successful builds and submit just the *links* to us for hosting on OpenSim website.
 * submit the binaries to us for hosting on OpenSim website.
  
 These people will be compiling OpenSim sources. This issue is mostly about this group.
For this group, we could use CMake *super-build* to resolve dependencies. The advantages:
 * We control where to download the dependency from.
 * We control which version of the dependency to use.
 * We control what build parameters to use when building the dependency.
 * Potentially reduce number of build related questions and therefore our time in answering them.

 Here is how it could work:
 * Building OpenSim will be two-step procedure, which can be wrapped into one step for most users in this group.
 * Step one is -- *DownloadAndBuildDependencies*. This is handled by CMake *super-build* using the Cmake module named *ExtenalProject*. For advanced users, this step can be configured to skip/include certain dependencies.
 * Step two is -- *BuildOpenSim*. At this point dependencies are already in place and it is just about passing correct paths to the dependencies for OpenSim build.

 In step one, dependencies are downloaded. We could host binaries for dependencies for most (if not all) platforms. The download step will check if binaries are available and proceed to compiling sources only if no pre-built binaries are available. For making the dependency sources available, the alternatives are:
 * Copy the sources from the dependency repo into Opensim. 
   *  :+1: No risk of losing access to sources. Links can die.
    *  :-1: We have to re-copy if we want to use newer versions of the dependency.
    * :-1: Dependencies get downloaded with OpenSim whether they are needed or not.
[...]

PR #715, opened 2015-12-01
Superbuild for dependencies.

klshrinidhi:
This PR implements some of the things from #702.

With this change the current build procedure will not break. The change allows a new way to build OpenSim along with its dependencies.

The new build procedure will be as follows. These steps can be wrapped into one shell/powershell script to work on Linux/OSX/Windows.

----------------------------------------------------------
##### Step 1 -- Dependencies
```shell
# Create build directory next to opensim-core.
mkdir opensim_dependencies_build
cd opensim_dependencies_build
cmake ..\opensim-core/dependencies `
        -G"Visual Studio 14 2015 Win64" `
        -DCMAKE_INSTALL_PREFIX=..\opensim_dependencies_install
cmake --build . --config Release
```
##### Step 2 -- OpenSim
```shell
# create build directory next to opensim-core.
mkdir opensim_build
cd opensim_build
cmake ..\opensim-core `
        -DOPENSIM_DEPENDENCIES_DIR=..\opensim_dependencies_install `
        -DCMAKE_INSTALL_PREFIX=<install-location>
cmake --build . --target INSTALL --config Release
```
##### Step 1 -- Dependencies
```shell
mkdir ~/opensim_dependencies_build
cd ~/opensim_dependencies_build
cmake ~/opensim-core/dependencies \
        -DCMAKE_INSTALL_PREFIX=~/opensim_dependencies_install \
        -DCMAKE_BUILD_TYPE=RelWithDebInfo
# OR selectively superbuild dependencies. 
ccmake ~/opensim-core/dependencies
make
```
##### Step 2 -- OpenSim
```shell
mkdir ~/opensim_build
cd ~/opensim_build
cmake ~opensim-core \
        -DOPENSIM_DEPENDENCIES_DIR=~/opensim_dependencies_install \
        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
        -DCMAKE_INSTALL_PREFIX=<install-location>
make
make install
```
----------------------------------------------------------
Some questions that came up in meeting:
**Is it possible to selectively super-build dependencies ?**
Yes. CMake GUI offers one check-box per dependency. 
Currently BTK is *required*. Un-checking BTK will not super-build it but it is still *required*. I can make it *disable-able* if you like.
**When re-building, does the super-build happen again ?**
No. The dependencies are not downloaded, built and installed repeatedly.
**Is it possible to supply build parameters to dependencies ?**
Yes. By passing `CMAKE_ARGS` in call to `AddDependency` in `CMakeLists.txt`, it is possible to set build parameters for any dependency. CMake GUI does not allow setting any parameters for any dependency selectively. The parameters user specify in CMake GUI are applied to all dependencies.
**Is it possible to get all the sources into one compressed file ?**
Yes. During super-build, dependency sources are downloaded into the directory `dependencies` under `opensim-core`. Each dependency will get its own directory.
**Can super-build be skipped ?**
Yes. If binaries are available for dependencies, one may directly go to step 2 above and specify the path to individual dependencies.
**Does super-build take care of literally *all* dependencies ?**
No. For example `simbody` depends on BLAS, LAPACK, GLUT etc. These are not currently in super-build. On Linux & Mac, these are available as binaries through package managers. Super-build currently only handles dependencies whose binaries are not available, at-least in the versions OpenSim needs them.
**Does it handle multi-configuration generators ?**
Yes. Multi-configuration generators will generate rules for all chosen configurations at once. During build/install, user chooses the build-type to build/install.
**Is it possible to build multiple configurations without conflict ?**
Yes.
On Linux/Mac, use different different build/install directories for different build types. For example, the following directory structure could work:
```
opensim_build
        Debug
        Release
        RelWithDebInfo

opensim_install
        Debug
        Release
        RelWithDebInfo
```
On Windows, msbuild (used by MSVC) builds different build-types in different directories under the same build directory. For installation, choose different directories for different build types. The above directory structure could work here too.

TODO: 
* Add ability to download pre-built binaries for dependencies if available.


klshrinidhi:
Ready for review.

dependencies/CMakeLists.txt
```diff
@@ -0,0 +1,119 @@
...
+endfunction()
+
+
+SetDefaultCMakeInstallPrefix()
+SetDefaultCMakeBuildType()
+
+####################### Add dependencies below.
+
+AddDependency(NAME BTK
+              URL  https://github.com/klshrinidhi/BTKCore.git
```
tkuchida: Will this eventually reside on opensim-org?

klshrinidhi: Yes. This is temporary until BTK author merges my pull request fixing a problem.

aymanhab:
@klshrinidhi Is there a way to tell what the install layout looks like (release, debug) without actually building the branch since that's what really matters at the end? Thanks.

klshrinidhi:
@aymanhab 
The dependencies download/build/install step will:
* Download the sources to the directory *dependencies* under opensim-core. Each dependency gets its own directory. 
* Build the dependency sources in build directory for dependencies (named *opensim_dependencies_build* in my example) given by the user. Each dependency gets a directory here too.
* Install the dependencies in the install directory (named *opensim_dependencies_install* in my example) given by the user. Each dependency gets a directory here too.

The build and install of OpenSim is exactly same as it was before. If dependencies are built using super-build, they will be installed in their own directory, which user could choose to be within OpenSim installation directory. 
Overall, there will be two install directories, one for all the dependencies, the other for OpenSim. User can mix and match different installations.
Does this answer your question ?

aymanhab:
Thanks @klshrinidhi. Yes, it answers but the story doesn't end up there since there are users of the install (e.g. Packaging or user CMake files) that now rely on env var. OPENSIM_HOME that's the top level install dir of OpenSim and all its dependencies, including Simbody, will these need to specify/find dependencies separately? We should at least discuss and hopefully sort out how things will look from end user or API-user perspective not just to the developers in-house. The system we have in place for integrating Simbody seems relatively smooth, do you see that changed by the added dependencies?

chrisdembia:
This worked without hiccups on Ubuntu 14.04. Also, the `dependencies/CMakeLists.txt` is beautiful and surprisingly short. I still want to test on Windows and OSX (with both make and Xcode).

1. One usability downside I see is the need to create a separate build directory for the dependencies.

2. To be consistent with [the default name for the OpenSim install directory](https://github.com/opensim-org/opensim-core/blob/master/CMakeLists.txt#L57), what about naming the default dependencies install dir as `${CMAKE_SOURCE_DIR}-install-dependencies`?

1. Can you edit the README.md to describe the new install instructions? Also, document the `OPENSIM_DEPENDENCIES_DIR` in the way that other cmake variables are documented (for each platform).

3. Can you add a comment to the top of dependencies/CMakeLists.txt clarifying that this file is NOT part of the OpenSim cmake project?

3. As for packaging up the sources: I like that you've structured everything so that we can just zip up the git repo and have a source distribution. The next step would be to still leverage the `dependencies/CMakeLists.txt` to automatically build that source code *without* downloading again. Is it possible to set up the `dependencies/CMakeLists.txt` to recognize if the source is already present, and then skip the download?

2. Don't gitignore `dependencies/CMakeLists.txt` (with maybe `!dependencies/CMakeLists.txt`).

3. When running CMake for the first time to build dependencies, CMake already creates the install directory (opensim_dependencies_install, as well as empty BTK and simbody dirs), even before I start building. Can this be avoided? Ideally these directories would only be created after I run `cmake --build .`.

appveyor.yml
```diff
@@ -54,15 +60,20 @@ install:
   - nuget install simbody -Version 0.0.0 -ExcludeVersion -OutputDirectory C:\
 
 build_script:
+  ## Superbuild dependencies.cd
```
chrisdembia: "cd" is typo?

appveyor.yml
```diff
@@ -54,15 +60,20 @@ install:
...
   # for some reason.
-  - mkdir build
-  - cd build
+  - cd %OPENSIM_BUILD_DIR%
   # Configure.
-  - cmake .. -G"Visual Studio 14 2015 Win64" -DSIMBODY_HOME=C:\simbody -DCMAKE_INSTALL_PREFIX=%OPENSIM_HOME% -DBUILD_JAVA_WRAPPING=ON -DBUILD_PYTHON_WRAPPING=ON # TODO -DBUILD_SIMM_TRANSLATOR=ON
+  - cmake %OPENSIM_SOURCE_DIR% -G"Visual Studio 14 2015 Win64" -DOPENSIM_DEPENDENCIES_DIR=%OPENSIM_DEPENDENCIES_INSTALL_DIR% -DCMAKE_INSTALL_PREFIX=%OPENSIM_INSTALL_DIR% -DBUILD_JAVA_WRAPPING=ON -DBUILD_PYTHON_WRAPPING=ON # TODO -DBUILD_SIMM_TRANSLATOR=ON
   # Build.
-  - cmake --build . --target ALL_BUILD --config Release -- /maxcpucount:4 /verbosity:quiet
+  - cmake --build . --config Release -- /maxcpucount:4 /verbosity:quiet
```
chrisdembia: Cool; I didn't realize the `target` was not necessary.

CMakeLists.txt
```diff
@@ -423,6 +423,20 @@ if(APPLE AND NOT (${CMAKE_VERSION} VERSION_LESS 2.8.12))
     endif()
 endif()
 
+set(OPENSIM_DEPENDENCIES_DIR 
+    "" 
+    CACHE 
+    PATH
+    "Directory containing installed binaries of OpenSim dependencies.")
```
chrisdembia: Can you clarify that this related to the superbuild and should not be set if you installed the dependencies yourself? People might get confused if they built their dependencies themselves; they might think they ALSO have to set this.

CMakeLists.txt
```diff
@@ -423,6 +423,20 @@ if(APPLE AND NOT (${CMAKE_VERSION} VERSION_LESS 2.8.12))
...
+
+find_package(BTK 
+    REQUIRED
+    HINTS ${OPENSIM_DEPENDENCIES_DIR}/BTK/share/btk-0.4dev)
+
+if(NOT SIMBODY_HOME AND OPENSIM_DEPENDENCIES_DIR)
+    set(SIMBODY_HOME ${OPENSIM_DEPENDENCIES_DIR}/simbody)
```
chrisdembia: Paths should be in quotes in case they contain spaces: `"${OPENSIM_DEPENDENCIES_DIR}/simbody"`. Same goes for the BTK HINTS.

.travis.yml
```diff
@@ -114,10 +116,16 @@ before_install:
   ## If we are building in debug, there are some tests to ignore.
   - if [ "$BTYPE" = "Debug" ]; then export TESTS_TO_EXCLUDE="$TESTS_TO_EXCLUDE|testInitState|testPrescribedForce|testInducedAccelerations|testAssemblySolver|testCMC|testOptimizationExample|testComponents|testForward"; fi
 
+  # Run superbuild to download, configure, build and install dependencies.
+  - mkdir $OPENSIM_DEPENDENCIES_BUILD_DIR
+  - cd $OPENSIM_DEPENDENCIES_BUILD_DIR
+  - cmake $TRAVIS_BUILD_DIR/dependencies -DCMAKE_INSTALL_PREFIX=$OPENSIM_DEPENDENCIES_INSTALL_DIR -DCMAKE_BUILD_TYPE=$BTYPE
+  - make -j$NPROC
```
chrisdembia: You don't need to use NPROC here; you can use 8. I use NPROC just when building wrapping since building the wrapping uses up a lot of memory.

chrisdembia:
On travis and appveyor, can you avoid using superbuild for simbody? We already download binaries for simbody (on both travis and appveyor) so superbuild is not necessary for it. However, we still should use superbuild for btk on travis/appveyor (and can even store btk binaries for use with continuous integration).

dependencies/CMakeLists.txt
```diff
@@ -0,0 +1,119 @@
+cmake_minimum_required(VERSION 2.8)
+
```
chrisdembia: Can you give this project a name? Perhaps `project(OpenSimDependencies)`? Otherwise, it's just named "Project".

chrisdembia:
> On Windows, msbuild (used by MSVC) builds different build-types in different directories under the same build directory. For installation, choose different directories for different build types.

I'm not clear how this works for installation. Choosing a different variable value for different configurations doesn't work. We can't expect people to change cmake variable values when they use a different configuration. However, people should not *need* to use different install directories for different configurations. All configurations should be installed into the same install directories, and the `SimbodyTargets-debug.cmake`, etc files should take care of finding the appropriate files.

chrisdembia:
On OSX I had issues with Xcode (multiconfig generator) when building the dependencies in relwithdebinfo and debug, and then build opensim in debug: cmake tried adding `opensim_dependencies_install/simbody/lib/Debug` to the link line, which  didn't exist and is a bug. It's not too surprising that this occurred though, since CMake appends `Debug` to certain paths when using a multiconfig generator. Anywho, this is a very common use case for Visual Studio users, so I think we should iron it out.

klshrinidhi:
@chrisdembia 
>One usability downside I see is the need to create a separate build directory for the dependencies.

I actually don't see this as a downside. Yes. There is need for another build directory but this allows users to link OpenSim against different build types of dependencies easily. My intention was to separate the dependencies build and install from OpenSim.

>To be consistent with [the default name for the OpenSim install directory](https://github.com/opensim-org/opensim-core/blob/master/CMakeLists.txt#L57), what about naming the default dependencies install dir as `${CMAKE_SOURCE_DIR}-install-dependencies`?

I think `${CMAKE_SOURCE_DIR}-install-dependencies` will translate to `opensim-core/dependencies-install-dependencies`. Correct ? I would like the build directories to end with `build` and install directories to end with `install`.

>Can you edit the README.md to describe the new install instructions? Also, document the `OPENSIM_DEPENDENCIES_DIR` in the way that other cmake variables are documented (for each platform).

Yes. Will do.

>Can you add a comment to the top of dependencies/CMakeLists.txt clarifying that this file is NOT part of the OpenSim cmake project?

Yes. Will do.

>As for packaging up the sources: I like that you've structured everything so that we can just zip up the git repo and have a source distribution. The next step would be to still leverage the `dependencies/CMakeLists.txt` to automatically build that source code *without* downloading again. Is it possible to set up the `dependencies/CMakeLists.txt` to recognize if the source is already present, and then skip the download?

I don't think the download happens repeatedly. If the particular commit mentioned in `dependencies/CMakeLists.txt` exists in the directory `dependencies/<dependency-name>`, it will not download the sources again. It will just check it and move to build step.

>Don't gitignore `dependencies/CMakeLists.txt` (with maybe `!dependencies/CMakeLists.txt`).

Oh. I was not sure about this. I thought because I have added CMakeLists.txt to version control, git will not ignore it and track its changes. I will double-check.

>When running CMake for the first time to build dependencies, CMake already creates the install directory (opensim_dependencies_install, as well as empty BTK and simbody dirs), even before I start building. Can this be avoided? Ideally these directories would only be created after I run `cmake --build .`.

I found this annoying too. But CMake tries to make sure it can create (with permission) the directories in the configure step. If you change the variable CMAKE_INSTALL_PREFIX, I make sure to delete this unused directory during build. I didn't see any other way to avoid it. The default setting was `/usr/local` which gives permission denied error during configure.

chrisdembia:
> I think ${CMAKE_SOURCE_DIR}-install-dependencies will translate to opensim-core/dependencies-install-dependencies. Correct ? I would like the build directories to end with build and install directories to end with install.

Sorry, I meant that `CMAKE_SOURCE_DIR` is the `opensim-core` directory, so you'd end up with `opensim-core-install-dependencies` next to the `opensim-core` directory.  I'm fine with `opensim-core-dependencies-install` as well.

chrisdembia:
> I found this annoying too. But CMake tries to make sure it can create (with permission) the directories in the configure step.

Ah okay. Sounds fine then.

klshrinidhi:
@chrisdembia 
>Sorry, I meant that `CMAKE_SOURCE_DIR` is the `opensim-core` directory, so you'd end up with `opensim-core-install-dependencies` next to the `opensim-core` directory.  I'm fine with `opensim-core-dependencies-install` as well.

Okay. Will do.

aymanhab:
@klshrinidhi The main point I'm making regarding installation of dependencies is that all dlls/so/dylibs/executables should end up in the same directory by default (though users can override) so that they can be located off one CMake variable, Path env. variable etc. (Java, Matlab, Python and GUI all make this assumption as well as existing user CMake files, I don't see a reason to change that behavior. I'd prefer not to have OpenSim installer add multiple paths to user env. or require more, potentially error prone steps from the end-user side.

klshrinidhi:
@aymanhab 
Currently I have dependencies installed in a separate directory. This means:
* On Windows, PATH has to be edited to included those directories. I am looking into making this better by using alternatives that equivalent to RPATH in Linux. Will get back on this.
* On Linux/Mac, CMake edits the RPATH of executables and libraries so they can find their own dependencies. So we should be good here.

The problem with having all executables and libraries from all dependencies in the same directory is that libraries/executables from different dependencies can conflict with each other. We may have to do extra work to avoid these situations. Looking at other projects, we can see that they are organizing their binaries in directories and have some way to locate them at run-time. I will investigate and get back on this.

klshrinidhi:
I have been looking at manifests on Windows in the hope to get behavior equivalent to RPATH on Linux/Mac. So far it has been frustrating. I read a lot of msdn documentation. VS2015 is currently embedding manifests into exe/dll by default but doesn't seem to insert full path to dependencies (found at link time) into those binaries. I couldn't find a way to do this that is as simple as enabling/disabling a flag to the linker.
I am now working on writing ugly platform dependent code just for Windows so CMake copies all the dependency dlls into OpenSim.

chrisdembia:
> I am now working on writing ugly platform dependent code just for Windows so CMake copies all the dependency dlls into OpenSim.

To accommodate the current copying we do with simbody libraries (which I don't like), I tried writing this cmake macro: https://github.com/opensim-org/opensim-core/compare/cmake-simbody-libs-debug to make our own Config file that adapts to where we copied the libraries.

It's really ugly and bug-prone.

chrisdembia:
https://github.com/opensim-org/opensim-core/compare/cmake-simbody-libs-debug#diff-f808c5dde690cfe904903e625de2aa48R9

chrisdembia:
I still *really* *really* think that this PR is separate from hooking up the dependencies at run time. This PR is good to go no matter what we do to solve runtime issues (though there might be other issues that I forgot). People building from source need to build dependencies, regardless of where they exist at runtime.

klshrinidhi:
@chrisdembia 
In its current state, on Windows, this PR allows download, configure and build of dependencies, and then build OpenSim to link against the dependency libraries. The problem starts when we run tests through ctest. Runtime loader cannot find the dependencies to complete the tests and all tests fail. Same problem if we run any executable in OpenSim. This problem of finding dependencies has to be figured for this PR to be merged.

chrisdembia:
That problem exists anyway. Say Ajay wants to build the latest source with BTK. He needs to get BTK and find a way to solve the runtime issue. Right now, Ajay would likely solve the issue by editing his PATH variable. That's what people could do with this PR. Solving the runtime issue is then an enhancement over setting PATH.

It's your choice if you want to incorporate the runtime issue in this PR. I'm just trying to help by getting one of your PR's merged quicker.

klshrinidhi:
Thanks. I think I will get the copy thing done and make it complete.
The data adapter PR will need this (or modifying PATH which we don't want) in order to pass the tests.

klshrinidhi:
I added code to copy DLLs over from a dependency into OpenSim for Windows only.

**Ready for review.**

Lets try and merge this soon.

chrisdembia:
I'll review soon. 

When merging this I'd like to open a new issue to improve our runtime issue. I'm happy we have something that works now but I want us to keep examining alternatives (thanks for looking into manifests also). It might be worthwhile to study how conda has decided to organize libraries (on all platforms): http://www.slideshare.net/teoliphant/effectively-using-open-source-with-conda.

chrisdembia:
On both travis and appveyor, I vote for not building simbody using superbuild. Both travis and appveyor are set up to use prebuilt simbody binaries to speed up the build process.

klshrinidhi:
I created an organization at https://bintray.com/nmbl to host binaries for our dependencies. Right now it only contains binaries for BTK on Linux. After we upload binaries for Mac & Windows, we can disable super-build on Travis and Appveyor.

chrisdembia:
Can we still use `-DSUPERBUILD_SIMBODY=off` for now?

chrisdembia:
@klshrinidhi as for the copying of libraries into OpenSim's lib dir: I think I am OK with this so long as we *also* ship the dependencies intact (that means we ship two copies of the libraries for our dependencies). That way, those building a project off of OpenSim can use the *dependency* directories to get the dependency libraries, and we do *not* need to modify the Config files for the dependencies. Also, to help users with copying the dependency libraries, we could provide a macro for their use that will do this for them. We could also use the same macro in this PR.

I still want to think about other solutions though, such as statically linking (but not as a part of this PR).

.travis.yml
```diff
@@ -114,10 +116,16 @@ before_install:
   ## If we are building in debug, there are some tests to ignore.
   - if [ "$BTYPE" = "Debug" ]; then export TESTS_TO_EXCLUDE="$TESTS_TO_EXCLUDE|testInitState|testPrescribedForce|testInducedAccelerations|testAssemblySolver|testCMC|testOptimizationExample|testComponents|testForward"; fi
 
+  # Run superbuild to download, configure, build and install dependencies.
+  - mkdir $OPENSIM_DEPENDENCIES_BUILD_DIR
+  - cd $OPENSIM_DEPENDENCIES_BUILD_DIR
+  - cmake $TRAVIS_BUILD_DIR/dependencies -DCMAKE_INSTALL_PREFIX=$OPENSIM_DEPENDENCIES_INSTALL_DIR -DCMAKE_BUILD_TYPE=$BTYPE -DSUPERBUILD_SIMBODY=OFF
```
chrisdembia: Thank you!

appveyor.yml
chrisdembia: bump

appveyor.yml
```diff
@@ -54,15 +60,20 @@ install:
   - nuget install simbody -Version 0.0.0 -ExcludeVersion -OutputDirectory C:\
 
 build_script:
+  ## Superbuild dependencies.cd 
+  - mkdir %OPENSIM_DEPENDENCIES_BUILD_DIR%
+  - cd %OPENSIM_DEPENDENCIES_BUILD_DIR%
+  - cmake %OPENSIM_DEPENDENCIES_SOURCE_DIR% -G"Visual Studio 14 2015 Win64" -DCMAKE_INSTALL_PREFIX=%OPENSIM_DEPENDENCIES_INSTALL_DIR%
```
chrisdembia: `-DSUPERBUILD_SIMBODY=OFF` could go here too; simbody binaries are set up for appveyor as well (the `nuget install simbody` line). I'm fine with migrating to a new binary artifact mechanism (which I think you are planning), but in the interim, can we use the binary system that exists now?

cmake/OpenSimMacros.cmake
```diff
@@ -253,3 +253,29 @@ function(OpenSimAddApplication APPNAME)
...
+            set(DEST_DIR ${CMAKE_BINARY_DIR}/${CMAKE_CFG_INTDIR})
+            add_custom_command(OUTPUT ${DLL_NAME}
+                               COMMAND cmake -E copy ${DLL} ${DEST_DIR}
+                               COMMENT "Copying ${DLL_NAME} to ${DEST_DIR}")
+            list(APPEND DLL_NAMES ${DLL_NAME})
+        endforeach()
+        add_custom_target("Copy_${DEP_NAME}_DLLs" ALL DEPENDS ${DLL_NAMES})
+        install(FILES ${DLLS} DESTINATION ${CMAKE_INSTALL_BINDIR})
+    endif()
+endfunction()
```
chrisdembia: So emacs lets you save a file that doesn't have a newline at the end? vim > emacs!

cmake/OpenSimMacros.cmake
```diff
@@ -253,3 +253,29 @@ function(OpenSimAddApplication APPNAME)
     endif()
 
 endfunction()
+
+
+# Function to copy DLL files from dependency install directory into OpenSim 
+# build and install directories. This is a Windows specific function enabled 
+# only for Windows platform. Intention is to allow runtime loader to find all 
+# the required DLLs without need for editing PATH variable.
```
chrisdembia: Awesome function. One issue we've had in the past with copying dll's is we used to GLOB all the libraries in a directory and this ended up copying ALL libraries on UNIX machines (if Simbody was installed into `/usr/local` or something). Since this function only does anything on windows, this is unlikely to be an issue. A slightly more robust solution would be to loop through all the libraries that a package exports. I think this would be possible, but is unnecessary for now. I'm just mentioning this to consider if we run into problems in the future.

dependencies/CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
...
+
+AddDependency(NAME simbody
+              URL  https://github.com/simbody/simbody.git
+              TAG  ed67885db8983405e0724e63e62bf8e9316b2493)
+
+
+#######################
+
+RemoveDefaultInstallDirIfEmpty(${DEFAULT_CMAKE_INSTALL_PREFIX})
```
chrisdembia: In general, variables that contain paths should be enclosed in double quotes. Otherwise, spaces in a path will make CMake think that a single path is actually multiple arguments.

chrisdembia:
Some questions (I don't know the answers):

1. For simbody, should we disable the building of examples to speed up the build?
2. Should the install directory be named `opensim-core_dependencies_install` to be consistent with the name of the repo, and to be unambiguous with the GUI repo?

This PR worked great on OSX with Xcode using both Debug and Release builds. I'm now going to make some edits to your README changes.

chrisdembia:
I'm done reviewing (see my PR about the README).

chrisdembia:
Does anyone else plan to review this as well? Looks good to me, now.

PR #717, opened 2015-12-11
Pseudo code ayman01

aymanhab:
Hypothetical code to be written by users to report MusclesLineOfAction, in prep for pseudo code session.Internals and used classes are either need to be written from scratch or replaced by some existing classes.


chrisdembia:
Ayman: people have requested this for 3.2.
Ajay: we would have to give more information..from c3d, need "force", etc. trajectories to be specified.

Ajay: how does a user digest this information?

Jen: you want the path of the muscle, or the force direction?

Ajay: the complete path changes, b/c the number of points changes.

Sherm: what do people mean by line of action?
Ajay: what's on the bone side. Just two.
Ayman: could have a few fixed points on a bone.
Ayman: should it report all the points?

James: Luca uses it for FE analysis.
Sherm: normal to line of action?

PR #720, opened 2015-12-14
Added pseudo code for performing a kinematics analysis (without using…

aseth1:
An example of kinematics reporting without use of analyses for this week's pseudo code session.

### DIFF
--- a/OpenSim/Sandbox/pseudoKinematicsReporting.cpp
+++ b/OpenSim/Sandbox/pseudoKinematicsReporting.cpp
@@ -1,0 +1,51 @@
+
+#include <OpenSim/OpenSim.h>
+
+main() {
+    // A study is the top level component that contains the model and other 
+    // computational components. 
+    Study kinematicsStudy;
+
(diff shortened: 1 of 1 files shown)


OpenSim/Sandbox/pseudoKinematicsReporting.cpp
```diff
@@ -0,0 +1,51 @@
...
+        bodyKinematics.addOutput(body.getOutput("transfom"));
+    }
+    kinematicsStudy.addReporter(bodyKinematics);
+
+
+    // Print out a list of all the available outputs from Frames in the model
+    for (const auto& frame : model.getComponents<Frame>()) {
+        for (const auto& output : frame.getOutputs()) {
+            cout << "Frame " frame.getName();
+            cout << " output: " << output.getName() << endl;
```
chrisdembia: Perhaps all Components can have a method like `getPath()` that would concatenate the frame name and output name as a convenience. To disambiguate with filesystem path, the method could also be called `getComponentPath()`.

PR #723, opened 2015-12-14
Pseudocode for reporting user-defined calculations on outputs

tkuchida:
Hypothetical code for computing the metabolic power attributable to each lower extremity joint.

### DIFF
--- a/OpenSim/Sandbox/pseudoJointMetabolics.cpp
+++ b/OpenSim/Sandbox/pseudoJointMetabolics.cpp
@@ -1,0 +1,100 @@
+/* -------------------------------------------------------------------------- *
+ *                    OpenSim:  pseudoJointMetabolics.cpp                     *
+ * -------------------------------------------------------------------------- *
+ * The OpenSim API is a toolkit for musculoskeletal modeling and simulation.  *
+ * See http://opensim.stanford.edu and the NOTICE file for more information.  *
+ * OpenSim is developed at Stanford University and supported by the US        *
+ * National Institutes of Health (U54 GM072970, R24 HD065690) and by DARPA    *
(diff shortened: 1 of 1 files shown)


chrisdembia:
Discussion in meeting:

- Tom: channels are like oscilliscopes.
- Chris: I like the concept of "ReporterChannel"
- Ajay: individual channels, bc mixed final files, moment arms, powers, etc. We already have mixed data like that... I've been thinking about that as individual reporters for each one. Each one would have a data tables, and then you could concatenate the data tables with similar types.
- Tom: add any types to one reporter, then underneath it will output multiple tables for types.
- Jen: thinking about matlab. using structs, can combine different types of data. precedent for keeping things separated by type. 
- Tom: can't have doubles and vec3's in one table right now.
- Sherm: how is reporter different from data table?
- Ajay: Reporter is a component. It's a DataSink. Wires up outputs to columns in a data table.
- Chris: If we
- Jen: rationale for computations in reporters instead in modeling space? once all the other stuff is computed, it goes into the oscilliscope, then i just need to do what the oscilliscope requires..don't need Simbody.
- Chris: if your oscilliscope were integrating it might use Simbody's measures.
- Jen likes Channels.
- Tom: channels could be good for plotter.
- Ajay: operator with "a + b + c" with outputs.
- Sherm: computations should be in reporters. reporting norms of a bunch of vectors.
- Ajay: take output of models.
- Sherm: what owns both model and normalizer? Ajay: a Study. Doesn't exist yet.
- Chris: if we called model as "System", then we would only need one class. We define lines somewhat arbitrarily.
- Ayman: what's available to reporter? Simbody quantities as well?
- Sherm: nice to have one concept, we might need to write a bunch of small wrapper classes.

PR #724, opened 2015-12-14
Inverse kinematics solver doxygen update

aseth1:
Render the IK objective and constraint functions as Latex style equations.


chrisdembia:
The doxygen will be uploaded to http://myosin.sourceforge.net/724 when the travis-ci/pr test is done.

chrisdembia:
I'm happy to merge this b/c it's better than it was, but the equation could still be improved. For example, what's `Wm`?

sherm1:
Equations look beautiful (evaluating just form, not content).

aseth1:
> I'm happy to merge this b/c it's better than it was, but the equation could still be improved. For example, what's Wm?

Good point, I will address that.

Issue #149, opened 2014-08-06
Serialize trajectories of States in their exact and complete form


aseth1:
Includes discrete state variables but not cached variables. A States trajectory must be tied to a model.
It shouldn't be easily human readable (e.g. zipped) and not bound by model name but by other unique key. It would also be handy if it maintained the version of OpenSim used to generate the States as well. Party addresses issue raised in #104 for robust post-hoc analysis of a simulation result.

chrisdembia:
Just found this binary format MessagePack: https://github.com/msgpack/msgpack/blob/master/spec.md#formats-array

PR #730, opened 2015-12-28
StatesTrajectory

chrisdembia:
I am actively working on a StatesTrajectory class. It's far from done, but I wanted to create a PR in case anyone has early feedback.

EDIT: This PR is now ready for review. 

## Summary

1. I've added a class StatesTrajectory (though there is no way to serialize/deserialize it yet).
2. The class has a static method that can populate the trajectory from a states storage file.
3. I've wrapped the class for scripting.
4. In python, one can also iterate through a trajectory:
  ```py
  states = osim.StatesTrajectory.createFromStatesStorage(
                   model, "subject01_states.sto")
  for state in states:
      model.calcMassCenterPosition(state)
  ```

5. I've written C++ and python tests.
6. I also created a sandbox file that compiles and runs without issue, to serve as example code:
  ```cpp
     Model model("gait10dof18musc_subject01.osim");
     ...
     auto& initState = model.initSystem();
     StatesTrajectory states = simulate(model, initState, 0.05);
 
     // Use trajectory.
     // ---------------
     for (const auto& state : states) {
         std::cout << state.getTime() << "s: "
                   << model.getStateVariableValue(state,
                                                  "ankle_r/ankle_angle_r/value")
                   << std::endl;
     }
  ```

7. The doxygen for this class is available at http://myosin.sourceforge.net/730/classOpenSim_1_1StatesTrajectory.html.
8. I made some small changes to Storage to support what I was doing with the trajectory.

## Discussion points
1. It may seem that a StatesTrajectory class belongs in Simbody. However, I talked with @sherm1 about this for a while and we decided that a sequential container of states is not too useful for simbody users, and that putting the container in OpenSim lets us make it more specific/powerful for OpenSim users. The idea is that Simbody will be able to de/serialize individual states as elements in an XML file, and StatesTrajectory's de/serialization will leverage that.
2. When appending a new state to the trajectory, I check that the time of the new state is greater/equal to that of the last state. However, it's easy to alter the trajectory so that the times are not nondecreasing (e.g., `states[5].setTime(5)`). So we have 3 options: (1) drop any notion that the times are nondecreasing, (2) leave as-is with the weak enforcement, (3) find a way to truly enforce an invariant that the times are nondecreasing.
3. What about creating a StatesTrajectory from kinematics storages (or kinematics and speed storages)? I don't think we should do this; instead, we should have a formal concept of "trajectory builders", like IK, muscle equilibrium, etc., and one could be creating a trajectory from just kinematics info (e.g., IK output).
4. Should cache variables be saved in the serialization? Some cache variables would be silly to recompute (CMC controls) and thus make sense to save). This question is a little bit beyond the scope of this PR.
5. Should `createFromStatesStorage()` have an option to `assemble()` each state?
6. I created a method `consistent()` that checks if all the states have the same number of Q's, U's, etc. The guts of this method really belong in Simbody. Do we all agree? Should the `consistent()` method even exist?
7. File extension for trajectory files? In the past I've like 'osimstates', but I ended up going with Ajay's 'ost'. We should discuss, though, since there might be even more opensim-related file extensions in the future.


chrisdembia:
This PR is now ready for review. Hopefully, at the next dev meeting, I can present this PR and get everyone's feedback.

sherm1:
Sweet! Just looked at the high-level stuff so far. I hope to have some States for you to (de)serialize soon.

Minor: should start bool methods with a verb, like `isConsistentWith()` or `hasWhatever()`.

chrisdembia:
> I hope to have some States for you to (de)serialize soon.

Cool! Looking forward to it.

BTW the doxygen comments in StatesTrajectory.h reflect what I think the class should become (once we can properly de/serialize, etc.), not what it is as of this PR.

chrisdembia:
### Feedback from dev meeting

Ajay, Matt, others: what about a method to access the nearest state to a given time? Tools need to essentially crop the trajectory.

Ajay: Tighter coupling to a model could make `createFromStatesStorage()` more efficient, since we could make more assumptions. However, people really shouldn't be using `createFromStatesStorage()` too much.

Since some methods provide a writeable reference, it's easy to violate the "sequential times" invariant. Everyone seemed to be okay with removing the ability to get a writeable reference, so that means we can properly enforce the "sequential times" invariant. 

Whenever someone appends a new state, we should make sure it's consistent with the other states. So this will become a new invariant as well.

Small bug: I forgot to document that `createsFromStatesStorage()` can throw an exception when the number of rows in the storage are not all the same.

Jen asked about how people can edit the trajectory, suggesting a GUI. I think that's a neat idea down the line, but for now people can build their own trajectories by creating a states storage file.

There seems to be consensus that `createFromStatesStorage()` should not try to assemble or equilibrate..that's the job of a more advanced trajectory builder. Ajay thinks that `createFromStatesStorage()` might be doing too much.

Jen asked about how tools deal with NaNs for state variables. It might be worthwhile to have an option with `createFromStatesStorage()` to set unspecified values to the default value instead of NaN (for IK?).

#### Filename extension

We discussed alternatives to `.ost`; Jen mentioned that this is confusingly close to `.sto`. Matt suggested `.states`. Jen likes `.ostates`. It's important to be consistent across all our file extensions. We discussed changing the extension for models away from `.osim`, and repurposing the `.osim` extension, but I don't think these would be good to do.

There was a discussion of a file (called a "simulation") that contains a model and any number of related trajectories.

Exceptions should go into the namespace of the class, and I should edit our doxygen style to make this look clean.

As of now, this class does NOT allow us to replace `formCompleteStorages()` that appears in other parts of the code (SimbodyEngine), though this PR takes a step towards getting rid of those methods.

Ayman mentioned filtering and padding done by tools to storage files. This could still be done to storage files before creating a StatesTrajectory from the storage.

#### Pairing with a model

Ayman: How should we deal with external forces, controllers, etc that get added by tools and that we consider to not be part of a model?
Chris: We can embed the exact model into the states trajectory file.

Tom: The states trajectory file could contain a metadata section like the data tables, to facilitate matching a trajectory with a model.

Matt mentioned creating a hash to pair a model with a states trajectory. We discussed that the hash would have to be on the topology, not on the xml file.

#### StatesContainer vs StatesTrajectory vs Simulation

Matt and Ajay brought up having tighter constraints on the contents of the trajectory, to certify if the trajectory came from a forward integration and thus have a known relationship between each other (Tom suggested that we'd also want to keep around the constraint tolerance, etc.). They discussed a new type, called Simulation, that would be subject to such constraints. Tom mentioned the idea of certificates of authenticity.

chrisdembia:
@sherm1 , we held off on discussing whether or not the cache should be stored in the StatesTrajectory file so that you could be part of the discussion.

sherm1:
> @sherm1 , we held off on discussing whether or not the cache should be stored in the StatesTrajectory file so that you could be part of the discussion.

Thanks! Actually some of that may not be a choice. It depends on whether it is feasible to recreate a functional State given only the non-cache info. It is always possible to recreate cache *values* from state, but it isn't clear to me yet whether the cache objects themselves (especially if they were allocated at Topology stage) can all be properly re-*allocated*.

chrisdembia:
Notes from meeting on 1/14:

Cropping a trajectory should use a tolerance.

Matt suggested that the need to reuse/repurpose a state is more immediate than ensuring a state came from a given model.

We decided not to worry about an equality operator for States, and also not to save the cache values in the state XML.

chrisdembia:
This PR is now ready for detailed review. There are still a few changes I'd like to make, but the review can start without them.

@klshrinidhi @aseth1 @sherm1 @jimmyDunne @msdemers

jenhicks:
Fixes #149

chrisdembia:
Just an update on where I am with this PR: I just fixed build failures on appveyor. The remaining tasks are to move `IteratatorRange` and `StatesTrajectory::isConsistent()` to Simbody, and to fix an issue with GCC related to `IteratorRange`. These changes should have little effect on the review.

OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -0,0 +1,641 @@
...
+
+/** A sequence of SimTK::State%s that can be saved to a plain-text OSTATES
+ * file.  The trajectory can also be populated from such a file. You may obtain
+ * a StatesTrajectory from a simulation, or other numerical methods whose task
+ * is to produce a trajectory of states. Users can modify the trajectory by
+ * appending states to it, but users cannot modify the individual states that
+ * are already in the trajectory.
+ *
+ * This class was introduced in OpenSim version 4.0, and enables scripting
+ * (Python/MATLAB) and C++ users to postprocess their results much more
```
aseth1: results with greater ease and more versatility?

OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -0,0 +1,641 @@
...
+ * flexibly than with an Analysis.
+ *
+ * ### Guarantees
+ * This class is designed to ensure the following:
+ * - The states are ordered nondecreasing in time (adjacent states *can* have
+ *   the same time).
```
aseth1: why is time not strictly increasing? What does it mean to have multiple state values that share the same time?

OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -0,0 +1,641 @@
...
+ * with a generic gait2392 model as well as with a scaled (subject-specific)
+ * gait2392 model. This flexibility may be beneficial in some scenarios, but
+ * also allows one to accidentally use the wrong model with a given states
+ * trajectory, potentially leading to silent errors that could compromise a
+ * scientific study.
+ *
+ * To increase your confidence that a StatesTrajectory matches a given Model,
+ * you can perform some weak checks with isCompatibleWith().
+ *
+ * TODO acceleration-level calculations?
```
aseth1: comment seems orphaned. Are you considering performing acceleration-level calcs as a way of checking something?

OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -0,0 +1,641 @@
...
+ *
+ * ### File format
+ * StatesTrajectory files use the file extension `.OSTATES`, with the XML
```
aseth1: my preference would be not to capitalize extensions.

OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -0,0 +1,641 @@
...
+ * 
+ * type of data                 | saved in OSTATES?
+ * ---------------------------- | -------------
+ * (continuous) state variables | yes
+ * discrete state variables     | yes
+ * modeling options             | yes
+ * cache variables              | no
+ *
+ * The cache variables (e.g., total system mass, control signals) are not saved
+ * to the OSTATES file because they can be regenerated from the other data and
```
aseth1: could refer to the definition of a cache variable as being a variable that can be computed/realized from state values and is conceptually separate from the state (although it is tucked in the State because they are intimately related - cache on state).

OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -0,0 +1,641 @@
...
+ *
+ * Without a model, you can access the state variable values, but you won't
+ * know the identity of such state variables.
+ * @code{.cpp}
+ * int numGeneralizedCoordinates = states[0].getNQ();
+ * const SimTK::Vector& generalizedCoordinateValues = states[0].getQ();
+ * @endcode
+ *
+ * To do most things with the StatesTrajectory, you'll need a model. **It is
+ * essential that you've called `Model::initSystem()` before you try to use any
```
aseth1: you'll need a model with a system. It is therefore required that `Model::initSystem()` becalled before ...

OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -0,0 +1,641 @@
...
+    // TODO this function should be pushed to the SimTK::State class, so that
+    // the check can evolve with the State class.
+    static bool isConsistent(const SimTK::State& stateA,
+                             const SimTK::State& stateB);
+
+    std::vector<SimTK::State> m_states;
+
+public:
+
+    /** Thrown when asking for a state at a time that is out of range. */
```
aseth1: Nested classes are shown at the top in Doxygen and it would be great if it had a heading like: "Exceptions thrown by StatesTrajectory" or "StatesTrajectory Exceptions". I really like these!

OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -0,0 +1,641 @@
...
+
+    /** Thrown when trying to append a state that is not consistent with the
+     * rest of the trajectory. */
+    class InconsistentState : public OpenSim::Exception {
+    public:
+        InconsistentState(const std::string& file, size_t line,
+                const std::string& func, const double& stateTime) :
+                OpenSim::Exception(file, line, func) {
+            std::ostringstream msg;
+            msg << "Cannot append the provided state (time " <<
```
aseth1: .. (at time = " ...

OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -0,0 +1,641 @@
...
+    
+    /** Thrown by createFromStatesStorage(). */
+    class StatesStorageIsInDegrees : public OpenSim::Exception {
+    public:
+        StatesStorageIsInDegrees(const std::string& file, size_t line,
+                const std::string& func) : OpenSim::Exception(file, line, func) {
+            addMessage("States Storage is in degrees, but this is inappropriate "
+                    "for creating a StatesTrajectory. Edit the Storage so that "
+                    "angles are in radians, and set 'inDegrees' to "
+                    "yes in the header.");
```
aseth1: should be set 'inDegrees=' to no! right?

OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -0,0 +1,641 @@
...
+    const SimTK::State& back() const { 
+        return m_states.back();
+    }
+
+    /** Get the index of the state just before (or at) the given time.
+     * @throws TimeOutOfRange If there are no states before the given time.
+     * @throws Exception If the trajectory is empty.
+     * @see getIndexBefore(const double&, const double&)
+     */
+    size_t getIndexBefore(const double& time) {
```
aseth1: why the two separate methods and not just have the tolerance default to 0 or `SimTK::Eps` or `SimTK::SignificantReal` if unspecified? User would still have the option to set the tolerance if they didn't like the default behavior.

OpenSim/Simulation/Test/testStatesTrajectory.cpp
```diff
@@ -0,0 +1,877 @@
...
+        state.setTime(3.999999);
+        states2.append(state); // 1
+        state.setTime(4.5);
+        states2.append(state); // 2
+        state.setTime(5.000001);
+        states2.append(state); // 3
+        state.setTime(9.3);
+        states2.append(state); // 4
+        SimTK_TEST(states2.getIndexAfter(4.0) != 1);
+        SimTK_TEST(states2.getIndexBefore(5.0) != 3);
```
aseth1: with slop (tolerance) > 1e-6 you would expect this to return 3, yes? maybe 'getIndexAtOrBefore(time)' would better match the use/intent?

OpenSim/Common/Storage.cpp
```diff
@@ -432,43 +422,52 @@ getStateIndex(const std::string &aColumnName, int startIndex) const
...
-            // try replacing the '/' with '.' in the last connection
-            shortName = aColumnName;
-            shortName.replace(back, 1, ".");
-            back = shortName.rfind("/");
-            shortName = shortName.substr(back + 1, shortName.length() - back);
-            thisColumnIndex = _columnLabels.findIndex(shortName);
-        }
+    thisColumnIndex = _columnLabels.findIndex(shortName);
+    if (thisColumnIndex >= 0)
+        return thisColumnIndex - 1;
```
aseth1: I understand the preference to short circuit and return sooner rather than later, but at least the original structure was isolating the `index -1` goofiness to one place that was documented. Now the same 'if' block is repeated, so I don't see the benefit of the rearrangement. The logic is less clear to me now.

OpenSim/Simulation/StatesTrajectory.cpp
```diff
@@ -0,0 +1,393 @@
...
+    for (int itime = 0; itime < sto.getSize(); ++itime) {
+
+        // Extract data from Storage into the working Vector.
+        sto.getData(itime, numDependentColumns, dependentValues);
+
+        // Set the correct time in the state.
+        sto.getTime(itime, state.updTime());
+
+        // Put in NaNs for state variable values not in the Storage.
+        for (const auto& stateName : missingColumnNames) {
```
aseth1: Should this be NaN or model's default for this state value? NaN may be better so that the value is not blinded used as though it were read in - when in fact it did not exist. A StateTrajectoryBuilder could read in a StateTrajectory and produce a new with NaNs resolved based on convenient rules, like use default values and solver for equilibrium, etc ... is that what you are thinking?

OpenSim/Common/Exception.h
```diff
@@ -137,11 +137,18 @@ class OSIMCOMMON_API Exception  : public std::exception {
               int aLine=-1);
 
     /** Call this constructor from derived classes to add file, line and 
-    function information to the error message                                 */
+    function information to the error message.                                */
     Exception(const std::string& file,
               size_t line,
               const std::string& func);
 
+    /** Use this when you want to throw an Exception (with OPENSIM_THROW or
```
aseth1: Do we need another method vs. default `msg` to be empty?

Bindings/Python/tests/test_states_trajectory.py
```diff
@@ -0,0 +1,243 @@
...
+        states.append(state)
+        state.setTime(0.9)
+        states.append(state)
+        state.setTime(0.9)
+        states.append(state)
+        state.setTime(0.9)
+        states.append(state)
+        state.setTime(1.2)
+        states.append(state)
+
```
aseth1: Isn't the use case that states are not full precision and therefore very close to use perceived start/end times that are? I am not convinced the tolerance works as expected in that case where you would allow some slop above or below the actual value as a match.

aseth1:
@chrisdembia  I am done reviewing at it is really nice work! The documentation is easy to follow.
My one concern is whether the finding the time tolerance works as expected. Also the expected before and after iterator designations with a tolerance behavior is unclear to me. 

For getting an index, i would prefer something like: `getIndexAtTime(time, TimeMatchOption, tolerance)` method where `TimeMatchOption = {AtOrBefore, AtOrAfter, Closest} and tolerance is used to evaluate the closeness such that `abs(time- state.getTime() <= tolerance` in all cases otherwise you get an exception thrown that there is no matching time.

OpenSim/Simulation/StatesTrajectory.h
chrisdembia: I tried adding a heading but it didn't work (I guess doxygen member groups don't work for nested classes). However, I just opened a separate PR to move the nested classes further down: https://github.com/opensim-org/opensim-core/pull/755

OpenSim/Simulation/StatesTrajectory.h
chrisdembia: I was on the fence about this, and I'm happy to change it. Any other opinions out there?

To some extent, allowing the states to have the same time allows one to use the trajectory as just any container of states (perhaps multiple iterations of inverse kinematics for a single pose). On the other hand, if someone is appending multiple states with the same time, perhaps it means they are doing something unintended (a bug)? It may also be possible that some algorithms/simulations produce two final states with the same time; I feel like I may have seen this in the past.

OpenSim/Simulation/StatesTrajectory.h
chrisdembia: I deleted the comment. I just had trouble realizing to the acceleration level with a states trajectory and I'm not sure why (there's a TODO related to this in the test case as well: https://github.com/opensim-org/opensim-core/pull/730/files#diff-b84db054c2c290c48a2508428576311eR214). Maybe because I didn't specify controls?

I was planning on not addressing the acceleration-level issue for now.

OpenSim/Simulation/StatesTrajectory.h
sherm1: Technically, discontinuous events do produce two states at the same time: before and after the event is handled. Simbody's integrators generally avoid this by making the event cross a very small but finite amount of time, but in general it does make sense to have (no more than) two adjacent states at the same time, with the implication that something that was modeled as instantaneous occurred at that time.

OpenSim/Simulation/StatesTrajectory.h
chrisdembia: Could you clarify if you mean just when extension appears with the period (`.OSTATES`), or even when it appears without the period (`OSTATES`)? 

I'm indfferent in the case that the period is there (`.ostates`), but I still prefer upper-case when leaving out the period (`OSTATES`); perhaps just because it seems more common. Though, I'll follow whatever your preference is.

OpenSim/Simulation/StatesTrajectory.h
chrisdembia: Oh interesting. Well, I could add a check that no more than two adjacent states have the same time. This would, however, still allow accidents where someone appends the same state twice unintentionally, etc.

chrisdembia: There's also the super minor issue that we're using finite precision numbers. Someone might not be able to represent two slightly-different times by two different `double`s. Doesn't seem like this is a practical issue, though.

OpenSim/Simulation/StatesTrajectory.h
chrisdembia: I would prefer to leave this part of the documentation as-is. Two reasons: (1) you probably need more than just the model and system; you also need some basic realization (model, instance); (2) I believe that most users don't know what the "system" is, and don't know that `initSystem()` creates  a `SimTK::System` (I think they view "system" as just a generic term).

OpenSim/Simulation/StatesTrajectory.cpp
chrisdembia: > NaN may be better so that the value is not blindly used as though it were read in - when in fact it did not exist.

I agree.

> A StateTrajectoryBuilder could read in a StateTrajectory and produce a new with NaNs resolved based on convenient rules, like use default values and solver for equilibrium, etc ... is that what you are thinking?

I hadn't thought of that per-se, but I like the overall idea. One potential issue is if some of the NaNs really should be NaNs (perhaps b/c foot is not in contact or something like that? I can't think of an actual example where this would be the case). Another option is that this method, in the future, could have an optional parameter for what to use for unspecified state variables.

OpenSim/Common/Exception.h
chrisdembia: I started consolidating the two constructors, but as I was doing it I started to feel that the two constructors might be clearer as separate constructors. The two constructors have distinct uses: one is for calling from the constructor of a derived class, and the other is a convenience when you don't want to make a derived class. I felt that consolidating the two constructors made this distinction confusing.

chrisdembia:
@aseth1 I had considered using "slop" instead of "tolerance", and I see that you also used "slop." Which do you think I should use? Slop or tolerance? I went back and forth on this.

OpenSim/Simulation/Test/testStatesTrajectory.cpp
chrisdembia: My documentation of this test is a little confusing. This particular test is making sure that the methods give the *wrong* indices when an appropriate slop is *not* provided. That is, the user would expect to get 3, but didn't, because they did not provide the appropriate slop. I'm going to address this comment by improving the test's documentation. I'll also address the method names, but separately.

Bindings/Python/tests/test_states_trajectory.py
chrisdembia: Could you clarify your question? I think there was a typo.

This python test is much weaker than the related C++ test; the C++ test `testAccessByTime()` attempts to ensure that the tolerance works as expected.

OpenSim/Simulation/StatesTrajectory.h
chrisdembia: The use of separate methods for the tolerance was intentional. I think it's rare that someone wants to use the tolerance, and thus that in most cases it would only make the methods look more complicated than they are. If the users run into some unexpected behavior, they can look at the doxygen again and find out that there's another variant (that accepts a tolerance) to help them. This is distinct from the `createFromStatesStorage()` method, where I think the default arguments are fairly easy to comprehend and that users are likely to want to alter them.

chrisdembia:
> For getting an index, i would prefer something like: getIndexAtTime(time, TimeMatchOption, tolerance) method where TimeMatchOption = {AtOrBefore, AtOrAfter, Closest} and tolerance is used to evaluate the closeness of the match such that abs(time- state.getTime() <= tolerance in all cases, otherwise you get an exception thrown that there is no matching time.

I like this, and I'll try it out, though I'll still create separate methods for the tolerance. A small issue is that these methods will be harder to use in python, but this is a minor issue. I'll also use a default value for `TimeMatchOption` as `Closest`.

I'm a little unsure about using `AtOrBefore` (instead of `Before`). Say the user is looking at a STO file that they also have in memory, and the STO file lists a time of 0.53333. Say that, in memory, this state is represented as 0.533330001. If the user calls `getIndexAtTime(0.53333, AtOrBefore)`, they will expect to get the 0.53333 state, but actually won't. Whereas if the enum value were named `Before`, they would be more likely to do something like `getIndexAtTime(0.534, Before)` or `getIndexAtTime(0.53333, Closest)`. That is, `AtOrBefore` might give users the illusion that they'll be able to specify an exact time successfully (same problem for a  method named `getIndexAtTime()`).

chrisdembia:
> Also the expected before and after iterator designations with a tolerance behavior is unclear to me.

Would this be addressed by a similar `getIteratorAtTime(time, TimeMatchOption)` method?

OpenSim/Common/Storage.cpp
chrisdembia: These changes were mainly to fix a bug where the function returned `-2` instead of `-1` when the desired column was not found. I agree that the repeated code is undesirable; I'll play around with this a bit more.

OpenSim/Simulation/StatesTrajectory.h
aseth1: I think most times users do require a tolerance (due to imprecision) but have no idea what it should be. Setting a default value keeps the usage simple but allows control if/when needed. It will likely cause more frustration unless you also bake in an internal tolerance in this case.

OpenSim/Simulation/StatesTrajectory.h
aseth1: i don't like saying something is "essential" without any remark as to why. The association of the state is with the system as you remark [here](https://github.com/opensim-org/opensim-core/pull/730/files#diff-bb1a7ce0ff41c3218c6887695c614ad1R104). I acknowledge that we should either make the model<->system relationship more prominent/clear in the documentation or we should attempt to hide it completely.

Bindings/Python/tests/test_states_trajectory.py
aseth1: update the comment above. Sorry it was totally unclear before!

chrisdembia: > Isn't the use case that actual states are truncated and have a loss of precision (e.g. 4.99999999999) but very close to user intended start/end times that are crisp/precise (like 5)?

Yes, I think so.

> I am not convinced the tolerance works as expected in that case where you would allow some slop above or below the truncated value as a match to the user crisp value.

I think the tolerance works in this case (that, or I don't understand this scenario). I'll respond to this in more detail in the main thread.

OpenSim/Simulation/Test/testStatesTrajectory.cpp
chrisdembia: > with slop (tolerance) > 1e-6 you would expect this to return 3, yes?

To answer your question directly, yes. This test is checking the behavior when the slop is < 1e-6, for sanity.

chrisdembia:
@aseth1 can we discuss the tolerance, etc. in person?

OpenSim/Simulation/StatesTrajectory.h
chrisdembia: Addressed.

OpenSim/Simulation/StatesTrajectory.h
chrisdembia: Addressed by uncapitalizing `.OSTATES`.

OpenSim/Simulation/StatesTrajectory.h
chrisdembia: I updated the description of cache variables, and also refer the reader to related documentation in Component.

OpenSim/Simulation/StatesTrajectory.h
chrisdembia: I've replaced "essential" with "required" and also I now mention the SimTK::System.

OpenSim/Simulation/StatesTrajectory.h
chrisdembia: Fixed.

OpenSim/Simulation/StatesTrajectory.h
chrisdembia: Yes! Thank you. This is fixed now.

OpenSim/Simulation/StatesTrajectory.h
chrisdembia: This will be addressed in #773.

OpenSim/Simulation/Test/testStatesTrajectory.cpp
chrisdembia: Moved to #773.

OpenSim/Common/Storage.cpp
chrisdembia: I updated this function to reduce code repetition.

chrisdembia:
Here's a summary of @aseth1's feedback and what I've addressed:

EDIT: TODO after Ajay's first review.

- [x] "results with greater ease and more versatility?"
- [x] Remove `TODO acceleration-level calculations?`
- [x] Un-capitalize `.OSTATES`.
- [x] Update description of cache variables.
- [x] Update `Model::initSystem()` warning (explain why this call is essential).
- [x] Nested exception classes #755.
- [x] ".. (at time = " ..."
- [x] "should be set 'inDegrees=' to no! right?"
- [x] Storage::getStateIndex()
- [x] StatesTrajectory::createFromStatesStorage() value for unspecified columns (NaN).
- [x] Exception constructors; default argument `msg = ""`.
- [ ] "why is time not strictly increasing?"

So, the only thing to address for this PR is if time should be nondecreasing or strictly increasing. My feeling is that it's a little too strict to enforce strictly increasing times; I imagine there may be use cases where you want the same time in multiple states. Though I'm happy to change this and revisit after this class has been in use.

aseth1:
I still believe that having stationary time (remaining the same over multiple rows) is an error case. It is physically impossible. If this was simply a collection of states then order in time would not matter (say from a monte carlo process) but trajectory to me means evolving through time and a model can not be in two poses at the same time.

aseth1:
@chrisdembia happy to leave enforcing time as strictly increasing as a to do post hackathon.

tkuchida:
> I still believe that having stationary time (remaining the same over multiple rows) is an error case

I disagree. I think an integrator could, in theory, report two states with the same time if an event is processed in between. I don't think StatesTrajectory should be making the determination as to whether this is permitted. :palm_tree:

chrisdembia:
Okay I am happy to enforce strictly increasing time after the hackathon. Thanks for reviewing and merging.

Issue #692, opened 2015-11-02
DataTable warnings with MSVC.


chrisdembia:
```
C:\projects\opensim-core\OpenSim/Common/DataTable.h(368): warning C4267: '=': conversion from 'size_t' to 'unsigned int', possible loss of data [C:\projects\opensim-core\build\OpenSim\Examples\DataTable\example1DataTable.vcxproj]
C:\projects\opensim-core\OpenSim/Common/DataTable.h(368): warning C4267: '=': conversion from 'size_t' to 'unsigned int', possible loss of data [C:\projects\opensim-core\build\OpenSim\Examples\DataTable\example2DataTable.vcxproj]
```

There are additional similar warnings as well.

@klshrinidhi 

Thanks @aymanhab

sherm1:
I think I fixed this in PR #736 -- Chris, can you verify and close if so?

chrisdembia:
Thanks.

PR #736, opened 2016-01-16
Fix some warnings and uses of SimTK::Xml

sherm1:
As part of the State serialization project, the SimTK::Xml *class* will be the SimTK::Xml *namespace*. In most cases this is backwards compatible because other classes like Element were local to Xml and hence referenced `Xml::Element`, which will still work. There were a few `using` statements and the like that will have to change; I believe the changes here will work now and after Xml gets promoted to namespace.

Also fixed miscellaneous new warnings caught by VC++.

### DIFF
--- a/OpenSim/Common/DataTable.h
+++ b/OpenSim/Common/DataTable.h
@@ -258,7 +258,7 @@ public:
     RowVectorView getRowAtIndex(size_t index) const {
         OPENSIM_THROW_IF(isRowIndexOutOfRange(index),
                          RowIndexOutOfRange, index, 0, _indData.size())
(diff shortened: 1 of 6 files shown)


chrisdembia:
Looks good to me, assuming the tests pass.

This PR *might* also fix #692.

sherm1:
Yes, #692 looks like the same warnings I just fixed. I'll merge this and make a note there.

chrisdembia:
Did you intend to not delete the branch?

sherm1:
> Did you intend to not delete the branch?

Nope -- thanks. Gone now.

PR #737, opened 2016-01-17
Upgrade travis compilers. clang-3.4 -> clang-3.7 , gcc-4.8 -> gcc-5.2

klshrinidhi:
Upgrading compilers used on travis for Linux. Binaries are available for both clang and gcc.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@
 language: cpp
 
 # Use container-based infrastructure to allow caching (for ccache).
-sudo: false
+# sudo: false
     
 matrix:
   include:
@@ -38,20 +38,12 @@ addons:
   # Dependencies on linux.
   apt:
     sources:
-      # For gcc >= 4.8
-      - ubuntu-toolchain-r-test
       # For cmake >= 2.8.8 (for CMakePackageConfigHelpers)
       - kubuntu-backports
     packages:
       - cmake
       # For Simbody.
       - liblapack-dev
-      # C++11 on Ubuntu. Update to gcc 4.8, which provides full C++11 support.  We
-      # use this script because if building Simbody with C++11, we need gcc-4.8,
-      # and the Travis Ubuntu 12.04 machines have an older version of gcc. Even if
-      # building with Clang, we need the newer libstdc++ that we get by updating to
-      # gcc-4.8.  See https://github.com/travis-ci/travis-ci/issues/979.
-      - g++-4.8
       # In case someone wants to check for memory leaks.
       - valgrind
       # To build doxygen documentation.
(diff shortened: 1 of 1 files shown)


klshrinidhi:
Travis builds are exceeding maximum allowed time limit (49min approx). Build is successful otherwise. Not sure how to solve this time limit problem.

chrisdembia:
Clang 3.7 and GCC 5.2 are available in Ubuntu 15.10 but not 15.04. So we have to decide if we are willing to limiting our Ubuntu support to the latest Ubuntu. Are these changes necessary for the data adapters?

If we decide to upgrade, we should also update the README.

The slowdown is probably because we are not using the container-based machines (sudo: false), and thus are not benefitting from ccache. 

We shouldn't suppress all warnings; only the ones we don't want. I think the clang warnings are just coming from two different kinds of warnings.

klshrinidhi:
@chrisdembia 
clang-3.7 is available for 12.04, 14.04, 15.04 and 15.10 from [here](http://llvm.org/apt/). Upcoming clang-3.8  nightly builds are available too.
gcc-5 is available for all of the above except 15.04 from [here](https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test). The one entry for gcc-5 without distribution name may be the one for 15.04.

I think we should upgrade and keep the compilers on travis the same as the ones we use on our machines. 
I have been using clang-3.7 and gcc-5.3 to develop. clang-3.4 on travis was throwing errors which I could not reproduce.

sherm1:
We need to pick a minimum Ubuntu version that will be required by OpenSim 4.0 and it should be one of the long term support versions (ending in "04"). Using compilers that aren't available until 15.10 means the minimum will be 16.04, which won't be released until April of this year. 

Why can't OpenSim be built using the compilers in 15.04? That would avoid forcing end users to upgrade to the very latest Ubuntu in order to use OpenSim.

chrisdembia:
> Why can't OpenSim be built using the compilers in 15.04? That would avoid forcing end users to upgrade to the very latest Ubuntu in order to use OpenSim.

LTS releases are every two years; the latest LTS is 14.04.



> The packages would install but the name of the binaries for new gcc and clang was not clear.

Maybe using `apt-update-alternatives` would help?

sherm1:
> LTS releases are every two years; the latest LTS is 14.04.

Thanks -- I misunderstood. I see that has support through 4/2019 and 16.04 will be supported through 4/2021. In that case 15.04 is no better than 15.10. But we should pick a particular release and not require users to upgrade the compilers from what came with that release. OpenSim is not likely to be the only software being run on a user's machine and we should not assume they can make system changes to accommodate it. This is particularly important for users running on clusters since they may not have admin rights.

chrisdembia:
That's a good point. We should also check what's available on the Stanford Sherlock cluster, since we use them ourselves.

klshrinidhi:
This issue is applicable to a very small group of users -- developers on Linux/Mac who are building OpenSim from sources to write code and link against OpenSim libraries. Binaries are available for both Linux and Mac. The latest Xcode also comes with LLVM 3.7 (https://en.wikipedia.org/wiki/Xcode#Toolchain_versions). On Linux, multiple versions of a compiler can coexist. 
I think we should upgrade compilers to a version whose binaries are available. I don't think this will break any existing software to user's machine.

chrisdembia:
I think we should continue discussing in person.

aymanhab:
I'd like to be included in this discussion because that's a policy decision
that affects everybody on the dev side and users as well. If the code
compiles and runs with earlier versions and it's a matter of cleaning up
warnings then it's less of an issue than if the code fails to compile/run
with widely used/supported compilers.

While developers can have a branch or local environment that uses the most
bleeding edge compilers, I think default Travis builds should use some
agreed upon/mature release(s) for the reasons that Sherm described earlier.

klshrinidhi:
@aymanhab 
The newer compilers increase the support for C++11/14. 
May be not related to C++11/14 support, but I was having problems with clang-3.4 on Travis previously.

I think it is best for us to release OpenSim 4.0 with latest compilers whose binaries are readily available at the time of release.  
* Developing on newer compiler versions on local machines and having travis use older compiler versions just increases the amount of work we have to do.
* I think this issue is applicable to very few users -- those who are building OpenSim sources in Linux. Mac users already get latest clang with Xcode. https://gist.github.com/yamaya/2924292

chrisdembia:
> Mac users already get latest clang with Xcode. https://gist.github.com/yamaya/2924292

If they decide to upgrade to the latest OSX / Xcode, though I think it is very common for users to do so.

Fortunately, we have sent out a survey and perhaps that'll help us see how many people use Ubuntu/Linux.

@jenhicks @jimmyDunne have we posted the survey to the forum?

klshrinidhi:
@chrisdembia 
I think it is okay for us to require latest tools if one wants to *develop* with OpenSim. I think we need not freeze the compilers to what users have today.

chrisdembia:
The latest gcc available on the sherlock cluster is 4.9.1 (the latest clang is 3.4). While the sherlock cluster admins are very responsive to me @carmichaelong, other people on other clusters might not be so lucky.

Of course it's *possible* to download gcc binaries, but that's a lot of friction and could halt someone from putting opensim on their cluster. Consider that many of the people using opensim on linux are beginners, and are almost certainly not CS people.

Mostly this sounds like a time tradeoff. How much time are we willing to invest versus forcing some users to invest.

chrisdembia:
> @chrisdembia 
> I think it is okay for us to require latest tools if one wants to develop with OpenSim. I think we need not freeze the compilers to what users have today.

I agree the picture starts to change once we start releasing binaries for Ubuntu. It's true that I am stuck in the mindset that all linux users have to build from source, b/c that's the state of our current linux support.

sherm1:
I need c++11 standard regular expressions (`<regex>`) for the State serialization project. That requires at least gcc 4.9 because they weren't implemented in 4.8 (the headers were there but the library was unimplemented). 4.9 was first included in Ubuntu 15.04, I believe.

Issue #119, opened 2014-07-23
Component does not handle Connector, Input and Output copying upon clone()


aseth1:
A component that is currently cloned will find it has not Inputs or Outputs. I don't believe the current use of a map containing unique_ptrs resolves this issue although it does fix the problem of leaking Inputs and Outputs.

chrisdembia:
What exactly is the desired behavior upon copy of, e.g., connected inputs? If I make a copy of a model that contains connected inputs, I want the inputs to be wired up to the NEW model's outputs, not the previous one's.

aseth1:
The connection should not be copied. The connector and its property that identifies its connectee should be copied. You should be forced to call connect() or connectToModel() on the model that copy is added to. This guarantees that the copy has no effect on the original component, its connections and the model it belongs to.

Issue #536, opened 2015-07-16
Component iterator of its subcomponents should be valid during Model building


aseth1:
As it is now the iterator tree is only built at Model::connect(). We should discuss employing a suitable underlying std:: data structure instead of an adhoc linked list. In that way we could support inserts and deletes of subcomponents (as references) .  A related problem is that most subcomponents are also properties, and a reference to a component in the property list is not guaranteed to be valid if the property list is edited (e.g. another element is added) even if the component itself is untouched. Guaranteering  validity of a reference would allow us to have a valid iterator tree on every add such that the tree was iteratable immediately instead of waiting until connect (which guarantees properties will no longer be edited). 

Relates to #418.

chrisdembia:
How would the `std::` data structure be notified of the updated property list?

aseth1:
We could require the use `<ConcreteComponent>::add<C>()` methods to append to the property list and also update the list of subcomponents (which would be the `std::list` or  `std::tree`) e.g. `Component::markAsSubcomponent`. We would have to make the list property protected so users cannot `upd_<prop_list>()` directly otherwise they could be out of synch. Model already has `add<C>()` for all its subcomponent types (e.g. `addBody()`, `addJoint()`, ...) and these could also mark a component as a subcomponent instead of waiting until `finalizeFromProperties`. It does require the reference in the property list (or Set) to remain valid, however.

aseth1:
A more comprehensive solution is to enable Component to have a special property list (like we are doing for connectors) for all subcomponents intended to serialize them and to iterate over them. Then component builders would not have to define them as properties and then manually mark them as subcomponents - anything in that list is a subcomponent and serialized. Uber component's like Model then filter the large flat list using the component iterator based on type to handle Bodies, Joints, etc.. in required order. That would: 1) centralize/unify all all handling of subcomponents. A `markAsSubcomponent()` method could still exist to allow "hidden" subcomponents (however, there are no examples of that atm).   2) it would eliminate most of the use of sets except for the notion of groups. If subcomponent list could be easily grouped (which is trivial for types) then we could also have model files more or less preserved by grouping all Bodies, Joints, Constraints, etc... together, and also allow convenient access to useful sublists. Again, this may be better handled if Component also has a notion of membership to groups, then an iterator filter could be used to find Components that have the same group affiliation.

aymanhab:
[...]

Issue #421, opened 2015-04-30
When and where to mark Components as subcomponents is not clear.


aseth1:
There are a couple of issues here:

1. Components should only be marked as subcomponents during `finalizeFromProperties()` (e.g. your Component's `extendFinalizeFromProperties()`. The marking of a component simply adds a reference to that component to a list. If where the Component lives (in memory) changes (due to array resizing which allocates and copies) then that reference becomes invalid. That is one of the issues raised in #417.
2. This fact is not well documented and furthermore there are instances e.g. [here in Model](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/Model.cpp#L835)  where `addComponent()` is being used (incorrectly) but seems completely sensible. Model was getting away with it because it was explicitly clearing and rebuilding the subcomponent list in its `extendFinalizeFromProperties()`.

I propose we:
- [ ] remove all calls to `addComponent()` that exist outside of an `extendFinalizeFromProperties()` implementation.
- [ ] rename `addComponent` to `markAsSubcomponent()` which is more consistent with its documentation and clearly state it should only be called from the component's `extendFinalizeFromProperties()` and add a similar comment there to call `markAsSubcomponent()`.

sherm1:
:+1:

jenhicks:
This currently being adressed by @aseth1

PR #740, opened 2016-01-21
Properties that are components are marked as subcomponents

aseth1:
The main objective of this PR is to simplify handling of subcomponents (#421) and to make them immediately traverseable to serve model building and other tasks (#536). To this end, the following principles were instituted:

1. Components own their subcomponents.
 - a Component owns it properties and properties can be components: these are subcomponents.
   - user/developer *never* has to mark components that are properties as subcomponents
 - components that are data members of a Component are also its subcomponents (also do not need to be marked)
 - any other owned Component can be marked as a subcomponent (e.g. Model generated loop breaking constraints)

2. The ability to iterate through components is only dependent on the subcomponents list

**The main priorities for review are:**

1. Changes to Component so that:
 - components that are properties are listed as subcomponents automatically 
 - subcomponents (and all other structural lists) are copied appropriately
 - no requirement to connect() to traverse through subcomponents 

2. Update to Model:
 - removed the requirement to connect() in Model::extendFinalizeFromProperties()
 - subcomponent list and multibody tree are decoupled (tree building can now be delayed to connecting stage)
 - adding a Component assumes model takes ownership of it and it is a member of a property list.

3. Related test cases: testComponentInterface, testSerializeOpenSimObjects, testComponents, testIterators to verify subcomponents (as a property and internal member) are handled correctly and are traverseable. 

4. Additional Intention/consequence:
  -  always finalize Component properties upon construction (if Components are properties they need to be properly in place first). Only exception is for deserialization testing where we know there isn't a valid model there and we only care about comparing property values not their validity. 

5. Known Hacks/Bugs
 - I have not dealt the with the ComponentList to traverse subcomponents. To enable getting a valid ComponentList at anytime I force the creation of Component's traversal tree whenever a ComponentList is asked for. I put in some ugly const casting that MUST be removed in a subsequent PR, but that is independent of the the current changes. 
 - The Model leaks secondarily created Components needed to create a valid multibody tree. That is instead of adding to properties (and thus the serialized version of the Model) I just mark it as a subcomponent but don't hang on to it in order to delete it. There are several easy options to fix this but wanted to discuss separately and not delay this PR.

The remainder of the changes are to handle the fallout of the above changes, which were many. I also apologize for any spurious changes made as cleanup while debugging and testing.

Thanks and happy reviewing!


OpenSim/Simulation/Model/ContactMesh.cpp
```diff
@@ -37,7 +37,7 @@ ContactMesh::ContactMesh() :
     setupProperties();
 }
 
-ContactMesh::ContactMesh(const std::string& filename, const SimTK::Vec3& location, const SimTK::Vec3& orientation, Body& body) :
+ContactMesh::ContactMesh(const std::string& filename, const SimTK::Vec3& location, const SimTK::Vec3& orientation, PhysicalFrame& body) :
```
tkuchida: Perhaps change `body` to `frame` (here and on line 60) to be consistent with ContactMesh.h.

OpenSim/Simulation/Model/ContactSphere.cpp
```diff
@@ -33,7 +33,7 @@ ContactSphere::ContactSphere() :
     setNull();
 }
 
-ContactSphere::ContactSphere(double radius, const SimTK::Vec3& location, Body& body) :
+ContactSphere::ContactSphere(double radius, const SimTK::Vec3& location, PhysicalFrame& body) :
```
tkuchida: `body` -> `frame` (and on line 45)?

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -688,8 +640,10 @@ void Model::extendConnectToModel(Model &model)
                 WeldConstraint* weld = new WeldConstraint(outb->getName()+"_weld",
                                                           *outbMaster, o, *outb, o);
 
-                // Add to list of subcomponents but not serialize ConstraintSet
-                updModel().addComponent(weld);
+                // TODO: add all added compoents to a private list of owned components
```
tkuchida: compoents -> components. Have you filed an issue?

OpenSim/Simulation/SimbodyEngine/PlanarJoint.h
```diff
@@ -1,7 +1,7 @@
 #ifndef OPENSIM_PLANAR_JOINT_H_
 #define OPENSIM_PLANAR_JOINT_H_
 /* -------------------------------------------------------------------------- *
- *                          OpenSim:  SliderJoint.h                           *
+ *                         OpenSim:  PlanarrJoint.h                           *
```
tkuchida: Planarr -> Planar

OpenSim/Common/Component.cpp
```diff
@@ -838,26 +836,61 @@ void Component::constructOutputForStateVariable(const std::string& name)
...
+                }
+                //Set<Component>* objects = dynamic_cast<Set<Component>*>(&obj)
+                else {
+                    std::string objType = obj.getConcreteClassName();
+                    if (obj.hasProperty("objects")) {
+                        auto& objectsProp = obj.updPropertyByName("objects");
+                        for (int k = 0; k < objectsProp.size(); ++k) {
+                            Object& obj = objectsProp.updValueAsObject(k);
+                            if (Component* comp = dynamic_cast<Component*>(&obj) )
+                                markAsSubcomponent(comp);
```
tkuchida: I was expecting a recursive call here. It isn't necessary to call `comp->markPropertiesAsSubcomponents();`?

OpenSim/Common/Component.h
```diff
@@ -935,6 +947,17 @@ template <class T> friend class ComponentMeasure;
         constructOutputs();
     }
 
+    /**
+    * Mark another Component as a subcomponent of this Component. Component
+    * methods (e.g. addToSystem(), initStateFromProperties(), ...) are
+    * therefore invoked on subcomponents when called on this Component.
+    * Realization is also performed automatically on subcomponents. This
+    * member function does not take ownership, it is assumed that the
+    * the component is a data member or property of this Component.
```
tkuchida: the the

OpenSim/Common/Component.cpp
aseth1: I currently have the marking as part of finalizeFromProperties(), which is already called recursively, but if we chose to decouple the two, then yes. I wanted to minimize what you have to call in the construct so that is my primary reason for lumping them.

tkuchida: Hm, not sure I understand completely. Here's my question more precisely:

Line 853 checks to see whether the property is a Component; if it is, then mark it as a subcomponent. Otherwise, line 859 checks to see whether the property might have Components buried within it, and line 863 (which is identical to line 853) checks for more Components at that level. But what if there are Components buried under _two_ layers of the `hasProperty("objects")` test? Is it not necessary to call lines 857 to 864 recursively in the event that there are Components many layers deep?

OpenSim/Common/Component.cpp
```diff
@@ -838,26 +836,61 @@ void Component::constructOutputForStateVariable(const std::string& name)
...
+        auto& prop = updPropertyByIndex(i);
+        if (prop.isObjectProperty()) {
+            for (int j = 0; j < prop.size(); ++j) {
+                Object& obj = prop.updValueAsObject(j);
+                if (Component* comp = dynamic_cast<Component*>(&obj) ) {
+                    markAsSubcomponent(comp);
+                }
+                //Set<Component>* objects = dynamic_cast<Set<Component>*>(&obj)
+                else {
+                    std::string objType = obj.getConcreteClassName();
```
tkuchida: `objType` doesn't seem to be used anywhere.

OpenSim/Common/Component.cpp
aseth1: The reason I go one level down is because the Set property has another level of "objects", which are the actual subcomponents. The subcomponents of those components are added when those components are asked to finalizeFromProperties() which is propogated down its subtree of components (of course the marking as subcomponents is the first thing to happen in finalizeFromProperties). Does that make more sense?

tkuchida: Ah, ok I see now. Thanks for the explanation. :gift:

chrisdembia: Given the confusion, it sounds like there should be a related comment in the code. Also, given the 6 close-brackets at the end of the function, is it possible to break up this method somehow? Potentially using lambdas?

OpenSim/Simulation/Model/ContactMesh.cpp
aseth1: Then I would have to also rename the data member to `frame` and which would also cause a request to change the property to `frame_name` which would beg the question why isn't it using a Connector to a Frame instead? And also why ContactGeometry doesn't simply require Frame now that we have OffsetFrames we don't need pass along location and orientation all over the place. Given that we are likely to revamp contact geometry (post dev week) to leverage Geometry I thought i would leave it alone. I could address making it coniststently use frame (instead of body) as a quick side PR.

tkuchida: Either way; not a big deal. I just thought you meant to change `body` to `frame` here because you changed it in the header file. :snowman:

aymanhab:
@aseth1 Thanks for the in person explanation yesterday regarding parent/ownership language. A few high level items based on reading the description, and Component.cpp:
1.  A good test for a hierarchical model building is to be able to take a model and another model and combine them into one model (maybe sharing ground only). Is that possible after this PR with real models?
2.  "Adding a component causes transfer of ownership", did anything change in the interface for how it's done. In the past we had to tell the scripting/SWIG about that so memory management doesn't blow up for scripting. This could be unrelated to this PR, but needs handling in SWIG. A related issue is handling stack allocated objects and adding them to the model.
3.  Does this PR handle or do the plumbing to allow for "internal" unexposed components. I'm interested because the "frame-geometry" can possibly use this functionality if available (a separate PR of course).
4.  Recomputing the tree-traversal on every call to getComponentList is expensive and a bit excessive especially that the tree doesn't change at all post initialization. A caching mechanism can be implemented to improve performance but I still think this can be optimized if you know exactly when components are added to the tree, so there's no cost post model building.
5.  Everything here works with Components while models have been made of ModelComponents, is there a long term future for this distinction?
6.  The ownership paradigm implies no sharing/re-reuse, what happens when a user tries to add a component already owned? Does the user need to "new" components to add them to the model? does the ownership switch again?

Thanks much, sorry for the long list, it maybe better to discuss in person if you feel github PR is not the right medium.

OpenSim/Common/Component.h
```diff
@@ -198,19 +198,16 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
     Component(const std::string& aFileName,
         bool aUpdateFromXMLNode = true) SWIG_DECLARE_EXCEPTION;
 
+    /** Construct Component as a subcomponent member of another Component.
+        Automatically marks the component as a subcomponent of the owner.
+        A subcomponent MUST be named. **/
+    explicit Component(const std::string& name, Component* owner=nullptr);
```
sherm1: There is something wrong with this constructor. From the comment I would have expected that the parent Component would be essential and that it would be passed as a writable reference. But it is optional and a pointer. I guess you intended this constructor to do double duty as some other kind of constructor but the comment doesn't describe that use. Probably there should be two distinct constructors something like this:
```c++
/** Create an empty Component base class object with a given name. */ // is this protected?
explicit Component(const std::string& name);
/** Create an empty Component base class object with a given name and make it a
subcomponent of a given parent Component. */
Component(const std::string& name, Component& parent);
```
That said, I still think the second constructor is awkward. Wouldn't this be more natural as an operation on the parent, something like `parent.adoptSubcomponent(Component* subcomponent);`?

OpenSim/Common/Component.h
```diff
@@ -198,19 +198,16 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
+
     /** Construct Component from a specific node in an XML document. **/
     explicit Component(SimTK::Xml::Element& aNode);
 
-    /** Copy Constructor. Required to perform custom handling of 
-        internal references to subcomonents and system indices.
-        The copy has to be connected in order to function. */
-    Component(const Component& source);
+    /** Use default copy constructor and assignment operator. */
```
sherm1: You can make that explicit now with:
```c++
Component(const Component&) = default;
Component& operator=(const Component&) = default;
```

OpenSim/Common/Component.h
```diff
@@ -198,19 +198,16 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
-    Component(const Component& source);
+    /** Use default copy constructor and assignment operator. */ 
 
-    /** Copy assignment.  Required to disconnect Connectors.
-        and reset indices. Musct call connect() on Component
-        after is has been assigned to another. */
-    Component& operator=(const Component &component);
-    
     /** Destructor is virtual to allow concrete Component to cleanup. **/
     virtual ~Component() {}
```
sherm1: This could also be `virtual ~Component() = default;`.

OpenSim/Common/Component.h
chrisdembia: I think I agree with @sherm1's comments, but I also think I know why @aseth1 made this constructor. The constructor is supposed to be used during the in-class member initialization of a member subcomponent. Since the in-class member initializer is called during construction of the parent (owner), is it permissible to call a method on the parent as a side effect of the in-class member initialization?

OpenSim/Common/Component.h
```diff
@@ -327,6 +324,8 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
      * must be done before calling this method on the top model. */
     template <typename T = Component>
     ComponentList<T> getComponentList() const {
+        Component* mutableThis = const_cast<Component*>(this);
+        mutableThis->initComponentTreeTraversal(*mutableThis);
```
sherm1: Can this be cached? The method here could be `mutableThis->calcComponentListIfNeeded()` or whatever. Maybe you could write it that way now and add the actual caching and invalidating later?

OpenSim/Common/Component.h
```diff
@@ -429,6 +428,16 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
         return getProperty_connectors().size();
     }
 
+    /** Access the number of Inputs that this component has. */
```
sherm1: "Access" doesn't seem like the right verb here. Maybe "Get" or "Return"?

OpenSim/Common/Component.h
```diff
@@ -916,6 +925,9 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
     // End of Model Component State Accessors.
     //@} 
 
+
+    void dumpSubcomponents(int depth=0) const;
```
sherm1: This probably needs a doxygen comment since it will appear there.

OpenSim/Common/Component.h
```diff
@@ -935,6 +947,17 @@ template <class T> friend class ComponentMeasure;
...
 
+    /**
+    * Mark another Component as a subcomponent of this Component. Component
+    * methods (e.g. addToSystem(), initStateFromProperties(), ...) are
+    * therefore invoked on subcomponents when called on this Component.
+    * Realization is also performed automatically on subcomponents. This
+    * member function does not take ownership, it is assumed that the
+    * component is a data member or property of this Component.
+    */
+    void markAsSubcomponent(Component* component);
```
sherm1: I don't understand why this is a pointer if there is no ownership change happening. Shouldn't it be a reference? Also in view of the PR description, this seems wrong since it would allow "marking" of a non-owned subcomponent. I'm not sure I get the term "mark" here either -- my understanding is that any component owned by a parent is a subcomponent of that parent. If that's right then shouldn't subcomponent-ness be a side effect of ownership?

Is this just an obsolete method at this point?

OpenSim/Common/Component.h
```diff
@@ -946,9 +969,9 @@ template <class T> friend class ComponentMeasure;
     with the line "Super::extend<xxx>(args);" to ensure that the parent class
     is called before the child class method.
     
-    The base class implementations ensures that the 
-    corresponding calls are made to any subcomponents that have been specified 
-    by derived %Component objects, via calls to the addComponent() method. 
+    The base class implementations ensures that the corresponding calls are made
+    to any subcomponents that have been specified by derived %Component objects,
+    according to the markAsSubcomponent() method.
```
sherm1: is this comment right? `markAsSubcomponent()` is still needed?

aseth1:
@aymanhab Thank you for the nice list of issues. I am going to repost most of these as an issue for further discussion outside this PR. The objective for this PR was narrower in scope and did not specifically target many of these, but I can provide more details how each of the items were impacted.

>1.A good test for a hierarchical model building is to be able to take a model and another model and combine them into one model (maybe sharing ground only). Is that possible after this PR with real models?

I agree and it is definitely a capability I want to work in OpenSim. The Component interface supports this, see [here](https://github.com/opensim-org/opensim-core/pull/740/files#diff-074825ae2a365226376a3224512e6329R533). But,with current models it will not work. One issue there is that model owns and expects to own the system. Multiple Ground components should be OK since they all would reference the same multibody system Ground and it would allow each model to have its own Frames (e.g. offsets) and Geometry associated with Ground that the other or an Uber model would have to know. For the time being it is a relatively low priority but we are making steady progress in that direction.

>2."Adding a component causes transfer of ownership", did anything change in the interface for how it's done. In the past we had to tell the scripting/SWIG about that so memory management doesn't blow up for scripting. This could be unrelated to this PR, but needs handling in SWIG. A related issue is handling stack allocated objects and adding them to the model.

No change here whatsover. The same issues remain and warrant a separate discussion.

>3.Does this PR handle or do the plumbing to allow for "internal" unexposed components. I'm interested because the "frame-geometry" can possibly use this functionality if available (a separate PR of course).

Yes and no. If you have an new member that is a Component and use the new constructor to initialize it like [this](https://github.com/opensim-org/opensim-core/pull/740/files#diff-074825ae2a365226376a3224512e6329R118) then you are done. It will be a subcomponent and it will have Component calls propagated to it. If you need to create Components on the fly due what happens after finalizeFromProperties (e.g realize you need to break a Loop in the case of Model) then you still have to manually markAsSubcomponent(). My sense is that this will be so rarely used as to not warrant the effort. You could easily manage it for Frame's builtin FrameGeometry to keep it a private component of Frame.

>4.Recomputing the tree-traversal on every call to getComponentList is expensive and a bit excessive especially that the tree doesn't change at all post initialization. A caching mechanism can be implemented to improve performance but I still think this can be optimized if you know exactly when components are added to the tree, so there's no cost post model building.

Absolutely! That is why I listed it in Hack/Bugs. It is not intended to stay as is. It was shortest path to show that the interface works. We can correct and optimize the backend as necessary. I am not as concerned about the inefficiency as with the guarantee of correctness with other moving parts. For all its flaws it at least guarantees that the list is up-to-date and there is no order dependency, flags, or side-effects to worry about! 

>5.Everything here works with Components while models have been made of ModelComponents, is there a long term future for this distinction?

For the time being you can use them synonymously. The distinction will be more clear when there are Components that are not ModelComponents. That is they are purely calculating Components (still need system resources). These would be things like Filters, Signal generators, Multiplexers  and mathematical operators.

> 6.The ownership paradigm implies no sharing/re-reuse, what happens when a user tries to add a component already owned? Does the user need to "new" components to add them to the model? does the ownership switch again?

Same as 2. it behaves as it does now, unfortunately- failing during destruction when the second Set tries to delete its objects. This is a problem and it is not being addressed here. It could be remedied by using a data structure of smart pointers (instead of or inside of Sets) or use handles - but that is for another discussion/fix/PR.

OpenSim/Common/Component.cpp
```diff
@@ -838,26 +836,74 @@ void Component::constructOutputForStateVariable(const std::string& name)
...
+                    // otherwise it may be a Set (of objects), and
+                    // would prefer to do something like this to test:
+                    //  Set<Component>* objects = dynamic_cast<Set<Component>*>(obj)
+                    // Instead we can see if the object has a property called
+                    // "objects" which is a PropertyObjArray used by Set<T>.
+                    // knowing the object Type is useful for debugging
+                    // and it could be used to strengthen the test (e.g. scan  
+                    // for "Set" in the type name). 
+                    std::string objType = obj.getConcreteClassName();
+                    if (obj.hasProperty("objects")) {
```
sherm1: Will this still be necessary once the old property system is gone? If not this should be marked somehow so we'll know to remove it then.

OpenSim/Common/Component.cpp
```diff
@@ -838,26 +836,74 @@ void Component::constructOutputForStateVariable(const std::string& name)
...
+                } // end of if/else property value is an Object or something else
+            } // loop over the property list
+        } // end if property is an Object
+    } // loop over properties
+}
+
 // Include another Component as a subcomponent of this one. If already a
 // subcomponent, it is not added to the list again.
-void Component::addComponent(Component* component)
+void Component::markAsSubcomponent(Component* component)
```
sherm1: The name of this method and the fact that `component` is passed in writable makes me think that it is `component` that will be modified. Could this method be something like `recordInSubcomponentList()`?

Also it would be good to use the term `subcomponent` whenever it can help distinguish parent from child, so I think the argument here should be `Component& subcomponent`.

OpenSim/Common/Component.cpp
```diff
@@ -838,26 +836,74 @@ void Component::constructOutputForStateVariable(const std::string& name)
...
-    // Only add if the Component is not already a part of the model
+    // Only add if the component is not already a part of this Component
     // So, add if empty
     if ( _components.empty() ){
         _components.push_back(component);
     }
-    else{ //otherwise check that it isn't apart of the component already        
+    else{ //otherwise check that it isn't a part of the component already
         SimTK::Array_<Component *>::iterator it =
             std::find(_components.begin(), _components.end(), component);
```
sherm1: This is implemented with an O(n^2) algorithm which could be very slow if many subcomponents were added to a particular component (because the list being searched keeps growing). I assume you aren't expecting long lists here so this is probably fine, but there should be some hard-to-miss warning comments here pointing out the potential for disaster with this algorithm.

OpenSim/Simulation/Model/PhysicalFrame.cpp
```diff
@@ -45,16 +45,27 @@ PhysicalFrame::PhysicalFrame() : Frame()
...
-    frm.set_display_radius(.004);
-    frm.upd_Appearance().set_visible(false);
-    append_geometry(frm);
 }
 
 void PhysicalFrame::constructProperties()
 {
     constructProperty_WrapObjectSet(WrapObjectSet());
+
+    updProperty_geometry().setAllowableListSize(1, 1000);
```
sherm1: Why does this specify a list size?

OpenSim/Simulation/Model/PhysicalFrame.cpp
```diff
@@ -45,16 +45,27 @@ PhysicalFrame::PhysicalFrame() : Frame()
...
+
+    FrameGeometry frm;
+    frm.setName("frame_geometry");
+    frm.upd_Appearance().set_visible(false);
+    updProperty_geometry().setValue(0, frm);
+    //updProperty_geometry().setValueIsDefault(true);
```
sherm1: delete this commented-out line?

OpenSim/Simulation/Model/PhysicalFrame.cpp
```diff
@@ -128,3 +135,203 @@ void PhysicalFrame::scale(const SimTK::Vec3& aScaleFactors)
...
+    SimTK::Xml::Element translationNode("translation", ss.str());
+    ss.clear();
+    ss << localXform[0] << " " << localXform[1] << " " << localXform[2];
+    SimTK::Xml::Element orientationNode("rotation", ss.str());
+    frameNode.insertNodeAfter(frameNode.element_end(), translationNode);
+    frameNode.insertNodeAfter(frameNode.element_end(), orientationNode);
+    frameSetIter->insertNodeAfter(frameSetIter->element_end(), frameNode);
+    XMLDocument::addConnector(frameNode, "Connector_PhysicalFrame_", "parent", bodyName);
+
+}
```
sherm1: missing newline at eof

sherm1:
Ajay, I feel that I was only able to provide a shallow review; I hope that is still useful. I would probably need an in-person-with-whiteboard explanation in order to comment usefully on the overall design. Also with 49 files modified it was hard to tell where to focus; please let me know if there is somewhere I should look that I didn't review.

OpenSim/Common/Component.h
sherm1: > Since the in-class member initializer is called during construction of the parent (owner), is it permissible to call a method on the parent as a side effect of the in-class member initialization?

Oh. Cheeky. Probably not. In that case the address of the parent is probably the only thing you can count on. But then I think this constructor needs a new comment explaining what it is for and how to use it. And that second argument should definitely *not* be optional!

sherm1:
Discussed with @aseth1. We decided that the best approach would be: merge this big PR (after whatever local changes Ajay wants to make); make an issue with a checklist for remaining changes; then do a series of small, focused, easier-to-review PRs with remaining changes.

chrisdembia:
It seems preferable to reverse that process: first make small PRs that are necessary for an upcoming feature, then make a PR for the feature. Is that possible here?

aseth1:
@chrisdembia that is the ideal but it isn't always possible. For example, a small change in the call sequence of finalizeFromProperties caused many fixes. I feel this is a relatively small/simple PR with lots of changes and testing to accommodate it which accounts for the bulk of the files involved. Now that the bulk change is done I can focus more on the correct implementation of the iterator and it won't get buried.

sherm1:
> first make small PRs that are necessary for an upcoming feature, then make a PR for the feature. Is that possible here?

We discussed that. That is the ideal process, but very difficult to accomplish in practice. I am having the same problem with state serialization: I started out to add the new feature and while developing it I ran into a bunch of unanticipated subproblems requiring a bunch of unimportant changes to a lot of files. Some of those subproblems were hard and made me rethink things so I could undo them. By the time I had anything ready to push, I had at least 50 changed files and the PR was going to be hard to review. I started untangling some of the final set of subproblems so I could make small PRs, which makes it look as though I cleverly thought of those things first, but that was not the case. It just happened that there wasn't a lot of file overlap for the subproblems so it was feasible to untangle them afterwards. It was still a lot of work though, and Ajay's subproblems aren't as easily decoupled.

chrisdembia:
Thanks for explaining.

FYI A good git tool/interface should allow you to easily stage only parts of an individual file. My git tool allows me to stage individual lines of files, which makes it easier to separate out distinct features/changes.

aymanhab:
@aseth1 and @sherm1 Is there a specific reason to try to merge this as is to master vs. create a separate feature branch, fix everything on it and then merge together? Is anybody else on depending on these changes being on master? I sure like the cleanup and better definition of the MCI methods but since future PRs are tightly coupled, it may be easier to review on a branch. Thoughts?

aseth1:
@ayman since all tests pass and tests were strengthened it is worth merging. I believe creating a feature branch will delay and make merging more difficult.

aseth1:
@chrisdembia - that sounds cool! Mind showing me how you manage staging individual lines into different PRs? That would be super useful for sure.

aseth1:
@chrisdembia @aymanhab @sherm1 I would like to reiterate that the scope of this PR was to get Components that are properties to be automatically treated as subcomponents and to traverse them immediately, which it does.  Both issues of tree traversal and managing "newed" subcomponents were also flagged as hacks/bugs and I intended them to be addressed next.  I opened issue (#748) to capture these explicitly. Please feel free add to comments/concerns there.

I am still addressing other comments so please go ahead and add more if you wish (except add those related to #748 there, please). Thanks!

aymanhab:
@aseth the part I'm having hard time accepting (as we all agreed is a hack) is the change to getComponentList because a "get" method have the connotation that it's fast and not destructive, which the old method was (I'm using it inside my live animation loop so performance is critical) and the new method isn't. if you split it into two a "get" and "upd/construct/..." where the former keeps the old behavior and the latter is used during model building I'd be fine. It doesn't seem like sacrificing runtime performance for model building elegance is a required compromise.

sherm1:
> My git tool allows me to stage individual lines of files, which makes it easier to separate out distinct features/changes.

TortoiseGit claims to be able to do this. See:
> 2.5.2. Commit only parts of files

In TortoiseGit help.

I'm not sure how easy it would be to use this in practice because it seems like it would be hard to find all the right pieces that would allow the selected changes to build properly (like I'd forget that I had to add `#include <algorithm>` at the top of a file or whatever).

sherm1:
> @aseth the part I'm having hard time accepting (as we all agreed is a hack) is the change to getComponentList because a "get" method have the connotation that it's fast and not destructive

Ayman, I was thinking of this in terms of how best to use GitHub to get the important code reviewed in digestible chunks. Leaving giant PRs around like this one makes the whole process very awkward. There needs to be code for correctly obtaining all the components *prior* to any hope for caching the resulting list. Both pieces ought to get a good review, but if that involves people looking repeatedly at a 50+ file PR with mostly uninteresting changes I am skeptical that the important code will get the reviewing it deserves.

Merging into a branch other than master (as you suggested) could work though if merging to master messes up someone else's work.

aymanhab:
@aseth1 I agree overall, that's why I'm not addressing any details but I'm not sure if that excludes making the specific relatively small change I suggested to isolate what we agreed is a hack and avoid slowing down the code that's used in the GUI/animation loop?

chrisdembia:
My opinion is to just create a feature branch to hold off these changes from master until the performance concerns are addressed. I think this would satisfy everyone, especially since I believe feature branches are easy to create and use.

sherm1:
> My opinion is to just create a feature branch to hold off these changes from master until the performance concerns are addressed.

We should watch out for premature optimization here! My understanding of the algorithm is that the uncached version should be O(n log d) where n is the total number of components and d the tree depth. But d is likely to be very small and essentially constant because most components are owned directly by Model. The cached algorithm is O(n) but involves lots of bug-prone update/invalidate code. It could turn out that a straightforward iterator-based walking of the subcomponent tree could be both reliable and fast.

chrisdembia:
Good point. @aseth1 what is the difference in runtime for testIterators before and after your changes?

aymanhab:
@sherm1 Good point but I'd also keep in mind that the claim that most components are owned by the model though true today will start breaking soon, and especially if we create hierarchical models as we think will happen in the future.

sherm1:
> keep in mind that the claim that most components are owned by the model though true today will start breaking soon

Agreed. I wouldn't argue against *eventual* optimization; just that on general principles we shouldn't optimize until we know there is a reason to do so!

aseth1:
Appveyor
Recent:
build (2141)
44/68 Test #47: testIterators ................................   Passed    2.14 sec
 build (2166)
43/68 Test #47: testIterators ................................   Passed    1.20 sec

This branch (2125):
45/68 Test #47: testIterators ................................   Passed    1.59 sec

Gcc Travis:
Previous (2837.2)
38/66 Test #47: testIterators ................................   Passed    0.32 sec
Recent (2873.2)
37/66 Test #47: testIterators ................................   Passed    1.82 sec 

This branch (2840.2)
37/66 Test #47: testIterators ................................   Passed    0.51 sec

My machine:
(master): 48/66 Test #47: testIterators ................................   Passed    0.37 sec
(branch): 5/66 Test #47: testIterators ................................   Passed    0.14 sec
(No these are not reversed.)

There seems to be little indication that change slows down performance but there is a lot of variability.

aymanhab:
Sounds good @aseth, though it seem suspect that recomputing anything is faster than reusing it but you may have made improvements somewhere else. Feel free to merge to move along if others are ok.

testIterators cycles thru the list of components in Arm26 so the numbers are small anyway but it's good that the numbers don't look that different (though an estimate of complexity would scale better to large models).

aseth1:
@sherm1 I addressed some of your earlier comments:
1. used `default` 
2. removed Constructor that required name and owner (turned out I could remove it)
3. added `constructComponent<C>(name)` initializer automatically sets `this` to be its owner and marks it at a subcomponent.
4. added `Component::adoptSubcomponent(Component *) and corresponding list of _adoptees to handle internal components created due to do optional settings are based on the validity of the properties. E.g. model breaks a loop of joints and adds a constraint, which is now adopted.
5. `markAsSubcomponent()` is not intended for external use and is now private and all reference to it in doxygen has been removed.
6. revised Component comments accordingly

OpenSim/Common/Component.h
```diff
@@ -946,9 +976,16 @@ template <class T> friend class ComponentMeasure;
     with the line "Super::extend<xxx>(args);" to ensure that the parent class
     is called before the child class method.
     
-    The base class implementations ensures that the 
-    corresponding calls are made to any subcomponents that have been specified 
-    by derived %Component objects, via calls to the addComponent() method. 
+    The base class implementations ensures that the corresponding calls are made
+    to any subcomponents which are owned by this Component. Ownership is
+    established by the subcomponent being a data member (not serialized), a 
+    property (serialized), of created and adopted based on other settings
```
sherm1: of -> or

sherm1:
> @sherm1 I addressed some of your earlier comments

Thanks, Ajay. Looks good to me. I vote for merging and moving on!

jenhicks:
Fixes #119

aseth1:
With @sherm1 and @aymanhab OK to merge I will do that now.

PR #747, opened 2016-01-24
[travis] libblas, liblapack

chrisdembia:
This change is to accomodate Simbody's move to 14.04 with travis. Right now, all travis builds are failing. This change is somewhat temporary, and is just to get the builds to work again.

This change is also somewhat separate from the upgrade in compiler versions.

Another way to fix this problem would be for OpenSim to use 14.04 with travis. However, we can't use caching with 14.04, and caching (ccache) speeds up our travis builds.

Appveyor failed b/c I cancelled it.


sherm1:
Looks good -- mergable.

PR #750, opened 2016-01-27
[travis] SUPERBUILD_SIMBODY -> SUPERBUILD_simbody

chrisdembia:
CMake variables are case sensitive. SUPERBUILD_SIMBODY was being ignored.


chrisdembia:
The appveyor builds failed because I cancelled them.

PR #754, opened 2016-02-05
[cmake] Remove link_directories() for Simbody.

chrisdembia:
No need to call `link_directories()` for Simbody anymore (see https://github.com/simbody/simbody/pull/449) . Thanks @elen4 for pointing out that this call is deprecated.


chrisdembia:
@klshrinidhi could you review/merge?

PR #757, opened 2016-02-08
Installation instructions for impatient

klshrinidhi:
Update readme to include build from source instructions for users who are familiar with their development environment and want to get started quickly.

Instructions may need updating for Python/Matlab bindings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -717,22 +717,24 @@ Your changes will only take effect in new terminal windows.
 
 #### For the impatient (Ubuntu)
 
-* In **Terminal** --
- * `sudo add-apt-repository ppa:george-edison55/cmake-3.x`
- * `sudo apt-add-repository ppa:fenics-packages/fenics-exp`
- * `sudo apt-get update`
- * `sudo apt-get install git cmake cmake-curses-gui clang-3.6 freeglut3-dev libxi-dev libxmu-dev liblapack-dev swig3.0 python-dev openjdk-7-jdk`
- * `sudo rm -f /usr/bin/cc /usr/bin/c++`
- * `sudo ln -s /usr/bin/clang-3.6 /usr/bin/cc`
- * `sudo ln -s /usr/bin/clang++-3.6 /usr/bin/c++`
- * `git clone https://github.com/opensim-org/opensim-core.git`
- * `mkdir opensim_dependencies_build`
- * `cd opensim_dependencies_build`
- * `cmake ../opensim-core/dependencies/ -DCMAKE_INSTALL_PREFIX='~/opensim_dependencies_install' -DCMAKE_BUILD_TYPE=RelWithDebInfo`
- * `make -j8`
- * `cd ..`
- * `mkdir opensim_build`
- * `cd opensim_build`
- * `cmake ../opensim-core -DCMAKE_INSTALL_PREFIX="~/opensim_install" -DCMAKE_BUILD_TYPE=RelWithDebInfo -DOPENSIM_DEPENDENCIES_DIR="~/opensim_dependencies_install" -DBUILD_PYTHON_WRAPPING=ON -DBUILD_JAVA_WRAPPING=ON`
- * `make -j8`
- * `ctest -j8`
+In **Terminal** --
+```shell
+ sudo add-apt-repository ppa:george-edison55/cmake-3.x
+ sudo apt-add-repository ppa:fenics-packages/fenics-exp
+ sudo apt-get update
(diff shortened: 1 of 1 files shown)


chrisdembia:
I like this a lot. My main feedback is: how about making just one huge code block for each of the platforms, instead of individual bullets? Then people could easily copy and paste multiple lines into a text file or into the terminal.

README.md
```diff
@@ -675,3 +714,24 @@ Your changes will only take effect in new terminal windows.
...
+
+#### For the impatient (Ubuntu)
+
+* In **Terminal** --
+ * `sudo add-apt-repository ppa:george-edison55/cmake-3.x`
+ * `sudo apt-get update`
+ * `sudo apt-get install git cmake cmake-curses-gui clang-3.6 freeglut3-dev libxi-dev libxmu-dev liblapack-dev`
+ * `sudo rm -f /usr/bin/cc /usr/bin/c++`
+ * `sudo ln -s /usr/bin/clang-3.6 /usr/bin/cc`
+ * `sudo ln -s /usr/bin/clang++-3.6 /usr/bin/c++`
```
chrisdembia: I don't think we should suggest that people should alter their system in this way. There should be a way to do this using `sudo update-alternatives --config c++` and `sudo update-alternatives --config cc`.

README.md
```diff
@@ -675,3 +714,24 @@ Your changes will only take effect in new terminal windows.
 [running_gif]: doc/images/opensim_running.gif
 [simple_example_gif]: doc/images/opensim_double_pendulum_muscle.gif
 [java]: http://www.oracle.com/technetwork/java/javasebusiness/downloads/java-archive-downloads-javase6-419409.html
+
+#### For the impatient (Ubuntu)
```
chrisdembia: Could you list which version of Ubuntu these instructions are for? On more recent versions of Ubuntu, users probably don't need to add a ppa for cmake.

README.md
chrisdembia: Another alternative is to define `CC=clang-3.5` and `CXX=clang++-3.6`. I actually think this is the best option, since the user might also have software elsewhere on their computer that they want to compile with gcc, etc.

README.md
```diff
@@ -675,3 +714,37 @@ Your changes will only take effect in new terminal windows.
...
+ sudo add-apt-repository ppa:george-edison55/cmake-3.x
+ sudo apt-add-repository ppa:fenics-packages/fenics-exp
+ sudo apt-get update
+ sudo apt-get install git cmake cmake-curses-gui clang-3.6 \
+                      freeglut3-dev libxi-dev libxmu-dev \
+                      liblapack-dev swig3.0 python-dev openjdk-7-jdk
+ sudo rm -f /usr/bin/cc /usr/bin/c++
+ sudo ln -s /usr/bin/clang-3.6 /usr/bin/cc
+ sudo ln -s /usr/bin/clang++-3.6 /usr/bin/c++
+ export JAVA_HOME=/usr/lib/jvm/java-7-openjdk-amd64
```
chrisdembia: You might also be able to set a CMake variable (`-DJAVA_HOME`), based on the documentation here: https://cmake.org/cmake/help/v3.0/module/FindJava.html.

klshrinidhi:
Updated instructions for following OSes:
* Ubuntu 14.04 LTS
* Ubuntu 15.10
* OS X 10.10
* OS X 10.11

Looks merge-able now ?

README.md
```diff
@@ -288,6 +288,28 @@ directory to your `PATH` environment variable.
...
+ * `mkdir opensim_dependencies_build`
+ * `cd .\opensim_dependencies_build`
+ * `cmake ..\opensim-core\dependencies -G"Visual Studio 14 2015 Win64" -DCMAKE_INSTALL_PREFIX="..\opensim_dependencies_install"`
+ * `cmake --build . --config RelWithDebInfo -- /maxcpucount:8`
+ * `cd ..`
+ * `mkdir opensim_build`
+ * `cd .\opensim_build`
+ * `cmake ..\opensim-core -G"Visual Studio 14 2015 Win64" -DCMAKE_INSTALL_PREFIX="..\opensim_install" -DOPENSIM_DEPENDENCIES_DIR="..\opensim_dependencies_install"`
+ * `cmake --build . --config RelWithDebInfo -- /maxcpucount:8`
+ * `ctest -C RelWithDebInfo --parallel 8`
```
chrisdembia: I still prefer that this appears as one code block. Markdown should support indented multi-line code blocks as part of a bulleted list. Anyway, I'll merge this as-is; this is good enough.

PR #767, opened 2016-02-16
Fix link to BTK in README.md

klshrinidhi:
The link previously pointed to original BTK but needs to point to my fork because the author has not merged my pull request yet.


chrisdembia:
That link appears in two other places as well (for each platform).

klshrinidhi:
@chrisdembia 
Thanks. I fixed the other two links.

PR #768, opened 2016-02-16
Don't build examples and testing for simbody during super-build.

klshrinidhi:
To save time, turn OFF building examples and tests for simbody when built as a dependency during super-build.

### DIFF
--- a/dependencies/CMakeLists.txt
+++ b/dependencies/CMakeLists.txt
@@ -119,9 +119,11 @@ AddDependency(NAME       BTK
               CMAKE_ARGS -DBUILD_SHARED_LIBS:BOOL=ON)
 
 
-AddDependency(NAME simbody
-              URL  https://github.com/simbody/simbody.git
-              TAG  ed67885db8983405e0724e63e62bf8e9316b2493)
+AddDependency(NAME       simbody
+              URL        https://github.com/simbody/simbody.git
+              TAG        ed67885db8983405e0724e63e62bf8e9316b2493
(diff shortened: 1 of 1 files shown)


klshrinidhi:
Additional note -- BTK configuration defaults to not building examples and tests.

dependencies/CMakeLists.txt
```diff
@@ -119,9 +119,11 @@ AddDependency(NAME       BTK
...
 
 
-AddDependency(NAME simbody
-              URL  https://github.com/simbody/simbody.git
-              TAG  ed67885db8983405e0724e63e62bf8e9316b2493)
+AddDependency(NAME       simbody
+              URL        https://github.com/simbody/simbody.git
+              TAG        ed67885db8983405e0724e63e62bf8e9316b2493
+              CMAKE_ARGS -DBUILD_EXAMPLES:BOOL=OFF
+              CMAKE_ARGS -DBUILD_TESTING:BOOL=OFF)
```
chrisdembia: You don't have to use multi-value args to get this to work? See here for example: https://github.com/opensim-org/opensim-core/blob/master/cmake/OpenSimMacros.cmake#L39

klshrinidhi: You are right. I just pushed the bug fix.

klshrinidhi: I verified that Simbody is not building examples and tests.

chrisdembia:
Are there any issues that can be closed now?

PR #771, opened 2016-02-18
Enable bindings for TimeSeriesTable

klshrinidhi:
Add TimeSeriesTable header to bindings.


chrisdembia:
+1

aymanhab:
Would be great to visually inspect the output and see if the correct methods/signatures were produced and that the resulting class(es) are usable. Maybe add a test case to make sure what works now doesn't break later.

chrisdembia:
@klshrinidhi FYI the swig output is easier to visually inspect with via java rather than swig, b/c there is a separate .java file for each class.

klshrinidhi:
I have wrapped all classes necessary to wrap `DataTable` and `TimeSeriesTable` except for `ValueArrayDictionary`. Wrapping `ValueArrayDictionary` requires wrapping `SimTK::AbstractValue` and `SimTK::Value` classes, which in turn require enumerating all the types that going to be *type-erased*. This needs some discussion in my opinion.

chrisdembia:
Cool. I think it would be good to add at least some basic python tests. Just creating a table, adding some rows, and getting some values out. I'm happy to discuss.

chrisdembia:
> This needs some discussion in my opinion.

I agree.

jenhicks:
It's ok not to wrap the metadata (at least for the hackathon). Key for hackathon is the table and column labels. Should try one of the examples to see if it works.

klshrinidhi:
Wrapping for `DataTable` and `TimeSeriesTable` is functional now.
See [Bindings/Python/tests/test_DataTable.py](https://github.com/opensim-org/opensim-core/blob/datatable_wrapping/Bindings/Python/tests/test_DataTable.py) for usage in Python.

Should be merge-able.

Bindings/Python/tests/test_DataTable.py
```diff
@@ -0,0 +1,177 @@
...
+        assert (str(row1[0]) == str(row[0]) and
+                str(row1[1]) == str(row[1]) and
+                str(row1[2]) == str(row[2]))
+        # Append another row to the table with a timestamp
+        # less than the previous one. Exception expected.
+        try:
+            table.appendRow(0.15, row)
+            assert False
+        except RuntimeError:
+            pass
```
chrisdembia: No need to change this, but FYI there is a `self.assertRaises()` that you could use.

chrisdembia:
The python test is awesome. By the way there is a failure:

```
======================================================================
ERROR: test_TimeSeriesTableVec3 (test_DataTable.TestDataTable)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/opensim-core/share/doc/OpenSim/python/opensim/tests/test_DataTable.py", line 145, in test_TimeSeriesTableVec3
    table = osim.TimeSeriesTableVec3()
AttributeError: 'module' object has no attribute 'TimeSeriesTableVec3'
----------------------------------------------------------------------
```

klshrinidhi:
All :white_check_mark: !!!! :smile: 
Can we merge ?

PR #774, opened 2016-02-18
StatesTrajectory methods to access states by time.

chrisdembia:
These changes were originally part of #730, but I thought it'd be easier to review these changes in a separate PR. These changes can be cleanly separated from the rest of #730. Ready for review.

### DIFF
--- a/OpenSim/Simulation/StatesTrajectory.h
+++ b/OpenSim/Simulation/StatesTrajectory.h
@@ -156,8 +156,8 @@ class Model;
  * before `timeB`:
  * @code{.cpp}
  * for (const auto& state : SimTK::makeIteratorRange(
- *                              states.getIteratorAt(timeA),
- *                              states.getIteratorBefore(timeB)) {
+ *                              states.findAt(timeA),
+ *                              states.findNearestBefore(timeB)) {
  *     ...
  * }
  * @endcode
@@ -229,7 +229,8 @@ public:
 
     /// @name Accessing states by time
     /// @{
-    /** Get the index of the state just before (or at) the given time.
+    /** Get the index of the nearest state before the given time
+     * (possibly *at* the given time).
      *
      * The tolerance exists to handle imprecise
      * times. For example, if the trajectory contains states with times 1.3
@@ -245,58 +246,64 @@ public:
      * the console as 1.500; you might be tempted to get this state by asking
      * for the one that is before/at time 1.5, but this would end up giving you
      * the state at time 1.3. You could remedy this issue by setting the
-     * tolerance to 1e-3 (the precision of the console output).
+     * tolerance to 1e-3 (the precision of the console output). So, this
+     * function actually gives the nearest state before `time + tolerance`.
      *
      * For a trajectory with times 0.2, 0.5, 0.7, this code will get the state
      * at time 0.5:
      * @code
-     * const auto& state = states[states.getIndexBefore(0.6)];
+     * const auto& state = states[states.findIndexNearestBefore(0.6)];
      * @endcode
      *
      * @throws TimeOutOfRange If there are no states before the given time.
      * @throws Exception If the trajectory is empty.
      */
-    size_t getIndexBefore(const double& time,
-                          const double& tolerance = 0) const;
-    /** Get the index of the state just after (or at) the given time.
+    size_t findIndexNearestBefore(const double& time,
+                                  const double& tolerance = 0) const;
+    /** Get the index of the nearest state after the given time
+     * (possibly *at* the given time).
      *
-     * See getIndexBefore() for an explanation of the tolerance.
+     * See findIndexNearestBefore() for an explanation of the tolerance; this
+     * function gives the nearest state after `time - tolerance`.
      *
      * For a trajectory with times 0.2, 0.5, 0.7, this code will get the state
      * at time 0.5:
      * @code
-     * const auto& state = states[states.getIndexAfter(0.4)];
+     * const auto& state = states[states.findIndexNearestAfter(0.4)];
      * @endcode
      *
      * @throws TimeOutOfRange If there are no states after the given time.
      * @throws Exception If the trajectory is empty.
-     * */
-    size_t getIndexAfter(const double& time,
-                         const double& tolerance = 0) const;
+     */
+    size_t findIndexNearestAfter(const double& time,
+                                 const double& tolerance = 0) const;
 
-    /** Get the index of the state closest to the given time, within a given
-     * tolerance. If there are multiple states exactly at the given time, the
+    /** Get the index of the state closest to the given time within the
+     * interval [time - tolerance, time + tolerance].
+     * If there are multiple states exactly at the given time, the
      * index is that of the first of these states.
      *
-     * The tolerance should be set to the precision of the time where you
-     * observed it (perhaps in console output or storage files).
+     * This tolerance is *different* from the one used in
+     * findIndexNearestBefore(), but should still be set to the precision of
+     * the desired time where you observed it (perhaps in console output or
+     * storage files).
      *
-     * For a trajectory with times 0.2, 0.501, 0.7, this code gives the index for
-     * the state at time 0.5:
+     * For a trajectory with times 0.2, 0.501, 0.7, this code gives the index
+     * for the state at time 0.5:
      * @code
-     * states.getIndexAt(0.5, 0.01);
+     * states.findIndexAt(0.5, 0.01);
      * @endcode
      *
      * @throws NoStateAtTime If there is no state at the given time.
      * @throws Exception If the trajectory is empty.
      */
-    size_t getIndexAt(const double& time, const double& tolerance = 0) const;
+    size_t findIndexAt(const double& time, const double& tolerance = 0) const;
     /** Iterate over the states in a given time interval (inclusive).
      *
      * For a trajectory with times 0.2, 0.5, 0.7, 1.5, 1.8, this code will
      * iterate through the states at times 0.5, 0.7, 1.5:
      * @code
-     * for (const auto& state : states.getBetween(0.5, 1.5)) {
+     * for (const auto& state : states.findBetween(0.5, 1.5)) {
      *     state.getTime();
      * }
      * @endcode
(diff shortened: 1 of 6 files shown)


OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -221,6 +230,164 @@ class OSIMSIMULATION_API StatesTrajectory {
...
+    /// @{
+    /** Get the index of the nearest state before the given time
+     * (possibly *at* the given time).
+     *
+     * The tolerance exists to handle imprecise
+     * times. For example, if the trajectory contains states with times 1.3
+     * and 1.5 and you ask for the index of the state before time 1.4, you'll
+     * normally get the state with time 1.3. However, if you set a tolerance of
+     * 0.15, then you'll get the state with time 1.5 (because it is the state
+     * just before 1.55 = 1.4 + 0.15).
```
aseth1: This seems wrong or at least and awkward use case. First, using a tolerance of 0.15 should not be advocated - its purpose is to deal with imprecision not to expand the range.   Second the nearest **before** should still be enforced such that if 1.3 and 1.5 are equidistant fro 1.4 then 1.3 should prevail. In the case that 1.4 is 1.400000001 then the tolerance should reflect expected/typical truncation error. I don't  think the tolerance should default to 0 because users will not know what a reasonable tolerance should be and are going to be. We should use SimTK::Eps or the like as the default tolerance.

OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -221,6 +230,164 @@ class OSIMSIMULATION_API StatesTrajectory {
...
+     *
+     * @throws TimeOutOfRange If there are no states after the given time.
+     * @throws Exception If the trajectory is empty.
+     */
+    size_t findIndexNearestAfter(const double& time,
+                                 const double& tolerance = 0) const;
+
+    /** Get the index of the state closest to the given time within the
+     * interval [time - tolerance, time + tolerance].
+     * If there are multiple states exactly at the given time, the
```
aseth1: "multiple states exactly at the given time" should not be possible IMHO

OpenSim/Simulation/Test/testStatesTrajectory.cpp
```diff
@@ -508,6 +508,386 @@ void testCopying() {
...
+        states.append(state); // 5
+        state.setTime(8.0);
+        states.append(state); // 6
+
+        // If you ask for a time at which there are multiple states (with
+        // the same exact time), you get the first of these states.
+        SimTK_TEST(states.findIndexAt(7.5) == 1);
+        SimTK_TEST(states.findIndexAt(7.9) == 4);
+        SimTK_TEST(states.findIndexAt(7.4, 0.1) == 1);
+        SimTK_TEST(states.findIndexAt(7.56, 0.06) == 3);
```
aseth1: Not sure what you mean by:

> you get the first of these states
index 3 is the third state with time 7.5 not the first (1)

OpenSim/Simulation/Test/testStatesTrajectory.cpp
```diff
@@ -508,6 +508,386 @@ void testCopying() {
...
+        // Both the 8.04 and 8.05 states are within the tolerance for the
+        // requested times below, but we expect to get the closer of the two.
+        SimTK_TEST(states.findIndexAt(8.044, 0.01) == 1);
+        SimTK_TEST(states.findIndexAt(8.046, 0.01) == 2);
+
+        SimTK_TEST(states.findIndexAt(8.04499999999, 0.01) == 1);
+        SimTK_TEST(states.findIndexAt(8.04500000001, 0.01) == 2);
+
+        // It actually doesn't matter what the tolerance is too much; the
+        // tolerance can even extend beyond the range of times in trajectory.
```
aseth1: The tolerance matters a lot in defining what is an acceptable match. findIndexAt(8.02, 1.0e-5) will throw an Exception that there is no state at time 8.02. but if I use a tolerance of 1e-2. Using large tolerances is not the intended use case IMO.

OpenSim/Simulation/Test/testStatesTrajectory.cpp
```diff
@@ -508,6 +508,386 @@ void testCopying() {
...
+        state.setTime(8.06);
+        states.append(state); // 3
+
+        // Both the 8.04 and 8.05 states are within the tolerance for the
+        // requested times below, but we expect to get the closer of the two.
+        SimTK_TEST(states.findIndexAt(8.044, 0.01) == 1);
+        SimTK_TEST(states.findIndexAt(8.046, 0.01) == 2);
+
+        SimTK_TEST(states.findIndexAt(8.04499999999, 0.01) == 1);
+        SimTK_TEST(states.findIndexAt(8.04500000001, 0.01) == 2);
```
aseth1: I don't see how these two conditions is any different from the previous two.

OpenSim/Simulation/Test/testStatesTrajectory.cpp
```diff
@@ -508,6 +508,386 @@ void testCopying() {
...
+    // Make sure we get the right state when there are multiple within
+    // the tolerance.
+    {
+        // Assume the user views the times with only two decimal places (say
+        // in a console or storage file); they'll see:
+        //      0.50, 1.30, 3.50, 3.50, 3.50, 4.10.
+        // Therefore, the appropriate tolerance
+        // is 0.01. If they ask for the nearest state before 3.50, they should
+        // get the state that is actually at 3.502, since, from the
+        // console/file, this appears to be the first state "before" 3.50.
```
aseth1: I don't understand why 3.502 is the nearest to 3.499, when 3.500 is closer.

OpenSim/Simulation/Test/testStatesTrajectory.cpp
```diff
@@ -508,6 +508,386 @@ void testCopying() {
...
+            StatesTrajectory::TimeOutOfRange);
+    SimTK_TEST_MUST_THROW_EXC(states.findIndexNearestBefore(.499999),
+            StatesTrajectory::TimeOutOfRange);
+
+    SimTK_TEST_MUST_THROW_EXC(states.findIndexNearestAfter(4.10000001),
+            StatesTrajectory::TimeOutOfRange);
+    SimTK_TEST_MUST_THROW_EXC(states.findIndexNearestAfter(1050.9),
+            StatesTrajectory::TimeOutOfRange);
+
+    // Simulate the scenario that the tolerance is intended to handle.
```
aseth1: Why is this block here? Why interject states2 at this point when the test case resumes with states? It seems to unnaturally interrupt the flow.

OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -221,6 +230,164 @@ class OSIMSIMULATION_API StatesTrajectory {
...
+     *     state.getTime();
+     * }
+     * @endcode
+     *
+     * The startTime must be less than or equal to the endTime, otherwise an
+     * exception is thrown.
+     *
+     * See findIndexNearestBefore() for an explanation of the tolerance.
+     */
+    IteratorRange findBetween(const double& startTime, const double& endTime,
```
aseth1: is this interpreted as startTime-tolerance <= x <= endTime+tolerance?

OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -221,6 +230,164 @@ class OSIMSIMULATION_API StatesTrajectory {
...
+     * If there are multiple states exactly at the given time, the
+     * index is that of the first of these states.
+     *
+     * This tolerance is *different* from the one used in
+     * findIndexNearestBefore(), but should still be set to the precision of
+     * the desired time where you observed it (perhaps in console output or
+     * storage files).
+     *
+     * For a trajectory with times 0.2, 0.501, 0.7, this code gives the index
+     * for the state at time 0.5:
```
aseth1: at time 0.501

OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -155,6 +155,16 @@ class Model;
...
  * }
  * @endcode
+ *
+ * Here's an example of iterating from a state at `timeA` to the last state
+ * before `timeB`:
+ * @code{.cpp}
+ * for (const auto& state : SimTK::makeIteratorRange(
+ *                              states.findAt(timeA),
+ *                              states.findNearestBefore(timeB)) {
+ *     ...
```
aseth1: This is  very helpful and likely what Tools will adopt to cycle through available states.

aseth1:
There are many test conditions but I don't believe they reflect the use cases you described to me.
If looping through the states by index, from some desired start to final time, I should write something like:

````
for(auto ix = states.findIndexNearestAfter(_start_time, tol);
              ix <=states.findIndexNearestBefore(_final_time, tol); 
              ++ix) {
... // do something with states[ix]
}
````
Is that correct? 
In the interest of minimally complete interface why encourage looping using the index when that can be replace with:

````
for (const auto& state : states.findBetween(_start_time, _final_time, tol)) {
    doSomethingWith(state);
}
````

OpenSim/Simulation/Test/testStatesTrajectory.cpp
chrisdembia: They aren't different...the second two were just for my own comfort. I could remove them.

OpenSim/Simulation/Test/testStatesTrajectory.cpp
chrisdembia: Good point; I'll rearrange.

chrisdembia:
I think it would be better for people to use the variants that give iterators rather than indices. I'm not sure if the iterators work in MATLAB, though. How about if I rearrange the methods so that the ones that give indices appear farther down?

OpenSim/Simulation/StatesTrajectory.h
chrisdembia: Yes, but based on our discussion yesterday, I'll change this so that if there are multiple states between `startTime - tolerance` and `startTime`, I'll use the one closest to `startTime` within that interval. And likewise for the last state.

OpenSim/Simulation/Test/testStatesTrajectory.cpp
chrisdembia: Based on our discussion yesterday, I'll change the behavior so you get 3.5 instead.

aseth1:
I forgot about MATAB - good point! The first case is how someone would loop through in MATLAB. A test case might show these two methods yield the identical times.

chrisdembia:
> I forgot about MATAB - good point! The first case is how someone would loop through in MATLAB. A test case might show these two methods yield the identical times.

Good idea. I'll add that test case and make it more clear that the indices are intended for MATLAB (and maybe people uncomfortable with iterators).

OpenSim/Simulation/Test/testStatesTrajectory.cpp
chrisdembia: You're right. I'll rephrase the comment. I'm leaving this test in because it's still the desired behavior.

OpenSim/Simulation/Test/testStatesTrajectory.cpp
chrisdembia: Good catch. This issue will go away when I enforce strictly increasing times.

chrisdembia:
I'll also add a test using realistic times from a simulation.

PR #777, opened 2016-02-22
Move methods to attach Geometry to frames into Frame class and rename methods for clarity.

aymanhab:
Address issues brought up by @chrisdembia while creating hopper example in particular:
1. Add a convenience method attachGeometry to allow for adding Geometry to any "Frame" object that sets the frame name on the Geometry internally.
2. Renamed the method addMeshGeometry to attachMeshGeometry for consistency with 1. above and to avoid having methods that have same name (addGeometry) and different meaning/function which was quite confusing.
3. Both methods above have been moved to Frame instead of Body for added flexibility.
4. Removed extra methods to access properties of concrete Geometry classes as Property accessor methods should be enough. The meat of the change is in Frame.* and elimination of method in Body.*, the rest is all text substitution due to renaming and simplifying examples as a result.

### DIFF
--- a/OpenSim/Simulation/Model/Frame.cpp
+++ b/OpenSim/Simulation/Model/Frame.cpp
@@ -88,11 +88,13 @@ void Frame::attachMeshGeometry(const std::string& aGeometryFileName, const SimTK
 }
 
 
-void Frame::attachGeometry(OpenSim::Geometry& geom, const SimTK::Vec3 scale)
+void Frame::attachGeometry(const OpenSim::Geometry& geom, const SimTK::Vec3 scale)
 {
-    geom.set_scale_factors(scale);
-    geom.setFrameName(getName());
-    addGeometry(geom);
+    SimTK::ClonePtr<Geometry> clone = SimTK::ClonePtr<Geometry>(geom);
+    clone->set_scale_factors(scale);
+    clone->setFrameName(getName());
+    addGeometry(clone.updRef());
+
 }
 
 //=============================================================================
--- a/OpenSim/Simulation/Model/Frame.cpp
+++ b/OpenSim/Simulation/Model/Frame.cpp
@@ -79,9 +79,17 @@ void Frame::extendAddGeometry(OpenSim::Geometry& geom) {
 }
 
 
-void Frame::addMeshGeometry(const std::string& aGeometryFileName, const SimTK::Vec3 scale)
+void Frame::attachMeshGeometry(const std::string& aGeometryFileName, const SimTK::Vec3 scale)
 {
     Mesh geom(aGeometryFileName);
+    geom.set_scale_factors(scale);
+    geom.setFrameName(getName());
+    addGeometry(geom);
+}
+
+
+void Frame::attachGeometry(OpenSim::Geometry& geom, const SimTK::Vec3 scale)
+{
     geom.set_scale_factors(scale);
     geom.setFrameName(getName());
     addGeometry(geom);
--- a/OpenSim/Simulation/Model/Frame.h
+++ b/OpenSim/Simulation/Model/Frame.h
@@ -193,7 +193,11 @@ public:
     */
     void attachMeshGeometry(const std::string &aGeometryFileName, const SimTK::Vec3 scale = SimTK::Vec3(1));
     /** Add a piece of Geometry to the list of Geometry owned by the Frame.
-    Scale defaults to 1.0 but can be changed on the call line, for convenience.
+    This function is a convenience for ModelComponent::addGeometry() for the case 
+    of adding Geometry directly to a Frame, and thus sets the "frame name" of the 
+    Geometry to this frame. The provided geom is copied into the Frame, which will
+    own a copy of the Geometry so any changes you make to geom after calling this 
+    method will not have an effect.
     */
     void attachGeometry(OpenSim::Geometry& geom, const SimTK::Vec3 scale = SimTK::Vec3(1));
 
--- a/OpenSim/Simulation/Model/Frame.h
+++ b/OpenSim/Simulation/Model/Frame.h
@@ -189,9 +189,13 @@ public:
     ///@}
 
     /** Add a Mesh specified by file name to the list of Geometry owned by the Frame.
-    Scale defaults to 1.0 but can be changed on the call line. For convenience 
+    Scale defaults to 1.0 but can be changed on the call line for convenience.
     */
-    void addMeshGeometry(const std::string &aGeometryFileName, const SimTK::Vec3 scale = SimTK::Vec3(1));
+    void attachMeshGeometry(const std::string &aGeometryFileName, const SimTK::Vec3 scale = SimTK::Vec3(1));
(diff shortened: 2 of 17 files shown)


OpenSim/Simulation/SimbodyEngine/Test/testJoints.cpp
```diff
@@ -1946,11 +1946,12 @@ void testAutomaticJointReversal()
     //thigh.addGeometry(Cylinder(0.035, 0.4));
     thigh->upd_geometry(0).setColor(SimTK::Vec3(0, 0, 1));   // BLUE
     
-    //TODO fix Bug: ModleComponent::addGeometry assumes ownership but does not
-    // take a pointer to heap allocated Geometry
-    shank->addGeometry(*new Cylinder(0.02, 0.243800));
+    //ModelComponent::addGeometry makes a copy of the passed in Geometry
+    Cylinder cyl(0.02, 0.243800);
+    shank->addGeometry(cyl);
```
chrisdembia: Is this supposed to be `attachGeometry()`?

OpenSim/Simulation/Model/Frame.h
```diff
@@ -189,9 +189,13 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Frame, ModelComponent);
...
 
     /** Add a Mesh specified by file name to the list of Geometry owned by the Frame.
-    Scale defaults to 1.0 but can be changed on the call line. For convenience 
+    Scale defaults to 1.0 but can be changed on the call line for convenience.
     */
-    void addMeshGeometry(const std::string &aGeometryFileName, const SimTK::Vec3 scale = SimTK::Vec3(1));
+    void attachMeshGeometry(const std::string &aGeometryFileName, const SimTK::Vec3 scale = SimTK::Vec3(1));
+    /** Add a piece of Geometry to the list of Geometry owned by the Frame.
+    Scale defaults to 1.0 but can be changed on the call line, for convenience.
+    */
```
chrisdembia: I think the statement about scaling is evident from the function declaration. 

Could you add a bit to this doxygen documentation more clearly stating that this is a convenience over `ModelComponent::addGeometry()` for the case of adding geometry to a Frame, and that this sets the frame name? Also, a statement about memory management would be good too.
Perhaps something like:

> This function is a convenience for ModelComponent::addGeometry() for the case of adding Geometry directly to a Frame, and thus sets the "frame name" of the Geometry to *this* frame.  The provided `geom` is copied into this Frame, so any changes you make to `geom` after calling this method will not have an effect.

chrisdembia:
Aside from my one comment, this looks good to me.

OpenSim/Simulation/SimbodyEngine/Test/testJoints.cpp
aymanhab: It's possible, I was mainly replacing the addGeometry calls only if they were setting the framename. In general many instances of addGeometry will stay in cases where the object is not a subclass of Frame or if we don't want the Frame to "own" the Geometry.

chrisdembia: Okay. Your comment indicates that there is a memory leak.

chrisdembia: I suppose the leak is out of the scope of this PR though.

aymanhab: The old code snippet was leaking because it was assuming the ModelComponent was taking ownership of the passed in Geometry, the new code doesn't leak since the geometry is allocated on the stack and freed when out of scope, and the copy made downstream is handled by the model (properly I assume)..

chrisdembia: Oh I'm so sorry! I was having trouble reading the diff. Thanks for fixing the leak.

OpenSim/Simulation/Model/Frame.h
```diff
@@ -189,9 +189,13 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Frame, ModelComponent);
...
+    void attachGeometry(OpenSim::Geometry& geom, const SimTK::Vec3 scale = SimTK::Vec3(1));
```
chrisdembia: I just realized...couldn't this be `const OpenSim::Geometry& geom` if it's going to make a copy anyway? This would have allowed you to use a one-liner like I think you wanted to do in `testJoints`.

aymanhab: Good point @chrisdembia, the problem is that the copying is done internally by the list property so I had to use a non-const reference to set values before copying (in Frame::attachGeometry). I also thought that making it non-const makes it clear that the method modifies the Geometry by setting the frame name and maybe scale. Do you think it'd be better to make it const, have the property create and add a copy then retrieve the copy and set the attributes on it?

chrisdembia: > Do you think it'd be better to make it const, have the property create and add a copy then retrieve the copy and set the attributes on it?

Oh, good idea! I was thinking of just copying it twice.

aymanhab:
Addressed feedback, thanks @chrisdembia Can you please review the use of ClonePtr in Frame.cpp. Thanks

OpenSim/Simulation/Model/Frame.cpp
```diff
@@ -87,6 +87,16 @@ void Frame::addMeshGeometry(const std::string& aGeometryFileName, const SimTK::V
     addGeometry(geom);
 }
 
+
+void Frame::attachGeometry(const OpenSim::Geometry& geom, const SimTK::Vec3 scale)
+{
+    SimTK::ClonePtr<Geometry> clone = SimTK::ClonePtr<Geometry>(geom);
+    clone->set_scale_factors(scale);
+    clone->setFrameName(getName());
+    addGeometry(clone.updRef());
```
chrisdembia: I don't think pointers need to be involved:

```
Geometry geomCopy(geom);
geomCopy.set_scale_factors(scale);
geomCopy.setFrameName(getName());
addGeometry(geomCopy);

aymanhab: Here's the concern, Geometry is just base class, clone returns the correct concrete subclass. We want to make sure we preserve the actual concrete subclass, not sure you can do it with the above code, does that sound right?

chrisdembia: Oh I see. Then this looks good. The ClonePtr constructor you use will make a copy of `geom`.

chrisdembia:
This looks good to me now. Thanks for responding to the feedback that @klshrinidhi and I had.

OpenSim/Simulation/Model/Frame.cpp
chrisdembia: Actually, there might be two copies: one by the constructor, and one by the assignment. The compiler might optimize away the copy assignment, though I'm not sure.

aymanhab:
Based on the fact that this PR adds methods that simplify the examples, plugs a memory leak and improves the documentation and the great feedback from @chrisdembia  I'm merging this PR.

Issue #785, opened 2016-02-27
example{1|2}DataTable should reside in the Examples folder


tkuchida:
![solutionexplorer](https://cloud.githubusercontent.com/assets/4203505/13372278/6a69c726-dcf2-11e5-9769-b4d1290209fc.png)

chrisdembia:
Yes.

tkuchida:
This issue was addressed by #786. Thanks for merging, @chrisdembia. :penguin:

PR #786, opened 2016-02-27
example{1|2}DataTable projects now appear in Examples folder

tkuchida:
Addresses #785.

### DIFF
--- a/OpenSim/Examples/DataTable/CMakeLists.txt
+++ b/OpenSim/Examples/DataTable/CMakeLists.txt
@@ -1,7 +1,9 @@
-include_directories(${PROJECT_SOURCE_DIR})
+file(GLOB EXAMPLES_TO_COMPILE "*.cpp")
+include_directories(${OpenSim_SOURCE_DIR} ${OpenSim_SOURCE_DIR}/Vendors)
 
-add_executable(example1DataTable example1.cpp)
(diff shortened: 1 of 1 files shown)


klshrinidhi:
@tkuchida 
I think moving examples is a good idea and not sure why I didn't have it this way in the first place.
But I would avoid using `GLOB` to grab all the cpp files to compile. [It is not recommended by CMake](https://cmake.org/cmake/help/v3.4/command/file.html).

tkuchida:
> But I would avoid using `GLOB` to grab all the cpp files to compile.

Fixed. I followed the other CMakeLists.txt files in the project, many of which use `GLOB`. Should an issue be created to eradicate `GLOB`?

klshrinidhi:
Thanks for fixing.
I think an issue to track *elimination of `GLOB`* would be good.

tkuchida:
TODO: if the determination in #787 is that `GLOB` is actually preferred, undo the change made in the second commit to this branch before merging.

chrisdembia:
It's fine to use not use GLOB here I think; @sherm1's comment in #787 is more related to `OpenSim/Common/`, `OpenSim/Simulation/`, etc.

tkuchida:
> It's fine to use not use GLOB here I think

In that case, this PR can be merged IMO (the AppVeyor failure is irr:elephant: -- same thing happens [on master](https://ci.appveyor.com/project/opensim-org/opensim-core/branch/master)).

Issue #697, opened 2015-11-06
Tools crash Matlab if model_file is not specified


aymanhab:
This may become obsolete if we move away from using Tools but for the time being, some tools (e.g. RRATool) crashes rather than throw an exception if model file is not specified in XML. A meaningful error message should be reported instead ( reported by user on the forum https://simtk.org/forums/viewtopic.php?f=91&t=5192&view=unread#unread)

GUI created files do not have <model_file> specified since the GUI works on live models.

chrisdembia:
This might be more about exception handling in SWIG; as I've indicated before, at least with python, we should move toward handling all C++ exceptions through swig.

aymanhab:
@chrisdembia AFAIK all C++ exceptions are translated through SWIG to Java exceptions and so wouldn't cause a crash. In this case I think we just dereference a null pointer directly, and that's what we need to fix/change I think.

jenhicks:
@jimmyDunne Can you verify that this is fixed? Matlab shouldn't crash and you should get a useful exception warning message (e.g., tool is missing a model)

jimmyDunne:
Not having a model specified in the Fwd Setup XML does **still** crash Matlab (tested using the gait2392 pipeline).

jimmyDunne:
Can confirm, NOT having the model specified in the Setup XMLs for RRATool, CMCTool, and ForwardTool, all cause a fatal Matlab crash when Tool is instantiated.

chrisdembia:
Darnit. Is anything useful printed? Is there a crash log (in your home directory) that might show a stack trace?

jimmyDunne:
MATLAB crash file:/Users/jimmy/matlab_crash_dump.3303-1:


------------------------------------------------------------------------
       Segmentation violation detected at Thu Feb 23 19:55:08 2017
------------------------------------------------------------------------

Configuration:
  Crash Decoding      : Disabled
  Crash Mode          : continue (default)
  Current Graphics Driver: Unknown hardware 
  Current Visual      : Quartz
  Default Encoding    : ISO-8859-1
  Host Name           : mpu
  MATLAB Architecture : maci64
  MATLAB Root         : /Applications/MATLAB_R2015b.app
  MATLAB Version      : 8.6.0.267246 (R2015b)
  OpenGL              : hardware
[...]

PR #793, opened 2016-03-02
Throw exception if model file is not specified in xml when constructing tools

klshrinidhi:
Addressing #697.


klshrinidhi:
Additional note: The crash is a *seqfault* and it happens during construction of the tools, at the first non-comment line in the function `Component::disconnect()`. It happens when attempting to access the data member `_components`.

OpenSim/Simulation/Model/AbstractTool.cpp
```diff
@@ -110,6 +110,10 @@ AbstractTool::AbstractTool(const string &aFileName, bool aUpdateFromXMLNode):
     _analysisSet.setMemoryOwner(false);
     setNull();
     if(aUpdateFromXMLNode) updateFromXMLDocument();
+
+    if(_modelFile.empty())
```
aseth1: we can now use `OPENSIM_THROW_IF_FRMOBJ` and the message could be a bit clearer, e.g:
*"No model file was specified (<model_file> tag is empty) in the Tool's Setup file."*

Issue #686, opened 2015-11-02
Throw exception if xml parsing encounters an xml tag that does not correspond to the class's properties


chrisdembia:
We can have a global override to let these errors pass silently, for use in testing.

tkuchida:
Should also catch unrecognized property values (e.g., anything other than "true" or "false" for a bool).

jenhicks:
closing as this is a duplicate

PR #803, opened 2016-03-04
[WIP] Error checking while parsing xml

klshrinidhi:
Addressing #686.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/803)

### DIFF
--- a/OpenSim/Common/Object.cpp
+++ b/OpenSim/Common/Object.cpp
@@ -839,6 +839,8 @@ UpdateXMLNodeTransform(const Property_Deprecated*   aProperty,
 //-----------------------------------------------------------------------------
 void Object::updateFromXMLNode(SimTK::Xml::Element& aNode, int versionNumber)
 {
+    checkForInvalidPropertiesInXml(aNode);
+
 try {
     // NAME
     const string dName = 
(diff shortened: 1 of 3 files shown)


chrisdembia:
Looks great to me. However, I think we shouldn't close #686 until we set the default to "true".

klshrinidhi:
@chrisdembia 
Default is indeed `true` right now.

OpenSim/Common/Object.cpp
```diff
@@ -61,7 +61,7 @@ std::map<string,string>     Object::_renamedTypesMap;
 bool                        Object::_serializeAllDefaults=false;
 const string                Object::DEFAULT_NAME(ObjectDEFAULT_NAME);
 int                         Object::_debugLevel = 0;
-
+bool                        Object::_invalidPropertyException{false};
```
chrisdembia: @klshrinidhi it says `false` here.

klshrinidhi: Oh. Right. I haven't committed the change.
Will do today.

chrisdembia:
Also, I think this change could be mentioned in the CHANGELOG

aymanhab:
This was addressed by issue #2911

PR #806, opened 2016-03-06
Macros for outputs (OpenSim_DECLARE_OUTPUT())

chrisdembia:
I've created three macros for creating outputs:

- OpenSim_DECLARE_OUTPUT
- OpenSim_DECLARE_OUTPUT_FLEX
- OpenSim_DECLARE_OUTPUT_FOR_STATE_VARIABLE

They work just fine, and once this is merged, then other people can start using these macros in prep for the hackathon.  To be consistent, I tried to follow our conventions for properties pretty closely.

I also documented these macros so they appear nicely in doxygen, just like properties. You can view the doxygen here: http://myosin.sourceforge.net/806/classOpenSim_1_1Model.html

I didn't spend too much time on the documentation because it is likely to change when we add multi-outputs and create similar macros for inputs, connectors, etc.

Other than the FLEX macro above, I don't think these macros need to take comments, since member functions and state variables should be documented elsewhere.

Two tasks that might be good to do but that I'd prefer to leave to a subsequent PR:

1. Allow default value for the dependsOn stage through the macros; this requires overloaded/variadic macros and is somewhat complicated but is definitely do-able.
2. The member variable created by `constructOutput()` is an int whose value is always 0. This doesn't make too much sense, but it may also be OK. Luckily, the macro lets us change this in the future, and so I'd rather leave that for a separate PR. Also, we can deal with macro-generated accessor methods in a separate PR (probably after the hackathon).

Once this is merged, I'll move all existing `constructProperty()` calls to the macro.

Also, we can create new macros for list outputs once we figure out how to do those.

### DIFF
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -820,7 +820,8 @@ setDiscreteVariableValue(SimTK::State& s, const std::string& name, double value)
     }
 }
 
-int Component::constructOutputForStateVariable(const std::string& name)
+Component::OutputIndex
+Component::constructOutputForStateVariable(const std::string& name)
 {
     return constructOutput<double>(name,
             std::bind(&Component::getStateVariableValue,
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -820,9 +820,9 @@ setDiscreteVariableValue(SimTK::State& s, const std::string& name, double value)
     }
 }
 
-void Component::constructOutputForStateVariable(const std::string& name)
+int Component::constructOutputForStateVariable(const std::string& name)
 {
-    constructOutput<double>(name,
+    return constructOutput<double>(name,
             std::bind(&Component::getStateVariableValue,
                 this, std::placeholders::_1, name),
             SimTK::Stage::Model);
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -820,8 +820,7 @@ setDiscreteVariableValue(SimTK::State& s, const std::string& name, double value)
     }
 }
 
-Component::OutputIndex
-Component::constructOutputForStateVariable(const std::string& name)
+bool Component::constructOutputForStateVariable(const std::string& name)
 {
     return constructOutput<double>(name,
             std::bind(&Component::getStateVariableValue,
--- a/OpenSim/Common/ComponentOutput.h
+++ b/OpenSim/Common/ComponentOutput.h
@@ -168,7 +168,7 @@ private:
 /// Use these macros at the top of your component class declaration,
 /// near where you declare @ref Property properties.
 /// @{
-/** Create an output for a member function of the same component. 
+/** Create an output for a member function of this component.
  *  The following must be true about componentMemberFunction, the function
  *  that returns the output:
  *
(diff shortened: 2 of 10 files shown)


chrisdembia:
Here's what the doxygen documentation looks like for outputs:
image

sherm1:
:+1:

chrisdembia:
Here's a question: should the macro automatically fill in the `SimTK::Stage::` part of the stage? That way, people would invoke the macro as:

```cpp
OpenSim_DECLARE_OUTPUT(force, double, getForce, Dynamics);
```

OpenSim/Common/Component.cpp
```diff
@@ -820,9 +820,9 @@ setDiscreteVariableValue(SimTK::State& s, const std::string& name, double value)
     }
 }
 
-void Component::constructOutputForStateVariable(const std::string& name)
+int Component::constructOutputForStateVariable(const std::string& name)
```
aseth1: Should we settle now to use `SimTK_DEFINE_UNIQUE_INDEX_TYPE()`? instead of an int?
With the potential of having many indices I like using the unique index to thwart accidentally mixing indices up.

OpenSim/Common/ComponentOutput.h
```diff
@@ -164,6 +164,114 @@ class Output : public AbstractOutput {
 //=============================================================================
 };  // END class Output
 
+/// @name Creating Outputs for your Component
+/// Use these macros at the top of your component class declaration,
+/// near where you declare @ref Property properties.
+/// @{
+/** Create an output for a member function of the same component. 
+ *  The following must be true about componentMemberFunction, the function
```
aseth1: same -> this component?

OpenSim/Common/ComponentOutput.h
```diff
@@ -164,6 +164,114 @@ class Output : public AbstractOutput {
 //=============================================================================
 };  // END class Output
 
+/// @name Creating Outputs for your Component
+/// Use these macros at the top of your component class declaration,
+/// near where you declare @ref Property properties.
+/// @{
+/** Create an output for a member function of the same component.
```
aseth1: same -> this component.

chrisdembia:
I've addressed your feedback. I've created an OutputIndex class, but I still set its value to 0 for all outputs. After the hackathon, we should revisit what we do with this return value.

chrisdembia:
Updated to replace OutputIndex with bool.

aseth1:
Don't forget to remove OpenSim_DECLARE_OUTPUT_FLEX in a subsequent PR,

OpenSim/Common/ComponentOutput.h
```diff
@@ -164,6 +164,114 @@ class Output : public AbstractOutput {
...
+    /** @name Outputs                                                    */ \
+    /** @{                                                               */ \
+    /** Provides the value of func##() and is available at stage ostage. */ \
+    /** This output was generated with the                               */ \
+    /** #OpenSim_DECLARE_OUTPUT macro.                                   */ \
+    OpenSim_DOXYGEN_Q_PROPERTY(T, oname)                                    \
+    /** @}                                                               */ \
+    /** @cond                                                            */ \
+    bool _unused_output_##oname { constructOutput<T>(#oname, &Self::func, ostage) }; \
+    /** @endcond                                                         */
```
aseth1: minor but instead of `_unused_output...` you could rename it to `has_output_...` since it is a bool anyway. Might be less confusing than having a bool member named `_unused` and has a value of `true`.

chrisdembia:
Updated; tests pass.

aseth1:
Looks great!

chrisdembia:
Thank you.

PR #808, opened 2016-03-06
New Exception ctor that takes the name of a type and object.

chrisdembia:
It's very common that we want to throw an exception in which the message contains the name of the class and the name of an instance of that class. It's somewhat tedious to create error messages that include these names, and we are often inconsistent about how we format such messages.

This exception constructor allows us to include such names in exception messages more easily and uniformly.

@klshrinidhi


OpenSim/Common/Exception.cpp
```diff
@@ -70,8 +70,21 @@ exception()
...
-    addMessage(file + ":" + std::to_string(line) + "\n" + "In function '" + 
-               func + "'");
+    addMessage("\tIn file " + file + ":" + std::to_string(line) + "\n" +
+               "\tIn function '" + func + "'");
+}
+
+Exception::Exception(const std::string& file,
+              size_t line,
+              const std::string& func,
+              const std::string& className,
```
aseth1: What do you think of passing a const ref to the object itself? And strip the className and objectName out for us?

chrisdembia: Okay yeah. That will work for all Objects.

klshrinidhi:
When added the previous constructor, I wanted to add the class name too. I didn't add it at the time because there was no standard way to automatically infer class name at the throw site using macros like I have [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Exception.h#L62). It had to be inferred by demangling result of `typeid`. There is a library to do this. I wouldn't like it if throw sites had to specify class name all the time.

chrisdembia:
`SimTK::NiceTypeName<T>::namestr()` returns the demangled name.

> There is a library to do this. I wouldn't like it if throw sites had to specify class name all the time.

I agree.

What about the name associated with an Object/Component (not its type name, its instance name)?

chrisdembia:
Should we hold off on this PR then?

chrisdembia:
Merge conflict resolved.

klshrinidhi:
I personally would hold off on this until compilers give us macros to do this. I think the *file-name, function name and line-number* are enough information until then. 
If you want to go ahead with this, I think there will be need for another macro `OPENSIM_THROW_<something>` that uses this constructor. Every throw site will have to explicitly disambiguate by calling the right macro.

klshrinidhi:
:+1: 
I think `OPENSIM_THROW_<something>` macro to pick up this constructor is the only thing left here.

klshrinidhi:
I have:
* Added another constructor similar to the one Chris added. This constructor is intended for derived classes. That makes a pair of these constructors just like there was a pair before that does not take object.
* I have added two macros for throw sites to use. The macros pick up the `this` pointer so throw sites do not have to specify explicitly. I mentioned this in my first comment on this PR

Take a look at the macros. I have not used them yet.

chrisdembia:
Looks good to me! Thank you for doing this.

klshrinidhi:
Thanks for convincing me.
All :white_check_mark: !!!
Merging now.

Issue #748, opened 2016-01-25
Issues related to subcomponents management and traversal


aseth1:
PR #740 addressed the issue of marking subcomponents as properties and enabling immediate traversal of those subcomponents. But several issues remain:

1. Subcomponents that are data members are being lost because the list of subcomponents is being redefined in finalizeFromProperties. 

2. Dynamically allocated subcomponents are being leaked. 

3. The traversal tree is being rebuilt on every call to getComponentList and it is faking being a const method.

The solution to these are coupled and I propose tackling them in the following way:

1. Create an internal (my) subcomponents iterator that traverses three internal component lists: members, properties, and adoptees (of dynamically allocated components during processing of properties and/or wiring up (connecting) the Component and does not modify its properties).  The iterator is valid over the duration that the Component is not modified. We can then remove the list of subcomponents since figuring out when to generate the complete list (e.g. mark) and maintain its validity is confusing/error prone. Access to the iterator is protected. The adoptees are owned and destroyed. The markAsSubcomponent() method would be eliminated.

2. A public ComponentList iterator to traverse all subcomponents including those of the belonging to subcomponents will use the the internal iterators and call each recursively to traverse the complete tree live. No tree or data structure is created or cached. The Component::_nextComponent pointer can then be removed and getting the iterator is const.

klshrinidhi:
Iterator class to traverse a list of containers as if they were concatenated.

https://github.com/opensim-org/opensim-core/blob/container_chain_iterator/OpenSim/Common/ContainerChainIterator.cpp

sherm1:
> Iterator class to traverse a list of containers as if they were concatenated.

Sweet! One question: could the initializer list take references rather than pointers?

Issue #390, opened 2015-03-09
All Components should have a name


chrisdembia:
We can either throw an error if an object does not have a name or autogenerate a name. See #387 for related discussion.

This could happen upon `addComponent`.

chrisdembia:
If we autogenerate a name we'll have to be very careful. I think there are some places where a class changes its own name and that it wasn't good.

jimmyDunne:
need to choose the naming methods

PR #810, opened 2016-03-07
Subcomponent creation management and traversal addresses #748

aseth1:
The purpose of this PR is to address the issues in  #748.

There are now 3 private lists of subcomponents according to whether they are properties (member of a `ComponentSet`), class members or dynamically created and adopted components (e.g. adding joints and constraints to break loops).  Therefore, member and adopted components are not being lost when properties are changed and `finalizeFromProperties` and adopted (dynamically allocated subcomponents) are no longer being leaked

Added `Component::constructSubcomponent()` non-static initializer to create internal subcomponents that are not serialized. It returns the index of the component in the private _memberSubcomponents list. Members and properties are automatically marked as subcomponents and adopted components are marked upon adoption, so there is no requirement to mark a component as a subcomponent anymore, which addresses #421. 

Extended tree traversal to include member and adopted components. Moved implementation to .cpp and removed need to const_cast. But, the traversal tree is still built everytime getComponent is called. This enables us to iterate over subcomponents when the component is constructed or immediately after `finalizeFromProperties` if the component has been edited or is a fresh cop of another component. Address #536.

Added `traversePathToComponent<C>()` facility to Component to traverse the tree to get subcomponents by unique path name and type. `get/updComponent()` now use it.

Direct traversal of the component path requires knowing the path of the Component with respect to the Connector that it satisfies and `Connector<C>.connect(const Component& c)` now computes the relative path name to store as the `connectee_name`. Added `Component::getFullPathName()` and `getRelativePathName` to serve Connectors and to replace the old path facility which was prone to being stale and now is never cached.

As part of using paths have to enforce that Component names are never left empty which addresses #390.

Added Exceptions for components that have no name and if component cannot be found along specified path.

Remainder of the many changes are present to handle the consequences of these changes in test cases.

@chrisdembia , @aymanhab, @klshrinidhi please review that the interface is sufficient and clear for:
1. defining subcomponents (as properties, members or adoptees without marking) and member initializer
2. accessing subcomponents by path name (traversal) and type
3. finding subcomponents by name and type only so that we can load old models (should throw if more than one match is found)

The rest of the Component API should have remained the same.

Your comments will be particularly useful for improving the doxygen comments which are quite thin.

I will also start a "to do" issue related to this PR that can be revisited after the hackathon, for example to optimize iterators through subcomponents, adding member accessors, etc...


OpenSim/Common/Component.cpp
```diff
@@ -127,6 +127,24 @@ Component::Component(SimTK::Xml::Element& element) : Object(element)
 void Component::finalizeFromProperties()
 {
     reset();
+
+    // TODO use a flag to set whether we are lenient on having nameless
```
chrisdembia: Is this TODO documented somewhere on GitHub?

OpenSim/Common/Component.cpp
```diff
@@ -134,11 +152,19 @@ void Component::finalizeFromProperties()
...
-    for (unsigned int i = 0; i<_components.size(); i++){
-        _components[i]->finalizeFromProperties();
+    for (auto& comp : _memberSubcomponents) {
+        const_cast<Component*>(comp.get())
+            ->finalizeFromProperties();
+    }
+    for (auto& comp : _propertySubcomponents) {
+        comp->finalizeFromProperties();
+    }
+    for (auto& comp : _adoptedSubcomponents) {
```
chrisdembia: Another name to consider: "runtime subcomponents" I think "adopted subcomponents" is also fine.

OpenSim/Common/Component.cpp
```diff
@@ -134,11 +152,19 @@ void Component::finalizeFromProperties()
 }
 
 // Base class implementation of virtual method.
-// Call extendFinalizeFromProperties on all components
+// Call finalizeFromProperties on all subcomponents
 void Component::componentsFinalizeFromProperties() const
 {
-    for (unsigned int i = 0; i<_components.size(); i++){
-        _components[i]->finalizeFromProperties();
```
chrisdembia: I don't understand why the const casts are now necessary.

OpenSim/Common/Component.cpp
```diff
@@ -457,8 +479,8 @@ int Component::getNumStateVariables() const
     //Get the number of state variables added (or exposed) by this Component
     int ns = getNumStateVariablesAddedByComponent(); 
     // And then include the states of its subcomponents
-    for(unsigned int i=0; i<_components.size(); i++)
-        ns += _components[i]->getNumStateVariables();
+    for(unsigned int i=0; i<_propertySubcomponents.size(); i++)
+        ns += _propertySubcomponents[i]->getNumStateVariables();
```
chrisdembia: How come this doesn't include adopted or member subcomponents?

aseth1: Good find! It most definitely should have to go through the other two! 
I was waiting to replace with a "my subcomponents" iterator instead of looping through three lists. I may be able use that in multiple places see above. Maybe I will just make it correct first then replace with the iterator.

OpenSim/Common/Component.cpp
aseth1: `_propertySubcomponents` is an array of `SimTK::ReferencePtr`s and `->` operator returns a writeable reference even though it is a const method. The other two arrays use `SimTK::ClonePtr` and here `get()` is const and returns const reference. `componentsFinalizeFromProperties` itself is const to reflect that `*this` is not being modified and `finalizeFromProperties` is being called on its subcomponents. I am on the fence on whether that is important/relevant distinction or not.

OpenSim/Common/Component.h
```diff
@@ -335,56 +367,53 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
+
+    /** Get the complete pathname for this Component to its ancestral Component,
+     *  which is the root of the tree to which this Component belongs.
+     * For example: a Coordinate Components would have a full path name like:
+     *  `/arm26/elbow_r/flexion`. Accessing a Component by its fullPathName from
+     * root is guaranteed to be unique. */
+    std::string getFullPathName() const;
+
+
+    /** Get the relative pathname of this Component with respect to another one */
```
chrisdembia: What does this do if `wrt` is not part of the same tree? Does it just return the full path to `this` component?

aseth1: Yes, it should like [this](https://github.com/opensim-org/opensim-core/pull/810/files#diff-074825ae2a365226376a3224512e6329R691).

OpenSim/Common/Component.h
```diff
@@ -335,56 +367,53 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
 
         // Only error cases remain
-        std::string msg = getConcreteClassName() + " '" + getName() + "': ERROR- ";
-
-        // too many components of the right type with the same name
-        if (foundCs.size() > 1) {
-            msg +="Found multiple '" + name + "'s of type "+ C::getClassName() + ".";
-            throw Exception(msg, __FILE__, __LINE__);
-        }
+        OPENSIM_THROW(ComponentNotFoundOnSpecifedPath, name,
```
chrisdembia: Specifed -> Specified

OpenSim/Common/Component.h
```diff
@@ -335,56 +367,53 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
+    std::string getFullPathName() const;
+
+
+    /** Get the relative pathname of this Component with respect to another one */
+    std::string getRelativePathName(const Component& wrt) const;
+
     /**
-     * Get a subcomponent of this Component by its name. 
+     * Get a subcomponent of this Component by its path name. 
      * Note using a component's full "path" name is faster and will provide a
```
chrisdembia: Doesn't this always yield a unique result now?

OpenSim/Common/Component.h
```diff
@@ -925,24 +954,27 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
     template<class C=Component>
-    C& constructSubcomponent(const std::string& name) {
+    MemberSubcomponentIndex constructSubcomponent(const std::string& name) {
         C* component = new C();
         component->setName(name);
-        this->markAsSubcomponent(component);
-        return *component;
+        component->setParent(*this);
+        _memberSubcomponents.push_back(SimTK::ClonePtr<Component>(component));
+        return MemberSubcomponentIndex(_memberSubcomponents.size()-1);
```
chrisdembia: I understand now why the member can't be a reference.

If you add those accessors by index, then Tom can use them in his PR.

OpenSim/Common/Component.h
```diff
@@ -1790,7 +1917,7 @@ template <class T> friend class ComponentMeasure;
     // These are just references, don't delete them!
     // TODO: subcomponents should not be exposed to derived classes to trash.
     //       Need to provide universal access via const iterators -aseth
```
chrisdembia: Is this TODO also an issue on GitHub?

OpenSim/Common/Component.h
```diff
@@ -1996,10 +2136,34 @@ void ComponentListIterator<T>::advanceToNextValidComponent() {
...
 template<class C>
 void Connector<C>::findAndConnect(const Component& root) {
-    const C& comp = root.getComponent<C>(get_connectee_name());
-    connectee = comp;
+ 
+    const std::string& path = get_connectee_name();
+    const C* comp = nullptr;
+
+    try {
+        if (path[0] == '/') { //absolute path name
```
chrisdembia: We should (eventually) have a central place for making queries about a path so that this check could be `ComponentPathUtil::isAbsolute(path)` or something like that. Or, we can have a new class, "ComponentPath"; then we could do `path.isAbsolute()`. Not for this PR...

OpenSim/Common/Component.h
```diff
@@ -1996,10 +2136,34 @@ void ComponentListIterator<T>::advanceToNextValidComponent() {
...
+            comp =  &root.template getComponent<C>(path);
+        }
+        else { // relative path name
+            comp =  &getOwner().template getComponent<C>(path);
```
chrisdembia: What would happen if you gave a full path to `owner.getComponent()`?

OpenSim/Common/Test/testComponentInterface.cpp
```diff
@@ -601,3 +653,146 @@ int main() {
     cout << "Done" << endl;
     return 0;
 }
+
+
+void testComponentPathNames()
```
chrisdembia: Awesome!

OpenSim/Common/Component.h
chrisdembia: Ah so the path name stuff doesn't account for the actual tree; it just deals with strings.

chrisdembia:
@aymanhab @klshrinidhi: Ajay updated the main description of this PR with some things to look for. I'm not sure if GitHub sends out notifications when you are @mentioned in a followup edit.

Applications/Scale/test/testScale.cpp
```diff
@@ -240,6 +240,13 @@ void scaleModelWithLigament()
     std.print("std_toyLigamentModelScaled_latest.osim");
     comp.print("comp_toyLigamentModelScaled_latest.osim");
 
+    // the latest model will not match the standard because the naming convention has
```
tkuchida: latest model -> latest model (comp)

Applications/Scale/test/testScale.cpp
```diff
@@ -240,6 +240,13 @@ void scaleModelWithLigament()
...
+    // been updated to store path names and connecting a model results in connectors
+    // storing relative paths so that collections of components are more portable.
+    // The models must be equivalent after being connected.
```
tkuchida: after being connected -> after being connected (i.e., after setup() is called)

OpenSim/Common/Component.cpp
```diff
@@ -134,11 +152,19 @@ void Component::finalizeFromProperties()
 }
 
 // Base class implementation of virtual method.
-// Call extendFinalizeFromProperties on all components
+// Call finalizeFromProperties on all subcomponents
 void Component::componentsFinalizeFromProperties() const
```
tkuchida: Is it possible (through some circuitous route) for this method to result in an infinite loop of `finalizeFromProperties()` calls? For example, is there anything preventing an adopted subcomponent from adopting its grandparent? I would have expected something like
```cpp
if (!_finalizedFromProperties) {
    _finalizedFromProperties = true;
    ...
}
```
I'm still learning the Component API, so please ignore if the machinery is already in place to prevent this. :frog:

OpenSim/Common/Component.cpp
```diff
@@ -202,19 +208,25 @@ void Component::connect(Component &root)
 
 
 // Call connect on all components and find unconnected Connectors a
```
tkuchida: Comment is incomplete. Also, I don't see where unconnected Connectors are detected, either here or in `connect()`.

OpenSim/Common/Component.cpp
```diff
@@ -202,19 +208,25 @@ void Component::connect(Component &root)
...
-void Component::componentsConnect(Component& root) const
+void Component::componentsConnect(Component& root) 
 {
     // First give the subcomponents the opportunity to connect themselves
```
tkuchida: Obsolete comment? (Where's `// Next, ...`?)

OpenSim/Common/Component.cpp
```diff
@@ -202,19 +208,25 @@ void Component::connect(Component &root)
...
+        _adoptedSubcomponents[i].upd()->connect(root);
     }
 }
 
 void Component::disconnect()
 {
     // First give the subcomponents the opportunity to disconnect themselves
-    for (unsigned int i = 0; i<_components.size(); i++){
-        _components[i]->disconnect();
+    for (unsigned int i = 0; i<_propertySubcomponents.size(); i++){
```
tkuchida: Why aren't `memberSubcomponents` and `adoptedSubcomponents` disconnected also?

OpenSim/Common/Component.h
```diff
@@ -1577,10 +1595,82 @@ template <class T> friend class ComponentMeasure;
...
          by owning component(s). */
-    const Component* findComponent(const std::string& name, 
-                                   const StateVariable** rsv = nullptr) const;
+    template<class C = Component>
+    const C* findComponent(const std::string& name, 
+                           const StateVariable** rsv = nullptr) const {
+        std::string msg = getConcreteClassName() + "'" + getName() +
+                          "'::findComponent() ";
+        if (name.empty()) {
+            msg += "cannot find a nameless subcomponent.";
```
tkuchida: If this is a user-facing message, perhaps something like "ensure all subcomponents have been named" would be friendlier?

OpenSim/Common/Component.h
```diff
@@ -1577,10 +1595,82 @@ template <class T> friend class ComponentMeasure;
         of StateVariables by name. 
         
         NOTE: If the component name or the state variable name is ambiguous, the 
-         first instance found is returned. To disambiguate use the full name provided
+         an exception is thrown. To disambiguate use the full name provided
```
tkuchida: This comment ("if ambiguous, an exception is thrown") is not entirely true: line 1618 will immediately return if `this` matches, regardless of whether ambiguity exists. Perhaps the immediate return on line 1618 is necessary for performance reasons, but it seems undesirable that the method has two potential outcomes when there's ambiguity (peaceful return if `this` happens to match, and an exception otherwise). In any event, this comment doesn't currently match the behavior of the method.

aseth1: Yes, there is a bug that it should verify against the the full path name instead of just assuming it is the top of a relative path in the return case, otherwise it should have added itself to the `foundCs` list. Thanks!

As background 1618 was added to handle the corner case where you are calling findComponent() on the component you actually wanted to find. It almost never happens (why would you call find for a Component if you already had a handle to it to ask it to find) but I thought it would be weird for it to say it could not find itself. It's not there for performance.

aymanhab:
Would it be clearer @aseth if we rename Component::constructSubcomponent() to Component::constructInternalComponents() since there're many "sub"components now (based on second paragraph of description)?

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -659,34 +657,42 @@ void Model::extendConnectToModel(Model &model)
...
-            Component* compToMoveOut = _components.at(m+nb);
+            Component* compToMoveOut = _propertySubcomponents.at(m+nb).get();
             // reorder the joint components in the order of the multibody tree
             Joint* jointToSwap = static_cast<Joint*>(mob.getJointRef());
-            it = std::find(_components.begin(), _components.end(), jointToSwap);
-            if (it != _components.end()){
+            auto it = std::find(_propertySubcomponents.begin(),
+                                _propertySubcomponents.end(), 
+                                SimTK::ReferencePtr<Component>(jointToSwap));
+            if (it != _propertySubcomponents.end()){
```
chrisdembia: There's nothing preventing someone for adding joints to their model as member or adopted subcomponents, right? So those should be handled as well. I think this can be taken care of in a separate PR, perhaps one that also addresses the issues that @klshrinidhi and I ran into when making the hopper device example (replacing access to joints with getComponentList).

OpenSim/Simulation/Model/ModelComponent.cpp
```diff
@@ -124,8 +123,11 @@ void ModelComponent::addGeometry(OpenSim::Geometry& geom) {
         }
         
     }
-    append_geometry(geom);
-    return;
+
+    int ix = append_geometry(geom);
+    finalizeFromProperties();
```
chrisdembia: I think it's OK to require whoever called `addGeometry()` to have to finalize their model again. What if they were adding tons of geometry and thus only need to finalize once after they've added everything?

OpenSim/Simulation/Model/TwoFrameLinker.h
```diff
@@ -80,6 +80,20 @@ class TwoFrameLinker : public C {
         const std::string& frame1Name,
         const std::string& frame2Name);
 
+    /** Convenience Constructor with two Frames
+    Construct a TwoFrameLinker where the two frames are specified by name
+    and offset transforms on the respective frames.
```
chrisdembia: This doxygen was copy-pasted and is incorrect; the frames are not specified by name.

aymanhab:
Traversal question: If the component returns all other "internal components" in traversal then it's leaking its internals and can't be refactored in a different way, seems like some "scope" like rule is needed. where "internal" components (number, names, states, inputs, outputs) are not exposed to the outside world. Hard to tell if that already holds, though it appeared earlier in comments. Some clarification would be helpful.

OpenSim/Simulation/Model/TwoFrameLinker.h
```diff
@@ -252,6 +266,33 @@ TwoFrameLinker<C, F>::TwoFrameLinker(const std::string &name,
...
+    PhysicalOffsetFrame frame1Offset(frame1.getName() + "_offset",
+        frame1, transformInFrame1);
+
+    PhysicalOffsetFrame frame2Offset(frame2.getName() + "_offset",
+        frame2, transformInFrame2);
+
+    // Append the offset frames to the Joints internal list of frames
+    int ix1 = append_frames(frame1Offset);
+    int ix2 = append_frames(frame2Offset);
+    this->finalizeFromProperties();
```
chrisdembia: The doxygen for `Component::finalizeFromProperties()` should explain that it is expected to call `Component::finalizeFromProperties()` in scenarios like this. What specifically does this do that enables you to call the lines below?

OpenSim/Simulation/Model/TwoFrameLinker.h
chrisdembia: Can you add to the comment that this constructor will make intermediate `PhysicalOffsetFrames` and add them to the `frames` property?

chrisdembia:
> Traversal question: If the component returns all other "internal components" in traversal then it's leaking its internals and can't be refactored in a different way, seems like some "scope" like rule is needed. where "internal" components (number, names, states, inputs, outputs) are not exposed to the outside world. Hard to tell if that already holds, though it appeared earlier in comments. Some clarification would be helpful.

I don't think member subcomponents are intended to be "internal." They are just not serialized. That is, they are not configurable except by the component that owns them. But I think they should be exposed. Though all access to the member subcomponents from outside that component should be const.

OpenSim/Common/Component.h
chrisdembia: I think it would be clearer if this method took a path name instead of a component, then.

aymanhab:
Thanks, @chrisdembia Some clarification/doc would help because the comments about "internal" components gave me the connotation that they are "private".

chrisdembia:
> Thanks, @chrisdembia Some clarification/doc would help because the comments about "internal" components gave me the connotation that they are "private".

I agree; perhaps we should chat with @aseth1 when he gets in.

Once we are clear on the distinctions between the 3 types of subcomponents, I'll write up some doxygen.

OpenSim/Simulation/SimbodyEngine/Joint.cpp
```diff
@@ -84,14 +87,18 @@ Joint::Joint(const std::string &name,
     // from the root of the tree.
```
chrisdembia: Can this TODO be removed now?

OpenSim/Common/Component.cpp
aymanhab: The names that are auto-generated are not unique increasing the potential of downstream problems with bogus messages (e.g. Path_ component is not found or is not unique etc.) while the user has no such name in his/her model. Is that possible?

OpenSim/Common/Component.cpp
```diff
@@ -127,6 +127,24 @@ Component::Component(SimTK::Xml::Element& element) : Object(element)
...
+
+    OPENSIM_THROW_IF( getName().empty(), ComponentHasNoName,
+                      getConcreteClassName() );
+
+    for (auto& comp : _memberSubcomponents) {
+        comp->setParent(*this);
+    }
+    for (auto& comp : _adoptedSubcomponents) {
+        comp->setParent(*this);
+    }
```
chrisdembia: Add comment that the property subcomponents get their parent set elsewhere. When we move to a unified iterator, it should not be used here.

OpenSim/Common/Component.cpp
aymanhab: What are valid names? including "/" in a name would mess us up as far as I can tell, don't know about ".", "..", " ", "," etc. if used in component name. Would be good to have some error checks, generation of valid names in a central location.

chrisdembia: There's an open issue for this somewhere...

OpenSim/Simulation/SimbodyEngine/Joint.cpp
```diff
@@ -62,6 +62,9 @@ Joint::Joint(const std::string &name, const std::string& parentName,
                                       const std::string& childName,
                                       bool reverse) : Joint()
 {
+    OPENSIM_THROW_IF( name.empty(), ComponentHasNoName,
+                      getClassName());
+
```
chrisdembia: I think the elegant way to handle this is that the constructor has a signature

```
Joint::Joint(const ComponentPathName& name, const ComponentPathName& parent, const ComponentPathName& child)
```

and the constructor to `ComponentPathName` does all the validation.

chrisdembia: Nothing to act on for this PR though; just making a comment for a future PR.

chrisdembia:
Here are my main comments:

1. `Component::finalizeFromProperties()` should be documented explaining its seemingly new use case. In a future PR, we may revisit how this method is named, since how it is handling things that are not properties. It almost seems to make more sense to place the `setParent()` stuff in `connect()` in the future, since it feels more like connecting.

2. There should be a separate PR for a central naming mechanism, perhaps through `ComponentPathName`, that can validate paths and also provide absolute/relative paths.

3. There should be documentation at the top of Component.h explaining the 3 types of components. I'll give a crack at this.

OpenSim/Common/Component.cpp
tkuchida: See #596

OpenSim/Common/Component.cpp
aseth1: @aymanhab it is already possible to get useless messages when a component is unnamed. The individual component name does not have to be unique but its path name must be. So unless there is already a duplicate path having a non-unique names should be just fine. For example, all GeometyPaths were unnamed, and there is no reason to name it something <MuscleName>_geometry_path since the path name would be <ModelName>/<MuscleName>/geometry_path anyways. If you have a good scheme for unique naming let me know. Randomly generating a number to tack on or using time will cause de/serialization test failures were we are comparing to an old standard file that has no names, we have to guarantee the same names are autogenerated also when a nameless component is copied.

aseth1: @tkuchida the sucbomponents list is about ownership. It is an invalid tree if the child owns the parent or any other component in its ancestry. I will add a check so that this isn't attempted. It is a good point that a component to be adopted cannot already by a component in the tree. There should be an exception for that. Thanks.

OpenSim/Common/ComponentConnector.h
```diff
@@ -133,9 +134,18 @@ class OSIMCOMMON_API AbstractConnector : public Object {
     /** Disconnect this Connector from the connectee object */
     virtual void disconnect() = 0;
 
+protected:
+    const Component& getOwner() const { return _owner.getRef(); }
+    void setOwner(const Component& o) { _owner.reset(&o); }
```
chrisdembia: I'm adding an analogous method in Outputs to fix the output copying bug #809. However, I wrote it to take a `const Component*` instead of a reference. I think it's more natural that way, and results in less `&`'s and `*`'s. Whatever we do we should be consistent. So let me know what you think we should go with.

aseth1: From the users point of view I prefer the getter to return a const ref and then for consistency I would like the setter to take a const ref. I think most, if not all, of our get<something> methods return a const reference.

chrisdembia: OK I'll use references.

OpenSim/Common/Component.cpp
aseth1: @tkuchida Good question. The reason is that unlike the `_propertySubcomponents` the size of member and adopted are fixed (there is no unadopt/abondon subcomponent yet) so there is no concern that I will have a refrence to stale Component in a connector as could happen with a property component that could change or be removed. 

It wouldn't hurt disconnect them also since connect will be called and it disconnects them first before connecting.

aseth1: Thinking about it more, I think you are right and I should disconnect all subcomponents for completeness. Model building/editing in the GUI could require that all connectors of a Component (including its subs) must be disconnected before removing/deleting, for example.

OpenSim/Common/Component.cpp
aseth1: A Component  name "validater" is important, but will be for a subsequent PR. Adding it to the TODO list (#596).

OpenSim/Common/Component.cpp
aseth1: Yes, it is listed in #811

OpenSim/Common/Component.h
aseth1: Yes definitely. It does not do a find anymore. I will update the comment.

OpenSim/Common/Component.cpp
aymanhab: @aseth1 True that we get bad messages now, and that's why we're doing this. Already an improvement. GeometryPath is special in that it's unnamed single object, what I ran into with Geometry is that you can have multiple unnamed Mesh(es) or Sphere(s) so the class name doesn't do the trick (don't know if possible else where). ModelComponent::addGeometry generates a unique new name (was done pre naming convention).
We can open a separate issue for naming and move past this particular issue.

aseth1:
PR builds are passing. @chrisdembia @aymanhab please merge unless there are outstanding issues that have not been addressed and are not in #811.

Issue #393, opened 2015-03-12
Debug test failures


chrisdembia:
- [x] testMomentArms
- [x] testInitState
- [x] testPrescribedForce #304 
- [x] testAssemblySolver
- [x] testModelCopy
- [x] testComponents
- [x] testIterators

Tested on Ubuntu 14.04 clang 3.4.

Timeouts:
- testOptimizationExampleRuns (testOptimizationExample also fails, likely b/c this one times out).
- testCMCArm26
- testCMCSingleMuscle
- testCMCSingleRigidTendonMuscle

_Note added by tkuchida:_
See [travis.yml L128](https://github.com/opensim-org/opensim-core/blob/master/.travis.yml#L128) for list of excluded tests.

jimmyDunne:
Break up and start going through the list.

aymanhab:
As of 08/10, on windows x64 and VS 2013 Debug builds
testMomentArms passes
testInitState passes
testPrescribedForce passes
testAssemblySolver fails
testModelCopy passes
testComponents passes

Not clear if the fixes were intentional and work cross-platform or coincidental!

chrisdembia:
Woohoo! We should fix all our bugs this way..just wait :)

jenhicks:
Can we close?

chrisdembia:
I think we should leave open. There are still a number of debug failures the last time I checked. At least now we are running the debug tests on travis, so we won't get *new* failures.

jenhicks:
@aymanhab Can you update the list to the latest?

chrisdembia:
We run the debug tests on travis now, and I have omitted the ones that fail or that take too long in debug (https://github.com/opensim-org/opensim-core/blob/master/.travis.yml#L125). The list could still be outdated though.

aymanhab:
Updated list on Windows 7 Pro x64, VS 2015 community on 3/7/2016:
The following tests FAILED:
1>  	 14 - testAssemblySolver (Failed)
1>  	 15 - testContactGeometry (Timeout)
1>  	 38 - testOptimizationExampleRuns (Timeout)
1>  	 39 - testOptimizationExample (Failed)
1>  	 49 - testWrapping (Timeout)
1>  	 50 - testAnalyzeTutorialOne (Timeout)
1>  	 51 - testInducedAccelerations (Failed)
1>  	 53 - testStaticOptimization (Failed)
1>  	 58 - testCMCArm26 (Timeout)
[...]

PR #812, opened 2016-03-08
Fix testAssemblySolver failure in debug due to incorrect sequence of calls.

aseth1:
We cannot modify properties on the model and then run a simulation. Property edits must be finalized and have `initSystem` called. Also halved the simulation time to speed up the test.


aseth1:
@chrisdembia @aymanhab this should address the fourth check box in the list: #393

chrisdembia:
Looks good to me.

aymanhab:
Looks good, merging.

PR #814, opened 2016-03-09
Allow Output functions to be used after copying a component.

### LINKED ISSUE #809: When copying a component, the functions within outputs point to the original component
One possible solution might be possible through this `std::mem_fn()` thing I just found: http://en.cppreference.com/w/cpp/utility/functional/mem_fn

Basically, if the output function is a member function, we won't embed the `this` pointer into the output; we'll instead use `mem_fn` and pass the appropriate `this` pointer only when you actually call the function. The only hurdle I foresee is knowing how to downcast to the proper concrete type.

@klshrinidhi , maybe we can look at this together tomorrow?

chrisdembia:
Fixes #809.

Outputs cannot be used until after finalizeFromProperties() has been called. Let me know if there is a better place for the `setOwner()` calls.

I also removed the OpenSim_DECLARE_OUTPUT_FLEX macro.

- [x] use default template argument for `CompType`
- [x] rename copytesting2
- [x] enable the commented-out exception.


chrisdembia:
Thank you to @klshrinidhi for helping me figure out how to do this.

OpenSim/Common/Component.h
```diff
@@ -1764,66 +1760,42 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
+
+        You can ask outputs for their value only after you call
+        `finalizeFromProperties()`.
 
        @see constructOutputForStateVariable()
      */
 
 #ifndef SWIG // SWIG can't parse the const at the end of the second argument.
-    template <typename T, typename Class>
+    template <typename T, typename CompType>
```
aseth1: can default ComptType = Component which also serves to document expected types

chrisdembia:
Ready for review again.

aseth1:
Looks great!

chrisdembia:
Thank you for merging.

chrisdembia:
And thank you again to @klshrinidhi.

PR #818, opened 2016-03-10
[appveyor] Remove treat warnings as errors.

chrisdembia:
This PR tries to fix the weird failures on AppVeyor. I'm not sure yet if I found a solution to the problem.


chrisdembia:
I'm merging this in case this helps fix the appveyor errors.

PR #815, opened 2016-03-09
Create macro for specifying inputs.

chrisdembia:
This PR moves the interface for specifying inputs from `constructInputs()` to the header file, and uses a macro to document the input. Component implementers can now also provide comments for their inputs.

PR #830, opened 2016-03-14
List inputs and outputs via channels.

chrisdembia:
This PR introduces list inputs and outputs via channels.

I had taken a shot at doing list connectors as well, but I undid those changes. They are in the separate branch (I'll post information about that separately).

See **`futureIKListOutputs.cpp`** for the most complete example of how list inputs and outputs could work together. This example includes an "IK" component that takes inputs from a variety of sources, and also has a variety of reporters, etc. The review should start from this file.

That example made clear the need for "vector" outputs, and I had hoped to implement this but ran out of time. I have thought about this a lot though, and you can see some of those thoughts in `sandboxIKVectorOutputs.cpp`. As such, the design in this PR may need to be revisited.

This PR includes changes from PR #815, so it might make sense to merge that PR first. That PR is fairly small though.

The diff makes it look like I destroyed testComponentInterface(), but rest assured that I did not destroy this test. I simply rearranged stuff and added some tests for list inputs and this messes up the diff.


chrisdembia:
Next tasks following this PR are making sure the inputs and outputs are wrapped and usable from wrapping, and creating a proper DataSource component (and wrapping it). @klshrinidhi, some of this may be well-suited to you.

chrisdembia:
Fixed merge conflict.

aseth1:
I am likely to experiment with some changes, but I think it is a great place to start and it already works!
In it goes ...

chrisdembia:
Please do experiment. I think there's lots of room for improvement. Thanks for merging.

PR #834, opened 2016-03-15
Suppress clang warning about friend declaration.

chrisdembia:
@klshrinidhi could you review?

### DIFF
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -1572,8 +1572,16 @@ protected:
 
     // End of System Creation and Access Methods.
 
+#if !defined(SWIG) && defined(__clang__)
+#pragma clang diagnostic push
+    // clang says this friend declaration is invalid but it's not.
+#pragma clang diagnostic ignored "-Wunsupported-friend"
+#endif
     template<class C>
     friend void Connector<C>::findAndConnect(const Component& root);
+#if !defined(SWIG) && defined(__clang__)
+#pragma clang diagnostic pop
+#endif
 
     /** Utility method to find a component in the list of sub components of this
         component and any of their sub components, etc..., by name or state variable name.


klshrinidhi:
Is it possible to set compiler flags when compiling this file ?
It would be good to avoid compiler specific code if possible.

klshrinidhi:
`unsupported-friend` sounds like clang is not supportive of friends 😯 :laughing:

chrisdembia:
Yes, but it would be nice to keep the suppression as local as possible, especially since this header uses friend declarations in other places as well. It is kinda ugly as-is though...


> unsupported-friend sounds like clang is not supportive of friends 😯

hahah yeah

klshrinidhi:
I compiled the master with clang-3.8 (development build) to the see what exactly the warning is.
```
opensim-core/OpenSim/Common/Component.h:1633:31: warning: dependent nested name specifier 'Connector<C>::' for friend class declaration is not supported; turning off access control for 'Component' [-Wunsupported-friend]
    friend void Connector<C>::findAndConnect(const Component& root);
                ~~~~~~~~~~~~~~^
1 warning generated.
```
I think it is okay to leave the warning un-suppressed. The code required to suppress it looks bit of an overkill. 
If it were an error and not a warning, this would have been necessary. But I think this is okay for now. It will go away with newer versions of clang.

klshrinidhi:
@chrisdembia 
As I am working on getting rid of warnings, I am seeing this one appear over and over again as `Component.h` is used everywhere. I now think this PR should go in. To me, the benefit of not having to see that many warnings outweighs ugly code in `Component.h`.
I will look at this PR again.

chrisdembia:
Okay cool. You can go ahead and merge this then.

chrisdembia:
I'm a little preoccupied at the moment, so if you want to make changes to this PR, could you take care of them?

klshrinidhi:
No problem. I am actually using this in my PR about removing warnings.

Issue #824, opened 2016-03-11
Allow creating outputs for functions that return T by const ref instead of by value


chrisdembia:
See #796

aseth1:
I am not sure I agree, lets discuss why this should be necessary.

chrisdembia:
Okay; I'm labelling as "needs discussion."

aymanhab:
Essentially most methods that retrieve values from the Cache return values by const reference (e.g. Frame::getGroundTransform). Requiring the function to return by value will lead to duplicate methods and confusion to users (as to why) unless there's a fundamental reason why const references need to be excluded I strongly recommend allowing this route for specifying outputs (in addition to avoiding an extra copy).

aseth1:
@aymanhab The issue in my mind wasn't to duplicate the method (that was never an option) but to have the Output make a local private copy (using stages to validate it) instead of hanging on to a potentially stale reference. If the Output value is a direct reference to that returned by the bound member function we have to have guarantees that it will be around at reporting time (and earlier of course). This is likely to be true for virtually anything cached in the state - but nothing else. If component writers think they have to return a const reference for any reason (thinking it will be faster for example) they are going to mess it up, guaranteed. I am not a fan of the argument "well you can do anything in C++ to violate the intent" because it was that kind of reasoning that lead us to properties that were brittle and impossible to debug. And if values are not carefully cached and managed by the state then it is likely that the values will become invalid (even if the reference is fine) but there would be no way to know that! That is not a problem I would want to debug! 

I want Outputs to continue to bind to member functions that return by value. To accommodate those that return a reference I still prefer to make a local copy, but I am also OK with adding a (secret) alternate option that short-circuits directly to the reference returned from trusted members (i.e. most Simbody methods fishing values out of the cache). If there was a systematic way we could test if the referenced value were contained in the passed in state (i.e. state cache), that would be ideal, then I would make reference option on par with the value.

Let me know if that makes sense.

aymanhab:
@aseth1 I understand the sentiment. My thought is that the easiest way to avoid developers making mistakes is to have a clean/simple mental picture. 

In my mind, the simplest such picture is a plain "wire" that does no caching whatsoever. Once you have caches (e.g. to convert from const_ref to value), the picture becomes muddy and is more error prone (why caching?, when to invalidate?, what if I cascade them? etc.). The wire picture also has the benefit that we don't need to invent/document-and-explain new semantics, but instead rely on the semantics of the language directly.

[...]

PR #839, opened 2016-03-16
Allow outputs functions to return const T&.

chrisdembia:
This addresses #824 for the Hackathon, and must be revisited afterwards. This change is in accord with what I discussed with @aseth1 and includes warnings in Doxygen to scare people away from using this variant of the Output.

### DIFF
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -917,11 +917,12 @@ setDiscreteVariableValue(SimTK::State& s, const std::string& name, double value)
 
 bool Component::constructOutputForStateVariable(const std::string& name)
 {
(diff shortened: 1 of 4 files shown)


aymanhab:
Looks fine but I'm not very familiar with the gymnastics involved, can you review and merge @klshrinidhi ? Thanks.

chrisdembia:
@aseth1 I added a comment to the code. I suggest we merge it before the tests pass, since I only added a comment.

I also added our discussion to #824 (the associated issue) to keep track of what we must do after the hackathon.

aymanhab:
Great, thanks @aseth1 & @chrisdembia

chrisdembia:
Thank you @aseth1

PR #840, opened 2016-03-16
Comment out findAndConnect message for the hackathon.

### LINKED ISSUE #828: updateFromXMLNode() can use relative paths to connectees to avoid invoking the search for the components.
I just discussed this with @aseth1 . Since we know the structure of pre-v4.0 models, we can edit the updateFromXMLNode() code to put in relative paths for `connectee_name`. This will avert the backwards-compatible searching for the component by name.

chrisdembia:
This message is very good to have but prints too much output. After the hackathon, if we resolve #828, then this message will no longer clutter test output, and we can re-enable the message.

### DIFF
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -2344,7 +2344,7 @@ void Connector<C>::findAndConnect(const Component& root) {
         }
     }
     catch (const ComponentNotFoundOnSpecifiedPath& ex) {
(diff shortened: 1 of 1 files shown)


chrisdembia:
Would anyone like to merge this?

aseth1:
Sure!

chrisdembia:
Thank you.

Issue #838, opened 2016-03-15
Remaining Hackathon TODO


chrisdembia:
This issue does not include tasks related to the didactic example.

- [x] Reporter: @aseth1 hands off this afternoon to @chrisdembia @klshrinidhi (PR #843)
- [x] Outputs for certain components: @aymanhab (PR #796)
- [x] Outputs for return by `const T&`: @chrisdembia @aymanhab (PR #839)
- [ ] TableSource: @klshrinidhi 
- [x] Adapter wrapping: @klshrinidhi 
- [ ] Wrapping ComponentList, Output, Input: @chrisdembia @aymanhab 
- [ ] PR StatesCollector review @aseth1 @klshrinidhi 
- [ ] PR StatesTrajectory export to DataTable @aseth1 @klshrinidhi 
- [ ] Simulator class: @aseth1 
- [ ] GUI build: @aymanhab (btw you can use AppVeyor's automated build).
- [ ] Adding Components to ModelComponents @aseth1

chrisdembia:
Adding an item above: allowing an easy way to add Components to a Model (Models can only contain ModelComponents).

aymanhab:
@chrisdembia Do you have a branch for item 5 above (Wrapping lists) that we discussed yesterday? Would be good to get  PR #858 in first so we can try some examples. Tests still running.

chrisdembia:
> @chrisdembia Do you have a branch for item 5 above (Wrapping lists) that we discussed yesterday?

I'm going to work on it later today.

chrisdembia:
@aymanhab with regard to wrapping component lists, outputs, inputs, what's of the highest priority for you? Especially given @jenhicks's reminder that people aren't necessarily using wrapping for their hackathon projects? I still plan to work on wrapping, but perhaps just not as completely.

aymanhab:
@chrisdembia I wanted first to rename the methods per our conversations (BodiesList to BodyList etc.), add test case to exercise them, and then try to show output list in the GUI for the hackathon, which will in theory make it possible to use outputs in Python/Java/Matlab. Just wasn't sure if you started down this road of implementing the plan/doc we discussed yesterday.

chrisdembia:
I have not started down the road yet.

chrisdembia:
Well, sorry, yes I have started, but I'm not ready to show anything. Please just go ahead.

chrisdembia:
[...]

PR #843, opened 2016-03-16
Reporter components and list Connector/Input enhancements

aseth1:
This PR defines an `AbstractReporter` and two concrete reporters: `TableReporter` and `ConsoleReporter` as Components. Addresses the first point in #838.

Input (which are now all list connectors with Connectees to Output::Channels) was updated so that the path name to the channel is stored in a now private `connectee_name` (now a list) property, so that they Input connections can now be serialized (like any Connector). 

Examples of using the Reporters are in `testComponentInterface` and `futureMuscleMetabolicsResponse` (in Sandbox). 

These should be usable for the Hackathon where-after they can be revisited to improve error handling and documentation. 

@chrisdembia @klshrinidhi please review for any blatant bugs or misuse (e.g. of DataTables).


OpenSim/Common/Reporter.h
```diff
@@ -0,0 +1,252 @@
...
+        return _outputTable;
+    }
+
+protected:
+    void implementReport(const SimTK::State& state) const override {
+        const auto& input = getInput<InputT>("inputs");
+        SimTK::Vector_<ValueT> result(int(input.getNumConnectees()));
+
+        for (auto idx = 0; idx < input.getNumConnectees(); ++idx) {
+              const auto& chan = Input<InputT>::downcast(input).getChannel(idx);
```
chrisdembia: No need to downcast; the `input` is already downcasted.

OpenSim/Common/Reporter.h
```diff
@@ -0,0 +1,252 @@
...
+
+    OpenSim_DECLARE_PROPERTY(report_time_interval, double,
+        "The recording time interval (s). If interval < 0 or NaN defaults to"
+        "every valid integration time step.");
+
+//=============================================================================
+// PUBLIC METHODS
+//=============================================================================
+    /** Report values given the state and top-level Component (e.g. Model) */
+    void report(const SimTK::State& s) const;
```
chrisdembia: It's not clear to me that this method is necessary. But perhaps it's useful if someone is writing their own homebrewed solver in Python or MATLAB.

OpenSim/Common/Reporter.h
```diff
@@ -0,0 +1,252 @@
...
+            }
+            labels.push_back(label);
+        }
+        _outputTable.setColumnLabels(labels);
+    }
+
+private:
+
+    // Hold the output values in a table with values as columns and time rows
+    mutable TimeSeriesTable_<ValueT> _outputTable;
```
chrisdembia: I prefer to just const-cast this at the time it's needed. Otherwise, it's too easy to modify in member functions that *shouldn't* modify it.

chrisdembia: I'll also add a comment describing why it's okay for this to be written to in a const method.

OpenSim/Common/Reporter.cpp
```diff
@@ -0,0 +1,120 @@
...
+    constructProperty_report_time_interval(0.0);
+}
+
+// Create an underlying SimTK::Reporter to represent the OpenSim::Reporter in the 
+// computational system.  Create a SimTK::Reporter::Custom by default.
+void AbstractReporter::extendAddToSystem(SimTK::MultibodySystem& system) const
+{
+    Super::extendAddToSystem(system);
+    if (get_is_disabled())
+        return;
```
chrisdembia: I think this check should maybe also occur in `extendRealizeReport()`. As is, the reporter will still spew stuff out if not using a reporting interval. This can then be revisited once we are using a proper Event to handle the reporting at each successful step.

OpenSim/Common/Reporter.h
```diff
@@ -0,0 +1,252 @@
...
+};
+
+
+template <typename T>
+class ConsoleReporter : public Reporter<T> {
+    OpenSim_DECLARE_CONCRETE_OBJECT_T(ConsoleReporter, T, Reporter<T>);
```
chrisdembia: I'll add some documentation for this class.

OpenSim/Common/ComponentConnector.h
```diff
@@ -354,12 +374,27 @@ class  Input : public AbstractInput {
...
+                std::string pathName =
+                    output.getOwner().getRelativePathName(getOwner());
+                pathName = pathName + "/" + chan.second.getChannelName();
+
+                // set the connectee name so that the connection can be
+                // serialized
+                if (ix < getNumConnectees())
+                    setConnecteeName(pathName, ix);
+                else
+                    appendConnecteeName(pathName);
```
chrisdembia: Could maybe get rid of this if statement and just append.

OpenSim/Common/Reporter.cpp
aseth1: I prefer to get rid of is_disabled altogether. It makes the logic more complex for no reason and we if we cannot get handling the flag correctly, I don't expect others to do any better.

chrisdembia: In discussion with @aseth1 ; I'll remove this.

chrisdembia:
Okay I made a bunch of mostly inconsequential changes. I think this can be merged once the tests pass.

chrisdembia:
*uncontroversial (not inconsequential).

aseth1:
Thank you @chrisdembia !

PR #847, opened 2016-03-17
WITH_BTK=ON on Appveyor and Travis

klshrinidhi:


chrisdembia:
Unfortunately you should not use multi-line continuation stuff in those yml files. https://github.com/travis-ci/travis-ci/issues/497.

There are multiple subtle problems with splitting a command into multiple lines.

chrisdembia:
Looks like we should have enabled the appveyor build! Appveyor is failing due to issues with Eigen when compiling osimJavaJNI.

PR #851, opened 2016-03-17
Add outputs

aymanhab:
Add Outputs to Frame and PathSpring


OpenSim/Simulation/Model/Frame.h
```diff
@@ -185,6 +193,11 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Frame, ModelComponent);
     */
     SimTK::Transform findTransformInBaseFrame() const;
 
+    /** Accessor for position of Frame in Ground frame as output. */
```
chrisdembia: The comment should clarify that this means the origin of the frame. Also, Station has a method called `findLocation...`. But Model also already has a `com_position` output. After the hackathon, we should choose one and be consistent.

OpenSim/Tests/Components/testComponents.cpp
```diff
@@ -245,7 +245,7 @@ void testComponent(const Component& instanceToTest)
         {
             model.getSystem().realize(state,
                     thisOutput->getDependsOnStage().prev());
-            ASSERT_THROW(Exception,
+            ASSERT_THROW(SimTK::Exception::StageTooLow,
```
chrisdembia: :+1:

OpenSim/Simulation/Model/Frame.h
chrisdembia: Also, this method itself doesn't need to mention anything about "output";

aymanhab:
Appveyor failure is spurious unrelated. Restarting.

OpenSim/Simulation/Model/Frame.h
aymanhab: Agreed @chrisdembia on the need to unify names :+1:
I left the reference to Output due to the discussion regarding what's allowed as output function, but I'm happy to remove it now.
If Output functions will change semantics to mean more than a plain wire/bind then we should give them a distinct name and include comment to signify that.

aymanhab:
Changed comment per @chrisdembia Please merge when tests pass. Thanks.

OpenSim/Simulation/Model/GeometryPath.h
```diff
@@ -57,6 +57,12 @@ class PointForceDirection;
  */
 class OSIMSIMULATION_API GeometryPath : public ModelComponent {
 OpenSim_DECLARE_CONCRETE_OBJECT(GeometryPath, ModelComponent);
+    //=============================================================================
+    // OUTPUTS
+    //=============================================================================
+    OpenSim_DECLARE_OUTPUT(length, double, getLength, SimTK::Stage::Position);
+    // 
+    OpenSim_DECLARE_OUTPUT(speed, double, getLengtheningSpeed, SimTK::Stage::Velocity);
```
chrisdembia: I prefer lengthening_speed, for consistency with PathSpring, and also because it's reasonable for people to think it's shortening speed otherwise.

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -112,6 +112,11 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
         "to the Joint. When the joint is delete so are the Frames in this list.");
 
 //=============================================================================
+// OUTPUTS
+//=============================================================================
+    OpenSim_DECLARE_OUTPUT(power, double, calcPower, SimTK::Stage::Acceleration);
```
chrisdembia: This should be available at Dynamics; @aseth1 what do you think?

OpenSim/Simulation/Model/Actuator.h
```diff
@@ -55,6 +55,10 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Actuator, Force);
 //=============================================================================
 // NO PROPERTIES
 //=============================================================================
+//=============================================================================
+// OUTPUTS
+//=============================================================================
+OpenSim_DECLARE_OUTPUT(power, double, getPower, SimTK::Stage::Velocity);
```
chrisdembia: @aseth1 should this be the Power stage?

aseth1: this should be available after forces are realized in dynamics

aymanhab:
After feedback on proper Stage name above by @aseth1 I only have joint loads to add AFAIK.

OpenSim/Simulation/SimbodyEngine/Joint.h
aseth1: Comment says acceleration because it needs to compute constraint forces, which well depend on applied forces. The constraint multipliers are solved concurrently witch acceleration so I think this is correct.

chrisdembia:
@aymanhab feel free to merge this.

OpenSim/Simulation/Model/Model.h
```diff
@@ -761,7 +768,12 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Model, ModelComponent);
     SimTK::Vec3 calcMassCenterPosition(const SimTK::State &s) const;
     SimTK::Vec3 calcMassCenterVelocity(const SimTK::State &s) const;
     SimTK::Vec3 calcMassCenterAcceleration(const SimTK::State &s) const;
-
+    double calcKineticEnergy(const SimTK::State &s) const {
+        return getMultibodySystem().calcKineticEnergy(s);
+    }    
+    double calcPotentialEnergy(const SimTK::State &s) const {
+        return getMultibodySystem().calcPotentialEnergy(s);
+    }
```
chrisdembia: If you're planning on adding more changes to this branch anyway, could you add comments here simply saying that these quantities are for the total system?

aymanhab:
Thanks @chrisdembia Will merge now but will keep the branch for the last set of outputs (that I'm still working on)

OpenSim/Simulation/Model/Model.h
aymanhab: Ok, thanks @chrisdembia will add in the next commit.

PR #853, opened 2016-03-17
Hide btk headers from interface.

chrisdembia:
Hoping this will fix appveyor build.

@klshrinidhi is there a reason why the btk headers needed to be in the header of the C3D adapter? I think it was causing the Java wrapping c++ files to not compile on Appveyor.

### DIFF
--- a/OpenSim/Common/C3DFileAdapter.cpp
+++ b/OpenSim/Common/C3DFileAdapter.cpp
@@ -1,5 +1,10 @@
 #include "C3DFileAdapter.h"
 
+#include "btkAcquisitionFileReader.h"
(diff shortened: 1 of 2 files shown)


klshrinidhi:
:+1: 

It does solve the problem of `Eigen` being ambiguous. BTK uses Eigen library which is under namespace `Eigen`. Simbody has a class `SimTK::Eigen`. These two collide when `using namespace SimTK` is used. In future, we need to eliminate `using namespace` declarations to avoid these problems.

Can't merge until CI builds say :green_heart:.

chrisdembia:
Appears that these headers are needed for the C3D example. I'm closing this PR.

PR #855, opened 2016-03-17
Update powershell instructions with wrapping

klshrinidhi:
Windows Powershell instructions currently do not include wrapping but instructions for OS X and Ubuntu do include wrapping. Adding it to Windows too for consistency.


chrisdembia:
That's neat that there's syntax highlighting for powershell.

PR #856, opened 2016-03-17
Update Simbody commit id for superbuild

klshrinidhi:


aymanhab:
Awesome, thanks @klshrinidhi @chrisdembia

PR #857, opened 2016-03-18
Include property list of components in component

aseth1:
This allows API users/model builders to tack on any other component as a subcomponent. 
That means that Matt can add frames to a body, we can add reporters to models, add a device to a model, etc...

Eventually we will need to get the multibody tree building code in Model to traverse ComponentLists for Bodies and Joints that could be nested below in other subcomponents, instead of just those in the Model level sets.

### DIFF
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -111,17 +111,20 @@ private:
 Component::Component() : Object()
 {
     constructProperty_connectors();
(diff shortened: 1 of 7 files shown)


chrisdembia:
@aseth1 don't forget to open an issue regarding catching properties with the same name down the inheritance hierarchy.

chrisdembia:
Nevermind, I just created the issue; #859 .

PR #860, opened 2016-03-18
Create luxo example

aymanhab:
Example model from Matt. Python wrapping temporarily broken... fixing.


chrisdembia:
I think the failure on travis is unrelated and is with the java bindings. Why do you think there is an issue with python wrapping? I think this is good to go once some of the "PR" tests pass.

aseth1:
@aymanhab @msdemers when I synced to your branch (before this PR) the `ExampleLuxoMuscle` project did not appear in my VS2015 solution. I cleared my cache and ran CMake again to check.
It was late so I did not investigate further. Anyone else have that issue?

Also, can you commit a version of the model that I can grab.

aymanhab:
@aseth1 I'll merge the branch so we can all see the same thing and since it's a standalone. @chrisdembia you're right the failure is unrelated.

chrisdembia:
Based on the CMake files, the example is not part of the OpenSim solution. You should "install" OpenSim, then build the example from the installation.

msdemers:
@aseth1 @chrisdembia you can create a build environment directly from the luxo example directory in the source.  You do not need to INSTALL to use the example and I DO NOT recommend it.  If you build directly from the directory within OpenSim's source example folder, your changes will be reflected in the git diffs. Not true if you build and develop from the installation.

chrisdembia:
Ah good point...sorry about that.

msdemers:
Also, I intentionally left this example out of the set of tested examples because the tested examples are only a subset of all examples. Unfortunately, it looks like the tested examples are simply duplicate copies of the same code in the broader examples directory. This is unfortunate, because it means the same code must be maintained in two places.

chrisdembia:
The source code for the examples is not duplicated. There are two CMakeLists.txt for those examples: one to build the example as part of the OpenSim solution/project, and one to be used when using the example as part of the distribution. Same source code both times.

aymanhab:
At this point we should probably delete the old branch since it was way behind, and create a new branch to do the fixes on, for CMake etc.

msdemers:
@chrisdembia I'm looking at the source and see multiple examples, including their cpp files, in both OpenSim/Examples and OpenSim/Tests. For example, ExampleMain, MuscleExample and ControllerExample have source in both places. Looks like some duplication to me.  Am I missing something?

msdemers:
@aymanhab I'm fine with deleting or keeping.  It's up to date with master as of 2 pm yesterday, even though I started it a while ago. Shouldn't be way behind at this point.

msdemers:
@chrisdembia Never mind. I see how it's working now.

aymanhab:
Thanks @msdemers I didn't know when you merged from master  but as of now it shows 17 commits behind. I mainly want to make sure we're all looking at the same version of the code since switching environments and building takes a long time on my (now old) laptop.

msdemers:
wow! so many commit's since yesterday afternoon. Awesome!

aseth1:
@msdemers can you add two PhysicalFrames (can be Offsets) to the hopper to denote the attachment sites for the device. Thanks!

msdemers:
sure thing. added to the model level bin of frames? is that working now?
wasn't when we talked a few days ago.

aseth1:
@msdemers yes, it should be working now. I'll try it if you added the frames.

msdemers:
k. making the change on a new branch

aseth1:
great. if you can give me the names I can prepare the device to attach automatically.

msdemers:
they are in the PR I just submitted on the new branch.

name: ”device_superior_attachment" attaches to the head body

name: "device_inferior_attachment" attaches to the bottom_bracket

chrisdembia:
I deleted the create_luxo_example branch; that's OK right? It's been merged.

msdemers:
Totally cool. I made a new branch for he latest and future changes.

aymanhab:
Right, thanks @chrisdembia

aseth1:
The model runs excruciatingly slow in in debug - that's unfortunate since it does serve as simple model for debug their stuff.

PR #861, opened 2016-03-18
Add Joint reaction output and fix doxygen per PR

aymanhab:
Remaining outputs for Joint per #851 

Please review the names of the methods used from Simbody since Joint has Child & Parent frames and the SimTK::MobilizedBody refers to Body & Child. I assumed we get MobilizedBody from Child frame.

### DIFF
--- a/OpenSim/Simulation/SimbodyEngine/Joint.h
+++ b/OpenSim/Simulation/SimbodyEngine/Joint.h
@@ -115,6 +115,10 @@ public:
 // OUTPUTS
 //=============================================================================
     OpenSim_DECLARE_OUTPUT(power, double, calcPower, SimTK::Stage::Acceleration);
+    OpenSim_DECLARE_OUTPUT(reaction_child_origin_in_ground, SimTK::SpatialVec, 
+        calcReactionOnChildAtOriginInGround, SimTK::Stage::Acceleration);
+    OpenSim_DECLARE_OUTPUT(reaction_parent_origin_in_ground, SimTK::SpatialVec, 
+        calcReactionOnParentAtOriginInGround, SimTK::Stage::Acceleration);
 
 //=============================================================================
 // METHODS
@@ -250,8 +254,19 @@ public:
     calcEquivalentSpatialForce(const SimTK::State &s, 
                                const SimTK::Vector &mobilityForces) const;
 
+    // Reaction forces (direct calls into Simbody methods)
+    /** Reaction forces on Child Body at Origin in Ground frame @see SimTK::MobilizedBody method. */
+    SimTK::SpatialVec
+        calcReactionOnChildAtOriginInGround(const SimTK::State &s) const {
+        return getChildFrame().getMobilizedBody().findMobilizerReactionOnBodyAtOriginInGround(s);
+    }
+    /** Reaction forces on Parent Body at Origin in Ground frame @see SimTK::MobilizedBody method. */
+    SimTK::SpatialVec
+        calcReactionOnParentAtOriginInGround(const SimTK::State &s) const {
+        return getChildFrame().getMobilizedBody().findMobilizerReactionOnParentAtOriginInGround(s);
+    }
 
-    /** Joints in general do not contribute power since the reaction space 
+    /** Joints in general do not contribute power since the reaction space
         forces are orthogonal to the mobility space. However, when joint motion 
         is prescribed, the internal forces that move the joint will do work. In 
         that case, the power is non-zero and the supplied SimTK::State
--- a/OpenSim/Simulation/SimbodyEngine/Joint.h
+++ b/OpenSim/Simulation/SimbodyEngine/Joint.h
@@ -115,10 +115,12 @@ public:
 // OUTPUTS
 //=============================================================================
     OpenSim_DECLARE_OUTPUT(power, double, calcPower, SimTK::Stage::Acceleration);
-    OpenSim_DECLARE_OUTPUT(reaction_child_origin_in_ground, SimTK::SpatialVec, 
-        calcReactionOnChildAtOriginInGround, SimTK::Stage::Acceleration);
-    OpenSim_DECLARE_OUTPUT(reaction_parent_origin_in_ground, SimTK::SpatialVec, 
-        calcReactionOnParentAtOriginInGround, SimTK::Stage::Acceleration);
+    // reaction_load_at_parent_frame
+    OpenSim_DECLARE_OUTPUT(reaction_load_at_parent_frame_in_ground_frame, SimTK::SpatialVec,
+        calcReactionAtParentFrameInGround, SimTK::Stage::Acceleration);
+    // reaction_load_at_child_frame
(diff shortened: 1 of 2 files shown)


OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -115,6 +115,10 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
 // OUTPUTS
 //=============================================================================
     OpenSim_DECLARE_OUTPUT(power, double, calcPower, SimTK::Stage::Acceleration);
+    OpenSim_DECLARE_OUTPUT(reaction_child_origin_in_ground, SimTK::SpatialVec,
```
msdemers: I'd recommend a name like reaction_load_on_child_at_origin_in_ground

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -115,6 +115,10 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
 // OUTPUTS
 //=============================================================================
     OpenSim_DECLARE_OUTPUT(power, double, calcPower, SimTK::Stage::Acceleration);
+    OpenSim_DECLARE_OUTPUT(reaction_child_origin_in_ground, SimTK::SpatialVec, 
+        calcReactionOnChildAtOriginInGround, SimTK::Stage::Acceleration);
+    OpenSim_DECLARE_OUTPUT(reaction_parent_origin_in_ground, SimTK::SpatialVec,
```
msdemers: again, something like reaction_load_on_parent_at_origin_in_ground

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -250,8 +254,19 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
     calcEquivalentSpatialForce(const SimTK::State &s, 
                                const SimTK::Vector &mobilityForces) const;
 
+    // Reaction forces (direct calls into Simbody methods)
+    /** Reaction forces on Child Body at Origin in Ground frame @see SimTK::MobilizedBody method. */
+    SimTK::SpatialVec
+        calcReactionOnChildAtOriginInGround(const SimTK::State &s) const {
+        return getChildFrame().getMobilizedBody().findMobilizerReactionOnBodyAtOriginInGround(s);
```
msdemers: looks correct to me

OpenSim/Simulation/SimbodyEngine/Joint.h
aymanhab: I'm perfectly fine with any name we agree upon, you probably have used these the most.

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -250,8 +254,19 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
...
+    }
+    /** Reaction forces on Parent Body at Origin in Ground frame @see SimTK::MobilizedBody method. */
+    SimTK::SpatialVec
+        calcReactionOnParentAtOriginInGround(const SimTK::State &s) const {
+        return getChildFrame().getMobilizedBody().findMobilizerReactionOnParentAtOriginInGround(s);
```
msdemers: yes, per your question in the comments section, this is the right approach. The child mobilized body matches the mobilizer of the joint. Asking it for it's child and parent reaction loads is appropriate.

msdemers:
The approach looks great and correct, but I'd recommend slight name changes. I'm worried that these two convenience outputs are less useful than the other conveniences simbody provides. Users will usually want the reaction load at the joint frames, not at the parent or child body origins.  I think this needs an output reaction_load_on_child_at_joint_frame_in_joint_frame which would use the very similar simbody method MobilizedBody::findMobilizerReactionOnBodyAtMInGround(const State& state) and re-express in the joint's physical frame on child. Also a similar one for the load on parent.

aymanhab:
@msdemers Thanks for the great feedback. As you saw it's very little overhead to add more outputs but it probably make sense to expose only the most commonly used 1 or 2 of the four variants for each. In summary, your suggestion:
- [ ]  reaction_load_on_parent_at_origin_in_ground 
- [ ] reaction_load_on_parent_at_joint_frame_in_joint_frame
and repeat for child (with corresponding output names). 

My follow up question is: Is the first output (at origin) useful at all or should we just do at joint frame? Thanks.

chrisdembia:
I understand almost nothing about joint reaction analysis in Simbody/OpenSim, but given that Joints now connect to physical frames, aren't the parent/child frames for the joint already the joint frames? Perhaps it's just the documentation of the new method calcReactionOnParentAtOriginInGround that should say "parent frame" instead of "parent body".

It's not clear to me what the "joint frame" is. Aren't there two joint frames, one on each of the two bodies? And aren't these called "parent frame" and "joint frame"?

chrisdembia:
(also, excuse my ignorance)

msdemers:
@aymanhab There are cases where you might want a reaction load at the MobilizeBbody origin, but usually only if you are adding forces back to the simbody system, which uses "body forces at origin" as it's language. most opensim users/devs shouldn't ever have to deal with that, but a few specialty dev task may require it.

msdemers:
@chrisdembia it's a really good question. Remember that simbody knows nothing about OpenSim::Frames. The parent/child joint frames are already at the joint, but these likely don't coincide with the origin of underlying MobilizedBody. The simbody methods we can use either express the joint reaction at the inboard or outboard joint frames, or at the parent or child origins of the MobilizedBodies.

For example:: Ayman's reaction_load_on_parent_at_origin_in_ground is curently returning the load computed at the parent MobilizedBody's origin, not the OpenSim::Joint::parent_frame origin. 

One nice answer could be to provide the outputs named this: 
reaction_load_at_parent_frame_in_ground_frame
reaction_load_at_parent_frame_in_parent_frame 


also the equivalent for child_frame

This would disambiguate the issue of PhysicalFrame vs Body, and hide the existence of MobilizedBody.  I think I like this option best. If a user wants to convert to an equivalent load at the underlying body origin, they can compute it. The load shifted to the body origin is rarely needed by the end users I've helped.

chrisdembia:
@msdemers your latest proposal makes sense to me. :+1: We can also have a component in the future that just transforms things between frames, alleviating the need for Joint to handle all the cases.

msdemers:
I like that plan

aymanhab:
@msdemers I added two outputs with the names you suggested, the other two are not available straight from Simbody as far as I can tell, so we can add them after the hackathon. Please let me know if I got these 2 correctly (considering all the parenting involved! Thanks.

msdemers:
@aymanhab looking good! Should be really useful.  I'll help make a test case for these and expand on these outputs later on.

aymanhab:
Awesome, thanks Reaction guru @msdemers Will merge when the tests pass.

PR #862, opened 2016-03-18
Added physical frames for attaching actuator

msdemers:
@aseth1 here you go! Added two physical frames for attaching an assistive device actuator.

name: ”device_superior_attachment" attaches to the head body

name: "device_inferior_attachment" attaches to the bottom_bracket


aseth1:
Thanks!

PR #817, opened 2016-03-10
Added StatesCollector class

chrisdembia:
This is a new-style reporter that accumulates states during a simulation. The Collector holds onto a StatesTrajectory that someone could access at the end of a simulation.

The StatesTrajectory within the collector is cleared when realizing to Instance.

I chose the name StatesCollector instead of StatesReporter because we already have a StatesReporter analysis.

PR #863, opened 2016-03-18
StatesAccumulator

chrisdembia:
This is a redo of #817. I addressed @tkuchida's feedback and renamed the StatesCollector to StatesAccumulator. It now uses the AbstractReporter class, and can thus report states at a constant reporting interval.


chrisdembia:
IMO, it would be useful to have this for the Hackathon so that people can visualize their simulations in the GUI.

OpenSim/Simulation/StatesAccumulator.h
```diff
@@ -0,0 +1,65 @@
...
+ * This class was introduced in v4.0 and is intended to replace the
+ * StatesReporter analysis.
+ *
+ * @ingroup reporters
+ */
+class OSIMSIMULATION_API StatesAccumulator : public AbstractReporter {
+OpenSim_DECLARE_CONCRETE_OBJECT(StatesAccumulator, AbstractReporter);
+
+public:
+    /** Access the recorded states. */
```
tkuchida: "recorded" -> "accumulated" for consistency with class name (here and elsewhere)?

tkuchida:
`</review>` Just the one minor suggestion. Looks good :+1:

chrisdembia:
d'oh you made that comment last time..sorry for forgetting about it.

chrisdembia:
This is ready for review again.

OpenSim/Simulation/StatesAccumulator.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+void StatesAccumulator::clear() {
+    m_states.clear();
+}
+
+const StatesTrajectory& StatesAccumulator::getStates() const {
+    return m_states;
+}
+
+/*
```
tkuchida: Might we need this method in the future? If not, perhaps delete it?

tkuchida: Oh, nevermind, I just saw the comment below.

tkuchida:
:+1: Looks good :penguin:

chrisdembia:
Thank you for reviewing. @aseth1, would you like to review this also?

OpenSim/Simulation/StatesAccumulator.cpp
```diff
@@ -0,0 +1,46 @@
...
+void StatesAccumulator::extendRealizeInstance(const SimTK::State& state) const {
+    Super::extendRealizeInstance(state);
+    const_cast<StatesAccumulator*>(this)->clear();
+}
+*/
+
+void StatesAccumulator::implementReport(const SimTK::State& state) const {
+    const_cast<StatesAccumulator*>(this)->m_states.append(state);
```
aseth1: Love to see it is so easy to implement a reporter! As and aside I am not convinced that `const_cast` is  better than using `mutable` especially for a private variable. There are other implications as well, see [here](http://stackoverflow.com/questions/27783236/const-castthis-performance-hit).

aseth1:
Other than a side comment, it looks good to me.

aseth1:
I think the correct name is `StatesReporter` and not `StatesAccumulator`. Otherwise, we should rename the base `Reporter` to be `Accumulator`. I realize that you want the preserve backward compatibility, which is appreciated, but I think at the cost of added confusion. We have a uesful abstraction that is `Reporter` for reporting values of interest and it is not going to be obvious that the `StatesAccumulator` is what you use to report out states. You could deprecate the existing `StatesReporter` and/or  name this class `StatesTrajectoryReporter` especially since you get a StatesTrajectory out.  @chrisdembia @tkuchida opinions?
Sorry I was about to merge and keep quiet, but I am thinking of the pain of renaming later once its in full use - and I do want to see it used everywhere.

chrisdembia:
Oh don't merge yet; I was thinking of switching out the `const_cast` for a `mutable`.

I agree about the confusion; I would prefer `StatesTrajectoryReporter`.

aseth1:
> Oh don't merge yet; I was thinking of switching out the const_cast for a mutable.
> I agree about the confusion; I would prefer StatesTrajectoryReporter.

:+1:

tkuchida:
> @tkuchida opinions?

IMHO, StatesTrajectoryAccumulator would be ideal since reporting doesn't necessarily imply that the same object is doing the accumulating, but StatesTrajectoryReporter seems best given the existing notion of Reporters. (My original suggestion came about because the class was called StatesCollector but the comments referred to "recording" states.)

chrisdembia:
Should we discuss naming of reporters in the next dev meeting?

aseth1:
> Should we discuss naming of reporters in the next dev meeting?

Not sure that is necessary. If `StatesTrajectoryReporter` strikes a balance between us, we can go with that.  Up to you.

chrisdembia:
I have addressed all the feedback:

- [x] Clarified why a method is commented out in the cpp file.
- [x] Renamed the class to StatesTrajectoryReporter.
- [x] Used `mutable` instead of `const_cast`

Ready for review again!

tkuchida:
:+1: Looks good to me.

aseth1:
sweet! :candy:

chrisdembia:
Thank you for merging.

PR #865, opened 2016-03-19
Remove warnings

klshrinidhi:


klshrinidhi:
This version compiles successfully and passes all tests on my machine.
Review and merge it if you like.
I will start another PR to cover remaining warnings.

aymanhab:
Thanks @klshrinidhi Way better than before, but would be good to get rid of the remaining warnings outside BTK and Wrapping (in another PR). Will merge after tests pass.

PR #867, opened 2016-03-19
Hide BTK in implementation

klshrinidhi:
Currently metadata in `DataTable`(s) obtained from `C3DFileAdapter` contains `Eigen::matrix`. This PR is to stop exposing `Eigen::matrix` directly to users of `C3DFileAdapter`. This change completely hides BTK in implementation of `C3DFileAdapter`.

This PR also solves the problem of `Eigen` symbol collision during Java wrapping when `WITH_BTK` is turned `ON` because this PR moves the BTK headers from `C3DFileAdapter.h` into `C3DFileAdapter.cpp` where they are directly needed.


chrisdembia:
+1

All PRs have been failing on Travis (just the one clang build that tests wrapping), including this PR. The failure is occurring because clang is running out of memory on the VM when trying to build the huge osimJavaJNI, which increased in size after including C3D. We have seen this problem before, and it is the reason why we split the python wrapper into multiple modules. The solution is to do the same splitting for java (if possible). 

@aymanhab and I have been ignoring this failure, given the deadline. I'd like for this to be merged, and I think it *should* be merged, despite the one failure that we know is unimportant.

A temporary fix would be decreasing the parallelization on travis; that might give clang more memory when building the java wrapping; this has worked for us in the past.

aymanhab:
Thanks @chrisdembia I think that's true, that the failure is due to size which grew for two reasons (additional BTK related classes and exposure of more SimTK classes). I don't mind splitting into modules as long as it doesn't complicate the client side (GUI or Matlab which is now a one line import).. But likely this's after the hackathon. Can we turn BTK wrapping off on Travis to get the build to succeed?

aymanhab:
Another option is to turn Java wrapping off and kep Python wrapping since it's handled more gracefully as of now.

klshrinidhi:
I am inclined to not merge anything until all CI builds are successful. I am not sure what to do to for short term (Hackathon) but for future, we may need to have CI builds for both:
* `WITH_BTK` ON
* `WITH_BTK` OFF

If we turn BTK OFF for CI builds, then we can not be sure if future changes work with BTK successfully. We encountered a problem in Java wrapping by turning BTK, which is how this PR came about in first place.

I made this PR to make sure C3D support is available from Python and Matlab for Hackathon.

chrisdembia:
> I made this PR to make sure C3D support is available from Python and Matlab for Hackathon.

One way to ensure that C3D support is not available in the hackathon is to not merge this PR. I completely agree that we should not normally merge things unless CI builds are successful; I've even considered using GitHub to block merges to master unless all the builds are green. But I think it's silly to worry about this stuff one day away from the hackathon. Do you have a way of getting the unrelated clang failure to go away before the hackathon?

klshrinidhi:
I will compile master (with this branch merged) on my machine with clang on Linux and OS X with `WITH_BTK` ON and Java/Python wrapping ON.
I will post the build output. We can merge it then.

klshrinidhi:
I just compiled this branch merged with master on Ubuntu 14.04 and OS X El Capitan, with clang on both platforms. Compilation succeeds. All tests pass on Ubuntu. One tests (*testInitState*) fails on OS X.
Outputs are too long to post here but following were the commands used on both platforms.
```bash
git clone git@github.com:opensim-org/opensim-core.git
cd opensim-core/dependencies
git checkout hide_btk_in_implementation
mkdir build
cd build
cmake ..                               \
    -DCMAKE_INSTALL_PREFIX=../install  \
    -DCMAKE_BUILD_TYPE=RelWithDebInfo
make -j12
cd ../..
mkdir build
cd build
cmake ..                                                 \
    -DCMAKE_INSTALL_PREFIX=../install                    \
    -DCMAKE_BUILD_TYPE=RelWithDebInfo                    \
    -DOPENSIM_DEPENDENCIES_DIR=../dependencies/install   \
    -DBUILD_PYTHON_WRAPPING=ON                           \
    -DBUILD_JAVA_WRAPPING=ON                             \
    -DWITH_BTK=ON
make
ctest
```
Go ahead and merge if you are okay with these results.

chrisdembia:
Thank you for testing manually. In it goes!

PR #868, opened 2016-03-19
use DecorativeMeshFile from Geometry::Mesh in sync. with latest Simbody. 

aymanhab:
And also to be consistent with the GUI.

### DIFF
--- a/OpenSim/Simulation/Model/Appearance.h
+++ b/OpenSim/Simulation/Model/Appearance.h
@@ -38,7 +38,11 @@ VisualRepresentation is the OpenSim name used across the OpenSim API, it is an
 that describes in what form is Geometry displayed:  
 DrawPoints, DrawWireframe, DrawSurface are supported.
 */
+#ifndef SWIG
 using VisualRepresentation = SimTK::DecorativeGeometry::Representation;
+#else
+typedef VisualRepresentation SimTK::DecorativeGeometry::Representation;
+#endif
 
 /**
 SurfaceProperties class holds the appearance properties of a piece of Geometry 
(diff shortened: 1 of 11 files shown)


aymanhab:
These are mainly changes to wrapping necessary to make the GUI and API consistent (in using DecorativeMeshFile, workaround SWIG not supporting the "using" typdef. Failure on Appveyor of testComponents(BallJoint) is unrelated.

aymanhab:
@chrisdembia Please let me know when you have a chance to review.

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -133,9 +137,9 @@ class OSIMSIMULATION_API Appearance : public Object {
     OpenSim_DECLARE_PROPERTY(color, SimTK::Vec3,
         "The color, (red, green, blue), [0, 1], used to display the geometry. ");
 
-    OpenSim_DECLARE_PROPERTY(surface_properties, SurfaceProperties,
+    OpenSim_DECLARE_UNNAMED_PROPERTY(SurfaceProperties,
         "Visuals applied to surfaces associated with this Appearance.");
-    OpenSim_DECLARE_PROPERTY(curve_properties, CurveProperties,
+    OpenSim_DECLARE_UNNAMED_PROPERTY(CurveProperties,
```
chrisdembia: Oh, a bonus! This change likely requires updating updateFromXML(); perhaps that can be a separate PR.

OpenSim/Simulation/Model/ModelVisualizer.cpp
```diff
@@ -245,7 +245,7 @@ bool ModelVisualizer::
 findGeometryFile(const Model& aModel, 
                  const std::string&          geoFile,
                  bool&                       geoFileIsAbsolute,
-                 SimTK::Array_<std::string>& attempts) const
+                 SimTK::Array_<std::string>& attempts)
```
chrisdembia: Logically, this method is const. The other option is to const_cast `attempts` whenever you use it. But I think this is fine for now. No change neessary.

chrisdembia:
My review is done. This PR looks good to go, but at least one related issue should be opened I think.

OpenSim/Simulation/Model/ModelVisualizer.cpp
chrisdembia: OH sorry I get it...it's static now. And `attempts` is passed in. Sorry.

chrisdembia: I like that the method is static.

aymanhab: Thanks @chrisdembia Yes, now a static method so can't be const since there's no object under it.

aymanhab:
Thanks @chrisdembia Happy to open issue for XML conversion though models I have worked just fine, ignoring the name.

PR #873, opened 2016-03-21
Find station in Ground

jimmyDunne:
Add a convenience method to station to get station position in Ground


OpenSim/Simulation/Model/Station.h
```diff
@@ -68,6 +68,9 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Station, ModelComponent);
     void setReferenceFrame(const OpenSim::PhysicalFrame& aFrame);
     /** Find this Station's location in any Frame */
     SimTK::Vec3 findLocationInFrame(const SimTK::State& s, const OpenSim::Frame& aFrame) const;
+    /** FInd thus Station's location in Ground */
```
tkuchida: "FInd thus" -> "Find this" :cactus:

carmichaelong:
Can have 3 different cases for tests (two tests currently test same thing):
1. Test location of a station in ground, when only ground exists.
2. Test location of a station in ground, when the station was defined in a different frame (i.e. frameA).
3. Test location of a station in a different frame (i.e. frameA), when defined in another frame (i.e. ground or frameB).

OpenSim/Simulation/Test/testFrames.cpp
```diff
@@ -85,25 +86,31 @@ int main()
...
+    
     try { testStationOnFrame(); }
     catch (const std::exception& e){
         cout << e.what() << endl; failures.push_back("testStationOnFrame");
     }
-
+    
+    try { testStationInGround(); }
+    catch (const std::exception& e){
+        cout << e.what() << endl; failures.push_back("testStationOnFrame");
```
tkuchida: testStationOnFrame -> testStationInGround

OpenSim/Simulation/Test/testFrames.cpp
```diff
@@ -316,88 +323,145 @@ void testPhysicalOffsetFrameOnBodySerialize()
...
 }
 
 
+
+
 void testStationOnFrame()
 {
     SimTK::Vec3 tolerance(SimTK::Eps);
+    
+    cout << "Running testStationInFrame" << endl;
```
tkuchida: testStationInFrame -> testStationOnFrame

OpenSim/Simulation/Test/testFrames.cpp
```diff
@@ -316,88 +323,145 @@ void testPhysicalOffsetFrameOnBodySerialize()
...
+        // now compare the original station location in rod1 (com) with the transformed station.
+        SimTK_TEST_EQ(stationPosition, com1);
+    }
+}
 
+void testStationInGround()
+{
+    SimTK::Vec3 tolerance(SimTK::Eps);
+    
     cout << "Running testStationOnFrame" << endl;
```
tkuchida: testStationOnFrame -> testStationInGround

chrisdembia:
OpenSim is not passing right now.

OpenSim/Simulation/Model/Station.cpp
```diff
@@ -112,3 +112,39 @@ SimTK::Vec3 Station::findLocationInFrame(const SimTK::State& s,
...
+    // Get stations current location
+    SimTK::Vec3 currentLocation = get_location();
+    
+    // Get station's physical frame
+    const PhysicalFrame& frame = getReferenceFrame();
+    // Get the frame's mobilized body
+    auto&  mb = frame.getMobilizedBody();
+    
+    // Use simbody method to get the acceleration in ground
+    return mb.findStationAccelerationInGround(s, currentLocation);
```
chrisdembia: The flavor here should be something like `mb.findStationAccelerationInGround(s, getReferenceFrame().findLocationInAnotherFrame(s, currentLocation, getReferenceFrame().getBaseFrame()))`. Not ACTUALLY this, b/c this is messy, but this is the flavor. Other people more familiar with such methods might have an idea of how to do this more cleanly.

Same for velocity.

OpenSim/Simulation/Test/testFrames.cpp
```diff
@@ -316,88 +332,145 @@ void testPhysicalOffsetFrameOnBodySerialize()
...
+    
+    
+    SimTK::Vec3 stationPosition;
+    
+    // Compare the vector between two stations, in two different frames.
+    SimTK::State& s = pendulum->initSystem();
+    for (double ang = 0; ang <= 90.0; ang += 10.){
+        double radAngle = SimTK::convertDegreesToRadians(ang);
+        const Coordinate& coord = pendulum->getCoordinateSet().get("q1");
+        coord.setValue(s, radAngle);
```
chrisdembia: You may want to play with the other degree of freedom also.

OpenSim/Simulation/Test/testFrames.cpp
```diff
@@ -316,88 +332,145 @@ void testPhysicalOffsetFrameOnBodySerialize()
...
+        stationPosition = stationInFrame2->findLocationInFrame(s, pendulum->getGround() );
+        // Set the location of the ground's station
+        stationInGround->set_location(stationPosition);
+        
+        //get the position of the ground's station in rod1
+        stationPosition = stationInGround->findLocationInFrame(s, pendulum->getBodySet().get("rod1") );
+        
+        // now compare the original station location in rod1 (com) with the transformed station.
+        SimTK_TEST_EQ(stationPosition, com1);
+    }
```
chrisdembia: I'm not sure I completely understand this test, but I think you're revisiting it in a separate PR.

OpenSim/Simulation/Test/testFrames.cpp
```diff
@@ -414,13 +487,51 @@ void testStationOnFrame()
...
+    manager.setInitialTime(0);
+    manager.setFinalTime(1);
+    manager.integrate(state);
+    
+    testBed.realizeAcceleration(state);
+    
+    // after simulating for 1 second the velocity should be
+    SimTK_TEST_EQ(SimTK::Vec3(0,-9.80665,0), myStation->findVelocityInGround(state))
+    // and the acceleration should
+    SimTK_TEST_EQ(SimTK::Vec3(0,-9.80665,0), myStation->findAccelerationInGround(state))
```
chrisdembia: As a reminder, if you simulate to a time other than 1.0 seconds, you can really ensure that velocity and acceleration are different.

OpenSim/Simulation/Model/Station.h
```diff
@@ -31,53 +31,64 @@
 #include "OpenSim/Simulation/Model/PhysicalFrame.h"
 
 namespace OpenSim {
+
```
chrisdembia: It's standard for us to not indent the contents of the namespace's scope. Could you remove this indent? If your editor did this automatically, it might also have an option to prevent indenting the namespace's scope.

chrisdembia:
Other than the tests (which you are revisiting), this looks pretty good to me.

jimmyDunne:
Thanks, Chris. I will address the issues with the tests in a PR into this branch.

chrisdembia:
I thought the intent was to update the test so it fails (since the new methods in Station right now do not consider intermediate PhysicalOffsetFrames), and then update the Station methods to properly consider intermediate PhysicalOffsetFrames (then the tests would pass).

OpenSim/Simulation/Model/Station.cpp
chrisdembia: See the above note; I was expecting a change like the above before merging this PR.

tkuchida: > I was expecting a change like the above before merging this PR.

:confused: This PR hasn't yet been merged.

chrisdembia: I know; @jimmyDunne said it was ready for review.

jimmyDunne:
?? I updated the test and it failed, then I updated station to make it work. I am confused. I have moved on to something else now so won't be able to keep working on this. If it is not suitable then just kill it.

chrisdembia:
Sorry...there was a misunderstanding, in part because I didn't manage to review your other PR (#900). I think `findLocationInGround()` was initially doing the right thing; it was the velocity and acceleration methods that need to change. I think the changes to make aren't so bad; I gave example code for `findAccelerationInGround()` as one of my comments in this PR. I just looked at #900, and the test seems to be much better but still does not ensure that it works for the scenario if intermediate bodies. The issue is the use of the method `mb.findStationLocationVelocityAndAccelerationInGround(state, point, l, v, a)`, which does the same thing as the current implementations of `Station::findLocationInGround()`, `Station::findVelocityInGround()`, and `Station::findAccelerationInGround()` and so has the same issue: the `point` that is provided to this method is expressed in `myExtraFrame`, which is a different frame from `mb`. Perhaps it'd be better to chat about this in person.

aseth1:
@jimmyDunne thanks for making these changes. In hindsight I think the Frames were not sufficient to support the computations you needed to create the outputs. I am adding velocity and acceleration calculations to the Frame. Then you won't have to worry about tunneling down to the base frame and talking directly to a `SimTK::MobilizedBody`.

I will take what you've done and absorb it into the changes I am making including adding the base `Point` abstraction. Please don't think this exercise was a waste of time- good things will come of it!

PR #875, opened 2016-03-22
Check for tabs in source code at the beginning on Travis

klshrinidhi:
* Move the checking for tabs to the beginning so this silly problem is caught without compiling.
* Add line number to the output of grep.


chrisdembia:
+1; merge when ready!

PR #885, opened 2016-03-23
[WIP] FileSystemPath

carmichaelong:
Class for handling file system paths. Most methods are implemented (getRelativePathFromOtherDir() is definitely not done). Any comments (especially on interface) are appreciated.

### DIFF
--- a/OpenSim/Common/FileSystemPath.h
+++ b/OpenSim/Common/FileSystemPath.h
@@ -1,0 +1,97 @@
+#ifndef OPENSIM_FILESYSTEMPATH_H_
+#define OPENSIM_FILESYSTEMPATH_H_
+/* -------------------------------------------------------------------------- *
+ *                             OpenSim: FileSystemPath.h                      *
+ * -------------------------------------------------------------------------- *
+ * The OpenSim API is a toolkit for musculoskeletal modeling and simulation.  *
+ * See http://opensim.stanford.edu and the NOTICE file for more information.  *
+ * OpenSim is developed at Stanford University and supported by the US        *
+ * National Institutes of Health (U54 GM072970, R24 HD065690) and by DARPA    *
+ * through the Warrior Web program.                                           *
+ *                                                                            *
+ * Copyright (c) 2005-2014 Stanford University and the Authors                *
+ * Author(s): Carmichael Ong                                                  *
+ *                                                                            *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may    *
+ * not use this file except in compliance with the License. You may obtain a  *
+ * copy of the License at http://www.apache.org/licenses/LICENSE-2.0.         *
+ *                                                                            *
+ * Unless required by applicable law or agreed to in writing, software        *
+ * distributed under the License is distributed on an "AS IS" BASIS,          *
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.   *
+ * See the License for the specific language governing permissions and        *
+ * limitations under the License.                                             *
+ * -------------------------------------------------------------------------- */
+
+// INCLUDES
+#include <OpenSim/Common/osimCommonDLL.h>
+#include "OpenSim/Common/Object.h"
+//#include "OpenSim/Common/Property.h"
+//#include "OpenSim/Common/ComponentConnector.h"
+//#include "OpenSim/Common/ComponentOutput.h"
+//#include "ComponentList.h"
+#include "Simbody.h"
+//#include <functional>
+//#include <memory>
+
+namespace OpenSim {
+
+//==============================================================================
+//                            OPENSIM FILESYSTEMPATH
+//==============================================================================
+/**
+
+*
+* @author Carmichael Ong
+*/
+class FileSystemPath : public Object {
+    OpenSim_DECLARE_CONCRETE_OBJECT(FileSystemPath, Object);
+
+public:
+    //==============================================================================
+    // METHODS
+    //==============================================================================
+    /** Default constructor **/
+    FileSystemPath();
+
+    /** Construct Component from an XML file. **/
+    FileSystemPath(const std::string& aFileName);
+
+    /** Use default copy constructor and assignment operator. */
+    FileSystemPath(const FileSystemPath&) = default;
+    FileSystemPath& operator=(const FileSystemPath&) = default;
+
+    /** Destructor. **/
+    ~FileSystemPath() = default;
+
+    // Return the absolute pathname relative to the working directory
+    FileSystemPath getAbsolutePathName();
+
+    // Return the absolute pathname relative to some specified relativeDir
+    FileSystemPath getAbsolutePathNameWithRelativeDir(const FileSystemPath relativeDir);
+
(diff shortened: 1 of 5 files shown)


OpenSim/Common/FileSystemPath.h
```diff
@@ -0,0 +1,122 @@
...
+
+    /** Use default copy constructor and assignment operator. */
+    FileSystemPath(const FileSystemPath&) = default;
+    FileSystemPath& operator=(const FileSystemPath&) = default;
+
+    /** Destructor. **/
+    ~FileSystemPath() = default;
+
+    // Get a FileSystemPath for the current working directory.
+    FileSystemPath getWorkingDirPath();
```
chrisdembia: could be a static method. Or, `IO::getCurrentWorkingDirectory()` could just return a `FileSystemPath`.

OpenSim/Common/FileSystemPath.h
```diff
@@ -0,0 +1,122 @@
...
+
+    // Get absolute path to a file or directory. getAbsoluteFilePath() and
```
chrisdembia: As is, these comments won't show up in doxygen. Use one of the following styles: `/** */`, `/*!  */`, `///`

chrisdembia:
Some helpful references for a filesystem API:

- python: https://docs.python.org/2/library/os.path.html
- c++'s upcoming filesystem: http://en.cppreference.com/w/cpp/filesystem

OpenSim/Common/FileSystemPath.h
```diff
@@ -0,0 +1,122 @@
...
+    // getAbsoluteDirPath() checks if the FileSystemPath is a file or dir
+    // first. getAbsolutePath() checks if the FileSystemPath is a file or dir
+    // then calls the corresponding function. If the path was not an absolute 
+    // path, then the directory is resolved with respect to the current working 
+    // directory.
+    FileSystemPath getAbsolutePath();
+    FileSystemPath getAbsoluteFilePath();
+    FileSystemPath getAbsoluteDirPath();
```
chrisdembia: I'm not sure these last two are necessary as part of the public interface; if you need these methods still, they could be private.

OpenSim/Common/FileSystemPath.h
```diff
@@ -0,0 +1,122 @@
...
+
+    // Get the absolute path for the directory of a file. Checks if it is
+    // given a file first, then gives the absolute path to the directory
+    // containing the file. If the path was not an absolute path, then the
+    // directory is resolved with respect to the current working directory.
+    FileSystemPath getAbsoluteDirPathForFile();
+
+    // Return the absolute path relative to some specified relativeDir.
+    FileSystemPath getAbsolutePathWithRelativeDir(FileSystemPath relativeDir);
```
chrisdembia: take the argument by const reference. This just appends two paths together? This is a little confusing (to say "absolute path relative to x"), but I think I get what you're saying. If I have a path "x/y/z" and I know it is relative to "u/v", then what you want is "abs(pwd)/u/v/x/y/z".?

OpenSim/Common/FileSystemPath.h
```diff
@@ -0,0 +1,122 @@
...
+    // Get relative pathname
+    FileSystemPath getRelativePathFromOtherDir(FileSystemPath otherDir);
+
+    // Convience methods for getting parts of the path as strings.
+    std::string getDirString();
+    std::string getFileString();
+    std::string getExtString();
+    void deconstructFilePathName(std::string& directory,
+                                 std::string& fileName,
+                                 std::string& extension);
```
chrisdembia: Could maybe be private.

OpenSim/Common/FileSystemPath.h
```diff
@@ -0,0 +1,122 @@
...
+    // Return the absolute path relative to some specified relativeDir.
+    FileSystemPath getAbsolutePathWithRelativeDir(FileSystemPath relativeDir);
+
+    // Get relative pathname
+    FileSystemPath getRelativePathFromOtherDir(FileSystemPath otherDir);
+
+    // Convience methods for getting parts of the path as strings.
+    std::string getDirString();
+    std::string getFileString();
+    std::string getExtString();
```
chrisdembia: I like these! Getting the extension, etc. :)

OpenSim/Common/FileSystemPath.h
```diff
@@ -0,0 +1,122 @@
...
+    // GET AND SET MEMBER VARIABLES
+    std::string getPathString() { return _pathStr; }
+    bool isFile() { return _isFile; }
+
+    void setPathString(std::string& aFileName) 
+    {
+        IO::TrimWhitespace(aFileName);
+        _pathStr = aFileName; 
+    }
+    void setIsFile(bool isFile) { _isFile = isFile; }
```
chrisdembia: No need for this setter. Should be known upon construction, etc.

OpenSim/Common/FileSystemPath.h
```diff
@@ -0,0 +1,122 @@
...
+    std::string getExtString();
+    void deconstructFilePathName(std::string& directory,
+                                 std::string& fileName,
+                                 std::string& extension);
+
+    // GET AND SET MEMBER VARIABLES
+    std::string getPathString() { return _pathStr; }
+    bool isFile() { return _isFile; }
+
+    void setPathString(std::string& aFileName)
```
chrisdembia: No need for this either. I like the idea of instances of this class being immutable.

chrisdembia:
It would be useful to see this used somewhere in the API, to help see if its interface is correct. Here is a good place: https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Control/ControlSetController.cpp#L229

OpenSim/Common/FileSystemPath.h
carmichaelong: The method currently uses `SimTK::Pathname::getCurrentWorkingDirectory()` rather than `IO::getCurrentWorkingDirectory()`. This is so that whenever this class looks for a current working directory, it will always use SimTK's method rather than a mix of methods. This might need more discussion.

OpenSim/Common/FileSystemPath.h
carmichaelong: You're generally right about what the intended use of this function is, but this does more than append two paths, since if the first path is an absolute path, it should ignore the `relativeDir`. For example, we should not join `C:\dir\file.ext` and `relativeDir` to become `abs(cwd)\relativeDir\C:\dir\file.ext`. Rather, it should still return `C:\dir\file.ext`. This is done by calling the methods in `SimTK::Pathname` and using its disambiguation rules.

aymanhab:
@carmichaelong This's outside the scope of this PR but inline with @chrisdembia request of how to use. Can this class be used as URL? I'd also look at the Java Path doc. at https://docs.oracle.com/javase/tutorial/essential/io/pathOps.html

aseth1:
@carmichaelong with #1215 merged can we close this WIP?

carmichaelong:
Yes. If we decide to have a FileSystemPath, it'll be based on #1215 rather than this.

PR #891, opened 2016-03-24
Fix Cooordinate's motion type initialization that caused hackathon example bug

aseth1:
which caused simulation in debug and release and different platforms to be different.  In particular, the Joints was correctly setting the `MotionType` of its coordinates upon construction, BUT when deserializing we use the default (registered) `Coordinate` (clone) and fill in its property values (effectively wiping out the Joint constructed coordinates), in which case the `Coordinate`'s  `_motionType` was uninitialized, since only a Joint can specify the correct motion type.

Changes:
1. Specified `Coordinate::MotionType::Undefined(0)` to be the default constructed value (so it is never uninitialized). This caused both release and debug versions to be consist (but now wrong).
2. For the `Coordinate::MotionType` after deserialization to be correct, it must get its value from its owning Joint. Thus, when the `Joint` constructs its coordinates (as member initializers,  except for `CustomJoint` which must wait until `finalizeFromProperties`) it also keeps track of the `MotionType` for the same coordinate (by `CoordinateIndex`).
3. `Cooridinate::getMotionType()` remains for convenience but it delegates to its owning `Joint`.
4. Removed the `motion_type` property from `Coordinate`.

### DIFF
--- a/OpenSim/Sandbox/ExampleHopperDevice/exampleHopperDevice.cpp
+++ b/OpenSim/Sandbox/ExampleHopperDevice/exampleHopperDevice.cpp
@@ -39,7 +39,7 @@ static const double GAIN{ 1.0 };
 static const double LOAD{ 2500.0 };
 static const double SPRINGSTIFF{ 5000 };
 static const double SIGNALGEN{ 0.33 };
-static const double REPORTING_INTERVAL{ 0.2 };
+static const double REPORTING_INTERVAL{ 0.20 };
 
 // Configure which hopper model to use and the attachments (by frame name)
 // and any signals (Outputs) for the device that will be created.
@@ -54,7 +54,7 @@ static const std::string HopperHeightOutput{
 
 namespace OpenSim {
 
-/* We begin by creating a class to contain all the parts for the model of
+/* We begin by creating a class to conatin all the parts for the model of
 our device. This is similar to how OpenSim uses Model to hold the parts
 of a musculoskeletal model. In a Model the parts are ModelComponents.
 Since the device will eventually be mounted on a musculoskeletal Model,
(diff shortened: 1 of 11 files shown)


OpenSim/Sandbox/ExampleHopperDevice/exampleHopperDevice.cpp
```diff
@@ -112,30 +145,30 @@ class PropMyoController : public OpenSim::Controller {
     }
 
     double computeControl(const SimTK::State& s) const {
-        // Compute the proportional control of GAIN * activation (Input)
-        // TODO
-        return 0;
+        double activation = getInputValue<double>(s, "activation");
+        if (activation < 0.31)
```
tkuchida: A comment explaining the significance of 0.31 would be helpful here.

OpenSim/Sandbox/ExampleHopperDevice/exampleHopperDevice.cpp
```diff
@@ -54,7 +54,7 @@ static const std::string HopperHeightOutput{
 
 namespace OpenSim {
 
-/* We begin by creating a class to contain all the parts for the model of
+/* We begin by creating a class to conatin all the parts for the model of
```
tkuchida: conatin -> contain

OpenSim/Sandbox/ExampleHopperDevice/exampleHopperDevice.cpp
```diff
@@ -70,20 +70,51 @@ class Device : public ModelComponent {
...
-
-    // TODO: add other outputs.
+    /** The lengthening speed of the device. */
+    OpenSim_DECLARE_OUTPUT(speed, double, getSpeed, SimTK::Stage::Velocity);
+    /** The force transmitted by the device. */
+    OpenSim_DECLARE_OUTPUT(tension, double, getTension, SimTK::Stage::Dynamics);
+    /** The power produced(+)/dissipated(-) by the device. */
+    OpenSim_DECLARE_OUTPUT(power, double, getPower, SimTK::Stage::Dynamics);
+    /** The height of the model that device is attached to. */
+    OpenSim_DECLARE_OUTPUT(height, double, getHeight, SimTK::Stage::Position);
```
tkuchida: Unclear how `height` differs from `com_height` (or whether they are meant to be equal). Perhaps `foot_height` or `clearance` is intended?

OpenSim/Sandbox/ExampleHopperDevice/exampleHopperDevice.cpp
```diff
@@ -70,20 +70,51 @@ class Device : public ModelComponent {
...
+        return getComponent<PathActuator>("cableAtoB").getLengtheningSpeed(s);
+    }
+    double getTension(const SimTK::State& s) const {
+        return getComponent<PathActuator>("cableAtoB").computeActuation(s);
+    }
+    double getPower(const SimTK::State& s) const {
+        return getComponent<PathActuator>("cableAtoB").getPower(s);
+    }
+
+    //device can read model height as measured from the block to ground
```
tkuchida: "as measured from the block to ground" -> "(the distance between the bottom of the lower block and the ground)"

OpenSim/Sandbox/ExampleHopperDevice/exampleHopperDevice.cpp
```diff
@@ -277,33 +306,44 @@ OpenSim::Device* createDevice() {
...
+    device->addComponent(pathActuator);
 
     // A controller that specifies the control to the actuator
-    // TODO
-    // TODO: finish implementing the PropMyoController class, above.
+    auto controller = new OpenSim::PropMyoController();
+    controller->setName("controller");
+    controller->set_gain(GAIN);
 
     // Connect the the controller to the device actuator
```
tkuchida: the the -> the

aseth1: Whoops, I did not mean to commit my local changes to the example! Ignore these, Sorry @tkuchida. I will close and make a new PR.

PR #894, opened 2016-03-25
Fix Cooordinate's motion type initialization that caused hackathon example bug

aseth1:
which caused simulation in debug and release and different platforms to be different. In particular, the Joints was correctly setting the MotionType of its coordinates upon construction, BUT when deserializing we use the default (registered) Coordinate (clone) and fill in its property values (effectively wiping out the Joint constructed coordinates), in which case the Coordinate's _motionType was uninitialized, since only a Joint can specify the correct motion type.

**Changes:**
1. Specified `Coordinate::MotionType::Undefined(0)` to be the default constructed value (so it is never uninitialized). This caused both release and debug versions to be consist (but now wrong).
2. For the `Coordinate::MotionType` after deserialization to be correct, it must get its value from its owning `Joint`. Thus, when the `Joint` constructs its coordinates (as member initializers, except for `CustomJoint` which must wait until `finalizeFromProperties`) it also keeps track of the `MotionType` for the same coordinate (by `CoordinateIndex`).
3. `Cooridinate::getMotionType()` remains for convenience but it delegates to its owning `Joint`.
4. Removed the `motion_type` property from `Coordinate`

### DIFF
--- a/OpenSim/Simulation/SimbodyEngine/Joint.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/Joint.cpp
@@ -238,7 +238,7 @@ void Joint::setMotionType(CoordinateIndex cix, Coordinate::MotionType mt)
         "Joint::setMotionType() for an invalid CoordinateIndex");
     // Grow the size of _motionTypes (array) if it is less than the number of
     // coordinates. Joint's _motionTypes must correspond to its CoordinateSet.
-    if (_motionTypes.size() < unsigned int(nc))
+    if (_motionTypes.size() < static_cast<unsigned>(nc))
         _motionTypes.resize(nc);
 
     _motionTypes[cix] = mt;
--- a/OpenSim/Simulation/SimbodyEngine/Joint.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/Joint.cpp
@@ -230,10 +230,16 @@ Coordinate::MotionType Joint::getMotionType(CoordinateIndex cix) const
 
 void Joint::setMotionType(CoordinateIndex cix, Coordinate::MotionType mt)
 {
-    OPENSIM_THROW_IF(cix >= numCoordinates(), Exception,
+    int nc = numCoordinates();
+
+    // Ensure that coordinate index is less than the number of coordinates
+    // this Joint has in its CoordinateSet.
+    OPENSIM_THROW_IF(cix >= nc, Exception,
         "Joint::setMotionType() for an invalid CoordinateIndex");
-    if (_motionTypes.size() <= cix)
-        _motionTypes.resize(numCoordinates());
+    // Grow the size of _motionTypes (array) if it is less than the number of
+    // coordinates. Joint's _motionTypes must correspond to its CoordinateSet.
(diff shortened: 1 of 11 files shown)


tkuchida:
:+1: hopper hops (Windows 32-bit RelWithDebInfo) :rabbit2:

OpenSim/Simulation/SimbodyEngine/Joint.cpp
```diff
@@ -222,6 +221,24 @@ const OpenSim::PhysicalFrame& Joint::getParentFrame() const
...
+    OPENSIM_THROW_IF(cix >= _motionTypes.size(), Exception,
+        "Joint::getMotionType() given an invalid CoordinateIndex");
+    return _motionTypes[cix];
+}
+
+void Joint::setMotionType(CoordinateIndex cix, Coordinate::MotionType mt)
+{
+    OPENSIM_THROW_IF(cix >= numCoordinates(), Exception,
+        "Joint::setMotionType() for an invalid CoordinateIndex");
+    if (_motionTypes.size() <= cix)
```
tkuchida: :confused: Why isn't this `if (_motionTypes.size() != numCoordinates())`?

aseth1: this method is primarily being used to set the `MotionType` as Coordinates are constructed during `Joint` construction which grows `_motionTypes` and will be ` < numCoordinates()` until it is added to the list.   `Joint::constructCoordinate()` checks (asserts) that the list of coordinates and motionTypes are are the same size after the motion type after this method is called.

tkuchida: Then perhaps the following line should be `_motionTypes.resize(cix+1);`? What doesn't make sense to me is why the `if` statement catches situations where `_motionTypes.size() <= cix` but the result is to increase the size to `numCoordinates()`. The resulting action doesn't immediately follow from the logic of the test. Perhaps a comment could be added to explain what's going on.

aseth1: `numCoordinates() == (cix+1)` so I can replace the to condition to be `_motionTypes.size() < numCoordinates()`.  I agree that would be easier to understand. The important thing is that the array must be resized to accommodate the `MotionType` of Coordinates the `Joint` is adding to its list (`CoordinateSet`). I'll make the change and add comments. Thanks.

tkuchida:
Travis is complaining about the typecast [here](https://github.com/opensim-org/opensim-core/pull/894/commits/c374e8d81048a43badd0888d54114f03c946d3fc#diff-2efad9acab01926bd89af8b957b3ac16R241). Otherwise, it looks good :+1:

klshrinidhi:
Fixed the compilation error.

aseth1:
@tkuchida thanks for the review
@klshrinidhi thanks for fixing the compilation error

aseth1:
Finally both Appveyor builds passed. @tkuchida good to merge?

aseth1:
Thank you @tkuchida

Issue #880, opened 2016-03-22
ASSERT_EQUAL does not evaluate Vec3


jimmyDunne:
If you pass ASSERT_EQUAL two difference Vec3's, it incorrectly evaluates them as equal. [This section of code](https://github.com/opensim-org/opensim-core/blob/b47cfb4cd5f3a2bc4632578c01f2f03adcd5763d/OpenSim/Auxiliary/auxiliaryTestFunctions.h#L36-L39) needs to be updated to use either SimTK_TEST_EQ or SimTK_TEST_EQ_TOL. 

I tried to update but couldn't get it working

tkuchida:
Could we leave `ASSERT_EQUAL()` as it is and instead define the comparison operators for Vec3 to compare 2-norms?

chrisdembia:
We could also specialize the ASSERT_EQUAL for Vec3 etc to use 2-norm.

PR #899, opened 2016-03-28
ASSERT_EQUAL overload for SimTK::Vec

klshrinidhi:
Addressing #880.

* Added two overloads for `ASSERT_EQUAL` for `SimTK::Vec` --
 * One that accepts tolerance.
 * One that does not accept tolerance and uses the default.

 Delegate to `SimTK_TEST_EQ` and `SimTK_TEST_EQ_TOL` respectively.

* Cleaned up the header file --
 * Moved `using namespace` declarations into function. If `using namespace` declaration is used in the header like it was, any file including this header will be forced to include the namespaces, potentially leading to name collisions.
 * Split lines to fit 80 char width.

### DIFF
--- a/OpenSim/Auxiliary/auxiliaryTestFunctions.h
+++ b/OpenSim/Auxiliary/auxiliaryTestFunctions.h
@@ -29,15 +29,20 @@
 #include <OpenSim/Common/PropertyObjArray.h>
 #include "getRSS.h"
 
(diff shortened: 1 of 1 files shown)


chrisdembia:
Looks good to me.

tkuchida:
:+1: Can be merged if tests pass

PR #903, opened 2016-03-30
Generate decorations fix

aymanhab:
Cache mesh loading from file to speed animation and add test case to verify speedup.


aymanhab:
@chrisdembia Please take a look whenever you have a chance. This's internal caching to speed up repeated calls to generate decorations and a test case to verify the behavior. 
Main issue I'm not quite satisfied with is when to invalidate the cache but that's hard to test now since there's no such context. Thanks.

OpenSim/Simulation/Model/Geometry.cpp
```diff
@@ -171,50 +171,63 @@ void FrameGeometry::implementCreateDecorativeGeometry(SimTK::Array_<SimTK::Decor
...
+        const std::string& file = get_mesh_file();
+        bool isAbsolutePath; string directory, fileName, extension;
+        SimTK::Pathname::deconstructPathname(file,
+            isAbsolutePath, directory, fileName, extension);
+        const string lowerExtension = SimTK::String::toLower(extension);
+        if (lowerExtension != ".vtp" && lowerExtension != ".obj" && lowerExtension != ".stl") {
+            std::clog << "ModelVisualizer ignoring '" << file
+                << "'; only .vtp .stl and .obj files currently supported.\n";
+            return;
+        }
```
aseth1: Why are you trying to load geometry at drawing time? This seems like a good case for extending finalize from properties to handle the geometry loading prior to simulation and visualization time. It would push these checks earlier where they belong IMHO.

aymanhab:
@aseth I think the rationale for delay loading (this hasn't changed since first introduced AFAIK) is that you don't incur any cost unless there's actual visualization going on.

aseth1:
> I think the rationale for delay loading (this hasn't changed since first introduced AFAIK) is that you don't incur any cost unless there's actual visualization going on

I agree with that rationale and that is also why there is a flag if the model is to be visualized or not. Perhaps the flag doesn't belong on the model long term, but checking it now would also save unnecessary loads and would free drawing code from unnecessary checks.

aymanhab:
Sure, we can move the flag down the road so that components don't need to do the checking. We may also need to account for the possibility that Geometry could be used for wrapping and/or Contact even if not visualized. Any specific proposal on this PR, though, or should I open a separate issue for future discussion? Thanks.

aseth1:
For this PR, I would like to see the logic and attempts to load a geometry file/mesh to go in the Geometry's `extendFinalizeFromProperties` and not be in the `Geometry::generateDecorations()`. For now you can simply check `Model::getUseVisualizer()` to decide even to attempt loading. In the future we could add logic to decide if the geometry should be loaded for other reasons (contact, wrapping, etc..) in any case that logic should be in finalizeFromProperties, so best to move it there now. Thanks!

aymanhab:
Thanks @aseth1 I will try that. One caveat is that when using the GUI, getUseVisualizer() will return false but I'll see what I can do to work around it.

aseth1:
Good point. The decision to use the API visualizer and/or to visualize can/should be separated. Geometry itself could maintain a flag whether or not to visualize - that seems cleaner than relying on the Model. This could be generalized for any use of the geometry for computation (e.g. wrapping, contact, etc...) and/or visualization - since if any of these apply the file must be loaded prior to the system being created.

OpenSim/Simulation/Model/Geometry.h
```diff
@@ -611,7 +615,12 @@ class OSIMSIMULATION_API Mesh : public Geometry
     /// Method to map Mesh to Array of SimTK::DecorativeGeometry.
     void implementCreateDecorativeGeometry(
         SimTK::Array_<SimTK::DecorativeGeometry>& decoGeoms) const override;
-
+private:
+    // We cache the DecorativeMeshFile and a flag indicating if we tried to
+    // load the mesh from file so we don't try loading from disk every frame.
+    // These are mutable since they're not part of the public interface.
+    mutable std::shared_ptr<SimTK::DecorativeMeshFile> cachedMesh;
```
chrisdembia: Could you remind me why `shared_ptr` is used here instead of `unique_ptr`?

OpenSim/Simulation/Model/Geometry.h
```diff
@@ -611,7 +615,12 @@ class OSIMSIMULATION_API Mesh : public Geometry
...
+    mutable bool triedToLoad;
```
chrisdembia: Would it be possible to use the pointer `cachedMesh` as the flag (and so remove this member variable)? The smart pointers (both unique and shared) have an operator bool to tell you if they are managing an object:

image
http://en.cppreference.com/w/cpp/memory/unique_ptr

So you might be able to replace `if (!triedToLoad)` with `if (!cachedMesh)`.

OpenSim/Simulation/Model/Geometry.h
chrisdembia: If only the Mesh object is going to need to access the `cachedMesh`, it'd be better to make it a unique_ptr.

OpenSim/Simulation/Model/Geometry.cpp
```diff
@@ -171,50 +171,63 @@ void FrameGeometry::implementCreateDecorativeGeometry(SimTK::Array_<SimTK::Decor
...
+        decoGeoms.push_back(*dmesh);
+        cachedMesh.reset(dmesh);
     }
-    catch (const std::exception& e) {
-        std::clog << "Visualizer couldn't read "
-            << attempts.back() << " because:\n"
-            << e.what() << "\n";
-        return;
+    else { // Not first time, reuse only if valid cachedMesh
+        if (cachedMesh.get() != nullptr) {
```
chrisdembia: Could also do `if (cachedMesh)`, but this is also fine.

OpenSim/Simulation/Model/Geometry.h
chrisdembia: This (as well as some of my other comments) might become irrelevant based on @aseth1's feedback.

chrisdembia:
In general, I agree that the loading of the mesh should happen during `finalizeFromProperties()`. However, I also think this PR makes the code better than it was, and if the changes required to move the caching to `finalizeFromProperties()` is non-trivial, then a separate issue should be created for that.

OpenSim/Simulation/Model/Geometry.cpp
```diff
@@ -171,50 +171,63 @@ void FrameGeometry::implementCreateDecorativeGeometry(SimTK::Array_<SimTK::Decor
...
+            std::clog << "Visualizer couldn't read "
+                << attempts.back() << " because:\n"
+                << e.what() << "\n";
+            return;
+        }
+
+        DecorativeMeshFile* dmesh = new DecorativeMeshFile(attempts.back().c_str());
+        dmesh->setScaleFactors(get_scale_factors());
+        decoGeoms.push_back(*dmesh);
+        cachedMesh.reset(dmesh);
```
chrisdembia: Lines 225 and 223 should not be repeated; they should just be outside of the `if (!triedToLoad)`:

```
if (!cachedMesh) {
    cachedMesh.reset(new ...);
}
cachedMesh->setScaleFactors(...);
decoGems.push_back(*cachedMesh);
```

I know this all might change anyway (if you move the loading to finalizeFromProperties()).

chrisdembia:
I'm done with my review.

OpenSim/Simulation/Model/Geometry.h
aymanhab: @chrisdembia The reason there is a separate flag is that if we try to fail once and fail then the pointer will be null and we'll try again with every frame making the drawing excessively slow even if there's only one missing or failed-to-load mesh file. That's why the flag is used to indicate that we "tried". I don't expect the objects with  loaded cachedMesh to be copies so I'm fine with either shared_ptr or unique_ptr.

chrisdembia: Ah okay I get it now.

chrisdembia: My personal preference is that there is an exception if the file is not found, but that's unrelated to this PR.

aymanhab:
@aseth1 One complication of trying to move the mesh loading/caching code to extendFinalizeFromProperties is that the "model pointer" is not set until connect time which is too late for extendFinalizeFromProperties. Mesh needs to know the directory that the model came from since search paths look there first. Longer term we should probably have a "working directory" or "search path" available to all components during initialization but that's likely outside the scope of this PR. Is there another convenient/plausible place? Thanks.

aymanhab:
@aseth1 Can you please advise as to how to handle the unavailability of the model during extendFinalizeFromProperties(). Seems like similar functionality would be needed for ControlSetController or any object that needs a file (which would need a more general solution).
Thanks. 

@chrisdembia I changed the smart pointer per your review. Thanks

aseth1:
> Can you please advise as to how to handle the unavailability of the model during extendFinalizeFromProperties().

Does Geometry reinterpret its properties based on the model? How would you visualize Geometry without a model, say with just attached to Ground? Do we need to require a model just to find a geometry file in the first place? Seems over restrictive considering you can assemble a model in the API without an input model file.

As a workaround you could delay the loading of the geometry file to connect() at which time you would be handed the model.

I would prefer that the loading of files didn't require a reference to a model.

aymanhab:
@aseth1 The loading doesn't require the "model object itself" but requires to know the directory where the model came from since files are assumed to have relative path to the model by default, that's why the discussion above referred to this being common need not only for Geometry files. It may be addressed longer term by a File path property (PR  #885)

Regardless of whether using API or XML, using a relative path to a file necessitates specifying "relative to what", unless there's a separate library search path somewhere else. 

I'll try using extendConnect until we converge/decide on long term solution to FilePath issues but that seems a outside the scope of this PR. Thanks.

aseth1:
It would be a cleaner solution to keep the file loading in `extendFinalizeFromProperties` and designate to static the file loader that doesn't require a Model. You could easily modify `ModelVisualizer::findGeometryFile` to not take in a model but to take in the Component that needs the the geometry loaded. You can then traverse up the ownership tree (calling `getParent()`) from that component. If the top-level component is a `Model` then nothing changes, otherwise you jump to the case where the Model input file is "unassigned". This solution would allow us to clean up the file path issue without having to revisit Geometry again- which is likely to be forgotten.

The scope of changes seems similar to me with the added benefits that Geometry doesn't have to be revisited and we don't have provide a model to visualize meshes.

aymanhab:
@aseth1 Thanks for the feedback and suggestions. I implemented your first suggestion to move the caching functionality to extendConnect and it worked smoothly and led to removal of an extra flag. :+1:  The build failure is unrelated.

I'm hesitant to make the move to  extendFinalizeFromProperties and introduce traversal mechanism (that would be generally useful and widely used but not Geometry specific) while addressing this Geometry specific PR. I'd suggest we review and merge this instead and open a separate issue regarding all "File" properties so that we can design and implement one fix that addresses Geometry, ContactGeometry, ExternalLoads, ControlSetController and the tools/analyses (that do not belong to any model) in a uniform way. Please let me know what you think.

aymanhab:
@aseth1 in trying to use getParent for traversal to the root model I ran into the issue that the method Component::getParent is protected so it can't be invoked to traverse the tree up outside the Component class, do you think the method should be made public or that a new public method Component::getRootComponent() should be added? Thanks.

aseth1:
OK, please go ahead and make `Component::getParent()` and `hasParent()` public. I had no reason to make it public earlier, but likely to be more cases. Thanks.

aymanhab:
@aseth1 I did move the mesh loading operation to FinalizeFromProperties and all test cases pass, but that exposed two runtime problems:
1. extendFinalizeFromProperties is called twice during the initialization sequence which doubles the initialization time if dominated by file loading as is the case here. If I add a flag to check if visualization is needed and not load until later then it'd defeat the whole purpose of this move.
2. I added a test case for editing from the GUI which goes thru the initialization sequence on any simple edit. Doing 1 edit now takes more than 2 seconds.
Neither of these issues seem specific to geometry (I'll be dealing with them in PrescribedController, PrescribedForce etc.) as well as in ContactMesh so a longer term efficient solution is likely still needed. I'm all open to ideas.

aseth1:
If the Geometry `isObjectUpToDateWithProperties` is true and the mesh is already loaded, there is no reason to reload EXCEPT in the case that file itself has changed (and you want those changes to take effect). For now I would create an issue that in the near future we should use a FilePath to ask if the file has been modified, if not, then do not reload.

aymanhab:
Good point @aseth1 I completely forgot about that flag. I'll use it and let you know how it goes. Thanks :+1:

aymanhab:
Comments/feedback has been addressed, ready for review.

aseth1:
This looks good to me @aymanhab. I will merge, but I have some comments that we could discuss offline, or if you agree, create related issue(s).

1. Is `implementCreateDecorativeGeometry` inside of `Geometry::generateDecorations()` still necessary? If so is it appropriately named?
2. Why create geometry on each call to `generateDecorations` simply to scale to fixed scale factors (properties which are not allowed to change post 'finalizeFromProperties'). If each Geometry kept a reference to its underlying fixed Geometry then Mesh would not be special it would just be creating its underlying Mesh instead. It also seems that scaling should be decoupled from creation, since someone could easily forget to apply scale factors when implementing `implementCreateDecorativeGeometry` for their new Geometry type.
3.  OpenSIm Frames already provide the Transform in Ground, so I don't understand why `Geometry::setDecorativeGeometryTransform` has to dynamic_cast for a `PhysicalFrame` and go fishing for mobilzed body index etc.. at this stage you could simply draw all geometry in Ground.

aymanhab:
Awesome, thanks much @aseth1 Will discuss in person tomorrow and then we can decide to open issues accordingly.

Issue #872, opened 2016-03-21
API Programs should not rely on Path names to do connections


aymanhab:
It was evident in hackathon that using names/fullPathName(s) to establish connections is error prone (typos, tree/scope, case-sensitivity). This is particularly avoidable in the case when the user/devloper already has a reference to the object(s) that they're connecting to. We should look for alternatives.

aseth1:
I agree. If you have a reference you should always be able to satisfy a connection with it and not specify its name. There are many constructors and accessors that  require names and these would have to change.

PR #905, opened 2016-04-02
Removed use of frame names from `Joint` constructor

aseth1:
Begin to address issue #872. Starting by replacing Joint constructors to use references to the PhysicalFrames that it joins. The use of names (strings) in the API to create models is a remnant from when properties were the only way to specify dependencies.  The API supports direct connections to other Components (by reference) and it is confusing to have a splattering of both (reference and string based) in the current API.  

The most significant changes are to Joint.h and the uses in test cases and example. 

There are two temporary "hacks" that I want to get rid of in subsequent PR(s), but they will not change the interface:

1. `Model::createMultibodyTree()` had to also call connect on the frames that the Joint is connected to in order for those frames to provide their base frame's name. In the past, this wasn't an issue because any offset frames were subcomponents of the Joint (so connect() on the joint causes them to connect()). If the `PhysicalOffsetFrame` is added to a Body it will not be able to provide its base frame until it is connected to that body. One solution might be to build the multibody tree later after components have had a chance to connect.

2. A component, like  a `PhysicalOffsetFrame` can be constructed "on the fly" and used to satisfy a Connector, for example, of a Joint when the Joint is constructed. That offset can the be added to (adopted by) another component like the Body frame it is offsetting or to the Joint itself (which we have been doing till now). When the Joint connects to the offset frame (e.g. during construction) it also needs to store its path name (e.g. `connectee_name`) but its path is not known. Rather assuming it is from root ('/') i recognize it as an orphan and only store its name. When the owning Model is connected it will have to find the offset frame and forcing it to search from '/' was wrong, it should search the local component (e.g. the joint) instead, which it does if there is no `/`. I added a hack to strip the leading `/`. I don't think this is a robust solution, and will replace it soon (once I come up with something better). It isn't independent of `1.` either.

I also reinstated warnings for not finding a component on its path and the Exception if duplicate components (with same name and type) are found.

### DIFF
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -1684,7 +1684,7 @@ public:
                 // when what appears to be the complete path.
                 std::string details = msg + " Found '" + compFullPathName + 
                     "' as a match for:\n Component '" + name + "' of type " + 
-                    comp.getConcreteClassName() + " found, but it "
+                    comp.getConcreteClassName() + ", but it "
                     "is not on specified path.\n";
                 //throw Exception(details, __FILE__, __LINE__);
                 std::cout << details << std::endl;
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -1679,17 +1679,15 @@ public:
               // where only names were used (not path or type)
               // TODO replace with an exception -aseth
             else if (comp.getName() == subname) {
-                if (foundCs.size() == 0) {
-                    foundCs.push_back(&comp);
-                    // Silenced this Message for Hackathon
-                    // TODO Revisit why the exact match isn't found when
-                    // when what appears to be the complete path.
-                    /*msg += "a match for Component '" + name + "' of type " +
-                        comp.getConcreteClassName() + " found, but it "
-                        "is not on specified path."; */
-                    //throw Exception(msg, __FILE__, __LINE__);
-                    //std::cout << msg << std::endl;
-                }
+                foundCs.push_back(&comp);
+                // TODO Revisit why the exact match isn't found when
+                // when what appears to be the complete path.
+                std::string details = msg + " Found '" + compFullPathName + 
+                    "' as a match for:\n Component '" + name + "' of type " + 
+                    comp.getConcreteClassName() + " found, but it "
+                    "is not on specified path.\n";
+                //throw Exception(details, __FILE__, __LINE__);
+                std::cout << details << std::endl;
             }
         }
 
--- a/OpenSim/Simulation/SimbodyEngine/Joint.h
+++ b/OpenSim/Simulation/SimbodyEngine/Joint.h
@@ -139,15 +139,15 @@ public:
 
         @param[in] name     the name associated with this joint (should be
                             unique from other joints in the same model)
-        @param[in] parentName   the name of the parent PhysicalFrame for the joint
-        @param[in] childName    the name of the child PhysicalFrame for the joint
+        @param[in] parent   the parent PhysicalFrame that joint connects to
+        @param[in] child    the child PhysicalFrame that joint connects to
         @param[in] reverse  Advanced optional flag (bool) specifying the 
                             direction of the Joint in the multibody tree. 
                             Default is false (that is, parent to child).
         */
     Joint( const std::string& name,
-           const std::string& parentName,
-           const std::string& childName,
+           const PhysicalFrame& parent,
+           const PhysicalFrame& child,
            bool reverse = false);
 
     /** Backwards compatible Convenience Constructor 
@@ -201,20 +201,12 @@ public:
     virtual ~Joint();
 
     // GET & SET
-
-    void setChildFrameName(const std::string& name);
-    const std::string& getChildFrameName() const;
-
     /**
      * Get the child joint frame.
      *
      * @return const PhysicalFrame reference.
      */
     const PhysicalFrame& getChildFrame() const;
-
-    // Relating to the parent joint frame
-    void setParentFrameName(const std::string& aName);
-    const std::string& getParentFrameName() const;
 
     /**
      * Get the parent frame to which this joint attaches.
--- a/OpenSim/Simulation/SimbodyEngine/Joint.h
+++ b/OpenSim/Simulation/SimbodyEngine/Joint.h
@@ -115,12 +115,14 @@ public:
 // OUTPUTS
 //=============================================================================
     OpenSim_DECLARE_OUTPUT(power, double, calcPower, SimTK::Stage::Acceleration);
-    // reaction_load_at_parent_frame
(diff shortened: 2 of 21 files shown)


OpenSim/Common/Component.h
```diff
@@ -1679,17 +1679,15 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
-                        "is not on specified path."; */
-                    //throw Exception(msg, __FILE__, __LINE__);
-                    //std::cout << msg << std::endl;
-                }
+                foundCs.push_back(&comp);
+                // TODO Revisit why the exact match isn't found when
+                // when what appears to be the complete path.
+                std::string details = msg + " Found '" + compFullPathName + 
+                    "' as a match for:\n Component '" + name + "' of type " + 
+                    comp.getConcreteClassName() + " found, but it "
```
tkuchida: Delete ` found`

tkuchida:
`</review>` Looks good to me :deciduous_tree:. I pushed a small change (very minor—hope you don't mind).

aseth1:
Don't mind at all @tkuchida - thank you for the edit!

aseth1:
As the proponent of #872, @aymanhab can you review these changes?  Thanks!

aymanhab:
@aseth1 This turned out to be much bigger than I expected, primarily because I would have expected an "extra" constructor that takes references for convenience rather than replace the old constructors which required changes to a bunch of test cases, examples etc that broke. Do you believe that was necessary?.
If the Frames have no names, would they still survive the disconnect/connect paradigm used in initialization? (for example after any unrelated edit?)
Do the frames need to be added to the model first? or does it matter? wasn't clear from the code and there was no documentation change in Joint.h as far as I can tell.

aseth1:
> Do you believe that was necessary?

@aymanhab one of the complaints levied against the API, which you have echoed, is that there are multiple pathways to do the same thing and it isn't clear which to choose - and I completely agree. Since our examples and test cases are the first place people look for usage, having nothing change and just adding the new constructors was terribly unsatisfying and I don't believe achieves the goal of making the API cleaner/more clear without removing unnecessary options. Adding the new constructors is trivial especially if they don't get used. The most comprehensive way to validate the new constructors work is to make their use the standard way to construct Joints, so that is what I did.

> This turned out to be much bigger than I expected

I was actually surprised how few things broke and it was pretty much the same fix everywhere!

> If the Frames have no names, would they still survive the disconnect/connect paradigm used in initialization? (for example after any unrelated edit?)

To the extent that this may already be happening in examples and test cases, I think yes it does work, but is likely not robust. I would like to discuss a scheme that does not allow API users to construct nameless components- that would avoid a lot of headaches. 

> Do the frames need to be added to the model first? or does it matter? wasn't clear from the code and there was no documentation change in Joint.h as far as I can tell.

It should not matter. Dependencies must be part of the Component (its tree) who's connect() method is being invoked. For our examples, that means it has to be somewhere in the model by the time you call `Model::connect()` which is called by `initSystem()` (if one doesn't need to know about finer granularity).

The plan is for path names to be used only to archive/remember the connection and serialize them and not to be exposed via the API. The only exception would be for generic access/finding of components (e.g.  `getComponent()` and `findComponent()`), but otherwise unnecessary for putting a model together with "live" components.

Let me know what Joint document improvements you would like to see made. Thanks Ayman!

aymanhab:
Thanks for the explanation @aseth1 The doxygen for Joint constructor says " where the parent and child frames are specified by name" that doesn't seem to be true anymore.  Would also be good to reiterate something like "Make sure the PhysicalFrame(s) are added to the model (as was done earlier near the top of the file). The example uses in cpp, .python at the top the file still refer to names so are not valid anymore.

One issue with using references in general is that users may allocate them on the stack and then they get deleted/scoped-out. but  that's outside the scope of this PR. 

Other than the documentation fixes I'm done with my review.

aseth1:
Thanks for finding those gaps in the doxygen comments - I will fix them.

I agree that the reference passing has issues. For now the rule I am trying to establish is that you pass a reference for satisfying connections but must provide pointers for adoption/ownership. I am happy to discuss that more with you offline.

aseth1:
@aymanhab the documentation has been updated. @tkuchida le me know if you have any other comments, otherwise it can be merged.

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -115,12 +115,14 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
...
-    OpenSim_DECLARE_OUTPUT(reaction_load_at_parent_frame_in_ground_frame, SimTK::SpatialVec,
-        calcReactionAtParentFrameInGround, SimTK::Stage::Acceleration);
-    // reaction_load_at_child_frame
-    OpenSim_DECLARE_OUTPUT(reaction_load_at_child_frame_in_ground_frame, SimTK::SpatialVec,
-        calcReactionAtChildFrameInGround, SimTK::Stage::Acceleration);
+    /** get the joint reaction force and moment experienced on the parent frame
+        and expressed in ground. */
+    OpenSim_DECLARE_OUTPUT(reaction_on_parent, SimTK::SpatialVec,
+        calcReactionOnParentExpressedInGround, SimTK::Stage::Acceleration);
+    /** alternatively, get the joint reaction force and moment experienced on
```
tkuchida: Remove "alternatively" (it won't make sense unless you happened to read the description for `reaction_on_parent` first).

tkuchida:
`</review>` Looks good with minor change noted above

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -115,12 +115,14 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
...
+        the child frame and expressed in ground. */
+    OpenSim_DECLARE_OUTPUT(reaction_on_child, SimTK::SpatialVec,
+        calcReactionOnChildExpressedInGround, SimTK::Stage::Acceleration);
```
chrisdembia: I specifically created the OUTPUT macros so that they do not have documentation associated with them. I think all documentation for them should be placed with the associated method. Otherwise, we'll have two places documenting the method and that could get confusing. Could you move these comments to the respected methods?

chrisdembia:
I also made a small comment.

OpenSim/Simulation/SimbodyEngine/Joint.h
aseth1: Good point. I think  the method documentation is already better anyways. I will clear it out.

aymanhab:
@aseth1 If you looked at the doxygen and the issues have been fixed feel free to merge when the tests pass.

aseth1:
Thanks @aymanhab - in it goes!

Issue #904, opened 2016-03-31
SWIG wrapping long-term plan


klshrinidhi:
This is more of a proposal for long-term regarding SWIG wrapping.
- I think it will be good in the long run to have **one interface file per class**.
  - :+1: The whole process of wrapping will be lot more comprehensible.
  - :+1:  Adding new classes will be relatively easier.
  - :+1: Tracking problems/bugs will be relatively easier as it will be relatively easier to understand effects of editing interface files.
  - :+1: Build will be faster as there will be as many CXX files as there are interface files.
  - :+1: Dependency management in CMake can be much more fine grained just like we do it with CXX code.
  - :+1: We can write unit tests to test each wrapped class for Python/Java.
  - :-1: There is initial investment of time/effort in creating individual interface files and giving CMake an appropriate dependency graph.
- I think we need lot more unit tests to make sure changes to interface files don't break anything.

Let me know your thoughts.

aymanhab:
@klshrinidhi Of the 300+ classes we wrap/expose today, likely the majority are handled seamlessly by including the header files in the common interface file(s). That's reasonably streamlined. 

Breaking this into a handful of interface files (one per library) seams like a perfect granularity, while one per class would be excessive unless the whole process is fully automated. If we follow the full automation route, users will have to have other custom tools to indicate what to wrap and troubleshoot so complexity would increase. I'm not sure who are we targeting here, users have looked and followed our swig interface files in the past (the few who wanted to). Totally agree on need for unit testing but not sure about the rest.

klshrinidhi:
Currently, I am finding the interface files pretty confusing to get SWIG to generate multiple CXX files instead of one. There seems to lot of trial and error involved. This issue is coming from that perspective.

For classes that can be exposed by just including the header file, we can make one interface file for a group of maybe a maximum of 10 classes. Grouping classes into one interface file can lead to problems we are having right now about build taking too long. For these kinds of classes, we could explore possibility of having CMake generate the interface file and split them up into appropriate number of chunks.

For classes that need more than just including header file, we should probably create one interface file per class. It has more benefits than downsides I think.

Issue #790, opened 2016-02-29
Wrapping step not executing when headers are edited


klshrinidhi:
I am having this issue on Linux currently. Haven't checked on OS X and Windows.

Editing the headers (I am currently editing `Bindings/SWIGSimTK/BigMatrix.h`) and compiling (`make`) again does not re-run the *wrapping* step and ends up just compiling the previously generated wrapper-code. Right now, I delete the wrapper code file (`Bindings/Java/OpenSimJNI/OpenSimJNI_wrap.cxx`) to make sure it is generated again.

chrisdembia:
The best solution would be to use `swig -M` to automatically detect the files that the wrapping depends on. `swig -M` generates a Makefile that lists all the dependencies. We'd have to write a script to clean up the Makefile to just get a simple list of filenames.

chrisdembia:
Thank you Jen.

PR #906, opened 2016-04-02
Generate multiple CXX files for Java wrapping.

klshrinidhi:
Travis currently seems to choke clang while compiling the SWIG generated cxx file for Java wrapping. The cxx file is about 3,80,000 lines. This PR is to generate multiple smaller cxx files, which also has the benefit of parallel compilation. 

*The strange thing is that SWIG generated cxx files for Python (after splitting) are pretty large too and they seem to compile fine.*

Note that this PR bumps up the SWIG version requirement from 3.0.5 to 3.0.6 because generating multiple modules and linking them together does not work without [this fix](https://github.com/swig/swig/issues/353) in SWIG.

##### Before
SWIG generated cxx files:

| file | num-lines |
| ----- | ----- |
| Bindings/Java/OpenSimJNI/OpenSimJNI_wrap.cxx | 378104 |
| Bindings/Python/python_actuators_wrap.cxx | 96527 |
| Bindings/Python/python_analyses_wrap.cxx | 25310 |
| Bindings/Python/python_common_wrap.cxx | 101556 |
| Bindings/Python/python_simbody_wrap.cxx | 96924 |
| Bindings/Python/python_simulation_wrap.cxx | 379393 |
| Bindings/Python/python_tools_wrap.cxx | 48260 |

##### After
SWIG generated cxx files:

| file | num-lines |
| ----- | ----- |
| build/Bindings/Java/OpenSimJNI/java_actuators.cxx | 83849 |
| build/Bindings/Java/OpenSimJNI/java_common.cxx | 47035 |
| build/Bindings/Java/OpenSimJNI/java_simbody.cxx | 26517 |
| build/Bindings/Java/OpenSimJNI/java_simulation1.cxx | 74175 |
| build/Bindings/Java/OpenSimJNI/java_simulation2.cxx | 62843 |
| build/Bindings/Java/OpenSimJNI/java_simulation3.cxx | 66061 |
| Bindings/Python/python_actuators_wrap.cxx | 96519 |
| Bindings/Python/python_analyses_wrap.cxx | 25302 |
| Bindings/Python/python_common_wrap.cxx | 95934 |
| Bindings/Python/python_simbody_wrap.cxx | 96924 |
| Bindings/Python/python_simulation_wrap.cxx | 383874 |
| Bindings/Python/python_tools_wrap.cxx | 48252 |


klshrinidhi:
CI builds fail because SWIG 3.0.6 is not available.

klshrinidhi:
*testComponents* failing on Appveyor PR build.

[Travis PR build has Java wrapping OFF](https://travis-ci.org/opensim-org/opensim-core/jobs/120743333#L2411).

aseth1:
Appveyor PR failure seems spurious and unrelated to the changes. As for the breakup into multiple swig generated CXX files, it would be good to get @aymanhab's blessing before merging.

aymanhab:
Thanks @klshrinidhi I'll need to try the result with the GUI to see if anything breaks since the changes to the interface file(s) are massive, it's practically impossible to compare. The test cases for java bindings were started few weeks ago so the real test is GUI integration. Stay tuned.

In general the number of lines may not correlate to the complexity or number of code-blocks etc. so it's possible large files from Python binding are better handled than those from Java binding. I would've preferred to have one Java/simulation interface file similar to Python so the breakup is logical and more maintainable but I don't know if that was achievable.

aseth1:
I would like to see the same breakdown/layout for both Java and Python wrapping.

aymanhab:
Testcase "TestBasics" fails with TestBasics.java:13: error: cannot find symbol
9>           StdVecString outNames = fi.getOutputNames();
9>           ^
9>    symbol:   class StdVecString
9>    location: class TestBasics
9>  1 error
Was there a name change? If so why?

klshrinidhi:
@aymanhab 
Breaking simulation into 3 pieces was intentional to get the compile time down. It is currently taking a long time to compile with wrapping. I compile the whole source code tens of times everyday with/without wrapping to test my changes. This splitting seems to have improved my compile time because cxx files are compile in parallel. 
I split the interfaces files to generate cxx files of roughly less than 100,000 lines each.

Now, I see that Python cxx files are taking the longest to compile. Previously when I built OpenSim with wrapping ON, Java wrapping was the last thread to finish. Now it is Python simulation. Like I proposed in the issue #904, I think it would be good split both Python & Java even more for:
* Better compilation time.
* Better dependency management in CMake. If I change a source file, wrappers are not generated again. It would be good to have SWIG generated cxx files *depend* on their respective interface files and source files. For every change in source file, I have to delete the all the generated files to make sure they are re-generated, which is inefficient for development.

All the tests passed on my machine and they are passing on CI builds. I will double-check and get back in the morning.

chrisdembia:
> It would be good to have SWIG generated cxx files depend on their respective interface files and source files.

SWIG targets *do* depend on the right interface files, just not the right OpenSim header files.

I agree with @aseth1 that both python and java should be split in the same way. It'd also be good to reorganize the Simulation headers into more flat categories instead of the nested categories.

aymanhab:
More issues detected when compiling GUI:
All "Sets" are not wraped correctly by SWIG (check Model.get/set/upd BodySet, JointSet, .....)
Iterators: .all deref() method return SWIG proxy classes rather than native API classes.
There could be other issues but compilation stops after first 100 errors.

chrisdembia:
This could be an excellent opportunity to convert those failures into java test cases in opensim-core. Perhaps it makes sense for @klshrinidhi to build the GUI so he can see those error messages (without cycling with @aymanhab)? I'm happy to set up the GUI build on your computer, @klshrinidhi .

aymanhab:
Totally agree @chrisdembia I'd also revert the simulation library breakup into one interface file/cxx similar to python (even at the expense of compile time for now) until we rearrange our libraries into more balanced layout in future PRs.

klshrinidhi:
If there are too many errors, then maybe this task needs to be re-done. I am hesitant to take up the task again if there are no tests for me to verify my changes. Looking GUI will not work for me because I build way too many times with small changes.

chrisdembia:
The GUI takes maybe 20 seconds to compile.

klshrinidhi:
Thanks. I would like automated tests to verify my changes. It is not practical for my workflow to manually look at the GUI every time I make a change.

chrisdembia:
I agree it's impractical. This project is a great opportunity to convert the GUI's compilation failures into test cases in opensim-core (although it could be a grueling process).

aymanhab:
Just to be clear @klshrinidhi you don't need to launch the GUI, just build it and see if there're compilation failures. This can all be done from the command line with ant.

chrisdembia:
@klshrinidhi Independent of this PR, I'd be happy to show you how to build the GUI (for fun, etc.).

aymanhab:
@klshrinidhi I could take this over and the test cases myself but I'll need your help automating the process in CMake which will be helpful for the longer term.

klshrinidhi:
Yes. I can definitely help with CMake automation.

aymanhab:
Ok, @klshrinidhi please stop by when you have a chance and I'll describe the steps to get started. Thanks,

jenhicks:
Will break into 5 groups reflecting current opensim-core directory structure. In future will break (post-dev week) into larger number of logical groups.

aymanhab:
@chrisdembia there's no details of the reported failure on AppVeyor, is it just time out? Thanks

klshrinidhi:
Yes. It does look like a timeout.

chrisdembia:
You can tell if it's a timeout if the duration in the upper left says 59 minutes 59 seconds or 1 hour

aymanhab:
Thanks @chrisdembia and @klshrinidhi. The generated code now works with only minor changes so it's ready for review. Automation pending (Issue in opensim-gui repo). One more Java test case added.

aymanhab:
The failure on AppVeyor is due to Python wrapping/compilation taking too long while the Java wrapping finished and successfully ran the tests but they both could be competing for resources on the build machine.

aymanhab:
@klshrinidhi I restarted the appveyor build of the branch which timed out, although the integrated build seems to have passed (in 55 min). Unless I hear otherwise I'll merge this after the build passes so we don't keep chasing master changes in swig interface files that were changed significantly in this PR. 
@chrisdembia Can you look at the cmake files to make sure they fit our conventions? Thanks.

.travis.yml
```diff
@@ -10,7 +10,7 @@ matrix:
   include:
     - os: linux
       compiler: clang
-      env: BTYPE=RelWithDebInfo WRAP=on  DOXY=on  NPROC=1 DEPLOY=yes
+      env: BTYPE=RelWithDebInfo WRAP=off  DOXY=on  NPROC=1 DEPLOY=yes
```
chrisdembia: Wrapping is not being tested on travis; I thought the intent of this PR was to enable testing of wrapping on travis?

Bindings/Java/OpenSimJNI/CMakeLists.txt
```diff
@@ -4,20 +4,22 @@ set(UKIT JAVAJNI)
 # For building C++ wrapper.
 find_package(JNI 1.6 REQUIRED)
 
+function(GenerateJavaWrapper
```
chrisdembia: Our convention for naming cmake macros/functions would give `OpenSimGenerateJavaWrapper`.

chrisdembia: Also, it'd be helpful to indent the body of this function.

Bindings/Java/OpenSimJNI/CMakeLists.txt
```diff
@@ -30,21 +32,44 @@ add_custom_command(
             "${CMAKE_SOURCE_DIR}/Bindings/OpenSimHeaders_actuators.h"
             "${CMAKE_SOURCE_DIR}/Bindings/OpenSimHeaders_analyses.h"
             "${CMAKE_SOURCE_DIR}/Bindings/OpenSimHeaders_tools.h"
-    COMMENT "Generating java bindings source code with SWIG."
```
chrisdembia: These dependencies are incorrect, I think. Generating the wrapper for `java_simbody` doesn't depend on `OpenSimHeaders_tools.h`. See how I handle this for python: https://github.com/opensim-org/opensim-core/blob/master/Bindings/Python/CMakeLists.txt#L188.

chrisdembia:
I made a few comments about CMake. My only other comment is that it's slightly confusing that the "actuators" wrapper actually wraps "actuators," "analyses," and "tools". Perhaps it's fine though.

Bindings/Java/OpenSimJNI/CMakeLists.txt
chrisdembia: I'd like for this to be cleaned up but it's also fine if it's left as-is. We should revisit the handling of dependencies anyway, since as @klshrinidhi has noted, this list of dependencies is incomplete. So it's fine if we just leave this to be dealt with when we address #790.

aymanhab:
Thanks @chrisdembia for the great feedback. I'd definitely remove the obviously wrong dependencies and handle the missing dependencies when we address #790

aymanhab:
Timed out again!

klshrinidhi:
Maybe it is time to do the more logical split we were talking about right now ?
I can help.

aymanhab:
@klshrinidhi Looking at the build log, the test case testC3DFileAdapter by itself takes more than 6 min. is that expected? Also I'm not sure how the latest breakup of Java compares to python, that would be another place to look before taking on a major restructuring effort especially when we'r enot sure where things will end up long term.

klshrinidhi:
I have noticed that the test `testC3DFileAdapter` takes different amount of time on different platforms. My implementation is not different for different platforms. It is likely BTK does different things on different platforms. It takes only few seconds on my Ubuntu/Linux machine. I will look into it.

chrisdembia:
Let me clarify that our appveyor timeout used to be 2 hours (for many months); I had specifically requested an increase in the timeout. So, our builds had been hovering around 1hr, and had somewhat frequently exceeded 1 hour (maybe 1 hour 5 minutes).

When we purchased two additional concurrent builds, our timeout seemed to go back to the default 1 hour. I will ask if we can have the timeout increased.

The timeout is unrelated to this PR.

chrisdembia:
Okay, the timeout for our build has now been increased. I restarted the build.

klshrinidhi:
I have made changes to the CMake files as per comments.

To keep a record, here is split after this PR:

| num-lines | file |
| ----- | ------ |
|   83849 |  Bindings/Java/OpenSimJNI/java_actuators.cxx |
|   46714 |  Bindings/Java/OpenSimJNI/java_common.cxx |
|   26517  |  Bindings/Java/OpenSimJNI/java_simbody.cxx |
|  207895 |  Bindings/Java/OpenSimJNI/java_simulation.cxx |
|   96527  |  Bindings/Python/python_actuators_wrap.cxx |
|   25310  | Bindings/Python/python_analyses_wrap.cxx |
|   95293  | Bindings/Python/python_common_wrap.cxx |
|   96924  | Bindings/Python/python_simbody_wrap.cxx |
|  386010 | Bindings/Python/python_simulation_wrap.cxx |
|   48260 | Bindings/Python/python_tools_wrap.cxx |

chrisdembia:
Thanks for those CMake changes. @klshrinidhi do you plan to also update the dependencies for generating swig wrapping? @aymanhab said earlier


> I'd definitely remove the obviously wrong dependencies and handle the missing dependencies when we address #790

klshrinidhi:
I can fix dependencies in another PR as we split the files even more. I think this is one of the low hanging fruits to improve build time and build system at the same time.
We need to merge this now to avoid conflicts.

aymanhab:
@klshrinidhi Based on discussion with @chrisdembia regarding the issues with dependencies, I agree to move ahead and merge this PR once tests pass and address dependencies in a different PR. Thanks.

Issue #350, opened 2015-02-07
testCMCWithControlConstraintsRunningModel periodically fails on appveyor


chrisdembia:
The first time I started seeing this was with the recent Frames PR. I'm not saying the Frames PR caused it, though.

chrisdembia:
This is the message that I see on appveyor:

```
61/63 Test #62: testCMCWithControlConstraintsRunningModel ....***Exception: SegFault 63.70 sec
 
******************************************************************
*                     testRunningModel                     *
******************************************************************
 
AbstractTool runningModel_CMC_test loading model 'runningModel_CMC_test.osim'
Updating Model file from 20302 to latest format...
Loaded model modelModified from file runningModel_CMC_test.osim
Adding force object set from runningModel_CMC_Actuators.xml
Running tool runningModel_CMC_test.
Old version storage/motion file encountered
Storage: file=runningModel_GRF_data.mot (nr=691 nc=19)
.. assuming rotations in Degrees.
ExternalForce::RightFoot_ExternalForce
Data source being set to runningModel_GRF_data.mot
ExternalForce::LeftFoot_ExternalForce
Data source being set to runningModel_GRF_data.mot
```

chrisdembia:
Can someone on a windows machine run this test a few times to see if it fails periodically?

tkuchida:
Passed 4 consecutive times.

tkuchida:
The next line in the output is `taskSet size = 29`… maybe the task set file isn't being read properly? No hypothesis as to why that would cause a periodic failure, though.

chrisdembia:
Thanks for looking into this. Yeah...I don't know. And every time I've rerun the appveyor build after failure, it succeeds.

Maybe appveyor is just timing out and I didn't notice?

chrisdembia:
Now testCMCGait10dof18musc failed: https://ci.appveyor.com/project/opensim-org/opensim-core/build/192

```
AbstractTool walk_subject loading model 'gait10dof18musc_subject01.osim'
Updating Model file from 30000 to latest format...
Loaded model walk_subject01 from file gait10dof18musc_subject01.osim
Adding force object set from gait10dof_Reserve_Actuators.xml
Running tool walk_subject.
Storage: file=gait10dof18musc_subject01_walk_grf.mot (nr=1501 nc=19)
ExternalForce::right
Data source being set to subject01_walk1_grf.mot
ExternalForce::left
Data source being set to subject01_walk1_grf.mot
```

[...]

PR #908, opened 2016-04-06
Remove signed index in Connector classes

klshrinidhi:
* Change Connector classes to use `unsigned` index instead of `signed` index. This provides for a cleaner interface without need for `-1` as default index showing up in the interface of the function. `unsigned` also makes it clear to compiler and user that this number is always non-negative.
* The above change eliminates need to templatize `IndexOutOfRange`. Making it a non-template class required changes to other code.
* Fix some compiler warnings while at it.

I encountered this while working on #889.

### DIFF
--- a/OpenSim/Actuators/Test/testActuators.cpp
+++ b/OpenSim/Actuators/Test/testActuators.cpp
@@ -307,7 +307,7 @@ void testTorqueActuator()
     }
 
     // determine the initial kinetic energy of the system
-    double iKE = model->getMatterSubsystem().calcKineticEnergy(state);
+    /*double iKE = */model->getMatterSubsystem().calcKineticEnergy(state);
 
     RungeKuttaMersonIntegrator integrator(model->getMultibodySystem());
     integrator.setAccuracy(integ_accuracy);
@@ -322,7 +322,7 @@ void testTorqueActuator()
 
     model->computeStateVariableDerivatives(state);
 
-    double fKE = model->getMatterSubsystem().calcKineticEnergy(state);
+    /*double fKE = */model->getMatterSubsystem().calcKineticEnergy(state);
 
     // Change in system kinetic energy can only be attributable to actuator work
     //double actuatorWork = (powerProbe->getProbeOutputs(state))[0];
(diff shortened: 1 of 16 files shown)


OpenSim/Actuators/Test/testActuators.cpp
```diff
@@ -322,7 +322,7 @@ void testTorqueActuator()
 
     model->computeStateVariableDerivatives(state);
 
-    double fKE = model->getMatterSubsystem().calcKineticEnergy(state);
+    /*double fKE = */model->getMatterSubsystem().calcKineticEnergy(state);
```
aseth1: The difference in final and initial KE should be used to compare to the actuator work. This is a major part of the test.

klshrinidhi: @aseth1 
This variable is currently unused. Maybe the author forgot to use it to test. Feel free to correct it by adding the appropriate test if you like. If you think it should be another PR, :+1:.

klshrinidhi:
Appveyor build failing with test `testCMCWithControlConstraintsGait2354` :disappointed:.

chrisdembia:
That's another intermittent appveyor failure (#350). Restart the build and it'll almost surely pass.

klshrinidhi:
Thanks. I just restarted. I :pray: it works. This is probably the 3rd time I restarted.

OpenSim/Actuators/Test/testActuators.cpp
aseth1: The test is there (few lines below) but have been commented out. I added those tests long ago- it might be it needed to be revised- but should not have been omitted altogether.

Not the issue you were addressing, I realize, but I don't see the point of commenting out the now unused variables when the underlying problem is that the test was gutted.

I raised a separate issue #910. It would be a good exercise @klshrinidhi for you to tackle after.

aseth1:
@klshrinidhi I am happy with the change to `unsigned int`
@chrisdembia did you have any reason for using an `int` we may have overlooked? If not, feel free to merge.

chrisdembia:
> @chrisdembia did you have any reason for using an int we may have overlooked?

No. Looks good.

PR #909, opened 2016-04-06
Attempt to fix appveyor error creating Release directory

### LINKED ISSUE #870: Appveyor build using two different variables for installation dir
In appveyor.yml, OPENSIM_HOME and OPENSIM_INSTALL_DIR should have the same value. They are different right now, which causes the upload of the OpenSim binaries to Appveyor to fail. This doesn't show up in any PR's because it only happens when building the master branch (after a merge).

chrisdembia:
This PR attempts to fix the issue with Appveyor where MSBuild has trouble creating the Release directory.

This PR also fixes #870


chrisdembia:
@klshrinidhi could you review/merge?

klshrinidhi:
If the problem was because multiple processes were *racing* to create the directory, then :+1: :green_heart: :smile:.

chrisdembia:
I don't know if that was the issue, but I think this solves it nonetheless, as somewhat of a hack.

chrisdembia:
Thank you for merging.

PR #911, opened 2016-04-06
Update Simbody commit-id for superbuild.

klshrinidhi:
This will pull the latest Simbody which has fixes for compiler warnings we see in Simbody headers while compiling OpenSim.


chrisdembia:
Please merge either immediately or after the tests pass (tests don't use superbuild for simbody).

Issue #902, opened 2016-03-30
Enable additional builds with Appveyor


chrisdembia:
We now have 4 concurrent appveyor jobs. We can use this to run a separate BTK build on Appveyor, or a separate Debug build, or a 32bit build. @klshrinidhi do you want to choose and add some more Appveyor jobs?

tkuchida:
Added a 32-bit build in #921. Should we add a Debug bulid?

chrisdembia:
No, I don't think so; at least not now. We have a Debug build on Travis.

Issue #784, opened 2016-02-27
testStaticOptimization and testCMCGait10dof18musc fail on fresh build


tkuchida:
(Win7, VS2015 32-bit, Release mode) I followed the instructions in README.md including the instructions for superbuild (not [this one](http://www.superbuild.co.nz/)), with the following exception: in step 9, running `cmake --build . --config Release` returned `Error: could not find CMAKE_MAKE_PROGRAM in Cache` so I built in Visual Studio instead. All but two tests pass; output follows.

_testStaticOptimization_
```
...
Loading coordinates from file walk_subject01_Kinematics_q.sto.
Storage: file=walk_subject01_Kinematics_q.sto (nr=2537 nc=11)
Found 2536 state vectors with time stamps ranging from 0 to 2.499.
Executing the analyses from 0.5 to 1.3...
SimTK Exception thrown at InteriorPointOptimizer.cpp:264:
  Optimizer failed: Ipopt: Maximum iterations exceeded (status -1)
OPTIMIZATION FAILED...
...
```

_testCMCGait10dof18musc_
```
...
================================================================
Computing initial values for muscles states (activation, length)
Start time = Sat Feb 27 00:41:56 2016

================================================================
CMC.computeControls:  t = 0.6
CMC::computeControls WARNING- small force range for iliopsoas_r (0 to 0)

CMC::computeControls WARNING- small force range for rect_fem_r (0 to 0)

CMC::computeControls WARNING- small force range for gastroc_r (0 to 0)

CMC::computeControls WARNING- small force range for soleus_r (0 to 0)

CMC::computeControls WARNING- small force range for iliopsoas_l (0 to 0)

CMC::computeControls WARNING- small force range for rect_fem_l (0 to 0)

CMC::computeControls WARNING- small force range for gastroc_l (0 to 0)

CMC::computeControls WARNING- small force range for soleus_l (0 to 0)

SimTK Exception thrown at InteriorPointOptimizer.cpp:264:
  Optimizer failed: Ipopt: Infeasible problem detected (status 2)
OPTIMIZATION FAILED...
...
```

chrisdembia:
> (not this one)

haha

Hmm...you built 64-bit libraries or 32-bit libraries (I'm not sure if "VS2015 32-bit" refers to the fact that VS itself is 32-bit)? We only test 64-bit libraries on Appveyor, so maybe it's because you might be building as 32-bit?

tkuchida:
[...]

PR #912, opened 2016-04-08
Add 32-bit build to AppVeyor

tkuchida:
See #902

TODO:
- [ ] Open issues to resolve known test failures on 32-bit Windows and remove them from the `OPENSIM_TESTS_TO_EXCLUDE` and `PYTHON_TESTS_TO_EXCLUDE` environment variables in appveyor.yml.


appveyor.yml
```diff
@@ -44,7 +51,7 @@ install:
   - choco install swig --version 3.0.5 -y
   
   ## Install python-nose for python testing.
-  - C:\Python27-x64\Scripts\pip install nose
+  - %PYTHON_DIR%\Scripts\pip install nose
```
chrisdembia: Maybe put the whole path in quotes.

chrisdembia: `"%PYTHON_DIR%\Scripts\pip"`

tkuchida: Thanks. I found [this page](http://help.appveyor.com/discussions/problems/2869-appdata-causes-error-parsing-appveyoryml-while-scanning-for-the-next-token-find-character-that-cannot-start-any-token), which uses single-quotes, but I'll try double-quotes… must be twice as good.

appveyor.yml
```diff
@@ -21,13 +21,20 @@ os: Visual Studio 2015
 
 platform: x64
 
+environment:
+  matrix:
+    - PYTHON_DIR: "C:\\Python27"
+      CMAKE_GENERATOR: "Visual Studio 14 2015"
+    - PYTHON_DIR: "C:\\Python27-x64"
+      CMAKE_GENERATOR: "Visual Studio 14 2015 Win64"
+
```
klshrinidhi: I think we need not specify a generator for CMake if we want to build 32-bit. It is the default generator chosen anyway with MSVC.
It is still good to keep it explicit so readers of this file can know what is being used.

klshrinidhi:
All :x: :anguished:

appveyor.yml
chrisdembia: > It is still good to keep it explicit so readers of this file can know what is being used.

I agree; the Appveyor machines may have multiple versions of MSVC also.

tkuchida:
Need to get 32-bit version of Simbody built and uploaded first. Working on it [here](https://github.com/simbody/simbody/tree/appveyor-add-32bit).

tkuchida:
Status:
- 64-bit build successful; all tests pass
- 32-bit build successful; testC3DFileAdapter and python_Release segfaulted (the latter seems to be related to testC3DFileAdapter also)

@chrisdembia - Can you tell whether [the segfaults](https://ci.appveyor.com/project/opensim-org/opensim-core/build/3120/job/9spibei8tx6mppnq) are due to a problem with the yml file? If not, perhaps this PR is ready for review and a separate issue should be raised for the test failures. Unfortunately, these failures are different from the ones reported in #784.

chrisdembia:
We are unable to see the output of the failed tests because our appveyor build cuts off at 1 hour (well, 59min 59sec). This also means that there may be more tests failures than we currently can see. 

The issue may be a bug with BTK when building it as 32bit. The only potential issue with the yml file I think would have been if you did not properly set the generator for the superbuild, but you did. 

> perhaps this PR is ready for review and a separate issue should be raised for the test failures.

I agree; the C3D failure should be debugged locally, not in this PR. but I think you should also use CTEST to avoid running the tests with the known failures on the 32-bit build. [We already do this on travis to exclude tests that are known to fail in debug](https://github.com/opensim-org/opensim-core/blob/master/.travis.yml#L158); search for all occurrences of "TESTS_TO_EXCLUDE." I think it'd be cleaner to do this as an environment variable as part of `matrix` in `appveyor.yml`, instead of the messy way I chose to do it for travis.

tkuchida:
Ready for review (AppVeyor timed out on one of the two x64 jobs). Perhaps wait for review/merge of [associated Simbody PR](https://github.com/simbody/simbody/pull/481) before merging this one.

chrisdembia:
Looks good to me. Did you notice that the 32-bit build takes almost half the amount of time on AppVeyor? (33 minutes vs over an hour)? I wonder why that is.

tkuchida:
> Did you notice that the 32-bit build takes almost half the amount of time on AppVeyor? (33 minutes vs over an hour)? I wonder why that is.

[The most recent AppVeyor build on master](https://ci.appveyor.com/project/opensim-org/opensim-core/branch/master) took 55½ minutes, so this PR hasn't affected the duration of the 64-bit build.

chrisdembia:
> The most recent AppVeyor build on master took 55½ minutes, so this PR hasn't affected the duration of the 64-bit build.

I should have clarified. I know that the 64-bit build takes about an hour. That's why I was happily surprised to see the 32-bit build taking much less time. I don't understand why this is though.

chrisdembia:
Looking at the appveyor build log, the speedup mostly comes from a reduced amount of time required for compiling (the tests take roughly the same amount of time). If you hover over the line number in the appveyor log, the hovertext tells you the timestamp for that line.

tkuchida:
There are 17 more build warnings on 64-bit compared to 32-bit (261 vs. 244).

chrisdembia:
Oh weird...

klshrinidhi:
I think this PR should wait and see if PR #916 fixes the 32-bit `C3DFileAdapter` failure issue. It would be good to not exclude the test.

tkuchida:
@chrisdembia Did you intend to merge this? I think it should at least wait until the corresponding PR has been merged in Simbody.

klshrinidhi:
Remember to nuke ( :fire: ) the branch when you are done.

chrisdembia:
No I did not intend to merge this. And I didn't, at least not manually. GitHub has counted this as merged when I merged #916. However, the changes are reverted (on master).

chrisdembia:
The reason GitHub counted this as merged is that @klshrinidhi merged your branch into his branch (at my request). So I think one way to proceed is to create a new branch off of `appveyor-add-32bit`, and create a new PR for that?

PR #915, opened 2016-04-10
New Point Component base class

aseth1:
Added `Point` as the location/translation only analog of `Frame` and attempted to make their calls consistent. Since we are adding Outputs to Station I thought it would be good time to introduce Point and let it provide the underlying structure. It was also evident that Frame should calculate and provide the velocity and acceleration of the Frame and that points (like Station) should use the Frame calculations to do most of the work.

Also incorporated from #873 into `testPoints.cpp` separate from `testFrames`.

The intention is to use Point to also PathPoints (`ConditionalPathPoint`, `PathWrapPoint`, `MovingPathPoint`) and enable new Points of interest to be computed and easily reported, for example, the center of pressure.

### DIFF
--- a/OpenSim/Simulation/Model/Frame.h
+++ b/OpenSim/Simulation/Model/Frame.h
@@ -95,6 +95,10 @@ public:
         SimTK::Stage::Position);
     OpenSim_DECLARE_OUTPUT(transform, SimTK::Transform, getTransformInGround,
         SimTK::Stage::Position);
+    OpenSim_DECLARE_OUTPUT(velocity, SimTK::SpatialVec, getVelocityInGround,
+        SimTK::Stage::Velocity);
+    OpenSim_DECLARE_OUTPUT(acceleration, SimTK::SpatialVec, getAccelerationInGround,
+        SimTK::Stage::Acceleration);
 
     /** @name Spatial Operations for Frames
     These methods allow access to the frame's transform and some convenient
@@ -235,22 +239,10 @@ public:
         const SimTK::Vec3 scale = SimTK::Vec3(1));
 
 protected:
-    /** @name Component Extension methods.
-        Frame types override these Component methods. */
-    /**@{**/
-    void extendAddToSystem(SimTK::MultibodySystem& system) const override;
-    void extendRealizeTopology(SimTK::State& s) const override;
-
-    /// override default extendAddGeometry to set Frame to this Object
-    void extendAddGeometry(OpenSim::Geometry& geom) override;
-
-    /**@}**/
-
-private:
-    /** @name Extension methods.
+    /** @name Extension of calculations of Frame kinematics.
     Concrete Frame types must override these calculations.
     Results of the calculations are cached by the Frame and made accessible
-    via the corresponding getTransformInGround, getVelocityInGround, 
+    via the corresponding getTransformInGround, getVelocityInGround,
     getAccelerationInGround public methods (above). */
     /**@{**/
 
@@ -262,18 +254,30 @@ private:
     virtual SimTK::Transform
         calcTransformInGround(const SimTK::State& state) const = 0;
 
-    /** The spatial velocity {omega; v} for this Frame in ground. */
+    /** The spatial velocity {omega; v} of this Frame in ground. */
     virtual SimTK::SpatialVec
         calcVelocityInGround(const SimTK::State& state) const = 0;
 
     /** The spatial acceleration {alpha; a} for this Frame in ground */
     virtual SimTK::SpatialVec
         calcAccelerationInGround(const SimTK::State& state) const = 0;
+    /**@}**/
 
+    /** @name Component Extension methods.
+    Frame types override these Component methods. */
+    /**@{**/
+    void extendAddToSystem(SimTK::MultibodySystem& system) const override;
+    void extendRealizeTopology(SimTK::State& s) const override;
+
+    /// override default extendAddGeometry to set Frame to this Object
+    void extendAddGeometry(OpenSim::Geometry& geom) override;
+    /**@}**/
+
+private:
     /** Extend how concrete Frame determines its base Frame. */
     virtual const Frame& extendFindBaseFrame() const = 0;
     virtual SimTK::Transform extendFindTransformInBaseFrame() const = 0;
-    /**@}**/
+
 
     SimTK::ResetOnCopy<SimTK::CacheEntryIndex> _transformIndex;
     SimTK::ResetOnCopy<SimTK::CacheEntryIndex> _velocityIndex;
--- a/OpenSim/Simulation/Model/Frame.h
+++ b/OpenSim/Simulation/Model/Frame.h
@@ -222,7 +222,8 @@ public:
     /** Add a Mesh specified by file name to the list of Geometry owned by the Frame.
     Scale defaults to 1.0 but can be changed on the call line for convenience.
     */
-    void attachMeshGeometry(const std::string &aGeometryFileName, const SimTK::Vec3 scale = SimTK::Vec3(1));
+    void attachMeshGeometry(const std::string &aGeometryFileName,
+        const SimTK::Vec3 scale = SimTK::Vec3(1));
     /** Add a piece of Geometry to the list of Geometry owned by the Frame.
     This function is a convenience for ModelComponent::addGeometry() for the case 
     of adding Geometry directly to a Frame, and thus sets the "frame name" of the 
@@ -230,7 +231,8 @@ public:
     own a copy of the Geometry so any changes you make to geom after calling this 
     method will not have an effect.
     */
-    void attachGeometry(const OpenSim::Geometry& geom, const SimTK::Vec3 scale = SimTK::Vec3(1));
+    void attachGeometry(const OpenSim::Geometry& geom,
+        const SimTK::Vec3 scale = SimTK::Vec3(1));
 
 protected:
     /** @name Component Extension methods.
@@ -246,7 +248,10 @@ protected:
 
 private:
     /** @name Extension methods.
-        Concrete Frame types must override these methods. */
+    Concrete Frame types must override these calculations.
+    Results of the calculations are cached by the Frame and made accessible
+    via the corresponding getTransformInGround, getVelocityInGround, 
+    getAccelerationInGround public methods (above). */
     /**@{**/
 
     /** Calculate the transform related to this Frame with respect to ground.
--- a/OpenSim/Simulation/Model/Frame.h
+++ b/OpenSim/Simulation/Model/Frame.h
@@ -106,7 +106,7 @@ public:
     It transforms quantities expressed in F into quantities expressed
     in G. This is mathematically stated as:
         vec_G = X_GF*vec_F ,
-    where X_GF is the transform returned by getGroundTransform.
+    where X_GF is the transform returned by getTransformInGround.
 
     @param state       The state applied to the model when determining the
                        transform.
--- a/OpenSim/Simulation/Model/Frame.h
+++ b/OpenSim/Simulation/Model/Frame.h
@@ -93,7 +93,7 @@ public:
     //=============================================================================
     OpenSim_DECLARE_OUTPUT(position, SimTK::Vec3, getPositionInGround,
         SimTK::Stage::Position);
-    OpenSim_DECLARE_OUTPUT(transform, SimTK::Transform, getGroundTransform,
+    OpenSim_DECLARE_OUTPUT(transform, SimTK::Transform, getTransformInGround,
         SimTK::Stage::Position);
 
     /** @name Spatial Operations for Frames
@@ -112,7 +112,25 @@ public:
                        transform.
     @return transform  The transform between this frame and the ground frame
     */
-    const SimTK::Transform& getGroundTransform(const SimTK::State& state) const;
+    const SimTK::Transform&
+        getTransformInGround(const SimTK::State& state) const;
+
+    /** The spatial velocity V_GF {omega; v} for this Frame in ground.
+        It can be used to compute the velocity of any stationary point on F,
+        located at r_F (Vec3), in ground, G, as:
+            v_G = V_GF(0)*r_F + V_GF(1);
+        Is only valid at Stage::Velocity or higher. */
+    const SimTK::SpatialVec&
+        getVelocityInGround(const SimTK::State& state) const;
+
+    /** The spatial acceleration A_GF {alpha; a} for this Frame in ground.
+        It can also be used to compute the acceleration of any stationary point
+        on F, located at r_F (Vec3), in ground, G, as:
+            a_G = A_GF(0)*r_F + A_GF(1);
+        Is only valid at Stage::Acceleration or higher. */
+    const SimTK::SpatialVec&
+        getAccelerationInGround(const SimTK::State& state) const;
+
 
     /**
     Find the transform that describes this frame (F) relative to another
@@ -195,7 +213,7 @@ public:
 
     /** Accessor for position of the origin of the Frame in Ground. */
     SimTK::Vec3 getPositionInGround(const SimTK::State& state) const {
-        return getGroundTransform(state).p();
+        return getTransformInGround(state).p();
     };
 
     // End of Base Frame and Transform accessors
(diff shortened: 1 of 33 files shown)


OpenSim/Simulation/Model/Station.cpp
```diff
@@ -107,8 +109,27 @@ void Station::setReferenceFrame(const OpenSim::PhysicalFrame& aFrame)
...
+    Vec3 r = getReferenceFrame().getTransformInGround(s).R()*get_location();
+    const SimTK::SpatialVec& V_GF = getReferenceFrame().getVelocityInGround(s);
+    return V_GF[1] + V_GF[0] % r;
+}
+
+SimTK::Vec3 Station::calcAccelerationInGround(const SimTK::State& s) const
+{
+    const SimTK::SpatialVec& V_GF = getReferenceFrame().getVelocityInGround(s);
+    const SimTK::SpatialVec& A_GF = getReferenceFrame().getAccelerationInGround(s);
+    Vec3 r = getReferenceFrame().getTransformInGround(s).R()*get_location();
```
jimmyDunne: Can you add some comments on how these are calculated?

OpenSim/Simulation/Model/Point.h
```diff
@@ -0,0 +1,139 @@
...
+ * SimTK::Stage (i.e. Position, Velocity and Acceleration). 
+ *
+ * @author Ajay Seth
+ */
+class OSIMSIMULATION_API Point : public ModelComponent {
+OpenSim_DECLARE_ABSTRACT_OBJECT(Point, ModelComponent);
+
+public:
+    OpenSim_DECLARE_OUTPUT(location, SimTK::Vec3, getLocationInGround,
+        SimTK::Stage::Position);
```
klshrinidhi: Is this *output* intended to call `getLocationInGround` or `calcLocationInGround` like other outputs ?

aseth1: they all should be `get...InGround` since these are the cached values and avoids unnecessary recalculation. That was the mistake I noticed earlier .. will push the update shortly

aseth1:
@jimmyDunne let me know if the added comments are helpful

OpenSim/Simulation/Model/Point.h
```diff
@@ -0,0 +1,137 @@
...
+// INCLUDE
+#include <OpenSim/Simulation/osimSimulationDLL.h>
+#include <OpenSim/Simulation/Model/ModelComponent.h>
+
+namespace OpenSim {
+
+//=============================================================================
+//=============================================================================
+/**
+ * A Point is an OpenSim abstraction for any point location in space. Points
```
chrisdembia: I think you could remove "location."

OpenSim/Simulation/Model/Point.h
```diff
@@ -0,0 +1,137 @@
...
+    OpenSim_DECLARE_OUTPUT(acceleration, SimTK::Vec3, getAccelerationInGround,
+        SimTK::Stage::Acceleration);
+    //--------------------------------------------------------------------------
+    // CONSTRUCTION
+    //--------------------------------------------------------------------------
+    Point();
+
+    virtual ~Point() {};
+
+    /** @name Spatial Operations for Point
```
chrisdembia: By Spatial, do you mean Cartesian? To me, Spatial means two Vec3's.

OpenSim/Simulation/Model/Point.h
```diff
@@ -0,0 +1,137 @@
...
+    Convenient access to spatial kinematics of Points.*/
```
chrisdembia: This comment is redundant with title for the group.

OpenSim/Simulation/Model/Station.h
```diff
@@ -37,38 +36,56 @@ class Body;
...
-OpenSim_DECLARE_CONCRETE_OBJECT(Station, ModelComponent);
+class OSIMSIMULATION_API Station : public Point {
+OpenSim_DECLARE_CONCRETE_OBJECT(Station, Point);
 public:
     //==============================================================================
     // PROPERTIES
     //==============================================================================
+    /** @name Property declarations
+    These are the serializable properties associated with a Station. **/
+    /**@{**/
```
chrisdembia: @aseth1 we do not use these groups for properties anymore. I made changes to the property macros so that the properties are automatically grouped together. Please remove the doxygen group.

OpenSim/Simulation/Model/Point.h
```diff
@@ -0,0 +1,137 @@
...
+                       acceleration of the Point.
+    @return velocity   The acceleration of the point expressed in Ground */
+    const SimTK::Vec3& getAccelerationInGround(const SimTK::State& state) const;
+
+    // End of Point's Spatial Kinematics
+    ///@}
+
+protected:
+    /** @name Point Extension methods.
+    Concrete Point types must override these methods. */
```
chrisdembia: Can you add a comment that Point takes care of caching the return value of these functions? This is useful to those implementing their own Points, so they know not to do their own caching.

OpenSim/Simulation/Model/Station.cpp
```diff
@@ -40,13 +40,21 @@ using SimTK::Vec3;
 /**
  * Default constructor.
  */
-Station::Station() :
-   ModelComponent()
+Station::Station() : Super()
```
chrisdembia: No need to change, but we don't use this convention in general (of using the `Super` typedef in constructors); we should be consistent.

OpenSim/Simulation/Model/Point.h
```diff
@@ -0,0 +1,137 @@
...
+    const SimTK::Vec3& getLocationInGround(const SimTK::State& state) const;
+
+    /** The velocity v_G of this Point expressed in ground.
+        Is only valid at Stage::Velocity or higher.
+    @param state       The state applied to the model when determining the
+                       velocity of the Point.
+    @return velocity   The velocity of the point expressed in the Ground. */
+    const SimTK::Vec3& getVelocityInGround(const SimTK::State& state) const;
+
+    /** The acceleration a_G, of this Point expressed in ground.
```
chrisdembia: The correct Simbody notation is probably `a_GP` (https://simtk.org/api_docs/simbody/3.5/classSimTK_1_1SimbodyMatterSubsystem.html#a84cc002312ec1dda2a7382a214253095). I think these comments should be as precise as the ones for the `calc` methods (with respect to and expressed in).

OpenSim/Simulation/Model/Station.h
```diff
@@ -37,38 +36,56 @@ class Body;
...
 /**
- * A class implementing a Station. A Station is a point fixed to and defined with 
- * respect to a reference frame. The reference frame can be a Body (fixed 
- * to origin of a Body), a FixedFrame (affixed to another PhysicalFrame) or any 
- * other PhysicalFrame. The main functionality provided by Station is to find its 
- * location in any Frame.
+ * A Station is a Point fixed to and located on a PhysicalFrame, which can be a 
+ * Body (w.r.t. the Body origin), Ground or any PhysicalOffsetFrame.
+ * A Station is an analogous physical Point to a PhyscialFrame where joints,
+ * constraints and forces can be attached and/or applied.
```
chrisdembia: This sentence is confusing to me. Consider: "Stations are similar to PhysicalFrames in that constraints and forces can be specified between stations."

OpenSim/Simulation/Model/Station.h
```diff
@@ -37,38 +36,56 @@ class Body;
 //=============================================================================
 //=============================================================================
 /**
- * A class implementing a Station. A Station is a point fixed to and defined with 
- * respect to a reference frame. The reference frame can be a Body (fixed 
- * to origin of a Body), a FixedFrame (affixed to another PhysicalFrame) or any 
- * other PhysicalFrame. The main functionality provided by Station is to find its 
- * location in any Frame.
+ * A Station is a Point fixed to and located on a PhysicalFrame, which can be a 
+ * Body (w.r.t. the Body origin), Ground or any PhysicalOffsetFrame.
```
chrisdembia: The `(w.r.t. ...)` is potentially confusing and could be removed.

OpenSim/Simulation/Model/Station.h
```diff
@@ -37,38 +36,56 @@ class Body;
...
+    SimTK::Vec3 findLocationInFrame(const SimTK::State& s,
+                                    const OpenSim::Frame& frame) const;
 private:
+    /** Calculate this Station's location in Ground */
+    SimTK::Vec3 calcLocationInGround(const SimTK::State& state) const override;
+    /** Calculate the velocity of this Station with respect to and expressed in
+        ground as a function of the state. */
+    SimTK::Vec3 calcVelocityInGround(const SimTK::State& state) const override;
+    /** Calculate the acceleration of this Station with respect to and 
+        expressed in ground as a function of the state. */
```
chrisdembia: IMO the 3 doxygen blocks above do not add anything. I think comments here are only necessary if they describe some quirk about this specific implementation of the virtual methods.

OpenSim/Tests/Components/testComponents.cpp
```diff
@@ -53,6 +53,13 @@ int main()
         availableComponents.push_back(availableFrames[i]);
     }
 
+    // starting with type Frame
+    ArrayPtrs<Point> availablePoints;
+    Object::getRegisteredObjectsOfGivenType(availablePoints);
+    for (int i = 0; i < availablePoints.size(); ++i) {
+        availableComponents.push_back(availablePoints[i]);
+    }
```
chrisdembia: +1

jimmyDunne:
Comments were really helpful in understanding what the variables are and how the calculation was getting made! Thanks!

OpenSim/Simulation/Model/Station.cpp
```diff
@@ -107,8 +109,42 @@ void Station::setReferenceFrame(const OpenSim::PhysicalFrame& aFrame)
...
+    const SimTK::SpatialVec& A_GF = getReferenceFrame().getAccelerationInGround(s);
+    // compute the local position vector of the point in its reference frame
+    // expressed in ground
+    Vec3 r = getReferenceFrame().getTransformInGround(s).R()*get_location();
+
+    // The acceleration of the station in ground is a function of its frame's
+    // linear (aF = A_GF[1]) and angular (alpha = A_GF[0]) accelerations and
+    // Corriolois acceleration due to the angular velocity (omega = V_GF[0]) of
+    // its frame, such that: a = aF + alphaF x r + omegaF x (omegaF x r)
+    return A_GF[1] + A_GF[0]%r +  V_GF[0] % (V_GF[0] % r) ;
```
chrisdembia: I would prefer that the location expressed in the mobilized body is cached in `finalizeFromProperties()` (or something like that) and that you call `findStationAccelerationInGround()`. Similar for velocity.

chrisdembia: Simbody is the dynamics engine. OpenSim is a modeling platform, etc.

OpenSim/Simulation/Model/Point.cpp
```diff
@@ -0,0 +1,129 @@
...
+            updCacheEntry(s, _locationIndex)).upd()
+                = calcLocationInGround(s);
+        // mark cache as up-to-date
+        getSystem().getDefaultSubsystem().
+            markCacheValueRealized(s, _locationIndex);
+    }
+    return SimTK::Value<SimTK::Vec3>::downcast(
+        getSystem().getDefaultSubsystem().
+        getCacheEntry(s, _locationIndex)).get();
+}
```
chrisdembia: Is this really the way that cache variables are to be handled? It is very verbose. Why not use `Component::isCacheVariableValid()` and `Component::markCacheVariableAsValid()`? If your concern is performance, then the interface for cache variables in Component should be updated as a separate issue, and this code should still just use the Component's current interface.

OpenSim/Simulation/Model/OffsetFrame.h
```diff
@@ -242,14 +248,28 @@ void OffsetFrame<C>::constructConnectors()
...
+// FRAME COMPUTATIONS must be specialized by concrete derived Offsets
 //=============================================================================
 // Implementation of Frame interface by OffsetFrame.
 template <class C>
 SimTK::Transform OffsetFrame<C>::
-calcGroundTransform(const SimTK::State& s) const
+calcTransformInGround(const SimTK::State& state) const
 {
-    return getParentFrame().getGroundTransform(s)*getOffsetTransform();
+    return SimTK::Transform(SimTK::Vec3(SimTK::NaN));
```
chrisdembia: Why can this no longer be implemented generically for all OffsetFrames?

OpenSim/Simulation/Model/OffsetFrame.h
```diff
@@ -242,14 +248,28 @@ void OffsetFrame<C>::constructConnectors()
...
+calcVelocityInGround(const SimTK::State& state) const
+{
+    return SimTK::SpatialVec(SimTK::Vec3(SimTK::NaN));
+}
+
+template <class C>
+SimTK::SpatialVec OffsetFrame<C>::
+calcAccelerationInGround(const SimTK::State& state) const
+{
+    return SimTK::SpatialVec(SimTK::Vec3(SimTK::NaN));
```
chrisdembia: Why do the two methods above need to be implemented in `OffsetFrame`? Could OffsetFrame just not implement them?

OpenSim/Simulation/Model/Frame.h
```diff
@@ -237,14 +255,24 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Frame, ModelComponent);
...
-        calcGroundTransform(const SimTK::State& state) const = 0;
+        calcTransformInGround(const SimTK::State& state) const = 0;
+
+    /** The spatial velocity {omega; v} for this Frame in ground. */
+    virtual SimTK::SpatialVec
+        calcVelocityInGround(const SimTK::State& state) const = 0;
+
+    /** The spatial acceleration {alpha; a} for this Frame in ground */
+    virtual SimTK::SpatialVec
+        calcAccelerationInGround(const SimTK::State& state) const = 0;
```
chrisdembia: See my related comment about mentioning that Frame takes care of the caching; no need for implementers to cache.

chrisdembia:
I am done reviewing. You may not agree with some of my comments; it's fine with me if some of those things are left as-is.

The Point class will be *really* helpful! It's very exciting to think about how it may clean up classes like GeometryPath.

OpenSim/Simulation/Model/Station.h
```diff
@@ -37,38 +36,56 @@ class Body;
...
+
+    /** Convenience constructor
+    @param[in] frame     PhysicalFrame in which the Station is located.
+    @param[in] location  Vec3 location of the station in its PhysicalFrame */
+    Station(const PhysicalFrame& frame, const SimTK::Vec3& location);
+
     virtual ~Station();
     /** getter of Reference Frame off which the Station is defined */
-    const OpenSim::PhysicalFrame& getReferenceFrame() const;
+    const PhysicalFrame& getReferenceFrame() const;
```
chrisdembia: I personally think we could have a better name for this method; I think `getFrame()` is fine (and the corresponding connector would also be `frame`). The main issue is that we don't use the term `ReferenceFrame` elsewhere in the code, so users may think it is something different from a Frame. I remember discussing this with @aymanhab and he had a reason for why he chose ReferenceFrame. @aymanhab, do you remember?

OpenSim/Simulation/Model/Station.cpp
chrisdembia: Since the Station test is pretty rigorous and checks against Simbody's method, I suppose this implementation is fine. For Frames though; is there an equally rigorous test against Simbody for the new kinematics equations?

chrisdembia:
It might also be good to update the CHANGELOG.md.

OpenSim/Simulation/Model/Station.h
aymanhab: Too bad, way too many discussions long ago that I forgot, feel free to rename as you see fit in the current context.

OpenSim/Simulation/Model/OffsetFrame.h
aseth1: Good point. Intermediate implementations had me move it out early on, but now that the OffsetFrame calculation are only dependent on its parent frame kinematics, the calcualtions have been restored on `OffsetFrame` and removed from `PhysicalOffsetFrame`. Thanks!

aseth1:
@chrisdembia thank you for the detailed review and insightful comments. The PR is greatly improved. Please let me know if you have other concerns.

chrisdembia:
🎆  Sorry for the delay.

aseth1:
Thank you!

PR #920, opened 2016-04-12
Add some useful methods to DataTable

klshrinidhi:
Adding some useful methods to `DataTable`. These methods can potentially be used in classes like `TableSource` which I am working on and is described in #889.


Bindings/common.i
```diff
@@ -93,6 +94,13 @@
 %ignore OpenSim::AbstractDataTable::setIndependentMetaData;
 %ignore OpenSim::AbstractDataTable::getDependentsMetaData;
 %ignore OpenSim::AbstractDataTable::setDependentsMetaData;
+%ignore OpenSim::AbstractDataTable::setColumnLabels(
+                                     const std::initializer_list<std::string>&);
+%extend OpenSim::AbstractDataTable {
+    void setColumnLabels(const std::vector<std::string>& columnLabels) {
+        $self->setColumnLabels(columnLabels);
+    }
+}
```
chrisdembia: +1

chrisdembia:
Looks good to me. Just one minor comment: What do you think about rearranging the methods in AbstractDataTable so that the "column label" methods appear above the "metadata" methods in doxygen? Also, what about using doxygen groups to group all the "column label" methods, etc? Similar grouping could also be useful in DataTable (row accessors/mutators, column accessors/mutators, matrix/submatrix accessors/mutators).

OpenSim/Common/DataTable.h
```diff
@@ -178,46 +178,175 @@ class AbstractDataTable {
...
     }
 
-    /** Set column labels.                                                    */
-    void setColumnLabels(const std::vector<std::string>& columnLabels) {
+    /** Get column label of a given column.                                   
+
+    \throws ColumnIndexOutOfRange If columnIndex is out of range of number of
+                                  columns.                                    
+    \throws KeyNotFound If column labels have not be set for the table.       */
+    const std::string& getColumnLabel(const size_t columnIndex) const {
```
aseth1: In this and the previous (and all column wise access) methods, It should be clearly documented that the columns indices, or iterate are only across the dependent columns of the table, with the independent column having its own accessors..

OpenSim/Common/DataTable.h
```diff
@@ -341,38 +485,151 @@ class DataTable_ : public AbstractDataTable {
         return _indData;
     }
 
-    /** Get dependent column.                                                 
+    /** Get dependent column at index.
 
-    \throws ColumnIndexOutOfRange If index is out of range.                   */
+    \throws ColumnIndexOutOfRange If index is out of range for number of columns
+                                  in the table.                               */
     VectorView getDependentColumnAtIndex(size_t index) const {
```
aseth1: Maybe use "dependent" for column label accessors as well, to be consistent with accessing the data.

OpenSim/Common/DataTable.h
```diff
@@ -341,38 +485,151 @@ class DataTable_ : public AbstractDataTable {
...
+    VectorView updDependentColumn(const std::string& columnLabel) {
+        return _depData.updCol(static_cast<int>(getColumnIndex(columnLabel)));
+    }
+
+    /** %Set independent column at index.                                      
 
     \throws RowIndexOutOfRange If index is out of range.                        
     \throws InvalidRow If this operation invalidates the row. Validation is
                        performed by derived classes.                          */
     void setIndependentColumnAtIndex(size_t index, const ETX& value) {
```
aseth1: I think this should be renamed to `setIndependentValueAtIndex` with a comment like: set the value at a specified index in the independent column of the `DataTable`.

chrisdembia: Maybe `index` should be named `rowIndex`. This is something that is generally confusing to me across the interface (when is the index a rowIndex or a colIndex?). I know that sometimes it only makes sense to be one or the other, but it takes some additional mental effort to realize that.

OpenSim/Common/DataTable.h
chrisdembia: I'm fine with `getColumnLabels()`, etc., Using a name like `getDependentColumnLabels()` would then make it sound like there should be `getIndependentColumnLabel()`. Also, `DependentColumnLabel` is very long; if we are renaming methods, I'd want to get rid of `Dependent`. I think it's sufficient just to document `getColumnLabels()` saying that it does *not* include the independent column.

OpenSim/Common/DataTable.h
chrisdembia: +1

klshrinidhi:
* I have grouped the functions in Doxygen.
* Renamed the member function.
* Resolved merge conflict.

klshrinidhi:
All :white_check_mark: !!!

chrisdembia:
Looks good to me; but @aseth1 may have more to say?

aseth1:
Yay! I would just add the comment about the column labels being just those of dependent columns in the comments for the column methods. Primarily because the current `Storage.getColumnLabels()` includes "time" which makes labels offset by 1 from the data (which has been so annoying!) and now they won't be!

klshrinidhi:
@aseth1 
The group description for column labels mentions this explicitly. I didn't want to write the same comment for 3 or 4 functions related to column labels.

klshrinidhi:
Sorry. I thought I added the description to mention this for column labels but I had done that for other groups.
I just push the change. 
Clear to merge ?

aseth1:
Yes!

Issue #682, opened 2015-10-29
Faster Building


mitkof6:
It is minor issue, but I find it productive when the library builds fast. I was curious, why OpenSim library takes a lot of time to build in contrast to Simbody, which builds quite fast. I had a quick search how to improve the building time and found the following guidelines:

A discussion on [Stackoverflow](http://stackoverflow.com/questions/373142/what-techniques-can-be-used-to-speed-up-c-compilation-times).

Also there is the [Unity Builds](http://buffered.io/posts/the-magic-of-unity-builds/), but I am not sure if it is a good choice.

In summary:
1. Pimpl Idiom
2. Forward Declarations
3. Unity Build

chrisdembia:
Thanks for the suggestions. I think the main reason OpenSim takes longer to compile than Simbody is because it is larger. I agree it's a real pain to compile OpenSim, and decreasing the build time will make the community better.

- Forward declarations is definitely something we could improve. If you see includes that can be turned into forward declarations, please submit a PR!
- I've tried unity builds (seems like a hack) and they didn't work out of the box so I gave up (they are not guaranteed to work for all projects). [cotire](https://github.com/sakra/cotire) helps here.
- I've also tried precompiled headers and the speed up was not significant (used [cotire](https://github.com/sakra/cotire)).
- PIMPL only speeds up incremental builds; it probably increases the build time of building the whole project.

Is your problem with building the whole project the first time or with incremental builds?

For now though, try using clang. Clang compiles our code up to 2 times faster than gcc or MSVC's compiler. You can also try using an LLVM frontend compiler with MSVC; I'm not sure if that will cause any speed up though.

chrisdembia:
If using MSVC, you might be able to get faster build times by compiling the projects in parallel (instead of simply the objects within each project in parallel). I know how to do this on the command line only. From the build directory:

```
cmake --build . --target install --config RelWithDebInfo -- /maxcpucount:8
```

chrisdembia:
I just found this google project that might help us find unnecessary #include's or #include's that could be converted to forward declarations: https://github.com/include-what-you-use/include-what-you-use.

It doesn't seem too easy to use, so I haven't tried it out yet. @mitkof6 if you get curious and you try `include-what-you-use`, let me know.

mitkof6:
> I just found this google project that might help us find unnecessary #include's or #include's that could be converted to forward declarations: https://github.com/include-what-you-use/include-what-you-use; http://include-what-you-use.org/

[...]

PR #923, opened 2016-04-13
Building OpenSim in less than a minute with unity build

mitkof6:
I tested building OpenSim using unity build. The libraries folder builds within 30s on my computer. There is a cmake function that can generate a single compilation unit containing all cpp files per library. I have added this option in the main cmake file so you can choose to build with unity build or without. Also, I have changed OpenSimAddLibrary function. Because all cpp files are included into a single cpp file, there were redefinition and ambiguous issues, which I tried to correct. It is possible to further speed up the compilation if we add all test*.cpp files into a single compilation unit, but I am not sure if there is a why to bypass multiple redefinition of the main function.


mitkof6:
@chrisdembia Regarding #682 I have uploaded a working version using unity builds. The compile time is very good! For some reason, the checks failed.

chrisdembia:
The checks fail because of the BTK dependency. Did you try building with BTK locally? BTK uses Eigen, and that conflicts with SimTK's Eigen.

chrisdembia:
A build time of 30 seconds is very impressive. Thanks for trying this; I'm surprised it worked! I think we'll have to discuss if supporting unity builds is desirable. One of the tradeoffs is that it means we cannot use `using` declarations as much in cpp files.

As for tests, I don't think there is a need to make a unity build for the tests; it's best to have those all as separate executables so they can be run individually.

mitkof6:
> The checks fail because of the BTK dependency. Did you try building with BTK locally? BTK uses Eigen, and that conflicts with SimTK's Eigen.

Yes I tried building with dependencies and it worked for me. 

> One of the tradeoffs is that it means we cannot use using declarations as much in cpp files.

From the changes that I made, it seems that there are only some classes that are in conflict, so if we rename or use OpenSim:: it will not be a problem.

chrisdembia:
Even if you build the dependencies (that is, BTK, using @klshrinidhi's superbuild), I think you must set the CMake variable `WITH_BTK` to ON to use BTK with OpenSim. Did you do that too?

mitkof6:
I built superbuild, but I didn't build with btk. I will check.

chrisdembia:
@mitkof6 before working on this any further, it might be good to wait until some of us can discuss in our next development meeting if we would like unity builds.

mitkof6:
@chrisdembia it seems that the only "problematic" file is C3DFileAdapter.cpp when building btk + unity build.

> before working on this any further, it might be good to wait until some of us can discuss in our next development meeting if we would like unity builds.

Just some comments:
1) There is also a problem when we change a single .cpp, the whole library will have to build again. To bypass this we can have an extra flag that will build some libraries with unity build, and other without (the current working).

2) It is possible to have a workaround solution when we have building  problems, like the case with C3D. We can have an exclude list which will not add some .cpp files (these will be built normally).

mitkof6:
@chrisdembia I have added support to exclude sources from unity build that may be in conflict when added in a single .cpp. I intended to do it for my projects, so I commuted in case you decide to use unity builds. Tests pass, the speedup is good, but most of the time is spent in building and testing the tests. I would suggest to specify build type Release for travis and appveyor. This could help if you want the checks to finish as fast as possible.

chrisdembia:
Thank you. We use the same optimization level (O2) for both Release and RelWithDebInfo; do you think there would still be a difference in the speed of the tests? @sherm1 found that O3 was only 2% faster than O2, which is why we changed Release to use O2. We also found that with MSVC, the time to compile as 32-bit is *much* shorter than for 64-bit. @sherm1 said he thinks the two compilers are different, so maybe the 32-bit one has been more optimized.

klshrinidhi:
@mitkof6 
We still haven't decided if we want to have unity build facility in OpenSim. We are yet to meet and go over the pros and cons of unity build in terms of usability and maintainability long-term because this is a major change. We are hesitant because we have not seen big C/C++ projects adopt this technique. My advice would be to hold off working on this until we have a decision.

mitkof6:
>do you think there would still be a difference in the speed of the tests?

I tested with travis, but no major improvement. It may depend on the underlying resource that are employed for testing. Proper measurements should be carried on the same machine both for Windows and Linux.  

>My advice would be to hold off working on this until we have a decision. 

Ok.

chrisdembia:
@mitkof6 we discussed this PR recently, and we decided to not merge it. Some of the reasons were that we still would not feel comfortable using unity builds for continuous integration or distribution, and that it requires programmers to account for unnecessary constraints when writing their cpp files. Thanks for submitting the PR. We definitely want to improve build time, and we are definitely interested in other PRs that may improve build time (for example, using http://include-what-you-use.org/).

mitkof6:
> Some of the reasons were that we still would not feel comfortable using unity builds for continuous integration or distribution, and that it requires programmers to account for unnecessary constraints when writing their cpp files.

Sure, a fair reason. If I stumble upon anything useful I will let you know.

PR #925, opened 2016-04-16
Using CMAKE_CXX_STANDARD to specify C++ standard

klshrinidhi:
This PR intends :

* To eliminate some code from top level CMakeLists.txt by replacing manual specification of compiler flags to set the C++ standard with the CMake variable [CMAKE_CXX_STANDARD](https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD).

* Make minimum CMake version required the same across platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -374,28 +374,11 @@ endif()
 
 include(InstallRequiredSystemLibraries)
 
-# C++11
-# -----
-# In revision 8629, we started using std::unique_ptr, which requires
-# C++11 features to be enabled when using GCC or Clang.
-if(${CMAKE_C_COMPILER} MATCHES "cc" OR ${CMAKE_C_COMPILER} MATCHES "clang")
-    # Using C++11 on OSX requires using libc++ instead of libstd++.
-    # libc++ is an implementation of the C++ standard library for OSX.
-    if(APPLE)
-        if(XCODE)
-            set(CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LANGUAGE_STANDARD "c++11")
-            set(CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LIBRARY "libc++")
-        else()
-            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
-            if(${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
-                set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
-            endif()
-        endif()
-    else() # not APPLE
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
-    endif()
-endif()
-
+# Minimum required C++ standard --> C++11
+set(CMAKE_CXX_STANDARD_REQUIRED 11
+    CACHE INTERNAL "Minimum C++ standard support required from compiler.")
+set(CMAKE_CXX_STANDARD 11
+    CACHE INTERNAL "C++ standard for all targets.")
 
 ## On APPLE, use MACOSX_RPATH.
 set(OPENSIM_USE_INSTALL_RPATH FALSE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -375,7 +375,7 @@ endif()
 include(InstallRequiredSystemLibraries)
 
 # Minimum required C++ standard --> C++11.
-set(CMAKE_CXX_STANDARD_REQUIRED 11
+set(CMAKE_CXX_STANDARD_REQUIRED ON
     CACHE INTERNAL "Minimum C++ standard support required from compiler.")
 set(CMAKE_CXX_STANDARD 11
     CACHE INTERNAL "C++ standard for all targets.")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -374,7 +374,7 @@ endif()
 
 include(InstallRequiredSystemLibraries)
 
-# Minimum required C++ standard --> C++11
+# Minimum required C++ standard --> C++11.
 set(CMAKE_CXX_STANDARD_REQUIRED 11
     CACHE INTERNAL "Minimum C++ standard support required from compiler.")
 set(CMAKE_CXX_STANDARD 11
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,14 +7,7 @@ project(OpenSim)
 #
 #
 #----------------------------------------------------
-
-if(MSVC)
-    # To properly support Visual Studio 2015.
-    set(OPENSIM_REQUIRED_CMAKE_VERSION 3.1.3)
-else()
-    set(OPENSIM_REQUIRED_CMAKE_VERSION 2.8.8)
-endif()
-cmake_minimum_required(VERSION ${OPENSIM_REQUIRED_CMAKE_VERSION})
+cmake_minimum_required(VERSION 3.1.3)
 
 if(COMMAND cmake_policy)
     cmake_policy(SET CMP0003 NEW)
--- a/dependencies/CMakeLists.txt
+++ b/dependencies/CMakeLists.txt
@@ -8,9 +8,8 @@ cmake_minimum_required(VERSION 3.1.3)
 include(ExternalProject)
 include(CMakeParseArguments)
 
-# Minimum required C++ standard --> C++11. OpenSim dependencies have to be
-# compiled with same standard requirement as OpenSim to have dependencies ABI 
-# compatible.
+# Minimum required C++ standard --> C++11. This is to make sure dependencies
+# are compiled with same minimum standard as OpenSim.
 set(CMAKE_CXX_STANDARD_REQUIRED 11
     CACHE INTERNAL "Minimum C++ standard support required from compiler.")
 set(CMAKE_CXX_STANDARD 11
--- a/dependencies/CMakeLists.txt
+++ b/dependencies/CMakeLists.txt
@@ -136,7 +136,8 @@ AddDependency(NAME       BTK
               # URL        https://github.com/Biomechanical-ToolKit/BTKCore.git
               URL        https://github.com/klshrinidhi/BTKCore.git
               TAG        ae0d2a43de0fe6c9ef2acce34a87c944dd113829
-              CMAKE_ARGS -DBUILD_SHARED_LIBS:BOOL=ON)
+              CMAKE_ARGS -DBUILD_SHARED_LIBS:BOOL=ON 
+                         -DCMAKE_CXX_FLAGS:STRING="/EHsc")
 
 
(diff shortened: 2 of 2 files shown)


klshrinidhi:
Travis CI builds will fail until #926 is merged.

chrisdembia:
The CMAKE_CXX_STANDARD variables are very nice.

This would make our minimum required Ubuntu version to be 15.10 (http://packages.ubuntu.com/search?keywords=cmake&searchon=names&suite=all&section=all; wily). We should discuss if we are okay with this in the dev meeting. The sherlock cluster that Carmichael uses has CMake 3.2.2.

Other feedback (perhaps wait until after the dev meeting to start acting on the feedback):

1. Are you using `CMAKE_CXX_STANDARD_REQUIRED` correctly? It is a boolean that says whether `CXX_STANDARD` must be defined: https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD_REQUIRED.html#prop_tgt:CXX_STANDARD_REQUIRED
4. What is the reason to make the variables `CACHE INTERNAL`?
1. Did you test this with Xcode? Xcode is particular about how to set the C++ standard; I'm not sure if CMake's new mechanism supports Xcode properly. Also, our OSX travis build does not use Xcode to build (though maybe it should, instead of unix makefiles).
2. That same old ugly CXX standard code is strewn all over the examples and in the SampleCMakeLists.txt; we should update it everywhere. Perhaps grep for all occurrences of `c++11`, etc.?
3. I think the ability to use different versions of CMake for different platforms/compilers is fine. Even if right now we use the same version of CMake on all platforms, that might not always be the case. MSVC usually lags behind GCC/Clang in terms of feature support, and supporting a new version of MSVC usually requires a newer version of CMake, but we may still want to support an older version of Ubuntu that doesn't have that newer CMake. Also, it's more reasonable to expect Windows users to have the latest version of CMake. I would prefer that we leave in the if-else, even if for now we use the same version in both cases.

Not for this PR: Early in the switch to C++11, some users ran into issues when they were trying to build their own project simbody-based that did not use C++11 when compiling simbody with C++11 (or something like that). Is it possible to force people who `find_package(OpenSim)` to also use C++11 in their projects?

Related: https://github.com/simbody/simbody/issues/325.

klshrinidhi:
>This would make our minimum required Ubuntu version to be 15.10 (http://packages.ubuntu.com/search?keywords=cmake&searchon=names&suite=all&section=all; wily). We should discuss if we are okay with this in the dev meeting. The sherlock cluster that Carmichael uses has CMake 3.2.2.

I don't agree. CMake does not come installed on any operating system. One has to install it anyway. It is just a matter of where you get it from. For Ubuntu 15.10 and later, Ubuntu repositories have CMake 3.x. For Ubuntu 14.04 and earlier, we have to add a repository and get it from there. Anyone who can get the required versions of the dependencies and build tools will be able to build OpenSim from source even on Ubuntu 12.04, which is what Travis runs.
Also our installation instructions already are to install CMake 3.x in all platforms.
OpenSim Ubuntu user/developer base is very small. So this is actually applicable to very small number of people. In general, I think Ubuntu users are used to getting tools from various repositories so I don't think this is making anything hard for anyone because our instructions are already doing this.

> Are you using `CMAKE_CXX_STANDARD_REQUIRED` correctly? It is a boolean that says whether `CXX_STANDARD` must be defined: https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD_REQUIRED.html#prop_tgt:CXX_STANDARD_REQUIRED

I think I am doing it right. I followed documentation [here for CMake 3.1.3](https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD). The variable `CMAKE_CXX_STANDARD_REQUIRED` is for us to tell CMake to mandate the `CXX_STANDARD` requirement. Without this, the CMake will use a previous standard if the compiler does not support the mentioned `CXX_STANDARD`.

> What is the reason to make the variables `CACHE INTERNAL`?

There are two things here -- 
* `CACHE` -- I can make *not a cache* variable if we are sure it will be propagated to entire project that way. I also wanted to keep a record of it in the cache in case we need to expose this in the config file. This is related to your last comment.
* `INTERNAL` -- Given that I made it a cache variable, I set it to `INTERNAL` because it is not an option that we want users to have.

>  Did you test this with Xcode? Xcode is particular about how to set the C++ standard; I'm not sure if CMake's new mechanism supports Xcode properly. Also, our OSX travis build does not use Xcode to build (though maybe it should, instead of unix makefiles).

I have not. I will get back on this.

>  That same old ugly CXX standard code is strewn all over the examples and in the SampleCMakeLists.txt; we should update it everywhere. Perhaps grep for all occurrences of `c++11`, etc.?

Good point. I can do this in this PR or in another one. No preference.

> I think the ability to use different versions of CMake for different platforms/compilers is fine. Even if right now we use the same version of CMake on all platforms, that might not always be the case. MSVC usually lags behind GCC/Clang in terms of feature support, and supporting a new version of MSVC usually requires a newer version of CMake, but we may still want to support an older version of Ubuntu that doesn't have that newer CMake. Also, it's more reasonable to expect Windows users to have the latest version of CMake. I would prefer that we leave in the if-else, even if for now we use the same version in both cases.

I think supporting different versions of CMake on different platforms will make our CMakeLists.txt harder to develop, maintain and read. For Linux/Mac, we use features of the older CMake but only for Windows, we write portions of CMakeLists.txt to use newer CMake.
I think that if it is possible to get the required build tools on all platforms, we should bump up the version to be the same for all platforms.

> Not for this PR: Early in the switch to C++11, some users ran into issues when they were trying to build their own project simbody-based that did not use C++11 when compiling simbody with C++11 (or something like that). Is it possible to force people who `find_package(OpenSim)` to also use C++11 in their projects?
> Related: https://github.com/simbody/simbody/issues/325.

Superbuild procedure will make sure to build dependencies with C++11. This change is included in this PR.

chrisdembia:
First, let me say that I think it may be fine if 15.10 is the minimum version we support (It's a little too recent for my comfort in general, but I think it's fine given the circumstances of C++11). However, I think we should all agree on this in a dev meeting.

> I don't agree. CMake does not come installed on any operating system. One has to install it anyway. It is just a matter of where you get it from. For Ubuntu 15.10 and later, Ubuntu repositories have CMake 3.x. For Ubuntu 14.04 and earlier, we have to add a repository and get it from there. Anyone who can get the required versions of the dependencies and build tools will be able to build OpenSim from source even on Ubuntu 12.04, which is what Travis runs.
> Also our installation instructions already are to install CMake 3.x in all platforms.
> OpenSim Ubuntu user/developer base is very small. So this is actually applicable to very small number of people. In general, I think Ubuntu users are used to getting tools from various repositories so I don't think this is making anything hard for anyone because our instructions are already doing this.

We had a similar discussion around gcc. I don't want the two of us to be the only ones having this discussion. We should decide as a team what it means to us to support (a given version of) Ubuntu.  Maybe we can do this in the next dev meeting. Here's my opinion: supporting a given version of Ubuntu is different than supporting a given version of Windows or OSX, because the assumptions/expectations are different. On Windows or OSX, it's reasonable to expect people to go to websites and get dependencies themselves (on OSX, this expectation is shifting I think; now the expectation is that you can get your dependencies through homebrew). On Ubuntu, the assumption is that you can get your dependencies through `apt` without any weird third party PPA's. This is more true for the typical OpenSim user than it is for a seasoned developer like you, @klshrinidhi. Of course; one can use older versions of Ubuntu if they can get the dependencies themselves. A separate issue is systems where users do not have sudo privileges (clusters). I think the general principle (for open, collaborative projects) is to support the minimum versions of dependencies as necessary.

Although our Ubuntu user base is small, we *want* it to be bigger.

> There are two things here --

Thanks for explaining. I'm pretty sure that variables do not need to be in the cache to be used in the config file.

> Good point. I can do this in this PR or in another one. No preference.

I have no preference either.

> but only for Windows, we write portions of CMakeLists.txt to use newer CMake.

I'm pretty sure we don't have logic in the CMakeLists like this; the only difference we should take advantage of on Windows should be support of the necessary MSVC versions.

> Superbuild procedure will make sure to build dependencies with C++11.

Yes (and I see you've done that), but I mean when OpenSim is a dependency of someone else's project. Maybe I'm misunderstanding something?

klshrinidhi:
I am happy to close this PR. I don't want to spend too much time on this if it this controversial.

> First, let me say that I think it may be fine if 15.10 is the minimum version we support ...

This is actually what I disagree on. I don't understand this minimum requirement issue. This PR doesn't make 15.10 the minimum requirement. CMake does not come installed in 15.10. One has to get it anyway. For 14.04 and below, one has to add another repository (our instructions currently do this already) and for 15.10 and above, Ubuntu repositories have it. But CMake still needs to be installed.

> We had a similar discussion around gcc. I don't want the two of us to be the only ones having this discussion. We should decide as a team what it means to us to support (a given version of) Ubuntu.  Maybe we can do this in the next dev meeting. 

I stated this PR for our team to discuss.

> Here's my opinion: supporting a given version of Ubuntu is different than supporting a given version of Windows or OSX, because the assumptions/expectations are different. On Windows or OSX, it's reasonable to expect people to go to websites and get dependencies themselves (on OSX, this expectation is shifting I think; now the expectation is that you can get your dependencies through homebrew). On Ubuntu, the assumption is that you can get your dependencies through `apt` without any weird third party PPA's. This is more true for the typical OpenSim user than it is for a seasoned developer like you, @klshrinidhi. Of course; one can use older versions of Ubuntu if they can get the dependencies themselves. A separate issue is systems where users do not have sudo privileges (clusters). I think the general principle (for open, collaborative projects) is to support the minimum versions of dependencies as necessary.

Third party PPA's are not weird in anyway. Launchpad officially wants people to contribute by uploading build-able source code. It will build the source code and make the binaries available for people to install. In Ubuntu world, it is natural for people to get binaries from various repositories. There is GUI support in Ubuntu to add repositories. 
I don't think we need to solve the issue of people not having admin access to their clusters. They can request their administrators to do this. Admins are there to provide such support anyway. Because multiple versions of CMake can exist on Ubuntu, admins can make this work. There is no technical challenge in getting new version of CMake into a Linux server.

> Thanks for explaining. I'm pretty sure that variables do not need to be in the cache to be used in the config file.

I was thinking about exposing the variable through config file so external projects can detect what cxx standard OpenSim was built on. If it can be done without caching the variable, I will change it.

> I'm pretty sure we don't have logic in the CMakeLists like this; the only difference we should take advantage of on Windows should be support of the necessary MSVC versions.

That means we have to make sure no PR gets through that takes advantage of newer CMake on Windows. 

> Yes (and I see you've done that), but I mean when OpenSim is a dependency of someone else's project. Maybe I'm misunderstanding something?

My intention to make those variables `CACHE INTERNAL` was to have them exposed in config file later so external projects can detect it.

chrisdembia:
I read through all your comments, but for now I just want to say that I think exposing the CXX_STANDARD variable in the config file is a great idea.

PR #927, opened 2016-04-18
Wrap Reporters and TableSource; move Component wrapper back to Common

chrisdembia:
This is an attempt to wrap the new reporter classes (TableReporter, ConsoleReporter, StatesTrajectoryReporter). I ran into a problem with the fact that `Component.h` is now wrapped in `simulation.i` instead of `common.i` (the Reporter.h is in Common): the Reporter classes do not have the methods they inherit from Component, Object, etc. I understand why the header was moved, but now that we have subclasses of Component in Common, it may be good to find a way to move Component.h back to `common.i`. 

@klshrinidhi, @aymanhab, any ideas? It is as if we need to "forward declare" Frame, etc. Perhaps the `getFrameList()`, etc. methods can be on Model instead of Component?


aymanhab:
@chrisdembia I think longer term we're trying to avoid treating Model as a special Component or ModelComponent so I'd prefer not to move these methods to Model. Doesn't really matter if on Component or ModelComponent to me since the distinction between them is unclear to me. Would it be possible to move Component (as a Modeling abstraction) to the Simulation library and then we don't have this problem altogether (since the separation into the libraries is not necessarily reflective of current abstractions/arrangement). Whjat do you think @aseth1 ?

aseth1:
`get####List()` methods on Component seems wrong to me. Component should only know about other Components in a generic way, e.g. templated `getComponentList()`.  The access to specific lists of components from the wrapping point of view is to replace the access to Sets of the Model. Keep `Component` and `Reporter` in common seems simplest to me and remove the type specific lists from Component please.

aymanhab:
@aseth1 It would be fine if the methods move to ModelComponent, I'm trying to avoid special methods on Model because that's what we advocated a while ago. Would that work @chrisdembia ? Would the ability to iterate thru and filter "subcomponets" be needed for Component or are ModelComponents the abstractions that we want users to deal with?

aseth1:
I m not sure that `ModelComponent` will need to exist for much longer. I think the distinction is perhaps to thin to justify. I also think it is fundamentally wrong for a base class to have specific awareness of its derived classes.

chrisdembia:
To summarize an in-person discussion, we decided that putting these instantiations on ModelComponent does not make sense, since that class may go away. We have decided to put the instantiations on Model for now, and move Component to the Common module in wrapping. The hope is that we find some other mechanism to provide generic downcasted access to ComponentLists in wrapping. The solution for python is straightforward; it's not as clear yet how we will achieve the same thing for java/matlab.

A side effect is that ModelDisplayHints must be moved to Common. This should be done anyway, since Component uses ModelDisplayHints. I have opened a new issue to consider renaming this class; #1029.

I have pushed commits that reflect what we agreed upon.

chrisdembia:
This PR is now ready for review (no longer WIP). It now covers more than just wrapping reporters:

- Fix memory ownership in bindings for `Component::addComponent`
- Move Component back to Common module for bindings
  - Move ModelDisplayHints to Common
  - Put getBodyList() etc on Model instead of Component
- Wrap Reporter classes
- Wrap TableSource classes.

chrisdembia:
Builds now pass. Please review when you get the chance.

Bindings/Python/tests/test_TableSourceReporter.py
```diff
@@ -0,0 +1,51 @@
...
+        c_rep.updInput("inputs").connect(source.getOutput("column").getChannel("col1"))
+        t_rep.updInput("inputs").connect(source.getOutput("column").getChannel("col2"))
+
+        # Realize.
+        s = m.initSystem()
+        s.setTime(0.5)
+        m.realizeReport(s)
+
+        # Test.
+        assert t_rep.getReport().getRowAtIndex(0)[0] == 2.5
```
aymanhab: Some comment about why this should be 2.5 would help (linear interpolation col 2 I assume)

aymanhab:
Looks good to me. With scripting tests passing feel free to merge.

chrisdembia:
Tests pass and I added the comment @aymanhab made. Based on @aymanhab's permission, I am now merging this.

PR #933, opened 2016-04-18
Add 'tension' output to pathactuator

klshrinidhi:
* Remove the function `updateFromXMLNode` as it is redundant.
* Add output named *tension* that retrieves the scalar(double) actuation from `computeActuation`.

Let me know if you need more changes.

### DIFF
--- a/OpenSim/Simulation/Model/PathActuator.cpp
+++ b/OpenSim/Simulation/Model/PathActuator.cpp
@@ -163,9 +163,6 @@ void PathActuator::addNewPathPoint(
  */
 double PathActuator::computeActuation( const SimTK::State& s ) const
 {
-    if(!_model)
-        return 0.0;
-
     // FORCE
     return( getControl(s) * get_optimal_force() );
 }
--- a/OpenSim/Simulation/Model/PathActuator.cpp
+++ b/OpenSim/Simulation/Model/PathActuator.cpp
@@ -262,6 +262,26 @@ SimTK::Vec3 PathActuator::computePathColor(const SimTK::State& state) const {
 
 
 //=============================================================================
+// XML
+//=============================================================================
+//-----------------------------------------------------------------------------
+// UPDATE FROM XML NODE
+//-----------------------------------------------------------------------------
+//_____________________________________________________________________________
+/**
+ * Update this object based on its XML node.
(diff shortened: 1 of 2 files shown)


chrisdembia:
Looks good to me.

aymanhab:
I agree the function updateFromXML looks redundant but it would be good to make sure we have a test case that (de)serializes an object of this type to make sure nothing is broken inadvertently.

aseth1:
@aymanhab `PathActuator` is the base for all Muscles and there are several test cases (testForward,testCMC, ...) that exercise deserialization of the model (including muscles). Do you think we need something beyond these?

aymanhab:
@aseth1 Nope that should be good enough for deserialization. 

Related issue, I'm not sure how we're going to test that "Owner" setting is still intact since this functionality is primarily used by the GUI/Muscle editor to go between the PathActuator and GeometryPath but maybe a different mechanism is available now.

aseth1:
> Related issue, I'm not sure how we're going to test that "Owner" setting is still intact since this functionality is primarily used by the GUI/Muscle editor to go between the PathActuator and GeometryPath but maybe a different mechanism is available now.

Good point. We should remove the actuator level ownership since it is redundant now with the Component's parent (which is its owner).

chrisdembia:
IMO in general it is better to keep PR's to one issue; a separate PR for the removal of updateFromXMLNode would have allowed this PR to be merged quickly. And, git allows for such a process to *not* affect the development workflow, since you need not commit all changes.

aseth1:
@chrisdembia. I agree, but in this case the PR was pretty small even with the updateFromXML. I don't see any harm done since we had to wait for the tests to pass anyways  and had time to discuss. The "owner" issue will be another PR.

If there are no objections we can merge now.

klshrinidhi:
Alright @chrisdembia, I have reverted the changes made with respect to removing the function `updateFromXMLNode`. I am going to start another PR for that. :+1: 

Can we merge this now ? :stuck_out_tongue:

klshrinidhi:
@aseth1 convinced me to revert the revert. Will split PRs properly from next time.
I am going to merge this now.

chrisdembia:
👍

PR #934, opened 2016-04-20
Test travis release speedup

mitkof6:
I wanted to check whether Release will be faster than RelWithDebInfo when the tests are performed.


mitkof6:
No major improvement, it may depend on the underlying resource that are employed for testing. Proper measurements should be carried on the same machine to test.

chrisdembia:
Good idea; thanks for doing this.

PR #935, opened 2016-04-20
Tablesource component

klshrinidhi:
Addressing issue #889.


OpenSim/Common/TableSource.h
```diff
@@ -0,0 +1,214 @@
...
+
+        addMessage(msg);
+    }
+}; // class TimeOutOfRange
+
+
+/** Component representing a source of data from a TimeSeriesTable_.
+
+This Component has two outputs:
+- A list output wih one channel per column of the TimeSeriesTable_.
```
chrisdembia: wih -> with

OpenSim/Common/TableSource.h
```diff
@@ -0,0 +1,214 @@
...
+public:
+    /** Type of the TimeSeriesTable_ this Component will hold.                */
+    typedef TimeSeriesTable_<ET> Table;
+    /** Type of the 'row' Output of this Component.                           */
+    typedef SimTK::Vector_<ET>   Vector;
+
+    OpenSim_DECLARE_OUTPUT(row, Vector, getRowAtTime, 
+                           SimTK::Stage::Time);
+    OpenSim_DECLARE_LIST_OUTPUT(column, ET, getColumnAtTime, 
+                                SimTK::Stage::Time);
```
chrisdembia: It'd be good to discuss in a dev meeting what these outputs should be called. Perhaps "column" and "all_columns"., or "indiv" and "all". "row" and "column" might be confusing.

klshrinidhi:
Fixed typo.
Can we merge ?

OpenSim/Common/Test/testTableSource.cpp
```diff
@@ -0,0 +1,216 @@
...
+    mutable SimTK::ReferencePtr<SimTK::SimbodyMatterSubsystem> matter;
+    mutable SimTK::ReferencePtr<SimTK::GeneralForceSubsystem> forces;
+
+    // keep track of the force added by the component
+    mutable SimTK::ForceIndex fix;
+
+    MemberSubcomponentIndex intSubix{constructSubcomponent<Sub>("internalSub")};
+}; // class TheWorld
+
+
```
aseth1: Although, I like having a separate test, I am not a fan of the duplication of `TheWorld` and its dependencies here. Would it not be more efficient to add additional cases (SUBTEST) for testing the `TableSource` to `testComonentInterface.cpp` which already provides the necessary "environment". My fear is that we have to maintain the same environment in two places.

OpenSim/Common/Test/testTableSource.cpp
```diff
@@ -0,0 +1,216 @@
...
+    // Define the Simbody system
+    MultibodySystem system;
+
+    TheWorld theWorld;
+    theWorld.setName("World");
+
+    theWorld.add(tableSource);
+    theWorld.add(tableReporter);
+
+    tableReporter->updInput("inputs").connect(tableSource->getOutput("column"));
```
aseth1: There seems to be a mismatch here where `getOutput("column")` returns a single column/channel and `updInput("inputs")` seems to be a list of inputs, so it isn't clear that all columns are being wired up to separate channels of the inputs. Otherwise it seems like a single "column" output is satisfying one input.

OpenSim/Common/Test/testTableSource.cpp
chrisdembia: We could put `TheWorld` in a header file.

OpenSim/Common/TableSource.h
aseth1: I agree. The list output also requires the index or name at access time, whereas the Vector of all values does not, but they both potentially can provide all the output values to satisfy a corresponding list input (e.g. [test-case](https://github.com/opensim-org/opensim-core/pull/935/files#diff-d4268f0f931b28653836e567b134a6b8R170) ) in which case a singular "column" seems incorrect. How about `vector` (for the vector of values)  vs `elements` (for many values accessible as individual elements)? I don't have my heart set on these. 

@tkuchida @jimmyDunne  @jenhicks any suggestions here?

OpenSim/Common/TableSource.h
tkuchida: I found these descriptions unhelpful. What is a "list output"? What do you mean by "channel"? I think the descriptions of the outputs should be more user-friendly. Also, I think they should appear in the doxygen for the methods themselves. Here, I would describe (at a high level) what the outputs can return (e.g., "This component has outputs for retrieving all or a specified subset of the columns in a TimeSeriesTable_ at a specified time, where the values are interpolated in time if necessary."). :snowman:

OpenSim/Common/TableSource.h
tkuchida: > "row" and "column" might be confusing.

I was (and still am) confused. Looking in [ComponentOutput.h](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/ComponentOutput.h#L393) helped, but I haven't been able to see where the user specifies the column labels when using the "column" output. Is `setTable()` :fork_and_knife: :fork_and_knife: :fork_and_knife: saying that all columns are reported by default (perhaps a comment is needed at L101)? Where in testTableSource are columns being selected? Sorry if I'm missing something obvious. :snowflake:

As for naming, perhaps "all_columns" and "named_columns"?

OpenSim/Common/TableSource.h
```diff
@@ -0,0 +1,214 @@
...
+                                                     colInd);
+            auto elt = ((time - prevTime) / (nextTime - prevTime)) * 
+                       (nextElt - prevElt) + prevElt;
+            return elt;
+        }
+    }
+
+    /** Retrieve a row of the TimeSeriesTable_ at a given time (specified by the
+    state). Linear interpolation is performed if the TimeSeriesTable_ does not
+    have an entry for the time mentioned by the state.
```
tkuchida: Is time included in the Vector? Perhaps include a `\returns` comment explicitly stating that a Vector is returned whose length is equal to… :snowboarder:

OpenSim/Common/TableSource.h
```diff
@@ -0,0 +1,214 @@
...
+        for(const auto& columnLabel : _table.getColumnLabels())
+            columnOutput.addChannel(columnLabel);
+    }
+
+    /// @}
+
+protected:
+    /** Retrieve value of a column at a given time(implicit in the State 
+    provided). Linear interpolation is performed if the TimeSeriesTable_ does
+    not contain an entry for the time mentioned by the state.
```
tkuchida: I'm confused. It seems from this comment that `getColumnAtTime()` returns the value from exactly one column (whose label is specified as the second argument) at exactly one time. But then

``` cpp
OpenSim_DECLARE_LIST_OUTPUT(column, ET, getColumnAtTime, SimTK::Stage::Time);
```

calls this method and assembles a list somehow. It would be helpful if the doxygen for this method provided some explanation. :icecream:

chrisdembia:
@tkuchida we discussed this in the dev meeting and started to settle on "all_columns" and "column." What was your rationale for the longer alternatives to "column"? My idea is that the fact that it is a list output conveys that "specified" nature of the output.

OpenSim/Common/TableSource.h
chrisdembia: I think your confusion stems from my poor documentation of list outputs, not of anything related to TableSource.

tkuchida:
> What was your rationale for the longer alternatives to "column"?

If I understand correctly, the list will be of length 1 or greater; I don't think "column" conveys this information. Both outputs will return columns, so it seems like both should have qualifiers: "all" vs. "only-those-that-I've-specified" (as opposed to the user asking herself, "Do I want _all columns_ or do I want _column_?").

Ideally, I would prefer a single output called "columns" where the user can either specify which subset of columns to return or specify nothing and all columns are returned (I think this would be a minimal and complete interface). :turtle:

OpenSim/Common/TableSource.h
tkuchida: I think the singular "column" was confusing me as well: the list could contain values from several columns (contradicts singular "column") or several time points (contradicts singular "getColumnAtTime"), or perhaps a single number is always returned but it's returned in a list container for some other reason (e.g., API compatibility). A code snippet demonstrating how to select the reported columns would avoid much of this confusion. :dog:

chrisdembia:
> Ideally, I would prefer a single output called "columns" where the user can either specify which subset of columns to return or specify nothing and all columns are returned (I think this would be a minimal and complete interface). :turtle:

This is indeed the interface, but I think we are focusing on the wrong side of the interface. If I have a TableSource and I want to connect its outputs to a reporter, I can do the following:

```cpp
// multiple Channels, each of which contains one column (annotated by column label).
scalarReporter.getInput("inputs").connect(tablesrc.getOutput("column"));
 // the knee_angle column
scalarReporter.getInput("inputs").connect(tablesrc.getOutput("column").getChannel("knee_angle"));
scalarReporter.getInput("inputs").connect(tablesrc.getOutput("column").getChannel("angle_angle"));
// a single Output that is of type Vector, containing each column. Not annotated by column label.
vectorReporter.getInput("inputs").connect(tablesrc.getOutput("all_columns"))

One can get all columns using the "column" list output. The "all_columns" non-list output is different: it's when you want all the columns as a single Vector (you can't separate them).
```

chrisdembia:
I'm realizing that that is kinda confusing, though. I suppose we need to document that clearly.

chrisdembia:
Of all those method calls I showed above, I think the most confusing is the first; since it actually wires up all columns to the reporter. But then again, I don't think "specified_columns" would make it clearer. I think maybe something like "indiv" could be clearer. Any other suggestions?

tkuchida:
> Of all those method calls I showed above, I think the most confusing is the first

Thanks; that code was super helpful! Not sure whether this is possible, but something like
``` cpp
scalarReporter.getInput("inputs").connect(tablesrc.getOutput("column").getChannel("*"));
```
might be clearer. Otherwise, perhaps the output names could be more explicit and analogous to "scalarReporter" and "vectorReporter" (e.g., "columnScalars" and "columnVector")? :dromedary_camel:

chrisdembia:
> Thanks; that code was super helpful! Not sure whether this is possible, but something like
> 
> scalarReporter.getInput("inputs").connect(tablesrc.getOutput("column").getChannel("*"));
> might be clearer.

I think this is a really good idea, and I hope we can implement something like this down the line.

Based on your suggestions, I realize I made a mistake: the "column" output could actually be Vec3's (not scalars) for a TableSource_<Vec3>.

klshrinidhi:
* Changed name of the output to "all_columns". 
* Moved test for `TableSource` into `testComponentInterface`.

klshrinidhi:
Does this need anything more to get merged ?

aseth1:
@klshrinidhi I am happy to merge now and have us get some use of this very useful class.  I do think their are outstanding issues with naming of list inputs based on the comments of @tkuchida. I believe they  are beyond the scope of this PR but will likely require TableSource to be revisited.

PR #938, opened 2016-04-21
Update prescribed force to component API

aymanhab:
Added Connector and outputs, included as Component in testComponents


OpenSim/Tests/Components/testComponents.cpp
```diff
@@ -74,6 +74,8 @@ int main()
         availableComponents.push_back(availableBushingForces[i]);
     }
 
+    // Test PrescribedForce
+    availableComponents.push_back(new PrescribedForce());
```
chrisdembia: I think this is a memory leak, since `availableComponents` doesn't own its objects (https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Object.h#L460). This might be a fix:

```
std::unique_ptr<PrescribedForce> f(new PrescribedForce());
availableComponents.push_back(f.get())
```

aymanhab:
@aseth1 Any feedback on this, let me know so we can move on. Thanks.

OpenSim/Common/XMLDocument.cpp
```diff
@@ -54,7 +54,8 @@ using namespace std;
 // 30502 for Changing serialization of Geometry
 // 30503 for Changing serialization of Ground
 // 30505 for Changing serialization of Joint to create offset frames
-const int XMLDocument::LatestVersion = 30505;   
+// 30506 for Changing serialization of PrescribedForce to use Connectors
+const int XMLDocument::LatestVersion = 30506;
```
aseth1: Is it necessary to increment the version number for every change vs. waiting until we are done for 4.0 and increment to a new version number then?

OpenSim/Simulation/Model/PrescribedForce.cpp
```diff
@@ -42,14 +42,13 @@ using namespace std;
 /**
  * Default constructor.
  */
-PrescribedForce::PrescribedForce(Body* body)
+PrescribedForce::PrescribedForce(const PhysicalFrame* body)
 {
```
aseth1: minor, but might as well change argument `body`->`frame`.

OpenSim/Simulation/Model/PrescribedForce.cpp
```diff
@@ -235,13 +256,15 @@ void PrescribedForce::computeForce(const SimTK::State& state,
...
+        getConnector<PhysicalFrame>("frame").getConnectee();
+
     if (hasForceFunctions) {
         Vec3 force(forceFunctions[0].calcValue(timeAsVector), 
                    forceFunctions[1].calcValue(timeAsVector), 
                    forceFunctions[2].calcValue(timeAsVector));
         if (!forceIsGlobal)
-            engine.transform(state, *_body,                 force, 
+            engine.transform(state, frame,                 force,
                                     getModel().getGround(), force);
```
aseth1: Could use the `Frame` itself to re-express the force in Ground. It would be much more readable and would start to remove our dependencies on the `SimbodyEngine`. Same is true for the transformations below that depend on the `engine`.

OpenSim/Simulation/Model/PrescribedForce.h
```diff
@@ -94,11 +89,19 @@ OpenSim_DECLARE_CONCRETE_OBJECT(PrescribedForce, Force);
     /** These are three functions providing the x,y,z measure numbers of the
     torque vector being applied to the body. The coordinate frame in which
     this vector is interpreted depends on the "torqueIsGlobal" property. **/
+
     // Would have been better for this to be a list property. 
     OpenSim_DECLARE_PROPERTY(torqueFunctions, FunctionSet,
```
aseth1: I agree with this comment. If you make it a sized list property we can enforce exactly 3 components.

chrisdembia: I agree too, but that's unrelated to this PR.

OpenSim/Simulation/Model/PrescribedForce.h
```diff
@@ -94,11 +89,19 @@ OpenSim_DECLARE_CONCRETE_OBJECT(PrescribedForce, Force);
...
         "Three functions describing the torque the PrescribedForce applies.");
+    /** The force applied by the PrescribedForce, this depends only on time */
+    OpenSim_DECLARE_OUTPUT(force_applied, SimTK::Vec3, getForceApplied, SimTK::Stage::Time);
```
aseth1: Would be good to know the reference frame in which the force is expressed in and indicating the effect of `forceIsGlobal`. Similarly for the torque and point.

OpenSim/Common/XMLDocument.cpp
chrisdembia: I had this question too... @klshrinidhi and I wanted to make some changes to the code in the 30505 and earlier versions, mainly to fix bugs, but it didn't seem like it would be worthwhile to add a new file version to fix a bug in an unreleased file version.

aymanhab: Here's the issue with not upgrading version number. If you save the file with the old version number, then the contents will be inconsistent with the version number in the header and there will be no way to tell that the file has been partially updated. the updatefromXML code will likely get triggered again which may try to recreate connections or duplicate xml structure.

Usually the code in updateFromXML is guarded by a check that the version is less than latest to avoid the conversion altogether on files conforming to the latest format. So if we don't bump we'll be putting  off many bugs related to conversion issues downstream until we bump the version number but I don't have a concrete example to back this up.

aseth1: I am not sure I follow. Is the issue that if I already have a 30505 update and I now want to have an additional update then their is ambiguity? I would sort of agree, but then the argument to increment should be that there is an existing update for the current version and you want to keep the next update separate.

In this case `PrescribedForce` had no versions to update at all, so why not include the update in 30505? I have not been incrementing the version numbers for updates I have been making and treating 30505 as the moving format as we make changes. So I have not been consistent with the a rule to increment the version for every update change. I think that would be a good rule to adopt going forward but maybe not during the overhaul itself.

Thanks for discussing in person @aymanhab and @chrisdembia. For now, I think we agreed to increment when updates to be applied to a given class already have an update for the latest version, then it makes sense to keep new changes separate and bump up the version number. 

We also agreed we should bump the version number up prior to having a version we are willing to support (e.g. for the Workshop).

aymanhab:
@aseth1 I undid the version change per our discussion as we assume current XML document version is not supported until we're done with this refactoring. I made the other doc changes and variable name changes. I didn't do the conversion to use Frame since the simbodyengine methods do special handling of inplace conversion that I don't know if tested with the generic Frame methods. If we decide to remove simbodyengine this is probably one of hundreds of places to fix/replace. Please let me know what you think.

OpenSim/Simulation/Model/PrescribedForce.cpp
```diff
@@ -74,6 +73,15 @@ PrescribedForce::PrescribedForce(SimTK::Xml::Element& aNode) : Super(aNode)
 void PrescribedForce::updateFromXMLNode(SimTK::Xml::Element& aNode, int versionNumber)
 {
     // Base class
+
```
aseth1: You still have to check that the version isn't 30505 (the current influx version, since you don't want to invoke this code block if deserializing files that have been converted during testing etc... even if it isn't a supported version.

OpenSim/Simulation/Model/PrescribedForce.cpp
```diff
@@ -42,14 +42,13 @@ using namespace std;
 /**
  * Default constructor.
  */
-PrescribedForce::PrescribedForce(Body* body)
+PrescribedForce::PrescribedForce(const PhysicalFrame* frame)
 {
```
aseth1: I've been switching passed in dependencies as const references so that there is not confusion about the the Component taking ownership. Although not backward compatible it would it more consistent with other Components (e.g. joints, bushings, etc.. that depend on PhysicalFrames)

chrisdembia: My suggestion: leave this method in, deprecate it using the `DEPRECATED_14` macro in simbody (usage shown here:  https://github.com/simbody/simbody/blob/master/SimTKcommon/include/SimTKcommon/internal/ClonePtr.h#L293), and create a new constructor that takes a const reference.

aseth1:
> I didn't do the conversion to use Frame since the simbodyengine methods do special handling of inplace conversion that I don't know if tested with the generic Frame methods. If we decide to remove simbodyengine this is probably one of hundreds of places to fix/replace.

Not sure what you mean by "inplace conversion", but the advantages of doing the conversion are: 1) we actually use the facilities (Frame calculations) we want users to use in their code; 2) it tests Frame calculations perhaps in ways the unit tests don't, and 3) it also provides one less class to convert when we do drop the SimbodyEngine (which can be tackled incrementally).  

I am fine with merging without the upgrade (to limit the scope of this PR) but I do think it should be part of the overall updating strategy to make Components use update capabilities whenever they can. So a follow up PR would be awesome.

OpenSim/Simulation/Model/PrescribedForce.cpp
```diff
@@ -42,14 +42,13 @@ using namespace std;
...
     setNull();
-    constructProperties();
+    constructInfrastructure();
 
-    _body = body;
-    if (_body)
-        setBodyName(_body->getName());
+    if (frame)
+        updConnector<PhysicalFrame>("frame").setConnecteeName(frame->getName());
```
aseth1: this should also be changed to:
`updConnector<PhysicalFrame>("frame").connect(frame);`
connect will determine the path name to store with the Connector.

OpenSim/Simulation/Model/PrescribedForce.cpp
aymanhab: This specific constructor performs double duty (as a default constructor if passed in pointer is null) and as actual constructor so practically can't be removed entirely instead replaced by 2 constructors.  For SWIG purposes we can't have a const_ref and const pointer signatures since they collide. Seems outside the scope of this PR.

aymanhab:
Thanks @aseth1 I made the change to use Frame methods (please review) and to use the object version of connect & XMl version check hack for now.

OpenSim/Simulation/Model/PrescribedForce.cpp
chrisdembia: I agree that changing to a reference is outside the scope of the PR. Perhaps a middle ground would be to add to the doxygen documentation (after moving it to the header) explaining that this constructor does not take ownership?

chrisdembia: Actually, I'm not sure anymore if it's outside the scope of the PR. I think it's fine we remove the logic that checks for null.

OpenSim/Simulation/Model/PrescribedForce.cpp
```diff
@@ -235,34 +257,36 @@ void PrescribedForce::computeForce(const SimTK::State& state,
...
+    const Ground& gnd = getModel().getGround();
     if (hasForceFunctions) {
         Vec3 force(forceFunctions[0].calcValue(timeAsVector), 
                    forceFunctions[1].calcValue(timeAsVector), 
                    forceFunctions[2].calcValue(timeAsVector));
         if (!forceIsGlobal)
-            engine.transform(state, *_body,                 force, 
-                                    getModel().getGround(), force);
+            force = frame.expressVectorInAnotherFrame(state, force, gnd);
+
```
aseth1: Ohhh so pretty ... :bouquet:

aseth1: But don't forget the best part - deleting the need for the SimbodyEngine on line 253! :boom:

aymanhab: Absolutely, now gone :boat:

OpenSim/Simulation/Model/PrescribedForce.cpp
```diff
@@ -370,35 +394,30 @@ OpenSim::Array<double> PrescribedForce::getRecordValues(const SimTK::State& stat
...
+        getConnector<PhysicalFrame>("frame").getConnectee();
     if (appliesForce) {
-        Vec3 force(forceFunctions[0].calcValue(timeAsVector), 
-                   forceFunctions[1].calcValue(timeAsVector), 
-                   forceFunctions[2].calcValue(timeAsVector));
+        Vec3 force = getForceApplied(state);
         if (!forceIsGlobal)
-            engine.transform(state, *_body, force, 
+            engine.transform(state, frame, force,
                              getModel().getGround(), force);
```
aseth1: :cry: forgot these

OpenSim/Simulation/Model/PrescribedForce.cpp
```diff
@@ -370,35 +394,30 @@ OpenSim::Array<double> PrescribedForce::getRecordValues(const SimTK::State& stat
     const double time = state.getTime();
     const SimbodyEngine& engine = getModel().getSimbodyEngine();
```
aseth1: tic tic tic :bomb:

OpenSim/Simulation/Model/PrescribedForce.cpp
aymanhab: Now we can :fireworks:

aymanhab:
@aseth1 Are you done reviewing this? Thanks.

OpenSim/Simulation/Model/PrescribedForce.cpp
aseth1: I agree it is cleaner to have default constructor that takes no arguments and a proper convenience constructor that  takes both the name of the `PrescribedForce` (and `const sted::string&`) and a `const` reference to a `PhysicalFrame`. I am OK with deprecating the constructor that takes a pointer OR removing it altogether.

aseth1:
Yep! I restarted the PR build (Friday) for a spurious failure and forgot to come back to it. It looks really good @aymanhab!

Do you want to address the separation of the default vs. convenience constructor in this PR. I think it would be a straightforward change to add.

aymanhab:
Sure, @aseth1 I'll do the constructor separation. Thanks

aymanhab:
ok, done.

OpenSim/Simulation/Model/PrescribedForce.cpp
```diff
@@ -37,19 +37,24 @@ using namespace std;
...
+
+}
+
 /**
- * Default constructor.
+ * Convenience constructor.
  */
-PrescribedForce::PrescribedForce(Body* body)
+PrescribedForce::PrescribedForce(const PhysicalFrame& frame)
 {
```
aseth1: It would be great if the convenience constructor required the Force name as well. This helps us avoid having to rename retroactively.

OpenSim/Simulation/Model/PrescribedForce.cpp
```diff
@@ -37,19 +37,24 @@ using namespace std;
...
+    constructInfrastructure();
+
+}
+
 /**
- * Default constructor.
+ * Convenience constructor.
  */
-PrescribedForce::PrescribedForce(Body* body)
+PrescribedForce::PrescribedForce(const PhysicalFrame& frame)
```
aseth1: can also delegate to the default constructor with ` : PrescribedForce()` and eliminate duplication of `setNull()` etc...

aymanhab:
@aseth1 I did the constructor delegation, to avoid code duplication. Good idea :+1:

I don't want to pass in a name into that one constructor because:
1. It seems unrelated to this issue/PR.
2. Objects coming from deserialization will have no name already.
3. The issue is all over the code base for almost every object type, we should address when we have a comprehensive solution. There are open issues related to this topic already. Other items that need to be addressed for example (what's a valid name? scope?, how to handle duplicates, ...etc.)

aseth1:
@aymanhab sorry for piling on more changes. I do want all convenience constructors to take the name of the Component. Handling nameless components is one thing, but using the API should have you do the right thing- and that is to have named Components. For example, I made all `Joint` and `BushingForce`'s  convenience constructors require the name.  Since you are adding the convenience constructor anyways, I thought it might as well have the desired behavior.
I agree there are other issues, for example should the name be pushed down to the Component for verification and throw at construction time, and these are beyond the scope of this PR. I will have to address these more holistically that is true, but it is also true that every Component will require constructors that allow you to specify the Component's name upon construction.

aymanhab:
Thanks @aseth1 I added it for consistency with the other classes that you've changed already. The fact that I can pass empty string or illegal names makes this less useful than we'd want but hopefully is a step toward the cure! Cheers.

aymanhab:
Failures are due to testWrapping timing out which is likely related to PathPoint changes and not to this PR. I think this can go in as is.

aymanhab:
@aseth1 Can you please confirm if this's ready to merge. Thanks.

aymanhab:
@aseth1 Can you please review again or merge, the build failures are unrelated timeouts. Thanks.

aseth1:
Sorry for the delay @aymanhab. It does not appear to be `testWrapping()` in this case:

See [#4059.4](https://travis-ci.org/opensim-org/opensim-core/builds/128969708)

> 61/70 Test #53: testWrapping .................................   Passed  1018.96 sec

I have no explanation except that perhaps Travis [timed out]( https://docs.travis-ci.com/user/common-build-problems/#My-builds-are-timing-out from) from waiting too long for unresponsive test/step - thanks to @tkuchida for sharing the link. I am not convinced that turning off testWrapping in debug is the correct response when it does not appear to be the culprit in this case.

I also agree that your changes likely have little to do with the failure. But since it is repeatable here, we should try to figure out the cause. I have not had a chance to investigate.

aseth1:
@aymanhab I shouldn't have held this up. In it goes!

aymanhab:
No worries, thanks @aseth1

Issue #897, opened 2016-03-27
"dereferenced null pointer" related warnings


klshrinidhi:
There are few places where it is assumed (and relied on) that a reference to a `nullptr` can be passed around. My understanding is that according to C++ standard, dereferencing a null pointer is _undefined-behavior_. Following are the cases I am noticing. There may be more.
- [Vendors/lepton/src/ExpressionTreeNode.cpp](https://github.com/opensim-org/opensim-core/blob/master/Vendors/lepton/src/ExpressionTreeNode.cpp#L62) --
  `getOperation()` returns a reference which can be a reference to a `nullptr` and this line checks if it so. Compiler says it may (in some optimization levels) assume that the condition always evaluates to `false`, which would be bad.
- [Vendors/lepton/src/ParsedExpression.cpp](https://github.com/opensim-org/opensim-core/blob/master/Vendors/lepton/src/ParsedExpression.cpp#L49) -- Same as above.
- [OpenSim/Simulation/Model/GeometryPath.cpp](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/GeometryPath.cpp#L77) -- `aModel` is a reference to a `Model` and from this line, it looks like `aModel` can be a reference to a `nullptr`.
- [OpenSim/Simulation/Model/AbstractTool.cpp](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/AbstractTool.cpp#L466) -- `get()` returns a reference. This line checks if it is returning a reference to a `nullptr`.
- [OpenSim/Simulation/Model/AbstractTool.cpp](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/AbstractTool.cpp#L495) -- `get()` returns a reference. This line checks if it is returning a reference to a `nullptr`.
- [OpenSim/Simulation/Model/ForceSet.cpp](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/ForceSet.cpp#L193) -- `updModel()` returns a reference. This line checks if it is a reference to a `nullptr`.
- [OpenSim/Simulation/Model/ForceSet.cpp](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/ForceSet.cpp#L217) -- `getModel()` returns a reference. This line checks if it is a reference to a `nullptr`.
[...]

PR #939, opened 2016-04-22
[travis] Turn warnings into errors, and ignore certain warnings.

chrisdembia:
@klshrinidhi could you review? This will help us prevent new warnings from entering the code base. If you want to make this more strict based on the warnings you removed, please do so.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -142,7 +142,7 @@ before_install:
 install:
   - mkdir ~/opensim-core-build && cd ~/opensim-core-build
   # Configure OpenSim.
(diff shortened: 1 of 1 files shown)


klshrinidhi:
* Are you sure that the setting of `CMAKE_CXX_FLAGS` on the command line will not unconditionally override any setting we may have in `CMakeLists.txt` ?
* The two extra flags -- `undefined-bool-conversion` and `tautological-undefined-compare` are important. We don't want to introduce more of those because they indicate suspicious code.

chrisdembia:
> Are you sure that the setting of CMAKE_CXX_FLAGS on the command line will not unconditionally override any setting we may have in CMakeLists.txt ?

Yes. A long time ago, I edited our CMakeLists.txt to ensure that we only ever append to this variable. Also, I think it would happen the other way: our CMakeLists.txt would override the value provided on the command line.

> The two extra flags -- undefined-bool-conversion and tautological-undefined-compare are important. We don't want to introduce more of those because they indicate suspicious code.

I think we should fix all cases where that warning occurs, and to stop ignoring those warnings (that is, we should address #897). This PR does not relax anything; it only makes us more strict than we currently are. Right now, we are letting *all* warnings through; this PR only allows two types of warnings through.

chrisdembia:
Thank you!

PR #941, opened 2016-04-23
OpenSim_DECLARE_CONNECTOR macro

chrisdembia:
This is a work-in-progress PR to allow users to create connectors using new OpenSim_DECLARE_CONNECTOR and OpenSim_DECLARE_CONNECTOR_FD/OpenSim_DEFINE_CONNECTOR_FD macros. The last two macros are for the case where the type T for the Connector must be forward-declared; the only place where this is currently the case is Geometry. However, I think it'll come up again, and so we must support forward-declared Connector types. These macros are intended to replace the current way that we specify connectors.

I'd like feedback on this PR, particularly about the OpenSim_DECLARE_CONNECTOR_FD/OpenSim_DEFINE_CONNECTOR_FD macros. I don't think @klshrinidhi will be very satisfied with this. I was hoping for a much better solution, but couldn't think of anything. I'm open to suggestions!

This PR is WIP because I'm missing one thing: handling custom copy constructors for component classes that have connectors. Right now, if you write a custom copy constructor, the Connector will be duplicated.


chrisdembia:
@aseth1 Based on your discussion, I have commented out the `_FD` macros. Before considering merging this, I want to discuss how we would deal with custom copy constructors. I can think of two ways that we could handle custom copy constructors:

1. Component-developers must put something in their initializer list (exposes the guts of the macro and makes it hard for us to make changes in the future to what the macro does):

  ```cpp
  Foo::Foo(const Foo& source) _connector_bar(source._connector_bar) {
  }
  ```

2. In the macro, change the in-class member initializer for `int Foo::_connector_bar` so that it first checks if a connector with the same name exists (prevents us from giving an error if someone tries to create two connectors with the same name):

  ```cpp
  #define OpenSim_DECLARE_CONNECTOR(name, T, ...) \
      ...
      int _connector_##name { hasConnector(#name) ?
                              getConnectorIndex(#name) : 
                              constructConnector<T>(#name) }
  ```

Let's chat in person about this.

chrisdembia:
This is ready for review, if we are going to table the treatment of custom copy constructors.

chrisdembia:
I opened #998 to resolve the issue with custom copy constructors.

OpenSim/Common/Property.h
```diff
@@ -546,8 +546,7 @@ class Property : public AbstractProperty {
         if (p) return *p;
         throw OpenSim::Exception
            ("Property<T>::getAs(): Property " + prop.getName() 
-            + " was not of type " 
-            + std::string(SimTK::NiceTypeName<T>::name()));
+            + " was not of type TODO");
```
aseth1: Why was `std::string(SimTK::NiceTypeName<T>::name())` replaced with `TODO`? Error message no longer makes sense.

OpenSim/Sandbox/ExampleHopperDevice/exampleHopperDevice.cpp_answers
```diff
@@ -134,6 +134,9 @@ public:
         "Gain used in converting muscle activation into a"
         " control signal (units depend on the device)");
 
+    OpenSim_DECLARE_CONNECTOR(actuator, Actuator,
+        "The ScalarActuator for which we're computing a control signal.");
```
aseth1: We are using a `ScalarActuator` primarily for ease of use in the example. I don't think it is a necessity especially since we have a `Connector` to any `Actuator`. We could limit the type to `ScalarAcuator` for the example and then the comment would be consistent.

OpenSim/Simulation/Model/OffsetFrame.h
```diff
@@ -73,6 +73,12 @@ class OffsetFrame : public C {
     "Orientation offset of this frame in its parent frame, expressed as a "
     "frame-fixed x-y-z rotation sequence.");
 
+//==============================================================================
+// CONNECTORS
+//==============================================================================
+    OpenSim_DECLARE_CONNECTOR(parent, C,
+    "The parent frame to this frame.");
```
aseth1: indent or fit on previous line (if <= 80)

OpenSim/Simulation/Control/ControlSet.cpp
```diff
@@ -34,7 +34,7 @@
 #include <OpenSim/Common/Storage.h>
 #include "ControlSet.h"
 #include "ControlLinear.h"
-#include <OpenSim/Simulation/SimbodyEngine/Constraint.h>
+// TODO #include <OpenSim/Simulation/SimbodyEngine/Constraint.h>
```
aseth1: Go ahead and remove this. I cannot imagine why the `ControlSet` needs to know about `Constraints`.

aseth1:
Done my review. I have very minor comments. Thanks for posting the custom copy issue. We will have to solve after Dev Week workshop, but in the meantime the Connector macros will make adding and documenting Component Connectors a lot easier. Thank you @chrisdembia !

OpenSim/Common/Property.h
chrisdembia: Oops. Reverted.

OpenSim/Sandbox/ExampleHopperDevice/exampleHopperDevice.cpp_answers
chrisdembia: Done.

OpenSim/Simulation/Model/OffsetFrame.h
chrisdembia: Done.

OpenSim/Simulation/Control/ControlSet.cpp
chrisdembia: Done.

chrisdembia:
@aseth1 I addressed all comments. Let's 🙏 the tests pass.

chrisdembia:
@aseth1 tests pass.

aseth1:
Nice job @chrisdembia!

chrisdembia:
Thank you for merging.

PR #942, opened 2016-04-24
Xsens IMU streaming

klshrinidhi:
This PR is into Sandbox.
Receive Xsens IMU data data, set the state of the model and visualize. This is currently functional only on Windows because Xsens SDK seems to work with the hardware only Windows.

The GIF image can be removed before merging this PR.

### DIFF
--- a/OpenSim/Sandbox/xsens/CMakeLists.txt
+++ b/OpenSim/Sandbox/xsens/CMakeLists.txt
@@ -1,0 +1,16 @@
+project(IMUStreaming)
+
+cmake_minimum_required(VERSION 3.1.3)
+
+set(CMAKE_CXX_STANDARD_REQUIRED 11)
+set(CMAKE_CXX_STANDARD 11)
+
+set(XSENS_ROOT "/home/shrik/tmp/MT_Software_Suite_Linux_4.5/mtsdk")
+
+include_directories("${XSENS_ROOT}/include")
+include_directories(${CMAKE_CURRENT_SOURCE_DIR})
+set(XSENS_LIBRARIES "${XSENS_ROOT}/lib/libxstypes.so" 
+                    "${XSENS_ROOT}/lib/libxsensdeviceapi.so")
+
+add_executable(IMUStreaming ImuStreaming.cpp)
+target_link_libraries(IMUStreaming ${XSENS_LIBRARIES})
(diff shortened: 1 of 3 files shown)


chrisdembia:
Woohoo! That's awesome, @klshrinidhi .

As for the gif: even if delete the gif before merging, the gif is still in the history. So I suggest we do one of the following:

1. Just leave the gif in the repo.
2. If you don't want the gif in the history (because it takes up space and is not as compressible as a binary blob), create a new branch that doesn't contain the gif in its history. So that we can see the excellent gif, you can upload it into the GitHub PR but not the source code history.

OpenSim/Sandbox/xsens/ImuStreaming.cpp
```diff
@@ -0,0 +1,320 @@
+// Important note: Have to include OpenSim before
+// xsens because otherwise there will be compiler
+// errors in Simbody headers. This happens because
+// xsens sysmbols are in global namespace.
```
chrisdembia: Ahhh!!!

chrisdembia:
Other than what to do with the gif, I think you should merge this! If the tests pass.

OpenSim/Sandbox/xsens/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
+cmake_minimum_required(VERSION 3.1.3)
```
chrisdembia: Remove this line (and the CXX_STANDARD variables); they should be unnecessary. Also, this line would cause OpenSim to require cmake 3.1.3, and we have not yet made a decision on this (PR #925).

chrisdembia:
I take that back; see my comment about the cmake version.

klshrinidhi:
You are right. The history will contain the gif file even if I delete it. If I create another branch without the gif and delete this branch, will we loose the file ?
I don't want to keep the gif around. Its value will reduce over time. It was just to show.

klshrinidhi:
Travis -- `testWrapping` is timing out.
Appveyor -- `testCMCWithControlConstraintsRunningModel` is *segfault*-ing.

They both are un-related to this PR.

chrisdembia:
> The history will contain the gif file even if I delete it. If I create another branch without the gif and delete this branch, will we loose the file ?

My understanding is yes, over time. Git eventually prunes parts of the history tree that are unretrievable (b/c they don't have any branches pointing to them).

PR #943, opened 2016-04-24
Xsens imu streaming without gif

klshrinidhi:
See #942.

This branch is same as the one in #942 but without the gif image.


chrisdembia:
Awesome! I merged despite a test failure because I think the test failure is unrelated.

I didn't delete the branch myself; I wasn't sure if you might still be using it.

klshrinidhi:
Thanks @chrisdembia. It is okay to delete the branch. I am deleting now.

Issue #859, opened 2016-03-18
Give an error when trying to create a property, inputs, outputs, connectors with the same name as one in a super class


chrisdembia:
Right now, a subclass can have a property with the same name as a superclass (if the superclass's property is protected). We could maybe give a warning when creating two properties with the same name in the same class.

chrisdembia:
This should also be done for outputs, inputs, connectors.

chrisdembia:
This shall be fixed by #1340, when it gets merged.

aseth1:
closed by #1340

PR #944, opened 2016-04-25
Add 'force' output to CoordinateActuator

klshrinidhi:


OpenSim/Actuators/CoordinateActuator.h
```diff
@@ -57,6 +57,11 @@ OpenSim_DECLARE_CONCRETE_OBJECT(CoordinateActuator, ScalarActuator);
         "Name of the generalized coordinate to which the actuator applies.");
     OpenSim_DECLARE_PROPERTY(optimal_force, double,
         "The maximum generalized force produced by this actuator.");
+//==============================================================================
+// OUTPUTS
+//==============================================================================
+    OpenSim_DECLARE_OUTPUT(force, double, computeActuation,
+                           SimTK::Stage::Acceleration);
```
chrisdembia: What about naming the output `generalized_force` or `actuation`? The name `force` could be confusing, since the CoordinateActuator may actually be applying a moment (this is actually more common). I know that this is already a problem with the `optimal_force` property, but one may expect output names to be more precise.

chrisdembia:
Cool! This might be redundant with ScalarActuator's `actuation` output.

aymanhab:
@chrisdembia Relying on personal memory to know if an output name is used does not seem scalable, we need a mechanism to lookup or a convention, I'll open an issue unless there's a documented policy for naming outputs.

chrisdembia:
You are definitely correct. One option is to search doxygen; the search should show all outputs with that name (along with all other symbols). The other solution is to throw an exception when creating an output with the same name as one up the hierarchy. There is already a related issue; see #859.

chrisdembia:
The first search result here is the output for `actuation`.

image

The ability to search doxygen for properties, outputs, etc. has not yet been documented.

aymanhab:
Thanks @chrisdembia that will be helpful indeed. 

From a user perspective there's a concern that built-in tools (e.g. code-completion) can't find what outputs are available, so we're trading compile-time/scripting-time checks with runtime checks. Should be fine but we need to make sure trying to get an output that's not available doesn't cause unexpected exceptions or mysterious error messages. How about if you try to get an output e.g "Actuation" that doesn't exist, the error message would display available outputs.. Just a thought.

chrisdembia:
I think that's a good idea.

klshrinidhi:
Closing this. There is an output named `actuation` in the base class.

PR #947, opened 2016-04-26
Create ScaleTool::run() method

chrisdembia:
I moved the meat of the `scale` executable to a new method `ScaleTool::run()`. This will allow people to easily run the scale tool from scripting. Also, this change is largely necessary in preparation for the upcoming command-line tool. 

To test this change, I made some parts of `testScale` use the new `ScaleTool::run()` method.


tkuchida:
:+1: This looks good to me.

chrisdembia:
Thank you for the review. Would anyone else like to review?

klshrinidhi:
What happens if I call `run()` multiple times in a row ?

chrisdembia:
Ah good question. The scale procedure runs again (no caching, same results). I made the `run()` method const (as well as a bunch of other methods).

chrisdembia:
Ready for review again.

chrisdembia:
This is ready to review again.

OpenSim/Tools/ScaleTool.h
```diff
@@ -101,34 +101,43 @@ OpenSim_DECLARE_CONCRETE_OBJECT(ScaleTool, Object);
 #endif
     void copyData(const ScaleTool &aSubject);
 
-    Model* createModel();
+    Model* createModel() const;
     /* Query the subject for different parameters */
-    GenericModelMaker& getGenericModelMaker()
+    GenericModelMaker& getGenericModelMaker() const
     {
```
aseth1: IMHO if this method is `const` it should not provide non `const` access to a data member - in this case its a property and the `const` "contract" is easily violated. It is no better than being non-const in IMO.

chrisdembia: Oh oops; right.

chrisdembia:
@aseth1 I have fixed the issue with accessors, so now they return const references. I thus had to propagate some consts throughout other scaling classes. As a result, I had to make the following two changes:

1. I changed `MarkerPlacer::_outputStorage` to mutable (and also made it a smart pointer). The functionality here should not change. This needs to be mutable since it's altered in `MarkerPlacer::processModel()` which is now const. I think it's fine for this member to be mutable, since it is a cache.
2. The GenericModelMaker used to update its `_fileName` property when `GenericModelMaker::processModel()` was called. Now, `_fileName` is not altered in `processModel()`. I assume that this functionality was for when users write out model files from the GUI's Scale Tool dialog box after running the scale tool. However, I think this change may be for the better, since it does not alter the setup file that the user provides. Also, upon loading the printed setup file in the GUI a second time, the correct path should be calculated again anyway. @aymanhab please let me know if this change is OK. If it is not OK, then it may not be OK for users to call `ScaleTool::run()` multiple times.

aymanhab:
@chrisdembia As far as I can tell, the GUI does not set or change the _filename (by looking at the code) and when you do scale a model in the GUI, the resulting model has no file associated with it, the association happens when saving if ever, but users of the ScaleTool from scripting or command line tools are very likely changing this since that's the only way to get a scaled model out of the tools. Did you consider that workflow? Thanks.

chrisdembia:
Good point. I just tested the ability to use relative paths in the XML Setup file for the OSIM file and then run the command line tool, and there was no issue. So, I have now considered and tested that workflow.

chrisdembia:
The tests have passed on this branch now. Anyone interested in merging this?

chrisdembia:
@aseth1 could you look at this briefly again? I tried to address your comment about constness. @aymanhab and @klshrinidhi seem happy with the PR.

aseth1:
LGTM

chrisdembia:
Woohoo! Thank you.

Issue #896, opened 2016-03-26
TwoFrameLinker::scale overrides sometimes and doesn't override other times


klshrinidhi:
- `TwoFrameLinker<Force, PhysicalFrame>::scale` does not override.
- `TwoFrameLinker<Constraint, PhysicalFrame>::scale` overrides because its base class method `Constraint::scale` is virtual.

This makes it impossible to solve the compiler warning about `override`. Marking the function `override` gives an error for the first case because base class has not virtual function named `scale`. Not marking the function `override` gives a warning in the second case because base class has a virtual function named `scale`.

I think it will be good to discuss and make this consistent for better understanding.

chrisdembia:
Maybe one of these should be `extendScale()`?

aseth1:
I am happy for the virtual method to be `extendScale()`. There are other instances where `scale()` is overridden and we may consider changing those as well.

klshrinidhi:
This is reported as a [bug in Clang](https://llvm.org/bugs/show_bug.cgi?id=22582). I am not sure if Clang developers acknowledge it as such.

aseth1:
Fixed by #1994

PR #950, opened 2016-04-28
Remove warnings

klshrinidhi:
* Add missing override to `ContactMesh`.
* Suppress missing override warning from clang for `TwoFrameLinker::scale()`. This **does not** address #896, rather just suppresses the warning. The issue still stands.

### DIFF
--- a/OpenSim/Simulation/Model/ContactMesh.h
+++ b/OpenSim/Simulation/Model/ContactMesh.h
@@ -94,7 +94,7 @@ public:
 private:
     // INITIALIZATION
     void setNull();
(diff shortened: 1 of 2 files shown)


chrisdembia:
Looks good to me. As I mentioned earlier, I think it would be good not to suppressing unrecognized pragmas on MSVC, even though it would make it even more ugly in the code to suppress warnings.

chrisdembia:
Though I agree that can be a separate PR.

klshrinidhi:
So this PR is now just adding one missing `override`. Can we merge ?

PR #952, opened 2016-04-29
Call cmake executable in a portable way when copying DLLs

elen4:
Currently, copy of BTK DLLs fails on Windows, when CMake is not in the PATH (default condition).
Thanks @lucaTagliapietra for noticing!


chrisdembia:
Looks good to me. Another reason why this is good: people could have multiple versions of CMake on their computer, and `${CMAKE_COMMAND}` would ensure use of the correct version.

I'll let @klshrinidhi merge.

klshrinidhi:
Thanks :bow: for the PR.
Keep it coming. :+1:

Issue #945, opened 2016-04-26
Make it easier to get OpenSim's version number from bindings.


chrisdembia:
Do we have this ability right now? This would be helpful for people who may have multiple versions of OpenSim installed.

aymanhab:
opensimModelCommonJNI.GET_OSIM_VERSION_get() in Java

aymanhab:
@chrisdembia If you can make similar call in Python we can close this issue. Thanks

chrisdembia:
Ah the python bindings have `opensim.GetVersion()`, `opensim.GetVersionAndDate()`, etc. Pretty good. Do you have these methods in Java/MATLAB?

chrisdembia:
Another conventional way of getting version info in Python is `opensim.__version__`. I'd like to leave this issue open until we add this.

aymanhab:
This issue can now be closed,

PR #954, opened 2016-05-01
Provide python package version in __version__.

chrisdembia:
This PR allows python users to get the version of OpenSim in a way that is somewhat standard in the python community:

```
import opensim
print(opensim.__version__)
```

Addresses part of #945.

### DIFF
--- a/Bindings/Python/CMakeLists.txt
+++ b/Bindings/Python/CMakeLists.txt
@@ -200,12 +200,15 @@ OpenSimAddPythonModule(MODULE tools
 # Note: some of the commands to do this copying (for the swig-generated py
 # files) appear above.
 
-# Configure setup.py
-configure_file(${CMAKE_CURRENT_SOURCE_DIR}/setup.py.in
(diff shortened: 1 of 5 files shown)


chrisdembia:
Oh I should say that I based this PR in part by what [pydy](https://github.com/pydy/pydy) does.

klshrinidhi:
Looks great.
Can you add a test to the Python tests to check the version ? It can serve both as example & test.

chrisdembia:
Ah good idea. Yes I'll do that.

aymanhab:
@chrisdembia This looks good but I want to make sure it doesn't change the installation process/commands or require documentation changes? Thanks

chrisdembia:
It does not change the installation process (I tested installing the package). I think it could be nice to document this but I wasn't sure how. I can't update the confluence page about python scripting because that page is for released versions.

aymanhab:
Great, thanks.

chrisdembia:
> Can you add a test to the Python tests to check the version ?

Done.

Ready for review again.

chrisdembia:
Thank you.

PR #955, opened 2016-05-02
[Do Not Merge] Precompiled headers to reduce build time

klshrinidhi:
This PR is just to document the failed attempt the reduce build time using precompiled headers. I did see some improvement in build time but it was not significant enough to justify using this technique. I tried only with GCC and Clang. I didn't think it was worth trying MSVC after that. Here are the important points I realized:
* Not all headers benefit (build time) by pre-compiling. it is not about how big (number of lines) the header is. There is trial and error involved in choosing which headers to precompile. This means we may need a *mixed* setup where some headers are pre-compiled and some are not, which may be hard to maintain over time. Precompiling all headers like I have done does not produce significant improvement through it never underperforms.
* GCC produced precompiled headers of size >350MB. Clang produced headers of size >45MB. Since the compiler needs to read in the precompiled header for every cpp file, it may be inefficient for many cpp files when compared to reading just the required headers for the cpp and compiling them on the fly. If there were a way to tell compiler to keep the precompiled header in memory, I guess it would help a lot, especially with computers hosting mechanical hard-drives.
* With precompiled headers, it is possible to explicitly instantiate templates so compiler saves time instantiating frequently used templates. *[Explicit instantiation is possible in other ways too using `extern template` but this is different.]* With lack of tools to profile template instantiations, this also becomes trial and error. I found that `SimTK::Matrix_` hierarchy benefited from explicit instantiation. I will try to explore just explicit instantiation in another PR.

I will close this PR and delete the branch. :disappointed:


chrisdembia:
+1 cool effort!

PR #956, opened 2016-05-03
Assembly and InverseKinematics Solver cleanup

aseth1:
Update the `AssemblySolver` and `InverseKinematics` Solvers as well as the Markers and CoordinateReference classes employed by the Solvers to use the new property system and make internal members private.


OpenSim/Simulation/AssemblySolver.h
```diff
@@ -126,6 +106,9 @@ OpenSim_DECLARE_CONCRETE_OBJECT(AssemblySolver, Solver);
         to track a desired trajectory of coordinate values. */
     virtual void track(SimTK::State &s);
 
+    /** Read access to the underlying SimTK::Assembler. */
+    const SimTK::Assembler& getAssembler() const { return *_assembler; }
```
chrisdembia: No need to make a change for this PR, but in the future, we should consider throwing an exception when we mask a pointer as a reference (since references should never be null).

OpenSim/Simulation/AssemblySolver.h
```diff
@@ -137,9 +120,27 @@ OpenSim_DECLARE_CONCRETE_OBJECT(AssemblySolver, Solver);
...
+    double _accuracy;
+
+    // Weight for built-in constraints to be satisfied
+    double _constraintWeight;
+
+    // The coordinates reference value and weighting. This is full copy rather than ref.
+    SimTK::Array_<CoordinateReference> _coordinateReferencesp;
+
+    // Underlying SimTK::Assembler that will perform the assembly
+    SimTK::ResetOnCopy< std::unique_ptr<SimTK::Assembler>> _assembler;
```
chrisdembia: +1

OpenSim/Simulation/InverseKinematicsSolver.cpp
```diff
@@ -187,29 +184,29 @@ double InverseKinematicsSolver::computeCurrentSquaredMarkerError(int markerIndex
...
-/** Internal method to convert the MarkerReferences into additional goals of the 
+/* Internal method to convert the MarkerReferences into additional goals of the 
     of the base assembly solver, that is going to do the assembly.  */
 void InverseKinematicsSolver::setupGoals(SimTK::State &s)
 {
     // Setup coordinates performed by the base class
     AssemblySolver::setupGoals(s);
 
-    _markerAssemblyCondition = new SimTK::Markers();
+    _markerAssemblyCondition.reset(new SimTK::Markers());
```
chrisdembia: This type of situation has come up before ( (in Coordinate)[https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/SimbodyEngine/Coordinate.cpp#L182]), since the Assembly is actually the owner. @sherm1 came up with a way to deal with this; see the link. Can you change the code to be like the following?

```cpp
std::unique_ptr<SimTK::Markers> condOwner(new SimTK::Markers());
_markerAssemblyCondition.reset(condOnwer.get());
...
updAssembler().adoptAssemblyGoal(condOwner.release()); // condOwner gives up ownership
```

OpenSim/Simulation/InverseKinematicsSolver.h
```diff
@@ -140,16 +128,23 @@ class OSIMSIMULATION_API InverseKinematicsSolver: public AssemblySolver
...
+
+    // Markers collectively form a single assembly condition for the SimTK::Assembler
+    // and the memory is managed by the Assembler
+    SimTK::ReferencePtr<SimTK::Markers> _markerAssemblyCondition;
+
     // Non-accessible cache of the marker values to be matched at a given state
     SimTK::Array_<SimTK::Vec3> _markerValues;
 
-
+    std::unique_ptr<SimTK::OrientationSensors> _orientationAssemblyCondition;
```
chrisdembia: As far as I can tell, this is unused. Also, I didn't think you could make a `std::unique_ptr` member variable because I though `std::unique_ptr` is not copyable. Anywho, how about commenting this out for now until we support orientation sensors?

OpenSim/Tools/MarkerPlacer.cpp
```diff
@@ -242,17 +242,17 @@ bool MarkerPlacer::processModel(Model* aModel, const string& aPathToSubject)
     /* Load the static pose marker file, and average all the
      * frames in the user-specified time range.
      */
-    MarkerData staticPose(aPathToSubject + _markerFileName);
+    MarkerData* staticPose = new MarkerData(aPathToSubject + _markerFileName);
```
chrisdembia: IMO It's slightly better to use a unique_ptr and then release it when you pass this pointer onto the `MarkersReference`. The reason is that the memory is still leaked if an exception is thrown before you construct the `MarkersReference` (for example, 2 lines down, there's an exception).

```cpp
std::unique_ptr<MarkerData> staticPose(new MarkerData(...));
...
MarkersReference markersReference(staticPose.release(), &markerWeightSet);
```

OpenSim/Simulation/MarkersReference.h
```diff
@@ -152,32 +113,42 @@ OpenSim_DECLARE_CONCRETE_OBJECT(MarkersReference, Reference_<SimTK::Vec3>);
...
     //--------------------------------------------------------------------------
     // Convenience Access
     //--------------------------------------------------------------------------
     double getSamplingFrequency() {return _markerData->getDataRate(); }
-    Set<MarkerWeight> &updMarkerWeightSet() {return _markerWeightSet; }
-    void setMarkerWeightSet(Set<MarkerWeight> &markerWeights);
-    void setDefaultWeight(double weight) {_defaultWeight = weight; }
+    Set<MarkerWeight> &updMarkerWeightSet() {return upd_marker_weights(); }
+    void setMarkerWeightSet(const Set<MarkerWeight>& markerWeights);
+    void setDefaultWeight(double weight) { upd_default_weight() = weight; }
```
chrisdembia: No need to change, but why not use `set_default_weight(weight);`?

OpenSim/Simulation/InverseKinematicsSolver.h
aseth1: Whoops .. how did that sneak in there :open_mouth:  I will remove.

OpenSim/Simulation/MarkersReference.h
```diff
@@ -152,32 +113,42 @@ OpenSim_DECLARE_CONCRETE_OBJECT(MarkersReference, Reference_<SimTK::Vec3>);
...
 
     //--------------------------------------------------------------------------
     // Convenience Access
     //--------------------------------------------------------------------------
     double getSamplingFrequency() {return _markerData->getDataRate(); }
-    Set<MarkerWeight> &updMarkerWeightSet() {return _markerWeightSet; }
-    void setMarkerWeightSet(Set<MarkerWeight> &markerWeights);
-    void setDefaultWeight(double weight) {_defaultWeight = weight; }
+    Set<MarkerWeight> &updMarkerWeightSet() {return upd_marker_weights(); }
+    void setMarkerWeightSet(const Set<MarkerWeight>& markerWeights);
```
chrisdembia: What do you think about a doxygen comment saying that starting in 4.0 this method now makes a copy (instead of referencing the objects in the passed-in set) in case people previously had code where they continued to modify their set, thinking it affected the MarkersReference?

OpenSim/Simulation/MarkersReference.h
```diff
@@ -123,24 +90,18 @@ OpenSim_DECLARE_CONCRETE_OBJECT(MarkersReference, Reference_<SimTK::Vec3>);
...
-    // Convenience load markers from a file
-    MarkersReference(const std::string filename, Units modelUnits=Units(Units::Meters));
-
-    MarkersReference(MarkerData& aMarkerData, const Set<MarkerWeight>* aMarkerWeightSet=NULL);
+    /** Convenience load markers from a file */
+    MarkersReference(const std::string& markerFileName,
+                     Units modelUnits=Units(Units::Meters));
+    /** Form a Reference from Marker data (taking ownership) and weights */
+    MarkersReference(MarkerData* markerData,
+        const Set<MarkerWeight>* markerWeightSet=nullptr);
```
chrisdembia: This constructor is a little confusing because `MarkersReference` takes ownership of `markerData` but makes a copy of `markerWeightSet`. Can you make this more clear in the doxygen?

chrisdembia:
Done reviewing. The code looks a lot prettier now.

chrisdembia:
Also, no need to change this in this PR, but I prefer usage of `set_prop_name(x)` over `upd_prop_name() = x` when possible (for single-value properties).

chrisdembia:
Looks good to me. Thanks for addressing those comments.

aseth1:
You are free to merge since this PR does not introduce any new functionality.

PR #959, opened 2016-05-05
Fix memory leak in testmuscles

klshrinidhi:
* Fix memory leak in `testMuscles`. <br />
I think it will be good to change the interface of `Model::add<>` methods to accept only `std::unique_ptr`s if the intention is to take over the ownership of the object passed. This will make the caller code clear to read since passing a `std::unique_ptr` means transferring ownership. 

* Fix warning on *missing override*.

### DIFF
--- a/OpenSim/Actuators/Test/testMuscles.cpp
+++ b/OpenSim/Actuators/Test/testMuscles.cpp
@@ -301,7 +301,7 @@ void simulateMuscle(
 
     // Create a prescribed controller that simply 
     //applies controls as function of time
-    PrescribedController * muscleController = new PrescribedController();
+    std::unique_ptr<PrescribedController> muscleController{};
     if(control != NULL){
         muscleController->setActuators(model.updActuators());
         // Set the individual muscle control functions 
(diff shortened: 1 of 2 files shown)


chrisdembia:
Looks good to me.

I agree we need to consider how adoption works with those methods. I'm not sure how we would handle the fact that `unique_ptr` isn't available with SWIG. I think it may be sufficient to rename those methods to `Model::adopt<>`.

klshrinidhi:
Good point. SWIG has support for `std::shared_ptr`. Maybe we should change the interface to use that until `std::unique_ptr` is supported. That will still be a great improvement over current situation. Right now, 
* One has to read the documentation to know what `adoptee` means in the parameters of these methods. Ownership is not mentioned in most of these methods.
* More importantly, the caller code has to pass a raw pointer to a dynamically allocated object which may lead to memory leaks (like the one this PR fixes) if caller is not careful. Forcing use of *any* smart pointer will alleviate the situation.

klshrinidhi:
Can we merge ?

PR #960, opened 2016-05-05
Fix memory leaks in testforces

klshrinidhi:
In `testForces', objects of class `Model` are dynamically allocated in every test function without any need. Though these leaks will not have any effect on main code, it will be good to get rid of these leaks anyway because users may try to copy this code and these leaks may lurk into their code.

Going through the tests, I find it confusing to keep track of ownership of dynamically allocated objects. It is easy to the make the mistake of passing the address of a statically allocated object to one of these functions and enter the *undefined-behavior* land.


chrisdembia:
LGTM

klshrinidhi:
Talking about LGTM, what do you think about using:
https://lgtm.co
for our repository ?

chrisdembia:
Oh interesting...I'm not sure what I think. We could at least try it out for a little bit? Let's discuss quickly tomorrow in the meeting?

aymanhab:
The changes look good (for a test case), much nicer. 

I would strongly suggest against doing something similar for examples that users look into where the syntax of using smart-pointers and things like std::move are incomprehensible to most of our users and do not translate into scripting transparently.

PR #961, opened 2016-05-05
[travis] Disable testWrapping in debug

chrisdembia:
It has been timing out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -131,7 +131,7 @@ before_install:
   ## On OSX, we know that testInitState fails; exclude it.
   - if [ "$TRAVIS_OS_NAME" = "osx" ]; then export TESTS_TO_EXCLUDE="$TESTS_TO_EXCLUDE|testInitState"; fi
   ## If we are building in debug, there are some tests to ignore.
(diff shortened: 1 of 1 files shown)


aseth1:
It still baffles me why we are getting `The job exceeded the maxmimum time limit for jobs, and has been terminated` [here](https://travis-ci.org/opensim-org/opensim-core/jobs/127594741), even though it did not exceed 50min.

chrisdembia:
Yeah I don't get it either. The other weird thing is that *other* travis builds regularly go over 1 hour. 😖

tkuchida:
I'm less interested in why Travis bails after 48min49s rather than 50min, and more interested in why testWrapping takes 17min—are we getting 17min worth of information from the test? Also, note [from this page](https://docs.travis-ci.com/user/common-build-problems/#My-builds-are-timing-out) that a timeout can also occur if no log output is produced for 10min. :watch:

chrisdembia:
I wonder if it's a recent regression.

tkuchida:
Hm… testWrapping only takes a long time (17.5min) on clang in Debug mode; in all other AppVeyor and Travis environments, it takes no more than 4min. :hourglass:

chrisdembia:
Would someone like to merge this?

PR #962, opened 2016-05-06
[cmake] Use correct file dependencies for swig.

chrisdembia:
During cmake's configure step, I run swig on the interface files and ask it to give back a list of the file dependencies for that interface file. Then I pass those files onto CMake as dependencies for generating swig source code. This may substantially improve the process of developing bindings.

The only downside is that the time spent configuring has increased from 1.4 s to 3.4 s (when building only python bindings). However, I think this delay is worthwhile, since much more than 2 seconds are saved when developing the bindings, and this delay is not present if you're not building the bindings.

### DIFF
--- a/Bindings/Java/OpenSimJNI/CMakeLists.txt
+++ b/Bindings/Java/OpenSimJNI/CMakeLists.txt
@@ -6,32 +6,30 @@ find_package(JNI 1.6 REQUIRED)
 
 function(OpenSimGenerateJavaWrapper
         NAME INPUT_INTERFACE_FILE OUTPUT_CXX_FILE OUTPUT_H_FILE)
+
+    set(_swig_common_args -c++ -java
+            -package ${OPENSIM_JAVA_WRAPPING_PACKAGE}
+            -I${OpenSim_SOURCE_DIR}
+            -I${OpenSim_SOURCE_DIR}/Bindings
+            -I${Simbody_INCLUDE_DIR}
+            ${SWIG_FLAGS}
+            ${INPUT_INTERFACE_FILE}
+            )
+
+    OpenSimFindSwigFileDependencies(_${NAME}_dependencies ${NAME}
+        "${_swig_common_args}"
+        )
+
     add_custom_command(
         # This target actually creates a lot more (all the produced .java files)
         # but we will just use these two files as a proxy for all of those.
         OUTPUT ${OUTPUT_CXX_FILE} ${OUTPUT_H_FILE}
-        COMMAND ${SWIG_EXECUTABLE} -v -c++ -java
-                -package ${OPENSIM_JAVA_WRAPPING_PACKAGE}
-                -I${OpenSim_SOURCE_DIR}
-                -I${OpenSim_SOURCE_DIR}/Bindings
-                -I${Simbody_INCLUDE_DIR}
-                ${SWIG_FLAGS}
-                -o ${OUTPUT_CXX_FILE}
-                -outdir ${SWIG_JAVA_SOURCE_BUILD_OUTPUT_DIR}
-                ${INPUT_INTERFACE_FILE}
-        DEPENDS ${INPUT_INTERFACE_FILE}
-        "${CMAKE_SOURCE_DIR}/Bindings/preliminaries.i"
-        "${CMAKE_SOURCE_DIR}/Bindings/simbody.i"
-        "${CMAKE_SOURCE_DIR}/Bindings/common.i"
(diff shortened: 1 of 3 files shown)


klshrinidhi:
This looks like a glorified GLOB like solution to me.
Can we not just list the header files as dependencies ? 
If the concern is that list is too long, we could keep the list of header and cpp files in two separate  CMake files and read them when needed. When developers add new headers/cpp, they need to append their file to the list.

chrisdembia:
I think GLOBs might give incomplete dependencies; the nice thing about this solution is that the dependencies are correct. We could also ask each library target for its list of source files and list those as dependencies, but I'm not sure we wrap everything. Right now, we manually list classes in many places (RegisterTypes_<lib>.cpp, osim<lib>.h, interface files, and fake header files for swig). I'm weary of yet another manual place to list files. I would prefer the glob solution to another manual list. There would be chance for error, and we would have to be careful that we do not cause a rebuild of *all* wrapping libraries if only changing a header in the Tools library.  I think this solution reduces the mental burden of managing dependencies.

Could you clarify what you mean by " we could keep the list of header and cpp files in two separate CMake files and read them when needed."?

Also note that the delay in configuring only happens when re-running CMake, as occurs when editing a CMakeLists.txt. If just re-building (e.g., `make PythonBindings`), then CMake does not need to re-run and the delay doesn't happen.

aymanhab:
@chrisdembia on Mac/XCode it's not clear how to run swig, I basically have to clean and rebuild the target ALL_BUILD, do you think this will be fixed by this PR? Are there other open issues similar to what I described? I think @klshrinidhi was running into similar issue where SWIG was never ran unless forced. Thanks.

chrisdembia:
I don't know if I have tried building the bindings with Xcode; you should be able to build the PythonBindings or JavaBindings target. This PR does not address any potential issue with Xcode not running swig.

chrisdembia:
This PR only addresses issue #790, as far as I know.

klshrinidhi:
> Could you clarify what you mean by " we could keep the list of header and cpp files in two separate CMake files and read them when needed."?

I am suggesting that we keep a file with list of headers we wrap.
```
set(INCLUDES
    file1.h
    file2.h
    file3.h
    ...)
```
And use this along with any other targets wrapping depends on. I think it will be simple and straightforward to *see* the dependencies and add/remove entries as we go.

In general, I prefer listing dependencies explicitly in CMake rather than getting them through GLOB or SWIG. I would like to see the GLOBs go away in CMakeLists.txt in Common, Simulation, etc. For that, I suggest we keep a file in each directory (Common, Simulation, ...) which sets two variables.
```
set(INCLUDES
    file1.h
    file2.h
    file3.h
    ...)
set(SOURCES
    file1.cpp
    file2.cpp
    file3.cpp
    ...)
```
and use it to add library targets. Developers will have to add/remove entries here which is not much extra work when he/she spent time creating a source file.

chrisdembia:
I'm happy to close this PR and accept a PR that manually lists file dependencies for SWIG, and we can see how that works.

I know that CMake thinks using GLOBs is a bad idea for adding files to a target, and says to list source files manually. We can discuss that separately, I think. 

I am realizing that the SWIG situation is different from the `add_library()` and `add_executable()` cases, because it can be completely automated; I don't see a clear argument why we should do something manually that can be automated well. If the issue is *seeing* a list of these dependencies, it is easy to print out the list of automatically-detected dependencies (I did this during testing of this PR). Furthermore, all the dependencies are the files specified in the interface file; so they are available to see there to.

As for the `add_library()` and `add_executable()` scenarios, the CMake user/developer really is the source of the knowledge of the dependencies (this is not true for SWIG; for SWIG, the interface file is the source of this knowledge). I think it would still be neat if we could make CMake "watch" an entire directory; I think it's a fine convention to say "I *know* all source files in this directory are exactly the dependencies for the target." Just seems good to automate things that can be automated. For my upcoming PR on a command line interface, I do manually list all source file dependencies.

</ramble>

klshrinidhi:
If others like this solution, feel free to merge. We can always come back and fix it if we don't like.

aymanhab:
I like @chrisdembia solution since it's more automated and less files to maintain. The main time consuming issue I've been running into (on regular basis) is that there's no direct dependency of bindings on header files which costs way more than extra 2 min in configuration step. I'd lean to merge this as is and see how it works.

chrisdembia:
Thank you for merging. We'll see how this goes. If we don't like it we can revert.

PR #963, opened 2016-05-06
Update ContactGeometry to use a Connector to a PhysicalFrame.

chrisdembia:
I also performed a bunch of cleanup.

This PR is part of the list of components to update that @aseth1 made.

I also deprecated a bunch of accessors/mutators in ContactGeometry.


OpenSim/Simulation/Model/ContactGeometry.cpp
```diff
@@ -81,90 +74,110 @@ void ContactGeometry::constructProperties()
...
-SimTK::Transform ContactGeometry::getTransform()
+SimTK::Transform ContactGeometry::findTransformInBaseFrame() const
 {
-    return SimTK::Transform(Rotation(SimTK::BodyRotationSequence,
-        get_orientation()[0], SimTK::XAxis,
-        get_orientation()[1], SimTK::YAxis,
-        get_orientation()[2], SimTK::ZAxis), get_location());
+    // B: base
+    // F: frame specified in the connector for this object.
+    // P: the (imaginary) frame that is defined (relative to F) by the location
```
aseth1: imaginary->intermediate

OpenSim/Simulation/Model/ContactGeometry.cpp
```diff
@@ -81,90 +74,110 @@ void ContactGeometry::constructProperties()
...
-void ContactGeometry::setBody(PhysicalFrame& body)
+void ContactGeometry::setFrame(const PhysicalFrame& frame)
 {
-    _body = &body;
-    set_body_name(body.getName());
+    updConnector<PhysicalFrame>("frame").connect(frame);
 }
 
-const std::string& ContactGeometry::getBodyName()
+const std::string& ContactGeometry::getFrameName() const
```
aseth1: Having `get/setFrame` is minimally complete. Now that we can connect directly to the frame and we aren't setting string properties anymore, we don't need the name-based interface anymore. Now seems like a good time to remove these.

OpenSim/Simulation/Model/ContactGeometry.h
```diff
@@ -33,120 +33,91 @@ namespace OpenSim {
...
+     *                     the provided `frame`
      */
-    void setOrientation(const SimTK::Vec3& orientation);
+    ContactGeometry(const SimTK::Vec3& location,
+                    const SimTK::Vec3& orientation,
+                    const PhysicalFrame& frame);
+
+
+    // ACCESSORS
 #ifndef SWIG
```
aseth1: Is the SWIG guard still necessary?

OpenSim/Simulation/Model/ContactGeometry.cpp
chrisdembia: Good point. Is it OK if I remove `getFrameName()` and `setFrameName()` but leave in the deprecated `getBodyName()` and `setBodyName()`? They would do the same thing as `getFrameName()` and `setFrameName()` are doing now, with deprecation messages saying to use `getBody().getConnecteeName()` and `updConnector("frame").setConnecteeName()`? If you don't like this, then I can just remove all 4 methods.

OpenSim/Simulation/Model/ContactGeometry.cpp
chrisdembia: I don't really like my choice of "imaginary" but is the frame intermediate? It's a leaf. The important thing I want to convey is that this is not an actual OpenSim `Frame` object; it's just a frame defined by the `location` and `orientation` properties. I'll try to make this clearer.

chrisdembia:
@aseth1, I have addressed your feedback.

- I removed "imaginary" and instead used "Frame" and "PhysicalFrame" to refer to classes, and lower-case "frame" to refer to the "imaginary" one. I felt that "intermediate" could be confusing, since the lower-case "frame" is necessarily a leaf in the tree of frames.
- I removed the SWIG guards :)
- I removed the 4 methods `getFrameName()`, `setFrameName()`, `getBodyName()`, `setBodyName()`

OpenSim/Simulation/Model/ContactGeometry.h
```diff
@@ -33,120 +33,84 @@ namespace OpenSim {
...
-    SimTK::Transform getTransform();
+
+    /** Get a Transform representing the position and orientation of the
+     * geometry within the base Frame (e.g., a Body or Ground) it is attached
+     * to (*not* the "frame" that the geometry is connected to). If `B` is the
+     * base or Body Frame, `F` is the PhysicalFrame that this geometry is
+     * connected to, and `P` is the frame that is defined (relative to `F`) by
+     * the location and orientation properties of this class, then this returns
+     * `X_BP = X_BF * X_FP`. */
+    SimTK::Transform findTransformInBaseFrame() const;
```
aseth1: Why is this necessary? The `ContactGeometry` could provide its transform in Ground for convenience and use that draw itself in `generateDecorations()`. I think it is confusing and unnecessary to introduce the user of a `ContactGeometry` to the concept of a base frame.

chrisdembia: I think you are right, that the geometry could be drawn with a transform to ground instead. But this method is still nice to have when building the [`SimTK::HuntCrossleyForce`](https://github.com/opensim-org/opensim-core/pull/963/files/13bdeebd722bcff3278763b204762c57d10bed09#diff-82843d67d539b05cee170376c728fe38R79) or [`SimTK::ElasticFoundationForce`](https://github.com/opensim-org/opensim-core/pull/963/files/13bdeebd722bcff3278763b204762c57d10bed09#diff-df1955ed2d738ca938e55a7d6d8c7863R74), which I don't think we could specify relative to ground:

```c++
contacts.addBody(set, geom.getFrame().getMobilizedBody(),
                     geom.createSimTKContactGeometry(),
                     geom.findTransformInBaseFrame());
```

Another reason for having this method is that it does nearly the same thing as the previous `ContactGeometry::getTransform()`, reducing the chance that users will introduce a bug into their code when upgrading to 4.0. 

I agree that the name is unfortunate. I would have preferred `findTransformInBody()`. The reason I chose to go with `findTransformInBaseFrame()` was because the base frame may not be an `OpenSim::Body`; it may be `OpenSim::Ground`. I attempted to document this in this method's doxygen.

Considering all this, let me know what you'd like me to do.

chrisdembia: @aseth1 see my comment about `ContactGeometry::findTransformInBaseFrame()`

OpenSim/Simulation/Model/ContactGeometry.cpp
```diff
@@ -81,90 +74,95 @@ void ContactGeometry::constructProperties()
...
-}
+{ set_orientation(orientation); }
 
-const PhysicalFrame& ContactGeometry::getBody() const
+SimTK::Transform ContactGeometry::findTransformInBaseFrame() const
 {
-    return *_body;
+    // B: base Frame (Body)
+    // F: PhysicalFrame specified in the connector for this object.
+    // P: the frame that is defined (relative to F) by the location
```
aseth1: I still do not understand why it is the ContactGeometry's responsibility to provide X_BP. If the user only provides P in F then return that. If the ContactForce implementation requires a transform wr.t. MobilizedBody it can always ask any F for its BaseFrame. I would prefer that in order to simplify ContactGeometry.

OpenSim/Simulation/Model/ContactGeometry.h
aseth1: I would prefer that the guts of the contact implementation become a tad more complex than introduce the concepts of `BaseFrame` or even that of a `MobilizedBody` frame to interface of ContactGeometry. ContactGeometry is conceptually complete with Geometry attached to a (Physical)Frame (with a build in convenience to be offset from that frame). Whoever consumes the `ContactGeometry` (e.g. `HuntCrossleyForce`) should drill down to a "Body" if/when it needs it.

I rather have the complexity in the implementation of the Force than in the interface of `ContactGeometry`. For example, I see nothing wrong with:

```cpp
contacts.addBody(set, geom.getFrame().getMobilizedBody(),
                     geom.createSimTKContactGeometry(),
                     geom.getFrame().findTransformInBaseFrame());
```
since you had to ask the frame to get its `MobilizedBody`  anyways.

Let's discuss it in person tomorrow.

chrisdembia: I agree with your points (about the scope of the ContactGeometry class). My concern is reducing the chance that users introduce a bug in their code because they have to do some transform multiplication themselves. I'll try to remedy that with good doxygen comments. I will:

1) remove `findTransformInBaseFrame()`.
2) create `findTransformInFrame()`
3) give a doxygen comment to `findTransformInFrame()` that describes how to get "transform in base frame" (this would still involve mentioning base/Ground/Body frames in the header for ContactGeometry).
4) as for the old `getTransform()` method, I assume your preference is to delete it but my preference would be to implement it so that it still has the same functionality but deprecate it. Let me know what you think.

And yeah; we can discuss in person (should be quicker that way); I just wanted to get down these thoughts before I forgot them.

chrisdembia:
I've updated this PR based on our discussion earlier today: I removed `findTransformInBaseFrame()` and returned `getTransform()` to its original implementation.

OpenSim/Simulation/Model/ContactGeometry.h
```diff
@@ -33,120 +33,92 @@ namespace OpenSim {
...
+     *
+     * If you want the transform of this geometry relative to the Body (or
+     * Ground) `B` in which this geometry is fixed, you can use the following
+     * code:
+     * @code{.cpp}
+     * const auto& X_BF = geom.getFrame().findTransformInBaseFrame();
+     * const auto X_FP = geom.getTransform();
+     * const auto X_BP = X_BF * X_FP;
+     * @endcode
+     *
```
aseth1: :+1:

aseth1:
@chrisdembia changes look good. The test passing with older models is reassuring that the meaning of `getTransform()` for the older models has not changed. My only concern is that for new models where a `ContactGeometry` can be offset by many intermediate frames that case is not explicitly being tested since old models always had geometry bolted onto Bodies. The code looks right, but it would be reassuring to add a test case where `ContactGeometry` is affixed to a `PhysicalOffsetFrame` (at the end of series of them, even) just to verify the new capability. What do you think?

chrisdembia:
> What do you think?

Yes; definitely. Especially to detect regressions.

chrisdembia:
@aseth1 I have added a test. Please review again.

OpenSim/Simulation/Test/testContactGeometry.cpp
```diff
@@ -339,16 +354,206 @@ void compareHertzAndMeshContactResults()
...
+    }
+
+    // The X and Y offsets are both accomplished with the PhysicalOffsetFrame.
+    SimTK::State stateIntermedFrameXY;
+    {
+        Model model = createBaseModel();
+        auto* frame = new PhysicalOffsetFrame("intermed",
+                model.updBodySet().get("point"),
+                // Frame is up 1m in the y direction.
+                SimTK::Transform(Vec3(0.5, 1, 0)));
```
aseth1: I was expecting you to use two `PhysicalOffsetFrames` to extend from the Y only offset from before. In which case you could just compare this answer to the weld and skip the Y only test. You could also combine rotational offsets. Since you are comparing to the welded mechanism anyway, you could use some random-like transform in both cases (Weld vs. Offset(s)).

chrisdembia: I like the idea of including a rotation offset. I might need some in-person clarification. In general, I like the idea of using random-like transforms to increase coverage of possible scenarios and to avoid trivial transforms that may weaken tests. But I think that in this case, a random-like transform doesn't improve the coverage, and makes the test harder to understand. I don't know if I've fully thought it through, so I'm not sure.

chrisdembia:
@aseth1 please review again. I added a test for the ContactGeometry's orientation property. I originally planned to apply the orientation offset to the ball geometry, but that doesn't make sense, since the sphere is symmetrical. So I applied the orientation offset to the ContactHalfSpace.

OpenSim/Simulation/Test/testContactGeometry.cpp
```diff
@@ -380,80 +380,101 @@ void compareHertzAndMeshContactResults()
...
+//                    .                -----(ball)
+//                    .   (platform)   |
+//                    . 90deg.-`       |
+//                    .   .-`          | 1 m
+//                    ..-`     1 m     |
 //                  (hinge)----------(mass)
-// 
-// The pendulum starts at 27 degrees and then the bar drops down and hits the
-// ground.
+//
```
aseth1: The 45 degree line is confusing especially with the `90deg` annotation. Otherwise, your description is clear. You could try to draw the plane.

chrisdembia: Updated.

chrisdembia:
Tests pass. Ready for review.

chrisdembia:
🎆  Thank you.

PR #964, opened 2016-05-06
use attachGeometry() instead of attachMeshGeometry()

klshrinidhi:
The calls to `attachMeshGeometry()` are referring to a file that does not exist in the repository. It is now possible to use `attachGeometry()` without needing to read files.

*This came up while working on another task.*

### DIFF
--- a/OpenSim/Actuators/Test/testActuators.cpp
+++ b/OpenSim/Actuators/Test/testActuators.cpp
@@ -373,7 +373,7 @@ void testClutchedPathSpring()
     // body the path spring is connected to at both ends
     OpenSim::Body* block =
         new OpenSim::Body("block", mass ,Vec3(0),  mass*Inertia::brick(0.2, 0.1, 0.1));
-    block->attachGeometry(Brick{Vec3{0.2, 0.1, 0.1}});
+    block->attachGeometry(Brick(Vec3(0.2, 0.1, 0.1)));
     block->scale(Vec3(0.2, 0.1, 0.1), false);
 
     //double dh = mass*gravity_vec(1)/stiffness;
@@ -573,9 +573,9 @@ void testBodyActuator()
                                              blockMassCenter, blockInertia);
 
     // Add display geometry to the block to visualize in the GUI
-    block->attachGeometry(Brick{Vec3{blockSideLength/2,
-                                    blockSideLength/2, 
-                                    blockSideLength/2}});
+    block->attachGeometry(Brick(Vec3(blockSideLength/2,
+                                     blockSideLength/2, 
+                                     blockSideLength/2)));
 
     Vec3 locationInParent(0, blockSideLength / 2, 0), orientationInParent(0), 
         locationInBody(0), orientationInBody(0);
@@ -758,9 +758,9 @@ void testActuatorsCombination()
                                     blockMassCenter, blockInertia);
 
     // Add display geometry to the block to visualize in the GUI
-    block->attachGeometry(Brick{Vec3{blockSideLength/2, 
+    block->attachGeometry(Brick(Vec3(blockSideLength/2, 
                                      blockSideLength/2, 
-                                     blockSideLength/2}});
+                                     blockSideLength/2)));
 
     // Make a FreeJoint from block to ground
     Vec3 locationInParent(0, blockSideLength/2, 0), orientationInParent(0), //locationInParent(0, blockSideLength, 0)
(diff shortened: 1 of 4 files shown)


aymanhab:
Thanks for doing this @klshrinidhi 

My main concern is with the use of the aggregate constructors {} which are spreading quickly, instead of standard constructor when available primarily because these are not available to Matlab (or Python) users, and will look intimidating to anybody who's not up-to-date with C++11 and later (vast majority).  What do you think @chrisdembia ?

chrisdembia:
Last summer, I briefly expressed my preference for `()` over `{}`. I asked @klshrinidhi about this and he said that `{}` won't do implicit conversions (or something like that), and thus might be a safer alternative. I advocate limiting our use of `{}` to internal code, and using `()` in user-facing code/examples.

chrisdembia:
A similar issue is that, in MATLAB, you can construct an object without *any* parenthesis; e.g., `model = Model`. I also think we should avoid that in our examples, and instead use `model = Model()`, etc.

aymanhab:
@klshrinidhi Can you comment on why you preferred value-initialization over standard constructors when both are available. I agree with @chrisdembia that we should keep this for internal code and tests but want to make sure we're in sync. and also that the functionality not-using value-initialization will stay around and be tested. Thanks.

klshrinidhi:
I am happy to close this PR if you don't want to merge it.

I think `{}` is the improved syntax C++11 introduced for initialization. I would refer you to the [wikipedia article](https://en.wikipedia.org/wiki/C%2B%2B11#Uniform_initialization) about it. Later standards are making it more widely applicable. That is the reason I use it and I think we should use widely across OpenSim.

I think we should promote using C++ with modern tools/techniques. I don't agree people will be intimidated by such things when they have decided to write/read C++. It is just matter to of searching on internet to find out what it is and technical people like the ones this domain are good at that. Once they realize it is actually better, they will like too.

On a general note, I don't think we should encourage people to learn C++ through OpenSim. One has to know basics of C++ (up to C++11) before starting. I think popular projects like bootstrap, django, ruby on rails etc assume basic knowledge of the language too. People learn the language and use their software because they are that good.

chrisdembia:
I suppose the reason why we have many novice C++ programmers using OpenSim is (1) biomechanists are farther from programming in their education than are engineers, etc.; (2) the API was mostly in C++ originally. I suppose as we make our python and MATLAB interfaces more capable, we will have fewer novices using the C++ interface. 

> I think we should promote using C++ with modern tools/techniques.

For some reason, I am fine with exposing `auto` and range-for in our example code, but I do not like using the value-initialization. The main reason is maybe that the `auto` and range-for features are fairly clear, while value-initialization is confusing, since braces are used for scopes, etc.

What I don't think is OK is for people to learn *programming* through OpenSim. I think it is fine if they learn C++ through OpenSim. Other than taking a class, how do you learn a programming language other than working with real-world libraries?

klshrinidhi:
If they have decided to use OpenSim in C++, I think they have to learn programming and the language before they start with OpenSim. I would not advice anyone to learn C++ through OpenSim.

C++, Python and Matlab are three different languages with different contructs. I think you are trying to make them look the same. I don't think it works that way. If we want to make Python and Matlab experience better, we have to focus on better documentation and providing much more native syntax/constructs for users in those languages.

I think this PR is getting too out of hand. I am happy to close this and move on.

chrisdembia:
Another option that I'd be happy with: in our examples, we can give a comment in some of the places where we use value-initialization (perhaps the first 2 times or so that we construct an object in an example):

```cpp
// Construct object using C++11 value initialization
Thelen2003Muscle musce{...};
```

This would allow us to use the new features and also teach our users.

Also, I don't think @aymanhab is against this PR; he's only talking about the value initialization. I don't think this PR is getting out of hand; we all need to be on the same page about our approach to our users.

aymanhab:
@klshrinidhi Absolutely I'm not against this PR, it's definitely an improvement, the question was which syntax to use for initialization which is quite minor. 

Regarding C++, the fact is that our users are mostly Matlab users rather than C++ programmers (as evident by most of our surveys and workshops), even the ones that do C++ are most familiar with C syntax. It's a significant limitation but denying it or forcing users to learn C++ is impractical.

aseth1:
@klshrinidhi I also agree with @aymanhab that the initialization syntax was a minor objection compared to the overall improvement introduced.  I am fine with the value initialization, but frankly I don't think it matters. I think consistency matters more- and to have a sprinkling of value initialization in examples could be annoying. Internal code doesn't matter - we should do what is best/up-to-date with the standard.
@klshrinidhi please reopen the PR.

klshrinidhi:
I have changed all the constructor calls to function style syntax.

aymanhab:
Thanks @klshrinidhi Looks great. Feel free to merge if there are no other objections.

Issue #965, opened 2016-05-06
List of classes needing update from Storage to TimeSeriesTable/DataTable


klshrinidhi:
List of classes that need update from `Storage` to `DataTable`. This list is based on classes used in Examples directory.

* `Model`
 * `printControlStorage()`
* `Manager`
 * `getStateStorage()`
* `ForceReporter`
 * `getForceStorage()`
* `SymbolicExpressionReporter` (in Examples)
 * Has a `Storage` data member
* `AnalysisPlugin_Template` (in Examples)
 * Has a `Storage` data member

klshrinidhi:
@aseth1 
Let me know what you think.

aseth1:
Looks like the right list to target. For the top 3 leave the method that returns the storage but `deprecate` them. The `AnalysisPlugin_Template` example we should discuss. I am in favor of removing it altoghter since we no longer want people to write new `Analysis` classes since we now have outputs and reporters.

chrisdembia:
I think we should create an example that exactly replicates (using outputs/reporters) the functionality of creating an Analysis before removing the example. It would still be good to have easy-to-use packaged "studies" that aren't customizable (i.e., "just give me all the kinematics"), and that could still be the Analysis, in which case it may make sense to modernize the AnalysisPlugin examples. If it's not the Analysis, I think maybe we should figure out what it is before removing Analysis example.

For example, what about convenience classes like `KinematicsSomething` that would gather up all coordinate values into one output for reporting?

Anyhow I think the `AnalysisPlugin_Template` does not need to be updated to use DataTable.

chrisdembia:
I think it would be great to have a new API example that has (1) a new Component "calculator" class, that computes "force generation ability", and (2) has a sample main program that adds the new component to a model, combines it with a reporter, and writes the data out to a CSV file.

chrisdembia:
@klshrinidhi can this be closed?

PR #966, opened 2016-05-10
Replace Storage with TimeSeriesTable in examples

klshrinidhi:
Replace usage of `Storage` class with `TimeSeriesTable` in `OpenSim/Examples`. This needed adding/editing methods to other classes.

Changes are as per #965.
Classes derived from `Analysis` are not updated as per comments in #965.

* Added some functions to `DataTable` & `AbstractDataTable`.
* Made the `read` and `write` functions `static` in all the file adapters for ease of calling them.
* Added a function to `Storage` to return a `TimeSeriesTable` equivalent of the `Storage`.
* Added functions to `Model`, `Manager`, `ForceReporter`, `ControllerSet` to get a `TimeSeriesTable` in place of a `Storage`.
* Replaced calls to retrieve and print `Storage` with equivalent calls for `TimeSeriesTable`.


OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp
```diff
@@ -978,6 +978,16 @@ void SimbodyEngine::convertRadiansToDegrees(Storage &rStorage) const
...
     rStorage.setInDegrees(true);
 }
+
+void SimbodyEngine::convertRadiansToDegrees(TimeSeriesTable& table) const {
+    constexpr double PI{3.141592653589793};
+    constexpr double RAD_TO_DEG{180 / PI};
+
+    table.updMatrix() *= RAD_TO_DEG;
+    table.removeTableMetaDataKey("inDegrees");
+    table.addTableMetaData("inDegrees", std::string{"yes"});
```
chrisdembia: Does the equivalent method for Storage (just above) modify *all* columns, or just ones that have radians in the numerator?

OpenSim/Simulation/Model/ControllerSet.h
```diff
@@ -83,6 +84,7 @@ OpenSim_DECLARE_CONCRETE_OBJECT(ControllerSet, ModelComponentSet<Controller>);
     void constructStorage();
     void storeControls( const SimTK::State& s, int step );
     void printControlStorage( const std::string& fileName) const;
+    TimeSeriesTable getControlTable() const;
```
chrisdembia: For consistency with the methods in this class and in Model, this should be `getControlsTable()`.

OpenSim/Simulation/Model/ControllerSet.h
```diff
@@ -25,6 +25,7 @@
 
 
 // INCLUDES
+#include "OpenSim/Common/TimeSeriesTable.h"
```
chrisdembia: Quotes are supposed to be used for paths that are relative to this file, right? Brackets are used for searching the specified include paths?

OpenSim/Common/DataTable.h
```diff
@@ -422,6 +481,39 @@ class DataTable_ : public AbstractDataTable {
...
+    template<typename RowIter>
+    void appendRow(const ETX& indRow, RowIter begin, RowIter end) {
+        RowVector row{static_cast<int>(std::distance(begin, end))};
+        int ind{0};
+        for(auto it = begin; it != end; ++it)
+            row[ind++] = *it;
+
+        appendRow(indRow, row);
+    }
+
```
chrisdembia: These seem nice to have, but I think these should be farther down in the header file, perhaps in an "Advanced" doxygen group (or at least below the other method in this doxygen group). I think it'll be much more common to use the variant below that takes a RowVector.

chrisdembia:
Looks pretty good to me. A few small comments:

1. StatesTrajectory has a method `exportToTable()`; [see here](http://myosin.sourceforge.net/964/classOpenSim_1_1StatesTrajectory.html#a378427d8c4afc0881d6e5ff86530dd23). For consistency, what do you think about renaming the method on Storage to `Storage::exportToTable()`? We are trying to use [`get`](https://github.com/opensim-org/opensim-core/blob/master/CONTRIBUTING.md#functions-and-methods) only for methods that are supposed to return something that already exists, though we don't really follow this convention too well yet. Anyway, I'm fine with `getAsTimeSeriesTable()` if you prefer it.
2. A lot of the output storage files were renamed to use a `.mot` extension, and I think this will be confusing for most users. I think we should leave the extensions as they were. As a separate issue (definitely not this PR), I think it makes more sense to rename `MOTFileAdapter` to `STOFileAdapter` or `StorageFileAdapter`. I don't know if there were discussions about this already. Another option is to create `typedef MOTFileAdapter STOFileAdapter`. MOT files are usually only for a certain type of data (I think joint angles?), not for states or forces.
3. Rename `getControlTable()` to `getControlsTable()`.
4. Reordering of the new iterator-based `appendRow` methods.


A small topic for discussion: I am in the process of adding a method to StatesTrajectory to be able to write them to file. For consistency with other parts of our code (Objects, Storage, etc.) I plan on naming the method `StatesTrajectory::print()`. I think `write()` makes sense for DataTable; perhaps I should use `write()` as well. Just wanted to bring this up; no need to take action about this.

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp
klshrinidhi: I don't think there is an equivalent method in storage. This operation is not modifying any columns. It is just adding metadata to the table.

OpenSim/Simulation/Model/ControllerSet.h
klshrinidhi: Changed the name to `getControlsTable`.

OpenSim/Simulation/Model/ControllerSet.h
klshrinidhi: I don't think so. The one with quotes searches more directories than one with angle brackets.

OpenSim/Common/DataTable.h
klshrinidhi: I think marking these methods as "advanced" will discourage people from using them. 
I found them way more useful than the one that just takes a `RowVector`. It is now possible to append a row with pretty much any sequence of elements without having to create a `RowVector` every time. These methods are the kind found in standard library. I think I have enough error-checking in place to help debug problems.
I can group these `appendRow` overloads and write examples on using them in Doxygen.

klshrinidhi:
> StatesTrajectory has a method exportToTable(); see here. For consistency, what do you think about renaming the method on Storage to Storage::exportToTable()? We are trying to use get only for methods that are supposed to return something that already exists, though we don't really follow this convention too well yet. Anyway, I'm fine with getAsTimeSeriesTable() if you prefer it.

Currently the name is consistent with ones used in `SimTK::Matrix` hierarchy. To me `export` sounds like a file operation of exporting to a different format.

> A lot of the output storage files were renamed to use a .mot extension, and I think this will be confusing for most users. I think we should leave the extensions as they were. As a separate issue (definitely not this PR), I think it makes more sense to rename MOTFileAdapter to STOFileAdapter or StorageFileAdapter. I don't know if there were discussions about this already. Another option is to create typedef MOTFileAdapter STOFileAdapter. MOT files are usually only for a certain type of data (I think joint angles?), not for states or forces.

Because `Storage` is being replaced, I don't think people will find it as confusing if files are `mot` instead of `sto`.

Also, the formats of STO and MOT files are pretty much same. I don't think MOT/STO files have any structure (like TRC) associated with them. Their header section is unstructured and anything can go there. So I don't think MOT files are limited to a particular type of data.

OpenSim/Common/DataTable.h
klshrinidhi: I have added code examples with comments in Doxygen.

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp
aseth1: I believe @chrisdembia is referring to `convertRadiansToDegrees(Storage &)` vs this implementation for the table. This method is incorrectly converting the entire table to degrees (and is using a different constant - should also use `SimTK_RADIAN_TO_DEGREE` here. `SimbodyEngine::scaleRotationalDofColumns()` only converts columns that are associated with rotational coordinates.

OpenSim/Common/DataTable.h
chrisdembia: Okay cool.

chrisdembia:
> Currently the name is consistent with ones used in SimTK::Matrix hierarchy. To me export sounds like a file operation of exporting to a different format.

Okay sounds good. Perhaps I will change the name of the method in StatesTrajectory.

> Because Storage is being replaced, I don't think people will find it as confusing if files are mot instead of sto.

The Storage class is being replaced; I don't know if the Storage file format is being replaced.



> So I don't think MOT files are limited to a particular type of data.

I know that STO and MOT are nearly the exact same format; I mean that we have different conventions for the data that MOT/STO files contain. In particular, I think "MOT" stands for "motion," and so it seems weird to me to have forces in a "motion" file.

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp
chrisdembia: Line 986 is modifying all the columns; I should have put the comment there.

klshrinidhi:
> I know that STO and MOT are nearly the exact same format; I mean that we have different conventions for the data that MOT/STO files contain. In particular, I think "MOT" stands for "motion," and so it seems weird to me to have forces in a "motion" file.

Got it. I am not sure. We can talk about this in next meeting.

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp
klshrinidhi: I fixed it.
* Added an overload of `SimbodyEngine::scaleRotationalDofColumns` for `TimeSeriesTable` that does the same thing as the other overload.
* Now using the correct scale factor.

I think I got away with this previously because all the columns of the table were rotational dof and so comparison tests passed.

klshrinidhi:
Travis timing out on one build. Everything else builds/tests fine.

aymanhab:
@klshrinidhi We should try to build the GUI against this branch to make sure nothing broke since the GUI uses Storage for a variety of purposes (tools, motions/results, experimental data, plotting etc.) This testing builds on the work that we did earlier to integrate the GUI and API. Please let me know if you need help trying it out. Thanks.

klshrinidhi:
This PR does not delete any existing functions or rename them. This PR only adds new functions. So I don't think it will break anything. It may still be good to try building GUI.

klshrinidhi:
I get the following error from CMake when building the `Visualizer` branch.
```
CMake Error at CMakeLists.txt:530 (INCLUDE):
   include could not find load file:

     ApiDoxygen.cmake

 CMake Error at CMakeLists.txt:592 (SUBDIRS):
   subdirs Incorrect SUBDIRS command.  Directory: Models does not exist.

 CMake Error at Vendors/CMakeLists.txt:2 (SUBDIRS):
   subdirs Incorrect SUBDIRS command.  Directory: lepton does not exist.
```

aymanhab:
@klshrinidhi The top level CMakeLists refers to the one under Gui/opensim neither have reference to Doxygen, Models or Lepton. Plase open an issue on the gui repo to make sure we're in sync. and do whatever needed cleanup there. Thanks.

aseth1:
> The Storage class is being replaced; I don't know if the Storage file format is being replaced.

  >  > So I don't think MOT files are limited to a particular type of data.

>I know that STO and MOT are nearly the exact same format; I mean that we have different conventions for the data that MOT/STO files contain. In particular, I think "MOT" stands for "motion," and so it seems weird to me to have forces in a "motion" file.

If anything I would like to us stop the native use `.mot` which is a SIMM defined format for its motion files. We adopted it to be "compatible" with SIMM. I think we can stick with ".sto" as OpenSim's stock output format since the bulk of files generated are all `.sto` - we might as well "own" it and update it as necessary.

It would be good if OpenSim could in fact read in genuine `.mot` files from SIMM (I am not sure it does anymore). I prefer us to *import/export* `.mot` files even if the  `.sto` is very similar. Separating the two formats will give us more freedom to evolve the `.sto` format if necessary. I never liked `Storage` treating the two as if they were the same format.

klshrinidhi:
> If anything I would like to us stop the native use .mot which is a SIMM defined format for its motion files. We adopted it to be "compatible" with SIMM. I think we can stick with ".sto" as OpenSim's stock output format since the bulk of files generated are all .sto - we might as well "own" it and update it as necessary.
>
> It would be good if OpenSim could in fact read in genuine .mot files from SIMM (I am not sure it does anymore). I prefer us to import/export .mot files even if the .sto is very similar. Separating the two formats will give us more freedom to evolve the .sto format if necessary. I never liked Storage treating the two as if they were the same format.

Sounds good. We can have one adapter for each -- 
* `MOTFileAdapter` -- already exists but not sure if it can read genuine MOT files but I can make it read them if I know the format.
* `STOFileAdapter` -- OpenSim's home cooked format. I can add this one in another PR.

Can we merge this now ?

chrisdembia:
> We can have one adapter for each --

+1

> Can we merge this now ?

Do you mean that you would change the file extensions from `.mot` back to `.sto` in that subsequent PR?

klshrinidhi:
> Do you mean that you would change the file extensions from .mot back to .sto in that subsequent PR?

Correct. This PR is not about adding another adapter. It is about replacing `Storage`. Adding `STOFileAdapter` can be subject of my next PR. I can change format of files in these examples from *mot* to *sto* in that PR.

klshrinidhi:
I still have to build the GUI with this branch. Forgot about that. Will get back soon.

klshrinidhi:
:fireworks: I am able to build GUI successfully against this branch.

Can we merge this now ? :panda_face:

aymanhab:
Great, thanks @klshrinidhi Looks good to me.

klshrinidhi:
I am going to merge this if there are no more comments.
I will start another PR to add STOFileAdapter and change the extension of files in the examples to reduce confusion for users.

aseth1:
Let me review ...

OpenSim/Examples/ControllerExample/ControllerExample.cpp
```diff
@@ -317,8 +318,11 @@ int main()
         manager.integrate( si );
 
         // Save the simulation results.
-        osimModel.printControlStorage( "tugOfWar_controls.sto" );
-        manager.getStateStorage().print( "tugOfWar_states.sto" );
+        auto controlsTable = osimModel.getControlsTable();
+        MOTFileAdapter::write(controlsTable, "tugOfWar_controls.mot");
```
aseth1: I would like us to avoid using `.mot` as our "output" format. We have total control over `.sto` but `.mot` was originally defined and used by SIMM (motion file) and if we want to maintain some compatibility with SIMM we shouldn't use it as our own format. Can you revert these to `.sto`?  We can make `.sto` stand  for standard table opensim/output or whatever.

I also don't mind if a `STOFileAdapter` is just an alias of an `MOTFileAdapter` for the time being. I don't think it will be long before these two become more distinct.

aseth1: To summarize, I would like us to:
 1. revert changes to `.mot` back to `.sto`
 2. create a stand in for STOFileAdpater (e.g. aliasing MOTFileAdapter is good enough). 
 
I am happy for you to create the separate STO/MOT adapters as a future PR. :crystal_ball:

klshrinidhi: Currently the files are being written by MOTFileAdapter. I think it will incorrect to change the extension to `.sto`. My suggestion is to have another PR to introduce STOFileAdapter and change these file extensions there. Just creating a type alias STOFileAdapter will not make sense since we have to introduce that adapter as a separate class anyway.

If I can get the precise format of MOT files, I can start an another PR pretty quickly.

aseth1: I think it is confusing and unnecessary to permit the rename simply to revert in subsequent PR. Why not revert to `.mot` now and then rename `MOTFileAdpater` to `STOFileAdapter` since we are certain it handles `.sto` files. The correctness of the `MOTFileAdapter` to read SIMM generated motion files is of relatively low priority, so I don't want you to go down that route at present. I am OK with using our `STOFileAdapter` to read in our version of `.mot` files.

klshrinidhi: Okay. Will push the changes soon. :+1:

klshrinidhi:
* MOTFileAdapter -->> STOFileAdapter
* File extensions are back to `.sto`.

OpenSim/Examples/ControllerExample/OutputReference/ControllerExample_Solution.cpp
```diff
@@ -285,8 +286,11 @@ int main()
         manager.integrate( si );
 
         // Save the simulation results.
-        osimModel.printControlStorage( "tugOfWar_controls.sto" );
-        manager.getStateStorage().print( "tugOfWar_states.sto" );
+        auto controlsTable = osimModel.getControlsTable();
+        MOTFileAdapter::write(controlsTable, "tugOfWar_controls.mot");
```
aseth1: Missed these

OpenSim/Examples/CustomActuatorExample/OutputReference/toyLeg_example.cpp
```diff
@@ -216,14 +216,15 @@ int main()
         manager.integrate(si);
 
         // Save results
-        osimModel.printControlStorage("SpringActuatedLeg_controls.sto");
-        Storage statesDegrees(manager.getStateStorage());
-        osimModel.updSimbodyEngine().convertRadiansToDegrees(statesDegrees);
-        //statesDegrees.print("PistonActuatedLeg_states_degrees.mot");
-        statesDegrees.print("SpringActuatedLeg_states_degrees.mot");
+        auto controlsTable = osimModel.getControlsTable();
+        MOTFileAdapter::write(controlsTable, "SpringActuatedLeg_controls.mot");
```
aseth1: And here

OpenSim/Examples/OptimizationExample_Arm26/OutputReference/OptimizationExample.cpp
```diff
@@ -93,12 +94,15 @@ class ExampleOptimizationSystem : public OptimizerSystem {
         
         // Store and print the  results of the first step.
         if( stepCount == 1){ 
-            manager.getStateStorage().print("Arm26_noActivation_states.sto");
+            auto statesTable = manager.getStatesTable();
+            MOTFileAdapter::write(statesTable, "Arm26_noActivation_states.mot");
```
aseth1: here

aseth1:
Looks good.

PR #967, opened 2016-05-12
Update WrapObject to a ModelComponent

aseth1:
Changes mostly due to update to new properties and conforming to `Component` interface:
 - converted old properties to new 
 - removed custom copy constructor and assignment operator, which are no longer necessary 
 - use `Connector<PhysicalFrame>` to replace dependency on Body (used to be naked pointer)
 - implement `extendFinalizeFromProperties()` for converting properties to enum or other values (e.g. `wrapAxis`, `wrapSign`, etc...)
 - update count of components in `testIterators`
 - some cleanup of unnecessary includes


OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp
```diff
@@ -82,7 +82,9 @@ int main()
             try {
                 testModel.addModelComponent(ModelComponent::safeDownCast(randClone));
             } //Ignore the validity of the property values
-            catch (const SimTK::Exception::ErrorCheck&) {
+            // TODO this should specifically handle "InvalidPropertyValue" exceptions
+            // once we have that in place.
+            catch (const std::exception&) {
```
chrisdembia: What was the new exception that was getting thrown?

OpenSim/Simulation/Wrap/WrapObject.h
```diff
@@ -82,89 +88,83 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(WrapObject, Object);
...
-    //void setGeometryQuadrants(AnalyticGeometry *aGeometry) const;
+   /** Determine the appropriate values of _quadrant, _wrapAxis, and _wrapSign,
+     * based on the name of the quadrant. finalizeFromProperties() should be
+     * called whenever the quadrant property changes. */
+    void extendFinalizeFromProperties();
+
 private:
-    void setNull();
     void constructProperties();
+    void constructConnectors();
```
chrisdembia: Add the `override` to reduce clang warnings.

OpenSim/Simulation/Wrap/WrapObject.h
```diff
@@ -56,13 +44,31 @@ class PhysicalFrame;
...
+        "Translation of the WrapObject.");
+
+    OpenSim_DECLARE_PROPERTY(display_preference, int,
+        "Display Pref. 0:Hide 1:Wire 3:Flat 4:Shaded");
+
+    OpenSim_DECLARE_LIST_PROPERTY_SIZE(color, double, 3,
+        "Display Color");
+
+    OpenSim_DECLARE_PROPERTY(quadrant, std::string,
+        "The quadrant over which the wrap object is active.");
```
chrisdembia: Can you give examples of valid values in the documentation string?

OpenSim/Simulation/Wrap/WrapObject.h
```diff
@@ -56,13 +44,31 @@ class PhysicalFrame;
...
+        "Body-fixed Euler angle sequence for the orientation of the WrapObject");
+
+    OpenSim_DECLARE_PROPERTY(translation, SimTK::Vec3,
+        "Translation of the WrapObject.");
+
+    OpenSim_DECLARE_PROPERTY(display_preference, int,
+        "Display Pref. 0:Hide 1:Wire 3:Flat 4:Shaded");
+
+    OpenSim_DECLARE_LIST_PROPERTY_SIZE(color, double, 3,
+        "Display Color");
```
chrisdembia: `[0, 1]` or `[0, 255]`? RGB?

chrisdembia:
It is not clear to me that the WrapObject should use a connector to connect to a PhysicalFrame. Is it the case that the WrapObject can only connect to the PhysicalFrame that contains it? If so, then creating the connector will allow users to mess up the model by specifying some other frame as the connectee. I suppose there is a similar problem with Geometry. Why not just use the `getParent()` pointer?

chrisdembia:
Note: There's a similar issue with the MetabolicsCalculator I'm working on. It seems like we have a common situation where we know a component is a subcomponent of another component, and we need a pointer to that parent component (WrapObject, MetabolicsWhatever, Geometry, and also probably with HuntCrossleyForce and ElasticFoundationForce).

chrisdembia:
Maybe there should be some sort of `ParentConnector` that is not serialized, or something like that.

OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp
aseth1: testSerializeOpenSimObjects fills all string type properties with a generic "ABCXYZ" string which is invalid for a quadrant name. Then was "activated" because I know resolve the quadrant at `finalizeFromProperties` instead of side-effect of `connect()`

chrisdembia: OK thanks for explaining.

aymanhab:
Main issue with GUI build so far is the lack of wrapObject.getBody(), I'll try getFrame() instead and follow up with other downstream effects if any. How are wrap objects supposed to display themselves?

aseth1:
> Maybe there should be some sort of ParentConnector that is not serialized, or something like that.

@chrisdembia thanks for pointing this out. I agree that the `Connector` in this case would be misleading since users could not effectively change the `PhysicalFrame` it is bound to. I had thought that this restriction was unnecessary and we could eliminate the `WrapObjectSet` on `PhysicalFrame`, but that would require changes to the wrapping infrastructure- so I reverted it back to a `ReferencePtr` that the owner can set (which is how it was anyways).

I don't think there is a need for a `ParentConnector` type. It seems reasonable for the parent component to run through its list components and satisfy unspecified connections. For example, Geometry added to a Frame (which owns and is the the parent of Geometry now) could set unsatisfied connections of its geometry to itself- which would be the correct default behavior. Alternatively, `Geometry` could have no `Connector` (like WrapObject) and must be added to a Frame to be displayed.

chrisdembia:
Okay great. LGTM.

aseth1:
>Main issue with GUI build so far is the lack of wrapObject.getBody(), I'll try getFrame() instead and follow up with other downstream effects if any. 

`getFrame()` should behave identically to `getBody()

> How are wrap objects supposed to display themselves?

Good question @aymanhab. They should implement `generateDecorations`. Currently, this appears to be delegated to `DefaultGeometry` defined in `ModelVisualizer.h`. I think it would great if you can move the implementations for each WrapObject type into its respective Component as a follow up PR.

aseth1:
@aymanhab I verified that the `ModelVisualizer` was not broken and still displays wrapping and wrap points as before. I agree that this functionality should be moved to and appropriately implemented in `generateDecorations()` of the respective `WrapObject` components. Want to give that a try after merging?

aymanhab:
@aseth1 Yes, please merge this and let me know when API changes in Contact and muscle wrapping are done and I will look into how to make them use Geometry classes.

PR #970, opened 2016-05-12
Add version mapping for MATLAB R2016a

sebastianskejoe:


klshrinidhi:
Thanks for the PR.
Looks good to me. 
@chrisdembia take a look.

chrisdembia:
Awesome! Thank you, @sebastianskejoe . It seems that the FindMatlab.cmake script anticipates this scenario, and that it suggests instead to define

```cmake
set(MATLAB_ADDITIONAL_VERSIONS
         "R2016a=9.0")
```
.

Could you add the above line in the root CMakeLists.txt at [this line](https://github.com/opensim-org/opensim-core/blob/master/CMakeLists.txt#L374)? 

Also, have you been able to successfully run the OpenSim MATLAB tests?

sebastianskejoe:
@chrisdembia The way I understand [this](https://github.com/opensim-org/opensim-core/blob/master/cmake/FindMatlab.cmake#L247) comment MATLAB_ADDITIONAL_VERSIONS should be used when the **user** is using a not-supported MATLAB version, while MATLAB_VERSIONS_MAPPING contains the mapping for supported versions. I don't know if R2016a is supported yet, but if it is it should be added to MATLAB_VERSIONS_MAPPING IMHO. If it is not, I'm not sure it should be allowed to be built by default, but instead users can be hinted at setting MATLAB_ADDITIONAL_VERSIONS, if they want to try out other versions than the supported ones. However, I might be wrong.

With regards to the test, I have not been able to run it successfully. It fails with the following error: 
```
Required library failed to load. Check that the dynamic library osimJavaJNI is in your PATH
java.lang.UnsatisfiedLinkError: C:\OpenSim-get\bin\osimJavaJNI.dll: The specified procedure could not be found
```
and the output file says:
```
An exception has been thrown during the execution
  JavaException with properties:

    ExceptionObject: [1x1 java.lang.UnsatisfiedLinkError]
         identifier: 'MATLAB:Java:GenericException'
            message: 'Java exception occurred: ...'
              cause: {}
              stack: [1x1 struct]

    file: 'C:\Users\au323720\Desktop\OpenSim-git\Bindings\Java\Matlab\examples\OpenSimCreateTugOfWarModel.m'
    name: 'OpenSimCreateTugOfWarModel'
    line: 37
```
If I run OpenSimCreateTugOfWarModel.m in MATLAB I get the first error along with this in the console:
```
Error using OpenSimCreateTugOfWarModel (line 37)
Java exception occurred:
java.lang.UnsatisfiedLinkError: org.opensim.modeling.opensimModelSimulationJNI.new_Model__SWIG_0()J
	at org.opensim.modeling.opensimModelSimulationJNI.new_Model__SWIG_0(Native Method)
	at org.opensim.modeling.Model.<init>(Model.java:814)
```

I suspect that this is more of a setup issue, but I'm not sure and I don't know how to fix it.

aymanhab:
Thanks @sebastianskejoe The error is due to the osimJavaJNI.dll either not in your path or an incorrect version is found by Matlab. Can you fix that and retry? Thanks.

sebastianskejoe:
Ah yes, now it works and the test runs successfully. I only checked the MATLAB path, and not the system path before. Turns out OpenSim3.3 was before the the git version on the system path and therefore the old osimJavaJNI was used. Thanks @aymanhab!

aymanhab:
Thanks, looks good to me @sebastianskejoe 

Feel free to merge @chrisdembia

chrisdembia:
Thanks for clarifying, @sebastianskejoe, and also for testing that 2016a is indeed supported. The build failure is unrelated; I am merging!

PR #972, opened 2016-05-13
Add LoadOpenSimLibraryExact()

chrisdembia:
This PR adds a method that loads an OpenSim library using an exact path. The pre-existing LoadOpenSimLibrary methods expect the user *not* to provide a file extension for the library file (except on OSX, which was not handled well and so one still had to provide an extension). The previous methods attempt to detect whether or not you want a debug or release variant of the library, which is neat.

However, I think there is a good reason to use full paths (including the extension) instead of trying to add the extension automatically:

1. It is more natural and less confusing for users to provide a full path to a file. Usually, one specifies the name of a library separately from its path (e.g., LD_LIBRARY_PATH, DYLD_LIBRARY_PATH, PATH), and specifying the name of a library), and so our previous solution of combining the path to the library with the library name is some middle-ground which may be confusing to users. On UNIX, we also automatically insert a "lib" prefix to the library name, which may also be confusing for users.
2. It is actually really hard to add the correct file extension automatically. This is especially true on UNIX, where the SONAME is an additional suffix (e.g., `libFoo.so.1.0.0`; I don't know how we'd figure out to correctly add the `.1.0.0`).

I also considered just having the current methods detect if a full path (with extension) was provided, and to then just load the library at that full path. However, I decided it would be somewhat hard and error-prone to detect if a full path with extension were given (because of (2) above), and so I think it makes more sense to have a separate method that is clear about what it does.

I left in the old methods because I think it is still neat and potentially useful that they try to automatically try to use the correct release/debug variant (and for backwards compatibility).

This PR also fixes the old LoadOpenSimLibrary methods to handle dylib files (OSX).

This PR is in preparation for the upcoming command line tool PR. I have tested the changes in this PR through the command line tool branch.

### DIFF
--- a/OpenSim/Common/LoadOpenSimLibrary.cpp
+++ b/OpenSim/Common/LoadOpenSimLibrary.cpp
@@ -30,30 +30,38 @@ using namespace OpenSim;
 using namespace std;
 
 #ifndef _WIN32
-// Current solution for Linux compatibility is to remap LoadLibrary/GetProcAddress to dlopen/dlsym
-// using macros.  Also use macros for portable handles.
-//
-// LoadLibrary used to be a macro for dlopen but we want to transparently support
-// adding the "lib" prefix to library names when loading them, so made it a function
+// For Linux compatibility is to remap
+// LoadLibrary/GetProcAddress to dlopen/dlsym.
+static void *LoadLibrary(const char* name) {
+    return dlopen(name, RTLD_LAZY | RTLD_GLOBAL);
+}
+
+// We want to transparently support adding the "lib" prefix to library names
+// when loading them.
 static void *LoadLibrary(const std::string &name, std::string &actualNameLoaded) {
-    void *lib = dlopen(name.c_str(), RTLD_LAZY | RTLD_GLOBAL);
+    actualNameLoaded = name;
+    void *lib = LoadLibrary(name.c_str());
     if(!lib) {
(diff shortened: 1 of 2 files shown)


chrisdembia:
@aseth1 says implement `LoadOpenSimLibrary` to use `LoadOpenSimLibraryExact`.

chrisdembia:
@aseth1 I looked at this PR again.

1. To actually load a library, both `LoadOpenSimLibrary()` and `LoadOpenSimLibraryExact()` ultimately cal the same method, `LoadLibrary()`, and I think this achieves the code reuse that you want. It is helpful to have this extra level of indirection, as `LoadOpenSimLibrary()` must call out to an intermediate method in order to try appending `lib` to the library name on UNIX.
2. I tested `LoadOpenSimLibrary()` locally and it still works fine.
3. Simbody's Pathname has a `getAbsolutePathname()` that also canonicalizes the path and thus would convert slashes. But it also has a lot of other logic that I would like to avoid for `LoadOpenSimLibraryExact()`; mainly because we wouldn't really be using the exact path provided and would become susceptible to any potential (though unlikely) bugs in Simbody's Pathname code.

chrisdembia:
@aseth1 could you review this again? Once this gets in, I can open the PR about the command-line tool.

aseth1:
LGTM

chrisdembia:
Thank you!

PR #980, opened 2016-05-18
Cache build directories on Travis

klshrinidhi:
This PR is only applicable to CI build on Travis.

This PR improves CI build time by:
* Caching the build directory of dependencies and OpenSim.
* Building only the changed files for each branch.
* Running only relevant tests for each branch.

Following is the procedure --
* *Superbuild* dependencies. If the version of the dependency required is:
 * *cached*, download it and use it as build directory.
 * *not cached*, proceed to build sources as usual.
* If above build is not *cached*, upload it the build to cache.
* Build OpenSim.
 * If building *master* OR if commit comment contains *[make clean]*, do not use cached build. Build sources as usual.
 * If building a *branch* (that is not master), find the base of this branch in master, i.e. the commit in master that this branch started with. If build is: 
   * *cached* for that commit, download and use it as build directory. In this case, only the sources that this branch touched will be built, saving build time as a result.
   * * not cached* for that commit, proceed to build sources as usual.
* If the above build was for *master*, cache the build if it is not already on cache. We only cache builds of *master* branch.
* Test OpenSim -- find the tests that have not been touched by the build above and exclude them, saving time as a result.

Cache is versioned using the tuple:
* Machine-type -- `x86_64-pc-linux-gnu` or `x86_64-apple-darwin13_clang_RelWithDebInfo`.
* Compiler -- `gcc` or `clang`.
* Build-type -- `RelWithDebInfo` or `Debug`.
* Git commit-id -- eg `7649794a63d23fb8b1d5ee041ae5c2ff350a48a3`

Cache is maintained on Bintray under OpenSim organization -- https://bintray.com/opensim. This cache is not meant for public use. I think it is a matter of time before we run out of space in Bintray. I don't know how much space we have right now. If we decide to use this, we could use some space/time on simtk server potentially.

Here are a few tests to see this working:
* https://travis-ci.org/opensim-org/opensim-core/builds/130981934 <br/>
https://github.com/opensim-org/opensim-core/compare/cache_build_dirs_travis <br/>
This branch has no changes to any sources. The changes are only to files that are not build.
* https://travis-ci.org/opensim-org/opensim-core/builds/130987027 <br/>
https://github.com/opensim-org/opensim-core/compare/cache_test_common <br/>
This branch changes one file in Common.
* https://travis-ci.org/opensim-org/opensim-core/builds/130991263 <br/>
https://github.com/opensim-org/opensim-core/compare/cache_test_simulation <br/>
This branch changes one file in Simulation.
* https://travis-ci.org/opensim-org/opensim-core/builds/130991985 <br/>
https://github.com/opensim-org/opensim-core/compare/cache_test_analyses <br/>
This branch changes one file in Analyses.
* https://travis-ci.org/opensim-org/opensim-core/builds/130992279 <br/>
https://github.com/opensim-org/opensim-core/compare/cache_test_actuators <br/>
This branch changes one file in Actuators.
* https://travis-ci.org/opensim-org/opensim-core/builds/130992666 <br/>
https://github.com/opensim-org/opensim-core/compare/cache_test_tools <br/>
This branch changes one file in Tools.
* https://travis-ci.org/opensim-org/opensim-core/builds/130993607 <br/>
https://github.com/opensim-org/opensim-core/compare/cache_test_tests <br/>
This branch changes few tests.


chrisdembia:
Wow the `travis-ci/push` builds all took less than 10 minutes!!!

chrisdembia:
I don't think I have the mental capacity to do a thorough review but I do think we should try this out! 

I like that you created the `[make clean]` flag. Could you add a note to CONTRIBUTING.md near the [note about `[ci valgrind]`](https://github.com/opensim-org/opensim-core/blob/master/CONTRIBUTING.md#checking-for-memory-leaks-through-github)? Also, for consistency, could we rename it to `[ci clean]` or `[ci make clean]`?

1. Are you *sure* we can't leverage travis' caching mechanism? Does the travis caching mechanism inherit the cache from `master` when you create a new branch?
2. I think it would be nice if all of the caching were toggleable with one change to the travis.yml file (e.g, define an env var `USE_CACHE`). This is the situation I want to avoid: it's 5 days before the developer's workshop and we find a bug in the mechanism. It would be unreasonable in this situation to tell everyone to keep putting `[make clean]` in their commit messages (also, `[make clean]` doesn't prevent all of the new mechanisms from running). You have already done a good job of isolating your changes, but I still think they could be further isolated. Especially since we may want to only temporarily disable the cache and re-enable it in the future (that is, I hope that reverting the history isn't the only way to toggle).
3. Could you put all of the shell scripts in a subdirectory; perhaps `.travis_scripts` or `.build_scripts` (so it's hidden)? I was also going to suggest `.ci_scripts` but I think that's too cryptic. I think it is nice for the root of the repo to have few files (one reason is so that the README is not buried on the github site).

Other than those points, I think we should just try this out.

klshrinidhi:
>I like that you created the [make clean] flag. Could you add a note to CONTRIBUTING.md near the note about [ci valgrind]? Also, for consistency, could we rename it to [ci clean] or [ci make clean]?

Changed `[make clean]` to `[ci make clean]`.

>Are you sure we can't leverage travis' caching mechanism? Does the travis caching mechanism inherit the cache from master when you create a new branch?

*ccache* seems to work really well for *gcc*. I did not see it working for *clang* with exactly same settings. I did not see anything wrong with the way *ccache* is currently setup for *clang*. There may be space limitations coming to play here or maybe just that *ccahe* is not that good with *clang*. 

> I think it would be nice if all of the caching were toggleable with one change to the travis.yml file (e.g, define an env var USE_CACHE). This is the situation I want to avoid: it's 5 days before the developer's workshop and we find a bug in the mechanism. It would be unreasonable in this situation to tell everyone to keep putting [make clean] in their commit messages (also, [make clean] doesn't prevent all of the new mechanisms from running). You have already done a good job of isolating your changes, but I still think they could be further isolated. Especially since we may want to only temporarily disable the cache and re-enable it in the future (that is, I hope that reverting the history isn't the only way to toggle).

I just added a global variable `USE_CACHE` to control everything. Setting this variable to `false` will disable download and upload of cache completely.

> Could you put all of the shell scripts in a subdirectory; perhaps .travis_scripts or .build_scripts (so it's hidden)? I was also going to suggest .ci_scripts but I think that's too cryptic. I think it is nice for the root of the repo to have few files (one reason is so that the README is not buried on the github site).

All the scripts are under `travis_scripts` right now. It is not hidden as you point out. Once I do the same caching for Appveyor, I think it will be good to make a directory named `ci_scripts` (as you suggest) and have directories named `travis` and `appveyor` under it.  `ci_scripts` need not be hidden. What do you think ?

chrisdembia:
> ccache seems to work really well for gcc. I did not see it working for clang with exactly same settings. I did not see anything wrong with the way ccache is currently setup for clang. There may be space limitations coming to play here or maybe just that ccahe is not that good with clang.

Sorry, I don't mean to use `ccache`. I mean to use Travis-CI's caching mechanism, which is not specific to ccache. Travis-CI can cache anything for you; you just list the directories that you want to have cached: https://docs.travis-ci.com/user/caching#Arbitrary-directories

> All the scripts are under travis_scripts right now.

Oh d'oh. Sorry I did not notice. I think this is fine. The directory does not need to be hidden. 


Feel free to merge once the tests pass.

klshrinidhi:
> Sorry, I don't mean to use ccache. I mean to use Travis-CI's caching mechanism, which is not specific to ccache. Travis-CI can cache anything for you; you just list the directories that you want to have cached: https://docs.travis-ci.com/user/caching#Arbitrary-directories

I saw this while working on this PR. They have size limitations and we may hit the limit there too if we want to stash multiple versions of build directory. The good thing about this PR is that if we want to use a different server to store the cache later (because Bintray refused to do so for us), then we can switch easily.

chrisdembia:
Good point about the modularity. Yeah I think we should merge this PR using bintray to store the cache for now, and we can try using travis if we run into problems.

Also, travis' stated size limitation is not enforced I think. I think we are already violating it.

PR #981, opened 2016-05-18
Updated prescribed controller to use ModelComponent API

aymanhab:
Removed constructors and data members that don't belong, added outputs, modified test case accordingly.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/981)

### DIFF
--- a/Applications/Analyze/test/testInducedAccelerations.cpp
+++ b/Applications/Analyze/test/testInducedAccelerations.cpp
@@ -25,7 +25,7 @@
 #include <OpenSim/Common/Constant.h>
 #include <OpenSim/Simulation/Model/Model.h>
 #include <OpenSim/Simulation/Model/BodySet.h>
-#include <OpenSim/Simulation/Control/PrescribedController.h>
+#include <OpenSim/Simulation/Control/FunctionBasedController.h>
 #include <OpenSim/Tools/AnalyzeTool.h>
 #include <OpenSim/Auxiliary/auxiliaryTestFunctions.h>
 #include <OpenSim/Analyses/InducedAccelerationsSolver.h>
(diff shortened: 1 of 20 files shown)


aseth1:
@aymanhab I agree that the double duty of the PrescribedController between a *FunctionBased* and *FileBased* controller was confusing/ambiguous and it is worthwhile to separate. 

I don't think it is a good idea, however, to eliminate tests and use cases. There is now a vacuum where we cannot obtain a Controller from controls that are in a file. We introduced the file capability to `PrescribedController` to get away from the `ControlSetController` and I don't want users to have to resort to using it again.  My guess is that majority of uses of the `PrescribedController` are for reading controls in a storage file (from CMC or from MATLAB) in order to run a forward simulation.

We must bring back the *FileBased* capability in some form. Some choices are:
 1. Use the  `TableSource` to generate Ouputs that you plug into to corresponding Actuator Inputs 
 2. Add a `FileBasedController ` with a controls file property
 3. Conversely restore `PrescribedController` to be the file-based controller and create a separate `FunctionBasedController`.

In any case we should restore the PrescribedController test cases so that we have to handle previous uses of `PrecribedController` and replace old/nonexistent functionality with the appropriate (new) Component(s) with a suitable `updateFromXMLNode` implementation.

aymanhab:
@aseth1 I agree that we should restore the functionality of the old PrescribedController that read from file (before 4.0). The easiest way would be to rename the new Controller in this PR to FunctionBasedController, restore the old one so users don't see a hole in functionality and when a FileBasedController becomes available (that depends on DataTable/Source/...) we can offer it as alternative.

chrisdembia:
That solution sounds like it would be clear to API users.

jenhicks:
I agree with the proposed solution to keep PrescribedController. @chrisdembia did you review the code?

chrisdembia:
No I haven't reviewed this.

aymanhab:
Branch is pre devweek so 700 commits behind. I think you should wait @chrisdembia until I update it first and run against latest API/tests.

jenhicks:
Sounds good. One piece of feedback: I might update the documentation for PrescribedController to point out that it should only be used for control inputs from files. The FunctionBasedController should be used otherwise.

aymanhab:
This was addressed as a part of major refactoring of Controllers in 2024.

PR #982, opened 2016-05-19
Fix repeated StateVariable lookups by name

aseth1:
During integration, the `Manager` accesses the model's state variables by name and writes them to Storage(Table) with names as column labels. Accessing each `StateVariable` by name within `getStateVariableValues()` (on each integration step) was accounting for 46% of CPU time according to VS2015's profiler (sample based).

This PR has `Component` hold on to a flat (private) list of `StateVariables` in the order of `getStateVariableNames()` , whichcreated (once) the first time the `StateVariable` values are requested.

Several (4) conditions, including that the component is up-to-date with its properties and that the underlying `System` has not changed are used to determine the validity of list and rebuilds the list if the conditions are not met.

Profiling again shows that `getStateVariableValues()` now accounts for 17% (16% just from `findStateVariable()`) of CPU time.


OpenSim/Common/Component.cpp
```diff
@@ -865,11 +880,29 @@ SimTK::Vector Component::
...
+    // 1. Component is up-to-date with its Properties
+    // 2. a System has been associated with the list of StateVariables
+    // 3. The list of all StateVariables is correctly sized (initialized)
+    // 4. The System associated with the StateVariables is the current System
+    bool valid = isObjectUpToDateWithProperties() &&   // 1.
+                _statesAssociatedSystem.empty()  &&     // 2.
+                _allStateVariables.size() == nsv &&     // 3.
+                getSystem().isSameSystem(_statesAssociatedSystem.getRef());// 4.
+    // if the StateVariables are invalid (see above) rebuild the list 
+    if (!valid) {
```
tkuchida: Does it make any sense to force `!valid` and then confirm that the new `_allStateVariables` is as expected (e.g., in Debug mode or a regression test)? My concern is that, somewhere down the line, the four conditions above may be insufficient. :crystal_ball:

chrisdembia: I think a regression test is a good idea. I also think it makes sense to break this check out into another method, so it can be used in other similar caching scenarios.

aseth1: 1. Do you want to see a test for each condition to verify the logic is correct? What tests do you have in mind? The existing tests (e.g. testForward, testCMC, testMuscle, and any other case involving the Manager at the least) form the regression to verify the results are the same as before. The StateVariables list is rebuilt many times during these test cases. Are you asking to test the counter conditions specifically as part of additional unit tests?
  
2. I don't think breaking this validity check out as its own method is a good idea. It is specific to this method since this and only this method should be dealing with the private StatesVariables list. A more generic consistency check might be useful downstream, but I think we should hold off until we need it.

tkuchida: >What tests do you have in mind?

Perhaps this idea isn't applicable here, but my intuition says that if a shortcut is being taken based on some checks, then a test should be included that takes the long way around to verify that the shortcut leads to the same place (now and evermore). :earth_americas:

aseth1: I don't disagree with your point.  In this case the "short cut" leads to the same destination as measured by existing regression tests (they test the answers remained the same). So my question is what other answers should I be checking? Your statement implies that we are at some new location and the current tests are insufficient at identifying the location adequately. What would be a sufficient test? You also suggest that it should be a regression test which means we should be looking for results from the previous code base that would have changed do to the short cut. I could use a hand in figuring out which would be sufficient results to compare since previous simulation results are not sufficient in your view, correct?

aseth1: I think you are asking for more coverage of the conditions and to do that we can come up with atypical but plausible use cases like:
 1. repeated property changes and performing a brief simulation upon each repeat where answers are known to change
 2. similar to 1. but a body is added (via a joint) so that the number of states is changing
 3. purposed wiping out the system (1. and 2. should have this as a side-effect)

Let me know if you have additional cases you would like me to try.
Should these be baked into testComponents or testComponentInterface or make a new test?

tkuchida: Hm, okay. I thought it might be useful to add a Debug mode–only assert here, just in case there's some other condition the needs to be checked in the future (and assuming the existing tests don't provide sufficiently detailed diagnostic information). I don't know the new API well enough to know whether these four conditions constitute a necessary and sufficient set, but the comments didn't immediately convince me that there isn't something missing. One specific question, though: why isn't `_statesAssociatedSystem.empty()` negated? Don't you want to check that it is _not_ empty? :water_buffalo:

aseth1: You are right `_statesAssociatedSystem.empty()`  should NOT be empty. The conditions are not independent since the `_allStateVariables.size() == nsv` condition is also false initially so either would trigger a rebuild of the list. Test 3. or something like it is needed to exercise where a list may already exist and try to sweep the system out beneath it.

tkuchida: I see. I like your approach of adding use cases that exercise this condition (rather than the low-level comparison I suggested).

aseth1:
@tkuchida I tested the effect of removing each condition in isolation to see if the current tests were sufficient to catch expected failures. For example, the last check that the `System` at the time the StateVariables list was built is equivalent to the current `System` backing the Component does in fact protect from accessing invalid `StateVariables` (reference pointers are to garbage since StateVariables are destroyed and new ones created with new indices from the new System during `Model::initSystem`). This is exercised in `testForces::testExternalForce()` where the Model is simulated twice each time integrating with a different system (`initSystem` called between the two) .

It also turned out that in most cases (models with muscles) that the `isObjectUpToDateWithProperties()` was `false` forcing a rebuilding of the list. This is related to a bug in GeometryPath (there is already a TODO there) where computing the path (defined by PathPoints) requires an update of its `PathPointSet` property. Long story short, the caching of the StateVariables can temporarily ignore the properties change. We need a general mechanism to invalidate (wipeout) the `System` when properties are changed, in which case we would not need this check either. But first we need to fix the GeometryPath's mangling of properties!

Also, not all property changes are equal. We definitely want to catch the ones that change the number of state variables.  Luckily, when a component is added, `finalizeFromProperties` is being invoked which clears all the Component's indices into the system (since they should be defined in `addToSystem` after `finalizeFromProperties` is done. In which case, you cannot access StateVariables since they have not been allocated. Without the checking the System equivalence condition we segfault for the same reason that the StateVariables list has stale references to Component StateVariables which were destroyed at `finalizeFromProperties` and redefined at `addToSystem`.

Then for all intents and purposes the existing tests already exercise the three cases I was intending to add to verify the conditions for rebuilding the cached list of StateVariables.

Let me know if you rather discuss in person.

tkuchida:
Discussed in person. I'm convinced, but someone else should review as well. :game_die:

aseth1:
@chrisdembia we pulled you into the conversation yesterday and value your feedback. I liked your ideas of structurally imposing the order of state variables (i.e. using the same traversal path each time) which I think is doable long term, but when I attempted this there were many secondary issues, such as  compatibility of naming with existing results, which started to affect methods like formCompleteStorage etc... 

This change achieves the performance improvement we are looking for with no changes to the API and no side effects. The immediate goal is to get the tests to execute faster so we can waste less time dealing with failing CI tests.

Similarly, I have follow up changes that can re-enable tests in debug and begin to reverse the trend of increasing test times and ignoring tests in debug.

Please let me know if you are OK with merging.

chrisdembia:
I'm OK with merging.

aseth1:
Thanks!

PR #984, opened 2016-05-25
Reduce use of setStateVariableValue

aseth1:
Similar to PR #982, when setting `StateVariable` values en mass, we should use `Component::setStateVariableValues()` with a `Vector` of values instead of looping through each state variable by name. In which case, the values should be mapped directly to StateVariables instead of finding each StateVariable by name. The same "caching" of the list of StateVariables used to speedup calls to `getStateVariableValues()` is used here to set the state variable values.

The `AnalyzeTool` was modified to map the order of read in state values to the order expected by the Model, once, before looping through each row of the states Storage. This was then used to set the vector of state values in the Model order expected by `setStateVariableValues()` (i.e. same order as `getStateVariableNames()`) at every time frame of the analyses.

I also (accidentally) included minor changes to `GeometryPath` to use methods of `Point` and some other cleanup (e.g. using references instead of making copies of Transforms).

### DIFF
--- a/OpenSim/Simulation/Model/GeometryPath.cpp
+++ b/OpenSim/Simulation/Model/GeometryPath.cpp
@@ -919,8 +919,11 @@ void GeometryPath::computeLengtheningSpeed(const SimTK::State& s) const
         start->getVelocity(s, velStartLocal);
         end->getVelocity(s, velEndLocal);
 
-        velStartMoving = start->getBody().getTransformInGround(s).R()*velStartLocal;
-        velEndMoving = end->getBody().getTransformInGround(s).R()*velEndLocal;
+        velStartMoving = start->getBody()
+            .expressVectorInAnotherFrame(s, velStartLocal, getModel().getGround());
+
+        velEndMoving = end->getBody()
+            .expressVectorInAnotherFrame(s, velEndLocal, getModel().getGround());
 
         // Calculate the relative positions and velocities.
         posRelative = posEndInertial - posStartInertial;
--- a/OpenSim/Simulation/Model/GeometryPath.cpp
+++ b/OpenSim/Simulation/Model/GeometryPath.cpp
@@ -907,17 +907,11 @@ void GeometryPath::computeLengtheningSpeed(const SimTK::State& s) const
         end   = currentPath[i+1];
 
         // Find the positions and velocities in the inertial frame.
-        posStartInertial =
-            start->getBody().getTransformInGround(s)*start->getLocation();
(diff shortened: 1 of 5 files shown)


aseth1:
Also the motivation for this change is to bring more tests inline with their previous run-times. In particular, This change reduced `testInducedAccelerations` from 167s to 70s (locally, and prior to MCI changes this was 63s). Besides reducing CI cycle time, the hope is to also re-enable tests in debug.

OpenSim/Simulation/Model/GeometryPath.cpp
```diff
@@ -907,29 +907,20 @@ void GeometryPath::computeLengtheningSpeed(const SimTK::State& s) const
...
         start->getVelocity(s, velStartLocal);
         end->getVelocity(s, velEndLocal);
 
-        velStartMoving = start->getBody()
-            .expressVectorInAnotherFrame(s, velStartLocal, getModel().getGround());
-
-        velEndMoving = end->getBody()
-            .expressVectorInAnotherFrame(s, velEndLocal, getModel().getGround());
+        velStartMoving = start->getBody().getTransformInGround(s).R()*velStartLocal;
+        velEndMoving = end->getBody().getTransformInGround(s).R()*velEndLocal;
```
chrisdembia: I prefer the original version of these two.

OpenSim/Tools/AnalyzeTool.cpp
```diff
@@ -613,6 +613,20 @@ void AnalyzeTool::run(SimTK::State& s, Model &aModel, int iInitial, int iFinal,
     SimTK::Vector stateData;
     stateData.resize(numOpenSimStates);
 
+    const Array<std::string>& stateNames = aStatesStore.getColumnLabels();
+    Array<std::string> modelStateNames = aModel.getStateVariableNames();
+
+    int nsData = stateNames.size() - 1;  //-1 since time is a column
```
chrisdembia: Can you add a comment saying why this mapping is done? Otherwise, someone else might come here and scrap it, thinking it's unnecessarily complicated.

chrisdembia:
I made a few small comments. If we don't try re-enabling those tests as part of this PR, we may never remember to do so. May also want to update #393.

aseth1:
I have another change coming up that also addresses part of the debug build issue and then I will update #393 and include (remove remove from the exclusion list) the tests that don't timeout.

aseth1:
@chrisdembia I made changers per your comments. Ready for review when you have a chance.

aseth1:
Thanks!

PR #986, opened 2016-05-27
Remove isObjectUpToDateWithProperties() checks (asserts) from Thelen2003Muscle

aseth1:
Checking `isObjectUpToDateWithProperties()` prior to any calculation in a `Component` is not a sustainable approach to maintaining validity. That is a separate issue from this PR, but why I think we are justified in removing the asserts.

The asserts fail because `Thelen2003Muscle::isObjectUpToDateWithProperties` is almost always `false` since `GeometryPath` updates its `PathPointSet` property during a simulation.
This was causing any test case using the `Thelen2004Muscle` to fail in debug. IMHO it is more important to run these tests in debug than have these asserts in place, so they were removed. The GeometryPath rummaging through properties is also another issue we should fix, but separate from this one.

Locally, I still get tests that timeout but others like `testInducedAccelerations` now pass.

>
1>  87% tests passed, 9 tests failed out of 70
1>
1>  Total Test time (real) = 3143.60 sec
1>
1>  The following tests FAILED:
1>  	 42 - testOptimizationExampleRuns (Timeout)
1>  	 53 - testWrapping (Timeout)
1>  	 62 - testCMCArm26 (Timeout)
1>  	 63 - testCMCEMGDrivenArm (Timeout)
1>  	 64 - testCMCGait10dof18musc (Timeout)
1>  	 65 - testCMCSingleMuscle (Timeout)
1>  	 66 - testCMCSingleRigidTendonMuscle (Timeout)
1>  	 68 - testCMCWithControlConstraintsGait2354 (Timeout)
1>  	 69 - testCMCWithControlConstraintsRunningModel (Timeout)

Once the enabled CI tests (in debug) pass I will restore those tests that do not appear in the above list and then update #393, with a change from 15 exclusions (failures) to the 9 above.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -117,7 +117,7 @@ before_install:
   ## On OSX, we know that testInitState fails; exclude it.
   - if [ "$TRAVIS_OS_NAME" = "osx" ]; then export TESTS_TO_EXCLUDE="$TESTS_TO_EXCLUDE|testInitState"; fi
   ## If we are building in debug, there are some tests to ignore.
-  - if [ "$BTYPE" = "Debug" ]; then export TESTS_TO_EXCLUDE="$TESTS_TO_EXCLUDE|testInitState|testPrescribedForce|testInducedAccelerations|testAssemblySolver|testCMC|testOptimizationExample|testComponents|testForward|testWrapping"; fi
+  - if [ "$BTYPE" = "Debug" ]; then export TESTS_TO_EXCLUDE="$TESTS_TO_EXCLUDE|testCMC|testOptimizationExample|testWrapping"; fi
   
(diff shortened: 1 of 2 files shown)


aseth1:
@chrisdembia @aymanhab @klshrinidhi I was able to restore 6 tests that were being excluded, namely: testInitState, testPrescribedForce, testInducedAccelerations, testAssemblySolver, testComponents, testForward.

They still take a long time to execute, but at least they are being tested again. Let me know what you think.

chrisdembia:
LGTM.

aymanhab:
Looks good @aseth1 Merging.

aseth1:
Thanks!!

PR #987, opened 2016-05-27
Add constructors to DataTable/TimeSeriesTable to accept filename

klshrinidhi:
As discussed in today's meeting.


OpenSim/Common/TimeSeriesTable.h
```diff
@@ -123,6 +123,35 @@ class TimeSeriesTable_ : public DataTable_<double, ETY> {
                          TimeColumnNotIncreasing);
     }
 
+    /** Construct TimeSeriesTable_ from a file.                           
+    
+    \param filename Name of the file. File should contain only one table. For
+                    example, trc, csv & sto files contain one table whereas a 
+                    c3d file can contain more than.                           
+
```
aseth1: Did you consider adding `tableName` as an optional argument to handle the case that multiple tables are returned by the adapter? Then, there would be a single interface to handle all supported file formats.

For example, it would be great to do IK with a data source that could be a: .trc, .c3d, or .sto.

chrisdembia:
LGTM (once tests pass). Will the TableSource filename property be in a separate PR?

klshrinidhi:
This PR is not about `TableSource`. Changes to `TableSource` will come later.

chrisdembia:
> Changes to TableSource will come later.

Sounds good.

klshrinidhi:
* Added filename parameter to the constructor.
* Made `testC3DFileAdapter` faster. It was already slow and it got even slower after adding more tests in this PR.

klshrinidhi:
:question: :question:

aseth1:
I noticed that the treadmillRunning.C3D was eliminated. I think that is OK if conditions are covered by other files.

@klshrinidhi @jimmyDunne are there differences in the C3Ds beyond the activity being tested? Not for this PR, but I think it would be better to rename the files for the particular aspects being tested. For example, do some have multiple force-plates, or w/out EMG, and/or other auxiliary data? Are there assumptions that C3Ds must contain Markers? Maybe a C3D without marker data would be appropriate as well.

Other than that it looks good to me and the speedup is welcomed!

jimmyDunne:
The c3d's do have different orientations and types of forceplates. This will need to be tested. I will have to look at each c3d and get a scheme for what we are testing.

klshrinidhi:
After this PR added more tests to C3D, it slowed down the (already slow) C3D test a lot. So tried speeding it up by removing a test file.
Right now the C3D test is not taking into account differences in the files. I just wanted to have as many files as possible in the tests. I measured the time each c3d file was taking to read and removed the slowest one.
We can later add more smaller c3d files to test on if we see problems.

aseth1:
@klshrinidhi @jimmyDunne it is great to have sample files, I agree, and we can be more systematic in our coverage so we aren't just loading files that are effectively the same from the C3DAdapter's point of view.
I will make a separate issue as a reminder. In the meantime, I'll merge this. Thanks!

PR #989, opened 2016-05-31
Add ability to skip Travis build using [skip travis]

klshrinidhi:
Right now, Travis does not have an equivalent of [skip appveyor].

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/989)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,6 +62,8 @@ addons:
 before_install:
   ## Ensure that there are no tabs in source code.
   - cd $TRAVIS_BUILD_DIR
+  # Stop build if comment contains [skip travis].
+  - if $(git log -n1 --format="%B" | grep --quiet '\[skip travis\]'); then exit; fi
   # GREP returns 0 (true) if there are any matches, and
   # we don't want any matches. If there are matches,
   # print a helpful message, and make the test fail by using "false".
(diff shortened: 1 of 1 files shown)


klshrinidhi:
Travis Linux builds are failing becuase http://llvm.org/apt/llvm-snapshot.gpg.key is not reachable.
Will restart the build later.

chrisdembia:
Should travis give green or red when skipped?

klshrinidhi:
Travis will give green when [skip travis] is specified. It will finish the build in something like 35sec.

tkuchida:
Does `[ci skip]` still work? If so, perhaps `[travis skip]` and `[appveyor skip]` for consistency?

chrisdembia:
Appveyor only has `[skip appveyor]`, so this PR is consistent:

image

Travis and Appveyor both detect `[skip ci]` and `[skip ci]`, but Appveyor does not detect `[appveyor skip]`.

chrisdembia:
Though, @klshrinidhi could you update this PR so the tests pass?

tkuchida:
@chrisdembia Thanks for the explanation. :snowboarder:

klshrinidhi:
:white_check_mark:

chrisdembia:
@klshrinidhi it's not clear to me that the travis builds will be green ✅ if one uses `[skip travis]`. Could you push a commit that skips travis so we can verify?

chrisdembia:
This will be very helpful to have! The builds still queue up but then take only a minute or so before aborting.

PR #990, opened 2016-06-01
Update GeometryPath computations to use Point methods

aseth1:
Computations by `GeometryPath` to determine the location of and speed between points are delegated to `Point`, which is now performing these calculations and reduce duplication. This greatly simplifies `GeometryPath`, but more importantly allows us to eliminate a major source of property mangling during a simulation: `GeometryPath` no longer updates a `MovingPathPoint`'s `location` property during a simulation. We are a lot closer towards enforcing that properties must be `const` during a simulation (`GeometryPath` being the primary violator). I agree with arguments that `MovingPathPoint` should not even have a `location` property, which it inherits from its `Station` super class. `PathPoint` then, should not be a `Station` either. I will leave that for another PR. 

I also made  `PathPoint` (and its children) conform to the Component API and eliminated custom *connect* like methods, which required a confusing call sequence (e.g. called when adding a Point). From the API perspective there was no need or use of the `PathPoint`'s reference (pointer)  back to a `GeometryPath` and all the `connectToModelAndPath` and `connectToModelAndBody` were removed. We are one step closer to eliminating `PathPoint` altogether and having `GeometryPath` being composed by `Points`. 

Nonetheless, the updated PathPoints are included in `testComponents` with some added handling for `Connector<Coordinate>`.

I detected no change in test performances (timings) due to these changes.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/990)

### DIFF
--- a/OpenSim/Simulation/Model/ConditionalPathPoint.cpp
+++ b/OpenSim/Simulation/Model/ConditionalPathPoint.cpp
@@ -110,6 +110,17 @@ void ConditionalPathPoint::setCoordinate(const Coordinate& coordinate)
     updConnector<Coordinate>("coordinate").connect(coordinate);
 }
 
+bool ConditionalPathPoint::hasCoordinate() const
+{
+    return getConnector<Coordinate>("coordinate").isConnected();
+}
+
+const Coordinate& ConditionalPathPoint::getCoordinate() const
+{
+    return getConnectee<Coordinate>("coordinate");
+}
+
+
 //_____________________________________________________________________________
 /**
  * Set the range min.
--- a/OpenSim/Simulation/Model/ConditionalPathPoint.cpp
+++ b/OpenSim/Simulation/Model/ConditionalPathPoint.cpp
@@ -100,10 +100,8 @@ void ConditionalPathPoint::constructConnectors()
 }
 
 //_____________________________________________________________________________
-/**
+/*
  * Set the coordinate that this point is linked to.
- *
- * @return Whether or not this point is active.
  */
 void ConditionalPathPoint::setCoordinate(const Coordinate& coordinate)
 {
--- a/OpenSim/Simulation/Model/ConditionalPathPoint.cpp
+++ b/OpenSim/Simulation/Model/ConditionalPathPoint.cpp
@@ -26,11 +26,6 @@
 //=============================================================================
 #include "ConditionalPathPoint.h"
 #include <OpenSim/Common/XMLDocument.h>
-#include <OpenSim/Simulation/Model/Model.h>
-#include <OpenSim/Simulation/Model/GeometryPath.h>
-#include <OpenSim/Simulation/Model/CoordinateSet.h>
-#include <OpenSim/Simulation/SimbodyEngine/SimbodyEngine.h>
-#include <OpenSim/Simulation/SimbodyEngine/Body.h>
 #include <OpenSim/Simulation/SimbodyEngine/Coordinate.h>
 
 //=============================================================================
@@ -59,7 +54,7 @@ ConditionalPathPoint::~ConditionalPathPoint()
 {}
 
 //_____________________________________________________________________________
-/**
+/*
  * Override default implementation by object to intercept and fix the XML node
  * underneath the model to match current version
  */
@@ -84,7 +79,7 @@ void ConditionalPathPoint::updateFromXMLNode(SimTK::Xml::Element& node,
 }
 
 //_____________________________________________________________________________
-/**
+/*
  * Connect properties to local pointers.
  */
 void ConditionalPathPoint::constructProperties()
@@ -120,10 +115,8 @@ const Coordinate& ConditionalPathPoint::getCoordinate() const
 
 
 //_____________________________________________________________________________
-/**
+/*
  * Set the range min.
- *
- * @param aRange range min to change to.
  */
 void ConditionalPathPoint::setRangeMin(double minVal)
 {
@@ -131,10 +124,8 @@ void ConditionalPathPoint::setRangeMin(double minVal)
 }
 
 //_____________________________________________________________________________
-/**
+/*
  * Set the range max.
- *
- * @param aRange range max to change to.
  */
 void ConditionalPathPoint::setRangeMax(double maxVal)
 {
@@ -142,11 +133,9 @@ void ConditionalPathPoint::setRangeMax(double maxVal)
 }
 
 //_____________________________________________________________________________
-/**
+/*
  * Determine if this point is active by checking the value of the
  * coordinate that it is linked to.
- *
- * @return Whether or not this point is active.
  */
 bool ConditionalPathPoint::isActive(const SimTK::State& s) const
 {
--- a/OpenSim/Simulation/Model/ConditionalPathPoint.cpp
+++ b/OpenSim/Simulation/Model/ConditionalPathPoint.cpp
@@ -46,13 +46,9 @@ using namespace OpenSim;
 /**
  * Default constructor.
  */
-ConditionalPathPoint::ConditionalPathPoint() :
-   _range(_rangeProp.getValueDblArray()),
-    _coordinateName(_coordinateNameProp.getValueStr()),
-    _coordinate(NULL)
+ConditionalPathPoint::ConditionalPathPoint() : PathPoint()
 {
-    setNull();
-    setupProperties();
+    constructInfrastructure();
 }
 
 //_____________________________________________________________________________
@@ -60,117 +56,47 @@ ConditionalPathPoint::ConditionalPathPoint() :
  * Destructor.
  */
 ConditionalPathPoint::~ConditionalPathPoint()
-{
-}
-
-//_____________________________________________________________________________
-/**
- * Copy constructor.
- *
- * @param aPoint ConditionalPathPoint to be copied.
- */
-ConditionalPathPoint::ConditionalPathPoint(const ConditionalPathPoint &aPoint) :
-   PathPoint(aPoint),
-   _range(_rangeProp.getValueDblArray()),
-    _coordinateName(_coordinateNameProp.getValueStr()),
-    _coordinate(NULL)
-{
-    setNull();
-    setupProperties();
-    copyData(aPoint);
-}
-
-//=============================================================================
-// CONSTRUCTION METHODS
-//=============================================================================
-//_____________________________________________________________________________
-/**
- * Copy data members from one ConditionalPathPoint to another.
- *
- * @param aPoint ConditionalPathPoint to be copied.
- */
-void ConditionalPathPoint::copyData(const ConditionalPathPoint &aPoint)
-{
-    _range = aPoint._range;
-    _coordinateName = aPoint._coordinateName;
-    _coordinate = aPoint._coordinate;
-}
-
-//_____________________________________________________________________________
-/**
- * Initialize a ConditionalPathPoint with data from a PathPoint.
- * This function should not do anything to invalidate the path,
- * because the via point may not be added to it.
- *
- * @param aPoint PathPoint to be copied.
- */
-void ConditionalPathPoint::init(const PathPoint& aPoint)
-{
-    PathPoint::copyData(aPoint);
-
-    // If aPoint is a ConditionalPathPoint, then you can copy all of its members over.
-    // Otherwise, set the range and coordinate name to default values (using the first
-    // coordinate in the model, if there is one).
-   const ConditionalPathPoint* mmp = dynamic_cast<const ConditionalPathPoint*>(&aPoint);
-    if (mmp) {
-        copyData(*mmp);
-    } else {
-        _range[0] = 0.0;
-        _range[1] = 0.0;
-        GeometryPath* path = aPoint.getPath();
-        if (path) {
-            ModelComponent* comp = dynamic_cast<ModelComponent*>(path->getOwner());
-            if (comp) {
-                const CoordinateSet& coords = comp->getModel().getCoordinateSet();
-                if (coords.getSize() > 0) {
-                    int index = 0;
-                    _coordinateName = coords.get(index).getName();
-                    _range[0] = coords.get(index).getRangeMin();
-                    _range[1] = coords.get(index).getRangeMax();
-                }
-            }
-        }
-    }
-}
-
-//_____________________________________________________________________________
-/**
- * Set the data members of this ConditionalPathPoint to their null values.
- */
-void ConditionalPathPoint::setNull()
-{
-}
+{}
 
 //_____________________________________________________________________________
 /**
  * Override default implementation by object to intercept and fix the XML node
  * underneath the model to match current version
  */
-void ConditionalPathPoint::updateFromXMLNode(SimTK::Xml::Element& aNode, int versionNumber)
+void ConditionalPathPoint::updateFromXMLNode(SimTK::Xml::Element& node, 
+                                             int versionNumber)
 {
-    if ( versionNumber < XMLDocument::getLatestVersion()){
-        if (versionNumber<=20001)
+    if (versionNumber <= 20001) {
         // Version has to be 1.6 or later, otherwise assert
-        XMLDocument::renameChildNode(aNode, "coordinates", "coordinate");
-        }
+        XMLDocument::renameChildNode(node, "coordinates", "coordinate");
+    }
+    if (versionNumber < 30505) {
+        // replace old properties with latest use of Connectors
+        SimTK::Xml::element_iterator coord = node.element_begin("coordinate");
+        std::string coord_name("");
+        if (coord != node.element_end())
+            coord->getValueAs<std::string>(coord_name);
+        XMLDocument::addConnector(node, "Connector_Coordinate_", "coordinate", coord_name);
+    }
+
     // Call base class now assuming _node has been corrected for current version
-    PathPoint::updateFromXMLNode(aNode, versionNumber);
+    PathPoint::updateFromXMLNode(node, versionNumber);
 }
 
 //_____________________________________________________________________________
 /**
  * Connect properties to local pointers.
  */
-void ConditionalPathPoint::setupProperties()
+void ConditionalPathPoint::constructProperties()
 {
-    const double defaultRange[] = {0.0, 0.0};
-    _rangeProp.setName("range");
-    _rangeProp.setValue(2, defaultRange);
-    _rangeProp.setAllowableListSize(2);
-    _propertySet.append(&_rangeProp);
+    Array<double> defaultRange(0.0, 2); //two values of the range
+    constructProperty_range(defaultRange);
+}
 
-    _coordinateNameProp.setName("coordinate");
-    _propertySet.append(&_coordinateNameProp);
+
+void ConditionalPathPoint::constructConnectors()
+{
+    constructConnector<Coordinate>("coordinate");
 }
 
 //_____________________________________________________________________________
@@ -179,13 +105,9 @@ void ConditionalPathPoint::setupProperties()
  *
  * @return Whether or not this point is active.
  */
-void ConditionalPathPoint::setCoordinate(const SimTK::State& s, Coordinate& aCoordinate)
+void ConditionalPathPoint::setCoordinate(const Coordinate& coordinate)
 {
-    if (&aCoordinate != _coordinate)
-    {
-       _coordinate = &aCoordinate;
-       _coordinateName = _coordinate->getName();
-    }
+    updConnector<Coordinate>("coordinate").connect(coordinate);
 }
 
 //_____________________________________________________________________________
@@ -194,12 +116,9 @@ void ConditionalPathPoint::setCoordinate(const SimTK::State& s, Coordinate& aCoo
  *
  * @param aRange range min to change to.
  */
-void ConditionalPathPoint::setRangeMin(const SimTK::State& s, double aMin)
+void ConditionalPathPoint::setRangeMin(double minVal)
 {
-    if (aMin <= _range[1])
-    {
-        _range[0] = aMin;
-    }
+    set_range(0, minVal);
 }
 
 //_____________________________________________________________________________
@@ -208,12 +127,9 @@ void ConditionalPathPoint::setRangeMin(const SimTK::State& s, double aMin)
  *
  * @param aRange range max to change to.
  */
-void ConditionalPathPoint::setRangeMax( const SimTK::State& s, double aMax)
+void ConditionalPathPoint::setRangeMax(double maxVal)
 {
-    if (aMax >= _range[0])
-    {
-        _range[1] = aMax;
-    }
+    set_range(1, maxVal);
 }
 
 //_____________________________________________________________________________
@@ -225,58 +141,11 @@ void ConditionalPathPoint::setRangeMax( const SimTK::State& s, double aMax)
  */
 bool ConditionalPathPoint::isActive(const SimTK::State& s) const
 {
-    if (_coordinate)
-    {
-        double value = _coordinate->getValue(s);
-        if (value >= _range[0] - 1e-5 &&
-             value <= _range[1] + 1e-5)
+    if (getConnector<Coordinate>("coordinate").isConnected()) {
+        double value = getConnectee<Coordinate>("coordinate").getValue(s);
+        if (value >= get_range(0) - 1e-5 &&
+             value <= get_range(1) + 1e-5)
             return true;
     }
-
     return false;
-}
-
-//_____________________________________________________________________________
-/**
- * Perform some set up functions that happen after the
- * object has been deserialized or copied.
- *
- * @param aModel model containing this ConditionalPathPoint.
- */
-void ConditionalPathPoint::connectToModelAndPath(Model& aModel, GeometryPath& aPath)
-{
-    // base class
-    Super::connectToModelAndPath(aModel, aPath);
-
-    // Setup() can be called before the model's coordinate set has been constructed, in which
-    // case you don't want to throw an exception if the coordinate is not found.
-    if (aModel.getCoordinateSet().getSize() > 0) {
-        // Look up the coordinate by name and store a pointer to it.
-        if (aModel.getCoordinateSet().contains(_coordinateName)) {
-            _coordinate = &aModel.getCoordinateSet().get(_coordinateName);
-        } else {
-            string errorMessage = "Error: Coordinate " + _coordinateName + " referenced in muscle " + aPath.getOwner()->getName() +
-                " does not exist in model " +   aModel.getName();
-            throw Exception(errorMessage);
-        }
-    }
-}
-
-//=============================================================================
-// OPERATORS
-//=============================================================================
-//_____________________________________________________________________________
-/**
- * Assignment operator.
- *
- * @return Reference to this object.
- */
-ConditionalPathPoint& ConditionalPathPoint::operator=(const ConditionalPathPoint &aPoint)
-{
-    // BASE CLASS
-    PathPoint::operator=(aPoint);
-
-    copyData(aPoint);
-
-    return(*this);
 }
--- a/OpenSim/Simulation/Model/MovingPathPoint.cpp
+++ b/OpenSim/Simulation/Model/MovingPathPoint.cpp
@@ -27,14 +27,12 @@
 #include <OpenSim/Common/XMLDocument.h>
 #include "MovingPathPoint.h"
 #include <OpenSim/Common/Function.h>
+#include <OpenSim/Common/Constant.h>
 #include <OpenSim/Common/MultiplierFunction.h>
 #include <OpenSim/Simulation/Model/Model.h>
 #include <OpenSim/Simulation/Model/GeometryPath.h>
-#include <OpenSim/Simulation/Model/CoordinateSet.h>
-#include <OpenSim/Simulation/SimbodyEngine/SimbodyEngine.h>
 #include <OpenSim/Simulation/SimbodyEngine/Coordinate.h>
-#include <OpenSim/Common/SimmSpline.h>
-#include <OpenSim/Simulation/SimbodyEngine/Body.h>
+
 
 //=============================================================================
 // STATICS
@@ -50,16 +48,9 @@ using SimTK::Vec3;
 /**
  * Default constructor.
  */
-MovingPathPoint::MovingPathPoint() : PathPoint(),
-   _xLocation(_xLocationProp.getValueObjPtrRef()),
-    _xCoordinateName(_xCoordinateNameProp.getValueStr()),
-   _yLocation(_yLocationProp.getValueObjPtrRef()),
-    _yCoordinateName(_yCoordinateNameProp.getValueStr()),
-   _zLocation(_zLocationProp.getValueObjPtrRef()),
-    _zCoordinateName(_zCoordinateNameProp.getValueStr())
+MovingPathPoint::MovingPathPoint() : PathPoint()
 {
-    setNull();
-    setupProperties();
+    constructInfrastructure();
 }
 
 //_____________________________________________________________________________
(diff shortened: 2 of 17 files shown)


chrisdembia:
*[OpenSim/Simulation/Wrap/PathWrap.cpp, line 87 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJH7T6AYUkZrEGzafIq:-KJH7T6AYUkZrEGzafIr:867476730) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Wrap/PathWrap.cpp#L87)):*
> ```cpp
> void PathWrap::extendConnectToModel(Model& model)
> {
>     _path = dynamic_cast<const GeometryPath*>(&getParent());
> ```

Missing a call to `Super::extendConnectToModel()`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
*[OpenSim/Simulation/Model/ConditionalPathPoint.h, line 76 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJH8Nl2LKzEJiXschfH:-KJH8Nl3FppiSEeZC9sR:512643733) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/ConditionalPathPoint.h#L76)):*
> ```objc
>     void constructConnectors() override;
>     void updateFromXMLNode(SimTK::Xml::Element& aNode,
>                            int versionNumber = -1) override;
> ```

As far as I can tell, we do not have a convention of having a default value for `versionNumber`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
*[OpenSim/Simulation/Model/MovingPathPoint.cpp, line 337 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJH9axFlkg7AKI3ycyr:-KJH9axFlkg7AKI3ycys:94168765) ([raw file](https://github.com/opensim-org/opensim-core/blob/a0a0d99ab746cf9eae42d06509d74c56568af8e3/OpenSim/Simulation/Model/MovingPathPoint.cpp#L337)):*
> ```cpp
>         msg += "must depend on same the coordinate. Condition: ";
>         msg += "x_coordinate == y_coordinate == z_coordinate  FAILED.";
>         throw Exception(msg, __FILE__, __LINE__);       
> ```

Could you explain why it is okay to remove this check?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
*[OpenSim/Simulation/Model/MovingPathPoint.h, line 98 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJHZJdoxReDGg2Ww1sq:-KJHZJdoxReDGg2Ww1sr:359729847) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/MovingPathPoint.h#L98)):*
> ```objc
>     // Get the local location of the MovingPathPoint in its Frame
>     SimTK::Vec3 getLocation(const SimTK::State& s) const;
>     // Get the local velocity of the MovingPathPoint in its Frame
> ```

I think this should be a doxygen comment. And it should also clarify if whether the derivative is taken in the PhysicalFrame.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
*[OpenSim/Simulation/Model/MovingPathPoint.h, line 98 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJHZJdoxReDGg2Ww1sq:-KJHZaqayjSGydm1Jane:-1339279731) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/MovingPathPoint.h#L98)):*
> I think this should be a doxygen comment. And it should also clarify if whether the derivative is taken in the PhysicalFrame.
</details>
Also, the comment for `getLocation()` should be a doxygen comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
*[OpenSim/Simulation/Model/MovingPathPoint.h, line 61 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJHZjeuF587tHdBJwso:-KJHZjeuF587tHdBJwsp:-1324464029) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/MovingPathPoint.h#L61)):*
> ```objc
> //==============================================================================
>     OpenSim_DECLARE_OPTIONAL_PROPERTY(x_location, Function,
>         "Function defining the x component of the point's location.");
> ```

Clarify that the location is expressed in the PhysicalFrame of the point.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
*[OpenSim/Simulation/Model/MovingPathPoint.h, line 61 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJHZjeuF587tHdBJwso:-KJHZoDnV8oNKmkgQbOu:-1704313357) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/MovingPathPoint.h#L61)):*
> Clarify that the location is expressed in the PhysicalFrame of the point.
</details>
er, the PhysicalFrame to which the point is connected.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
*[OpenSim/Simulation/Model/MovingPathPoint.cpp, line 359 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJH_3l4-WdepA4WVMli:-KJH_3l4-WdepA4WVMlj:1604271775) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/MovingPathPoint.cpp#L359)):*
> ```cpp
>     //TODO: Enable Exception or Implement the method and add accompanying test.
>     //OPENSIM_THROW(Exception, "MovingPathPoint::calcAccelerationInGround not implemented.");
>     return Vec3(SimTK::NaN);
> ```

Feel free to ignore, but the exception is preferable to the NaN. I assume you ran into some issue when trying to use the exception. Perhaps it would be good to document that issue with your TODO.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
*[OpenSim/Simulation/Model/MovingPathPoint.cpp, line 353 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJH_HSfo8KGxfMg2Cs-:-KJH_HSgSSkY_zWWwgSt:1599241507) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/MovingPathPoint.cpp#L353)):*
> ```cpp
>     // linear (vF = V_GF[1]) and angular (omegaF = V_GF[0]) velocity, such that
>     // velocity of the station: v = vF + omegaF x r
>     return V_GF[1] + V_GF[0] % r + v;
> ```

I still dislike all of this math in classes like this. It just seems like we should be using methods in Simbody or on Station to do these calculations.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
*[OpenSim/Simulation/Model/MovingPathPoint.cpp, line 337 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJH9axFlkg7AKI3ycyr:-KJHa43yDYmyz3XYekGW:343838915) ([raw file](https://github.com/opensim-org/opensim-core/blob/a0a0d99ab746cf9eae42d06509d74c56568af8e3/OpenSim/Simulation/Model/MovingPathPoint.cpp#L337)):*
> Could you explain why it is okay to remove this check?
</details>
Perhaps removing this check requires mention in the CHANGELOG.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
*[OpenSim/Simulation/Model/Point.h, line 111 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJHbtk83p_tRa_2WxEU:-KJHbtk83p_tRa_2WxEV:273072193) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/Point.h#L111)):*
> ```objc
>         const SimTK::Vec3& location) const;
>     /** Calculate the speed of motion between this Point and some other Point.
>     A positive speed is growing apart and negative speed is coming closer.
> ```

It would be clearer to me to say "the rate of change of the distance," if that is what it is. I'm fine with the name of the method being `calcSpeedBetween()` though. I'm not sure I understand the implementation of this method.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
*[OpenSim/Simulation/Model/Point.cpp, line 134 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJHdA9oqzmtphXr9MKN:-KJHdA9oqzmtphXr9MKO:-47187412) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/Point.cpp#L134)):*
> ```cpp
>     auto r = getLocationInGround(s) - o.getLocationInGround(s);
>     double d = r.norm();
>     auto v = getVelocityInGround(s) - o.getVelocityInGround(s);
> ```

Nitpicky: you could use `const auto r` and `const auto v`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
My biggest issue is that `MovingPathPoint` is a `Station`. It is *not* a Station. A Station is a point fixed to a Body.

> #### Update GeometryPath computations to use Point methods
> 
> Computations by `GeometryPath` to determine the location of and speed between points are delegated to `Point`, which is now performing these calculations and reduce duplication. This greatly simplifies `GeometryPath`, but more importantly allows us to eliminate a major source of property mangling during a simulation: `GeometryPath` no longer updates a `MovingPathPoint`'s `location` property during a simulation. We are a lot closer towards enforcing that properties must be `const` during a simulation (`GeometryPath` being the primary violator). I agree with arguments that `MovingPathPoint` should not even have a `location` property, which it inherits from its `Station` super class. `PathPoint` then, should not be a `Station` either. I will leave that for another PR. 
> 
> 
> 
> I also made  `PathPoint` (and its children) conform to the Component API and eliminated custom *connect* like methods, which required a confusing call sequence (e.g. called when adding a Point). From the API perspective there was no need or use of the `PathPoint`'s reference (pointer)  back to a `GeometryPath` and all the `connectToModelAndPath` and `connectToModelAndBody` were removed. We are one step closer to eliminating `PathPoint` altogether and having `GeometryPath` being composed by `Points`. 
> 
> 
> 
> Nonetheless, the updated PathPoints are included in `testComponents` with some added handling for `Connector<Coordinate>`.
> 
> 
> 
> I detected no change in test performances (timings) due to these changes.
> 
> 
> 
</details>


---







*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-:-KJHdScr2CoI8RalaNZ2:-1809171097)*

chrisdembia:
I realize that this PR did not introduce the fact that `MovingPathPoint` is a `Station`. Nonetheless, I'm stopping my review for now until we can discuss that.

> My biggest issue is that `MovingPathPoint` is a `Station`. It is *not* a Station. A Station is a point fixed to a Body.
</details>


---







*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-:-KJHdmVR8DtLt_INFxNc:1642502505)*

chrisdembia:
Reviewed 6 of 17 files at r1.
Review status: 6 of 17 files reviewed at latest revision, 9 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
Ah I see you mentioned this in your PR description. Could you elaborate on any possible consequences of this, and temporarily document this in doxygen as a bug in MovingPathPoint? A component that wants to connect to a Station could actually be connected to a MovingPathPoint, causing unphysical behavior.

> I realize that this PR did not introduce the fact that `MovingPathPoint` is a `Station`. Nonetheless, I'm stopping my review for now until we can discuss that.
</details>


---







*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-:-KJHeCZqKzeI9-Dzebvi:1578285339)*

chrisdembia:
*[OpenSim/Simulation/Wrap/PathWrap.cpp, line 87 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJH7T6AYUkZrEGzafIq:-KJHfgektaVbrXdReulG:-1875309029) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Wrap/PathWrap.cpp#L87)):*
> Missing a call to `Super::extendConnectToModel()`?
</details>
Consider throwing an exception if the parent is not a GeometryPath.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
*[OpenSim/Simulation/Wrap/PathWrap.cpp, line 87 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJHfvxPD3aI2agrTK7F:-KJHfvxPD3aI2agrTK7G:-1868760122) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Wrap/PathWrap.cpp#L87)):*
> ```cpp
> void PathWrap::extendConnectToModel(Model& model)
> {
>     _path = dynamic_cast<const GeometryPath*>(&getParent());
> ```

Missing a call to Super::extendConnectToModel()?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
It would be good to know your plan for addressing the issue that MovingPathPoint is a Station. One hack to help prevent mis-use for now is for MovingPathPoint to override the `location` property and set it to NaN. Not sure if that would mess up other things...

> Ah I see you mentioned this in your PR description. Could you elaborate on any possible consequences of this, and temporarily document this in doxygen as a bug in MovingPathPoint? A component that wants to connect to a Station could actually be connected to a MovingPathPoint, causing unphysical behavior.
</details>


---







*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-:-KJHh1XO0g3mu1wp6Csj:-1379753681)*

chrisdembia:
*[OpenSim/Simulation/Model/MovingPathPoint.cpp, line 336 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJHha4LPW7l_wsUwrhW:-KJHha4LPW7l_wsUwrhX:1135854336) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/MovingPathPoint.cpp#L336)):*
> ```cpp
> SimTK::Vec3 MovingPathPoint::calcLocationInGround(const SimTK::State& s) const
> {
>     return getParentFrame().getTransformInGround(s)*getLocation(s);
> ```

This tells me (as does similar code elsewhere) that there should be a `Frame::expressVectorInGround()`. I don't remember if that was discussed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
*[OpenSim/Simulation/Model/MovingPathPoint.h, line 112 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJHj6BKGf0TjZINvZI6:-KJHj6BLRfKXro6eE0am:-919867296) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/MovingPathPoint.h#L112)):*
> ```objc
>     SimTK::Vec3 calcLocationInGround(const SimTK::State& state) const override;
>     SimTK::Vec3 calcVelocityInGround(const SimTK::State& state) const override;
>     SimTK::Vec3 calcAccelerationInGround(const SimTK::State& state) const override;
> ```

These methods should be `final` in Station, once we remove `Station` as `MovingPathPoint`'s base class.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
*[OpenSim/Simulation/Model/ConditionalPathPoint.cpp, line 79 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJHjbZpgUh_4Ub6OmV4:-KJHjbZqVCBjVJ8gE4VV:139744598) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/ConditionalPathPoint.cpp#L79)):*
> ```cpp
>         if (coord != node.element_end())
>             coord->getValueAs<std::string>(coord_name);
>         XMLDocument::addConnector(node, "Connector_Coordinate_", "coordinate", coord_name);
> ```

The connectee name specified here would ideally be the path to the coordinate, not just its name. Don't worry about it for this PR though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
Reviewed 11 of 17 files at r1.
Review status: all files reviewed at latest revision, 13 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

aseth1:
*[OpenSim/Simulation/Model/MovingPathPoint.cpp, line 359 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJH_3l4-WdepA4WVMli:-KJI5zfoK1QCQaFqYCHD:1782883518) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/MovingPathPoint.cpp#L359)):*
> Feel free to ignore, but the exception is preferable to the NaN. I assume you ran into some issue when trying to use the exception. Perhaps it would be good to document that issue with your TODO.
</details>
Yes, the Exception causes testComponents' automatic verification of Outputs to fail. My preference is to simply implement it but need to also come up with test case(s) so holding off for now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

aseth1:
*[OpenSim/Simulation/Model/MovingPathPoint.cpp, line 353 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJH_HSfo8KGxfMg2Cs-:-KJI76RscR8Z9LyKHHTA:-751393493) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/MovingPathPoint.cpp#L353)):*
> I still dislike all of this math in classes like this. It just seems like we should be using methods in Simbody or on Station to do these calculations.
</details>
It looks like Sherm intended to have similar methods in Simbody (e.g. `MobilizedBody::calcBodyMovingPointVelocityInBody()` but it is not implemented yet.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

aseth1:
*[OpenSim/Simulation/Model/MovingPathPoint.cpp, line 336 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJHha4LPW7l_wsUwrhW:-KJIcKuxOnqDVjeKGZbo:155462457) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/MovingPathPoint.cpp#L336)):*
> This tells me (as does similar code elsewhere) that there should be a `Frame::expressVectorInGround()`. I don't remember if that was discussed.
</details>
I agree that Frame::findLocationInGround and expressVectorInGround would be generally useful. Can that be a separate PR to augment Frame?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
*[OpenSim/Simulation/Model/MovingPathPoint.cpp, line 336 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJHha4LPW7l_wsUwrhW:-KJIdiAfjQOX48UaAlVd:160394146) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/MovingPathPoint.cpp#L336)):*
> I agree that Frame::findLocationInGround and expressVectorInGround would be generally useful. Can that be a separate PR to augment Frame?
</details>
Yes

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

aseth1:
We should not have Path specific Points of any kind. They should be considered "path" Points by virtue of being subcomponents of a GeometryPath. Given the current pace of thinning out PathPoints and updates to GeometryPath this could be achieved before releasing 4.0.

I like you idea of overriding any user setting for location and making the location NaN until we have a proper fix in place.

> It would be good to know your plan for addressing the issue that MovingPathPoint is a Station. One hack to help prevent mis-use for now is for MovingPathPoint to override the `location` property and set it to NaN. Not sure if that would mess up other things...
</details>


---

Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks broke.

---

*[OpenSim/Simulation/Model/ConditionalPathPoint.cpp, line 79 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJHjbZpgUh_4Ub6OmV4:-KJIfRp2awzUlewxJKJZ:-864019388) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/ConditionalPathPoint.cpp#L79)):*
> The connectee name specified here would ideally be the path to the coordinate, not just its name. Don't worry about it for this PR though.
</details>
True. But the update code cannot get the fullPathName() from old models, just its `coordinate` property value (string).

---

*[OpenSim/Simulation/Model/ConditionalPathPoint.h, line 76 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJH8Nl2LKzEJiXschfH:-KJIgPP42EBt0chpixqk:1675637715) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/ConditionalPathPoint.h#L76)):*
> As far as I can tell, we do not have a convention of having a default value for `versionNumber`.
</details>
removed.

---

*[OpenSim/Simulation/Model/MovingPathPoint.cpp, line 337 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJH9axFlkg7AKI3ycyr:-KJIdUnQur-5AN4n3wW7:-1954417350) ([raw file](https://github.com/opensim-org/opensim-core/blob/a0a0d99ab746cf9eae42d06509d74c56568af8e3/OpenSim/Simulation/Model/MovingPathPoint.cpp#L337)):*
> Perhaps removing this check requires mention in the CHANGELOG.
</details>
It was not OK to remove. Thanks for flagging and making me double check. I restored and added more details in the comments.

---

*[OpenSim/Simulation/Model/MovingPathPoint.h, line 98 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJHZJdoxReDGg2Ww1sq:-KJIc7zLBHbsVmChPut-:-499146009) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/MovingPathPoint.h#L98)):*
> Also, the comment for `getLocation()` should be a doxygen comment.
</details>
Done.

---

*[OpenSim/Simulation/Model/Point.cpp, line 134 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJHdA9oqzmtphXr9MKN:-KJIc4U-qsS9CgCX9tCE:-499146009) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/Point.cpp#L134)):*
> Nitpicky: you could use `const auto r` and `const auto v`.
</details>
Done.

---

*[OpenSim/Simulation/Model/Point.h, line 111 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJHbtk83p_tRa_2WxEU:-KJIbzGdME2QsO8QAMDb:-499146009) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/Point.h#L111)):*
> It would be clearer to me to say "the rate of change of the distance," if that is what it is. I'm fine with the name of the method being `calcSpeedBetween()` though. I'm not sure I understand the implementation of this method.
</details>
Done.

---

*[OpenSim/Simulation/Wrap/PathWrap.cpp, line 87 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJH7T6AYUkZrEGzafIq:-KJIc04pd1jSfzaA1NZg:-499146009) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Wrap/PathWrap.cpp#L87)):*
> Consider throwing an exception if the parent is not a GeometryPath.
</details>
Done.

---

*[OpenSim/Simulation/Wrap/PathWrap.cpp, line 87 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJHfvxPD3aI2agrTK7F:-KJIc2bGpfmsPfMtImGY:-499146009) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Wrap/PathWrap.cpp#L87)):*
> Missing a call to Super::extendConnectToModel()?
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-:-KJIexQTi6a1MvcVNc_Y:588181028)*

chrisdembia:
*[OpenSim/Simulation/Wrap/PathWrap.cpp, line 87 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJH7T6AYUkZrEGzafIq:-KJNTq1sbwExfcpj3I_8:215993149) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Wrap/PathWrap.cpp#L87)):*
> Done.
</details>
No need to change this here, but this `msg` will be created even if the condition is false. In the future, I think we should avoid such overhead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
:lgtm:

> We should not have Path specific Points of any kind. They should be considered "path" Points by virtue of being subcomponents of a GeometryPath. Given the current pace of thinning out PathPoints and updates to GeometryPath this could be achieved before releasing 4.0.
> 
> I like you idea of overriding any user setting for location and making the location NaN until we have a proper fix in place.
</details>


---







*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-:-KJNU2SqV_sJkNw24QFr:1417088721)*

chrisdembia:
Reviewed 6 of 6 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

chrisdembia:
I updated this PR so that the travis builds should pass.

I think @aseth1 is looking for one more reviewer for this PR.

aseth1:
*[OpenSim/Simulation/Model/MovingPathPoint.h, line 112 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/990#-KJHj6BKGf0TjZINvZI6:-KKFJ9YREZTzFIkeQC0e:-556781233) ([raw file](https://github.com/opensim-org/opensim-core/blob/b942da9ff9db614004e6c6526989bdcf38a8ba80/OpenSim/Simulation/Model/MovingPathPoint.h#L112)):*
> These methods should be `final` in Station, once we remove `Station` as `MovingPathPoint`'s base class.
</details>
Yes, I agree.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/990)*

aseth1:
> I think @aseth1 is looking for one more reviewer for this PR.

Yes and looking to @aymanhab for feedback on the effect of these changes on the GUI.

aseth1:
Thanks for merging in the updates @chrisdembia! Nice to see tests passing again. :clap:

aymanhab:
@aseth1 As long as the OpenSimContext class interface hasn't changed the GUI code will likely compile. I'm building against master to see the behavior before this PR and then will redo with this branch/PR.

aseth1:
@aymanhab you are right that  I did not modify `OpenSimContext.h` but there were many changes to its implementation. I just wanted to see those changes exercised as a sanity check. If you feel `testContext` provides adequate coverage then are we OK to merge this PR?

aymanhab:
@aseth1 I'd say yes, please go ahead and merge. The closer we are to the final API the better we are in general, even if we need to go back and fix some implementation of methods on the OpenSimContext class.

aymanhab:
I'll open specific issues if I run into problems while testing.

aseth1:
Thanks @aymanhab. 

@chrisdembia let me know if there is anything that needs a closer look. :microscope:

chrisdembia:
LGTM.

OpenSim/Simulation/Model/ConditionalPathPoint.cpp
```diff
@@ -46,160 +46,90 @@ using namespace OpenSim;
...
+    constructConnector<Coordinate>("coordinate");
 }
 
 //_____________________________________________________________________________
 /**
- * Override default implementation by object to intercept and fix the XML node
- * underneath the model to match current version
+ * Set the coordinate that this point is linked to.
+ *
+ * @return Whether or not this point is active.
```
tkuchida: I don't think this `void()` should have a `@return` comment.

OpenSim/Simulation/Model/ConditionalPathPoint.cpp
```diff
@@ -46,160 +46,90 @@ using namespace OpenSim;
...
 
+
 //_____________________________________________________________________________
 /**
  * Set the range min.
  *
  * @param aRange range min to change to.
  */
-void ConditionalPathPoint::setRangeMin(const SimTK::State& s, double aMin)
+void ConditionalPathPoint::setRangeMin(double minVal)
```
tkuchida: `@param aRange` should be changed to `@param minVal`

OpenSim/Simulation/Model/ConditionalPathPoint.cpp
```diff
@@ -208,12 +138,9 @@ void ConditionalPathPoint::setRangeMin(const SimTK::State& s, double aMin)
  *
  * @param aRange range max to change to.
  */
-void ConditionalPathPoint::setRangeMax( const SimTK::State& s, double aMax)
+void ConditionalPathPoint::setRangeMax(double maxVal)
```
tkuchida: `@param aRange` -> `@param maxVal` (or just delete the comment because it's obvious)

OpenSim/Simulation/Model/MovingPathPoint.cpp
```diff
@@ -72,364 +63,211 @@ MovingPathPoint::~MovingPathPoint()
...
-    } else // type == Constant
-        upd_location()[1] = _yLocation->calcValue(SimTK::Vector(1, 0.0));
+            _yCoordinate->getValue(s),
+            _yCoordinate->getRangeMax());
+        pInF[1] = get_y_location().calcValue(SimTK::Vector(1, yval));
+    }
+    else // type == Constant
+        pInF[1] = get_y_location().calcValue(SimTK::Vector(1, 0.0));
 
     if (_zCoordinate) {
```
tkuchida: `_zCoordinate` -> `!_zCoordinate.empty()` to agree with changes to x and y cases?

OpenSim/Simulation/Model/MovingPathPoint.cpp
```diff
@@ -47,472 +43,337 @@ using SimTK::Vec3;
...
     updateGeometry();
 }
+
+
+SimTK::Vec3 MovingPathPoint::calcLocationInGround(const SimTK::State& s) const
+{
+    return getParentFrame().getTransformInGround(s)*getLocation(s);
+}
+
+SimTK::Vec3 MovingPathPoint::calcVelocityInGround(const SimTK::State& s) const
```
tkuchida: Is there a test case for this method?

OpenSim/Simulation/Model/MovingPathPoint.cpp
```diff
@@ -47,472 +43,337 @@ using SimTK::Vec3;
...
+    const Vec3 r = R_GF*getLocation(s);
+
+    const SimTK::SpatialVec& V_GF = getParentFrame().getVelocityInGround(s);
+    const Vec3 v = R_GF*getVelocity(s);
+
+
+    // The velocity of the station in ground is a function of its frame's
+    // linear (vF = V_GF[1]) and angular (omegaF = V_GF[0]) velocity, such that
+    // velocity of the station: v = vF + omegaF x r
+    return V_GF[1] + V_GF[0] % r + v;
```
tkuchida: I'm confused here. Why isn't `r` the position vector from Ground to the origin of ParentFrame?

tkuchida: Also, isn't this calculation missing the "omega cross r" contribution from the angular velocity of the point in the ParentFrame?

OpenSim/Simulation/Model/MovingPathPoint.h
```diff
@@ -101,49 +77,50 @@ OpenSim_DECLARE_CONCRETE_OBJECT(MovingPathPoint, PathPoint);
...
-                                                                override;
-    void update(const SimTK::State& s) override;
-    void getVelocity(const SimTK::State& s, SimTK::Vec3& aVelocity) override;
-#endif
+
+    /** Get the local location of the MovingPathPoint in its Frame */
+    SimTK::Vec3 getLocation(const SimTK::State& s) const;
+    /** Get the local velocity of the MovingPathPoint w.r.t to and 
+        expressed in its Frame. To get the velocity of the point w.r.t.
+        and expressed in Ground, call getVelocityInGround(). */
```
tkuchida: I don't see a `getVelocityInGround()`. Was `calcVelocityInGround()` intended?

OpenSim/Simulation/Model/PathPoint.cpp
```diff
@@ -40,51 +40,25 @@ using namespace OpenSim;
...
-                newPoint->init(*aPoint);
-            }
-        }
-    }
+    // Preserve location means to switch bodies without changing
+    // the location of the point in the inertial reference frame.
+    upd_location() = currentFrame.findLocationInAnotherFrame(s, get_location(), body);
 
-    return newPoint;
+    // now assign this point's body to point to aBody
```
tkuchida: Perhaps clearer as `// now make "body" this PathPoint's parent Frame`

OpenSim/Simulation/Model/MovingPathPoint.cpp
aseth1: I did not include additional unit tests. `testMomentArms` was my working test case for checking my implementation was correct (or at least produces consistent moment-arms).  These functions are also required to evaluate muscle path lengths and speeds (for models with MovingPathPoints) which are cases in `testForward`, `testStaticOptimization`, `testCMC`, etc...  

My intention is to eliminate `MovingPathPoint`, and have all Points provide their Jacobian with respect to any set of Coordinates. The implementation of `MovingPathPoint` makes little sense with exactly one Coordinate assigned to each component of the location vector. Once points can provide their partials I plan to test that more extensively.

tkuchida:
`</review>`. I was confused by the implementation and comments in `MovingPathPoint::calcVelocityInGround()`; other suggestions are minor.

OpenSim/Simulation/Model/MovingPathPoint.cpp
aseth1: > I'm confused here. Why isn't r the position vector from Ground to the origin of ParentFrame?

Because that isn't the local rotating vector due to the Frame's angular velocity. Think about it - if it was r to Ground origin you would get different velocity contribution by virtue of its location w.r.t Ground.  For example,  if we had two identical balls spinning at identical angular velocity you would get the velocity of a point on the surface to be different by virtue of where the origins of the spheres are w.r.t. ground - that cannot be right!

aseth1: >Also, isn't this calculation missing the "omega cross r" contribution from the angular velocity of the point in the ParentFrame?

Perhaps you overlooked line 367 and comments on lines 364-366?

OpenSim/Simulation/Model/MovingPathPoint.h
aseth1: > I don't see a getVelocityInGround(). Was calcVelocityInGround() intended?

`getVelocityInGround()` is implemented by the base and handles caching the calculated value. The calculation is virtual and the responsibility of the derived Point classes. Perhaps I missed an `override`? I will check.

aseth1:
@tkuchida thank you for the careful review! Please let me know how I might make the comments or implementation of `MovingPathPoint::calcVelocityInGround()` less confusing.

OpenSim/Simulation/Model/MovingPathPoint.cpp
tkuchida: > Think about it

I did think about it. Based on your response, this method is meant to express the MovingPathPoint's local velocity in the Ground frame; in my opinion, such a method should not be named `calcVelocityInGround()`. I suggest (1) renaming the method to `expressLocalVelocityInGround()` or `calcLocalVelocityExpressedInGround()` and (2) including a description of the method on line 116 of MovingPathPoint.h.

aseth1: Bad choice of words. I meant imagine the scenario not that you did not think.

But you're still not right. Only `v` in the method is the local velocity expressed in ground. The value returned by the method is the velocity of the point in ground.

aseth1:
@tkuchida I attempted to clarify the comments and steps for calculating the velocity of the moving point in Ground (`MovingPathPoint::calcVelocityInGround()`). Let me know if it is better now. Thanks!

tkuchida:
> Let me know if it is better now.

Ok, looks good :+1: Thanks for clarifying the comments.

OpenSim/Simulation/Model/MovingPathPoint.cpp
tkuchida: > But you're still not right.

Yep, I was super confused. Thanks for clarifying.

aseth1:
Woohoo!  Thank you @tkuchida.

Issue #985, opened 2016-05-26
Range Based Loops on OpenSim containers


mitkof6:
OpenSim has many custom container facilities (i.g. `Array<>, Set<>`). It would be nice if we can loop through these containers with a  range based loops (`for(auto a : b) { ... `}) and support other types of operations such as with the std library.

chrisdembia:
+1. Would you be interested in submitting a PR for `Set<>`? If you are interested, we should wait to make sure others think this is a good idea.

Note: over time, we hope to reduce the usage of both of these classes in favor of other classes.

chrisdembia:
For looping through components, we hope to move users to `Component::getComponentList()`, which supports range based loops.

mitkof6:
> Would you be interested in submitting a PR for Set<>? 

I will try to submit a PR when I find some free time.

> For looping through components, we hope to move users to Component::getComponentList(), which supports range based loops.

This may be true for future code/users, but there may be some users who have written code through Set, Array, ... so it would be good to support this if you think of keeping these containers in the future. Apart from Set, Array, do you use other containers that need this feature?

chrisdembia:
The Property could use it too, but that is more complicated.

PR #993, opened 2016-06-04
Custom iterators for ArrayPtrs, Array, Set

mitkof6:
With regards to  #985, I have implemented a constant and non-constant iterators for the following classes: ArrayPtrs<T>, Array<T>, Set<T> (which is an aggregation to ArrayPtrs<T>). There is a simple testSetIterator.cpp that tests some basic operation.

---

This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/993)

### DIFF
--- a/OpenSim/Common/ArrayPtrs.h
+++ b/OpenSim/Common/ArrayPtrs.h
@@ -9,8 +9,9 @@
  * National Institutes of Health (U54 GM072970, R24 HD065690) and by DARPA    *
  * through the Warrior Web program.                                           *
  *                                                                            *
- * Copyright (c) 2005-2012 Stanford University and the Authors                *
+ * Copyright (c) 2005-2016 Stanford University and the Authors                *
  * Author(s): Frank C. Anderson                                               *
+ * Contributer(s): Dimitar Stanev                                             *
  *                                                                            *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may    *
  * not use this file except in compliance with the License. You may obtain a  *
@@ -57,6 +58,11 @@
  * @author Frank C. Anderson
  */
 namespace OpenSim { 
+
+/**
+* Forward declaration of custom iterator
+*/
+template<class T, bool> class ArrayPtrsIterator;
 
 template<class T> class ArrayPtrs
 {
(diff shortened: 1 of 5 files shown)


OpenSim/Common/ArrayPtrs.h
```diff
@@ -1024,6 +1064,121 @@ int searchBinary(const T &aObject,bool aFindFirst=false,
...
+* @author Dimitar Stanev
+*/
+template<class T, bool is_const_iterator = true>
+class ArrayPtrsIterator : public std::iterator<std::forward_iterator_tag, T>
+{
+    /**
+    * For const_iterator:   define ArrayPointerType to be a   pointer to a const pointer to T
+    * For regular iterator: define ArrayPointerType to be a   T**
+    */
+    typedef typename std::conditional<is_const_iterator, T* const *, T**>::type ArrayPointerType;
```
mitkof6: Also, I am not very confident with T* const *, as I accidentally find a way to bypass the compiler errors.

OpenSim/Common/ArrayPtrs.h
```diff
@@ -1024,6 +1064,121 @@ int searchBinary(const T &aObject,bool aFindFirst=false,
...
 };  // END of class ArrayPtrs
 
+
+//=============================================================================
+// Custom Iterator
+//=============================================================================
+//_____________________________________________________________________________
+
+/**
+* Inner class that describes a const_iterator and 'regular' iterator at the same time,
```
chrisdembia: This one is not inner, right?

chrisdembia:
Cool!  I'll let @klshrinidhi review this, who I think is very knowledgeable about iterators.

1. Why is one of the iterators a nested class, while the other isn't?
2. I'm not too familiar with this, but would it be possible to accomplish const/non-const with `ArrayIterator<T>` and `ArrayIterator<const T>`?

chrisdembia:
What is the reason to use `begin()` for the const version instead of `cbegin()`? Does only the former get used by the range-based for-loop?

OpenSim/Common/ArrayPtrs.h
mitkof6: If I make it inner, there are some template initializations that produce errors:

https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Object.h#L54
https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/FunctionSet.cpp#L41

mitkof6:
> What is the reason to use begin() for the const version instead of cbegin()? Does only the former get used by the range-based for-loop?

Do you think that we should use cbegin() cend() instead of begin() const?

chrisdembia:
> Do you think that we should use cbegin() cend() instead of begin() const?

I don't know; it depends on how it affects range for-loops. Perhaps @klshrinidhi has an opinion.

OpenSim/Common/ArrayPtrs.h
chrisdembia: Ah, okay.

klshrinidhi:
@mitkof6 
I looked at your changes to the class `OpenSim::Array`. I think this class is an implementation of `std::vector` and should be removed altogether. I think same applies to `OpenSim::ArrayPtrs`, which can be replaced by an array of smart pointers.

I will get back with comments on your changes to `OpenSim::Set`.

aymanhab:
We're moving away from all these constructs and there's a ton of code using existing accessors so adding iterators now may have little added value

PR #994, opened 2016-06-07
Add properties -- filename & tablename -- to Tablesource

klshrinidhi:
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/994)


OpenSim/Common/TableSource.h
```diff
@@ -65,6 +65,13 @@ class TableSource_ : public Component {
     OpenSim_DECLARE_CONCRETE_OBJECT_T(TableSource_, ET, Component);
 
 public:
+    OpenSim_DECLARE_PROPERTY(filename, std::string,
+                             "Path to the file to populate the TableSource_ "
+                             "with.");
+    OpenSim_DECLARE_PROPERTY(tablename, std::string,
+                             "Name of the table in the file to populate the "
+                             "TableSource_ with. Ex. 'markers', 'forces'.");
```
chrisdembia: Can you clarify that this is necessary only for certain file types (here and elsewhere)?

chrisdembia:
+1 Looks really good to me. Two things:

1. Could you create an issue for eventually parsing the file name relative to the XML file (rather than relative to the current working dir), and perhaps document this as a bug in the doxygen (`\note`) for this class?
2. Could you document in the doxygen for the class how the two ways of setting the table are resolved (xml vs in-memory)?

klshrinidhi:
* Added comment to the 'filename' property saying the path is relative to working directory.
* Added comment to `setTable()` overloads saying what to do if filename property is already set.

chrisdembia:
Tests failed.

klshrinidhi:
Fixed the tests.

OpenSim/Common/TableSource.h
```diff
@@ -75,16 +84,33 @@ class TableSource_ : public Component {
...
+    /** Construct the TableSource_ from a file.
+
+    \param filename Name of the file.
+    \param tablename Name of the table in the file to populate the TableSource
+                     with. Ex. 'markers', 'forces'.
+
+    \throws KeyNotFound If table provided does not have column-labels.        */
+    TableSource_(const std::string& filename,
+                 const std::string& tablename) {
+        constructProperties();
```
aseth1: In many cases a file will contain only one table, but it is necessary to provide a tablename. Perhaps a default empty string should be provided and in the case that their are more tables throw if the specific table name was not provided.

OpenSim/Common/TableSource.h
```diff
@@ -65,6 +65,15 @@ class TableSource_ : public Component {
     OpenSim_DECLARE_CONCRETE_OBJECT_T(TableSource_, ET, Component);
 
 public:
+    OpenSim_DECLARE_PROPERTY(filename, std::string,
+                             "Path to the file to populate the TableSource_ "
+                             "with. The path is relative to the working "
```
aseth1: underscore is not necessary in the description intended for users.

OpenSim/Common/TableSource.h
```diff
@@ -95,21 +121,52 @@ class TableSource_ : public Component {
...
+    'filename' using `upd_filename()` to empty string and call this function.
 
-    \throws KeyNotFound If table provided does not have column-labels.        */
+    \throws InvalidCall If property `filename` is set. This call is not allowed
+                        if `filename` property is set.                        
+    \throws KeyNotFound If table provided does not have column labels.        */
     void setTable(const Table& table) {
+        OPENSIM_THROW_IF(!get_filename().empty(),
+                         InvalidCall,
+                         "Property 'filename' is set. Cannot set Table now.");
```
aseth1: What is the rationale for throwing here if you want to reset the underlying table? Why not empty the filename property automatically OR require a filename in setTable?

OpenSim/Common/TableSource.h
```diff
@@ -95,21 +121,52 @@ class TableSource_ : public Component {
...
+
+    \throws InvalidCall If the property 'filename' is empty.
+    \throws KeyNotFound If table provided does not have column-labels.        */
+    void setTable() {
+        OPENSIM_THROW_IF(get_filename().empty(),
+                         InvalidCall,
+                         "Property 'filename' is empty.");
+
+        setTable_impl(TimeSeriesTable_<ET>{get_filename(), get_tablename()});
+    }
```
aseth1: The is only valid/necessary inside `extendFinalizeFromProperties`, so why not make this call directly there? What is the extra indirection buying here?

OpenSim/Common/Test/TestTableSource.osim
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<OpenSimDocument Version="30500">
+	<TheWorld name="TestTableSource">
+          <components>
+            <TableSource__Vec3_ name="tablesource">
+              <filename>aStaticTrial_2.c3d</filename>
```
aseth1: We should make it's tag serialize as `<TableSourceVec3>`

aseth1:
@klshrinidhi Looks good. I just had a few comments/questions above.

OpenSim/Common/TableSource.h
chrisdembia: I like this check here. I think the user may be doing something wrong if they set the table programmatically and in XML. It may be an accident; which table did they really want to use? If they didn't want to use the one specified in XML, then the filename should be empty. One option is to have a second argument to this method, like `bool overrideIfFilenameGiven`.

OpenSim/Common/TableSource.h
chrisdembia: Could you explain your comment more? Maybe @klshrinidhi understands but I don't think I do.

OpenSim/Common/TableSource.h
chrisdembia: +1.

OpenSim/Common/Test/TestTableSource.osim
chrisdembia: This tag results from the auto-generated implementation of `getConcreteClassName()`. 

How do you do this? `Object::renameType("TableSource__Vec3__", "TableSourceVec3")`?

OpenSim/Common/TableSource.h
aseth1: I am not sure I agree. I read a Model with a TableSource and it has an associated filename. I run that simulation (with Table read in from file) and decide to change the Table and run it again. I am making a conscious and deliberate action to change it.  I would find the exception surprirsing and the extra step to clear the property  annoying.  If the property is not longer in sync with the table then the Component should clear the property. I am telling it which Table to use!

It is also inconsistent with other Components. If I set the frame for a Joint I don't get an exception that the Joint I am setting is conflict with the property, instead the property is updated to represent what I am setting.

OpenSim/Common/TableSource.h
aseth1: Sorry for being unclear. We do not need this  `setTable()` method IMO. It only serves  `extendFinalizeFromProperties` so why not call the one liner to load it. We are already checking there that the property is not empty.

OpenSim/Common/TableSource.h
chrisdembia: Okay I agree with you now. So long as the documentation of the method is really clear about what happens to the property.

OpenSim/Common/TableSource.h
chrisdembia: Oh okay; thanks for clarifying..

OpenSim/Common/TableSource.h
klshrinidhi: Added another overload of the constructor that takes only 'filename'.

OpenSim/Common/TableSource.h
klshrinidhi: Removed underscore.

OpenSim/Common/TableSource.h
klshrinidhi: I have changed the functions to not throw. I think the only argument for it to be consistent with existing operations.

I am in favor of throwing an exception during a call to read from a filename when a value for filename already exists, instead of resetting the filename during the operation. 
* It can lead to mistakes (as Chris points out). 
* It also makes the code unreadable because a reader has to go through the documentation of every operation to know if properties are being changed. If code includes an explicit reset of the filename, then reader will can see clearly.
* As a result of above, debugging becomes harder if one desires to track a property.

aseth1:
Changes LGTM.

I am still interested in making `TableSource` serialize like `<TableSource>` for doubles and  `<TableSourceVec3>`,   `<TableSourceSpatialVec>`, ... for other types.  I think we could use `Object::renameType()` upon deserialization to interpret  `<TableSourceVec3>` XML as  a `TableSource__Vec3_`, but it doesn't solve the serialization which would still write out  `<TableSource__Vec3_)>`, I think. @aymanhab is that right? Any other tricks we could do here?

Alternatively, we would have to derive the common types from `TableSource_` so they have the "right" concrete class name. They would be virtually implementation free except to invoke the `OpenSim_DECLARE_CONCRETE_OBJECT` macro.

chrisdembia:
> They would be virtually implementation free except to invoke the OpenSim_DECLARE_CONCRETE_OBJECT macro.

A little more complicated than that: you'd also have to duplicate the constructors. You could also use `using TableSource_<double>::TableSource_<double>;` to inherit the constructors, but then swig won't parse this correctly, and we'd need to create the constructors for swig.

chrisdembia:
The tag name for serialization is a little separate from this PR. I think if this PR is updated so that the `osim` file is not committed into the repo with the undesirable tag names (instead, the `osim` file can be created on the fly inside the test), then we can open a separate issue for the tag names.

aymanhab:
I would think rename type can go a long way toward this since it would map a nicely formatted name to a narly complicated/templatized type. True this's on deserialization only unless we instantiate the class and have its getConcreteClassName method return the nice name. Conjectured and untested of course.

chrisdembia:
Oh interesting...perhaps we can have a new macro `OpenSim_DECLARE_CONCRETE_CLASS` that can take an alternative name to be returned by `getConcreteClassName()`.

aseth1:
> but then swig won't parse this correctly, and we'd need to create the constructors for swig.

And this is also the case now, correct? You cannot use SWIG to construct `TableSource_<double>` for example so `TableSource` is not accessible via MATLAB of Python scripting. It seems then we need the derived classes then anyways.

aseth1:
> if this PR is updated so that the osim file is not committed into the repo with the undesirable tag names (instead, the osim file can be created on the fly inside the test), then we can open a separate issue for the tag names

:+1: agreed.

chrisdembia:
This PR introduces the use of the c3d adapter in testComponentInterface, which is only available if one compiles with BTK support. Could we use `#ifdef WITH_BTK` guards around the test that uses the c3d adapter?

klshrinidhi:
* I don't think removing the file `TestTableSource.osim` and generating it on the fly adds any value. It just adds more work for this PR in my opinion. I think having a physical and reading it during the test is easier to maintain. I have not made any changes regarding this. Future PR can easily edit the file to change the tags appropriately.

* I have changed the file `TestTableSource.osim` to use a TRC file instead of a C3D file. This is to make sure `TableSource` is tested regardless of whether C3D support is available.

aseth1:
>I don't think removing the file TestTableSource.osim and generating it on the fly adds any value. It just adds more work for this PR in my opinion. I think having a physical file and reading it during the test is easier to maintain. I have not made any changes regarding this. Future PR can easily edit the file to change the tags appropriately.

Having the test case write the TableSource file is one line of code which you likely used to write the file in the first place. The value is to decouple the otherwise approved operation of the TableSource from how its tags serialize, which we are happy to consider as a separate issue (PR). This would avoid requiring you to update that file as you revise the tag format. Serializing and deserializing in the test also ensures that the TableSource is self consistent with the latest version (and why update from a format that has yet to be specified?). Otherwise you have to have to update the test file as you test and in the repo. I think it would be less work to do what @chrisdembia recommends.

> I have changed the file TestTableSource.osim to use a TRC file instead of a C3D file. This is to make sure TableSource is tested regardless of whether C3D support is available.

:+1:

klshrinidhi:
I thought you meant having the xml file as string in the source file. I didn't realize you mean to serialize (write to file) and de-serialize (read from file) the model and check with the model in memory that was serialized in the first place. Currently I think the test ensures both serialization and de-serialization work.

> Having the test case write the TableSource file is one line of code which you likely used to write the file in the first place. 

I actually manually typed out the osim file with Chris's help. I did not generate it. It was very simple to type the file.

> The value is to decouple the otherwise approved operation of the TableSource from how its tags serialize, which we are happy to consider as a separate issue (PR). 

I don't understand. All the operations supported by TableSource are being tested separately and are decoupled from serialization test already. 

> This would avoid requiring you to update that file as you revise the tag format. 

I don't think editing tag names is much work. It is just search and replace.

> Serializing and deserializing in the test also ensures that the TableSource is self consistent with the latest version (and why update from a format that has yet to be specified?). Otherwise you have to have to update the test file as you test and in the repo. I think it would be less work to do what @chrisdembia recommends.

Currently I think the test ensures that TableSource can be both serialized and de-serialized. I don't think it is much work to edit the file when needed -- just search and replace. It is meant to be a test file anyway.

I don't know how to do what you are asking. I will try to get it done soon.

klshrinidhi:
Removed the test file.

aseth1:
>I actually manually typed out the osim file with Chris's help. I did not generate it. It was very simple to type the file

>Currently I think the test ensures that TableSource can be both serialized and de-serialized. I don't think it is much work to edit the file when needed -- just search and replace. It is meant to be a test file anyway.

If you hand write and update the file then you are not testing that the `TableSource` can deserialize from what it serialized.  Anyhow, `testComponents` will check that. You should add `TableSource` to the list of components in `testComponents` as well.

klshrinidhi:
Like I mentioned before, I think the test I had previously ensures both serialization and de-serialization.

chrisdembia:
I think we should merge this when the tests pass. TableSource is sufficiently tested in this PR. Adding this class to `testComponents` can be done later. Adding another commit to this PR will cause us to wait for CI and lengthen the CI queue and delay everyone's PRs.

aseth1:
@chrisdembia I was going to merge when tests pass.

Issue #876, opened 2016-03-22
Hackathon Feedback => Developer Week Example and Lecture Materials TODOs


aseth1:
Initial comments by @jenhicks but please feel free to add!

**Example Issues**
- [x] Get example working consistently in release and debug on all supported platforms: "it was a bummer that the example didn't work correctly for a lot of us" (PR #894) 

- [x] Would be helpful to have an overview of the model and functions used in the example. It was hard to get oriented to the various functions. It could even be something that participants draw themselves with some guidance. If we had to draw a picture of the connectors, inputs/outputs, and frames work before trying to write code (or reviewed one on slides, or a handout), it would really help reinforce the concepts.

- [ ] It would be cool to use Luxo geometry, if we can (if not the full mechanism).   
- [x] Include instruction to run the example target after building OpenSim so we "Have participants test getting the example running  beforehand"
- [ ] The mix of pointers and references can be confusing. Depending on audience, may want a primer or cheat-sheet for those whose primarily language isn't C++
- [x]  Example requires a lot of code before compiling -- would be helpful to break it into smaller pieces. 
- [x] Split the code into separate files (device, controller, etc.) with meaningful comments at the top of each to help organize. Would also reduce the need to scroll around when showing answers to the class (which is difficult to follow). (Tom)
- [x] Include more information in the "TODO" comments (e.g., what code needs to be added, why we need to add it, what the code will do). (Tom)
- [x] Brace "TODO" code blocks in the answer sheet; collapse blocks beforehand and expand to reveal answers to the class. This would avoid copy-and-paste errors and allow you to concentrate on explaining things. (Tom)

**Notes on the lecture** (many of them are minor style comments):

- [x]  Conceptual overview slide can be simplified: "does it need to be 3d? Hard to read some parts of the figure"
- [x]  keep in the challenges of current OpenSim -- people can relate to them
- [ ]  Reusability concept: can you give a better example of meaning in OpenSim (or at least for coding)
- [ ] Overview component diagram: maybe use colors or some other mechanism to give the notion that inputs/outputs have different types?
- [x] Connectors (detail slide that followed): rotate so it's the same orientation as the intro figure (to reinforce that connectors and inputs/outputs are different)
- [x] Would be helpful to give an overview of new frames at some point; lecture assumed people knew what these were
- [ ] Can we provide a guide/table to help people different stages? Maybe on cheat-sheet?
- [ ] Clarify what is meant by "Sources and reporters" -> Flip one
[...]

PR #1002, opened 2016-06-13
[WIP] Updates to hopper example for Dev Week

tkuchida:
(Not ready for review)
Addresses part of #876

### DIFF
--- a/OpenSim/Sandbox/ExampleHopperDevice/BouncingLeg.osim
+++ b/OpenSim/Sandbox/ExampleHopperDevice/BouncingLeg.osim
@@ -10,7 +10,7 @@
                     <connectors>
                         <Connector_Frame_ name="frame">
                             <!--Name of the component this Connector should be connected to.-->
-                            <connectee_name>ground</connectee_name>
+                            <connectee_name>../</connectee_name>
                         </Connector_Frame_>
                     </connectors>
                     <!--Default appearance for this Geometry-->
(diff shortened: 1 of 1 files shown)


aymanhab:
@tkuchida I worry about checking updated .osim files that have most up to date contents when the document version number at the top of the file is not updated as this may mess up future conversion. If we're close to done with xml changes, I'd update the XML document version number before saving so we can see what model files go through the version upgrade successfully.

tkuchida:
@aymanhab I don't understand your concern. The version number was originally set to 30505 for the hackathon, but changes were made to the way models are defined (in particular, adding frames to joints) without incrementing the version. Thus, the version number in this file was 30505 (and, consequently, `updateFromXMLNode()` was not triggered) but the file did not conform to the 30505 standard. I manually changed the version number to 30504, loaded it, then saved it. What would you like me to do instead?

aymanhab:
@tkuchida My concern is that version 30505 is a moving target now that we agreed upon (mainly @aseth1, @chrisdembia  and I) because so many changes are occurring with the understanding that we will not commit or honor files written in this continuously changing version. The invariant we try to keep is that the version number at the top of the file is consistent with the code that wrote it. I think we're close enough to the devweek that we can change the number in XMLDocument.cpp to be 30506 and let the conversion code kick in rather than subvert it by changing the header manually only to be surprised later.

tkuchida:
> I think we're close enough to the devweek that we can change the number in XMLDocument.cpp to be 30506 and let the conversion code kick in rather than subvert it by changing the header manually only to be surprised later.

That sounds fine, but please discuss on a separate thread. For now, I'm just trying to get the example working again on master so I can start updating it.

tkuchida:
Addressing XML document number first at @aymanhab's request.

PR #1004, opened 2016-06-14
Cleanup of ContactGeometry classes.

chrisdembia:
This PR gets rid of unnecessary copy constructors in ContactGeometry classes, and also replaces old-style properties with the new property system. 

Also, ContactGeometry constructors now take frames by const reference (instead of non-const ref).

Should be a quick review.


OpenSim/Simulation/Model/ContactSphere.h
```diff
@@ -37,11 +36,9 @@ class OSIMSIMULATION_API ContactSphere : public ContactGeometry {
 OpenSim_DECLARE_CONCRETE_OBJECT(ContactSphere, ContactGeometry);
 
 //=============================================================================
-// DATA
+// PROPERTIES
 //=============================================================================
-protected:
-    PropertyDbl _radiusProp;
-    double& _radius;
+    OpenSim_DECLARE_PROPERTY(radius, double, "Radius of the sphere.");
```
tkuchida: Not critical, but some property comments mention the default value. Might be particularly helpful in this case because the default value is zero (so the user won't see anything unless the radius has been set).

tkuchida:
Just one suggestion. Otherwise, LGTM and IMO can be merged if tests pass. :snowflake:

chrisdembia:
I have addressed your comment. Thank you for the review.

OpenSim/Simulation/Test/testContactGeometry.cpp
```diff
@@ -549,7 +549,7 @@ void testIntermediateFrames() {
 
         // Scaffolding for the ball.
         auto* linkOffset = new PhysicalOffsetFrame("link_offset",
-                model.updBodySet().get("point"),
+                model.getBodySet().get("point"),
```
aseth1: could replace this with `model.getComponent<Body>("point")` since you are updating anyways

aseth1:
LGTM. Did you verify that you still need all includes? e.g.:
```cpp
 #include <fstream>
 #include <OpenSim/Common/IO.h>
```

chrisdembia:
@aseth1 I think perhaps you should just merge this. We can do a sweep through the repo later to remove the use of sets. And those includes you mentioned are needed.

aseth1:
@chrisdembia :a::ok:

chrisdembia:
Thank you.

PR #1005, opened 2016-06-14
Geometry 4 wrap objects

aymanhab:
This PR is basically a refactoring /no-op to move custom visualization code from ModelVisualiser to proper generateDecorations methods of classes (Wrap{Sphere, Cylinder,Ellipsoid} and ContactSphere). WrapTorus was never implemented so an exception/place holder is introduced and thrown. With this we can test if WrapObject(s) and ContactSpheres show up in the GUI/Visualizer. Later PRs will address:
- Parts of Analytical shapes (Quadrants etc. that were not handled in the past).
- WrapTorus
- ContactHalfSpace
- ContactMesh.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/1005)


aymanhab:
Ready for review. @chrisdembia and/or @aseth1 no functionality change just moving visualization code where it belongs so that GUI and API visualizer behave the same.

chrisdembia:
Looks good to me. I like the new `NotImplementedYet` exception (and its optimistic tone) but I'm not sure we need to use it for WrapTorus. I think it's not such a fatal thing, compared to e.g. an actuator not providing its power output. Perhaps it's fine to just document in the WrapTorus class that visualization is not provided for it.

aymanhab:
Thanks @chrisdembia fro the prompt review. Agreed that not all unsupported/unimplemented options are critical, so I'll remove the throw line and document in the .cpp since this's temporary. Thanks.

aymanhab:
@chrisdembia testContactGeometry failed due to timeout in debug mode on Travus/Clang, does that sound like a fluke? I couldn't force a rebuild.

chrisdembia:
Yeah that's been happening on a lot of builds. I tested locally and testContactGeometry took only 255 seconds in Debug. Maybe it's an Ubuntu issue? @klshrinidhi would you be able to try running testContactGeometry in debug from latest master on your computer?

tkuchida:
> Yeah that's been happening on a lot of builds.

testContactGeometry also failed due to timeout in #990 (only on Travis in Debug mode). I restarted and [it passed in 1143 seconds](https://travis-ci.org/opensim-org/opensim-core/jobs/137678410#L1937).

aymanhab:
Thanks @tkuchida. Please let me know if there's a reason not to merge now @chrisdembia.

chrisdembia:
LGTM but @aseth1 may want to take a look.

OpenSim/Common/Exception.h
```diff
@@ -266,6 +266,11 @@ class IOError : public Exception {
     using Exception::Exception;
 };
 
+class NotImplementedYet : public Exception {
+public:
```
aseth1: This is a generally useful `Exception` type and should:
1. be named more accurately, such as: `MethodNotImplemented` or `MethodNotOverriden` or `OverrideNotImplemented`
2. construct with the method name and format the message consistent with existing exceptions in `Exception.h`

OpenSim/Simulation/Wrap/WrapCylinder.cpp
```diff
@@ -32,7 +33,7 @@
 #include <OpenSim/Common/SimmMacros.h>
 #include <OpenSim/Common/Mtx.h>
 #include <sstream>
-
+#include "OpenSim/Simulation/Model/ModelDisplayHints.h"
```
aseth1: If adding headers are there some that can be removed? E.g. `SimmMacros.h`, `Mtx.h`, ...

OpenSim/Simulation/Wrap/WrapCylinder.cpp
```diff
@@ -800,3 +801,23 @@ bool WrapCylinder::_adjust_tangent_point(SimTK::Vec3& pt1,
...
+// in ModelVisualizer
+void WrapCylinder::generateDecorations(bool fixed, const ModelDisplayHints& hints, const SimTK::State& state,
+    SimTK::Array_<SimTK::DecorativeGeometry>& appendToThis) const 
+{
+    
+    if (hints.get_show_wrap_geometry()) {
+        const Vec3 color(SimTK::Cyan);
+        SimTK::Transform ztoy;
+        ztoy.updR().setRotationFromAngleAboutX(SimTK_PI / 2);
+        const SimTK::Transform& X_GB = getFrame().getMobilizedBody().getBodyTransform(state);
```
aseth1: This is not correct. The `MobilizedBody` and a `PhysicalFrame` do not, in general, have the same transform w.r.t. Ground - can be a `PhysicalOffsetFrame` for example.  You need to call `Frame:: getTransformInGround(state)` instead.

OpenSim/Simulation/Wrap/WrapCylinder.cpp
```diff
@@ -800,3 +801,23 @@ bool WrapCylinder::_adjust_tangent_point(SimTK::Vec3& pt1,
...
+// Implement generateDecorations by WrapCylinder to replace the previous out of place implementation 
+// in ModelVisualizer
+void WrapCylinder::generateDecorations(bool fixed, const ModelDisplayHints& hints, const SimTK::State& state,
+    SimTK::Array_<SimTK::DecorativeGeometry>& appendToThis) const 
+{
+    
+    if (hints.get_show_wrap_geometry()) {
+        const Vec3 color(SimTK::Cyan);
+        SimTK::Transform ztoy;
+        ztoy.updR().setRotationFromAngleAboutX(SimTK_PI / 2);
```
aseth1: Would be good to add a comment for why this Rotation offset is necessary, otherwise it seems tempting to get rid of it.

OpenSim/Simulation/Wrap/WrapEllipsoid.cpp
```diff
@@ -1294,3 +1296,22 @@ double WrapEllipsoid::closestPointToEllipse(double a, double b, double u,
...
+void WrapEllipsoid::generateDecorations(bool fixed, const ModelDisplayHints& hints, const SimTK::State& state,
+    SimTK::Array_<SimTK::DecorativeGeometry>& appendToThis) const 
+{
+
+    if (hints.get_show_wrap_geometry()) {
+        const Vec3 color(SimTK::Cyan);
+        SimTK::Transform ztoy;
+        ztoy.updR().setRotationFromAngleAboutX(SimTK_PI / 2);
+        const SimTK::Transform& X_GB = getFrame().getMobilizedBody().getBodyTransform(state);
+        SimTK::Transform X_GW = X_GB*getTransform()*ztoy;
```
aseth1: Same as above - use `Frame:: getTransformInGround(state)`

OpenSim/Simulation/Wrap/WrapSphere.cpp
```diff
@@ -589,3 +591,23 @@ int WrapSphere::wrapLine(const SimTK::State& s, SimTK::Vec3& aPoint1, SimTK::Vec
...
+// Implement generateDecorations by WrapSphere to replace the previous out of place implementation 
+// in ModelVisualizer
+void WrapSphere::generateDecorations(bool fixed, const ModelDisplayHints& hints, const SimTK::State& state,
+    SimTK::Array_<SimTK::DecorativeGeometry>& appendToThis) const 
+{
+    if (hints.get_show_wrap_geometry()) {
+        const Vec3 color(SimTK::Cyan);
+        SimTK::Transform ztoy;
+        ztoy.updR().setRotationFromAngleAboutX(SimTK_PI / 2);
+        const SimTK::Transform& X_GB = getFrame().getMobilizedBody().getBodyTransform(state);
```
aseth1: Same as above - use `Frame:: getTransformInGround(state)`

aseth1: Should also consider having the base Wrap object providing a helper to place the decorative geometry since the code is identical for three type implemented so far. All that differs is the creation of the Decorative shape and seems like that's all the derived classes should specify. I imagine Torus would then follow the same pattern

aseth1:
@aymanhab I am done my review. Let me know if you want to discuss.

OpenSim/Common/Exception.h
aymanhab: Thanks @aseth1 There are multiple use cases, some could be for a branch in the code inside a method and others where a full method isn't implemented or overriden. The first case is handled by the base class as is. For the sake of keeping this PR as narrow as possible. I undid this change and will have a separate PR for it since I ended up not using it anyway.

OpenSim/Simulation/Wrap/WrapCylinder.cpp
aymanhab: These headers are actually needed.

aymanhab:
Thanks for the prompt review @aseth1 I modified the Frame method call as you suggested (thanks for the find) though it's hard to test since we have no models with WrapObjects on offset frames. I refrained from any other refactoring to make this easily reviewable/mergeable so I can start the meat of the work described up top.

tkuchida:
> it's hard to test since we have no models with WrapObjects on offset frames

There is a WrapCylinder in the hopper example that must be attached to a PhysicalOffsetFrame.

aymanhab:
Perfect, is the file checked in somewhere or should I run the example?

tkuchida:
> Perfect, is the file checked in somewhere or should I run the example?

I'm still working on it (locally); hope to have it finished today. I'll make a PR and @mention you when it's ready.

aymanhab:
Ready for review/merge,  tests pass , feedback addressed.

tkuchida:
:fireworks: I pulled this branch and it fixes a problem I have been struggling to resolve since yesterday (attaching a `WrapCylinder` to a `PhysicalOffsetFrame`)! My hypothesis is that everything was being connected correctly but the visualizer was still drawing the cylinder at the body's origin.

One issue: The visualizer is drawing two cylinders when only one has been added to the model: one is drawn in the initial location and one moves with the Frame. Any ideas?

aymanhab:
Thanks Tom.

Method to generate decorations is called twice once with fixed true, then
again with fixed set to false. Need to check on top so we draw in only one
of the two calls.

tkuchida:
> Method to generate decorations is called twice once with fixed true, then again with fixed set to false.

Thanks, @aymanhab. Adding `if (fixed) { return; }` to the top of `WrapCylinder::generateDecorations()` (to match `ContactSphere::generateDecorations()`) worked. It looks like the same change needs to be made in `WrapEllipsoid::generateDecorations()`, and perhaps elsewhere.

aymanhab:
Sure, will do. Thanks

aymanhab:
Pushed a fix @tkuchida Thanks for testing.

tkuchida:
[testComponents failed on AppVeyor x64](https://ci.appveyor.com/project/opensim-org/opensim-core/build/3758/job/0jwa5frbelovsxdb#L1479) ("BushingForce.initSystem() increased memory use by 1.1e+12%."). Restarted…

aymanhab:
All tests passed. Ready to merge IMHO.

tkuchida:
b04cd72 looks good. You have my :+1: but perhaps @aseth1 should confirm/merge since he raised some issues above.

OpenSim/Simulation/Wrap/WrapCylinder.cpp
```diff
@@ -800,3 +801,29 @@ bool WrapCylinder::_adjust_tangent_point(SimTK::Vec3& pt1,
...
+    Super::generateDecorations(fixed, hints, state, appendToThis);
+    if (fixed) return;
+
+    if (hints.get_show_wrap_geometry()) {
+        const Vec3 color(SimTK::Cyan);
+        SimTK::Transform ztoy;
+        // Make transform that takes z axis to y axis due to different
+        // assumptions between DecorativeCylinder aligned with y  and
+        // WrapCylinder aligned with z
+        ztoy.updR().setRotationFromAngleAboutX(SimTK_PI / 2);
```
aseth1: This Rotation rotates Y to Z!  You get the correct effect because you post multiply by the transform.

OpenSim/Simulation/Wrap/WrapEllipsoid.cpp
```diff
@@ -1294,3 +1296,25 @@ double WrapEllipsoid::closestPointToEllipse(double a, double b, double u,
...
+    SimTK::Array_<SimTK::DecorativeGeometry>& appendToThis) const 
+{
+
+    Super::generateDecorations(fixed, hints, state, appendToThis);
+    if (fixed) return;
+
+    if (hints.get_show_wrap_geometry()) {
+        const Vec3 color(SimTK::Cyan);
+        SimTK::Transform ztoy;
+        ztoy.updR().setRotationFromAngleAboutX(SimTK_PI / 2);
```
aseth1: I understand the reason for rotating the Cylinder but why the Ellipsoid? Are the radii for the WrapEllipsoid (rx, ry, rz)  not consistent with the DecorativeEllipsoid radii? The old ModelVisualizer code was not doing this either, which makes think this might be copy and paste bug.

OpenSim/Simulation/Wrap/WrapSphere.cpp
```diff
@@ -589,3 +591,27 @@ int WrapSphere::wrapLine(const SimTK::State& s, SimTK::Vec3& aPoint1, SimTK::Vec
...
+
+    Super::generateDecorations(fixed, hints, state, appendToThis);
+    if (fixed) return;
+
+    if (hints.get_show_wrap_geometry()) {
+        const Vec3 color(SimTK::Cyan);
+        SimTK::Transform ztoy;
+        ztoy.updR().setRotationFromAngleAboutX(SimTK_PI / 2);
+        const SimTK::Transform& X_GB = getFrame().getTransformInGround(state);
+        SimTK::Transform X_GW = X_GB*getTransform()*ztoy;
```
aseth1: I am pretty sure the ztoy transform is unnecessary for a Sphere!

aymanhab:
Thanks for the catch @aseth1 :+1: Indeed copy/paste bug.  
May need to introduce transforms to handle parts of Spheres/Cylinders/Torii later depending on how the primitives on the visualizer side are parameterized, for now keeping functionality exactly as in old ModelVisualizer..

aymanhab:
Tests pass minus bogus failure in testComponents.

aymanhab:
Awesome, thanks @aseth1 :+1:
On Jun 18, 2016 1:27 PM, "Ajay Seth" <notifications@github.com> wrote:

Merged #1005 <https://github.com/opensim-org/opensim-core/pull/1005>.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/opensim-org/opensim-core/pull/1005#event-696808240>,
or mute
the thread
<https://github.com/notifications/unsubscribe/ADyccEgQiAm7uIxHQqV-SDvmK2f3l3dCks5qNFS9gaJpZM4I1kjE>
.

PR #1011, opened 2016-06-15
Silence clang warning about static function.

chrisdembia:
Removes `static` from the `randomize()` function in `auxiliaryTestFunctions.h` to silence a warning from clang.

### DIFF
--- a/OpenSim/Auxiliary/auxiliaryTestFunctions.h
+++ b/OpenSim/Auxiliary/auxiliaryTestFunctions.h
@@ -122,7 +122,7 @@ do { \
     } \
 } while(false) 
 
-static OpenSim::Object* randomize(OpenSim::Object* obj)
+OpenSim::Object* randomize(OpenSim::Object* obj)
 {
     using namespace OpenSim;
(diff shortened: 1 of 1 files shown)


chrisdembia:
Should be a quick and easy review! Any takers?

tkuchida:
Seems innocuous enough, but why is `static` wrong here?

chrisdembia:
Here, `static` means the function should not be available outside the translation unit being compiled (the cpp file and everything it includes). So putting `static` on a function in a header file is kinda weird, since multiple cpp files could include it, and then it's not available in only one translation unit. I may have introduced this issue. This function was originally in a cpp file, I think.

tkuchida:
Ah, ok. Thanks.

klshrinidhi:
:+1: 
Can be merged I think.

chrisdembia:
:) would you like to do the honors?

tkuchida:
Merging. Thanks @chrisdembia and @klshrinidhi for the lesson about the `static` keyword :smiley:

Issue #974, opened 2016-05-13
Replace reference to Sets during model building with Iterators


aymanhab:
Per dev team meeting, this is needed for correct wiring of nested models. @aseth1 will specify what specific functions are considered "model building".

chrisdembia:
I found an instance of using `updContactGeometrySet()` in `ElasticFoundationForce::extendAddToModel()`. It's probably used in HuntCrossleyForce also.

jenhicks:
Given discussion of support for handling models with hierarchical components, we need to decide if this task is in or out for 4.0. Leaning towards hold for post-4.0

chrisdembia:
I thought that we decided we would support putting any component (including Bodies, Joints, Constraints) in any other component for 4.0, but that we would not support putting a Model in another component. 

The former still requires replacing the use of Sets internally (mostly in Model) with iterators.

We could have a more nuanced situation for 4.0, where we say something like "we support putting arbitrary components inside other components, except for ContactGeometry and Model". In this case, ElasticFoundationForce could still use Model::updContactGeometrySet().

aymanhab:
Thanks @chrisdembia the question that was raised was whether having a Body inside a Joint (e.g. Body internal to the shoulder joint) should cause that Body to show in the Bodies list on navigator, topology-view or Tools that operate on Bodies. If this internal Body shouldn't be changed and is completely internal to the Joint then using the Sets will be ok, or switching to iterator if the iterator knows how to handle it. Still up for discussion as far as I can tell.

chrisdembia:
Thanks for the clarification. I think I understand now.

To support "internal" subcomponents, we introduced "member subcomponents." These are subcomponents that a user cannot remove or edit, and we have two functioning member subcomponents in Thelen2003Muscle. The idea is that Thelen2003Muscle gets to choose how much to expose of the internal components. Right now, I am almost positive the iterator *does* iterate through these subcomponents. But IMO the iterator should *not* iterate through member subcomponents. I'm not really sure though!

Issue #914, opened 2016-04-10
API is littered with `add`, `attach`, `adopt` methods with subtle differences


aseth1:
that results in confusion. Friedl ran into this a lot at the hackathon not knowing when to use: `addComponent()` , `addGeometry` or `attachGeometry` to include geometry to visualize her knee model. Similarly for adding a `Frame`, we have: `addComponent()`, `addModelComponent`, `addFrame`. 

We (temporarily) made the problem worse because we enabled any Component to be its own composite (with its own methods/list to build composites) which is a good thing, but Model (and intermediaries like ModelComponent) Set based methods still persist.

It would simplify the API immensely if we only had `Component::adopt(Component* comp)` with the understanding of taking over ownership of `comp`. Since all the add/attach Geometry methods are new they could be reworked and the Model's `Set` specific ('addBody', 'addJoint', etc...) can be deprecated and aliased to call `adopt`. Derived Components could implement `extendAdopt` so that `Model`, for example, could move the component types it wanted out of the generic property list of Components into type specific lists or even into `Sets` to preserve the current serialization format.

From that point it would also be easier to eliminate Sets altogether but would not be necessary for 4.0.

As part of the work/process of updating Components in 4.0 we would also implement `extendAdopt` accordingly and convert to using `ComponentList<C>` instead of `Set<T>` and its derivatives.

PR #1014, opened 2016-06-16
Enable Component::addComponent() to replace type specific Model::add###() methods

aseth1:
This PR partially addresses issue #914, #974 and other related concerns by making it possible to assemble the same `System` from the model being assembled by `Component::addComponent()` instead of component specific `Model::add###()`. Since a `Body`, for example, is a `Frame`, a `ModelComponent`, and a `Component`, it is unclear which method from `addBody()`, `addFrame()`, `addModelComponent()`, `addComponent()` should a user use? Having individual `addFrame()`, `addBody()`, `addJoint()`, ... methods is also not sustainable with new Component types. By making the base most `addComponent()` work we can eliminate the need for the other methods.

The main driver for the changes is testAddComponents, which is a reimplementation of the tug-if-war example (exampleMain) where all the `add###()` where replaced with `addComponent()`. The most significant changes where to update look-ups for dependencies from using Model's type specific `Set`s to `getComponentList<C>` to  traverse a list or `getComponent<C>()` to locate a component of type, `C`.

With these changes in place we still need to address these TODOs:
1. handle adding of Geometry (e.g. redundancy between `addGeometry()`, `attachGeometry()`, `attachGeometryMesh()`)
2. deprecate all the `add###()` methods so it is clear that `addComponent()` is the correct (long-term) way to add to your Components (including Model)

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/1014)


chrisdembia:
Here's my record of which types of components are tested in `testAddComponents`:

- [x] Body
- [x] Joint
- [x] ContactGeometry
- [x] Force
- [x] Controller
- [x] Constraint
- [x] Geometry
- [x] Frame
- [ ] Marker
- [ ] Probe
- [ ] ModelComponent

I think it may be fine if `ModelComponent` is not tested, but I think Marker and Probe should be tested (maybe an ActuatorPower probe?).

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-:-KKQaA6A0oN1Zg7usvxb:1870061881)*

chrisdembia:
*[OpenSim/Simulation/Model/ContactMesh.h, line 102 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQcHeUtlU3qF1Z8Ux8:-KKQcHeVhBrl8KYM2D6p:1449937922) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Simulation/Model/ContactMesh.h#L102)):*
> ```objc
>     void extendFinalizeFromProperties() override;
> 
>     //Load the mesh from disk.
> ```

I think it is fine for this method to have a doxygen comment. I think the comment should say that the returned object is heap-allocated and not owned.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
*[OpenSim/Simulation/Control/Controller.cpp, line 126 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQh-jeaDPTI1-AnasS:-KKQh-jeaDPTI1-AnasT:2118774334) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Simulation/Control/Controller.cpp#L126)):*
> ```cpp
>                         found = true;
>                         break;
>                     }
> ```

Consider one of the following:
- `getComponent<Actuator>(get_actuator_list(i))` (and catching the exception it throws if the component can't be found)
- creating a public `findComponent<Actuator>(std::string)`, similar to the existing `findComponent()`.
- using std::find(), as @klshrinidhi [has done before](https://github.com/opensim-org/opensim-core/commit/6e46bfebd71dffa1ceb94bda47a2a922114874a8#diff-42ab0181181a8c731e017a187770b3e6):

```cpp
auto iter = std::find_if(components.begin(),
                             components.end(),
                             [&] (const Body& body) {
                                 return body.getName() == _bodyName;
                             });
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
*[OpenSim/Simulation/Control/Controller.cpp, line 115 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQoY8jv7xByGpOqx3A:-KKQoY8jv7xByGpOqx3B:-2003758199) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Simulation/Control/Controller.cpp#L115)):*
> ```cpp
>             for (auto& actuator : actuators) {
>                 addActuator(actuator);
>             }
> ```

I think you should build an ActuatorSet in a loop and then call `setActuators()`. Otherwise, this might be something like O(n^2), since `addActuators()` loops through the properties.

And it actually is the wrong thing to do here, because `addActuator()` will add each actuator to the `actuators_list`, but the `actuators_list` should remain as `ALL`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
*[OpenSim/Simulation/Control/Controller.cpp, line 123 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQpI1p4ThE9m4YiFC3:-KKQpI1qrZfRnyoC-j5B:1854140140) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Simulation/Control/Controller.cpp#L123)):*
> ```cpp
>                 for (auto& actuator : actuators) {
>                     if (get_actuator_list(i) == actuator.getName()) {
>                         addActuator(actuator);
> ```

This might not be what you want here either; since you already *know* that `actuator` is in the `actuators_list`; you don't need the check that `addActuator()` does.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
*[OpenSim/Common/Component.h, line 407 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQrsaBcJr1Mnbd-eqM:-KKQrsaCtrYo_hvqUQ62:261603338) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Common/Component.h#L407)):*
> ```objc
>     * serialized as part of this component.
>     * If the addition is already owned by this component, an Exception
>     * is thrown.
> ```

What does "addition" mean? This is going to be one of the most important methods in the API now, probably more important than `initSystem()`. Could you add some more details here? Could you include the fact that this component will be added to the `components` property, and thus will appear in XML? And that one should use this method for all types of components (bodies, joints, etc.)? Perhaps mention that the resulting component is a "subcomponent" of the containing component?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
*[OpenSim/Common/Component.h, line 410 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQt01oHmaJdsLAJ8Ue:-KKQt01oHmaJdsLAJ8Uf:942597982) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Common/Component.h#L410)):*
> ```objc
>     *
>     * @throws ComponentAlreadyPartOfOwnershipTree  */
>     void addComponent(Component* addition);
> ```

FYI This method will almost certainly cause a crash in scripting, because of ownership issues. I can try to address this when I look into wrapping. I created #1017 to track this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
*[OpenSim/Common/Component.h, line 1177 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQtgT1EdJ1RxYjuhIm:-KKQtgT2BYWP1N4DL99h:2000998568) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Common/Component.h#L1177)):*
> ```objc
>     inserting it into a particular internal list (for grouping), after adding
>     an addition (Component) to this component.
> ```

Consider mentioning that one should really *not* to implement this method; it's really for special components like Model.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
Could you edit the doxygen for all the old `Model::add*()` methods to say to use `addComponent()` instead? I don't think we should use the `DEPRECATE_14` macro for them yet, since they are *so* widely used. But a doxygen comment is necessary, I think, to address Friedl's confusion.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-:-KKQuUPnMQZ4ztzsrFzm:571544744)*

chrisdembia:
*[OpenSim/Common/Component.cpp, line 149 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQvW08YOSJy2nFXtkN:-KKQvW08YOSJy2nFXtkO:-1932305640) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Common/Component.cpp#L149)):*
> ```cpp
>     finalizeFromProperties();
> 
>     // allow the concrete Component perform secondary operations
> ```

not necessarily concrete; another abstract could implement this method.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
*[OpenSim/Simulation/Model/Model.cpp, line 747 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQwn9Aj_LTK69BKbUp:-KKQwn9BxDCgUoVQllrM:2130955157) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Simulation/Model/Model.cpp#L747)):*
> ```cpp
> void Model::extendAddToSystem(SimTK::MultibodySystem& system) const
> {
>     Model *mutableThis = const_cast<Model *>(this);
> ```

Missing a call to `Super::extendAddToSystem(system)`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
*[OpenSim/Simulation/Model/Model.cpp, line 750 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQwqX7mtKEVQYiu57c:-KKQwqX7mtKEVQYiu57d:862493193) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Simulation/Model/Model.cpp#L750)):*
> ```cpp
> 
>     // Ensure Ground is added before all other Components
>     getGround().addToSystem(system);
> ```

I think I know why this is necessary (same bug we had before the hackathon?), but doesn't this mean Ground will get added to the system twice? Doesn't Component invoke `addToSystem()` on all propertySubcomponents?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
*[OpenSim/Simulation/Model/Model.cpp, line 746 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQxHvZEpa1N0Ct8roQ:-KKQxHvZEpa1N0Ct8roR:-1101166137) ([raw file](https://github.com/opensim-org/opensim-core/blob/17aba0e165b8e75b41e587559be83c454c6f2a8f/OpenSim/Simulation/Model/Model.cpp#L746)):*
> ```cpp
>     updFrameSet().invokeConnectToModel(*this);
>     updMarkerSet().invokeConnectToModel(*this);
>     updContactGeometrySet().invokeConnectToModel(*this);
> ```

Just double-checking: these are not needed since Model handles invoking `extendConnectToModel()` on all its property subcomponents? So, these components were being connected twice?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
*[OpenSim/Simulation/Model/Model.cpp, line 748 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQy8yrl1pEC6N1VFC9:-KKQy8yrl1pEC6N1VFCA:-971291593) ([raw file](https://github.com/opensim-org/opensim-core/blob/17aba0e165b8e75b41e587559be83c454c6f2a8f/OpenSim/Simulation/Model/Model.cpp#L748)):*
> ```cpp
>     updContactGeometrySet().invokeConnectToModel(*this);
>     updForceSet().setupGroups();
>     updControllerSet().setActuators(updActuators());
> ```

Are we sure this doesn't need any changes? What does the ControllerSet do with these actuators?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
*[a discussion](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQyuTyHFjEHUylSGQD:-KKQyuTyHFjEHUylSGQE:155418670) (no related file):*
I am done reviewing this PR. It was an easier review than I thought it would be! I think most of my comments are minor.  

My biggest question while reviewing was if/how `addComponent()` affected the sets. I think it should be documented somewhere (Component::getComponent, perhaps) that components added to the model with `getComponent` are *not* accessible from the API via the Sets in Model. A user might expect that `Model.addComponent(body)` would be accessible in `Model.getBodySet()`. 

It should be documented that there are two ways of adding components:

1. Model::add###()
2. Component::addComponent()

and two ways of accessing components:

a. Model::get###Set()
b. Component::getComponent() and Component::getComponentList()

and most importantly, that components added via (1) are available via (a) and (b), but components added via (2) are only available via (b).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
*[a discussion](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQzyB1TdDGHpTNJj9s:-KKQzyB1TdDGHpTNJj9t:247285314) (no related file):*
Could you add mention of this change to the CHANGELOG?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
Reviewed 22 of 22 files at r1.
Review status: all files reviewed at latest revision, 13 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
What happened to the history here? Did you do a rebase?

aseth1:
@chrisdembia I didn't rebase but I merged the branch with master.

aseth1:
*[OpenSim/Simulation/Model/Model.cpp, line 746 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQxHvZEpa1N0Ct8roQ:-KKu7S3P5-gOljLTmARZ:1721742657) ([raw file](https://github.com/opensim-org/opensim-core/blob/17aba0e165b8e75b41e587559be83c454c6f2a8f/OpenSim/Simulation/Model/Model.cpp#L746)):*
> Just double-checking: these are not needed since Model handles invoking `extendConnectToModel()` on all its property subcomponents? So, these components were being connected twice?
</details>
Yes that's right.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

aseth1:
*[OpenSim/Simulation/Model/Model.cpp, line 748 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQy8yrl1pEC6N1VFC9:-KKuH3L0wXMJGXJjGoZz:2002552169) ([raw file](https://github.com/opensim-org/opensim-core/blob/17aba0e165b8e75b41e587559be83c454c6f2a8f/OpenSim/Simulation/Model/Model.cpp#L748)):*
> Are we sure this doesn't need any changes? What does the ControllerSet do with these actuators?
</details>
It appears that the ControllerSet uses the Actuators to write controls output to storage. I will make sure that `updActuators` returns the complete set.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

aseth1:
*[OpenSim/Simulation/Model/Model.cpp, line 747 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQwn9Aj_LTK69BKbUp:-KKuSdJ8Vi3io_hj11bI:-499146009) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Simulation/Model/Model.cpp#L747)):*
> Missing a call to `Super::extendAddToSystem(system)`.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

aseth1:
*[OpenSim/Simulation/Model/Model.cpp, line 750 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQwqX7mtKEVQYiu57c:-KKuSiPlRczokMLoZzlS:-1150908314) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Simulation/Model/Model.cpp#L750)):*
> I think I know why this is necessary (same bug we had before the hackathon?), but doesn't this mean Ground will get added to the system twice? Doesn't Component invoke `addToSystem()` on all propertySubcomponents?
</details>
The issue is order. I need Ground to be first. I don't have a good way to ensure that other than doing it directly. It does call Ground::extendAddToSystem() twice, but this has no detrimental effect since it gets assigned the same MobilizedBodyIndex. I don't like it being called twice either, but at least it works. Happy to discuss a solution after the workshop. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

aseth1:
*[OpenSim/Common/Component.cpp, line 149 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQvW08YOSJy2nFXtkN:-KKugdv-1fGpNxr-0k7B:-499146009) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Common/Component.cpp#L149)):*
> not necessarily concrete; another abstract could implement this method.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

aseth1:
*[OpenSim/Common/Component.h, line 407 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQrsaBcJr1Mnbd-eqM:-KKugh9GhL6Il5xzFlp0:-604371748) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Common/Component.h#L407)):*
> What does "addition" mean? This is going to be one of the most important methods in the API now, probably more important than `initSystem()`. Could you add some more details here? Could you include the fact that this component will be added to the `components` property, and thus will appear in XML? And that one should use this method for all types of components (bodies, joints, etc.)? Perhaps mention that the resulting component is a "subcomponent" of the containing component?
</details>
reverted to subcomponent. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

aseth1:
*[OpenSim/Simulation/Control/Controller.cpp, line 126 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQh-jeaDPTI1-AnasS:-KKu1wVZeLoInv6FuKr3:867741285) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Simulation/Control/Controller.cpp#L126)):*
> Consider one of the following:
> - `getComponent<Actuator>(get_actuator_list(i))` (and catching the exception it throws if the component can't be found)
> - creating a public `findComponent<Actuator>(std::string)`, similar to the existing `findComponent()`.
> - using std::find(), as @klshrinidhi [has done before](https://github.com/opensim-org/opensim-core/commit/6e46bfebd71dffa1ceb94bda47a2a922114874a8#diff-42ab0181181a8c731e017a187770b3e6):
> 
> ```cpp
> auto iter = std::find_if(components.begin(),
>                              components.end(),
>                              [&] (const Body& body) {
>                                  return body.getName() == _bodyName;
>                              });
> ```
</details>
Tried this unsuccessfully. Reverted back to what I had. It ends up being O(n*m) in both cases.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

aseth1:
*[OpenSim/Simulation/Control/Controller.cpp, line 115 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQoY8jv7xByGpOqx3A:-KKu0vW7dJ2rVSqWrj7F:36051832) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Simulation/Control/Controller.cpp#L115)):*
> I think you should build an ActuatorSet in a loop and then call `setActuators()`. Otherwise, this might be something like O(n^2), since `addActuators()` loops through the properties.
> 
> And it actually is the wrong thing to do here, because `addActuator()` will add each actuator to the `actuators_list`, but the `actuators_list` should remain as `ALL`.
</details>
I agree. Mangling the properties here is wrong. In the future we should have a list Connector to Actuators so that his connection logic can go away and be handled by the Component machinery. I will make fix to not edit property in the meantime.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

aseth1:
*[OpenSim/Simulation/Control/Controller.cpp, line 123 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQpI1p4ThE9m4YiFC3:-KKugxCyRl5qLE-2d9FI:-499146009) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Simulation/Control/Controller.cpp#L123)):*
> This might not be what you want here either; since you already *know* that `actuator` is in the `actuators_list`; you don't need the check that `addActuator()` does.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

aseth1:
*[OpenSim/Common/Component.h, line 1177 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQtgT1EdJ1RxYjuhIm:-KKugqEJZDp7-rAP727S:-499146009) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Common/Component.h#L1177)):*
> Consider mentioning that one should really *not* to implement this method; it's really for special components like Model.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
*[OpenSim/Simulation/Model/Model.cpp, line 750 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQwqX7mtKEVQYiu57c:-KKvlKKIXSrHQaCjdhFK:1110214866) ([raw file](https://github.com/opensim-org/opensim-core/blob/b0c6c9ad5f441c65ff6a23002b130a5bd007f069/OpenSim/Simulation/Model/Model.cpp#L750)):*
> The issue is order. I need Ground to be first. I don't have a good way to ensure that other than doing it directly. It does call Ground::extendAddToSystem() twice, but this has no detrimental effect since it gets assigned the same MobilizedBodyIndex. I don't like it being called twice either, but at least it works. Happy to discuss a solution after the workshop. 
</details>
+1 after the workshop!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
*[OpenSim/Simulation/Model/Model.cpp, line 748 \[r1\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKQy8yrl1pEC6N1VFC9:-KKvm1i0lAcFB38mbo9D:-1888467195) ([raw file](https://github.com/opensim-org/opensim-core/blob/17aba0e165b8e75b41e587559be83c454c6f2a8f/OpenSim/Simulation/Model/Model.cpp#L748)):*
> It appears that the ControllerSet uses the Actuators to write controls output to storage. I will make sure that `updActuators` returns the complete set.
</details>
I created #1047 to track this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
*[OpenSim/Common/Component.h, line 432 \[r3\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKvmpKDsGUyErCMvv9C:-KKvmpKDsGUyErCMvv9D:-172707796) ([raw file](https://github.com/opensim-org/opensim-core/blob/c8595b73ef4a5448a7c9ced1eab8e513db365b62/OpenSim/Common/Component.h#L432)):*
> ```objc
>     *
>     * @throws ComponentAlreadyPartOfOwnershipTree
>     * @param subcomponent is the Component to be added. */
> ```

Excellent doxygen!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
Reviewed 24 of 171 files at r2, 1 of 2 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

chrisdembia:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-:-KKvtxw2-zYZXOYnApqm:1417088721)*

chrisdembia:
*[a discussion](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014#-KKvtzKOYFFQPv65xNSr:-KKvtzKOYFFQPv65xNSs:133003446) (no related file):*
I pushed one more commit that fixes tests and doxygen warnings. I also created a doxygen group for your description of Model::add### methods.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1014)*

aseth1:
Thanks @chrisdembia for the review and fixes. In the interest of time I will go ahead and merge.

PR #1020, opened 2016-06-17
Avoid unsigned int wrap through zero, causing testComponents failures

tkuchida:
testComponents has been reporting huge memory leaks (e.g., https://ci.appveyor.com/project/opensim-org/opensim-core/build/3758/job/0jwa5frbelovsxdb#L1479), presumably because this subtraction of unsigned ints wraps through zero.

### DIFF
--- a/OpenSim/Tests/Components/testComponents.cpp
+++ b/OpenSim/Tests/Components/testComponents.cpp
@@ -303,8 +303,13 @@ void testComponent(const Component& instanceToTest)
             Component* copy = instance->clone();
             delete copy;
         }
-        const size_t increaseInMemory = getCurrentRSS() - initMemory;
-        const long double leakPercent = (100.0*increaseInMemory/instanceSize)/nCopies;
+
+        // Catch a possible decrease in the memory footprint, which will cause
+        // size_t (unsigned int) to wrap through zero.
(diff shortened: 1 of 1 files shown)


klshrinidhi:
How can the for loop be decreasing memory usage ? 
It doesn't seem to free previously allocated memory. Does the result of function `getCurrentRSS()` include other things that are out of control of this program ?

tkuchida:
> How can the for loop be decreasing memory usage ?

On the surface, it doesn't look like a decrease in memory use is possible, but I think heap space can be freed behind the scenes. See [this webpage](http://nadeausoftware.com/articles/2012/07/c_c_tip_how_get_process_resident_set_size_physical_memory_use) for details.

klshrinidhi:
Got it. 
Then maybe use a `long long` as type for `increaseInMemory`.

tkuchida:
> Then maybe use a `long long` as type for `increaseInMemory`.

It doesn't look like `increaseInMemory` could ever be negative, so I thought it would be best to catch it immediately. I can make it a `long long` and do `max(0., ...)` if you prefer.

klshrinidhi:
I think it is fine as is. Don't worry about it.
:+1:

tkuchida:
Thanks for merging.

PR #1021, opened 2016-06-17
Randomize bool properties when testing Components

tkuchida:
Current `randomize()` method randomizes non-Boolean properties but _always_ toggles Boolean properties.


klshrinidhi:
:+1:

PR #1027, opened 2016-06-20
Add Component::updComponentList(), mutable iterator.

chrisdembia:
This PR introduces a ComponentList that allows modifying components:

```cpp
for (auto& muscle : model.getComponentList<Muscle>()) {
    muscle.set_max_isometric_force(...);
}
```

Basically, `getComponentList<T>()` now returns a `ComponentList<const T>`, and `updComponentList<T>()` returns a `ComponentList<T>`. The former provides access only to `ComponentListIterator<const T>`, while the latter also provides access to `ComponentListIterator<T>`.

The utility of this feature is shown in PR #1014, which works around the lack of this feature by using const casts. 

For now, python/java still only have access to the const iterators. We can extend the non-const iterators to wrapping in a future PR.


aseth1:
LGTM

chrisdembia:
Thank you @aseth1 & @klshrinidhi

Issue #1025, opened 2016-06-20
Make findComponent() private?


tkuchida:
I think it's confusing to present users with both `getComponent()` and `findComponent()`. These methods can be used (almost) interchangeably if the path is known ("almost" because one returns a reference and the other returns a pointer). I think it would be preferable to have a single `getComponent()` method that calls `findComponent()` if necessary.

chrisdembia:
I recently advocated for making it public (I thought it was protected or private). I would like two public methods:

- `getComponent()` must be given either a valid relative or absolute path. If there is no component at that path, it throws an exception.
- `findComponent()` takes the name (not a path) of a component, and searches recursively downward for a match, and returns nullptr if no component is found.

I think there are good use cases for both of these public methods.

tkuchida:
Currently, the doxygen for [`getComponent()`](http://myosin.sourceforge.net/1021/classOpenSim_1_1Component.html#a8fe4d7cab7fe752c69cd1e84ea07e5fd) and [`findComponent()`](http://myosin.sourceforge.net/1021/classOpenSim_1_1Component.html#a22d5f46e6b831efa27864da89dc3767c) indicates that both methods can take a path, so it seems that either method could be used if the full path is known. Having a single method would ensure the fast "get" is used whenever possible, and the slower "find" is used only when necessary. A suggestion to include the full path name could be printed when "find" is called.

I would expect a user to try `getComponent()` first and, if an exception is thrown, to just switch to `findComponent()` (if they know it exists) and then change all references to pointers.

chrisdembia:
If I want to check if a component `"foo"` is in a model, I can do either:

```cpp
try {
   getComponent("foo");
} catch (...) {
    ...
}
```

or

```cpp
auto* foo = findComponent("foo");
if (foo != nullptr) {
    ...
}
```

The latter is cleaner.


> Having a single method would ensure the fast "get" is used whenever possible, and the slower "find" is used only when necessary.

I think this is more how @aseth1 had it originally and I was opposed to it. You want one method that gets the component you asked for (unambiguously), and a separate method that will search for a component. Those are different requests, and so I think it is cleaner to have separate methods. 

It'd be better to chat in person I thin.

tkuchida:
> You want one method that gets the component you asked for (unambiguously), and a separate method that will search for a component.

[...]

PR #1030, opened 2016-06-21
Create methodComponent::printComponentsMatching().

chrisdembia:
This PR introduces a method on Component to help users discover the full path names of components in their model. This came out of the discussion in #1025. To facilitate this method, I created a new ComponentList filter: ComponentFilterFullPathNameRegex.

@tkuchida @klshrinidhi please review!

### DIFF
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -444,6 +444,33 @@ public:
     /** Get the relative pathname of this Component with respect to another one */
     std::string getRelativePathName(const Component& wrt) const;
 
+    /** Query if there is a component (of any type) at the specified
+     * path name. For example,
+     * @code 
+     * bool exists = model.hasComponent("right_elbow/elbow_flexion");
+     * @endcode
+     * checks if `model` has a subcomponent "right_elbow," which has a
+     * subcomponent "elbow_flexion." */
+    bool hasComponent(const std::string& pathname) const {
+        const auto* comp = this->template traversePathToComponent<Component>(pathname);
+        return comp;
+    }
+
+    /** Query if there is a component of a given type at the specified
+     * path name. For example,
+     * @code 
+     * bool exists = model.hasComponent<Coordinate>("right_elbow/elbow_flexion");
+     * @endcode
+     * checks if `model` has a subcomponent "right_elbow," which has a
+     * subcomponent "elbow_flexion," and that "elbow_flexion" is of type
+     * Coordinate. This method cannot be used from scripting; see the
+     * non-templatized hasComponent(). */
+    template <class C>
+    bool hasComponent(const std::string& pathname) const {
+        const C* comp = this->template traversePathToComponent<C>(pathname);
+        return comp;
+    }
+
     /**
      * Get a unique subcomponent of this Component by its path name and type 'C'. 
      * Throws ComponentNotFoundOnSpecifiedPath exception if the component at
@@ -455,7 +482,7 @@ public:
      * returns coord which is a Coordinate named "elbow_flexion" from a Joint
      * named "right_elbow" given it is a child of the Component (Model) model.
      * If unsure of a Component's path or whether or not it exists in the model,
-     * use findComponent() 
+     * use printComponentsMatching() or hasComponent().
      *
      * @param  pathname        a pathname (string) of a Component of interest
      * @return const reference to component of type C at 
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -485,6 +485,15 @@ public:
         return *const_cast<C*>(&(this->template getComponent<C>(name)));
     }
 
+    /** Print a list to the console of all components whose full path name
+     * contains the given string. You might use this if (a) you know the name of
+     * a component in your model but don't know its full path, (b) if you want
+     * to find all components with a given name, or (c) to get a list of all
+     * components on the right leg of a model (if all components on the right
+     * side have "_r" in their name).
+     * @returns The number of matches. */
+    int printComponentsMatching(const std::string& substring);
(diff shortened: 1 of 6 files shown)


OpenSim/Common/Component.h
```diff
@@ -485,6 +485,15 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
     }
 
+    /** Print a list to the console of all components whose full path name
+     * contains the given string. You might use this if (a) you know the name of
+     * a component in your model but don't know its full path, (b) if you want
+     * to find all components with a given name, or (c) to get a list of all
+     * components on the right leg of a model (if all components on the right
+     * side have "_r" in their name).
+     * @returns The number of matches. */
+    int printComponentsMatching(const std::string& substring);
```
tkuchida: Should this be a const method?

tkuchida:
`</review>` Cool! :snowflake: Just the one comment. Looks like this will be really useful, especially once `findComponent()` has been made private.

OpenSim/Common/Component.h
klshrinidhi: * Return type should be `unsigned`.
* Maybe be explicit that argument `substring` can be regex. If you don't want users to know, then maybe change implementation to not use regex because simple substring find is faster.
* Maybe add a code example to show how to call this thing and what result one might get.

OpenSim/Common/ComponentList.h
```diff
@@ -76,6 +78,21 @@ class ComponentFilterMatchAll : public ComponentFilter {
         return new ComponentFilterMatchAll(*this);
     }
 };
+
+/** A component is considered a match if its full path name matches the 
+provided regular expression. */
+class ComponentFilterFullPathNameRegex : public ComponentFilter {
+public:
+    ComponentFilterFullPathNameRegex(const std::regex& expr) : _expr(expr) {}
```
klshrinidhi: Maybe accept a `string` rather than reference to `regex` itself. That way, user of this class need not create an object of type `regex` to use this class. It also avoids copying of `regex` object. They can just pass a string and internally, this class can either:
* Create a `regex` and keep it as private member.
* Create a use a `regex` on the fly having a `string` as private member.

OpenSim/Common/Component.h
chrisdembia: Good points with (1) and (3). 

My intent with (2) is that advanced users can use ComponentFilterFullPathNameRegex directly, but that this method is a convenience when you just want to check for a substring (and is also available in python/java/matlab--std::regex does not get wrapped). I could write a new ComponentFilterFullPathNameContainsSubstring or add another method on Component that takes a regex. What do you think?

OpenSim/Common/ComponentList.h
chrisdembia: How about if I create a second constructor that takes a string? I'd like to leave one constructor that takes a regex, as someone could conceivably want to [set their own flags](http://www.cplusplus.com/reference/regex/basic_regex/basic_regex/) on the regex (e.g., case-insensitive).

klshrinidhi: I think keeping it simple will be better. May be accept only a string for now and always perform a case-insensitive search or have another parameter (with default) for case.

If complaints arise in future, update it accordingly.

chrisdembia: Based on this comment and your other one, then perhaps I shouldn't use a regex at all? What do you think? I can make it just do a substring search. And in the future, we can add the regex variant if need be.

klshrinidhi: If @tkuchida agrees, go ahead and remove regex. I don't know if it worth having or not. If you do want to have regex, using a simple string will be better.

chrisdembia: I'll wait for @tkuchida to respond but I think I do prefer not using the regex... my regex is probably wrong! I didn't try to escape any characters that the user may have provided in their substring.

chrisdembia: And I'm more inclined to have a weaker form of this feature than figure out how to get the regex correct.

tkuchida: Good point about regex. I agree: simpler is best; we can add regex in the future if desired/necessary. (We may then also want to prohibit special characters in Component names as well.)

chrisdembia: Great. Thanks for the cross reference.

OpenSim/Common/Component.h
chrisdembia: I have addressed your feedback:

1. Made the method const.
2. Changed return type to unsigned.
3. Added a code example with sample output.

OpenSim/Common/ComponentList.h
chrisdembia: I have addressed this feedback: I replaced this filter with one that just checks for a substring.

chrisdembia:
@tkuchida and @klshrinidhi I have addressed your feedback.

@tkuchida I also made `findComponent()` protected. Upon merging, please close #1025 if you think this PR addresses that issue.

Ready for review again.

tkuchida:
Looks great! :+1:
The only remaining issue that came up in #1025 was potentially needing/wanting a `hasComponent()` method.

chrisdembia:
> The only remaining issue that came up in #1025 was potentially needing/wanting a hasComponent() method.

I pushed a commit that adds this method.

tkuchida:
> I pushed a commit that adds this method.

Ok, great! LGTM.
#1025 can :boom: on merge.

chrisdembia:
Would someone like to merge this?

chrisdembia:
Thank you. 💥

Issue #1033, opened 2016-06-22
Adapters.h should be #included in osimCommon.h


tkuchida:
It's currently necessary to include
```cpp
#include <OpenSim/Common/Adapters.h>
```
(in addition to OpenSim.h) if you want to use `CSVFileAdapter`.

chrisdembia:
@carmichaelong, interested?

carmichaelong:
Sure I'll take a look at it.

tkuchida:
Fixed by #1035

PR #1035, opened 2016-06-22
Add adapters header to common.

klshrinidhi:
Addressing #1033.


tkuchida:
Thanks :+1: I would have made the change but I didn't know whether there were any other headers that should be included as well. :frog:

chrisdembia:
I'm merging this. OSX builds are stalled, but including an extra header file is extremely unlikely to only cause OSX builds to fail.

Thanks @klshrinidhi .

Issue #172, opened 2014-08-14
Need an interface for visualizing (consuming) geometry (lines, 3D shapes, meshes, ...) 


aseth1:
There are: Geometry, VisibleObject, DisplayGeometry, DecorativeGeometry classes just to visualize mesh files (as far as I can tell) and then we have ContactGeometry and WrapObject for specific analytical types of contact and wrapping. That's a lot of classes but it isn't the main problem- the problem is that the concept(s) they are trying to abstract are lost and relationships between them are obtuse. We have objects with an interface like:

```
//--------------------------------------------------------------------------
// Visible Object Support
//--------------------------------------------------------------------------
virtual VisibleObject* getDisplayer() const;
virtual void updateDisplayer(const SimTK::State& s);
virtual void updateGeometry(const SimTK::State& s);
```

why is a VisibleObject returned by getDisplayer. What is a Displayer? There is no such class/concept defined. This interface is not very usable. 

We should define Geometry.  Is it simply an abstraction for a shape? Or is a shape a type of Geometry? What is the main purpose of geometry in OpenSim?
What do we mean when we want to draw, display or visualize geometry? Are these synonymous?

I believe the main (and possibly the only purpose) for geometry in OpenSim going forward is to visualize Components (that is to see them on screen). Except for WrapObjects this seems to be true. The plan is to have Simbody handle wrapping as well someday. It also seems to me that the main purpose for geometry in Simbody is to compute something (contact (constraint) condition, a geodesic, a force) and not really to see it. 

Given the purpose of geometry is to visualize Components in OpenSim, we should make every effort to make it easy to add geometry to any component for the purpose of being seen.  We could addGeometry to a Component, and Geometry could connect to a Frame. The primary job of a Frame is to describe its transform in the ground.

The job of the Geometry then is to create its instantaneous shape(s) (as a function of the state). Here, I am using Geometry as an abstraction for something that make shapes (for example, someting like `generateDecocrations(const State&)`, but on a Geometry level instead of at the Component level). A Shape is any primitive shape expressed in ground (screen frame?) that can be rendered on screen (hopefully lightening fast). (I am not sure it makes sense to use VTK to do transformations when we can do them virtually for free and accurately ourselves). A Visualizer/GUI eats Shapes (to draw them) and knows which Shapes are in view, selected, etc.... The Shape knows its Geometry and the Geometry its Component, so now we have a way to identify a Component by its visualization. 

[...]

PR #1037, opened 2016-06-22
Cleanup add geometry and attachment to frames

aseth1:
This PR addresses most of the remaining issues of #172 and #914. The two main use cases it aims to support are:  1) adding geometry (e.g. bone mesh) that is fixed on a Frame which governs the transform of the geometry during visualization, and 2) add and move geometry that is not fixed to any Frame (e.g. representing experimental "floating" Markers) with the transform being an input.  

`ModelComponent::addGeometry` was removed and `addComponent` can be used instead but one must either 1) connect the `Geometry` to `Frame` using its `Connector<Frame>` (via the convenience `setFrame()` does this) OR 2) satisfy the newly added `Geometry::Input<SimTK::Transform>` in which case the `Output` must be a transform of Geometry w.r.t and expressed in the `Ground`.  In the second case, you can add the Geometry to any Component like any other using `addComponent`.

It is also very handy to just *attach* `Geometry` directly to a `Frame`. Here *attach* specifically means fixed (not moving w.r.t.) to a `Frame`, which takes ownership and uses itself to satisfy the Geometry's `Connector<Frame>` (leaving the Input disconnected). The input `transform` can be used if there is no `Frame` satisfying the connector, in which case the `Geometry` is not *attached* to any Frame and is drawn according to its "live" transform w.r.t. Ground. 

The current visualization code is dependent on (at least) decroative meshes being fixed to a mobilized body and my attempt to have all Geomtry display with its transform w.r.t ground was disastrous. @aymanhab assured that is was a very bad idea and that being fixed to a mobilized body was quite essential. A few floaty things seems OK.

At this  time there is one test failure `testVisualization` but it isn't clear what the test is doing and it may take awhile to figure out and debug. The exampleMain and Luxo example and other models appear to visualize fine, so we can at least begin the review process.


chrisdembia:
The diff for this PR includes your changes in enable_addComponent_to_replace_type_specific_add_methods. I'm going to do some branchwork so that this change is more reviewable (I will close this PR and open a new one).

chrisdembia:
@aseth1 I'm closing this because I don't want to have to review files from your other PR.

chrisdembia:
See #1038.

PR #1039, opened 2016-06-22
Bindings for component accessors, connectors, inputs, and outputs

chrisdembia:
This PR ensures that a number of the new component features are accessible via Python/Java/MATLAB. It includes tests for Python and Java that test:

1. Component::getComponent and Component::updComponent
2. ComponentList
3. Access connectors (Component::getConnector(), Component::getConnectee(), ...)
4. Connecting connectors to other components.
3. Accessing Outputs (Component::getOutputNames(), Component::getOutput(name), ...)
4. Accessing Inputs (Component::getInputNames(), Component::getInput(name), ...)
5. Connecting outputs to inputs.
6. Accessing AbstractChannels.
7. Setting the annotation for an Input.

To enable this functionality, I made a number of changes to the accessors in Component. For example, I created non-templatized variants of existing member functions. I also changed the way for iterating through a Component's outputs.

In python, I was able to use a typemap so that the return type of certain functions is the concrete type rather than the generic type. This will help avoid the confusion around using `safeDownCast()`, at least in python.

There is definitely more that I would like for the Python/Java/MATLAB interface. For example, it'd be great to have a for-each loop in MATLAB. I have made TODO comments where I haven't been able to get something to work. The emphasis in this PR is having access to subcomponents/connectors/inputs/outputs at all, over a good/complete interface (especially for MATLAB).

This PR also fixes the CMakeLists for the MATLAB tests, which had not been working.


chrisdembia:
@klshrinidhi do you have time to review this today? If @aseth1 does not have time to look at the changes to Component.h, then maybe @tkuchida has a moment?

klshrinidhi:
I will look at it. It will be good if another person looks at it too.

chrisdembia:
@klshrinidhi I agree. Perhaps you can focus on the bindings and the other reviewer can focus on the changes to Component.h etc.

OpenSim/Common/Component.h
```diff
@@ -501,8 +509,28 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
                                                        getName());
     }
 
+    /** Similar to the templatized getComponent(), except this returns the
+     * component as the generic Component type. This can be used in
+     * Python/Java/MATLAB. Here is an example of using this in MATLAB:
```
tkuchida: Is the code supposed to be Matlab? If so, comments should begin with `%`.

tkuchida:
I looked only at Component.h and didn't find anything awry.

OpenSim/Common/Component.h
chrisdembia: Ah good catch. Thanks.

OpenSim/Common/Component.h
```diff
@@ -570,53 +609,83 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
-    @see getNumConnectors()
+    /**
+    * Get the "connectee" object that the Component's Connector
+    * is bound to. Guaranteed to be valid only after the Component
+    * has been connected (that is connect() has been invoked).
+    * If the Connector has not been connected, an exception is thrown.
+    *
+    * This method is for getting the concrete connectee object, and is not
+    * available in scripting. If you want generic access to the connectee as an
+    * Object, use the non-templated version.
```
tkuchida: Out of curiosity, providing a default template argument doesn't allow the method to be used in scripting?

OpenSim/Common/Component.h
```diff
@@ -627,44 +696,97 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
+    int getNumInputs() const {
+        return int(_inputsTable.size());
+    }
+
+    /** Access the number of Outputs that this component has. */
+    int getNumOutputs() const {
+        return int(_outputsTable.size());
+    }
+
+    /** Collect and return the names of Outputs in this component as an
```
tkuchida: names of Outputs -> names of Inputs

OpenSim/Common/Component.h
```diff
@@ -627,44 +696,97 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
+     * std::vector. */
+    std::vector<std::string> getInputNames() const {
+        std::vector<std::string> names;
+        for (const auto& it : _inputsTable) {
+            names.push_back(it.first);
+        }
+        return names;
+    }
+
+    /** Collect and return the names of Outputs in this compoonent as an
```
tkuchida: com:hankey:nent -> component

tkuchida:
`</review>` Minor comments only. @klshrinidhi please include the python in your review.

OpenSim/Common/Component.h
chrisdembia: No; this method can be used in scripting if it has a default template argument, but I would need to add a line to the SWIG interface file:

```
%template(getConnectee) Component::getConnectee<Component>;
```

I thought a long time about whether I should do that, or do what I chose to do. In the end, I chose to create a separate non-template method:

1. In at least one case, there is good reason for the `C = Component` variant to be implemented differently from the more generic variant.
2. For consistency with (1), then it makes sense for all similar methods to have a non-template variant.
3. Scripting users may be scared by the template function in doxygen, and not think that the function is available in scripting. By creating the non-template function, we are explicit about its existence in scripting.

tkuchida: Ah, ok. Good points. Thanks for the explanation.

klshrinidhi:
I can't anything obvious. Go ahead. :+1:

chrisdembia:
Thanks for looking.

chrisdembia:
Oh darnit I forgot to commit my fixes for Tom's comments. Coming up.

PR #1042, opened 2016-06-22
Remove conversion of ptr to bool in hasComponent()

### LINKED ISSUE #1040: Warning C4800 on Visual Studio
Compiler reports `warning C4800: 'const OpenSim::Component *': forcing value to bool 'true' or 'false' (performance warning)` [due to coercion of pointer into bool](https://msdn.microsoft.com/en-us/library/b6801kcy.aspx) in [Component::hasComponent()](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Component.h#L456). Perhaps do the conversion manually:
```cpp
bool hasComponent(const std::string& pathname) const {
    const auto* comp = this->template traversePathToComponent<Component>(pathname);
    return comp != 0; //instead of "return comp;"
}
```
or include a #pragma at the top of Component.h:
```cpp
#ifdef WIN32
#pragma warning( disable : 4800 ) //forcing value to bool
#endif
```
@chrisdembia

chrisdembia:
Fixes #1040.

### DIFF
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -452,8 +452,7 @@ public:
      * checks if `model` has a subcomponent "right_elbow," which has a
      * subcomponent "elbow_flexion." */
     bool hasComponent(const std::string& pathname) const {
-        const auto* comp = this->template traversePathToComponent<Component>(pathname);
-        return comp;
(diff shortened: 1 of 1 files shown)


tkuchida:
:+1:

PR #1043, opened 2016-06-22
Updated simple example in README.md

### LINKED ISSUE #884: Update README cpp example to reflect latest API changes.

tkuchida:
Added ConsoleReporter and made other small improvements (e.g., more meaningful names). Changed from arm to leg because "humerus" was too long for line 56. Added console output and updated code in README.md.

Fixes #884 
@chrisdembia @jimmyDunne

### DIFF
--- a/OpenSim/Tests/README/testREADME.cpp
+++ b/OpenSim/Tests/README/testREADME.cpp
@@ -7,8 +7,9 @@
  * National Institutes of Health (U54 GM072970, R24 HD065690) and by DARPA    *
  * through the Warrior Web program.                                           *
  *                                                                            *
- * Copyright (c) 2005-2014 Stanford University and the Authors                *
+ * Copyright (c) 2005-2016 Stanford University and the Authors                *
  * Author(s): Chris Dembia                                                    *
+ * Contributor(s): Thomas Uchida                                              *
  *                                                                            *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may    *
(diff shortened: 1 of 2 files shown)


OpenSim/Tests/README/testREADME.cpp
```diff
@@ -43,48 +44,57 @@
...
     model.addForce(muscle);
     model.addController(brain);
 
+    // Add a console reporter to print the muscle fiber force and knee angle.
+    ConsoleReporter* reporter = new ConsoleReporter();
+    reporter->set_report_time_interval(1.0);
+    reporter->updInput("inputs").connect(muscle->getOutput("fiber_force"));
+    reporter->updInput("inputs").connect(
+        knee->getCoordinateSet()[0].getOutput("value"));
+    model.addComponent(reporter);
```
tkuchida: `addComponent()` should come before `connect()`.

chrisdembia:
LGTM except I would prefer to leave it as an arm example. The gif definitely looks like an arm more than a leg, and I don't think we should go through the hassle of updating the gif. Also, the muscle is still called "biceps" (although I suppose this could be the BFSH/LH).  Also, the README displays as more than 80 columns on github, so I am not too concerned about column width here.

tkuchida:
> LGTM except I would prefer to leave it as an arm example.

Ok, I will change it back and violate our column width rule.

chrisdembia:
LGTM please merge when tests pass.

PR #1038, opened 2016-06-22
Cleanup add geometry and attachment to frames

chrisdembia:
I am creating this PR for @aseth1 to replace #1037. The issue was that #1037 goes into master, even though it includes changes from a branch that has a separate PR.

Original description from @aseth1:

This PR addresses most of the remaining issues of #172 and #914. The two main use cases it aims to support are:  1) adding geometry (e.g. bone mesh) that is fixed on a Frame which governs the transform of the geometry during visualization, and 2) add and move geometry that is not fixed to any Frame (e.g. representing experimental "floating" Markers) with the transform being an input.  

`ModelComponent::addGeometry` was removed and `addComponent` can be used instead but one must either 1) connect the `Geometry` to `Frame` using its `Connector<Frame>` (via the convenience `setFrame()` does this) OR 2) satisfy the newly added `Geometry::Input<SimTK::Transform>` in which case the `Output` must be a transform of Geometry w.r.t and expressed in the `Ground`.  In the second case, you can add the Geometry to any Component like any other using `addComponent`.

It is also very handy to just *attach* `Geometry` directly to a `Frame`. Here *attach* specifically means fixed (not moving w.r.t.) to a `Frame`, which takes ownership and uses itself to satisfy the Geometry's `Connector<Frame>` (leaving the Input disconnected). The input `transform` can be used if there is no `Frame` satisfying the connector, in which case the `Geometry` is not *attached* to any Frame and is drawn according to its "live" transform w.r.t. Ground. 

The current visualization code is dependent on (at least) decroative meshes being fixed to a mobilized body and my attempt to have all Geomtry display with its transform w.r.t ground was disastrous. @aymanhab assured that is was a very bad idea and that being fixed to a mobilized body was quite essential. A few floaty things seems OK.

At this  time there is one test failure `testVisualization` but it isn't clear what the test is doing and it may take awhile to figure out and debug. The exampleMain and Luxo example and other models appear to visualize fine, so we can at least begin the review process.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/1038)

PR #1045, opened 2016-06-23
Fix testVisualization

chrisdembia:
@aseth1 this PR fixes testVisualization. Also, this PR is into your other visualization branch; #1038. 

### Description of changes

testVisualization was failing for a few reasons:

1. The removal of the <geometry> list property from ModelComponent, which
   (since updateFromXMLNode was not updated to do so), caused all geometry in
   this list to get lost.
2. The model used for this test (BuiltInGeometry.osim) overrode the <geometry>
   property for two of the Frames in the model (to contain Sphere geometry,
   etc.), causing the FrameGeometry for these Frames to not appear (since the
   FrameGeometry was stored in the default <geometry> list).
3. Previously, the FrameGeometry that got inserted during updateFromXMLNode set
   the display representation of the FrameGeometry to 0, since the older models
   did not have FrameGeometry and perhaps it would be surprising to suddenly
   show new geometry for old models. However, with the move of FrameGeomtry
   from the <geometry> list to a single-value property in Frame, the
   representation became 3 instead of 0. So even old models will now show
   FrameGeometry.

For (2) and (3), I updated the test's standard (vis_BuiltInGeometry.txt) by
changing the FrameGeometry's rep from 0 to 3, and by adding two more
FrameGeoemtrys to the standard (for the two Frames who had overriden the
<geometry> list in the model file).

For (1), I wrote some updateFromXMLNode code that moves everything that was in
`<geometry> into <components>`, where it now belongs. This also means that this
now tests that Geometry in `<components>` will visualize correctly!

An alternative fix for these issues would have been to use a model file for
this test that is not based on an intermediate XML version (30503). But the
model file used in this example could *not* have been written in the v3.3
format, since it contains the new Geometry classes. Future commits can revisit
this test, but at least for now it passes and does test some important
functionality. We should revisit the decision of if FrameGeometry gets
visualized by default (rep = 3).

### Note about why this PR has so many changes.

I merged this branch with master, which is why this PR has so many changes. If you merge #1038 with master, then this PR's diff will become much simpler. In any case, all the important changes are in the last commit, which you can view here: https://github.com/opensim-org/opensim-core/commit/883e8c30c03267b95740c2e3cfeb01e693ed5fd2.


chrisdembia:
I cancelled the travis-ci/push build.

OpenSim/Simulation/Model/PhysicalFrame.cpp
```diff
@@ -195,11 +195,36 @@ void PhysicalFrame::updateFromXMLNode(SimTK::Xml::Element& aNode, int versionNum
                     SimTK::Xml::element_iterator geomSetIter = visObjElement.element_begin("GeometrySet");
                     if (geomSetIter != visObjElement.element_end()) {
                         convertDisplayGeometryToGeometryXML(aNode, outerScaleFactors, outerTransform, *geomSetIter);
-                        geomSetIter->setElementTag("geometry");
                     }
                 }
             }
         }
```
aseth1: Just double checking that we agreed that this should be moved to `ModelComponent` since `<geometry>` was housed there? The test passes because the Model only had `Geometry` on `PhysicalFrames`, which was just luck.
Otherwise, we can revert the test file to an earlier format (30000) then this issue goes away since `<geometry>` on `ModelComponent` did not exist then. Something to resolve after dev week.

chrisdembia: K I made that change.

chrisdembia:
I cancelled the branch test.

PR #1046, opened 2016-06-23
add #ifdef WITH_BTK to Adapters.h

carmichaelong:
Necessary when WITH_BTK is unchecked since C3DFileAdapter.h will not be installed.

@chrisdembia


chrisdembia:
LGTM. We can come up with a more robust fix after the hackathon.

PR #1052, opened 2016-06-24
Cast to ScalarActuator to avoid crash.

chrisdembia:
Same as the PR I just created, except this goes into the hopper branch instead of master.


tkuchida:
I didn't encounter this issue… will it only appear in Debug mode or on certain platforms?

tkuchida:
Regardless, this should go in.

chrisdembia:
> I didn't encounter this issue… will it only appear in Debug mode or on certain platforms?

Dunno. I ran in debug and on a certain platform. :)

PR #1053, opened 2016-06-24
Allow building and running the answers for hopper.

chrisdembia:
I changed my mind about creating a cmake target for the answers. It is very useful for development to have a target for the answers (for debugging, and future incorporating fixes from the API). This PR creates a target for the answers.

If we really want to, we can remove the target before the workshop.


tkuchida:
:+1: Great!

PR #1031, opened 2016-06-21
Update hopper example for Dev Week

tkuchida:
Ready for review. Branch name is **tkuchida-updateHopperExample**.

The example consists of 5 files (in the "Examples/Hopper Device" project in Visual Studio), 3 of which you will need to modify. Please see the top of exampleHopperDevice.cpp for instructions. If you get stuck, the answers can be found in the `*_answers.*` files (they appear in the "Example answers/Hopper Device Answers" project in Visual Studio).

Hop to it! :rabbit:

PR #1064, opened 2016-06-25
Enable path point forces and visualization on phyiscal frames

aseth1:
Path point associated with `WrapObjects` where applying forces directly to the point as if it was expressed in the `MobilizedBody`. In the case of `PhysicalOffsetFrame` we have to account for the potential series of offset transforms by using `findTransformInBaseFrame()` and using it to transform the point in to the base mobilized body frame. This correction was necessary for both applying forces to the points and visualizing them correctly.


OpenSim/Simulation/Model/Force.cpp
```diff
@@ -137,25 +137,32 @@ double Force::computePotentialEnergy(const SimTK::State& state) const
 //-----------------------------------------------------------------------------
 // METHODS TO APPLY FORCES AND TORQUES
 //-----------------------------------------------------------------------------
-void Force::applyForceToPoint(const SimTK::State &s, const PhysicalFrame &aBody, const Vec3& aPoint, 
-                                    const Vec3& aForce, Vector_<SpatialVec> &bodyForces) const
+void Force::applyForceToPoint(const SimTK::State &s, const PhysicalFrame &frame,
+                              const Vec3& point, const Vec3& forceInG,
```
tkuchida: `point` -> `pointInF` would be clearer, and would self-document L145.

chrisdembia:
I don't really understand this stuff but the tests looks good. Might be good to test with the hopper device (create a new branch in which both the hopper branch and this branch are combined) to see if it fixes the issue for the hopper example.

Also, maybe it's fine to merge the hopper example into master now, but then we can still keep working on it on the tkuchida-updateHopperExample branch. But this way, anything that gets merged into master could be tested against the hopper example.

tkuchida:
`</review>` LGTM aside from the one comment. I'll test on the hopper branch now…

tkuchida:
:+1: Great! Successfully wrapped `PathActuator`s around a `WrapCylinder` that lives on a `PhysicalOffsetFrame`. I suggest we merge this PR. I'll then merge master into the hopper branch and apply my local changes, whereupon we can merge the hopper PR #1031.

chrisdembia:
Sweet. @aseth1 says "thank you!" If we wait 10 minutes, we can make sure at least one of the `travis-ci/pr` builds passes.

PR #1067, opened 2016-06-26
Fix wrapping of ArrayPtrs for GUI.

chrisdembia:
This PR fixes the wrapping of `ArrayPtrs<Object*>` for the GUI.

Also, this increases the required version of Java/JNI to 1.7, since our Java test cases rely
on a module only available in 1.7.

Also, I tried to ignore a warning that is generated when compiling the java bindings, but this did not succeed (we already ignore the same warning for gcc).


chrisdembia:
Would anyone like to review/merge this?

tkuchida:
LGTM but someone who is more familiar with the GUI should merge.

Bindings/Java/tests/TestBasics.java
```diff
@@ -23,6 +23,12 @@ public static void main(String[] args) {
 
     m.addComponent(pj);
 
+    // Change from ArrayPtrs<Object*> being non-const const.
```
aseth1: Looks like this block was added to verify ArrayConstObjPtr, but doesn't really belong here. Did you mean to leave it in?

chrisdembia: Yes I meant to leave it in, since this is similar to the code in the GUI that I wanted to make sure would work. The comment describing this "test" is lacking, though.

aseth1: As long as it is meant to be here I am happy to merge it. I imagine you will you or Ayman will be testing all the sets after dev week. So you no point adding better documentation now. I will merge it.

PR #1071, opened 2016-06-26
Improvements to Reporter labels

### LINKED ISSUE #1034: TableReporter does not write meaningful column labels
Currently, an output like this
```cpp
reporter->updInput("inputs").connect(myJoint->getCoordinateSet()[0].getOutput("value"));
```
produces a column labeled "value" in the resulting table. Better would be to use either the full path name or the annotation provided to `connect()` (if one is provided).

tkuchida:
ConsoleReporters now set labels using Input annotations, which can be specified by the user; TableReporters use full path names because, presumably, a user would want this detailed information in a file (and we are currently unable to determine the user's intent if they happen to specify an annotation that is the same as the default; added TODO in Reporter.h). Updates to hopper example and README.

Fixes #1034

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ int main() {
     reporter->set_report_time_interval(1.0);
     reporter->updInput("inputs").connect(biceps->getOutput("fiber_force"));
     reporter->updInput("inputs").connect(
-        elbow->getCoordinateSet()[0].getOutput("value"));
+        elbow->getCoordinateSet()[0].getOutput("value"), "elbow_angle");
     model.addComponent(reporter);
 
     // Configure the model.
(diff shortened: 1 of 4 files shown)


chrisdembia:
LGTM. +1 for remembering to update the README!

chrisdembia:
@aseth1 might want to chime in, so I won't merge.

README.md
```diff
@@ -104,7 +104,7 @@ This code produces the following animation:
 and prints the following information to the console:
 
         [reporter]
-                time|  fiber_force|       value|
+                time|  fiber_force| elbow_angle|
```
aseth1: Nice!

aseth1:
LGTM

aseth1:
Feel free to merge when tests pass.

tkuchida:
Thanks for reviewing. All :white_check_mark:; merging.

PR #1075, opened 2016-06-27
More hopper example updates

tkuchida:
I'll keep committing updates to this branch and merge this PR later tonight. [ci skip]


OpenSim/Sandbox/ExampleHopperDevice/exampleHopperDevice.cpp
```diff
@@ -178,6 +178,7 @@ int main()
     {
         // Build the hopper.
         auto hopper = buildHopper();
+        hopper.finalizeFromProperties(); //required due to known bug
```
aseth1: I think this would be more accurate:
```cpp
        // Get a copy of the hopper built programmatically.
        auto hopper = buildHopper();
        // Finalize the hopper's definition, which includes identifying 
        // its subcomponents from its properties.
        hopper.finalizeFromProperties();
```

tkuchida: I will make the requested change, though I think it looks like a bug from the perspective of the model builder. Also note that the only page that comes up in the doxygen when searching for "finalizeFromProperties" is [this one](http://myosin.sourceforge.net/1072/classOpenSim_1_1Component.html#a94531c6dfb2490ef235d05e7911fcc94). Based on this description, it is unlikely someone will know they have to call it after building a model.

tkuchida: We should also add `model.finalizeFromProperties()` to testREADME to be consistent.

chrisdembia: This is a larger discussion.

tkuchida:
Closing without merging to avoid polluting repo with large gifs.

Issue #1069, opened 2016-06-26
Introduce classes for managing paths of Components, Outputs, Channels, etc.


tkuchida:
Paths are currently strings that are passed around and parsed throughout the Component API, which is error prone. It might be better to introduce classes for managing paths, annotations on Inputs, etc.

chrisdembia:
related: #596.

tkuchida:
Also: perhaps use a different delimiter between Component path and Output name; a path to a Component is currently indistinguishable from a path to an Output.

Issue #1073, opened 2016-06-26
TODO from hopper example


tkuchida:
The "Hopper Device" project used to run without making any changes, but now crashes before the simulation starts. Also, each step should execute successfully when `if (false)` is changed to `if (true)` (though won't necessarily do anything interesting at first).

chrisdembia:
Successfully reproduced. "patellaFrame" does not have a mobilized body index.

tkuchida:
Interesting… it works as long as the `ConsoleReporter` has been added.

chrisdembia:
Occurs in `simulate()`, in `PathActuator::computeForce()`

tkuchida:
> Occurs in `simulate()`, in `PathActuator::computeForce()`

Sounds like it might be related to #1064. @aseth1: any ideas?

chrisdembia:
@tkuchida when you tested #1064 against the hopper device example, you tested it against `_answers.*`?

tkuchida:
> @tkuchida when you tested #1064 against the hopper device example, you tested it against `_answers.*`?

Yes, IIRC. Sorry, it's difficult for me to imagine an issue like this could occur. Why does adding a `ConsoleReporter` affect whether forces are computed correctly?

tkuchida:
It works if I replace `addConsoleReporterToHopper(hopper);` with `hopper.initSystem();` (i.e., calling `initSystem()` twice).

chrisdembia:
I don't know why this is, but adding the ConsoleReporter causes a reordering of the components (when traversing with the ComponentList()). I don't know if this is the issue though.

tkuchida:
Ok, thanks for investigating @chrisdembia. I'll include a workaround for now.

chrisdembia:
Had a lengthy discussion about this with @aseth1. He may try a fix, but IMO we should not rely on it for tomorrow.

chrisdembia:
The issue is that `addToSystem()` is invoked on `patellaFrame` before it is invoked on the Joints. The fix for now is to make sure `addToSystem()` is invoked on Joints before it is invoked on anything else.

chrisdembia:
@tkuchida my opinion is that at this point, you could push right to master if you are only editing the hopper device. The tests don't test the hopper device.

tkuchida:
Ok, thanks @chrisdembia. We could also consider elevating the hopper to a test case, since it seems that's how it's being used now.

chrisdembia:
[...]

Issue #1066, opened 2016-06-25
Simplify getting Coordinates from Joints


tkuchida:
### TODO
- [x] Improve interface for accessing Coordinates (PR #1116)
- [x] Replace CoordinateSet with a list property (PR #1210)
- [x] Add `updCoordinate()` methods (analogous to `getCoordinate()`) (PR #1222)
- [x] Replace `getCoordinateSet()[#]` with `getCoordinate(enum)` (or `updCoordinate(enum)`) (PR #1222)

-----

We currently need to get the entire `CoordinateSet` from a Joint:
```cpp
reporter->updInput("inputs").connect(myJoint->getCoordinateSet()[0].getOutput("value"));
```
It would look cleaner if we could replace the `()[0]` with just `(0)`:
```cpp
reporter->updInput("inputs").connect(myJoint->getCoordinate(0).getOutput("value"));
// or
reporter->updInput("inputs").connect(myJoint->getCoordinate("kneeFlexion").getOutput("value"));
```

aseth1:
This could be done pretty easily if we did not use the Set but use a property list of `<coordinates>`; then it would be `get_coordinates(0)`. I was thinking of doing that at some point.

tkuchida:
Ok, I'll look into this. I would also like to do
```cpp
myPinJoint.getCoordinate();
```
if there is only one Coordinate (e.g., `double getCoordinate(int idx = -1) const;` and return `get_coordinates(0)` if the list is of length 1.

tkuchida:
We could also introduce enums to improve readability, e.g.
```cpp
myFreeJoint.getCoordinate(FreeJoint::rotation1);
```

aymanhab:
Perfect, thanks @tkuchida :+1:

chrisdembia:
Most alternatives would be better than what we have now.

tkuchida:
Fixed by PRs #1116, #1210, and #1222.

PR #1078, opened 2016-06-28
Improvements to hopper example

tkuchida:
Changed from "height" to "yCoord" to avoid confusion with output not being "height" but rather "height/value". We may also want a unique delimiter between Component path and Output; added to discussion at #1069.

[ci skip]

### DIFF
--- a/OpenSim/Sandbox/ExampleHopperDevice/buildDeviceModel.cpp
+++ b/OpenSim/Sandbox/ExampleHopperDevice/buildDeviceModel.cpp
@@ -63,8 +63,8 @@ Device* buildDevice() {
     auto sphere = new Sphere(0.01);
     sphere->setName("sphere");
     sphere->setColor(SimTK::Red);
-    sphere->setFrame(*cuffA); cuffA->attachGeometry(sphere);
-    //sphere->setFrame(*cuffB); cuffB->attachGeometry(sphere->clone());
+    cuffA->attachGeometry(sphere);
+    //cuffB->attachGeometry(sphere->clone());
 
     // Create a WeldJoint to anchor cuffA to the hopper.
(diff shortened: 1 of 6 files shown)


aymanhab:
Sorry @tkuchida but I don't like ycoord since it implies some inherent coordinate-system/alignment. I think Joints should have built in enums that map to meaningful native names (e.g. Rotation1:0, Rotation2:1, Rotation3:2, Translation1:4.... etc.). You can decide to rename these to a more meaningful names but using an index does not feel very user friendly. If we follow this convention a Pin would use "Rotation" rather than 0! How does that sound?

tkuchida:
> I don't like ycoord since it implies some inherent coordinate-system/alignment.

What would you prefer? We've already written `hopper.setGravity(Vec3(0, -9.80665, 0));` at the top, so the model builder knows about the orientation of the coordinate system.

aymanhab:
@tkuchida I'm fine with the custom name being "yCoord" but I'd like a separate issue where the Coordinates have builtin meaningful default names rather than indices for readability.

tkuchida:
> I'd like a separate issue where the Coordinates have builtin meaningful default names rather than indices for readability.

Added to #1066, which already raised the issue.

chrisdembia:
Is this ready to be merged?

chrisdembia:
Just wondering; in case you were waiting for someone to merge this.

tkuchida:
> Is this ready to be merged?

Yes. One of the three tasks listed at the bottom of #1073 is to run the hopper example as a test (without a visualizer), but I think that can be done as a separate PR.

chrisdembia:
@tkuchida I did not delete your branch.

PR #1086, opened 2016-06-28
Fix momementum in component.h

MohammadhosseinSaadatzi:


tkuchida:
:scream_cat: You accidentally committed changes to the hopper example. Please commit only the typo change and create a new PR.

PR #1098, opened 2016-06-29
github homework, add new component to sandbox, fix task space [ci skip]

mitkof6:
Git PR for the OpenSim workshop, made some fixes to sandbox.


chrisdembia:
Thanks for updating the task space stuff and the function component! Neat idea. The main issue with it is there is no generic way to serialize a `std::function`. The component could still be useful though for users not interested in serializing (just writing a main). Another neat component would be a Lepton Expression (string), which *could* be serialized.

PR #1102, opened 2016-06-29
Update testComponents to use addComponent

aseth1:
testComponents is no longer limited to ModelComponents.  Now that any component can be added to Model, we do not need to limit to tested components to be of type `ModelComponent`. It now throws if the object being tested is not a `Component` and causes the test to fail.

### DIFF
--- a/OpenSim/Tests/Components/testComponents.cpp
+++ b/OpenSim/Tests/Components/testComponents.cpp
@@ -217,7 +217,8 @@ void testComponent(const Component& instanceToTest)
 
             if (dependency == nullptr){
                 // Get a concrete instance of a PhysicalFrame, which is a Body
-                if (dependencyTypeName == "PhysicalFrame"){
+                size_t n = dependencyTypeName.length();
+                if (n > 4 && dependencyTypeName.substr(n-5) == "Frame") {
                     dependency = Object::newInstanceOfType("Body");
                 }
             }
@@ -467,7 +468,7 @@ void addObjectAsComponentToModel(Object* instance, Model& model)
     const string& className = instance->getConcreteClassName();
     cout << "Adding " << className << " to the model." << endl;
 
-    try{
+   // try{
         if (Object::isObjectTypeDerivedFrom< Analysis >(className))
             model.addAnalysis(dynamic_cast<Analysis*>(instance));
         else if (Object::isObjectTypeDerivedFrom< Body >(className))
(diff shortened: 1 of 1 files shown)


OpenSim/Tests/Components/testComponents.cpp
```diff
@@ -217,7 +217,8 @@ void testComponent(const Component& instanceToTest)
 
             if (dependency == nullptr){
                 // Get a concrete instance of a PhysicalFrame, which is a Body
-                if (dependencyTypeName == "PhysicalFrame"){
+                size_t n = dependencyTypeName.length();
+                if (n > 4 && dependencyTypeName.substr(n-5) == "Frame") {
```
chrisdembia: It'd be great if we could use `std::is_base_of` or casts instead of this.

chrisdembia:
Done reviewing. Made 1 comment.

OpenSim/Tests/Components/testComponents.cpp
aseth1: I agree, but here we do not even have an object, just the type name of the dependency.
Maybe there is a better way without using the registered types.

OpenSim/Tests/Components/testComponents.cpp
```diff
@@ -467,40 +468,28 @@ void addObjectAsComponentToModel(Object* instance, Model& model)
...
-    // It is more than likely that connect() will fail, but the subcomponents tree
-    // will be traversable, so we can continue to resolve dependencies by visiting
-    // subcomponents' connectors
-    catch (const std::exception& e) {
-        cout << "testComponents: Model unable to connect after adding ";
-        cout << instance->getName() << endl;
-        cout << "ERROR: " << e.what() << "'" << endl;
-        cout << "Possible that dependency was not added yet. Continuing...." << endl;
+    if (Object::isObjectTypeDerivedFrom< Analysis >(className))
+        model.addAnalysis(dynamic_cast<Analysis*>(instance));
```
chrisdembia: Would be good to add a comment here. Connectors *can* connect to Objects but it is not tested and may not work (only components have paths).

OpenSim/Tests/Components/testComponents.cpp
chrisdembia: You can do the same thing as below: `isObjectDerivedFrom<Frame>(dependencyTypeName)`

chrisdembia:
I restarted some of the builds.

tkuchida:
AppVeyor is likely failing because of MSVC Update 3 bug (see #1115).

chrisdembia:
I restarted the appveyor build b/c we fixed the appveyor MSVC update issue in master.

chrisdembia:
appveyor/pr now passes. Merging.

aseth1:
Many thanks!

PR #1106, opened 2016-06-30
[gitignore] Ignore settings directory from CLion IDE

chrisdembia:
The CLion C++ IDE creates a settings directory `.idea` that should not be committed to the repo. This makes sure it doesn't happen by accident.

I cancelled the CI because I forgot to do `[ci skip]`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@
 tmp
 # Ignore build trees.
 build*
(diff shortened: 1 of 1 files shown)


chrisdembia:
@klshrinidhi could you review?

klshrinidhi:
:+1:

PR #1116, opened 2016-07-06
Add convenience accessors to underlying Coordinates

tkuchida:
Begins to address #1066. This PR implements the following:
```cpp
//enums are unique to each joint type
myBallJoint.getCoordinate(BallJoint::Coord::Rotation1X)

//zero arguments permitted only if there is exactly one Coordinate
myPinJoint.getCoordinate()
mySliderJoint.getCoordinate()

//not permitted
myFreeJoint.getCoordinate()
myPlanarJoint.getCoordinate(0)
myPinJoint.getCoordinate(FreeJoint::Coord::TranslationX)
```
Thanks @aymanhab, @klshrinidhi, and @aseth1 for the help!


OpenSim/Simulation/SimbodyEngine/BallJoint.h
```diff
@@ -52,14 +52,36 @@ class OSIMSIMULATION_API BallJoint : public Joint {
...
     using Joint::Joint;
 
+    using Joint::getCoordinate;
+
+    /** Indices of Coordinates; can be used with getCoordinate() for
+        convenience. */
+    enum class Coord {
+        Rotation1,
+        Rotation2,
+        Rotation3
```
chrisdembia: Is there a benefit to hard-coding these enum values? Is that possible with enum classes?

OpenSim/Simulation/SimbodyEngine/Joint.cpp
```diff
@@ -207,6 +207,27 @@ const OpenSim::PhysicalFrame& Joint::getParentFrame() const
...
+}
+
+const Coordinate& Joint::getCoordinate(unsigned idx) const {
+    const auto& coordSet = get_CoordinateSet();
+
+    OPENSIM_THROW_IF(coordSet.getSize() == 0,
+                     EmptyCoordinateSet);
+    OPENSIM_THROW_IF(idx >= coordSet.getSize(),
+                     IndexOutOfRange,
+                     idx, 0, static_cast<unsigned>(coordSet.getSize() - 1));
```
chrisdembia: This will not behave well if `coordSet.getSize() == 0`.

chrisdembia: Oh I see there is a separate exception for that.

OpenSim/Simulation/SimbodyEngine/BallJoint.h
chrisdembia: The names here are not consistent with the names of the `CoordinateIndex`es (`rx`, `ry`, `rz`).

OpenSim/Simulation/SimbodyEngine/BallJoint.h
```diff
@@ -52,14 +52,36 @@ class OSIMSIMULATION_API BallJoint : public Joint {
...
+    };
+
+    /** Get the Coordinates associated with this Joint. */
+    const Coordinate& getCoordinate(Coord idx) const {
```
chrisdembia: A question: Do you think this method should be on the base `Joint` class (they could be virtual, and implemented by the subclasses). For example:

```
const auto& joint = model.getComponent<Joint>("/arm26/shoulder");
joint.getCoordinate(BallJoint::Coord::Rotation1);
```

I am undecided.

My gut is that if the methods are generic (e.g., they take `int`s), they belong in the base class. If they are in the subclasses, why not make them even more tailored to the specific subclasses? E.g., `BallJoint::getCoordinateR1()` or `getRotation1()`. I understand that you are using the `Coord` enum as an argument, which means that the method with the current signature could not be in the base class. But the signature could easily take an int, and one could pass a `Coord` for that int.

chrisdembia:
Done reviewing. I am unsure about the design, but I do think this is an improvement over what exists so I will not halt a merge. Personally, I am not sure if I will use that more verbose form. The purpose of the enum is more evident to me if `getCoordinate()` and its overloads are on the base Joint; otherwise, it seems unnecessary.

OpenSim/Simulation/SimbodyEngine/BallJoint.h
tkuchida: > The names here are not consistent with the names of the `CoordinateIndex`es (`rx`, `ry`, `rz`).

That is true. IMHO, `rz` is not sufficiently descriptive to help the user avoid errors since it is not immediately apparent about _which_ Z-axis the rotation was performed (it could have occurred anywhere in a rotation sequence). `Rotation3`, on the other hand, implies nothing more than what is immediately apparent from the name. Thus, we would be enforcing that the user knows the rotation sequence.

> Is there a benefit to hard-coding these enum values? Is that possible with enum classes?

@klshrinidhi: Can you answer this?

OpenSim/Simulation/SimbodyEngine/BallJoint.h
tkuchida: > Do you think this method should be on the base `Joint` class (they could be virtual, and implemented by the subclasses).

We were trying to accomplish a couple things with this design. First, it should not be possible to do something like
```cpp
myBallJoint.getCoordinate(PlanarJoint::SomeEnum)
```
For example, a user could change the joint type without changing the argument to `getCoordinate()` and end up with the wrong index into the CoordinateSet. Secondly, it doesn't seem like the base Joint class should know anything about the derived classes, so including all the enums in Joint seems undesirable.

Methods like `BallJoint::getCoordinateR1()` would work as well, but I thought it would be most elegant to have a single `getCoordinate()` that was common to all joint types.

tkuchida:
> The purpose of the enum is more evident to me if `getCoordinate()` and its overloads are on the base Joint; otherwise, it seems unnecessary.

True: this entire PR is unnecessary since we can do
```cpp
myBallJoint.getCoordinateSet()[0]
```
already, but this PR will hopefully encourage more readable code while attempting to prevent users from making indexing errors (e.g., several joints allow "translation along X", but the corresponding index into the CoordinateSet is not guaranteed to be consistent between joint types).

OpenSim/Simulation/SimbodyEngine/BallJoint.h
chrisdembia: Thanks for explaining. Silly me...I just finally saw that there *is* a method in the base class.

> For example, a user could change the joint type without changing the argument to getCoordinate() and end up with the wrong index into the CoordinateSet.

That is a nice feature of the current implementation.

> including all the enums in Joint seems undesirable.

I agree; this would not be a good idea.


> most elegant to have a single getCoordinate() that was common to all joint types.

[This post](http://stackoverflow.com/questions/411103/function-with-same-name-but-different-signature-in-derived-class), which is about the `using Joint::getCoordinate` line, disagrees: 

> It is generally considered bad practice to have have functions in derived classes which have the same name as functions in the bass class which aren't intended to override the base class functions as what you are seeing is not usually desirable behaviour. It is usually preferable to give different functions different names.

I can understand that sentiment. It's not so bad in this case because the methods really are doing pretty much the same thing. If we leave in that `using` line, we should perhaps explain why it's necessary.


Anyway, I am fine with this PR.

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -226,6 +226,11 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
     // Coordinate Set
     const CoordinateSet& getCoordinateSet() const {return get_CoordinateSet();}
 
+    /** Get the Coordinate(s) associated with this Joint. If there is only one
+        Coordinate, you can call getCoordinate() with no argument. */
+    const Coordinate& getCoordinate() const;
+    const Coordinate& getCoordinate(unsigned idx) const;
```
chrisdembia: This second method will not have a doxygen comment. So, the comment above will seem weird to users.

klshrinidhi:
I think enum classes/structs can be assigned values at compile time -- hard coding -- but the values we want to assign these enums are not available at compile time.

* I think this PR makes user code more readable through use of enums. I think other projects use enums this way as well. 
* I saw the link to the stackoverflow post. In our case, these function(s) in base class are not declared virtual. So adding functionality in derived class has to be done through overloading and not through overriding. I think in this class hierarchy, if every joint has its own set of coordinates, appropriate enums and overloads that accept enums will make the user code less error prone and more readable.

Overall, I think it looks good. It may need some more Doxygen documentation.

tkuchida:
> If we leave in that `using` line, we should perhaps explain why it's necessary.

Added comment.

> This second method will not have a doxygen comment. So, the comment above will seem weird to doxygen viewers.

Fixed.

> It may need some more Doxygen documentation.

Added.

chrisdembia:
Thanks Tom.

tkuchida:
Only test failure is testContactGeometry timeout; added comment to #393

aseth1:
Thanks @tkuchida for the nice improvements to the interface. I like the `Set` free interface. I believe we should also replace the `Set` with a fixed sized (private) property list according to the number of coordinates provided by the `Joint`. I believe that is an important change since it fixes a potential bug of associating more coordinates with a Joint than it supports.

IMHO enums will be more work to add and maintain than the clarity we would hope to attain.

1.  the enums themselves may not be more informative (and in some cases more confusing) since each joint will have different enums and different context so one must still read the doxygen to understand what the Joint does and what each enum means in that specific context. For example, you need to know what `Rotation1` means for a given Joint where it is a rotation about the joint's Z in a `PlanarJoint` and a rotation about X for a `BallJoint`, etc...

 2. it does not extend to the `CustomJoint` - we cannot generate enums dynamically.
  
 3. we collate all the `Coordinates` into a flat list  (Set) in the `Model` in which, case the enums are not practical for accessing individual coordinates from the Model and you will still need index access.

I think these issues could be addressed with a significant amount of work, but in the end I don't see the added value over accessing the coordinate by name or index given that the `Joint` doxygen adequately describes what the coordinates are and their order in either case.

klshrinidhi:
I mostly disagree with your comments.

* To me, looking at the user code with named coordinates is far more intuitive to read and less error prone to write than indices like 0, 1, 2 and so on. 
* I disagree that enums are hard to maintain. I have seen this use of enums in other projects before and they worked very well. They may be hard to maintain as a result of poor design of classes and their interface.

Overall, I feel this PR is good improvement and I think changes should be made to the design to use more named indices like this than raw numbers.

aseth1:
@klshrinidhi Thank you for sharing your opinion. 

1. I don't disagree with you that enums are nicer than indices and could be an improvement. However, they are particularly useful if they mean something more than simply a replacement for an index. I think that "meaning" is hard to get across consistently given all the different `Joint` types. If I already know and understand the Joint type then using an enum over an index (0 to 5) is a relatively small improvement. Usually the context/meaning of coordinate is associated with its name "knee/felxion" is more meaningful than `knee.getCoordinate(Rotation1)` and this is marginally better than `knee.getCoordinate(0)`.

2.. I don't believe enums are hard to maintain. I believe the definitions/names to each Joint type and maintaining consistency with new Joint types added by new users may not justify the effort. The labels associated with the enums are not trivial.

3. And most importantly there are at least two other practical points (listed above) that need to be addressed.

I am happy to discuss in person.

aseth1:
@klshrinidhi @chrisdembia you are bringing me around that this would be more useful than not. I do still think it is big job to update each Joint and our examples. Can you come up with candidate names for the different Joints? The current `CoordinateIndex` member should also become the enum.

tkuchida:
> I believe we should also replace the `Set` with a fixed sized (private) property list according to the number of coordinates provided by the `Joint`. I believe that is an important change since it fixes a potential bug of associating more coordinates with a Joint than it supports.

I agree, but such a change will affect hundreds of lines of code. I thought it would make the most sense to agree on a cleaner interface first so that these hundreds of changes can be made using the new interface. The changes proposed in this PR do not preclude changing the underlying data structure later, so it isn't clear what concern is being raised here. Replacing the `Set` has already been noted in #1066.

> Usually the context/meaning of coordinate is associated with its name "knee/felxion" is more meaningful than `knee.getCoordinate(Rotation1)` and this is marginally better than `knee.getCoordinate(0)`.

Are you implying that there should be a `Joint::getCoordinate(const std::string& name)` method that iterates through the Joint's Coordinates looking for a Coordinate with the specified name?

> Can you come up with candidate names for the different Joints?

I don't think anyone has proposed changing the names of the Joints.

> The current CoordinateIndex member should also become the enum.

Do you mean that the name of the member variable should match the enum? There were some technical issues with the enums (e.g., cannot use the `CoordinateIndex` member variables directly because they are not defined until runtime); @klshrinidhi can explain better than I can.

chrisdembia:
> I don't think anyone has proposed changing the names of the Joints.

I'll let @aseth1 respond to the other comments but he meant the names of coordinates (enums) for the rest of the Joint classes.

tkuchida:
> I'll let @aseth1 respond to the other comments but he meant the names of coordinates (enums) for the rest of the Joint classes.

Proposed enum names:
- PinJoint - Rotation
- BallJoint - Rotation1, Rotation2, Rotation3
- FreeJoint - Rotation1, Rotation2, Rotation3, TranslationX, TranslationY, TranslationZ
- WeldJoint - (none)
- CustomJoint - (none)
- GimbalJoint - Rotation1, Rotation2, Rotation3
- PlanarJoint - Rotation, TranslationX, TranslationY
- SliderJoint - Translation
- EllipsoidJoint - Rotation1, Rotation2, Rotation3
- UniversalJoint - Rotation1, Rotation2

An alternative like "RotationX, RotationY, RotationZ" for BallJoint, FreeJoint, etc. (to be analogous with the Translation enums) would be ambiguous, but something like "RotationX, RotationYprime, RotationZdoubleprime" (or something shorter) could work.

chrisdembia:
Consider "RotationX1, RotationY2, RotationZ3."

tkuchida:
> Consider "RotationX1, RotationY2, RotationZ3."

I considered this option, but an "X-Y-Z" rotation sequence is also commonly referred to as a "1-2-3" sequence, so the suffixes just seem redundant. What about "RotationX, RotationY1, RotationZ2" (since the first rotation is about the _original_ X-axis)?

chrisdembia:
> an "X-Y-Z" rotation sequence is also commonly referred to as a "1-2-3" sequence

OpenSim does not use the convention of labeling the 3 cartesian directions using integers.



> What about "RotationX, RotationY1, RotationZ2" (since the first rotation is about the original X-axis)?

I'm fine with this.

tkuchida:
> OpenSim does not use the convention of labeling the 3 cartesian directions using integers.

I didn't claim that it does. My statement (an "X-Y-Z" rotation sequence is also commonly referred to as a "1-2-3" sequence) refers to the world beyond OpenSim (e.g., a body-fixed "313" rotation sequence is the default in Adams). The interface needs to be designed from the perspective of a user, not an OpenSim developer.

tkuchida:
[Just [kicking the watchdog](https://en.wikipedia.org/wiki/Watchdog_timer#Watchdog_restart)] @aseth1: any thoughts about the outstanding questions (search for "tag1" on this page) or proposed enum names ("tag2")?

aseth1:
@tkuchida @chrisdembia 

1. I think it is OK to make the interface changes you propose and then change the Joint's CoordinateSet into a property list of fixed size.  The reason I suggested doing it the other way around is that you will want to keep track of the index of the Coordinate in the list. The Coordinate index and the dof enum have to be equivalent. It might be easier/cleaner to update the Coordinate initializer (or use a macro) to provide the label and return the enum (that corresponds to the index in the coordinate list) before you go specifying enums for each joint class.

2.  I like the additional info about the sequence of Rotations where RotationY1 means rotation about Y' (after rotation about X) and that RotationZ2 is then about Z'', for example, but am I am not sure users will get the equivalence of 1 and 2, in the labels as ' and '' - I guess that's what Doxygen is for :grinning:.  It may be more straightforward to have Rotation1X, Rotation2Y, Rotation3Z to capture both the order (1, 2, 3) and axes (X, Y, Z) of the rotations. Then `ballJoint.getCoordinate(BallJoint::Rotation2Y)` specifies the second rotation in a sequence which is about the Y axis. If just Rotation2 you have to know that it is about Y and if just RotationY you need to know that it is the second rotation in a sequence. Then we can imagine other Joint types, like the ScapulothoracicJoint with `Rotation1Y, Rotation2X, Rotation3Z, Rotation4Y`, which might be more clear than `RotationY, RotationX1, RotationZ2, RotationY3`,  I could live with either. We could ask others for their opinions here, like @jenhicks @carmichaelong. You could discuss it in the dev meeting.

klshrinidhi:
Any updates on this ? Can we merge ?

tkuchida:
> Any updates on this ? Can we merge ?

I don't know how to address Ajay's first comment and I haven't received input from Jen or Carmichael.

carmichaelong:
On the proposed enums, I think any of the choices are defensible, so with good documenting any choice would be fine. The only other random thought that came to me was denoting body-fixed vs. space-fixed rotations, but good documentation should cover this without unnecessarily lengthening variable names.

Personally, my gut says the `Rotation1X` feels the best as I can read it as "the first rotation about axis X". The shortest way for me to read `RotationY1` is "the rotation about the Y axis of the first intermediate axis".

klshrinidhi:
> > Any updates on this ? Can we merge ?
>
> I don't know how to address Ajay's first comment and I haven't received input from Jen or Carmichael.

I am not in favor of writing more macros. I don't think it can be done another way. I think it is completely fine to define an enum for every Joint for which it makes sense (for CustomJoint, it may not).

tkuchida:
> Personally, my gut says the `Rotation1X` feels the best as I can read it as "the first rotation about axis X". The shortest way for me to read `RotationY1` is "the rotation about the Y axis of the first intermediate axis".

I would read `RotationY1` as "the rotation about the Y1-axis" (analogous to `TranslationX` = "the translation along the X-axis"), but Ajay prefers `Rotation{1|2|3}{X|Y|Z}`. @carmichaelong's preference agrees with @aseth1's preference (as well as the current proposal described in comment "tag2"); I will update the code accordingly.
- [ ] Update enum names to `Rotation{1|2|3}{X|Y|Z}`.

> I think it is completely fine to define an enum for every Joint for which it makes sense (for CustomJoint, it may not).

I think it is reasonable to expect users to define their own enums for `CustomJoint`s; @klshrinidhi, this would be a good thing to demonstrate in the proposed "example gallery".

klshrinidhi:
> I think it is reasonable to expect users to define their own enums for CustomJoints; @klshrinidhi, this would be a good thing to demonstrate in the proposed "example gallery".

:+1:

klshrinidhi:
Great. I can add the enums as per this discussion to rest of the Joints.

tkuchida:
> Great. I can add the enums as per this discussion to rest of the Joints.

Might want to hold off until @aseth1 is satisfied with the design to avoid redoing all the code; Ajay may have had an idea in mind about how to address his comment :one: above:

:speech_balloon: It might be easier/cleaner to update the Coordinate initializer (or use a macro) to provide the label and return the enum (that corresponds to the index in the coordinate list) before you go specifying enums for each joint class.

klshrinidhi:
Okay. I think I understand Ajay's comment and I don't want to add macros to add coordinate index and enum at the same time. Adding enums to Joints is not mandatory. It is a convenience, as this PR says. I think the current scheme is pretty good and can be copied over to other Joints.

aseth1:
@tkuchidachida @klshrinidhi 

>Might want to hold off until @aseth1 is satisfied with the design to avoid redoing all the code; Ajay may have had an idea in mind about how to address his comment :one: above:

The Coordinate initializers return a `CoordinateIndex` already. The enum must correspond to the same index or the interface will be broken. I would prefer to not enforce that consistency by hand and construct the necessary test cases to verify the correct correspondence. If the initializer could be modified or utilized by a macro so that the enum label and the index value our systematically bound, it will: 1) reduce the likelihood of errors 2) make it much easier to add enums to all Joints and 3) be used by others who create new `Joint` types that have enums and are consistent.  I am sure we can come up with something relatively simple. 

>  I don't want to add macros to add coordinate index and enum at the same time. Adding enums to Joints is not mandatory. It is a convenience, as this PR says.

With the exception of the `CustomJoint` each `Joint` should have enums for its coordinates or there should be none at all. It is either part of the `Joint` interface or it is not.

OpenSim/Simulation/SimbodyEngine/BallJoint.h
```diff
@@ -52,14 +52,44 @@ class OSIMSIMULATION_API BallJoint : public Joint {
...
+    /** Indices of Coordinates; can be used as arguments to getCoordinate() for
+        convenience.
+
+        <b>C++ example</b>
+        \code{.cpp}
+        const auto& rx = myBallJoint.getCoordinate(BallJoint::Coord::Rotation1X);
+        \endcode
+    */
+    enum class Coord {
+        Rotation1X,
```
tkuchida: @klshrinidhi will try `Rotation1X = constructCoordinate(Coordinate::MotionType::Rotational)`

klshrinidhi:
Enumerators can only be initialized with a constant-expression (compile time constant), an expression (including function calls) that can be evaluated at compile time.

This means we cannot initialize the enumerators Rotation1X, Rotation2Y, Rotation3Z with values currently held by rx, ry, rz. The correspondence has to be through the if-else ladder in the overload `getCoordinate()`.

aseth1:
@tkuchida @klshrinidhi Summarizing what we discussed. Rather than initialize the enumerators using CoordinateIndex, use the enumerated type to be the definitive index (as int). Then have the Coordinate initializer take the enum as an argument and assert that the returned index in the list (the initializer creates and adds the Coordinate to the list) is the expected enum.  For list access safety only provide `getCoordinate()` by enum. One can iterate through the list generically using `getComponentList<Coordinate>`. I think this will: 1) ensure consistency (the underlying enumerated value is guaranteed to be its index in the coordinate list), 2) eliminate custom logic for each Joint type, 3) minimize the code  a Joint maker has to add/duplicate for his/her new Joint type.

tkuchida:
Based on discussion with @aseth1 and @klshrinidhi: the initial design specifications (stated in the PR comment at the top) were incorrect. TODO: remove the first accessor (`myJoint.getCoordinate(0)`) and pass the enum to the CoordinateIndex constructor.

klshrinidhi:
> @tkuchida @klshrinidhi Summarizing what we discussed. Rather than initialize the enumerators using CoordinateIndex, use the enumerated type to be the definitive index (as int). Then have the Coordinate initializer take the enum as an argument and assert that the returned index in the list (the initializer creates and adds the Coordinate to the list) is the expected enum. For list access safety only provide getCoordinate() by enum. One can iterate through the list generically using getComponentList<Coordinate>. I think this will: 1) ensure consistency (the underlying enumerated value is guaranteed to be its index in the coordinate list), 2) eliminate custom logic for each Joint type, 3) minimize the code a Joint maker has to add/duplicate for his/her new Joint type.

Here are my thoughts:
* The function `constructCoordinate()` cannot take enum as an argument because different Joints will have different enums associated and they are considered to be different 'types'. The function `constructCoordinate()` will have to take an unsigned integer.
* If the function `constructCoordinate()` takes unsigned integer, then users are not forced to provide this argument in their Joints and therefore users are forced to create enums for all their Joints even if they don't care. 
<br/>The current solution is flexible in that it allows users to not have enums in their Joints.
* The solution you mention requires users to declare the indices and enums in the same order. If they don't , they will see an exception thrown by Joint and error message will have to tell them to keep order of indices and enums the same. The message may have to be verbose enough to 'explain' why user has to do this.
<br/>The current solution allows users to not care about the order of indices and enums. They only have to write a function that maps enums to corresponding indices.

I will implement you solution thought I don't think it is necessarily an improvement.

aseth1:
> The solution you mention requires users to declare the indices and enums in the same order. If they don't , they will see an exception thrown by Joint and error message will have to tell them to keep order of indices and enums the same. The message may have to be verbose enough to 'explain' why user has to do this.

That's the point! The purpose of the enum is to subsitute for an index in a list. If the enum does not correspond to the correct DOF the interface is broken - we will not be accessing the Coordinate we expect and will get no error to that effect. It could go unnoticed and be very difficult to debug so I do not want it to be possible in the first place. 

In the proposed scheme, the Joint maker only defines the enumerated list, and as long as they call the initializers in the same order (that is the coordinate is the order of their enumerators) everything should be fine. If for some reason they forget to the change the enum when initializing the next Coordinate they will get an exception that the "slot" is already filled/does not correspond and that is what we want. I don't see why this is in any way bad/inferior than allowing a possible mismatch to occur. 

>If the function constructCoordinate() takes unsigned integer, then users are not forced to provide this argument in their Joints and therefore users are forced to create enums for all their Joints even if they don't care.

I don't agree with the notion of "optional" enums in this case. You convinced me enums were much better than integers and you repeated how ambiguous/error prone access of the list was otherwise, and now you want to let users waft between using enums and ints. We should choose one or the other.

klshrinidhi:
> That's the point! The purpose of the enum is to subsitute for an index in a list. If the enum does not correspond to the correct DOF the interface is broken - we will not be accessing the Coordinate we expect and will get no error to that effect. It could go unnoticed and be very difficult to debug so I do not want it to be possible in the first place.

The interface is not broken. In the new scheme, users will be accessing coordinates by index just like it is right now. The enums add convenience of accessing the coordinates by name. Making the interface dependent on order of declaration of these indices is result of poor design. 
I don't think this PR is adding problems that don't exist. 

> I don't agree with the notion of "optional" enums in this case. You convinced me enums were much better than integers and you repeated how ambiguous/error prone access of the list was otherwise, and now you want to let users waft between using enums and ints. We should choose one or the other.

I am in favor of flexibility. I was arguing for enums where they make sense. If a user just wants to march through all coordinates and operate on them, they should be able to do so. Forcing them to declare enums makes them write unnecessary extra code.

I think we should drop the idea. This is getting too time consuming to arrive at a good agreeable decision.

aseth1:
> Making the interface dependent on order of declaration of these indices is result of poor design. 

The interface is dependent on order because  Joint type has a specific sequence to its coordinates. The CoordinateIndex is not arbitrary. If for any reason the enums does not correspond to that order then you do not have a valid Joint.  I am simply trying to help prevent that from happening.

jenhicks:
If I understand the discussion correctly, the API user-facing interface (i.e., interface for people using existing joints) is not at question, but how to implement the interface, in particular how to make it the process of creating a new joint user-friendly. @klshrinidhi or @aseth1 Can you provide an example of a new joint someone might want to create and explain what would be required?

tkuchida:
AppVeyor test failures ("An internal error has occurred in the compiler.") appear to be unrelated to this PR. All tests pass locally on Windows.

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -226,14 +226,12 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
     // Coordinate Set
     const CoordinateSet& getCoordinateSet() const {return get_CoordinateSet();}
```
aseth1: Consider removing this (redundant) accessor, since we: 1) do not want to promote access to individual Coordinates that bypasses the less error prone Coord enum interface , and 2) want to make it easier to convert to using a fixed sized property list.  Can be done as subsequent PR (just noting here so I don't forget).

tkuchida: Will submit as a separate PR to avoid polluting this one.

OpenSim/Simulation/SimbodyEngine/Joint.cpp
```diff
@@ -152,6 +152,17 @@ Joint::CoordinateIndex Joint::constructCoordinate(Coordinate::MotionType mt)
     return cix;
 }
 
+Joint::CoordinateIndex Joint::constructCoordinate(Coordinate::MotionType mt,
+                                                  unsigned idx)
+{
+    const auto cix = constructCoordinate(mt);
```
aseth1: Are you keeping the previous incarnation of `constructCoordinate` so that all the other Joints remain functional while you test this change? Are you then planning to have just one implementation of `constructCoordinate` after updating the other Joint types? I think the incremental changes are good but you should open an accompanying issue for what changes are forthcoming. You can also make `constructCoordinate(mt)` private to identify all the Joints that fail. Leaving it in as is risks having some Joints still on the "old" system without enumerators that we would not know about, or new Joints being implemented in the old way.

tkuchida: Discussed with @aseth1. I'll remove the single-argument version of `constructCoordinate()` and just pass in the loop counter for CustomJoint (since it won't have a `Coord` enum).

klshrinidhi:
I don't think passing enum's underlying integer to `constructCoordinate()` in the initializer of coordinates every Joint is a good idea. It looks like a hack. Looking the derived classes like BallJoint, it is not clear why this being done and why it has to be done by new Joint author and when it is done wrong, there will have to be lot of explaining. *This is kind of prevailing problem with rest of architecture as well. Things have to be called at exactly the right time/place and there is really no explanation. A good architecture in my opinion would not require such steps and avoid explanation.*

I can totally see that in future, there will be a PR to stick a macro in there *to help users do the right thing* and as a result make it even more opaque.

tkuchida:
> I don't think passing enum's underlying integer to `constructCoordinate()` in the initializer of coordinates every Joint is a good idea. It looks like a hack.

Can you suggest an alternative design, or are you proposing that we close this PR without merging?

klshrinidhi:
> Can you suggest an alternative design, or are you proposing that we close this PR without merging?

I don't have an alternate solution right now. It may need more thought on how the coordinates are managed internally.

jenhicks:
Let's merge this if (1) @aseth1 is happy with the implementation and (2) it improves the usability of the interface for @tkuchida. Future design suggestions are welcome @klshrinidhi , but let's move on to the rest of the big list of 4.0 tasks.

klshrinidhi:
I think it is important to think about maintenance too, which is the problem with the current solution. It is not straightforward. The solution we had before when this PR was started was easier to understand/maintain in my opinion.

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -322,8 +328,8 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
     /** Utility for derived Joints to add Coordinate(s) to reflect its DOFs.
     Derived Joints must construct as many Coordinates as reflected by the
     Mobilizer Qs. */
```
aseth1: The comment should be updated and document the additional argument. A constructed Coordinate requires the enum to identify its DOF in the Joint to ensure its list of Coordinates is consistent with its DOFs. Perhaps also mention that if their is a mismatch with the enumerator and the order in the Joint's list (when the Coordinate is added) an Exception is thrown.

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -568,6 +574,17 @@ inline int Joint::getNumMobilities(const SimTK::MobilizedBody::Free& mobod) cons
     return 6;
 }
 
+class EmptyCoordinateSet : public Exception {
```
aseth1: Try making the Exception type agnostic to the underlying data structure containing the Coordinates, e.g. `HasNoCoordinates`. If you change to property list of Coordinates do you want to have to change the Exception?

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -315,19 +315,32 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
...
-    Derived Joints must construct as many Coordinates as reflected by the
-    Mobilizer Qs. */
+    /** Utility for a derived Joint to add the Coordinates that correspond to
+    the motion it permits. Derived Joints must construct as many Coordinates as
+    are reflected by the underlying Mobilizer Qs. The index of the corresponding
+    enumeration (enum) is required at construction to ensure the Joint's
+    internal list of Coordinates is consistent with its interface; an exception
+    is thrown if the Coordinates are not constructed in the same order as the
+    enums have been defined. */
     CoordinateIndex constructCoordinate(Coordinate::MotionType mt,
```
aseth1: :+1:

aseth1:
LGTM

@klshrinidhi while agree that the static_casts are ugly and even "hacky" the current solution is less error prone. If you accidentally associate the enum with the wrong index in the list of coordinates you get an Exception. The original proposal had no way of guaranteeing correspondence and it was up to the implementer to not make a mistake. Furthermore, once such a mistake occurs there would be virtually no way to track it down until the user begins to see weird results. Because the Coordinates of a Joint are order dependent (that is just the nature of 3D rotations and not by design) a single rotation may appear to be correct and only a particular sequence result in something obviously wrong.

This code is objectively easier to maintain since the correspondence is created at time of construction through a single shared mechanism across all Joints and can be maintained, modified there. The alternative was to implement and update an if/else ladder in some 10 Joint classes.

tkuchida:
- [ ] Add comments about why we chose this design and how a user would add a new Joint type

aseth1:
@tkuchida @klshrinidhi @jenhicks IMHO this PR can be merged. Adding the documentation is not necessary in this PR and #1200 captures the need for better component base class documentation to serve as "guides" for deriving new components of those types.  Subsequent PRs should address the points in #1200.

tkuchida:
I think [the exception message in Joint.cpp (L155)](https://github.com/opensim-org/opensim-core/pull/1116/files#diff-2efad9acab01926bd89af8b957b3ac16R155) and [the comment in Joint.h (L337)](https://github.com/opensim-org/opensim-core/pull/1116/files#diff-31e3d687c1da78d898daf827d6930458R337) are sufficient for now, with #1200 covering the additional desired documentation.

tkuchida:
Thanks for merging, @aseth1 :fireworks:

chrisdembia:
🎆

PR #1122, opened 2016-07-14
Visualize wrap objects and contact geometry

aymanhab:
This PR allows WrapObjects and ContactMeshes to be visualized in API visualizer and GUI and also associate an Appearance Object to control the display instead of the old ad-hoc preference and color properties.


aymanhab:
@tkuchida You used/referred to this leading to the DevWeek if I remember correctly. Any feedback? Thanks.

tkuchida:
> Any feedback?

I don't see any new test cases here, so I'm assuming some existing tests were broken and are fixed by this PR. Which test cases should I be looking at?

aymanhab:
Good point @tkuchida, visualization is verified mostly "visually". There is a test case that does textual comparison of visualization primitives but the model there doesn't include either a ContactMesh or a WrapObject. I'll add another model to beef up the test of these. Thanks,

tkuchida:
> I'll add another model to beef up the test of these.

That sounds good. It might also be helpful to have a simple example program that displays every geometry primitive in every major configuration (solid vs. wireframe, attached to offset frames, etc.) so we can quickly verify that all cases are properly handled.

aymanhab:
Added a ContactMesh to test case and verified visually and textually. Had issues with WrapObjects, opened issue #1139 to discuss/address.

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -59,8 +59,8 @@ class OSIMSIMULATION_API SurfaceProperties : public Object {
     //==========================================================================
     OpenSim_DECLARE_PROPERTY(representation, int,
      "The representation (1:Points, 2:Wire, 3:Shaded) used to display the object.");
-    OpenSim_DECLARE_OPTIONAL_PROPERTY(texture_file, std::string,
-        "Name of file containing texture.");
+    OpenSim_DECLARE_OPTIONAL_PROPERTY(texture, std::string,
+        "Name of texture e.g. metal, bone. This is a hint to the GUI/Visualizer, implementation is Visualization dependent.");
```
aseth1: > This is a hint to the GUI/Visualizer, implementation is Visualization dependent

Is this not true for each attribute of the Appearance? You could include these of similar details to the Appearance property's comment to clarify  "default appearance".

OpenSim/Simulation/Model/ContactSphere.cpp
```diff
@@ -95,10 +95,10 @@ void ContactSphere::generateDecorations(bool fixed, const ModelDisplayHints& hin
...
     geometry.push_back(SimTK::DecorativeSphere(getRadius())
                            .setTransform(X_BP)
-                           .setRepresentation(SimTK::DecorativeGeometry::DrawWireframe)
+                           .setRepresentation(get_Appearance().get_representation())
                            .setBodyId(getFrame().getMobilizedBodyIndex())
-                           .setColor(SimTK::Vec3(0,1,0))
-                           .setOpacity(0.5));
+                           .setColor(get_Appearance().get_color())
+                           .setOpacity(get_Appearance().get_opacity()));
 }
```
aseth1: :+1:  so nice to see hard coded values disapear

OpenSim/Tools/Test/BuiltinGeometry.osim
```diff
@@ -402,14 +523,50 @@
...
+            <!--The opacity used to display the geometry between 0:transparent, 1:opaque.-->
+            <opacity>1</opacity>
+            <!--The color, (red, green, blue), [0, 1], used to display the geometry. -->
+            <color>0 1 1</color>
+            <!--Visuals applied to surfaces associated with this Appearance.-->
+            <SurfaceProperties>
+              <!--The representation (1:Points, 2:Wire, 3:Shaded) used to display the object.-->
+              <representation>2</representation>
+            </SurfaceProperties>
+          </Appearance>
```
aseth1: Is it that we only see Appearance for the first time here in the model file because it is the only changed appearance (others are all property value default so do not appear)? Or is it that only ContactGeometry Mesh has an Appearance? The test should verify the effect of Appearance on Geometry.

OpenSim/Simulation/Model/Appearance.h
```diff
@@ -186,7 +152,6 @@ class OSIMSIMULATION_API Appearance : public Object {
         constructProperty_opacity(1.0);
         // White by default, shows as a shade of gray
         constructProperty_color(SimTK::Vec3(1.0));
```
tkuchida: Perhaps `SimTK::White` rather than `SimTK::Vec3(1.0))`?

OpenSim/Simulation/Model/ContactGeometry.cpp
```diff
@@ -67,11 +67,11 @@ void ContactGeometry::constructProperties()
 {
     constructProperty_location(Vec3(0));
     constructProperty_orientation(Vec3(0));
-    constructProperty_display_preference(1);
+    Appearance defaultAppearance;
+    defaultAppearance.set_color(SimTK::Vec3(0., 1.0, 1.0));
```
tkuchida: `SimTK::Cyan` instead? The `SimTK` colors are defined [here](https://github.com/simbody/simbody/blob/master/SimTKcommon/Geometry/src/DecorativeGeometry.cpp#L34).

OpenSim/Simulation/Wrap/WrapObject.cpp
```diff
@@ -67,12 +67,10 @@ void WrapObject::constructProperties()
...
 
     constructProperty_quadrant("Unassigned");
-
-    constructProperty_display_preference(1);
-    Array<double> defaultColor(1.0, 3); //color default to 0, 1, 1
-    defaultColor[0] = 0.0; 
-
-    constructProperty_color(defaultColor);
+    Appearance defaultAppearance;
+    defaultAppearance.set_color(SimTK::Vec3(0., 1.0, 1.0));
```
tkuchida: `SimTK::Cyan`?

OpenSim/Tools/Test/BuiltinGeometry.osim
aymanhab: @aseth1 I believe only changed properties will show up.  I got this one though thru the XML browser in the GUI since I wasn't sure what's the latest XML layout for ContactMesh references/connections/frames/etc. I agree that the test case should verify the effect of appearance but probably only for limited set of primitives since they all use the same code.

OpenSim/Simulation/Model/Appearance.h
aymanhab: Sure, thanks @tkuchida though they end up in XML as three numbers but that's more readable. :+1:

OpenSim/Simulation/Model/ContactGeometry.cpp
aymanhab: Agreed, will also have to check if these definitions are accessible through scripting so the code is not C++ handicapped.

aymanhab:
SimTK colors are actually constants not enums so they don't wrap cleanly (you'll have to use a static getCyan() method instead of SimTK::Cyan from scripting. Is it worth writing a macro or using a more involved enum. @klshrinidhi any ideas?

OpenSim/Tools/Test/BuiltinGeometry.osim
aymanhab: Indeed, verified that in case no Appearance changes are made from default then no Appearance shows in XML altogether.

aymanhab:
Added case in testVisualization that displays a Sphere with various combinations of Appearance attributes and verified the result visually and then added test case. Should be ready to go AFAIK.

OpenSim/Simulation/Model/Appearance.h
tkuchida: Can we list the textures that are supported by the standard OpenSim visualizer, or tell users how to find this list?

tkuchida:
`</review>`
* One suggestion [in Appearance.h](https://github.com/opensim-org/opensim-core/pull/1122/files#diff-04203e15f10330090be952ca0ceca1ce).
* It might be good to replace "body" with "frame" throughout (e.g., `ContactGeometry::getBody()`, `void ContactGeometry::setFrame(const PhysicalFrame& body);`), but that's unrelated to this PR.

OpenSim/Simulation/Model/Appearance.h
aymanhab: @aseth1 The change here is moving from a specific file for texture (which was available in old models/format) to a name based texture that can be reused and that's why the comment was updated. @tkuchida supported textures will depend on the visualizer, e.g. API visualizer supports none while I foresee GUI supporting "bone", "sand", "metal", "wood" etc. None of these is implemented yet pending this change (and also texture coordinates assigned to geometry) but we should open an issue to make sure that this important documentation doesn't fall through the cracks.

aymanhab:
I agree @tkuchida regarding changing Body to PhysicalFrame in new/modified files while at it (I was mainly looking at diffs so I didn't see it). Thanks for the find.

OpenSim/Simulation/Model/Appearance.h
tkuchida: > we should open an issue to make sure that this important documentation doesn't fall through the cracks.

Sounds good.

aymanhab: Issue #1149 opened.

aymanhab:
Only failure is bogus failure in testComponents/UniversalJoint that was never touched by this PR. All feedback has been addressed AFAIK. Thanks for the review, and please let me know if ready to merge.

tkuchida:
> bogus failure in testComponents/UniversalJoint

Fixed by #1155

tkuchida:
@aseth1: LGTM. Merge?

OpenSim/Tools/Test/testVisualization.cpp
```diff
@@ -101,6 +103,9 @@ int main()
     try {
         LoadOpenSimLibrary("osimActuators");
         testVisModel("BuiltinGeometry.osim");
+        std::string testAppearanceModelFile = "AppearanceTest.osim";
+        createModel4AppearanceTest(testAppearanceModelFile);
```
aseth1: I think we could easily avoid writing out more model files unnecessarily by making `testVisModel()` take a model and make `createModel4AppearanceTest()` return a Model. It wouldn't preclude using models from file, you would just load em up prior to sending to `testVisModel()` and would allow us to add more visualization tests/models without creating a litany of files as proposed in #1151. We should beef up serialization tests to verify serialization as per #1150.

OpenSim/Tools/Test/testVisualization.cpp
```diff
@@ -142,3 +147,33 @@ void testVisModel(string fileName)
     delete model;
     ASSERT(same == 0, __FILE__, __LINE__, "Files do not match.");
 }
+
+void createModel4AppearanceTest(const std::string& filename)
+{
```
aseth1: This model creation needs to be better documented so that there is unambiguous correspondence between the visualization of Geometry (explicit and implied) and the vis file standard below.  Describe every piece of geometry and how it should appear (what transform, rep, opacity, scale) it should have. It should not be up to the reviewer and future tester to deduce that correspondence.

OpenSim/Tools/Test/vis_AppearanceTest.txt
```diff
@@ -0,0 +1,42 @@
...
+[0,1,0,0]
+[0,0,1,0]
+[0,0,0,1]
+
+DecorativeSphere:1 bodyId:0 color:~[1,1,1] indexOnBody:0 Opacity:1 Rep:2 Scale:~[1,1,1] Transform:
+[1,0,0,3]
+[0,1,0,0]
+[0,0,1,0]
+[0,0,0,1]
+
```
aseth1: I realize that this is the "standard" file for the test, but corresponding this text file to the geometry in the model is challenging. Same goes for the previous test. Having tried to debug the visualization myself in the past (e.g. with changes to attach Geometry to Frames) I found it extremely time consuming to map the geometry in the model to corresponding text rendering. For example, why is everything `DecorativeFrame:1` or `DecorativeSphere:1`? Could you not use different bodies and transforms so that not everything is `bodyId:0` and what is `indexOnBody` and why is it always 0? if we can only see color as `[1, 1, 1]` then don't use use the SimTK::Color constants because again the correspondence is not obvious. You can use comments in test model creation block above to describe what you should see in the vis file. For example that `DecorativeGeometry::DrawWireframe` is `Rep:2` and the default (Shaded?) is `Rep:3`.  As a reviewer I am supposed to help verify that the standard is indeed correct, and I don't think I can do that given the potential ambiguities. That is another reason to use models that are created programmatically because you can document what the visualization should draw for each Component added to the Model.

aymanhab:
Test failure (in testVisualization) due to failing to load osimActuators on one platform after merging with master, any clues @tkuchida or @chrisdembia ?

tkuchida:
> any clues?

Nothing immediately comes to mind, but the failure occurs only in Debug mode and it appears to be due to a [Debug mode–only assert](https://travis-ci.org/opensim-org/opensim-core/jobs/150721783#L1954).

aymanhab:
The assert in Simbody is downstream from the first issue which is failing to load the library osimActuators (hard to tell if it's related, though probably not). The failure to load library is unjustified unless something changed in LoadOpenSimLibrary. Does it require "_d" suffix on all platforms now? and are the libraries created with such suffix?

tkuchida:
> hard to tell if it's related, though probably not

Okay, sorry. That was my only idea.

chrisdembia:
The message about failing to load osimActuators appears even if the test passes. It can be ignored here and in other tests as well. It's caused by the line `LoadOpenSimLibrary("osimActuators");`, but this line is only necessary if using MSVC/GCC I think. @carmichaelong and I investigated this a while ago. It's likely because those compilers choose not to link those libraries since it cannot see that that library is actually being used (via XML deserialization).

aymanhab:
Thanks @chrisdembia Verified indeed since I made another change that made the test pass but the message still appears. Would be good to find cross-compiler/platform solution since the message will be misleading to users.

chrisdembia:
Well it only really occurs in test cases; we should avoid `LoadOpenSimLibrary("osimActuators")` it when possible in c++ examples. What I've done at least once before is to surround it by `#if !defined(__clang__) ... #endif` or something like that.

aymanhab:
Thanks @aseth1 @tkuchida and @chrisdembia for the feedback/review.

I made the change so the test case doesn't write the new model to a file, in line with the new testing strategy #1151 For testing visualization, the ground truth is really nothing short of putting up a window and taking a snapshot but that's not doable for many reasons. 

The test verifies that whatever primitives sent down the pipe to visualizer are correct (after verifying visually). The dump of the primitives is not intended for human consumption as such. The numbers you see are just a dump of whatever in the DecorativeGeometry class from simbody, I don't know for sure how/if they are used in generateDecorations calls.  Hope this helps.

aymanhab:
@aseth1 With the change to testVisualization per your feedback, so no new .osim file is created, please let me know if this's ready to merge. Thanks.

jenhicks:
Is there someway we can include a reference visual image for debugging? Or can we capture a render of the model and compare pixels to a standard?

aymanhab:
We can include one image for our own visual comparison (if we can set camera position, view angle, view up, etc.....), automating is much harder since it depends also on resolution and so is not cross-platform (API visualizer didn't work on VMs for a while due to old OpenGL drivers etc.) To do cross platform we have to use a surrogate measure and right now this's listing the primitives sent to the visualizer. In general for a reasonable model with many bodies/joints/geometry objects it'd be next to impossible to hypothesize/compose where things are in 3D and even more if objects will be occluded by others or within view frustum including lighting etc. to decide if the output is "correct" without doing actual visualization.

jenhicks:
Including an image seems reasonable to me.

aymanhab:
Sure @jenhicks I'll add two images for the two sub-tests.

aymanhab:
Tw images have been added per @jenhicks

chrisdembia:
This PR now has merge conflicts.

aymanhab:
Conflict was due to headers refactoring and has been resolved, merged and pushed.

aymanhab:
All tests have passed and comments/review addressed including adding images to verify visualization.  Last bit of feedback was only related to test which have been addressed. I'd like to move on, if there're issues with the test open a separate issue, please let me know what you think @chrisdembia @tkuchida

tkuchida:
:+1: This PR is getting soggy; I think it should be merged upon verifying that issues have been opened to track any remaining tasks. @aseth1 raised some issues on line 42 of vis_AppearanceTest.txt and elsewhere; I think he should review.

aymanhab:
Issue #1190 added to address the standards file and explain the rationale but that's no reason to hold this PR as far as I can tell.

jenhicks:
@aymanhab For this PR, I would make one additional update to address concerns:

> You can use comments in test model creation block above to describe what you should see in the vis file. For example that DecorativeGeometry::DrawWireframe is Rep:2 and the default (Shaded?) is Rep:3

I would add a brief description of what a tester should be looking for and note the jpgs to check.

Other than that, I am OK with merging, and putting additional feedback in PR #1190

aymanhab:
Done, and merged. Thanks all for the feedback.

PR #398, opened 2015-03-13
A first shot at a Delay component

chrisdembia:
I have created a Delay component that basically wraps Simbody's Delay Measure and allows component developers to use the input/output system to delay a quantity of their choosing. This PR addresses #52. A large reason for creating this Delay component was to test the "new MCI". As a result, the points below don't only concern this specific Delay component, but also the design of the new MCI as it affects this Delay component and other future components.

The Delay component is functional but has a few limitations; see below.
### Here's a summary of changes:
1. I created an InputMeasure<T>, which I think will be useful for other operator components.
2. Created a Delay component with one input and one output. The output is simply the input delayed by a specified amount of time. This delaying is implemented via a InputMeasure<T>.
3. The Delay is templatized by the type of quantity one wants to delay. I've created typedefs for what I expect to be common types: ScalarDelay, VectorDelay.
4. Added an Object_GetClassName() specialization for Vector, since I use Vector as a template argument (Delay<Vector>).
5. Created a testOperators test to test the Delay component. Tests that the delay is correct and that it has an effect on the behavior of the system through the controller.
   1. Tests the creation of one's own type of Delay: Delay<Vec3>.
   2. Tests the use of a Delay within a custom controller.
   3. Tests the delay with various dependsOn stages (commented out for now; see below).
   4. Tests copying and serialization of a model with a Delay  (commented out for now; see below).
   5. Test that a negative value for the delay duration throws an exception.
6. Created a doxygen group for Operators.
### Some points for discussion:
1. This Delay is the first operator so far. I decided that an Operator abstract class is not needed; I couldn't discern what would be special about such a class aside from the fact that the inputs and outputs are very generic (so it's not really possible to give meaningful names to the inputs and outputs). This is why I created a Doxygen group.
2. Are the names "input" and "output" for the Delay's "input" and "output" okay? The component is so general it's not possible to come up with meaningful names.
3. Right now the Delay holds onto a MeasureIndex. In other places, classes that use measures instead hold onto a Measure handle. Is one of these alternatives preferable?
4. Should InputMeasure be a plain Measure, as it is, or should it be a Measure::Result?
5. Should there be a minimum dependsOnStage for InputMeasure (Time), regardless of the Input's connectAt stage?
6. The fact that this Delay component is not a ModelComponent means it can only be a subcomponent; it cannot be a root component in a Model. This seems fine and perhaps desirable. Any other opinions?
[...]

PR #1123, opened 2016-07-14
Delay branch updates from workshop

stavness:
- moved testDelay.cpp from "OpenSim/Common/Test" to  "OpenSim/Simulation/Test" (because it requires osimSimulation.h and osimActuators.h)
- moved Delay.h from OpenSim/Common to OpenSim/Simulation/Model
- renamed constructInfrastructure to constructProperties; removed override keyword
- include only required headers
- in extendConnectToModel added if statements to check for a connection before connecting inputs, because method is called multiple times
- fixed testNegativeDelayDurationException — SimTK_TEST_MUST_THROW_EXC for model.addController() call
- in DiscontinuousInputTesting: added set name method; changed three Delay components to be member subcomponents; added memberSubcomponentIndex to class; added new calls to get Delay component output value


OpenSim/Simulation/Test/testDelay.cpp
```diff
@@ -603,14 +632,19 @@ void testDiscontinuousInput() {
     // The "step" output switches at 0.5 seconds, but the delay is 0.2 seconds,
     // so the delay output should still be 1.5.
     //std::cout << comp->stepDelay.getValue(ts.getState()) << " " << std::endl;
-    SimTK_TEST_EQ(comp->stepDelay.getValue(ts.getState()), 1.5);
-    //std::cout << comp->discontDelay.getValue(ts.getState()) << " " << std::endl;
-    SimTK_TEST_EQ(comp->discontDelay.getValue(ts.getState()), 0.46);
 
+    const Delay& stepDelay = comp->getComponent<Delay>("/discontinuousInputTesting/stepDelay");
+    const Delay& discontDelay = comp->getComponent<Delay>("/discontinuousInputTesting/discontDelay");
```
chrisdembia: The current way to fish out a member subcomponent is shown [here](https://github.com/opensim-org/opensim-core/blob/de24648dbc6c1342d74bf7ba8b792fd4c08edbc6/OpenSim/Actuators/Thelen2003Muscle.cpp#L103).

chrisdembia:
> in extendConnectToModel added if statements to check for a connection before connecting inputs, because method is called multiple times

I think this is fine but I am curious. Is this for performance? Was reconnecting slow?

Thanks for updating this, @stavness. I'll merge this now.

chrisdembia:
Now #398 is updated, and further work on the Delay component can happen there. Let me know when you think #398 is ready to be reviewed, or if you have things planned or other things that you think should be done first.

mshabani:
@chrisdembia  
"in extendConnectToModel added if statements to check for a connection before connecting inputs, because method is called multiple times"
I think this is fine but I am curious. Is this for performance? Was reconnecting slow?

Removing these if statements would result in an exception complaining about connecting input multiple time. That's because "extendConnectToModel " is called several times internally and every time it tries to connect all inputs. first time it's ok because those are not connected. in the second call exception happens.
Sorry I don't have access to reply to your comment so I wrote a new one.

chrisdembia:
Thank you for explaining.

PR #1125, opened 2016-07-15
Proposal and thoughts about an interface for implicit form of dynamics.

chrisdembia:
This PR introduces a markdown file into the sandbox that describes ways of introducing implicit differential equations to OpenSim.


chrisdembia:
I'm merging this; it's just a markdown file in the sandbox.

PR #1126, opened 2016-07-16
Allow SWIG to recognize that Mat33, etc is wrapped.

chrisdembia:
Previously, SWIG did not know about the Mat33 typedef, so if SWIG saw
Mat33 in a header file, it would not know that it was the same type that it
had already wrapped.

@aymanhab or @klshrinidhi, could you review please?

This bug was discovered by @humphreysb.


aymanhab:
Sure, looks good. Merging as having little to no side effects AFAIK.

chrisdembia:
Thank you.

PR #1127, opened 2016-07-18
Split testCMCArm26 into two to speed it up.

klshrinidhi:
Splitting the test into two -- Thelen and Millard. Ctest will run these tests in parallel and result in some speed up of about 2x (as per my machine).

Running Thelen and Millard in parallel within the same test did not work right away. The threads conflict with each other due to shared objects in memory. The solution of splitting the test into two separate was the quickest and easiest.


Applications/CMC/test/testCMCArm26_Millard.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    Storage results(resultFile), temp("std_arm26_states.sto");
+    Storage *standard = new Storage();
+    cmc.getModel().formStateStorage(temp, *standard);
+
+    Array<double> rms_tols(0.02, 2*2+2*6); // activations within 2%, angles within .6 degrees
+    const string& muscleType = cmc.getModel().getMuscles()[0].getConcreteClassName();
+    string base = "testArm26 "+ muscleType;
+
+    if(muscleType != "Thelen2003Muscle"){
```
chrisdembia: This check should no longer be necessary, since the model should only contain Millard muscles. A corresponding change could be made in testArm26.cpp.

Applications/CMC/test/testCMCArm26_Millard.cpp
```diff
@@ -0,0 +1,115 @@
...
+    Object::renameType("Thelen2003Muscle", "Millard2012EquilibriumMuscle");
+
+    std::ifstream inFile{"arm26_Setup_CMC.xml"};
+    std::ofstream outFile{"arm26_Setup_CMC_Millard.xml"};
+    std::string line{};
+    while(std::getline(inFile, line)) {
+        // Find and replace muscle name.
+        {
+            std::string orig{"Thelen2003Muscle"};
+            std::string repl{"Millard2012EquilibriumMuscle"};
```
chrisdembia: `arm26_Setup_CMC.xml` doesn't use Thelen2003Muscle except in the <defaults> section, which doesn't actually matter. It is confusing to have this replacement here, I think, because it makes it seem like more is happening. Either remove or give a comment clarifying that this block doesn't actually change the muscle used in the model.

Applications/CMC/test/testCMCArm26_Millard.cpp
```diff
@@ -0,0 +1,115 @@
...
+        // Find and replace result directory.
+        {
+            std::string orig{"Results_Arm26"};
+            std::string repl{"Results_Arm26_Millard"};
+            auto pos = line.find(orig);
+            if(pos != std::string::npos)
+                line.replace(pos, orig.length(), repl);
+        }
+        outFile << line << "\n";
+    }
```
chrisdembia: It would be cleaner to use the OpenSim API to edit the tool setup file rather than plain text files; this would make the intent of this code more clear (e.g., `AbstractTool::setResultsDir()`).

chrisdembia:
++1. I made some small comments. Other than those comments, LGTM.

Applications/CMC/test/testCMCArm26_Millard.cpp
klshrinidhi: Done.

If it possible to not have "Thelen..." in the setup xml file, it should be removed it from setup file.

Applications/CMC/test/testCMCArm26_Millard.cpp
klshrinidhi: Done.

Applications/CMC/test/testCMCArm26_Millard.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+#include <fstream>
+#include <thread>
+
+using namespace OpenSim;
+using namespace std;
+
+void testArm26() {
+    cout<<"\n******************************************************************" << endl;
+    cout << "*                             testArm26                          *" << endl;
```
tkuchida: testArm26 -> testArm26_Millard (to avoid confusion when reading output, particularly if both tests fail)?

Applications/CMC/test/testCMCArm26_Millard.cpp
```diff
@@ -0,0 +1,87 @@
...
+    cout << "******************************************************************\n" << endl;
+    CMCTool cmc("arm26_Setup_CMC.xml");
+    cmc.setResultsDir("Results_Arm26_Millard");
+    cmc.run();
+
+    Storage results("Results_Arm26_Millard/arm26_states.sto"), temp("std_arm26_states.sto");
+    Storage *standard = new Storage();
+    cmc.getModel().formStateStorage(temp, *standard);
+
+    Array<double> rms_tols(0.02, 2*2+2*6); // activations within 2%, angles within .6 degrees
```
tkuchida: I don't understand the "angles within .6 degrees" part of the comment. Perhaps that refers to lines 52–56, but it looks like those tolerances are either 0.05 degrees or 0.05 rad (=2.86 degrees).

Applications/CMC/test/testCMCArm26_Millard.cpp
```diff
@@ -0,0 +1,87 @@
...
+    const string& muscleType = cmc.getModel().getMuscles()[0].getConcreteClassName();
+    string base = "testArm26 "+ muscleType;
+
+    rms_tols[6] = 0.05;
```
tkuchida: Add comment indicating what these indices correspond to and what the units are? Also, it looks like lines 52–56 should appear immediately after line 48.

tkuchida:
`</review>` One additional minor comment: should `testCMCArm26` be renamed to `testCMCArm26_Thelen`? (Without a suffix, it looks like `testCMCArm26` is doing something categorically different than `testCMCArm26_Millard`.)

Applications/CMC/test/testCMCArm26_Millard.cpp
klshrinidhi: Done.

Applications/CMC/test/testCMCArm26_Millard.cpp
klshrinidhi: I don't know about this. Others will have to chime in.

I just split the previous code into two. Haven't changed anything if I did it right.

Applications/CMC/test/testCMCArm26_Millard.cpp
klshrinidhi: I actually don't know what these indices represent. Others will have to chime in.

chrisdembia: IMO it's fine to leave this as-is; the PR does not introduce any regression here. A separate PR could address the clarity of the original tests.

tkuchida: > A separate PR could address the clarity of the original tests.

I agree with opening a new issue to investigate and at least improve the comments here.

Applications/CMC/test/testCMCArm26_Millard.cpp
tkuchida: > Done.

My comment was about updating the string here so the output message agrees with the filename. Perhaps you forgot to push a commit?

klshrinidhi: Oh. Right. I just pushed the commit.

Corrected file header and output string.

tkuchida:
:+1: IMHO, can be merged once tests pass.

Applications/CMC/test/testCMCArm26_Millard.cpp
chrisdembia: I agree the defaults should be removed from the setup file, but that we should not worry about it for this PR.

Applications/CMC/test/testCMCArm26_Millard.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+#include <fstream>
+#include <thread>
+
+using namespace OpenSim;
+using namespace std;
+
+void testArm26() {
+    cout<<"\n******************************************************************" << endl;
+    cout << "*                     testArm26_Millard                          *" << endl;
```
tkuchida: Sorry I missed this before, but "testArm…" should be "testCMCArm…" here to match the filename (makes it easier to locate the file when the test fails). Similar comment applies to testCMCArm26_Thelen.cpp.

Issue #195, opened 2014-09-23
Allow installing OpenSim-Core into UNIX /usr/ or /usr/local


chrisdembia:
The number of unix users is increasing. The organization of our installation currently does not permit one to safely install into /usr/ or /usr/local. The following changes should occur:
- [x] On UNIX, do not copy over simbody libraries. #343 
- [x] Clean up CMake for api examples. #229
- [x] Clean up CMake for lepton. #327
- [x] Clean up "USING_NAMESPACE". #343 
- [x] Allow copying libraries to be an option. #343
- [ ] Create versioned libraries (SOVERSION), as is now done with Simbody.
- [x] Follow the [FHS](http://www.pathname.com/fhs/) filesystem hierarchy standard for the layout of the installation. This is facilitated by using CMake GNUInstallDirs. (gnu-install-dirs; chrisdembia:unix-install-2) #671
  - [x] configureOpenSim.m
  - [x] location of Java wrapping (look at vtk).
  - [x] location of python bindings (look at vtk). http://anonscm.debian.org/cgit/debian-science/packages/vtk6.git/tree/debian/python-vtk6.install, http://packages.ubuntu.com/trusty/amd64/python-vtk/filelist
- [x] Rename the executables so their names are distinguishable. Right now, e.g., linux already has an `id` executable, and our `id` would overwrite this. I suggest a single executable that takes a tool name as an argument, or renaming the executables as `opensim-ik`, and creating an `opensim` executable that calls `opensim-ik` when `opensim ik` is called (this is how command-line git works I think).

chrisdembia:
Reminder to self: https://github.com/chrisdembia/opensim-core/tree/unix-install-2

chrisdembia:
next branches to merge: unix-dont-copy-simbody and unix-install-2

moorepants:
@chrisdembia just came across this: http://click.pocoo.org

I think it is setup to do command line interfaces with lots of nested subcommands. Same guy that wrote Flask wrote this, so it is likely good. There example implements the git subcommands:

http://click.pocoo.org/3/complex/#complex-guide

Maybe you can use it for Opensim's command line access (although you'd have to provide Python then).

chrisdembia:
Awesome!! Thanks for sharing that. That's exactly what I was thinking about. Having to use python might be a serious limitation though.

moorepants:
I'm not sure how Python is a limitation :)

[...]

PR #1137, opened 2016-07-23
Unified command-line tool 

chrisdembia:
### Background and motivation

This PR introduces a single unified command-line tool, currently named `opensim`. The impetus for this change was the ability to install OpenSim into system-wide directories on UNIX systems (#195). However, the primary benefit in the long term will be the ability to add new functionality to the command-line tool. Right now, this new unified tool has the same abilities as the union of all our current command-line tools, but in the future, we can add new features (just ideas; some are not great ideas b/c they duplicate GUI functionality):

1. visualizing/simulating a model
2. converting data files between formats (c3d to trc) using DataAdapters.
3. package manager for OpenSim plugins, etc.
4. plotting data from a data file or states trajectory.
5. creating template files for making custom plugins (autogenerating myPluginDLL.h, etc.).
6. managing OpenSim "workbook" zip files (if/when we have this feature).

### Overview of the changes

This PR introduces the following changes:

1. Introduces the dependency `docopt.cpp` into the repo. This is a command-line argument parser whose input is the documentation of the command itself, which ensures consistency between the documentation and the parsing. The package `docopt` was originally a python package, but has been ported to numerous languages.
2. A new executable `opensim`, with the following subcommands:
  1. `opensim run-tool` for running any Tool (ik, rra, cmc, etc.).
  2. `opensim print-xml` for creating a template XML file for any Object, including tools.
  3. `opensim info` prints the names and description of properties for any Object.
  4. `opensim update-file` for upgrading XMLDocuments and Storage files (previously, `versionUpdate`).
3. A test case that tests most possible command invocations, and also checks for the correct error messages.
4. On UNIX/OSX, the old command-line tools are not built/installed anymore. The old command-line tools are still built/installed but ONLY on Windows, for backwards compatibility. In a future release, we can delete the old command-line tools.
5. A fix to a bug in `Object::makeObjectFromFile()`; this command changes the working directory to the directory containing the file, but previously would not change back to your original working directory.

Perhaps the best way to start reviewing the PR is to look at the documentation for the commands in `opensim-cli.cpp`, `opensim_run_tool.h`, etc.; and to read the test `testCommandLineInterface.cpp`, which shows a bunch of different invocations of the tool and snippets of the expected output.

I did plenty of manual testing (running the commands myself with different arguments) on OSX.

### Decisions

1. **Choice of command-line parsing library** I considered building the command-line tool in python and using a python library for the command-line parsing, but this would have created a headache for packaging OpenSim. I considered TCLAP, which is a very nice C++ library but it contains *lots* of files. Docopt.cpp strikes a good balance between functionality and size.
1. **copying `docopt.cpp` into the repo**: I had to decide if it was a good idea to copy `docopt` into our repo. I chose to do so because:
  1. `docopt.cpp` is not available in popular package managers yet (brew, apt, chocolatey).
  2. I could have used superbuild, but it's not clear that docopt has proper CMake support for this. Also, currently, Docopt requires cmake 3.1, but we currently require support of cmake 2.8.8. As discussed, I am fine with us moving to cmake 3.1 under some conditions.
  3. `docopt.cpp` is only a few files, so it is not too much hassle to copy it into the repo.
  4. Users need not worry about obtaining `docopt.cpp` on their own or through superbuild; resilience to where `docopt.cpp` is located.

  If we upgrade to cmake 3.1, I would be happy to try using docopt through superbuild instead of copying it into the repo. 

### To discuss

What should we name the command-line tool?

1. The name `opensim` makes sense for opensim-core, but could cause issues when combining with the GUI, since the GUI executable is also named `opensim[.exe]`. We could rename the GUI as `opensim-gui[.exe]`.
2. I don't like `opensim-cli`. 
3. `osim`: This is somewhat short which introduces a greater chance of name conflicts again, and it also could cause confusion with our model file extension? Though we also prefix our libraries with `osim`, so perhaps it's not a big issue.


Applications/opensim/opensim_info.h
```diff
@@ -0,0 +1,97 @@
...
+        Object::PrintPropertyInfo(std::cout, "", false);
+        return EXIT_SUCCESS;
+    }
+
+    // Class was provided.
+    const std::string& className = args["<class>"].asString();
+    const Object* object = Object::getDefaultInstanceOfType(className);
+    if (object == nullptr) {
+        throw Exception( "No registered class with name '" + className +
+                "'. Did you intend to load a plugin?");
```
tkuchida: :+1: Helpful error message.

Applications/opensim/opensim_print_xml.h
```diff
@@ -0,0 +1,103 @@
...
+            true); // show help if requested
+
+    // Parse arguments.
+    // ----------------
+
+    // Tool or class.
+    std::string toolOrClass = args["<tool-or-class>"].asString();
+    std::string className;
+    if      (toolOrClass == "scale")   className = "ScaleTool";
+    else if (toolOrClass == "ik")      className = "InverseKinematicsTool";
```
tkuchida: Perhaps more user-friendly to check for uppercase "IK" as well.

Applications/opensim/opensim_run_tool.h
```diff
@@ -0,0 +1,97 @@
...
+
+#include <iostream>
+
+#include <docopt.h>
+#include "parse_arguments.h"
+
+#include <OpenSim/OpenSim.h>
+
+static const char HELP_RUN_TOOL[] = 
+R"(Run a tool (IK, CMC, ...) from an XML setup file.
```
tkuchida: opensim_print_xml.h demands lowercase "ik", "cmc", etc., so perhaps use lowercase throughout.

Applications/opensim/test/testCommandLineInterface.cpp
```diff
@@ -0,0 +1,449 @@
...
+        testCommand("--library x", EXIT_FAILURE, output);
+        testCommand("-L=x", EXIT_FAILURE, output);
+        testCommand("--library=y", EXIT_FAILURE, output);
+        testCommand("-L x --library y -L z", EXIT_FAILURE, output);
+        testCommand("-L=x --library=y -L=z", EXIT_FAILURE, output);
+    }
+
+    // Unrecognized command.
+    // =====================
+    testCommand("bleepbloop", EXIT_FAILURE,
```
tkuchida: :space_invader:

CHANGELOG.md
```diff
@@ -36,6 +36,9 @@ Converting from v3.x to v4.0
   other components they depend on (instead of string properties):
   - ContactGeometry (ContactSphere, ContactHalfSpace, ContactMesh)
 - Many of the methods in ScaleTool have now been marked const.
+- We created a new unified command line interface that will replace the
+  single-tool command line executables (`scale`, `ik`, `id`, `rra`, `cmc`,
+  etc.).
```
tkuchida: Perhaps indicate the change users will need to make if they're currently calling the command-line executables from their code.

tkuchida:
`</review>` Minor stuff only. I really like having a single `opensim` executable.

chrisdembia:
Thanks for the review, Tom. Any preference for the name of the executable?

tkuchida:
> Any preference for the name of the executable?

I like `opensim` and renaming the GUI to `opensim-gui`. I don't know about other platforms, but I would expect Windows users to run the GUI from the Start Menu or a shortcut and wouldn't ever need to type the `-gui` suffix.

moorepants:
Can you have `opensim

moorepants:
Can you have `opensim` open the gui if it is installed and simply print the help if it isn't? Then all other commands would be accessed via a sub-command? This is how we did it in the yeadon package.

chrisdembia:
Yeah we could do that.

klshrinidhi:
The docopt library should not be committed to the repository for following reasons

* This makes it look like we intend to fork, develop and maintain the dependency in our repository. This is certainly not our intention, as I understand. Unless I do a file comparison, I right now cannot know if you did edit docopt sources.
If you do intend fork this dependency, please do so as a separate project and submit a PR to the original docopt. If your changes are not getting accepted by docopt authors, please maintain this project as a separate repository so we can efficiently track the changes. This will also help us update docopt from original in future. 
* If we suck docopt into OpenSim, future contributors could edit the dependency sources in order to fix a bug or implement a feature. This is also not our intention as this will make updating the dependency harder with merge conflicts. We cannot always remember to not accept a PR if it edited docopt sources.
* We will surely want to keep the dependency updated with bug fixes and features. Every time, one will have to copy sources from original repository and commit it here, as if the person himself authored the changes.
* I do not like the fact that git thinks you authored those files.
* There can also be a differences in licenses of the dependency and OpenSim itself and I am not sure if this needs to be stated explicitly in OpenSim. This should just be avoided altogether.

Overall, sucking other projects into our repository burdens us and we should avoid this. I think all docopt files should be removed from this PR and and add docopt as a build dependency. This makes it easier for future contributors to understand the dependency graph well enough to suggest build system changes.

OpenSim source releases could contain sources for all dependencies. If we keep all old releases (sources & binaries) around, users should be able to download and use them anytime.

I will review the PR once the changes I mention above are made.

klshrinidhi:
I can make the changes I have mentioned.

chrisdembia:
If you'd like, sure :). We can discuss in person. Could you resurrect your PR to increase the minimum required CMake version?

After docopt is integrated into superbuild, I can also squash all the commits so as to avoid having docopt in our history.

klshrinidhi:
Sounds good. Will get back soon.

aymanhab:
@chrisdembia Not sure about possible issues with naming the executable because: earlier installations on windows will have opensim.exe refers to the GUI executable. Renaming the executable for the GUI may not be as trivial as it sounds since it's the name of a set of modules that refer to each other through xml configuration files all over the place, need to check if feasible/sensible first, maybe easy/straightforward or not, don't know yet.

chrisdembia:
I am fine with using a different name if necessary. I am also happy to do the investigation with the GUI to see if we can change the name of the GUI executable.

chrisdembia:
Please let me know if you would like to do it or if you would like me to do it.

Applications/opensim/test/testCommandLineInterface.cpp
chrisdembia: :)

PR #1145, opened 2016-07-28
clean up header includes and forward declarations

carmichaelong:
This PR is motivated by #682 to speed up build times by removing unused #include's, using more specific #include's, and using forward declarations if possible. Changes in this PR were limited to the Libraries (i.e. no effort was made to clean up Test, Applications, etc.). For my local setup, doing a 32-bit build with Java wrapping, times dropped from ~14 minutes to ~11 minutes.


tkuchida:
Warning to reviewers:

warning

chrisdembia:
Woohoo! ~14 to ~11 is pretty good I think. I'll look through this after the tests finish.

aseth1:
This is pretty awesome! Was it easy using [iwyu](http://include-what-you-use.org/downloads/)?

carmichaelong:
Not exactly, mainly because our heavy use of macros tricked iwyu many times into thinking incomplete types were fine. Still, it was pretty helpful in general with heading me in the right direction and sped up the process a bit.

OpenSim/Simulation/Model/ExpressionBasedCoordinateForce.cpp
```diff
@@ -26,6 +26,8 @@
 //=============================================================================
 #include "ExpressionBasedCoordinateForce.h"
 #include <OpenSim/Simulation/Model/Model.h>
+#include <Vendors/lepton/include/lepton/Parser.h>
+#include <Vendors/lepton/include/lepton/ParsedExpression.h>
```
chrisdembia: While this works, [I think you should be able to remove the `Vendors/`](https://github.com/opensim-org/opensim-core/blob/master/cmake/OpenSimMacros.cmake#L233).

OpenSim/Simulation/Model/ExpressionBasedCoordinateForce.h
```diff
@@ -24,7 +24,7 @@
  * -------------------------------------------------------------------------- */
 // INCLUDE
 #include "Force.h"
-#include <Vendors/lepton/include/Lepton.h>
+#include <Vendors/lepton/include/lepton/ExpressionProgram.h>
```
chrisdembia: See above comment.

OpenSim/Simulation/Model/ExpressionBasedCoordinateForce.cpp
chrisdembia: Actually, nevermind. I was incorrect. This is fine.

OpenSim/Simulation/Model/ExpressionBasedCoordinateForce.h
chrisdembia: Ignore.

OpenSim/Simulation/Model/FunctionBasedBushingForce.h
```diff
@@ -25,7 +25,6 @@
 
 
 // INCLUDE
-#include <OpenSim/Simulation/osimSimulationDLL.h>
```
chrisdembia: I think it is okay to leave in headers like this. Otherwise, we are relying on other headers (like Force.h) to have included this. Doesn't matter too much. The important thing is that `OSIMSIMULATION_API` is defined in this header.

chrisdembia:
I read through the files I could view. The only question I have is if `#include "OpenSim/Simulation/osimSimulationDLL.h"` should be removed. It seems to be removed from files that do need it, but it just so happens that another include also includes this file.

carmichaelong:
> I read through the files I could view.

Yeahh my bad... didn't realize doing all of these files at once would do this. I guess one way to work around that would just be to go to the branch and look at files that way? At least all the changes are limited to the top.

> The only question I have is if #include "OpenSim/Simulation/osimSimulationDLL.h" should be removed. It seems to be removed from files that do need it, but it just so happens that another include also includes this file.

This is something I went back and forth on, especially since this shouldn't be that big of an issue (unless you argue opening/closing files is super costly). The way I thought of it is that, for instance, say a file is in the Simulation folder, as long as it derives from another class in Simulation, I removed it. I'm happy to revert those changes, though.

chrisdembia:
I think we should merge this. Does anyone else want to review?

tkuchida:
> Does anyone else want to review?

I scanned through the files displayed in GitHub and didn't see anything suspicious. All tests pass, so I think it's safe to merge.

chrisdembia:
Cool. Thanks @carmichaelong .

carmichaelong:
Thanks for reviewing/merging!

PR #1147, opened 2016-07-30
[WIP] Component interface for providing implicit form of dynamics

chrisdembia:
This PR is a work-in-progress. There is a lot more testing and documenting to do. However, the key elements of the interface are in place.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/1147)


nickbianco:
@chrisdembia, I assume there is no plan for continued development here since is very out-of-date, but just confirming before I close.

PR #1155, opened 2016-08-02
Catch unsigned int wrap through zero following initSystem test in testComponents

### LINKED ISSUE #1152: testComponents incorrectly reported 1e12% memory leak
See [this failure](https://ci.appveyor.com/project/opensim-org/opensim-core/build/4136/job/7ro35rlu4ek12add#L1573) on AppVeyor (64-bit). #1020 appeared to fix this issue most of the time, but I believe there is still a subtle issue in [testComponents.cpp, line 312](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Tests/Components/testComponents.cpp#L312):
``` cpp
const size_t increaseInMemory = getCurrentRSS() > initMemory ?
    getCurrentRSS() - initMemory : 0;
```
It is possible for memory to be freed behind-the-scenes between evaluating the condition (`getCurrentRSS() > initMemory ?`) and the consequence if true (`increaseInMemory = getCurrentRSS() - initMemory`), thereby still resulting in the unsigned int wrapping through zero. We should call `getCurrentRSS()` _once_ and use the stored value in both the condition and the assignment if true. Self-assigned.

tkuchida:
Analogous to #1020 for copy/delete test. Also, now calling `getCurrentRSS()` only once when calculating increase in memory to guarantee that wrapping through zero is impossible. Updated messages in test 10 to match test 11. Fixes #1152

### DIFF
--- a/OpenSim/Tests/Components/testComponents.cpp
+++ b/OpenSim/Tests/Components/testComponents.cpp
@@ -309,21 +309,22 @@ void testComponent(const Component& instanceToTest)
 
         // Catch a possible decrease in the memory footprint, which will cause
         // size_t (unsigned int) to wrap through zero.
-        const size_t increaseInMemory = getCurrentRSS() > initMemory ?
-                                        getCurrentRSS() - initMemory : 0;
+        const size_t finalMemory = getCurrentRSS();
+        const size_t increaseInMemory = finalMemory > initMemory ?
+                                        finalMemory - initMemory : 0;
         const long double leakPercent = (100.0*increaseInMemory/instanceSize)
                                         /nCopies;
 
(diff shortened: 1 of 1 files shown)


tkuchida:
Can we review/merge this so we don't keep seeing the spurious failure on AppVeyor?
@klshrinidhi, @carmichaelong

klshrinidhi:
I just looked at this. :+1:

carmichaelong:
LGTM. Feel free to merge (or am I allowed to merge?)

tkuchida:
> or am I allowed to merge?

Second reviewer can merge; gopher it! Thanks for reviewing.

PR #1156, opened 2016-08-03
Throw exception if user forgot to call initSystem()

### LINKED ISSUE #729: Component::getStateVariableNames() returns empty Array if initSystem() hasn't been called.
Either `getStateVariableNames()` should throw an exception if `!hasSystem()`, or the requirement on the system should be removed. I think @aseth1 is planning the latter solution.

tkuchida:
Added checks to ensure initSystem() has been called before attempting to access state, discrete, or cache variables owned by a Component or its descendants. An exception will be thrown that instructs the user to call initSystem() first (rather than reporting that the specified state variable was not found, for example). Added tests to testComponentInterface.cpp. Fixes #729 and propagated to include the following methods:

StateVariable
-------------
* findStateVariable
* getNumStateVariables
* getStateVariableNames
* getStateVariableValue, setStateVariableValue
* getStateVariableValues, setStateVariableValues
* getStateVariableDerivativeValue

DiscreteVariable
----------------
* getDiscreteVariableValue, setDiscreteVariableValue

CacheVariable
-------------
* getCacheVariableValue, setCacheVariableValue, updCacheVariableValue
* markCacheVariableValid, markCacheVariableInvalid, isCacheVariableValid

### DIFF
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -566,7 +566,7 @@ unsigned Component::printComponentsMatching(const std::string& substring) const
 int Component::getNumStateVariables() const
 {
     // Must have already called initSystem.
-    OPENSIM_THROW_IF(!hasSystem(), ComponentHasNoSystem, getName().c_str());
+    OPENSIM_THROW_IF_FRMOBJ(!hasSystem(), ComponentHasNoSystem);
 
     //Get the number of state variables added (or exposed) by this Component
     int ns = getNumStateVariablesAddedByComponent(); 
@@ -727,7 +727,7 @@ const Component::StateVariable* Component::
     findStateVariable(const std::string& name) const
 {
     // Must have already called initSystem.
-    OPENSIM_THROW_IF(!hasSystem(), ComponentHasNoSystem, getName().c_str());
+    OPENSIM_THROW_IF_FRMOBJ(!hasSystem(), ComponentHasNoSystem);
 
     // first assume that the state variable named belongs to this
     // top level component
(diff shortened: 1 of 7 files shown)


OpenSim/Common/Component.cpp
```diff
@@ -565,6 +565,9 @@ unsigned Component::printComponentsMatching(const std::string& substring) const
 
 int Component::getNumStateVariables() const
 {
+    // Must have already called initSystem.
+    OPENSIM_THROW_IF(!hasSystem(), ComponentHasNoSystem, getName().c_str());
```
aseth1: Consider using `OPENSIM_THROW_IF_FRMOBJ` so you don't to pass in the name and also report the concrete Class Name, which can be helpful.

tkuchida:
@aseth1: testForward.cpp is failing [at line 118](https://github.com/opensim-org/opensim-core/blob/master/Applications/Forward/test/testForward.cpp#L118) (specifically, when `ForwardTool::run()` calls `Model::initSystem()` [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Tools/ForwardTool.cpp#L258)). `Model::buildSystem()` ends up calling `Component::getNumStateVariables()` on two Components that don't have a System ("GroundJoint" and "pendulum"), which will throw the new exception (previously, `getNumStateVariables()` would quietly return 0 if `!hasSystem()`). Explicitly returning 0 from `getNumStateVariables()` if `!hasSystem()` gets the test to pass, but I think `getNumStateVariables()` should be throwing an exception if `!hasSystem()`. I'm guessing that ForwardTool isn't correctly initializing the model somehow, but I haven't been able to untangle. Any hints?

aseth1:
@tkuchida I agree that throwing the Exceptions is much better and that now you have exposed legitimate bugs. Why those two Components is kind of baffling. "GroundJoint" is a `FreeJoint` and "pendulum" is a `Body`? Who's asking for the number of state variables and when? Either these two components are failing to hang on to a reference to the System (so apparently having no System) or the request for the values is happening at an incorrect time, in which case all the Components should be screaming the same thing. If you have it running on your machine swing by and grab me- might be easier to step through together.

tkuchida:
All tests now pass.

> Consider using `OPENSIM_THROW_IF_FRMOBJ` so you don't have to pass in the name and also report the concrete Class Name, which can be helpful.

Fixed.

aseth1:
LGTM.  Thanks @tkuchida

tkuchida:
Thanks for merging.

PR #1158, opened 2016-08-04
Add STO adapters for Vec3, Vec6 & SpatialVec

klshrinidhi:
Adding STO adapters for Vec3, Vec6 & SpatialVec.

Similar adapter for Transform cannot be done with current scheme of Adapters spitting out TimeSeriesTables because SimTK::Matrix_<Transform> cannot be created.


OpenSim/Common/STOFileAdapter.h
```diff
@@ -27,9 +27,35 @@
...
+                        const std::string& func) :
+        Exception(file, line, func) {
+        std::string msg = "DataType not specified in the header.";
+
+        addMessage(msg);
+    }
+};
+
 /** STOFileAdapter is a DelimFileAdapter that presets the delimiters 
 appropriately for STO files.                                                  */
```
chrisdembia: I think this class description (or the description for DelimFileAdapter) could explain that this file format handles a few data types, and perhaps even give a snippet of what a file might look like:

```
DataType=double
endheader
time marker1 marker2
```

chrisdembia:
Should we do anything about handling the sto file version? Perhaps the STOFileAdapter should print the STO version as 2 so that the Storage class will *not* try to read Storage files with version 2.

aseth1:
@chrisdembia Yes to both points. We should prevent the Storage class from trying to read new .STO versions and it should provide a message that it (Storage) cannot handle new file versions and to use the appropriate (STO) FileAdapter.

OpenSim/Common/DelimFileAdapter.h
```diff
@@ -25,16 +25,53 @@
...
 this class.                                                                   
 Header in the file is assumed to end with string "endheader" occupying a full
 line.                                                                         */
+template<typename T>
 class OSIMCOMMON_API DelimFileAdapter : public FileAdapter {
+    static_assert(std::is_same<T, double>::value ||
+                  std::is_same<T, SimTK::Vec3>::value ||
+                  std::is_same<T, SimTK::Vec6>::value ||
+                  std::is_same<T, SimTK::SpatialVec>::value,
+                  "Template argument T must be one of the following types : "
```
aseth1: I am fine with this limitation, but I think it is possibly too strict. I think any `Vec#` so that even `Quaternion` (`Vec4`) and `UnitVec` (Vec3 of norm 1) would be inherently valid. It is understandable that Matrix types are not supported, so perhaps a check that is not of type Matrix or Vector (not fixed sized) would be more adequate? Let me know if there are other reasons to limit to these 4 types specifically (which would still be great!).

klshrinidhi:
* Added format information to doxygen comments.
* Added version number (= 2) to STOFileAdapter. Files written by STOFileAdapter will have this version number. Storage will throw when reading files with incompatible version number.
* Added support for Vec2, Vec4, Vec5, UnitVec and Quaternion.

OpenSim/Common/DelimFileAdapter.h
```diff
@@ -47,11 +93,20 @@ class OSIMCOMMON_API DelimFileAdapter : public FileAdapter {
...
+                     const std::string& compDelimWrite);
+
     DelimFileAdapter* clone() const override;
 
     /** Key used for table associative array returned/accepted by write/read. */
     static const std::string _table;
 
+    /** Name of the data type T (template parameter).                         */
+    static const std::string dataTypeName();
+
```
aseth1: Did you consider a generic implementation, like:
```cpp
return SimTK::NiceTypeName<T>::namestr(); 
```
instead of having to specialize below?

OpenSim/Common/DelimFileAdapter.h
```diff
@@ -22,19 +22,65 @@
...
 given delimiters. CSVFileAdapter and MOTFileAdapter derive from this class and
 set the delimiters appropriately for the files they parse. The read/write
 functions return/accept a specific type of DataTable referred to as Table in 
 this class.                                                                   
 Header in the file is assumed to end with string "endheader" occupying a full
 line.                                                                         */
+template<typename T>
 class OSIMCOMMON_API DelimFileAdapter : public FileAdapter {
+    static_assert(std::is_same<T, double           >::value ||
+		  std::is_same<T, SimTK::Vec2      >::value ||
```
aseth1: Is there a generic way to test for any specialization of `Vec<T>`?

OpenSim/Common/DelimFileAdapter.h
klshrinidhi: I did. There needs to be *guarantee* that the string produces is going to be *exactly same*:
* Across platforms.
* Across compilers.
* Over time.

I think the current implementation does this.

OpenSim/Common/DelimFileAdapter.h
klshrinidhi: The implementation currently supports only Vec2, Vec3, Vec4, Vec5 & Vec6, The check here is according to that.

I was initially asked to support Vec3, Vec6 and SpatialVec (as title of this PR says) only and so I wrote explicit specializations to accommodate this. Now you asking me to support any Vec<T>. Not sure what you exactly want and how much of it is justified as use cases. If you want support for any Vec<T>, I will have to change the implementation to partial specializations.

OpenSim/Common/DelimFileAdapter.h
chrisdembia: Good point--`NiceTypeName<T>::namestr()` gives different results on different platforms.

aseth1: @klshrinidhi @chrisdembia thanks for the clarification.

OpenSim/Common/DelimFileAdapter.h
```diff
@@ -60,17 +115,650 @@ class OSIMCOMMON_API DelimFileAdapter : public FileAdapter {
...
+           << _compDelimWrite << std::setprecision(prec) << elem[2];
+}
+
+template<>
+inline void 
+DelimFileAdapter<SimTK::Vec4>::writeElem(std::ostream& stream,
+                                         const SimTK::Vec4& elem,
+                                         const unsigned& prec) const {
+    stream << std::setprecision(prec) << elem[0];
+    for(auto i = 1u; i < 4; ++i)
```
aseth1: Similarly why not use `Vec<M>::size()` to handle the supported `Vec` types instead of specializing for every type?

klshrinidhi: Again, because the initial goal for this PR was to add just 3 new types, I wrote *Full Specializations*. When you asked to support Quaternion and UnitVec, I just wrote more of them.

OpenSim/Common/DelimFileAdapter.h
```diff
@@ -60,17 +115,650 @@ class OSIMCOMMON_API DelimFileAdapter : public FileAdapter {
...
+    SimTK::RowVector_<double> elems{static_cast<int>(tokens.size())};
+    for(auto i = 0u; i < tokens.size(); ++i)
+        elems[static_cast<int>(i)] = std::stod(tokens[i]);
+
+    return elems;
+}
+
+template<>
+inline SimTK::RowVector_<SimTK::Vec2>
+DelimFileAdapter<SimTK::Vec2>::readElems(
```
aseth1: Again, the specializations may be redundant if used the size of the `Vec#` in the implementation instead of specializing for every `#`. Why would that not work or be less desirable?

OpenSim/Common/STOFileAdapter.h
```diff
@@ -27,28 +27,141 @@
...
+};
+
 /** STOFileAdapter is a DelimFileAdapter that presets the delimiters 
-appropriately for STO files.                                                  */
-class OSIMCOMMON_API STOFileAdapter : public DelimFileAdapter {
+appropriately for STO files. The format of the file is as follows:
+\code
+.................................
+.................................
+........ <metadata> .............
```
aseth1: Would be good to briefly describe what metadata is and how it should be specified, e.g:  `<key>=<value>`.

OpenSim/Common/DelimFileAdapter.h
klshrinidhi: I think I have responded to why these are specializations in previous comments. Not sure why you keep making the same comment over and over again. They are extension of what I initially wrote -- support 3 more types in title of this PR only. Suggesting to use partial specialization is basically asking support for all Vec<>.

OpenSim/Common/STOFileAdapter.h
klshrinidhi: I have provided an example of how this looks like in the same comment.

OpenSim/Common/STOFileAdapter.h
```diff
@@ -27,28 +27,141 @@
...
+.................................
+............ <data-rows> ........
+.................................
+.................................
+\endcode
+For example a TimeSeriesTable_<SimTK::Vec3> could look like:
+\code
+units=m
+DataType=Vec3
+endheader
```
aseth1: This example is very helpful. Are keys case sensitive.  If so should we have a convention that key name begin with uppercase and should be camel case for longer names: e.g.: `DataType` vs `dataType` and `data_type` (which is the convention for properties).

klshrinidhi: What goes in metadata is not in our control. The key-value pairs we insert will be of the format we decide. The key-value pairs users insert can be different. Therefore the convention (if any) does not apply to the entire header.
For now, I don't think this is important.

OpenSim/Common/Storage.cpp
```diff
@@ -130,6 +130,9 @@ Storage::Storage(const string &aFileName, bool readHeadersOnly) :
     // Motion files from SIMM are in degrees
     if (_fileVersion < 1 && (0 == aFileName.compare (aFileName.length() - 4, 4, ".mot"))) _inDegrees = true;
     if (_fileVersion < 1) cout << ".. assuming rotations in " << (_inDegrees?"Degrees.":"Radians.") << endl;
+    if(_fileVersion > 1)
```
aseth1: `fileVersion = 1` is intended to pass through normally, correct?

klshrinidhi: Yes. This can be seen working in various tests because many STO files in repository contain version = 1 in metadata.

OpenSim/Tests/ControllerExample/testControllerExample.cpp
```diff
@@ -32,10 +32,16 @@
 using namespace OpenSim;
 using namespace std;
 
+void changeVersionNumber(const std::string& filenameOld,
+                         const std::string& filenameNew);
```
aseth1: Instead of having to convert in the tests, some alternate approaches:
 1. have the option to force the version number on write (Manager and Tools must be updated to provide the version #) 
 2. write STOs of type `double` as version 1 by default and only non doubles as version 2
 3. add the conversion code to shared testing utilities 

My preference would be 1 or 2.

OpenSim/Tests/ControllerExample/testControllerExample.cpp
```diff
@@ -32,10 +32,16 @@
...
+
 int main()
 {
     try {
-        Storage result1("tugOfWar_controls.sto"), 
+      const std::string result1Filename{"tugOfWar_controls.sto"};
+      const std::string result1FilenameV2{"tugOfWar_controls_v2.sto"};
+      changeVersionNumber(result1Filename, result1FilenameV2);
```
aseth1: The naming makes this a bit hard to follow. One would expect  `result1FilenameV2` to be v2 or an update in versioning but you are in fact downgrading the version from 2 to 1.

OpenSim/Common/DelimFileAdapter.h
aseth1: > Not sure why you keep making the same comment over and over again.

I raise it repeatedly because it repeatedly indicates the code duplication could be reduce significantly in multiple places and it isn't an isolated benefit. 

> Suggesting to use partial specialization is basically asking support for all Vec<>.

Given all the types you are handling that are all `Vec<>` why would this be bad or wrong to do? Once you start seeing code duplication why not adapt?

klshrinidhi: If you were clear on the goals of this PR from the beginning, it would have saved me a lot of my time writing and re-writing code. I submit the PR and you decide you want to support more types. I wrote the implementation for the initial goal.

I will change the implementation to partial specialization and get back. Before, I do that I need to know EXACTLY what the goals of this PR are. I am not interested in keep re-doing things.

aseth1:
I am done my review.  Issues:
 1. code repetition in the full specialization, which is not necessary. That should have become obvious when you had to extend the specializations to handle more types. Why is code duplication a superior solution to partial specialization handling Vec<> generically?
 2. the necessity of adding conversion code to down version Storage files produced by test cases. Here there is the issue of having to do the conversion (opposed to writing version 1 files instead of 2 to begin with) and secondarily that the  conversion code (if necessary) should not be re-implemented in every test.

OpenSim/Common/STOFileAdapter.h
aseth1: The example does not state that that the metadata is composed of key value pairs. One could imply that but it just seems more clear to state it upfront.

OpenSim/Common/STOFileAdapter.h
aseth1: >The key-value pairs we insert will be of the format we decide.

I think a convention for key names for metadata that we use would be helpful. We should decide what that should be.  We did the same for the names of properties (e.g. lowercase and underscore separated terms). Since the serialization of metadata is being introduced here I think it is appropriate to discuss/decide here.

aseth1: I have no preferences other than it should be distinct from property name.  Starting uppercase with following camel-case is close to what you already have:
```
Units=m
DataType=Vec3
```

klshrinidhi: The only metadata we introduce are DataType and Version. Rest are user controlled metadata. I don't think this is important enough to hold this PR.

The other issue is that not all metadata is serializable. When DataTable was first implemented, I was told there was no requirement for metadata to be serializable.

klshrinidhi: > I have no preferences other than it should be distinct from property name.  Starting uppercase with following camel-case is close to what you already have:
```
Units=m
DataType=Vec3
```

The example here includes user's metadata. Particularly, `units` is user inserted metadata. We should not be changing it. Like I said before, this is not important at this point to hold this PR.

klshrinidhi:
> code repetition in the full specialization, which is not necessary. That should have become obvious when you had to extend the specializations to handle more types. Why is code duplication a superior solution to partial specialization handling Vec<> generically?

I am not sure you understand what this PR was initially supposed to be and why I wrote full specializations in the first place. You changed the goals/design during your review (which is never a good idea and results in re-work). As a result, I chose to add more specializations to support more types. If you were clear on what was required from the beginning, I would have wrote the implementation accordingly. If you review the implementation based on its original goals, there is nothing wrong here.
I don't see this as a good review process in general.

> the necessity of adding conversion code to down version Storage files produced by test cases. Here there is the issue of having to do the conversion (opposed to writing version 1 files instead of 2 to begin with) and secondarily that the conversion code (if necessary) should not be re-implemented in every test.

I will move the conversion function to a shared location.

klshrinidhi:
* I have reduced redundancy in code with reading/writing elems. Now all `SimTK::Vec<>` are read/written using same code.
* I have moved the function (that changes STO the version number to 1) to a header to avoid redundancy.  Once we remove Storage, this function can be easily removed as well.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
```diff
@@ -201,4 +205,20 @@ OpenSim::Object* randomize(OpenSim::Object* obj)
      return obj;
 }
 
+// Change version number of the file to 1 so that Storage can read it. This
+// function can be removed when Storage class is removed.
```
aseth1: Thank you for putting the version changing code in one place. I agree that it can be removed when our dependency on the Storage class is removed.  Please create an issue to replace the use of Storage with DataTable for comparing results of of test cases.

The comment should say that `Storage` can only read version 1 or earlier .STO files and that existing tests must revert the version number for newly created files in order to use the Storage class comparison methods.

Also renaming to:
 `revertToVersionNumber1(const std::string& currentVersionFilename,  const std::string& version1Filename)` 
would also be more informative.

OpenSim/Common/Test/testSTOFileAdapter.cpp.orig
```diff
@@ -0,0 +1,266 @@
+/* -------------------------------------------------------------------------- *
```
aseth1: Looks like this file has been mistakenly included.

aseth1:
Just a few easy comments to address: 
 1. clarify the name and comment of the utility function to revert current .STO files (version 2) to version 1 that enables tests to still use the Storage class for data comparison.
 2. open an issue to convert the use of Storage for test result comparisons to methods/utilities that use DataTable (or TimeSeriesTable)
 3. remove the accidentally included diff file

We need a second reviewer. @chrisdembia and @aymanhab you may also have more feedback on usage.

klshrinidhi:
> clarify the name and comment of the utility function to revert current .STO files (version 2) to version 1 that enables tests to still use the Storage class for data comparison.

Added comments.

> open an issue to convert the use of Storage for test result comparisons to methods/utilities that use DataTable (or TimeSeriesTable)

Done.

> remove the accidentally included diff file

Done.

OpenSim/Common/DelimFileAdapter.h
```diff
@@ -22,19 +22,72 @@
...
+
 /** DelimFileAdapter is a FileAdapter that reads and writes text files with
 given delimiters. CSVFileAdapter and MOTFileAdapter derive from this class and
 set the delimiters appropriately for the files they parse. The read/write
 functions return/accept a specific type of DataTable referred to as Table in 
 this class.                                                                   
 Header in the file is assumed to end with string "endheader" occupying a full
 line.                                                                         */
+template<typename T>
 class OSIMCOMMON_API DelimFileAdapter : public FileAdapter {
```
aseth1: Since the `DelimFileAdapter` is in header defined there is no need for the dllimport/export decls defined in `OSIMCOMMON_API`. In fact, this may be causing the AppVeyor [failure](https://ci.appveyor.com/project/opensim-org/opensim-core/build/4547/job/ef5e3h7aiu3sr09t#L303) since the complete definition is in this module when it is expecting it elsewhere.

aseth1:
LGTM. Just need to sort out AppVeyor failure.

jenhicks:
@aymanhab Can you review this PR?

aymanhab:
Looks good @klshrinidhi, what jumped at me looking at the code was:
1. While use of uniform initialization (with {}) for classes is justified and solves a problem, using it for Strings (e.g. in many test***examples.cpp) may not be necessary and we can avoid especially in user facing code since the syntax doesn't translate to scripting users. Not sure how prevalent is this but in these situations it is a coding style issue and so is subjective (stackoverflow http://programmers.stackexchange.com/questions/133688/is-c11-uniform-initialization-a-replacement-for-the-old-style-syntax)
2. Case sensitivity, while non-blocking to this PR should be addressed (e.g. by documentation and making sure errors/exceptions indicate the exact/expected match for supported types are Vec3, Vec6,...) or some other means.

Feel free to merge after opening issues to discuss these items.

Issue #335, opened 2015-01-26
Manager constructor that doesn't take integrator leads to exception


chrisdembia:
```
simbody/SimTKmath/Integrators/src/TimeStepperRep.h:66: SimTK::Integrator &SimTK::TimeStepperRep::updIntegrator(): Assertion `integ' failed.
```

I guess we have no use cases of this in the code.

See related https://github.com/opensim-org/opensim-core/pull/106

sherm1:
Per Ajay: would be better to replace this with a sensible Solver component in 4.0.

chrisdembia:
#156.

mizimmer:
I've just run into this error using the Java bindings.  Is there a workaround, e.g. a class other than Manager that can be used to integrate a model in time?  A change to the underlying C++ code might be in order...

chrisdembia:
Yes; directly use Simbody classes to do the integration (like [here](https://github.com/opensim-org/opensim-core/blob/ea91ce0fc8759fc249d26b232695f00e33b6e5ed/OpenSim/Common/Test/testDelay.cpp#L187)).

mizimmer:
chrisdembia, thanks so much for your reply!  More precisely, my issue is that the Java bindings do not expose any SimTK implementations of Integrator, so I cannot instantiate a Manager with an existing instance of Integrator.  Using the Manager constructor without an Integrator (which is presently my only option) leaves the manager with a NULL pointer to an Integrator, so I get a C++ segfault when calling methods like doIntegration(...) from Java.

There is also something strange about the SWIG-generated getIntegrator() method in the Manager Java class.  It creates a new Integrator and returns it; is this the expected behavior?

I wonder if it would be possible to add a convenience method to the Manager C++ class to generate a new Integrator internally. Another option would be to expose the RungeKuttaMersonIntegrator, etc. to SWIG for Java wrapping.

I could help if you have some guidance on which solution might be best.  I'm trying to integrate OpenSim into a couple of ongoing research projects and am happy to dive deep into the code base.

aymanhab:
@mizimmer Wonderful to have you on board. Should I assume you're working with the master branch? If so, can you describe at a high level what you're trying to accomplish so we know if you'll need more of the simbody internals?

mizimmer:
[...]

Issue #156, opened 2014-08-08
Allow easy integration of Model state in Matlab


aymanhab:
In version 3.2, users could instantiate a Manager(model) from Matlab and then integrate forward using a default integrator with minimal overhead. This functionality has been removed in master branch while refactoring the code. Either we need to restore this functionality to Manager or provide an easy alternative that doesn't involve the Manager class altogether.

chrisdembia:
Use unique_ptr and swap.

PR #1163, opened 2016-08-09
Restore internally-managed integrator to Manager, for use in MATLAB/Python

chrisdembia:
We used to have a constructor to Manager that created an Integrator internally.
This had been done in a way that led to memory leaks. We fixed the leak (#105) but
introduced a worse bug wherein the Manager no longer could create an Integrator
internally, even though the documentation said it could. Not a show-stopper for C++users, but this meant that MATLAB/Python users
could no longer use the Manager.

With this PR, the Manager can create its
own internal integrator, allowing MATLAB/Python users to use it (while also avoiding memory leaks).

I know that @aseth1 would prefer to not make improvements to Manager since we are planning to create something that replaces it. And in working on this PR, I definitely understood that sentiment, and acknowledge that a replacement or cleanup is necessary. However, it seems unlikely that we will have that in place for 4.0, and this is an easy-to-fix bug that has come up many times. Put another way, I think it would be bad if we release 4.0 and MATLAB/Python users are unable to easily integrate their models (the ForwardTool is not a sufficient workaround). I avoided a general cleanup of the Manager and tried to limit myself mostly to just fixing this bug.


aseth1:
> However, it seems unlikely that we will have that in place for 4.0, and this is an easy-to-fix bug that has come up many times. Put another way, I think it would be bad if we release 4.0 and MATLAB/Python users are unable to easily integrate their models (the ForwardTool is not a sufficient workaround).

Point taken and I agree. One idea (not for this PR) is to slowly refactor the `Manager` and evolve it into the new `ForwardSolver` (this PR being just the very beginning of that process). While a complete rewrite would be cleaner/faster (and what I always hoped to do) the issues of integration with the existing examples and Tools would be more difficult. This way all that use the Manager would benefit.

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -59,7 +59,14 @@ class OSIMSIMULATION_API Manager
...
     /** Integrator. */
-    SimTK::Integrator* _integ;
+    // This is the actual integrator that is used when integrate() is called.
+    // Its memory is managed elsewehere; either by the user or by the
+    // _defaultInteg smart pointer.
+    SimTK::ReferencePtr<SimTK::Integrator> _integ;
+
+    // The integrator that is used when using the model-only constructor.
+    // This is allocated only if necessary.
+    SimTK::ResetOnCopy<std::unique_ptr<SimTK::Integrator>> _defaultInteg;
```
aseth1: The copy of the `Manager` will still point to the original's Model but the default integrator will be reset (nullptr). Is that what you wanted? If you wanted both the and the used default to be retained in the copy you could use SimTK::ClonePtr for both the `_model` and `_defaultInteg`.

chrisdembia: There are many things that make this class behave poorly if copied. For example, the `_controllerSet` member variable would also need to be updated based on the cloned model. We would need to write a custom copy constructor. My preference would be to just disallow copying the Manager--it almost certainly does not work anyway--delete the copy constructor, etc. 

I cannot use a ClonePtr for the `_defaultInteg` because it requires the new `System` to function properly.

chrisdembia: I went ahead and deleted the copy constructor and copy assignment operator. Let me know what you think.

jenhicks:
@aseth1 Is this OK to merge?

@chrisdembia Does this address #156 or #335 ? They seem related, but I did not get a chance to dig in to answer my question (arriving at my 🚌 stop).

chrisdembia:
Thanks for looking that up; this PR addresses both of those issues.

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -122,13 +130,26 @@ class OSIMSIMULATION_API Manager
...
+    /** <b>(Deprecated)</b> A Constructor that does not take a model or
+     * controllerSet. This constructor also does not set an integrator; you
+     * must call setIntegrator() on your own. You should use one of the other
+     * two constructors. */
+    DEPRECATED_14("There will be no replacement for this constructor.")
     Manager();  
 
+    // This class would not behave properly if copied (we would need to write a
+    // complex custom copy constructor, etc.), so don't allow copies.
+    Manager(const Manager&) = delete;
```
aseth1: :+1:

chrisdembia:
Thank you @aseth1

chrisdembia:
@wagglefoot, you should now be able to use the Manager in python.

aymanhab:
@chrisdembia Thanks for the fix. Are there SimTK classes that we can expose to make Forward integration in python and Java parallels C++ (e.g. Expose Integrator, TimeStepper, or Study?)

wagglefoot:
@chrisdembia - excellent, thank you very much for highlighting this.

chrisdembia:
> Are there SimTK classes that we can expose to make Forward integration in python and Java parallel C++ (e.g. Expose Integrator, TimeStepper, or Study?)

Exposing the concrete integrators and the time steppers would be great.

PR #1164, opened 2016-08-10
Add docopt to superbuild and increase minimum cmake version to 3.2.

klshrinidhi:
As per discussion in #1137.


.travis.yml
```diff
@@ -64,6 +66,11 @@ before_install:
   # Stop build if comment contains [skip travis].
   - if $(git log -n1 --format="%B" | grep --quiet '\[skip travis\]'); then exit; fi 
   
+  - if [[ "$TRAVIS_OS_NAME" = "osx" ]]; then brew update; fi
+  - if [[ "$TRAVIS_OS_NAME" = "osx" ]]; then brew unlink cmake; fi
+  - if [[ "$TRAVIS_OS_NAME" = "osx" ]]; then brew install cmake; fi
+  - cmake --version
```
chrisdembia: Could you add a comment describing why this is necessary? I am surprised that travis's osx machines don't come with a recent cmake version.

chrisdembia:
Thanks for submitting this. I'll want to test the superbuild locally before finishing my review.

chrisdembia:
The README needs to be updated to mention the new CMake version we depend on. Also, I would like the Ubuntu description to be helpful. Right now, the README says:

```
operating system: Ubuntu 13.10 or later.
cross-platform build system: CMake >= 2.8.8; cmake-gui. Ubuntu 12.04 only has 2.8.6 available; download from the website or from this third party PPA.
```

I think it's reasonable for it to now say something like:

```
operating system: Ubuntu 15.10 is the oldest version that allows you to apt-get most of the dependencies.
cross-platform build system: CMake >= 3.1.3; cmake-gui. For earlier versions of Ubuntu, download from the website or from this third party PPA.
```

Also, ideally, we would use exactly cmake 3.1.3 on travis (at least on Ubuntu).

chrisdembia:
The README.md also needs to be updated to mention docopt as a dependency.

chrisdembia:
@klshrinidhi I have been testing this on my computer with the `cli-docopt` branch and it works great when using UNIX Makefiles on OSX. However, if I use Xcode for the superbuild, docopt is unable to install b/c the `libdocopt.dylib` does not get generated.

chrisdembia:
This may be the issue with Xcode:

> Some native build systems may not like targets that have only object files, so consider adding at least one real source file to any target that references `$<TARGET_OBJECTS:objlib>`.

- https://cmake.org/cmake/help/v3.0/command/add_library.html?highlight=add_library
- http://stackoverflow.com/questions/24078873/cmake-generated-xcode-project-wont-compile

chrisdembia:
I created a PR on docopt.cpp to fix this issue https://github.com/docopt/docopt.cpp/pull/56.

.travis.yml
klshrinidhi: I think the CMake that comes installed is 2.8.x. 
I now realized this is not what I intended. My intention was to install CMake 3.1.3 in particular. I will fix this.

chrisdembia: That would be ideal but if it's hard to use brew to get that specific version, then I think it's fine if we only test with 3.1.3 on at least one build (Ubuntu).

klshrinidhi: Okay. With this PR's changes, CMake version on Ubuntu is bumped to 3.2.x. That needs to get fixed too. I will get back.

klshrinidhi:
* I could not find a way to install CMake 3.1.3 in particular on Ubuntu Precise.
* I did find a way to do so on OS X. But this revealed possible bug in CMake 3.1.3 (https://cmake.org/Bug/view.php?id=15355) that got fixed in CMake 3.2. With this info, I now think that minimum CMake requirement for docopt is probably 3.2 and not 3.1.3.

chrisdembia:
Okay 3.2 sounds fine.

klshrinidhi:
Regarding editing the Readme, it is not true that 'minimum requirement' is Ubuntu 15.10. There is in fact no notion of "minimum operating system version" here as we distribute sources and any one with required build tools can build the sources on any operating system. CMake 3.2 is available through PPAs (which you mention) for Ubuntu 12.04 and higher. 

I think it would be better to not state a minimum OS requirement for Ubuntu and just state the build dependencies with versions. And then list the instructions for various Ubuntu releases we have tested on. 
I will be making this change.

klshrinidhi:
I updated the readme. I also almost wrote build instructions for Ubuntu 12.04 but decided to leave it for another PR.

klshrinidhi:
I have created an [issue](https://github.com/docopt/docopt.cpp/issues/57) on *docopt.cpp* regarding the CMake error on Windows.

README.md
```diff
@@ -584,13 +584,11 @@ On Ubuntu using Unix Makefiles
...
+* **operating system**: Ubuntu 12.04 or later. *(See OS specific instructions under 'For the impatient' below)*
 * **cross-platform build system**:
-  [CMake](http://www.cmake.org/cmake/resources/software.html) >= 2.8.8;
-  `cmake-gui`. Ubuntu 12.04 only has 2.8.6 available; download from the website
-  or from this [third party
-  PPA](https://launchpad.net/~robotology/+archive/ubuntu/ppa).
-* **compiler**: [gcc](http://gcc.gnu.org) >= 4.8; `g++-4.8`, or
+  [CMake](http://www.cmake.org/cmake/resources/software.html) >= 3.2;
+  `cmake-gui`. 
+* **compiler**: [gcc](http://gcc.gnu.org) >= 4.9; `g++-4.9`, or
```
chrisdembia: +1

README.md
```diff
@@ -584,13 +584,11 @@ On Ubuntu using Unix Makefiles
 Most dependencies can be obtained via the Ubuntu software repositories. On each
 line below, we show the corresponding package.
 
-* **operating system**: Ubuntu 13.10 or later.
+* **operating system**: Ubuntu 12.04 or later. *(See OS specific instructions under 'For the impatient' below)*
```
chrisdembia: What do you think about doing the following:

1. Removing the `**operating system**` bullet from the itemized list of dependencies.
2. Modifying the sentences that come before the list of the dependencies:
  Before:

  > Most dependencies can be obtained via the Ubuntu software repositories. On each line below, we show the corresponding package.

  After:

  > Most dependencies can be obtained via the Ubuntu software repositories; especially if you are using Ubuntu 15.10 or later. On each line below, we show the corresponding package name. You can find instructions for specific Ubuntu versions under 'For the impatient' below.

chrisdembia:
Thanks for updating the README and creating the docopt issue. I have made one comment on the README.

I have also submitted a PR  https://github.com/docopt/docopt.cpp/pull/60 to fix the docopt issue you reported.

Aside from agreeing on a resolution to the comment I made, I think the following items remain for this PR:

1. Updating the `docopt` commit hash once https://github.com/docopt/docopt.cpp/pull/60 is merged.
2. In the README, mentioning docopt as a dependency, and giving non-impatient instructions for it. I will add this.

chrisdembia:
@klshrinidhi for now, I have decided to use a fork of docopt.

README.md
chrisdembia: This is the last remaining issue for this PR. What do you think about the suggestion, @klshrinidhi? I think it aligns with the idea that there is no minimum operating system version. I don't think we should state "12.04" if we do not yet give detailed instructions for it.

chrisdembia: If you like the proposed wording, I can make the change.

klshrinidhi:
:+1: 
Once this is merged to master, #1137 can pull master into its branch so we can see it working.

README.md
chrisdembia: @klshrinidhi what do you think?

klshrinidhi: Sounds good.

chrisdembia: Okay thanks.

chrisdembia:
@klshrinidhi I have updated the README's description of Ubuntu versions, and I have reverted back to using the main `docopt.cpp` repo instead of our fork (the PR was merged).

Ready for review.

klshrinidhi:
:+1:

chrisdembia:
Thanks @klshrinidhi . Any other reviewers interested? This PR affects the README, travis/appveyor scripts, and CMakeLists.

jenhicks:
@aymanhab Can you review?

chrisdembia:
@aymanhab I'd like to respond to your concern about requiring users to obtain an additional dependency. In general, I agree that requiring users to get additional dependencies should be avoided. However, the superbuild process works very well. We could make the command-line tool optional, but I don't think that is a good idea: otherwise, people will only get the libraries by default and won't be to do much unless they write their own executable (or build an example). I agreed with Shrinidhi to try having docopt as an external dependency. We can try that for now and if we run into issues, we can reconsider. Of course, there are pros and cons on both sides.

aymanhab:
Thanks for the feedback @chrisdembia 
One comment: 
george-edison55-precise-backports is this going to be around for good?

No issues with the code/make file changes, though I would have preferred to absorb the files in our repository (sorry @klshrinidhi I know your preference) but will give this a try and hope for the best. How this fits into our build/install (there're no targets that depend on it unless this's a separate PR).
If not optional I'd like to have an option to build core-libraries only that excludes this and BTK and other extras for a minimal functional opensim-core footprint that runs embedded or on a phone for example. If we have an agreement let me know and I'll open an issue.

chrisdembia:
> george-edison55-precise-backports is this going to be around for good?

I don't know about the permanence of third-party ppa's; @klshrinidhi may know more.



> How this fits into our build/install (there're no targets that depend on it unless this's a separate PR).

The cmake target for building docopt is not part of the OpenSim project (so there could not be a target dependency) it is part of the separate "OpenSimDependencies" project. If a user tries to build OpenSim and they have not installed docopt, they will get a CMake error saying "couldn't find docopt." If you want @aymanhab, we could add an additional message saying "Follow the superbuild instructions in README.md to obtain docopt."

>  I'd like to have an option to build core-libraries only that excludes this and BTK and other extras 

We still have a `BUILD_API_ONLY` CMake flag that still works. Would that be fine?

Right now, even if `BUILD_API_ONLY` is ON, this PR would still require docopt. I can change this so that docopt is only required if BUILD_API_ONLY is OFF.

aymanhab:
Thanks @chrisdembia  Agreed on the the additional message, ideally the instructions indicate briefly what is the "functionality" that the dependency provides.
Making BUILD_API_ONLY avoid all the extras is indeed the right thing to do.

chrisdembia:
> Agreed on the the additional message, ideally the instructions indicate briefly what is the "functionality" that the dependency provides.

The README mentions what docopt is used for, but I will also add a note to the CMake error message.

klshrinidhi:
> george-edison55-precise-backports is this going to be around for good? 

This PPA is used only to install CMake 3.2 in Ubuntu 12.04. For newer versions of Ubuntu, it is not required. The PPA has been around for years now. If this PPA goes down/away, we can replace it with some other PPA. We could potentially have CMake package on our PPA (https://launchpad.net/~opensim/+archive/ubuntu/opensim). That way we can guarantee its existence.  

--------------------------------------

Regarding making docopt optional, super-build will allow docopt to be checked ON/OFF. So this PR is not mandating docopt. The PR #1137 has to make sure to keep docopt optional too if it is not doing so already.

Though it is probably late now, I think using Python to be the front end of the command line tools would have been a good idea. That would have avoided docopt altogether. Of-course that would make Python a *requirement*. Since Python comes with Linux & OS X, this applies only to Windows, which I don't think is much of a downside.

chrisdembia:
> Though it is probably late now, I think using Python to be the front end of the command line tools would have been a good idea. That would have avoided docopt altogether. Of-course that would make Python a requirement. Since Python comes with Linux & OS X, this applies only to Windows, which I don't think is much of a downside.

I investigated this fairly thoroughly. Python would only be a requirement for developers, not for end-users, so it wouldn't be that bad. However, developers would also *need* to build the python bindings (so they'd need swig, etc.). Docopt made it easy to make the command line tool in C++. Overall, I think we reduce complexity by using docopt instead of python.

We can always reimplement the command line tools in the future as our needs change. The more important thing is the new interface and functionality of the command line interface.

chrisdembia:
@aymanhab I have improved the CMake error message that is produced when docopt is not found. 

Ready for review.

aymanhab:
LGTM, thanks @chrisdembia and @klshrinidhi merged.

chrisdembia:
Thank you! One step closer to the future OpenSim command-line interface!

PR #1170, opened 2016-08-10
[bindings] Test SimbodyMatterSubsystem::calcResidualForce().

chrisdembia:
This PR just makes sure that the method `SimbodyMatterSubsystem::calcResidualForce()` and the class `OpenSim::InverseDynamicsSolver` are accessible from MATLAB/Python.


chrisdembia:
Oops; created the wrong PR. See #1171 instead.

Issue #991, opened 2016-06-01
Target/label C3D testing data files according to the intended condition 


aseth1:
and/or add comments in the test to indicate the different conditions presented by the files: e.g. different forceplate configurations, different units, w/out EMG data, other analog data, etc.... 

This issue was raised from review of #987 where the test was sped up by eliminating the reading of one particularly large data file. If the point of the test was to verify we can handle large files, then perhaps we it could become its own case that is tested only in release. 

The following files are tested
-    filenames.push_back("jogging.c3d"); **_--> multiple force-plates with different orientations_**
-    filenames.push_back("singleLeglanding_2.c3d");
-    filenames.push_back("aStaticTrial_2.c3d");
-    filenames.push_back("aStaticTrial.c3d");
-    filenames.push_back("doubleLegLanding.c3d");
-    filenames.push_back("sideCutting.c3d");
-    filenames.push_back("singleLegLanding.c3d");
-    filenames.push_back("walking2.c3d");
-    filenames.push_back("walking5.c3d"); **_--> huge data file of minutes of treadmill walking_**

The **_bold_** text are a sample annotation of the condition that helpful to document what coverage exists.

chrisdembia:
> If the point of the test was to verify we can handle large files, then perhaps we it could become its own case that is tested only in release.

It also seems that this test may be a good candidate for parallelizing by splitting into multiple tests. A header file could contain the actual testing function, and each file could have its own executable test.

jenhicks:
For 4.0, we should add comments about the purpose of each file in the test, per Ajay's comments.

jimmyDunne:
I passed these c3d files on for internal testing, they were not meant to be pushed to the main OpenSim repository. We should take them off immediately and I can gather some c3d files that I have had permission to distribute.

jenhicks:
@klshrinidhi Can you remove these files from the repo per @jimmyDunne comment?

klshrinidhi:
These files have been in the repository for several months now. I committed them as part of testing C3DFileAdapter. This comment could have been made in that PR.

I can remove the files when I have enough other files to test the adapter.

jenhicks:
@klshrinidhi Is starting a PR to remove extra files. walking2 and walking5 are ok to distribute so can remain in the test.

PR #1173, opened 2016-08-10
Remove undistributable c3d files.

klshrinidhi:
As per #991.

Removing c3d files from repo that we don't have permission to distribute.


tkuchida:
LGTM. We could also try [purging these files from the repo history](https://help.github.com/articles/remove-sensitive-data/)—though we can't do anything about existing clones and forks.

tkuchida:
The tests are failing because [test_DataAdapter.py](https://github.com/opensim-org/opensim-core/blob/remove_undistributable_c3d_files/Bindings/Python/tests/test_DataAdapter.py) is trying to read singleLeglanding_2.c3d and jogging.c3d.

chrisdembia:
> LGTM. We could also try purging these files from the repo history—though we can't do anything about existing clones and forks.

I used methods like this when I created the original git repo. But from what I remember, it's not really a good idea to delete files from a repo once it's already in use. I think doing something like that changes all the commit hashes (since a commit hash is generated by the contents of the files in the commit), etc. and basically creates a new repo that doesn't work with the previous one.

jenhicks:
What is the source of the CI failures? Besides that, let's go ahead and merge.

tkuchida:
> What is the source of the CI failures?

As mentioned above, test_DataAdapter.py is trying to read singleLeglanding_2.c3d and jogging.c3d, which are among the files that were removed. test_DataAdapter.py should be updated to read walking2.c3d and/or walking5.c3d instead.

tkuchida:
:+1: Looks good to merge if tests pass.

chrisdembia:
LGTM too.

jenhicks:
@klshrinidhi can you id the source of the ci failure?

klshrinidhi:
CI builds are passing except for segfault in clang. This is merge-able now.

PR #1174, opened 2016-08-10
[doxygen] Added missing escapes to `<` and `>`

tkuchida:
Doxygen was attempting to interpret `<attached_geometry>` as an html tag. [ci skip]


chrisdembia:
Thanks.

PR #1177, opened 2016-08-11
Fix to GeometryPath for MovingPathPoints

### LINKED ISSUE #1104: [visualizer] Muscles attached to MovingPathPoints are not drawn
Can be reproduced by loading gait10dof18musc.osim (the vasti and rect_fem use MovingPathPoints).

tkuchida:
This change ends up calling `MovingPathPoint::getLocation(state)` rather than `get_location()` (the latter returns the MovingPathPoint's unused "location" property). Full explanation at the bottom of #1104. Results from visual test using gait10dof18musc.osim:

| Before | After
| --- | ---
| before | after

Fixes #1104

### DIFF
--- a/OpenSim/Simulation/Model/GeometryPath.cpp
+++ b/OpenSim/Simulation/Model/GeometryPath.cpp
@@ -153,7 +153,7 @@ generateDecorations(bool fixed, const ModelDisplayHints& hints,
 
         // the body (PhysicalFrame) IS part of the actual Model and its system
         // so we can ask it for its transform w.r.t. Ground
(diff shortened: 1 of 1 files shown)


chrisdembia:
LGTM.

It is pretty important that MovingPathPoint not be a Station; I think this is a high priority bug fix for 4.0.

jenhicks:
@chrisdembia Can you create an issue with more details?

chrisdembia:
> @chrisdembia Can you create an issue with more details?

I created issue #1179 .

Issue #1165, opened 2016-08-10
Remove redundant Joint::getCoordinateSet() method


tkuchida:
Replace with either `Joint::get_CoordinateSet()` or new Coordinate accessors once #1116 has been completed and merged.

tkuchida:
`getCoordinateSet()` is also called in Bindings/Python/tests/test_swig_additional_interface.py and potentially other python tests as well (tagged from #1178).

tkuchida:
Fixed by #1210.

PR #1178, opened 2016-08-11
Fix python memory management for prescribeControlForActuator.

chrisdembia:
This PR addresses memory management for `PrescribedController::prescribeControlForActuator`, which adopts the passed-in Function.

This bug came up in PR #1162.

I also changed the error-handling within `prescribeControlForActuator`, since the pre-existing asserts would not cause any error message to bubble up to python.

### DIFF
--- a/Bindings/Python/swig/python_simulation.i
+++ b/Bindings/Python/swig/python_simulation.i
@@ -77,6 +77,14 @@ MODEL_ADOPT_HELPER(Controller);
     geom._markAdopted()
 %}
 
+// PrescribedController::prescribeControlForActuator takes ownership of
+// the passed-in function.
+// There are two overloads of this function; we append to both of them.
+// arg[1] is `Function* prescribedFunction`.
+%pythonappend OpenSim::PrescribedController::prescribeControlForActuator %{
+    args[1]._markAdopted()
+%}
+
 
 // Compensate for insufficient C++11 support in SWIG
 // =================================================
--- a/Bindings/Python/swig/python_simulation.i
+++ b/Bindings/Python/swig/python_simulation.i
@@ -80,7 +80,7 @@ MODEL_ADOPT_HELPER(Controller);
 // PrescribedController::prescribeControlForActuator takes ownership of
 // the passed-in function.
 // There are two overloads of this function; we append to both of them.
-// arg[1] is `Function* prescribedFunction`.
+// args[1] is `Function* prescribedFunction`.
 %pythonappend OpenSim::PrescribedController::prescribeControlForActuator %{
     args[1]._markAdopted()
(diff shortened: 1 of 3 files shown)


aymanhab:
@chrisdembia I had a PR #981 a while ago to replace PrescribedController with FunctionBasedController since current PRescribedControl does double-duty to allow specifying a file for input or a set of Functions. I mention this here so that when we get back to that we make sure this bandaid in swig is maintained or that the interface is made better so that it's not needed anymore.

chrisdembia:
Ah thanks for the reminder; hopefully we can find a way to avoid the messy adoption issues. Do you think it's fine if we still merge this PR?

jenhicks:
I think it is ok to merge this PR. Would be good to get PR #981 merged soon so that PR #1162 can use the FunctionBasedController instead, right?

aymanhab:
Yes, definitely, I wouldn't hold this one, it fixes an existing bug/crash on current master :+1:

Bindings/Python/swig/python_simulation.i
```diff
@@ -77,6 +77,14 @@ MODEL_ADOPT_HELPER(Controller);
     geom._markAdopted()
 %}
 
+// PrescribedController::prescribeControlForActuator takes ownership of
+// the passed-in function.
+// There are two overloads of this function; we append to both of them.
+// arg[1] is `Function* prescribedFunction`.
```
tkuchida: `arg` -> `args`?

Bindings/Python/tests/test_swig_additional_interface.py
```diff
@@ -272,6 +272,30 @@ def test_markAdoptedSets(self):
...
+        # Body.
+        body = osim.Body('b1', 1.0, osim.Vec3(0, 0, 0), osim.Inertia(0, 0, 0))
+        model.addBody(body)
+        # Joint.
+        joint = osim.PinJoint('j1', model.getGround(), body)
+        model.addJoint(joint)
+        # Actuator.
+        actu = osim.CoordinateActuator()
+        actu.setName('actu')
+        actu.setCoordinate(joint.getCoordinateSet().get(0))
```
tkuchida: [Tagging] Will need to update to `get_CoordinateSet()` in #1165.

tkuchida:
`</review>` One minor comment. @chrisdembia, I think you're clear to merge with or without the change.

tkuchida:
[Spurious AppVeyor failure](https://ci.appveyor.com/project/opensim-org/opensim-core/build/4284/job/dxj6y2fny4316uyo#L15) was not caused by this PR; merging.

chrisdembia:
Thank you.

Issue #1128, opened 2016-07-18
Move hopper example to Examples folder and add as test


tkuchida:
The hopper example currently resides in the Sandbox; it should be moved into the Examples folder at this point. We should at least be ensuring that the "student" and "teacher" versions continue to build, but perhaps they should also be tests. In that case, we will need to disable the visualizer (could simply add a condition [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Sandbox/ExampleHopperDevice/helperMethods.h#L168)), but ideally a "student" should be able to run the example _without making any changes_ and the visualizer should appear—is this possible? Telling the "student" to toggle a flag at the top of some file would be highly undesirable for this didactic example.

chrisdembia:
It should be easy to use an environment variable or a compile definition to cause the visualizer not to show in testing. CMake lets you set environment variables to use when running tests.

klshrinidhi:
I think the `main()` could take arguments (which is standard) that say whether or not visualizers should show. In the absence of an argument, visualizer is shown. This will allow users to run the example separately and not provide any arguments.

chrisdembia:
Ah +1. That's a much better idea.

klshrinidhi:
I can do this after I am done with my current task.

tkuchida:
Possibly related to organization of proposed Example Gallery; see #1144.

PR #1180, opened 2016-08-11
Added missing #include to hopper example

tkuchida:
Fallout from #1145; good motivation to tackle #1128. [ci skip]


carmichaelong:
Just ran into the same issue. LGTM.

PR #1183, opened 2016-08-15
Simple Python API example (take two)

wagglefoot:
This is the proper PR, and essentially the same as #1162 (which was deleted because of reasons).

This example is provided to demonstrate the utility of the python OpenSim API. It mimics the simple-arm example found on the core landing page. Though it does not include forward simulation, but instead saves the model to an .osim file which can be used in the GUI for fort further analysis.

### DIFF
--- a/Bindings/Python/examples/simple-arm.py
+++ b/Bindings/Python/examples/simple-arm.py
@@ -1,0 +1,135 @@
+# ----------------------------------------------------------------------- #
+# The OpenSim API is a toolkit for musculoskeletal modeling and           #
+# simulation. See http://opensim.stanford.edu and the NOTICE file         #
+# for more information. OpenSim is developed at Stanford University       #
+# and supported by the US National Institutes of Health (U54 GM072970,    #
+# R24 HD065690) and by DARPA through the Warrior Web program.             #
+#                                                                         #
+# Copyright (c) 2005-2012 Stanford University and the Authors             #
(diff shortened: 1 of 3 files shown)


Bindings/Python/examples/simple-arm.py
```diff
@@ -0,0 +1,135 @@
...
+# ----------------------------------------------------------------------- #
+
+# simple-arm.py
+# Author: Neil Dhir
+# ------------------------------------------------------------------------#
+# ABSTRACT: This short piece of OpenSim python API example mimics         #
+# the simple-arm example found on the core landing page. Though it does   #
+# not include forward simulation, but instead saves the model to an .osim #
+# file which can be used in the GUI for further analysis.		  #
+# ------------------------------------------------------------------------#
```
chrisdembia: I expect most people will find this file in a distribution of OpenSim--not in the opensim-core source code repository; "core landing page" will not have much meaning in that context. I think this description should explain that the model has two bodies, two joints, a muscle, and a controller.

You could also consider giving the file a name that indicates it is a model-building example (`build-simple-arm-model.py`?).

chrisdembia:
I am going to merge this now. @wagglefoot, you accidentally put two copies of the file in the repo (rather than just renaming the one file), but that's okay--I'll resolve that after merging. Thanks for contributing this!

Issue #1140, opened 2016-07-25
Speedup CI Testing by reducing simulation period of tests


aseth1:
I believe a more effective strategy than splitting tests into smaller bits (which achieves more granularity so more cores are used effectively) is to reduce the total test times by reducing the simulation period for many test. (Not that splitting tests is bad, but it has less impact than reducing test duration and currently it is leading to code duplication).  

Here is an ongoing list:
- [ ]  `testCMCArm26`: simulates for 1s. This could be reduced to 0.25s duration, (e.g. 0.40 to 0.65) and the comparison tolerances tightened.
- [ ]  `testCMCEMGDrivenArm`: simulates for 1s. This can be reduced to 0.25s duration, (0.65 to 0.90) to target time with high "EMG".
- [ ]  `testCMCGait10dof18musc`: simulates from 0.6 to 1.1s and can be reduced to 0.25s duration, (0.75 to 1.0).

@klshrinidhi you could start working through this list.

klshrinidhi:
:+1:

PR #1184, opened 2016-08-15
Reduce simulation time from 1s to 0.25s for testCMCArm26

klshrinidhi:
Addressing issue #1140.


Applications/CMC/test/testCMCArm26_Millard.cpp
```diff
@@ -45,15 +45,26 @@ void testCMCArm26() {
     Storage *standard = new Storage();
     cmc.getModel().formStateStorage(temp, *standard);
 
-    Array<double> rms_tols(0.02, 2*2+2*6); // activations within 2%, angles within .6 degrees
+    Array<double> rms_tols(0, 2*2+2*6); // activations within 2%, angles within .6 degrees
```
tkuchida: Please remove obsolete, potentially confusing trailing comment.

Applications/CMC/test/testCMCArm26_Thelen.cpp
```diff
@@ -41,10 +41,27 @@ void testCMCArm26() {
     Storage *standard = new Storage();
     cmc.getModel().formStateStorage(temp, *standard);
 
-    Array<double> rms_tols(0.02, 2*2+2*6); // activations within 2%, angles within .6 degrees
+    Array<double> rms_tols(0, 2*2+2*6); // activations within 2%, angles within .6 degrees
```
tkuchida: As above.

Applications/CMC/test/testCMCArm26_Thelen.cpp
```diff
@@ -41,10 +41,27 @@ void testCMCArm26() {
...
     const string& muscleType = cmc.getModel().getMuscles()[0].getConcreteClassName();
     string base = "testCMCArm26 "+ muscleType;
 
+    rms_tols[0]  = 0.0002;    // r_shoulder/r_shoulder_elev/value
+    rms_tols[1]  = 0.002;     // r_shoulder/r_shoulder_elev/speed
+    rms_tols[2]  = 0.0006;    // r_elbow/r_elbow_flex/value
+    rms_tols[3]  = 0.006;     // r_elbow/r_elbow_flex/speed
+    rms_tols[4]  = 0.0000001; // TRIlong/activation
```
tkuchida: How were these tolerances determined? Are we sure an RMS error of, say, 1.1e-7 is unacceptably large?

klshrinidhi: No solid rationale. I just looked at the rms errors manually after I changed the simulation times and set these values to be quite tight.

tkuchida: > I just looked at the rms errors manually after I changed the simulation times and set these values to be quite tight.

Ok, seems fine provided all platforms arrive at the same answer—which may not be the case (IIRC, different LAPACK versions on different platforms can cause slightly different optimizer behavior). Another option could be to update the standards based on this new time range and set all tolerances to the same value.

tkuchida:
`</review>` I think the ideal case would be to take the CMC results and perform tests using *knowledge* rather than particular *data* (e.g., certain muscles should have higher activations than certain other muscles; running a forward simulation using the CMC results should produce kinematics that are similar to the kinematics that were tracked, etc.), but perhaps that belongs in a separate (potentially post-4.0) issue.

tkuchida:
I opened #1188 to track the issue I raised above.

Applications/CMC/test/testCMCArm26_Thelen.cpp
```diff
@@ -41,10 +41,27 @@ void testCMCArm26() {
     Storage *standard = new Storage();
     cmc.getModel().formStateStorage(temp, *standard);
 
-    Array<double> rms_tols(0.02, 2*2+2*6); // activations within 2%, angles within .6 degrees
+    Array<double> rms_tols(0, 2*2+2*6); 
     const string& muscleType = cmc.getModel().getMuscles()[0].getConcreteClassName();
```
aseth1: The original tolerances are justifiable. Id they could be tightened to 1% or 0.5 degree that would be more than sufficient. It is not necessary to ratchet down to detect minor changes likely do to different math libraries in different environments.

Applications/CMC/test/testCMCArm26_Millard.cpp
```diff
@@ -45,15 +45,26 @@ void testCMCArm26() {
...
+    rms_tols[8]  = 0.03;   // TRImed/activation
+    rms_tols[9]  = 0.0003; // TRImed/fiber_length
+    rms_tols[10] = 0.05;   // BIClong/activation
+    rms_tols[11] = 0.0008; // BIClong/fiber_length
+    rms_tols[12] = 0.0071; // BICshort/activation
+    rms_tols[13] = 0.0005; // BICshort/fiber_length
+    rms_tols[14] = 0.003;  // BRA/activation
+    rms_tols[15] = 0.0001; // BRA/fiber_length
 
     CHECK_STORAGE_AGAINST_STANDARD(results, *standard, rms_tols, __FILE__, __LINE__,
```
aseth1: It is not necessary to individualize the tolerances.  The original test gave the numerical rationale for the values chosen. Being within 1 degree of the standard angles and with 5% of the Millard muscles seems fine to me. The standard was created using the Thelen muscle so we would not expect the answers to be identical.

Applications/CMC/test/testCMCArm26_Millard.cpp
```diff
@@ -45,15 +45,15 @@ void testCMCArm26() {
     Storage *standard = new Storage();
     cmc.getModel().formStateStorage(temp, *standard);
 
-    Array<double> rms_tols(0.02, 2*2+2*6); // activations within 2%, angles within .6 degrees
+    Array<double> rms_tols(0.015, 2*2+2*6); // activations within 1%, angles within .6 degrees
```
aseth1: It would be ideal if one tolerance could be set here (say 2%) and the individual tolerances not being adjusted below (as in the Thelen case). If that is not possible then this fine. Should say 1.5% in the comment though.

aseth1:
Please don't forget to fix the merge conflicts. Thanks.

jenhicks:
@klshrinidhi Please resolve conflicts and merge.

Issue #1185, opened 2016-08-16
Add tests for Millard2012AccelerationMuscle or label "experimental"


tkuchida:
Millard2012AccelerationMuscle isn't being tested thoroughly (e.g., with CMC) and many changes have been made to EquilibriumMuscle that have not been propagated to AccelerationMuscle. We should either add tests or indicate in the doxygen that AccelerationMuscle is still experimental.

jenhicks:
I think we should indicate in doxygen that AccelerationMuscle is experimental.

PR #1186, opened 2016-08-16
Throw exception if equilibrium cannot be calculated for Millard and Thelen muscles

### LINKED ISSUE #391: Throw an exception if equilibrate method fails in Millard2012*Muscle
Currently, the `computeInitialFiberEquilibrium()` method in `Millard2012AccelerationMuscle` and `Millard2012EquilibriumMuscle` sets the fiber length to `optimal_fiber_length` if the initialization routine fails. Throwing an exception may be a safer alternative.

tkuchida:
Millard2012EquilibriumMuscle, Millard2012AccelerationMuscle, and Thelen2003Muscle currently display a warning and quietly set activation to zero and fiber length to optimal_fiber_length if equilibrium cannot be computed. This warning could go unnoticed and lead to unexpected results. This PR replaces the warning message with an exception.

Commented out Millard2012AccelerationMuscle test in testStaticOptimization. This is the only test that ended up failing. The Millard2012AccelerationMuscle class needs to be updated and tested more thoroughly (see #1185).

This PR fixes #391 and #1195

[Diff ignoring whitespace changes](https://github.com/opensim-org/opensim-core/pull/1186/files?w=1)


tkuchida:
### Note to reviewers
[Click here](https://github.com/opensim-org/opensim-core/pull/1186/files?w=1) to see the diff with whitespace changes ignored.

tkuchida:
The [test failures](https://travis-ci.org/opensim-org/opensim-core/jobs/152651424#L8073) are unrelated to this PR (timeouts from testAssemblySolver and testContactGeometry on clang in Debug mode).

OpenSim/Actuators/Millard2012AccelerationMuscle.h
```diff
@@ -51,7 +51,39 @@ The parent class, Muscle.h, provides
 
 namespace OpenSim {
 
+//==============================================================================
+//                   MILLARD2012ACCELERATIONMUSCLE EXCEPTIONS
+//==============================================================================
+class Millard2012AccelerationMuscleCannotEquilibrate : public Exception {
```
aseth1: Why is this Exception specific to Millard2012AccelerationMuscle? There could be a single `MuscleCannotEquilibrate` or `MuscleCannotInitialize` Exception defined by Muscle that all muscles could use.

tkuchida:
Moved `MuscleCannotEquilibrate` exception to Muscle class and propagated changes to Thelen2003Muscle. Ready for re-review.

OpenSim/Actuators/Millard2012AccelerationMuscle.h
```diff
@@ -51,7 +51,24 @@ The parent class, Muscle.h, provides
 
 namespace OpenSim {
 
+//==============================================================================
+//                   Millard2012AccelerationMuscle Exceptions
+//==============================================================================
+class Millard2012AccelerationMuscleInvalidFlag : public Exception {
```
aseth1: this and the `Thelen2003MuscleInvalidFlag`  are identical (and subtlly different from `Millard2012EquilibriumMuscleInvalidFlag`), which indicates that they should have a common definition.  I am also not sure of the difference between these exceptions and the general `MuscleCannotEquilibrate` since they will be thrown in `computeInitialFiberEquilibrium` in which case the *invalid flag* context seems more mystifying than informative. There is no added information for what causes the invalid flag and the user is being told about an internal private method they have no control over.

My suggestion is to enable the general `MuscleCannotEquilibrate` to take additional details (string) about why the failure occurred. If more information can be added by the failure of internal helper methods then that could be tacked on as details. I would be satisfied if the specific contents of the "details" are a TODO (simply stating the internal method that failed) for when the initialization/equilibrium methods are revisited. At least then you get the correct top level error/message that the MuscleCannotEquilibrate and you get the internal method as a detail rather than a mysterious `###MuscleInvalidFlag` exception.

Please chat with me if this is unclear or seems unnecessary.

tkuchida: > Please chat with me if this is unclear

Thank you for the explanation. I will do as you say.

tkuchida: A more logical design is for the internal method to return an enum so it is impossible for the caller to receive an unexpected flag. From your response, it seems this is considered to be an important piece of code, so I will implement the redesign and remove the MuscleInvalidFlag exceptions.

aseth1: For this PR the internal method and its implementation/code are not important.  What is important is that we not have the `###MuscleInvalidFlag` exceptions, which are less informative than the `MuscleCannotEquilibrate`, which should be thrown. If using enums prevents the invalid flag case from ever being reached (which I agree was a bad implementation to begin with) then that seems like a reasonable solution so that there is no need for that type or exception.  You could also fix/address the internal methods at a later data. Which ever you prefer is fine with me.

tkuchida: I think it would be incorrect to replace `[baseClassName]InvalidFlag` with `MuscleCannotEquilibrate`. The `[baseClassName]InvalidFlag` exception is thrown in a condition that can be reached only if the concrete class developer made a programming error such as

  1. Assigning a value other than 0, 1, or 2 to the first element of the SimTK::Vector that is returned by `initMuscleState()` (Millard2012AccelerationMuscle and Thelen2003Muscle) or `estimateMuscleFiberState()` (Millard2012EquilibriumMuscle); or
  2. Neglecting to handle one of the valid return values in the `switch` statement in `computeInitialFiberEquilibrium()`.

Neither of these errors relate to whether an equilibrium can be found, which I believe is what the `MuscleCannotEquilibrate` exception implies. In this situation, I believe a "fatal internal error encountered"-type of exception should be thrown (as defined on lines 57–67).

I am implementing a change that will remove the condition altogether to avoid further discussion about which exception should be thrown and what it should say.

Issue #1113, opened 2016-07-01
Support both python2 and python3


chrisdembia:
I thought we already had an issue for this but I guess not.

@clnsmith @juzhang

wagglefoot:
Hey Chris!

Speaking of this very issue; when will examples appear? 

From the [documentation](http://simtk-confluence.stanford.edu:8080/display/OpenSim/Scripting+in+Python) we have this:

> We are currently in the process of developing example Python scripts for OpenSim and will share these with the community soon.

If I go into Sherlock mode, it would appear that that page was lasted edited in August last year.

Any updates?

chrisdembia:
There are plenty of examples in the tests folder: https://github.com/opensim-org/opensim-core/tree/master/Bindings/Python/tests

One roadblock, though silly, has been figuring where examples should go in our distribution. In the repo, they should be arranged in a way that makes it easy to test them; perhaps they should be in the `tests` folder. In the installation/distribution, I suppose they could they go in `sdk/python/opensim/examples`, as is typical for python packages. We can always rearrange them later.

Would you be interested in contributing any examples? The example in the README would be a great place to start.

@moorepants any suggestions for how to make sure examples run successfully (just making sure no exceptions are thrown) as part of running the tests?

wagglefoot:
I would be quite keen to contribute actually, as I am having to learn how to do it anyway, and in that process I might as well write something up. 

Naturally, someone more in the know, should double-check them whence they are done.

chrisdembia:
> I would be quite keen to contribute actually

Excellent! Please take a quick look at our [contributing guidelines](https://github.com/opensim-org/opensim-core/blob/master/CONTRIBUTING.md); most of it is not relevant if you are writing python examples, so just look at the beginning.

If you are planning a larger contribution, it may be good to open an Issue in which you propose what you want to contribute. Then we can discuss and agree on it so that you don't waste your time creating something that doesn't match what we want. We should not have the discussion on this Issue, since it is specifically about python3 support.

> Naturally, someone more in the know, should double-check them whence they are done.

Of course. We review everything that users want to contribute.

moorepants:
I typically run examples in the build process for releases, since they often take longer than unit tests. You could have the examples run through a release/build script, for example.

wagglefoot:
[...]

Issue #979, opened 2016-05-16
Document the layout of the OpenSim installation on all platforms.


chrisdembia:
We have done something similar for Simbody in its [README](https://github.com/simbody/simbody#layout-of-installation):

![image](https://cloud.githubusercontent.com/assets/846001/15303918/38997eba-1b6f-11e6-92bd-98fac13a75ae.png)

Places to be updated:
- changelog
- Confluence
- README

Related: #975 #978

aymanhab:
@chrisdembia While this works for Simbody which is a toolkit with no specific entry point and we're free to place folders as we like as long as we can find them, it is quite different for the application framework we use (Netbeans Platform) where there's a assumed directory structure (e.g. the relative placement and names of the folders bin, etc, ide, opensim, platform  in OpenSim 3.3- is not modifiable). My current game plan is to make our application layout mirror Netbeans on all platforms and then backfill the missing pieces (includes, libs, python, Geometry etc.) by placing them in sensible locations. Other option is to have a separate application installer and an API installer which is just a zip of opensim-core installation. What do you think?

chrisdembia:
Gah this stuff is so complex; it's hard for me to get it all straight in my head. I'm fine with using the same directory structure for the GUI on Windows/OSX. For the copy of opensim-core that we use with the GUI, we can set `OPENSIM_INSTALL_UNIX_FHS` to OFF, so that opensim-core also has the same layout on Windows/OSX. So far, I am undecided as to whether or not that's a good idea; it may be fine.

I don't think we should shuffle around the opensim-core layout when it gets distributed with the GUI. Because, then the OpenSimConfig.cmake file will be incorrect. A better option is to give opensim-core a CMake option like `OPENSIM_INSTALL_LAYOUT_FOR_GUI`; then we can choose whatever layout we like, and the OpenSimConfig.cmake will be correct. The only exception I think is okay is copying OpenSim dlls. But ideally, opensim-core and the GUI would just share the same bin dir.

I have no opinion (yet) about Geometry, or the `opensim-models` stuff.

If/when we support Ubuntu, it will be Ubuntu that decides the installation layout, not Netbeans, etc. But for OSX and Windows, we get to choose (or let Netbeans choose). So our policy for Windows/OSX may be different than for Ubuntu. But we can deal with that when we get there.

As for an "API installer," I think we can just distribute API binaries through package managers; homebrew (OSX) and Chocolatey (windows). What do you think?

[...]

PR #1187, opened 2016-08-16
Copy python example scripts into installation

chrisdembia:
This PR edits CMake files so that the python example files are copied into the installation. Example scripts are placed into the `share/doc/OpenSim/Scripts/Python` directory on UNIX, and into the `sdk/Scripts/Python` directory on Windows. Thus, this PR partially addresses a concern raised in #1113.

This PR also deletes a duplicate file that entered the repository through #1183.

I was going to also include running the python examples as part of our tests, but that ended up being more complicated than I thought and so I would like to punt on that for now.


tkuchida:
[AppVeyor failure](https://ci.appveyor.com/project/opensim-org/opensim-core/build/4304/job/vr7ac6gbwtj80gii#L1380) is due to unsigned int wrap through zero; reported in #1196.

tkuchida:
:+1: LGTM

chrisdembia:
Would anyone else like to review this?

jimmyDunne:
I will

aymanhab:
@chrisdembia not sure if there's an issue dedicated to installation layout but if we're going to make an Alpha build for API only then we should have a planned layout for all platforms, including models, scripts, sdk (headers and libraries) etc.

chrisdembia:
Yeah; I agree that we need to decide on an installation layout for all platforms. I'm happy to change where the python examples go in future PRs.

aymanhab:
Related issue #979

chrisdembia:
Ah thanks for the link.

chrisdembia:
I have fixed merge conflicts. This PR is ready for review.

chrisdembia:
Thank you!

PR #1189, opened 2016-08-16
Reduce simulation time from 1s to 0.25s for testCMCEMGDrivenArm

klshrinidhi:
Addressing #1140.

Similar to PR #1184.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/1189)

PR #1191, opened 2016-08-16
Reduce simulation time from 1s to 0.25s for testCMCGait10dof18musc

klshrinidhi:
Addressing issue #1140.

Similar to #1184 & #1189.

### DIFF
--- a/Applications/CMC/test/testCMCGait10dof18musc.cpp
+++ b/Applications/CMC/test/testCMCGait10dof18musc.cpp
@@ -66,10 +66,63 @@ void testGait10dof18musc() {
     Storage *standard = new Storage();
     cmc.getModel().formStateStorage(temp, *standard);
 
-    Array<double> rms_tols(0.02, 2*10+2*18); // activations within 2%
-    for(int i=0; i<20; ++i){
-        rms_tols[i] = 0.01;  // angles and speeds within .6 degrees .6 degs/s 
-    }
+    Array<double> rms_tols(0, 2*10+2*18);
+    rms_tols[0]  = 0.00004;  // ground_pelvis/pelvis_tilt/value
+    rms_tols[1]  = 0.001;    // ground_pelvis/pelvis_tilt/speed
+    rms_tols[2]  = 0.000002; // ground_pelvis/pelvis_tx/value
+    rms_tols[3]  = 0.00002;  // ground_pelvis/pelvis_tx/speed
+    rms_tols[4]  = 0.000004; // ground_pelvis/pelvis_ty/value
+    rms_tols[5]  = 0.0001;   // ground_pelvis/pelvis_ty/speed
+    rms_tols[6]  = 0.00006;  // hip_r/hip_flexion_r/value
+    rms_tols[7]  = 0.002;    // hip_r/hip_flexion_r/speed
+    rms_tols[8]  = 0.00006;  // hip_l/hip_flexion_l/value
+    rms_tols[9]  = 0.002;    // hip_l/hip_flexion_l/speed
+    rms_tols[10] = 0.00004;  // back/lumbar_extension/value
+    rms_tols[11] = 0.001;    // back/lumbar_extension/speed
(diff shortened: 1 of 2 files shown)


Applications/CMC/test/testCMCGait10dof18musc.cpp
```diff
@@ -66,10 +66,63 @@ void testGait10dof18musc() {
     Storage *standard = new Storage();
     cmc.getModel().formStateStorage(temp, *standard);
 
-    Array<double> rms_tols(0.02, 2*10+2*18); // activations within 2%
-    for(int i=0; i<20; ++i){
-        rms_tols[i] = 0.01;  // angles and speeds within .6 degrees .6 degs/s 
-    }
```
aseth1: You can restore the previous tolerances and then halving them based on type (angle, speed, activation, ...) . There is no rationale for changing each one individually.

Applications/CMC/test/testCMCGait10dof18musc.cpp
```diff
@@ -66,63 +66,9 @@ void testGait10dof18musc() {
...
-    rms_tols[47] = 0.00001;  // rect_fem_l/fiber_length
-    rms_tols[48] = 0.0004;   // vasti_l/activation
-    rms_tols[49] = 0.00001;  // vasti_l/fiber_length
-    rms_tols[50] = 0.002;    // gastroc_l/activation
-    rms_tols[51] = 0.00001;  // gastroc_l/fiber_length
-    rms_tols[52] = 0.002;    // soleus_l/activation
-    rms_tols[53] = 0.00003;  // soleus_l/fiber_length
-    rms_tols[54] = 0.004;    // tib_ant_l/activation
-    rms_tols[55] = 0.00004;  // tib_ant_l/fiber_length
+    Array<double> rms_tols(0.035, 2*10+2*18); // Activations within 2%.
```
aseth1: According to your previous tolerances it was not necessary to loosen the tolerances to 0.035 from 0.02. I am in favor of tightening the tolerances just not tightening them individually (as before). A tolerance of 1% on activation and 0.01 rad and 0.01 rad/s  for speeds is doable here (all set to 0.01). Then the following lines are not necessary either.

klshrinidhi: The tests were not passing with previous tolerance (0.02). You can see it in CI builds. I relaxed the tolerance accordingly and ended up this number after several attempts.

aseth1: Thanks for letting me know and sorry that you ran into this problem. I was expecting it to be a quick fix. I am trying to figure out why 0.02 was successful for the longer duration (previous long test) and shortening the duration requires larger tolerances? I am quite certain it is not due to these changes. I also made some other changes to the test in #1205 so let me investigate and hold off on merging this PR for the moment.

chrisdembia: It may be the case that `Storage::compareColumnRMS()` should be numerically integrating the RMS. That is, something like`trapz(time, diff * diff) / (endTime - startTime)` I see that it is *somewhat* doing this "integration" by dividing by the `endIndex - startIndex`, but this does not account for non-uniform time steps.

tkuchida: The RMS error will increase if there happened to be low error in the parts of the simulation that were removed. Also, IIRC, the feedback loops in CMC are such that the reference signal can drift; changing the start time may have lasting downstream effects. :ocean:

jenhicks:
@aseth1 Check on why tolerances had to be relaxed.

aseth1:
@klshrinidhi there have been changes made to the Tools and to make this test more complete from other PRs. Can you update this branch and we can look at the tolerances again. Thanks!

jenhicks:
If you update and the tolerances still aren't sufficient, we will table this PR for now.

aseth1:
This PR was outdated by #1529 and #1627 and is being closed.

Issue #1175, opened 2016-08-10
Error reading v3.3 model into v4.0 API due to <VisibleObject /> 	<WrapObjectSet />


clnsmith:
auto model = Model(model_file);

Im using the line above to read a model into a C++ main using the 4.0 API. The model reads into the 3.3 GUI with no problem, but throws this error in the 4.0 API:

SimTK Exception thrown at Xml.cpp:820:
Error detected by Simbody method Xml::Element::getAs(): The given Node was not an Element; it is a NoNode. Use Element::isA()  to check before calling Element::getAs().
(Required condition 'isA(node)' was not met.)

I have tracked it down to "VisibleObject /" and "WrapObjectSet /". If these lines are not included under a body, the error is thrown in 4.0. If they are included, no errors. Sample model attached. [gait2392_simbody_test.zip](https://github.com/opensim-org/opensim-core/files/412337/gait2392_simbody_test.zip)

				<Body name="ground">
					<mass>0</mass>
					<mass_center> 0 0 0</mass_center>
					<inertia_xx>0</inertia_xx>
					<inertia_yy>0</inertia_yy>
					<inertia_zz>0</inertia_zz>
					<inertia_xy>0</inertia_xy>
					<inertia_xz>0</inertia_xz>
					<inertia_yz>0</inertia_yz>
					<VisibleObject />
					<WrapObjectSet />
				</Body>

chrisdembia:
Thank you for reporting.

carmichaelong:
Errors come from [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/Model.cpp#L169) and [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/Model.cpp#L173) for VisibleObject and WrapObjectSet, respectively. If the XML element does not exist in the file, the method attempts to perform a `clone` on `element_end()`, which throws the Exception.

I imagine a line to check for this like [the one above for geometry](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/Model.cpp#L163) could help, but I don't think I know enough about the downstream effects in loading models to know what could break.

aymanhab:
Definitely, this code should all be defensive since you don't know if these
xml elements existed in old format files. Please submit a PR with this
guard/change and I'll review it, or take over if it turns ourt to be more
involved. Thanks.

carmichaelong:
[...]

PR #1193, opened 2016-08-17
fix model load from file with missing <VisibleObject> or <WrapObjectSet> tags

carmichaelong:
Addresses #1175. I tested this locally by using testModelCopy. I removed the <VisibleObject> and <WrapObjectSet> tags from arm26.osim, then ran the test in Debug mode, which confirmed an Exception being thrown at the lines mentioned in the issue. I ran the test after the committed changes with the tags missing, and the test passed.

@aymanhab


aymanhab:
Looks good and safe, thanks @carmichaelong Will merge after tests pass. :+1:

PR #1198, opened 2016-08-18
Prevent wrap through zero when subtracting unsigned ints

### LINKED ISSUE #1196: testInitState incorrectly reported memory leak due to unsigned int wrap through zero
[This recent AppVeyor failure](https://ci.appveyor.com/project/opensim-org/opensim-core/build/4304/job/vr7ac6gbwtj80gii#L1380) is due to [this test in testInitState.cpp](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Test/testInitState.cpp#L161), which does not consider that memory can sometimes be freed "behind the scenes". The subtraction `getCurrentRSS() - mem0` can, therefore, wrap the unsigned int through zero and throw a spurious Exception.

Same problem as reported in issue #1152 and fixed in PR #1155. Also search for any other tests that use `getCurrentRSS()`.

tkuchida:
This PR adds "wrap through zero" protection everywhere `getCurrentRSS()` is used (#1155 already fixed testComponents). Fixes #1196.

### DIFF
--- a/OpenSim/Simulation/Test/testInitState.cpp
+++ b/OpenSim/Simulation/Test/testInitState.cpp
@@ -153,12 +153,12 @@ void testMemoryUsage(const string& modelFile)
     //=========================================================================
     // Setup OpenSim model
     // base footprint
-    size_t mem0 = getCurrentRSS( );
-    size_t model_size{ 0 };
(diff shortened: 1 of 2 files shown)


aseth1:
LGTM

Issue #1195, opened 2016-08-18
Throw an exception if equilibrate method fails in Thelen2003Muscle


tkuchida:
Analogous to issue #391 (PR #1186) for Millard2012*Muscle.

PR #1201, opened 2016-08-20
Throw exceptions if muscles cannot compute equilibrium

tkuchida:
## Summary
This PR fixes issues #391 and #1195, and addresses the concerns raised in the previous proposal (PR #1186). [[diff ignoring whitespace](https://github.com/opensim-org/opensim-core/pull/1201/files?w=1)]

## Primary Changes
Three muscles currently display a warning and quietly modify their state if the initialization method fails:

  1. Thelen2003Muscle
    - `computeInitialFiberEquilibrium()` calls `initMuscleState()`
    - actuation=0 and fiberLength=optimal_fiber_length on failure
  2. Millard2012EquilibriumMuscle
    - `computeInitialFiberEquilibrium()` and `computeFiberEquilibriumAtZeroVelocity()` call `estimateMuscleFiberState()`
    - actuation=0 and fiberLength=optimal_fiber_length on failure
  3. Millard2012AccelerationMuscle
    - `computeInitialFiberEquilibrium()` calls `initMuscleState()`
    - actuation=0, fiberLength=optimal_fiber_length, and fiberVelocity=0 on failure

A warning message could go unnoticed and lead to unexpected results. This PR replaces the warning messages with exceptions (`MuscleCannotEquilibrate`) that provide additional details of the failure. The internal methods were also modified so they return their status flags as enums, thereby eliminating the need to catch an "internal error" exception if the calling method encounters an unexpected status flag.

## Secondary Changes
- Commented out the Millard2012AccelerationMuscle test in testStaticOptimization. This is the only test that ended up failing. The Millard2012AccelerationMuscle class needs to be updated and tested more thoroughly (see #1185).
- Added tests to testMuscles that exercise the new `MuscleCannotEquilibrate` exception for Thelen2003Muscle and Millard2012EquilibriumMuscle. An analogous test causes the Millard2012AccelerationMuscle to throw a *different* exception that would need further investigation; covered by #1185.
- SWIG 3.0.6 (up to 3.0.8) doesn't allow an enum class that is private to a class. The issue was raised [here](https://github.com/swig/swig/issues/594) and fixed [here](https://github.com/swig/swig/commit/1d6279092816cb3e0a47254d9db545ce3e447f95) in 3.0.9, but the most recent package on chocolatey is 3.0.6. In my investigations I noticed instances of 3.0.5 in README.md, which are corrected to 3.0.6 here.


OpenSim/Actuators/Millard2012AccelerationMuscle.cpp
```diff
@@ -702,131 +702,70 @@ double  Millard2012AccelerationMuscle::
...
+            "  Solution error %e exceeds tolerance of %e\n"
+            "  Newton iterations reached limit of %d\n"
+            "  Activation is %f\n"
+            "  Fiber length is %f\n";
+        const int messageLength = std::snprintf(nullptr, 0, fmt,
+            abs(solnErr), tol, maxIter, activation, fiberLength);
+        std::vector<char> buf(messageLength+1); //+1 for null terminator
+        std::snprintf(&buf[0], buf.size(), fmt,
+            abs(solnErr), tol, maxIter, activation, fiberLength);
+        OPENSIM_THROW_FRMOBJ( MuscleCannotEquilibrate,
```
aseth1: :+1:

aseth1:
LGreatTM

aseth1:
@chrisdembia @klshrinidhi care to review and merge if satisfied?

tkuchida:
Add test case to bindings/Java/Test

jenhicks:
Add a java or matlab test to see if try/catch works to catch exception. @klshrinidhi can help if needed.

aseth1:
@tkuchida my understanding was that you were relieved of having to test the handling of these Exceptions in Java. There is a separate issue (#1212) to address the broader problem (potentially dozens and dozens of untested Exceptions) which is beyond the scope of this PR.  @aymanhab @klshrinidhi was that your understanding as well, to not hold up the PR on this issue.

aymanhab:
Indeed, no need to hold, this PR doesn't materially change the potential issue of unhandled exceptions.

tkuchida:
> Indeed, no need to hold

Glad to see this decision, as I'm 2 hours in with no end in sight (I haven't built wrapping before).

jenhicks:
That was my impression as well; sorry for the confusion and 2 hours. 😞

chrisdembia:
No need to get *wrapped* up in all of that.

OpenSim/Actuators/Millard2012AccelerationMuscle.cpp
```diff
@@ -1256,31 +1195,20 @@ void Millard2012AccelerationMuscle::
...
-                    int aMaxIterations,
-                    double aNewtonStepFraction) const
+Millard2012AccelerationMuscle::ResultOfInitMuscleState
+Millard2012AccelerationMuscle::initMuscleState(SimTK::Vector& result,
+                                               const SimTK::State& s,
+                                               double aActivation,
+                                               double aSolTolerance,
+                                               int aMaxIterations,
+                                               double aNewtonStepFraction) const
 {
```
klshrinidhi: I think it will be good to *return the result* rather than *take the result by reference and edit it*. It improves readability by separating inputs and outputs of a function. The return type could be `std::pair<ResultOfInitMuscleState,SimTK::Vector>`.

OpenSim/Actuators/Millard2012AccelerationMuscle.cpp
```diff
@@ -1536,93 +1464,83 @@ SimTK::Vector Millard2012AccelerationMuscle::
...
+    if (abs(ddlce_dtt) < aSolTolerance) {  // The solution converged.
+
+        result[0] = ddlce_dtt;
+        result[1] = (double)iter;
+        result[2] = lce;
+        result[3] = dlce_dt;
+        result[4] = ami.fpeVEM * fiso;
+        result[5] = ami.fseVEM * fiso;
+        return ResultOfInitMuscleState::Success_Converged;
+    }
```
klshrinidhi: Maybe consider returning an associative array instead of a sequence if values represent different quantities. The associative array will improve readability.
```
results["solution_error"]
results["fiber_velocity"]
.....so on
```
If you have another reason to return a sequence, consider `std::vector` instead of `SimTK::Vector`.

OpenSim/Actuators/Millard2012AccelerationMuscle.h
```diff
@@ -897,27 +900,34 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Millard2012AccelerationMuscle, Muscle);
...
-    // Private Utility Class Members
-    //      -Computes activation dynamics and fiber kinematics
-    //=====================================================================
+    // Status flag returned by initMuscleState().
+    enum ResultOfInitMuscleState {
+        Success_Converged,
+        Warning_FiberAtLowerBound,
+        Failure_MaxIterationsReached
+    };
+
```
klshrinidhi: If this enum can be generalized to be applicable to other all muscles, consider moving the enum out of the class and making it an `enum class` (It offers lot more type safety and regular enum too).
Moving it out of the class may have implications toward bindings.

OpenSim/Actuators/Millard2012AccelerationMuscle.h
```diff
@@ -897,27 +900,34 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Millard2012AccelerationMuscle, Muscle);
...
-                             double aSolTolerance, int aMaxIterations,
-                             double aNewtonStepFraction) const;
-    
-
+    ResultOfInitMuscleState initMuscleState(SimTK::Vector& result,
+                                            const SimTK::State& s,
+                                            double aActivation,
+                                            double aSolTolerance,
+                                            int aMaxIterations,
+                                            double aNewtonStepFraction) const;
```
klshrinidhi: Could make the parameters `const` if they represent constant values inside the function. Compiler will help if function violates const-ness.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -425,111 +425,74 @@ computeInitialFiberEquilibrium(SimTK::State& s) const
...
+            "  Fiber length is %f\n";
+        const int messageLength = std::snprintf(nullptr, 0, fmt,
+            abs(solnErr), tol, maxIter, clampedActivation, fiberLength);
+        std::vector<char> buf(messageLength+1); //+1 for null terminator
+        std::snprintf(&buf[0], buf.size(), fmt,
+            abs(solnErr), tol, maxIter, clampedActivation, fiberLength);
+        OPENSIM_THROW_FRMOBJ( MuscleCannotEquilibrate,
+                              string(buf.begin(), buf.end()) );
+        break;
     }
```
klshrinidhi: Would be better to replace C code with C++ code using `std::string`, `std::stringstream` and so on. It will improve readability as well as correctness.

OpenSim/Simulation/Model/Muscle.h
```diff
@@ -36,8 +36,28 @@
...
+                            const std::string& detail) :
+        Exception(file, line, func, obj, detail) {
+        std::string msg = "Unable to compute equilibrium for this muscle.\n";
+        msg += "Please verify that the initial activation is valid and that ";
+        msg += "the length of the musculotendon actuator doesn't produce a ";
+        msg += "pennation angle of 90 degrees or a negative fiber length.";
+        addMessage(msg);
+    }
+};
+
```
klshrinidhi: Consider deriving from this exception for each muscle type. The advantage is that when multiple muscles are involved and one of them throws (failure to reach equilibrium), the enclosing try-catch block(s) can catch/do different things (or not) depending on muscle type.

klshrinidhi:
:+1:

OpenSim/Actuators/Millard2012AccelerationMuscle.cpp
tkuchida: > I think it will be good to *return the result* rather than *take the result by reference and edit it*. It improves readability by separating inputs and outputs of a function. The return type could be `std::pair<ResultOfInitMuscleState,SimTK::Vector>`.

Done. The return type is now `std::pair<StatusFromInitMuscleState, ValuesFromInitMuscleState>`, where
- `StatusFromInitMuscleState` is the status flag (enum)
- `ValuesFromInitMuscleState` is a `std::map<std::string, double>` containing the calculated values

OpenSim/Actuators/Millard2012AccelerationMuscle.h
tkuchida: > If this enum can be generalized to be applicable to other all muscles, consider moving the enum out of the class and making it an `enum class` (It offers lot more type safety and regular enum too).
Moving it out of the class may have implications toward bindings.

The enum is not a common feature of all muscles since the internal `initMuscleState()` and `estimateMuscleFiberState()` methods will not necessarily appear in every muscle model. In fact, `Millard2012AccelerationMuscle::initMuscleState()` and `Thelen2003Muscle::initMuscleState()` are both private methods that are called in exactly one place, so there does not seem to be any reason not to embed them in their calling methods (`computeInitialFiberEquilibrium()` in both cases). However, I think it is beyond the scope of this PR to do further refactoring.

OpenSim/Actuators/Millard2012AccelerationMuscle.h
tkuchida: > Could make the parameters `const` if they represent constant values inside the function. Compiler will help if function violates const-ness.

I generally do this in my own code but the author of these models did not follow this convention. I have made these parameters `const` but did not apply to the other methods in these classes; if more time is invested in this code, I think algorithmic improvements would provide greater impact.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
tkuchida: > Would be better to replace C code with C++ code using `std::string`, `std::stringstream` and so on. It will improve readability as well as correctness.

Done. My initial change was to remove the fixed 999-character buffer, but yes ostringstream looks better.

OpenSim/Simulation/Model/Muscle.h
tkuchida: > Consider deriving from this exception for each muscle type. The advantage is that when multiple muscles are involved and one of them throws (failure to reach equilibrium), the enclosing try-catch block(s) can catch/do different things (or not) depending on muscle type.

At present, there does not appear to be a need to handle the exception differently depending on muscle type.

tkuchida:
@klshrinidhi Thank you for reviewing. Addressed and replied to comments; ready for re-re-review.

klshrinidhi:
:+1:

tkuchida:
Thanks, @klshrinidhi.
@aseth1: okay to merge with latest changes?

aseth1:
The additional changes LGTM!

PR #1205, opened 2016-08-23
Fix path failure when WrapObject is attached to a PhysicalOffsetFrame

### LINKED ISSUE #1105: "Array index out of bounds" error when reordering Joints in the multibody tree
Attaching the device from the hopper example to the leg6dof9musc model crashes at initSystem(), specifically at [line 689 of Model.cpp](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/Model.cpp#L689). Presumably, this "swap" code is attempting to swap one of the model's Joints with one of the Joints in the device (which is not in the model's JointSet).

### LINKED ISSUE #1129: Unable to wrap around a WrapCylinder attached to a PhysicalOffsetFrame
Wrapping around a `WrapCylinder` attached to a `PhysicalOffsetFrame` fails in general (it works in the hopper example only when a `ConsoleReporter` has been added; could find no other workaround). According to @chrisdembia: "The issue is that `addToSystem()` is invoked on `patellaFrame` before it is invoked on the Joints. The fix for now is to make sure `addToSystem()` is invoked on Joints before it is invoked on anything else." (see discussion thread in #1073).

In addition to a formal test case, we should modify the hopper example (see [this block](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Sandbox/ExampleHopperDevice/buildHopperModel.cpp#L150) in buildHopperModel.cpp) by creating a `PhysicalOffsetFrame` at the distal end of the thigh and adding the patella (a `WrapCylinder`) to this `PhysicalOffsetFrame`.

aseth1:
This PR fixes #1129. 

The main changes are:
 1. attached a patella `WrapCylinder` to a `PhyscialOffsetFrame` on the thigh `Body` instead of using the `WrapObject`'s internal transform to the `exampleHopperDevice` and trapping Exceptions thrown during its execution.
 2. added a private list of ordered subcomponents that specifies the the order in which subcomponents are added to the System. Bodies and Joints anywhere in the Model, for example, must be added to the System in the order of a valid MultibodyTree, followed by Frames, Constraints and Forces. All Components can have an opinion about their order, but the top-level (Model) has the final say. Corresponding methods to reset the list and build it were added.
 3.  The major fallout of the change was that CMC was assuming coordinate data in the order of the CoordinateSet corresponds directly with the order of q's in the State. The Coordinates in CoordinateSet are in order of the traversal order of `getComponentList<Joint>` which has nothing to do with the State order, so this assumption cannot be made and was removed from CMC.
 4. Some general cleanup, correction and tightening of tolerances in related test cases.
 5. The order of the `DecorativeGeometry` to be drawn was changed due changes in Component order, so the standard for `testVisualization` had to be updated.

This PR likely fixes #1105 as well, but it is not explicitly tested. A subsequent PR will address that and will attempt to remove the now unnecessary reordering of the Joints in the model's JointSet in Model.cpp [lines 591-702](https://github.com/opensim-org/opensim-core/commit/09197e1969601320d738125582cf3c7ba89e8d04#diff-447a828d94f7bb607768e454b9d331cbR691)

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/1205)


OpenSim/Sandbox/ExampleHopperDevice/buildHopperModel.cpp
```diff
@@ -151,13 +151,17 @@ Model buildHopper() {
     // vastus muscle can wrap. In the future, we will be able to attach wrap
     // objects to PhysicalOffsetFrames; for now, we will use the existing
     // "translation" property.
+    auto patellaFrame = new PhysicalOffsetFrame("patellaFrame",
```
tkuchida: The "In the future…" sentence can now be removed from the above comment.

OpenSim/Sandbox/ExampleHopperDevice/buildHopperModel.cpp
```diff
@@ -151,13 +151,17 @@ Model buildHopper() {
...
+        *thigh, SimTK::Transform(linkDistalPoint));
     auto patella = new WrapCylinder();
     patella->setName("patella");
     patella->set_radius(0.08);
     patella->set_length(linkRadius*2.);
-    patella->set_translation(linkDistalPoint);
+    //patella->set_translation(linkDistalPoint);
```
tkuchida: I think these commented lines (160 and 162) can be deleted. (We wouldn't want a workshop attendee to uncomment them.)

OpenSim/Sandbox/ExampleHopperDevice/exampleHopperDevice.cpp
```diff
@@ -168,12 +168,28 @@ void addDeviceConsoleReporterToModel(Model& model, Device& device,
 } // namespace OpenSim
 
 
+void executeExample();
```
tkuchida: Renaming `main()` to `executeExample()` requires corresponding modifications to [README.md](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Sandbox/ExampleHopperDevice/README.md) (bottom), exampleHopperDevice.cpp ([line 35](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Sandbox/ExampleHopperDevice/exampleHopperDevice.cpp#L35)), and exampleHopperDevice_answers.cpp ([line 35](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Sandbox/ExampleHopperDevice/exampleHopperDevice_answers.cpp#L35)).

Applications/CMC/test/testCMCGait10dof18musc.cpp
```diff
@@ -58,21 +66,25 @@ void testGait10dof18musc() {
...
+    int nstates = standard->getColumnLabels().size() - 1;
 
-    CHECK_STORAGE_AGAINST_STANDARD(results, *standard, rms_tols, __FILE__, __LINE__, "testArm26 failed");
+    // angles and speeds within .6 degrees .6 degs/s; activations within 1%
+    Array<double> rms_tols(0.01, nstates);
 
-    const string& muscleType = cmc.getModel().getMuscles()[0].getConcreteClassName();
-    cout << "\ntestGait10dof18musc "+muscleType+" passed\n" << endl;
+    CHECK_STORAGE_AGAINST_STANDARD(results, *standard, rms_tols,
+        __FILE__, __LINE__, "testGait10dof18musc failed");
```
tkuchida: `"testGait10dof18musc failed"`  :arrow_right:  `"testGait10dof18musc " + muscleType + " failed"` ?

OpenSim/Common/Component.cpp
```diff
@@ -1053,6 +1067,17 @@ bool Component::constructOutputForStateVariable(const std::string& name)
     return constructOutput<double>(name, func, SimTK::Stage::Model);
 }
 
+// helper method to specify the order of subcomponents.
```
tkuchida: `(append subcomponent to the end of Component's _orderedSubcomponents list if it isn't already in the list)`?

tkuchida: Nevermind - the comment for this method in Component.h looks good.

OpenSim/Common/Component.h
```diff
@@ -2487,11 +2501,18 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
     // subsystem.
     SimTK::ResetOnCopy<SimTK::MeasureIndex> _simTKcomponentIndex;
 
+    // list of subomponents that are contained in this Component's properties
```
tkuchida: subomponents -> subcomponents

OpenSim/Simulation/Control/TrackingController.cpp
```diff
@@ -76,7 +77,9 @@ void TrackingController::setNull()
 //=============================================================================
 void TrackingController::setDesiredStatesStorage(const Storage* aYDesStore)
 {
-    _desiredStatesStorage = aYDesStore;
+    auto* old = _desiredStatesStorage.release();
+    delete old;
```
tkuchida: :potable_water: :arrow_right: :non-potable_water: :+1:

OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp
```diff
@@ -146,14 +142,21 @@ void CMCActuatorSubsystemRep::setSpeedTrajectories(FunctionSet *aSet) {
         _qSet->evaluate(_uWork,1,t);
     }
 
-    for(i=0;i<nq;i++) q[i] = _qWork[i] + _qCorrections[i];
-    for(i=0;i<nu;i++) u[i] = _uWork[i] + _uCorrections[i];
+    /* Hack to obtain a modifiable state in a const method */
```
tkuchida: Ideally, would `realizeSubsystemDynamicsImpl()` just take a non-const state? The comment seems sufficient if we don't want to start fiddling with CMC's architecture.

OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp
```diff
@@ -146,14 +142,21 @@ void CMCActuatorSubsystemRep::setSpeedTrajectories(FunctionSet *aSet) {
...
+    State& modCompState = const_cast<SimTK::State&>(_completeState);
 
+    // Update the coordinate values to pose the model while computing muscle
+    // controls
+    const CoordinateSet& coords = _model->getCoordinateSet();
+    for (int i = 0; i < nq; ++i) {
+        coords[i].setValue(modCompState, _qWork[i] + _qCorrections[i], i==(nq-1));
```
tkuchida: Why is the "enforce constraints" flag set for the last coordinate (as opposed to, say, `i==0`)? Perhaps a comment would be helpful here, since this seems to be a minor change to the algorithm.

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -656,38 +666,31 @@ void Model::extendConnectToModel(Model &model)
             Joint* free = new FreeJoint(jname, *ground, *child);
             free->upd_reverse() = mob.isReversedFromJoint();
             addJoint(free);
+            setNextSubcomponentInSystem(*free);
```
tkuchida: I was expecting this code to be more similar to the code above (lines 645–650) when a WeldConstraint is added. The comment above about JointSet (lines 611–612) suggests that JointSet will eventually disappear, so should `addJoint(free)` be `adoptSubcomponent(free)` instead? (In `addJoint()`, it looks like `free` will be adopted by the JointSet, which will eventually disappear.)

OpenSim/Sandbox/ExampleHopperDevice/exampleHopperDevice.cpp
aseth1: I also arbitrarily chose `executeExample()` and I am happy to change it while I am revising. Any preferences: `runHopperExample` or `hopToIt()` ?

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -744,7 +764,7 @@ void Model::extendAddToSystem(SimTK::MultibodySystem& system) const
     Model *mutableThis = const_cast<Model *>(this);
 
     // Ensure Ground is added before all other Components
-    getGround().addToSystem(system);
+    //getGround().addToSystem(system);
```
tkuchida: Was this line commented out to allow nested Models (i.e., Ground isn't necessarily at the top of the Component list)? Perhaps delete altogether or add a "TODO" for when nested Models are fully implemented.

OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp
aseth1: Since we are setting all the coordinate values, I wait to the last coordinate before enforcing the constraints. The default is to enforce constraints on every `setValue()`.  I will add a clarifying comment.

OpenSim/Simulation/Model/Model.cpp
aseth1: That's a good point. It currently would serialize the FreeJoint (having been added to the JointSet), so I agree adopting directly would not have that affect (like the hidden Welds that are added to break loops). I will update.

OpenSim/Sandbox/ExampleHopperDevice/exampleHopperDevice.cpp
tkuchida: I like `hopToIt()` :rabbit2:, but it isn't descriptive so we probably shouldn't encourage in a workshop example. `runHopperExample()` sounds good to me.

OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp
tkuchida: > Since we are setting all the coordinate values, I wait to the last coordinate before enforcing the constraints.

Ah, ok. I think the algorithm is slightly different now because, previously, all the coordinate values could have changed slightly for every `0 <= i < nq-1`—but the tolerance is probably so small that the results don't change substantially (and the new way seems correct, since the projection should be independent of the order in which the coordinates are set).

tkuchida:
`</review>` A few comments above; LGTM otherwise.

OpenSim/Simulation/Model/Model.cpp
aseth1: It was a previous hack to force Ground to be added first. I will delete!

OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp
aseth1: The issue is that the CMCActuatorSubsystemRep has it own secret internal State (_completeState) which it updates. Computation is not supposed change the System (or subsystem). It is a bad implementation and I would rather not change it. All I did was gather it to one `const_cast` instead of having three.

tkuchida: > All I did was gather it to one `const_cast` instead of having three.

Ok, sounds good. Thanks for explaining.

aseth1:
Thank-you @tkuchida for the fast review. I have attempted to address each of your comments. 
@chrisdembia can I entice you to review for a pastry/coffee from Peet's?

tkuchida:
@aseth1 Does this PR repair the hopper example? The "Examples/Hopper Device" project should spawn a visualizer and simulate when run. Current behavior on master is for `Component::getSystem()` to throw a `!hasSystem()` exception.

aseth1:
@tkuchida thanks for raising the the `PathWrapPoint::getSystem()` issue. I was not aware of the problem on master either so there was not attempt to address the issue here. Would you like that issue addressed in this PR?

tkuchida:
> Would you like that issue addressed in this PR?

Let's create a new issue for that and move this PR through. You have my :+1:

Applications/CMC/test/testCMCWithControlConstraintsGait2354.cpp
```diff
@@ -65,28 +65,27 @@ void testGait2354() {
...
-    for (int i = 23; i < 46; ++i){
-        rms_tols2[i] = 0.75; // velocities
-    }
-    for (size_t i = 46; i < rms_tols2.size(); ++i){
-        rms_tols2[i] = 0.15; // muscle activations and fiber-lengths
+
+    // tolerance for joint angles 0.02rad ~= 1.15 degs and activation within 2%
+    std::vector<double> rms_tols2(col_labels.getSize()-1, 0.02);
+    for (int i = 0; i < nq; ++i){
+        rms_tols2[2*i+1] = 0.25; // velocities rad/s
```
chrisdembia: I'm afraid the order of the states could change again in the future.

OpenSim/Common/Component.cpp
```diff
@@ -319,6 +331,11 @@ void Component::disconnect()
 
 void Component::addToSystem(SimTK::MultibodySystem& system) const
 {
+    // If being asked to be added to the same System that it is already
+    // a part, there is nothing to be done.
+    if (hasSystem() && (&getSystem() == &system)) {
+        return;
+    }
```
chrisdembia: Seems okay to me but I'm curious. What caused this to be necessary? Are you sure that silently passing is the correct thing to do, instead of throwing an exception?

OpenSim/Common/Component.h
```diff
@@ -2487,11 +2501,18 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
+    SimTK::Array_<SimTK::ReferencePtr<Component> >  _propertySubcomponents;
     // Keep fixed list of data member Components upon construction
     SimTK::Array_<SimTK::ClonePtr<Component> > _memberSubcomponents;
     // Hold onto adopted components
     SimTK::Array_<SimTK::ClonePtr<Component> > _adoptedSubcomponents;
 
+    // A flat list of all the subcomponents that comprise this Component.
+    // The list must be complete and in the correct System order by 
+    // addToSystem().
+    mutable std::vector<SimTK::ReferencePtr<const Component> > _orderedSubcomponents;
```
chrisdembia: Why don't member or adopted subcomponents need to be ordered with the property subcomponents?

chrisdembia: Sorry; forget this comment. I need to get a better understanding of the changes first.

OpenSim/Simulation/Control/TrackingController.cpp
chrisdembia: Despite Tom's excitement, I don't think this is what we should do. `_desiredStatesStorage` is a non-owning pointer and it's effectively being used as an owning pointer now. If the memory management has changed, then `_desiredStatesStorage` should be changed to a `SimTK::ResetOnCopy<std::unique_ptr<Storage>`, or something like that.

Did you discover a leak? If so, where?

I think it would be good to also edit the doxygen for this method to mention the memory management.

tkuchida: > Despite Tom's excitement, I don't think this is what we should do.

I didn't perform a thorough analysis of this change. I checked [the commit where this change was made](https://github.com/opensim-org/opensim-core/pull/1205/commits/02b4c1cc8509d758a245e65acd6d63c0251dfa79), and it seemed as though @aseth1 just happened to find the leak and patched it up without redoing the whole kitchen. If further analysis is necessary, I suggest opening a new issue to investigate since this class seems unrelated to the point of this PR.

chrisdembia: Yeah...I would prefer that this change is undone and a new Issue be created.

Applications/CMC/test/testCMCWithControlConstraintsGait2354.cpp
aseth1: I agree. But that wasn't the reason for the change. Tightening the tolerances was the main purpose. As far as I can tell the order has been wrong since we started using the `MultibodyGraphMaker` but the tolerances were to loose make a difference.

OpenSim/Common/Component.cpp
aseth1: Now that Component writers can dictate the order of subcomponents this will often mean cycling through Component lists of specific types, and it will be difficult to to guarantee they are attempting to add components only once. Since property changes should wipe out the system, this should be safe to do. My inclination is that it is too harsh to throw an Exception in this case.

`Component::setNextSubcomponentInSystem()` only appends the component to the list of ordered subcomponents if the subcomponent is not already in the list, so it is unique for a given Component. However, each subcomponent also has its own list.

But there is nothing stopping the component from directly calling `addToSystem` on a subcomponent. Note that previously, we didn't check and simply added the component to the system again. So this at least fixes that problem.

Here is the use case. A Model has Ground, Bodies, PhysicalOffsetFrames (POFs), Joints, Forces and user defined Components (like a Device). Some Joints have their own POFs (as subcomponents) and others connect to POFs somewhere else in the Model (like on Bodies or a Device). The MultibodyGraphMaker dictates which Bodies and POFs are to be added first to define each Joint in multibody tree order until the POFs, Bodies and Joints that make up the multibody tree are part of the System. At this point there are still POFs (not specifically part of the multibody tree) that have not been added to the System yet (think POF of the Device for attaching it). So a reasonable thing to do is to cycle through all POFs and add them next sinces, POFs act as attachment sites for other components (e.g. Stations, Forces, Device). POFs that were added as a requirement for the building the tree will already be present and I think it makes sense to now ignore them as the Model does another pass to make sure all POFs are added before attempting to add other Components to the System.

Let me know if you agree.

OpenSim/Common/Component.h
aseth1: `_orderedSubcomponents` is just a way for a Component to dictate the order in which subcomponents get added to the System. If the Component writer has no requirement on order then the list is auto generated in ownership traversal order. It includes all subcomponents regardless of internal designation (member, property, adopted)

aseth1:
@chrisdembia I reverted the memory leak "fix". Thanks for pointing it out - it was an oversight on my part.

Applications/CMC/test/testCMCWithControlConstraintsGait2354.cpp
chrisdembia: I know; it's not a new problem. It's fine if it's not addressed for this PR but it would be good to make the tests not depend on the order.

OpenSim/Common/Component.cpp
chrisdembia: Thanks for the explanation. I need to read more of the PR and keep thinking before giving an opinion.

chrisdembia:
*[OpenSim/Simulation/Model/PhysicalFrame.cpp, line 73 \[r6\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KR4ZFrMpIJZ38t5Z8T2:-KR4ZFrNezvSbEgYS2gP:b346fe7) ([raw file](https://github.com/opensim-org/opensim-core/blob/9dd7ef465817707c6a9772dd0d8095d82298394d/OpenSim/Simulation/Model/PhysicalFrame.cpp#L73)):*
> ```cpp
> {
>     OPENSIM_THROW_IF_FRMOBJ(!mbix.isValid(), Exception,
>         "Assigned an invalid SimTK::MobilizedBodyIndex");
> ```

BTW +1.

(I have to say "BTW" to tell Reviewable to bypass this comment)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
*[OpenSim/Tools/CorrectionController.cpp, line 277 \[r6\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KR4Zqp2-YwXurv07wH_:-KR4Zqp3iKbuQusIBb7A:bpp1sle) ([raw file](https://github.com/opensim-org/opensim-core/blob/9dd7ef465817707c6a9772dd0d8095d82298394d/OpenSim/Tools/CorrectionController.cpp#L277)):*
> ```cpp
>             // the controller is removed, so are all the actuators it added.
>             adoptSubcomponent(actuator);
>             setNextSubcomponentInSystem(*actuator);
> ```

Seems like a potentially good use of `adoptSubcomponent()`. 

As a result of this change, serializing a model after calling `Model::initSystem()` will no longer cause the added actuators to be printed out---a backwards-incompatible change.

I think this is okay though, mainly because I don't think the correction controller is widely used. I also think this new behavior is better.

Do you think this change should be mentioned in the changelog?

Have you tested the CorrectionController with this change? I can see that the CorrectionController is used in testControllers, but there is no check that it is doing the right thing (e.g., that the actuators are added and used).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
*[OpenSim/Tools/CMCTool.cpp, line 653 \[r6\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KR4c1itirk2AnAQG35Q:-KR4c1itirk2AnAQG35R:b159xyd) ([raw file](https://github.com/opensim-org/opensim-core/blob/9dd7ef465817707c6a9772dd0d8095d82298394d/OpenSim/Tools/CMCTool.cpp#L653)):*
> ```cpp
>     const CoordinateSet& coords = _model->getCoordinateSet();
>     for (int i = 0; i < nq; ++i) {
>         coords[i].setValue(s, q[i], i==(nq-1));
> ```

I understand it, but the boolean expression in the third argument warrants a comment in the code.

Do you think it might read better to pass `false` every time in the loop, and then outside the loop, call `model.assemble()` or something like that?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
I have finished reviewing all the EASY stuff. I'll come back later to review the meat of the PR.

aseth1:
> Have you tested the CorrectionController with this change? I can see that the CorrectionController is used in testControllers, but there is no check that it is doing the right thing (e.g., that the actuators are added and used).

Yes, `testForward` has an explicit test, `testGait2354WithController`,  for the `CorrectionController`. The previous test is without the controller and with the controller tracking tolerances are tightened to make sure the controller (and its actuators) are having an effect. Note, this check was already in place.

aseth1:
> Do you think this change should be mentioned in the changelog?

I have no issue with adding this fix to the changelog, but I feel this is more of a bug fix than a change to the feature. I don't think users would have expected their models to suddenly have extra actuators after running forward. If we decide  to add it to the changelog, I will frame it as a bug fix.

chrisdembia:
*[OpenSim/Common/Component.cpp, line 1186 \[r6\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KR9iXjNIFR1exd5APHg:-KR9iXjOQlangKywMsHO:b-aekisz) ([raw file](https://github.com/opensim-org/opensim-core/blob/9dd7ef465817707c6a9772dd0d8095d82298394d/OpenSim/Common/Component.cpp#L1186)):*
> ```cpp
>         mySubcomponents.push_back(
>             SimTK::ReferencePtr<const Component>(compRef.get()) );
>     }
> ```

To discuss later: I'm not sure we want to give users access to the member subcomponents. Otherwise, they could change the properties of that subcomponent. One reason we created the member subcomponents was so that the owning component could have control over its property values.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
*[OpenSim/Common/Component.cpp, line 1186 \[r6\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KR9iXjNIFR1exd5APHg:-KR9im4sYSofqRb79Qkr:bxyye5v) ([raw file](https://github.com/opensim-org/opensim-core/blob/9dd7ef465817707c6a9772dd0d8095d82298394d/OpenSim/Common/Component.cpp#L1186)):*
> To discuss later: I'm not sure we want to give users access to the member subcomponents. Otherwise, they could change the properties of that subcomponent. One reason we created the member subcomponents was so that the owning component could have control over its property values.
</details>
I know that they only get a const reference here, and that's better. But we should discuss.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
*[OpenSim/Common/Component.h, line 1462 \[r6\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KR9mgL_W897U7AG7T4k:-KR9j6aKHqqfaevtBMIG:b-b88bpy) ([raw file](https://github.com/opensim-org/opensim-core/blob/9dd7ef465817707c6a9772dd0d8095d82298394d/OpenSim/Common/Component.h#L1462)):*
> ```objc
>     /** Access this Component's immediate subcomponents (not those owned by
>         subcomponents) */
>     std::vector<SimTK::ReferencePtr<const Component>> getMySubcomponents() const;
> ```

Not for this PR, but it might be good to come up with consistent terminology that we like for this; it is likely to come up again. I like "immediate", "local", "direct children".

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
*[OpenSim/Simulation/Model/Model.cpp, line 511 \[r6\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KR9tQjDe6a-bmLx3xjM:-KR9tQjDe6a-bmLx3xjN:b6w2veg) ([raw file](https://github.com/opensim-org/opensim-core/blob/a2613cd69cd68f8713c8357128e26896dbf9c71e/OpenSim/Simulation/Model/Model.cpp#L511)):*
> ```cpp
>     }
> 
>     updCoordinateSet().populate(*this);
> ```

This line of code seems like it should stay in `extendFinalizeFromProperties()`. How come it had to move?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
*[OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp, line 146 \[r6\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KR9wIN3lusddnJt4sBV:-KR9wIN3lusddnJt4sBW:b-62ucsh) ([raw file](https://github.com/opensim-org/opensim-core/blob/9dd7ef465817707c6a9772dd0d8095d82298394d/OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp#L146)):*
> ```cpp
> 
>     /* Hack to obtain a modifiable state in a const method */
>     State& modCompState = const_cast<SimTK::State&>(_completeState);
> ```

What about naming this `mutCompState` instead, since I think we use that (e.g., `mutableThis`) more commonly after a const_cast.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
*[OpenSim/Simulation/Model/Model.cpp, line 749 \[r6\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KRAgVgjYI-UizuTahZP:-KRAgVgk43z33TOTH7In:b-78cfbw) ([raw file](https://github.com/opensim-org/opensim-core/blob/9dd7ef465817707c6a9772dd0d8095d82298394d/OpenSim/Simulation/Model/Model.cpp#L749)):*
> ```cpp
>     for (const auto& comp : components) {
>         setNextSubcomponentInSystem(comp.getRef());
>     }
> ```

From my understanding, `Model::_orderedSubcomponents` holds *all* Joints and POFs *anywhere* in the model, but only contains the "immediate" components for all other types (e.g., a Constraint inside a Joint will not be in `Model::_orderedSubcomponents`). I would expect `Model::_orderedSubcomponents` to contain either only "immediate" components, or to contain all components anywhere in the model.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
*[OpenSim/Common/Component.h, line 2458 \[r6\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KRAhCw85rcdi0zxDqYK:-KRAhCw9qW3oixP9wXb5:btsly3k) ([raw file](https://github.com/opensim-org/opensim-core/blob/9dd7ef465817707c6a9772dd0d8095d82298394d/OpenSim/Common/Component.h#L2458)):*
> ```objc
>     /// Helper method to enable Component maker to specify the order of
>     /// subcomponents to be added to the System used by addToSystem().
>     /// Use this method in extendConnect() of your Component to set the
> ```

Probably in `extendConnectToModel()` instead for ModelComponents.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
*[OpenSim/Common/Component.cpp, line 275 \[r6\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KRAhgWzl0ZI8ds-IzvL:-KRAhgX-KARWSoSywvvl:b-7ejbow) ([raw file](https://github.com/opensim-org/opensim-core/blob/9dd7ef465817707c6a9772dd0d8095d82298394d/OpenSim/Common/Component.cpp#L275)):*
> ```cpp
>     for (const auto& comp : allMyComponents) {
>         setNextSubcomponentInSystem(comp);
>     }
> ```

It seems inefficient that this code is run even for components that will completely override this behavior. For example, this code is run for `Model`, even though model will clear the list and come up with its own order. Even if the performance penalty is small, it seems messy.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
*[OpenSim/Simulation/Model/Model.cpp, line 742 \[r6\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KRAia01zk0fyCK5HcoR:-KRAia026v3PG3NWeNdK:b1woj1f) ([raw file](https://github.com/opensim-org/opensim-core/blob/9dd7ef465817707c6a9772dd0d8095d82298394d/OpenSim/Simulation/Model/Model.cpp#L742)):*
> ```cpp
>     for (const auto& pof : poFrames) {
>         setNextSubcomponentInSystem(pof);
>     }
> ```

POFs depend on other POFs, namely its parent frame. Question: does a POF's parent frame need to have `addToSystem()` invoked first before its own `addToSystem()` is invoked? Or can POFs be `addToSystem()`'d in any order? Do we have a test to verify the behavior of POFs' addToSystem() order being different (or backwards) from the parent-child relationship?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
*[OpenSim/Common/Component.cpp, line 338 \[r5\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KR4Y3JwQ1aETikSHt4A-r5-338:-KRAjogzHBggxYdDEurd:b-pxhiq7) ([raw file](https://github.com/opensim-org/opensim-core/blob/9cadc0fcf1bf85a9ce4df8e14c7cf5f29f325bba/OpenSim/Common/Component.cpp#L338)):*
> Thanks for the explanation. I need to read more of the PR and keep thinking before giving an opinion.
</details>
After reading some of the new code in Model, I now understand.  I don't like the fact that an `_orderedSubcomponents` in some components may contain non-immediate subcomponents. It would be cleaner if each Component's `_orderedSubcomponents` contained exactly all of its non-immediate subcomponents (and no others), or if only Model's `_orderedSubcomponents` list is populated (and populated with *all* components).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
*[OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp, line 155 \[r6\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KRAkttS9iPdqKx1OY0d:-KRAkttS9iPdqKx1OY0e:b-3eoes) ([raw file](https://github.com/opensim-org/opensim-core/blob/9dd7ef465817707c6a9772dd0d8095d82298394d/OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp#L155)):*
> ```cpp
>         // or not. It is being set to true when we set the last coordinate value.
>         coords[i].setValue(modCompState, _qWork[i] + _qCorrections[i], i==(nq-1));
>         coords[i].setSpeedValue(modCompState, _uWork[i] + _uCorrections[i]);
> ```

Does this new code (invoking setValue() and setSpeedValue()) affect performance? I would think not, but I'm not sure.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
*[OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp, line 155 \[r6\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KRAkttS9iPdqKx1OY0d:-KRAlDlpzx1Iz2HzprB6:b-2z7dma) ([raw file](https://github.com/opensim-org/opensim-core/blob/9dd7ef465817707c6a9772dd0d8095d82298394d/OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp#L155)):*
> Does this new code (invoking setValue() and setSpeedValue()) affect performance? I would think not, but I'm not sure.
</details>
I guess in the context of CMC, it really shouldn't matter. But I am curious how much longer it takes to use these setters than to play with Arrays/Vectors directly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

aseth1:
>>    Does this new code (invoking setValue() and setSpeedValue()) affect performance? I would think not, but I'm not sure.

> I guess in the context of CMC, it really shouldn't matter. But I am curious how much longer it takes to use these setters than to play with Arrays/Vectors directly.

In debug the profiler indicates it takes 3.2% of all CPU time - so not free. Should be less in release, but profiler is not reading in symbols properly so cannot attribute to function names (just addresses)

chrisdembia:
> In debug the profiler indicates it takes 3.2% of all CPU time - so not free. Should be less in release, but profiler is not reading in symbols properly so cannot attribute to function names (just addresses)

Okay; thanks for checking. Sounds like more than we would want for simply moving data around. But I think it's fine...just something to keep in mind for later.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-:-KRAzV1A2wsbx_yur9d0:b-7snz05)*

chrisdembia:
*[a discussion](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KRCIhZogLfxV38DG-_Z:-KRCIhZogLfxV38DG-__:b-9h1wrs) (no related file):*
To summarize my discussion with @aseth1 earlier today:

1. The most compelling example we could think of so far of a component that would override the default `_orderedSubcomponents` would be a root-like component similar to Model, like a Device component that would need to serve as the root component. It seems that most of the dependencies between Simbody classes are just on MobilizedBodies (e.g., while Measures depend on other Measures, we have decoupled them).
2. I'm skeptical that it is necessary for each component to have an `_orderedSubcomponents` list, but we will proceed with that for now and keep an eye on whether or not we find it to be necessary.
3. @aseth1  will see if there are ways to hide the `_orderedSubcomponents` even more than they are currently.
4. I have a better understanding of @aseth1's desire to put the `_orderedSubcomponents` member variable in `Component`: to make Component more capable without the Model.
5. If we eventually have an Environment class, the ordering could be handled there exclusively. But we wouldn't do this for 4.0.

That's all I can remember right now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
*[a discussion](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KRCRoUbWQk-1DzGJrrJ:-KRCRoUbWQk-1DzGJrrK:b-xt9ap) (no related file):*
As far as I can tell, the `_orderedSubcomponents` are only *used* during `addToSystem()`, after which that ordered list is not needed. Would it make sense to clear it after `addToSystem()` is done? That could help avoid any use of the `_orderedSubcomponents` list for other unintended purposes. On the other hand, if one calls `initSystem()` twice in a row and does not need to re-connect, etc., then the previously-formed `_orderedSubcomponents` could be re-used and should not need to be formed again.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
*[OpenSim/Common/Component.h, line 2514 \[r5\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KR4Y3JwQ1aETikSHt4B-r5-2514:-KRCSnE8eCCIn2Vzbegw:buul22f) ([raw file](https://github.com/opensim-org/opensim-core/blob/9cadc0fcf1bf85a9ce4df8e14c7cf5f29f325bba/OpenSim/Common/Component.h#L2514)):*
> `_orderedSubcomponents` is just a way for a Component to dictate the order in which subcomponents get added to the System. If the Component writer has no requirement on order then the list is auto generated in ownership traversal order. It includes all subcomponents regardless of internal designation (member, property, adopted)
</details>
The doxygen description for this member variable could be clearer. Consider:

"A flat list of subcomponents (immediate and otherwise) under this Component. This list must be populated prior to addToSystem(), and is used strictly to decide the order in which addToSystem() is invoked on components."

It would be good explain (beyond what I did above) the fact that the vector can contain both direct/immediate and non-immediate subcomponents, and does not need to include *all* subcomponents.

If you use the phrase "System order," it would be good to provide a definition of what this means. E.g., "Simbody objects depend on other Simbody objects; e.g., Forces depend on MobilizedBodies, and so MobilizedBodies must be created prior to the Forces."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
*[OpenSim/Common/Component.h, line 2465 \[r7\]](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205#-KRCV7F_lhSraxbVjv11:-KRCV7F_lhSraxbVjv12:bccpopz) ([raw file](https://github.com/opensim-org/opensim-core/blob/8a00188f0d716c03129ffc0c666a9fc5c57c36cd/OpenSim/Common/Component.h#L2465)):*
> ```objc
>     /// subcomponent.
>     /// If the subcomponent already appears in the order list setting it
>     /// later in the list has no effect. The list remains unique.
> ```

Just a thought (no changes to code necessary): It doesn't *feel* like this step really belongs in the "connecting" phase. I understand that it must happen after connections are satisfied, but it does not itself have to do with the connections. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
Reviewed 12 of 24 files at r1, 3 of 7 files at r2, 1 of 1 files at r3, 3 of 5 files at r4, 7 of 7 files at r5, 1 of 1 files at r6.
Review status: 26 of 27 files reviewed at latest revision, 29 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensim-org/opensim-core/1205)*

chrisdembia:
I have finished my review. It's an interesting solution to the problem that I had not thought of. I've expressed my concerns. While I am skeptical that the flexibility is necessary or warranted by the slightly confusing nature of the solution, I think we should try it and monitor it. In general, the documentation of the change was very good. The main thing that could be improved about the documentation is the fact that any Component's `_orderedSubcomponents` could contain non-immediate children.

aseth1:
>>
 > ```cpp
 >     }
 >     updCoordinateSet().populate(*this);
 > ```

>This line of code seems like it should stay in `extendFinalizeFromProperties()`. How come it had to move?

Had to move because Joints can exist anywhere not only in the `JointSet`. As such I have to wait until all Components are done finalizing from their properties (which marks properties as subcomponents and could include Joint(s) internally) before I can iterate over Joints . This is now performed after `finalizeFromProperties` in the Model constructors so I know i have a valid and complete `ComponentList<Joint>`. 

The alternative is to change the order of  operations in `Component::finalizeFromProperties()` such that `finalizeFromProperties()` is invoked on subcomponents before the Component. Then, in the case of a top-level `Model` it will have all its Joints finalized before `extendFinalizeFromProperties()` is called on itself. At which  time the `getComponentList<Joint>()` will be complete. I will try that, especially if we are going to remove `finalizeFromProperties` from `Model` constructors.

aseth1:
> The alternative is to change the order of operations in Component::finalizeFromProperties() such that finalizeFromProperties() is invoked on subcomponents before the Component. Then, in the case of a top-level Model it will have all its Joints finalized before extendFinalizeFromProperties() is called on itself. At which time the getComponentList<Joint>() will be complete. I will try that, especially if we are going to remove finalizeFromProperties from Model constructors.

It worked. `updCoordinateSet().populate(*this);` was reverted back to `Model::extendFinalizeFromProperties` and removed from Model constructors.

aseth1:
@chrisdembia. I believe I have addressed the issues you raised except for the `getMySubcomponents` (now renamed `getImmediateSubcomponents`) also providing const access to member subcomponents. I think this should stay in. As a protected method, its main purpose is to delegate addToSysetm calls to subcomponents and we would not want to exclude member subcomponents here.

OpenSim/Simulation/Test/testInitState.cpp
```diff
@@ -146,23 +146,44 @@ void testStates(const string& modelFile)
...
 }
 
+size_t getMedianOfCurrentRSSOverNSamples(size_t nsamples) {
+    // NOTE: getCurrentRSS() can be unreliable at evaluating the memory usage
+    // of a current process because of caching, shared memory and a litany of 
+    // other reasons including some platform specific details.
+    // Taking an average over nsamples just in case we get a flaky estimate.
+    // A stable estimate of memory usage is necessary so that we do not find
+    // cases where loading model either decreases or has no change in memory
+    // usage, which is impossible.
```
tkuchida: Not sure this is the right approach, and I don't see how this change is related to the topic of the PR ("Fix path failure when WrapObject is attached to a PhysicalOffsetFrame"). Concerns:

1. It is still possible for the `size_t mem0 = getMedianOfCurrentRSSOverNSamples(nsamples);` line to be executed immediately before a chunk of memory is freed, in which case the `ASSERT(current_size > mem0, ...);` will be thrown. We have lots of false alarms on CI (e.g., testContactGeometry in Debug mode) and it would be good to avoid creating more of these.

2. Instead of calling `getCurrentRSS()` 10 times in a row, perhaps the entire operation (set `mem0`, load model, compute memory increase) should be repeated.

3. It isn't clear why the median is being computed rather than, say, the maximum (e.g., if memory is freed after the first iteration, `samples[nmedian] < samples[1]` and perhaps `samples[nmedian] < mem0`).

I suggest backing off this commit and investigating in a separate PR.

chrisdembia: I agree; this change does not belong in this PR.

If this PR will not pass tests no matter what without such a change, then I would prefer a separate PR into master, which then also gets merged into this PR before it gets merged itself.

chrisdembia:
> I will try that, especially if we are going to remove finalizeFromProperties from Model constructors.

I cannot think clearly about this right now, but it might be that it *is* okay to call `finalizeFromProperties()` from `Model`, so long as Model is `final`. That is, as long as  `finalizeFromProperties()` does not invoke the virtual method `extendFinalizeFromProperties()` that is implemented in types derived from Model. I think the previous issue was that we were calling virtual functions from the `Component` constructor, and that is not okay because it is a base class. https://www.securecoding.cert.org/confluence/display/cplusplus/OOP50-CPP.+Do+not+invoke+virtual+functions+from+constructors+or+destructors

OpenSim/Common/Component.h
```diff
@@ -362,12 +362,16 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
     ///@{
 
-    /** Update Component's internal data members based on properties.
-        Marks the Component as up to date with its properties. */
+    /** Define a Component's internal data members and structure according to
+        its properties. This includes its subcomponents as part of the component
+        ownership tree and identifies its parent (if present) in the tree.
+        finalizeFromProperties propagates to all of the component's
+        subcomponents prior to invoking the virtual extendFinalizeFromProperties()
+        on itself.*/
```
chrisdembia: 👍

OpenSim/Common/Component.h
```diff
@@ -2508,9 +2512,18 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
-    // A flat list of all the subcomponents that comprise this Component.
-    // The list must be complete and in the correct System order by 
-    // addToSystem().
+    // A flat list of subcomponents (immediate and otherwise) under this
+    // Component. This list must be populated prior to addToSystem(), and is
+    // used strictly to specify the order in which addToSystem() is invoked
+    // on its subcomponents. The order is necessary for the construction
+    // of Simbody elements (e.g. MobilizedBodies, Constraints, Forces, 
+    // Measures, ...) which cannot be added to the SimTK::MultibodySystem
+    // arbitrarily. In the case of MobilizedBodies, for example, the parent
```
chrisdembia: "arbitrarily" -> "in arbitrary order"

OpenSim/Common/Component.cpp
```diff
@@ -369,8 +359,26 @@ void Component::baseAddToSystem(SimTK::MultibodySystem& system) const
 
 void Component::componentsAddToSystem(SimTK::MultibodySystem& system) const
 {
-    for (const auto& compRef : _orderedSubcomponents) {
-        compRef->addToSystem(system);
+    // If _orderedSubcomponents is specified, then use this Component's
+    // specification for the order in which subcomponents are added. At a
+    // minimum the order for all immediate subcomponents must be specified.
```
chrisdembia: The doxygen for `setNextSubcomponentInSystem()` should mention this requirement as well.

OpenSim/Common/Component.cpp
```diff
@@ -369,8 +359,26 @@ void Component::baseAddToSystem(SimTK::MultibodySystem& system) const
...
+        auto mySubcomponents = getImmediateSubcomponents();
+        for (const auto& compRef : mySubcomponents) {
+            compRef->addToSystem(system);
+        }
+    }
+    else {
+        OPENSIM_THROW_FRMOBJ(Exception, 
+            "_orderedSubcomponents specified, but does not include all "
+            "immediate subcomponents."
+        )
```
chrisdembia: I am fine with this implementation but the comments and this exception message are not exactly correct. If this exception is thrown, all we know is that the number of pointers in that vector is the same as the number of immediate subcomponents. I don't think a more rigorous/exact check is needed, but the comments/exception message should be weaker I think.

OpenSim/Common/Component.h
```diff
@@ -2463,19 +2465,20 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
-    /// according to the multibody tree and add bodies and joints in order
+    /// Use this method in extendConnect() of your Component (or within 
+    /// your extendConnectToModel() for ModelComponents) to set the
+    /// order of your subcomponents. For example, Model orders subcomponents
+    /// according to the multibody tree and adds bodies and joints in order
     /// starting from Ground and growing outward.
-    /// Note: call resetSubcomponentOrder() prior to setting the first
-    /// subcomponent.
     /// If the subcomponent already appears in the order list setting it
     /// later in the list has no effect. The list remains unique.
```
chrisdembia: Can you add a comment here to the effect of "You should *rarely* need to use this; it is mainly used by Model."

OpenSim/Common/Component.h
```diff
@@ -2463,19 +2465,20 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
     void setNextSubcomponentInSystem(const Component& sub) const;
 
+    /// resetSubcomponentOrder clears this Component's list of ordered
+    /// subcomponents (which leaves subcomponents untouched). Rebuild the
```
chrisdembia: Slight preference for wording: "which leaves..." -> "but leaves...".

OpenSim/Simulation/Test/testInitState.cpp
aseth1: OK

aseth1:
> What do you think about adding a test related to the order of PhysicalOffsetFrames? Despite the improvements in this PR so far, I think that it is still possible for PhysicalOffsetFrames to be added in the incorrect order. Especially if there is no Body downstream of the PhysicalOffsetFrames.

I think it is a necessary test and this PR introduces a mechanism for handling it (appropriately order the subcomponents) but I think doing that here is just going to make this PR huger than it already is.

I have the test in place and as expected it fails, with the PhysicalOffsetFrames not added in tree order (that is the frame it is dependent on is not added in first). I have a good idea of what a fix would look like (traverse the parentFrame(s) and ensure that it (they) are in the ordered list) but I don't think it will be a small incremental change.

Mainly it has implications on when we create the multibody tree and the ordered list. There is a TODO in `Model::createMultibodyTree()` to postpone creation of the Multibody tree to after all connections are established and there are two direct ways to achieve this: in `Component::connect` push `extendConnect()` (by extension `extendConnectToModel()`) to after `componentsConnect()` which means `Model::extendConnectToModel()` is the last to be called. Alternatively we could keep the current order of operations in `connect()` and move the Tree building and ordering operation to be the first thing that happens in addToSystem (where `Model::extendAddToSystem()` is called prior to its `componentsAddToSystem()`). I may try both (unless you see a clear benefit one way or the other) to see which allows us to preserve the order of operations across the propagated methods. Anyhow, I think it warrants another PR.   It's been on my radar but lower priority than this PR (although clearly related).

I can post the current test that fails as an issue to make it clear what needs to be achieved in the followup PR. Let me know if that is reasonable.

chrisdembia:
@aseth1 I am done with my review and the changes look good. I think this can be merged. I am not going to merge this myself yet because two of the tests were failing. I understand that you had made commits to make the tests pass, so it may be okay that the tests are failing, but I am unsure of the best way to proceed. I've restarted the tests. 

Ideally, the PR to fix the tests would come separately and could be merged before this gets merged. But I am also okay with merging this despite the test failures, so long as the subsequent test-fixing PR comes quickly.

aseth1:
@chrisdembia thank you for your review. I unwound all the changes to `testInitState`, which is still failing on gcc with wrapping on. It fails on the ASSERT [here](https://github.com/opensim-org/opensim-core/pull/1205/files#diff-44383ef8b20005e92ec3c8ce2cbe123dL197) that model_size > 0

I cannot see any direct connection with this failure an the changes made in this PR. I am happy to address the `testInitState` failure instituting suggestions from @tkuchida to attempt repeat the test, when this occurs.

chrisdembia:
> I am happy to address the testInitState failure instituting suggestions from @tkuchida to attempt repeat the test, when this occurs.

I'm not sure what course of action you are suggesting. Feel free to merge this and handle the test failures as you see fit. It is likely that the tests on master will now fail (affecting all other PRs), so I would think the improvement to testInitState would need to follow soon.

tkuchida:
> I'm not sure what course of action you are suggesting.

Created issue #1263 with detailed explanation and pseudocode.

aseth1:
I merged PR branch for #1267 into this PR to see if it actually fixes the issue here. Of course everything is fine locally on Windows. Let's see what happens!

aseth1:
@chrisdembia @tkuchida. Given that all the tests now pass, I think it is fair to say that the previous failure was due to `getCurrentRSS()` anomalies in `testInitState` and not due the the principal changes of the PR.

There is definitely more work to be done with memory usage test in #1267 that I will continue to do. But I would now like to decouple the two PRs. I suggest merging this as is and then I will merge master into #1267 and continue to address issues/comments raised there.

If you agree please merge. Thanks!

tkuchida:
Did you intend to push your merge of #1267? I would prefer to merge this PR at bf9061c and then have the new testInitStates code reviewed on the other branch before merging it to master.

aseth1:
I did intend to merge #1267 here to test that it would have an effect- no other way for me to verify on CI AFAIK. I can unwind, wait for CI to fail again and then merge previous. I have no intention of ditching #1267 -it will simply continue from this point (also where the PR is now). Up to you if you want to wait to revert but I don't think it buys us much. Yes, `testInitState` is in temporarily in limbo.

aseth1:
@tkuchida @chrisdembia forget merging this now. I will fix #1267 first. Get it merged to master and then merge master here as @chrisdembia originally requested.

tkuchida:
> no other way for me to verify on CI AFAIK

I would have merged locally, commited to a new throwaway branch, then pushed the branch to GitHub to get CI to do the testing for you. I don't think anybody notices if you do this overnight :wink:

aseth1:
@tkuchida no worries. Lets get #1267 in first. Thanks!

aseth1:
@tkuchida @chrisdembia, master (with #1267) was merged with this PR and `testInitState` is now passing across the board (Yay!). There is the `testContactGeometry`  failure in debug, which is timing out on travis (known issue reported in #393). If there are no further issues please merge.

tkuchida:
Great! :+1:

aseth1:
Thank you! :kissing_closed_eyes:

tkuchida:
@aseth1: Please confirm that #1105 was supposed to close. You said "This PR likely fixes #1105", which triggered the issue to close automatically (GitHub doesn't parse the modifier "likely").

aseth1:
No should not be closed. I reopened #1105. Thanks for pointing it out!

Issue #775, opened 2016-02-18
Need Schema for osim files


aymanhab:
The XML files we use in OpenSim are called "Well-formed" that is, they don't have a way to validate them. This option is flexible but leaves users at a loss as to what can be in the file. We should create and publish a schema for built in types that users can disable if extending classes/format.

chrisdembia:
+1 !!! I've worked on this. I'd say I'm 80% done with having a complete schema. With a complete schema, we can also provide autocompletion in an XML text editor.

Here's the code: https://github.com/opensim-org/opensim-core/tree/xmlschema

Want to work on this (with me)? 

The *big* issue I've run into is that I don't think it's valid XML to have `::` in a tag name, since `:` is used for xml namespaces, or something like that. Therefore, there's an issue with `HuntCrossleyForce::ContactParameters`, etc (nested types). Either we have to have an alias for `::` when writing to XML, an alias for each individual nested type, or we need to avoid registering nested types.

A smaller issue that I'm confused about is where we would host the schemas (online? in the installation?).

We can actually automatically generate a new schema for people who extend classes (with plugins), so long as they register their new types.

aymanhab:
Awesome, sure would love to work on it. I'd host it on confluence or github
for standalone validation. I thought  Sherm used something else for :: in
class names, I have this on an issue somewhere.

chrisdembia:
I talked to him about it and I forgot what came of it...He writes `::`'s to his upcoming state xml files. @sherm1

chrisdembia:
Oh, the other issue is that single-value Object properties can be written in two different ways, and it maybe that only one of these ways can be easily accomodated by a schema. See #707 for a description and a related bug.

sherm1:
I don't use class names as tags for the State serialization stuff. The class names are attributes, e.g. `<Value type="Some::Type::SubType"> blah </Value>`. Everything is allowed in attributes, although I substitute `{}` for `<>` because XML is weird about angle brackets.

chrisdembia:
Oh okay. That helps.

aseth1:
Is this a **need** for the 4.0 GUI? Would the GUI provide XML editing where it could be validated against the schema?

jenhicks:
This issues is in the "Next Release" milestone, so not in consideration for 4.0 release. As you are looking through issues, the ones to look at are those already in the OpenSim 4.0 milestone and those that have no assigned milestone.

PR #1210, opened 2016-08-25
Changed Joint's CoordinateSet to "coordinates" list property

tkuchida:
## Summary
This PR addresses #1165 and the second item in #1066 (items 3 and 4 to come in a future PR).

## List of changes
- Joint's Coordinates now live in a list property (`coordinates`).
- Added conversion to [`Joint::updateFromXMLNode()`](https://github.com/opensim-org/opensim-core/pull/1210/files#diff-2efad9acab01926bd89af8b957b3ac16R633) and incremented version number to 30507.
- Replaced `get_CoordinateSet()` calls with `getCoordinate()`, `get_coordinates()`, or `upd_coordinates()` as appropriate (`updCoordinate()` to be added in a future PR; tracked in #1066).
- Removed redundant (and now obsolete) `getCoordinateSet()` method (#1165).
- Removed useless (tautological) `ASSERT` [in Joint.h](https://github.com/opensim-org/opensim-core/pull/1210/files#diff-31e3d687c1da78d898daf827d6930458L476).


aseth1:
I am done my review. LGTM :+1:
I am ambivalent on upping the version. @aymanhab do you feel strongly one way or another?

(The now tautological Joint.h was a remnant from when `numCoordinates()` was virtual and set by the Joint maker. It became obsolete when initializers were introduced to fix the number of Coordinates. Thanks for removing it!)

aymanhab:
If it's easy I'd up the version (since the devweek was using 30506 so we may actually have user models with that format), but I don't feel strongly about it since we're still in development cycle..

tkuchida:
> the devweek was using 30506 so we may actually have user models with that format

Good point! I'll increment.

aseth1:
@aymanhab @tkuchida I agree.

aseth1:
@aymanhab i reviewed the mapping of the `Set` to corresponding property list calls which were very straightforward.  I also verified the tautological nature of the removed assert. I compared the updateFromXMLNode() implementation to recent changes I made to geometry, but I think your double check there would be most beneficial. If you are happy with it I think is safe to merge.

OpenSim/Simulation/SimbodyEngine/Joint.cpp
```diff
@@ -640,6 +629,26 @@ void Joint::updateFromXMLNode(SimTK::Xml::Element& aNode, int versionNumber)
                 childNameElt->setValue(childFrameName + "_offset");
             }
         }
+
+        // Version 30506 replaced Joint's CoordinateSet with a "coordinates"
+        // list property.
+        if (documentVersion <= 30506) {
+            if (aNode.hasElement("CoordinateSet")) {
+                auto coordSetIter = aNode.element_begin("CoordinateSet");
+                auto coordIter = coordSetIter->getRequiredElement("objects")
```
aymanhab: If w ehave a WeldJoint, will it have an empty "CoordinateSet"? Maybe check that coordSetIter is not at end before dereferencing?

tkuchida: > If w ehave a WeldJoint, will it have an empty "CoordinateSet"? Maybe check that coordSetIter is not at end before dereferencing?

AFAIK WeldJoints can never have a "CoordinateSet" element because they have no Coordinates and, therefore, cannot have any non-default Coordinate-related properties to serialize. Line 636 ensures the "CoordinateSet" element exists; if it does, I believe the XML schema (if we had one) would demand that it holds a non-empty "objects" element (hence my use of `getRequiredElement()`).

aseth1: I am pretty sure `objects` can be empty. A `WeldJoint` does have a `CoordinateSet` that is empty. There is also nothing stopping users from deleting all Coordinates in a `CoordinateSet` and in that case it should not attempt to add Coordinates to the property list. I think the check that `coordIter != objects_node->element_end()` is prudent.

tkuchida: The XML Browser in 3.3 says
```
<WeldJoint>
    ...
    <CoordinateSet>
        <objects />
        <groups />
    </CoordinateSet>
    ...
</WeldJoint>
```
but the CoordinateSet element is not actually serialized when you print a Model containing a WeldJoint. Currently, this PR requires only that, if a CoordinateSet element exists, then it contains an "objects" element; the `while` loop (below) will prevent an exception if the "objects" element is empty. I'll add more checks, but I would be curious to know how a developer is expected to know this since AFAIK none of the serialized WeldJoints in the repo have a CoordinateSet element.

aymanhab:
It's a good point @tkuchida that our models in release 3.3 don't have these elements but users edit files manually or using Matlab scripts and without a Schema these would be valid so it's possible that such models exist, that's why I suggested being defensive. Longer term we should publish a schema with the release (issue #775) and validate files against it. A recent issue reported a crash due to empty tags in a model file (fixed in PR #1193 by adding a similar check). 

Thanks for your patience @tkuchida :camel: LGTM

tkuchida:
> users edit files manually or using Matlab scripts and without a Schema these would be valid so it's possible that such models exist

Good point. I also like the idea of publishing a schema so users and developers are all speaking the same dialect.

chrisdembia:
The schema in the `xmlschema` branch is 80-90% complete! Mostly all that's left to do is make some tough decisions.

aymanhab:
Thanks @chrisdembia for the pointer and your work on the schema, will revisit to account for XML changes when the dust settles :ice_cream:

tkuchida:
Have we converged?
Can this be merged?

aymanhab:
I have no issues with the latest updateFromXML incarnation :+1: Feel free to merge if @aseth1 has no other outstanding issues.

aymanhab:
Merging as the issues seem to have been resolved. Thanks @tkuchida

tkuchida:
Thanks for merging :fireworks:

Issue #1209, opened 2016-08-25
Provide informative messages when encountering orphaned Components


aseth1:
Although the use of `auto` is convenient it is leading to more errors in user code. When a method returns a typed pointer `auto` yields a correctly typed pointer, however, if the type returns a `const T&` (reference) auto is happy to give you a copy of type T if you do not explicitly request a reference, `auto&`.  This is very problematic for all our methods that return references into a Model with a context (order, topology, connections) since a copy is technically not a part of that Model. The copy is "orphaned" (has no parent) but otherwise acts like any other Component until you try to hook it up to anything. First, calls to its finalizeFromProperties, connect, ... will not be invoked automatically from its parent (it has none). If connect is called directly on an orphan its dependencies cannot be found (it isn't part of anything) OR things dependent on it cannot find it OR, worst of all, the original component is found (satisfying the dependency) but all the property changes exist(ed) on an orphaned component that has no effect on your Model. This can happen easily in range for loops (e.g. `for(auto musc : model.updComponentList<Muscle>) { ...` where you think you are editing each muscle in the Model when you are in fact editing a copy of each muscle which will be out of scope at the end of the loop.  

This issue first came up a couple of times at the developer's workshop and recently appeared in a litany of Collin's recent problems. "Who knew one little '&' could wreak so much havoc" he said. No kidding.

I don't think this issue can be easily prevented (short of using the PIMPL idiom) but it would be possible to determine if you were calling connect() on an orphan (easiest) and flag dependencies that cannot be found as potential orphans (but could be for other reasons like typos). But because we routinely edit properties before we add a Component to its parent, we cannot prevent the editing of accidentally orphaned components, which can be the most annoying/difficult to debug.

If you have a potential solution please raise it.

tkuchida:
Some ideas come to mind… don't know whether any of these would be appropriate or feasible. Tagged with :apple:, :banana:, and :cherries: for commenting.

## Strict stages to model building
Could be implemented locally or globally.
  1. :apple: *Local* – Each Component has knowledge of whether or not it is an orphan, and its properties/connectors cannot be modified until it has been adopted by another Component (Device, Model, World, …). Alternatively, the Component could have knowledge of whether it had ever been adopted and an exception could be thrown from the destructor when deleting a Component that was always an orphan, which could indicate that the Component went out of scope unexpectedly (though perhaps there are situations where it is acceptable to delete a Component that was never adopted).
[...]

PR #1214, opened 2016-08-26
Removed "auto" from user-facing code

tkuchida:
## Main changes
Removed "auto" from examples and example code in doxygen to avoid encouraging its use (and potential misuse), per discussion in #1209.

## Minor changes
1. Suppressed compiler warnings about unsigned ints.
2. Corrected headers in hopper example.

## Omissions
This PR does not repair the hopper example, which is again broken on master; noted in #1205. The hopper example should be added as a test so it doesn't keep breaking (see #1128).

### DIFF
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -755,7 +755,7 @@ Array<std::string> Component::getStateVariableNames() const
         stateNames[i] = (getFullPathName() + "/" + stateNames[i]);
     }
 
-    for (auto& comp : getComponentList<Component>()) {
+    for (Component& comp : getComponentList<Component>()) {
         const std::string& pathName = comp.getFullPathName();// *this);
         Array<std::string> subStateNames = 
             comp.getStateVariablesNamesAddedByComponent();
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -590,7 +590,7 @@ public:
      * that path name location does not exist OR it is not of the correct type.
      * For example, 
      * @code 
-     *    Coordinate& coord = model.getComponent<Coordinate>("right_elbow/elbow_flexion");
+     *    auto& coord = model.getComponent<Coordinate>("right_elbow/elbow_flexion");
      * @endcode
      * returns coord which is a Coordinate named "elbow_flexion" from a Joint
      * named "right_elbow" given it is a child of the Component (Model) model.
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -495,7 +495,7 @@ public:
      * component is traversed before its subcomponents.
      *
      * @code{.cpp}
-     * for (const auto& muscle : model.getComponentList<Muscle>()) {
+     * for (const Muscle& muscle : model.getComponentList<Muscle>()) {
      *     muscle.get_max_isometric_force();
      * }
      * @endcode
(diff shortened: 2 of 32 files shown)


klshrinidhi:
I don't think removing `auto` is going to solve the problem of users *unintentionally* making a copy of objects. This feature (`auto`) is considered very helpful, currently widely used and there are few instances where `auto` is the required. C++ standard is moving more toward using *automatic typed deduction*.

I think the solution here is to make a section somewhere (the currently unused Wiki on github or somewhere else) that lists the most common mistakes found and enlightening users about the right use of this feature `auto`.

klshrinidhi:
Another solution requiring architectural changes is to write classes (Components) using some variation of the PIMPL idiom. This will make it very cheap to perform copies and maintain references (or pointers) to original.

chrisdembia:
In my opinion, we should do both: avoid `auto` in examples and list common mistakes in documentation. Of course, in the code base, we should use `auto` as we please.

If we can figure out a "poor man's" PIMPL, that would be good too.

klshrinidhi:
I don't think it sends the right message if often end typing long names in examples and internally use `auto`. OpenSim is not the only way for a user to learn C++. One could have learned about `auto` somewhere else too.

aymanhab:
@klshrinidhi we need to keep in mind who are our users (or developers) and you have seen them in the dev week, very capable biomechanists, beginner/intermediate C++ at best. If using auto causes mysterious errors/issues that could be a show stopper. Learning C++ to the latest standard is not an option, unfortunately. An improved Matlab interface would probably fit them better but we're trying to hit the middle ground that lets us develop without hindering others.

chrisdembia:
> An improved Matlab interface would probably fit them better

I agree that this is probably the best way to solve problems related to the complexity of C++ for our users. On the other hand, those writing new components will probably not be able to use MATLAB.

tkuchida:
I think the most compelling reasons for removing `auto` in the examples are :one: to explicitly show the object types when it might not be immediately obvious otherwise (e.g., [here](https://github.com/opensim-org/opensim-core/pull/1214/files#diff-273445e80413bbe09be2b325b103d89eR254)) and :two: so that the C++ code parallels the scripting code.

chrisdembia:
> so that the C++ code parallels the scripting code.

In scripting, there are no types on the left hand side of the equals sign `=`. So, this would only increase the parallel to Java code.

chrisdembia:
> to explicitly show the object types when it might not be immediately obvious otherwise

This is definitely good for examples. There are cases where I do think it is fine to use auto in examples (like [here](https://github.com/opensim-org/opensim-core/pull/1214/files#diff-27aff0ebd0924330bf42bcf1069558adR125)): 

```
 auto* reporter = new ConsoleReporter();
```

The type is more or less obvious.

chrisdembia:
I do prefer `auto* rep = new Blah()` over `auto rep = new Blah()`.

aymanhab:
Sorry if my comment about similarity to scripting was a bit misleading, so let me clarify. It is true that you don't need to declare the type of object on lhs in python and Matlab (while you do in Java due to static typing), however, if the actual class underneath the handle is not wrapped in SWIG then you're dead in the water and can't use the handle you have. So, if we try to reproduce examples in python or Matlab, it will be easier to troubleshoot and to reproduce these examples if they don't use 'auto' or use it only in places where the class is obvious as described above by @chrisdembia .

tkuchida:
> use it only in places where the class is obvious

Ok, I can revert back to `auto` in cases like `auto* blah = new Blah();`. Any other cases?

tkuchida:
Merged master and reverted back to "auto" in situations where the class is obvious.

tkuchida:
This appears to be a controversial issue that requires more discussion before implementing. The issue is already being tracked as part of #1209.

Issue #596, opened 2015-09-03
Property names and paths: disallow ".", "/" "\", " " in object names.


chrisdembia:

tkuchida:
Perhaps prohibit `*`, `+`, and other such special characters (particularly to avoid potential confusion if regex searching is implemented—see #1030).

PR #1215, opened 2016-08-26
Path and ComponentPath classes

carmichaelong:
This PR addresses issue #1069 and tangentially issue #596.

This PR adds two classes. The Path class is an abstract object from which more specific types of Path's can derive. It is intended to contain much of the logic that any generic Path may need. ComponentPath is a concrete object that should be used specifically for Path's related to the Component interface. The Path class could also serve as a base class for a FileSystemPath class in a later PR.

Specifically towards #1069, ComponentPath is used to replace any of the path logic in Component without changing any of the function signatures.


chrisdembia:
Cool stuff. What is reasoning for making `Path` derive from `Object`? As is, is the class serialized? If not, is it planned that this class will be serialized? If so, it should have properties. Perhaps if there is a "path" class that needs to be serialized, it would use this class internally (in which case these classes do not need to derive from Object?). Just ideas...I'm not sure myself about what is the right inheritance hierarchy.

OpenSim/Common/Test/testPath.cpp
```diff
@@ -0,0 +1,125 @@
...
+/* The purpose of this test is strictly to check that the Path functions work
+ * outside of the context of the objects/components they are meant to service. 
+ */
+
+using namespace OpenSim;
+using namespace std;
+
+void testComponentPath() {
+    // Test that input matches printing out
+    string absStr1 = "/a/b/c/d";
```
chrisdembia: To help future developers with preserving the integrity of these tests, can you add comments explaining what each of these paths are intended to test?

OpenSim/Common/Path.h
```diff
@@ -0,0 +1,152 @@
...
+
+public:
+    /// Default constructor
+    Path() = default;
+
+    /// Construct Path from a string, given separator character and a string
+    /// of invalidChars. Performs a cleanPath() at the end of construction.
+    Path(const std::string path,
+         const char separator,
+         const std::string invalidChars);
```
chrisdembia: Are the invalid characters expected to be different for different instances of the same derived class? If not, then consider a different scheme for specifying them: use a virtual function that returns the invalid characters.

OpenSim/Common/ComponentPath.cpp
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.                                             *
+ * -------------------------------------------------------------------------- */
+
+#include "ComponentPath.h"
+
+using namespace OpenSim;
+using namespace std;
+
+const char separator = '/';
+const std::string invalidChars = "\\/*+";
```
chrisdembia: Would be slightly better to make these static member variables.

Also, if it helps, I think you could also use `std::vector<char>` to hold the invalid characters. Not sure though...

carmichaelong:
> What is reasoning for making Path derive from Object?

I think this was a bit mixed up in my line of thinking. At first, I thought it would be nice if it serialized, but in the use case here, Path and derived classes do not (or even should not) serialize. I'll drop it from deriving from Object in a future commit.

OpenSim/Common/ComponentPath.cpp
carmichaelong: The choice for `std::string` here was mainly due to convenience with working with many of the `string` functions. For example, `find_first_of`, `find_last_of`, etc. take in the characters to look for as `std::string`'s.

chrisdembia:
> I'll drop it from deriving from Object in a future commit.

Maybe wait till others chime in.

OpenSim/Common/Component.cpp
```diff
@@ -615,86 +615,27 @@ void Component::setParent(const Component& parent)
...
-    }
-
-    last = 0;
-    // process all the '/' nodes in the wrtP up to the common path
-    while ((jx = strj.find('/', last)) < std::string::npos) {
-        prefix += "../";
-        last = jx + 1;
-    }
+    ComponentPath thisP(getFullPathName());
+    ComponentPath wrtP(wrt.getFullPathName());
```
aseth1: :+1: It makes me so happy to see this hack of mine go! Thanks @carmichaelong

OpenSim/Common/Component.h
```diff
@@ -2686,26 +2696,26 @@ void ComponentListIterator<T>::advanceToNextValidComponent() {
...
 void Connector<C>::findAndConnect(const Component& root) {
  
-    const std::string& path = getConnecteeName();
+    ComponentPath path(getConnecteeName());
     const C* comp = nullptr;
 
     try {
-        if (path[0] == '/') { //absolute path name
-            comp =  &root.template getComponent<C>(path);
+        if (path.isAbsolute()) {
```
aseth1: nice.

OpenSim/Common/Component.h
```diff
@@ -2686,26 +2696,26 @@ void ComponentListIterator<T>::advanceToNextValidComponent() {
...
+            comp =  &root.template getComponent<C>(path.getString());
```
aseth1: I am now tempted to make `getComponent<C>()` take a `ComponentPath` instead of a `string`. Next PR.

OpenSim/Common/ComponentPath.h
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    bool operator!=(const ComponentPath& other) const
+    {
+        return this->getString() != other.getString();
+    }
+
+    /// Get an absolute path by resolving it relative to a given otherPath.
+    /// If the current Path is already absolute, return the same Path.
+    ComponentPath getAbsolutePath(ComponentPath* otherPath);
```
aseth1: I am having trouble understanding why another path is required to determine the absolute path. Is this more of a `formAbsolutePath` operation where the the current relative path is evaluated relative to  the `otherPath`? This would also require `otherPath` to be an absolute path, yes?

OpenSim/Common/ComponentPath.h
```diff
@@ -0,0 +1,90 @@
...
+    /// absolute.
+    ComponentPath getRelativePath(ComponentPath* otherPath);
+
+    /// Return the the sub-path that contains all sub components except for
+    /// the last one.
+    ComponentPath getParentPath();
+
+    /// Return a ComponentPath object of a subcomponent by index. This will
+    /// always be a relative path.
+    ComponentPath getSubcomponent(size_t index);
```
aseth1: Relative to what? My sense is that this is just the name of the directory (component) at depth = index. Could this just return a string then? I might rename to `getPathElementAtDepth` and return the string.

OpenSim/Common/ComponentPath.h
```diff
@@ -0,0 +1,90 @@
...
+
+    /// Return the last part of a ComponentPath. This will always be a relative
+    /// path.
+    ComponentPath getLastSubcomponent();
```
aseth1: How is returning an orphaned path here useful? For Path comparisons, perhaps? I am thinking I would use this method to extract just the Component's name from the path. In that case, should it return a string?

OpenSim/Common/Path.h
```diff
@@ -0,0 +1,152 @@
...
+*
+* This class stores the list of levels as a vector of strings. One char is used
+* to denote a path separator when either reading in or writing the Path out to
+* a string. A bool is stored to determine whether this Path should be resolved
+* relative to the root (denoted with a leading separator char) or not.
+*
+* @author Carmichael Ong
+*/
+class OSIMCOMMON_API Path : public Object {
+    OpenSim_DECLARE_ABSTRACT_OBJECT(Path, Object);
```
aseth1: My preference is to have `Path` be a built in type rather than an Object. It would be nice to have a corresponding `Property<Path>` as a SimpleProperty vs an ObjectProperty for serialization.

OpenSim/Common/Path.h
```diff
@@ -0,0 +1,152 @@
...
+
+    /// Return true if this Path is an absolute path.
+    bool isAbsolute() { return _isAbsolute; };
+
+    /// Return the number of elements in the Path
+    size_t getPathLength() const { return _path.size(); };
+
+    /// Push a string to the back of a path (i.e. the end). First checks if the
+    /// pathElement is legal. Does not change if path is absolute or relative.
+    void pushBack(std::string pathElement) {
```
aseth1: consider using `const std::string& pathElement` since `std::vector<std::string>::push_back` will make a copy anyways (I think).

OpenSim/Common/Path.h
carmichaelong: I don't think they should, except if you count maybe the case where a FileSystemPath would want different invalid characters depending on platform? In any case, the platform dependent part could be handled by the function you're suggesting.

OpenSim/Common/ComponentPath.h
carmichaelong: Sounds like you get what's going on here. Unlike with a path related to a file system, there is no global call for something like getCurrentWorkingDirectory(), and furthermore for `ComponentPath`, these functions won't really know what the equivalent getCurrentComponentPath() would be (hence why the Component should pass in this `otherPath`).

You're also correct that `otherPath` must be an absolute path. The implementation right now does check for this and throws an exception if that's not the case.

OpenSim/Common/ComponentPath.h
carmichaelong: It wasn't meant to be necessarily relative to something in particular, just that it would return `false` for `isAbsolute()`. I think you're right that the use cases really only need a string, so it could be good to just return one.

Concerning the name of the function, I'm open to ideas. I wanted to pick something that kind of hid the `pathElement` name and make the name more specific for Components, but if you think `pathElement` is clear, I can change it to your suggested name.

aseth1: I only saw that you used `pathElement` later.  Anything that conveys the path is in parts so `getPathElementAtDepth` or `getPathComponentAtDepth` might work. @chrisdembia  @tkuchida preferences?

tkuchida: > I only saw that you used `pathElement` later.  Anything that conveys the path is in parts so `getPathElementAtDepth` or `getPathComponentAtDepth` might work. @chrisdembia  @tkuchida preferences?

As someone who hasn't been following this thread, it isn't immediately obvious what a "path element" might be. A "path component" sounds like Path is a new type of Component, which doesn't appear to be the case (because Path derives from Object). I also wouldn't know what "depth" means in a call like `getPathComponentAtDepth(3)` and wouldn't know what to expect back.

Based on the comment on L78, I'm guessing that this method returns a path of any "descendant" (child, grandchild, great-grandchild, etc.) relative to the top-level Component. (If the method is referring only to immediate children of the Component, then it seems like the method would always trivially return the name of the subcomponent.)

Perhaps this isn't the intended purpose of the method, but I think something like `ComponentPath getRelativeSubcomponentPath(const Component& subcomp);` would be easily understood.

carmichaelong: Sounds like the function name and comments are unclear. As an example:
`ComponentPath path("/a/b/c/d");`
`ComponentPath subcomponentPath(2);`
`std::string str = subcomponentPath.getString(); // this returns "c" specifically not "/c"`
`subcomponentPath.isAbsolute(); // this returns false`

The purpose of this method was to walk through and compare parts of a path, as seen [here](https://github.com/opensim-org/opensim-core/pull/1215/files#diff-a8c6cc5aa99c4c61fcc4d4b0a45a4813R2095).

aseth1: > As someone who hasn't been following this thread, it isn't immediately obvious what a "path element" might be. A "path component" sounds like Path is a new type of Component, which doesn't appear to be the case (because Path derives from Object).

Unfortunately "Path" is highly overloaded. That is why the line "path" for muscles, ligaments, etc.. is a `GeometryPath` Component and see the confusion with using "PathComponent". Given a traversal path (like a file path for Components -I think that is why @carmichaelong used `ComponentPath`) and you want the subcomponent (like a sub-directory) name between the root and the leaf (not its relative path), what would you call it? Perhaps `getDescendantName(unsigned int level)`? I like having the `atDepth` or `atLevel` in the method name to indicate there are many descendants/elements/levels. The class description for `Path` uses "levels" as the string elements in a list.

tkuchida: > Unfortunately "Path" is highly overloaded. That is why the line "path" for muscles, ligaments, etc.. is a `GeometryPath` Component and see the confusion with using "PathComponent".

The word "path" is clear; what is not immediately obvious (to me, at least) is what you mean by "element" or "component" since AFAIK these terms deviate from the terminology typically used to discuss paths. I suggest using established terminology if possible.

I would either use the standard *file system* terminology throughout (e.g., `getSubdirectoryNameAtLevel()`) with the understanding that the Component tree is directly analogous to a file system, or use standard *tree data structure* terminology throughout (e.g., `getSubtreeNodeNameAtLevel()`).

chrisdembia: I prefer the file system terminology over the tree data structure terminology (this tree only has one branch).

Another term you could use is `segment`, though this is not part of either file system or data structure terminology.

carmichaelong:
Thanks for all the comments @tkuchida @aseth1 @chrisdembia. I've made some commits to address your comments.

@chrisdembia I wasn't completely sure what you meant with the virtual functions, so I took a stab at it and it may be completely wrong with what you were thinking, so let me know how to fix it if it needs to be fixed.

OpenSim/Common/Component.h
chrisdembia: I like that idea but it could also prove verbose, especially since `C` could be `Millard2012EquilibriumMuscle`. Although it is good to minimize redundancy, I would be okay with an overload; one that takes a string and converts to a `ComponentPath` internally, and one that takes a `ComponentPath`.

OpenSim/Common/Component.h
chrisdembia: 👍

jenhicks:
@tkuchida and @chrisdembia Do you have a vote on the question in Ajay's comment?

tkuchida:
> Do you have a vote on the question in Ajay's comment?

My understanding is that a *subcomponent* is defined as a Component that is owned by another Component; if a Component is deleted, then so too are its subcomponents. According to this definition, the "get" method in question is dealing strictly with subcomponents, so I think `getSubcomponentNameAtLevel` is more precise. Nevertheless, Ajay wrote the dictionary for the new Component API, so I'll leave it up to him.

aseth1:
> so I think getSubcomponentNameAtLevel is more precise. 

@tkuchida I am happy with that. Thanks!

aseth1:
LGTM. Thanks @carmichaelong! 
@chrisdembia you had reviewed early on. Are you OK with latest changes?

OpenSim/Common/Component.cpp
```diff
@@ -615,86 +615,27 @@ void Component::setParent(const Component& parent)
 
 std::string Component::getFullPathName() const
```
chrisdembia: @carmichaelong do you think it would be useful to have variants of these methods that give back a `ComponentPath`? In your experience of making these changes, can you think of a use case for such a variant? Just wondering. Perhaps it's best for us to leave out such variants for now, and add them in the future if they seem useful/necessary.

chrisdembia: Ah, for example, `getRelativePathName()` must convert the resulting string back into a `ComponentPath`, even though `getFullPathName()` had created a `ComponentPath` initially. So perhaps such a variant could be useful?

chrisdembia: Considering that you use the term "absolute" in the Path classes, what about renaming this function to `getAbsolutePathName()`?

aseth1: I like getting Full (or Absolute) and Relative Paths and `ComponentPath`s. I think we should transition to use `ComponentPath` instead of `std::string`, but we need to de/serialize ComponenPaths to do that effectively.  Keeping string variants with get<...>Name() may not be necessary if there is `ComponentPath::toString()`. I think we can discuss that as a subsequent improvement.

aseth1: Seems reasonable to switch to `getAbsolutePathName()` (better now than later!) but make sure to change it everywhere (including comments) where we use/describe a Component's "full" path name.

OpenSim/Common/Path.h
chrisdembia: Ah good point. This is fine as-is.

OpenSim/Common/Path.h
```diff
@@ -0,0 +1,162 @@
...
+    std::string getString() const;
+
+    /// Return true if this Path is an absolute path.
+    bool isAbsolute() { return _isAbsolute; };
+
+    /// Return the number of elements in the Path
+    size_t getPathLength() const { return _path.size(); };
+
+    /// Push a string to the back of a path (i.e. the end). First checks if the
+    /// pathElement is legal. Does not change if path is absolute or relative.
```
chrisdembia: `legal` -> `valid`

Consider: `Does not change if path is absolute or relative.` -> `This function does not alter whether this path is absolute or relative.`

OpenSim/Common/Path.h
```diff
@@ -0,0 +1,162 @@
...
+
+    /// Write out the path to a string with each element separated by the
+    /// specified separator.
+    std::string getString() const;
+
+    /// Return true if this Path is an absolute path.
+    bool isAbsolute() { return _isAbsolute; };
+
+    /// Return the number of elements in the Path
+    size_t getPathLength() const { return _path.size(); };
```
chrisdembia: I slightly prefer` getNumElements()` or `getNumLevels()`. At first, I thought "path length" could be `getString().size()`.

OpenSim/Common/Path.cpp
```diff
@@ -0,0 +1,223 @@
...
+    if (_isAbsolute) {
+        return _path;
+    }
+    
+    if (!otherPath->_isAbsolute) {
+        OPENSIM_THROW(Exception, "getOatherPathVec: otherPath must be an absolute path");
+    }
+
+    std::vector<std::string> pathVec;
+    for (auto pathElement : otherPath->_path) {
```
chrisdembia: `auto pathElement` -> `const auto& pathElement`. And similarly for other places where this comes up.

OpenSim/Common/Path.cpp
```diff
@@ -0,0 +1,223 @@
...
+}
+
+std::vector<std::string> Path::formAbsolutePathVec(Path* otherPath) 
+{
+    if (_isAbsolute) {
+        return _path;
+    }
+    
+    if (!otherPath->_isAbsolute) {
+        OPENSIM_THROW(Exception, "getOatherPathVec: otherPath must be an absolute path");
```
chrisdembia: typo `getOatherPathVec`

OpenSim/Common/Path.h
```diff
@@ -0,0 +1,162 @@
...
+    /// Pure virtual function that returns a char for the designated separator.
+    virtual const char getSeparator() const = 0;
+
+    /// Pure virtual function that returns a string of invalid characters.
+    virtual const std::string getInvalidChars() const = 0;
+
+protected:
+    /// Get an absolute path by resolving it relative to a given otherPath.
+    /// If the current Path is already absolute, return the same Path.
+    std::vector<std::string> formAbsolutePathVec(Path* otherPath);
```
chrisdembia: I think the parameter should be `const Path& otherPath`. Also, this function should be const. Similar for `formRelativePathVec`.

OpenSim/Common/Path.h
```diff
@@ -0,0 +1,162 @@
...
+* @author Carmichael Ong
+*/
+class OSIMCOMMON_API Path {
+
+public:
+    /// Default constructor
+    Path() = default;
+
+    /// Construct Path from a string, given separator character and a string
+    /// of invalidChars. Performs a cleanPath() at the end of construction.
```
chrisdembia: Since `cleanPath()` is private, users may not be able to see it (e.g., in user doxygen). Perhaps replace that with something like "Cleans up the path by removing blah blah."

OpenSim/Common/Path.h
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /// Return true if pathElement does not contain any chars from the list
+    /// of _invalidChars
+    bool isLegalPathElement(const std::string& pathElement);
+
+    // Path variables
+    std::vector<std::string> _path;
+    char _separator;
+    std::string _invalidChars;
```
chrisdembia: This could be `const std::string` I think. Not sure if the language will allow it though...

OpenSim/Common/Path.h
```diff
@@ -0,0 +1,162 @@
...
+    void appendPathElement(const std::string& pathElement);
+
+    /// Cleans up a path. This includes removing "." and resolving ".." if
+    /// possible (i.e. it will not remove leading ".." but otherwise will
+    /// remove the previous pathElement from _path. This method also checks
+    /// if a path is invalid due to an absolute path starting with "..".
+    void cleanPath();
+
+    /// Return the sub-path, on the range [first, last).
+    std::vector<std::string> getSubPathVec(size_t first, size_t last)
```
chrisdembia: This function doesn't edit the member variables, and should thus be marked const.

OpenSim/Common/Path.h
```diff
@@ -0,0 +1,162 @@
...
+    {
+        return getSubPathVec(0, getPathLength() - 1);
+    }
+
+    /// Return the pathElement from the specified position as a string. This
+    /// does not do any checks on the bounds.
+    std::string getPathElement(size_t pos) { return _path[pos]; };
+
+    /// Return the last pathElement as a string.
+    std::string getPathName() { return _path[getPathLength() - 1]; };
```
chrisdembia: const function.

OpenSim/Common/Path.h
```diff
@@ -0,0 +1,162 @@
...
+    /// Return the the sub-path that contains all pathElements except for
+    /// the last one.
+    std::vector<std::string> getParentPathVec()
+    {
+        return getSubPathVec(0, getPathLength() - 1);
+    }
+
+    /// Return the pathElement from the specified position as a string. This
+    /// does not do any checks on the bounds.
+    std::string getPathElement(size_t pos) { return _path[pos]; };
```
chrisdembia: const function.

OpenSim/Common/Path.h
```diff
@@ -0,0 +1,162 @@
...
+    std::vector<std::string> formAbsolutePathVec(Path* otherPath);
+
+    /// Find the relative Path between this Path and another Path (otherPath)
+    /// (i.e. the Path to go FROM otherPath TO this Path). Both Paths must be 
+    /// absolute.
+    std::vector<std::string> formRelativePathVec(Path* otherPath);
+
+    /// Return the the sub-path that contains all pathElements except for
+    /// the last one.
+    std::vector<std::string> getParentPathVec()
```
chrisdembia: const

OpenSim/Common/Path.h
```diff
@@ -0,0 +1,162 @@
...
+
+    /// Destructor.
+    ~Path() = default;
+
+    /// Write out the path to a string with each element separated by the
+    /// specified separator.
+    std::string getString() const;
+
+    /// Return true if this Path is an absolute path.
+    bool isAbsolute() { return _isAbsolute; };
```
chrisdembia: const

OpenSim/Common/Path.cpp
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        // otherwise, move on
+        else {
+            ++i;
+        }
+    }
+
+    if (!_path.empty() && _path[0] == ".." && _isAbsolute) {
+        OPENSIM_THROW(Exception, "absolute path cannot start with ''..''");
```
chrisdembia: Most of OpenSim's exception messages use proper-ish sentences. Consider starting exception messages with a capital letter.

OpenSim/Common/Path.h
chrisdembia: Consider adding "Assumes the path is not an absolute path."

OpenSim/Common/Path.cpp
```diff
@@ -0,0 +1,223 @@
...
+        pathVec.push_back(pathElement);
+    }
+
+    return pathVec;
+}
+
+std::vector<std::string> Path::formRelativePathVec(Path* otherPath) 
+{
+    if (!this->_isAbsolute || !otherPath->_isAbsolute) {
+        OPENSIM_THROW(Exception, "getRelativePathVec: both paths must be absolute to find relative path");
```
chrisdembia: Violates 80-column limit.

OpenSim/Common/ComponentPath.h
```diff
@@ -0,0 +1,106 @@
...
+        return this->getString() != other.getString();
+    }
+
+    // Override virtual functions
+    const char getSeparator() const override { return separator; };
+    const std::string getInvalidChars() const override { return invalidChars; };
+
+    /// Get an absolute path by resolving it relative to a given otherPath.
+    /// If the current Path is already absolute, return the same Path.
+    ComponentPath formAbsolutePath(ComponentPath* otherPath);
```
chrisdembia: Should take `const ComponentPath& otherPath`, and should also be a const function.

OpenSim/Common/ComponentPath.h
```diff
@@ -0,0 +1,106 @@
...
+    /// Default Constructor
+    ComponentPath() = default;
+    
+    /// Construct a ComponentPath from a string. This will clean up the
+    /// path, removing and resolving "." and ".." when possible.
+    ComponentPath(const std::string path);
+
+    /// Constructor a ComponentPath from a vector that contains all subtree
+    /// node names and a bool that indicates if the path is an absolute path.
+    ComponentPath(std::vector<std::string> pathVec, bool isAbsolute);
```
chrisdembia: Should take a `const std::vector<std::string>& pathVec` (missing the reference).

OpenSim/Common/ComponentPath.h
```diff
@@ -0,0 +1,106 @@
...
+class OSIMCOMMON_API ComponentPath : public Path {
+
+public:
+    // Constructors
+    /// Default Constructor
+    ComponentPath() = default;
+    
+    /// Construct a ComponentPath from a string. This will clean up the
+    /// path, removing and resolving "." and ".." when possible.
+    ComponentPath(const std::string path);
```
chrisdembia: Should take a `const std::string&`

OpenSim/Common/ComponentPath.h
```diff
@@ -0,0 +1,106 @@
...
+    /// (i.e. the Path to go FROM otherPath TO this Path). Both Paths must be 
+    /// absolute.
+    ComponentPath formRelativePath(ComponentPath* otherPath);
+
+    /// Return the sub-path that contains all subdirectory levels 
+    /// except for the last one.
+    ComponentPath getParentPath();
+
+    /// Return the parent path as a string.
+    std::string getParentPathStr();
```
chrisdembia: Personal preference: `getParentPathString()`.

chrisdembia:
I apologize for taking so long to review this. I really wanted to be able to delve into it, so I had been waiting for a long block of time.

OpenSim/Common/Component.cpp
carmichaelong: There is a `getString()` method in place already. It was named this way to avoid the `toString()` since this originally derived from `Object`. I'm tempted to change it to `toString()` now that it does not derive from anything. @aseth1 @chrisdembia thoughts?

aseth1: I have no objections with `getString()` either.

chrisdembia: I am fine with either `getString()` or `toString()`. Perhaps `toString()` makes a little more sense, since `get<...>()` is usually used for accessing member variables, whereas `toString()` indicates that a "type conversion" is happening.

To clarify: @aseth1, you are saying to *not* add `ComponentPath ComponentPath::getRelativePathName()` for this PR, correct? I think that is fine. We can add them later.

aseth1: > Perhaps toString() makes a little more sense

:+1: 

> you are saying to not add `ComponentPath::getRelativePathName()`

Yep! I agree about adding later if necessary.

OpenSim/Common/Component.h
```diff
@@ -1973,45 +1974,51 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
                 break;
             } // otherwise, we just have a type and name match
               // which we may need to support for compatibility with older models
               // where only names were used (not path or type)
               // TODO replace with an exception -aseth
-            else if (comp.getName() == subname) {
+            else if (compName == subname) {
                 foundCs.push_back(&comp);
                 // TODO Revisit why the exact match isn't found when
                 // when what appears to be the complete path.
```
carmichaelong: I don't think I know enough about this issueto remove the TODO.

OpenSim/Common/Path.h
carmichaelong: Well the IDE allowed it, but apparently it quietly deletes the move constructor which causes issues when compiling...

chrisdembia:
@carmichaelong let me know when this is ready for review again.

carmichaelong:
@chrisdembia This should be ready for another review. The test failure looks to be a timeout of testContactGeometry in Debug mode.

OpenSim/Common/ComponentPath.h
```diff
@@ -51,21 +51,21 @@ class OSIMCOMMON_API ComponentPath : public Path {
     
     /// Construct a ComponentPath from a string. This will clean up the
     /// path, removing and resolving "." and ".." when possible.
-    ComponentPath(const std::string path);
+    ComponentPath(const std::string& path);
 
     /// Constructor a ComponentPath from a vector that contains all subtree
     /// node names and a bool that indicates if the path is an absolute path.
-    ComponentPath(std::vector<std::string> pathVec, bool isAbsolute);
+    ComponentPath(std::vector<std::string>& pathVec, bool isAbsolute);
```
chrisdembia: Sorry I neglected to say this in my previous review, but this should also be const (`const std::vector<std::string>&`).

chrisdembia:
Looks good to me.

carmichaelong:
Looks like random failure with trying to install swig for both of the build fails. @aseth1 there were some changes done since you reviewed last as suggested by @chrisdembia. Let me know if you want to look at it again or if you're okay with merging.

OpenSim/Common/Component.cpp
```diff
@@ -708,6 +649,8 @@ const Component::StateVariable* Component::
     // top level component
     std::string::size_type back = name.rfind("/");
     std::string prefix = name.substr(0, back);
+    // this is a bug (or misuse?)! if the name does not have a "/", then
+    // name = prefix = varName. back+1 = 0 due to uint
```
aseth1: This was intentional (mis)use so that `varName` is `name` if there are no `/` in the `name`. The effect when `back = string::max_size` (not found) is that of `name.length() - back = name.length() +1 `. I stumbled upon on that fact and left it.

aseth1:
LGTM @carmichaelong. Thanks for the changes and cleanup! I addressed a comment you made in the code concerning a bug or misuse of size_t. If you wish to clarify the comment (and [ci skip]) that would be great.

carmichaelong:
@aseth1 I updated the comments in the function. It might be more verbose than needed (it explains the unsigned int math), so let me know if you want it to be limited to more "functional" comments only.

chrisdembia:
Yay!

Issue #1194, opened 2016-08-18
Limit and reduce use of OpenSim::Array


aseth1:
In addition to OpenSim::Set there is wide use of OpenSim:;Array. We should eventually move to exclusive use of `std::` containers so that we don't have to maintain unnecessary code and run into unexpected bugs (e.g. #1119) from less than ideal implementations.

Also came up in #1146 where our test utilities use OpenSim::Array for containing tolerances:
>  OpenSim::Arrays are used to define the tolerances as expected by CHECK_STORAGE_AGAINST_STANDARD() and updating that method would impact numerous test cases. Making that update perhaps should wait until we are closer to eliminating OpenSim::Array altogether

carmichaelong:
@aseth1 I looked at `CHECK_STORAGE_AGAINST_STANDARD()` and tests using this, and it doesn't look too time consuming to change it to use `std::vector` instead. If you think that we should just do that (instead of backing `OpenSim::Array` with `std::vector`) I can work on it.

chrisdembia:
@klshrinidhi has worked on replacing `OpenSim::Array` with `std::vector`. He may have input here.

klshrinidhi:
I have a local branch on this. It is a pretty time consuming task to replace OpenSim::Array with std::vector completely. I stopped working on it several weeks ago and my branch now is severely out of sync with master. You can go ahead if you have the time. I am probably not going to do this anytime soon.

carmichaelong:
> It is a pretty time consuming task to replace OpenSim::Array with std::vector completely.

I agree that to **completely** do it would take quite awhile, and shouldn't be done all in one PR.

It seemed to me that for this case specifically, changing `CHECK_STORAGE_AGAINST_STANDARD` is a low hanging fruit that would also be efficient in preventing further use of OpenSim::Array. I think I'll go ahead and work on this specific case then.

jenhicks:
Back OpenSim Array with std::vec

PR #1217, opened 2016-08-30
use std::vector with CHECK_STORAGE_AGAINST_STANDARD

carmichaelong:
This PR starts to address #1194 by accomplishing two things:
1) CHECK_STORAGE_AGAINST_STANDARD now expects tolerances to be passed in by `std::vector` (instead of `OpenSim::Array`)
2) Changes `Storage::compareWithStandard()` to use `std::vector` internally.

### DIFF
--- a/Applications/CMC/test/testCMCArm26_Millard.cpp
+++ b/Applications/CMC/test/testCMCArm26_Millard.cpp
@@ -45,7 +45,7 @@ void testCMCArm26() {
     Storage *standard = new Storage();
     cmc.getModel().formStateStorage(temp, *standard);
 
-    std::vector<double> rms_tols(2*2+2*6, 0.02); // activations within 2%, angles within .6 degrees
+    std::vector<double> rms_tols(2*2+2*6, 0.02);
     const string& muscleType = cmc.getModel().getMuscles()[0].getConcreteClassName();
     string base = "testCMCArm26 "+ muscleType;
 
--- a/Applications/CMC/test/testCMCArm26_Millard.cpp
+++ b/Applications/CMC/test/testCMCArm26_Millard.cpp
@@ -45,7 +45,7 @@ void testCMCArm26() {
     Storage *standard = new Storage();
     cmc.getModel().formStateStorage(temp, *standard);
 
-    Array<double> rms_tols(0.02, 2*2+2*6); // activations within 2%, angles within .6 degrees
+    std::vector<double> rms_tols(2*2+2*6, 0.02); // activations within 2%, angles within .6 degrees
     const string& muscleType = cmc.getModel().getMuscles()[0].getConcreteClassName();
     string base = "testCMCArm26 "+ muscleType;
 
--- a/Applications/CMC/test/testCMCArm26_Thelen.cpp
+++ b/Applications/CMC/test/testCMCArm26_Thelen.cpp
@@ -41,7 +41,7 @@ void testCMCArm26() {
     Storage *standard = new Storage();
     cmc.getModel().formStateStorage(temp, *standard);
 
-    std::vector<double> rms_tols(2*2+2*6, 0.02); // activations within 2%, angles within .6 degrees
+    std::vector<double> rms_tols(2*2+2*6, 0.02);
     const string& muscleType = cmc.getModel().getMuscles()[0].getConcreteClassName();
     string base = "testCMCArm26 "+ muscleType;
 
--- a/Applications/CMC/test/testCMCArm26_Thelen.cpp
+++ b/Applications/CMC/test/testCMCArm26_Thelen.cpp
@@ -41,7 +41,7 @@ void testCMCArm26() {
     Storage *standard = new Storage();
     cmc.getModel().formStateStorage(temp, *standard);
 
-    Array<double> rms_tols(0.02, 2*2+2*6); // activations within 2%, angles within .6 degrees
+    std::vector<double> rms_tols(2*2+2*6, 0.02); // activations within 2%, angles within .6 degrees
     const string& muscleType = cmc.getModel().getMuscles()[0].getConcreteClassName();
     string base = "testCMCArm26 "+ muscleType;
 
--- a/Applications/CMC/test/testCMCWithControlConstraintsGait2354.cpp
+++ b/Applications/CMC/test/testCMCWithControlConstraintsGait2354.cpp
@@ -82,7 +82,7 @@ void testGait2354() {
     for (int i = 23; i < 46; ++i){
         rms_tols2[i] = 0.75; // velocities
     }
-    for (int i = 46; i < rms_tols2.size(); ++i){
+    for (size_t i = 46; i < rms_tols2.size(); ++i){
         rms_tols2[i] = 0.15; // muscle activations and fiber-lengths
     }
 
--- a/Applications/CMC/test/testCMCWithControlConstraintsGait2354.cpp
+++ b/Applications/CMC/test/testCMCWithControlConstraintsGait2354.cpp
@@ -66,26 +66,28 @@ void testGait2354() {
     int nq = results.getColumnLabels().getSize()-1;
 
     // Tracking kinematics angles in degrees should be within 2 degrees
-    Array<double> rms_tols(2.0, nq);
+    std::vector<double> rms_tols(nq, 2.0);
     rms_tols[3] = 0.005; // pelvis translations in m should be with 5mm
     rms_tols[4] = 0.005;
     rms_tols[5] = 0.005;
 
-    CHECK_STORAGE_AGAINST_STANDARD(results, standard, rms_tols, __FILE__, __LINE__, "testGait2354 tracking failed");
+    CHECK_STORAGE_AGAINST_STANDARD(results, standard, rms_tols, 
+        __FILE__, __LINE__, "testGait2354 tracking failed");
 
     Storage results2("subject01_ResultsCMC/subject01_walk1_states.sto");
     Storage standard2("std_subject01_walk1_states.sto");
 
     Array<string> col_labels = standard2.getColumnLabels();
-    Array<double> rms_tols2(0.1, col_labels.getSize()-1);
+    std::vector<double> rms_tols2(col_labels.getSize()-1, 0.1);
     for (int i = 23; i < 46; ++i){
         rms_tols2[i] = 0.75; // velocities
     }
-    for (int i = 46; i < rms_tols2.getSize(); ++i){
+    for (int i = 46; i < rms_tols2.size(); ++i){
         rms_tols2[i] = 0.15; // muscle activations and fiber-lengths
     }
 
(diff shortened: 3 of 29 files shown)


chrisdembia:
LGTM. +1 for indenting all those lines of code!

chrisdembia:
I think this PR needs one more reviewer.

Applications/CMC/test/testCMCArm26_Millard.cpp
```diff
@@ -45,7 +45,7 @@ void testCMCArm26() {
     Storage *standard = new Storage();
     cmc.getModel().formStateStorage(temp, *standard);
 
-    Array<double> rms_tols(0.02, 2*2+2*6); // activations within 2%, angles within .6 degrees
+    std::vector<double> rms_tols(2*2+2*6, 0.02); // activations within 2%, angles within .6 degrees
```
tkuchida: I know you didn't put it there, but the trailing comment is obsolete and should be deleted.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
```diff
@@ -79,16 +79,16 @@ inline void ASSERT(bool cond,
  */
 void CHECK_STORAGE_AGAINST_STANDARD(OpenSim::Storage& result, 
                                     OpenSim::Storage& standard, 
-                                    OpenSim::Array<double> tolerances, 
+                                    std::vector<double> tolerances, 
                                     std::string testFile, 
                                     int testFileLine, 
                                     std::string errorMessage)
 {
```
klshrinidhi: I think the parameters need to be received as `const T&`. I understand this PR is not introducing the problem but good to fix it if possible.

Applications/CMC/test/testCMCArm26_Thelen.cpp
```diff
@@ -41,7 +41,7 @@ void testCMCArm26() {
     Storage *standard = new Storage();
     cmc.getModel().formStateStorage(temp, *standard);
 
-    Array<double> rms_tols(0.02, 2*2+2*6); // activations within 2%, angles within .6 degrees
+    std::vector<double> rms_tols(2*2+2*6, 0.02); // activations within 2%, angles within .6 degrees
```
tkuchida: Please delete trailing comment or update `.6` to `1.5`.

OpenSim/Common/Storage.cpp
```diff
@@ -3328,23 +3328,19 @@ double Storage::compareColumnRMS(Storage& aOtherStorage, const std::string& aCol
  * errors for columns occurring in both storage objects, and record the
  * values and column names in the comparisons and columnsUsed Arrays.
  */
-void Storage::compareWithStandard(Storage& standard, Array<string> &columnsUsed, Array<double> &comparisons)
+void Storage::compareWithStandard(Storage& standard, std::vector<string> &columnsUsed, std::vector<double> &comparisons)
 {
```
klshrinidhi: Same comment here. Receive arguments as `const T&`.

OpenSim/Common/Storage.cpp
```diff
@@ -3328,23 +3328,19 @@ double Storage::compareColumnRMS(Storage& aOtherStorage, const std::string& aCol
...
     int maxColumns = _columnLabels.getSize();
-    columnsUsed.ensureCapacity(maxColumns);
-    comparisons.ensureCapacity(maxColumns);
+    columnsUsed.reserve(maxColumns);
+    comparisons.reserve(maxColumns);
```
klshrinidhi: No need to reserve space. This kind of thing is only helps with huge sized arrays of the order of 1000s or elements.

OpenSim/Common/Storage.cpp
```diff
@@ -3328,23 +3328,19 @@ double Storage::compareColumnRMS(Storage& aOtherStorage, const std::string& aCol
...
-    int columns = 0;
     for (int i = 1; i < maxColumns; ++i) {
         double comparison = compareColumnRMS(standard, _columnLabels[i]);
         if (!SimTK::isNaN(comparison)) {
-            comparisons[columns] = comparison;
-            columnsUsed[columns++] = _columnLabels[i];
+            comparisons.push_back(comparison);
+            columnsUsed.push_back(_columnLabels[i]);
         }
     }
```
klshrinidhi: This loop could use iterators instead.

OpenSim/Common/Storage.h
```diff
@@ -303,7 +303,7 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Storage, StorageInterface);
                          const std::string& aColumnName,
                          double startTime=SimTK::NaN, double endTime=SimTK::NaN);
     //void checkAgainstStandard(Storage standard, Array<double> &tolerances, std::string testFile = "", int testFileLine = -1, std::string errorMessage = "Exception");
-    void compareWithStandard(Storage& standard, Array<std::string> &columnsUsed, Array<double> &comparisons);
+    void compareWithStandard(Storage& standard, std::vector<std::string> &columnsUsed, std::vector<double> &comparisons);
     /** Force column labels for a Storage object to become unique. This is done
```
klshrinidhi: Could receive arguments as `const T&`.

klshrinidhi:
:+1:

Applications/CMC/test/testCMCWithControlConstraintsGait2354.cpp
```diff
@@ -66,26 +66,28 @@ void testGait2354() {
...
     Storage standard2("std_subject01_walk1_states.sto");
 
     Array<string> col_labels = standard2.getColumnLabels();
-    Array<double> rms_tols2(0.1, col_labels.getSize()-1);
+    std::vector<double> rms_tols2(col_labels.getSize()-1, 0.1);
     for (int i = 23; i < 46; ++i){
         rms_tols2[i] = 0.75; // velocities
     }
-    for (int i = 46; i < rms_tols2.getSize(); ++i){
+    for (int i = 46; i < rms_tols2.size(); ++i){
```
tkuchida: `size()` returns an unsigned so I believe some compilers will give a warning when attempting to compare `i` with `rms_tols2.size()`. Not a big deal (we have lots of warnings), but perhaps `i < (int)rms_tols2.size()`?

OpenSim/Common/Storage.cpp
```diff
@@ -3328,23 +3328,19 @@ double Storage::compareColumnRMS(Storage& aOtherStorage, const std::string& aCol
...
+    comparisons.reserve(maxColumns);
 
-    int columns = 0;
     for (int i = 1; i < maxColumns; ++i) {
         double comparison = compareColumnRMS(standard, _columnLabels[i]);
         if (!SimTK::isNaN(comparison)) {
-            comparisons[columns] = comparison;
-            columnsUsed[columns++] = _columnLabels[i];
+            comparisons.push_back(comparison);
+            columnsUsed.push_back(_columnLabels[i]);
```
tkuchida: Much clearer! :eyeglasses:

tkuchida:
`</review>` :+1: Minor suggestions only; can be ignored.

tkuchida:
Oops, sorry I didn't notice @klshrinidhi was already reviewing :see_no_evil:

Applications/CMC/test/testCMCWithControlConstraintsGait2354.cpp
carmichaelong: You're right, this should lead to a warning. My instinct would be to change it to `for (size_t i = 46; i < rms_tols2.size(); ++i)` to avoid casting. Thoughts on which is better?

tkuchida: > Thoughts on which is better?

I would make `i` a `size_t` in my own code, but OpenSim tends not to do this (in fact, there are methods like `numCoordinates()` that return ints when they should really return unsigned ints). There are reasons to avoid unsigned ints (see, e.g., #1155 and #1198), but in this particular case I don't see any problem with using `size_t`.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
carmichaelong: I'm not sure if this should be templatized. In the end, `tolerances` is compared to `comparisons`, which is necessarily a vector of `double` because of [this line](https://github.com/opensim-org/opensim-core/blob/4c416d4d000e4cfcbbe59b64f40c6f4140359076/OpenSim/Common/Storage.cpp#L3338). I could easily be missing something though?

chrisdembia: @klshrinidhi is not suggesting to make this templated. He is just using `T` as a placeholder for `std::vector<double>`, `std::string`, etc.

chrisdembia: @klshrinidhi is saying: pass by const reference instead of pass by value.

carmichaelong: That makes way more sense. Thanks!

carmichaelong:
Addressed comments in the last commit and tests are passing. Re: iterator for the for loop: the member variable looped over is an OpenSim::Array and not a std::vector. I think changing that member variable is out of scope for this PR.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
```diff
@@ -77,18 +77,18 @@ inline void ASSERT(bool cond,
...
-void CHECK_STORAGE_AGAINST_STANDARD(OpenSim::Storage& result, 
-                                    OpenSim::Storage& standard, 
-                                    OpenSim::Array<double> tolerances, 
-                                    std::string testFile, 
-                                    int testFileLine, 
-                                    std::string errorMessage)
+void CHECK_STORAGE_AGAINST_STANDARD(const OpenSim::Storage& result, 
+                                    const OpenSim::Storage& standard, 
+                                    const std::vector<double>& tolerances, 
+                                    const std::string testFile,
```
tkuchida: I think the two `const std::string`s should be `const std::string&`.

tkuchida:
`</re-review>` One minor comment; can be merged without it IMO. :+1:

OpenSim/Auxiliary/auxiliaryTestFunctions.h
carmichaelong: Ah, yes. Thanks, I got a little distracted by the `const` and forgot the references there. This also made me look at the `int` line, and apparently there's some thought that passing primitives by reference may not speed things up (one of many sources [here](http://stackoverflow.com/questions/3009543/passing-integers-as-constant-references-versus-copying))

tkuchida: > there's some thought that passing primitives by reference may not speed things up

:+1: This thought jibes with the convention used in Simbody and OpenSim.

carmichaelong:
Thanks @tkuchida @klshrinidhi @chrisdembia!

PR #1219, opened 2016-08-30
[DevProp] Improved interface for connecting Components

tkuchida:
Seeking feedback on proposal and any anticipated issues with bindings.

### DIFF
--- a/DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
+++ b/DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
@@ -1,74 +1,84 @@
-## Project: Improved interface for connecting Components
+## Project: Improved interface for connecting Outputs to Inputs
 ***OpenSim Development Proposal***
 
 ### Overview
-The interface for connecting Components currently looks something like the following:
+The interface for connecting Outputs to Inputs currently looks like the following:
 ```cpp
 // "update this Input to my ConsoleReporter and connect it to this Output from my Model"
 reporter->updInput("inputs").connect(hopper.getComponent(hopperHeightCoord).getOutput("value"), "height");
-
-// "update this Connector to my Component and connect it to this object"
-controller->updConnector("actuator").connect(*pathActuator);
 ```
 
 Some users may find this design difficult to learn for two reasons:
-- The verb "connect" appears in the middle of a potentially long line, not the beginning where it would be more visible.
+- The verb "connect" appears in the middle of a potentially long line.
 - The verb "connect" does not imply directionality, yet `connect()` can be called only in one direction: `consumer.connect(producer)`.
 
-This project proposes to change the interface for connecting Components while retaining all current functionality.
+This project proposes to change the interface for connecting Outputs to Inputs while retaining all current functionality.
 See [issue #1118](https://github.com/opensim-org/opensim-core/issues/1118).
 
 ### Requirements
 - No change to what can be connected or how the underlying connections are implemented.
 - No redundancy in the interface (i.e., the new interface replaces the existing interface).
 
-### Architecture
-**Changes to `Component`** -- 
-Create new methods for forming connections between Components.
+### Architecture and Interfaces
+There are several possible designs.
+
+**Design 1**
+
+Create convenience methods to simplify the code that precedes the word `connect`.
+Where applicable, make the existing `connect()` method private.
 ```cpp
-// 1. Plugging an AbstractOutput into an AbstractInput.
-//    -> calls Input::connect(const AbstractOutput& output, const std::string& annotation = "")
-static void connectToInput(AbstractInput& input, const AbstractOutput& output,
-                           const std::string& annotation = "");
+// Current design (for reference).
+const auto& output = hopper.getComponent(hopperHeightCoord).getOutput("value");
+reporter->updInput("inputs").connect(output, "height");
 
-// 2. Plugging an AbstractChannel into an AbstractInput.
-//    -> calls Input::connect(const AbstractChannel& channel, const std::string& annotation = "")
-static void connectToInput(AbstractInput& input, const AbstractChannel& channel,
-                           const std::string& annotation = "");
-
-// 3. Plugging an Object into a Connector.
-//    -> calls Connector::connect(const Object& object)
-static void connectToConnector(Connector& connector, const Object& object);
+// Alternative designs (mutually exclusive).
+reporter->connect("inputs", output, "height");  //of the form [register] = [expression]
+reporter->connect(output, "inputs", "height");  //mirrors left-to-right signal flow diagram
+// Specific to Reporters, which have only one Input.
+reporter->connectToInput(output, "height");     //hides the Reporter's Input
+reporter->satisfyInput(output, "height");       //avoids the overloaded verb "connect"
+reporter->addToReport(output, "height");        //might match user psychology
 ```
 
-**Changes to `Component`** -- 
-Rename the existing `void connect(Component& root)` method to `finalizeConnections()`.
+**Design 2**
 
-**Changes to `AbstractConnector`, `Connector`, `AbstractInput`, and `Input`** -- 
-Move the existing `connect()` methods from public to protected.
-Note that `Component` is already a friend of `AbstractConnector`.
-
-### Interfaces
-Example from hopper:
+Create a new static method on Component for connecting Outputs to Inputs.
 ```cpp
-// Before
-reporter->updInput("inputs").connect(hopper.getComponent(hopperHeightCoord).getOutput("value"), "height");
-controller->updConnector("actuator").connect(*pathActuator);
+// Current design (for reference).
+const auto& output = hopper.getComponent(hopperHeightCoord).getOutput("value");
+reporter->updInput("inputs").connect(output, "height");
 
-// After
-Component::connectToInput(reporter->updInput("inputs"),
-                          hopper.getComponent(hopperHeightCoord).getOutput("value"), "height");
-Component::connectToConnector(controller->updConnector("actuator"), *pathActuator);
+// Alternative designs (mutually exclusive).
+// Would rename the existing connect() method to, e.g., finalizeConnections().
+Component::connect(reporter->updInput("inputs"), output, "height");  //not both
+Component::connect(reporter->output, updInput("inputs"), "height");  //of these
+Component::connect(reporter->updInput(), output, "height");
+Component::connect(reporter->output, updInput(), "height");
+Component::satisfyInput(reporter->updInput(), output, "height");
+Component::satisfyInput(output, reporter->updInput(), "height");
+```
+
+**Design 3 (unable to wrap)**
+
+Overload `operator<<` or `operator>>`.
+```cpp
+// Current design (for reference).
+const auto& output = hopper.getComponent(hopperHeightCoord).getOutput("value");
+reporter->updInput("inputs").connect(output, "height");
+
+// Alternative designs.
+output >> reporter->updInput();  //would need a separate setAnnotation() method
+reporter->updInput() << output;
 ```
 
 ### Bindings
-No anticipated issues.
+No anticipated issues for Designs 1 or 2.
 
 ### Backwards Compatibility
 No backwards compatibility concerns because Components are new to 4.0.
 
 ### Lifecycle and Ownership / Memory Management
-No new objects; only adding static methods to an existing class.
+No new objects.
 
 ### Performance Considerations
 No expected performance issues because the same methods will be called under the hood.
@@ -80,4 +90,4 @@ The necessary updates to testComponentInterface.cpp will retain the current test
 No anticipated issues.
 
 ### Pull Requests
-A single PR should suffice.
+A single PR should be sufficient.
--- a/DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
+++ b/DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
@@ -2,41 +2,56 @@
 ***OpenSim Development Proposal***
 
 ### Overview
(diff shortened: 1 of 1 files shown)


DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Overview
+Connecting Components currently looks something like "update the Input of my ConsoleReporter and connect it to this Output from my Model".
+Some users may find it more intuitive to (i) begin with the verb *connect*, and (ii) swap the order of the Input and Output.
+This project proposes to change the user interface for connecting Components while retaining all current functionality.
+See [issue #1118](https://github.com/opensim-org/opensim-core/issues/1118).
+
+### Requirements
+- No change to what can be connected or how the underlying connections are formed.
+- No redundancy in the interface (i.e., the new interface replaces the existing interface).
```
chrisdembia: Do you propose removing (or hiding) the existing mechanism for connecting?

DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Interfaces
+Example from hopper:
+```cpp
+// before
+reporter->updInput("inputs").connect(device.getOutput(outputName));
+// after: either
+Component::connect(reporter->updInput("inputs"), device.getOutput(outputName));
+// or
+Component::connect(device.getOutput(outputName), reporter->updInput("inputs"));
```
chrisdembia: I think we should only allow one of the two.

DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
```diff
@@ -0,0 +1,62 @@
...
+  - `static void connect(const Object& object, Object& object)` and `static void connect(Object& object, const Object& object)`,
+    which will call [this method](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/ComponentConnector.h#L278).
+  - `static void connect(const AbstractOutput& output, AbstractInput& input, const std::string& annotation = "")` and
+    `static void connect(AbstractInput& input, const AbstractOutput& output, const std::string& annotation = "")`,
+    which will call [this method](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/ComponentConnector.h#L494).
+  - `static void connect(const AbstractChannel& channel, AbstractInput& input, const std::string& annotation = "")` and
+    `static void connect(AbstractInput& input, const AbstractChannel& channel, const std::string& annotation = "")`,
+    which will call [this method](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/ComponentConnector.h#L497).
+- Change the name of the existing `void connect(Component& root)` method to `finalizeConnections()`.
```
chrisdembia: What about connecting components? (i.e., Connectors)?

DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
```diff
@@ -0,0 +1,62 @@
...
+See [issue #1118](https://github.com/opensim-org/opensim-core/issues/1118).
+
+### Requirements
+- No change to what can be connected or how the underlying connections are formed.
+- No redundancy in the interface (i.e., the new interface replaces the existing interface).
+
+### Architecture
+Changes to `Component`:
+- Create a new method for forming connections between Components. There will be several variants:
+  - `static void connect(const Object& object, Object& object)` and `static void connect(Object& object, const Object& object)`,
```
chrisdembia: Components can require a connection to Objects, but Objects cannot require a connection to other Objects. So you would want `static void connect(const Object& object, Component& object)`.

DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
```diff
@@ -0,0 +1,62 @@
...
+// before
+reporter->updInput("inputs").connect(device.getOutput(outputName));
+// after: either
+Component::connect(reporter->updInput("inputs"), device.getOutput(outputName));
+// or
+Component::connect(device.getOutput(outputName), reporter->updInput("inputs"));
+```
+
+### Bindings
+I don't know. Are there issues wrapping static methods?
```
chrisdembia: IIRC I have used static functions in both python and MATLAB. They should be automatically wrapped without any special treatment.

DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
tkuchida: > I now see your message about this below.

Based on the template, my understanding is that we are supposed to list only the needs in the "Requirements" section, with the strategy for meeting these needs discussed in later sections.

DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
tkuchida: > I think we should only allow one of the two

That sounds okay in theory, but I think @aseth1 prefers the input first ("satisfy this input with this output") whereas I prefer the output first ("plug this output into this input"). It wasn't clear that we would be able to agree on one or the other.

DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
tkuchida: > Components can require a connection to Objects, but Objects cannot require a connection to other Objects.

Correct.

DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
chrisdembia: Yeah I think you're right. Oops.

DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
chrisdembia: Then we have redundancy, violating the requirements. I prefer the latter (output -> input).

aymanhab: Would it make it clearer if we call the methods connectInputToOutput & connectOutputToInput respectively?

DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
aymanhab: Sorry, but I think we have enough "finalize" methods, it feels quite diluted. If we want to call it topLevel, or root or world or any other connotation to indicate actual distinction from Component level that would be clearer to users I think.

DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
tkuchida: > Would it make it clearer if we call the methods connectInputToOutput & connectOutputToInput respectively?

That could work for Inputs and Outputs. What about Connectors like [these](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Sandbox/ExampleHopperDevice/defineDeviceAndController_answers.h#L184)?

aymanhab: Aha, good one. The input/output relation seems like a dataflow diagram. The example here is a different kind of connection (like a depends on) at least conceptually, does that warrant a different presentation/API?

DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
chrisdembia: I was unsure about `finalize` here but I think @tkuchida is trying to achieve something important. There are two distinct operations: (i) model-developers specifying what should be connected to what, and (ii) model-consumers telling their model to "find the things that the model-developer said you're supposed to be connected to". Right now, we use the word `connect()` for both of these operations, and that is confusing. I personally am fine with `finalizeConnections()` since it reduces the number of verbs in our interface, and the operation is similar to `finalizeFromProperties()`.

One option: remove this method altogether, and include "satisfying connections" as part of "finalizing from properties." This would be a larger project I think, though. We could either have a separate proposal for that, or have this proposal include multiple PRs.

DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
tkuchida: > The input/output relation seems like a dataflow diagram. The example here is a different kind of connection (like a depends on) at least conceptually, does that warrant a different presentation/API?

I think the distinction between Inputs and Connectors is somewhat artificial because both must be satisfied (just by different types). On a block diagram, Inputs and Connectors would both look like "inputs", so perhaps Connector should be renamed. Also, "input" implies directionality whereas "connector" does not, which could lead to further confusion. This proposal does not attempt to address this issue.

I've updated the proposal, replacing the single `connect()` method with separate `connectToInput()` and `connectToConnector()` methods.

DevelopmentProposals/componentConnectInterface/componentConnectInterface.md
tkuchida: > One option: remove this method altogether (perhaps make it internal/private), and include "satisfying connections" as part of "finalizing from properties." This would be a larger project I think, though. We could either have a separate proposal for that, or have this proposal cover multiple PRs.

Worth exploring, but perhaps as a separate DevProp for discussing user workflow and disciplined model-building "stages".

tkuchida:
Revised proposal ready for review.

jenhicks:
Thanks @tkuchida for tackling this and writing what I think is our first DevProp! I'm still struggling a little bit with understanding the interface when I look at the code. 

Here is what I might expect as a user (in psuedocode):

1) I want to report some values from a muscle or other component.
myReporter.setFields(Muscle.getOutputs("outputName"))

Not sure if "Fields" is the right word, but in the case of a reporter, input is confusing to me, because the "inputs" are actually the things that I want to output to the console or to a file.

2) I want to set up data flow between two Components (say a muscle and a myo-electric controller)
myController.setInput("muscleExcitation", Muscle.getOutput("name"))

"muscleExcitation" is the name of the input I've defined for my controller.

@tkuchida @aseth1 @chrisdembia 
Let me know what you think. If these suggestions are naive or impossible to implement, please feel free to say so! We can also discuss more tomorrow, if it is easier to talk in person.

tkuchida:
> Not sure if "Fields" is the right word, but in the case of a reporter, input is confusing to me, because the "inputs" are actually the things that I want to output to the console or to a file.

An Input is a "port" on a Component; it is part of the nomenclature of the Component Interface. Inputs have names, though, so we could assign the name "fields" to the Input of Reporters, and create a convenience method like `setFields()` that connects to the Reporter's Input.

> "muscleExcitation" is the name of the input I've defined for my controller.

Copied from my response to Ayman (above): "I think the distinction between Inputs and Connectors is somewhat artificial because both must be satisfied (just by different types). On a block diagram, Inputs and Connectors would both look like "inputs", so perhaps Connector should be renamed. Also, "input" implies directionality whereas "connector" does not, which could lead to further confusion. This proposal does not attempt to address this issue."

tkuchida:
Per conversation with @aseth1 and @chrisdembia, I will prepare 5 possible designs for further discussion.

aseth1:
> 1) I want to report some values from a muscle or other component.
myReporter.setFields(Muscle.getOutputs("outputName")

@jenhicks @tkuchida I agree that when looking at a Reporter in isolation it is weird to connect Inputs, because as you stated you expect the Reporter to report Outputs (which it does). But, if you look at a Reporter as any other Component, then it has Inputs which are any Outputs from your model you want to report. 

A convenience that did not require "fields" or "inputs" on a Reporter might work here since it only has one list Input. E.g:
`myReporter.connect(muscle.getOutput("name"));`

> 2) I want to set up data flow between two Components (say a muscle and a myo-electric controller)
myController.setInput("muscleExcitation", Muscle.getOutput("name"))

The existing method is somewhat similar. 
`myController.updInput("muscleExcitation").connect(muscle.getOutput("name"))`
which @tkuchida dislikes.

A hybrid:
`myController.connectInput("muscleExcitation", muscle.getOutput("name"))`

@tkuchida I don't think you were assigned the whole task of coming up with five options. @chrisdembia and I agreed to come up with a couple too (total 4-5 options seems good to discuss tomorrow)

tkuchida:
- Rename Connector to Socket
- Create `connectSocket_name_of_socket()` and `connectInput_name_of_input()` methods in existing macro

I will add these points to the proposal, along with the other points that were raised in the dev meeting.

tkuchida:
Updated based on discussion in dev team meeting. Please comment with your favorite ideas.

jenhicks:
Thanks Tom for the nice summary of our discussion last week. I still like Design 6. Let's check in today at the Dev Team meeting about any remaining thoughts and how we will accomplish implementation.

tkuchida:
> Let's check in today at the Dev Team meeting about any remaining thoughts and how we will accomplish implementation.

On hold. Prerequisites for this project include work on :one: serialization and :two: redesign of Component lists.

tkuchida:
Summarized the four main changes the Dev Team seems to have agreed on (listed immediately beneath the "Architecture and Interfaces" heading). These four changes could be implemented in separate PRs.

aseth1:
@klshrinidhi a nice next project would be to to implement the first two steps of this proposal:

The current proposal involves four main changes:

1. Rename Connector to Socket; avoids the awkward-sounding action of connecting connectors.
2. Rename existing Component::connect() method to finalizeConnections(). Aligns with finalizeFromProperties() and avoids confusion Socket (Connector) and Input connect() methods.

I am have to chat about the task in person 😄

klshrinidhi:
Sounds good.

aseth1:
> Sounds good.

Let me know if you need any background or help deciphering anything. Thanks!

PR #1222, opened 2016-08-31
Add non-const accessors for Coordinates and disseminate

tkuchida:
## Summary
- Added `updCoordinate()` methods and corresponding test cases (analogous to the `getCoordinate()` methods added in #1116).
- Replaced existing accessors with the new enum-argumented ones to improve code readability.

Addresses last two checkboxes of #1066.

### DIFF
--- a/OpenSim/Simulation/SimbodyEngine/Joint.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/Joint.cpp
@@ -217,9 +217,8 @@ const Coordinate& Joint::getCoordinate() const {
                      JointHasNoCoordinates);
     OPENSIM_THROW_IF(numCoordinates() > 1,
                      InvalidCall,
-                     "Coordinate set has more than one coordinate. Use "
-                     "getCoordinate method defined in the concrete class "
-                     "instead.");
+                     "Joint has more than one coordinate. Use the getCoordinate "
+                     "method defined in the concrete class instead.");
 
     return get_coordinates(0);
 }
@@ -229,9 +228,8 @@ Coordinate& Joint::updCoordinate() {
                      JointHasNoCoordinates);
     OPENSIM_THROW_IF(numCoordinates() > 1,
                      InvalidCall,
-                     "Coordinate set has more than one coordinate. Use "
-                     "updCoordinate method defined in the concrete class "
-                     "instead.");
+                     "Joint has more than one coordinate. Use the updCoordinate "
+                     "method defined in the concrete class instead.");
 
     return upd_coordinates(0);
 }
--- a/OpenSim/Simulation/SimbodyEngine/Joint.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/Joint.cpp
@@ -224,6 +224,18 @@ const Coordinate& Joint::getCoordinate() const {
     return get_coordinates(0);
 }
 
+Coordinate& Joint::updCoordinate() {
+    OPENSIM_THROW_IF(numCoordinates() == 0,
+                     JointHasNoCoordinates);
+    OPENSIM_THROW_IF(numCoordinates() > 1,
+                     InvalidCall,
+                     "Coordinate set has more than one coordinate. Use "
+                     "updCoordinate method defined in the concrete class "
+                     "instead.");
+
+    return upd_coordinates(0);
+}
+
 Coordinate::MotionType Joint::getMotionType(CoordinateIndex cix) const
 {
     OPENSIM_THROW_IF(cix >= _motionTypes.size(), Exception,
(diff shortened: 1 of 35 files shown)


OpenSim/Simulation/SimbodyEngine/Joint.cpp
```diff
@@ -224,6 +224,18 @@ const Coordinate& Joint::getCoordinate() const {
...
 
+Coordinate& Joint::updCoordinate() {
+    OPENSIM_THROW_IF(numCoordinates() == 0,
+                     JointHasNoCoordinates);
+    OPENSIM_THROW_IF(numCoordinates() > 1,
+                     InvalidCall,
+                     "Coordinate set has more than one coordinate. Use "
+                     "updCoordinate method defined in the concrete class "
+                     "instead.");
+
```
klshrinidhi: Maybe good to say "*Joint* ~~Coordinates set~~ has more than one coordinate. ......".

klshrinidhi:
:+1: 

Following is just a thought. Feel free to ignore.
One improvement (or not) could be to make the class `Coord` available outside the Joints by:
```c++
namespace OpenSim {
    // Cannot name the alias 'FreeJoint' because it already exists.
    using FreeJointCoord = FreeJoint::Coord;
}
```
This way the enums can be used as follows:
```c++
blockToGround->updCoordinate(FreeJointCoord::Rotation1X).setRange(angleRange);
blockToGround->updCoordinate(FreeJointCoord::Rotation2Y).setRange(angleRange);
blockToGround->updCoordinate(FreeJointCoord::Rotation3Z).setRange(angleRange);
blockToGround->updCoordinate(FreeJointCoord::TranslationX).setRange(positionRange);
blockToGround->updCoordinate(FreeJointCoord::TranslationY).setRange(positionRange);
blockToGround->updCoordinate(FreeJointCoord::TranslationZ).setRange(positionRange);
```

OpenSim/Simulation/SimbodyEngine/Joint.cpp
tkuchida: > Maybe good to say "*Joint* ~~Coordinates set~~ has more than one coordinate. ......".

Done.

tkuchida:
> One improvement (or not) could be to make the class `Coord` available outside the Joints

Where would you like me to add the `using` statements? At the top of *Joint.h (after `namespace OpenSim`)?

chrisdembia:
IMO If you add them they should go in the header for the concrete joints. E.g., `FreeJointCoord` should be in FreeJoint.h.

chrisdembia:
I'm not sure that that adds much though; saves only two characters in C++. The most convincing reason to use them would be if they make the python/matlab interfaces better.

klshrinidhi:
They should be right after the class declarations in headers (BallJoint.h, EllipsoidJoint.h, ...).

tkuchida:
> The most convincing reason to use them would be if they make the python/matlab interfaces better.

Ok, I will build wrapping and investigate.

aseth1:
> The most convincing reason to use them would be if they make the python/matlab interfaces better.

How would they improve python/matlab interfaces?

I am OK with or without the change. If it makes scripting easier than great. Also could be done subsequently.

tkuchida:
> they should go in the header for the concrete joints. E.g., `FreeJointCoord` should be in FreeJoint.h.

That's what I meant by "*Joint.h". Sorry, I should have specified what the asterisk meant.

> How would they improve python/matlab interfaces?

According to [this documentation](http://www.swig.org/Doc3.0/CPlusPlus11.html#CPlusPlus11_strongly_typed_enumerations), the current implementation will result in Python code that looks like `myFreeJoint.getCoordinate(FreeJoint_Coord_TranslationX)` and @klshrinidhi's change will eliminate the first underscore: `myFreeJoint.getCoordinate(FreeJointCoord_TranslationX)`.

tkuchida:
> They should be right after the class declarations in headers (BallJoint.h, EllipsoidJoint.h, ...).

I thought there might be a clever way of putting them at the top so they're more visible, but it seems like I would need to forward declare the enum class in that case, [which isn't possible](http://stackoverflow.com/questions/27019292/is-in-class-enum-forward-declaration-possible). If we want the enum classes to be within the OpenSim namespace so we don't need the first two colons, should I move the enum class declarations outside the joint class declarations?

klshrinidhi:
I am not sure if the type aliases are picked up the SWIG and exposed like you say. My guess is nothing will change with respect to wrapping. The aliases are more of a convenience to C++ users. 

> I thought there might be a clever way of putting them at the top so they're more visible, but it seems like I would need to forward declare the enum class in that case, which isn't possible. If we want the enum classes to be within the OpenSim namespace so we don't need the first two colons, should I move the enum class declarations outside the joint class declarations?

If the enum classes are moved out into OpenSim namespace, then extra pair of colons will go away for sure, just like it would go away if type aliases are used. But moving the enums outside may have implications to wrapping. It will require to change some code inside the classes. Maybe drop the idea for now. Go ahead and merge this. It looks good as is.

aseth1:
> It looks good as is.

I agree. I appreciate the effort to provide convenience, but I am happy for it to be a separate PR. 
@tkuchida feel free to merge as is. I would do it now, but I am not sure if you are in the process of making more commits.

tkuchida:
Ok, thanks @klshrinidhi and @aseth1. I'll merge this PR and create an issue to investigate how the wrapping can be improved.

PR #1223, opened 2016-08-31
Remove clutter in AppVeyor/Travis output

tkuchida:
## Summary
- Suppressed "ModelVisualizer couldn't find file" warning messages.
- Suppressed compiler warning about signed/unsigned mismatch.
- Also replaced `clog` with `cout` in Geometry.cpp because we don't use `clog` in any disciplined way (`clog` doesn't appear *anywhere else* in OpenSim).


klshrinidhi:
:+1:

Issue #1221, opened 2016-08-30
TableReporter doesn't respect column label string


jimmyDunne:
TableReporter doesn't respect the string that is used for the column label. Using the ConsoleReporter, I can set the name of the input and it shows up in the console output:

```
>>reporter = ConsoleReporter();
>>reporter.set_report_time_interval(0.1);
>>reporter.updInput('inputs').connect(model.getCoordinateSet.get(0).getOutput('value'), 'pin1_angle' );
>>reporter.updInput('inputs').connect(model.getCoordinateSet.get(1).getOutput('value'), 'q2' );
>>model.addComponent(reporter);
...

time|   pin1_angle|          q2|
           0|  -0.78539816|  -1.0471975|
         0.1|  -0.73474627|   -1.027509|
         0.2|  -0.60741367| -0.92328623|
         0.3|  -0.47806726| -0.61092782|
         0.4|  -0.43760328| 0.037672947|
         0.5|  -0.29961138|  0.56629776|
         0.6|  0.087742457|  0.53901707|
         0.7|   0.61300325| 0.045838744|
         0.8|      1.05583| -0.42200138|
```
However, it I use a table reporter and try to get the column name, it uses the full path instead:

```
>> coordinateReporter = TableReporter()
>> coordinateReporter.set_report_time_interval(0.1)
>> coordinateReporter.updInput('inputs').connect(model.getCoordinateSet.get(0).getOutput('value'), 'pin_angle' );
>> coordinateReporter.updInput('inputs').connect(model.getCoordinateSet.get(1).getOutput('value'), 'q2' );
...
>> table = coordinateReporter.getReport()
>> table.getNumColumns
ans = 
2
>> table.getColumnLabel(0)
ans = 
/double_pendulum/pin1/q1/value
```

tkuchida:
Related to #1071

tkuchida:
Addressing this issue involves a design decision. Please provide discussion on the Development Proposal PR (#1226).

PR #1226, opened 2016-09-01
[DevProp] Improve column labels used by TableReporter

tkuchida:
Seeking feedback on proposal and possible designs.


tkuchida:
There are also considerations related to serialization; I will add these to the proposal.

tkuchida:
Added notes about serialization following discussion with @chrisdembia.

jimmyDunne:
I really like Design 4, especially:

```
+    _annotation.first  = Output.getName()
 +    _annotation.second = Output.getFullPathName()
```

jenhicks:
Thanks @tkuchida for summarizing the issues and proposing several solutions.

One question: What are the other uses (if any) for the annotations?

chrisdembia:
> One question: What are the other uses (if any) for the annotations?

I was pleasantly surprised to find that I wrote some [doxygen documentation](http://myosin.sourceforge.net/1244/classOpenSim_1_1AbstractInput.html#ac94317c46c0bbbb29b22421ed7e590f1) for them, though it's pretty brief:

> An Annotation is a description of a channel that is specific to how this input should use that channel.

> For example, the component containing this Input might expect the annotations to be the names of markers in the model. 

Say that an inverse kinematics component has an input for the experimental marker trajectories, and that this component expects the channels of the input to be labeled with the names of the markers in the model. But that some of the experimental markers do not come directly from data files, but instead from an intermediate component that does some calculation to compute a "fake" experimental marker (e.g., computing the knee joint center using lateral and medial knee markers). This intermediate component would be some very generic thing, like Summer, and so its output would have a generic name, like "sum". Now when passing an output named "sum" to the inverse kinematics component, one needs a way to "rename" or "relabel" that output (channel) to have the name of a marker in the model. That's what the annotation is.

tkuchida:
> One question: What are the other uses (if any) for the annotations?

We may end up introducing additional annotation-like properties (e.g., "legendLabel" used by a real-time plotter) so perhaps terminology like "{long|detailed}Label" and "{short|brief}Label" would be better?

Discussion of this PR is slated for Sep. 15.

chrisdembia:
> We may end up introducing additional annotation-like properties (e.g., "legendLabel" used by a real-time plotter) so perhaps terminology like "{long|detailed}Label" and "{short|brief}Label" would be better?

That sounds fine to me. That use-case is good, but it would be helpful to think of more use cases for a long label. Perhaps it could go in the metadata of a table. It's not clear to me if the long label belongs with the input or the output.

tkuchida:
Discussion postponed until Sep. 22.

chrisdembia:
Building on Tom's comment here about long/detailed and short/brief, how about this proposal:

```cpp
void Input::connect(const Output& output, std::string anno="") {
    ...connect...
    _anno = anno;
    _shortLabels[i] = anno.empty() ? output.getName() : anno;
    _longLabels[i] = anno.empty() ? output.getFullPathName() : anno;
}

// allow changing annotation after connecting (but not before).
void Input::setAnnotation(const Output& output, std::string anno) {
    ...find index of given output...
    ...throw exception if that output is not wired to this input...
    _anno = anno;
    _shortLabels[i] = anno.empty() ? output.getName() : anno;
    _longLabels[i] = anno.empty() ? output.getFullPathName() : anno;
}

// In ConsoleReporter:
columnLabel = input.getShortLabel();
// In TableReporter:
columnLabel = input.getLongLabel();
```

This avoids using logic later on when requesting the labels, and allows reporters to choose whether to use the short or long label (which is always the annotation if provided).

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
```diff
@@ -0,0 +1,148 @@
...
+Several designs would satisfy the requirements; four are proposed here.
+As a general note, it may be useful to add a `setAnnotation()` method (e.g., if a fully connected model was loaded from file).
+
+**Design 1**
+
+Store an empty string in `_annotations` if no annotation was provided when `Input::connect()` was called.
+When writing the column label, use the `_annotations` value if it is not empty; otherwise,
+the ConsoleReporter will use the Output's name and the TableReporter will use the Output's full path name.
+This design will change the current behavior.
+Currently, the name of the Output can differ from the column label if the name is changed after `Input::connect()` has been called.
```
aseth1: Changing names currently is a change in a Component's properties. That should invalidate all connections and require a connect() to be invoked.

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
```diff
@@ -0,0 +1,148 @@
...
+It seems natural to generate a default column label when reporting rather than when connecting,
+so there is presumably a good reason for wanting to generate a default when `Input::connect()` is called
+(though always using the Output's full path name in the TableReporter is inconsistent).
```
aseth1: > It seems natural to generate a default column label when reporting rather than when connecting

If we take a step back, the job of a Reporter is to report the values of the model Outputs it has plugged into it (its Inputs). The identity of those Outputs is established by connecting.
If you use attributes of the Output to provide reasonable defaults for labeling, it follows then that those defaults are also a direct consequence of connecting. 

Introducing more indirection between the Output and the label has he potential to raise inconsistencies. If we want user supplied names to override the Output's information, then we may want the TableReporter to maintain a map between the Outputs it is reporting on the the columns labels it uses to report those Outputs and have that serialized.

tkuchida:
### Summary of action items from discussion in Dev Team meeting:
1. Rename "annotation" to "alias".
2. If user provides optional second argument, store as alias; otherwise, alias is null.
3. Create `getShortLabel()` and `getLongLabel()` methods on Input. The former is called by ConsoleReporter and returns `isNull(alias) ? output_name : alias`; the latter is called by TableReporter and returns `isNull(alias) ? full_path_name : alias`.

### Additional action items:
1. Create `setAlias()` method so the user can set/reset the alias after (but not before) connecting.
2. Open a separate issue to consider moving alias to Output (would require substantial design change because Outputs are not currently serialized).
3. Update the DevProp to describe the selected design.

chrisdembia:
That summary looks good to me.

tkuchida:
Updated document with selected design. Merged master but "devprop" branch prefix did not prevent CI builds from firing.

tkuchida:
> Merged master but "devprop" branch prefix did not prevent CI builds from firing.

"exclude" should be "except". I'll test on this branch…

tkuchida:
The selected design was implemented in PR #1279.

chrisdembia:
We agreed on this devprop in the meeting, so I'm merging this.

aseth1:
Reopening branch to revise DevProp

tkuchida:
> Reopening branch to revise DevProp

I thought we had reached an agreement on the design in the Dev Team meeting on 9/22. I summarized the selected design in the DevProp. Revising the design after the implementation has been completed wastes a lot of time. Please let me know what process I was supposed to follow.

aseth1:
You didn't do anything wrong. I was slow in reviewing and commenting.

tkuchida:
> You didn't do anything wrong. I was slow in reviewing and commenting.

I spent my weekend writing code to meet a spec that has now changed. I recognize this doesn't affect the entire Dev Team, but I see it as a problem because it was my time that was wasted. It would have been helpful to alert me if there were still concerns with the design following the Dev Team meeting on 9/22, as I thought we had reached an agreement.

chrisdembia:
I suggest the following additions to the devprop procedure:

1. The devprop should be merged before working on the actual feature.
2. The reviewers of the devprop should be the same reviewers of the subsequent feature PRs. The devprop can list the reviewers.

tkuchida:
> I suggest the following additions to the devprop procedure:

Item 1 would not have prevented this issue because we all agreed on the design in the Dev Team meeting and anyone could have merged the devprop. The original issue (#1221) was opened 4 weeks ago and the devprop was written a few days later, so it seemed to me that everyone had sufficient time to consider the design.

PR #1228, opened 2016-09-01
Remove OpenSim::Array use from Component

carmichaelong:
Continue to address #1194. Removes use of and returning OpenSim::Array from the Component class, in favor of std::vector. With the heavy use of the Component class, this should limit further usage of OpenSim::Array.

### DIFF
--- a/Bindings/Java/OpenSimJNI/OpenSimContext.cpp
+++ b/Bindings/Java/OpenSimJNI/OpenSimContext.cpp
@@ -415,13 +415,14 @@ void OpenSimContext::restoreStateFromCachedModel()
     _model->initSystem();
     clonedModel->initSystem();
 
-    Array<std::string> modelVariableNames = _model->getStateVariableNames();
-    Array<std::string> clonedModelVariableNames = clonedModel->getStateVariableNames();
+    std::vector<std::string> modelVariableNames = _model->getStateVariableNames();
+    std::vector<std::string> clonedModelVariableNames = clonedModel->getStateVariableNames();
 
-    for(int i = 0; i < modelVariableNames.getSize(); i++)
+    for(auto const& name : modelVariableNames)
     {
-        std::string name = modelVariableNames.get(i);
-        if(clonedModelVariableNames.findIndex(name) >= 0)
+        std::vector<std::string>::iterator it;
+        it = std::find(clonedModelVariableNames.begin(), clonedModelVariableNames.end(), name);
+        if(it != clonedModelVariableNames.end())
         {
             double value = clonedModel->getStateVariableValue(clonedState, name);
             _model->setStateVariableValue(_model->updWorkingState(), name, value);
--- a/OpenSim/Examples/SymbolicExpressionReporter/SymbolicExpressionReporter.cpp
+++ b/OpenSim/Examples/SymbolicExpressionReporter/SymbolicExpressionReporter.cpp
@@ -222,11 +222,11 @@ record(const SimTK::State& s)
     // MAKE SURE ALL QUANTITIES ARE VALID
     _model->getMultibodySystem().realize(s, SimTK::Stage::Velocity );
     // Get state variable names and values in same order and use that to update map
-    Array<std::string> stateNames = _model->getStateVariableNames();
+    std::vector<std::string> stateNames = _model->getStateVariableNames();
     SimTK::Vector rStateValues = _model->getStateVariableValues(s);
 
     // update map between names and values, then pass it to expression evaluator
-    for(int i=0; i<stateNames.getSize(); i++){
+    for(int i=0; i<stateNames.size(); i++){
         _variables.find(stateNames[i])->second = rStateValues[i];
     }
     double value = Lepton::Parser::parse(_expressionStr).evaluate(_variables);
@@ -260,9 +260,9 @@ begin( SimTK::State& s)
     // RESET STORAGE
     _resultStore.reset(s.getTime());
     // Populate list of names, initial values from s
-    Array<std::string> stateNames = _model->getStateVariableNames();
-    for(int i=0; i< stateNames.getSize(); i++){
-        _variables.insert(pair<string, double>(stateNames[i], 0.0));
+    std::vector<std::string> stateNames = _model->getStateVariableNames();
+    for(auto const& name : stateNames){
+        _variables.insert(pair<string, double>(name, 0.0));
     }
     // RECORD
     int status = 0;
--- a/OpenSim/Simulation/Model/Model.cpp
+++ b/OpenSim/Simulation/Model/Model.cpp
@@ -1462,9 +1462,11 @@ void Model::printDetailedInfo(const SimTK::State& s, std::ostream &aOStream) con
     aOStream<<"\nCONTACTS ("<<n<<")" << std::endl;
 
 */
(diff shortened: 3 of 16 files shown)


Bindings/Java/OpenSimJNI/OpenSimContext.cpp
```diff
@@ -415,13 +415,14 @@ void OpenSimContext::restoreStateFromCachedModel()
...
+    std::vector<std::string> modelVariableNames = _model->getStateVariableNames();
+    std::vector<std::string> clonedModelVariableNames = clonedModel->getStateVariableNames();
 
-    for(int i = 0; i < modelVariableNames.getSize(); i++)
+    for(auto const& name : modelVariableNames)
     {
-        std::string name = modelVariableNames.get(i);
-        if(clonedModelVariableNames.findIndex(name) >= 0)
+        std::vector<std::string>::iterator it;
+        it = std::find(clonedModelVariableNames.begin(), clonedModelVariableNames.end(), name);
```
klshrinidhi: Could use:
`auto it = ...`
and remove previous line.

aymanhab:
@carmichaelong Before you spend any more time on this issue we need to find out if the std::vector bindings in  Matlab/Java and Python are usable or support the same functionality as provided by our OpenSim::Array. The previous PR was all internal to testing and doesn't affect users, this potentially changes most user programs, I advise extreme caution while proceeding! Thanks.

klshrinidhi:
std::vector is wrapped by SWIG.

chrisdembia:
Might be worthwhile to add some python/Java tests just to see how the interface changes for the bindings.

aymanhab:
Thanks @klshrinidhi but we need to instantiate the templates for the types we use and see that we can use them in similar fashion. Also if the operations involve using iterators then iterators may need to be wrapped and/or instantiated as well. In the past users suffered through the differences between SimTK::Vector, OpenSim::Array, adding a third one to the mix is probably not a benefit unless we remove either or both IMHO.

chrisdembia:
Ah yes; it would be great to test iterating through the state variable names in python:

```py
for name in model.getStateVariableNames():
    print name
```

This should work now.

OpenSim/Examples/SymbolicExpressionReporter/SymbolicExpressionReporter.cpp
```diff
@@ -260,9 +260,9 @@ begin( SimTK::State& s)
     // RESET STORAGE
     _resultStore.reset(s.getTime());
     // Populate list of names, initial values from s
-    Array<std::string> stateNames = _model->getStateVariableNames();
-    for(int i=0; i< stateNames.getSize(); i++){
-        _variables.insert(pair<string, double>(stateNames[i], 0.0));
+    std::vector<std::string> stateNames = _model->getStateVariableNames();
+    for(auto const& name : stateNames){
```
chrisdembia: We use the convention `const auto&`

chrisdembia: The `const` applies to whatever is just to its left, with the exception that putting a const all the way to the left means it applies to the thing to its right. So `const auto&` and `auto const&` are the same.

aymanhab:
Definitely @chrisdembia we should add tests in both Python and Java (Matlab), however there are methods that return an OpenSim::Array that the user code manipulates, we basically need to address most/all the possible ways that users (would) have used it or declare that we'll live with whatever std::vector provides. We haven't made this decision yet or studied the implications.

OpenSim/Examples/SymbolicExpressionReporter/SymbolicExpressionReporter.cpp
```diff
@@ -260,9 +260,9 @@ begin( SimTK::State& s)
...
+        _variables.insert(pair<string, double>(name, 0.0));
```
klshrinidhi: Could use:
```
_variables.emplace(name, 0);
```
instead of `insert`.

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -1462,9 +1462,11 @@ void Model::printDetailedInfo(const SimTK::State& s, std::ostream &aOStream) con
     aOStream<<"\nCONTACTS ("<<n<<")" << std::endl;
 
 */
-    Array<string> stateNames = getStateVariableNames();
-    aOStream<<"\nSTATES (total: "<<stateNames.getSize()<<")"<<std::endl;
-    for(int i=0;i<stateNames.getSize();i++) aOStream<<"y["<<i<<"] = "<<stateNames[i]<<std::endl;
+    std::vector<string> stateNames = getStateVariableNames();
+    aOStream << "\nSTATES (total: " << stateNames.size() << ")" << std::endl;
+    for (int i = 0; i < stateNames.size(); i++) {
```
klshrinidhi: `unsigned i` or `size_t i` instead of `int i`.

aymanhab:
We might be able to utilize some community effort (.i) files since this must have been done by many SWIG users already. This may enable us to work around operators in C++ that are not supported cross language.

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -1890,8 +1892,12 @@ void Model::formStateStorage(const Storage& originalStorage, Storage& statesStor
...
         statesStorage.append(*stateVec);
     }
-    rStateNames.insert(0, "time");
-    statesStorage.setColumnLabels(rStateNames);
+    rStateNames.insert(rStateNames.begin(), "time");
+    OpenSim::Array<std::string> rStateNamesArray;
+    for (auto const& name : rStateNames) {
+        rStateNamesArray.append(name);
+    }
+    statesStorage.setColumnLabels(rStateNamesArray);
```
klshrinidhi: Could add a method to `OpenSim::Array` that accepts `std::vector<std::string>` for column-labels. It avoids repeated use of this snippet.

OpenSim/Tools/AnalyzeTool.cpp
```diff
@@ -304,7 +304,11 @@ createStatesStorageFromCoordinatesAndSpeeds(const Model& aModel, const Storage&
     const SimTK::State &s = aModel.getWorkingState();
 
     Storage *statesStore = new Storage(512,"states");
-    statesStore->setColumnLabels(stateNames);
+    Array<string> stateNamesArray;
+    for (auto const& name : stateNames) {
+        stateNamesArray.append(name);
+    }
+    statesStore->setColumnLabels(stateNamesArray);
```
klshrinidhi: With the method in OpenSim::Array, this could be avoided.

klshrinidhi:
:+1:

chrisdembia:
> We might be able to utilize some community effort (.i) files since this must have been done by many SWIG users already. This may enable us to work around operators in C++ that are not supported cross language.

I think SWIG comes with the necessary .i files to wrap std::vector nicely, and gives iterators etc.

jenhicks:
Dev team discussion: back off this PR and instead back OpenSim Array with std::vec. If it proves tricky, we can hold off on this.

chrisdembia:
Thanks @carmichaelong for doing this but as we decided to hold off on this, I am closing this PR to clean up the list of open PRs. We can re-open at a later time.

PR #1231, opened 2016-09-02
Do not test branches starting with `devprop`

chrisdembia:
I noticed that @tkuchida was `[ci skip]`ing his commits to his devprops. This PR tells travis and appveyor not to test any branches whose name starts with `devprop`. Accordingly, I think future devprops should be named with `devprop` at the front.

The only downside I foresee is that people will be able to put tabs into the devprop folder, and it won't be caught until someone later on tries to make a PR unrelated to the devprop.


aymanhab:
LGTM, test failure is bogus and unrelated.

tkuchida:
:+1: LGTM

Issue #1230, opened 2016-09-01
Hopper example throws a !hasSystem() exception out-of-the-box


tkuchida:
The "Examples/Hopper Device" project should spawn a visualizer and simulate when run. The current behavior is for `Component::getSystem()` to throw a `!hasSystem()` exception on `viz.drawFrameNow(state);`:
```
Hopper Example Failed to run due to the following Exception: PathWrapPoint::getS
ystem()  has no reference to a System.
Make sure you added the Component to the Model and called Model::initSystem().
```
The expected behavior (minus the hop) is observed when `vastus->updGeometryPath().addPathWrap(*patella);` (line 164 in buildHopperModel.cpp) is commented out.

aseth1:
Fixed by #1260

PR #1233, opened 2016-09-02
Add hopper example as a test, suppress visualizer through cmd line args.

klshrinidhi:
Addressing #1128.

* Move hopper example to Examples directory.
* Add hopper example as a test.
* Add command line arguments to hopper example to allow:
 * Suppressing visualizer while running as a test.
 * Simulating just once instead of multiple times until user stops.
* Fix broken code due to recent changes in the interface for Joints.


OpenSim/Examples/ExampleHopperDevice/buildHopperModel.cpp
```diff
@@ -88,15 +88,16 @@ Model buildHopper() {
 
     // Set the coordinate names and default values. Note that we need "auto&"
     // here so that we get a reference to the CoordinateSet rather than a copy.
-    auto& sliderCoord = sliderToGround->upd_CoordinateSet()[0];
+    auto& sliderCoord =
```
tkuchida: Hm I'm pretty sure I fixed this already. Anyway, the comment should be updated ("CoordinateSet" -> "Coordinate").

klshrinidhi: Maybe the PR you fixed it in has not been merged yet. The master currently has this problem:
https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Sandbox/ExampleHopperDevice/buildHopperModel.cpp#L91

tkuchida: > Maybe the PR you fixed it in has not been merged yet.

Ah, you are right. I slipped it into #1214 because I wasn't expecting that to be controversial.

tkuchida:
LGTM. Thanks! :rabbit:

OpenSim/Examples/ExampleHopperDevice/exampleHopperDevice.cpp
```diff
@@ -173,8 +173,20 @@ void addDeviceConsoleReporterToModel(Model& model, Device& device,
 // the exercise. The project should execute without making any changes (you
 // should see the unassisted hopper hop slightly).
 //------------------------------------------------------------------------------
-int main()
-{
+int main(int argc, char* argv[]) {
+    //==========================================================================
+    // Command line argument parsing.
```
chrisdembia: Could you give examples of the different possible invocations of the resulting executable?

chrisdembia:
I'm done reviewing.

Looks like a nice elegant solution that doesn't interfere much with the code.

But travis/appveyor tests have not yet passed for this PR. It is fine that the latest edit to comments does not need to be tested, but perhaps an earlier commit should pass all tests. I'd like to ensure that this is tested via the travis/appveyor logs.

The main issue is that, when installed, this example will have the wrong CMakeLists.txt: it'll have a CMakeLists that only works with the opensim-core repo, when it should have a standalone "downstream project" CMakeLists.txt. 

I proposed a solution to this a long time ago and did start working on it but I haven't worked on it in a while. Anyway, I am happy to consider that a separate issue, but a separate GitHub Issue should be created.

OpenSim/Examples/ExampleHopperDevice/helperMethods.h
```diff
@@ -184,7 +195,7 @@ inline void simulate(Model& model, SimTK::State& state, bool saveStatesFile)
 
     // Simulate until the user presses ESC (or enters 'q' if visualization has
     // been disabled).
-    while (true) {
+    while (!simulateOnce) {
```
aseth1: Perhaps I am missing something, but this will result in no simulation, not even once. For it to pass through once the condition would have to start true and become false after the first pass.

OpenSim/Examples/ExampleHopperDevice/helperMethods.h
```diff
@@ -160,11 +163,19 @@ inline void showAllOutputs(const Component& comp, bool includeDescendants)
...
+inline void simulate(Model& model,
+                     SimTK::State& state,
+                     const bool showVisualizer,
+                     const bool simulateOnce,
+                     const bool saveStatesFile) {
     SimTK::State initState = state;
     SimTK::Visualizer::InputSilo* silo;
 
+    if(!showVisualizer) {
+        model.setUseVisualizer(false);
```
chrisdembia: Have you tested this in the interactive scenario where the visualization does show? I think this line must be done before `initSystem()`; I'm not sure it has an effect after the System is built.

OpenSim/Examples/ExampleHopperDevice/helperMethods.h
tkuchida: > this will result in no simulation, not even once

Hm, perhaps we need to revise [point 3 in CONTRIBUTING.md under "Making a Pull Request (PR)"](https://github.com/opensim-org/opensim-core/blob/master/CONTRIBUTING.md#making-a-pull-request-pr) to indicate that tests must pass *and* the developer must ensure that the code executes as designed?

chrisdembia: > Hm, perhaps we need to revise point 3 in CONTRIBUTING.md under "Making a Pull Request (PR)" to indicate that tests must pass and the developer must ensure that the code executes as designed?

Sounds good to me. I think that point 3 is really about making sure *other* / *all* tests pass (tests aside from the ones related to the feature being worked on). But I do think it would be good to add such a statement to that document.

klshrinidhi: I intended to change `while` to `do-while`. 

The current version works without visualizer with one simulation. It is broken if visualizer is turned on.

OpenSim/Examples/ExampleHopperDevice/exampleHopperDevice.cpp
klshrinidhi: Added description of arguments that can be passed. Because the two arguments are optional, listing all different ways to call the executable could be too much.

OpenSim/Examples/ExampleHopperDevice/exampleHopperDevice.cpp
```diff
@@ -173,8 +173,27 @@ void addDeviceConsoleReporterToModel(Model& model, Device& device,
...
+    //==========================================================================
+    // Following are arguments to this executable:
+    // (1) 'noVisualizer' -- [Optional] Do no show visualizer. Default
```
tkuchida: Do no -> Do not

klshrinidhi: Done.

Can you see why it breaks with visualizer in this version ?

OpenSim/Examples/ExampleHopperDevice/exampleHopperDevice.cpp
```diff
@@ -173,8 +173,27 @@ void addDeviceConsoleReporterToModel(Model& model, Device& device,
...
+    // (2) 'simulateOnce' -- [Optional] Simulate once and exit the program.
+    //                       Default behavior is to simulate multiple times
+    //                       until user requests to stop.
+    
+    // Suppress/show visualizer.
+    bool showVisualizer{true};
+    // Simulate just once or until user stops.
+    bool simulateOnce{false};
+    for(int i = 0; i < argc; ++i)
+        if(strcmp(argv[i], "noVisualizer") == 0)
```
tkuchida: According to the instructions above, the text should be surrounded by single quotes. Is it overkill to use a regex here to retain only `[a-zA-Z]+`?

klshrinidhi: Quote is not required when calling the executable. I had to emphasize the argument.
I have removed the quote now.

tkuchida:
> The current version works without visualizer with one simulation. It is broken if visualizer is turned on.

Do you get the exception reported in #1230? If so, I think this PR can be merged once tests pass on the most recent non-[skip ci] commit (5d279c5).

klshrinidhi:
> > The current version works without visualizer with one simulation. It is broken if visualizer is turned on.
>
>Do you get the exception reported in #1230? If so, I think this PR can be merged once tests pass on the most recent non-[skip ci] commit (5d279c5).

It is exactly that exception.

PR #1235, opened 2016-09-05
Update copyright year to 2016

klshrinidhi:


OpenSim/Actuators/About.cpp
```diff
@@ -36,7 +36,7 @@
 #define STR(var) #var
 #define MAKE_VERSION_STRING(maj,min,build)  STR(maj.min.build)
 #define MAKE_COPYRIGHT_STRING(y,a) \
-    "Copyright (c) " STR(y) " Stanford University, " STR(a)
+    "Copyright (c) 2005-2016 Stanford University, " STR(a)
```
chrisdembia: What was the reason for changing this?

Bindings/Java/OpenSimJNI/About.cpp
```diff
@@ -36,7 +36,7 @@
 #define STR(var) #var
 #define MAKE_VERSION_STRING(maj,min,build)  STR(maj.min.build)
 #define MAKE_COPYRIGHT_STRING(y,a) \
-    "Copyright (c) " STR(y) " Stanford University, " STR(a)
+    "Copyright (c) 2005-2016 Stanford University, " STR(a)
```
chrisdembia: I would prefer to roll back this change here. Otherwise, I think the macro should be updated to not have an unused parameter.

chrisdembia:
Cool! Related issue: #672. I made one comment. Once that is addressed, I think this should be merged.

klshrinidhi:
Reverted changes to macros.

chrisdembia:
LGTM.

tkuchida:
Have you (a) described how to do this somewhere, or is it (b) obvious to "real" programmers or (c) too complicated to explain?

klshrinidhi:
It is "search & replace" done multiple times. :stuck_out_tongue_winking_eye:

chrisdembia:
> It is "search & replace" done multiple times. 😜

In emacs? You went file by file???

> Have you (a) described how to do this somewhere, or is it (b) obvious to "real" programmers or (c) too complicated to explain?

Ah yes it would be good to share any script you may have used, since we are likely to need to do this again.

tkuchida:
> It is "search & replace" done multiple times.

Ah, ok. IIRC this was done in the past using some kind of code/script that is now lost.

jenhicks:
This looks OK to me. 

There was a larger change in the past to change the whole license block. @aymanhab do you know if we still have that script anywhere? I think Kevin worked on it.

aymanhab:
Will look it up and report back.

chrisdembia:
FYI I spent some time looking for it also but did not find it. I started working on a python script to change the license for the GUI: https://github.com/opensim-org/opensim-gui/compare/apache-2.0. It would work for this repo too with a few small changes.

jenhicks:
Thanks @chrisdembia and @aymanhab for the update. This discussion is separate from the current PR though, I think. (so no need to hold up merge.)

aymanhab:
For the record, the scripts are on confluence page 
http://simtk-confluence.stanford.edu:8080/display/DevPlans/Kevin%27s+notes

chrisdembia:
Ah cool!

Issue #1241, opened 2016-09-07
prescribeControlForActuator is not recognized in python


nikorose87:
Hi, I've been working with the opensim software in windows but now I'm moving to the API, specifically transferring to python. I'm trying to run the example moving the code of elbow arm algorithm to python, as it is shown below:

```
import opensim

model = opensim.Model()
model.setUseVisualizer(True)
humerus = opensim.Body("humerus",1, opensim.Vec3(0), opensim.Inertia(0))
radius = opensim.Body("radius",1, opensim.Vec3(0), opensim.Inertia(0))
shoulder = opensim.PinJoint("shoulder", model.getGround(), opensim.Vec3(0), opensim.Vec3(0), humerus, opensim.Vec3(0,1,0), opensim.Vec3(0))
elbow = opensim.PinJoint("elbow", humerus, opensim.Vec3(0), opensim.Vec3(0), radius, opensim.Vec3(0,1,0),opensim.Vec3(0))

#Add a muscle that flexes the elbow
biceps = opensim.Millard2012EquilibriumMuscle("biceps", 200, 0.6, 0.55, 0)
biceps.addNewPathPoint("origin", humerus, opensim.Vec3(0,0.8,0))
biceps.addNewPathPoint("insertion", radius, opensim.Vec3(0,0.7,0))
print(type(biceps))

# Add a controller that specifies the excitation of the muscle.
#opensim.Controller.addActuator(1,biceps)
brain = opensim.PrescribedController().addActuator(biceps).prescribeControlForActuator("biceps", opensim.StepFunction(0.5, 3, 0.3, 1))

# Muscle excitation is 0.3 for the first 0.5 seconds, then increases to 1.
model.addBody(humerus)
model.addBody(radius)
model.addJoint(shoulder)
model.addJoint(elbow)
model.addForce(biceps)
#model.addController(brain)

# Add a console reporter to print the muscle fiber force and elbow angle.
reporter = opensim.ConsoleReporter().set_report_time_interval(1.0)
#reporter.updInput("inputs").connect(biceps.getOutput("fiber_force"))
#reporter.updInput("inputs").connect(
#    elbow.getCoordinateSet()[0].getOutput("value"), "elbow_angle")
#model.addComponent(reporter)

```
Unfortunately,  the library did not recognize the  prescribeControlForActuator class, showing the following error: 
```
[...]

PR #1242, opened 2016-09-07
[readme] Mention python version of README example

chrisdembia:
In #1241, a user started converting the README example into python, presumably because the user did not know that a python version already exists. This PR attempts to avoid such duplicated effort in the future by putting a note in the README about the python example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,10 @@ include source code for the OpenSim GUI.
 
 Simple example
 --------------
-Let's simulate a simple arm whose elbow is actuated by a muscle:
+Let's simulate a simple arm whose elbow is actuated by a muscle, using 
+the C++ interface (You can find a python version of this example at 
+`Bindings/Python/examples/build_simple_arm_model.py`):
+
 ```cpp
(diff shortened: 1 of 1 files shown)


chrisdembia:
I cancelled the CI builds because this PR only affects the README.

jenhicks:
Looks good! Ok by me to merge.

carmichaelong:
It could merge as is, or I would consider moving it to the top of the section where it says "Let's simulate a simple arm whose elbow is actuated by a muscle:" so that a Python user won't go through the whole C++ code first and maybe miss the comment at the end. Maybe it could be something like:

Let's simulate a simple arm whose elbow is actuated by a muscle using the C++ libraries. (You can find a python version of this example at Bindings/Python/examples/build_simple_arm_model.py).

chrisdembia:
Thanks @jenhicks and @carmichaelong.

> I would consider moving it to the top of the section

Addressed.

Ready for review again.

chrisdembia:
Thank you!

PR #1243, opened 2016-09-08
Remove Model's reordering of Joints in its JointSet to fix #1105

### LINKED ISSUE #1105: "Array index out of bounds" error when reordering Joints in the multibody tree

aseth1:
This PR is intended to fix issue #1105 where joints in a nested Component (device) results in an invalid index when reordering joints in the Model's JointSet. 

The first step is adding a test that reproduces #1105. (First commit). 

The remainder of the PR will include:
 - merging #1205 to remove the need to reorder Joints
 - removing the reordering of the Joints in `Model::extendConnectToModel` 
 - dealing with secondary effects like Tools expecting Joints (or their Coordinates) in MultibodyTree order


aseth1:
@tkuchida @chrisdembia @aymanhab this PR is ready for review and addresses #1105

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -489,6 +494,46 @@ void Model::createMultibodySystem()
...
+        auto mbix = coord.getBodyIndex();
+        auto mqix = coord.getMobilizerQIndex();
+
+        int cix = matter.getMobilizedBody(mbix).getFirstUIndex(s) + mqix;
+
+        SimTK_ASSERT_ALWAYS(cix < nc, "Index exceeds the number of Coordinates "
+            "in this Model.");
+
+        // Set the coordinate in the right slot in the CoordinateSet
+        coordSet.set(cix, &coord);
```
chrisdembia: Could you add a comment to `Model::getCoordinateSet()` (or anywhere that you think it is relevant for the user) that describes this ordering? of the CoordinateSet()?

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -489,6 +494,46 @@ void Model::createMultibodySystem()
...
+        "Attempting to reorder Coordinates according to an invalid System.");
+
+    auto& coordSet = updCoordinateSet();
+
+    // We have a valid MultibodySystem underlying the Coordinates
+    int nc = coordSet.getSize();
+    const SimTK::State& s = getModel().getWorkingState();
+    SimTK_ASSERT_ALWAYS(nc <= s.getNQ(),
+        "Number of Coordinates exceeds the number of mobilities in "
+        "the underlying MultibodySystem.");
```
chrisdembia: I think that, technically, `getNU()` is the number of mobilities, not `getNQ()`. Perhaps:

"Number of OpenSim Coordinates exceeds the number of generalized coordinates..."

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -489,6 +494,46 @@ void Model::createMultibodySystem()
...
+            "in this Model.");
+
+        // Set the coordinate in the right slot in the CoordinateSet
+        coordSet.set(cix, &coord);
+        cnt++;
+    }
+
+    SimTK_ASSERT_ALWAYS(cnt == nc,
+        "Reordered Coordinates does not correspond to the number of "
+        "Coordinates in the Model.");
```
chrisdembia: Unless I misunderstand this function, this check should not be necessary, given the `cix < nc` check above.

OpenSim/Simulation/Model/Model.cpp
chrisdembia: Are you sure that this check isn't performed somewhere else as well?

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -489,6 +494,46 @@ void Model::createMultibodySystem()
     addToSystem(*_system);
 }
 
+
+void Model::reorderCoordinatesAccordingToSystemMobilities()
```
chrisdembia: Why does this function rely on Topology? Is it because you need a State? Could you document this at the declaration of `Model::reorderCoordinatesAccordingToSystemMobilities()`?

aymanhab:
@aseth1 I'm a bit confused reviewing this PR as to why the order of coordinates matter to the Model at all rather than to the Tools (where ordering/mapping of tasks, data, and model coordinates could all be different and the tool does the mapping/matching by name). Should it matter to the model whether left-hip or right-hip is attached first to pelvis (same level)? Maybe a comment along with reorderCoordinatesAccordingToSystemMobilities() would help explain. Thanks.

chrisdembia:
> why the order of coordinates matter to the Model at all rather than to the Tools (where ordering/mapping of tasks, data, and model coordinates could all be different and the tool does the mapping/matching by name).

Ah, what if you could simply ask the model for a list of the coordinates in the order the tools want, and not actually change the order in the `CoordinateSet()`? 

`std::vector<const Coordinate&> Model::getCoordinatesInMultibodyTreeOrder()`?

aymanhab:
Exactly, @chrisdembia so the model can provide the service to the tools but its internals are hidden can can change as needed.

aseth1:
@chrisdembia @aymanhab You are right that the model should not care about the order of Joints, Coordinates or any of its Components, for that matter, and it really doesn't it.  We handled the reporting of the states based only on names. In this case, the issue is with generalized forces. The ID, RRA, SO expect a vector of gen. forces in tree order. I was reluctant to significantly change the Tools so opted to reorder the Coordinates. I could easily be convinced not to do it. I also like the idea of just having a method (service) that returns coordinates in the desired order/convention.  I will give that a try. Thanks!

chrisdembia:
> I also like the idea of just having a method (service) that returns coordinates in the desired order/convention. I will give that a try.

Without having really thought about it, this seems to be as easy/difficult as the current solution. So I think it makes sense to try it. If it turns out to be substantially more difficult, we can reconsider.

aymanhab:
Awesome, thanks @aseth1 for trying it out, and sorry for the late review. Cheers.

OpenSim/Simulation/Test/testNestedModel.cpp
```diff
@@ -0,0 +1,121 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.   *
+ * See the License for the specific language governing permissions and        *
+ * limitations under the License.                                             *
+ * -------------------------------------------------------------------------- */
+
+/*=============================================================================
+
+Tests Include:
+    1. Pendulum Model with nested Device derived from ModelComponent 
+    2. Pendulum Model with a Device that is a Model
```
aseth1: Done.

OpenSim/Simulation/Test/testNestedModel.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+    // add the device to the pendulum model
+    pendulum->addModelComponent(device);
+
+    // Connect the device to the pendulum
+    auto bodies = pendulum->getComponentList<OpenSim::Body>();
+    auto bodyIter = bodies.begin();
+    anchorA->updConnector("parent_frame").connect(*bodyIter);
+    bodyIter++;
+    anchorB->updConnector("parent_frame").connect(*bodyIter);
```
aseth1: Done.

OpenSim/Simulation/Model/Model.cpp
aseth1: A user-facing comment is provided with the new public `getCoordinatesInMultibodyTreeOrder()` method and the `CoordinateSet` is left untouched.

aseth1:
@chrisdembia @aymanhab thank you for your feedback. 

I reverted the changes to reorder the Model's CoordinateSet directly. Instead there is now a public `getCoordinatesInMultibodyTreeOrder()` that enables IDTool, StaticOptimization, InverseDynamics (Analysis) and RRATool to account for the order of generalized forces returned by their solvers versus that of the CoordinateSet in the Model.  There should now be no assumption about the order of Coordinates in the Model's CoordinateSet across all Tools. I think this is a better solution than I had before, so thank you for proposing the alternative.

It also exposed a bug in `RRATool` when assigning desired kinematics to Coordinates, it was assuming that Coordinates were in MultibodyTree order. RRATool was updated to reflect changes made to other Tools like `CMCTool` to decouple `CoordinateSet` order from the `State` (or Tree) order.

This PR is ready for a follow-up review. Thank you.

aymanhab:
Thanks @aseth1 much more understandable now. 
In old "Analyses" (InverseDynamics, StaticOptimization etc.) we were using the coordinates from a copy of the model maintained internally, these were changed to calls to getCoordinatesInMultibodyTreeOrder on the original model, is that intended? Is there a chance the two models could be out of sync.? 
The other question I have is whether you expect the function getCoordinatesInMultibodyTreeOrder  to be called from scripting since the return type maybe hard to SWiG wrap. Other than these two questions/clarifications, looks great. Thanks.

aseth1:
> In old "Analyses" (InverseDynamics, StaticOptimization etc.) we were using the coordinates from a copy of the model maintained internally, these were changed to calls to getCoordinatesInMultibodyTreeOrder on the original model, is that intended? Is there a chance the two models could be out of sync.?

I did not realize that I inadvertently changed that. It was not intentional, so I am reverting so they access `_modelWorkingCopy` like before.

aseth1:
> The other question I have is whether you expect the function getCoordinatesInMultibodyTreeOrder to be called from scripting since the return type maybe hard to SWiG wrap.

Hadn't thought about using via scripting, but I would not want to restrict its use (it is public) either. Any opinions for making it more wrapable? I have to be honest that `std::reference_wrapper<const Coordinate>` did not provide the use syntax of `coordinates[i].getValue()` that I wanted, and instead it resulted in `coordinates[i].get().getValue()`. I could easily convert to `SimTK::ReferencePtr` if that is easier to wrap. @aymanhab @chrisdembia any preferences?

aymanhab:
The SWIG doc says there's no special treatment for thesestd::reference_wrappers and you have to write your own "typemaps" which makes me lean toward the SimTK one if starting from scratch, but I wouldn't hold this PR for it, we can see how the wrapping looks after this PR is in.

aseth1:
I am leaning towards switching to `SimTK::ReferencePtr<const Coordinate>` for the following reasons:
 1. it is already in use in the API and likely we have or will have a solution to wrap it.
 2. this would be the only use of `std::reference_wrapper` in the API and it doesn't provide the direct syntax we wanted

OpenSim/Simulation/Test/testNestedModel.cpp
```diff
@@ -0,0 +1,121 @@
+/* -------------------------------------------------------------------------- *
+ *                      OpenSim:  testNestedModel.cpp                         *
```
chrisdembia: Yes I would prefer that.

OpenSim/Analyses/InverseDynamics.cpp
```diff
@@ -338,8 +337,11 @@ record(const SimTK::State& s)
         for(int i=0; i<nacc; i++) _constraintMatrix(i,j) = (c[i] - _constraintVector[i]);
         f[j] = 0;
     }
+
+    auto coordinates = _modelWorkingCopy->getCoordinatesInMultibodyTreeOrder();
```
chrisdembia: This is re-evaluated for each time step, right? How does this affect performance? Could the InverseDynamics class hold onto the list during initialization?

OpenSim/Analyses/StaticOptimizationTarget.cpp
```diff
@@ -586,9 +583,11 @@ computeConstraintVector(SimTK::State& s, const Vector &parameters,Vector &constr
     Vector actualAcceleration(getNumConstraints());
     computeAcceleration(s, parameters, actualAcceleration);
 
+    auto coordinates = _model->getCoordinatesInMultibodyTreeOrder();
```
chrisdembia: Could this be expensive at all?

OpenSim/Simulation/Model/Model.h
```diff
@@ -802,6 +802,20 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Model, ModelComponent);
...
+    /** Obtain a list of Model's Coordinates in the order they appear in the
+        MultibodySystem after Model::initSystem() has been called. 
+        Coordinates in the CoordinateSet do not have a predefined order. In
+        some instances it is helpful to get the coordinates in order of 
+        generalized coordinates in the Multibody Tree as defined in the 
+        underlying MultibodySystem. For example, computing the generalized
+        forces from the System, yields a vector of generalized forces
+        in order of the Multibody Tree and now that can be attributed to 
+        corresponding generalized Coordinates of the Model. 
+        Throws if the MultibodySystem is not valid. */
```
chrisdembia: Looks good.

OpenSim/Analyses/InverseDynamics.cpp
aseth1: I do think it introduces a small overhead, but the `InverseDynamic`s Analysis is only being used by RRA (RRA should be using `InverseDynamicsSolver`, but that is another issue) so I am reluctant to optimize for something that should be removed. 

Current PR Appveyor x86 `testRRA`: [11.08s](https://ci.appveyor.com/project/opensim-org/opensim-core/build/4921/job/w95h8obwmfecua0t#L610)
PR #1279 Appveyor x86  `testRRA`: [9.89s](https://ci.appveyor.com/project/opensim-org/opensim-core/build/4918/job/5cbpa88fnqra9cox#L544)

Is this a tolerable hit? If/when we switch to `InverseDynamicsSolver` the issue also goes away.

OpenSim/Analyses/StaticOptimizationTarget.cpp
aseth1: Good point. Let's compare:
Current PR `testStaticOptimization` on Appveyor x86: [114.84](https://ci.appveyor.com/project/opensim-org/opensim-core/build/4921/job/w95h8obwmfecua0t#L596)
PR #1279  `testStaticOptimization` on Appveyor x86: [112.95] (https://ci.appveyor.com/project/opensim-org/opensim-core/build/4918/job/5cbpa88fnqra9cox#L530)

The overhead of rummaging through coordinates appears to be small compared to performing the optimization at each step. The `computeConstraintVector` is called prior to optimization in `prepareToOptimize`.

OpenSim/Analyses/InverseDynamics.cpp
chrisdembia: Thanks for checking.



> Is this a tolerable hit?

Yes I think so.

aseth1:
@chrisdembia @aymanhab a friendly reminder that the requested changes are complete.
Test failures are due 1) sporadic failure in testContactGeometry timing out in debug and 2) error in Lepton on OSX. Both of which are unrelated this PR. Please let me know if there is anything else to be done on this PR. Thanks!

aymanhab:
LGTM, Thanks @aseth1 I'll merge it unless @chrisdembia has any outstanding concerns.

aseth1:
Thank you @chrisdembia and @aymanhab.

PR #1244, opened 2016-09-08
Fix warnings from MSVC

klshrinidhi:


OpenSim/Common/TableSource.h
```diff
@@ -202,21 +202,30 @@ class TableSource_ : public Component {
                          TimeOutOfRange, 
                          time, timeCol.front(), timeCol.back());
 
-        const auto colInd = _table.getColumnIndex(columnLabel);
+        const auto colInd = 
+            static_cast<int>(_table.getColumnIndex(columnLabel));
```
tkuchida: Out of curiosity, why does `_table.getColumnIndex()` return a size_t (unsigned) while `_storage.getSize()` returns an int (see the change in Storage.cpp, above)? It seems to be equally impossible for a size and a column index to be negative.

tkuchida:
:+1: LGTM

chrisdembia:
LGTM. I would merge this myself but I will let @tkuchida or @klshrinidhi do so because @klshrinidhi may want to respond to @tkuchida's comment.

OpenSim/Common/TableSource.h
klshrinidhi: I believe unsigned or size_t is the correct return type. It represents index correctly -- as a non-negative integer. I have used this in my code.

I am not sure why Storage was written differently. I think there are a lot of places where an int is (incorrectly) used as an argument or return type in OpenSim.

tkuchida: > I think there are a lot of places where an int is (incorrectly) used as an argument or return type in OpenSim.

I thought a potential rationale for returning an int from `getSize()` was to avoid the possibility of accidentally wrapping through zero when subtracting (which could go unnoticed, e.g., as an intermediate calculation in a long expression)—but I agree with your use of size_t. Anyway, I was just curious; no need to hold up the PR.

Issue #1159, opened 2016-08-04
getComponentList<C> breaks for nested loops


aseth1:
We currently cannot cycle through the Joints and concurrently cycle through each of their Coordinates using the `ComponentList`, as in:

```cpp
        for (auto& joint : hopper.getComponentList<Joint>()) {
            std::cout << "Joint:" << joint.getFullPathName() << std::endl;
            for (auto& coord : joint.getComponentList<Coordinate>()) {
                std::cout << "   coord:" << coord.getFullPathName() << std::endl;
            }
        }
```
Because the Component's `_nextComponent` pointer is reset every time the list is created. The list is recreated each time for reliability as to never allow the tree to be "stale" so the solution is not revert back to building the list once at some special time and build in smarts to detect changes.  Instead this is a strong case for removing the `_nextComponent` member in `Component`.  We should not have mixed (stuffed in as convenience) the linked link structure of the `ComponentList` with the Component itself! The Component should not partly implement nor have any effect on any data structure it is a part of. 

We should add this to `testIterators.cpp` and eliminate `Component::_nextComponent`  as a 'cache' for the ComponentList's underlying linked list. Subsequently I would like to see the need for `initComponentTreeTraversal()` to go away. The iterator should simply walk the same path used to set the `_nextComponent` pointers, which is all the init method does. @klshrinidhi can you add test case(s) to `testIterators.cpp` and try to fix this?

klshrinidhi:
I will work on this.

jenhicks:
@klshrinidhi Please check in with @aymanhab about the implementation before proceeding so you don't chase 👻

jenhicks:
The immediate bug is fixed by the PR referenced above, so I am in favor of working on decoupling Component from ComponentList after 4.0. @aseth1 does that sound reasonable?

aseth1:
> I am in favor of working on decoupling Component from ComponentList after 4.0. @aseth1 does that sound reasonable?

Yes. But I think we should be prepared for this issue to creep up again and then address it once and for all. As @aymanhab noted in the last meeting, reassigning the `_nextComponent` pointer can lead to undefined behavior. I am unsure of the circumstances that guarantee that the `_nextComponent` pointer is being preserved on repeated calls. It seems to be OK but it has not been tested extensively. The band-aid seems to be holding, for now.

jenhicks:
Thanks for the input @aseth1 . I will defer this issue for now.

aseth1:
[...]

PR #1248, opened 2016-09-12
Fix bug that breaks nested loops with ComponentListIterator. Includes DevProp.

klshrinidhi:
Addressing #1159.

The fix provided here is probably the simplest requiring minimum change in my opinion. The issue above suggests removing `_nextComponent` and introducing logic to handle the tree traversal. This PR does not implement the suggestion.

I have included the fix and dev-prop in the same PR as it was a pretty small fix. I didn't think dev-prop was necessary for this one but here it is anyway.


aseth1:
Thank you @klshrinidhi for addressing this bug.

My request for a design proposal was to address the decoupling of the `Component` from the `ComponentList` data structure used to iterate through subcomponents. I apologize if that intent was not clear in the issue.  IMHO, decoupling would not only fix this bug, it would be better code hygiene and prevents other issues.  I realize that you found a more targeted solution, which is great, but I think we could avoid this and other problems by removing the `_nextComponent` pointer from `Component` altogether.

For instance,  the `_nextComponent` pointer cannot guarantee that the traversal of the outer list is preserved because creating the inner list resets it. I think that is the real problem here and it seems fragile that we are relying on the `_nextComponent` to be reset to the same value (when the next list is created) otherwise the previous list is not conserved.  Removing the partial linked list implementation of the `Component` (resulting in the `_nextComponent` data member) would be more robust and eliminate related issues of using concurrent lists of Components.

klshrinidhi:
Okay. I understand.

I think now it is better to:
* Merge this PR as a solution to the bug as it addresses the bug. This will allow other code to use the fix immediately without having to wait for the long term solution. This PR is not introducing new problems.
* Open another issue describing the intention to decouple the classes. We could discuss it in a dev meeting if it is worth doing now that the bug is fixed. We could potentially  wait till we see another problem/bug to work on long term solution. I could spend time working on something more high priority meanwhile.

aseth1:
>I think now it is better to:
* Merge this PR as a solution to the bug as it addresses the bug. This will allow other code to use the fix immediately without having to wait for the long term solution. This PR is not introducing new problems.
* Open another issue describing the intention to decouple the classes. We could discuss it in a dev meeting if it is worth doing now that the bug is fixed. We could potentially  wait till we see another problem/bug to work on long term solution. I could spend time working on something more high priority meanwhile.

This sounds reasonable. Before we merge, can you add one or more test case to verify:

1. that the fix works when Joints are nested in subcomponents of the Model. (E.g. the Joints inside of a device).
2. that in general the outer list is preserved. This can be done by:

klshrinidhi:
Sure.
I basically wrote this PR as a solution to the issue's specific problem. I haven't thought of anything else in general.
If you can provide code snippets like in the issue, I can look into it.

aseth1:
> If you can provide code snippets like in the issue, I can look into it.

OK something like this would be good:

```cpp 
// Model to add the device to
Model model"leg6dof9musc.osim"); 
...
auto* device = new Device();
...
// Device (like Hopper example) used to nest other joints (any types other 
// than Weld since these have no coordinates). Joints should already have
// the bodies (Thigh and Shank) satisfying their `parent_frame`  Connectors.
// and the respective bodies on the device (e.g. cuffs) as the `child_frame` 
// satisfied.
device->addJoint(ballJointWithThigh);
device->addJoint(ballJointWithShank);
...
model.addModelComponent(device);

model.finalizeFromProperties();

std::vector<const Joint *> joints1;
std::vector<const Joint *> joints2;
std::vector<const Coordinate*> coords;

for (auto& joint : model.getComponentList<Joint>()) {
    joints1.push_back(&joint);
}

for (auto& joint : model.getComponentList<Joint>()) {
    std::cout << "Joint:" << joint.getFullPathName() << std::endl;
    joints2.push_back(&joint);
            
    for (auto& coord : joint.getComponentList<Coordinate>()) {
        std::cout << "   coord:" << coord.getFullPathName() << std::endl;
        coords.push_back(&coord); 
      }
}

// Verify that joints1 list has 6 elements (4 from model and 2 from device)
// and joints1 is unique.
...
// Verify that joints1 and joints2 lists are identical
...
// Verify coords.size()==12 (6 from base model + 2*3 coords from ball joints of the device)
...
// and that coords is unique (i.e. we don't inadvertently traverse the same branch)
...

```

Let me know if you have any questions. May be easier to chat in person for 5min.
Also @aymanhab was intentionally setting root's `_nextComponent` pointer to `nullptr` and I think we should check in with his reasoning/issue for doing that (if you haven't done so already).

Thanks!

klshrinidhi:
I have added the test into a function. The model used is the same as the one rest of the tests use in the file.

OpenSim/Tests/testIterators/testIterators.cpp
```diff
@@ -62,6 +63,110 @@ const int expectedNumJntComponents = 3;
...
+    }
+
+protected:
+    // Change the color of the device's path as its tension changes.
+    void extendRealizeDynamics(const SimTK::State& s) const override {
+        const auto& actuator = getComponent<PathActuator>("cableAtoB");
+        double level = fmin(1., getTension(s) / actuator.get_optimal_force());
+        actuator.getGeometryPath().setColor(s, SimTK::Vec3(0.1, level, 0.1));
+    }
+
```
aseth1: For the purpose of this test the Outputs and related methods as well as the color changing code is not unnecessary. Better to get rid of it to reduce clutter in the test.

OpenSim/Tests/testIterators/testIterators.cpp
```diff
@@ -62,6 +63,110 @@ const int expectedNumJntComponents = 3;
...
+        joints2.insert(&joint);
+        std::cout << "    Joint: " << joint.getFullPathName() << std::endl;
+        for(const auto& coord : joint.getComponentList<Coordinate>()) {
+            std::cout << "        Coord: "
+                      << coord.getFullPathName() << std::endl;
+            coords.insert(&coord);
+        }
+    }
+
+    ASSERT(joints1.size() == 4);
```
aseth1: :+1:

OpenSim/Tests/testIterators/testIterators.cpp
```diff
@@ -454,6 +558,7 @@ int main() {
         SimTK_SUBTEST(testComponentListNonConstWithConstIterator);
         SimTK_SUBTEST(testComponentListNonConstWithNonConstIterator);
         SimTK_SUBTEST(testComponentListComparisonOperators);
+        SimTK_SUBTEST(testComponentListRefs);
```
aseth1: without comments a more descriptive test name would be helpful, like: `testNestedComponentListConsistency` or something along those lines.

OpenSim/Tests/testIterators/testIterators.cpp
```diff
@@ -62,6 +63,110 @@ const int expectedNumJntComponents = 3;
...
+        std::cout << "    Joint: " << joint.getFullPathName() << std::endl;
+        for(const auto& coord : joint.getComponentList<Coordinate>()) {
+            std::cout << "        Coord: "
+                      << coord.getFullPathName() << std::endl;
+            coords.insert(&coord);
+        }
+    }
+
+    ASSERT(joints1.size() == 4);
+    ASSERT(joints1.size() == joints2.size());
```
aseth1: I think it is important to verify that the component lists are identical, including that they produce the same order. If the order changes for any reason it means that the inner list of coordinates is affecting the outer list, which is what we are trying to check does not happen.  While `std::set` is useful to ensure uniqueness it will also order the elements, so comparing sets will not verify if list order was preserved.

DevelopmentProposals/FixingBugWithNestedComponentListIterator/readme.md
```diff
@@ -0,0 +1,36 @@
+## Project: Address problem/bug with nested loops using `ComponentListIterator`
```
aseth1: I agree for this ' bug fix' (which is not a proposal to decouple the Component from the `ComponentList` data structure) it is not necessary to include this DevProp.

klshrinidhi:
Done for all.

OpenSim/Tests/testIterators/testIterators.cpp
```diff
@@ -142,28 +97,31 @@ void testComponentListRefs() {
     model.addModelComponent(device);
     model.finalizeFromProperties();
 
-    std::set<const Joint*> joints1{}, joints2{};
-    std::set<const Coordinate*> coords{};
+    std::vector<const Joint*> joints1{}, joints2{};
+    std::vector<const Coordinate*> coords{};
```
aseth1: set was fine for coords, we do want them to be unique

klshrinidhi:
Done.

aymanhab:
LGTM Merging.

PR #1259, opened 2016-09-17
Replaced DecorativeEllipsoids with Ellipsoids in README example

### LINKED ISSUE #1254: In README, add Ellipsoids to the Model rather than viz directly
Update testREADME.cpp, README.md, and test in Matlab.

tkuchida:
Attached Ellipsoids to PhysicalOffsetFrames located at the center of each Body. A positive side effect is that we are now demonstrating Frames in the README example. Did not generate new gif because appearance of visualization did not change substantially. Fixes #1254.

### DIFF
--- a/OpenSim/Tests/README/testREADME.cpp
+++ b/OpenSim/Tests/README/testREADME.cpp
@@ -94,19 +94,17 @@ int main() {
     model.addComponent(reporter);
 
     // Add display geometry.
-#ifdef VISUALIZE
-    Ellipsoid* bodyGeometry = new Ellipsoid(0.1, 0.5, 0.1);
-    bodyGeometry->setColor(Gray);
+    Ellipsoid bodyGeometry(0.1, 0.5, 0.1);
+    bodyGeometry.setColor(Gray);
     // Attach an ellipsoid to a frame located at the center of each body.
     PhysicalOffsetFrame* humerusCenter = new PhysicalOffsetFrame(
         "humerusCenter", "humerus", Transform(Vec3(0, 0.5, 0)));
     humerus->addComponent(humerusCenter);
-    humerusCenter->attachGeometry(bodyGeometry->clone());
+    humerusCenter->attachGeometry(bodyGeometry.clone());
     PhysicalOffsetFrame* radiusCenter = new PhysicalOffsetFrame(
         "radiusCenter", "radius", Transform(Vec3(0, 0.5, 0)));
     radius->addComponent(radiusCenter);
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -77,21 +77,30 @@ int main() {
...
     model.addComponent(reporter);
 
+    // Add display geometry.
+    Ellipsoid* bodyGeometry = new Ellipsoid(0.1, 0.5, 0.1);
+    bodyGeometry->setColor(Gray);
+    // Attach an ellipsoid to a frame located at the center of each body.
+    PhysicalOffsetFrame* humerusCenter = new PhysicalOffsetFrame(
+        "humerusCenter", "humerus", Transform(Vec3(0, 0.5, 0)));
+    humerus->addComponent(humerusCenter);
+    humerusCenter->attachGeometry(bodyGeometry->clone());
```
chrisdembia: I think this means that the initial `Ellipsoid* bodyGeometry` is leaked. Perhaps (i) allocate the Ellipsoid on the stack, or (ii) remove `clone()` here but leave it in the second time.

OpenSim/Tests/README/testREADME.cpp
```diff
@@ -93,22 +93,33 @@ int main() {
         "elbow_angle");
     model.addComponent(reporter);
 
+    // Add display geometry.
+#ifdef VISUALIZE
```
aseth1: I don't think you have to guard against adding the Geometry. If the Model is not visualized the Geometry won't be visible but it will still be part of the Model.

README.md
aseth1: I have a slight preference for the stack allocated Ellipsoid with both `attachGeometry` calls getting clones. Seems more uniform to me. Otherwise, overall LGTM :+1:

OpenSim/Tests/README/testREADME.cpp
tkuchida: Yes, you are correct.

README.md
tkuchida: Yes, I originally had `clone()` on only one of the calls to `attachGeometry()` but changed it to preserve symmetry, which I thought would be appropriate for this didactic example. I've adopted @chrisdembia's (i) and @aseth1's slight preference.

tkuchida:
Addressed concerns; ready for re-review.

chrisdembia:
LGTM (I have approved the changes).

tkuchida:
@aseth1: Merge or submerge?

aseth1:
Definitely merge!

PR #1260, opened 2016-09-17
Fix hopper visualization failure 

aseth1:
In #1230 the `!hasSystem()` Exception was being thrown because `PathWrapPoint`s were being asked their location in Ground, but they are not "Components" of the Model. They are "ethereal" in the sense that a `WrapObject` creates points on its surface (for the purpose of being drawn it would seem). These points (as Vec3s) were being used to construct PathPoints and being cached as a `display_path` but they are not part of the `Model` or the `System` being simulated- they were only being used to be drawn. It seems completely unnecessary to cache `display_path` especially now that the the `GeometryPath` can draw  itself based on the actual (current) path.

There are two significant changes:

1. `GeometryPath::generateDecorations()` was updated so the the surface points are simply drawn and are no longer required to exist as ethereal/orphan `PathWrapPoint`s.
2.  Since the surface points can be drawn directly from the current path, there is no need for the cached `display_path` and all related methods and members have been removed, including from the OpenSimContext. @aymanhab please let me know if this is a problem for the GUI. I don't think PathWrapPoints were ever select-able or editable, if they were it would have had little effect since on the next frame they would have been replaced.

### DIFF
--- a/OpenSim/Simulation/Model/GeometryPath.cpp
+++ b/OpenSim/Simulation/Model/GeometryPath.cpp
@@ -144,9 +144,11 @@ generateDecorations(bool fixed, const ModelDisplayHints& hints,
         if (pwp) {
             // A PathWrapPoint provides points on the wrapping surface as Vec3s
             Array<Vec3>& surfacePoints = pwp->getWrapPath();
-            // The surface points are expressed w.r.t. the wrap surface's body 
+            // The surface points are expressed w.r.t. the wrap surface's body frame.
+            // Transform the surface points into the ground reference frame to draw
+            // the surface point as the wrapping portion of the GeometryPath
             const Transform& X_BG = pwp->getBody().getTransformInGround(state);
-            // Cycle through each surface point and draw it
+            // Cycle through each surface point and draw it the Ground frame
             for (int j = 0; j<surfacePoints.getSize(); ++j) {
                 // transform the surface point into the Ground reference frame
                 pos = X_BG*surfacePoints[j];
--- a/OpenSim/Simulation/Model/GeometryPath.cpp
+++ b/OpenSim/Simulation/Model/GeometryPath.cpp
@@ -136,28 +136,27 @@ generateDecorations(bool fixed, const ModelDisplayHints& hints,
         DefaultGeometry::drawPathPoint(mbix, lastPos, getColor(state), appendToThis);
 
     Vec3 pos;
-    for (int j = 1; j < pathPoints.getSize(); j++) {
-        PathPoint* point = pathPoints[j];
-        PathWrapPoint* mwp = dynamic_cast<PathWrapPoint*>(point);
+    int cnt = 0;
+    for (int i = 1; i < pathPoints.getSize(); ++i) {
+        PathPoint* point = pathPoints[i];
+        PathWrapPoint* pwp = dynamic_cast<PathWrapPoint*>(point);
 
-        if (mwp) {
-            // If the point is a PathWrapPoint and has surfacePoints,
-            // then this is the second of two tangent points for the
-            // wrap instance. So add the surface points to the display
-            // path before adding the second tangent point.
-            // Note: the first surface point is coincident with the
-            // first tangent point, so don't add it to the path.
-            Array<Vec3>& surfacePoints = mwp->getWrapPath();
-            const Transform& X_BG = mwp->getBody().getTransformInGround(state);
-            for (int j = 1; j<surfacePoints.getSize(); j++) {
-                pos = X_BG*surfacePoints.get(j);
+        if (pwp) {
+            // A PathWrapPoint provides points on the wrapping surface as Vec3s
+            Array<Vec3>& surfacePoints = pwp->getWrapPath();
+            // The surface points are expressed w.r.t. the wrap surface's body 
+            const Transform& X_BG = pwp->getBody().getTransformInGround(state);
+            // Cycle through each surface point and draw it
+            for (int j = 0; j<surfacePoints.getSize(); ++j) {
+                // transform the surface point into the Ground reference frame
+                pos = X_BG*surfacePoints[j];
                 if (hints.get_show_path_points())
                     DefaultGeometry::drawPathPoint(mbix, pos, getColor(state),
                         appendToThis);
                 // Line segments will be in ground frame
                 appendToThis.push_back(DecorativeLine(lastPos, pos)
                     .setLineThickness(4)
-                    .setColor(getColor(state)).setBodyId(0).setIndexOnBody(j));
+                    .setColor(getColor(state)).setBodyId(0).setIndexOnBody(cnt++));
                 lastPos = pos;
             }
         } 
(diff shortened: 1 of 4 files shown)


OpenSim/Simulation/Model/GeometryPath.cpp
```diff
@@ -124,44 +121,57 @@ generateDecorations(bool fixed, const ModelDisplayHints& hints,
...
+        PathWrapPoint* mwp = dynamic_cast<PathWrapPoint*>(point);
 
-        lastPos = pos;
+        if (mwp) {
+            // If the point is a PathWrapPoint and has surfacePoints,
+            // then this is the second of two tangent points for the
+            // wrap instance. So add the surface points to the display
+            // path before adding the second tangent point.
+            // Note: the first surface point is coincident with the
+            // first tangent point, so don't add it to the path.
```
tkuchida: The points are coincident so it may not matter algorithmically, but it isn't clear whether "so don't add it to the path" refers to the first surface point or the first tangent point.

OpenSim/Simulation/Model/GeometryPath.cpp
```diff
@@ -124,44 +121,57 @@ generateDecorations(bool fixed, const ModelDisplayHints& hints,
...
-        appendToThis.push_back(DecorativeLine(lastPos, pos)
-            .setLineThickness(4)
-            .setColor(getColor(state)).setBodyId(0).setIndexOnBody(j));
+    for (int j = 1; j < pathPoints.getSize(); j++) {
+        PathPoint* point = pathPoints[j];
+        PathWrapPoint* mwp = dynamic_cast<PathWrapPoint*>(point);
 
-        lastPos = pos;
+        if (mwp) {
+            // If the point is a PathWrapPoint and has surfacePoints,
```
aseth1: I had copied the code verbatim from the previous implementation in `updateDisplayPath` which was removed. I agree it is cryptic and will attempt to replace with something more intelligible.

OpenSim/Simulation/Model/GeometryPath.cpp
```diff
@@ -124,44 +121,57 @@ generateDecorations(bool fixed, const ModelDisplayHints& hints,
...
+            // then this is the second of two tangent points for the
+            // wrap instance. So add the surface points to the display
+            // path before adding the second tangent point.
+            // Note: the first surface point is coincident with the
+            // first tangent point, so don't add it to the path.
+            Array<Vec3>& surfacePoints = mwp->getWrapPath();
+            const Transform& X_BG = mwp->getBody().getTransformInGround(state);
+            for (int j = 1; j<surfacePoints.getSize(); j++) {
```
aseth1: Yiche! I missed that. Will fix it.

aseth1:
@tkuchida thank you for your review. Please let me know if the recent changes are sufficient to address the issues you raised.

@aymanhab I would appreciate your review since the drawing of `GeometryPath`  has changed and may have implications for the visualizer and/or GUI.

OpenSim/Simulation/Model/GeometryPath.cpp
```diff
@@ -136,28 +136,27 @@ generateDecorations(bool fixed, const ModelDisplayHints& hints,
...
-                pos = X_BG*surfacePoints.get(j);
+    int cnt = 0;
+    for (int i = 1; i < pathPoints.getSize(); ++i) {
+        PathPoint* point = pathPoints[i];
+        PathWrapPoint* pwp = dynamic_cast<PathWrapPoint*>(point);
+
+        if (pwp) {
+            // A PathWrapPoint provides points on the wrapping surface as Vec3s
+            Array<Vec3>& surfacePoints = pwp->getWrapPath();
+            // The surface points are expressed w.r.t. the wrap surface's body
```
tkuchida: Would prefer to see a more direct explanation of what the following line of code does. Perhaps something like "Get the body's transform w.r.t. ground because the surface points are expressed w.r.t. the wrap surface's body and we need to draw them in the ground frame".

OpenSim/Simulation/Model/GeometryPath.cpp
```diff
@@ -136,28 +136,27 @@ generateDecorations(bool fixed, const ModelDisplayHints& hints,
...
+        PathPoint* point = pathPoints[i];
+        PathWrapPoint* pwp = dynamic_cast<PathWrapPoint*>(point);
+
+        if (pwp) {
+            // A PathWrapPoint provides points on the wrapping surface as Vec3s
+            Array<Vec3>& surfacePoints = pwp->getWrapPath();
+            // The surface points are expressed w.r.t. the wrap surface's body 
+            const Transform& X_BG = pwp->getBody().getTransformInGround(state);
+            // Cycle through each surface point and draw it
+            for (int j = 0; j<surfacePoints.getSize(); ++j) {
```
tkuchida: This loop started at 1 before: are you now drawing overlapping points?

aymanhab:
@aseth1 Thanks for checking. Indeed, the render loop used to ask the GeometryPath object for pathpoints and was drawing them accordingly and that breaks since these are not PathPoints anymore but currently can't test the effect of this change on GUI since the GUI hasn't been updated to draw muscles/Path objects since they were refactored into separate components. I'd say if the API visualizer looks as before, in terms of visualizing a model with wrapping then merge this to move along and we'll revisit when we address muscle/path visualization. WrapPoints were computed and so were not selectable, and they needed to be recreated fast for interactive path editing, we may have used some caching to save time during visualization but that's all changing anyway.

OpenSim/Simulation/Model/GeometryPath.cpp
aseth1: My understanding  is that there are two "real" `PathWrapPoints` in a `GeometryPath` associated with a `WrapObject` and these are the "tangent points" where the path contacts and leaves the wrap surface. The path on the wrap surface is described by surface points (Vec3 in the WrapObject's Frame). Why this array is stuffed in one (the 2nd) of the `PathWrapPoint`'s is beyond me. The surface path apparently includes the "tangent" points and the first was not to being converted into an intermediate "PathWrapPoint" by starting the index at 1.

In the current version `PathWrapPoint` is not draw, instead only the surface points it contains are drawn. I think the new way is less confusing.

tkuchida: Ah, ok I think I understand now. Thanks for the explanation.

aseth1:
@aymanhab the rendering in the GUI will also rely on `generateDecorations` moving forward, correct?
Had `PathWrapPoint`s been selectable and editable that would have been an exercise in futility since in the next instance all those `PathWrapPoint`s would have been deleted! I am not sure how the previous implementation provided "fast interactive path editing" unless users were able to edit the display path independent of the actual path (which also seems scary to me). As far as I can tell the *display_path* was primarily being used to convert the array of surface points into PathWrapPoints since the drawing code just cycled through PathPoints and drew their locations. I believe removing these "ethereal" `PathWrapPoint`s is simpler and does not introduce/force any more computation at display time.

aymanhab:
That's the plan (to use generateDecorations) and to modify it as needed.

aseth1:
@aymanhab if you are OK with these changes please go ahead and merge otherwise let me know if you have any additional comments.

aseth1:
Thank you @aymanhab and @tkuchida

Issue #1225, opened 2016-08-31
Matlab crashing during multiple simulation due to Reporters/TimeSeriesTable


jimmyDunne:
[double_pendulum_markers.osim.zip](https://github.com/opensim-org/opensim-core/files/448366/double_pendulum_markers.osim.zip)

Running multiple forward simulations in a loop (with reporters) continually crashes Matlab. 

If I pass the function a model object, it will not run a second time (fwd_simulation function at bottom of post). 

```
import org.opensim.modeling.*

model = Model('double_pendulum_markers.osim');
time_interval = 0.01;
default_coordinates = [0.5 0.3];

for i = 1 : 100
    [coordinatetable, markertable] = fwd_simulation(model, time_interval, default_coordinates, false)
end
```
 On the second iteration I get this error;

```
Error using fwd_simulation (line 57)
Java exception occurred:
java.lang.RuntimeException: Timestamp at row 10001 with value 0.000000 is less-than/equal to
timestamp at row 10000 with value 10.000000
	In file
        /Users/jimmy/repository/opensim-master/opensim-core/OpenSim/Common/TimeSeriesTable.h:194
	In function 'validateRow'
	at org.opensim.modeling.opensimModelSimulationJNI.Manager_integrate__SWIG_1(Native Method)
	at org.opensim.modeling.Manager.integrate(Manager.java:183)
```

I am interpreting this as there is a pesky TimesSeriesTable hanging around somewhere, even though everything should be getting wiped at the end of the function call. 

I managed to get it running more than once by passing the string name for the model and creating the Model object on the side of the function

```
import org.opensim.modeling.*

time_interval = 0.01;
default_coordinates = [0.5 0.3];

for i = 1 : 100
    [coordinatetable, markertable] = fwd_simulation('double_pendulum_markers.osim', time_interval, default_coordinates, false)
end
```

This allows me to run the function more than once (great!), but it will crash Matlab after 10 iterations without warning or sufficient error message. 

If I do not create any reporters in the function, it will run continually (tested up to 100 iterations)

Something funky is going on with Reporters and tables, so if anyone can give it a test that would be super. 

```
function [coordinatetable, markertable] = fwd_simulation(modelname, time_interval, default_coordinates, plotMarkers)

import org.opensim.modeling.*

%% add a table reporter for the coordinates
coordinateReporter = TableReporter();
coordinateReporter.set_report_time_interval(0.001)
[...]

Issue #1220, opened 2016-08-30
Please add getTable to TableReporter's


jimmyDunne:
The TableReporter holds tables, so my instinct would be to ask something like getTable(). It took Chris telling me about getReport() to figure out that is what gives me a table. 

Since we don't have reports, we have tables, we should be calling the method getTable()

tkuchida:
Another potential challenge is that the method appears in the doxygen as
```cpp
const TimeSeriesTable_<ValueT>& OpenSim::TableReporter_< InputT, ValueT >::getReport( ) const
```
and has no comments.

jimmyDunne:
The only way I can figure out how to use the tables and adapters is to use methodsview() in Matlab, or search through the github repo, hoping there is an example of use. 

I understand that Macros and templates reduce code redundancy, but it sure makes it harder to find public methods.

jenhicks:
Change getReport() to getTable()

Issue #1068, opened 2016-06-26
[doxygen] Improve documentation for new classes (AbstractChannel, Input, etc.)


tkuchida:
The documentation for AbstractChannel, Channel, Input, TableReporter, etc. should explain what they are, how they relate to each other (e.g., "Channel" vs. "Output"), how they are used (including scripting examples, where appropriate), etc.

tkuchida:
There is also example code sprinkled throughout the doxygen that is now out of date. For example, due to #1181:
```cpp
// code like this
device->updInput("controller/activation").connect(
    hopper.getOutput("/hopper/vastus/activation"));
// changes to this
device->updComponent("controller").updInput("activation").connect(
    hopper.getComponent(vastus).getOutput("activation"));
```
Addressing #1118 may further change the code to something like this:
```cpp
Component::connect(device->updComponent("controller").updInput("activation"),
                   hopper.getComponent(vastus).getOutput("activation"));
```

PR #1265, opened 2016-09-19
Add clearReport() to TableReporter. Throw meaningful exception while appendRow.

klshrinidhi:
Addressing #1225.

This fix will allow Matlab code to clear the report at the end of each iteration when running simulations in a for loop. Not clearing the report will cause an exception as before but with a different, hopefully more meaningful message.


OpenSim/Common/Reporter.h
```diff
@@ -163,10 +163,24 @@ OpenSim_DECLARE_CONCRETE_OBJECT_T(TableReporter_, InputT, Reporter<InputT>);
     TableReporter_() = default;
     virtual ~TableReporter_() = default;
 
+    /** Retrieve the report.                                                  */
     const TimeSeriesTable_<ValueT>& getReport() const {
         return _outputTable;
     }
 
+    /** Clear the report. This can be used for example in loops where iteration
+    is intended to create a fresh report.                                     */
```
chrisdembia: +1 for the example, but I think the example can be a little clearer ("performing multiple simulations starting from the same initial time"). I think the comment should mention that the input will still be connected; that the channels set up are still wired up.

OpenSim/Common/Reporter.h
```diff
@@ -163,10 +163,24 @@ OpenSim_DECLARE_CONCRETE_OBJECT_T(TableReporter_, InputT, Reporter<InputT>);
...
+    void clearReport() {
+        auto columnLabels = _outputTable.getColumnLabels();
+        _outputTable = TimeSeriesTable_<ValueT>{};
+        _outputTable.setColumnLabels(columnLabels);
+    }
+
+    /** Same as clearReport().                                                */
+    void clearTable() {
+        clearReport();
+    }
```
chrisdembia: I don't think we need this method with the alternate name. If we don't like `clearReport()`, then we should also change `getReport()` to `getTable()`, as part of a separate PR.

OpenSim/Common/Reporter.h
```diff
@@ -177,7 +191,16 @@ OpenSim_DECLARE_CONCRETE_OBJECT_T(TableReporter_, InputT, Reporter<InputT>);
...
         }
-        const_cast<Self*>(this)->_outputTable.appendRow(state.getTime(), result);
+        try {
+            const_cast<Self*>(this)->_outputTable.appendRow(state.getTime(),
+                                                            result);
+        } catch(const InvalidTimestamp&) {
+            OPENSIM_THROW(Exception,
+                          "Attempting to update reporter with rows having "
+                          "invalid timestamps. Hint: If running simulation in "
+                          "a loop, use clearReport() to clear report at the end"
```
chrisdembia: Personal wording preference: change

```
use clearReport() to clear report
```

to either

```
call clearReport()
```

or

```
call clearReport() to clear data from previous simulation
```

OpenSim/Common/Test/testComponentInterface.cpp
```diff
@@ -830,6 +830,8 @@ void testListInputs() {
 
     cout << "  TableReporterMixedOutputs (contents)" << endl;
     cout << tabReporter->getReport() << endl;
+
+    tabReporter->clearReport();
```
chrisdembia: Could you add a simple one-line check after this to make sure the table is now empty?

```cpp
SimTK_TEST(tabReporter->getReport().getNumRows() == 0);
```

OpenSim/Common/Reporter.h
```diff
@@ -163,10 +163,24 @@ OpenSim_DECLARE_CONCRETE_OBJECT_T(TableReporter_, InputT, Reporter<InputT>);
     TableReporter_() = default;
     virtual ~TableReporter_() = default;
 
+    /** Retrieve the report.                                                  */
```
tkuchida: I think this user-facing documentation could be improved by providing an example that helps the user understand what `_<ValueT>` means. There is already an issue to track the lack of meaningful documentation (#1068), so we may just want to leave this for later.

OpenSim/Common/Reporter.h
```diff
@@ -177,7 +191,16 @@ OpenSim_DECLARE_CONCRETE_OBJECT_T(TableReporter_, InputT, Reporter<InputT>);
...
               result[idx] = value;
         }
-        const_cast<Self*>(this)->_outputTable.appendRow(state.getTime(), result);
+        try {
+            const_cast<Self*>(this)->_outputTable.appendRow(state.getTime(),
+                                                            result);
+        } catch(const InvalidTimestamp&) {
+            OPENSIM_THROW(Exception,
+                          "Attempting to update reporter with rows having "
+                          "invalid timestamps. Hint: If running simulation in "
```
tkuchida: What is the default "InvalidTimestamp" exception message? Does it tell the user what makes a timestamp invalid? If the only requirement is for the time to be non-decreasing, perhaps just say that here and/or report the current and previous times; otherwise, maybe the exception should be re-thrown and also caught elsewhere to get the default message as well—though I don't know whether that is good software design.

OpenSim/Common/Reporter.h
klshrinidhi: I have added a sentence or two more. 
I don't think this is the place to explain details about component connections. Documentation on connections (or connectors) should do that.

OpenSim/Common/Reporter.h
klshrinidhi: I don't see this function causing a problem. It is not in conflict with anything else. I will leave it as is.

OpenSim/Common/Reporter.h
klshrinidhi: I think the current comment is pretty clear. I will leave it as is.

OpenSim/Common/Reporter.h
klshrinidhi: The class documentation does say what the template argument `ValueT` is in one sentence.
Providing examples in documentation could be topic of another PR. I have not done anything about this here.

tkuchida: > Providing examples in documentation could be topic of another PR.

Ok, seems fine to consider later in #1068.

OpenSim/Common/Reporter.h
klshrinidhi: The default message from `TimeSeriesTable` will be:
```
Timestamp at row <new-row-index> with value <new-timestamp> is less-than/equal to timestamp at row <previous-row> with value <previous-row-timestamp>.
```
I have changed throw statement to *append* to the message from the caught exception rather than replace it.

tkuchida: Ok, LGTM :+1:

OpenSim/Common/Reporter.h
aseth1: Did we rule out the "clear on read" case, @tkuchida. I thought that was handy in most cases and would avoid having to manually call clearReport(). Performing multiple simulations in a loop without calling `getReport()` seems like a case we need to cover (thus an explicit clearReport()) but I think calling `getReport()` will be the typical use case. Let me know if this was already discussed between you guys and you decided to not do that.

tkuchida: @klshrinidhi and I discussed the clear-on-read philosophy. The potential issue is that some other process might want/need to read the report without clearing, in which case we'd need a sniff/read-without-clearing method. Throwing an exception if `time_to_append < time_last_written` and telling the user to call `clearReport()` if they're sure they want to overwrite the data seems like a good solution.

klshrinidhi: I have summarized the changes in issue #1225 after talking to Tom.

aseth1: @tkuchida thanks for the discussion details.

OpenSim/Common/Reporter.h
```diff
@@ -177,7 +192,17 @@ OpenSim_DECLARE_CONCRETE_OBJECT_T(TableReporter_, InputT, Reporter<InputT>);
               const auto& value = chan.getValue(state);
               result[idx] = value;
         }
-        const_cast<Self*>(this)->_outputTable.appendRow(state.getTime(), result);
+        try {
+            const_cast<Self*>(this)->_outputTable.appendRow(state.getTime(),
+                                                            result);
+        } catch(const InvalidTimestamp& exception) {
+            OPENSIM_THROW(Exception,
+                          std::string{exception.what()} +
```
aseth1: Minor comment, but it may be easier for the user to read/react to the main message first and append the actual lower-level exception as "details".

klshrinidhi: Done.

klshrinidhi:
Since it was a small addition, I have added `getTable()`  to `TableReporter` as per #1220 in this PR.

jenhicks:
@klshrinidhi has addressed all the comments. @aseth1 can you give a final look and merge if all looks ok? Thanks!

jenhicks:
@klshrinidhi has kept getReport and clearReport for now. One remaining question is whether to keep or remove.

chrisdembia:
Personally, I think having just `getReport()` is better than having both.

chrisdembia:
If we have both, users will think "what's the difference between a report and a table? Does the report come with additional information?"

jimmyDunne:
>Personally, I think having just getReport() is better than having both.

if you are going to pick one, I would suggest having getTable(). You don't get a report from a Reporter, you get a table.

aseth1:
My preference is to replace `getReport` and `clearReport` with `getTable` and `clearTable`. Since the method retrieves and operates on a Table and there is no thing of type Report, I believe it is the most straightforward.

jenhicks:
Based on input, let's have getTable() and clearTable() only. @klshrinidhi will this create any issues implementing? If it is more complicated, it could become a separate PR.

klshrinidhi:
> If we have both, users will think "What's the difference between a report and a table? Does the report come with additional information?"

Documentation says what each method does.

klshrinidhi:
My opinion is `getTable()` is good to have but `getReport()` does not hurt. So I think having both is okay. Documentation is clear enough to accommodate both methods.

I will remove `getReport()` anyway.

jenhicks:
Documentation is of course great, but if the interface is clear without documentation that seems preferable. @klshrinidhi if it's just a simple rename, please make the update.

jenhicks:
@klshrinidhi thanks! just saw your comment.

jenhicks:
Tests are passing. Is this ready to merge @chrisdembia @aseth1 ?

chrisdembia:
I've now "approved" the PR. I'll leave it the merging to you, @aseth1, in case you have more comments.

jimmyDunne:
This branch is really unstable in Matlab. I am getting segfaults almost at random. Noting it here since once this is merged it may effect master.

aseth1:
> This branch is really unstable in Matlab. I am getting segfaults almost at random. Noting it here since once this is merged it may effect master. 

That doesn't sound good. Are there particular calls that cause the segfault? Can you post or send a link to the code you were executing. We are suspecting the attempt at double deletion of the table owned by the reporter after it has been cleared but have to see your code to verify. 
Thanks!

OpenSim/Common/Reporter.h
```diff
@@ -163,10 +163,20 @@ OpenSim_DECLARE_CONCRETE_OBJECT_T(TableReporter_, InputT, Reporter<InputT>);
     TableReporter_() = default;
     virtual ~TableReporter_() = default;
 
-    const TimeSeriesTable_<ValueT>& getReport() const {
+    /** Retrieve the report as a TimeSeriesTable.                             */
+    const TimeSeriesTable_<ValueT>& getTable() const {
```
aseth1: It seems that this could be problematic- giving back a reference to the table that can be wiped out below (`clearTable()`).
I suspect that the crashes that @jimmyDunne is experiencing is because in Java/MATLAB the table reference into the Reporter becomes invalid after `clearTable()` and any use (even *clear* of the MATLAB variable?) would segfault. 

Would it be safer and more inline with expected use cases if `clearTable()` only cleared the data contents (leaving the table and meta data intact) then the reference would remain valid. Are there any downsides to just clearing the matrix data of the `TimeSeriesTable`?  I am curious if that might solve the issue that @jimmyDunne is running into.

klshrinidhi: Once PR #1288 is merged, I can debug this issue on Java side.

klshrinidhi:
I added Java tests to run the `realize()` step multiple times in a loop. I am able to clear the table and re-populate the report successfully.

jenhicks:
@klshrinidhi If I am understanding correctly, you added some java tests but this is not reproducing the issues that James reported? Is that correct? If so, any ideas on why this might be the case and thoughts on how to move forward?

klshrinidhi:
I think this PR addresses the issue reported in #1225.
The Java tests I added test the `clearTable()` functionality introduced by this PR in the use case described in #1225 -- being able use TableReporter in a loop.

OpenSim/Common/Reporter.h
jenhicks: @klshrinidhi To be a little more specific, I was wondering if you had done any more debugging/testing on the Java side in response to this conversation.

klshrinidhi: This is addressed by the Java tests I added.

jenhicks:
@klshrinidhi Thanks for clarifying. Would you like to have the PR reviewed and merged and then investigate the reported segfaults? Or would you like to have @jimmyDunne (or you could) investigate now?

klshrinidhi:
> @klshrinidhi Thanks for clarifying. Would you like to have the PR reviewed and merged and then investigate the reported segfaults? Or would you like to have @jimmyDunne (or you could) investigate now?

Either is fine with me. The important thing is that debugging the reported random seg-faults needs more context to proceed. I need enough information to reproduce the problem in Java/Python.

jenhicks:
I think the odd thing was that the segfaults seemed to happen at random. If I am remembering my conversation with @jimmyDunne correctly, he tried to figure out the specific conditions but couldn't find a pattern. @jimmyDunne Is that correct? Maybe you could show him the code you were using or describe in person?

aymanhab:
In my test of a similar memory corruption issue #1298 I added explicit
calls to System.gc (), and that reproduced the crashes.

jimmyDunne:
> specific conditions but couldn't find a pattern

Correct. I will play with this branch again and see what I come up with.

jimmyDunne:
[Matlab] clearTable works well. If you don't clearTable then you get the expected error message (though the error message refers to clearReport rather than clearTable). I am still getting segfaults but when running the below code, but that doesn't seem to have much to do with clearTable()

```
%% Read a model into memory, run a fwd simulation, 


import org.opensim.modeling.*


time_interval = 0.1;

model = Model('double_pendulum_markers.osim');
%  Add a console reporter to print the muscle fiber force and elbow angle.
reporter = ConsoleReporter();
reporter.set_report_time_interval(0.1);
reporter.updInput('inputs').connect(model.getCoordinateSet.get(0).getOutput('value'), 'pin1_angle' );
reporter.updInput('inputs').connect(model.getCoordinateSet.get(1).getOutput('value'), 'q2' );
model.addComponent(reporter);

%% add a table reporter for the coordinates
coordinateReporter = TableReporter();
coordinateReporter.set_report_time_interval(time_interval)
coordinateReporter.updInput('inputs').connect(model.getCoordinateSet.get(0).getOutput('value'), 'q1' );
coordinateReporter.updInput('inputs').connect(model.getCoordinateSet.get(1).getOutput('value'), 'q2' );
model.addComponent(coordinateReporter);

stofileadapter = STOFileAdapter();

for n = 1 : 10
   
    disp(['This is loop ' num2str(n) ''])
       
    % Simulate.
    state = model.initSystem();
    manager = Manager(model);
    manager.setInitialTime(0); manager.setFinalTime(n);
    manager.integrate(state);

    %%%%% If I comment out the next few cmd's, the for loop is fine. 
    %%%%% Otherwise, I generate a segfault. 
    % get the coordinate and marker tables
    coordinatetable = coordinateReporter.getTable; 
    % print coordinates to file
    stofileadapter.write(coordinatetable,'pendulum_coordinates.sto');
    clear coordinatetable
    %%%%%%%

    % clear the table
    coordinateReporter.clearTable ; 
    
end
```

chrisdembia:
To build on what @jimmyDunne reported, the following lines of his script can be commented out and MATLAB will still crash:

```
stofileadapter.write(coordinatetable,'pendulum_coordinates.sto');
clear coordinatetable
```

OpenSim/Common/Reporter.h
```diff
@@ -177,7 +187,16 @@ OpenSim_DECLARE_CONCRETE_OBJECT_T(TableReporter_, InputT, Reporter<InputT>);
...
         }
-        const_cast<Self*>(this)->_outputTable.appendRow(state.getTime(), result);
+        try {
+            const_cast<Self*>(this)->_outputTable.appendRow(state.getTime(),
+                                                            result);
+        } catch(const InvalidTimestamp& exception) {
+            OPENSIM_THROW(Exception,
+                          "Attempting to update reporter with rows having "
+                          "invalid timestamps. Hint: If running simulation in "
+                          "a loop, use clearReport() to clear report at the end"
```
klshrinidhi: I have changed *clearReport* to *clearTable*.

klshrinidhi:
I will try to reproduce the seg-fault in Java by re-writing the above snippet in Java.

jenhicks:
@aymanhab will create a c++ prototype as well. Can discuss more this afternoon.

aymanhab:
Plan is for @klshrinidhi to move initSystem inside the loop in Java and see if it reproduces teh crash. If not on first attempt, add System.gc(0 at end of loop iteration. If that still works will try with exact code from James. If these fail to find the problem will revisit.

klshrinidhi:
Sorry. I didn't realize yesterday that the call to `initSystem()` is already in the for loop.
I added `System.gc()` to the end of the loop and increased the number of iterations to 40. It did not reproduce the problem.

I will try to reproduce the snippet above in Java.

klshrinidhi:
I have added a Java test equivalent to the snippet above. It does not reproduce the seg-fault.

jimmyDunne:
It works!! I can run the pendulum sim in a loop without segfaults! I ran the loop 300 times and everything went great! Thanks!! 🎉 👏

aymanhab:
Any reason not to merge this PR @aseth1 @chrisdembia or @tkuchida ? I assume from your comment @jimmyDunne that the branch is stable, please let us know otherwise.

tkuchida:
@aymanhab The changes that were made since I approved LGTM.

jimmyDunne:
LGTM as well. The issues I was having have been solved. Thanks, @klshrinidhi

aseth1:
Awesome. I am glad it is working! 
I am not a reviewer on this PR, but I am curios what was causing the segfaults and what changes stopped them from occurring? Thanks.

jenhicks:
@klshrinidhi I think this PR can be merged once conflicts are resolved.

klshrinidhi:
Tests pass. Merging.

PR #827, opened 2016-03-13
Convert Thelen2003Muscle properties into private subcomponents

tkuchida:
Converted `MuscleFixedWidthPennationModel` and `MuscleFirstOrderActivationDynamicModel` properties into private (non-serialized) subcomponents of `Thelen2003Muscle`. The properties of these subcomponents are overridden by `Thelen2003Muscle` in `extendFinalizeFromProperties()`. Test cases updated/augmented.

Fixes #687

Issue #687, opened 2015-11-02
Bugs in properties with Thelen2003Muscle (optimal_fiber_length and activation_time_constant)


chrisdembia:
Two critical bugs:

1. The `optimal_fiber_length` property exists in both Muscle and in MuscleFixedWidthPennationModel, so the Thelen2003Muscle has this property twice. Somehow, an old model gets the correct value in both places (when the model is updated).
2. The `activation_time_constant` and `deactivation_time_constant` properties were moved from Thelen2003Muscle to MuscleFirstOrderActivationDynamicModel *but are not copied over when updating the model file*. Thus, the updated model just has the default time constants.

Before:

```xml
				<Thelen2003Muscle name="hamstrings_r">
					<!--Flag indicating whether the force is disabled or not. Disabled means that the force is not active in subsequent dynamics realizations.-->
					<isDisabled>false</isDisabled>
					<!--Minimum allowed value for control signal. Used primarily when solving for control values.-->
					<min_control>0</min_control>
					<!--Maximum allowed value for control signal. Used primarily when solving for control values.-->
					<max_control>1</max_control>
					<!--The set of points defining the path of the muscle.-->
					<GeometryPath>
						<!--The set of points defining the path-->
						<PathPointSet>
							<objects>
								<PathPoint name="bifemlh_r-P1">
									<location> -0.129056 -0.105091 0.0711468</location>
									<body>pelvis</body>
								</PathPoint>
								<PathPoint name="bifemlh_r-P2">
									<location> -0.0297545 -0.0355868 0.0290922</location>
									<body>tibia_r</body>
								</PathPoint>
								<PathPoint name="bifemlh_r-P3">
									<location> -0.0231314 -0.0556538 0.0339063</location>
									<body>tibia_r</body>
								</PathPoint>
							</objects>
							<groups />
						</PathPointSet>
						<!--The wrap objecs that are associated with this path-->
						<PathWrapSet>
							<objects />
							<groups />
						</PathWrapSet>
						<!--Used to display the path in the 3D window-->
						<VisibleObject name="display">
							<!--Set of geometry files and associated attributes, allow .vtp, .stl, .obj-->
							<GeometrySet>
								<objects />
								<groups />
							</GeometrySet>
							<!--Three scale factors for display purposes: scaleX scaleY scaleZ-->
							<scale_factors> 1 1 1</scale_factors>
[...]

PR #1266, opened 2016-09-20
Updating Millard2012EquilibriumMuscle to Component API

tkuchida:
### Background
In #827, the pennation and activation dynamics models in Thelen2003Muscle were converted into private (non-serialized) subcomponents of the muscle, in part to avoid mishandling properties (see #687). The present PR comprises an analogous update to Millard2012EquilibriumMuscle.

### Primary changes
- Converted pennation and activation dynamics models into private subcomponents, whose properties are overridden by the muscle's properties in `extendFinalizeFromProperties()`. Added `maximum_pennation_angle` property to muscle as a result.
- Moved all property value checks into `extendFinalizeFromProperties()` to avoid bypassing checks by calling `set_property_name()` rather than `setPropertyName()`. Retained existing `setPropertyName()` methods for backwards compatibility (they now simply call `set_property_name()`).

### Secondary changes
- Removed `buildMuscle()` method, which was called only once.
- Removed property value checks that are already handled by subcomponents.
- Updated component count in test_component_interface.py to account for new subcomponents.

### Exclusions
- The muscle still owns the activation state variable. Ideally, this state would be owned by the activation dynamics model and would be allocated only if activation dynamics are being computed. The eventual plan is to begin using the ZerothOrderMuscleActivationDynamics and FirstOrderMuscleActivationDynamics classes (which own 0 and 1 states, respectively). This change would also allow users to easily create new activation dynamics models (perhaps fatigue models as well) and plug them into the muscle without creating a new muscle model. The present PR does not address that issue.


OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -878,9 +838,10 @@ calcMuscleDynamicsInfo(const SimTK::State& s, MuscleDynamicsInfo& mdi) const
         // Compute dynamic quantities.
         double a = SimTK::NaN;
         if(!get_ignore_activation_dynamics()) {
-            a = clampActivation(getStateVariableValue(s, STATE_ACTIVATION_NAME));
+            a = getActivationModel().clampActivation(
+                    getStateVariableValue(s, STATE_ACTIVATION_NAME));
```
aseth1: now with a separate activationModel it would make sense for the activation state to be part of it and also take its activation time constant properties with it.

OpenSim/Actuators/Millard2012EquilibriumMuscle.h
```diff
@@ -712,11 +703,13 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Millard2012EquilibriumMuscle, Muscle);
...
-    // This object defines the pennation model used for this muscle model. Using
-    // a ClonePtr saves us from having to write a destructor, copy constructor,
-    // or copy assignment. This will be zeroed on construction, cleaned up on
-    // destruction, and cloned when copying.
-    MuscleFixedWidthPennationModel penMdl;
+    // Subcomponents owned by the muscle. The properties of these subcomponents
+    // are set in extendFinalizeFromProperties() from the properties of the
+    // muscle.
+    MemberSubcomponentIndex penMdlIdx{
+      constructSubcomponent<MuscleFixedWidthPennationModel>("penMdl") };
```
aseth1: I see you mentioned these in your exclusions! You can ignore my comments

aseth1:
I totally agree with your exclusions. You can pretty much ignore my comments which repeat what you already mentioned. Thank you for these improvements!

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -812,15 +771,16 @@ calcFiberVelocityInfo(const SimTK::State& s, FiberVelocityInfo& fvi) const
...
-            dlce, mli.sinPennationAngle, mli.cosPennationAngle, dphidt);
+        double dlceAT = getPennationModel().calcFiberVelocityAlongTendon(
+            mli.fiberLength, dlce, mli.sinPennationAngle, mli.cosPennationAngle,
+            dphidt);
         double dmcldt = getLengtheningSpeed(s);
         double dtl = 0;
 
         if(!get_ignore_tendon_compliance()) {
-            dtl = penMdl.calcTendonVelocity(mli.cosPennationAngle,
+            dtl = getPennationModel().calcTendonVelocity(mli.cosPennationAngle,
```
tkuchida: I agree that this PR does not address all the design issues. The primary objective with this PR was to ensure that users don't encounter issues due to (1) properties being out of sync between the muscle and its helper Components, or (2) inadvertently bypassing property value checks by calling the `set_property_name()` methods instead of the value-checking `setPropertyName()` methods.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
tkuchida: I agree. Please read the "Exclusions" paragraph at the top. I have already built a state-owning activation dynamics model so, in my opinion, it would be a waste of time to modify the old model to own the activation state. In my opinion, the next logical step is to swap in the new model.

aseth1: I did read your exclusions (see my last comment at the bottom) sorry about that!

chrisdembia:
> This change would also allow users to easily create new activation dynamics models (perhaps fatigue models as well) and plug them into the muscle without creating a new muscle model.

I have mentioned this before, but I suppose people agree with me: I think that the Millard muscle class should not be customizable. People may read a paper that says "we used the Millard muscle in OpenSim (Millard et al., 2013)." A reader will then read that paper and think they know what activation dynamics model was used. But in fact the activation model could have been something else. 

I am fully in support of other muscle classes that do support customization.

OpenSim/Actuators/Millard2012EquilibriumMuscle.h
tkuchida: I have not claimed that the design is "desirable", but there were several practical reasons for making these private subcomponents and not removing the original properties of Millard2012EquilibriumMuscle; here are two:

1. The design philosophy of the Millard2012EquilibriumMuscle class is focused around a particular type of muscle model (one fiber, one tendon, etc.). The muscle can operate in a few configurations and its curves can be adjusted, but the class was not designed in a modular fashion. I think that modifying this class would be more work than creating a new model that has been designed to be modular.
2. Tests use `Object::RenameType()`, which assumes that the properties of Millard2012EquilibriumMuscle have the same names as those of Thelen2003Muscle. I disagree with using `Object::RenameType()` in tests, and there should be exceptions thrown if unexpected xml elements are encountered during deserialization; however, changing the tests and deserializer would be more work than is reasonable for this PR.

This PR improves how properties are handled; it is not a redesign of the muscle.

aseth1:
> ... "we used the Millard muscle in OpenSim (Millard et al., 2013)." ...

That is a good reason for keeping the MillardMuscle as unmodifiable (with activation and pennation models fixed internally as member subcomponents). 

However,  the model already has many knobs to ignore tendon compliance and to turn on/off or change damping, and all the curve parameters are editable, and so it remains insufficient to just cite Millard et al. 2013 without providing the parameter set used. It may then be sufficient that the default activation dynamics model and pennation model are consistent with the paper. If someone wants to mix and match models, I don't think we should necessarily prevent that.

tkuchida:
> I have mentioned this before, but I suppose people agree with me: I think that the Millard muscle class should not be customizable. People may read a paper that says "we used the Millard muscle in OpenSim (Millard et al., 2013)." A reader will then read that paper and think they know what activation dynamics model was used. But in fact the activation model could have been something else.

> I am fully in support of other muscle classes that do support customization.

What modifications are you requesting? The Millard2012EquilibriumMuscle class always permitted users to modify the muscle curves, which violates your desire for an uncustomizable model. Users are also able to adjust/ignore tendon compliance, adjust/ignore activation dynamics, etc., which may go unreported in a publication.

This PR does not change the extent to which the muscle can be customized.

chrisdembia:
> all the curve parameters are editable

Good point... a large part of the Millard model in my opinion *is* the specific definition of the curves (to match experimental data), so I agree with you.

@tkuchida @aseth1 What makes the Millard muscle the Millard muscle, then?

> What modifications are you requesting?

I am not requesting modifications. I try to make it very clear if I am suggesting modifications. I'll try to make it clearer when I am not suggesting modifications. I support this PR, and my point was that the "exclusions" should remain as such.

tkuchida:
> What makes the Millard muscle the Millard muscle, then?

I don't think there is such an entity as "the Millard muscle": there is the Millard2012EquilibriumMuscle class, which allows users to model single-fiber, single-tendon muscles with default/adjusted characteristic curves, with/without tendon compliance, with/without fiber damping, etc. If these parameters are not specified, I would assume "the Millard muscle" refers to the Millard2012EquilibriumMuscle class with the default parameters as of the date of publication.

> my point was that the "exclusions" should remain as such

Okay, I will summarize everyone's comments in a DevProp for a new, customizable muscle model.

chrisdembia:
> Okay, I will summarize everyone's comments in a DevProp for a new, customizable muscle model.

Maybe hold off...I could be convinced that the Millard2012EquilibriumMuscle could be more customizable. Perhaps an in-person discussion should occur before a DevProp. It seems that others may have a different mental model than I do for these...models.

aseth1:
> @tkuchida @aseth1 What makes the Millard muscle the Millard muscle, then?

I would say it is primarily the use of fiber damping to avoid singularity at low (zero) activation and its parameterization of muscle curves (which were used to get a reasonable match to physiological data).

aseth1:
@chrisdembia I am happy to merge this. You?  We agree that these change are improvements.  

Where these changes will ultimately lead seems open. Modular Muscle model(s) are not on the 4.0 priority list so we should document ideas/potential problems in an issue and revisit. I don't think a DevProp is necessary at this time.

chrisdembia:
I know I joined in on the discussion, but I have not had a chance to look at the code changes in this PR. Feel free to merge without my review, though.

aseth1:
@chrisdembia Not a problem. I was not sure. Please review at your convenience.

tkuchida:
> Maybe hold off...I could be convinced that the Millard2012EquilibriumMuscle could be more customizable.

Ok, let me know what you and Ajay decide and whether you want it in 4.0. This PR leaves the muscle in an improved and stable condition, so I would be fine with leaving it for now.

jenhicks:
My vote is to merge this PR and document discussion in an issue. I also don't think a DevProp is high priority at this time, given other tasks on our plate for 4.0. Thanks for making the improvements @tkuchida

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -49,132 +49,126 @@ void Millard2012EquilibriumMuscle::constructProperties()
...
+    // Set properties of pennation model subcomponent. Values of
+    // optimal_fiber_length, pennation_angle_at_optimal, and
+    // maximum_pennation_angle are checked by
+    // MuscleFixedWidthPennationModel::extendFinalizeFromProperties().
+    auto& penMdl =
+        updMemberSubcomponent<MuscleFixedWidthPennationModel>(penMdlIdx);
+    penMdl.set_optimal_fiber_length(getOptimalFiberLength());
+    penMdl.set_pennation_angle_at_optimal(getPennationAngleAtOptimalFiberLength());
+    penMdl.set_maximum_pennation_angle(get_maximum_pennation_angle());
+    penMdl.finalizeFromProperties();
```
chrisdembia: I believe `Component::componentsFinalizeFromProperties()` takes care of invoking `finalizeFromProperties()` on the pennation model (and the activation model).

chrisdembia:
@tkuchida I made one small comment. Other than that, this looks good to me.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
chrisdembia: If possible, we should remove these calls so that we don't invoke it twice.

tkuchida: Yes, `finalizeFromProperties()` will be called on the pennation model automatically, but I need it to be called between L146 (when the properties are set) and L165 (when `getMinimumFiberLength()` is called) because `getMinimumFiberLength()` returns `m_minimumFiberLength` [which is set in `MuscleFixedWidthPennationModel::extendFinalizeFromProperties()`](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Actuators/MuscleFixedWidthPennationModel.cpp#L103).

Perhaps I should move `penMdl.finalizeFromProperties();` down (immediately before L165) and add a comment about why it needs to be called explicitly. I think I included `actMdl.finalizeFromProperties();` on L159 for symmetry with the pennation model, but that line can be removed.

chrisdembia: Ah okay. Yeah I think it is fine to just add a comment on that line.

tkuchida: I moved `penMdl.finalizeFromProperties();` down to the line that required it to be called, added a comment explaining as much, and removed `actMdl.finalizeFromProperties();` which was indeed unnecessary since it's called automatically by the Component machinery.

Issue #1263, opened 2016-09-19
Improve evaluation of memory usage in testInitState


tkuchida:
See [testInitState.cpp L156](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Test/testInitState.cpp#L156).

### Issue

If memory is freed behind the scenes immediately after L156, then `mem0` will not represent the "base footprint" as assumed and the evaluation of memory usage will be ineffective (currently, if `current_size < mem0`, we set `model_size = 0` and leaks could slip through). We also don't want to throw an exception if `current_size < mem0`, as we already have lots of false alarms on CI (e.g., testContactGeometry).

### Suggestions

1. Perhaps the entire operation of setting `mem0`, loading a model, and computing the memory increase should be repeated until a nonzero memory increase is detected.

2. If the model is small, perhaps it can always fit within the memory already allocated (I don't know the details on every platform). I suggest loading *n* copies of the model and then dividing the memory increase by *n*. Perhaps use the following algorithm:

```
n <- 1
while (1) {
    mem0 <- current memory footprint
    loadNCopiesOfModel(n)
    mem1 <- current memory footprint
    if (mem1 > mem0)
        break
    ++n
}
memoryIncrease <- (mem1-mem0) / n
```

tkuchida:
An analogous issue exists in testComponents.cpp, which would benefit from an analogous solution.

aseth1:
Both seem reasonable to me, but 2. does not address the problem if `mem0` is inaccurate.  In which case we will loop N times and remain unsuccessful. This is what is currently happening in #1205.

> We also don't want to throw an exception if current_size < mem0, as we already have lots of false alarms on CI (e.g., testContactGeometry).

What if you have looped through N times and it still evaluates to less than mem0? 

We could bake in the memory estimate of a model based on the mean (or median) value (over N samples) by adding a utility auxiliary testing so it could be used by other test. The median value would be more accurate if a few of the evaluations are completely wrong (like 0 or even negative). We could also evaluate mem0 in a similar way,  estimating over N calls with a brief/mandatory wait. 

@tkuchida are you tackling this now? Otherwise, I will do it.

tkuchida:
> but 2. does not address the problem if `mem0` is inaccurate.  In which case we will loop N times and remain unsuccessful.

[...]

PR #1267, opened 2016-09-20
Improve memory usage estimates and apply to testInitStates

aseth1:
This PR addresses #1263 (arising from the current failure in #1205) and intermittent failures in CI of `testInitStates.cpp`. The main issue is that a single call to `getCurrentRSS()` can be unreliable, but it is generally consistent over repeated calls (at least on Windows). 

Therefore, when evaluating the memory usage of statement it makes sense to make several samples and to use the samples to select the typically "expected" value. Here we use the median value to represent the typical memory usage as estimated from `getCurrentRSS()`. The utility method wraps the *command* in a loop and estimates the change in memory use due to executing the command. The utility is `estimateMemoryChangeForCommand(T command, const size_t nSamples)` and is located in `auxiliaryTestFunctions.h` so other tests (such as `testComponents`) can also use it and be simplified. 

`testInitState` was simplified to use `estimateMemoryChangeForCommand` instead of employing its own loops in hopes of addressing #1205.

A method was also added to check that getRSS is valid by computing the change in memory usage after explicitly allocating a block of known size. It is currently being used as a guard in `testInitStates` to decide whether memory usage should even be tested.  Currently, if it is invalid the test are skipped. I am looking for reviewer comments on whether something like this should stay or not. Leaving it may cause the the memory usage to never be tested on Mac or Linux.

Adapting `testComponents` (and any other tests) will be part of subsequent PRs.

### DIFF
--- a/OpenSim/Auxiliary/auxiliaryTestFunctions.h
+++ b/OpenSim/Auxiliary/auxiliaryTestFunctions.h
@@ -286,7 +286,15 @@ void validateMemoryUseEstimates(const size_t nSamples = 11)
         return new Block(size);
     };
 
-    size_t delta = estimateMemoryChangeForCreator<Block>(creator, nSamples);
+    size_t delta = 0;
+    try {
+        delta = estimateMemoryChangeForCreator<Block>(creator, nSamples);
+    }
+    catch (const std::exception& ex) {
+        OPENSIM_THROW(OpenSim::Exception,
+            "Failed to estimate change in memory usage. Details:\n"
+            + std::string(ex.what()) );
+    }
 
     OPENSIM_THROW_IF(delta < size/2, OpenSim::Exception,
         "Cannot estimate memory usage due to invalid getRSS() evaluation."
--- a/OpenSim/Auxiliary/auxiliaryTestFunctions.h
+++ b/OpenSim/Auxiliary/auxiliaryTestFunctions.h
@@ -27,7 +27,6 @@
 #include <OpenSim/Common/Function.h>
 #include <OpenSim/Common/LinearFunction.h>
 #include <OpenSim/Common/PropertyObjArray.h>
-#include <OpenSim/Simulation/Model/Model.h>
 #include "getRSS.h"
 
 #include <fstream>
@@ -210,68 +209,35 @@ OpenSim::Object* randomize(OpenSim::Object* obj)
 // Storage can only read files with version <= 1.
 // This function can be removed when Storage class is removed.
 inline void revertToVersionNumber1(const std::string& filenameOld,
-                                   const std::string& filenameNew) {
-  std::regex versionline{R"([ \t]*version[ \t]*=[ \t]*\d[ \t]*)"};
-  std::ifstream fileOld{filenameOld};
-  std::ofstream fileNew{filenameNew};
-  std::string line{};
-  while(std::getline(fileOld, line)) {
-    if(std::regex_match(line, versionline))
-      fileNew << "version=1\n";
-    else
-      fileNew << line << "\n";
-  }
-}
-
-bool isGetRSSValid() {
-    auto s = sizeof(char);
-    // chat is a byte on ALL platforms
-    ASSERT(s == 1);
-
-    size_t mem0 = getCurrentRSS();
-
-    size_t size = 20 * 1024; // 20 * 1KB;
-    // This should yield a change in memory usage of exactly 20KB;
-    char* charBlock = (char*)std::malloc(size * s);
-
-    // do something with charBlock so that compiler does not optimize it away
-    for (size_t i = 1; i < size; ++i) {
-        charBlock[i] = charBlock[i - 1];
+    const std::string& filenameNew) {
+    std::regex versionline{ R"([ \t]*version[ \t]*=[ \t]*\d[ \t]*)" };
+    std::ifstream fileOld{ filenameOld };
+    std::ofstream fileNew{ filenameNew };
+    std::string line{};
+    while (std::getline(fileOld, line)) {
+        if (std::regex_match(line, versionline))
+            fileNew << "version=1\n";
+        else
+            fileNew << line << "\n";
     }
-
-    size_t mem1 = mem0;
-    size_t count = 0;
-
-    while(mem1 == mem0) {
-        mem1 = getCurrentRSS();
-        std::this_thread::sleep_for(std::chrono::milliseconds(5));
-        ++count;
-    }
-
-    std::cout << "Took " << count << " tries to register change in getRSS()" << std::endl;
-    std::cout << "mem1=" << mem1 << "  mem0=" << mem0 << " | " << mem1 - mem0 << std::endl;
-
-    free(charBlock);
-
-    return (mem1-mem0) == size;
 }
-
 
 // Estimate the change in memory usage for a given command
 template <typename T>
-size_t estimateMemoryChangeForCommand(T command, const size_t nSamples = 100) {
+size_t estimateMemoryChangeForCommand(T command, const size_t nSamples = 100)
+{
     std::vector<size_t> deltas;
+    std::cout << "command of type: " << typeid(T).name() << std::endl;
 
     for (size_t i = 0; i < nSamples; ++i) {
         size_t mem0 = getCurrentRSS();
-
         // Execute the desired command
         command();
-
+        // wait a millisecond to allow memory usage to settle
         std::this_thread::sleep_for(std::chrono::milliseconds(1));
+        // poll the change in memory usage
         size_t mem1 = getCurrentRSS();
         deltas.push_back(mem1 > mem0 ? mem1 - mem0 : 0);
-        //std::cout << "i: " << i << " | " << "delta = " << deltas[i] << std::endl;
     }
 
     size_t nmedian = nSamples / 2;
@@ -283,17 +249,22 @@ size_t estimateMemoryChangeForCommand(T command, const size_t nSamples = 100) {
     return deltas[nmedian];
 }
 
-// Estimate the change in memory caused by loading a model
-size_t estimateMemoryChangeForModelLoading( const std::string& modelFileName,
-    size_t nSamples) {
-
-    auto command = [modelFileName]() {
-        OpenSim::Model model(modelFileName);
-        model.finalizeFromProperties();
+bool isGetRSSValid() 
+{
+    size_t size = 20 * 1024; // 20 * 1KB;
+    // This should yield a change in memory usage of exactly 20KB;
+    char* charBlock = nullptr;
+    auto command = [size, charBlock]() {
+        // This should yield a change in memory usage of exactly 20KB;
+        char* charBlock = (char*)std::malloc(size * sizeof(char));
+        for (size_t i = 1; i < size; ++i) {
+            charBlock[i] = charBlock[i - 1];
+        }
     };
-    return estimateMemoryChangeForCommand(command, nSamples);
+    size_t delta = estimateMemoryChangeForCommand(command, 10);
 
+    free(charBlock);
+    return delta == size;
 }
-
 
 #endif // OPENSIM_AUXILIARY_TEST_FUNCTIONS_H_
--- a/OpenSim/Auxiliary/auxiliaryTestFunctions.h
+++ b/OpenSim/Auxiliary/auxiliaryTestFunctions.h
@@ -312,10 +312,18 @@ size_t estimateMemoryChangeForCommand(T command, const size_t nSamples = 100)
         size_t mem0 = getCurrentRSS();
         // Execute the desired command
         command();
-        // wait a millisecond to allow memory usage to settle
-        std::this_thread::sleep_for(std::chrono::milliseconds(1));
-        // poll the change in memory usage
-        size_t mem1 = getCurrentRSS();
+        // initialize post-command memory usage to an error causing size
+        size_t mem1 = std::numeric_limits<std::size_t>::max();
+        int cnt = 0;
+        // wait up to 100ms total for memory usage to settle
+        do {
+            // poll the change in memory usage
+            mem1 = getCurrentRSS();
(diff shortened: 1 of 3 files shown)


OpenSim/Auxiliary/auxiliaryTestFunctions.h
```diff
@@ -209,17 +209,62 @@ OpenSim::Object* randomize(OpenSim::Object* obj)
...
+    return deltas[nmedian];
+}
+
+bool isGetRSSValid() 
+{
+    size_t size = 20 * 1024; // 20 * 1KB;
+    // This should yield a change in memory usage of exactly 20KB;
+    char* charBlock = nullptr;
+    auto command = [size, charBlock]() {
+        // This should yield a change in memory usage of exactly 20KB;
```
tkuchida: I don't think it's necessary to include the same comment here and on L255.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
```diff
@@ -209,17 +209,62 @@ OpenSim::Object* randomize(OpenSim::Object* obj)
...
+    std::string line{};
+    while (std::getline(fileOld, line)) {
+        if (std::regex_match(line, versionline))
+            fileNew << "version=1\n";
+        else
+            fileNew << line << "\n";
+    }
+}
+
+// Estimate the change in memory usage for a given command
```
tkuchida: Does [getCurrentRSS()](http://nadeausoftware.com/articles/2012/07/c_c_tip_how_get_process_resident_set_size_physical_memory_use#HowtogettheresidentsetsizeRSS) measure only heap space? Perhaps provide a clarifying comment here.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
```diff
@@ -209,17 +209,62 @@ OpenSim::Object* randomize(OpenSim::Object* obj)
...
+template <typename T>
+size_t estimateMemoryChangeForCommand(T command, const size_t nSamples = 100)
+{
+    std::vector<size_t> deltas;
+    std::cout << "command of type: " << typeid(T).name() << std::endl;
+
+    for (size_t i = 0; i < nSamples; ++i) {
+        size_t mem0 = getCurrentRSS();
+        // Execute the desired command
+        command();
```
tkuchida: If we're measuring the increase in heap space between L233 and L239, aren't we leaking memory here? Does `command()` need to return a pointer so the allocated memory can be freed?

OpenSim/Simulation/Test/testInitState.cpp
```diff
@@ -50,8 +50,10 @@ int main()
     try {
         LoadOpenSimLibrary("osimActuators");
         testStates("arm26.osim");
-        testMemoryUsage("arm26.osim");
-        testMemoryUsage("PushUpToesOnGroundWithMuscles.osim");
+        if (isGetRSSValid()) {
```
tkuchida: I don't think this is a good idea, as it will just silently skip the tests if `!isGetRSSValid()`. My understanding is that part of the motivation behind his PR was to be more strict about testing memory usage (e.g., not just allowing the test to pass if `mem1 < mem0`). If you want to check whether `getCurrentRSS()` is operating as expected, I suggest moving this part of the test to a new test case and throwing an exception if `!isGetRSSValid()`. We could then decide to merge a PR even if `getCurrentRSS()` malfunctions, but at least we would have that information.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
aseth1: I don't believe `getRSS` can distinguish between stack and heap memory. I think it can only observe the memory associate with the current process.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
aseth1: True.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
tkuchida: Ah, okay. Seems fine then.

OpenSim/Simulation/Test/testInitState.cpp
aseth1: I tend to agree @tkuchida. The flip-side is: should CI testing *fail* because `getRSS` is flaking out?

I do prefer to have the explicit exception as you suggest and then the reviewers could make the call to merge a change or not based on the test passing/failing on other platforms. 

If we do break validity of getRSS out into a separate test it may be possible that `isGetRSSValid()` is then true but had it been evaluated in another process (the actual memory test) it would have been false. Then, another option is to bake the `isGetRSSValid()` check into every memory evaluation and throw a `getRSS() is not currently valid` exception.  As a reviewer you would not have to determine if the test failed because of a legitimate leak or some quirk with getRSS() - our instrumentation would tell you it was getRSS(). It could potentially result in more failures, but it would be more clear which ones can be ignored.

I mainly want to avoid having to rebuild PRs just to get these memory tests to pass on all platforms. If it continues to be a problem, we should consider just having it on Windows and/or use other tools on Linux/Mac.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
```diff
@@ -209,17 +209,62 @@ OpenSim::Object* randomize(OpenSim::Object* obj)
...
+        // wait a millisecond to allow memory usage to settle
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+        // poll the change in memory usage
+        size_t mem1 = getCurrentRSS();
+        deltas.push_back(mem1 > mem0 ? mem1 - mem0 : 0);
+    }
+
+    size_t nmedian = nSamples / 2;
+    std::nth_element(deltas.begin(), deltas.begin() + nmedian, deltas.end());
```
chrisdembia: Consider adding a comment before this function call explaining that this reorders the elements of `deltas`.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
```diff
@@ -209,17 +209,62 @@ OpenSim::Object* randomize(OpenSim::Object* obj)
...
+
+bool isGetRSSValid() 
+{
+    size_t size = 20 * 1024; // 20 * 1KB;
+    // This should yield a change in memory usage of exactly 20KB;
+    char* charBlock = nullptr;
+    auto command = [size, charBlock]() {
+        // This should yield a change in memory usage of exactly 20KB;
+        char* charBlock = (char*)std::malloc(size * sizeof(char));
+        for (size_t i = 1; i < size; ++i) {
```
chrisdembia: Is this to avoid `unused variable` warnings, etc.? If so, could you add a comment?

aseth1:
Thank you @tkuchida and @chrisdembia for your reviews. I agree with your comments and suggestions. I attempted appropriate changes:
 1. no more `isGetRSSValid()`. Instead there is a `validateMemoryUseEstimates()` that gets called at the beginning of a test case (one that will require memory estimates like in `testInitState.cpp`'s  `testMemoryUsage()`).
 2. created a separate estimator for "creators" (`estimateMemoryChangeForCreator`) so that memory management is part of the testing function and not up to the tester.
 3. Have `validateMemoryUseEstimates` use `estimateMemoryChangeForCreator` with a known sized `Block`.
 4. Added more procedural and usage comments
 5. Chose defaults for sizes and number of samples that provide repeatable/robust performance (on Windows).
 6. re-enabled `testInitState` for OSX to see if the these fixes also address issues on OSX

I look forward to your comments.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
```diff
@@ -209,17 +209,116 @@ OpenSim::Object* randomize(OpenSim::Object* obj)
...
+
+// Determine if getRSS is providing reliable estimates of memory usage by
+// testing against an allocation of known size and verifying that the change
+// memory use is detected. Employ this method to validate the use of
+// memory use estimators (e.g. estimateMemoryChangeForCreator and
+// estimateMemoryChangeForCommand). Do this at the beginning of your test 
+// involving checks for memory use.
+void validateMemoryUseEstimates()
+{
+    size_t nSamples = 5;
```
tkuchida: Should `nSamples` be higher? [testInitState currently fails on OSX](https://travis-ci.org/opensim-org/opensim-core/jobs/161733598#L2246). The exception message isn't sufficiently detailed to determine what failed, but perhaps more than 5 samples are necessary to get at least 2 valid measures (by default, `estimateMemoryChangeForCreator()` requires only 2 of 100).

Perhaps make `nSamples` a default argument for symmetry with estimateMemoryChangeForC{reator|ommand}().

OpenSim/Auxiliary/auxiliaryTestFunctions.h
```diff
@@ -209,17 +209,116 @@ OpenSim::Object* randomize(OpenSim::Object* obj)
...
+        }
+        // member is pointer to allocated block
+        char* p{};
+    };
+
+    auto creator = [size]() { 
+        return new Block(size);
+    };
+
+    size_t delta = estimateMemoryChangeForCreator<Block>(creator, nSamples);
```
tkuchida: I suggest enclosing this line in a try/catch and re-throwing the exception here so the message indicates whether the "Insufficient number of nonzero samples to estimate memory change" exception occurred pre- or post-"validation".

OpenSim/Auxiliary/auxiliaryTestFunctions.h
```diff
@@ -209,17 +209,116 @@ OpenSim::Object* randomize(OpenSim::Object* obj)
...
+size_t estimateMemoryChangeForCommand(T command, const size_t nSamples = 100)
+{
+    std::vector<size_t> deltas;
+
+    for (size_t i = 0; i < nSamples; ++i) {
+        size_t mem0 = getCurrentRSS();
+        // Execute the desired command
+        command();
+        // wait a millisecond to allow memory usage to settle
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
tkuchida: Not sure how long the nap should be, but I suggest the following, which is a more explicit check that the memory usage has settled:
``` cpp
size_t mem1, mem2;
const int maxIter = 1000;
int ctr = 0;

while (1) {
    mem1 = getCurrentRSS();
    std::this_thread::sleep_for(std::chrono::milliseconds(1));
    mem2 = getCurrentRSS();
    if (mem1 == mem2 || ++ctr == maxIter) { break; }
}
```

tkuchida:
> re-enabled testInitState for OSX to see if the these fixes also address issues on OSX

If the [testInitState failure](https://travis-ci.org/opensim-org/opensim-core/jobs/161733598#L2246) occurred in `validateMemoryUseEstimates()`, perhaps try increasing `nSamples` from 5 to 100 like the other memory tests. I think it's also fine to un-re-enable the test on OSX and leave #473 open.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
tkuchida: Resolved below (`getCurrentRSS()` does not distinguish between stack and heap).

chrisdembia:
I'm happy for this to be merged once the tests pass reliably. I think it is fine to allow the OSX failure for now.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
aseth1: > Perhaps make nSamples a default argument for symmetry with ...

Done. `nSamples` also defaults to a bigger number (11) but we would like to keep it as small as possible. Not sure if it will have an effect on OSX

OpenSim/Auxiliary/auxiliaryTestFunctions.h
aseth1: Good idea. Will add that.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
aseth1: Makes sense, but perhaps only wait up to 100ms? Otherwise 1s inside a loop (that defaults to 100x) will lead to a 100s of waiting to fail, most likely .

tkuchida: I agree: if we can't get a reliable memory usage estimate within a couple of seconds, another 98 seconds is unlikely to help. However, I'm confused about the logic here. If the theory is that an *n*-millisecond wait might be necessary for the memory footprint to stabilize, then why isn't mem0 treated similarly?

```cpp
size_t mem0, mem1;
const int maxIter = 10;

int ctr = 0;
do {
    mem0 = getCurrentRSS();
    std::this_thread::sleep_for(std::chrono::milliseconds(1));
} while ((getCurrentRSS() != mem0) && (++ctr < maxIter));

command();

ctr = 0;
do {
    mem1 = getCurrentRSS();
    std::this_thread::sleep_for(std::chrono::milliseconds(1));
} while ((getCurrentRSS() != mem1) && (++ctr < maxIter));
```

Based on the current algorithm, I assume you've found that `getCurrentRSS()` is unstable only *after* `command()` is called (perhaps because of the time required to store a new entry in `deltas`). Regardless, I think the other checks will catch situations where mem0 is unreliable (if that is an issue) so I think it's fine as it is.

tkuchida:
I agree with @chrisdembia: let's merge once tests pass and monitor its reliability on master.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
tkuchida: Ok, sounds good.

aseth1:
@tkuchida and @chrisdembia the only failure (once again) was on OSX. I have reinstated the exclusion on travis.

aseth1:
thank you guys!

Issue #821, opened 2016-03-10
Calling DataTable_::getColumnLabels() on a new data table gives an exception about a missing "labels" key


chrisdembia:
What if DataTable_'s are default-constructed to have a "labels" metadata already? Or was I doing something wrong?

Not for the hackathon, though.

klshrinidhi:
I don't think this is a bug. It was designed that way. `DataTable` is not assumed to contain labels at all times. One has to add labels to metatdata first in order to retrieve them.

chrisdembia:
What do you think about having an easy way to query if the table has column labels, then? I think the easiest way to check currently is via:

```cpp
const auto& keys = _table.getDependentsMetaData().getKeys();
if (std::find(keys.begin(), keys.end(), "labels") != keys.end())
```

klshrinidhi:
Yes. That is currently the only way to check if a key exists in metadata.

jenhicks:
@klshrinidhi Can you check the wording of the exception? Maybe add a hasColumnLabels check.

PR #1269, opened 2016-09-20
hasColumnLabels() for DataTable

klshrinidhi:
Addressing #821.

### DIFF
--- a/OpenSim/Common/AbstractDataTable.cpp
+++ b/OpenSim/Common/AbstractDataTable.cpp
@@ -77,6 +77,11 @@ namespace OpenSim {
         validateDependentsMetaData();
     }
 
+    bool
+    AbstractDataTable::hasColumnLabels() const {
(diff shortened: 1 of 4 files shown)


chrisdembia:
This looks good to me. What do you think about adding to the doxygen for `getColumnLabels()` that it throws an exception if there are no labels? You have nicely documented exceptions elsewhere. I know this is related to #821 and not this specific PR.

klshrinidhi:
There is already a sentence in the documentation for `getColumnLabels()` that it throws if table has no column labels. You can see it [here in master](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/AbstractDataTable.h#L261).

chrisdembia:
Oh oops! Thanks for pointing that out.

klshrinidhi:
I will make some changes to provide a bit more meaningful exception and message for 'labels not found'. Will push the changes soon.

PR #1272, opened 2016-09-21
[doxygen] Add Python and Matlab example code to Joints

### LINKED ISSUE #1232: Investigate wrapping of Coordinate enum classes

tkuchida:
Added usage examples for Python and Java/Matlab. Also note that the enum block wasn't being shown in user documentation, hence the change to doxygen-layout-user.xml. Here's the documentation for BallJoint:

coordinates

Fixes #1232.

### DIFF
--- a/doc/doxygen-layout-user.xml
+++ b/doc/doxygen-layout-user.xml
@@ -77,10 +77,10 @@
       <typedefs title=""/>
       <functions title=""/>
       <properties title="OpenSim Property, Connector, Output, Input Documentation"/>
-      <!-- <enums title=""/>
+      <enums title=""/>
       
       
-      <related title=""/>
+      <!-- <related title=""/>
       <variables title=""/>
       <properties title=""/>
       <events title=""/> -->
--- a/OpenSim/Simulation/SimbodyEngine/BallJoint.h
+++ b/OpenSim/Simulation/SimbodyEngine/BallJoint.h
@@ -59,6 +59,17 @@ public:
         \code{.cpp}
         const auto& rx = myBallJoint.getCoordinate(BallJoint::Coord::Rotation1X);
         \endcode
+
+        <b>Python example</b>
+        \code{.py}
+        import opensim
+        rx = myBallJoint.getCoordinate(opensim.BallJoint.Coord_Rotation1X)
+        \endcode
+
+        <b>Java/Matlab example</b>
+        \code{.java}
+        rx = myBallJoint.getCoordinate(BallJoint.Coord.Rotation1X);
+        \endcode
     */
     enum class Coord: unsigned {
         Rotation1X,
--- a/OpenSim/Simulation/SimbodyEngine/EllipsoidJoint.h
+++ b/OpenSim/Simulation/SimbodyEngine/EllipsoidJoint.h
@@ -56,6 +56,17 @@ public:
         const auto& rx = myEllipsoidJoint.
                          getCoordinate(EllipsoidJoint::Coord::Rotation1X);
         \endcode
+
+        <b>Python example</b>
+        \code{.py}
+        import opensim
+        rx = myEllipsoidJoint.getCoordinate(opensim.EllipsoidJoint.Coord_Rotation1X)
+        \endcode
+
+        <b>Java/Matlab example</b>
+        \code{.java}
+        rx = myEllipsoidJoint.getCoordinate(EllipsoidJoint.Coord.Rotation1X);
+        \endcode
     */
     enum class Coord: unsigned {
         Rotation1X,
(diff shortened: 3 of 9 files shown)


doc/doxygen-layout-user.xml
```diff
@@ -77,10 +77,10 @@
       <typedefs title=""/>
       <functions title=""/>
       <properties title="OpenSim Property, Connector, Output, Input Documentation"/>
-      <!-- <enums title=""/>
+      <enums title=""/>
       
       
-      <related title=""/>
+      <!-- <related title=""/>
```
chrisdembia: I'm pretty sure this means that the enums appear at the very very bottom of the pages. Is this desirable? Should they also be under the declaration section (`<memberdecl>`)? Have you checked to see how this affects the documentation of any other enums? Are there cryptic enums that should be hidden from users that are now shown?

chrisdembia:
> This documentation looks excellent. The image you show reveals that you could document the individual entries of the enum (there's a table with a second column, all empty). See here for an example of how to do this.

Let me add: I'm fine with not documenting these; I just wanted to make you aware of the possibility.

tkuchida:
> The image you show reveals that you could document the individual entries of the enum (there's a table with a second column, all empty).

I am aware of this. Would you like me to add a detailed description for each enum?

> As an aside: since [ci skip] was used, the generated doxygen is not viewable on myosin.sourceforge.net.

Yes, I included a screencap instead because myosin is often unresponsive (e.g., right now).

chrisdembia:
> I am aware of this. Would you like me to add a detailed description for each enum?

I'm indifferent; just did not know if you were aware.

> ...because myosin is often unresponsive (e.g., right now).

I'm not sure what you mean. The site is loading for me, but (as expected) there is no page for this PR.

doc/doxygen-layout-user.xml
tkuchida: > I'm pretty sure this means that the enums appear at the very very bottom of the pages.

That is correct.

> Is this desirable?

Yes, as far as I can tell. Previous to this PR, enums were not being shown in the user doxygen. I, therefore, concluded that the author of doxygen-layout-user.xml established that users would not need to know about enums (though I disagree; see below). Thus, it did not seem reasonable to include the enums at the top of the page. For the joints, the `getCoordinate()` and `updCoordinate()` methods appear at the bottom of the "Member Function Documentation" section; the "Member Enumeration Documentation" section follows immediately thereafter, which is ideal.

> Should they also be under the declaration section (<memberdecl>)?

I don't think it's necessary to show the enum documentation anywhere but where it currently appears.

> Have you checked to see how this affects the documentation of any other enums?

Yes. One example is Thelen2003Muscle, whose `printCurveToCSVFile()` method takes a `CurveType` enum as its first argument. The table of possible `CurveType`s now appears at the bottom of the Thelen2003Muscle doxygen page and is hyperlinked from the first argument of `printCurveToCSVFile()`.

> Are there cryptic enums that should be hidden from users that are now shown?

Here is a list of all the documented enums:

enums

I think the following entries are useful: Coord, CurveType, MotionType, UnitType, ValueType; the remaining entries are harmless (the Property_Deprecated and Tool pages are unlikely to be visited by users anyway).

tkuchida:
> I'm not sure what you mean.

I mean that the site does not load on my machine.

tkuchida:
I've amended the last commit so the doxygen will build on myosin. If anyone else wants to view it and does not have access to myosin, I can upload it somewhere public.

doc/doxygen-layout-user.xml
chrisdembia: Okay thanks for the responses.

chrisdembia:
LGTM.

> I've amended the last commit so the doxygen will build on myosin.

I'm not certain but I think you did this via something like `git commit --amend`. This changes the contents of a commit (e.g., one would have had to force the push). This is okay to do locally, but should not be done for commits that have been shared (on opensim-org). For example, someone may have pulled this branch to their computer before an amend, and now the two histories are different. I'm almost certain that nothing bad has happened this time. If you did not do this or if you already knew this, then I apologize for explaining.

tkuchida:
> someone may have pulled this branch to their computer before an amend, and now the two histories are different

Ok, I'll close this PR, delete the branch, and start again.

chrisdembia:
Alright; sorry for being such a pain!

tkuchida:
It wasn't clear how to proceed. There weren't any tasks pending for this PR, so I would have had to either commit an unrelated change or insert/delete whitespace somewhere (in which case the text displayed next to the filename on GitHub would be a meaningless message like "Forcing CI to run"). I'll try to be more careful.

PR #1273, opened 2016-09-22
[doxygen] Add Python and Java/Matlab example code to Joints

### LINKED ISSUE #1232: Investigate wrapping of Coordinate enum classes

tkuchida:
Added usage examples for Python and Java/Matlab. Also note that the "enum" block wasn't being shown in user documentation, hence the change to doxygen-layout-user.xml. Here's the documentation for BallJoint:

img1

Fixes #1232. See #1272 for initial review and detailed explanation of design decisions.

### DIFF
--- a/OpenSim/Simulation/SimbodyEngine/BallJoint.h
+++ b/OpenSim/Simulation/SimbodyEngine/BallJoint.h
@@ -59,6 +59,17 @@ public:
         \code{.cpp}
         const auto& rx = myBallJoint.getCoordinate(BallJoint::Coord::Rotation1X);
         \endcode
+
+        <b>Python example</b>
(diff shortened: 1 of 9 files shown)


tkuchida:
If a reviewer is interested (and patient), the developer's doxygen built from this PR can be viewed [here](http://myosin.sourceforge.net/1273/). @chrisdembia: please correct me if I'm wrong, but I believe the user's doxygen can be viewed only by pulling the branch and building locally.

chrisdembia:
> I believe the user's doxygen can be viewed only by pulling the branch and building locally.

Correct.

chrisdembia:
I'm merging under rule 5.

PR #1281, opened 2016-09-26
update metabolics_pythonic.py

gautam1858:
Its from time[1] right instead of time[-1] ?


tkuchida:
`duration = final_time - initial_time`. In Python, the last element of a list is index -1, so `time[-1]` is correct.

chrisdembia:
Thanks Guatam for being willing to fix a bug! Please let us know about other things you find.

gautam1858:
Thank you :)

PR #1279, opened 2016-09-26
Improve column labels used by TableReporter

tkuchida:
### Summary
Implements design that was proposed and approved in DevProp #1226. Fixes #1221.

### Primary changes
- Renamed "annotation" to "alias".
- Added `getShortLabel()`, `getLongLabel()`, and `setAlias()` methods.
- Updated `connect()` and `getAlias()` methods and doxygen.
- Removed existing code that created default aliases if none was provided by the user.
- Updated Reporters to use short label (ConsoleReporter) or long label (TableReporter).

### Tests added
- Created testReporters.cpp to check column headings reported by ConsoleReporter and TableReporter.
- Added tests for aliases and labels to testComponentInterface.cpp.
- Updated related Python test.

### Minor changes
- Silenced compiler warning in auxiliaryTestFunctions.h.

PR #1284, opened 2016-09-26
[DevProp] revision to Reporter column label design based on Input alias

aseth1:


aseth1:
@tkuchida @chrisdembia I believe these changes are not a major shift from the implementation in #1279. I hope you agree. If you prefer I make these changes there, I am fine with that.

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
```diff
@@ -54,14 +54,15 @@ and forcing the user to specify an annotation for each Output may not be practic
...
 
 1. Rename "annotation" to "alias" (e.g., `AbstractInput::getAnnotation()` becomes `AbstractInput::getAlias()`).
 2. If the user provides the optional second argument when connecting an Output to an Input, store the string as the Input's alias; otherwise, the alias remains null.
-3. Create `getShortLabel()` and `getLongLabel()` methods on Input:
-  - `getShortLabel()` will be called by ConsoleReporter and will return `isNull(alias) ? output_name : alias`.
-  - `getLongLabel()` will be called by TableReporter and will return `isNull(alias) ? full_path_name : alias`.
-4. Create `setAlias()` methods analogous to the current `getAnnotation()` methods so the user can change the alias after (but not before) connecting.
+3. Create `getReportedLabel()` on Input, which will always return the alias if one is provided:
+  - `getReportedLabel()` will be called by a Reporter and will return `isNull(alias) ? full_path_name : alias`.
+4. Create `setAlias()` methods analogous to the current `getAnnotation()` methods so the user can change the alias after (but not before) connecting. Connecting without an alias has the effect of clearing any previous alias.
```
chrisdembia: Right now, when you connect an output/channel to an input, we do not check if you already connected it. That is, a single output can be connected a list input multiple times. That might be OK behavior; not sure... Anyway, there would be no clearing of a previous alias, since the Input would just regard the connection as new.

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
```diff
@@ -54,14 +54,15 @@ and forcing the user to specify an annotation for each Output may not be practic
...
-The selected design involves four main changes:
+The selected design involves six main changes:
 
 1. Rename "annotation" to "alias" (e.g., `AbstractInput::getAnnotation()` becomes `AbstractInput::getAlias()`).
 2. If the user provides the optional second argument when connecting an Output to an Input, store the string as the Input's alias; otherwise, the alias remains null.
-3. Create `getShortLabel()` and `getLongLabel()` methods on Input:
-  - `getShortLabel()` will be called by ConsoleReporter and will return `isNull(alias) ? output_name : alias`.
-  - `getLongLabel()` will be called by TableReporter and will return `isNull(alias) ? full_path_name : alias`.
-4. Create `setAlias()` methods analogous to the current `getAnnotation()` methods so the user can change the alias after (but not before) connecting.
+3. Create `getReportedLabel()` on Input, which will always return the alias if one is provided:
```
chrisdembia: For some reason, I'm uncomfortable with the past-tense "reported." What about `getReportingLabel()` or `getLabelForReporting()`?

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
```diff
@@ -54,14 +54,15 @@ and forcing the user to specify an annotation for each Output may not be practic
...
+  - `getReportedLabel()` will be called by a Reporter and will return `isNull(alias) ? full_path_name : alias`.
```
tkuchida: I believe we agreed on `getShortLabel()` and `getLongLabel()` to accommodate any consumer. I don't think Input should need to know who is requesting the label (or even what a Reporter is). Perhaps `getDescriptiveLabel()`?

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
aseth1: Either is fine with me. I just want the naming to reflect its usage for reporting.

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
aseth1: I don't think we should check if was already connected.  On the Input side the *alias* becomes empty if you do not provide one when you (re)connect, correct?

chrisdembia: What do you mean by reconnect? If you call `initSystem()` (or `connect()`) multiple times, the alias should be preserved. This is because the alias is stored in the `connectee_name` property.

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
```diff
@@ -54,14 +54,15 @@ and forcing the user to specify an annotation for each Output may not be practic
...
+4. Create `setAlias()` methods analogous to the current `getAnnotation()` methods so the user can change the alias after (but not before) connecting. Connecting without an alias has the effect of clearing any previous alias.
+5. The ConsoleReporter has to truncate the ReportedLabel() because the label (either alias or the full_path_name) may exceed the colum width. For the time being, this can be the last `_width` characters of the label and a future PR can address more informative truncation approaches. For the time being, the users and testers of the ConsoleReporter can provide short aliases.
```
tkuchida: I think the last `_width` characters of the label could be misleading, depending on the names that have been assigned. I propose using the label if it does not exceed `_width` characters, and using the last `_width - 3` characters of the label preceded by "`...`" otherwise.

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
```diff
@@ -54,14 +54,15 @@ and forcing the user to specify an annotation for each Output may not be practic
...
+6. The TableReporter will provide the full_path_name for its Outputs as additional column meta data (in addition to labels) so that an Output can always we be traced back to a specific Component, regardless of its alias.
```
tkuchida: "we be" -> "be"

tkuchida: Is there an example demonstrating how this metadata should appear when the table is written to file?

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
tkuchida: "always" is unnecessary, and "is" should be "has been" since the alias is not passed to `getReportedLabel()`.

tkuchida: I don't think Input should care what type of Object is requesting the label, hence our previous design of "shortLabel" and "longLabel".

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
chrisdembia: Sounds good.

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
jenhicks: I can see pros and cons to both. Could we stick with the original design for this one and get some more user testing?

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
aseth1: The previous version said:
```.md
3. Create `getShortLabel()` and `getLongLabel()` methods on Input:
  - `getShortLabel()` will be called by ConsoleReporter and will return `isNull(alias) ? output_name : alias`.
  - `getLongLabel()` will be called by TableReporter and will return `isNull(alias) ? full_path_name : alias`.
```
That sounds specific to reporting. 

I think having four methods: `Input::` ` getName()`, `getAlias()`, `getLongLabel()`, `getShortLabel()` is confusing. I believe having one *label* method primarily associated with the act of reporting would be easier to understand, especially since `getLongLabel()` and `getShortLabel()` yield the same value whenever there is an alias. IMHO `getDescriptiveLabel()` will not be any more descriptive than the alias, but I am fine with being outvoted for the name. The more important design aspect, for me, is that we do not introduce two methods where one can suffice.

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
aseth1: I mean if you do:
`input.connect(output)` or `input.connect(channel)` (i.e. with no alias).

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
tkuchida: > That sounds specific to reporting.

"Label" is a generic term that is *not* specific to Reporters; the original design had methods called `getShortLabel()` and `getLongLabel()`. In lines 2 and 3, I indicated where `getShortLabel()` and `getLongLabel()` would be called in the implementation of the devprop; however, as we discussed, there may be other consumers that will want a label, with a preference for a short or long version. In our design meeting, it was decided that Input would handle the logic of preparing a short or long label. The new item 5 ("The ConsoleReporter has to truncate the ReportedLabel() […] a future PR can address more informative truncation approaches") moves this logic back to the ConsoleReporter, which contradicts the previous design decision.

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
aseth1: I believe string valued column meta data will appear as delimiter separated strings, just as *labels* currently appear (one per dependent column) . @klshrinidhi are additional meta data and particularly column meta data (in `dependentsMetaData`) being written out to .STO files?

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
aseth1: > as we discussed, there may be other consumers that will want a label, with a preference for a short or long version

Why would a user not request the *alias* directly that they assigned and use either `getShortName()` or `getLongName()` which will give the same value if an alias is provided? What are the non Reporter use cases for  `getShortLabel()` and `getLongLabel()`? Who else needs to truncate other than the `ConsoleReporter`? That issue was punted to `getShortLabel()`. 

My interpretation of the meeting was that short and long name were used to refer to the alias or output name and alias or full_path_name in the discussion and not that they would be the actual method names.  When I viewed the revised DevProp I thought the names were stand ins to be revised. I was surprised when the DevProp got merged today without review and that you had a PR in.

I'm very sorry for taking a weekend to mull it over. If you believe that I raise valid points then I will make the changes so your time is not wasted. Otherwise, leave it as is and I will kill this PR.

tkuchida: > Why would a user not request the *alias* directly that they assigned and use either `getShortName()` or `getLongName()` which will give the same value if an alias is provided? What are the non Reporter use cases for `getShortLabel()` and `getLongLabel()`?

Potential use cases include displaying information about connections in the GUI and generating plot legends. It was my understanding that we wanted to avoid code like `text = !getAlias().empty() ? getAlias() : getOutput().getName()` and path string parsers outside Input. Instead, Input would be asked for a short or long label and the logic for determining the short and long versions of the label would be centralized in Input.

> My interpretation of the meeting was that short and long name were used to refer to the alias or output name and alias or full_path_name in the discussion and not that they would be the actual method names. When I viewed the revised DevProp I thought the names were stand ins to be revised.

I thought the devprop was very clear about what the methods would be named (see item 3: "Create `getShortLabel()` and `getLongLabel()` methods on Input").

> I was surprised when the DevProp got merged today without review and that you had a PR in.

I was unaware that there were any outstanding design concerns following the Dev Team meeting. @chrisdembia reviewed and merged and devprop, which reflected what seemed to be a unanimous agreement on the design. Apparently I misinterpreted the conversation during the meeting, or perhaps there was a follow-up discussion that I did not hear and that was not summarized in a comment on the devprop PR.

> If you believe that I raise valid points then I will make the changes

I don't think that is how this process is supposed to work. I will fix the implementation once we all agree on a new design.

jenhicks:
Let's discuss this at the end of the OpenSim meeting today (after the visualizer demo/discussion).

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
aseth1: There was no secret follow up discussion. I should have raised my concerns earlier, but they did not come to me immediately at the meeting. I thought a bit about it over the weekend. I tried to discuss it with you in person first but you demanded I update the DevProp, which I did. A 15-20min direct conversation could have established agreement to revise/or not. As the principle on this PR if you don't agree then there isn't an agreement.  I voiced my suggestions/concerns and they were rejected. I am totally fine with that.

I apologize for the disruption and I will strive to be more timely in my feedback. As a matter of team dynamics, we should all feel free to discuss any concerns we have at anytime in the dev process.  Those concerns can be addressed, squashed or postponed, but discussing it with each other should be easy and should not be reserved for weekly meetings.

Thank you for carefully reviewing the proposal and providing your feedback. @jenhicks also recommended against the revision, which I believe is your position as well. I will close it.

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
tkuchida: An alternative design would be for the ConsoleReporter to break long labels over multiple lines and default to the full path:
<pre>
             |             | a/very/long/
        time |  activation |  output/path
</pre>
which would not be difficult to do, would avoid adding context-aware truncation logic to the ConsoleReporter, and would eliminate one of the "getLabel" methods.

chrisdembia: > An alternative design would be for the ConsoleReporter to break long labels over multiple lines and default to the full path:

I like this idea! We might need to try it out first to see if it'll work well in practice. I think all full output paths will start with the model's name (e.g., `/Gait2392_Simbody/hip_r/flexion/value`), so it seems that all column labels will be at least two lines.

DevelopmentProposals/reporterColumnLabels/reporterColumnLabels.md
tkuchida: > you demanded I update the DevProp, which I did. A 15-20min direct conversation could have established agreement to revise/or not

The whole Dev Team was involved in selecting the design, so we cannot decide to change it without everyone's involvement. I think one of the objectives of the DevProps is to make the design process more transparent and inclusive.

> we should all feel free to discuss any concerns we have at anytime in the dev process

I disagree. The DevProp was written ~3 weeks before the meeting and we spent an hour discussing designs. This is not a critical piece of the software, and it seems like re-considering the design is not an ideal allocation of resources leading up to Friday's alpha release. Re-designing aliases takes time away from other issues that have not been addressed at all.

aseth1:
Created  #1287 to summarize discussion to be discussed later. The current merged DevProp stands for the review of PR #1279 and this PR is being closed.

PR #1285, opened 2016-09-27
Unified command-line tool, this time without putting docopt in the repo

chrisdembia:
### Background and motivation

This PR introduces a single unified command-line tool, currently named `opensim`. The impetus for this change was the ability to install OpenSim into system-wide directories on UNIX systems (#195). However, the primary benefit in the long term will be the ability to add new functionality to the command-line tool. Right now, this new unified tool has the same abilities as the union of all our current command-line tools, but in the future, we can add new features (just ideas; some are not great ideas b/c they duplicate GUI functionality):
1. visualizing/simulating a model
2. converting data files between formats (c3d to trc) using DataAdapters.
3. package manager for OpenSim plugins, etc.
4. plotting data from a data file or states trajectory.
5. creating template files for making custom plugins (autogenerating myPluginDLL.h, etc.).
6. managing OpenSim "workbook" zip files (if/when we have this feature).
### Overview of the changes

This PR introduces the following changes:
1. Introduces the dependency `docopt.cpp` into the repo. This is a command-line argument parser whose input is the documentation of the command itself, which ensures consistency between the documentation and the parsing. The package `docopt` was originally a python package, but has been ported to numerous languages.
2. A new executable `opensim`, with the following subcommands:
   1. `opensim run-tool` for running any Tool (ik, rra, cmc, etc.).
   2. `opensim print-xml` for creating a template XML file for any Object, including tools.
   3. `opensim info` prints the names and description of properties for any Object.
   4. `opensim update-file` for upgrading XMLDocuments and Storage files (previously, `versionUpdate`).
3. A test case that tests most possible command invocations, and also checks for the correct error messages.
4. On UNIX/OSX, the old command-line tools are not built/installed anymore. The old command-line tools are still built/installed but ONLY on Windows, for backwards compatibility. In a future release, we can remove the old command-line tools.
5. A fix to a bug in `Object::makeObjectFromFile()`; this command changes the working directory to the directory containing the file, but previously would not change back to your original working directory.

Perhaps the best way to start reviewing the PR is to look at the documentation for the commands in `opensim-cli.cpp`, `opensim_run_tool.h`, etc.; and to read the test `testCommandLineInterface.cpp`, which shows a bunch of different invocations of the tool and snippets of the expected output.

I did plenty of manual testing (running the commands myself with different arguments) on OSX.
### Decisions
1. **Choice of command-line parsing library** I considered building the command-line tool in python and using a python library for the command-line parsing, but this would have created a headache for packaging OpenSim. I considered TCLAP, which is a very nice C++ library but it contains _lots_ of files. Docopt.cpp strikes a good balance between functionality and size.
### To discuss

What should we name the command-line tool?
1. The name `opensim` makes sense for opensim-core, but causes complications when combining with the GUI, since the GUI executable is also named `opensim[.exe]`. The easiest method for renaming the GUI executable to `OpenSimGUI` causes the whole application to be named `OpenSimGUI` (in the start menu, etc.). To preserve the name of the application as OpenSim, we would need to either:
   i.  manually overwrite some of the files that Netbeans auto-generates for us, or
   ii. put the GUI's `opensim` executable in a separate directory, which would require manually overriding Netbean's default application layout.
2. I don't like `opensim-cli`. 
3. `osim`: This is somewhat short which introduces a greater chance of name conflicts again, and it also could cause confusion with our model file extension? Though we also prefix our libraries with `osim`, so perhaps it's not a big issue.
### History of this PR/feature

This feature first showed up in PR #1137, which was closed so that docopt would not be copied into the repository. In this new PR, I addressed @tkuchida's previous comments (about uppercase/lowercase for tool names, and a more user-friendly changelog note).


aymanhab:
We will remove old command line executables at some point or note dprecated and point to the command line (seems 4.0 is reasonable to remove old execs but may affect old user scripts)

aymanhab:
We may add a line to confirm which tool we are running.

aymanhab:
Question on whether to prompt user before overwriting output file, seems unnecessary

aymanhab:
Path specification clearup maybe helpful e.g. handling spaces, relative-path etc.

aymanhab:
name discussion: either osim (short but could confuse with model format) or opensim-cmd or opensim-command

aymanhab:
conclusion opensim-cmd

aseth1:
Possibly update existing command line exe's to indicate their deprecation and to provide instruction on how to invoke their setup using the new unified command line tool.

aseth1:
Perhaps include "Setup" in the default_<ToolName>.xml generate Tool setup files.

chrisdembia:
Thank you very  much @aymanhab and @aseth1 for taking notes.

aseth1:
First example(s) in help could be of just invoking `run-tool` without the other options for simplicity

aseth1:
When "pattern" is not satisfied and tool does not run, do not spew out the complete help but instead suggest the user run help on just the command they tried to execute.

chrisdembia:
For myself to keep track of the comments I've addressed, I've made a checklist:

- [ ] Renamed tool to `opensim-cmd`.
- [ ] For `run-tool`, print the name of the tool being run.
- [ ] Print brief message if arguments are not valid; don't print the whole HELP text.
- [ ] Add simpler examples to main HELP text (e.g., no plugins).
- [ ] Put "Setup" in the name of default XML files for tools.
- [ ] Clarify the syntax for plugin paths.
- [ ] Add "deprecated" messages to old tools.
- [ ] In old tools, add guide on how to use new command-line tools.

chrisdembia:
I have now addressed all feedback from the dev meeting today. This PR is ready for review.

@aymanhab: Clarification about the syntax for paths for the `--library` option:

1. Spaces are allowed, so long as you enclose the path in quotes (single or double). This is now mentioned in the HELP text.
2. The path can be absolute or relative to the current directory. This is mentioned in the HELP text.
3. On UNIX (only tested on OSX), the path separator must be forward slash `/`. On Windows, the separator can be either `\` or `/`. In the interest of keeping the HELP text somewhat brief, I did not put this in the HELP text.

chrisdembia:
Any takers for reviewing this? 💻

chrisdembia:
> I'll review :)

Woohoo! thank you :)

chrisdembia:
@jimmyDunne has said he will review this, but this PR still needs one more reviewer. Anyone else interested?

jenhicks:
@klshrinidhi can you give one more quick pass?

klshrinidhi:
:+1:

chrisdembia:
Thank you @klshrinidhi

jimmyDunne:
Works on OSX. Didn't have an issue with any of the major calls.

chrisdembia:
Thanks for testing, @jimmyDunne. I hope you like the interface.

PR #1278, opened 2016-09-24
flatten() for DataTables & TimeSeriesTables

klshrinidhi:
Addressing #1252.

The PR adds functions `flatten()` to `DataTable` and `TimeSeriesTable`. The function is accessible from Python and Java.

PR #1286, opened 2016-09-27
Matlab test for the table.flatten

jimmyDunne:
flattening a c3d generated Vec3 times series table causes an exception. Would be great to get some advice on the exception

```
Java exception occurred:
java.lang.RuntimeException: Key = units expected = 18 received = 6
	In file
        /Users/jimmy/repository/opensim-master/opensim-core/OpenSim/Common/DataTable.h:738
        	In function 'validateDependentsMetaData'
	at
        org.opensim.modeling.opensimModelCommonJNI.DataTableVec3_flatten__SWIG_0(Native
        Method)
	at
        org.opensim.modeling.DataTableVec3.flatten(DataTableVec3.java:53) 

```

### DIFF
--- a/OpenSim/Sandbox/MatlabScripts/c3d_reading.m
+++ b/OpenSim/Sandbox/MatlabScripts/c3d_reading.m
@@ -1,124 +1,29 @@
 
 import org.opensim.modeling.*
 
-
-
 %% Use a c3dAdapter to turn read a c3d file
 adapter = C3DFileAdapter()
 
-tables = adapter.read('test_walking.c3d')
-
+tables = adapter.read('test_walking.c3d');
 
 %% get the Markers
 markers = tables.get('markers');
-
-markerdata = opensimTimeSeriesTableToMatlab(markers);
-markerlabels = fieldnames(markerdata);
 
 % Print the (unrotated) markers to trc file
 trcfileadapter = TRCFileAdapter();
 trcfileadapter.write(markers,'test_walking.trc');
 
-%% Get the force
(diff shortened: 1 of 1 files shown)


klshrinidhi:
I don't think this PR should go into mine. Testing Matlab scripts is  not implemented in our CI. So this PR will not add any testable code.
I think you should close this PR and start one after my PR is merged to master.

jenhicks:
Thanks @jimmyDunne . I think it is still helpful to see what the test of a Tool should look like even if the Matlab PR doesn't make it into this PR.

jimmyDunne:
This PR in sandbox, It doesn't need to contribute — it needs to show that #1278 works in Matlab. If I can't get this working on a simple example, then the original PR should not be merged.

jenhicks:
Thanks for clarifying @jimmyDunne. I didn't realize the change was into sandbox. That seems like the right place to add this kind of code. If there are other proposals for how to test Matlab use cases, we can discuss at an upcoming dev team meeting. 

@klshrinidhi can you comment on the exception that James pointed out?

klshrinidhi:
I will debug and get back what the exception is about. It is thrown by DataTable. I will add appropriate Java tests to PR #1278. 

While I appreciate testing it in Matlab, I am not going to be accepting this PR into #1278.

jenhicks:
@jimmyDunne Thanks again for catching this issue and providing code to reproduce it. Would you mind closing this PR for now and submitting a sandbox PR after PR #1278 is merged? @klshrinidhi is also going to add a corresponding java test.

Issue #1250, opened 2016-09-13
Prevent segfault in Matlab when attempting to access bad index


tkuchida:
Because Matlab is 1-indexed and C++ is 0-indexed, Matlab users are likely to make off-by-one errors (OBOEs) when indexing into OpenSim Tables, etc. Currently, Matlab just kerplodes with an unhelpful segfault message when attempting to access a bad index, as reported by @jimmyDunne:

![jamesmatlabcrash](https://cloud.githubusercontent.com/assets/4203505/18487629/d41a55d2-79aa-11e6-91ca-d0d8801288d5.png)

@aymanhab suggests throwing/catching exceptions to prevent :boom:. See #1249.

tkuchida:
Try replicating @jimmyDunne's example (the `for` loop in opensimVec3TableToStruct.m, line 12) in the API: is an index-out-of-bounds exception thrown and just not being caught correctly in Matlab? cc/ @aymanhab

carmichaelong:
Likely that SimTK::VectorView has bound-checking and non-bound checking operators. Check if there's an Exception thrown (might be Release vs Debug too)

chrisdembia:
> is an index-out-of-bounds exception thrown and just not being caught correctly in Matlab?

This sounds likely. In general, if OpenSim throws an exception but SWIG/Java doesn't know about it, MATLAB will crash.


> Likely that SimTK::VectorView has bound-checking and non-bound checking operators. Check if there's an Exception thrown (might be Release vs Debug too)

The non-checking method is usually `operator[]`, which probably isn't wrapped or available in the wrapping. We probably don't want to make it available in MATLAB either.

Might be best to tell SWIG/Java about the exception using `SWIG_DECLARE_EXCEPTION`, as is done [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/Model.h#L206).

aymanhab:
@jimmyDunne What's the minimal piece of code to test this. I can't create a new SimTK::RowVector (you get them from something else) and I'd like to write a minimal piece of code to get it (either build a table, add to it live, or use files if necessary), please advise @klshrinidhi or @jimmyDunne Thanks.

klshrinidhi:
I think it is possible to create an object of `RowVector` since it is being wrapped [here](https://github.com/opensim-org/opensim-core/blob/master/Bindings/simbody.i#L86).

[...]

PR #1288, opened 2016-09-28
Java tests for DataTable to catch/prevent seg-fault problems

klshrinidhi:
Addressing #1250.

This PR adds tests in Java that parallel the tests in Python. The seg-fault problem in #1250 is handled with functions `get()` and `set()` for `RowVector`, `RowVectorView`, `VectorView` etc, which now throw an exception when accessing elements out of bounds of these arrays.

One problem was discovered while working on this:
`table.getRowAtIndex()` causes seg-fault if the argument is out-of-bounds though the function throws an exception in C++. Other exceptions thrown by DataTable do propagate up to Java so it is not common to all exceptions. This problem is yet to be debugged.

I would like to expand these tests to cover changes in PR #1278.

**Edit:**
I have fixed the above mentioned problem. Here is the seg-fault for the record:
```
83: Test command: /usr/lib/jvm/java-8-openjdk-amd64/bin/java "-enableassertions" "-Djava.library.path=<path-to>/opensim-core/build" "-classpath" "TestTables.jar:<path-to>/opensim-core/build/Bindings/Java/src/org-opensim-modeling.jar" "TestTables"
83: Test timeout computed to be: 1500
83: #
83: # A fatal error has been detected by the Java Runtime Environment:
83: #
83: #  SIGSEGV (0xb) at pc=0x00007f2e12b3c3e2, pid=7294, tid=139836563777280
83: #
83: # JRE version: OpenJDK Runtime Environment (8.0_91-b14) (build 1.8.0_91-8u91-b14-3ubuntu1~16.04.1-b14)
83: # Java VM: OpenJDK 64-Bit Server VM (25.91-b14 mixed mode linux-amd64 compressed oops)
83: # Problematic frame:
83: # C  [libSimTKcommon.so.3.6+0x1d83e2]  SimTK::MatrixHelper<double>::MatrixHelper(SimTK::MatrixCommitment const&, SimTK::MatrixHelper<double>&, SimTK::MatrixHelper<double>::ShallowCopy const&)+0x12
83: #
83: # Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
83: #
83: # An error report file with more information is saved as:
83: # <path-to>/opensim-core/build/Bindings/Java/tests/TestTables/hs_err_pid7294.log
83: #
83: # If you would like to submit a bug report, please visit:
83: #   http://bugreport.java.com/bugreport/crash.jsp
83: # The crash happened outside the Java Virtual Machine in native code.
83: # See problematic frame for where to report the bug.
83: #
```


Bindings/simbody.i
```diff
@@ -32,22 +32,25 @@ namespace SimTK {
...
 
      void __setitem__(size_t i, double value) {
          if(i >= $self->nelt())
-             throw SimTK::Exception::Base{"Index out of Range."};
+             throw std::out_of_range{"Index out of Range."};
 
          $self->updElt(0, i) = value;
      }
  }
+%exception;
```
chrisdembia: This clears all previous exception-handling macros, right? Perhaps it would be helpful to add a comment that mentions that, since it is not obvious what this does.

klshrinidhi: I added it during debugging and did not remove it. It should not be there now.

aseth1:
> I would like to expand these tests to cover changes in PR #1278.

👍  
Just a reminder to include a test that also invokes a Tool that takes a flattened Vec3 table, like the forces.mot file (as in @jimmyDunne's example). I believe @jimmyDunne has or will add such a use case (of running a Tool through Matlab) which you can look at to make a similar Java side test.

klshrinidhi:
> Just a reminder to include a test that also invokes a Tool that takes a flattened Vec3 table, like the forces.mot file (as in @jimmyDunne's example). I believe @jimmyDunne has or will add such a use case (of running a Tool through Matlab) which you can look at to make a similar Java side test.

Yes. That is what I meant by my comment:
> I would like to expand these tests to cover changes in PR #1278.

jenhicks:
@klshrinidhi Just clarifying ... are you working on adding the test for #1278 ? Then we can merge this PR and you can track down the cause of the segaults reported for the branch in PR #1265 ? Let us know if you are waiting on any input.

klshrinidhi:
I have explained the changes in this PR to @aymanhab. He will get back with his review/changes.

jenhicks:
Sounds great. Thanks for the update @klshrinidhi !

aymanhab:
@klshrinidhi I'm debugging this now, after reverting the introduction of proxy classes. One question, who owns the memory returned in DataTable.getRowAtIndex()? SWIG flags indicate this is an independent copy rather than a reference that's likely why later editing crashes. Where should I look for documentation regarding who owns what on the DataTable side? Thanks.

klshrinidhi:
@aymanhab 
The function `getRowAtIndex()` returns a `SimTK::RowVectorView` extracted from a `SimTK::Matrix`. So simbody documentation have details. My understanding is that `RowVectorView` **does not** own the memory. It is merely a view to matrix owned by `DataTable`.

aymanhab:
I fixed the example and undid the proxy class, this should be good to go. Test failure is bogus though may indicate overall slowdown but totally unrelated to this PR. @jenhicks @klshrinidhi @jimmyDunne

klshrinidhi:
@aymanhab 
Looks great. 
* I removed the file `FixedSizeArray.h` I had added since it is now not required.
* I added the out-of-bounds tests back. You probably removed them for debugging and did not add them back.

aymanhab:
Great, thanks much Shrinidhi, yes I reverted the whole commit. Thanks much
for adding the tests back.

aymanhab:
All looks good, tests added and pass. Bogus failure due to missing SWIG on travis. Merged.

PR #1289, opened 2016-09-28
Update VisualizeModel.cpp

gautam1858:
adding the muscle controller

### DIFF
--- a/OpenSim/Tests/VisualizeModel/VisualizeModel.cpp
+++ b/OpenSim/Tests/VisualizeModel/VisualizeModel.cpp
@@ -60,6 +60,19 @@ int main(int argc, char **argv)
         osimModel.getVisualizer().show(si);
         getchar(); // Keep Visualizer from dying until we inspect the visualization window..
     }
+    // Add the muscle controller to the model
+    
+        osimModel.addController(muscleController);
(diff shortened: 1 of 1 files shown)


jimmyDunne:
This is a specific test for the visualizer, what does adding a muscle controller do to improve that?

gautam1858:
An additional component to control the muscle along with visualiser would be useful and thought test meaning we can try out different codes.Adding a muscle controller is really helpful when there will be end to end control with visualisation and muscle control i.e with the model. It opens up ideas for researchers and a bonus for the researches, in general if you have visualisation then it would be nice to have controller as well in the test.

jimmyDunne:
What you are looking to do is develop an example, not a test. The tests are designed to test the compilation and need to be small and specific. 

Examples need to be added to opensim-core/OpenSim/Examples/

chrisdembia:
@gautam1858 thank you for your eagerness to contribute! It is very much appreciated. However, @jimmyDunne is correct. As a result, I am closing this PR.

 @gautam1858, the README example contains a muscle, a controller that controls the muscle, and also visualization. What does this example miss that you would like to see?

gautam1858:
@jimmyDunne @chrisdembia ok thank you

PR #1291, opened 2016-10-02
Fix MATLAB tests on Windows (and with Xcode)

chrisdembia:
This PR fixes the configuration of the MATLAB tests so they successfully run on Windows. I tested this locally.

Unfortunately, we cannot test this with continuous integration. So you'll have to trust me a little.


carmichaelong:
👍 Just tested this and it fixed MATLAB tests for me (32-bit R2010b).

aymanhab:
Looks reasonable, and an improvement. Merging

chrisdembia:
Thank you @aymanhab.

PR #1292, opened 2016-10-02
Update Readme.md

gautam1858:
It should use "--force -yes"

### DIFF
--- a/README.md
+++ b/README.md
@@ -818,7 +818,7 @@ In **Terminal** --
 sudo add-apt-repository --yes ppa:george-edison55/cmake-3.x
 sudo apt-add-repository --yes ppa:fenics-packages/fenics-exp
 sudo apt-get update
-sudo apt-get --yes install git cmake cmake-curses-gui clang-3.6 \
+sudo apt-get --force -yes install git cmake cmake-curses-gui clang-3.6 \
                            freeglut3-dev libxi-dev libxmu-dev \
                            liblapack-dev swig3.0 python-dev \
                            openjdk-7-jdk


klshrinidhi:
Thanks for PR @gautam1858.
I think you intend to propose replacing `--yes` with `--force-yes`. We **do not** want to use `--force-yes`. The documentation says:
```
--force-yes
   Force yes; this is a dangerous option that will cause apt to continue without 
   prompting if it is doing something potentially harmful. It should not be used 
   except in very special situations. Using force-yes can potentially destroy 
   your system! Configuration Item:
   APT::Get::force-yes. This is deprecated and replaced by --allow-downgrades, 
   --allow-remove-essential, --allow-change-held-packages in 1.1.

-y, --yes, --assume-yes
   Automatic yes to prompts; assume "yes" as answer to all prompts and run 
   non-interactively. If an undesirable situation, such as changing a held package, 
   trying to install a unauthenticated package or removing an essential package 
   occurs then apt-get will abort.
   Configuration Item: APT::Get::Assume-Yes.
```

gautam1858:
But when I try to run this it shows error and when I use --force-yes it will run without any errors, so asked for the changes

chrisdembia:
@gautam1858 what are the errors you get? Hopefully there is an alternative solution.

klshrinidhi:
Got it. I guess this is specific to your environment. We have successfully tested the instructions on fresh installations of respective Ubuntu versions. 

Great that you solved it for your system. If you happen find out what exactly was causing the error, please post it here so we can try to understand the cause.

gautam1858:
@chrisdembia @klshrinidhi Sure I will re-run and get back to you

gautam1858:
This time it worked properly but if i find it happening again i will get back to you

PR #1299, opened 2016-10-04
Fix attached geometry with offset

aymanhab:
Old models that had Geometry attached to Bodies allowed for specifying a relative transform (in 2 locations). When these became attached_geometry, these transformations were lost (causing double_pendulum model and a few others to display incorrectly). This PR fixes the problem and beefs up the test case testVisualization to make sure the display is corrected.


aymanhab:
This fixes a bug in updateFromXML of PhysicalFrame, in particular the method PhysicalFrame::convertDisplayGeometryToGeometryXML, the rest is adding a test case to verify the fix which has also been verified visually displaying the double_pendulum model and Carmichael's model which both exhibited the problem of losing transforms. Any feedback is appreciated @aseth1 @chrisdembia

OpenSim/Simulation/Model/PhysicalFrame.cpp
```diff
@@ -267,9 +283,23 @@ void PhysicalFrame::convertDisplayGeometryToGeometryXML(SimTK::Xml::Element& bod
...
                     frameSetObjectsIter = frameSetNode.element_begin("objects");
                 }
-                createFrameForXform(frameSetObjectsIter, frameName, localXform, bodyName);
+                // Create Frame from composed transform
+                createFrameForXform(frameSetObjectsIter, frameName, composedXformVec6, bodyName);
 
                 XMLDocument::addConnector(meshNode, "Connector_Frame_", "frame", frameName);
+                SimTK::Xml::element_iterator parnetFrame = frameSetObjectsIter->element_begin("PhysicalOffsetFrame");
+                while (parnetFrame->getRequiredAttributeValue("name") != frameName)
+                    parnetFrame++;
```
chrisdembia: Typo: parnet -> parent.

aymanhab: Thanks @chrisdembia fixed.

OpenSim/Simulation/Model/PhysicalFrame.cpp
```diff
@@ -237,22 +238,37 @@ void PhysicalFrame::convertDisplayGeometryToGeometryXML(SimTK::Xml::Element& bod
...
+                && (outerTransformVec6.norm() < SimTK::Eps);
+            // In practice no models were found where both innerTransform and outerTransform were
+            // non trivial or scale factors were specified in both
+            if (!attachToThisFrame) {
+                // Create a Vec6 by composing inner and outer transforms
+                SimTK::Transform outerTransform(outerTransformVec6.getSubVec<3>(3));
+                outerTransform.updR().setRotationToBodyFixedXYZ(outerTransformVec6.getSubVec<3>(0));
+                SimTK::Transform innerTransform(innerXformVec6.getSubVec<3>(3));
+                innerTransform.updR().setRotationToBodyFixedXYZ(innerXformVec6.getSubVec<3>(0));
+                SimTK::Transform composed = outerTransform.compose(innerTransform);
```
aseth1: This looks correct to me and it is what I expect was happening when both internal mesh and the containing `VisibleObject` had associated transforms. Is there anyway to verify that this was in fact the case? The set of models is likely insufficient to test this since most transforms in example models are identity or require a simple translation or at most a single axis rotation.

A test case that beginning with a 3.3 model with two non-trivial transforms that fully exercises this code to convert to a single (combined) `PhysicalOffsetFrame` would be helpful. This could be added to `testVisualization` where the 3.3 transform of the DecorativeGeometry is compared to that from these updates.

aymanhab:
Bogus test failure unrelated to this PR, ready for review @aseth1

OpenSim/Tools/Test/sphere.vtp
```diff
@@ -0,0 +1,883 @@
+<?xml version="1.0"?>
+<VTKFile type="PolyData" version="0.1" byte_order="LittleEndian" compressor="vtkZLibDataCompressor">
```
aseth1: What is the purpose of the adding the sphere mesh to this test? I believe the actual mesh is somewhat irrelevant to the test, but a sphere is particularly bad because it impossible to visually inspect the result for errors in orientation.  My preference would be to remove both sphere and cylinder for a mesh that is easy to visually distinguish a rotational error. Consider a box with differing half-lengths.

OpenSim/Simulation/Model/PhysicalFrame.cpp
aymanhab: Thanks @aseth1 for the review. The two models I tested with that failed to visualize properly had transforms in exactly one of the two places. One of them was added as a test case (double_pendulum) the other tested locally (basically Carmichael's model that I didn't want to add to the repo). We can manufacture models that have both non-trivial transforms but to go thru this only to collapse into one-transform doesn't seem quite justified. Considering how rare that may have happened (if ever) I'd be happy to throw an exception instead if both transforms are non-trivial.

chrisdembia: > Considering how rare that may have happened (if ever) I'd be happy to throw an exception instead if both transforms are non-trivial.

My personal opinion is that this has happened, and we should support it. I thought there were good reasons for specifying two separate transforms (similar to the benefit of frames): e.g., to put a bunch of geometry at one place.

OpenSim/Tools/Test/sphere.vtp
aymanhab: Great point regarding the sphere being invariant to rotations,  but since we haven't had non-mesh objects pre 4.0, whatever model that exercises conversion will have to use a mesh file. I can make one that's a sparse Mesh (since we don't have one off-the-shelf) if you think that strengthens the test. Please let me know.

OpenSim/Simulation/Model/PhysicalFrame.cpp
aymanhab: Thanks @chrisdembia my opinion is that if we can't find such models in the house  (because models are either created by hand/GUI (e.g. Carmichael/Tim only one transform exposed) or programmatically (no published API to clearly set both transforms) or imported from SIMM years ago. If none of these resulted in both transforms being non-trivial then spending time to make this hypothetical scenario work (where we have no way to verify correctness) doesn't seem justified. Yes we hypothesized a good reason for having both transforms earlier but that need is subsumed by using Frames in 4.0

aseth1: > where we have no way to verify correctness

I believe we can verify if the composition implemented is what the previous code base was doing quite easily:

1. Create a simple model with just ground in 3.3. Add a `VisibleObject` and assign it a non-simple transform. Inside the `VisibleObject` add a `Mesh` (a box or block) with its own non-simple transform.
2. In 3.3 Visualize (1.) and record its `DecorativeGeometry` transform. This could be copied and pasted into to the test as the `expectedTransform`.
3. Load the model in the current code base that includes the composition.
4. Visualize and extract the `DecorativeGeometry` transform again as the `currentTransform`. 
5. Test that  `currentTransform == expectedTransform`.
6. Visually inspect the block position and orientation between 3.3 and current visualizations

OpenSim/Tools/Test/sphere.vtp
aseth1: > whatever model that exercises conversion will have to use a mesh file

Sorry for the confusion. I meant use a *box* mesh, but I realize now that box is built-in and I should have said `block.vtp`. I think that or one that you make that is fine- which ever is easiest.

aymanhab: Sure, will do. Though keep in mind the test will only be visual since the standard actually compares transforms rather than actual display, so we'll see nothing in the code base to reflect this other than the change of the file name. I may add a png. You'll have to trust me ;)

aseth1: The first thing I usually do to debug is to visualize! So if it is immediately (visually) apparent that a transform is off it saves time - I think. Otherwise seeing spheres I would just be scratching my head and would have to diff the visualization text files and also factor in a possible change in order, etc...

aymanhab: Agreed, and that change has been made to use non-uniformly scaled block.vtp. Screen capture from 3.3 and master have been added as well. Thanks @aseth1

aymanhab:
Thanks for the feedback @chrisdembia and @aseth1 your concerns have been addressed. Ready for another pass on the test.

OpenSim/Tools/Test/doubletransform33.osim
```diff
@@ -0,0 +1,150 @@
...
+									<!--Display opacity between 0.0 and 1.0-->
+									<opacity>1</opacity>
+								</DisplayGeometry>
+							</objects>
+							<groups />
+						</GeometrySet>
+						<!--Three scale factors for display purposes: scaleX scaleY scaleZ-->
+						<scale_factors> 1 1 1</scale_factors>
+						<!--transform relative to owner specified as 3 rotations (rad) followed by 3 translations rX rY rZ tx ty tz-->
+						<transform> 0 0 0 0.4 0.5 0.6</transform>
```
aseth1: Same as above.

aymanhab:
Considering that this is a rather hypothetical case that we have NO models to exercise and will verify only visually, I preferred a composition of Translations that anybody can verify visually over a composition of both translations and rotations where I could be able to verify visually but  reviewers can't. Agreed it's not comprehensive but as a sanity check. Let me know what you think. Thanks.

aseth1:
At this stage, whether it is hypothetical or not, we decided to add a test to verify that the current composition is what you would have gotten from the old `VisibleObject`. IMHO it is more important that we compute the correct transform than have visual confirmation for this test case- I am willing to trust the DecorativeGeometry will be drawn correctly given the correct transform. 

From experience, it is hard/time consuming to to get transforms right and one relies heavily on visualization for confirmation. If a user opens their model in 4.0 and it now looks nothing like what the 3.3 model looks like, it isn't going to instill much confidence in 4.0.

It  seems plausible that some (very few) users could have painstakingly composed nice visuals by positioning/rotating individual meshes as part of a single VisibleObject (as a composite) and then scaled and transformed the VisibleObject to represent a Body. It seems perfectly reasonable to have done that. I'll try to find an example where I did that.

aymanhab:
That would be perfect @aseth1 as I definitely wouldn't want such work to be lost. If such model showed correctly in the API visualizer that would be even better since the API visualizer has not changed in 4.0 while we moved from VTK to DecorativeGeometry in the GUI, adding more unknowns.

aymanhab:
@aseth1 Considering the scope of the proposal to strengthen the test case (to potentially cover a corner case that may have existed), can we merge this PR and open an issue to track that if you can find a model as there are other PRs (Display of ContactGeometry, Frames and DisplayHints that all interact and are tested by the same test.) Thanks.

aseth1:
@aymanhab I agree we should merge this PR since it is a definite improvement/fix for our models.
I would like that we don't forget to test that the transforms are equivalent - can you open an issue and I will merge this PR. Thanks!

aymanhab:
Thanks much @aseth1, Issue #1347 has been opened to keep track of test improvements/needs.

aseth1:
@chrisdembia do you have any remaining concerns?

chrisdembia:
> @chrisdembia do you have any remaining concerns?

Nope!

aymanhab:
Great, thanks much @aseth1 and @chrisdembia

jenhicks:
@aymanhab Is there an associate issue to close?

aymanhab:
Nope, discovered during testing, trying to visualize old models from 3.3.

PR #1301, opened 2016-10-05
[travis] Do not reinstall cmake on OSX

chrisdembia:
Travis has updated their default OSX image, which now has cmake 3.6.2. 
This is sufficiently new for us, so we no longer need to install a newer 
cmake on our own. Not only this, but their update was causing our Travis build script to fail (we unlinked the existing cmake but homebrew does not re-install cmake b/c the latest version is already installed).

https://blog.travis-ci.com/2016-10-04-osx-73-default-image-live/

This will hopefully resolve the build failures on @aymanhab's #1299

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,10 +73,7 @@ before_install:
   # Stop build if comment contains [skip travis].
   - if $(git log -n1 --format="%B" | grep --quiet '\[skip travis\]'); then exit; fi 
   
-  - if [[ "$TRAVIS_OS_NAME" = "osx" ]]; then brew update; fi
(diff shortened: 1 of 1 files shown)


chrisdembia:
I cancelled the appveyor builds, since they are not affected by changes to only the travis script.

aseth1:
Looks reasonable to me. Merging since there are several PRs that are now failing and it would be good to have this fix in place ASAP. Thanks @chrisdembia

chrisdembia:
Thanks for merging. Looking at the tests, this PR did in fact address the `cmake`-related error, but there is still another related error causing OSX builds to fail.

PR #794, opened 2016-03-03
Save marker error to a .txt file, and update marker position .sto file name

ANKELA:
This update writes time, marker error squared, error rms and max error
to a .txt file with a filename appropriate to the trial data when the
"report_errors" flag is true.   It also modifies the name of the marker
location file to reflect the trial name when the
"report_marker_locations" flag is true.

PR #1311, opened 2016-10-10
[WIP] Matlab scripts for Sandbox

jimmyDunne:
A single PR for Matlab scripts used in testing and showcasing new functionality.


aseth1:
@jimmyDunne these examples and utilities will be super useful.  
I just glanced quickly to see if the flattened file being (written out)[https://github.com/opensim-org/opensim-core/pull/1311/files#diff-1fd8c205600d6b4d43fca075b6b0c56dR98] is being used by a Tool but could not find it. I likely missed it.

Knowing that the flattened file can be used for input forces to a Tool would be great.

jimmyDunne:
I haven't gotten there yet. Need to just print it off. Tools are next 😉

aseth1:
> I haven't gotten there yet. Need to just print it off. Tools are next

Sounds good. As you have a plan to use the flattened forces here,  we can merge #1278, which will get exercised by it. Thanks!

jimmyDunne:
@chrisdembia would you mind having a quick look at three files
- opensimTimeSeriesTableToMatlab.m
- matlabArrayToOpenSimTimeSeriesTableVec3.m
- matlabArrayToOpenSimTimeSeriesTable.m

They work, but I just feel like there is a better way to do this. Any thoughts would be appreciated?

OpenSim/Sandbox/MatlabScripts/opensimTimeSeriesTableToMatlab.m
```diff
@@ -1,12 +1,23 @@
...
+%% 
 data = struct;
-rowdata = zeros(nRow, 3);
 
+%%
+if strcmp( char(table.getClass), 'class org.opensim.modeling.TimeSeriesTableVec3')
+    rowdata = zeros(nRow, 3);
+elseif strcmp( char(table.getClass), 'class org.opensim.modeling.TimeSeriesTable')
+    rowdata = zeros(nRow, 1);
+end
```
chrisdembia: You may want to handle more than just these two types of tables. Consider using `flatten()` first, so you can handle all tables uniformly.

OpenSim/Sandbox/MatlabScripts/opensimTimeSeriesTableToMatlab.m
```diff
@@ -1,12 +1,23 @@
 function data = opensimTimeSeriesTableToMatlab(table)
 
 
+%%
+import org.opensim.modeling.*
+%%
 nCol = table.getNumColumns;
 nRow = table.getNumRows;
```
chrisdembia: I much prefer to use `()` at the end of method names, even though MATLAB lets you omit them. Otherwise, these look like variables, not functions.

OpenSim/Sandbox/MatlabScripts/opensimTimeSeriesTableToMatlab.m
```diff
@@ -33,15 +44,15 @@
        new_col_label = col_label(temp(end-1)+1:temp(end)-1);
```
chrisdembia: What will happen here if the table contains coordinate values? I think then data will contain only one column, called `data.value`. What do you think about converting the slashes into underscores, instead?

OpenSim/Sandbox/MatlabScripts/opensimTimeSeriesTableToMatlab.m
chrisdembia: You might also be able to make use of DataTable's `numComponentsPerElement()`.

OpenSim/Sandbox/MatlabScripts/matlabArrayToOpenSimTimeSeriesTable.m
```diff
@@ -0,0 +1,66 @@
+function table = matlabArrayToOpenSimTimeSeriesTableVec3(table_array)
```
chrisdembia: Is the input actually a matlab array? Or a struct?

chrisdembia: The filename is not consistent with the function name.

OpenSim/Sandbox/MatlabScripts/matlabArrayToOpenSimTimeSeriesTable.m
```diff
@@ -0,0 +1,66 @@
...
+%% Set the number of columns and the column labels
+colnames = StdVectorString;
+for i = 1 : length(labels)
+   
+    if ~strcmp(labels{i},'time')
+        colnames.add(labels{i})
+    end
+end
+
+table.setColumnLabels(colnames);
```
chrisdembia: This part looks good.

OpenSim/Sandbox/MatlabScripts/matlabArrayToOpenSimTimeSeriesTable.m
```diff
@@ -0,0 +1,66 @@
...
+
```
chrisdembia: It would be good to make sure all fields in `table_array` have the same number of entries.

OpenSim/Sandbox/MatlabScripts/matlabArrayToOpenSimTimeSeriesTable.m
```diff
@@ -0,0 +1,66 @@
...
+for ii  = 1 : length(table_array.(labels{1}))
+
+    table.appendRow(0, row)
+    
+end
+    
+%% write values to the OpenSim table
+for i  = 0 : length(labels) -1
+
+    if strcmp(labels(i+1),'time')
```
chrisdembia: Is there a reason why parentheses are used here, but curly braces are used on line 44?

OpenSim/Sandbox/MatlabScripts/matlabArrayToOpenSimTimeSeriesTable.m
```diff
@@ -0,0 +1,66 @@
...
+    column = table.getDependentColumnAtIndex(i);
+    
+   for iRow = 0 : table.getNumRows - 1
+
+        column.set( iRow , table_array.(labels{i+1})(iRow+1)  );    
+   end 
+  
+end 
+
+%% TODO: Decide to output a timerseries table or not
```
chrisdembia: Yes, I think TimeSeriesTable. After all, you look for the `time` column. We do not yet really have use cases that I know of for non-time DataTables, so we can hold off on converters for those.

jimmyDunne:
Thanks for looking through. I am going to write some documentation but wanted to get the design right first. I think using Matlab's table class may be worthwhile. The other is to use [Matlab's Times Series Objects](https://www.mathworks.com/help/matlab/data_analysis/time-series-objects.html). 

I hadn't seen `numComponentsPerElement()`, that will actually be a great help to make it more generic. 

The other thing to consider is the data types people would want to shove into an OpenSim table. I think we should only support 2-3 data formats as inputs. For example, a common input type may be an array of doubles and an accompanying vector of column labels.

chrisdembia:
> The other is to use Matlab's Times Series Objects.

Wow! That looks pretty cool. Seems like that could be even better.



> The other thing to consider is the data types people would want to shove into an OpenSim table. I think we should only support 2-3 data formats as inputs. For example, a common input type may be an array of doubles and an accompanying vector of column labels.

I'm not sure I understand. Are you saying it's sufficient to support only TimeSeriesTable and TimeSeriesTableVec3?

jenhicks:
Ready for review

chrisdembia:
@jimmyDunne I see you added the Dynamic Walker files into the sandbox. Is there a reason to use the new copy instead of modifying the Dynamic Walker files that are already in the repository in [this folder](https://github.com/opensim-org/opensim-core/tree/master/Bindings/Java/Matlab/Dynamic_Walker_Example)?

OpenSim/Sandbox/MatlabScripts/timesSeriesTable.m
```diff
@@ -0,0 +1,195 @@
...
+
+elseif  strcmp( class(generictable), 'tscollection') 
+   
+    output_table = tsc2table(generictable);
+    
+else
+    error('input must either be an OpenSim TimeSeriesTable or a Matlab tscollection')
+end
+
+end
```
chrisdembia: Is there a particular reason for using 1 function to support both directions of conversions? I think I prefer having two separate functions. That way, user code is less ambiguous:

```matlab
table_out = timeSeriesTable(table_in)  % which one is the OpenSim DataTable?
```

What about `convertFromOpenSimTimeSeriesTable()` and `convertToOpenSimTimeSeriesTable()`?

OpenSim/Sandbox/MatlabScripts/timesSeriesTable.m
```diff
@@ -0,0 +1,195 @@
+function output_table = timesSeriesTable(generictable)
+%% // Convert between OpenSim and Matlab Timeseries 
+%  
+%   Inputs can be either an OpenSim TimeSeriesTable, TimesSeriesTableVec3,
+%   or a Matlab times series collection. 
+%
+%   Outputs from OpenSim TimesSeriesTable's will be a Matlab times series
+%   collection.
```
chrisdembia: We should maybe give a link to MATLAB documentation about `tscollection` and `timeseries`, and maybe also have some examples of how to use it.

OpenSim/Sandbox/MatlabScripts/timesSeriesTable.m
```diff
@@ -0,0 +1,195 @@
...
+    %% data size
+    dataSize = tsc.size();
+    nRow = dataSize(1); nlabels = dataSize(2);
+    [m, nCol] = size(tsc.(labelnames{1}).Data);
+    
+    %% empty opensim table
+    if nCol == 1
+        table = DataTable();
+    elseif nCol == 3
+        table = DataTableVec3();
```
chrisdembia: I think these should be TimeSeriesTables, so that `tsc2table(table2tsc(T))` gives back something of the same type as T.

OpenSim/Sandbox/MatlabScripts/timesSeriesTable.m
```diff
@@ -0,0 +1,195 @@
...
+            for iCol = 1 : nlabels
+
+                rowdata = tsc.(labelnames{iCol}).Data(iRow,:);
+
+                row.set(iCol-1, rowdata);
+            end
+        
+            table.appendRow(iRow-1, row);
+            % set the time value    
+            timeColumn.set(iRow-1, double(tsc.Time(iRow)) );
```
chrisdembia: If using a TimeSeriesTable, you should be able to combine the two lines above.

OpenSim/Sandbox/MatlabScripts/timesSeriesTable.m
chrisdembia: EDIT: I now see that you do eventually convert to TimeSeriesTables. Is there a reason these cannot start as TimeSeriesTables? Is something about the interface preventing this?

jimmyDunne:
I got rid of the times series objects. Instead I just use structs.

jenhicks:
@jimmyDunne Can you provide a summary of what is included and key things to review?

Bindings/Java/Matlab/Dynamic_Walker_Example/DesignMainStarter_Manager.m
```diff
@@ -0,0 +1,75 @@
...
+reporter.updInput('inputs').connect( model.getCoordinateSet().get(3).getOutput('value'), 'Hip_Angle_L')
+reporter.updInput('inputs').connect( model.getCoordinateSet().get(4).getOutput('value'), 'Hip_Angle_R')
+reporter.updInput('inputs').connect( model.getCoordinateSet().get(5).getOutput('value'), 'Knee_Angle_L')
+reporter.updInput('inputs').connect( model.getCoordinateSet().get(6).getOutput('value'), 'Knee_Angle_R')
+
+% add the reporter to the model
+model.addComponent(reporter) 
+
+% re-initialize the state
+state = model.initSystem();
```
tkuchida: Should L38 be removed since `state` isn't being used (and the coordinates will have the wrong initial values)? However, it seems like you need to call `initSystem()` after `model.addComponent(reporter);`. Perhaps first create/add the reporter, then call `initSystem()`, and then set the coordinate values? Further, it may be clearer to configure the model first, and set the initial conditions immediately before simulating.

Bindings/Java/Matlab/Dynamic_Walker_Example/DesignMainStarter_Manager.m
```diff
@@ -0,0 +1,75 @@
...
+manager.integrate(s);
+
+% get the table last value of pelvis X (distance down the run way)
+table = reporter.getTable();
+nRows = table.getNumRows();
+
+%display(table.getDependentColumnAtIndex(0).getElt(nRows-1,0))
+
+% need to get close to 3 meters
+J = 3 - table.getDependentColumnAtIndex(0).getElt(nRows-1,0) ;
```
tkuchida: If you want "close to 3 meters" as indicated in the comment, you need `abs()` to penalize simulations that exceed 3 m. Perhaps simply set J to the negative of the distance traveled, where lower is better?

Bindings/Java/Matlab/Dynamic_Walker_Example/Main_WalkerForwardSim.m
```diff
@@ -0,0 +1,68 @@
...
+osimModel = Model('../Model/WalkerModel.osim');
+
+% Use the visualizer (must be done before the call to init system)
+osimModel.setUseVisualizer(true);
+
+% Initialize the system and get the initial state
+osimState = osimModel.initSystem();
+
+% Set the initial states of the model
+editableCoordSet = osimModel.updCoordinateSet();
```
tkuchida: `setValue()` is a const method, so AFAIK you don't need the CoordinateSet to be editable (i.e., you can use `getCoordinateSet()` and remove "editable").

Bindings/Java/Matlab/Dynamic_Walker_Example/DesignMainStarter_Manager.m
```diff
@@ -0,0 +1,75 @@
...
+import org.opensim.modeling.*;
+
+% Open a Model by name
+model = Model('../Model/WalkerModel.osim');
+
+% Initialize the system and get the initial state
+s = model.initSystem();
+
+% Set the initial states of the model
+editableCoordSet = model.updCoordinateSet();
```
tkuchida: `setValue()` is a const method, so AFAIK you don't need the CoordinateSet to be editable (i.e., you can use `getCoordinateSet()` and remove "editable").

Bindings/Java/Matlab/Dynamic_Walker_Example/DesignMainStarter_Manager.m
```diff
@@ -0,0 +1,75 @@
...
+editableCoordSet.get(3).setValue(s, states(1)); % LHIP
+editableCoordSet.get(4).setValue(s, states(1)); % RHIP
+editableCoordSet.get(5).setValue(s, states(1)); % LKnee
+editableCoordSet.get(6).setValue(s, states(1)); % RKnee
```
tkuchida: The index into `states` should presumably be `1` for only one of these coordinates.

Bindings/Java/Matlab/Dynamic_Walker_Example/Main_WalkerForwardSimWithControls.m
```diff
@@ -0,0 +1,78 @@
...
+% the motion data.  The model has an coordinate actuator on the left knee
+% and a position controller to set the angle to -135 degrees.
+
+% Make sure to run the AddCoordinateActuator.m script first!
+
+% Import
+import org.opensim.modeling.*;
+
+% Open a Model by name
+display('Make sure you run the script file "AddCoordinateActuator.m" first.')
```
tkuchida: Instead, perhaps read the model and throw an error if the expected components are missing?

Bindings/Java/Matlab/Dynamic_Walker_Example/Main_WalkerForwardSimWithControls.m
```diff
@@ -0,0 +1,78 @@
...
+osimModel = Model('../Model/WalkerModelTerrain_CoordAct.osim');
+
+% Use the visualizer (must be done before the call to init system)
+osimModel.setUseVisualizer(true);
+
+% Initialize the system and get the initial state
+osimState = osimModel.initSystem();
+
+% Set the initial states of the model
+editableCoordSet = osimModel.updCoordinateSet();
```
tkuchida: Does not need to be editable AFAIK.

Bindings/Java/Matlab/Dynamic_Walker_Example/RunForwardAnalysis.m
```diff
@@ -0,0 +1,58 @@
+import org.opensim.modeling.*
+
+% Open Model
+model= Model('../Model/WalkerModel.osim');
+
+% Add Analyses to the Model
+forceReporter = ForceReporter();
+model.addAnalysis(forceReporter);
```
tkuchida: Conflicts with name of script. Perhaps rename script to "RunForwardSimulation.m"?

Bindings/Java/Matlab/Dynamic_Walker_Example/RunForwardAnalysis.m
```diff
@@ -0,0 +1,58 @@
...
+
+% Initialize the underlying computational system
+state = model.initSystem();
+
+% Run a fwd simulation using the manager
+state = model.initSystem();
```
tkuchida: Already called on L11.

OpenSim/Sandbox/MatlabScripts/c3d_reading.m
```diff
@@ -1,76 +1,75 @@
...
-% % Print the (unrotated) markers to trc file
-% trcfileadapter = TRCFileAdapter();
-% trcfileadapter.write(markers,'test_walking.trc');
-
+% get the number of markers and rows
+nMarkers = markers.getNumColumns();
+markers_nRows = markers.getNumRows();
+% get the array of the metadata key names
+metadDataKeys_markers = markers.getTableMetaDataKeys();
+markerrate = str2num(markers.getTableMetaDataAsString(metadDataKeys_markers.get(0)));
```
tkuchida: Would be helpful to have consistent variable names (`nMarkers` vs. `markers_nRows` vs. `metadDataKeys_markers` vs. `markerrate`).

OpenSim/Sandbox/MatlabScripts/c3d_reading.m
```diff
@@ -111,13 +106,35 @@
...
+postfix.add('_x');
+postfix.add('_y');
+postfix.add('_z');
+
+% flatten the Vec3 table to a doubles table
+forces_flattened = forces.flatten(postfix);
+forces_rot_flattened = forces_rotated.flatten(postfix);
+
+% make a sto adapter and write the forces table to file.
+stofileadapater = STOFileAdapter();
```
tkuchida: Is it possible to do `STOFileAdapter().write(...);`? Conceptually, you're calling a static method on STOFileAdapter.

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,165 @@
...
+% Generate an instance of the model
+model = Model(modelFilePath);
+
+% Get the number of coordinates  and a handle to the coordainte set
+nCoord = model.getCoordinateSet.getSize();
+coordSet = model.getCoordinateSet();
+
+% Evaluate the ground body and get the mass center
+groundBodyName = model.getGround().getName();
+groundJoint = model.getJointSet.get(0);
```
tkuchida: I'm not sure it's still guaranteed that the first joint in the JointSet will be connected to Ground.

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,165 @@
...
+
+% Create some empty vec3's for later.
+massCenter = Vec3();
+axisValues = Vec3();
+
+% Create an empty Force set
+forceSet = ForceSet();
+% get the state
```
tkuchida: Create the underlying computational System and return a handle to the State

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,165 @@
+function createActuatorsFile
+%createActuatorsFile   Build and Print an OpenSim Actuator File from a Model
```
tkuchida: I think users will find this script very handy 🔧

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,165 @@
...
+    coordinate = coordSet.get(iCoord);
+    % If the coordinate is locked don't add an actuator.
+    if coordinate.getLocked(state)
+        continue
+    end
+    % If the coordinate is prescribed, don't add an actuator.
+    if coordinate.isPrescribed(state)
+        continue
+    end
+    % If the coodinate is constrained, don't add an actuator
```
tkuchida: Only need to check `isConstrained()`, not `getLocked()` or `isPrescribed()` (a Coordinate is constrained if it's locked, prescribed, or dependent on other Coordinates).

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,165 @@
...
+    if coordinate.isConstrained(state)
+        continue
+    end
+
+    % get the joint, parent and child names for the coordiante
+    joint = coordinate.getJoint;
```
tkuchida: `getJoint` -> `getJoint()`

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,165 @@
...
+
+               % Build a torque actuator for a rotational coordinate
+               if strcmp(motion, 'Rotational')
+                   newActuator = TorqueActuator(joint.getParentFrame(),...
+                                         joint.getParentFrame(),...
+                                         axisValues,...
+                                         1);
+    
+               % Build a point actuator for a translational coordainte.
+               else
```
tkuchida: The comment doesn't match the code. There are actually 4 MotionTypes, 3 of which might be encountered here (Rotational, Translational, and Coupled). The `else` should be an `elseif` and presumably Coupled should be handled as well.

OpenSim/Sandbox/MatlabScripts/muscleCurves.m
```diff
@@ -0,0 +1,259 @@
...
+    %%%
+    while isempty(name)
+        
+       
+        string = input('feed me a muscle name!?', 's');
+        
+        % get a valid name or 'exit' 
+        nameIndex = find( cellfun(@(s) ~isempty(strmatch(string, s)), musclelist) == 1);
+        
+        if strmatch(string, 'exit');
```
tkuchida: Logically, this check should occur before `find()` on L68. Also, the information that the user can type 'exit' to stop should appear in the prompt (L65).

OpenSim/Sandbox/MatlabScripts/muscleCurves.m
```diff
@@ -0,0 +1,259 @@
...
+        display('thanks for ending this program. See you Laterz!')
+        close all
+        clc
+        return
+    end
+        
+    %% Muscle Coordinate finder
+    %   Find the coordinate's that each muscle crosses. This is done by
+    %   examining the moment arm contribution of the muscle across all
+    %   coordinates. A muscle will contribute to any coodinate when the moment
```
tkuchida: Technically this is not true, since the muscle may happen to have a moment arm of zero at the current pose but not in general.

OpenSim/Sandbox/MatlabScripts/muscleCurves.m
```diff
@@ -0,0 +1,259 @@
...
+    % the state. 
+    momentArm_aCoord(k+1) = muscle.computeMomentArm(state,aCoord);
+end
+
+% round the numbers. This is needed because at some coordinates there
+% are moments generated at the e-18 level. This is most likely
+% numerical error. So to deal with this I round to 4 decimal points. 
+x = round((momentArm_aCoord*1000))/1000;
+% Find the coordinate index's that are non-zero
+muscCoord = find(x ~= 0)-1;
```
tkuchida: Could just use `find()` to locate entries of momentArm_aCoord that are within/beyond whatever your tolerance is (i.e., without rounding explicitly).

OpenSim/Sandbox/MatlabScripts/osimTableFromStruct.m
```diff
@@ -0,0 +1,131 @@
...
+%  Output is an OpenSim TimesSeriesTable
+
+% ----------------------------------------------------------------------- %
+% The OpenSim API is a toolkit for musculoskeletal modeling and           %
+% simulation. See http://opensim.stanford.edu and the NOTICE file         %
+% for more information. OpenSim is developed at Stanford University       %
+% and supported by the US National Institutes of Health (U54 GM072970,    %
+% R24 HD065690) and by DARPA through the Warrior Web program.             %
+%                                                                         %   
+% Copyright (c) 2005-2012 Stanford University and the Authors             %
```
tkuchida: 2012 -> 2016

OpenSim/Sandbox/MatlabScripts/osimTableFromStruct.m
```diff
@@ -0,0 +1,131 @@
...
+% implied. See the License for the specific language governing            %
+% permissions and limitations under the License.                          %
+% ----------------------------------------------------------------------- %
+
+% Author: James Dunne, Shrinidhi K. Lakshmikanth, Chris Dembia, Tom Uchida,
+% Ajay Seth, Ayman Habib, Jen Hicks. 
+
+%%
+import org.opensim.modeling.*
+%% get n col and rows
```
tkuchida: Move to L43?

OpenSim/Sandbox/MatlabScripts/osimTableFromStruct.m
```diff
@@ -0,0 +1,131 @@
...
+% get the index for the time array
+timeIndex = find(cellfun(@(s) ~isempty(strfind('time', s)),lower(labelnames)));
+
+if timeIndex == 1
+    [nRows,nCol] = size(structdata.(labelnames{2}));
+else
+    [nRows, nCol] = size(structdata.(labelnames{1}));
+end
+
+% if there is more than three columns, make an exception.
```
tkuchida: "Throw an error if the number of columns is neither 1 nor 3"

OpenSim/Sandbox/MatlabScripts/osimTableFromStruct.m
```diff
@@ -0,0 +1,131 @@
...
+
+            row.set(iCol-1, rowdata);
+        end
+
+        table.appendRow(iRow-1, row);
+        % set the time value    
+        timeColumn.set(iRow-1, structdata.(labelnames{timeIndex})(iRow) );
+    end
+
+elseif nCol == 3
```
tkuchida: At this point, `nCol` is guaranteed to be either 1 or 3.

OpenSim/Sandbox/MatlabScripts/osimTableToStruct.m
```diff
@@ -0,0 +1,88 @@
...
+
+%% make an empty data structure
+data = struct();
+
+%%
+if strcmp( char(table.getClass), 'class org.opensim.modeling.TimeSeriesTableVec3')
+    rowdata = zeros(nRow, 3);
+elseif strcmp( char(table.getClass), 'class org.opensim.modeling.TimeSeriesTable')
+    rowdata = zeros(nRow, 1);
+end
```
tkuchida: `else, error(...);`?

OpenSim/Sandbox/MatlabScripts/pendulum_simulation.m
```diff
@@ -0,0 +1,92 @@
...
+%% run in a loop, writing over the file each time. 
+for n = 1 : 40
+    % run a fwd simulation using the manager
+    state = model.initSystem();
+    manager = Manager(model);
+    manager.setInitialTime(0);
+    manager.setFinalTime(2);
+    manager.integrate(state);
+
+    % get the coordinate table from the reporter and write to file
```
tkuchida: Perhaps write the files on only the last iteration?

.gitignore
```diff
@@ -19,3 +19,6 @@ dependencies/*
 
 # Ignore log files from OpenSim crashing
 *.log
+
+# Ignore OSX attribute files
```
jimmyDunne: .DS_Store is some OS X based file that gets autogenerated by the OS X file system. Every time I make changes to the repo on an OSX machine; GIT recognizes .DS_Store as either an addition or modified file.

chrisdembia: The appearance of these files in git might be due to mixing operating systems, like accessing the repo, stored on an OS X drive, via Windows in Parallels?

You can maybe add .DS_Store to the .gitignore file.

OpenSim/Sandbox/MatlabScripts/muscleCurves.m
jimmyDunne: I updated this section to check the moment arm at three locations within the range of the coordinate.

chrisdembia:
Notes from lab meeting, and some of my own additional comments:

- MATLAB "dcm" type (direction cosine matrix)
- Simbody might have an operator to rotate a Vector_<Vec3>.
- createActuatorsFile: should be clear that the point/torque actuator is only added if the *parent* is ground. I think that behavior is fine (to ignore the case where the child is ground), but should be documented at the top of the file
- Consistent script naming: underscores or lowerCamelCase?
- muscleCurves: make sure the behavior is consistent with GUI.
- Could piggyback on Muscles' writing curves to CSV.
- Two underscores for slashes in structs.
- (later on) Add containers.Map (key value pair) to struct for metadata.
- Create MATLAB class for data tables. 
- Could write to a file to speed up to/from struct.
- DesignMainStarter_Manager should maybe document why a user would use this instead of the other DesignMainStarter.m, and also mention the differences (e.g, the Manager doesn't allow setting controls via a matlab function).
- Delete RunForwardTool.m
- IK errors file: https://github.com/opensim-org/opensim-core/pull/794/files
- Placement and styling of the license block should be consistent.

It's good to gitignore the DS_Store file, but they are still committed to the repo. Make sure to `git rm` them (or `git rm --cached` to not delete them from your hard drive). Mentioning that file name in gitignore does not automatically remove files from the repo that have already been committed.

PR #1314, opened 2016-10-10
Gui visualize contacts

aymanhab:
Create DecorativeGeometry for all types of ContactGeometry all the way to Visualizer.


aymanhab:
Bogus and unrelated test failures, ready for review @chrisdembia and/or @aseth1

OpenSim/Simulation/Model/ContactHalfSpace.cpp
```diff
@@ -59,4 +60,31 @@ SimTK::ContactGeometry ContactHalfSpace::createSimTKContactGeometry() const
...
+
+    // B: base Frame (Body or Ground)
+    // F: PhysicalFrame that this ContactGeometry is connected to
+    // P: the frame defined (relative to F) by the location and orientation
+    //    properties.
+    if (hints.get_show_contact_geometry()) {
+        const auto& X_BF = getFrame().findTransformInBaseFrame();
+        const auto& X_FP = getTransform();
+        const auto X_BP = X_BF * X_FP;
+        geometry.push_back(SimTK::DecorativeBrick(Vec3{.005, 0.5, 0.5})
```
chrisdembia: Do you think it makes sense to let users customize these dimensions? For example, a model of a rat or cat would be at a substantially different scale than that of a human.

chrisdembia: Would it be possible to show arrows to show the "up" direction of the half-space? The arrows that one sees in the GUI are very helpful.

OpenSim/Simulation/Model/ContactMesh.cpp
```diff
@@ -115,7 +118,7 @@ SimTK::ContactGeometry::TriangleMesh* ContactMesh::
     file.close();
     mesh.loadFile(filename);
     if (restoreDirectory) IO::chDir(savedCwd);
-        
+    _decorativeGeometry.reset(new SimTK::DecorativeMesh(mesh));
     return new SimTK::ContactGeometry::TriangleMesh(mesh);
```
chrisdembia: This is a little weird to me; the two pointers are handled differently. I would prefer if `loadMesh()` were changed so its return type is null, and that the last line is changed to:

```
_geometry.reset(new SimTK::ContactGeometry::TriangleMesh(mesh));
```

And that the doxygen for this method is updated accordingly.

OpenSim/Simulation/Model/ContactMesh.cpp
```diff
@@ -145,7 +148,7 @@ void ContactMesh::generateDecorations(bool fixed, const ModelDisplayHints& hints
         const auto& X_BF = getFrame().findTransformInBaseFrame();
         const auto& X_FP = getTransform();
         const auto X_BP = X_BF * X_FP;
-        geometry.push_back(SimTK::DecorativeMeshFile(get_filename())
+        geometry.push_back(SimTK::DecorativeMesh(*_decorativeGeometry)
```
chrisdembia: Ah....so is the reason that the class holds onto a `DecorativeMesh` that it is cheaper to copy construct it than to recreate it from a file for every visualization frame? Could the code look like this:

```cpp
void loadMesh() {
...
_decorativeGeometry.reset(new SimTK::DecorativeMesh(mesh).setTransform(X_BFP).setRepresentation(...)...
}

void generateDecorations(...) {
...
geometry.push_back(*_decorativeGeometry);
}
```
?

If that doesn't make sense for some reason, then it's fine to leave this as is...the unnecessary copy is bothering me though. Either way, I'm fine with how this is.

OpenSim/Tools/Test/vis_BuiltinGeometry.txt
```diff
@@ -118,9 +118,3 @@ DecorativeLine:~[0,0.05,0.35]~[0.19999,0.0480388,0.05] bodyId:0 color:~[0.05,0,0
 [0,0,1,0]
 [0,0,0,1]
 
-DecorativeMeshFile:blockMesh.obj  bodyId:1 color:~[0,1,1] indexOnBody:-1 Opacity:1 Rep:2 Scale:~[-1,-1,-1] Transform:
-[1,0,0,0]
-[0,1,0,0]
-[0,0,1,0]
-[0,0,0,1]
```
chrisdembia: I haven't been following the evolution of this test case too well...how come this block isn't replaced by a similar one for `DecorativeMesh` (instead of `DecorativeMeshFile`? And where is the block for a ContactHalfSpace? Correspondingly, how come it is okay to remove the contact geometry from the `osim` file? 

Is this all tested differently now?

OpenSim/Simulation/Model/ContactHalfSpace.cpp
aymanhab: Great feedback, and comments @chrisdembia 
The Contact visualization pathway uses generateDecorations (which has little or no customization except maybe thru display hints) In most cases that's enough because the dimensions are finite and map to Geometry. Not in this case. I could add a half_plane_size to display_hints and it will apply to all HalfSpaces, other option is to have a property Vec3 display_size in ContactHalfSpace, or what I conjecture would work in the new visualizer, is allow the user to change the size and texture etc. in visualizer and keep track of that outside the model (after all this's a kind of appearance related change that depends on Visualizer capabilities) rather than a part of the model. Would love to hear your thoughts.

aymanhab: Old visualizer was using a mesh file that has arrows, We can use that here instead of the Brick but stretching could be undesirable. Other possible options are: Make the material one-sided so the top shows but the bottom doesn't, or use different materials/textures/colors for top/bottom.

OpenSim/Simulation/Model/ContactMesh.cpp
aymanhab: Yes, I debated this. Basically I didn't want the DecorativeMesh to be created and displayed unless the SimTK::ContactGeometry is created as well since otherwise the display is misleading that there's Contact while the model actually doesn't have one. That also is the reason it was removed from  testVisualization because no ContactForce was defined in the model.

OpenSim/Tools/Test/vis_BuiltinGeometry.txt
aymanhab: As explained above the current Mesh display was misleading, I didn't want to make major change to testVisualization until my other PRs that modify it are merged in.  It maybe best to beef up the test when all ContactGeometry, and WrapGeometry are finalized instead of continuously changing/updating the tests but thanks for pointing out the potential hole in the current test.

aymanhab:
Any thoughts or feedback in response @chrisdembia or other interested parties @tkuchida @aseth1

OpenSim/Tools/Test/vis_BuiltinGeometry.txt
aymanhab: Now that another PR that changes testVisualization is in, I'll beef up the Model and tests to account for Contacts

OpenSim/Simulation/Model/ContactHalfSpace.cpp
chrisdembia: > use different materials/textures/colors for top/bottom

I like this idea! It's fine with me to leave this for another PR, but I do think users would really like such an indication. I definitely used to rely on those 4 little arrows.

OpenSim/Simulation/Model/ContactMesh.cpp
chrisdembia: I'm not sure I understand; it seems like changing the return type of this function would still satisfy the requirement that the DecorativeMesh is only created if the ContactGeometry is created (with which I completely agree). We can chat in person.

OpenSim/Tools/Test/vis_BuiltinGeometry.txt
chrisdembia: > I'll beef up the Model and tests to account for Contacts

Do you mean to do that in this PR or in a follow-up?

OpenSim/Simulation/Model/ModelComponentSet.h
```diff
@@ -91,9 +91,8 @@ OpenSim_DECLARE_CONCRETE_OBJECT_T(ModelComponentSet, T, Set<T>);
     {
     }
 
-#ifndef SWIG
     /** Does this Set have a Model associated with it? */
-    bool hasModel() const { return _model.get(); }
+    bool hasModel() const { return _model.get()!=nullptr; }
```
chrisdembia: 👍

OpenSim/Simulation/Model/ContactMesh.cpp
chrisdembia: Ah I see...but as is, the model display hints do not affect the `SimTK::DecorativeMesh` that gets sent to the visualizer. However, as this may change if you introduce a display hint for the overall size of the mesh, then it's fine with me to leave in the extra copying.

What happens when `get_show_contact_geometry() == true` but the Force `isDisabled = true`? In this case, it seems like `_decorativeGeometry` will be null, but this code will still try to use it.

chrisdembia:
I've responded to your responses. My main concern now is the behavior of `generateDecorations()` when the `get_show_contact_geometry() == true` but `Force::_isDisable == true`.

OpenSim/Simulation/Model/ContactHalfSpace.cpp
aymanhab: Great, thanks @chrisdembia I was already moving in this direction by including marker_color in display_hints (separate PR) so I can add marker_size and half_space_display_size.

OpenSim/Simulation/Model/ContactMesh.cpp
aymanhab: Agreed, however I think that needs to be handled somewhere else so that the component is not processed during traversal, otherwise each component will deal with it in an ad-hoc way.

jenhicks:
@carmichaelong Can you review?

aymanhab:
Added defensive check to avoid possible crash if mesh was never loaded, beefed up test case to include a model with Contacts. Ready for another -review.

carmichaelong:
Generally looks good to me. Just one small comment: In the methods `testVisModel` and `testVisModelAgainstStandard`, seems like if `visualDebug = true`, then the visualizer (or console) should probably wait for user input before moving to the next test.

aymanhab:
@carmichaelong Thanks for the review, actually since the visualizer is non-blocking, I find it convenient when visual debugging  to have all windows open at the same time to quickly see if something is wrong rather than babysit one case at a time, but I don't feel strongly about it since it's not the default out-of-the-box behavior. What are your thoughts?

carmichaelong:
When I flipped the flags to "true" and ran the test, all of the visualizer windows would open briefly but then close themselves when each part of the test passes (so the problem isn't that I want some input from the user, just that the windows don't show for long enough for me to inspect unless I added some lines). Is this a different behavior than your machine? I just thought it was weird to switch the flags to "true" but couldn't actually take a look at them!

aymanhab:
Thanks @carmichaelong it turns out I always had a breakpoint at the end while visual debugging. Fixed now. :+1:

aymanhab:
Changes done per feedback, ready for review. Bogus test failures. Customization through DisplayHints is on a separate PR.

aymanhab:
@chrisdembia and @carmichaelong please let me know if your concerns were addressed.

carmichaelong:
With the line `std::cin >> c;` I think you still would have to hit "enter" for it to go away and not just any key? Also, since `c` is a `char`, what happens if I put in a longer string and then hit enter?

carmichaelong:
LGTM (sorry for the nit-picky stuff). I think it's ready whenever issues are cleared up with Chris.

jenhicks:
@aymanhab Can you verify that you're not getting segfaults for the actions Chris noted? If so, then I think it is OK to merge.

OpenSim/Simulation/Model/ContactMesh.cpp
chrisdembia: This is the main comment I was hoping could still be addressed.

chrisdembia:
> @aymanhab Can you verify that you're not getting segfaults for the actions Chris noted? If so, then I think it is OK to merge.

I think @aymanhab did attempt to verify this but ran into other issues. @aymanhab you can merge this. However,  I am not too satisfied with `loadMesh()` returning the mesh pointer.

jenhicks:
Thanks @chrisdembia for reviewing. If you try this out in your workflows, let us know if issues arise. I'm sure @aymanhab is happy to chat more, too.

aymanhab:
Thanks much @chrisdembia

Issue #1255, opened 2016-09-15
Order of PhysicalOffsetFrames are not adequately handled


aseth1:
We have been fortunate in most cases to be applying `PhysicaloOffsetFrames` (POFs) to Bodies and Ground.  Bodies and Ground are part of the Multibody tree that we follow to appropriately add components to the MultibodySystem. In `testFrames` we also add a POF to another POF, but in that case we happen to be adding them to the Model in the correct order (proximal POF then distal POF).

If we invert the order in which they are added to the Model (distal prior to proximal) then `initSystem` fails to add the distal POF to the System.

For example, The following code fails:
```cpp
    Model pendulum("double_pendulum.osim");

    SimTK::Transform X_RO;
    X_RO.setP(SimTK::Vec3(0.1, 0.2, 0.3));
    X_RO.updR().setRotationFromAngleAboutAxis(SimTK::Pi / 4.0, SimTK::ZAxis);

    // create PhysicalOffsetFrames Distal and Proximal to identify their
    // relative location to the w.r.t a Body in the Model.
    PhysicalOffsetFrame* offsetFrameDistal = new PhysicalOffsetFrame();
    offsetFrameDistal->setName("offsetFrameDistal");
    offsetFrameDistal->setOffsetTransform(X_RO);
    pendulum.addComponent(offsetFrameDistal);

    PhysicalOffsetFrame* offsetFrameProximal = new PhysicalOffsetFrame();
    offsetFrameProximal->setName("offsetFrameProximal");
    offsetFrameProximal->setOffsetTransform(~X_RO);
    pendulum.addComponent(offsetFrameProximal);

    // Now attach them such that offsetFrameProximal is attached to rod2 of
    // the pendulum and offsetFrameDistal  is attached to offsetFrameProximal 
    offsetFrameProximal->setParentFrame(pendulum.getComponent<Body>("rod2"));
    offsetFrameDistal->setParentFrame(*offsetFrameProximal);

    SimTK::State& s = pendulum.initSystem();
```
with the following Exception:
```
Assigned an invalid SimTK::MobilizedBodyIndex
        In object 'offsetFrameDistal' of type PhysicalOffsetFrame.
        In file G:\OpenSimGit\opensim-core\OpenSim\Simulation\Model\PhysicalFrame.cpp:73
        In function 'setMobilizedBodyIndex'
```

[...]

PR #1315, opened 2016-10-11
Handle order of PhysicalOffsetFrame

aseth1:
Addresses #1255.

Only two significant changes:
 1. Introduced a test case in `testFrames` to produce a failure when `PhysicalOffsetFrame`s are added to the `Model` in an order that does not correspond to the Multibody topology.
 2. `Model::extendConnectToModel` now traverses its `PhysicalOffsetFrame` sub-components so that its parent `PhysicalFrame`s are queued up first to establish proximal to distal order. 

Should be a short review.

### DIFF
--- a/OpenSim/Simulation/Model/Model.cpp
+++ b/OpenSim/Simulation/Model/Model.cpp
@@ -733,11 +733,29 @@ void Model::extendConnectToModel(Model &model)
             "Unrecognized loop constraint type '" + joint.getConcreteClassName() + "'.");
     }
 
-    // Now include all remaining Components beginning with PhysicalFrames
+    // Now include all remaining Components beginning with PhysicalOffsetFrames
     // since Constraints, Forces and other components can only be applied to
-    // PhyicalFrames.
+    // PhyicalFrames, which include PhysicalOffsetFrames.
+    // PhysicalOffsetFrames require that their parent frame (a PhysicalFrame)
+    // be added to the System first. So for each PhysicalOffsetFrame locate its
+    // parent and verify its presence in the _orderedList otherwise add it first.
     auto poFrames = getComponentList<PhysicalOffsetFrame>();
     for (const auto& pof : poFrames) {
+        // Ground and Body type PhysicalFrames are handled by the Multibody graph
+        // PhysicalOffsetFrame can be listed in any order and be attched
+        // to any other PhysicalOffsetFrame, so we need to find their parent(s)
+        const PhysicalOffsetFrame* parentPof =
+            dynamic_cast<const PhysicalOffsetFrame*>(&pof.getParentFrame());
+        std::vector<const PhysicalOffsetFrame*> parentPofs;
+        while (parentPof) {
+            parentPofs.push_back(parentPof);
+            parentPof =
+                dynamic_cast<const PhysicalOffsetFrame*>(&parentPof->getParentFrame());
+        }
+        while (parentPofs.size()) {
+            setNextSubcomponentInSystem(*parentPofs.back());
+            parentPofs.pop_back();
+        }
         setNextSubcomponentInSystem(pof);
     }
 
--- a/OpenSim/Simulation/Model/Model.cpp
+++ b/OpenSim/Simulation/Model/Model.cpp
@@ -775,11 +775,13 @@ void Model::extendConnectToModel(Model &model)
     // PhysicalOffsetFrames require that their parent frame (a PhysicalFrame)
     // be added to the System first. So for each PhysicalOffsetFrame locate its
     // parent and verify its presence in the _orderedList otherwise add it first.
-    auto poFrames = getComponentList<PhysicalOffsetFrame>();
-    for (const auto& pof : poFrames) {
-        // Ground and Body type PhysicalFrames are handled by the Multibody graph
-        // PhysicalOffsetFrame can be listed in any order and be attched
+    auto poFrames = updComponentList<PhysicalOffsetFrame>();
+    for (auto& pof : poFrames) {
+        // Ground and Body type PhysicalFrames are included in the Multibody graph
+        // PhysicalOffsetFrame can be listed in any order and may be attached
         // to any other PhysicalOffsetFrame, so we need to find their parent(s)
+        // in the tree and add them first.
+        pof.connectToModel(*this);
         const PhysicalOffsetFrame* parentPof =
             dynamic_cast<const PhysicalOffsetFrame*>(&pof.getParentFrame());
         std::vector<const PhysicalOffsetFrame*> parentPofs;
--- a/OpenSim/Simulation/Model/Model.cpp
+++ b/OpenSim/Simulation/Model/Model.cpp
@@ -788,8 +788,8 @@ void Model::extendConnectToModel(Model &model)
         while (parentPof) {
             const auto found =
                 std::find(parentPofs.begin(), parentPofs.end(), parentPof);
-            OPENSIM_THROW_IF_FRMOBJ(found != parentPofs.end(), Exception,
-                "PhysicalOffsetFrames are not permitted to form loops.");
+            OPENSIM_THROW_IF_FRMOBJ(found != parentPofs.end(),
+                PhysicalOffsetFramesFormLoop, (*found)->getName());
             parentPofs.push_back(parentPof);
             // Given a chain of offsets, the most proximal must have Ground or 
             // Body as its parent. When that happens we can stop.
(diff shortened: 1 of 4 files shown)


OpenSim/Simulation/Model/Model.cpp
```diff
@@ -769,11 +769,31 @@ void Model::extendConnectToModel(Model &model)
...
+    // PhysicalOffsetFrames require that their parent frame (a PhysicalFrame)
+    // be added to the System first. So for each PhysicalOffsetFrame locate its
+    // parent and verify its presence in the _orderedList otherwise add it first.
+    auto poFrames = updComponentList<PhysicalOffsetFrame>();
+    for (auto& pof : poFrames) {
+        // Ground and Body type PhysicalFrames are included in the Multibody graph
+        // PhysicalOffsetFrame can be listed in any order and may be attached
+        // to any other PhysicalOffsetFrame, so we need to find their parent(s)
+        // in the tree and add them first.
+        pof.connect(*this);
```
chrisdembia: This will cause all subcomponents of `pof` to be connected, right? Do you want that, or do you specifically want to connect the `parent` connector?

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -769,11 +769,31 @@ void Model::extendConnectToModel(Model &model)
...
+        const PhysicalOffsetFrame* parentPof =
+            dynamic_cast<const PhysicalOffsetFrame*>(&pof.getParentFrame());
+        std::vector<const PhysicalOffsetFrame*> parentPofs;
```
chrisdembia: This is fine as-is. But: you are using this as a stack; I wonder if there is any benefit to using `std::stack`?

OpenSim/Simulation/Test/testFrames.cpp
```diff
@@ -332,6 +341,45 @@ void testPhysicalOffsetFrameOnBodySerialize()
         "testPhysicalOffsetFrameOnBodySerialize(): incorrect MobilizedBodyIndex");
 }
 
+void testPhysicalOffsetFrameOnPhysicalOffsetFrameOrder()
+{
+    Model pendulum("double_pendulum.osim");
+
+    SimTK::Transform X_RO;
+    X_RO.setP(SimTK::Vec3(0.1, 0.2, 0.3));
+    X_RO.updR().setRotationFromAngleAboutAxis(SimTK::Pi / 4.0, SimTK::ZAxis);
```
aseth1: No significance whatsoever. I *borrrowed* the code block from the `testPhysicalOffsetFrameOnBodySerialize()` case earlier in `testFrames`. The outcome of the transform is not being tested, so it doesn't matter what value it has..

OpenSim/Simulation/Test/testFrames.cpp
```diff
@@ -332,6 +341,45 @@ void testPhysicalOffsetFrameOnBodySerialize()
...
+    // relative location to w.r.t a Body in the Model.
+    PhysicalOffsetFrame* offsetFrameDistal = new PhysicalOffsetFrame();
+    offsetFrameDistal->setName("offsetFrameDistal");
+    offsetFrameDistal->setOffsetTransform(X_RO);
+    pendulum.addComponent(offsetFrameDistal);
+
+    PhysicalOffsetFrame* offsetFrameProximal = new PhysicalOffsetFrame();
+    offsetFrameProximal->setName("offsetFrameProximal");
+    offsetFrameProximal->setOffsetTransform(~X_RO);
+    pendulum.addComponent(offsetFrameProximal);
```
aseth1: I will add it.

aymanhab:
@aseth1 My main issue is the knowledge about the function of various Components/Order that's baked into Model::extendConnectToModel which may make it harder to interpret/extend in the future (e.g. if we introduce a new type of Frame will this code need to change? how? What about new types of Components?) If you believe there's no way around it then we can live with it.

One other high level question before diving into the details. Have you considered using an actual Simbody 'Body' underneath PhysicalOffsetFrame? This will delegate the ordering to one commonplace/ tree-builder on the Simbody side and may make all the issues with SimbodyEngine go away but I'm not sure what's the price/overhead.

aseth1:
>  My main issue is the knowledge about the function of various Components/Order that's baked into Model::extendConnectToModel which may make it harder to interpret/extend in the future (e.g. if we introduce a new type of Frame will this code need to change? how? What about new types of Components?)

It is a major concern for sure and ideally a Model or any top-level component would not have special knowledge and handling of any of its constituents.  However, the main task of an OpenSim Model is to translate between the world of the modeler/user (biomechanist, movement scientist, etc..) to something that can be solved (the MultibodySystem). A MultibodySystem requires a well-defined tree topology whereas a Model (ideally) does not. We have made a lot of progress in that direction and we can handle any order of bodies and joints, for example. Nonetheless, it is the Model's responsibility to construct the appropriate topology for the user's model, so a Model has to know what Components define the topology of a MultibodySystem and these primarily Joints and the rigid bodies they connect. A `PhysicalOffsetFrame` then is an extension of a rigid body and provides convenience for locating attachments and measuring motion in easier to interpret reference frames. As an extension of a rigid body, a `PhysicalOffsetFrame` is part of the tree and has to be treated as such.  

I think we have the rigid bodies (aka `PhysicalFrame`s) covered. They can and should be embedded wherever a rigid connection location is needed. As for new types of Frames, like representing the center of pressure or the center of mass are not part of the Multibody topology and are computed based on the system topology, so these would not require special handling. 

In the case of new Components, new Components should avoid (and perhaps we should prevent it completely) creating MobiliizedBodies themselves. Joints are the OpenSim abstraction for adding dgerees-of-freedom and new Components that need additional dofs and need new attachments should use Joints and PhysicalFrames internally. The exception to this is creating new types of Joint components, in which case deriving from Joint ensures that it is appropriated integrated into the System topology.

> One other high level question before diving into the details. Have you considered using an actual Simbody 'Body' underneath PhysicalOffsetFrame? This will delegate the ordering to one commonplace/ tree-builder on the Simbody side and may make all the issues with SimbodyEngine go away but I'm not sure what's the price/overhead.

It it is a good point and one we discussed a while ago when considering `PhysicalOffsetFrame` as a welded massless Body. The main issues with doing that, as I recall,  were: 1) complete redundancy with `WeldJoint`. 2) `MultibodySystem` size and book-keeping increases with the number of SimTK::Bodies (even though the number of ODEs does not change) whereas adding a `PhysicalOffsetFrame` has no impact on the System computations until it is asked to calculate its transform. In the typical scenario of connecting Bodies with Joints it would introduce Welds for every parent and child frame, where as currently they have no impact since they get baked into the creation of the underlying `MobilizedBody`; 3) asking the `MultibodySystem` for the number of bodies would not even remotely correspond to the number of Bodies in the Model, and could potentially double, triple or even quadruple the number of underlying bodies.

aseth1:
@aymanhab sorry for the lengthy response, but I hope it addressed your concerns.

@chrisdembia I added comments to describe what is being tested and how.

OpenSim/Simulation/Model/Model.cpp
chrisdembia: Okay; if performance becomes an issue, we can revisit later.

aymanhab:
Thanks for the explanation @aseth1  I had one small question above, other than that LGTM.

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -769,11 +769,31 @@ void Model::extendConnectToModel(Model &model)
...
+        while (parentPof) {
+            parentPofs.push_back(parentPof);
+            parentPof =
+                dynamic_cast<const PhysicalOffsetFrame*>(&parentPof->getParentFrame());
+        }
+        while (parentPofs.size()) {
```
chrisdembia: I think you might be referring to the while-loop above, the one that involves the dynamic cast. The one below should certainly terminate, as it pops off each element.

aseth1: The rationale is that moving inward the final PhysicalOffsetFrame has to be connected to a Body or Ground, which would cause the dynamic_cast to return nullptr and terminate the loop. That should work for any well formed tree, but I see now that it is certainly possible to *connect* PhysicalOffsetFrame's such that they form loops and that should be prevented. I am not sure that should happen here, but I agree a check is needed.

aseth1:
@aymanhab thanks for pointing out the potential for loops of `PhysicalOffsetFrames` which is indeed an error case that wasn't being checked. I've added the check and included an Exception when a loop is detected. I also added two more conditions in `testFrames.cpp` to verify that the exception is thrown when expected. 
@chrisdembia feel free to take a look too. Thanks!

aseth1:
@aymanhab @chrisdembia CI finally completed successfully.

aymanhab:
Thanks @aseth1 LGTM. Feel free to merge when tests pass.

aseth1:
:champagne: cheers @chrisdembia + @aymanhab

PR #1321, opened 2016-10-11
Fix memory leak -- ControlSet

klshrinidhi:


chrisdembia:
@klshrinidhi how did you discover the memory leaks? Did you use valgrind or a static analysis tool?

Issue #1312, opened 2016-10-10
Accessibility of table metadata values in Matlab 


jimmyDunne:
I have marker data in a table, and I would like to get the data rate. I can access `.getTableMetaDataKeys` to see what metadata is available. Unfortunately, I can't seem to find a way to access the metadata values (in this case the data rate). `.getTableMetaData` gives an exception and I am unsure on any other methods I could use to get the rate.

klshrinidhi:
This requires wrapping of the classes `OpenSim::ValueArrayDictionary`, `SimTK::AbstractValue`, `SimTK::Value` and may be more.

klshrinidhi:
I will try to add a templated member function to `AbstractDataTable` to retrieve table metadata based on the specified type in the call. I will then try to expose this method to Python/Java. This will avoid the need for wrapping more classes.

PR #1332, opened 2016-10-11
Add ability to access DataTable/TimeSeriesTable metadata from Java/Python

klshrinidhi:
Addressing #1312.
- Added ability to access table meta data from Java/Python/Matlab.
- Added ability to access dependents column metadata from Java/Python/Matlab.
- Added ability to add table meta data from Java/Python/Matlab and later retrieve it.
- Tests demonstrate retrieving following metadata read from C3D file:
  - Force plate calibration matrices.
  - Force plate corners.
  - Force plate origins.
  - Force plate types.


chrisdembia:
The new functionality looks excellent, and the tests look thorough. I look forward to seeing these new functions in MATLAB/Python examples.

I feel strongly that the new functions should be in the C++ code, not only in the SWIG wrapper layer. A large reason for this is to support discovery of these functions by users. I *know* they shouldn't be necessary in C++, but it's a compromise for having a small development team, etc. From what I can tell, there is nothing about the new functions that prohibits putting them directly in the C++ code.

jenhicks:
@chrisdembia You raise a good point that it can be hard for Java and Python users to discover these added functions. But the current implementation does follow how previous templated functions are exposed in wrapping. 

@aymanhab Could you review the PR and provide input?

Are there other (preferably automated) ways that we could document these type of wrapped classes?

chrisdembia:
> But the current implementation does follow how previous templated functions are exposed in wrapping.

As far as I can tell, the other functions that appear in common.i are replacing methods of the same name that exist in C++ and appear in Doxygen (e.g., `setColumnLabels()` and `flatten()` appear in doxygen). So I still think these methods belong in the C++ files.


> Are there other (preferably automated) ways that we could document these type of wrapped classes?

I have tried one automated method for generating python documentation (both within the python shell, and by generating a website with sphinx). But that is a large-scale solution. These methods will still show up via tab-completion.

If the methods are not moved to C++ files, then the C++ files' doxygen documentation could list them, perhaps in a section titled "Accessing the metadata in MATLAB and Python".

klshrinidhi:
I have added comments for Python/Java/Matlab use.

Cancelled CI builds as changes are only to comments.

klshrinidhi:
I have removed the relative paths in the tests as per our discussion.

klshrinidhi:
@jimmyDunne 
The Java tests [here](https://github.com/opensim-org/opensim-core/blob/issue_1312_table_metadata/Bindings/Java/tests/TestDataAdapter.java) show how to retrieve metadata from tables read from C3D files.

aymanhab:
@jimmyDunne Please let me know when you're good to go so we can merge this PR and move on. Thanks @klshrinidhi

jenhicks:
@jimmyDunne Now that the workshop has wrapped up, can you provide an update on your testing of this PR (if testing is still planned). Thanks!

jimmyDunne:
np. Will test as soon as possible.

jimmyDunne:
Tested in Matlab and works great. Am able to get the data rate. 

Thanks, @klshrinidhi!

PR #400, opened 2015-03-16
[DON'T-MERGE] Add stacktrace on windows to debug appveyor crashes.

chrisdembia:
To help address #350, I've made OpenSim print a stack trace upon segfault (and upon some other exit signals) on Windows. I did this kinda messily. It's not intended to be a permanent feature; I would remove it once we find the bug. If you'd prefer me to clean this up before merging, I can do that.

Any suggestions would be welcome too. Where should I put the `signal()` code? It should be called before any code gets executed.

PR #1333, opened 2016-10-12
Fix memory leak -- CMC

klshrinidhi:


OpenSim/Tools/CMC.h
```diff
@@ -113,11 +113,11 @@ OpenSim_DECLARE_CONCRETE_OBJECT(CMC, TrackingController);
...
     bool _checkTargetTime;
     /** Storage object for the position errors. */
-    Storage *_pErrStore;
+    std::shared_ptr<Storage> _pErrStore;
     /** Storage object for the velocity errors. */
-    Storage *_vErrStore;
+    std::shared_ptr<Storage> _vErrStore;
     /** Storage object for the stress term weight. */
-    Storage *_stressTermWeightStore;
+    std::shared_ptr<Storage> _stressTermWeightStore;
```
chrisdembia: What led you to choose shared instead of unique?

klshrinidhi: As per original version, the copy constructor of this class intends to do *shallow copy*. `shared_ptr` suits the need I think.

chrisdembia: Ah okay; looks good.

OpenSim/Tools/CMC_Task.cpp
```diff
@@ -940,9 +940,9 @@ updateFromXMLNode(SimTK::Xml::Element& aNode, int versionNumber)
     //setOn(_on);
     setActive(_active[0],_active[1],_active[2]);
     //setWeight(_w[0],_w[1],_w[2]);
-    setKP(_kp[0],_kp[1],_kp[2]);
-    setKV(_kv[0],_kv[1],_kv[2]);
-    setKA(_ka[0],_ka[1],_ka[2]);
+    //setKP(_kp[0],_kp[1],_kp[2]);
+    //setKV(_kv[0],_kv[1],_kv[2]);
+    //setKA(_ka[0],_ka[1],_ka[2]);
```
chrisdembia: Why are these commented out?

chrisdembia: Ah I see...how come you didn't comment out other ones like `setActive`? I think this is a bug: the arguments should be using the variables like `_propKP`. @aseth1 @aymanhab , do you know what's going on here?

klshrinidhi: I thought I covered all of it. Maybe not. I can comment out the others as well.

chrisdembia: I'm fine with either way. I created a new issue #1335 to track the possibility that those properties are being ignored.

aseth1: Since the only useful part is calling the Super implementation, I think the method could be removed.

klshrinidhi: As per comment in #1335, I removed all code from the function except call to base class.

klshrinidhi: Okay. I have completely removed the function.

aseth1: perfect! Feel free to merge when tests pass.

chrisdembia:
Before we merge this, can one of you address this question?

The function `updateFromXMLNode()` has the side effect of normalizing the direction properties `_r0`, etc. 

```
void CMC_Task::
setDirection_0(const SimTK::Vec3& aR)
{
    _r0 = aR; _r0.normalize();
}
```

Does this mean that `_r0`, etc will no longer be normalized?

chrisdembia:
Another comment: I think this PR might address the spurious testCMC* segfaults on appveyor (see #350). As #400 mentions, the segfault appeared to occur on the call to `setKP()`!

aseth1:
@chrisdembia good catch. It is normalizing and I honestly do not know how this worked since the default values are set to 0, which would cause a divide by zero by default.

aseth1:
@chrisdembia @klshrinidhi the `direction` properties are only used by `CMC_Point` ( a derived `CMC_Task`). None of our user facing documentation that I could find discusses `CMC_Point` and we have no examples that use it. I think we should merge the PR as is and consider removing `CMC_Point` altogether, but I think that is of low priority at this time.

klshrinidhi:
Tests pass. Merging.

Issue #1307, opened 2016-10-07
ComponentPath::getComponentName() crashes when getNumPathLevels() == 0


chrisdembia:
The following code causes a crash:

```cpp
ComponentPath path("");
path.getComponentName();
```

What should the expected behavior be? I can think of two options:

1. Return an empty string; `""`.
2. Throw an exception, saying the ComponentPath is empty.

carmichaelong:
Of the three uses of `getComponentName()` in the Component class, one function checks if there is an empty name (`findComponent()`) and two functions do not (`findAndConnect()` and `traversePathToComponent()`).

I think this currently points at doing option 1, since it seems like Component currently handles the empty name, but I don't know what the complete vision is for the Component class.

aseth1:
Closed by #1336

PR #1336, opened 2016-10-12
fix getComponentName() bug

carmichaelong:
Addresses #1307. When getComponentName() is called on a ComponentPath without any names, it now returns an empty string rather than crashing.


aseth1:
LGTM. CI failures are unrelated. Merging.

PR #1012, opened 2016-06-15
(De)serialize inputs connectee names.

chrisdembia:
This PR allows one to serialize and deserialize input connectee names, as is done for connectors. The major issue I ran into was handling the `isList` member variable of AbstractConnector, which is not restored across serialization. The way I handled this for now was to store the `isList` setting in the `_inputsTable`, which is then passed onto the AbstractInput after deserialization (through the `AbstractConnector::initialize()` method). However, I hope to fix that issue in #1009.

I also did a bunch of miscellaneous cleanup.

This PR builds on top of @klshrinidhi's TableSource PR #994, and should not be merged until after that PR is merged.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/1012)

Issue #1009, opened 2016-06-15
Move AbstractConnector's XML property to the containing component


chrisdembia:
This issue is to record a discussion I had with @aseth1 about a possibly large design change to the way Connectors and Inputs record their connectees. The idea is for the `connectee_name` property to live in the containing component rather than within the Input or Connector. We would not serialize Inputs or Connectors themselves, but they would get an associated property in the containing component. We could name that property something like `input_<name>_connectee_name`. Here's an example of the change in XML:

Before:

``` xml
<MyComponent>
  <connectors>
    <Connector_PhysicalFrame_ name="parent_frame">
      <connectee_name>../ground</connectee_name>
    </Connector_PhysicalFrame_>
  </connectors>
</MyComponent>
```

After:

``` xml
<MyComponent>
  <!-- Single-value Connector to PhysicalFrame. -->
  <connector_parent_frame_connectee_name>../ground</connector_parent_frame_connectee_name>
</MyComponent>
```

@aseth1 also mentioned that we could potentially view this property as the "default" connectees, which could be modified in-memory after loading the model.

Pros:
1. No need to register Connector and Input objects. In fact, they would not need to derive from `Object` anymore, and could potentially be nested in `Component`.
2. No error-prone xml format, where right now users could edit the names of connectors, their types, and could introduce new connectors that don't actually exist in the component.
3. Fixes an issue we have with trying to remember if an Input is a list input or a single-value input.
4. Less verbose way to express connectors in XML (#359).
5. We could automatically generate the comment for the xml property so that it mentions if the connector is a list connector, and the type that it connects to.

Cons:
1. The simplest solution would make the new property (`connector_<name>_connectee_name`) public in the containing class. Right now, the `connectee_name` property is private in `AbstractConnector`. But we still allow people to edit it through accessors/mutators we've created. We must be careful to keep the property in sync with the `_connectees` vector of references.
2. In XML, the inputs and connectors will not be grouped together anymore. Right now, all connectors and inputs show up together at the top of a Component's xml element. With this change, the connectors/inputs would be scattered throughout. This isn't really so bad. Also, we could update the serialization code to serialize connectors before all other properties.
3. What if we learn, down the road, that we actually want more properties associated with a connector other than just its `connectee_name`? Say connectors eventually get more features (e.g., enabled/disabled). Then we would need to introduce even more properties into the containing component, and this could get messy.

tkuchida:
[...]

Issue #825, opened 2016-03-11
Connectee name property for inputs


chrisdembia:
Right now, Connectors automatically perform connection during `initSystem()` because the names of the connectees are stored in an XML property. The same should be done for inputs.

@aseth1 and I also talked about using two different verbs for `Connector::connect()` (like `set_connectee()`) and for `Component::connect()`. We discussed that the `connectee_name` property could be more completely hidden from the API.

aseth1:
The API would be a lot cleaner and consistent if names (strings) were not used to make connections, but used instances to connect directly. The exception would be from deserialization which would then use the (path) name to locate the dependency and satisfy the connectee.

That requires that every Component that is constructed via the API has a valid name (so it can be serialized). Once there are instances of components around those instances are used to make connections for connectors and inputs.

The main barrier to achieving this are:
 1. that the vast majority of Components have properties with the dependency names exposed (a subset converted to Connector->connectee_name)

 2. the API is littered with dependency names in constructors and public accessors.
  
 3. the multitude of examples and test cases that explicitly use names to identify their dependencies

It would take a big and concerted effort to make components satisfy their connections directly and use (path) names strictly for serialization, but it would make the API and its examples less confusing. The current API is trying to do both (use instances and/or names) but for many components it seems to work one way or another depending on what was necessary to conform to test cases and examples. It is probably about time that the test cases and examples were updated to reflect the API.

We should discuss doing this after the hackathon.

Issue #359, opened 2015-02-21
Serialization of Connections is verbose


aymanhab:
Connections can be serialized as glorified Properties from the XML side with exactly three attributes: type, name and connected_to_name, instead of the current xml representation:

```
 <connectors>
    <Connector_PhysicalFrame_ name="frame">
        <!--Name of the component this Connector should be connected to.-->
        <connected_to_name>ground</connected_to_name>
    </Connector_PhysicalFrame_>
</connectors>
```

An example simpler syntax,

```
<Connector type="PhysicalFrame" name = "frame" connected_to="ground" />
```

would do the job, and avoids making up syntax to serialize templates.

It'd be good to investigate this before we propagate connections all over our xml files.

chrisdembia:
@aymanhab, `<` and `>` seem to have some special meaning in Markdown. I edited your post so that we can now see your code samples (put code samples between pairs of 3 backticks: 

![image](https://cloud.githubusercontent.com/assets/846001/6312766/e0b5175a-b93a-11e4-92e0-e0823b00f15c.png)

chrisdembia:
+1

I agree that the current syntax is very verbose, and I really like the proposed syntax.    Is it possible? What about a type with multiple template arguments?

sherm1:
+1
Nice syntax!

aymanhab:
Thanks @chrisdembia and @sherm1 for the feedback. I think it's possible, I just wanted to start the discussion to see how others feel about it, then we can discuss the pros, cons and implementation.

aseth1:
I also agree the brevity of the proposed syntax is nice. @aymanhab How easy is it for you to make the change? Would you the have a Connector factory to create the concrete type?

If we didn't require people to build models via XML editing, would the verbosity even matter?

chrisdembia:
> If we didn't require people to build models via XML editing, would the verbosity even matter?

Not as much but yes, to view their model. Also, I think if we support a plain-text serialization format, we should support it well. I actually wasn't thinking of the verbosity in terms of having to type up all of that when creating XML; I just thought it was too much to look at when reading XML. I like the idea that one line of XML contains roughly one unit of useful information.

aymanhab:
@aseth1 I think the implementation "should" be straightforward (basically after deserialization, pass XML into a Factory that creates the concrete class) but I may need some template gymnastics help from the template-jedi-master @sherm1), do you think that's feasible, @sherm1 ?

The issues that triggered this are:
[...]

PR #1340, opened 2016-10-13
Move connectee_name property out of AbstractConnector; Serialize Input connectee_name

chrisdembia:
### Background

This PR moves the `connectee_name` property from `AbstractConnector` to `Component`, and also now causes Inputs' connectee_name to be serialized and deserialized.

The work in this PR originates from PR #1012, whose aim was to serialize Inputs' connectee_name in the same manner as Connectors' connectee_name. In doing so, however, I ran into an issue with that serialization scheme. As a result, this PR changes the way in which the connectee_name is serialized, and also provides some additional benefits (see #359, #825, #1009).
### Change in XML connectee name syntax

Here's the previous syntax:

``` xml
<Geometry>
   <connectors>
      <Connector_PhysicalFrame_ name="frame">
          <!--Name of the component this Connector should be connected to.-->
          <connectee_name>ground</connectee_name>
      </Connector_PhysicalFrame_>
  </connectors>
</Geometry>
```

Here's the new syntax:

``` xml
<MyComponent>
  <!--Path to a Component to satisfy the Connector 'childFoo' of type Foo (description: The frame to which this geometry is attached. Note, being connected to a Frame means its transform is used to position this Geometry.).--> 
  <connector_frame_connectee_name>ground</connector_frame_connecteee_name>
</MyComponent>
```
#### Advantages in the new syntax and implementation:
1. Shorter (fewer lines); especially if the comment is removed.
2. The user input is limited to exactly what must be provided (user cannot put in the wrong type (`PhysicalFrame`), or wrong connector name (`frame`)).
3. The XML comment is specific to the connector (rather than generic), and includes the comment provided in `OpenSim_DECLARE_CONNECTOR`.
4. Removes the need tor register (`Object::registerType`) a Connector/Input for each new template argument.
5. The user could previously reorder the Connectors within the `connectors` property list, or add non-existant connectors (e.g., could add a `Connector_Foo_` and it would get deserialized, even though Foo shouldn't have such a connector).
6. I think that users _should_ be able to directly edit XML files (though I know others disagree), and this new syntax is hopefully easier to view/edit.
#### Disadvantages of the new syntax and implementation:
1. The new implementation relies on more macros, which are confusing to users (plugin developers).
2. Slight abuse of the property system (introducing two new macros to generate properties that are specific to connectee names.
3. In XML, the inputs and connectors are no longer grouped together anymore. Previously, all connectors and inputs showed up together at the top of a Component's xml element. Also, we could update the serialization code to serialize connectors before all other properties.
4. If we learn, down the road, that we actually want more properties associated with a connector other than just its connectee_name, then we could end up in a somewhat messy situation. But this is a hypothetical concern.
### What does doxygen look like?

Here's the top of the doxygen page for Geometry, which has a list input and a connector:

image

The `input_transform_connectee_name` property does not show up, as expected. However, the detailed description of the `transform` Input describes the `<input_transform_connectee_name>` XML element:
image

Here's an example detailed description for a connector:
image
### Summary of changes to the code

Note: Unfortunately, I was not very good about staying focused to the task at hand, and this PR includes a bunch of other smaller changes.
1. Wrote new property-generating macros, `OpenSim_DECLARE_PROPERTY_CONNECTEE_NAME` and `OpenSim_DECLARE_LIST_PROPERTY_CONNECTEE_NAME`. These generate `std::string` properties, but do _not_ generate a bunch of extra functions like `get_prop_name()`. As a result, the only way to edit the connectee name is through the Input. The C++ interface has not really changed (caveat: advanced users can still edit _any_ property via the property table).
   1. This property is constructed using an in-class member initializer, and thus it is not required to call `constructProperty_prop_name()` in the constructor.
2. Gave `AbstractConnector` access to the new property generated in the macro.
3. `Component::getInput()` and `Component::getConnector()` now have a "side effect" of setting the "owner" on the input/connector before returning it. This is to deal with the fact that the input's/connector's "owner" pointer is lost after copying the component. This is less than idea but is way better than writing a custom copy constructor for Component.
4. Wrote XML conversion code by implementing `updateFromXMLNode()` for `Component`, to update previous XML files to the new format. I bumped up the `XMLDocument::LatestVersion` to 30508.
5. Created more documentation within the `CONNECTOR` and `INPUT` macros.
6. Wrote tests to make sure that serializing -> deserializing input connectee names works.
7. Improved testComponents so that it now wires up inputs.

Unrelated changes:
1. Rolled back the model file `Applications/Analyze/test/subject01_adjusted_spring.osim` from version 30503 to 30000. Here's part of the commit message for that commit:
   
   > This commit reverts the model used in the testStaticOptimization passive force
   > test to version 30000. Previously, the model was at version 30503, but we
   > changed the Station connector's name from reference_frame to parent_frame after
   > 30503 was first created.
2. Made constructors for Connectors/Inputs protected, so that only Components will construct them.
3. Rearranged the implementation of `Input::connect` methods. This causes a really messy diff, but there are no significant changes here. I mostly just reorganized the code so I could reduce indentation.
### Looking forward

I am interested in further improving the XML syntax, as described in #1009 (thanks @tkuchida). I think we _could_ achieve the following:

``` xml
<MyComponent>
  <input_desired_angle_connectee_name>
    <InputConnecteeName path="../marker_data" output="column" alias="right_ankle"/>
  </input_desired_angle_connectee_name>
</MyComponent>
```

I don't want to tackle this here though.


chrisdembia:
I have finished updating the PR description, and this is now ready for review. 

(Note to @tkuchida: this PR likely conflicts with your #1316. I want your PR to be merged first, so that I can be the one to fix the conflicts)

chrisdembia:
You can view the doxygen for this PR on the myosin website: http://myosin.sourceforge.net/1340/classOpenSim_1_1Geometry.html

OpenSim/Common/ComponentConnector.h
```diff
@@ -200,45 +175,110 @@ class OSIMCOMMON_API AbstractConnector : public Object {
...
+    // constructor for Component.
+    void setOwner(Component& o) { _owner.reset(&o); }
+    /** This will be false immediately after copy construction or assignment.*/
+    bool hasOwner() const { return !_owner.empty(); }
+    
+    /** Check if entries of the connectee_name property contain spaces; if so,
+    print out a warning. */
+    void finalizeFromProperties() {
+        // TODO This check may go elsewhere once the connectee_name
+        // property is a ComponentPath (or a ChannelPath?).
```
chrisdembia: Fortunately, `:`, `(`, and `)` are not forbidden characters in `ComponentPath`, so yeah I can try that.

OpenSim/Common/ComponentConnector.h
```diff
@@ -200,45 +175,110 @@ class OSIMCOMMON_API AbstractConnector : public Object {
...
+    a writeable reference. */
+    // We could avoid the need for this function by writing a custom copy
+    // constructor for Component.
+    void setOwner(Component& o) { _owner.reset(&o); }
+    /** This will be false immediately after copy construction or assignment.*/
+    bool hasOwner() const { return !_owner.empty(); }
+    
+    /** Check if entries of the connectee_name property contain spaces; if so,
+    print out a warning. */
+    void finalizeFromProperties() {
```
chrisdembia: How about a more specific name like `checkConnecteeNameProperty()`? Keep in mind that this function should disappear if we make the property type into InputConnecteePath or ChannelPath instead of std::string.

OpenSim/Common/Component.h
```diff
@@ -2180,18 +2184,21 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
     template <typename T>
-    int constructConnector(const std::string& name, bool isList = false) {
-        int ix = updProperty_connectors().adoptAndAppendValue(
-            new Connector<T>(name, SimTK::Stage::Topology, *this));
-        // Add pointer to connectorsTable so we can access connectors easily by
-        // name.
-        _connectorsTable[name] = ix;
-        return ix;
+    bool constructConnector(const std::string& name,
+                            const PropertyIndex& connecteeNameIndex) {
```
chrisdembia: I started to attempt this and I think it will work! Good idea. The biggest question is whether doing so would preclude creating separate input/connector macros in the case that the developer wants a custom copy constructor. In such a case, the developer will need to call macro-generated methods like `constructInput_input_name()` in their constructors. I think it should work fine, but I just haven't tried it yet.

OpenSim/Common/Component.h
```diff
@@ -813,7 +811,19 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
+            // copying the component;
+            // e.g., myComponent.clone().getConnector("a").getConnecteeName().
+            // Since we use the default copy constructor for Component,
+            // the copied AbstractConnector cannot know its new owner
+            // immediately after copying.
+            if (!it->second->hasOwner()) {
+                // The `this` pointer must be non-const because the Connector
+                // will want to be able to modify the connectee_name property.
+                const_cast<AbstractConnector*>(it->second.get())->setOwner(
+                        const_cast<Self&>(*this));
```
chrisdembia: I agree, and that was my first plan. But then I ran into the nice code from the README:

```cpp
Ellipsoid bodyGeometry(0.1, 0.5, 0.1);
...
humerusCenter->attachGeometry(bodyGeometry.clone()); <--- no chance to call finalizeFromProperties()
```

This causes the following to be invoked:

```cpp
bodyGeometry.clone()->updConnector<Frame>("frame").setConnecteeName(...);
```

But since we're working with a copy of the original `bodyGeometry`, calling `setConnecteeName()` doesn't work as the copy doesn't know about its parent yet.


These are the options I've thought of so far:
1. Prohibit the concise syntax above and require calling `finalizeFromProperties()`.
2. Cause the same behavior for subcomponents; e.g., have `getComponent()`, `getComponentList()`, etc invoke `setParent()` before returning the subcomponent.
3. When a user calls `Geometry::setFrame()`, or more generally sets up connections via the API, do not store the connectee name at that time; hold onto the given reference. Only fill in the connectee_name property when serializing, etc. This change would have other benefits too (avoid issues with connecting to an object that isn't part of the model yet). However, this is also out of scope of this PR.

Another possibility to

OpenSim/Common/Component.h
```diff
@@ -706,34 +714,24 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
+     *     const AbstractConnector& conn = getConnector(name);
+     * }
+     * @endcode */
     std::vector<std::string> getConnectorNames() {
-        std::vector<std::string> names(getNumConnectors());
-        for (int i = 0; i < getNumConnectors(); ++i) {
-            names[i] = get_connectors(i).getName(); 
+        std::vector<std::string> names;
+        for (const auto& it : _connectorsTable) {
+            names.push_back(it.first);
```
chrisdembia: > now it is the order maintained by the map

Correct.

> so this might be OK.

I haven't thought of a problem with this yet. But if there is, we can change the data structure from a map to a vector.

OpenSim/Common/Component.cpp
```diff
@@ -1035,6 +1033,25 @@ void Component::setNextSubcomponentInSystem(const Component& sub) const
...
+        if (versionNumber < 30508) {
+            // Here's an example of the change this function might make:
+            // Previous: <connectors>
+            //               <Connector_PhysicalFrame_ name="parent">
+            //                   <connectee_name>...</connectee_name>
+            //               </Connector_PhysicalFrame_>
+            //           </connectors>
+            // New:      <connector_parent_connectee_name>...
+            //               </connector_parent_connectee_name>
+            //
```
chrisdembia: > which isn't totally clear about what is auto generated and what is user defined.

The comment is completely auto-generated. When the user gets a default XML representation for a component from the GUI's XML Browser (or from `opensim-cmd print-xml`), they will see:

```xml
<MyComponent>
  <!--Path to a Component to satisfy the Connector 'childFoo' of type Foo (description: The frame to which this geometry is attached. Note, being connected to a Frame means its transform is used to position this Geometry.).--> 
  <connector_frame_connectee_name></connector_frame_connecteee_name>
</MyComponent>
```

So the only thing that is user-defined is the value of the `<connector_frame_connectee_name>` element.


> but if the comments disappear (e.g. MATLAB XML editing) then determining the type that satisfies "parent" can be challenging.

It may be possible to make the field name contain the type: `<connector_PhysicalFrame_frame_connectee_name>`. I'm not sure if this is possible to implement, and might look bad for templated types (e.g., `<connector_TableSource_double__frame_connectee_name>`.

While I understand the sentiment of more directly putting the type name in the XML tags, I'm not sure it's the right thing to do. The type of the Connector is indeed documentation, not user-editable. So I think it is a benefit that the type name does not appear "user-editable" in the XML file. For example, users might mis-specify a Connector to Body with the erroneous tag `<Connector_Ground_ name="frame">`. 

I think it is reasonable to expect that the comments are necessary for users to know how to fill in the XML fields. I think we already hold this expectation for our XML files. For most existing properties, I would not feel comfortable editing them without seeing the XML comment (i.e., Body's `<inertia>`, Muscle's `<max_contraction_velocity>`). Also, the workflows that involve manually editing XML files should hopefully be the same workflows that preserve XML comments (namely, I'm thinking of using Models/setup files in the distribution's Models folder, and the GUI's XML Browser). Surely, if one edits an XML file in MATLAB using the OpenSim API and calls `Object::print()`, they will still see XML comments. And we want to discourage using XML libraries directly to edit the files.

We can also ensure that the error message obtained when specifying a path to an object of the wrong type is very clear. For example, "The component at path '/foo/bar' is of type Ground, but must be of type Body." (I didn't check the code yet to see if such an error message is already in place).

chrisdembia: > Previously, the Component had a property list of Connectors, which kept Connectors grouped.

We may be able to edit the serialization code to first serialize properties whose names contain `connectee_name`. What do you think?

I agree that this is definitely a downside, but if we must live with it: I feel that the proposed design has fewer downsides than the alternatives.

chrisdembia:
> Might be easier to discuss in person. 

I have responded to your comments. I do still want to chat in person, but I decided to also type out my responses for record-keeping.

OpenSim/Common/Component.h
aseth1: I see now, thanks. My preference then is 2. if it works consistently. I also think 3. would be great and addresses the ability to rename components. If the reference is the connection then we could extract the name for the purpose of serialization. 

For now let's leave your solution for Connectors/Inputs. After the PR is merged lets try 2. to establish consistency with components. Let's put 3. on the back-burner, for now.

OpenSim/Common/Component.cpp
aseth1: >We can also ensure that the error message obtained when specifying a path to an object of the wrong type is very clear. For example, "The component at path '/foo/bar' is of type Ground, but must be of type Body." (I didn't check the code yet to see if such an error message is already in place).

I think that would be very helpful. One can turn to the *XMLBrowser* in the GUI or the *Doxygen* to learn of the type in the few instances that the comments are wiped out.  So, I am OK with this change.

aseth1: I have not thought it through completely, but what if Component maintained list properties for `Connector` and `Input` `connectee_name`s?

OpenSim/Common/Component.h
aseth1: Great! I think calling macro generated methods in custom copy would be fine and we already have a precedent for that with `copyProperty_property_name` so I think it would at least be consistent.

OpenSim/Common/ComponentConnector.h
aseth1: I like the more specific name. Can you also place a TODO to remove the method once we move to storing Paths instead of strings?

aseth1:
It would also be good to get feedback from the group (@aymanhab @jimmyDunne  @jenhicks @tkuchida @carmichaelong) on the:

1. XML format of `Connector` (which will be renamed Socket in a future PR according to #1219) and `Input`
2. different order of `connectee_name` properties from their uniform listing at the top of the component in XML

before making any further changes.

chrisdembia:
> It would also be good to get feedback from the group

Good point. We could discuss in the dev meeting this Thursday?

chrisdembia:
Next steps:

- Move property construction out of the macro and into the `constructConnector()`, etc methods.
- Discuss with team (dev meeting?) about the ordering of properties in XML (should inputs be grouped together?).
- Improve error/documentation about connectee_names to quantities of the incorrect type.

chrisdembia:
Summary from dev meeting: Try to achieve the following format:

```xml
<MyComponent>
  <connectors>
  <!--Path to a Component to satisfy the Connector 'childFoo' of type Foo (description: The frame to which this geometry is attached. Note, being connected to a Frame means its transform is used to position this Geometry.).--> 
  <connector_frame_connectee_name>ground</connector_frame_connecteee_name>
  </connectors>
</MyComponent>
```

carmichaelong:
I think a further point is that if there is now a `<connectors>` section, then maybe we can further shorten the most inner tag without ambiguity? For instance, it could now be `<frame_connectee_name>` or maybe something shorter (but only if it's clear).

chrisdembia:
Ah yes good point. I remember that but I forgot to put that in the above comment. Thanks for the reminder.

OpenSim/Common/Component.h
chrisdembia: > After the PR is merged lets try 2. to establish consistency with components. 

I've opened #1393 to track this.

OpenSim/Common/Component.cpp
chrisdembia: I have checked the error messages to display the required type in case that the connectee cannot be found.

OpenSim/Common/ComponentConnector.h
chrisdembia: Done.

chrisdembia:
@aseth1 this is ready for review again. I have:

1. put the construction of the connectee_name property within the `constructConnector()`, etc. methods,

2. renamed `AbstractConnector::finalizeFromProperties()` to `AbstractConnector::checkConnecteeNameProperty()`, and

3. added tests to make sure an exception is thrown for type mismatches (something exists at the connectee_name path, but it's not of the correct type).

aseth1:
Would also be great to get feedback on impact of serialization and connection management via the API (@carmichaelong ?) and also from the GUI perspective (@aymanhab) .

OpenSim/Common/Component.h
aymanhab: Thanks for clarifying the order issue here @chrisdembia , you can ignore my earlier comment except for uniqueness guarantee.

OpenSim/Common/Component.h
aymanhab: Should probably add this to #1393 but having  a side-effect inside a getter method (that users would assume is const/safe) is less desirable, so I'd rather see option 3 but that's futuristic for now.

OpenSim/Common/ComponentConnector.h
```diff
@@ -295,38 +333,38 @@ class Connector : public AbstractConnector {
...
+        }
+        
+        connectee = *objT;
+
+        std::string objPathName = objT->getAbsolutePathName();
+        std::string ownerPathName = getOwner().getAbsolutePathName();
+
+        // check if the absolute pathname is just /name
+        if (objPathName.compare("/" + objT->getName()) == 0) { //exact match
+            // in which case we likely are connecting to an orphan
```
aymanhab: I thought we got rid of these explicit "/" and delegated this to PathName, is that possible here? Same with handling of "(",":" etc.

OpenSim/Common/Object.h
```diff
@@ -69,6 +69,10 @@ template class OSIMCOMMON_API OpenSim::ArrayPtrs<OpenSim::Object>;
     #define SWIG_DECLARE_EXCEPTION
 #endif
 
+// Forward-declare SimTK types.
+namespace SimTK {
+    typedef Vec<2, Vec3> SpatialVec;
+}
```
aymanhab: Why were these added here?

aymanhab:
Finished with review. Minor suggestions/clarifications indicated above. An example that shows a minimal model with input serialization would be helpful. Thanks @chrisdembia Nicely done.

OpenSim/Common/Component.h
chrisdembia: @aymanhab I have added your comment to #1393. For this PR, would it be sufficient to add a note to the function doxygen block saying that there is such a side effect?

Bindings/Java/Matlab/tests/testConnectorsInputsOutputs.m
```diff
@@ -43,7 +43,7 @@
 end
 
 % Access (and iterate through) a component's connectees as objects.
-expectedNames = {'ground', 'block'};
+expectedNames = {'block', 'ground'};
```
chrisdembia: > If the latter are we guaranteed uniqueness?

Ah...there is an issue here. I will add a check to see if a connector/input with the same name already exists, and throw an exception if so.  Related: #859.

OpenSim/Common/ComponentConnector.h
chrisdembia: I have removed the explicit "/". @aseth1 it might be good for you to check this change, as I think you wrote this code originally.

OpenSim/Common/Object.h
chrisdembia: Done.

Bindings/Java/Matlab/tests/testConnectorsInputsOutputs.m
chrisdembia: I have added thorough checks for uniqueness, and corresponding tests.

chrisdembia:
@aymanhab I have addressed your comments.

Bindings/Python/examples/wiring_inputs_and_outputs_with_TableReporter.py
```diff
@@ -0,0 +1,99 @@
...
+reporter = osim.TableReporterVec3.safeDownCast(model.getComponent('reporter'))
+# This returns the TimeSeriesTableVec3 that holds the history of positions.
+table = reporter.getTable()
+# Create a FileAdapter, which handles writing to (and reading from) .sto files.
+sto = osim.STOFileAdapterVec3()
+sto.write(table, 'wiring_inputs_and_outputs_with_TableReporter.sto')
+# You can open the .sto file in a text editor and see that both outputs
+# (position of body's origin, and position of system mass center) are the same.
+
+
```
aymanhab: Very nice test @chrisdembia I would have created a new variable roundtrip_model to deserialize just for clarity. and also query the reporter upon deserialization that it has the right connections/inputs.  You went a step further to run a simulation which is great but is beyond the task at hand (of serializing/deserializing inputs. We can't verify correctness unless we run and check output file which I'm sure you did but may be a  bit harder to maintain. Thoughts?

aymanhab:
Test failure appears unrelated. Finished with re-review, looking very good. I'm ok with merging as is or after addressing my last comment on the test, unless @aseth1 had other reservations.

Bindings/Python/examples/wiring_inputs_and_outputs_with_TableReporter.py
chrisdembia: Good points. I will rename the variable for clarity. I prefer `deserialized_model`. What do you think?

This file is an example, and so I don't really want it to look like a test (that is why I run a simulation). I had thought you were looking for a user-facing illustration of what this feature looks like. This new feature is thoroughly tested in C++ (testComponentInterface). If you'd like, I can also add a python (or Java) test.

aymanhab: I like deserialized_model better as well. Go for it. 
As to what to do after deserialization I agree not to test/assert but you may want to modify something upon deserialization  that a user would possibly do ( and to show users how to recover access to the input). A Java example (for Matlab users) would be perfect, but if your environment isn't set up we can do later when other related open issues are resolved. Thanks.

chrisdembia: Sounds good. Should I create a Java test or a MATLAB example? Or, a MATLAB test?

aymanhab:
I d say matlab example

chrisdembia:
@aymanhab I created a MATLAB example and improved the python example. I wanted to show editing of the alias in the examples, but ran into a bug wherein `setAlias()` does not affect the TableReporter's column labels. I have opened #1420 to track this bug. I do not want to tackle it here, because it will be easier to tackle in a follow-up PR that I have already started working on.

@aseth1 please review when you get the chance.

OpenSim/Common/ComponentConnector.h
```diff
@@ -344,36 +381,99 @@ class Connector : public AbstractConnector {
...
+            ../foo/angle
+        </input_desired_angle_connectee_name>
+        ...
+    </MyComponent>
+@endcode
+You use this field to specify the outputs/channels that should be connected to
+this input (that is, the connectees). The syntax for the connectee name
+property is as follows:
+@code
+<path/to/component/><output_name>[:<channel_name>][(<alias>)]
```
aseth1: We discussed briefly,  should we separate the path from output name? As it stands we have to test what comes after the last '/' to see if it is an Output and not a Component.  That is to say paths always identify components so we have: `<path/to/component>|<output_name>[:<channel_name>][(<alias>)]`. 
We could leave for a followup PR unless you think it is easier to do now.

OpenSim/Common/ComponentConnector.h
```diff
@@ -344,36 +381,99 @@ class Connector : public AbstractConnector {
...
+    /** For use in python/java/MATLAB bindings. */
+    // This method exists for consistency with Object's safeDownCast.
+    static Connector<T>* safeDownCast(AbstractConnector* base) {
+        return dynamic_cast<Connector<T>*>(base);
+    }
+    
+protected:
+    /** Create a Connector that can only connect to Object of type T with 
+    specified name and stage at which it should be connected. Only Component
+    should ever construct this class.
```
aseth1: should --> can?
As defined only a derived `Connector` or `Component` *can* invoke this constructor.

OpenSim/Common/XMLDocument.cpp
```diff
@@ -427,6 +428,39 @@ void XMLDocument::addConnector(SimTK::Xml::Element& element,
     //connectors_node->writeToString(debug);
 }
 
+void XMLDocument::updateConnectors30508(SimTK::Xml::Element& componentElt)
+{
+    using ElementItr = SimTK::Xml::element_iterator;
+    
+    std::string typeName = "PhysicalFrame";
```
aseth1: `typeName` is unused.

OpenSim/Common/XMLDocument.cpp
```diff
@@ -427,6 +428,39 @@ void XMLDocument::addConnector(SimTK::Xml::Element& element,
...
+        // Create new element for this connector's connectee name.
+        SimTK::Xml::Element newConnecteeNameElt(
+                "connector_" + connectorName + "_connectee_name");
+        newConnecteeNameElt.setValue(connecteeName);
+        componentElt.insertNodeAfter(connectors_node, newConnecteeNameElt);
+    }
+    
+    // No longer want the old syntax for connectors.
+    componentElt.eraseNode(connectors_node);
+}
```
aseth1: LGTM :+1:

OpenSim/Tests/Components/testComponents.cpp
```diff
@@ -212,29 +227,66 @@ void testComponent(const Component& instanceToTest)
...
+        Component& next = *itc;
         //Now keep checking the subcomponents
-        sub = const_cast<Component *>(&next);
-        it++;
+        sub = &next;
+        itc++;
     }
-
+    
+    // Now make sure Inputs are satisfied.
```
aseth1: Great addition @chrisdembia!!

Applications/Analyze/test/subject01_adjusted_spring.osim
```diff
@@ -1,1611 +1,1444 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<OpenSimDocument Version="30503">
+<OpenSimDocument Version="30000">
```
chrisdembia: I wish that were the case. Here is part of the commit message for the commit in which I reverted the model:

> This commit reverts the model used in the testStaticOptimization passive force
test to version 30000. Previously, the model was at version 30503, but we
changed the Station connector's name from reference_frame to parent_frame after
30503 was first created.

This model file uses the outdated `reference_frame` as the name for Station's Connector. Unfortunately, there is no updateFromXML code that converts `reference_frame` to `parent_frame`. So it is not good to have this intermediate model version in the repo (we should also separately implement `Station::updateFromXMLNode()`).

chrisdembia: I can alternatively rename `reference_frame` to `parent_frame` everywhere within the model file.

chrisdembia:
> I am glad to see the checks for name uniqueness, but I am still unclear as to why the order changed. Is it because child_frame connectee name appears before the parent_frame's in the property table (ordered alphabetically?) whereas the property list of connectors maintained order of elements as they were added?

This is because connectors are stored in an `std::map` (alphabetically). We discussed this during your initial review.

OpenSim/Common/ComponentConnector.h
chrisdembia: I started looking at implementing this, and I have decided that this change will fit much better in a subsequent PR.

Applications/Analyze/test/subject01_adjusted_spring.osim
aseth1: Thanks for explanation. No need to change.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
```diff
@@ -127,6 +127,34 @@ do { \
...
+    } \
+    catch (...) { \
+        throw OpenSim::Exception("TESTING: Expected exception " \
+            #EXPECTED_EXCEPTION " but caught different exception."); \
+    } \
+    if (!caughtExpectedException) { \
+        throw OpenSim::Exception("TESTING: Expected exception " \
+            #EXPECTED_EXCEPTION " but no exception thrown."); \
+    } \
+} while(false)
```
aseth1: Makes sense. Thanks!

OpenSim/Common/ComponentConnector.h
aseth1: No problem. Thanks for looking into it.

Bindings/Java/Matlab/tests/testConnectorsInputsOutputs.m
```diff
@@ -105,15 +105,15 @@
...
 names = rep.getInputNames();
-expectedAnnos = {'value', 'target', 'c1', 'second_col'};
+expectedAliases = {'', 'target', '', 'second_col'};
 for i = 0:(names.size() - 1)
     % Actually, there is only one Input, named 'inputs'.
     % We connected it to 4 channels.
     numConnectees = rep.getInput(names.get(i)).getNumConnectees();
     assert(numConnectees == 4);
     for j = 0:(numConnectees - 1)
-        assert(strcmp(rep.getInput(names.get(i)).getAnnotation(j), ...
```
chrisdembia: Updated to test label.

OpenSim/Common/ComponentConnector.h
chrisdembia: Done.

OpenSim/Common/ComponentConnector.h
chrisdembia: So, I actually tried again and went through with the change. The component path and the output name are now separated by `|`. I decided it would be good to have this change in this PR, to avoid introducing an excess XML format update.

OpenSim/Common/XMLDocument.cpp
chrisdembia: Removed.

chrisdembia:
@aseth1 I have addressed your comments. Waiting for the tests to pass...

OpenSim/Common/ComponentConnector.h
aseth1: Looks great to me.

aseth1:
Verified that CI failure is due to `testScale` that was just fixed. I will merging now.

chrisdembia:
Thank you.

PR #1341, opened 2016-10-13
Future marker previewer

aseth1:
This PR adds a simple previewer of markers using Components and the Visualizer through the 4.0 API.
It highlights some of the new capabilities in 4.0 from new data Components like `TableSource`, Inputs and Outputs, and Component-based visualization (opposed to being drawn by the Model).  It may seed a more detailed example of be the basis for an experimental marker viewer type component in OpenSim. I would also like to see us replicate this example via scripting where it could be immediately useful to users.


aymanhab:
@aseth1 Is this functional in some form, or is it completely hypothetical/futuristic? If the latter do you have specific missing functionality? Thanks.

aseth1:
It totally works! Try it out if you get a chance.

chrisdembia:
This is for the sandbox. Merging.

chrisdembia:
@aseth1 I'll let you delete the branch if you want. I don't want to delete it myself because I think you might still be using it.

PR #1343, opened 2016-10-13
[doxygen] Add note to AssemblySolver::setAccuracy()

tkuchida:
Accuracy is propagated to underlying `SimTK::Assembler` when calling `assemble()` but not `track()`.


tkuchida:
Re-implemented by @aseth1 in #1371.

Issue #1338, opened 2016-10-12
MarkerData::readStoFile() writes two lines to file header every time the file is read


tkuchida:
[`MarkerData::readStoFile()`](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/MarkerData.cpp#L481) calls [`addNumRowsNumColumns()`](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Auxiliary/auxiliaryTestFunctions.h#L229), which writes
```
nRows=[number of rows]
nColumns=[number of columns]
```
to the header every time a .sto file is read. `addNumRowsNumColumns()` should write these header rows only if they don't already exist.

klshrinidhi:
I think I added this to make sure the Tools are able to read STO file written by STOFileAdapter. I will look into this and get back.

tkuchida:
The issue appears to start at [MarkerData.cpp#L486](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/MarkerData.cpp#L486). As we discussed, the `revertToVersionNumber1()` method can return a bool indicating whether the version number was decremented; `addNumRowsNumColumns()` should be called only if `true`:
```cpp
if (revertToVersionNumber1(aFileName, tmpFileName))
    addNumRowsNumColumns(tmpFileName, aFileName);
```

tkuchida:
Fixed by #1349.

PR #1349, opened 2016-10-17
Fix bug in MarkerData::readSTOFile()

klshrinidhi:
Addressing #1338.

### DIFF
--- a/OpenSim/Common/MarkerData.cpp
+++ b/OpenSim/Common/MarkerData.cpp
@@ -486,8 +486,9 @@ void MarkerData::readStoFile(const string& aFileName)
     // If the file was written by STOFileAdapter, make the file readable by
     // Storage. Calls below have no effect otherwise.
     std::string tmpFileName{"tmp.sto"};
-    revertToVersionNumber1(aFileName, tmpFileName);
-    addNumRowsNumColumns(tmpFileName, aFileName);
+    bool versionChanged{revertToVersionNumber1(aFileName, tmpFileName)};
+    if(versionChanged)
+        addNumRowsNumColumns(tmpFileName, aFileName);
     std::remove(tmpFileName.c_str());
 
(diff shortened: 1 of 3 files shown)


jenhicks:
@aymanhab or @aseth1 This needs one more reviewer.

OpenSim/Common/MarkerData.cpp
```diff
@@ -486,8 +486,9 @@ void MarkerData::readStoFile(const string& aFileName)
     // If the file was written by STOFileAdapter, make the file readable by
     // Storage. Calls below have no effect otherwise.
     std::string tmpFileName{"tmp.sto"};
-    revertToVersionNumber1(aFileName, tmpFileName);
-    addNumRowsNumColumns(tmpFileName, aFileName);
+    bool versionChanged{revertToVersionNumber1(aFileName, tmpFileName)};
+    if(versionChanged)
+        addNumRowsNumColumns(tmpFileName, aFileName);
```
aseth1: If the requirement to `addNumRowsColumns` is a result of reverting the version number (which it seem it is) then why doesn't `revertToVersionNumber1` simply `addNumRowsColumns` itself when it flags `changedVersion = true` above? Then we would not have to expose `addNumRowsNumColumns`.

klshrinidhi: The function `revertToVersionNumber1()` was written earlier to get some of the tests working, which use STO files that don't need adding number of rows and columns to header. Tests currently using this function -- `testControllerExample`, `testCustomActuatorExample`, `testExampleMain` and so on. 
This PR adds the function `addNumRowsColumns()` as a compatibility (between `STOFileAdapter` and `MarkerData`) fix.
So both these functions need to exist, unless we change the STO files used by the tests I mention above. I can make the change if you like.

PR #1352, opened 2016-10-18
Forward CMake args to dependencies during super-build

klshrinidhi:
Forward commonly used CMake settings to dependencies during super-build.

This can be used for example to silence warnings while building dependencies.


chrisdembia:
LGTM. Feel free to merge once tests pass.

Issue #1353, opened 2016-10-18
python: Can't find libBTKCommon.so.0.4dev


kidzik:
After installing the current version on Ubuntu 16.04, installing python bindings and running 
`import opensim` in python I'm getting: 

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/opensim-4.0-py2.7.egg/opensim/__init__.py", line 2, in <module>
    from common import *
  File "/usr/local/lib/python2.7/dist-packages/opensim-4.0-py2.7.egg/opensim/common.py", line 28, in <module>
    _common = swig_import_helper()
  File "/usr/local/lib/python2.7/dist-packages/opensim-4.0-py2.7.egg/opensim/common.py", line 24, in swig_import_helper
    _mod = imp.load_module('_common', fp, pathname, description)
ImportError: libBTKCommon.so.0.4dev: cannot open shared object file: No such file or directory
```

This

```
sudo cp opensim_dependencies_install/BTK/lib/btk-0.4dev/* /usr/lib
```

solves the problem but maybe I should just set up some path? In any case it seems it's not documented (I will try to fix such documentation issues by myself as soon as I get through the documentation by myself and understand the architecture better :).

klshrinidhi:
Please post how you built OpenSim ( all the commands you used ) so I can reproduce the problem.

kidzik:
I followed the "impatient" path:
```
sudo apt-get update
sudo apt-get --yes install git cmake cmake-curses-gui \
                           freeglut3-dev libxi-dev libxmu-dev \
                           liblapack-dev swig python-dev \
                           openjdk-8-jdk
export JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
git clone https://github.com/opensim-org/opensim-core.git
mkdir opensim_dependencies_build
cd opensim_dependencies_build
cmake ../opensim-core/dependencies/ \
      -DCMAKE_INSTALL_PREFIX='~/opensim_dependencies_install' \
      -DCMAKE_BUILD_TYPE=RelWithDebInfo
make -j8
cd ..
mkdir opensim_build
cd opensim_build
cmake ../opensim-core \
      -DCMAKE_INSTALL_PREFIX="~/opensim_install" \
      -DCMAKE_BUILD_TYPE=RelWithDebInfo \
      -DOPENSIM_DEPENDENCIES_DIR="~/opensim_dependencies_install" \
[...]

PR #1354, opened 2016-10-19
Add install rpath for libraries on Linux

klshrinidhi:
Addressing #1353.


chrisdembia:
Cool stuff. I think overall this is a good idea, and a good way to address #1353 and reduce headache for users.

Could you describe what testing you did? I don't think the Travis-CI tests are not sufficient, because of the line `export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$OPENSIM_HOME/lib:$OPENSIM_HOME/lib/x86_64-linux-gnu:$OPENSIM_DEPENDENCIES_INSTALL_DIR/BTK/lib/btk-0.4dev;`.

I think the behavior should be more similar to the behavior for OSX, and that perhaps the two platforms should not be treated differently. For example, for APPLE, we have logic that avoids using RPATH unnecessarily if installing into system directories like `/usr`, and a CMake option to toggle the use of RPATH.

What about just changing the preceding if-statement from ` if(APPLE AND NOT (${CMAKE_VERSION} VERSION_LESS 2.8.12))` to `if(UNIX)`?

klshrinidhi:
> Could you describe what testing you did?

Testing was done locally on my machine.

> I think the behavior should be more similar to the behavior for OSX, ...

I disagree. I think the behavior should be to use RPATH. If I build OpenSim from source using my *non-standard* installation of dependencies, I would expect OpenSim installation to refer to same dependency binaries I specified during build over those available in *standard directories*. 

> ... and that perhaps the two platforms should not be treated differently.

I think the behavior in OSX should always use RPATH.

--------------------------------
I think this PR addresses the issue. If this can be tested in CI without losing more time, that will be great.

jenhicks:
@klshrinidhi will look into solution that won't create issues when distributing binaries (use relative path with rpath, use env variable, or other technique)

jenhicks:
can you also try to use python bindings (under the install, not build directory) during testing as a double check.

klshrinidhi:
I have updated the PR. 
I am able to:
```python
import opensim
```
from `opensim_install_dir/lib/python2.7/site-packages`. All the dynamic libraries have their rpath set appropriately.

CMakeLists.txt
```diff
@@ -483,6 +483,11 @@ if(APPLE AND NOT (${CMAKE_VERSION} VERSION_LESS 2.8.12))
         endif()
     endif()
 endif()
+if(PLATFORM_NAME MATCHES Linux)
+    set(CMAKE_INSTALL_RPATH
+        "${CMAKE_INSTALL_RPATH};${CMAKE_INSTALL_PREFIX}/lib")
```
chrisdembia: Should `lib` be `${CMAKE_INSTALL_LIBDIR}`?

klshrinidhi: Done.

chrisdembia:
It does seem like this fixes the original bug #1353. Thanks for testing the installed Python package.

I had put a lot of effort into properly setting up RPATH for OSX, following guidelines in https://cmake.org/Wiki/CMake_RPATH_handling, and I feel that the Linux solution ignores what was done there. For example, [I tried to make the installation relocatable despite the use of RPATH](https://github.com/opensim-org/opensim-core/blob/188a17e6495718af888986fd70e37c81413b2a45/cmake/OpenSimMacros.cmake#L258). I agree that it is far less likely that Linux packages need to be relocatable, but it could be done with `$ORIGIN` (see link above). If we do not make the package relocatable on Linux, this should at least be documented, perhaps [here](https://github.com/opensim-org/opensim-core/pull/1354/files#diff-af3b638bc2a3e6c650974192a53c7291L485). 

I would have thought that the line `set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)` would mean that it is not necessary to manually add BTK libs to the RPATH.

The variable `OPENSIM_USE_INSTALL_RPATH` is ignored on Linux (the variable is used in OpenSimMacros.cmake and when building Python and Java bindings); I would prefer a solution that made use of this variable on Linux, but if not, then this variable should be renamed so it is clear it is only used on OSX.

I would have preferred that the `OPENSIM_NO_LIBRARY_PATH_ENV_VAR` would be changed to apply to Linux as well, so that future packagers could decide if use of RPATH is necessary. But this is fine; we can make it an option later if necessary.

In summary, please document the fact that the Linux package is not relocatable, and that `OPENSIM_USE_INSTALL_RPATH` is ignored on Linux. If it's easy, please also determine why manually adding BTK libs to the RPATH is necessary, given that you are also setting `CMAKE_INSTALL_RPATH_USE_LINK_PATH` to TRUE.

chrisdembia:
Thank you for making those changes. Please merge after tests pass, especially any tests of bindings.

chrisdembia:
A thought for the future: do MATLAB bindings work fine with BTK?

klshrinidhi:
One build failure -- timeout after 80/81 tests pass.
Merging.

PR #1359, opened 2016-10-22
Add getThelen2003MuscleList() & getMillard2012EquilibriumMuscleList()

klshrinidhi:
Addressing #643.

Reviewer -- @aymanhab, @tkuchida

### DIFF
--- a/Bindings/OpenSimHeaders_simulation.h
+++ b/Bindings/OpenSimHeaders_simulation.h
@@ -138,5 +138,12 @@
 #include <OpenSim/Simulation/StatesTrajectory.h>
 #include <OpenSim/Simulation/StatesTrajectoryReporter.h>
 
+#include <OpenSim/Actuators/osimActuatorsDLL.h>
+#include <OpenSim/Actuators/MuscleFixedWidthPennationModel.h>
+#include <OpenSim/Actuators/MuscleFirstOrderActivationDynamicModel.h>
+#include <OpenSim/Actuators/Thelen2003Muscle.h>
+#include <OpenSim/Actuators/Millard2012EquilibriumMuscle.h>
+#include <OpenSim/Actuators/FiberCompressiveForceCosPennationCurve.h>
+#include <OpenSim/Actuators/FiberCompressiveForceLengthCurve.h>
 #endif // OPENSIM_OPENSIM_HEADERS_SIMULATION_H_
 
--- a/Bindings/simulation.i
+++ b/Bindings/simulation.i
@@ -236,3 +236,17 @@
 %include <OpenSim/Actuators/Thelen2003Muscle.h>
 %include <OpenSim/Actuators/Millard2012EquilibriumMuscle.h>
 
+%template(Thelen2003MuscleList)
+    OpenSim::ComponentList<const OpenSim::Thelen2003Muscle>;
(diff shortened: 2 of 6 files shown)


Bindings/OpenSimHeaders_simulation.h
```diff
@@ -138,5 +138,12 @@
 #include <OpenSim/Simulation/StatesTrajectory.h>
 #include <OpenSim/Simulation/StatesTrajectoryReporter.h>
 
+#include <OpenSim/Actuators/osimActuatorsDLL.h>
+#include <OpenSim/Actuators/MuscleFixedWidthPennationModel.h>
```
tkuchida: Is it necessary to include these Actuator headers in OpenSimHeaders_simulation.h? Currently, each OpenSimHeaders_[blah].h file strictly includes headers from OpenSim/[blah]. Why is it now necessary to deviate from this pattern?

Bindings/simulation.i
```diff
@@ -229,3 +229,24 @@
...
+%template(Thelen2003MuscleIterator)
+    OpenSim::ComponentListIterator<const OpenSim::Thelen2003Muscle>;
+
+%template(Millard2012EquilibriumMuscleList)
+    OpenSim::ComponentList<const OpenSim::Millard2012EquilibriumMuscle>;
+%template(Millard2012EquilibriumMuscleIterator)
+    OpenSim::ComponentListIterator<const OpenSim::Millard2012EquilibriumMuscle>;
+
+%template(getThelen2003MuscleList)
+  OpenSim::Model::getComponentList<Thelen2003Muscle>;
```
tkuchida: Probably doesn't matter, but based on the rest of this file, `Thelen2003Muscle` should be `OpenSim::Thelen2003Muscle`.

Bindings/OpenSimHeaders_simulation.h
klshrinidhi: I think it is necessary to have them here to add these functions to `Model`.

tkuchida: Ah, ok. Thanks.

aymanhab:
@klshrinidhi Is there a reason ClutchedPathSpring header was removed from the actuators.i file or did it go somewhere else? Thanks.

klshrinidhi:
>@klshrinidhi Is there a reason ClutchedPathSpring header was removed from the actuators.i file or did it go somewhere else? Thanks.

It is still in `actuators.i`. You can see it [here](
https://github.com/opensim-org/opensim-core/blob/issue_643_component_list_muscles/Bindings/actuators.i#L8). Not sure why GitHub doesn't show.

aymanhab:
Thanks @klshrinidhi for the clearup.
One last item, did you consider having the Java test perform "safeDownCast" to check the type (similar to the python test) so that property names are not hard coded and the test can stay intact regardless of changes in property names?

klshrinidhi:
> One last item, did you consider having the Java test perform "safeDownCast" to check the type (similar to the python test) so that property names are not hard coded and the test can stay intact regardless of changes in property names?

I have changed the test to look like the one in Python.

aymanhab:
Thank you @klshrinidhi and @tkuchida

PR #1362, opened 2016-10-25
Disable C3D Java test if WITH_BTK is OFF

klshrinidhi:
Currently building with `WITH_BTK=OFF` and `JAVA_WRAPPING=ON` fails because Java C3D test is unable to find the appropriate class (which was not built with BTK off).
This PR fixes this problem.

Also, this PR adds some Python tests for C3D to look similar to Java tests for C3D.


aymanhab:
LGTM, feel free to merge when tests pass.

PR #1363, opened 2016-10-25
Incorrect number format used in error message

tkuchida:
Using `%d` rather than `%g` produces a message like `max = -379567735` rather than `max = 1.571`.


tkuchida:
I think this is safe to go in (the :x: :x: :x: :x: are because I forgot `[ci skip]` and canceled the builds manually). Care to merge, @klshrinidhi?

aseth1:
LGTM

tkuchida:
Thanks! :jack_o_lantern:

Issue #1368, opened 2016-10-28
Got stuck with double types in python


nikorose87:
Hi 

I am trying to do the simple TugofWar exercise in python just for learning a bit more, Unfortunately,  I got stuck setting bounds on coordinates because the `setRange` asked for a double type. I have tried to change the type of the numbers with the next commands:
- c_double
- float
- ArrayDouble

but all of them didn't work, throwing the next error:

`TypeError: in method 'Coordinate_setRange', argument 2 of type 'double [2]'`

The part of the code is here in case it could be useful:

```

# set bounds on coordinates
#angleRange = osim.ArrayDouble.createVec3(-math.pi/2,math.pi/2,0)
# Creates a list containing 5 lists, each of 8 items, all set to 0
#w, h = 1, 2
#angleRange = [[0 for x in range(w)] for y in range(h)]
#positionRange = [[0 for x in range(w)] for y in range(h)]
#angleRange = (c_double(-math.pi/2),c_double(math.pi/2))
#angleRange[0][0] = c_double(-math.pi/2)
#angleRange[1][0] = c_double(math.pi/2)
#positionRange = (c_double(-1),c_double(1))
angleRange = osim.ArrayDouble(-math.pi/2)
angleRange.append(math.pi/2)
positionRange = osim.ArrayDouble(-math.pi/2)
print (angleRange,positionRange)
blockToGround.upd_coordinates(0).setRange(angleRange)
blockToGround.upd_coordinates(1).setRange(angleRange)
blockToGround.upd_coordinates(2).setRange(angleRange)
blockToGround.upd_coordinates(3).setRange(positionRange)
blockToGround.upd_coordinates(4).setRange(positionRange)
blockToGround.upd_coordinates(5).setRange(positionRange)

```

thanks!

chrisdembia:
Use `Coordinate::setRangeMin()` and `Coordinate::setRangeMax()`. Please close this issue if this works for you.

nikorose87:
Well Thanks

It works in this way:

```
blockToGround.upd_coordinates(0).setRangeMin(float(-math.pi/2))
blockToGround.upd_coordinates(0).setRangeMax(float(math.pi/2))
```

Thanks again!

aymanhab:
Chris, I'm ok with closing as long as we open an issue to address or remove
unusable methods from bindings if that's the case here.

Thanks
Ayman

nikorose87:
Dear Crhis

[...]

PR #1370, opened 2016-10-31
Remove method Coordinate::setRange from Python binding as it can't be…

aymanhab:
… called from Python. This addresses issue #1368 one way by removing the method from Python bindings. A possible more general fix is to use a higher level datastructure (e.g. SimTK::Vector or OpenSim::Array) so the interface from Python is not different.

### DIFF
--- a/Bindings/Python/swig/python_simulation.i
+++ b/Bindings/Python/swig/python_simulation.i
@@ -18,6 +18,9 @@ using namespace OpenSim;
 using namespace SimTK;
 %}
 
+// Ignore method that is not callable from Python (uses double[] arg)
+%ignore OpenSim::Coordinate::setRange;
+
 
 %include "python_preliminaries.i"
 
@@ -212,3 +215,4 @@ SET_ADOPT_HELPER(Analysis);
         return $self->get(i);
     }
 };
+


aymanhab:
@klshrinidhi Ideally we can change the type of Argument (e.g. to SimTK::Vector<double>) and use implicit  conversion but I haven't tried that. @chrisdembia any thoughts?

chrisdembia:
LGTM.

The ideal C++ interface is probably to just have two parameters:

```cpp
void Coordinate::setRange(double min, double max);
```

Alternatively, we could use a single parameter that is a container that can only have 2 entries:
```cpp
void Coordinate::setRange(std::pair<double>);
void Coordinate::setRange(std::tuple<double, double>);

```

klshrinidhi:
I think this function is worth changing on C++ side to take two arguments (min & max). I can do it if you like.

aymanhab:
Thanks @chrisdembia and @klshrinidhi I agree about the change but there're a couple related items:
1. It's disturbing that SWIG generates code that can't be run/invoked from Python when it knows the context, is this a bug or is there a workaround in Python that we don't know of, need some investigation.
2. Assuming there's no such workaround, it would be good to find all other instances where a raw array [] is passed through the API and replace those methods with either SimTK::Vector<double> if size is unknown or a SimTK::Vec3, Vec2. etc. or multiple arguments if size is known.
3. pair and tuple are also not language agnostic so I'd try to avoid introducing them in the API.

Issue #1350, opened 2016-10-18
Add unit test for InverseKinematicsSolver


tkuchida:
InverseKinematicsSolver is used by InverseKinematicsTool and MarkerPlacer but isn't being tested in isolation (unit test is missing); `InverseKinematicsSolver::updateMarkerWeight()` is not being tested anywhere. Possible test cases for InverseKinematicsSolver include the following:
- Call `assemble()`, use `setAccuracy()` to tighten the tolerance, and call `assemble()` again. The total marker error should be lower after the second call to `assemble()`. Note that the second call must be `assemble()` rather than `track()` because the latter will not propagate the new accuracy to the underlying `SimTK::Assembler` (see PR #1343).
- Create a multi-marker model and set only one marker weight high (:cow:). (For this test, it must be impossible to satisfy all marker targets simultaneously.) Call `assemble()` and verify that :cow: has been correctly positioned. Now use `updateMarkerWeight()` to set :cow:'s weight low and a different marker's weight high (:cow2:). Call `assemble()` again and verify that :cow2: is now correctly positioned.

jenhicks:
@aymanhab Can you take a look?

PR #1371, opened 2016-11-01
Test InverseKinematicsSolver handles updates to to accuracy and marker weights

aseth1:
Addresses #1350.

Includes three test cases:

1. test that the *accuracy* setting effects the number of digits to which the solution can be "trusted". It verifies between a loose and tight accuracy value that the number of digits recovered increases with accuracy. 
2. test that updating the marker weights effects subsequent call to `assemble()`. It verifies that the marker error for the marker with increased weight is reduced.
3. test that repeated calls to `updateMarkerWeights()` followed by `track()` also effects the solution. It verifies that increasing a marker's weight also reduces its error over a static trial.

The main modification was to update `MarkersReference`'s internal ordered list of weights if the properties have been modified (e.g. `updateMarkerWeights()`).


aseth1:
@tkuchida it would be  great if you could review. @aymanhab could you provide a secondary review as well?
Thank you.

OpenSim/Simulation/MarkersReference.cpp
```diff
@@ -77,19 +77,13 @@ void MarkersReference::populateFromMarkerData(const MarkerData& aMarkerData)
     _markerNames.assign(nm, "");
     _weights.assign(nm, get_default_weight());
 
-    int index = 0;
-    // Build flat lists (arrays) of marker names and weights in the same order as the marker data
-    for(int i=0; i<tempNames.getSize(); i++){
+    for (int i = 0; i < tempNames.getSize(); i++) {
         const std::string &name = tempNames[i];
         _markerNames[i] = name;
```
tkuchida: Should there be a check here to ensure the specified marker name exists in the model?

aymanhab:
Thanks @tkuchida. @aseth1 will review once @tkuchida  concerns are addressed.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
```diff
@@ -0,0 +1,457 @@
...
+        // collect results into reporter
+        m->realizeReport(state);
+    }
+
+    auto maxval = noise.max();
+
+    // make a copy of the reported table
+    auto results = markerReporter->getTable();
+
+    SimTK::Vec3 offset = noiseRadius*SimTK::Vec3(double(noise(gen)),
```
tkuchida: `offset` is defined within the loop (L439), so this line can be removed.

tkuchida: Oh I see what's happening: the same offset is used for all markers and at all time steps if the `fixed` flag is set to `true`. A comment here would be helpful. Also perhaps consider changing the name of the flag from "fixed" to something more descriptive, like "constantOffset".

OpenSim/Simulation/InverseKinematicsSolver.cpp
```diff
@@ -218,18 +218,13 @@ void InverseKinematicsSolver::setupGoals(SimTK::State &s)
...
     const MarkerSet &modelMarkerSet = getModel().getMarkerSet();
 
-    // get markers with specified tasks/weights
-    const Set<MarkerWeight>& mwSet = _markersReference.updMarkerWeightSet();
-    
     int index = -1;
-    int wIndex = -1;
     SimTK::Transform X_BF;
     //Loop through all markers in the reference
     for(unsigned int i=0; i < markerNames.size(); ++i){
```
tkuchida: Actually, I don't think `unsigned i=0;` will cause a warning on Linux. I think the warning I was remembering would occur if it were `for(int i=0; i<markerNames.size(); ++i)` because then `i` (an int) would be compared to `markerNames.size()` (an unsigned). I prefer to be very explicit now (e.g., `unsigned i=0u;`, but it is not necessary.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
```diff
@@ -0,0 +1,457 @@
...
+    cout << "Assembled " << coord.getName() << " value = "
+        << coordValue << endl;
+
+    SimTK::Array_<double> nominalMarkerErrors;
+    ikSolver.computeCurrentMarkerErrors(nominalMarkerErrors);
+
+    SimTK::Array_<double> markerWeights;
+    markersRef.getWeights(state, markerWeights);
+
+    for (unsigned int i = 0; i < markerNames.size(); ++i) {
```
tkuchida: Actually, `unsigned i=0;` is fine; see comment above.

OpenSim/Simulation/MarkersReference.cpp
```diff
@@ -77,19 +77,13 @@ void MarkersReference::populateFromMarkerData(const MarkerData& aMarkerData)
     _markerNames.assign(nm, "");
     _weights.assign(nm, get_default_weight());
 
-    int index = 0;
-    // Build flat lists (arrays) of marker names and weights in the same order as the marker data
-    for(int i=0; i<tempNames.getSize(); i++){
+    for (int i = 0; i < tempNames.getSize(); i++) {
```
aseth1: Done.

OpenSim/Simulation/MarkersReference.cpp
aseth1: The decision to track an individual marker reference is left to the `InverseKinematicsSolver` (`setupGoals()`), which operates on a `Model` and uses the common subset of markers (between the Model and Reference). But the `MarkersReference` itself has no knowledge of the model.

OpenSim/Simulation/MarkersReference.cpp
```diff
@@ -153,4 +148,37 @@ void MarkersReference::setMarkerWeightSet(const Set<MarkerWeight>& markerWeights
     upd_marker_weights() = markerWeights;
 }
 
+void MarkersReference::setDefaultWeight(double weight)
+{
+    set_default_weight(weight); 
+}
+
+void MarkersReference::updateInternalWeights() const
```
aseth1: The `MarkersReference` maintains a flat private list of all the marker weights in the same order as the marker values.  This was done for convenience since the weights come from either the marker weight set or default weight properties, which are generally not modified and so there is efficiency to be gained by maintaining the internal `_weights`. If the weight properties are updated the output weights (`getWeights()`) must make sure the list is up-to-date. To ensure weights are always up to date with property values,  the update is called within the `getWeights() const` and therefore must also be a const method. The `updateInternalWeights() const` does not change any of the properties of the `MarkersReference` it merely transcribes the property values (const) into a single list for convenience/efficiency.

OpenSim/Simulation/MarkersReference.cpp
```diff
@@ -153,4 +148,37 @@ void MarkersReference::setMarkerWeightSet(const Set<MarkerWeight>& markerWeights
...
+    // else to be done.
+    if (get_marker_weights().isObjectUpToDateWithProperties())
+        return;
+
+    int wix = -1;
+    int ix = 0;
+    // Build flat lists of marker weights in the same order as the marker names
+    for (const std::string &name : _markerNames) {
+        wix = get_marker_weights().getIndex(name, wix);
+        //Assign user weighting for markers that are user listed in the input set
```
aseth1: I did intend *are* but I've rephrased anyway:
```cpp
        // Associate user weights (as specified in the marker_weights property)
        // with the corresponding marker by order of marker names
```

OpenSim/Simulation/MarkersReference.cpp
```diff
@@ -153,4 +148,37 @@ void MarkersReference::setMarkerWeightSet(const Set<MarkerWeight>& markerWeights
...
+        if (wix >= 0)
+            _weights[ix++] = get_marker_weights()[wix].getWeight();
```
aseth1: Good point. I added details about the markers being assigned a default weight to begin with.

OpenSim/Simulation/Model/ModelComponentSet.h
```diff
@@ -92,14 +92,14 @@ OpenSim_DECLARE_CONCRETE_OBJECT_T(ModelComponentSet, T, Set<T>);
...
 
     /** Does this Set have a Model associated with it? */
-    bool hasModel() const { return _model.get()!=nullptr; }
+    bool hasModel() const { return !_model.empty(); }
     /**
      * Get this Model this set is part of.
      */
     const Model& getModel() const
     {
         if (_model){
```
aseth1: Done.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
```diff
@@ -0,0 +1,457 @@
...
+MarkerData* generateMarkerDataFromModelAndStates(const Model& model,
+    const StatesTrajectory& states, double noiseRadius=0, bool fixed=false);
+
+// Verify that accuracy improves the number of decimals points to which
+// the solver solution (coordinates) can be trusted as it is tightened.
+void testAccuracy();
+// Verify that the marker weights impact the solver and has the expected
+// effect of reducing the error for the marker weight that is increased. 
+void testUpdateMarkerWeights();
+// Verify that the track() solution is also effected by updating marker
```
aseth1: I think it is correct to verify the *effect* on the solution (which tests the affect of marker weights).

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
```diff
@@ -0,0 +1,457 @@
...
+{
+    SimTK::Array_<std::string> failures;
+
+    try { testAccuracy(); }
+    catch (const std::exception& e) {
+        cout << e.what() << endl; failures.push_back("testAccuracy");
+    }
+    try { testUpdateMarkerWeights(); }
+    catch (const std::exception& e) {
+        cout << e.what() << endl; failures.push_back("testUpdateGoalWeights");
```
aseth1: Whops! Fixed.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
```diff
@@ -0,0 +1,457 @@
...
+    }
+    try { testTrackWithUpdateMarkerWeights(); }
+    catch (const std::exception& e) {
+        cout << e.what() << endl; failures.push_back("testAccuracy");
```
aseth1: Sorry for copy and paste bug. Fixed.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
```diff
@@ -0,0 +1,457 @@
...
+//=============================================================================
+
+void testAccuracy()
+{
+    cout << "\ntestInverseKinematicsSolver::testAccuracy()" << endl;
+
+    std::unique_ptr<Model> pendulum{ constructPendulumWithMarkers() };
+    Coordinate& coord = pendulum->getCoordinateSet()[0];
+
+    double refVal = 0.123456789;
```
aseth1: `refVal0` wasn't necessary, so it has been removed. I used `refVal` for its correspondence with it being the `CoordinateReference` value.  `coordVal` was intended to indicate the model's coordinate value. Would you prefer `referenceValue` and `modelValue` (since they are both Coordinate values)?

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
```diff
@@ -0,0 +1,457 @@
...
+    double looseAccuracy = 1e-3;
+    double tightAccuracy = 1.0e-9;
```
aseth1: Considering that the this range in accuracy can be found in use from the model's assembly solver to the InverseKinematicsTool I had the opposite thought that we should verify that solutions were consistent with this range.

I am not convinced that it is the correct use of *accuracy* to reduce marker errors. Accuracy is the number of digits that can be trusted in the solution, and that is the coordinate value in this case. I believe the test correctly verifies that number of correct digits in the coordinate value increase with tightening accuracy.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
```diff
@@ -0,0 +1,457 @@
...
+
+    coordValue = coord.getValue(state);
+    cout << "Assembled " << coord.getName() << " value = " << coordValue << endl;
+
+    double accuracy = abs(coordValue - refVal);
+    cout << "Specified accuracy: " << looseAccuracy << "; achieved: "
+        << accuracy << endl;
+
+    // test should fail if we did not meet the target accuracy
+    SimTK_ASSERT_ALWAYS(accuracy <= looseAccuracy,
```
aseth1: Should *NOT* accept equivalence with specified accuracy? I don't understand why not?

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
```diff
@@ -0,0 +1,457 @@
...
+    coordValue = coord.getValue(state);
+    cout << "Assembled " << coord.getName() <<" value = "<< coordValue << endl;
+
+    accuracy = abs(coord.getValue(state) - refVal);
+
+    cout << "Specified accuracy: " << tightAccuracy << "; achieved: "
+        << accuracy << endl;
+
+    // test should fail if we did not meet the new tighter target accuracy
+    SimTK_ASSERT_ALWAYS(accuracy <= tightAccuracy,
```
aseth1: If one gets exactly the accuracy specified, which is very unlikely, why should that be a failure?

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
```diff
@@ -0,0 +1,457 @@
...
+    ikSolver.computeCurrentSquaredMarkerErrors(sqMarkerErrors);
+    for (const double& err : sqMarkerErrors) {
+        tightSumSqError += err;
+    }
+
+    cout << "For accuracy: " << tightAccuracy << "; Sum-squared Error: "
+        << tightSumSqError << endl;
+
+    // refining the accuracy should not increase tracking errors
+    SimTK_ASSERT_ALWAYS(tightSumSqError <= looseSumSqError,
```
aseth1: I am not sure that solution accuracy is affects marker error. I definitely would not expect it to get worse, but I am not even totally sure about that now.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
```diff
@@ -0,0 +1,457 @@
...
+            << ") error = " << nominalMarkerErrors[i] << endl;
+    }
+
+    // Increase the weight of the right marker 
+    markerWeights[1] *= 10.0;
+    ikSolver.updateMarkerWeights(markerWeights);
+
+    // Reset the initial coordinate value
+    coord.setValue(state, 0.0);
+    ikSolver.assemble(state);
```
aseth1: I have separate tests for `assemble()` and `track()` (see [`testTrackWithUpdateMarkerWeights()`](https://github.com/opensim-org/opensim-core/pull/1371/files#diff-322146385f71292958514f4a29ec2e4eR283)). For `assemble()` I want to verify that starting from the same initial conditions we get the expected change in marker errors due to change in marker weights. In `testTrackWithUpdateMarkerWeights()` I verify that subsequent `track()` calls refine the solution according to the change in weights for the use case you describe.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
```diff
@@ -0,0 +1,457 @@
...
+    m->addComponent(markerReporter);
+
+    SimTK::State s = m->initSystem();
+
+    for (const auto& state : states) {
+        // collect results into reporter
+        m->realizeReport(state);
+    }
+
+    auto maxval = noise.max();
```
aseth1: It was a remnant of previous characterizations. It's been removed. Thanks!

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
aseth1: good point. I renamed as you suggested.

Also there was related documentation for the method listed with the prototype (at the top of the .cpp) but perhaps I should move it down to the definition. Any preference?

OpenSim/Simulation/MarkersReference.cpp
tkuchida: I see. I would have made `updateInternalWeights()` non-const and done `const_cast<Self*>(this)` from within `getWeights()` to update, but I think you're treating `updateInternalWeights()` like a cache-updating method, which makes sense. Thanks for the explanation.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
tkuchida: Yes, it is correct to say "we verify the *effect* on the solution", but on L53 the word "effected" is used as a verb and therefore means "caused to happen" AFAIK. The comment is saying that updating marker weights *caused* the track() solution, which I don't think makes sense because a solution is computed, not *caused* (*effected*).

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
tkuchida: Ah, ok I see. Removing `refVal0` helps. I think replacing `refVal` with `coordRefVal` might also be helpful.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
tkuchida: > Considering that the this range in accuracy can be found in use from the model's assembly solver to the InverseKinematicsTool I had the opposite thought that we should verify that solutions were consistent with this range.

My suggestion is meant to ensure that tightening the accuracy does, indeed, refine the solution. If the accuracy were tightened a little and the solution improved a little, I would be more confident that things were working correctly than if the accuracy were tightened a lot and the solution improved. However, IIRC, the solution will be *at least* as accurate as the accuracy that was requested, so tightening the accuracy by only a little may not actually change the solution. I'm fine with leaving `looseAccuracy` and `tightAccuracy` as they are.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
tkuchida: Sorry, I misinterpreted this as if it were `OPENSIM_THROW_IF()`, I think because the comment is written that way around (i.e., conditions for failure). No changes necessary.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
tkuchida: As above: I misinterpreted this as if it were `OPENSIM_THROW_IF()` , I think because the comment describes the conditions for failure but the assert states conditions for success. No changes necessary.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
tkuchida: As above. No change necessary.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
tkuchida: > there was related documentation for the method listed with the prototype (at the top of the .cpp) but perhaps I should move it down to the definition. Any preference?

Ah, sorry I didn't remember it was up there. This is a stylistic issue so my preference is irr:elephant:.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
aseth1: I agree there was a possible inconsistency where a user could `setAccuracy()` with the expectation that it would affect the accuracy of a subsequent call to `track()`. I have updated the documentation to make it clear that a change in accuracy requires a call to assemble() before subsequent calls to `track()` will reflect the new accuracy. 

In addition, setting the accuracy and calling `track()` directly throws an exception stating that `assemble()` must be invoked first. 

The result of setting accuracy on `track()` is added to the accuracy test and verifies that an exception is thrown in the case that `assemble()` is not called prior to `track()`.

aseth1:
@tkuchida I have attempted to address the issue where `setAccuracy()` was not affecting subsequent calls to `track()`. Now `track()` will throw if the accuracy is changed and a call to `assemble()` is required (dosxygen was also updated). I added conditions in the `testAccuracy()` case to verify. 

@aymanhab please let me know if there are any additional concerns you may have.

tkuchida:
Changes look good. It is fine to ignore my comment in testInverseKinematicsSolver.cpp L53.

aseth1:
@tkuchida Thank you. The tests have been significantly strengthened because of your review.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
aymanhab: Test looks great but uses MarkerData class, will that need to change with pending phasing out of MarkerData class?

aymanhab:
Finished with review, looks great. Minor comments above.  :+1:

PR #1372, opened 2016-11-02
Replace MarkerData with TimeSeriesTable in MarkersReference

klshrinidhi:
This PR is one step toward removing MarkerData. It replaces usage of MarkerData with DataTable in the class MarkersReference which is used by IK, ID and Scale.

If this gets merged, the plan is to remove MarkerData completely in the next PR.


jenhicks:
@aseth1 can you review? @klshrinidhi also mentioned that you guys both started a PR at the same time that modified the same file, so we will have to merge those changes.

OpenSim/Common/TimeSeriesTable.h
```diff
@@ -183,11 +222,88 @@ class TimeSeriesTable_ : public DataTable_<double, ETY> {
...
+                               is greater than or equal to the last value in the
+                               time column, the row returned is the last row. 
+                               This operation only returns existing rows and 
+                               does not perform any interpolation. Defaults to
+                               'true'.
+
+    \throws TimeOutOfRange If the given value is out-of-range of time column.
+    \throws EmptyTable If the table is empty.                                 */
+    RowVectorView
+    getRowNear(const double& time,
```
aseth1: consider renaming `getNearestRow`. This reminds me of similar method for finding the nearest state at a given time that @chrisdembia was working on for the `StatesTrajectory` and it would be good to use consistent naming.

OpenSim/Simulation/MarkersReference.cpp
```diff
@@ -24,133 +24,168 @@
...
 
-
-MarkersReference::MarkersReference(MarkerData *markerData,
-    const Set<MarkerWeight>* markerWeightSet) : MarkersReference()
-{
-    _markerData.reset(markerData);
-    if (markerWeightSet!=nullptr) upd_marker_weights()= *markerWeightSet;
-    populateFromMarkerData(*markerData);
+MarkersReference::
+MarkersReference(const TimeSeriesTable_<SimTK::Vec3>& markerTable,
```
aseth1: Yay! Nice to see `MarkerData` go.

aseth1: If you use this constructor with a Table you read in with  TRCFileAdapter we will not have the Units reflected when asking for the reference value, which is assumed to be in SI units.

As a *note*, a `TableSource` created from a TRC file would have the same issue, where the column output values would be in *mm*, for example, but all the tools, algorithms expect *m*.

OpenSim/Simulation/MarkersReference.cpp
```diff
@@ -24,133 +24,168 @@
...
+        _markerTable = TimeSeriesTableVec3{markerFile};
+    } else {
+        try {
+            _markerTable = (TimeSeriesTable{markerFile}).pack<SimTK::Vec3>();
+        } catch(const IncorrectTableType&) {
+            _markerTable = TimeSeriesTable_<SimTK::Vec3>{markerFile};
+        }
+    }
+    if(_markerTable.hasTableMetaDataKey("Units"))
+        _units = Units{_markerTable.getTableMetaData<std::string>("Units")};
```
aseth1: Units are handled correctly if we load from file. Can we abstract this conversion and use it at the time the reference value is requested. Then it would not matter how the Table was supplied (directly or by file).

Minor, Units is a local helper and does not need to be data member.

OpenSim/Simulation/MarkersReference.cpp
```diff
@@ -24,133 +24,168 @@
...
+    return static_cast<int>(_markerTable.getNumColumns());
+}
+
+double
+MarkersReference::getSamplingFrequency() const {
+    if(_markerTable.hasTableMetaDataKey("DataRate")) {
+        auto datarate = _markerTable.getTableMetaData<std::string>("DataRate");
+        return std::stod(datarate);
+    } else
+        return 250;
```
aseth1: If no data rate is provided it should return NaN.

OpenSim/Tools/MarkerPlacer.cpp
```diff
@@ -242,6 +243,46 @@ bool MarkerPlacer::processModel(Model* aModel,
...
+        for(unsigned r = 0; r < staticPoseTable.getNumRows(); ++r)
+            staticPoseTable.updRowAtIndex(r) *= scaleFactor;
+
+        staticPoseUnits = aModel->getLengthUnits();
+        staticPoseTable.removeTableMetaDataKey("Units");
+        staticPoseTable.addTableMetaData("Units",
+                                         staticPoseUnits.getAbbreviation());
+    }
+    
     MarkerData* staticPose = new MarkerData(aPathToSubject + _markerFileName);
```
aseth1: if using the `staticPoseTable` below to create the `MarkersReference` why not remove `staticPose` now since it seems to be unused?

OpenSim/Common/DataTable.h
```diff
@@ -220,10 +221,164 @@ class DataTable_ : public AbstractDataTable {
...
+    3 consecutive columns of the given DataTable will form one column of this
+    DataTable. The column labels of this table will be formed by stripping out
+    the suffixes from the column labels of the given DataTable. For the same 
+    example above, if columns labels of the given DataTable are -- 
+    "col0.x", "col0.y", "col0.x", "col1.x", "col1.y", "col1.x" -- the column 
+    labels of this DataTable will be -- "col0", "col1" -- where suffixes are
+    stripped out. This constructor will try to guess the suffixes used. If 
+    unable to do so, it will throw an exception. Suffixes used can also be 
+    specified as arguments. 
+    This constructor only makes sense for DataTable_<double, "not double">.
```
klshrinidhi: I have elaborated on what "not double" means. It should be clear now.

OpenSim/Common/DataTable.h
```diff
@@ -220,10 +221,164 @@ class DataTable_ : public AbstractDataTable {
...
+                        }
+                    }
+                }
+                suffs.push_back(suff);
+            }
+        }
+
+        // Form column labels for this table from that table.
+        std::vector<std::string> thisLabels{};
+        for(unsigned c = 0; c < thatLabels.size(); ) {
```
klshrinidhi: I have added `reserve`.

OpenSim/Common/TimeSeriesTable.h
klshrinidhi: I think `getRowNear(<time>)` looks okay. Need not say `nearest`.

OpenSim/Simulation/MarkersReference.cpp
klshrinidhi: I think this problem existed with `MarkerData` as well. When using this constructor, (as per previous implementation) one is expected to make sure the quantities in `MarkerData` are in the correct units. I do this kind of unit conversion in `MarkerPlacer`.

OpenSim/Simulation/MarkersReference.cpp
klshrinidhi: > Units are handled correctly if we load from file. Can we abstract this conversion and use it at the time the reference value is requested. Then it would not matter how the Table was supplied (directly or by file).

If Units is known during construction of MarkersReference, the quantities can be converted during construction and the TimeSeriesTable can hold this information as metadata. Moving this conversion process to `getValues()` will require the function to know the *current* and *target*.

> Minor, Units is a local helper and does not need to be data member.

I have removed the data-member.

OpenSim/Tools/MarkerPlacer.cpp
klshrinidhi: `staticPose` is actually used later. I had to leave it (unfinished) like this because I had hold off on this until later due to other priorities as we discussed in meetings.

klshrinidhi:
@aymanhab 
I have addressed the comments. You can take a look now.

aymanhab:
Thanks @klshrinidhi Will take a look and get back to you. There are instances of using MarkeData directly in the GUI code as well (e.g. to preview data, associate motion data and in Scaling, IK) that I need to verify have a replacement.

OpenSim/Simulation/MarkersReference.cpp
aseth1: You may be right that the problem already existed, but the `MarkerData` had `Units` as a field and it was responsible for providing markers in SI units when asked for its data. Now that we have replaced it with a `Table` (which also maintains the units via its metadata) we need to make sure the units are handled correctly. It is definitely more flexible to pass the `MarkersReference` a `Table` and it should sensibly convert to internal (SI) units as occurs when loading from file. A straightforward solution would to have the Table-based constructor do the conversion and have the file-based constructor read a Table from file and delegate it to the Table-based constructor. Does that sound reasonable?

We should not dismiss this issue and either address it in this PR or a subsequent one.

OpenSim/Common/TimeSeriesTable.h
aseth1: @chrisdembia what name are you using to provide the state closest to the supplied time? We should strive for consistency. IMHO `near` is imprecise and `nearest` is unequivocal.

chrisdembia: In #774, we use `findNearestBefore()`, `findNearestAfter()`, and `findAt()`. The last of these will only return a state if the given time is within a _very_ small tolerance of one that exists in the trajectory

OpenSim/Tools/MarkerPlacer.cpp
```diff
@@ -242,6 +243,47 @@ bool MarkerPlacer::processModel(Model* aModel,
...
     staticPose->averageFrames(_maxMarkerMovement, _timeRange[0], _timeRange[1]);
     staticPose->convertToUnits(aModel->getLengthUnits());
```
aseth1: Is the sticking point using `TimeSeriesTable_<Vec3>::averageRow()` as a replacement for `MarkerData::averageFrame()`? If this is more complex than anticipated there should be a note or issue filed briefly describing the problem.

Also, the average frame and conversion of units will only apply to the `MarkerData` pose that is never used, since the `staticPoseTable` is used to create the `MarkersReference` used by the Solver. So effectively, these calls have no impact on the algorithm. It seems like we are lucky that the first row in the `staticPoseTable` is sufficiently close to the average frame of the `staticPose` (`MarkerData`) to still pass the tests.

The future PR should have a test that `Table::averageRow()` yields the single frame result of the removed `MarkerData::averageFrame()`.

aymanhab:
Instances of MarkerData, in GUI code base: Need to make sure these can be substituted before removal

            markerData = new MarkerData(fileName);
            markerData.convertToUnits(model.getLengthUnits());
            markerData.makeRdStorage(newStorage);
            markerData.getUnits()
            markerData.getDataRate()

OpenSim/Common/TimeSeriesTable.h
jenhicks: This issue is relatively minor, but I think getNearestRow, while longer, is a little more intuitive.

OpenSim/Simulation/MarkersReference.cpp
jenhicks: I am fine with implementing in a little PR and filing an issue to track.

klshrinidhi: Both constructors now do the unit conversion.

OpenSim/Tools/MarkerPlacer.cpp
klshrinidhi: I am not sure I understand this comment.

The function `TimeSeriesTable::averageRow()` is being tested to make sure it returns the *average* row in the specified time-interval. It is pretty simple function if I understand correctly. The added snippet in `MarkerPlacer::processModel` replicates the operations on `MarkerData`. 
The future PR will replace rest of the use of MarkerData in MarkerPlacer and rest of the code.

jenhicks:
@aseth1 this is ready for another review.
@aymanhab - will address your comment on GUI needs in a subsequent PR if we end up removing MarkerReference in 4.0

OpenSim/Tools/MarkerPlacer.cpp
aseth1: Apologies for the confusing comment.  The main point is that although we still create the `staticPose` `MarkerData*` it is not used for anything, so modifying its units and averaging its rows has no effect as far as I can see. It appears that `staticPose` is used only for its markerNames (L294) but that could be replaced by `staticPoseTable.getColumnLabels()`. 

To fully replicate the previous behavior we should be updating `staticPoseTable` (or creating another table) such that it has a single (average) row. I believed that current `scaleTest` (that employs the `MarkerPlacer` still passed because the first row of the `staticPoseTable` was sufficiently close to the average row so that the answers were within the tolerance. But checking it now, there is no regression test for marker positions from `MarkerPlacer` so the answer could be significantly different- there is no way of knowing atm. I have no issue with the `averageRow` implementation, but with the averaging step is now missing altogether. 

We should verify that static pose markers given to the `MarkersReference` (for the IK Solver) remains the same or otherwise verify the results have not changed.  This could be done by comparing the markers on the processed model (after scaling and marker placer) to a standard model. That is the markers locations match the previous (standard) locations. I am happy to help with that too.

klshrinidhi: > The main point is that although we still create the staticPose MarkerData* it is not used for anything, so modifying its units and averaging its rows has no effect as far as I can see. It appears that staticPose is used only for its markerNames (L294) but that could be replaced by staticPoseTable.getColumnLabels().

I know. Like I said, I did not complete the task.

 > To fully replicate the previous behavior we should be updating staticPoseTable (or creating another table) such that it has a single (average) row. 

It was being done in L247 - L266. Not sure if you noticed.

aseth1: > I know. Like I said, I did not complete the task.

I can understand removing it later if it was doing something, but think it could be removed with no effect.

> It was being done in L247 - L266. Not sure if you noticed.

I don't know how I missed it (I searched for averageRow()) - my apologies this looks good.

klshrinidhi:
`staticPose` is being used at the end of the function `MarkerPlacer::processModel` to create a `Storage` and `StatesReporter` to print files. Not sure if I can remove `staticPose` without replacing this part.

aseth1:
> staticPose is being used at the end of the function MarkerPlacer::processModel to create a Storage and StatesReporter to print files. Not sure if I can remove staticPose without replacing this part.

@klshrinidhi you are absolutely right. Let's leave that for later. Since the main algorithm is still using `MarkerData* staticPose` I don't think it merging the MarkerPlacer test before this will do much good. That test will help when we full use the `staticPoseTable` in `MarkerPlacer::processModel()`.

We should go ahead and merge this. @aymanhab can you take a quick look.

OpenSim/Simulation/MarkersReference.cpp
```diff
@@ -22,139 +22,157 @@
...
 }
 
-/* load the marker data for this MarkersReference from markerFile  */
-void MarkersReference::loadMarkersFile(const std::string markerFile, Units modelUnits)
-{
-    upd_marker_file() = markerFile;
-    _markerData.reset(new MarkerData(get_marker_file()));
+void MarkersReference::loadMarkersFile(const std::string markerFile,
+                                       Units modelUnits) {
+    auto fileExt = FileAdapter::findExtension(markerFile);
```
aymanhab: Are these case sensitive? We had issues in the past with users providing .TRC, .MOT files, does the function return lowercase regardless? I can look it myself but it's hard to navigate. The test is for ".trc" and the message says "TRC".

OpenSim/Simulation/MarkersReference.cpp
```diff
@@ -22,139 +22,157 @@
...
+    double scaleFactor = thisUnits.convertTo(Units{units});
 
-/* A convenience method to populate MarkersReference from MarkerData **/
-void MarkersReference::populateFromMarkerData(const MarkerData& aMarkerData)
-{
-    const Array<std::string> &tempNames = aMarkerData.getMarkerNames();
-    int nm = tempNames.getSize();
+    OPENSIM_THROW_IF(SimTK::isNaN(scaleFactor),
+                     Exception,
+                     "Model has unspecified units.");
```
aymanhab: Do you mean data/table has no units? Models have builtin units by default.

OpenSim/Simulation/MarkersReference.cpp
```diff
@@ -22,139 +22,157 @@
...
+    for(unsigned i = 0; i < rowView.ncol(); ++i)
+        values.push_back(rowView[i]);
 }
 
-/** get the speed value of the MarkersReference */
-void MarkersReference::getSpeedValues(const SimTK::State &s, SimTK::Array_<Vec3> &speedValues) const
-{
+void
+MarkersReference::getSpeedValues(const SimTK::State &s,
+                                 SimTK::Array_<Vec3> &speedValues) const {
```
aymanhab: Is there a reason to keep the unimplemented method here? A user looking at documentation will get the false assumption that it's implemented only to be surprised at runtime. I'd prefer to remove it unless there's work in progress to implement it (short term) or this class implements an interface that requires the method and there are other instances where  implementation actually exist. I agree it will be great when implemented but don't see benefit to users.

OpenSim/Tools/MarkerPlacer.cpp
aymanhab: MarkerPlacer is used in a few places in the ScaleTool GUI I'll assume the old interface still works or that there's a straightforward mapping to new methods, if not the case please open an issue with removed methods and I'll try to build GUI.

aymanhab:
Done with review, minor comments above, looks good otherwise.

OpenSim/Simulation/MarkersReference.cpp
aseth1: This is a good point @aymanhab and we can discuss the virtues of leaving base implementation pure virtual or to have a default implementation to avoid overriding here. @klshrinidhi did not introduce the implemented method in this PR (just made it fit within 80c) so I suggest either filing an issue or accepting the current scheme to keep the pure virtual method on the base `Reference`.

OpenSim/Simulation/MarkersReference.cpp
klshrinidhi: The unit conversion code is `DataTable` analogue of `MarkerData::convertToUnits` which throws a similar exception. So I copied the same here.
I have changed the message to say `TimeSeriesTable has unspecified units.`.

aymanhab:
LGTM, please merge when tests pass.

klshrinidhi:
CI passed except for timeout on Travis.
Merging.

PR #1373, opened 2016-11-02
Enclose "using" syntax in C++ headers within #ifndef SWIG to reduce warnings

aymanhab:
Using the "using" syntax to indicate inherited constructors is not recognized by SWIG and leads to excessive warnings. This PR encloses these in ifndef SWIG to reduce the clutter. We could use a pragma instead but these tend to be global and so may suppress other useful warnings.


aymanhab:
@klshrinidhi and/or @chrisdembia to review. Thanks.

aymanhab:
Thanks @klshrinidhi

PR #1374, opened 2016-11-03
Display hints improvements

aymanhab:
This PR adds a display_scale_factor to ModelDisplayHints object, allowing resizing of ContactHalfSpace and Markers to work for arbitrary size models.

### DIFF
--- a/OpenSim/Common/ModelDisplayHints.h
+++ b/OpenSim/Common/ModelDisplayHints.h
@@ -43,16 +43,16 @@ If you are writing a ModelComponent that generates some of its own geometry,
 and you would like people to consider it well-behaved, you should check whether
 any of the flags here might reasonably be expected to affect the geometry that
 your component produces. The currently-supported flags are:
-  - show wrap geometry
-  - show contact geometry
-  - show muscle paths (should apply to other path objects too)
-  - show path points
-  - show markers
-  - show forces
-  - show frames
-  - show labels
-  - show debug geometry
-  - display_scale_factor
+  - show_wrap_geometry
+  - show_contact_geometry
+  - show_muscle_paths (should apply to other path objects too)
+  - show_path_points
+  - show_markers
+  - show_forces
+  - show_frames
+  - show_labels
+  - show_debug_geometry
+  - decorations_scale_factor
 
 This class is intended to provide some minimal user control over generated
 geometry in a form that is easy for a ModelComponent author to deal with, since
@@ -102,8 +102,8 @@ public:
     OpenSim_DECLARE_PROPERTY(show_debug_geometry, bool,
         "Flag to indicate whether or not to show debug geometry, default to false.");
 
-    OpenSim_DECLARE_PROPERTY(display_scale_factor, double,
-        "Scale factor to apply to decorations (that have no inherent size) but are model size dependent (e.g. Marker, halfplane size defaults to 1.");
+    OpenSim_DECLARE_PROPERTY(decorations_scale_factor, double,
+        "Scale factor to apply to decorations that have no inherent size but should scale with the overall size of the model (e.g. Marker, ContactHalfSpace); defaults to 1.");
 
     /** Default construction creates a valid display hints object with all
     hints set to their default values. **/
@@ -123,7 +123,7 @@ private:
         constructProperty_show_labels(false);
         constructProperty_show_forces(true);
         constructProperty_show_debug_geometry(false);
-        constructProperty_display_scale_factor(1.0);
+        constructProperty_decorations_scale_factor(1.0);
     }
 };
 
--- a/OpenSim/Common/ModelDisplayHints.h
+++ b/OpenSim/Common/ModelDisplayHints.h
@@ -27,7 +27,6 @@
 #include <OpenSim/Common/osimCommonDLL.h>
 #include <OpenSim/Common/Object.h>
 
-
 namespace OpenSim {
 
 //==============================================================================
@@ -53,6 +52,7 @@ your component produces. The currently-supported flags are:
   - show frames
   - show labels
   - show debug geometry
+  - display_scale_factor
 
 This class is intended to provide some minimal user control over generated
 geometry in a form that is easy for a ModelComponent author to deal with, since
@@ -102,6 +102,9 @@ public:
     OpenSim_DECLARE_PROPERTY(show_debug_geometry, bool,
         "Flag to indicate whether or not to show debug geometry, default to false.");
 
+    OpenSim_DECLARE_PROPERTY(display_scale_factor, double,
+        "Scale factor to apply to decorations (that have no inherent size) but are model size dependent (e.g. Marker, halfplane size defaults to 1.");
+
     /** Default construction creates a valid display hints object with all
     hints set to their default values. **/
     ModelDisplayHints() { constructProperties(); }
@@ -120,6 +123,7 @@ private:
         constructProperty_show_labels(false);
         constructProperty_show_forces(true);
         constructProperty_show_debug_geometry(false);
+        constructProperty_display_scale_factor(1.0);
     }
 };
 
--- a/OpenSim/Simulation/Model/Marker.cpp
+++ b/OpenSim/Simulation/Model/Marker.cpp
@@ -203,7 +203,7 @@ void Marker::generateDecorations(bool fixed, const ModelDisplayHints& hints, con
     //const Vec3& p_BM = bTrans*get_location();
     appendToThis.push_back(
         SimTK::DecorativeSphere(.005).setBodyId(frame.getMobilizedBodyIndex())
-        .setColor(color).setOpacity(1.0)
+        .setColor(color).setOpacity(1.0).setScale(hints.get_display_scale_factor())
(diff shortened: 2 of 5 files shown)


aymanhab:
@chrisdembia This should address the issue you brought up with PR #1314 regarding the size of halfSpace but generalizes to other "non-physical" geometry.

aymanhab:
@chrisdembia Please let me know what you think when you have a chance. Trying to use the same mechanism for FrameGeometry doesn't work, since it is serialized (rather than use generateDecorations) for finer control over visibility, it will have to use the scale factors if needed.

OpenSim/Common/ModelDisplayHints.h
```diff
@@ -102,6 +102,9 @@ class OSIMCOMMON_API ModelDisplayHints : public Object {
     OpenSim_DECLARE_PROPERTY(show_debug_geometry, bool,
         "Flag to indicate whether or not to show debug geometry, default to false.");
 
+    OpenSim_DECLARE_PROPERTY(display_scale_factor, double,
+        "Scale factor to apply to decorations (that have no inherent size) but are model size dependent (e.g. Marker, halfplane size defaults to 1.");
```
chrisdembia: Consider this revision:

```
"Scale factor to apply to decorations that have no inherent size but that should scale with the overall size of the model (e.g., Marker, ContactHalfSpace); defaults to 1."
```

chrisdembia: If you intended to convey that the decoration for the ContactHalfSpace has length/width 1.0 by default, then I think this information is best conveyed in the documentation for ContactHalfSpace.

chrisdembia: Consider renaming the property to `decorations_scale_factor`.

OpenSim/Common/ModelDisplayHints.h
```diff
@@ -53,6 +52,7 @@ your component produces. The currently-supported flags are:
   - show frames
   - show labels
   - show debug geometry
+  - display_scale_factor
```
chrisdembia: The other items in this list do not have underscores. I think the list should either all use the exact property names, or all use the more casual phrase.

chrisdembia:
> Would it not be better for FrameGeometry to be a member-subcomponent, so that it does not get serialized (like Thelen2003Muscle's activation and fiber dynamics subcomponents).

To clarify, I think it is okay to merge this PR without handling FrameGeometry, but I am curious about your response.

I am fine with merging this PR after resolving the following two items:
- hearing your opinion about the name and comment for the new property
- understanding how this is tested (the new model file?).

OpenSim/Common/ModelDisplayHints.h
aymanhab: Sure, sounds more appropriate. :+1:

OpenSim/Common/ModelDisplayHints.h
aymanhab: The flags actually have "_" so I'd lean to add those rather than remove the ones in display_scale_factor. I sure will take your wording.

aymanhab:
The model introduced here is not for this PR but for testing WrapObjects when they're fully implemented (including quadrants) it is harmless but unrelated. I can take it out for now. The testing was done visually pending review but I can easily beef up the test case to account for it.

For FrameGeometry, the issue is granularity. If we use the Frame::generateDecorations mechanism then FrameGeometry doesn't even have to be a Component and we can use SimTK::DecorativeFrame directly but we don't have fine control over visibility or size, on the other extreme making it a Component allows for both but exposes extra things in XML and doesn't allow "global" customization". Not sure what would be possible with an internal Component. Will experiment and submit a different PR if it looks promising. Thank s @chrisdembia

chrisdembia:
> The model introduced here is not for this PR but for testing WrapObjects when they're fully implemented (including quadrants) it is harmless but unrelated. I can take it out for now. 

Thanks for explaining the reason for the model file. I think it's fine to leave the model file in this PR. 

>The testing was done visually pending review but I can easily beef up the test case to account for it.

If easy, it'd be nice to have a simple test for this. I know it's a simple implementation, so the test is more just to catch regressions.

> For FrameGeometry, the issue is granularity. If we use the Frame::generateDecorations mechanism then FrameGeometry doesn't even have to be a Component and we can use SimTK::DecorativeFrame directly but we don't have fine control over visibility or size, on the other extreme making it a Component allows for both but exposes extra things in XML and doesn't allow "global" customization". Not sure what would be possible with an internal Component. Will experiment and submit a different PR if it looks promising. Thank s @chrisdembia

Thanks for explaining the tradeoffs. I don't have a strong preference. An internal Component likely has the exact same tradeoffs as the `Frame::generateDecorations()` solution: selectively exposing the customization, while also allowing for global customization.

aymanhab:
Changes done and test case updated. Thanks @chrisdembia

aymanhab:
Need another reviewer, @tkuchida @klshrinidhi or @aseth1 This is a small localized change to enable marker size and contactHalfSpace display to scale based on model size, with test case updated accordingly.

OpenSim/Tools/Test/test_wrapAllVis.osim
```diff
@@ -0,0 +1,745 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
aseth1: Where is model being used? I could not find it in the visualization tests. By its name it looks like it is to be a test for visualizing wrapping, which maybe slipped in from a future PR?

OpenSim/Common/ModelDisplayHints.h
```diff
@@ -102,6 +102,9 @@ class OSIMCOMMON_API ModelDisplayHints : public Object {
     OpenSim_DECLARE_PROPERTY(show_debug_geometry, bool,
         "Flag to indicate whether or not to show debug geometry, default to false.");
 
+    OpenSim_DECLARE_PROPERTY(decorations_scale_factor, double,
+        "Scale factor to apply to decorations that have no inherent size but should scale with the overall size of the model (e.g. Marker, ContactHalfSpace); defaults to 1.");
```
aymanhab: Thanks for the feedback @aseth1 As indicated at the top of the PR, this is intended to handle models of arbitrary size. Mesh files have built-in/assumed units, same with any real/physical geometry that assumes model length units, some pieces of Geometry however doesn't have any physical meaning (e.g. a Marker)  even if it had a radius you'll have to place it in the xml file. Having multiple numbers that you need to change (one for Markers, another for muscles, a third for Frame etc. doesn't seem to "scale").
There's no infinite plane or plane primitive that's why I used a brick, the API visualizer uses an opengl program to draw, and the Three.js uses a "finite" size plane but these used for ground do not generalize if you have multiple ContactHalfSpaces for example.

OpenSim/Simulation/Model/Marker.cpp
```diff
@@ -203,7 +203,7 @@ void Marker::generateDecorations(bool fixed, const ModelDisplayHints& hints, con
     //const Vec3& p_BM = bTrans*get_location();
     appendToThis.push_back(
         SimTK::DecorativeSphere(.005).setBodyId(frame.getMobilizedBodyIndex())
```
aymanhab: The tradeoff is either put Sphere(Geometry) as a Component under Marker then it's more flexible, you can change size, or even shape but is verbose and heavyweight that needs connections/attachments to maintain or use generateDecorations route with implicit Geometry which we opted to a while ago to reduce XML impact and redundancy. My preference is to use a single knob (rather than one per decoration type) and allow limited customization.

OpenSim/Tools/Test/test_wrapAllVis.osim
aymanhab: The model is not used in this PR, and planned for WrapObject visualization test, it felt harmless at the time but I can remove it if it is confusing.

OpenSim/Common/ModelDisplayHints.h
aseth1: > even if it had a radius you'll have to place it in the xml file. Having multiple numbers that you need to change (one for Markers, another for muscles, a third for Frame etc. doesn't seem to "scale"

More frustrating in my view, than setting the size of markers and Frames individually (which the GUI could help with multi-select edits and other utilities), is the magical scaling of multiple potentially unrelated objects by a mysterious common factor "decorations_scale_factor". I could understand one knob to change ALL decorations (which is what this factor does not do) but to arbitrarily change some odd decorations seems bound to cause confusion.  I could imaging sizing a HalfSpace decoration to appear as a wall or goal, which could be huge, and not also wanting markers on my player's leg to quadruple in size. I just don't see how these geometry should be related. I think decorations should be sized (not scaled) and then a single scale could be applied to collection of different scales :mouse: vs. 🐘

aseth1:
How about separate global fags for marker size and contact half plane

aymanhab:
Thanks for the feedback @aseth1 Since this isn't a pressing issue now and the point of using one global flag was to avoid having separate global flags for multiple decorations (e.g. for Markers, then Frames, PathPoints, HalfSpaces, ...) I'll close this PR for now until other more pressing issues are resolved.

PR #1376, opened 2016-11-04
Replace or Avoid wrapping functions with double[] in the interface

klshrinidhi:
Addressing #1368.

@aymanhab


klshrinidhi:
> I'd prefer we stay with SimTK::Vector rather than std::vector to avoid introducing a new container in the API. If the SimTK::Vector and std::Vector are very close then we can swap or typedef in one shot but for now I'd stay with SimTK::Vector Vec3, Vec2

I can change to `SimTK::Vector`. Using a `typedef` to do that will cause confusion and problems in debugging for users/developers. I can replace `std::vector` with `SimTK::Vector`.

> The initialization with {} is not cross language construct, ideally we use something that can be invoked from both python and Matlab/Java, especially that some of these methods were introduced specifically for GUI access from Java.

I can make the functions have default arguments so that an empty `{}` is not required. 
More widely, I don't think our goal should be to force equivalence between 3 different languages which are different by definition. The similarities do not go too far. Lack of documentation for scripting users is better addressed directly by writing documentation rathan than expecting users to translate C++ code to Python/Java/Matlab code mentally using C++ Doxygen documentation.

aymanhab:
Thanks @klshrinidhi Great that we're in agreement on changing to SimTK::Vector, Vec3, Vec2. Will need to build the GUI against this version of the API to make sure nothing fell through the cracks or broke in transition. Thank you 👍

klshrinidhi:
Sorry. I prefer `std::vector` over `SimTK::Vector`, even if it introduces new type into interface. I am just addressing your comment to get this PR merged. 

I think I made all the changes you asked for -- 
* Replace `std::vector` with `SimTK::Vector`.
* Use `()` instead of `{}`.

OpenSim/Common/Storage.cpp
```diff
@@ -2623,7 +2615,7 @@ void Storage::interpolateAt(const Array<double> &targetTimes)
         StateVector vec;
         // INTERPOLATE THE STATES
         ny = getDataAtTime(t,ny,&y);
-        vec.setStates(t,ny,y);
+        vec.setStates(t, SimTK::Vector_<double>(ny, y));
```
aymanhab: Can you update the documentation so it doesn't refer to removed parameters? Thanks.

OpenSim/Common/Scale.h
```diff
@@ -96,15 +96,17 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Scale, Object);
 
     void getScaleFactors(SimTK::Vec3& aScaleFactors) const;
     SimTK::Vec3& getScaleFactors() { return _scaleFactors; }
+#ifndef SWIG    
     void getScaleFactors(double rScaleFactors[]){   // A variant that uses basic types for use by GUI
         getScaleFactors(SimTK::Vec3::updAs(rScaleFactors));
     }
-
+#endif
```
aymanhab: This method was introduced for GUI use, I'd remove it rather than idndef SWIG it.

OpenSim/Common/Scale.h
```diff
@@ -96,15 +96,17 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Scale, Object);
...
     void setScaleFactors(const SimTK::Vec3& aScaleFactors);
+#ifndef SWIG
     void setScaleFactors(const double aScaleFactors[]){ // A variant that uses basic types for use by GUI
         setScaleFactors(SimTK::Vec3::getAs(aScaleFactors));
     }
-
+#endif
```
aymanhab: Same comment. This method was likely introduced for GUI use, I'd remove it rather than idndef SWIG it.

OpenSim/Common/Storage.cpp
klshrinidhi: Done.

OpenSim/Simulation/SimbodyEngine/TransformAxis.h
```diff
@@ -117,7 +117,9 @@ OpenSim_DECLARE_CONCRETE_OBJECT(TransformAxis, Object);
     void getAxis(SimTK::Vec3& axis) const {axis = getAxis();}
     /** Alternate signature that writes the axis value to its argument as
     an ordinary C array. **/
+#ifndef SWIG
     void getAxis(double rAxis[]) const {SimTK::Vec3::updAs(rAxis)= getAxis();}
+#endif
```
klshrinidhi: Done.

OpenSim/Common/Scale.h
klshrinidhi: Done.

aymanhab:
Thanks @klshrinidhi LGTM :+1:

aymanhab:
@chrisdembia Can you take a look at the Python bindings to make sure the changes caused the desired effect? Thanks.

OpenSim/Tools/InverseDynamicsTool.cpp
```diff
@@ -367,13 +367,18 @@ bool InverseDynamicsTool::run()
         SpatialVec equivalentBodyForceAtJoint;
 
         for(int i=0; i<nt; i++){
-            StateVector genForceVec(times[i], nq, &((genForceTraj[i])[0]));
+            StateVector
+                genForceVec(times[i],
+                            SimTK::Vector_<double>(nq,
+                                                   &((genForceTraj[i])[0])));
```
chrisdembia: `genForceTraj[i]` is already a SimTK::Vector; I think you can pass it directly to the `StateVector` constructor.

klshrinidhi: `genForceTraj` is a `SimTK::Array_<Vector>`. It is defined [here](https://github.com/opensim-org/opensim-core/blob/a245657272f83a4976cab9662e1316fb91bed282/OpenSim/Tools/InverseDynamicsTool.cpp#L330).

klshrinidhi: Got it. You are right. Its done.

chrisdembia:
> @chrisdembia Can you take a look at the Python bindings to make sure the changes caused the desired effect?

I tested this locally and the StateVector methods seem to work as expected, using `opensim.Vector`. I reviewed the code and it looks good to me. 

@klshrinidhi has already addressed my one comment.

LGTM.

aymanhab:
Thanks @klshrinidhi and @chrisdembia will merge after the tests pass.

aymanhab:
Failure seems unrelated. Merging.

Issue #1111, opened 2016-07-01
[visualizer] Visualizer crashes if the model contains disabled muscles


tkuchida:

tkuchida:
Presumably, `GeometryPath::generateDecorations()` should return immediately if the muscle is disabled.

aseth1:
I would like us to work out a mechanism so that any Component can be "disabled". In particular, "disabled" would mean it is not part of the Component tree so *none* of the Component interface methods are invoked on the disabled Component. You would have to "enable" the Component and call `finalizeFromProperties()` to have it in included again. See also  #845.

I think this would be better than surgically handling what "disable" does on different Components.

@tkuchida @chrisdembia @aymanhab what do you think?

chrisdembia:
I like that idea a lot. This is more possible now that the "subcomponents" data structure is not the properties itself, but something filled up from the properties. So it should be fairly easy to handle. 

Should still probably fix this bug in the meantime.

tkuchida:
> "disabled" would mean it is not part of the Component tree so *none* of the Component interface methods are invoked on the disabled Component.

I like that idea. It sounds like that would allow you to disable a Component that isn't working and it would be entirely ignored (i.e., behavior would be identical if it weren't added to the Model at all), which could be useful for debugging.

> Should still probably fix this bug in the meantime.

This bug likely affects lots of Components, so it might be best to just tackle the problem at the Component level in one fell swoop.

aymanhab:
I completely agree as well. Deciding what that means for various components is a bit harder, but that's a topic for discussion.

aymanhab:
@aseth1 This will be important to solve in 4.0 The main issue is that the GeometryPath doesn't know that the Muscle is disabled. Any thoughts?

aseth1:
Possible solutions:

1. Implement DevProp #1382 so that any Component can be temporarily excluded without crashing.
2. Figure out which calls to `GeometryPath` *blow-up* and band-aid the problem by checking the *appliesForce* flag first. With the rename to `appliesForce`, there should be no expectations that the Muscle should not draw itself - it simply should not apply forces - and we could still have the muscle draw itself.

aymanhab:
Thanks @aseth1 I think we greed already that DevProp is outside the scope of 4.0 as it introduces new concepts. For solution 2, the crash happens while trying to get the CacheEntry  "current_path". Incidently the CacheEntry does exist but retrieving the value throws an exception. This can be made more consistent I think.

[...]

PR #1382, opened 2016-11-08
[DevProp] Exclude any Component from underlying System

aseth1:
Enclosed is a proposal for a new feature of the Component to enable it to be effectively excluded from the Model and its resulting System. I do not see the proposed changes as a replacement for disabling Forces and Constraints. Instead it proposes a new feature to correctly and unambiguously exclude the Component (and its subcomponents). I believe users were using *disable* because they had not other means of removing a Component temporarily other than editing XML directly. The intent of this proposal is to address that use case.

Any and all feedback is appreciated. Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/1382)

### DIFF
--- a/DevelopmentProposals/excludeComponentFromSystem/excludeComponentFromSystem.md
+++ b/DevelopmentProposals/excludeComponentFromSystem/excludeComponentFromSystem.md
@@ -1,4 +1,4 @@
-## Proposal: Exclude Components from the underlying Computational System
+## Proposal: Add a Feature to Exclude Components from the underlying Computational System
 ***OpenSim Development Proposal***
 
 ### The Need:
--- a/DevelopmentProposals/excludeComponentFromSystem/excludeComponentFromSystem.md
+++ b/DevelopmentProposals/excludeComponentFromSystem/excludeComponentFromSystem.md
@@ -1,13 +1,15 @@
 ## Proposal: Exclude Components from the underlying Computational System
 ***OpenSim Development Proposal***
 
+### The Need:
+
+Models are complicated and the Component paradigm is enabling the construction of ever more complex models and behaviours. Modeller have to troubleshoot and users must test models, which often requires isolating a problematic component. The easiest way to isolate a problematic Component's effect is to effectively remove it from the Model and see if the problem is alleviated. The current method is to *disable* the Component, which is applicable to a handful of Components (Constraints and Forces). Unfortunately, disabling does not equate to effectively removing the Component and the effect of disabling is inconsistent across the subset of Components.   
+
+The purpose of this proposal is to enable the fundamental ability to effectively remove a Component so one can accurately assess its impact on computation (timing, CPU usage) and results. Currently, users have little confidence that disabling a Component (if it has that option) effectively removes its calculations and it is often recommended to remove Components by hand from XML files.  
+
+
 ### Overview
-This proposal describes how OpenSim Component's can be excluded from computations and thereby replace 
-the adhoc mechanisms in place that attempt to "disable" certain components, like Forces and Constraints. 
-The adhoc nature of the current mechanism puts the burden of "disabling" a Component on the Component
-writer and results in inconsistent behaviours, including partially disabling by not applying the force of a Muscle, for example, to the the MultibodySystem, but still computing the force or other costly values such as its length.
-The overall desired behaviour is that a Component that is disabled (excluded) is excluded from any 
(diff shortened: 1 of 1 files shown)


chrisdembia:
This will be a good feature. The proposed test cases sound great.

1. Consider the following code:

```cpp
Model m;
Component& foo = m.updComponent("foo");
foo.set_is_excluded(true);
m.finalizeFromProperties();
foo.set_is_excluded(false);
m.finalizeFromProperties();
Component& fooAgain = m.updComponent("foo");
```

Would that throw an exception? That is, will a pointer to a component be the same across calls to exclude/include it? Or would copies, etc. occur? Consider using the test case to ensure such code will work (without re-fetching the excluded subcomponent from the model using `getExcludedComponentList()`, etc.).

2. What do you think about adding methods like `(get|upd)ExcludedComponent()`, so that it is not necessary to use the ComponentList?

3. How should the excluded components appear in the GUI? Should the GUI present components as they exist in XML? Or should the GUI present the excluded components as a separate tree? And if so, how would such a tree be organized, considering that the excluded components could be anywhere in the tree? Would it be a flat list with full paths? I imagine it is preferred that the GUI shows only 1 tree of all components. If so, then it is probably necessary for the API to give access to a list of *all* components, included excluded components. That is, a `getIncludedAndExcludedComponents()`.

4. What is the word we will use to describe non-excluded components? That is, if I ask you, "did you exclude the muscle component?" would you respond with "No, the muscle component is included."? Or "No, the muscle component is present." or "No, the muscle component is in-use."? Ideally, we would _not_ say "No, the muscle component is not excluded."

5. Related to the point above, I think we should at least explore terms that are "positive." For example, using "is_included" instead of "is_excluded" for the property. I currently do prefer "is_excluded" over "is_included", but we should think about this more. Double negatives are super confusing.

6. I definitely think there is a strong use-case for the full excluding of subcomponents, but I also do not think it can be a full replacement for "is_disabled" (which I think is your stance as well, as you mentioned IAA). So I think we should expect the two sorts of flags to co-exist, and we should have clear documentation explaining the difference.

7. My main concern with the `is_excluded` flag is that the resulting `SimTK::System` and `SimTK::State` are different. You will not be able to use the same State with two copies of the same model if one of them has a different value for one of the `is_excluded` flags. For example, imagine doing a simulation with two nearly identical models: they both have a `ClutchedPathSpring`, but model B has excluded the ClutchedPathSpring. Say you want to use the same `SimTK::State` for the initial state of the simulation of both models. This will not be possible, because the two models will have a different number of state variables. The solution for this specific issue is to use the `is_disabled` flag instead of `is_excluded`, but maybe we should think about this general problem a little.

8. I think our convention is to use `is_` or `has_` before boolean properties (or at least for member functions). If we do have this convention, then I prefer `is_excluded` over `excluded` for the property name.

aymanhab:
Thanks @aseth1 for the nice proposal. Few thoughts come to mind:

1. While the Component/Tree layout gives the connotation that all Components are equal, the truth is that they aren't. For example, what does it mean to exclude a Coordinate of a Joint or a MuscleForceLength Component from a single Muscle? If there's no clear intuitive answer that users can assume/understand without reading oodles of documentation then these Components shouldn't be allowed to be "excluded".
2. We should consider practical use cases for users creating or assembling models, and using "exclusion" to troubleshoot.  We should discover what "actual" functionality or troubleshooting helpers that we can provide (initially at API and eventually the application) before we introduce yet another new concept.
3. The co-presence of disabled and excluded will likely confuse users as you pointed out.
4. What happens to "Connections, inputs, outputs" when a component is excluded. If the wiring is lost then very likely users will not do it, if not then all the connection/input/output wiring will need to handle the new excluded flag.
5. A possible useful and less drastic option is to better implement the disabled flag and make it available to few more Component types that make sense (e.g. Forces, Constraints, Controllers, Reporters,..) and make sure disabled means do nothing (including no-side-effects). These probably can be easily implemented at a few base classes but you know that code better to make this call.
6. From the GUI side, we currently assume fixed topology of the model once loaded, only few places where you can actually add new Components. Allowing "exclusion" means that Coordinate Window needs to be recreated repeatedly, same for plotter window (that shows Coordinates, Muscles), Navigator tree etc. Not impossible but is it justified?

tkuchida:
Related to #1111

nickbianco:
Very stale PR, closing.

Issue #1369, opened 2016-10-31
location/velocity/acceleration cache isn't being invalidated when integrating with the manager


carmichaelong:
Looked at this with @aseth1. Simple way to reproduce:

In `testPoints`, there is a test that uses a `TimeStepper` and compares the location, velocity, and acceleration of the center of mass of the pendulum between the `Station` method and `MobilizedBody` methods directly. This test passes as is, but does not pass if the integration is done by a `Manager` instead. Location, velocity, and acceleration match for the first (initial) time, but then all three fail to match on the immediate next time step.

jenhicks:
@carmichaelong Thanks for reporting. Sounds like something we should certainly fix for 4.0. Is it affecting your work? If so, we will prioritize.

aseth1:
I do not believe this is a bug in the Frame or Point classes and if we use the `TimeStepper` directly and only initialize(State) once, these lazy cache entries work as expected.

Looking a bit deeper, the `TimeStepper::initialize(State)` calls the Integrator's `initialize`, which makes a copy of the state passed in for its internal use, but unlike typical copies, it seems that the default system cache variables are not invalidated. The cache variables seem to be untouched during `Manager::doIntegration` (which invokes the `TimeStepper`) and it updates the passed in State reference to now refer to the integrator's internal copy (with now stale cache). 

If we explicitly make a copy of the internal state (instead of a getting a reference) we do get the cache to be invalidated and subsequent calls to the Frame and Point methods are forced to to compute their transform and location.

Unlike `GeometryPath`'s path and other lazily cached variables we are using (like StateVariable derivatives), the transform and location cache's of `Frame` and `Point` are unused during the evaluation of derivatives and the process of integration (there is no reason for the lazy cache variables to be recomputed), and so these cache entries appear to be untouched - not even invalidated.

@sherm1 do you have any comments/concerns/advice regarding this issue? Is it valid to create a new TimeStepper with the same Integrator repeatedly?  Thanks!

aseth1:
@jenhicks I agree this is a bug, but it isn't completely clear if it is because of a misuse of the TimeStepper (e.g. calling `ts.initialize(s)` more than once) or if there is a bug within `TimeStepper` or `Integrator` that should be invalidating the cache variables and it isn't. I hope @sherm1 can chime in.

sherm1:
[...]

PR #1386, opened 2016-11-09
Fix manager / cache bug

carmichaelong:
Aimed at resolving #1369. Currently does not pass testOptimizationExample. @aseth1

### DIFF
--- a/OpenSim/Examples/OptimizationExample_Arm26/OptimizationExample.cpp
+++ b/OpenSim/Examples/OptimizationExample_Arm26/OptimizationExample.cpp
@@ -58,7 +58,6 @@ class ExampleOptimizationSystem : public OptimizerSystem {
            // Create the integrator for the simulation.
            p_integrator = new RungeKuttaMersonIntegrator(osimModel.getMultibodySystem());
            p_integrator->setAccuracy(1.0e-7);
-           p_manager.reset(new Manager(osimModel, *p_integrator));
        }
                 
     int objectiveFunc(  const Vector &newControls, bool new_coefficients, Real& f ) const override {
@@ -70,12 +69,13 @@ class ExampleOptimizationSystem : public OptimizerSystem {
         osimModel.updDefaultControls() = newControls;
                 
         // Integrate from initial time to final time
-        p_manager->setInitialTime(initialTime);
-        p_manager->setFinalTime(finalTime);
+        Manager manager(osimModel, *p_integrator);
+        manager.setInitialTime(initialTime);
+        manager.setFinalTime(finalTime);
 
         osimModel.getMultibodySystem().realize(s, Stage::Acceleration);
 
-        p_manager->integrate(s);
+        manager.integrate(s);
 
         /* Calculate the scalar quantity we want to minimize or maximize. 
         *  In this case, we’re maximizing forward velocity of the 
@@ -107,7 +107,6 @@ private:
     int numControls;
     State& si;
     Model& osimModel;
-    std::unique_ptr<Manager> p_manager;
     SimTK::ReferencePtr<RungeKuttaMersonIntegrator> p_integrator;
 
  };
--- a/OpenSim/Simulation/Manager/Manager.cpp
+++ b/OpenSim/Simulation/Manager/Manager.cpp
@@ -506,7 +506,7 @@ setModel(Model& aModel)
     }
 
     if (_timeStepper) {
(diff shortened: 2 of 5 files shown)


OpenSim/Examples/OptimizationExample_Arm26/OptimizationExample.cpp
```diff
@@ -70,12 +69,13 @@ class ExampleOptimizationSystem : public OptimizerSystem {
         osimModel.updDefaultControls() = newControls;
                 
         // Integrate from initial time to final time
-        p_manager->setInitialTime(initialTime);
-        p_manager->setFinalTime(finalTime);
+        Manager manager(osimModel, *p_integrator);
```
carmichaelong: I think it was more specifically reuse of the `TimeStepper`. Before, a `TimeStepper` was created and initialized each time in the `Manager::integrate(s)` function. But now, you pass the initial state into the `Manager`, and now the `TimeStepper` isn't aware of it (since a new one isn't created).

OpenSim/Simulation/Manager/Manager.cpp
```diff
@@ -535,6 +542,11 @@ void Manager::
 setIntegrator(SimTK::Integrator& integrator) 
 {   
     if (_integ.get() == &integrator) return;
+    if (_timeStepper) {
```
aseth1: Not sure if  throwing here is the ideal behavior. Instead if you reset `_timeStepper` then `integrate` should throw that it has no `TimeStepper` and to call `initializeTimeStepper` first, or you could `setIntegrator()` do that for you after the reset.

OpenSim/Simulation/Manager/Manager.cpp
```diff
@@ -870,6 +882,18 @@ void Manager::initialize(SimTK::State& s, double dt )
 }
 //_____________________________________________________________________________
 /**
+* set and initialize a SimTK::TimeStepper
+*/
+void Manager::initializeTimeStepper(
+    const SimTK::System& sys, const SimTK::State& state)
```
carmichaelong: From my understanding, this is because of the inconsistency due to using the `Manager` for CMC as well (see [this line](https://github.com/opensim-org/opensim-core/pull/1386/files#diff-d43892ebaf9a963f51ea372ed57fcb78R746)).

aseth1: Yeah, I don't like using this inconsistency to slip in a CMCActuatorSystem. 
Where it is being used is in:
```cpp
void VectorFunctionForActuators::
evaluate( const SimTK::State& s, double *aX, double *rF)
```
And in this case I believe the `_model` is completely unnecessary. This may be easily replaced with using a TimeStepper directly. Here to the Manager is being created and destroyed over and over for every `dt` interval during CMC. Funny thing is a Manager is used but analyses and reporting have to be turned off since these live a model whose associate with the Manager but now being integrated. It is quite horrible.

carmichaelong: I totally agree. It does look like a search for `setSystem()` shows that this is the only use case. The question is whether or not this change belongs in this PR. I'm happy to add it on since (in theory) it should be easy to remove that functionality from `Manager`.

aseth1: I agree that the inconsistency is more general, but for the first time I think we have a handle of why that inconsistency was baked in (to handle integrating a CMCActuatorSystem). Correcting it has an impact on the interface being put in place here. I am happy for that to be a follow-up PR, but it could be done here especially `initializeTimeStepper` and the error handling that are being added here and are awkwardly working around the inconsistency. @carmichaelong it is up to you. As long as the we fix the inconsistency soon, either way is fine with me.

carmichaelong:
Updated to address review from @aseth1 and @chrisdembia and is ready to be looked at again. This PR is limited to fixing the problem with the cache with successive `integrate()` calls to a single `Manager`. Tangential issues that came up with this PR have been documented in #1405.

OpenSim/Simulation/Test/testManager.cpp
```diff
@@ -0,0 +1,139 @@
+/* -------------------------------------------------------------------------- *
+ *                          OpenSim:  testPoints.cpp                          *
```
aseth1: testPoints --> testManager

aseth1:
@chrisdembia do you have additional comments/concerns?

PR #1391, opened 2016-11-10
Bindings physicaloffsetframes

aymanhab:
Make custom constructors for PhysicalOffsetFrame  available to scripting (to work around SWIG lack of support for inherited constructors)


chrisdembia:
@aymanhab this looks pretty good. A few comments:

1. Since the fix is exactly the same for both languages, it could potentially go in `Bindings/simulation.i`. Though if that doesn't work for some reason, the duplication is fine with me.
2. Could you add a test for this? The simplest test would be to update the [build_simple_arm_model.py](https://github.com/opensim-org/opensim-core/blob/master/Bindings/Python/examples/build_simple_arm_model.py#L133) example to better reflect the example in the README. This example ran as part of CTest.
3. Could you fix indentation? It seems like the PR includes tabs instead of spaces.

aymanhab:
Bogus failures, timeout. Please take a look @chrisdembia when you have a chance. Thanks.

chrisdembia:
I just reviewed the PR at about the same time as you posted. See my comment above.

aymanhab:
Thanks @chrisdembia I thought about moving the fix to the top level .i file but noticed that you did the same fix in two places for the Joint constructors. I agree that changes that extend C++ classes should go into the common top level file rather than language specific. Will move both this and the joints fix to common place.

chrisdembia:
> but noticed that you did the same fix in two places for the Joint constructors.

I knowww...I don't remember how that came to be. Sorry!

aymanhab:
Not sure what's the convention on .i files formatting but the existing one already has a mix as well, let me know how you'd prefer it (since I use visual studio on Windows and don't have a preference) and I'll fix the file too. Thanks.

aymanhab:
Done, thanks @chrisdembia for review please let me know if ready to merge.

chrisdembia:
@aymanhab were you able to test the new constructors, at least locally?

aymanhab:
Yes I did, and the code compiled and the constructor was found indeed.

chrisdembia:
The PR tests passed, so I have merged this. Thank you.

aymanhab:
Awesome, thank you @chrisdembia

PR #1392, opened 2016-11-10
Silence compiler warnings

tkuchida:


OpenSim/Common/DataTable.h
```diff
@@ -843,8 +843,9 @@ class DataTable_ : public AbstractDataTable {
         const size_t totalWidth{std::accumulate(columnWidths.cbegin(),
                                                 columnWidths.cend(),
                                                 static_cast<size_t>(0))};
-        result.reserve((totalWidth * table.size() * excessAllocation) +
-                       result.capacity());
+        result.reserve(result.capacity() +
+            static_cast<unsigned>(std::roundf(
+                totalWidth * table.size() * excessAllocation)));
```
chrisdembia: @klshrinidhi might want to review this.

tkuchida: The existing implementation is implicitly casting a `float` to an `unsigned`; the compiler warns that information is being lost.

klshrinidhi: The logic doesn't require rounding.

tkuchida: Ok, I will change if you're fine with the silent truncation.

klshrinidhi: If you are concerned about readability, leave it as is. Its okay.

tkuchida: Actually, I have two concerns.

1. It isn't clear why `excessAllocation` has been set to 1.25 specifically. A comment on L762 or L845 would be helpful to explain the intention, as it seems like the most natural action (which would not require a comment) would be to allocate only as much memory as is required.

2. Because it isn't clear why `excessAllocation` has been set to 1.25, it therefore isn't immediately obvious that truncation via (implicit or explicit) casting is appropriate. I think I've used a language at some point that didn't allow casting to an int without floor/ceil/round first, so I'm probably biased to see something like `static_cast<unsigned>(3.14)` as a bug.

klshrinidhi:
I am done with review. :+1: 
Merge now or anytime after you make any changes you like.

OpenSim/Common/DataTable.h
tkuchida: > Feel free to add any documentation you think is helpful/necessary.

I have been unable to find a reference suggesting 1.25. The only other time I have looked into memory allocation is for Maple (not a "real" programming language, I admit), which allocates using a factor of 2. Can you point me to the reference you are using?

> My understanding is this is certainly not a bug.

Ok, I have removed `roundf()`.

klshrinidhi: This is not based on public reference. It is just my approximate calculation to make sure to over-allocate. Didn't want to go work out the exact amount of space needed.

I don't have anything to add in this issues. Feel free to do edit any way you think is helpful/necessary. :+1:

tkuchida:
Addressed feedback from @klshrinidhi and removed `roundf()`. I see two :+1:s and four :white_check_mark:s; merging.

Issue #1398, opened 2016-11-14
Need regression test to verify that marker placements after scaling are consistent 


aseth1:
testScale currently only verifies that scale factors are recomputed consistently (with results from previous versions), but there are no tests for whether or not markers are placed in the same locations as those from scaling in previous versions of OpenSim.

This is necessary to verify the effects of #1372 on marker placement.

aseth1:
closed by #1400

PR #1400, opened 2016-11-15
Test scale compares factors and markers

aseth1:
Addresses #1398.

Adds a condition to testScale cases that compares the marker locations post scale (and MarkerPlacer) and verifies that they match the standard. There was previously no coverage that marker placements were being performed consistently/correctly.

The test uncovered a bug in `MarkerPlacer` in its use of `Frame::findLocationInAnotherFrame()` to transform the marker's global position (location in ground) to its location in the Frame to which it is attached. The bug was also fixed.

Should be an easy review. @klshrinidhi can you review. It should help with coverage for  #1372. @aymanhab can you take a look too. Thanks!


klshrinidhi:
:+1:

aymanhab:
Thanks @aseth1 for the fix and the stronger test. Just curious how did it come up since there's no GUI to visualize, did you visualize the standard in 3.3? and why we don't get the Marker comparison/test for free when we compare models using == (considering that MarkerSet was a Property of Model)?

aseth1:
> Just curious how did it come up since there's no GUI

The issue came up during review #1372 which change the staticPose (`MarkerData`) to staticPoseTable. I looked for coverage and was surprised to find none. `testScale` was only comparing scale factors.

> did you visualize the standard in 3.3?

Yes, I ran the setup in 3.3 and generated the resulting scaled model in the GUI. There was no scaled model in the repo or used by the test. 

> why we don't get the Marker comparison/test for free when we compare models using == (considering that MarkerSet was a Property of Model)?

There was no such `==` comparison for models in `testScale` and my guess it would fail for a multitude of unrelated issues: e.g. order of bodies, joints,  and even markers, plus `==` is not well defined for double values.  I thought it more straightforward to compare the placement of markers directly.

chrisdembia:
> plus == is not well defined for double values.

I think a `==` could work, given how Properties are implemented. For properties, doubles are considered equal if their difference is less than 1e-7; see [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Property.cpp#L52).

aseth1:
> not well defined ...

I should have said unable to specify the comparison tolerance. A fixed tolerance makes sense for most properties that are set by the user, but in this case the locations are computed and depend on the assembly tolerance of the static pose IK (assemblySolver) tolerance which defaults to 1e-4.  

> I think a == could work

Perhaps.  I believe the current tests achieves the required coverage and that `==` will cause unelated issues (like fixed comparison tolerance and arbitrary order) to be have to be addressed unnecessarily. @chrisdembia are you advocating/requesting the change to use `==`?

chrisdembia:
> @chrisdembia are you advocating/requesting the change to use ==?

I am not advocating anything; just wanted to clarify. I have not reviewed this PR.

aymanhab:
My recollection is that we use == already in other tests, if thats not the case or there are other side effects due to assembly then we can put that discussion off.

aseth1:
> My recollection is that we use == already in other tests, if thats not the case or there are other side effects due to assembly then we can put that discussion off. 

Yes, we do often `==` in other tests to verify serialization/deserialization. However, the original authors of `testScale` did not see fit to use `==` for whatever reason and compared scale factors by first locating them by name. That was a clue to me they ran into or were concerned about issues of order. I have no real interest in finding out what happens with  `==` in this case. I am more concerned in making sure marker placements are adequately tested.

aymanhab:
Ideally we can reuse the comparison in variety of tests but that's not necessary for this PR just something to keep in mind if it comes up.  The rest is straightforward except for the fix to re-express marker location which may require either clearup of documentation or renaming. Feel free to merge, and open issues if you agree. Thanks.

aseth1:
> The rest is straightforward except for the fix to re-express marker location which may require either clearup of documentation or renaming.

Do you mean the fix [here](https://github.com/opensim-org/opensim-core/pull/1400/files#diff-1bb5e2ff6e416085795c391026a07e71R412)? Do you mean adding documentation here as well or that the method should be renamed/changed so that the existing order would have worked? Please clarify.

aymanhab:
@aseth1 Considering the symmetry of the syntax (before/after the fix) it looks like the method could be interpreted either-way.  Either we need to improve the documentation in Frame.h or better change the method name so it's less confusing.

aseth1:
Thanks @aymanhab. Yes, @chrisdembia and I did have similar concerns so it may be worth changing the method definition. Created issue #1412 to track and collect feedback.

PR #1401, opened 2016-11-15
[doxygen] Correction to FiberForceLengthCurve figure

### LINKED ISSUE #1389: FiberForceLengthCurve documentation is inconsistent
Strain is [defined](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Actuators/FiberForceLengthCurve.h#L76) as `(l-l0)/l0` and the default value of `strainAtZeroForce` is 0, so the "strain" label on the x-axis of [the accompanying figure](https://github.com/opensim-org/opensim-core/blob/master/doc/images/fig_FiberForceLengthCurve.png) may be confusing. This doxygen page should be reviewed (and perhaps the figure should be updated) to avoid this confusion. This issue arose on the Forum ([topic 7161](https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=7161)).

tkuchida:
Fixes #1389. The existing x-axis label is incorrect: strain has been defined as (L–L0)/L0, so a passive fiber force of zero corresponds to strain = 0 or, equivalently, normalized fiber length = 1. The definition of strain is consistent elsewhere on the page.


tkuchida:
Cool :sunglasses: Thanks for merging.

PR #1403, opened 2016-11-15
Visualize wrap-torus and ellipsoid

aymanhab:
Visualize wrap-torus and ellipsoid, also add a test case to testVisualization and make sure defaultAppearance is used rather than hardcoded values. This doesn't handle partial WrapObjects or change behavior on the API visualizer side to make WrapTorus visible, however the primitives are passed correctly to the GUI and display correctly there.


aymanhab:
Bogus test failure. Ready for review, and pretty straightforward. @chrisdembia @tkuchida @aseth1

OpenSim/Tools/Test/testVisualization.cpp
```diff
@@ -399,3 +431,68 @@ void populate_contactModelPrimitives(SimTK::Array_<DecorativeGeometry>& stdPrimi
...
+        .setIndexOnBody(0).setScale(0.2).setOpacity(1)
+        .setRepresentation(SimTK::DecorativeGeometry::DrawSurface));
+    // Frames for the Joint
+    stdPrimitives.push_back(
+        DecorativeFrame(1.0).setBodyId(1).setColor(SimTK::White)
+        .setIndexOnBody(0).setScale(0.2).setOpacity(1)
+        .setRepresentation(SimTK::DecorativeGeometry::DrawSurface)
+        .setTransform(Vec3({-.01, -.03, .03})));
+    Transform cylTransform;
+    cylTransform.updR().setRotationFromAngleAboutX(SimTK_PI / 2);
```
tkuchida: Should the rotations be (a) about all three axes, and (b) by amounts that are not multiples of 90° to ensure the geometry isn't accidentally getting flipped or reflected somehow? A few things can still be wrong even if a pi/2 rotation lands the cylinder in the right place.

aymanhab:
Totally agree @tkuchida on the need to load models and checking them visually during Alpha testing.

OpenSim/Tools/Test/testVisualization.cpp
aymanhab: Great point @tkuchida Actually the rotation here is to parallel the visualization code that applies a rotation to takes z axis to y axis due to different assumptions between DecorativeCylinder aligned with y  and WrapCylinder aligned with z. Making non trivial transforms in model and standard proved hard to incorporate into a test due to roundoff  especially that the rotations are in radians but I'm open to suggestions.

tkuchida: Ah, ok thanks for clarifying. IIRC the issue of how visualization should be tested was discussed at length a while back; this PR seems to be in keeping with the established approach.

aymanhab:
@chrisdembia Can you take a look. Thanks.

OpenSim/Simulation/Wrap/WrapObject.cpp
```diff
@@ -64,7 +64,8 @@ void WrapObject::constructProperties()
     constructProperty_quadrant("Unassigned");
     Appearance defaultAppearance;
     defaultAppearance.set_color(SimTK::Cyan);
-    defaultAppearance.set_representation(VisualRepresentation::DrawWireframe);
+    defaultAppearance.set_opacity(0.5);
+    defaultAppearance.set_representation(VisualRepresentation::DrawSurface);
```
chrisdembia: Could you explain the reason for changing this? Just curious. 

Does this change affect backwards compatibility?

OpenSim/Tools/Test/testVisualization.cpp
chrisdembia: Perhaps a comment should be added to the code so that future test maintainers won't need to figure that out.

OpenSim/Tools/Test/testVisualization.cpp
```diff
@@ -135,6 +144,12 @@ int main()
         Model modelWithContacts("visualize_contacts.osim");
         populate_contactModelPrimitives(standard);
         testVisModelAgainstStandard(modelWithContacts, standard);
+        
+        // Model with WrapObjects
+        Model modelWithWrap("test_wrapAllVis.osim");
+        modelWithWrap.updDisplayHints().set_show_frames(false);
```
chrisdembia: If show_frames == false, why are there so many DecorativeFrames in `populate_wrapModelPrimitives`?

OpenSim/Tools/Test/testVisualization.cpp
```diff
@@ -225,23 +241,39 @@ bool testVisModelAgainstStandard(Model& model, const SimTK::Array_<DecorativeGeo
...
             DecorativeGeometryImplementationText dgiTextFromModel;
+            dgiTextFromModel.setPrintTransforms(false);
             geometryToDisplay[i].implementGeometry(dgiTextFromModel);
             if (!(dgiTextFromStandard.getAsString() == dgiTextFromModel.getAsString()))
                 throw  OpenSim::Exception("failed comparing " + dgiTextFromStandard.getAsString() + "vs." + dgiTextFromModel.getAsString());
+            // Compare transforms, this has to be more lenient than String comparison due to roundoff
+            SimTK::Mat44 diffTransform = nextGeom->getTransform().toMat44() - geometryToDisplay[i].getTransform().toMat44();
+            double norm = diffTransform.norm();
+            if (norm > SimTK::Eps) 
+                throw  OpenSim::Exception("Difference in transforms of DecoratibeGeometry:" + std::to_string(norm));
```
chrisdembia: You should be able to use `SimTK_TEST_EQ` or `SimTK_TEST_EQ_TOL`, which will do an "almost-equal" check.

OpenSim/Tools/Test/testVisualization.cpp
```diff
@@ -98,21 +99,27 @@ class DecorativeGeometryImplementationText : public SimTK::DecorativeGeometryImp
     std::string getAsString() {
         return printout.str();
     }
+    void setPrintTransforms(bool toPrint) { printTransform = toPrint;}
 private:
     std::stringstream printout;
+    bool printTransform; // Flag to indicate whether or not to output transform as String
```
chrisdembia: You can get rid of the lines like `dgiTextFromStandard.setPrintTransforms(false);` by changing this member declaration to the following:

```
bool printTransform = false;
```

(C++11 non-static member initializer).

OpenSim/Simulation/Wrap/WrapObject.cpp
aymanhab: This is default appearance, users who cared and changed that will have their preserved AFAIK.  When updating the code to remove hardcoded settings, that broke the test cases. Based on the code, the API visualizer used to not set representation (default Surface) and a hard-coded opacity 0.5 while GUI used hard-coded (wire-frame) and no-opacity. Something has to give but I'm ok with changing to Wireframe and Opacity 0.5 and updating test cases accordingly. Please let me know your preference.

OpenSim/Tools/Test/testVisualization.cpp
aymanhab: Good point, will do, thanks @chrisdembia

OpenSim/Tools/Test/testVisualization.cpp
aymanhab: I believe there's a bug where the flag is not universally applied/respected, I have a separate branch to investigate/fix but I don't want it entangled with this one.

OpenSim/Tools/Test/testVisualization.cpp
aymanhab: Thanks @chrisdembia will try it.

OpenSim/Simulation/Wrap/WrapObject.cpp
chrisdembia: Okay; my preference is to leave this as-is.

chrisdembia:
Great; please merge when the tests pass.

aymanhab:
Tests pass, except for sporadic out of time and/or unrelated testScale issue #1419.

Thanks @chrisdembia Merging.

Issue #845, opened 2016-03-16
Consistent "isDisabled" vs "enabled" vs "is_disabled" properties


chrisdembia:
See Reporters, Controllers for examples.

aseth1:
I would like component writers to only focus on the affirmative case and NEVER check that flag to change their component's behavior. The enabling/disabling capability should be a feature of Component itself and it dictates whether (enabled) or not (disabled) that Component is included in the list (tree) of components used to build the system. It would then allow Components to be around for the sake of serialization but when disabled they have 0 impact on the system because they are simply not a part of the system. Then there would be no ambiguity about whether the flag works or not, which is the biggest frustration for users and then have to manually remove a component from XML.

chrisdembia:
+1, but I still think it should be `<enabled>` or `<is_enabled>` rather than something with a negative in it.

aymanhab:
I ran into this issue again when visualizing ContactGeometry as it turns out the isDisabled flag is likely ignored for ContactForce. We should have a design/plan for the long term even if we implement a small part of it in 4.0 (e.g. are all ModelComponents disablable? If yes what does it mean to disable a Body/Component-with-children? How does it affect visualization? Some SimTK level objects support the flag (e.g. Force, Constraint) but not others. Should we add a pass through the top level model component tree before "Connecting" and remove all disabled Components, so downstream functions can assume enabled all the time?

tkuchida:
> what does it mean to disable a Body/Component-with-children?

In the case of an actuator, it seems like "disable" would mean the actuator is still part of the model but just not applying any force. This may be the desired effect since ignoring the component entirely would have side effects (e.g., adjusting indices into the Model's ForceSet). If the flag removes the component entirely, perhaps something like `is_ignored` would be clearer?

jenhicks:
First step: rename disabled flags. I think we decided on the following.
Forces: appliesForce
Constraints: isEnforced
Controllers: enabled

jenhicks:
Also need to reverse the logic.

PR #1407, opened 2016-11-16
Force::isDisabled  -->>  Force::appliesForce

klshrinidhi:
Addressing #845 only for `Force`. 
Future PRs will address rest.


chrisdembia:
I don't think any osim files should be changed. Rather, the XMLDocument version number should be incremented and `Force::updateFromXMLNode()` should convert old models' `isDisabled` to `appliesForce`.

chrisdembia:
There are some notes about how to update the XML format here: https://github.com/opensim-org/opensim-core/blob/master/DEVELOPING.md

tkuchida:
> the XMLDocument version number should be incremented and `Force::updateFromXMLNode()` should convert old models' `isDisabled` to `appliesForce`

Oh, definitely! Good point.

klshrinidhi:
I think the point here is to use `Force::appliesForce` instead `Force::isDisabled` for all purposes in future. If we keep the osim/xml files with old property name, users who copy-and-edit these files will continue using the old names. I think the function `updateFromXMLNode` must read the old files but the repository should contain files in the latest format.

tkuchida:
> I think the function `updateFromXMLNode` must read the old files but the repository should contain files in the latest format.

Related to #206 (marked "Next Release"). We should have a folder of model files in all formats to rigorously exercise the `updateFromXMLNode()` methods in a dedicated test case, but the convention has been to leave model files in the repo as they are. Upon loading and re-saving, the model file will be updated to the latest version.

klshrinidhi:
Okay. I think if someone does a copy-edit on a file in the repo, he will end up with two or three different files which mean the exact same thing. It could lead to confusion. To me it doesn't make much sense If we change format and we don't adopt it in the repository.

chrisdembia:
> Okay. I think if someone does a copy-edit on a file in the repo, he will end up with two or three different files which mean the exact same thing. It could lead to confusion.

@tkuchida is talking about a separate area in the repo, presumably in which each file would be named in a way that clarified the fact that they are different XML versions of the same model (e.g., `arm_v30500.osim`, `arm_v30505.osim`).



> To me it doesn't make much sense If we change format and we don't adopt it in the repository.

I'm okay with adopting changed formats in the repository to a limited extent (see below), but when we do this the whole file should be updated all together. A file that says "version 30000" at the top should not have features of version 30500 sprinkled within it. I think that would be very confusing, and would be like time travel ⌛️ .

I think it would be good for our *examples* and reference materials to use the latest model version. It is important that some of the model files in the repo for testing remain with the old version number, because it is currently (unfortunately; see #206) our best way of (regression-) testing the updateFromXMLNode code.

aymanhab:
@klshrinidhi You may have figured this out already but the point here is that the contents of the XML file should be kept consistent with the header (version number) at the top of the same file so that the conversion code is triggered. If we update the contents of the XML file and leave the header unchanged then the conversion code is never triggered.

klshrinidhi:
> @tkuchida is talking about a separate area in the repo, presumably where each file would be named in a way that clarified the fact that they are different XML versions of the same model (e.g., arm_v30500.osim, arm_v30505.osim).

I know. I am not talking about testing reading of old files.
I am only talking about using latest format in the repository.

> I'm okay with adopting changed formats in the repository to a limited extent (see below), but when we do this the whole file should be updated all together. A file that says "version 30000" at the top should not have features of version 30500 sprinkled within it. I think that would be very confusing, and would be like time travel :hourglass: .
I think it would be good for our examples and reference materials to use the latest model version. It is important that some of the model files in the repo for testing remain with the old version number, because it is currently (unfortunately; see #206) our best way of (regression-) testing the updateFromXMLNode code.

I know the version number needs to be updated on all the XML files edited by this PR. It doesn't make sense to have new tag names in old versioned files.

klshrinidhi:
> @klshrinidhi You may have figured this out already but the point here is that the contents of the XML file should be kept consistent with the header (version number) at the top of the same file so that the conversion code is triggered. If we update the contents of the XML file and leave the header unchanged then the conversion code is never triggered.

I know the version number on these xml/osim files need to be updated. I will do so (and any changes needed to `updateFromXMLNode` and other functions) if we decide to adopt the latest format in the repository. I don't want to go ahead and make changes only to revert them back. So waiting for a firm decision on this.

chrisdembia:
I would not edit OSIM files in this PR. That will also allow you to test that your updateFromXMLNode code is testing.

We could have a separate project closer to the release (when the XML format is more stable) for updating example XML/OSIM files to the version that will be used for the 4.0 release. It would be bad to update those files multiple times (would waste space in the repo).

chrisdembia:
I updated my post above.

jenhicks:
> We could have a separate project closer to the release (when the XML format is more stable) for updating example XML/OSIM files to the version that will be used for the 4.0 release. It would be bad to update those files multiple times (would waste space in the repo's history).

This is what we have done for past (major) releases, as far as I know. My vote is to stick with this convention.

klshrinidhi:
> This is what we have done for past (major) releases, as far as I know. My vote is to stick with this convention.

Okay. That means 
* If I am correct, this PR cannot guarantee that the changes actually work on reading/writing files. 
* Doxygen comment and osim/xml files in the repository will have differ until the osim/xml files are brought up-to-date.

Also where is the issue that has been tracking all these changes so far so that addressing the issue will bring all files up-to-date before 4.0 release ?

chrisdembia:
> If I am correct, this PR cannot guarantee that the changes actually work on reading/writing files.

It is not necessary to test properties in general; the property system has its own tests. However, it is good to test specific changes to `updateFromXMLNode()`. By leaving old model files in the tests, we make sure that `updateFromXMLNode()` code works; though right now this is a very poor way of testing `updateFromXMLNode()`.

> Doxygen comment and osim/xml files in the repository will have differ until the osim/xml files are brought up-to-date.

Perhaps it would be a good idea to put a note about the change in the class description for `Force`. Something like `@note In 4.0, the property 'isDisabled' was renamed to 'appliesForce'`. That way, users who have an old model and are looking for documentation of the `isDisabled` property don't get lost.

tkuchida:
> Perhaps it would be a good idea to put a note about the change in the class description for `Force`.

Could also mention in [CHANGELOG.md](https://github.com/opensim-org/opensim-core/blob/master/CHANGELOG.md).

klshrinidhi:
> Perhaps it would be a good idea to put a note about the change in the class description for Force. Something like @note In 4.0, the property 'isDisabled' was renamed to 'appliesForce'. That way, users who have an old model and are looking for documentation of the isDisabled property don't get lost.

Has this been done for all previous property changes as well ?

> Could also mention in CHANGELOG.md.

Yes. My response is same as above.

aseth1:
> Also where is the issue that has been tracking all these changes so far so that addressing the issue will bring all files up-to-date before 4.0 release?

I don't think we ever want to update all the files in the repository, just primarily example models. Most model files exist for regression tests and starting with older versions ensures that user's models, which are undoubtedly older versions, will load and execute correct.

> If I am correct, this PR cannot guarantee that the changes actually work on reading/writing files.

Reading old versions will be tested by leaving them be, and for the new format it is good to serialize and deserialize in a local test (e.g. testForces) to verify that the latest version is also being written and read in correctly.

chrisdembia:
> Has this been done for all previous property changes as well ?

No but I think it could be a good idea.

klshrinidhi:
**To summarize:**
* I will revert changes to osim/xml files.
* I will update `updateFromXMLNode` accordingly to check for version number and so on.
* Add test for reading/writing files.

klshrinidhi:
I have made the changes I mention.

OpenSim/Simulation/Test/testForces.cpp
```diff
@@ -1670,3 +1677,20 @@ void testExternalForce()
             ASSERT_EQUAL(def, val, 10*accuracy);
     }
 }
+
+void testSerializeDeserialize() {
```
tkuchida: This isn't the test I would have expected. An old model is read in (which will indeed trigger the new `updateFromXMLNode()` method), then the file is written and read in again. The `ASSERT()` is checking that serializing a v30508 model and then deserializing it results in the same `ForceSet`—but correctness of the new `updateFromXMLNode()` method is not checked. Perhaps use a model that contains muscles that are both enabled and disabled, and confirm that the new `appliesForce` property is being set correctly?

jenhicks:
Add to the comment "Prior to 4.0 flag was is_disabled() with opposite logic"

OpenSim/Simulation/Test/testForces.cpp
klshrinidhi: Do you have any other model file in mind ?

The `ASSERT` is checking the property of the model read from the old file and new file.

tkuchida: > Do you have any other model file in mind ?

As far as I can tell, none of the models currently in the repo contain a muscle whose `isDisabled` property is set to `true`.

> The `ASSERT` is checking the property of the model read from the old file and new file.

I thought the `updateFromXMLNode()` code will be executed when the model is loaded (L1683), so the test is comparing a v30508 model to the same v30508 model (after serialization/deserialization). Not saying that's a bad thing to test, but it seems like that type of check should be done in a more general way for all properties (if it isn't being tested already).

tkuchida: > The test just makes sure that property values are same regardless of which version the model file is.

I don't think it does. Suppose a model has 5 "enabled" and 5 "disabled" muscles. The test will still pass if the `updateFromXMLNode()` code always erroneously set the `appliesForce` property to its default value of `true`.

klshrinidhi: Yes. So changing the model file will not make the test better. And I don't think we can load the file in the *old* style (as if version number on the file is the latest version) preserving *isDisabled* so it can be checked against *appliesForce*.

tkuchida: > So changing the model file will not make the test better.

My suggestion was to load an old-format model that contains both enabled and disabled muscles, and then check that the `appliesForce` property was set for each muscle as expected. It's not necessary to include this test (I don't think tests like this exist for other property changes), but it seems like that's what was intended here.

klshrinidhi: Okay. Got it. I will make the changes.

klshrinidhi: I have changed the test to flip the property in the old file and make sure the read-in model has the property flipped.

OpenSim/Simulation/Model/Force.cpp
```diff
@@ -56,7 +56,29 @@ void Force::setNull()
...
+            // negate the contained value.
+            std::string oldName{"isDisabled"};
+            std::string newName{"appliesForce"};
+            if(node.hasElement(oldName)) {
+                auto elem = node.getRequiredElement(oldName);
+                elem.setElementTag(newName);
+                if(elem.getValue().find("true") != std::string::npos)
+                    elem.setValue("false");
+                else
+                    elem.setValue("true");
```
chrisdembia: This logic looks pretty good to me. A question: If someone has `<isDisabled></isDisabled>` in their XML file, should we interpret this as `appliesForce == true`, as is currently done? What would happen before this PR if the XML file has `<isDisabled></isDisabled>`?  

For reference, this is the code that Simbody uses to parse bool properties (I think):

```cpp
bool String::tryConvertToBool(bool& out) const {                                
    const String adjusted = cleanUp(*this);                                     
    if (adjusted=="true")  {out=true;  return true;}                            
    if (adjusted=="false") {out=false; return true;}                            
    std::istringstream sstream(adjusted);                                       
    sstream >> out;                                                             
    return !sstream.fail();                                                     
}   
```

https://github.com/simbody/simbody/blob/master/SimTKcommon/src/String.cpp#L81

Thus, a more robust way to perform this conversion might be:

```cpp
elem.setValue(SimTK::String(!SimTK::String::tryConvertToBool(elem.getValue())));
```

OpenSim/Simulation/Model/Force.cpp
```diff
@@ -56,7 +56,29 @@ void Force::setNull()
...
+            }
+        }
+    }
+
+    ModelComponent::updateFromXMLNode(node, versionNumber);
```
chrisdembia: Can you use`Super::` instead of `ModelComponent::`? I once had to deal with a bug where the base class type name was hard-coded in `updateFromXMLNode()` and it was very confusing to figure out.

OpenSim/Simulation/Test/testForces.cpp
tkuchida: Oh cool :sunglasses: Looks good :+1:

OpenSim/Simulation/Model/Force.h
```diff
@@ -47,14 +47,18 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Force, ModelComponent);
...
-    /** A Force element is active (enabled) by default. **/
-    OpenSim_DECLARE_PROPERTY(isDisabled, bool,
-        "Flag indicating whether the force is disabled or not. Disabled means"
+    /** A Force element is active (enabled) by default.
+    NOTE: Prior to OpenSim 4.0, this property was named **isDisabled**.
+          If **isDisabled** is **true**, **appliesForce** is **false**.
+          If **isDisabled** is **false**, **appliesForce** is **true**.          */
+    OpenSim_DECLARE_PROPERTY(appliesForce, bool,
+        "Flag indicating whether the force is applied or not. Value of false means"
         " that the force is not active in subsequent dynamics realizations.");
```
chrisdembia: The new comment above is very clear. In the dev meeting, I think we decided to also put a brief note inside the XML comment as well, so that the change would appear in XML files. Did you check how the doxygen block looks in the generated HTML? There's a chance that doxygen blocks for properties don't show up nicely (because of the macro).

OpenSim/Simulation/Model/Force.cpp
klshrinidhi: I don't think empty tags should be allowed. Either the tag should contain a valid value or it should not be there so the default value is chosen.
I will do whatever is being done with rest of the properties. Not sure what that is.

OpenSim/Simulation/Model/Force.cpp
klshrinidhi: I don't think it is confusing at all. Also the class `Force` doesn't contain a type alias named `Super`.

OpenSim/Simulation/Model/Force.h
klshrinidhi: I looked at the generated Doxygen. It looks fine.

I have updated the property comment to mention the change in property name.

OpenSim/Simulation/Model/Force.cpp
chrisdembia: > macro magic

🎩 🐰 🎉

aseth1:
Appears the diff is with master at the time this PR was opened. Find the current diff with master [here](https://github.com/opensim-org/opensim-core/compare/issue_845_force_isDisabled_to_appliesForce). Sorry for the inconvenience.

chrisdembia:
In my first review, I asked if the doxygen output for Force looked okay. Here is how it looks:

image

I think this looks a little confusing (it is not clear that the comment refers to the appliesForce property) and messy, and is not what @klshrinidhi intended.

But this is a minor comment, and it can be ignored.

chrisdembia:
The XMLDocument number should be 30509; 30508 is from my PR about connector/input serialization.

aseth1:
@chrisdembia I updated the doxygen and description for the `appliesForce` property and incremented the version number as suggested. The Travis failure was due a temporary loss of connection.

Issue #1409, opened 2016-11-17
Usability of DataTable_<ETX, ETY>::getRow() and appendRow()


tkuchida:
Likely an issue that can be addressed by adding an example to the doxygen for `DataTable_`. Here's a MWE:
``` cpp
TimeSeriesTableVec3 tableA;
tableA.setColumnLabels({"col0", "col1"});
SimTK::RowVector_<Vec3> row{2, Vec3(0.1, 0.2, 0.3)};
tableA.appendRow(1.0, row);

TimeSeriesTableVec3 tableB;
tableB.setColumnLabels({"col0", "col1"});
tableB.appendRow(tableA.getIndependentColumn()[0],
    tableA.getRowAtIndex(0)); //looks like it should work, but doesn't
tableB.appendRow(tableA.getIndependentColumn()[0],
    static_cast<SimTK::RowVector_<Vec3>>(tableA.getRowAtIndex(0))); //works
```
This user interface is slightly awkward since it seems like `tableA.getRowAtIndex()` should return an object that is consumable by `tableB.appendRow()`. Similar to #1384.

klshrinidhi:
All this is mostly to do with `DataTable` not hiding its implementation details and using SimTK objects in its interface. I think the interface of `DataTable` is not good. It never was.

`getRowAtIndex` and `getRow` return a `SimTK::RowVectorView`, which is returned by `SimTK::Matrix` for efficiency. `RowVectorView` is a proxy class and provides a *view* into a block of the underlying matrix.  Whereas `appendRow` expects a `SimTK::RowVector`. You should have seen a compiler error (which I think is the earliest I can catch this problem).

Regarding your use of `static_cast`, I am not sure if it the right thing to do because `RowVectorView` and `RowVector` are siblings in the hierarchy. `static_cast` will blow up at run-time if I correct.

**Solution:**
I can add an overload to `appendRow` to accept a `RowVectorView`. It should be pretty simple fix. Will get back with a PR.

klshrinidhi:
Also worth mentioning is that my PR #1372 that I merged today adds functions `removeRowAtIndex` and `removeRow` if that can help avoid the above snippet until I add the overload.

tkuchida:
> `static_cast` will blow up at run-time if I correct.

`static_cast` is the only solution I have found that works (in Release mode, at least :wink:).

> I can add an overload to `appendRow` to accept a `RowVectorView`. It should be pretty simple fix. Will get back with a PR.

[...]

PR #1410, opened 2016-11-17
Add appendRow overload that accepts RowVectorView

klshrinidhi:
Addressing #1409.

### DIFF
--- a/OpenSim/Common/Test/testDataTable.cpp
+++ b/OpenSim/Common/Test/testDataTable.cpp
@@ -204,6 +204,23 @@ int main() {
                 "(\"Filename\").getValue<std::string>() != std::string"
                 "{\"/path/to/file\"}"};
 
+    {
+    std::cout << "Test feeding rows of one table to another [double]."
+              << std::endl;
+    TimeSeriesTable tableCopy{};
+    tableCopy.setColumnLabels(table.getColumnLabels());
+    for(unsigned row = 0; row < table.getNumRows(); ++row)
+        tableCopy.appendRow(table.getIndependentColumn()[row],
+                            table.getRowAtIndex(row));
+
+    ASSERT(tableCopy.getNumColumns()     == table.getNumColumns());
+    ASSERT(tableCopy.getNumRows()        == table.getNumRows());
+    for(unsigned r = 0; r < table.getNumRows(); ++r)
+        for(unsigned c = 0; c < table.getNumColumns(); ++c)
+            ASSERT(tableCopy.getRowAtIndex(r)[c] ==
+                       table.getRowAtIndex(r)[c]);
+    }
+
     std::cout << "Test numComponentsPerElement()." << std::endl;
     ASSERT((static_cast<AbstractDataTable&&>
             (DataTable_<double, double    >{})).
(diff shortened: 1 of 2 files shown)


OpenSim/Common/Test/testDataTable.cpp
```diff
@@ -251,6 +268,29 @@ int main() {
...
+        ASSERT(tableVec3Copy.getNumColumns() == tableVec3.getNumColumns());
+        ASSERT(tableVec3Copy.getNumRows()    == tableVec3.getNumRows());
+        for(unsigned r = 0; r < tableVec3.getNumRows(); ++r) {
+            for(unsigned c = 0; c < tableVec3.getNumColumns(); ++c) {
+                ASSERT(tableVec3Copy.getRowAtIndex(r).getElt(0, c)[0] ==
+                           tableVec3.getRowAtIndex(r).getElt(0, c)[0]);
+                ASSERT(tableVec3Copy.getRowAtIndex(r).getElt(0, c)[1] ==
+                           tableVec3.getRowAtIndex(r).getElt(0, c)[1]);
+                ASSERT(tableVec3Copy.getRowAtIndex(r).getElt(0, c)[2] ==
+                           tableVec3.getRowAtIndex(r).getElt(0, c)[2]);
```
aseth1: Would be easier to read if we used  `ASSERT_EQUAL<SimTK::Vec3>(...)`. Also why access a row element using `.getElt(0,c)` then using `[c]` operator?

klshrinidhi: I didn't know about `ASSERT_EQUAL`. I have been using `ASSERT` everywhere. Also my intention is to compare for bitwise equality here. Don't want floating point comparison of *almost equal*.

Replacing `getElt` with `operator[]` gives a exception in Debug mode. My latest commit was to fix exactly that by replacing `operator[]` with `getElt`.

aseth1: I thought ASSERT_EQUAL without a tolerance was a bitwise match, but it looks like in that case it defaults to some tolerance, so you can leave it as you have it.

It is troubling that the `[ ]` operator does not work in debug. That should be an issue, not for this PR. What assertion is failing?  Do you know why? If user's have to access elements of a row using `.getElt()` all the time that is going to be ☹️ for them.

klshrinidhi: > It is troubling that the [ ] operator does not work in debug. That should be an issue, not for this PR. What assertion is failing? Do you know why? If user's have to access elements of a row using .getElt() all the time that is going to be :frowning_face: for them.

I haven't looked at it much but if I remember correctly, `RowVectorView` is used to handle both `const` and `non-const` data using a runtime flag. The snippet I have will work if I store the row into a `const` variable and then use its `operator[]`. I think the fix is to use two different types to represent a row -- one for `const` and one for `non-const`. Sherm will know.

klshrinidhi: @sherm1

klshrinidhi: The problem here is that the following snippet throws an exception.
```c++
SimTK::Matrix_ matrix{};
/*--------------------*/
/* Fill up the matrix */
/*--------------------*/
/* Following throws   */
matrix.row(1)[0];
```
This is because `SimTK::Matrix_::row()` returns a non-const `RowVectorView` but actually represents const data. Because `RowVectorView` is non-const, the call to `operator[]` above is calling the non-const `operator[]` and throwing an exception.

I think I can fix it on `DataTable` by adding const to the `RowVectorView` that `SimTK::Matrix` returns. But it will be good to if `SimTK::Matrix` did that.

sherm1:
You’re right, Shrinidhi! I was going to say a return value of `const RowVectorView` wouldn’t be different than `RowVectorView` since both are value returns but it does matter when the temporary return object is used directly as it would be in this case. So the const version of matrix.row() ought to return a const RowVectorView.

 

Regards,

Sherm

OpenSim/Common/Test/testDataTable.cpp
```diff
@@ -217,8 +217,8 @@ int main() {
     ASSERT(tableCopy.getNumRows()    == table.getNumRows());
     for(unsigned r = 0; r < table.getNumRows(); ++r)
         for(unsigned c = 0; c < table.getNumColumns(); ++c)
-            ASSERT(tableCopy.getRowAtIndex(r).getElt(0, c) ==
-                       table.getRowAtIndex(r).getElt(0, c));
+            ASSERT(tableCopy.getRowAtIndex(r)[c] ==
+                       table.getRowAtIndex(r)[c]);
```
aseth1: This is a lot nicer @klshrinidhi. Thank you for digging into the issue and providing a fix.

aseth1:
Good job @klshrinidhi. I am looking into the `testScale` failure that is should be unrelated to this PR before merging. Thanks.

aseth1:
OK, `testScale` failure looks to be due to #1372 merge prior to the test in #1400. See #1419.
I am merging this PR.

PR #1411, opened 2016-11-17
Comment out MarkersReference::getSpeedValues() and MarkersReference::getAccelerationValues().

### LINKED ISSUE #1408: MarkersReference::getSpeedValues unimplemented
We need to either implement the method or remove it from the header to avoid giving impression that the method is actually implemented. Came up over discussion/review of PR #1372 though it dates back before the PR.

chrisdembia:
This PR fixes #1408 by commenting out these unimplemented functions.

In the dev meeting, we suggested moving the exceptions to Reference, but these functions are not on Reference; they first appear in MarkersReference. So, in chatting with @aseth1 , we decided to just comment them out.

Fixes #1408.


aymanhab:
Thanks @chrisdembia

PR #1414, opened 2016-11-18
Controller::isDisabled -->> Controller::enabled

klshrinidhi:
Addressing #845 for `Controller`.

Does not include test for `updateFromXMLNode`. Need files in old format with some controllers with *isDisabled* in them.


aseth1:
Only a handful of [changes](https://github.com/opensim-org/opensim-core/compare/issue_845_Controller_isDisabled_to_enabled) to review.

chrisdembia:
I have the same comments for this PR as I have for the other two.

OpenSim/Tools/Test/testControllers.cpp
```diff
@@ -212,7 +212,7 @@ void testPrescribedControllerOnBlock(bool disabled)
     actuatorController.setName("testPrescribedController");
     actuatorController.setActuators(osimModel.updActuators());
     actuatorController.prescribeControlForActuator(0, new Constant(controlForce));
-    actuatorController.setDisabled(disabled);
+    actuatorController.setEnabled(!disabled);
```
carmichaelong: Since I think we tell users to also look at our test code for examples, maybe consider changing the `disabled` variable to `enabled` to promote the change?

aseth1: good idea! Will do.

aseth1:
@chrisdembia I took you advice in using`String::tryConvertToBool()` i think this is a good idea. I also have `isDisabled` default to false so by the `enabled` should be true. I also updated the revision number and added a corresponding comment in `XMLDocument.cpp`.

@carmichaelong I updated `testControllers` to use `enabled`  instead of `disabled` in agreement with the Controller's flag.

chrisdembia:
LGTM. Please merge when tests pass.

PR #1418, opened 2016-11-20
Ensure Path::_separator and Path::_isAbsolute are always initialized

chrisdembia:
This PR fixes an issue where a default-constructed ComponentPath would have
random values for `Path::_separator` and `Path::_isAbsolute` (the default Path
constructor did not initialize these members). The fix is to disallow creating a Path without initializing these members.

Suggested reviewers: @carmichaelong and @klshrinidhi.


chrisdembia:
Thank you @carmichaelong

chrisdembia:
I just merged master into this branch so that perhaps the CI tests will pass.

chrisdembia:
The tests pass and there are two ✅ . Merging.

PR #1422, opened 2016-11-29
Make FrameGeometry::generateDecorations enforce hints.show_fames and …

aymanhab:
…update test case accordingly

### DIFF
--- a/OpenSim/Tools/Test/testVisualization.cpp
+++ b/OpenSim/Tools/Test/testVisualization.cpp
@@ -38,7 +38,7 @@ Model createModel4AppearanceTest();
 void populate_doublePendulumPrimitives(SimTK::Array_<DecorativeGeometry>&); 
 void populate_composedTransformPrimitives(SimTK::Array_<DecorativeGeometry>&);
 void populate_contactModelPrimitives(SimTK::Array_<DecorativeGeometry>&);
-void populate_wrapModelPrimitives(SimTK::Array_<DecorativeGeometry>&);
+void populate_wrapModelPrimitives(SimTK::Array_<DecorativeGeometry>&, bool includeFrames=true);
 bool testVisModelAgainstStandard(Model& model, const SimTK::Array_<DecorativeGeometry>& stdPrimitives);
 
 // Implementation of DecorativeGeometryImplementation that prints the representation to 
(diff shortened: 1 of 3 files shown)


aymanhab:
This fixes the issue reported in PR #1403 and earlier where the flag to remove display of frames is ignored.
@tkuchida or @chrisdembia please take a look when you have a chance. Thanks. 🦃

aymanhab:
Thanks for the feedback @chrisdembia the reason I did the change in FrameGeometry was for consistency (hints are used exclusively in generateDecorations rather than hidden a few levels deep) so it's easier to maintain, also hints will not be used for all other types of geometry so it will be confusing to pass them in and not use them except for FrameGeometry. In essence I wanted changes localized as much as possible to FrameGeometry that's why I implemented the method there.  What do you think?

aymanhab:
Agreed that the test for connection/input in FrameGeometry::generateDecorations is redundant. Will remove.

OpenSim/Tools/Test/testVisualization.cpp
```diff
@@ -429,35 +434,37 @@ void populate_contactModelPrimitives(SimTK::Array_<DecorativeGeometry>& stdPrimi
...
-        .setIndexOnBody(0).setScale(0.2).setOpacity(1)
-        .setRepresentation(SimTK::DecorativeGeometry::DrawSurface));
-    // Frames for the Joint
-    stdPrimitives.push_back(
-        DecorativeFrame(1.0).setBodyId(1).setColor(SimTK::White)
-        .setIndexOnBody(0).setScale(0.2).setOpacity(1)
-        .setRepresentation(SimTK::DecorativeGeometry::DrawSurface)
-        .setTransform(Vec3({-.01, -.03, .03})));
+    if (includeFrames) {
+        stdPrimitives.push_back(
```
tkuchida: Perhaps cleaner to use a `for` loop here? …though I see that all you did was indent the existing code.

aymanhab:
Thanks much @tkuchida 
Please let me know if I addressed your questions/suggestions @chrisdembia Thanks.

chrisdembia:
> Please let me know if I addressed your questions/suggestions @chrisdembia Thanks.

Yes; thank you.

aymanhab:
Thanks much @chrisdembia and @tkuchida Merging.

Issue #1405, opened 2016-11-15
Clean up Manager interface/usage


carmichaelong:
These are left over TODOs from #1386:

- [x] Remove `_system` and subsequently make CMC use its own `TimeStepper` rather than the `Manager` to do its integration
- [x] Prevent any chance of mixing the wrong system/model between the integrator and timestepper. This may involve removing/changing some of the constructors and methods such as `setIntegrator()` and `setModel()`. As per discussion with @aseth1, it may be best to only allow construction of a `Manager` when given both the Model and Integrator, just before integration.

aseth1:
Just to clarify point 1: The setting of the `_system` member of the `Manager` was a way for CMC to integrate the `CMCActuatorSystem` and not a Model's underlying `MultibodySystem`.  I believe the purpose of a `Manager` is to manage the integration of `Model` dynamics and this was a misuse.

chrisdembia:
It may be good to make sure the same integrator type is used for both systems.

aseth1:
> It may be good to make sure the same integrator type is used for both systems.

I agree, but they are currently not the same 😿.  Hard coded within `VectorFunctionForActuators`:
```cpp
    _integrator = new SimTK::RungeKuttaMersonIntegrator(*aActuatorSystem);
    _integrator->setAccuracy( 5.0e-6 );
    _integrator->setMaximumStepSize(1.0e-3);

    // Don't project constraints while inside the controller
    _integrator->setProjectInterpolatedStates( false );
```

PR #1425, opened 2016-12-01
Remove CMC usage from Manager and other Manager clean up

carmichaelong:
This PR is aimed at the first part of #1405 as well as other clean up of the `Manager` by removing unused variables and methods.


OpenSim/Simulation/Manager/Manager.h
```diff
@@ -95,38 +95,29 @@ class OSIMSIMULATION_API Manager
...
 
-   int _steps;
-   /** Number of integration step tries. */
-   int _trys;
-   /** Maximum number of steps in an integration. */
-   int _maxSteps;
-   /** Flag for signaling a desired halt. */
-   bool _halt;
-   /** Minimum step size. */
-   double _dtMin;
```
chrisdembia: So _dtMin was not used? Is _dtMax used?

chrisdembia: What happens when a user specifies a minimum time step in the ForwardTool XML file?

carmichaelong: _dtMin was not used, but _dtMax is used. As for minimum time step, a search on the code seems to show that CMCTool and RRATool have these (and not ForwardTool), and these are used to update the integrator before getting handed off to the Manager.

aseth1: The properties of the `AbstractTool` are assigned to the `SimTK::Integrator` by `ForwardTool`, `CMCTool` and `RRATool` that is passed into the `Manager`. The private `Manager` members appear to be unused. In fact, where `_dtMax` is being used to test the initial *dt* should be replaced, since `_dtMax` is always `1.0` AFAICT.

aseth1: @carmichaelong I believe it is a bug that the `ForwardTool` is not respecting `_minDT` since it derives from `AbstractTool` and exposes the property in XML. We should probably add that fix (here or separately).

carmichaelong: Noted. Oddly enough, `ForwardTool` does respect `_maxDT`, so I think this is just an oversight.

This might bring up a separate question then: should `AnalyzeTool` derive from `AbstractTool`? `AnalyzeTool` also won't do anything with `_minDT` or `_maxDT`.

carmichaelong: Also it does seem that `_dtMax` is always 1.0.

aseth1: > Oddly enough, ForwardTool does respect _maxDT, so I think this is just an oversight.

 I think so too.

> should AnalyzeTool derive from AbstractTool? 

That's a bigger issue. I think it was wrong to have integrator settings as `AbstractTool` properties. They should have been properties  of the `Manager` that creates the integrator. Then any `Tool` that has a `Manager` property would have the integrator settings where/when it needs it. This is beyond the scope of this PR so leave it for the time being.

aseth1: > Also it does seem that _dtMax is always 1.0.

I advocate for removing `_dtMax` from the Manager at this time. We should make sure that Tools that use the `Manager` relay their `_maxDT` to its `Integrator`.

carmichaelong: Will do. It does seem that `_dtMax` affects the local variable `dt`, but then `dt` actually has no effect on the integration itself. I think that `dt` could be removed safely, and further, the input variable `dtFirst` could also be removed and the signature can be changed safely (searching through the code, I only see uses of `Manager::integrate(state)`. @aseth1 would you be in favor of removing `dt` and changing the signature to remove `dtFirst` as well?

aseth1: Go for it @carmichaelong . I don't like having these variables around only for corner cases and side-effects. If there are none then more reason they should be eliminated.

carmichaelong:
@chrisdembia @aseth1 I've made the changes as per the discussion above. There did happen to be one use of `Manager::integrate(state, dtFirst)`, and I went ahead and removed it since `dtFirst` silently does nothing.

Issue #1419, opened 2016-11-20
TestScale is failing on Appveyor


chrisdembia:
TestScale is failing on [AppVeyor](https://ci.appveyor.com/project/opensim-org/opensim-core/build/5332/job/rfmmjm4j3jtblrji):

```
Step 3: Placing markers on model
Loaded marker file subject01_static.trc (49 markers, 300 frames)
Averaged frames from time 1 to 2 in subject01_static.trc (frames 61 to 121)
Deleted 0 unused markers from model subject01
Frame at (t=0):	total squared error = 0.0411326, marker error: RMS=0.0324759, max=0.0862369 (Top.Head)
Moved markers in model subject01 to match locations in marker file subject01_static.trc
Wrote model file subject01_simbody.osim from model subject01
Wrote marker file Unassigned from model subject01
Loaded model subject01 from file subject01_simbody.osim
Updating Model file from 30000 to latest format...
Loaded model subject01 from file std_subject01_simbody.osim
Marker `Sternum' location: ~[0.112046,0.308398,0.00138853] | ~[0.119011,0.307509,0.0012256] (standard) 
Scale model marker location failed to match the standard
```

This might be due to #1400?

aymanhab:
Likely true though all tests passed before merging, Could it be interaction with some other PR?

aseth1:
#1372 was merged prior to #1400 being merged. I am pretty sure this is the issue. In the interest of time, I think @klshrinidhi can create a new PR that fully switches from `staticPose` ` MarkerData` to the `staticPoseTable`. I believe the difference may be in how the average pose is computed and can now be verified due to the test added in #1400.

PR #1426, opened 2016-12-02
Fix testScale failure

klshrinidhi:
Addressing #1419.

`MarkerData::averageFrames()` computes the average frame for the given time-range and replaces *all the frames (not just the frames averaged)* with the average frame. The PR does the same possibly wrong thing to get the tests to pass.

### DIFF
--- a/OpenSim/Tools/MarkerPlacer.cpp
+++ b/OpenSim/Tools/MarkerPlacer.cpp
@@ -253,17 +253,10 @@ bool MarkerPlacer::processModel(Model* aModel,
                                         });
     const auto avgRow = staticPoseTable.averageRow(_timeRange[0],
                                                    _timeRange[1]);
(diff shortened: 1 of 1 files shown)


aseth1:
LGTM.
We should merge because it replaces exactly what was being done before. We can question the legitimacy of wiping out all the rows in the future, but it appears that is what the `MarkerPlacer` expects.
@aymanhab can you do a quick pass. I think we should merge ASAP so we can update current PRs that are failing.

aymanhab:
Looks fine. Merging to move on. Thanks @klshrinidhi

aseth1:
Thanks @aymanhab!

aseth1:
@opensim-org/dev-team if you have a pending PR that is failing CI due to `testScale` please merge in master to get the this fix. Thanks!

PR #1427, opened 2016-12-07
Throw exception when requesting value of unconnected input.

### LINKED ISSUE #883: If requesting the value of an input that is not wired up, you don't actually get an exception saying that the input was not wired up
Likely introduced by list inputs, etc.

chrisdembia:
This PR makes sure that an exception is thrown when requesting the value of an input that is not connected. The associated test case tests both single-value and list inputs.

Fixes #883.


chrisdembia:
> LGTM

Thank you

aseth1:
@chrisdembia Good to merge after resolving merge conflict.

chrisdembia:
Tests passed; I merged with the permission given by @aseth1 .

PR #1429, opened 2016-12-07
Get rid of TwoFrameLinker::scale() warning on clang.

chrisdembia:
For a while now, clang has given a warning in TwoFrameLinker/WeldConstraint about `TwoFrameLinker::scale()` not being marked as `override`. The base class `Constraint` has a virtual `scale()` function, and this was implemented for WeldConstraint via `TwoFrameLinker::scale()`--in this case, it would make sense to mark `TwoFrameLinker::scale()` as `override`. However, not all base classes of `TwoFrameLinker` have a `scale()` function (e.g., Force, for BushingForces).

This PR renames `TwoFrameLinker::scale()` to `TwoFrameLinker::scaleFrames()`, and `WeldConstraint::scale()` calls `TwoFrameLinker::scaleFrames()`. This gets rid of the warning. I also added notes to the TwoFrameLinker doxygen to help future developers make use of the renamed `scaleFrames()` function.


aseth1:
This solution seems fine to me. :+1:

aseth1:
still looking for a quick secondary review - @aymanhab @carmichaelong?

carmichaelong:
LGTM

chrisdembia:
Thank you.

PR #1430, opened 2016-12-07
Remove obsolete comment and messages

### LINKED ISSUE #1428: Test does not ensure potentially helpful warnings are displayed
On [testComponentInterface.cpp L1042](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Test/testComponentInterface.cpp#L1042), the comment "Should yield warnings about unconnected Inputs" is followed by a call to `world.connect();` but there is no check to ensure the warnings are printed. We are unlikely to catch a PR that accidentally removes these warnings. It would be easy to redirect cout to stringstream (see, e.g., [testReportersWithModel.cpp L55](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Test/testReportersWithModel.cpp#L55)) to at least throw an exception if nothing is printed.

tkuchida:
Fixes #1428

### DIFF
--- a/OpenSim/Common/Test/testComponentInterface.cpp
+++ b/OpenSim/Common/Test/testComponentInterface.cpp
@@ -1039,8 +1039,6 @@ void testInputOutputConnections()
 
     MultibodySystem mbs;
 
-    // Should yield warnings about unconnected Inputs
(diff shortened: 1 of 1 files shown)


aseth1:
:+1:

PR #1434, opened 2016-12-08
{Matlab} c3d reading function

jimmyDunne:
- Read a C3D file
- Get marker table (Vec3) and marker data rate
- Get force table (Vec3) and force data rate
- Define a rotation matrix
- Read marker from table, rotate values, write back to table
- Read force from table, rotate values, write back to table
- Write rotated marker data to TRC
- Flatten force data
- Write flattened, rotated, force data to MOT using STOFileAdapter.


OpenSim/Sandbox/MatlabScripts/c3d_reading.m
```diff
@@ -1,76 +1,75 @@
...
-% % Print the (unrotated) markers to trc file
-% trcfileadapter = TRCFileAdapter();
-% trcfileadapter.write(markers,'test_walking.trc');
-
+% get the number of markers and frames
+nMarkers = markers.getNumColumns();
+nMarkerFrames = markers.getNumRows();
+% get the array of the metadata key names
+MarkerMetaDataKeys = markers.getTableMetaDataKeys();
+MarkerRate = str2num(markers.getTableMetaDataAsString(MarkerMetaDataKeys.get(0)));
```
tkuchida: How would a user find out what index the marker rate is stored in? What's stored in the other keys? Could provide a list (or hyperlink) in the comment.

OpenSim/Sandbox/MatlabScripts/c3d_reading.m
```diff
@@ -1,76 +1,75 @@
...
+
+%% get the force data
+forces = tables.get('forces');
+% get the numner of forces and rows
+nForces = forces.getNumColumns();
+nForceFrames = forces.getNumRows();
+% get the array of the metadata key names
+ForceMetaDataKeys = forces.getTableMetaDataKeys();
+ForceRate = str2num(forces.getTableMetaDataAsString(ForceMetaDataKeys.get(2)));
```
tkuchida: Same comment as above.

OpenSim/Sandbox/MatlabScripts/c3d_reading.m
```diff
@@ -1,76 +1,75 @@
...
-        row[m++] = SimTK::Vec3{pt->GetValues().coeff(f, 0),
-                               pt->GetValues().coeff(f, 1),
-                               pt->GetValues().coeff(f, 2)};
-    }
-
-    
-marker_table.appendRow(0 + f * time_step, row);
 
+%% Rotate marker data
+% make a clean copuy to alter
```
tkuchida: copuy -> copy

OpenSim/Sandbox/MatlabScripts/c3d_reading.m
```diff
@@ -1,76 +1,75 @@
...
+nForces = forces.getNumColumns();
+nForceFrames = forces.getNumRows();
+% get the array of the metadata key names
+ForceMetaDataKeys = forces.getTableMetaDataKeys();
+ForceRate = str2num(forces.getTableMetaDataAsString(ForceMetaDataKeys.get(2)));
 
 %% Define a rotation matix
-Rot = 90;          
+Rot = 90;
 rotationMatrix = [1,0,0;0,cos(Rot*pi/180),-(sin(Rot*pi/180));0,sin(Rot*pi/180),cos(Rot*pi/180)];
```
tkuchida: Could use `cosd(Rot)` rather than `cos(Rot*pi/180)` to reduce clutter.

OpenSim/Sandbox/MatlabScripts/c3d_reading.m
```diff
@@ -80,28 +79,24 @@
         rotatedData = [rotationMatrix'*vectorData']';
```
tkuchida: It's easy to define rotation matrices the wrong way around, and the three transposes here could be confusing. Perhaps apply the transpose to `rotationMatrix` when it's defined and include a comment on L57 indicating whether new data = old data after rotation by Rot about X, or the other way around. The last transpose can be removed because L82 accesses the elements individually anyway.

jimmyDunne: This is actually an error. `rotationMatrix` should not be transposed. Would it be okay to just have `roatedData = [rotationMatrix*vectorData']';` ?

OpenSim/Sandbox/MatlabScripts/c3d_reading.m
```diff
@@ -1,124 +1,158 @@
...
-                               pt->GetValues().coeff(f, 2)};
-    }
-
-    
-marker_table.appendRow(0 + f * time_step, row);
+Rot = -90;
+rotationMatrix = [1,0,0;0,cosd(Rot),-(sin(Rot*pi/180));0,sin(Rot*pi/180),cosd(Rot)];
 
+%% Rotate marker data
+% make a clean copuy to alter
```
aseth1: copuy  -> copy

OpenSim/Sandbox/MatlabScripts/c3d_reading.m
```diff
@@ -1,124 +1,158 @@
...
-        rotatedData = [rotationMatrix'*vectorData']';
+        rotatedData = [rotationMatrix*vectorData']';
 
         % Write the rotated data back to the Vec3TimesSeriesTable
-        % THE BELOW LINES DO NOT WORK AND THERE IS CURRENTLY NO FIX
-        % force.updElt(0,iRow).set(0, rotatedData)
-             
+        elem = Vec3(rotatedData(1),rotatedData(2),rotatedData(3));
+        % set the value of the element
+        force.set(iRow, elem);
```
aseth1: lines 117-129 would seem to be very time consuming, because `rotatedData` and `elem` are allocated over and over again and not because of any math. 

Since you are converting to another `TimeSeriesTableVec3` (and not a Matlab data structure) I would consider doing the multiplies on the OpenSim/SimTK  side. For example, we can do this:
```cpp
        double ang = SimTK::Pi / 3;
        Rotation r(ang, SimTK::XAxis);
        SimTK::Vector_<SimTK::Vec3> packedV(5, Vec3(2));  // a packed Vector of Vec3s
        packedV.dump("packedV");
        auto rotatedV = r*packedV;                        // rotate all Vec3s in packedV at once 
        rotatedV.dump("rotateV");
 ```

I think this would speed things up a lot.

File: OpenSim/Sandbox/MatlabScripts/c3d_reading.m [hunk 8]
```diff
@@ -1,124 +1,158 @@
...
-        auto pt = *it;
-        row[m++] = SimTK::Vec3{pt->GetValues().coeff(f, 0),
-                               pt->GetValues().coeff(f, 1),
-                               pt->GetValues().coeff(f, 2)};
-    }
-
-    
-marker_table.appendRow(0 + f * time_step, row);
+Rot = -90;
+rotationMatrix = [1,0,0;0,cosd(Rot),-(sin(Rot*pi/180));0,sin(Rot*pi/180),cosd(Rot)];
```
aseth1: You're right. I would only `angle2dcm` and Matlab side utilities to demonstrate to users how they could form Rotations in Matlab and use them to transform table data. I agree if we want this to be the main C3D interface in Matlab we should consider doing the transformations with `SimTK::Rotation`.

OpenSim/Sandbox/MatlabScripts/c3d_reading.m
tkuchida: I would simplify to `rotatedData = rotationMatrix * vectorData;`

OpenSim/Sandbox/MatlabScripts/c3d_reading.m
aseth1: Matlab also has some performance [tips](https://www.mathworks.com/help/matlab/matlab_prog/techniques-for-improving-performance.html) worth looking at.

tkuchida: They don't mention that you can get a 10–100× improvement in performance by using C++ 😉

OpenSim/Sandbox/MatlabScripts/c3d_reading.m
jimmyDunne: angle2dcm is only available through the aerospace toolbox. I have 2015b and don't have access to it. 

What slows all these scripts down is Java accessing all the individual elements of the C++ tables. It would be great if OpenSim::Vec3tables had a method to rotate. That would make this script super fast.

OpenSim/Sandbox/MatlabScripts/c3d_reading.m
jimmyDunne: I don't think any of those methods you suggesting are available through Java/Matlab.

OpenSim/Sandbox/MatlabScripts/c3d_reading.m
tkuchida: > It would be great if OpenSim::Vec3tables had a method to rotate. That would make this script super fast.

Can you use a [SimTK::Transform()](https://simtk.org/api_docs/simbody/3.5/singletonSimTK_1_1Transform__.html)? Not sure whether all the necessary classes/methods are wrapped, though.

tkuchida: > What slows all these scripts down is Java accessing all the individual elements of the C++ tables.

Might also try using temporary variables to reduce the number of times you're accessing tables. For example, on L92:
```
% instead of
vectorData = [marker.getElt(0,iRow).get(0)...
              marker.getElt(0,iRow).get(1)...
              marker.getElt(0,iRow).get(2)];
% perhaps
turtle = marker.getElt(0,iRow);
vectorData = [turtle.get(0), turtle.get(1), turtle.get(2)];
```
AFAIK, Matlab won't do such optimizations.

jimmyDunne:
@aseth1 This is now a utility with a test. Can you please review?

jimmyDunne:
@aseth1 just a friendly reminder to review this. Thanks.

OpenSim/Sandbox/MatlabScripts/c3d_reading.m
```diff
@@ -1,124 +1,214 @@
...
-    force = markers.updDependentColumn(forcelabels(iForce+1) );
-
-    % go through each element of the table column, rotate the Vec3, and write
-    % back to the column.
-    for iRow = 0 : forces.getNumRows - 1
-        % get Matlab vector marker position
-        vectorData = [force.getElt(0,iRow).get(0)...
-                      force.getElt(0,iRow).get(1)...
-                      force.getElt(0,iRow).get(2)];
+function rotations = rotateCoordinateSys(axis,rot1,axis2,rot2)
```
aseth1: You can combine the rotations into a single `rotationMatrix`, which this method would provide. You could then return a `SimTK::Rotation` and use its convenience constructors to construct it for various sequences including space or body fixed. Since in L158 you apply each axis rotation one at a time, I think that is effectively space fixed. For example R_bf = *RxRyRz* would be body-fixed but you apply them in order separately  *(Rz(Ry(Rx(D))) = (RzRyRx)D = (R_bf')D* which is space-fixed.

OpenSim/Sandbox/MatlabScripts/c3d_reading.m
```diff
@@ -1,124 +1,214 @@
...
+else
+    rotate = true;
+    rotNames = fieldnames(rotations);
+    nRots    = length(rotNames);
+end
 
-new_table = TimeSeriesTableVec3()
+%% Rotate marker and force data
+if rotate
+    for i = 1 : nRots
```
aseth1: Consider combining to make a single rotation (see L139) which will provide an `nRots` factor speedup.

OpenSim/Sandbox/MatlabScripts/c3d_reading.m
```diff
@@ -1,124 +1,214 @@
...
-marker_table.appendRow(0 + f * time_step, row);
+import org.opensim.modeling.*
 
+nLabels = table.getNumColumns();
+nRows = table.getNumRows();
 
-for iMarker = 0 : length(markerdata) - 1
+table_rotated = table();
+
+for it = 0 : nLabels - 1
```
aseth1: a row is a contiguous vector of nColumns of Vec3s, so it might be more efficient to access a row once and process it then access each row nRows x nColumns times.

OpenSim/Sandbox/MatlabScripts/c3d_reading.m
```diff
@@ -1,124 +1,214 @@
...
     % back to the column.
-    for iRow = 0 : markers.getNumRows - 1
+    for iRow = 0 : nRows - 1
         % get Matlab vector marker position
-        vectorData = [marker.getElt(0,iRow).get(0)...
-                      marker.getElt(0,iRow).get(1)...
-                      marker.getElt(0,iRow).get(2)];
+        vectorData = [table_column.getElt(0,iRow).get(0)...
+                      table_column.getElt(0,iRow).get(1)...
+                      table_column.getElt(0,iRow).get(2)];
```
aseth1: On the C++ side you can multiply a vector of Vec3's, `Vector_<Vec3>` by a `SimTK::Rotation` to effectively transform a whole row at a time. 
 I posted this before, and it works.

```cpp
        double ang = SimTK::Pi / 3;
        Rotation r(ang, SimTK::XAxis);
        SimTK::Vector_<SimTK::Vec3> packedV(5, Vec3(2));  // a packed Vector of 5 Vec3s
        packedV.dump("packedV");
        auto rotatedV = r*packedV;                        // rotate all 5 Vec3s in packedV at once 
        rotatedV.dump("rotatedV");
```
yields
```
packedV
Matrix 5 X 1 3-scalar entries:
{2 2 2}
{2 2 2}
{2 2 2}
{2 2 2}
{2 2 2}
rotatedV
Matrix 5 X 1 3-scalar entries:
{2 -0.73205080756887697113 2.7320508075688776373}
{2 -0.73205080756887697113 2.7320508075688776373}
{2 -0.73205080756887697113 2.7320508075688776373}
{2 -0.73205080756887697113 2.7320508075688776373}
{2 -0.73205080756887697113 2.7320508075688776373}
```
I think we should exploit this to achieve high throughput.

jimmyDunne: Matlab can't perform some of the operations here (`auto rotatedV = r*packedV`). Any ideas on how I can rotate a Vector of Vec3's?

jimmyDunne:
@aseth1 I can now successfully write an Mot file that can be read by the 3.3 GUI. Below is an order of operations that is seen in c3d_reading.m. I have tested this now on 10 c3ds and they all seem to be working correctly. 

- Read a C3D file(s)
- Get marker table (Vec3) and marker data rate
- Get force table (Vec3) and force data rate
- Define a rotation matrix
- Read marker from table, rotate values, write back to table
- Read force from table, rotate values, write back to table
- Write rotated marker data to TRC

- Flatten force data
- Identify and change column headers to comply with older versions
- Change Point data from mm to m
- Change the order (force,point,force,point,torque,torque)
- Check data for NaNs and replace with 0
- Delete any existing Meta Data and write NColumns and nRows to Meta Data
- Write flattened, rotated, header adjusted, point adjusted, NaN replaced, force data to MOT using STOFileAdapter.

jimmyDunne:
@aseth1 can you please review.

Bindings/Java/Matlab/Utilities/osimC3Dconverter.m
```diff
@@ -0,0 +1,278 @@
...
+        end
+    end
+
+    %% change the order of the forces
+    % turn the timetable into a Maltab Struct
+    sdata = osimTableToStruct(forces_flattened);
+    % get the label names
+    labels =  fieldnames(sdata);
+    % find all the labels with string m (including time)
+    imoments = find(~cellfun(@isempty,strfind(labels,'m')));
```
aseth1: This seems to be error prone. You cannot anticipate what labels will have 'm' in them including forces, like *middle_plate*.

Bindings/Java/Matlab/Utilities/osimC3Dconverter.m
```diff
@@ -0,0 +1,278 @@
...
+                point = forces_flattened.getDependentColumnAtIndex(i).get(u);
+                % Divide the point by 1000 and set it back into the colomn
+                forces_flattened.getDependentColumnAtIndex(i).set(u,point/1000);
+            end
+        end
+    end
+
+    %% change the order of the forces
+    % turn the timetable into a Maltab Struct
+    sdata = osimTableToStruct(forces_flattened);
```
aseth1: Why is it necessary to form a MATLAB struct just to convert it back to a Table for writing?

Bindings/Java/Matlab/Utilities/osimC3Dconverter.m
```diff
@@ -0,0 +1,278 @@
...
+            warning(['Nan values found for ' labels2{i} ', at row(s) ' num2str(n') '. Replacing with zeros'])
+            % replace the NaN value with 0
+            sdata.(labels2{i})(n) = 0;
+        end    
+    end
+    
+    
+    
+    %% convert the structure back to an OpenSim time series table
+    forces_reorder = osimTableFromStruct(sdata);
```
aseth1: Roughly how much time is spent converting to and from the Struct?

Bindings/Java/Matlab/tests/testC3Dconverter.m
```diff
@@ -0,0 +1,138 @@
...
+%% Test script for utility function
+osimC3Dconverter('filepath', filepath);
+
+%% Test if trc and mot file were printed.
+assert( exist(fullfile(path,[file '.trc']), 'file') == 2, 'TRC was not printed ')
+assert( exist(fullfile(path,[file '.mot']), 'file') == 2, 'TRC was not printed ')
+
+%% Test rotations
+% test if roations about X, Y, and Z, respectively, maintain the correct
+% components. ie if rotatiing about X, X values remain the same.
```
aseth1: Testing that the component along the axis of rotation is unchanged is necessary but not sufficient as a test. For example, performing no rotation operation would still pass.

Bindings/Java/Matlab/Utilities/osimRotateTableData.m
```diff
@@ -0,0 +1,50 @@
...
+
+%% rotate the elements in each row
+% clone the table.
+T_R = T.clone();
+
+for iRow = 0 : T_R.getNumRows() - 1
+    % get a row from the table
+    R = T_R.getRowAtIndex(iRow);
+    % pass the row to Rotation. 
+    R_R = r.multiply(R);
```
jimmyDunne: Tested the change you suggested, but there is no appreciable performance improvemnt and it makes it much more difficult to read.

Bindings/Java/Matlab/tests/testC3Dconverter.m
jimmyDunne: That is not correct. If there is no rotation, the test would fail.

`assert(mkr_Yvalue_ref ~=  mkr_Yvalue, 'Y axis marker rotation is incorrect ')`

Bindings/Java/Matlab/tests/testC3Dconverter.m
```diff
@@ -0,0 +1,138 @@
...
+
+%% Get the unrotated reference values
+mkr = trc.read(fullfile(path,[file '.trc']));
+ana = sto.read(fullfile(path,[file '.mot']));
+
+mkr_Xvalue_ref = mkr.getDependentColumnAtIndex(0).getElt(238,0).get(0);
```
jimmyDunne: I picked a random, non-zero, row.

Bindings/Java/Matlab/Utilities/osimC3Dconverter.m
```diff
@@ -0,0 +1,278 @@
...
+        if i < 9
+            n = '1';
+        elseif  8 < i < 18
+            n = '2';
+        elseif 17 < i < 27
+           n = '3';
+        end
+
+        if ~isempty(strfind(label,'f'))
+            s = ['ground_force_' n '_v'];
```
jimmyDunne: I was _not_ able to get it to work without 'ground_force'     ¯\_(ツ)_/¯

Bindings/Java/Matlab/Utilities/osimC3Dconverter.m
jimmyDunne: There isn't a convenient method for reordering columns of a table. So I have to do the reordering on the Matlab side.

Bindings/Java/Matlab/Utilities/osimC3Dconverter.m
jimmyDunne: The C3DFileAdapter consistently names the forces {f1,m1,p1,f2,m2,mp2, ....fn,mn,pn}. The only time this would break is if the C3DFileAdapter changed.

Bindings/Java/Matlab/Utilities/osimC3Dconverter.m
jimmyDunne: Almost all the time (huge). But DataTable doesn't have reordering methods.

Bindings/Java/Matlab/Utilities/osimC3Dconverter.m
```diff
@@ -0,0 +1,278 @@
...
+        display('Either no axis or value for rotation has been been set; no rotations will be applied')
+        forces_rot = forces.clone();
+        markers_rot = markers.clone();
+    else
+        display(['Rotating Marker and Force data ' num2str(value) ' about the ' axis ' axis']); 
+        markers_rot = osimRotateTableData(markers, axis, value);
+        forces_rot = osimRotateTableData(forces, axis, value);
+    end
+    
+    %% Print the rotated markers to trc file
```
jimmyDunne: I appreciate what you are suggesting, but that is out of scope for this PR. The utility is to batch process peoples c3d's into .mot and .trc format. I feel splitting this up further is unnecessary at this stage and adds additional complexity when there is already too much.

Bindings/Java/Matlab/Utilities/osimC3Dconverter.m
aseth1: Hmmm can you send me the file you tried without `ground_force`? 
Just to confirm, the file attached didn't preview for you?

Bindings/Java/Matlab/Utilities/osimC3Dconverter.m
aseth1: I investigated further and you are right that the force and point must be sequential in the preview code, [GUI #336](https://github.com/opensim-org/opensim-gui/issues/336) which is the source of this problem. I had tested reordering the moments putting one at the beginning and one at the end, so it still preserved the sequence of columns from `_vx` through to `_pz`. Sorry that I had not probed deeper at that time.

Since we cannot go back to 3.3 to fix, I propose changing the order in which the `C3DFileAdaper` sets its columns since force, point, moment is just as legitimate as force, moment, point and would avoid this whole issue of reordering. I will do that.

Bindings/Java/Matlab/Utilities/osimC3Dconverter.m
aseth1: I agree the utility should be straightforward and not have too many steps. But, I do think it could call a separate utility for writing, which could be equally handy. Also OK to address in a future PR.

Bindings/Java/Matlab/Utilities/osimC3Dconverter.m
aseth1: Not an issue for you, but the doesn't seem to be the right thing to do- to rename columns as "f1", "m1", "p1", etc.... We should respect any names/labels that are supplied d to identify the data in the C3D. If labels are missing, then we can assume generic names like these.

jimmyDunne:
incorporated changes made in #1843. cTest passed, files (.trc & .mot) can be openned in 4.0 and 3.3 GUI. 

@aseth1 ready to review!

Bindings/Java/Matlab/Utilities/osimC3Dconverter.m
```diff
@@ -124,8 +125,8 @@ function osimC3Dconverter(varargin)
     postfix = StdVectorString(); postfix.add('_x');postfix.add('_y');postfix.add('_z');
     forces_flattened = forces_rot.flatten(postfix);
 
-    % update the column names to have v,p,m n them.
-
+    % Tge GUI (~3.3) looks for force labels containing '_v'. We add
```
aseth1: Tge -> The

Bindings/Java/Matlab/Utilities/osimC3Dconverter.m
```diff
@@ -174,61 +175,40 @@ function osimC3Dconverter(varargin)
...
-    labels(imoments) = [];
-    % append the moment labels
-    labels2 = [labels;templabels];
-    % reorder the structure given the new label order
-    sdata = orderfields(sdata, labels2);
-
+    
     %% Check for NaNs
     % If there are any NaNs in the forces data, OpenSim will interpret them as
     % the 'end' of the file. We must set NaNs to zero and report where they are
```
aseth1: Is this true that we cannot read in NaNs correctly? How does the marker reader/writer work then? We should not have to change NaNs to 0 and this is in fact incorrect to do. Please send me a .mot with some  NaNs to confirm this issue. This should be fixed on the API side.

Bindings/Java/Matlab/Utilities/osimC3Dconverter.m
```diff
@@ -0,0 +1,258 @@
+function osimC3Dconverter(varargin)
+% Utility function for converting c3d data to OpenSim format.
+% Inputs to function are pairs of string-string or string-number pairs. 
+% 'filepath', 'path2file'   reads c3d data at path2file.
+% 'axis', 'X'               applies first rotation to axis X (string)
+% 'value', 90               value, in degrees, to rotate data by
+%
```
aseth1: Missed this last time, but what are the Outputs of the converter? A markers.trc and a forces.mot file?

Bindings/Java/Matlab/tests/testC3Dconverter.m
```diff
@@ -47,7 +47,7 @@
 assert( exist(fullfile(path,[file '.mot']), 'file') == 2, 'TRC was not printed ')
```
aseth1: Should be `MOT was not printed` ?

Bindings/Java/Matlab/tests/testC3Dconverter.m
aseth1: Right, `~=` will require that it changed, that's good! But it can still be wrong and still pass if it is changed at all? E.g. rotated 90 instead of -90 would pass, right?

aseth1:
@jimmyDunne my comments were minor. I don't think it warranted closing the PR. The only big issue was handling of NaNs and I think that should be done on the API side (so not on your plate).

PR #1435, opened 2016-12-08
{Matlab} create actuator script

jimmyDunne:
- Find a model and read to into memory
- Loop through all the coordinates
       - Skip if coordinate is constrained
       - Determine if the coordinate is connected to ground
- Connected to ground and Rotational -> add TorqueAcuator
- Connected to ground and Translational -> add PointAcuator
- For all other coordinates, add a CoordinateActuator
- Write ForceSet to file

### DIFF
--- a/OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
+++ b/OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
@@ -1,12 +1,11 @@
-function createActuatorsFile
 %createActuatorsFile   make and Print an OpenSim Actuator File from a Model
 %
-%  createActuatorsFile attempts to make a template actuators file that can
-%  be used in Static Optimization, RRA and CMC. This tries to identify the
-%  coordinates  that are connected to ground and place point or torque
+%  createActuatorsFile makes a template actuators file that can
+%  be used in Static Optimization, RRA and CMC. This identifies the
+%  coordinates that are connected to ground and places point or torque
 %  actuators on translational or rotational coordinates, respectively. All
-%  other coordiantes will get coordinate actuators. Any locked, prescribed
-%  or constrained coordinates will be ignored.
+%  other coordiantes will get coordinate actuators. Any constrained 
+%  coordinates will be ignored.
 
 % ----------------------------------------------------------------------- %
 % The OpenSim API is a toolkit for musculoskeletal modeling and           %
@@ -30,14 +29,27 @@ function createActuatorsFile
 % permissions and limitations under the License.                          %
 % ----------------------------------------------------------------------- %
 
-% Author: James Dunne, Chris Dembia, Tom Uchida,
+% Author: James Dunne, Tom Uchida, Chris Dembia, 
 % Ajay Seth, Ayman Habib, Shrinidhi K. Lakshmikanth, Jen Hicks.
+
+function createActuatorsFile(varargin)
 
 % Import OpenSim Libraries
 import org.opensim.modeling.*
 
-% open dialog boxes to select the model
-[filename, pathname] = uigetfile('*.osim', 'Select an OpenSim Model File');
+if isempty(varargin)
+    % open dialog boxes to select the model
+    [filename, pathname] = uigetfile('*.osim', 'Select an OpenSim Model File');
+elseif nargin == 1
+    if exist(varargin{1}, 'file') == 2
+        [pathname,filename,ext] = fileparts(varargin{1});
+        filename = [filename ext];
+    else 
+        error(['Input file is invalid or does not exist']);
+    end
+else
+    error(['Number of inputs is > 1. Function only takes a single filepath']);
+end
 
 % get the model path
 modelFilePath = fullfile(pathname,filename);
@@ -66,7 +78,8 @@ for iCoord = 0 : nCoord - 1
 
     % get a reference to the current coordinate
     coordinate = coordSet.get(iCoord);
-    % If the coodinate is constrained, don't add an actuator
+    % If the coodinate is constrained (locked or prescribed), don't 
+    % add an actuator
     if coordinate.isConstrained(state)
         continue
     end
@@ -78,16 +91,17 @@ for iCoord = 0 : nCoord - 1
 
     % If the coordinates parent body is connected to ground, we need to
     % add residual actuators (torque or point).
-    if strmatch(parentName, model.getGround.getName() )
+    if strcmp(parentName, model.getGround.getName() )
 
-        % if the joint type is custom or free
+        % Custom and Free Joints have three translational and three
+        % rotational coordinates. 
         if strcmp(joint.getConcreteClassName(), 'CustomJoint') | strcmp(joint.getConcreteClassName(), 'FreeJoint')
                % get the coordainte motion type
-               motion = char(coordinate.getMotionType);
-               % to get the axis value for the coordiante, we need to drill
+               motion = char(coordinate.getMotionType());
+               % to get the axis value for the coordinate, we need to drill
                % down into the coordinate transform axis
                eval(['concreteJoint = ' char(joint.getConcreteClassName()) '.safeDownCast(joint);'])
-               sptr = concreteJoint.getSpatialTransform;
+               sptr = concreteJoint.getSpatialTransform();
                for ip = 0 : 5
                   if strcmp(char(sptr.getCoordinateNames().get(ip)), char(coordinate.getName))
                         sptr.getTransformAxis(ip).getAxis(axisValues);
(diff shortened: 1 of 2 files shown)


OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,156 @@
+function createActuatorsFile
```
tkuchida: Perhaps accept optional argument containing path/filename of model (e.g., so this function could be used in a loop without prompting the user with a dialog box).

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,156 @@
...
+%createActuatorsFile   make and Print an OpenSim Actuator File from a Model
+%
+%  createActuatorsFile attempts to make a template actuators file that can
```
tkuchida: "attempts" on L4 and "tries" on L5 imply that the function may fail. In what situations will failure occur? What are the prerequisites for success?

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,156 @@
...
+state = model.initSystem();
+
+optimalForce = 1;
+
+%% Start going through the coordinates, creating an actuator for each
+for iCoord = 0 : nCoord - 1
+
+    % get a reference to the current coordinate
+    coordinate = coordSet.get(iCoord);
+    % If the coodinate is constrained, don't add an actuator
```
tkuchida: Perhaps mention that "constrained" means the Coordinate is either locked, prescribed, or dependent on other coordinates.

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,156 @@
...
+    end
+
+    % get the joint, parent and child names for the coordiante
+    joint = coordinate.getJoint();
+    parentName = joint.getParentFrame().getName();
+    childName = joint.getChildFrame().getName();
+
+    % If the coordinates parent body is connected to ground, we need to
+    % add residual actuators (torque or point).
+    if strmatch(parentName, model.getGround.getName() )
```
tkuchida: `strmatch` -> `strcmp` (see "Note" at the top of [strmatch documentation](https://www.mathworks.com/help/matlab/ref/strmatch.html))

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,156 @@
...
+
+        % if the joint type is custom or free
+        if strcmp(joint.getConcreteClassName(), 'CustomJoint') | strcmp(joint.getConcreteClassName(), 'FreeJoint')
+               % get the coordainte motion type
+               motion = char(coordinate.getMotionType);
```
tkuchida: `getMotionType` -> `getMotionType()` (even if Matlab doesn't demand them, trailing parentheses are preferred here so it's clear that you're calling a method)

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,156 @@
...
+               % to get the axis value for the coordiante, we need to drill
```
tkuchida: coordiante -> coordinate

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,156 @@
...
+    % get the joint, parent and child names for the coordiante
+    joint = coordinate.getJoint();
+    parentName = joint.getParentFrame().getName();
+    childName = joint.getChildFrame().getName();
+
+    % If the coordinates parent body is connected to ground, we need to
+    % add residual actuators (torque or point).
+    if strmatch(parentName, model.getGround.getName() )
+
+        % if the joint type is custom or free
```
tkuchida: This comment is not useful. Perhaps provide a short explanation of what the code block does, or just delete the comment.

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,156 @@
...
+        if strcmp(joint.getConcreteClassName(), 'CustomJoint') | strcmp(joint.getConcreteClassName(), 'FreeJoint')
+               % get the coordainte motion type
+               motion = char(coordinate.getMotionType);
+               % to get the axis value for the coordiante, we need to drill
+               % down into the coordinate transform axis
+               eval(['concreteJoint = ' char(joint.getConcreteClassName()) '.safeDownCast(joint);'])
+               sptr = concreteJoint.getSpatialTransform;
```
tkuchida: `getSpatialTransform()`

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,156 @@
...
+                  end
+               end
+
+
+               % make a torque actuator if a rotational coordinate
+               if strcmp(motion, 'Rotational')
+                   newActuator = TorqueActuator(joint.getParentFrame(),...
+                                         joint.getParentFrame(),...
+                                         axisValues,...
+                                         1);
```
tkuchida: Prefer `true` rather than `1` so it's clear that the last argument is a `bool`. A comment would be helpful, too (though, since the last argument defaults to `true` anyway, you could just omit it).

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,156 @@
...
+
+               % make a point actuator if a translational coordainte.
```
tkuchida: coordainte -> coordinate

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,156 @@
...
+             elseif strcmp(motion, 'translational')
+                    % make a new Point actuator
+                    newActuator = PointActuator();
+                    % set the body
+                    newActuator.set_body(char(joint.getChildFrame().getName()))
+                    % set <point>      -0.07243760       0.00000000       0.00000000 </point>
```
tkuchida: Presumably, these comments should either be deleted or stated in a more general form.

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,156 @@
...
+                    newActuator.set_point(massCenter)
+                    % set <point_is_global> false </point_is_global>
+                    newActuator.set_point_is_global(0)
```
tkuchida: `0` -> `false`

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,156 @@
...
+                    % set <direction>      -0.00000000       1.00000000      -0.00000000 </direction>
+                    newActuator.set_direction(axisValues)
+                    % set <force_is_global> true </force_is_global>
+                    newActuator.set_force_is_global(1)
```
tkuchida: `1` -> `true`

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,156 @@
...
+              else % something else that we don't support right now
+                    error(['Motion Type ' char(motion) 'not supported yet'])
+              end
+        else % if the joint type is not free or custom, just add coordinate actuators
+                % make a new coordinate actuator for that coordinate
+                newActuator = CoordinateActuator();
```
tkuchida: Isn't it necessary to set the `coordinate` property here?

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,156 @@
...
+         % coordinate actuator.
+         newActuator = CoordinateActuator( char(coordinate.getName) );
+    end
+
+    % set the optimal force for that coordinate
+    newActuator.setOptimalForce(optimalForce);
+    % set the actuator name
+    newActuator.setName( coordinate.getName() );
+    % set the optimal force for that coordinate
+    newActuator.setOptimalForce(optimalForce);
```
tkuchida: The optimal force is already set on L134.

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
```diff
@@ -0,0 +1,156 @@
...
+    % set min and max controls
+    newActuator.setMaxControl(Inf)
+    newActuator.setMinControl(-Inf)
+
+    % append the new acuator onto the empty force set
```
tkuchida: acuator -> actuator

tkuchida: The force set will be empty only on the first iteration. Perhaps "append the new actuator to the force set"?

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
jimmyDunne: The properties get set outside of the if statement L146-L153 set properties for all actuators regardless of type (coordinate, point, torque).

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
jimmyDunne: From memory, `Append` used to cause Matlab crashing. When should we use `Append`, `cloneAndAppend`? This seems confusing.

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
tkuchida: `newActuator = CoordinateActuator();` creates a CoordinateActuator but doesn't specify which of the model's coordinates is being actuated. L146–153 set the optimal force (L147), the name of the actuator object (L149), and the min/max controls (L151–152), but not which coordinate to actuate.

OpenSim/Sandbox/MatlabScripts/createActuatorsFile.m
tkuchida: I'm talking about the comment. You're saying that the next line appends the new actuator to the *empty* ForceSet, but the ForceSet will not be empty when you're adding the second actuator.

PR #1437, opened 2016-12-08
{Matlab} table reading and writing utility

jimmyDunne:
Scripts to changing between Matlab and OpenSim data types


OpenSim/Sandbox/MatlabScripts/osimTableFromStruct.m
```diff
@@ -0,0 +1,127 @@
...
+    row = RowVector(nLabels, 1);
+
+    for iRow = 1 : nRows
+
+        % create and fill a row of data at a time
+        for iCol = indexArray
+
+            rowdata = structdata.(labelnames{iCol})(iRow,:);
+
+            row.set(iCol-1, rowdata);
```
tkuchida: If `timeIndex == 1`, the first entry in `indexArray` would be 2 and the first iteration over `iCol` would be setting element 1 of `row` (not 0). Suppose we have the following general case:
<pre>
nLabels = 5
timeIndex = 3
then... indexArray = [1, 2, 4, 5]
</pre>
Then this `for` loop runs four times (with `iCol = 1, 2, 4, 5`) and you're setting elements `iCol-1 = 0, 1, 3, 4` rather than `0, 1, 2, 3`. You should be iterating over the *index into* the variable named `indexArray`, as follows:
```
for idx = 1:length(indexArray)
    iCol = indexArray(idx);
    rowdata = ...
    row.set(idx-1, rowdata);
end
```
Test cases (e.g., calling `osimTableFromStruct()` followed by `osimTableToStruct()` and calling `osimTableToStruct()` followed by `osimTableFromStruct()`) could be helpful to ensure these functions will be reliable.

OpenSim/Sandbox/MatlabScripts/osimTableFromStruct.m
```diff
@@ -0,0 +1,127 @@
...
+        end
+
+        table.appendRow(iRow-1, row);
+        % set the time value
+        timeColumn.set(iRow-1, structdata.(labelnames{timeIndex})(iRow) );
+    end
+
+else nCol == 3
+    % if the data is in triples, then build a table of Vec3's
```
tkuchida: This comment is confusing since the description for the first branch (`nCol == 1`) is "the data is in doubles" (double-precision, not pairs). Perhaps triples -> triplets?

OpenSim/Sandbox/MatlabScripts/osimTableToStruct.m
```diff
@@ -0,0 +1,91 @@
+function data = osimTableToStruct(table)
+%% // Convert Matlab Struct to OpenSim time Series Table
```
tkuchida: Description is backwards (the function converts from table to struct).

OpenSim/Sandbox/MatlabScripts/osimTableToStruct.m
```diff
@@ -0,0 +1,91 @@
...
+%  Input is an OpenSim TimesSeriesTable
+%
+%  Output is a Maltab stucture where data.label = n X 1 or n x 3 array
+%  ie structdata.LASI = [2 3 4; 4 5 6, ...
```
tkuchida: ie -> e.g.,

OpenSim/Sandbox/MatlabScripts/osimTableToStruct.m
```diff
@@ -0,0 +1,91 @@
...
+
+% ----------------------------------------------------------------------- %
+% The OpenSim API is a toolkit for musculoskeletal modeling and           %
+% simulation. See http://opensim.stanford.edu and the NOTICE file         %
+% for more information. OpenSim is developed at Stanford University       %
+% and supported by the US National Institutes of Health (U54 GM072970,    %
+% R24 HD065690) and by DARPA through the Warrior Web program.             %
+%                                                                         %
+% Copyright (c) 2005-2012 Stanford University and the Authors             %
```
tkuchida: 2012 -> 2016

OpenSim/Sandbox/MatlabScripts/osimTableToStruct.m
```diff
@@ -0,0 +1,91 @@
...
+
+% Author: James Dunne, Shrinidhi K. Lakshmikanth, Chris Dembia, Tom Uchida,
+% Ajay Seth, Ayman Habib, Jen Hicks.
+
+%%
+import org.opensim.modeling.*
+
+%%
+nCol = table.getNumColumns;
+nRow = table.getNumRows;
```
tkuchida: Prefer `getNumColumns()` and `getNumRows()` (with trailing parens) so it's clear these are method calls.

OpenSim/Sandbox/MatlabScripts/osimTableToStruct.m
```diff
@@ -0,0 +1,91 @@
...
+% ----------------------------------------------------------------------- %
+
+% Author: James Dunne, Shrinidhi K. Lakshmikanth, Chris Dembia, Tom Uchida,
+% Ajay Seth, Ayman Habib, Jen Hicks.
+
+%%
+import org.opensim.modeling.*
+
+%%
+nCol = table.getNumColumns;
```
tkuchida: L37–38 should appear below the type checking (L44–51). If the input argument is of the incorrect type, an error may be thrown here already.

OpenSim/Sandbox/MatlabScripts/osimTableToStruct.m
```diff
@@ -0,0 +1,91 @@
...
+nRow = table.getNumRows;
+
+%% make an empty data structure
+data = struct();
+
+%%
+if strcmp( char(table.getClass), 'class org.opensim.modeling.TimeSeriesTableVec3')
```
tkuchida: `getClass` -> `getClass()` preferred

OpenSim/Sandbox/MatlabScripts/osimTableToStruct.m
```diff
@@ -0,0 +1,91 @@
+function data = osimTableToStruct(table)
+%% // Convert Matlab Struct to OpenSim time Series Table
+%  Input is an OpenSim TimesSeriesTable
```
tkuchida: "TimesSeriesTable" -> "TimesSeriesTable or TimesSeriesTableVec3"

OpenSim/Sandbox/MatlabScripts/osimTableToStruct.m
```diff
@@ -0,0 +1,91 @@
...
+data = struct();
+
+%%
+if strcmp( char(table.getClass), 'class org.opensim.modeling.TimeSeriesTableVec3')
+    rowdata = zeros(nRow, 3);
+elseif strcmp( char(table.getClass), 'class org.opensim.modeling.TimeSeriesTable')
+    rowdata = zeros(nRow, 1);
+else
+    disp(['Input is of type ' char(table.getClass)'])
+    error(['This function only deals with TimeSeriesTable or TimeSeriesTableVec3']);
```
tkuchida: "deals with... or..." -> "converts... and..."

OpenSim/Sandbox/MatlabScripts/osimTableToStruct.m
```diff
@@ -0,0 +1,91 @@
...
+end
+
+%%
+for iCol = 0 : nCol -1
+
+    if strcmp( char(table.getClass), 'class org.opensim.modeling.TimeSeriesTableVec3')
```
tkuchida: Would be more efficient to do this check once (i.e., before L54) and check a flag in the `for` loop to avoid performing a string comparison `nCol` times. Might not make a noticeable difference if running once, but users may call this function in a loop thousands of times.

OpenSim/Sandbox/MatlabScripts/osimTableToStruct.m
```diff
@@ -0,0 +1,91 @@
...
+
+        for iRow = 0 : nRow - 1
+            rowdata(iRow+1,1) = table.getDependentColumnAtIndex(iCol).getElt(iRow,0);
+        end
+    end
+
+    col_label  = char(table.getColumnLabels.get(iCol));
+
+    if isempty(strfind(col_label, '/'))
+        eval(['[data.' col_label '] = rowdata;']);
```
tkuchida: I think `eval` is very slow. From [Matlab documentation](https://www.mathworks.com/help/matlab/matlab_prog/techniques-for-improving-performance.html): "Avoid functions such as eval, evalc, evalin, and feval(fname). Use the function handle input to feval whenever possible. Indirectly evaluating a MATLAB expression from text is computationally expensive."

OpenSim/Sandbox/MatlabScripts/osimTableToStruct.m
```diff
@@ -0,0 +1,91 @@
...
+    else
+       temp = strfind(col_label, '/');
+       new_col_label = col_label(temp(end-1)+1:temp(end)-1);
```
tkuchida: If `col_label` is `a/b/c`, it looks like you're just using `b` as the field name. That strategy won't work in general since field names must be unique and there could be more than one `c` (Output) under each `b` (Component). Perhaps replace `/` with `_`?

OpenSim/Sandbox/MatlabScripts/osimTableFromStruct.m
jimmyDunne: There was definitely a bug at `for iCol = indexArray`. It should have been `iCol = 1 : indexArray`. 

I have tested and confirmed that the column length is the correct size and that the values are in the correct order. 

I will write a test case and add it to the PR.

tkuchida: Has the meaning of a `for` loop in the form `for i = 1 : [list]` changed in a recent version of Matlab? If so, we should check the Matlab version number (`ver('matlab')`); on my (old) version, it doesn't seem like this algorithm will work correctly. There are two cases to consider:
1. `timeIndex = 1` and thus `indexArray = [2, 3, ...]`, in which case `for iCol = 1 : indexArray` will iterate twice (`iCol = 1` and `iCol = 2`).
2. `timeIndex ~= 1` and thus `indexArray = [1, 2, ..., timeIndex-1, timeIndex+1, timeIndex+2, ...]` in which case `for iCol = 1 : indexArray` will iterate only once (`iCol = 1`).

jimmyDunne:
@aseth1 could you review?

jimmyDunne:
@aseth1 Can you please review.

PR #1441, opened 2016-12-09
Macro-generated connectInput_name()

chrisdembia:
This PR implements steps 3 and 4 of [this devprop](https://github.com/opensim-org/opensim-core/blob/master/DevelopmentProposals/componentConnectInterface/componentConnectInterface.md). This PR does *not* create similar macros for sockets (e.g., `connectSocket_name()`); that will come after the connector->socket rename.

Most of the changes are just basically find/replace; the meat of the changes are in ComponentConnector.h. View the updated doxygen [here](http://myosin.sourceforge.net/1441); here are some snippets from the Reporter class:

image

image

image

I worked on this PR as part of our effort to have an alpha release for Tuesday.

The main concern/question I have is that these new methods are only available on the concrete class; so for example, `model.getComponent("reporter").connectInput_inputs()` wouldn't work.

### DIFF
--- a/OpenSim/Common/Reporter.h
+++ b/OpenSim/Common/Reporter.h
@@ -113,18 +113,52 @@ public:
     OpenSim_DECLARE_LIST_INPUT(inputs, InputT, SimTK::Stage::Report,
         "Variable list of quantities to be reported.");
 
-    //=============================================================================
+    //==========================================================================
     // PUBLIC METHODS
-    //=============================================================================
+    //==========================================================================
 
     // Allow overloading of updInput
     using AbstractReporter::updInput;
 
+    /** Connect an output (single-valued or list) to this reporter. 
+    The output must be of type InputT.
+    If the output is a list output, this connects to all of the channels of the
+    output. You can optionally provide an alias that will be used by this
+    component to refer to the output; the alias will be used for all channels
+    of the output.                                          
+    @code
+    auto* reporter = new ConsoleReporter();
+    auto* src = new TableSource();
+    reporter->addToReport(src->getOutput("all_columns"));
+    @endcode
+    This method is equivalent to
+    connectInput_inputs(const AbstractOutput&, const std::string&). */
+    void addToReport(const AbstractOutput& output,
+                     const std::string& alias = "") {
+        connectInput_inputs(output, alias);
+    }
+
+    /** Connect an output channel to this reporter.
+    The output channel must be of type InputT.
+    You can optionally provide an alias that will be used by this component to
+    refer to the channel.
+    @code
+    auto* reporter = new ConsoleReporter();
+    auto* src = new TableSource();
+    reporter->addToReport(src->getOutput("column").getChannel("ankle"));
+    @endcode
+    This method is equivalent to
+    connectInput_inputs(const AbstractChannel&, const std::string&). */
+    void addToReport(const AbstractChannel& channel,
+                     const std::string& alias = "") {
+        connectInput_inputs(channel, alias);
+    }
+
     /** Convenience method that can be used in place of `updInput("inputs")`. 
     @code
     auto* reporter = new ConsoleReporter();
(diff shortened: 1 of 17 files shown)


OpenSim/Common/Reporter.h
```diff
@@ -113,18 +113,52 @@ class Reporter : public AbstractReporter {
...
+        connectInput_inputs(channel, alias);
+    }
+
     /** Convenience method that can be used in place of `updInput("inputs")`. 
     @code
     auto* reporter = new ConsoleReporter();
-    auto* src = new DataSource();
-    reporter->updInput().connect(src->getOutput("outputName"));
+    auto* src = new TableSource();
+    reporter->updInput().getLabel();
```
tkuchida: Couldn't you do `reporter->getInput().getLabel();` here? Perhaps provide an example where a writable reference is necessary?

chrisdembia:
> couldn't you do model.getComponent<Reporter>("reporter").connectInput_inputs()?

Ah good point. Yes, this will work in C++. The solution in MATLAB is a little clunkier.

OpenSim/Common/Reporter.h
chrisdembia: Reporter does not have a `getInput()` (with no arguments) yet; but I can add that as part of this PR. That would definitely look better.

chrisdembia:
Thanks for the review, @tkuchida

OpenSim/Common/Reporter.h
aseth1: I feel we added the convenience no argument `updInput()` to make the `connect() ` call less confusing. But now that we have `addToReport` to make the connection for the user, I think it would be simpler to get rid of `updInput()`.

OpenSim/Common/Test/testComponentInterface.cpp
```diff
@@ -696,16 +696,16 @@ void testMisc() {
...
     theWorld.add(reporter);
 
     MultibodySystem system3;
     cout << "Building theWorld's system:" << endl;
     theWorld.buildUpSystem(system3);
 
     // Connect our state variables.
-    foo.updInput("fiberLength").connect(bar.getOutput("fiberLength"));
-    foo.updInput("activation").connect(bar.getOutput("activation"));
+    foo.connectInput_fiberLength(bar.getOutput("fiberLength"));
```
aseth1: Looks good!

OpenSim/Simulation/Test/testReportersWithModel.cpp
```diff
@@ -46,10 +46,8 @@ void testConsoleReporerLabels() {
     // Create ConsoleReporter, and connect Outputs without and with alias.
     auto* reporter = new ConsoleReporter();
     reporter->set_report_time_interval(1.);
-    reporter->updInput("inputs").connect(
-        slider->getCoordinate().getOutput("value") );
-    reporter->updInput("inputs").connect(
-        slider->getCoordinate().getOutput("value"), "height" );
+    reporter->addToReport(slider->getCoordinate().getOutput("value"));
+    reporter->addToReport(slider->getCoordinate().getOutput("value"), "height");
```
aseth1: Should we support the same `Output` connect to the same list `Input` in multiple slots?

aseth1:
Overall LGTM. I only had a minor suggestion and a question. Neither have to be resolved in this PR.

OpenSim/Simulation/Test/testReportersWithModel.cpp
chrisdembia: Good question; I have opened #1445 to discuss this.

chrisdembia:
@tkuchida @aseth1 I have addressed the issue that both of you mentioned. Ready for review.

tkuchida:
LGTM :+1:

chrisdembia:
Thank you!

PR #1444, opened 2016-12-12
Fix visualization failure for disabled muscles

### LINKED ISSUE #1111: [visualizer] Visualizer crashes if the model contains disabled muscles

aseth1:
Fixes #1111.

Since we are renaming `Force::isDisabled` to `Force::appliesForce`  it now implies that the Muscle is still present but simply that it does not apply forces to the system.

The fix then is to allow the muscle (`GeometryPath`) to draw itself regardless if it is applying force or not. This could be helpful for visually debugging muscle path issues without bringing a simulation to a halt.

The related bug was due to the `GeometryPath::generateDecorations()` accessing the *current_path*
cache variable directly, when it was invalid and required to be recomputed. This has been corrected.

Should be a quick review.

### DIFF
--- a/OpenSim/Simulation/Model/GeometryPath.cpp
+++ b/OpenSim/Simulation/Model/GeometryPath.cpp
@@ -119,6 +119,10 @@ generateDecorations(bool fixed, const ModelDisplayHints& hints,
 {        
     // There is no fixed geometry to generate here.
     if (fixed) { return; }
+    
+    // Ensure that the state has been realized to Stage::Dynamics to give
+    // clients of this path a chance to calculate meaningful color information.
+    getModel().realizeDynamics(state);
 
     const Array<PathPoint*>& pathPoints = getCurrentPath(state);
 
--- a/OpenSim/Simulation/Model/GeometryPath.cpp
+++ b/OpenSim/Simulation/Model/GeometryPath.cpp
@@ -184,7 +184,7 @@ void GeometryPath::constructProperties()
 
     constructProperty_PathWrapSet(PathWrapSet());
     
-    Vec3 defaultColor = SimTK::White;
+    Vec3 defaultColor = SimTK::Gray;
     constructProperty_default_color(defaultColor);
 }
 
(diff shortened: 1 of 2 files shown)


aseth1:
The bug was reproduced in *ExampleMain*'s `TugOfWar_Complete.cpp` by adding these lines [295-298]:
```cpp
        // set use visualizer to true to visualize the simulation live
        osimModel.setUseVisualizer(true); // set to true for this test

        muscle1->set_isDisabled(true); // added for test
```

aymanhab:
Thanks @aseth1 the change looks good and rather safe,  ideally we incorporate this into testVisualization to make sure it doesn't pop up again but that could be done in a separate PR due to timeline. In your testing did the muscle show up at all and if so, what color? Not obvious from the code.  Thank you.

aseth1:
@aymanhab the muscle appears white but I agree with @tkuchida that this could be made SimTK::Gray

chrisdembia:
I agree about using Grey. Other than that, LGTM.

aseth1:
@aymanhab @tkuchida @chrisdembia I restored the realization to Stage::Dynamics to get the correct path color (which was causing testVisualization to fail) and made the default color Gray (when not applying force).

aymanhab:
LGTM if tested visually with both an enabled and a disabled muscle pending updating the test case, Do you conjecture the solution working for other components that utilize GeometryPath (e.g. Ligament, ClutchSpring, etc.) Thanks.

aseth1:
@aymanhab yes I did verify visually both with temporary change to `TugOfWar_Complete.cpp` (see above) and in `testVisualization`.  This fix is to `GeometryPath` and should resolve the same issue for all Components that consume it (nothing specific to Muscle in the change). That's not to say that they don't have other issues related to visualization and their handling of `isDisabled`.

aymanhab:
Sounds great. Thanks @aseth1 Feel free to merge when the tests pass.

aseth1:
Timeout failure on OSX is unrelated to these changes. I will merge. Thanks @aymanhab @tkuchida @chrisdembia.

PR #1452, opened 2016-12-15
[CMake] Update comment to help users building Matlab wrapping

tkuchida:
If building Python wrapping, it is clear that the `BUILD_PYTHON_WRAPPING` flag will need to be true, but there is no analogous `BUILD_MATLAB_WRAPPING` flag and the `BUILD_JAVA_WRAPPING` flag currently mentions only the GUI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -371,9 +371,9 @@ if(WIN32 AND ${CMAKE_C_COMPILER} MATCHES "cl")
 
 endif()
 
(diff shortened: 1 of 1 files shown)


chrisdembia:
👍

aymanhab:
Thanks @tkuchida and @chrisdembia Is there a place to indicate that you can't build debug with Python bindings?

chrisdembia:
It may be that if you build a debug version of python, then you can build debug python bindings. I'm not sure.

chrisdembia:
Definitely something that we should clarify for users, though.

PR #1454, opened 2016-12-16
Instruct users to check Matlab's JVM version

tkuchida:
Matlab's JVM version must be at least 1.7 (to agree with the version required by CMake). It's possible to build against an older version of JDK and modify the CMake file (or, alternatively, to update Matlab's JVM). I don't think it's necessary to explain all that in the README, but I do think it would be helpful to let users know about this issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,7 +194,9 @@ On Windows using Visual Studio
       intermediate; good for TortoiseSVN users;
     * [GitHub for Windows](https://windows.github.com/), easiest.
 * **Bindings** (optional): [SWIG](http://www.swig.org/) 3.0.6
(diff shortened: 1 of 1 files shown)


chrisdembia:
I have tried to spell MATLAB as all-caps, as this is what MathWorks seems to prefer (https://www.mathworks.com/products/new_products/latest_features.html?s_tid=hp_spot_R2016b_0914). 

I'm happy with merging this as-is, or with a change to the capitalization.

tkuchida:
> I have tried to spell MATLAB as all-caps, as this is what MathWorks seems to prefer

Ok, will do.

tkuchida:
> Thanks for cleaning up other capitalization as well.

Seemed like a capital idea 🎩 Thanks for merging.

PR #1460, opened 2016-12-16
[WIP] Script to convert all Thelen2003 muscles in a model with Millard2012A…

aymanhab:
…ccelerationMuscle muscles.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/1460)


aymanhab:
Current incarnation keeps all defaults in Millard muscle model, not sure if that's good enough, in essence all similarly named Properties (including GeometryPath) are maintained and the rest is default. Is there a better option @tkuchida ?

Bindings/Python/examples/convert_muscle_type.py
```diff
@@ -0,0 +1,111 @@
...
+# implied. See the License for the specific language governing            #
+# permissions and limitations under the License.                          #
+# ----------------------------------------------------------------------- #
+
+# convert_muscle_type.py
+# Author: Ayman Habib
+# ------------------------------------------------------------------------#
+# ABSTRACT: This short piece of OpenSim python API example demonstrates   #
+# how to load a model in the OpenSim python API, cycle through the        #
+# the muscles in the model and replace those of type Thelen2003 with      #
```
tkuchida: "the the" -> "the"

Bindings/Python/examples/convert_muscle_type.py
```diff
@@ -0,0 +1,111 @@
...
+# the more modern (true to publication) Millard2012AccelerationMuscle     #
```
tkuchida: We don't have comprehensive tests for `Millard2012AccelerationMuscle` and it has been marked "experimental" in the doxygen. I think this script should convert to `Millard2012EquilibriumMuscle` instead.

Bindings/Python/examples/convert_muscle_type.py
```diff
@@ -0,0 +1,111 @@
...
+    to_muscle.copyProperty_ignore_activation_dynamics(from_muscle)
+    # Properties inherited from PathActuator
+    to_muscle.set_GeometryPath(from_muscle.get_GeometryPath().clone())
+    to_muscle.copyProperty_optimal_force(from_muscle)
+    # Properties from ScalarActuator
+    to_muscle.copyProperty_max_control(from_muscle)
+    to_muscle.copyProperty_min_control(from_muscle)
+    # Properties from Force
+    to_muscle.copyProperty_appliesForce(from_muscle)
+    # print (to_muscle.dump())
```
tkuchida: There are also properties further up the hierarchy under `Component` (`connectors` and `components`). Should they be copied over as well?

Bindings/Python/examples/convert_muscle_type.py
```diff
@@ -0,0 +1,111 @@
...
+
+# Create ForceSet to contain the new muscles
+millard_muscles = osim.ForceSet()
+# Ownership will be transfered to the converted_model eventually
+millard_muscles.setMemoryOwner(False)
+
+
+for thelen_musc in thelen_muscles:
+    print ("Processing muscle:"+thelen_musc.getName())
+    millard_musc = osim.Millard2012AccelerationMuscle()
```
tkuchida: Millard2012AccelerationMuscle -> Millard2012EquilibriumMuscle

Bindings/Python/examples/convert_muscle_type.py
```diff
@@ -0,0 +1,111 @@
...
+# model may need to be tuned manually.
+# ------------------------------------------------------------------------#
+
+import opensim as osim
+
+import sys
+# Helper function to copy Properties between two objects
+def copy_common_muscle_properties(from_muscle, to_muscle):
+    print ("Copying Properties for muscle:" +from_muscle.getName())
+    # Common properties are in base class Muscle this includes
```
tkuchida: There are also six properties owned by `ActivationFiberLengthMuscle` and `Thelen2003Muscle` that map onto properties owned by `Millard2012EquilibriumMuscle`:

**`ActivationFiberLengthMuscle`**
- `default_activation`
- `default_fiber_length`

**`Thelen2003Muscle`**
- `maximum_pennation_angle`
- `activation_time_constant`
- `deactivation_time_constant`
- `minimum_activation`

Bindings/Python/examples/convert_muscle_type.py
```diff
@@ -0,0 +1,111 @@
...
+
+# convert_muscle_type.py
+# Author: Ayman Habib
+# ------------------------------------------------------------------------#
+# ABSTRACT: This short piece of OpenSim python API example demonstrates   #
+# how to load a model in the OpenSim python API, cycle through the        #
+# the muscles in the model and replace those of type Thelen2003 with      #
+# the more modern (true to publication) Millard2012AccelerationMuscle     #
+# There are differencs between the two muscle models so the resulting     #
+# model may need to be tuned manually.
```
tkuchida: I suggest also printing this message to the console so that users don't miss it. Perhaps also specifically state that all the new Millard2012EquilibriumMuscles will have default curves.

tkuchida:
> all similarly named Properties (including GeometryPath) are maintained and the rest is default. Is there a better option?

There are other properties that can be mapped over, as I've commented above, but there's no utility to translate the muscle curves. Actually, such a utility is impossible because of the constraints built into the Millard curves. We could try to find a set of Millard curves that approximate the Thelen curves, but I think it's fine to leave the muscle curves default; users can build on this script if they decide to adopt their own "Thelen-like" set of Millard curve parameters.

Bindings/Python/examples/convert_muscle_type.py
aymanhab: Does these properties map trivially between these 2 muscle models? I vaguely remember (de)activation time constant differences between Thelen and Schutte but don't know the relation to MillardEquilibriumMuscle.

Bindings/Python/examples/convert_muscle_type.py
tkuchida: > As far as I can tell Thelen2003Muscle has no embedded components or connections to copy.

Ok, seems fine to exclude. I think a user could create a new muscle that extends Thelen2003Muscle and has connectors/components, but I don't think this example script needs to cover that scenario.

Bindings/Python/examples/convert_muscle_type.py
tkuchida: > Does these properties map trivially between these 2 muscle models?

Yes, Thelen2003Muscle and Millard2012EquilibriumMuscle use the same underlying pennation (MuscleFixedWidthPennationModel) and activation (MuscleFirstOrderActivationDynamicModel) models; the properties are passed along to these subcomponents [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Actuators/Thelen2003Muscle.cpp#L103) and [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L138).

aymanhab:
Following up on this PR, I agree that using Millard2012EquilibriumMuscle would be better. Thinking about it however, I think a uility at the OpenSim::Object level that copies properties with similar names (including connections etc.) between Objects would be very useful in all operations that involve change types (e.g. PinJoint to FreeJoint or vice versa and all combinations of Muscle models). and would be useful in Matlab/Java, Python as well as C++.

 I started implementing 
Object::copyPropertiesFromObject(const OpenSim::Object& fromObject)

Your feedback would be appreciated. Thanks @tkuchida .
Thoughts on the name? @chrisdembia

tkuchida:
> I think a uility at the OpenSim::Object level that copies properties with similar names (including connections etc.) between Objects would be very useful in all operations that involve change types

Great idea! I'm wondering about the details. By default, I can imagine copying over any properties that match both type and name, but will there be a way to specialize (either by hard-coding or by passing in the desired mapping in, say, a vector of `std::pair<std::string, std::string>`)? A couple concerns:
- The same logical property could have different names. For example, `PointToPointSpring` has a `rest_length` (double) property and `Ligament` has a `resting_length` (double) property. I don't know whether a user would want to perform this specific conversion, but it's possible. There are likely other examples, and there isn't anything (e.g., a master list for registering all property names) controlling property names chosen by developers in the future.
- There could be properties in the target that aren't in the source, but where we can do better than the defaults. For example, if we had a two-fiber muscle, we might want to set `optimal_fiber_length_1` = `optimal_fiber_length_2` = `optimal_fiber_length` by default. There may already be situations like this (e.g., if converting from a "simple" object to an object with more properties).

aymanhab:
Thanks for the feedback @tkuchida 🥇 I think at such a low level class (Object) the functionality would be generic with no frills so I'd avoid passing dictionaries, maps, and error-prone mechanisms (e.g. due to typos, case-sensitivity etc.), though we can add these later. The only option I'd consider is copying the name (since it's not a property and maybe assumed).

The implementation I have so far copies primitive types correctly but fails to copy the GeometryPath using generic Property methods. I'll open an issue for that. 

@chrisdembia The tests fail because the python script is located under "python/examples" which is assumed that it must be a test case, which sounds surprising. Thoughts?

tkuchida:
> we can add these later

Ok, that seems fine and I think the design allows for this without any issues.

chrisdembia:
> @chrisdembia The tests fail because the python script is located under "python/examples" which is assumed that it must be a test case, which sounds surprising. Thoughts?

I recently added a ctest that runs the examples to just make sure they don't throw any exceptions (see [this CMakeLists.txt](https://github.com/opensim-org/opensim-core/blob/master/Bindings/Python/CMakeLists.txt#L268)). I thought this would be helpful in preventing our examples from becoming outdated.

Ideally, the examples should run for users as well, without needing to put a model file in place on their own.

You might be able to use the `OpenSimCopySharedTestFiles` CMake macro (https://github.com/opensim-org/opensim-core/blob/master/cmake/OpenSimMacros.cmake#L147).

OpenSim/Simulation/Model/Muscle.cpp
```diff
@@ -649,6 +649,46 @@ void Muscle::updateGeometry(const SimTK::State& s)
...
+    const int numProps = getNumProperties();
+    bool debug = false;
+    for (int px = 0; px < numProps; ++px) {
+        AbstractProperty& fromProp = updPropertyByIndex(px);
+        const std::string& pName = fromProp.getName();
+        if (hasProperty(pName)) {
+            AbstractProperty& myProp = updPropertyByName(pName); // Get writable reference to my property
+            if (myProp.isSamePropertyClass(fromProp) && !fromProp.getValueIsDefault()) {
+                if (fromProp.isOneObjectProperty()) { // Either recur or "clone the object
+                    // Expands to Property<GeometryPath>::updAs(fromProp).updValue() and works perfectly
```
aymanhab: These are the two ways to get GeometryPath from Muscle one using Properties, the other using custom/macro-based methods. @chrisdembia @aseth1 I maybe missing something obvious here, if not I'm open to suggestions. Thanks

Bindings/Python/examples/convert_muscle_type.py
aseth1: I am not sure I understand the rationale for using the low-level property machinery to do the copying opposed to using the interface provided by the Components themselves. You clearly need to know the differences between Muscles anyway (cannot map directly from one to the other derived Muscle model) but you could easily dynamic cast both to a common base (e.g. Muscle) and use the assignment operator. Then use public getters and setters to handle the leftover which are only in the derived classes.

aymanhab:
As I explained to Chris, I abandoned the copyProperty approach and I'm now
trying to use the AbstractProperty methods to do the copying (now in
Muscle.cpp) sorry about the misleading issue title.
@aseth1 the problem with your suggested approach is that it's class
specific and you'll need to do that for every common base class (Muscle,
Joint, Constraint, Probe, ......) which does not scale while doing it at
the Object or Component level would be generic and should scale
regardless). I envision something like that being used to change the type
of a component (e.g. from PinJoint to WeldJoint) without having to rewire
the Frames, locations or any other properties and without having to add new
methods for the Joint class or pair of Joints (and same for all other
types). If you believe that's impossible or unnecessary I'd be happy to
drop this (though I'd still want to understand why the code doesn't work as
expected for the future..) Thanks.

aseth1:
@aymanhab I agree a generic way would be ideal, but then calling `copyProperty_blah( )` doesn't seem like the right approach either and it sounds like you found a better way using the generic access to properties that Object provides.

aymanhab:
Totally agree @aseth1 and sorry for the misleading label that was made during early investigation.

aymanhab:
Bogus build failure. This PR shows the code that should be working but isn't that operates at the Properties level. I replicated the new method  Object::copyPropertiesFromObject to Muscle for debugging purposes.

aymanhab:
There're ModelOp operators to do the same thing now so no need for such script, in addition to the fact that every muscle model has its own parameters.

PR #1467, opened 2017-01-06
[doxygen] Remove "pre-4.0" in description of createFromStatesStorage

chrisdembia:
We had thought that 4.0 would contain an XML file format for state objects; we no longer think this will happen. So this PR removes the "pre-4.0" language with regard to states storage objects/files.


tkuchida:
Merging by [Rule 5](https://github.com/opensim-org/opensim-core/blob/master/CONTRIBUTING.md#making-a-pull-request-pr).

Issue #1450, opened 2016-12-13
attached_geometry ends up under "components" upon deserialization


aymanhab:
While this may not be an issue for the visualization code since it traverses the tree of components regardless, it affects how geometry is displayed in the navigator/GUI.

aseth1:
upon deserialization from 3.0 and early document versions or even in the latest version?

aymanhab:
I've seen it in older files (document number before 3.0) that were included in 3.3 distro. We don't have any saved model in 4.0 format. The issue is likely updateFromXML related so not critical for Alpha as far as I can tell.

aseth1:
This does not seem to be a problem for versionNumber 20303 up. I was able to replicate the problem for version 20001 with Neck3dof_point_constraint.osim in `testModelCopy`.  In `PhysicalFrame::updateFromXMLNode()` there is an `if (versionNumber < 20101) {` which does not correctly update to a 20303 or later model and it does not invoke the `convertDisplayGeometryToGeometryXML` method that creates the the `attached_geometry` list.

chrisdembia:
🕵️  🔍

aymanhab:
Thanks for the investigation @aseth1 It shouldn't be a problem to require users to save their models in version 3.2 or 3.3 to get around that, but then we broke some backward compatibility that was supposedly baked in. Do you think this would be fixable after all the changes we made in 4.0? We should discuss.

aseth1:
Closed by #1469

Issue #1347, opened 2016-10-15
Strengthen testVisualization test case


aymanhab:
testVisualization should be more extensive and cover the case where multiple non-trivial transforms are specified to transform meshes, as was done in version 3.3 and earlier.

aseth1:
In 3.3 and earlier, the `VisibleObject` had a transform associated with it and so did the individual `Mesh` geometry contained by the `VisibleObject`. The current solution (in #1299) assumes (and is likely correct) that the `Mesh` transform is expressed in the `VisibleObject` frame of reference and that the `VisibleObject` transform is expressed in the `Body` reference frame (prior to Frames), so the  when we update to replace with a new `Geometry` its transform is now composed from the two transforms to express the `Mesh` in the `Body` frame.

There are currently no tests to verify that the composition rule is correct and the test with translational offsets alone is insufficient since the order of transforms as no effect on the translation and only the rotational offset. We should verify that a 3.3. or earlier model with the two transforms defined are indeed composed as expected otherwise user define geometry will not visualize as expected.

jenhicks:
Solution for now is to throw an exception if there are multiple transforms for geometry that cannot be reproduced.

PR #1469, opened 2017-01-09
Fix attached_geometry updateFromXML

aseth1:
Addresses #1450. 

When `VisualObject` was removed, `Body::updateFromXML` was updated to put `Geometry` in a `geometry` list that was associated with every `Component`.  That code was migrated to `PhysicalFrame` and eventually we associated `attached_geometry` with `Frame`s and that block was not updated. Surprisingly, old model files were still being loaded successfully by being associated with general `components` lists and not with the appropriate `attached_geometry` list of the `Frame`. That has been corrected by simply renaming the node created `Body::updateFromXML` to `attached_geometry` from the original `geometry`.

Should be a short review @aymanhab @klshrinidhi.


aymanhab:
Thanks @aseth1 the changes look straightforward. The test change only checks for geometry attached to ground,  do you think this's sufficient? One other question, the change from "geometry" to "attached_geometry" had associated change in interpretation (attached have no extra transform while  geometry may have an extra transform). Did you check the case with an older model that had non-trivial transform? Thanks.

aseth1:
> The test change only checks for geometry attached to ground, do you think this's sufficient?

I think it was, but I updated the test so we can verify on any PhysicalFrame by name and I include "ground" for the arm26 model (which as the bulk of the geometry attached to it) and the "spine" in the neck model.

> Did you check the case with an older model that had non-trivial transform?

I did not. I agree we should do a better job verifying that the older `VisibleObject`s with potentially two transforms specified are consistent with their previous interpretation (i.e. #1347), but I think that is beyond the scope of this fix, which aims to put the `Geometry` in the correct list.

OpenSim/Tools/Test/testModelCopy.cpp
```diff
@@ -46,12 +50,14 @@ int main()
     return 0;
 }
 
-void testCopyModel(string fileName)
+void testCopyModel(const string& fileName, const int nbod, 
+    const string& physicalFrameName, const int ngeom)
 {
     const size_t mem0 = getCurrentRSS();
 
     // Automatically finalizes properties by default when loading from file
```
klshrinidhi: May remove this line since `finalizeFromProperties()` is called.

aymanhab:
Thanks @aseth1 Let's merge it and will see how old models convert.  LGTM

aseth1:
Thanks @aymanhab and @klshrinidhi.

Issue #1276, opened 2016-09-22
Mismatch in build-type of dependencies and OpenSim can lead to runtime errors like segfault


klshrinidhi:
For example, if Simbody is built in _Release_ and OpenSim is built in _Debug_, there may be runtime errors like segfault.

If the dependencies are built using superbuild, there may be a solution in writing out some information into the install directory so that OpenSim CMake can pick it up.

chrisdembia:
I'm hoping it is not necessary to write out additional information. All the detecting about the build configurations of the dependencies should exist, via the `SimbodyTargets-relwithdebinfo.cmake`, etc files. So I think it will just be a matter of picking up that information by querying for target properties. Perhaps there's something like `get_property(SimTKcommon PROPERTY CONFIGURATIONS)`.

chrisdembia:
Ah the configurations in which the dependencies are built is stored in the target property `IMPORTED_CONFIGURATIONS`.

So one can use `get_property(SimTKcommon PROPERTY IMPORTED_CONFIGURATIONS)`

https://cmake.org/cmake/help/v3.0/prop_tgt/IMPORTED_CONFIGURATIONS.html

chrisdembia:
@klshrinidhi @aymanhab There is a cmake variable for doing exactly what you want:

https://cmake.org/cmake/help/v3.0/prop_tgt/MAP_IMPORTED_CONFIG_CONFIG.html

I think all we have to do is add the following to our root CMakeLists.txt:

```cmake
set(CMAKE_MAP_IMPORTED_CONFIG_DEBUG Debug)
set(CMAKE_MAP_IMPORTED_CONFIG_RELWITHDEBINFO RelWithDebInfo)
set(CMAKE_MAP_IMPORTED_CONFIG_RELEASE Release)
set(CMAKE_MAP_IMPORTED_CONFIG_MINSIZEREL MinSizeRel)
```

This means that only the configs on the right can be used (from Simbody) for the configs on the left (for OpenSim's targets).

aymanhab:
:+1:

chrisdembia:
I discovered this via a Stack Overflow question: http://stackoverflow.com/questions/40029551/how-does-cmake-determine-the-fallback-build-configuration-for-target-dependencie/40032277#40032277

klshrinidhi:
Addressed by #1487.

PR #1475, opened 2017-01-13
[WIP] CMAKE_BUILD_TYPE mismatch issue

klshrinidhi:
Addressing #1276.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -568,18 +568,16 @@ if(NOT Simbody_FOUND)
 endif()
 
 get_property(Simbody_BUILD_TYPE
-  TARGET SimTKcommon
-  PROPERTY IMPORTED_CONFIGURATIONS)
-message(${Simbody_BUILD_TYPE})
-#message(${CMAKE_BUILD_TYPE})
+    TARGET   SimTKcommon
+    PROPERTY IMPORTED_CONFIGURATIONS)
 
-# if(${PLATFORM_NAME} STREQUAL Linux)
+string(CONCAT MESSAGE
+    "Checking if Simbody was built with same BUILD_TYPE or CONFIGURATION as "
+    "the current OpenSim BUILD_TYPE or CONFIGURATION.")
+
 add_custom_target(Simbody_CONFIG_check ALL
-  COMMAND "exit" "$<NOT:$<CONFIG:${Simbody_BUILD_TYPE}>>")
-# elseif(${PLATFORM_NAME} STREQUAL Windows)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -15,6 +15,13 @@ if(COMMAND cmake_policy)
     cmake_policy(SET CMP0005 NEW)
 endif(COMMAND cmake_policy)
 
+string(CONCAT DEP_BUILD_MSG
+  "\n-------------------------IMPORTANT----------------------------------\n"
+  "| Make sure that all OpenSim dependencies are built with the same  |\n"
+  "| CMAKE_BUILD_TYPE as OpenSim to avoid mysterious runtime errors.  |\n"
+  "-------------------------IMPORTANT----------------------------------\n")
+message(${DEP_BUILD_MSG})
+
```
chrisdembia: In CMake 3.0 and greater, it is possible to use `\` to extend a string over multiple lines. You could use that if you think it would be helpful.

klshrinidhi:
I know the message does not cover XCode, MSVC etc. It is not a complete PR. 
I started a PR to show that Windows build is failing to find LAPACK with the changes you suggested.

chrisdembia:
Oh I missed the WIP; sorry about that. Simbody creates the LAPACK target here: https://github.com/simbody/simbody/blob/master/cmake/SimbodyConfig.cmake.in

Maybe there is a way to add to that so that LAPACK can be used in any config?

klshrinidhi:
On a second thought, I think it will too restrictive if we introduce a general rule that all dependencies must be built with same BUILD_TYPE. I don't think such a thing is done in other projects thus allowing one to selectively build libraries in Debug mode.
It will be better if we:
* Apply it only to dependencies that are currently causing a problem (only Simbody I think) when there is mismatch in BUILD_TYPE.
* Print out a message during CMake configuration that in general, one may encounter problems if BUILD_TYPE don't match.

I will hold off until we have a decision.

chrisdembia:
@klshrinidhi I think that what you propose is a good idea. I agree...this is most likely a simbody-specific issue, and not a general issue.

aymanhab:
The need for all the stack of libraries to be built as Debug or Release is Windows specific and is due to the difference between the runtime libraries in both cases that are not compatible. This is intended for core developers who may forget what they built or switch between configurations to troubleshoot.

chrisdembia:
Hmm...I think at least with Simbody, it's an issue on other platforms as well. Not sure.

Issue #949, opened 2016-04-28
SimbodyEngine's getPosition, getVelocity etc are not accounting for frames correctly.


chrisdembia:
The SimbodyEngine has bugs! https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp#L190. This method is assuming internally that there are no intermediate frames, since it uses the Simbody MobilizedBody to do the transformation. This method should be reimplemented, and also deprecated.

aseth1:
I would be in favor of 1)  reverting it to use only Body so it is not wrong, and 2) deprecating since it is no longer needed. Wherever we can in the API we should use Frame and Point methods.

carmichaelong:
As per discussion with @aseth1, the plan is to do the following:
1) Keep PhysicalFrame as an argument, but throw an Exception when a PhysicalOffsetFrame is used. (Note: If we were to just change it to Body, then Ground could not be used).
2) Deprecate these functions.

chrisdembia:
See [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/ContactGeometry.h#L131) for the format for deprecating. Note the doxygen comment.

carmichaelong:
@aseth1 Is the intent to deprecate these specific methods (i.e. all methods within SimbodyEngine that take PhysicalFrame as an argument) or deprecate all methods within this class? (Or as a third option, deprecate the specific methods first, then in a later PR deprecate the rest.)

aseth1:
@carmichaelong It may be easiest to just deprecate the specific methods involving `PhysicalFrame` although it would be nice to deprecate the whole `SibodyEngine` class. With the individually deprecated methods you can also provide a more informative message about which `Frame` method to call instead, which would help ease the transition.

chrisdembia:
By the way, if you deprecate a method that OpenSim uses, then the Travis-CI tests will fail because we turn warnings into errors. So reducing the number of methods you deprecate will decrease how much extra code you have to fix.

carmichaelong:
Thanks for the heads up @aseth1 and @chrisdembia. Sounds like the first PR should just include the specific methods involving `PhysicalFrame` (which is also more specific to this issue).

jenhicks:
Can we close this issue now?

aseth1:
Yes, I believe so.

chrisdembia:
🎆

PR #1476, opened 2017-01-14
Deprecate and replace SimbodyEngine methods involving PhysicalFrame

carmichaelong:
This PR deals with #949. It takes the following steps:
* Throws an Exception if a `PhysicalOffsetFrame` is passed into a `SimbodyEngine` method that uses `PhysicalFrame` as an argument
* Deprecates the methods described above
* Adds methods to replace some of the deprecated methods (i.e. `findLocationInground`, `findVelocityInGround`, `findAccelerationInGround`) and adds tests for these methods
* Replace the use of the deprecated methods with `Frame` methods


OpenSim/Actuators/PointToPointActuator.cpp
```diff
@@ -165,24 +164,21 @@ void PointToPointActuator::computeForce(const SimTK::State& s,
...
     else
     {
         pointA_inBodyA = pointA;
         pointB_inBodyB = pointB;
-        engine.transformPosition(s, *_bodyA, pointA_inBodyA, 
-                                 getModel().getGround(), pointA_inGround);
-        engine.transformPosition(s, *_bodyB, pointB_inBodyB, 
-                                 getModel().getGround(), pointB_inGround);
+        pointA_inGround = _bodyA->findLocationInGround(s, pointA_inBodyA);
+        pointB_inGround = _bodyB->findLocationInGround(s, pointB_inGround);
```
carmichaelong: Performing a search through the code seems to indicate that `PointToPointActuator` is not being tested at all.

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.h
```diff
@@ -185,6 +162,53 @@ OpenSim_DECLARE_CONCRETE_OBJECT(SimbodyEngine, Object);
...
+    void getVelocity(const SimTK::State& s, const PhysicalFrame &aBody, const SimTK::Vec3& aPoint, SimTK::Vec3& rVel) const;
+    DEPRECATED_14("use Frame::findAccelerationInGround() instead")
+    void getAcceleration(const SimTK::State& s, const PhysicalFrame &aBody, const SimTK::Vec3& aPoint, SimTK::Vec3& rAcc) const;
+    DEPRECATED_14("use Frame::getTransformInGround().R() instead")
+    void getDirectionCosines(const SimTK::State& s, const PhysicalFrame &aBody, double rDirCos[3][3]) const;
+    DEPRECATED_14("use Frame::getTransformInGround().R() instead")
+    void getDirectionCosines(const SimTK::State& s, const PhysicalFrame &aBody, double *rDirCos) const;
+    DEPRECATED_14("use Frame::getVelocityInGround()(0) instead")
+    void getAngularVelocity(const SimTK::State& s, const PhysicalFrame &aBody, SimTK::Vec3& rAngVel) const;
+    DEPRECATED_14("use Frame::getVelocityInGround()(0) instead")
```
carmichaelong: I wasn't quite sure what to do here because if you look at the implementation of the two methods ([here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp#L270) and [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp#L283)) they seem to be implemented in exactly the same way. Simbody's doxygen [comment](https://simtk.org/api_docs/simbody/3.5/classSimTK_1_1MobilizedBody.html#ad07e64cc62fa2f76bd37894007abd378) suggests that they will both return the angular velocity in Ground frame. Thus, the true replacement is unfortunately the same function for both.

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.h
```diff
@@ -185,6 +162,53 @@ OpenSim_DECLARE_CONCRETE_OBJECT(SimbodyEngine, Object);
...
+    void getAngularVelocityBodyLocal(const SimTK::State& s, const PhysicalFrame &aBody, SimTK::Vec3& rAngVel) const;
+    DEPRECATED_14("use Frame::getAccelerationInGround()(0) instead")
+    void getAngularAcceleration(const SimTK::State& s, const PhysicalFrame &aBody, SimTK::Vec3& rAngAcc) const;
+    DEPRECATED_14("use Frame::getAccelerationInGround()(0) instead")
```
carmichaelong: Same issue here as with the angular velocity methods.

OpenSim/Simulation/Model/Frame.cpp
```diff
@@ -180,12 +180,42 @@ SimTK::Vec3 Frame::expressVectorInAnotherFrame(const SimTK::State& state,
     return findTransformBetween(state, frame).R()*vec;
 }
 
-SimTK::Vec3 Frame::findLocationInAnotherFrame(const SimTK::State& state, const
-        SimTK::Vec3& point, const Frame& otherFrame) const
+SimTK::Vec3 Frame::expressVectorInGround(const SimTK::State& state,
+                                const SimTK::Vec3& vec_F) const
+{
+    return getTransformInGround(state).R()*vec_F;
```
carmichaelong: The two functions here are meant to perform slightly different operations. `expressVectorInGround` is supposed to only express a `Vec3` in the ground basis, whereas `findLocationInGround` is supposed to treat the `Vec3` as a point and both express the `Vec3` in the ground basis and perform a translation if needed. The documentation [here](https://simtk.org/api_docs/simbody/3.5/singletonSimTK_1_1Transform__.html#a6431752e53d0aa5556082a0282a8a214) indicates that the latter occurs when a `Transform` * `Vec3` occurs. Is this redundant in the case when vectors are transformed to Ground?

OpenSim/Actuators/PointToPointActuator.cpp
tkuchida: :scream_cat:

OpenSim/Simulation/Model/Frame.cpp
tkuchida: Oh, sorry I hadn't noticed this distinction. It smelled like L186 might have been doing some unnecessary work, but I smell the difference now that you've defined `express` and `find`. I suggest adding `express` to the [list of naming conventions in CONTRIBUTING.md](https://github.com/opensim-org/opensim-core/blob/master/CONTRIBUTING.md#naming-conventions).

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.h
tkuchida: I'm confused why two methods with different names would be returning the same thing. I would expect a user to call `getAngularVelocity()` if they wanted it relative to Ground. For `getAngularVelocityBodyLocal()`, I suggest using the comment `"see Frame::getVelocityInGround()(0)"` on L183 so a user updating old code will (hopefully) check that the code is doing the right thing rather than blindly replacing with the new method.

OpenSim/Actuators/PointToPointActuator.cpp
tkuchida: The :cat: and I suggest just correcting L181 in this PR and opening a new issue about the missing test.

OpenSim/Simulation/Model/Frame.cpp
tkuchida: Addressed by c1c07c5

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.h
```diff
@@ -185,6 +162,53 @@ OpenSim_DECLARE_CONCRETE_OBJECT(SimbodyEngine, Object);
     void convertQuaternionsToDirectionCosines(double aQ1, double aQ2, double aQ3, double aQ4, double rDirCos[3][3]) const;
     void convertQuaternionsToDirectionCosines(double aQ1, double aQ2, double aQ3, double aQ4, double *rDirCos) const;
 
+    //-------------------------------------------------------------------------
+    // DEPRECATED METHODS
+    //-------------------------------------------------------------------------
+    /** @name Deprecated */
+    // @{
+    /** <b>(Deprecated)</b> Use Frame::findLocationInGround() instead. */
```
tkuchida: Addressed by c1c07c5

OpenSim/Simulation/Test/testPrescribedForce.cpp
```diff
@@ -162,9 +162,8 @@ void testPrescribedForce(OpenSim::Function* forceX, OpenSim::Function* forceY, O
         manager.setFinalTime(times[i]);
         manager.integrate(osim_state);
         osimModel->getMultibodySystem().realize(osim_state, Stage::Acceleration);
-        Vec3 accel, angularAccel;
-        accel = body.findAccelerationInGround(osim_state, Vec3(0));
-        angularAccel = body.getAccelerationInGround(osim_state)(0);
+        Vec3 accel = body.findAccelerationInGround(osim_state, Vec3(0));
+        Vec3 angularAccel = body.getAccelerationInGround(osim_state)(0);
```
tkuchida: Should be `[0]` here for consistency.

tkuchida: Should be `[0]` here for consistency.

carmichaelong:
> The 🐱 and I suggest just correcting L181 in this PR and opening a new issue about the missing test.

Issue #1480 has been created to address this and other possible testing issues.

OpenSim/Simulation/Model/Frame.h
```diff
@@ -193,6 +206,41 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Frame, ModelComponent);
...
                     const SimTK::Vec3& point_F, const Frame& otherFrame) const;
+
+    /**
+    Take a point located and expressed in this frame (F) and determine
+    its location expressed in Ground (G). This method is equivalent to
+    `findLocationInAnotherFrame()` where `otherFrame` is always Ground.
+
+    @param state       The state of the model.
+    @param point_F     The point to be re-expressed.
+    @return point_G    The re-expression of the point in Ground.
```
aseth1: I would prefer *location* instead of  *re-expression*

OpenSim/Simulation/Model/Frame.h
```diff
@@ -193,6 +206,41 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Frame, ModelComponent);
...
     SimTK::Vec3 findLocationInAnotherFrame(const SimTK::State& state, 
                     const SimTK::Vec3& point_F, const Frame& otherFrame) const;
+
+    /**
+    Take a point located and expressed in this frame (F) and determine
+    its location expressed in Ground (G). This method is equivalent to
+    `findLocationInAnotherFrame()` where `otherFrame` is always Ground.
+
+    @param state       The state of the model.
+    @param point_F     The point to be re-expressed.
```
aseth1: I would avoid re-expressed here. How about *The location of the point in F*?

OpenSim/Simulation/Model/Frame.h
```diff
@@ -193,6 +206,41 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Frame, ModelComponent);
...
+
+    /**
+    Take a point located and expressed in this frame (F) and determine
+    its velocity expressed in Ground (G).
+
+    @param state       The state of the model.
+    @param point_F     The point to be re-expressed.
+    @return point_G    The re-expression of the point in Ground.
+    */
+    SimTK::Vec3 findVelocityInGround(const SimTK::State& state,
```
aseth1: This is a bit confusing with `getVelocityInGround` which provides the Spatial Velocity of the Frame. I realize there is an extra argument which is the point, but perhaps we could rename to better differentiate? How about `findPointVelocityInGround` and similarly `findPointLocationInGround` and `findPointAccelerationInGround`? I am also OK with leaving it (already way better!).  Opinions @tkuchida, @chrisdembia @jenhicks?

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.h
```diff
@@ -185,6 +162,85 @@ OpenSim_DECLARE_CONCRETE_OBJECT(SimbodyEngine, Object);
...
+    DEPRECATED_14("use Frame::findLocationInGround() instead")
+    void transformPosition(const SimTK::State& s, const PhysicalFrame &aBodyFrom, const SimTK::Vec3& aPos, SimTK::Vec3& rPos) const;
+
+    /** <b>(Deprecated)</b> Use Point::calcDistanceBetween() or Frame::findLocationInGround() instead */
+    DEPRECATED_14("use Point::calcDistanceBetween() or Frame::findLocationInGround() instead")
+    double calcDistance(const SimTK::State& s, const PhysicalFrame& aBody1, const SimTK::Vec3& aPoint1, const PhysicalFrame& aBody2, const SimTK::Vec3& aPoint2) const;
+    
+    /** <b>(Deprecated)</b> Use Point::calcDistanceBetween() or Frame::findLocationInGround() instead */
+    DEPRECATED_14("use Point::calcDistanceBetween() or Frame::findLocationInGround() instead")
```
aseth1: The deprecated methods look great. I think we could also go ahead and deprecate the `SimbodyEngine` class as well so that users know to avoid using the class altogether.

OpenSim/Simulation/Test/testFrames.cpp
```diff
@@ -468,3 +476,57 @@ void testFilterByFrameType()
...
+        ao = rod2.findAccelerationInGround(s, p2);
+
+        cout << "t = " << s.getTime() << ": os_a = " << ao;
+        cout << " | sb_a = " << a << endl;
+
+        // Compare Simbody values to values from Station
+        SimTK_TEST_EQ(l, lo);
+        SimTK_TEST_EQ(v, vo);
+        SimTK_TEST_EQ(a, ao);
+    }
```
aseth1: 👍 Thanks!

OpenSim/Simulation/Model/Frame.h
tkuchida: Good point (pun intended). I like the more explicit `findPointVelocityInGround()`, etc.

carmichaelong: I think if we will be this explicit, perhaps `findStationLocationInGround()`, etc.? These methods would be incorrect if it's not a fixed point on the frame.

chrisdembia: I agree that renaming might be good, but I think I prefer:

- SpatialVec or Vec6: getSpatialVelocityInGround()
- SpatialVec[0]: getAngularVelocityInGround()
- SpatialVec[1]: getVelocityInGround()

Therefore, I think the name of this particular method is good.

tkuchida: Hm, there may be a logical conflict with these names. I see the preference for "velocity" when referring to linear velocity, but when written alongside the other two methods, it seems like the methods with the "spatial" and "angular" adjectives would return subsets of whatever is returned by the method with no adjective. Also, someone who generates formulations using spatial vectors may refer to the Vec6 as simply "the velocity". Perhaps `Vec6 getVelocityInGround()`, `Vec3 getAngularVelocityInGround()`, and `Vec3 getLinearVelocityInGround()` would be preferable?

aseth1: IMHO I would expect the velocity of a Frame to be 6 numbers - a SpatialVec, so calling it SpatialVelocity is somewhat unnecessary and confusing. We are then using the Frame (it has a handy helper) to find a velocity of a point fixed to that Frame, so that method could indicate what it operates on. I am happy with *Station* as well. I think *Point* is also OK because these methods do not take either a `Station` or `Point` (`Component`) but just a `Vec3` which cannot be a free floating point. The comments seem clear/unambiguous with `point_F` and `point_G` and so *Point* seems more natural.

jenhicks:
findStationVelocityInGround(s, station_F)

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.h
carmichaelong: Recap from meeting: There are some methods in `SimbodyEngine` (e.g., `formCompleteStorages`) that may be more involved to replace, so I'd like to not deprecate the whole class in this PR.

carmichaelong:
@aseth1 @tkuchida The PR is updated with the new names as discussed from the meeting and is ready to be reviewed again.

tkuchida:
:+1: Looks great.

PR #1477, opened 2017-01-16
Fix segfault when realizing to Dynamics after createFromStatesStorage()

### LINKED ISSUE #1455: Using a State from a StateTrajectory crashes on realizeDynamics call
Worked some with @aseth1 on this using MATLAB. If you get a State from a StateTrajectory and call `Model::realizeDynamics(state)`, a crash occurs with the top of the stack showing `SimTK::GeneralForceSubsystem::adoptForce`. If muscles set `appliesForce` to `false`, then no error occurs (and states can be looped to see the model move over multiple frames with the visualizer).

chrisdembia:
This PR fixes #1455.

This PR does the first of the two proposed fixes: requires the user to call `initSystem()` before invoking `createFromStatesStorage()`.

I had explored including the second change of invalidating the Instance-stage cache from the states in the trajectory. However, I ultimately decided against including this change. The reason is that I want users to be able to invoke `statesTraj.isConsistent()` immediately after `createFromStatesStorage()`, and `isConsistent()` requires the states to be realized to the Instance stage (for things like `getNQErr()`).

### DIFF
--- a/OpenSim/Simulation/StatesTrajectory.h
+++ b/OpenSim/Simulation/StatesTrajectory.h
@@ -321,6 +321,22 @@ public:
     };
 #endif
 
+    /** Thrown by createFromStatesStorage(). */
+    class ModelHasNoSystem : public Exception {
+    public:
+        ModelHasNoSystem(const std::string& file, size_t line,
+                const std::string& func,
+                const std::string& modelName) :
+                    OpenSim::Exception(file, line, func) {
+            std::string msg = "You must first call initSystem() on your Model";
+            if (!modelName.empty()) {
+                msg += " '" + modelName + "'";
+            }
+            msg += ".";
+            addMessage(msg);
+        }
+    };
+
     /** Thrown when trying to create a StatesTrajectory from a states Storage,
      * and the Storage does not contain a column for every continuous state
      * variable. */
@@ -414,6 +430,9 @@ public:
      * exactly reproduce results from the initial state trajectory; constraints
      * may not be satisfied, etc.
      *
+     * The states in the resulting trajectory will be realized to
+     * SimTK::Stage::Instance.
+     *
      * @note The naming convention for state variables changed in OpenSim v4.0;
      * `ankle_r/ankle_angle_r/speed` used to be `ankle_angle_r_u`,
      * `soleus_r/activation` used to be `soleus_r.activation`, etc. This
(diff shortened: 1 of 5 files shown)


chrisdembia:
@aseth1 @carmichaelong Could you review?

OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -321,6 +321,22 @@ class OSIMSIMULATION_API StatesTrajectory {
     };
 #endif
 
+    /** Thrown by createFromStatesStorage(). */
+    class ModelHasNoSystem : public Exception {
```
aseth1: This is a useful and more general Exception that could live with Model. Component has `ComponentHasNoSystem` but this is a good case for a Model specific one.

OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -414,6 +430,12 @@ class OSIMSIMULATION_API StatesTrajectory {
      * exactly reproduce results from the initial state trajectory; constraints
      * may not be satisfied, etc.
      *
+     * The states in the resulting trajectory will be realized to
+     * SimTK::Stage::Instance. You should not use the resulting trajectory with
+     * an instance of the model other than the you pass to this function; the
```
aseth1: "... other than the *one* you *passed* to this ..."

OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -414,6 +430,12 @@ class OSIMSIMULATION_API StatesTrajectory {
...
+     * state contains Instance-stage cache variables that point to force
```
aseth1: is it necessary to point to forces only? Any Component could have cache variables that point back into the Model.

aseth1: If it is just forces with raw pointers to `SimTK::Force`s we could disentangle so it works as expected. Perhaps post 4.0

chrisdembia: Ah...good question. I don't know...it seems that being able to cache pointers could have a performance benefit. What about PathPoints? Are they stored as cache variables?

I agree this is a larger issue to discuss, but as for this PR, do you want me to update the comment to be more general?

chrisdembia:
Thank you @carmichaelong.

chrisdembia:
I've addressed your comments. Now just waiting for the builds to pass.

chrisdembia:
The builds have pass. Ready for review.

chrisdembia:
Thank you!

PR #1478, opened 2017-01-16
Speedup StatesTrajectory's createFromStatesStorage() and exportToTable()

chrisdembia:
During the internal scripting hackathon, some noted that `StatesTrajectory::createFromStatesStorage()` is very slow (this is also true for my imiplementation of `StatesTrajectory::exportToTable()`). This was mainly because of many repeated calls to `setStateVariableValue()` (or `getStateVariableValue()`).

This PR speeds up these functions by using the much faster `setStateVariableValues()`/`getStateVariableValues()` (plural) instead.

The `testStatesTrajectory` takes 70 seconds to run without this PR. With this PR, the run time is reduced to 20 seconds (3.5x reduction).

This PR should depends on #1477 and should be merged *after* that PR is merged.


chrisdembia:
Ready for review.

OpenSim/Simulation/StatesTrajectory.cpp
```diff
@@ -143,16 +143,21 @@ TimeSeriesTable StatesTrajectory::exportToTable(const Model& model,
...
     for (size_t itime = 0; itime < getSize(); ++itime) {
         const auto& state = get(itime);
         TimeSeriesTable::RowVector row(static_cast<int>(numDepColumns));
 
         // Get each state variable's value.
-        for (unsigned icol = 0; icol < numDepColumns; ++icol) {
-            row[static_cast<int>(icol)] = 
-                model.getStateVariableValue(state, stateVars[icol]);
+        if (requestedStateVars.empty()) {
+            // This is *much* faster than getting the values one-by-one.
```
chrisdembia: Hmm...interesting.. 

> though I guess it depends on how many values the user has requested.

I suppose I could do some testing to determine the number of state variables at which to switch schemes.

tkuchida: > I suppose I could do some testing to determine the number of state variables at which to switch schemes.

Might vary depending on the model, etc. It would be slick if StatesTrajectory ran internal tests (e.g., during initialization) to determine whether to take an "additive" or "subtractive" approach, but I think there are more pressing issues for 4.0. Really, I was just imagining a question on the Forum asking how to speed up StatesTrajectory, and the nonintuitive reply to export all state variables instead of a subset.

chrisdembia: Ah well I could put a note about performance in the doxygen description for the function.

tkuchida: > I could put a note about performance in the doxygen description for the function.

:+1:

chrisdembia:
Thank you!

Issue #1443, opened 2016-12-12
Wrong marker weights value in MarkersReference


cpizzolato:
When using the `InverseKinematicsTool` the marker weights are read from the inverse kinematics setup file. However, the values of the marker weights in the xml and in the `MarkersReference` are different.

My understanding is that `MarkersReference::_markerNames` and `MarkersReference::_weights` should be in the same order, with the marker name matching in position with the respective weight, however this seems not to happen all the time. I have a [simple test case](/cpizzolato/opensim-core/tree/example-markerweightserror) if you want to take a look. Check for testInverseKinematicsTool among the other tests.

I think the problem is in `void MarkersReference::updateInternalWeights() const`, specifically, where names and weight are associated 
```
   // Begin by assigning default weight to each. Markers that do not have a
    // weight specified in the marker_weights property use the default weight.
    _weights.assign(getNumRefs(), get_default_weight());
    // Next fill in the marker weights that were specified in the
    // marker_weights property
    int wix = -1;
    int ix = 0;
    // Build flat lists of marker weights in the same order as the marker names
    for (const std::string &name : _markerNames) {
        wix = get_marker_weights().getIndex(name, wix);
        // Associate user weights (as specified in the marker_weights property)
        // with the corresponding marker by order of marker names
        if (wix >= 0)
            _weights[ix++] = get_marker_weights()[wix].getWeight();
    }
```

the index `ix` is incremented only when a weight in the xml is found, which makes it go out of sync when a name is not found. So, `ix` should be incremented at every iteration of the for loop.

Please, let me know it this makes sense :)

jenhicks:
@aseth1 will look into this. Thank you for reporting Claudio!

cpizzolato:
Hi Jen, no worries! 

To add some more details, I have a IK setup file where all the weights are 1, except for the the Top.Head marker, which has a weight of 10, see [file](https://github.com/cpizzolato/opensim-core/blob/example-markerweightserror/OpenSim/Tools/Test/subject01_IK_setup.xml).

If I print the content of `_markerNames` and `_weights` just before exiting the function  `MarkersReference::updateInternalWeights`
```
    for (int i(0); i < _markerNames.size(); ++i) {
        std::cout << _markerNames[i] << ": " << _weights[i] << std::endl;
    }
```
The marker weights should be in the same order as the marker names, however, I obtain the following output:

> R.ASIS: 1
> L.ASIS: 1
> V.Sacral: 1
[...]

PR #1481, opened 2017-01-18
Fix incorrect weighting assignments in MarkersReference

aseth1:
This PR addresses issue #1443 where the weights associated with the `MarkersReference` consumed by the `InverseKinematicsSolver` are inconsistent with the weightings in the `IKTaskSet` associated with the `InverseKinematicsTool`.

The PR adds test cases in `testInverseKinematicsSolver` and `testIK` to reproduce the inconsistency. The inconsistency appeared when the number of tasks is less than the number of markers present in the data, in which case the index into the `MarkersReference` internal list of weights was not being incremented.

Additional conditions such as order and superfluous tasks (one without a corresponding marker) was also included.

The most significant change was the creation of `InverseKinematicsTool::populateReferences()` so that the `MarkersReference` generated by the Tool could be tested directly (opposed to adding `cout`s in a *debug* mode, for example).

### DIFF
--- a/Applications/IK/test/testIK.cpp
+++ b/Applications/IK/test/testIK.cpp
@@ -28,14 +28,19 @@
 #include <OpenSim/Common/ScaleSet.h>
 #include <OpenSim/Simulation/Model/Model.h>
 #include <OpenSim/Tools/InverseKinematicsTool.h>
+#include <OpenSim/Tools/IKTaskSet.h>
 #include <OpenSim/Auxiliary/auxiliaryTestFunctions.h>
 
 using namespace OpenSim;
 using namespace std;
 
+void testMarkerWeightAssignments(const std::string& ikSetupFile); 
+
 int main()
 {
     try {
+
+        testMarkerWeightAssignments("subject01_Setup_InverseKinematics.xml");
 
         InverseKinematicsTool ik1("subject01_Setup_InverseKinematics.xml");
         ik1.run();
@@ -67,3 +72,82 @@ int main()
     cout << "Done" << endl;
     return 0;
 }
+
+void testMarkerWeightAssignments(const std::string& ikSetupFile)
+{
+    InverseKinematicsTool ik(ikSetupFile);
+
+    // Get a copy of the IK tasks
+    IKTaskSet tasks = ik.getIKTaskSet();
+
+    int nt = tasks.getSize();
+    for (int i{ 0 }; i < nt; ++i) {
+        tasks[i].setWeight(double(i));
+    }
+
+    cout << tasks.dump(true) << endl;
+    // update tasks used by the IK tool
+    ik.getIKTaskSet() = tasks;
+
+    MarkersReference markersReference;
+    SimTK::Array_<CoordinateReference> coordinateReferences;
+    
+    ik.populateReferences(markersReference, coordinateReferences);
+
+    SimTK::Array_<string> names = markersReference.getNames();
+
+    // Need a model to get a state, doesn't matter which model.
+    Model model;
+    SimTK::State& state = model.initSystem();
+
+    SimTK::Array_<double> weights;
+    markersReference.getWeights(state, weights);
+
+    SimTK_ASSERT_ALWAYS(names.size() == weights.size(),
+        "Number of markers does not match number of weights.");
+
+    for (auto i{ 0 }; i < names.size(); ++i) {
+        std::cout << names[i] << ": " << weights[i] << " in TaskSet: "
+        << tasks.get(names[i]).getWeight() << std::endl;
+        SimTK_ASSERT_ALWAYS(weights[i] == tasks.get(names[i]).getWeight(),
+            "Mismatched weight to marker");
+    }
+
+    IKTaskSet tasks2;
+    tasks2.setName("half_markers");
+
+    // Now reverse order and with half the number of tasks
+    for (int i{ nt / 2 }; i > -1 ; --i) {
+        tasks2.adoptAndAppend(tasks[2*i].clone());
+    }
+
+    cout << tasks2.dump(true) << endl;
+    ik.getIKTaskSet() = tasks2;
+
+    MarkersReference markersReference2;
+    SimTK::Array_<CoordinateReference> coordinateReferences2;
+
+    ik.populateReferences(markersReference2, coordinateReferences2);
+
+    SimTK::Array_<string> names2 = markersReference2.getNames();
+    SimTK::Array_<double> weights2;
+    markersReference2.getWeights(state, weights2);
+
+    for (auto i{ 0 }; i < names2.size(); ++i) {
+        std::cout << names2[i] << ": " << weights2[i];
+        int ix = tasks2.getIndex(names2[i]);
+        if (ix > -1) {
+            cout << " in TaskSet: " << tasks2[ix].getWeight() << endl;
+            SimTK_ASSERT_ALWAYS(weights2[i] == tasks2[ix].getWeight(),
+                "Mismatched weight to marker task");
+        }
+        else {
+            cout << " default: " << markersReference2.get_default_weight() << endl;
+            SimTK_ASSERT_ALWAYS(
+                weights2[i] == markersReference2.get_default_weight(),
+                "Mismatched weight to default weight");
+        }
+    }
+}
+
+
--- a/Applications/IK/test/testIK.cpp
+++ b/Applications/IK/test/testIK.cpp
@@ -84,7 +84,7 @@ void testMarkerWeightAssignments(const std::string& ikSetupFile)
 
     // assign different weightings so we can verify the assignments
     int nt = tasks.getSize();
-    for (int i{ 0 }; i < nt; ++i) {
+    for (int i=0; i < nt; ++i) {
         tasks[i].setWeight(double(i));
     }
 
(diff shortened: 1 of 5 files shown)


tkuchida:
(@aseth1 The CI failures appear to be legitimate.)

Applications/IK/test/testIK.cpp
```diff
@@ -67,3 +73,99 @@ int main()
...
+{
+    // Check the consistency for the IKTaskSet in the IKTool setup
+    InverseKinematicsTool ik(ikSetupFile);
+
+    // Get a copy of the IK tasks
+    IKTaskSet tasks = ik.getIKTaskSet();
+
+    // assign different weightings so we can verify the assignments
+    int nt = tasks.getSize();
+    for (int i{ 0 }; i < nt; ++i) {
```
tkuchida: Out of curiosity, why not simply `int i = 0;` as is more typical?

Applications/IK/test/testIK.cpp
```diff
@@ -67,3 +73,99 @@ int main()
...
+    }
+
+    cout << tasks.dump(true) << endl;
+    // update tasks used by the IK tool
+    ik.getIKTaskSet() = tasks;
+
+    // perform the check
+    checkMarkersReferenceConsistencyFromTool(ik);
+
+    // Now reverse the order and half the number of tasks
```
tkuchida: half -> halve

Applications/IK/test/testIK.cpp
```diff
@@ -67,3 +73,99 @@ int main()
...
+    // so that marker and tasks lists are no longer the same
+    IKTaskSet tasks2;
+    tasks2.setName("half_markers");
+
+    for (int i{ nt / 2 }; i > -1 ; --i) {
+        tasks2.adoptAndAppend(tasks[2*i].clone());
```
tkuchida: This code will segfault if `nt` is even (e.g., if `nt = 4`, then `i = 2` on the first iteration and L104 will attempt to access `tasks[4]`). I assume `nt` happens to be odd in all the examples you're using—fine for now, but it could break in the future. I suggest taking a simpler approach (and will be easier to read IMHO):
```cpp
for (int i = nt-1; i >= 0; i -= 2)
    tasks2.adoptAndAppend(tasks[i].clone());
```

Applications/IK/test/testIK.cpp
```diff
@@ -67,3 +73,99 @@ int main()
...
+    tasks.insert(nt / 2, newMarker);
+    newMarker = new IKMarkerTask();
+    newMarker->setName("C");
+    newMarker->setWeight(103);
+    // and one more at the end
+    tasks.adoptAndAppend(newMarker);
+    tasks.setName("Added superfluous tasks");
+    cout << tasks.dump(true) << endl;
+    
+    cout << tasks.dump(true) << endl;
```
tkuchida: Can one of these calls to `dump()` be removed?

Applications/IK/test/testIK.cpp
tkuchida: Actually, the code below doesn't result in half as many tasks when `nt` is odd (it will result in `(nt+1)/2`). Also see comment on L104. Perhaps simply "...and reduce the number of tasks..." since it isn't important exactly how many tasks remain.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
```diff
@@ -91,6 +99,71 @@ int main()
...
+void testMarkersReference()
+{
+    vector<std::string> labels{ "A", "B", "C", "D", "E", "F" };
+    size_t nc = labels.size();
+    size_t nr = 5;
+
+    TimeSeriesTable_<SimTK::Vec3> markerData;
+    markerData.setColumnLabels(labels);
+    for (size_t r{0}; r < nr; ++r) {
+    SimTK::RowVector_<SimTK::Vec3> row{ int(nc), SimTK::Vec3(0) };
```
tkuchida: Missing indents on L111–112.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
```diff
@@ -91,6 +99,71 @@ int main()
...
+    markerData.appendRow(0.1*r, row);
+    }
+
+    vector<int> order = { 3, 5, 1, 4, 0, 2};
+    Set<MarkerWeight> markerWeights;
+    for (size_t m{0}; m < nc; ++m)
+    markerWeights.adoptAndAppend(
+        new MarkerWeight(labels[order[m]], double(order[m])) );
```
tkuchida: `order` needs to be at least as long as `labels`. Perhaps L115 should appear immediately after L104 so their relationship is clear.

aseth1:
Thanks @tkuchida. Your points have been addressed.

aseth1:
@cpizzolato your issue should now be resolved. Care to review?

cpizzolato:
Hi @aseth1, looks really good to me, thanks for asking

aseth1:
@aymanhab please let me know if you have any comments.

OpenSim/Tools/InverseKinematicsTool.h
```diff
@@ -164,6 +166,11 @@ OpenSim_DECLARE_CONCRETE_OBJECT(InverseKinematicsTool, Tool);
     //--------------------------------------------------------------------------
     bool run() override SWIG_DECLARE_EXCEPTION;
 
+    // For testing/debugging it is necessary to know exactly what are the
+    // MarkersReference (set of marker trajectories and their weights) and
+    // CoordinateReferences that are being used by the InverseKinematicsSolver.
+    void populateReferences(MarkersReference& markersReference,
+        SimTK::Array_<CoordinateReference>&coordinateReferences) const;
```
aymanhab: Is there a reason for this method to be in the public interface? Could this be kept private?

OpenSim/Tools/InverseKinematicsTool.cpp
```diff
@@ -576,3 +516,71 @@ void InverseKinematicsTool::updateFromXMLNode(SimTK::Xml::Element& aNode, int ve
...
+    for (int i = 0; i < _ikTaskSet.getSize(); i++) {
+        if (!_ikTaskSet[i].getApply()) continue;
+        if (IKCoordinateTask *coordTask = dynamic_cast<IKCoordinateTask *>(&_ikTaskSet[i])) {
+            CoordinateReference *coordRef = NULL;
+            if (coordTask->getValueType() == IKCoordinateTask::FromFile) {
+                if (!coordFunctions)
+                    throw Exception("InverseKinematicsTool: value for coordinate " + coordTask->getName() + " not found.");
+
+                index = coordFunctions->getIndex(coordTask->getName(), index);
+                if (index >= 0) {
```
aymanhab: Is the intended behavior to pass index from previous iteration to start searching for new index? seems dependent on order of tasks. Did you consider dropping the second argument (index) so it will search from the start which is more predictable?

aymanhab:
Finished with review. I see the reason the method to populateReference was made public to support testing  but still feels that it doesn't belong to the public interface. Minor comments overall.

OpenSim/Tools/InverseKinematicsTool.cpp
aseth1: I did not change the previous functionality, I just moved it to its own method.

aseth1:
@aymanhab I agree that `populateReferences` doesn't really belong in the public interface but I also didn't want to make an intermediate (e.g. friend) class to access the private method. The method itself is `const` and populates references passed in, so calling does not impact the Tool or its behavior. Let me know if you feel strongly about hiding the method from the public interface.

aymanhab:
Thanks @aseth1 I understand and don't like the "Friend" solution either. If it's hard to isolate a test case that tests the method separately and rigorously we should put a comment in the header file indicating that the method is internal/used for debugging "only", Maybe mark it as "advanced" in Doxygen so it doesn't get mixed with the public interface ("contract") that users see.  Any suggestions from our Doxygen master @chrisdembia ?

OpenSim/Tools/InverseKinematicsTool.cpp
aymanhab: I understand, but does that look like a bug waiting to happen 💣 ? Maybe open a separate issue?

tkuchida: `Set::getIndex()` will begin searching at the provided index, but will still wrap to the beginning if the target isn't in `[index, end]`. It looks like the only chance this method would produce unpredictable results is if there are two objects in the Set with the same name, whereupon failure of this method would be the least of our worries.

aymanhab: Great point @tkuchida :+1:  I took the signature and argument names at face value (that it searches from startIndex) and didn't look at the code to see that it still wraps around! No further action on this issue then. Thanks.

chrisdembia:
You could hide the method from doxygen, or have it only show to developers (`@cond developer <method declaration> @endcond`)

aymanhab:
Perfect, thanks @chrisdembia Let's mark the method for developers and merge. Thank you,

aseth1:
I am happy to mark the method for developers, but there are currently no Doxygen comments in the Tool for any of its methods.

aymanhab:
Since  I don't know how widely spread is this lack of Doxygen, feel free to open a separate documentation PR to address. Thanks @aseth1

aseth1:
Decided to hide it from Doxygen.

aymanhab:
@aseth1 this looks good, unless you have some work to do on it please feel free to merge or let me know and I will merge it. Thanks.

aseth1:
Thanks @aymanhab - merging!

PR #1484, opened 2017-01-19
Speedup initSystem by delay loading mesh files

aymanhab:
Avoid actual loading of mesh file in Mesh::extendFinalizeFromProperties and rely on lazy loading as implemented in DecorativeMeshFile

### DIFF
--- a/Bindings/Java/OpenSimJNI/Test/testContext.cpp
+++ b/Bindings/Java/OpenSimJNI/Test/testContext.cpp
@@ -176,6 +176,14 @@ int main()
     assert(context->getLocked(dr_elbow_flexNew));
     ASSERT_EQUAL(0.5, context->getValue(dr_elbow_flexNew), 0.000001);
 
+    // Exercise Editing workflow
+    OpenSim::Body& bdy = model->updBodySet().get("r_humerus");
+    AbstractProperty& massProp = bdy.updPropertyByName("mass");
+    double oldValue = PropertyHelper::getValueDouble(massProp);
+    double v = oldValue + 1.0;
+    context->cacheModelAndState();
+    PropertyHelper::setValueDouble(v, massProp);
+    context->restoreStateFromCachedModel();
     return status;
   } catch (const std::exception& e) {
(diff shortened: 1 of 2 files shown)


Bindings/Java/OpenSimJNI/Test/testContext.cpp
```diff
@@ -176,6 +176,14 @@ int main()
     assert(context->getLocked(dr_elbow_flexNew));
     ASSERT_EQUAL(0.5, context->getValue(dr_elbow_flexNew), 0.000001);
 
+    // Exercise Editing workflow
+    OpenSim::Body& bdy = model->updBodySet().get("r_humerus");
+    AbstractProperty& massProp = bdy.updPropertyByName("mass");
```
aseth1: Is there a reason for using the property interface here instead `body.setMass(body.getMass()+1.0)`? Seems harder to read. But is this is actually what the GUI executes when an edit is made, it makes sense to leave it as is. Can you make a note of that, if that's the case?

OpenSim/Simulation/Model/Geometry.cpp
```diff
@@ -274,8 +274,9 @@ void Mesh::extendFinalizeFromProperties() {
         try {
             std::ifstream objFile;
             objFile.open(attempts.back().c_str());
-            pmesh.loadFile(attempts.back().c_str());
             // objFile closes when destructed
+            // if the file can be opened but had bad contents e.g. binary vtp
```
aseth1: We should probably update the exception message that the file could not be found and opened since the mesh won't be read in until display time.

Bindings/Java/OpenSimJNI/Test/testContext.cpp
aymanhab: Indeed that's what the GUI executes when editing from the "Properties" window. I'll add a note.

aymanhab:
Comments addressed. Thanks @aseth1 and @tkuchida

aymanhab:
Bogus failure due to timeout. Merging. Thanks much @tkuchida and @aseth1

Issue #968, opened 2016-05-12
OSX tests on Travis CI consistently timesout in < 50 min


aseth1:
We suspect it is a test that is taking over 10min without generating any output, see [here](https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received.)

jenhicks:
@aseth1 Is this still an issue?

PR #1486, opened 2017-01-23
Restore fixed flag to markers

aymanhab:
Restore flag "fixed" to Markers and restore its usage in Scaling.


aymanhab:
@chrisdembia Any conjecture as to why this specific build on clang fails due to timeout? This is not PR specific AFAIK

klshrinidhi:
Since the timeout is after 80/81 tests have passed, I think it is okay.

aymanhab:
Thanks @klshrinidhi but what's puzzling is that another build using Clang with no Wrapping and with fewer NPROC finishes in 10 min less. Since all tests take longer in this configuration it maybe a configuration/build-machine issue. Indeed digging into it you can tell that this's safe but it would be good to avoid it to begin with.

tkuchida:
> what's puzzling is that another build using Clang with no Wrapping and with fewer NPROC finishes in 10 min less

Related: #968 and #473. testInitState is still being skipped on OSX (see [here](https://github.com/opensim-org/opensim-core/blob/master/.travis.yml#L147)).

OpenSim/Simulation/Model/Marker.h
```diff
@@ -45,7 +45,13 @@ class OSIMSIMULATION_API Marker : public Station {
...
 
 class Body;
-
+public:
+    //==============================================================================
+    // PROPERTIES
+    //==============================================================================
+    OpenSim_DECLARE_PROPERTY(fixed, bool,
+        "Flag(true or false) specifying whether or not a marker "
+        "should be kept fixed in the marker placement step.  i.e. If false, the marker is allowed to move.");
```
aseth1: would prefer "Flag(true or false) that specifies whether or not the marker is fixed during marker placement (a step of the ScaleTool). If false, the MarkerPlacer is free to move the model marker's location to match its experimental counterpart." 
Or something like that.

aymanhab: Thanks @aseth1 I'll reword per your suggestion but I'll try to avoid hard-coding class names (ScaleTool, MarkerPlacer) in case the functionality moves around or is replicated somewhere else. Current wording is from 3.3 verbatim.

OpenSim/Simulation/Model/Marker.h
```diff
@@ -45,7 +45,14 @@ class OSIMSIMULATION_API Marker : public Station {
...
 class Body;
-
+public:
+    //==============================================================================
+    // PROPERTIES
+    //==============================================================================
+    OpenSim_DECLARE_PROPERTY(fixed, bool,
+        "Flag(true or false) specifying whether or not the marker "
+        "is fixed in the marker placement step of scaling.  If false, the marker is free to move "
+        "to match its experimental counterpart.");
```
tkuchida: This description makes it sound like the marker can be welded _to ground_, which I think could lead to confusion. Perhaps "Flag (true or false) indicating whether the marker is fixed in its parent Frame during the marker placement step of scaling. If false, the marker is free to move in its parent Frame to match the location of its experimental counterpart (if one exists)."?

PR #1483, opened 2017-01-19
[WIP] CMake BUILD_TYPE/CONFIGURATION mismatch issue 

klshrinidhi:
Addressing #1276.

Windows testing was done in #1475. Appveyor builds show that the solution successfully stops the build when a mismatch in CONFIG is found between Simbody and OpenSim.
Linux testing was done locally.

PR #1487, opened 2017-01-24
CMake BUILD_TYPE/CONFIGURATION mismatch issue

klshrinidhi:
Addressing #1276.
Replacing #1483.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,14 @@ if(COMMAND cmake_policy)
     cmake_policy(SET CMP0005 NEW)
 endif(COMMAND cmake_policy)
 
+string(CONCAT DEP_BUILD_MSG
+  "\n-------------------------IMPORTANT----------------------------------\n"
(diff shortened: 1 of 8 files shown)


chrisdembia:
Looks great to me except for one thing: on Mac and Linux, it is still possible to install OpenSim in multiple configurations; is this case handled? Xcode does not even need to be used to install in multiple configurations; someone could use separate build dirs with Makefiles and install multiple configurations.

klshrinidhi:
I have addressed your comment. I think multi-config builds should be handled in Linux/Mac too now.

aymanhab:
Thanks @klshrinidhi Looks good. Is there a way to handle these more generically (to avoid the check/branching based on PLATFORM_NAME)? If not then I'm fine with merging this solution as soon as builds/tests pass.

chrisdembia:
LGTM. I restarted the travis-ci/pr build that had failed. Once that passes, please merge!

klshrinidhi:
One build failed due to timeout after 78/82 tests passes.
Merging.

PR #1489, opened 2017-01-25
[WIP] ComponentPath and ChannelPath properties for Sockets and Inputs.

### LINKED ISSUE #1420: Input::setAlias is not reflected in TableReporter, etc.
Users can call `AbstractInput::setAlias()` and this will update the alias stored within `AbstractInput`. However, upon connecting, this alias is not obeyed; instead the alias (if any) from the `connectee_name` property is used. Therefore, calls to `setAlias()` will not be reflected in TableReporter's column labels.

The fix is to no longer store aliases as a separate member variable in `AbstractInput::_aliases`. Two solutions:

1. Only store the alias in the connectee_name property. This will be made easier with the `ChannelPath` class I am working on (PR is coming in a few weeks, maybe).
2. Move aliases to Outputs (instead of Inputs).

chrisdembia:
@aseth1 , @klshrinidhi and I just chatted about changes I had initiated to convert the property within  Sockets and Inputs to use ComponentPath and ChannelPath instead of std::string. The primary goal of these changes for 4.0 is to help fix bug #1420, but the changes also provide substantial cleanup of the code. The plan is for @klshrinidhi to finish this PR by tackling the tasks listed below, to get this PR merged, and then to fix #1420 in a separate pull request. If finishing these changes takes too long, then we should just fix #1420 the quick way and finish these changes after 4.0.

To-do list:

- [ ] Write documentation for the ChannelPath class.
- [ ] Some simple python/MATLAB tests to make sure it's easy to use ComponentPath/ChannelPath with Component methods (e.g., `Component::getRelativePath()`).
- [ ] Get rid of AbstractSocket::checkConnecteeNameProperty(); error checks should now occur in the deserialization code in Property.h (`readSimplePropertyFromStream()`).
- [ ] Create a base class for Sockets and Inputs whose template argument is the property type (either ComponentPath or ChannelPath); this base class would have string-based methods like `setConnecteePath(std::string)` and `appendConnecteePath(std::string)`.
- [ ] Create new methods on AbstractSocket and AbstractInput that use the path type (e.g., `setConnecteePath(ComponentPath)`), and use these in Input::findAndConnect(), etc.
- [ ] Get rid of the AbstractSocket constructor that takes `isList` as an argument; this should be obtained by the new base `Connector` by asking the property if it is a list (I can clarify this in person).
- [ ] Decide if `Channel::getPathName()` is unnecessary (replaced by `getAbsolutePath()` and `getRelativePath()`) and if so, delete it.
- [ ] Properly handle errors within `operator>>(std::istream& in, ComponentPath&)`
- [ ] Remove the `parseConnecteeName()` helper function, and related tests.

@klshrinidhi please feel free to chat with me about these changes. 

See the branch [ComponentPath_property_sandbox](https://github.com/opensim-org/opensim-core/tree/ComponentPath_property_sandbox) for my first shot at rearranging the class hierarchy for Sockets and Inputs.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/1489)


nickbianco:
Very stale PR, closing.

Issue #1461, opened 2016-12-20
Model::getFrameSet() may not find all Frames


carmichaelong:
The description for `FrameSet` says the following:

> "List of Frames that various objects can be anchored to or expressed in, Body frames are built-in and not included in this list."

Using the `SinglePin.osim` model, `dumpSubcomponents()` prints the following:
```
Model 'SinglePin'
        Ground 'ground'
                Mesh 'ground_geom_0'
                FrameGeometry 'frame_geometry'
        Body 'rod1'
                Mesh 'rod1_geom_0'
                FrameGeometry 'frame_geometry'
        CustomJoint 'pin1'
                Coordinate 'r1_z'
                PhysicalOffsetFrame 'rod1_offset'
                        FrameGeometry 'frame_geometry'
```

Looping through `getBodySet()` returns the correct output:
```
printing body names, size = 1
Body rod1
```

Looping through `getFrameSet()` correctly does not list `Body rod1` (as per its own description) but it does not grab the `PhysicalOffsetFrame rod1_offset`:
```
printing frame names, size = 0
```

I don't think this behavior fits the description of `FrameSet`, even though `rod1_offset` is auto-generated. Should the comment be changed or the behavior?

chrisdembia:
The FrameSet was created before it was possible to put any component within any other component. So at the time that the description was written, it was not ambiguous: such Frames couldn't really be anywhere else (they could be in the MiscModelComponents, but that wouldn't be the right place for them). I think the comment should be changed. I don't think it was ever intended that the FrameSet would be like the CoordinateSet, which is a non-owning bag of components that actually live elsewhere.

This is part of a larger discussion about what to do with Sets. And I don't remember what we decided.

aseth1:
I agree the comment should be updated. I don't quite understand what "Body frames are built-in" means either.  I also don't see the need for the FrameSet since (as the example illustrates) Frames are generally in service of another Component (in this case a Joint), in which case it makes sense for it to live with that Component or related (e.g. Body) component. We don't have to shove everything into the top-level model and maybe we don't need a top-level FrameSet especially since it did not exist prior to 4.0 no one would miss it.

jenhicks:
Some code in UpdateFromXML uses FrameSet. Ajay will look into changing that code and removing FrameSet completely. Also need to expose getFrameList.

aseth1:
Closed by #1491

PR #1491, opened 2017-01-26
Remove FrameSet from Model

aseth1:
Address issue #1461.

Changes:
 1. removed `FrameSet` from `Model` and all its access points in the API
 2. `Model::updateFromXMLNode` extended to translate Frames in `FrameSet` to ` components` list property
 3. incremented `XMLDocument::LatestVersion` to 30512 
 4. `PhysicalFrame::updateFromXMLNode` revised so that the `Model` and its (non-existent)  `FrameSet` node are no longer modified, instead the offset frame is added to the `components` list of its parent frame.
 5. The list order of DecorativeGeometry generate by the Model is now different and so `testVisualization` standards (file-based and programmatically generated) were also updated (painfully).


OpenSim/Simulation/Model/Marker.cpp
```diff
@@ -68,21 +68,12 @@ void Marker::setNull()
...
-        setParentFrame(*refFrame);
-    }
-    else
-    {
-        string errorMessage = "Markers must be fixed to PhysicalFrames. " + string(aName) + " is not a PhysicalFrame.";
-        throw Exception(errorMessage);
-    }
-
+    const auto& refFrame = getModel().getComponent<PhysicalFrame>(name);
+    setParentFrame(refFrame);
```
aymanhab: What would happen if the name (which is user specified in a text file) is not found due to user error or typo?

aymanhab:
@aseth1 As of version 3.3 the GUI/navigator has displayed the top level components list as a bag of "Other Components".  Specific lists are broken by category (e.g. Bodies, Joints etc.) In 4.0 I've been using the Lists/Iterators to locate Frames to show in Navigator, now this will overlap with the generic "Components". Any thoughts/suggestions? Is there a reason these new Frames don't live under a Body or Frame instead? Thanks.

OpenSim/Simulation/Model/Marker.cpp
aseth1: `getComponent()` will throw an Exception that Component could not be found, which is what `getFrameSet().get(aName)` does as well if not in the set, so that previous code was not going to execute as expected to throw the Marker specific exception.

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -1645,16 +1669,9 @@ int Model::replaceMarkerSet(const SimTK::State& s, MarkerSet& aMarkerSet)
...
-        if (getFrameSet().contains(frameName))
-        {
-            const OpenSim::PhysicalFrame* frame = dynamic_cast<const PhysicalFrame*>(&getFrameSet().get(frameName));
-            if(frame) marker->changeFrame(*frame);
-            upd_MarkerSet().adoptAndAppend(marker);
-            numAdded++;
-        }
+        Marker* marker = aMarkerSet[i].clone();
+        upd_MarkerSet().adoptAndAppend(marker);
+        ++numAdded;
```
aseth1: It is a good point, but I believe the previous implementation was not working as expected. Silently not including a `Marker` because its frame name was not in the `FrameSet` (which few are, most Markers are associated with Body frames not Frames in the Model's FrameSet property!) or was misspelled would lead to strange behavior without warning/notification to the user. Now we treat Marker as the Component it is and rely on the Socket mechanism to handle the parent frame dependency uniformly.

aseth1:
> Actually reading the description I'm not sure if the Frames will appear under "Components" of Frame/Body or Model. If the former then there's no issue here.

Any frames in the intermediate versions of the `Model` with `FrameSet` will be relocated to `components` property list. The `PhysicalOffsetFrame` frames created by handling the transforms of `VisibleObject` and its geometry are now added to the `components` list of their respective parent frames.

aseth1:
> In 4.0 I've been using the Lists/Iterators to locate Frames to show in Navigator, now this will overlap with the generic "Components". Any thoughts/suggestions? Is there a reason these new Frames don't live under a Body or Frame instead?

At the moment is the GUI "FrameSet" then a super set of the frames in the Model's FrameSet and any other Frames associated with its  subomponents obtained through the `ComponentList` but not in the Model's Components? What stops users from adding any Component to the `components` list? Isn't the same true for other Body, Force, etc.. Components added to the components list? It seems the inconsistency was present prior to this PR.

The issue of displaying lists of specific component types is important and it was definite concern/thought when adding the requirement for `ComponentList` and related iterators. I would be in favor of not adding a Frames node in the top level of the GUI in exchange for a filter option that provides a list of Components by type (i.e. the result of `getComponentList<C>`.

chrisdembia:
> ...filter option that provides a list of Components by type

Ideally, a dropdown pre-populated with all types detected in the model.

aseth1:
@chrisdembia & @aymanhab Travis is [servicing OSX](https://www.traviscistatus.com/incidents/k79mjcv403c4).  AS we wait, are there any other (or unaddressed) issues with the PR?

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -1645,16 +1670,9 @@ int Model::replaceMarkerSet(const SimTK::State& s, MarkerSet& aMarkerSet)
...
     }
```
aymanhab: Is there a reason this block and the call to marker->changeFrame has been removed?

aymanhab:
In general this looks like the first time we expose the addComponent (vs typed versions e.g. addBod, addJoint etc.) I worry about users seeing it here and being confused about it or if there're multiple calls to do the same thing (for other types). Just something to keep in mind. Otherwise this looks fine.

aymanhab:
For GUI/Navigator layout, I removed the generic list of Components from the top level model since it became meaningless, we can revisit later. Dropdown to filter could be useful (though was never planned) but defeats the purpose of having a quick access to what's in a model which is a major value added by the GUI. We can discuss somewhere else.

OpenSim/Simulation/Model/Model.cpp
aseth1: Yes. We don't rely on the `Model` using this method to both replace the MarkererSet (editing the model) and forming the connections. We rely on `finalizeConnections()` to validate and throw if parent frames cannot be found. I think that is more robust than simply dropping Markers and more flexible, since users could still be adding Frames to the Model, in which case it is premature to drop markers.

aseth1:
> In general this looks like the first time we expose the addComponent (vs typed versions e.g. addBod, addJoint etc.)

The hopper example and Sandbox prototypes use `addComponent`. That is the future. The specific types are still supported for backwards compatibility and we added `testAddComponents` to address the issue of equivalence. 

> I worry about users seeing it here and being confused about it or if there're multiple calls to do the same thing (for other types)

I agree. It may help to deprecate the typed versions. As part of preparing our examples and tutorial materials we could switch to `addComponent` uniformly. I am confident it will work for the most part, but the issue is serialization and XML editing. Until we have interactive model building/editing in the GUI people need to look at and organize their model in XML and the type specific Sets are likely still preferable.

OpenSim/Simulation/Model/Model.cpp
aymanhab: Thanks for the explanation @aseth1, would that change the workflow of Scaling  for users to call finalizeConnections() explcitly, or would that change be transparent to old user code?

aseth1: No need to `finalizeConnections()` explicitly and the only change from the user perspective will be an exception thrown if a Marker's frame is not present.

aymanhab:
LGTM, if @chrisdembia issues have been addressed feel free to merge.

chrisdembia:
LGTM>

chrisdembia:
@aseth1 , I'll let you close #1461 if you think it can be closed.

aseth1:
Thanks @chrisdembia and @aymanhab

PR #1493, opened 2017-01-27
Fix MATLAB example code for accessing coordinates from joints

### LINKED ISSUE #1232: Investigate wrapping of Coordinate enum classes

tkuchida:
Coordinate enums are available in Java but not in MATLAB. Fixes #1232.

[Example of updated doxygen on myosin](http://myosin.sourceforge.net/1493/classOpenSim_1_1BallJoint.html#aeafdd72ac449bc072bd0760404c37e3b)


klshrinidhi:
I looked at the changes too.
Explicitly numbering the enums improves readability as we can see the mapping right away.
:+1:

tkuchida:
:fireworks: Thanks @aseth1 and @klshrinidhi

PR #1496, opened 2017-01-31
Edit connecteeName in Input::setAlias

klshrinidhi:
Addressing #1420.

This is as per discussion with @aseth1 & @chrisdembia.


Bindings/Python/tests/test_sockets_inputs_outputs.py
```diff
@@ -228,22 +228,53 @@ def test_connecting_and_iterate_inputs(self):
...
-
-
-
-
-
-
-
-
+    def test_input_alias(self):
+        model_filename = 'wiring_inputs_and_outputs_with_TableReporter.osim'
```
chrisdembia: Could you rename this so it doesn't get confused with the python example of this name? Perhaps `test_input_alias.osim`.

klshrinidhi: Done.

OpenSim/Common/ComponentSocket.h
```diff
@@ -721,6 +721,19 @@ class Input : public AbstractInput {
         SimTK_INDEXCHECK_ALWAYS(index, getNumConnectees(),
                                 "Input<T>::setAlias()");
 
+        const auto& connecteeName = getConnecteeName(index);
+        std::string ignore1{}, ignore2{}, ignore3{};
+        std::string currAlias{};
+        parseConnecteeName(connecteeName,
+                           ignore1, ignore2, ignore3, currAlias);
+        if(currAlias.empty()) {
+            setConnecteeName(getConnecteeName(index) + "(" + alias + ")",
```
aseth1: It seems that is would be possible to get `/path_to_output/out_name|channel_name()(alias)` since there isn't an explicit check that the brackets are present.  I think it would be nice if there was a `composeConnecteeName` analog to the parse method with the same arguments and you simply passed in the new alias (last argument). It would be a lot more readable and easy to adapt if/when we move to `ChannelPath`

klshrinidhi: Done.
I have added a function to compose the path. The fix uses it to set the alias.

OpenSim/Common/ComponentSocket.h
```diff
@@ -722,17 +750,20 @@ class Input : public AbstractInput {
...
-                             "(" + alias + ")", index);
-        }
+                           componentPath,
+                           outputName,
+                           channelName,
+                           currAlias);
+        setConnecteeName(composeConnecteeName(componentPath,
+                                              outputName,
+                                              channelName,
+                                              alias),
```
aseth1: :+1:

chrisdembia:
Now that `composeConnecteeName()` exists, it would be nice to use it uniformly; e.g. [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Component.h#L2922) and [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Component.h#L2874). It is fine with me if we leave this for a follow-on PR.

aseth1:
> Now that composeConnecteeName() exists, it would be nice to use it uniformly; e.g. here and here. It is fine with me if we leave this for a follow-on PR.

Good point @chrisdembia. Might be easier to make the change here, but another PR is fine too. Let us know @klshrinidhi.

klshrinidhi:
I am working on the changes now. Will get back soon.

klshrinidhi:
> Now that `composeConnecteeName()` exists, it would be nice to use it uniformly; e.g. [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Component.h#L2922) and [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Component.h#L2874). It is fine with me if we leave this for a follow-on PR.

Done.
I have made changes to the two locations you mention.

chrisdembia:
Great. Please merge when tests pass.

klshrinidhi:
Tests pass. Merging.

PR #1506, opened 2017-02-05
Wrap SimTK::Visualizer and InputSilo for scripting

chrisdembia:
This PR addresses #1502: making `SimTK::Visualizer` and `SimTK::Visualizer::InputSilo` available in MATLAB. With these classes wrapped, it is possible to add DecorativeText to the Visualizer view, and to ask the InputSilo for key presses. There are no tests in this PR, but I have tested this on the tgcs2017 branch, and it works very nicely.

I had to create two functions in the swig interface file to work around the unavailability of enums in MATLAB, and the inability for SWIG to properly wrap functions that take `unsigned*` as an argument (I could have used swig typemaps instead, but I don't really know how to use them, and I thought this solution would be fine).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,7 +78,7 @@ before_install:
   ## Avoid bug on OSX (https://github.com/travis-ci/travis-ci/issues/6307)
   # This prevents the build from failing with:
   # `/Users/travis/build.sh: line 109: shell_session_update: command not found`
-  - rvm get head
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then rvm get head; fi
   
   ###########################################################################################################
   # Temporary fix until llvm.org/apt is back up. Manually download clang binaries and use them.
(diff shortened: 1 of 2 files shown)


aymanhab:
@chrisdembia What is the scope of the "flatnested" feature and is it cross language? I also don't understand why the change in the travis.yml was included in this PR. Thanks,

klshrinidhi:
@aymanhab 
The SWIG feature *flatnested* makes the bindings uniform across languages. Because not all languages support nested classes, SWIG, without this feature mentioned, will put the nested class outside/alongside the outer class in those languages. If we mention it, SWIG will do the same for all languages regardless of what they support -- it will pull the inner class out.

chrisdembia:
Thanks @klshrinidhi; I believe you are correct. The scope is limited to the class with which I used the feature. I think it is possible to specify the feature in such a way that it is applied for all classes, but that's not what I have done. The travis change was necessary to get the linux builds to pass (I think).

aymanhab:
Thanks @chrisdembia & @klshrinidhi If this has been tested (at least manually) feel free to merge.

chrisdembia:
I have tested this locally.

@klshrinidhi @aymanhab Thank you for your reviews.

Issue #1500, opened 2017-02-02
Move Hopper Example's SignalGenerator to osimCommon


chrisdembia:

tkuchida:
Fixed by #1507.

chrisdembia:
Thanks Tom.

PR #1507, opened 2017-02-06
Move SignalGenerator from Hopper example to osimCommon.

chrisdembia:
This PR addresses #1500: The SignalGenerator class used in the hopper example is generally useful, so I have moved it the osimCommon library. I also added tests for this class.

Ideally, the SignalGenerator could support types other than `double`, but Function can only provide `double`s.


OpenSim/Common/Test/testComponentInterface.cpp
```diff
@@ -1926,6 +1928,49 @@ void testAliasesAndLabels() {
...
+    const double phase = 0.3;
+    signalGen->set_function(Sine(amplitude, omega, phase));
+    world.addComponent(signalGen);
+    
+    auto* reporter = new TableReporter();
+    reporter->addToReport(signalGen->getOutput("signal"));
+    world.addComponent(reporter);
+
+    // Build the system.
+    // Check that the value of the input is the same as before.
```
tkuchida: There's no check (e.g., `SimTK_TEST_EQ()`) in this block and the system is being realized to only `Stage::Model` so it isn't clear (to me, at least) how L1951–1954 contribute to checking the value of the input. Perhaps remove this comment?

OpenSim/Common/Test/testComponentInterface.cpp
```diff
@@ -1926,6 +1928,49 @@ void testAliasesAndLabels() {
...
+    MultibodySystem system;
+    world.buildUpSystem(system);
+    State s = system.realizeTopology();
+    system.realize(s, Stage::Model);
+
+    // "Simulate."
+    const int numTimeSteps = 5;
+    for (int i = 0; i < numTimeSteps; ++i) {
```
tkuchida: For `numTimeSteps` to mean "the number of time steps taken", I suggest changing to `<=` (with corresponding changes to L1965–1966).

OpenSim/Common/Test/testSignalGenerator.cpp
```diff
@@ -0,0 +1,32 @@
...
+ *                                                                            *
+ * Unless required by applicable law or agreed to in writing, software        *
+ * distributed under the License is distributed on an "AS IS" BASIS,          *
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.   *
+ * See the License for the specific language governing permissions and        *
+ * limitations under the License.                                             *
+ * -------------------------------------------------------------------------- */
+
+void testSignalGenerator() {
+}
```
tkuchida: Perhaps include a comment explaining why this method is empty?

klshrinidhi: Wow!! Markdown is awesome!!

OpenSim/Common/Test/testComponentInterface.cpp
chrisdembia: Removed the comment...C&P error.

OpenSim/Common/Test/testComponentInterface.cpp
chrisdembia: numTimeSteps -> numTimePoint.

OpenSim/Common/Test/testSignalGenerator.cpp
chrisdembia: Didn't mean to commit this file!

chrisdembia:
@tkuchida I have tried to address your concerns 🏉

tkuchida:
> I have tried to address your concerns

Looks good. Super pull (request) :trophy:

OpenSim/Common/Test/testComponentInterface.cpp
```diff
@@ -1926,6 +1928,48 @@ void testAliasesAndLabels() {
     SimTK_TEST(foo->getInput("listInput1").getLabel(1) == "thud");
 }
 
+void testSignalGenerator() {
```
aseth1: :+1: for the test case. It really doesn't belong in `testComponentInterface`. We could start a new test suite (maybe `testFunctions`) for `Function`s and some of its primary clients, e.g. `SignalGenerator` to verify their correctness.

OpenSim/Tests/Components/testComponents.cpp
```diff
@@ -122,6 +122,8 @@ int main()
     }
     */
 
+    availableComponents.push_back(new SignalGenerator());
```
aseth1: :+1:

chrisdembia:
Thanks @aseth1. I agree that there should be a separate test case, and I think `testFunctions` is a good idea. 

I want to have `TheWorld` accessible in that separate test case (this is why I had sloppily put the test in testComponentInterface). I will either make a separate "World" class to be used in other test cases in the Common library, or I will copy `TheWorld` from `testComponentInterface`. I'll try to see which option makes more sense.

chrisdembia:
@aseth1 This is ready for review again.

OpenSim/Common/Test/testFunctions.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include <OpenSim/Common/SignalGenerator.h>
+#include <OpenSim/Common/Reporter.h>
+
+#include "ComponentsForTesting.h"
+
+using namespace OpenSim;
+using namespace SimTK;
+
+
+void testSignalGenerator() {
```
aseth1: I think this is fine for testing the SignalGenerator. 

I was also thinking that we could have:
```
testFunction(const Function& func, const std::vector<double>& times, 
                 const std::vector<double>& expectedValues) 
{
    //compare function evaluated at times provided to expected values provided
}

aseth1: Probably best to do consider as a subsequent PR.

aseth1: I added an issue #1518 to address Functions. I am merging this PR.

chrisdembia:
Thank you @aseth1 .

PR #1508, opened 2017-02-06
Fix doxygen warnings about DEPRECATED_14.

chrisdembia:
Doxygen was producing (incorrect) warnings related to the DEPRECATED_14 macro. This PR causes Doxygen to evaluate the macro to nothing, and thus gets rid of the warnings.

For more information, see https://www.stack.nl/~dimitri/doxygen/manual/preprocessing.html


klshrinidhi:
:+1:

chrisdembia:
Thanks @klshrinidhi; I'm merging based on rule 5.

Issue #1130, opened 2016-07-19
Add requirement for user to call finalizeFromProperties() explicitly in all cases


tkuchida:
In the hopper example, the user must call `finalizeFromProperties()` explicitly after creating a model; otherwise, an exception will be thrown downstream (and I believe the exception is likely to be unhelpful—e.g., indicating that a Component cannot be found on the specified path). Ideally, the user should not need to be concerned with how/when OpenSim organizes its internal data structures.

In addition to a formal test case, we should be able to remove the `finalizeFromProperties()` calls in [exampleHopperDevice.cpp](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Sandbox/ExampleHopperDevice/exampleHopperDevice.cpp#L171) and [exampleHopperDevice_answers.cpp](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Sandbox/ExampleHopperDevice/exampleHopperDevice_answers.cpp#L243).

chrisdembia:
It is an unfortunate reality of C++ that one should not call virtual methods in constructors. Therefore, we should not be calling `finalizeFromProperties()` in constructors (since it invokes the virtual `extendFinalizeFromProperties()`). It may still be possible to internally invoke `finalizeFromProperties()` farther downstream, but still before the exception gets thrown.

klshrinidhi:
Maybe it is good to re-think how this whole thing is laid out to eliminate this step needed by user. It is a result of previous decisions.

tkuchida:
Other potential alternatives are to (1) call `finalizeFromProperties()` automatically whenever necessary, or (2) track down the most likely downstream calls a user would make and include a "did you remember to call finalizeFromProperties()?" reminder in the exception message. Point (2) is related to existing issues with user-facing messages (e.g., see #521 and #337).

chrisdembia:
Related: #1010

jenhicks:
Will update API so that loadModel doesn't call finalizeFromProperties(), but users have to call explicitly to be consistent with copy and making edits.

tkuchida:
> Will update API so that loadModel doesn't call finalizeFromProperties(), but users have to call explicitly to be consistent with copy and making edits.

Sounds good. I suggest also catching/adding a meaningful error message if the user forgot to call `finalizeFromProperties()` to avoid cryptic exceptions thrown downstream.

Issue #1010, opened 2016-06-15
finalizeFromProperties() invokes virtual methods, and as such should maybe not be invoked in constructors.


chrisdembia:
There's a C++ guideline that says not to call virtual methods in constructors. `finalizeFromProperties()` invokes the virtual method `extendFinalizeFromProperties()`, so it may not be good to call this method in constructors. I haven't witnessed any bad reprecussions yet, though.

chrisdembia:
It might be okay to call `finalizeFromProperties()` from `Model`, so long as Model is `final`. That is, as long as  `finalizeFromProperties()` does not invoke the virtual method `extendFinalizeFromProperties()` that is implemented in types derived from Model. I think the previous issue was that we were calling virtual functions from the `Component` constructor, and that is not okay because it is a base class. https://www.securecoding.cert.org/confluence/display/cplusplus/OOP50-CPP.+Do+not+invoke+virtual+functions+from+constructors+or+destructors.

(I made the above comment in #1205.

Issue #606, opened 2015-09-08
Does Component need a constructor that takes a filename? Does Component need to call finalizeFromProperties() in its constructors?


chrisdembia:
It's possible that constructing Model invokes two calls of finalizeFromProperties().

jenhicks:
We should remove the constructor that takes a filename.

aymanhab:
I'm not sure which class we're referring to here but constructors from filenames are used all over the place from Models to Tools and in almost every single example piece of code we publish!

chrisdembia:
We're talking about removing `Component::Component(std::string)`. It's an abstract class, so it makes less sense to have this constructor. At least it should be protected (though since it's abstract, maybe it's already implied as protected).

chrisdembia:
The issue is that having this constructor causes too many invocations of `finalizeFromProperties()`

aseth1:
I think there are at least two issues being mangled here. First is whether or not Component should have a constructor which takes filename which it uses to construct as an Object that deserializes from that file.  I am fine if we decide to remove it but I don't think it is causing any problems, and if you derive a concrete Component and you want that deserializing constructor you have to explicitly call Object(filename) from your constructor instead of calling the Component(filename), which may be weird for some.

Second, `finalizeFromProperties()` was being called excessively because of a bug that was causing the isObjectUpToDateWithProperties() to be false and triggering subsequent finalizes to be invoked unnecessarily  (see #1100) . That issue has been addressed (#1120).

Third is whether or not Model constructors should have the additional *power* of finalizing itself upon construction (which invokes the virtual `extendFinalizeFromProperties()` on Model).  No other Component can do this. This was done mostly for convenience since a Model is dead (no traversable tree of subcomponents)  without a top-level `finalizeFromProperties()`. So immediately after constructing a Model you would have to call  `finalizeFromProperties()` always. That seemed to burdensome for me, but in hindsight I think it is the correct thing to do, because now if you copy construct a `Model` you **do** have to `finalizeFromProperties` yourself and it exposes an inconsistency (see  #1130) that is bothersome. 

[...]

PR #1511, opened 2017-02-09
Require Model::finalizeFromProperties() to be called explicitly 

aseth1:
Removes implicit call to `finalizeFromProperties()` within `Model` constructors and requires the call to be explicit (related to issues #606, #1010, #1130). Also adds exceptions for orphaned (accidental cloning) and empty root components (when forgetting to call `finalizeFromProperties()` on  a model).

### DIFF
--- a/Bindings/Java/tests/TestBasics.java
+++ b/Bindings/Java/tests/TestBasics.java
@@ -3,6 +3,7 @@ import org.opensim.modeling.*;
 class TestBasics {
     public static void testBasics() {
         Model m = new Model();
+        m.finalizeFromProperties();
         m.print("empty_model.osim");
         Frame gnd = m.getGround();
         FrameList frames = m.getFrameList();
--- a/Bindings/Java/tests/TestEditProperties.java
+++ b/Bindings/Java/tests/TestEditProperties.java
@@ -7,7 +7,6 @@ class TestEditProperties {
   public static void main(String[] args) {
     try {
         Model model = new Model("gait10dof18musc_subject01.osim");
-        model.finalizeFromProperties();
         OpenSimContext context=null;
         long t1  = System.currentTimeMillis();
         try {
--- a/Bindings/Java/tests/TestEditProperties.java
+++ b/Bindings/Java/tests/TestEditProperties.java
@@ -7,6 +7,7 @@ class TestEditProperties {
   public static void main(String[] args) {
     try {
         Model model = new Model("gait10dof18musc_subject01.osim");
+        model.finalizeFromProperties();
         OpenSimContext context=null;
         long t1  = System.currentTimeMillis();
         try {
--- a/Bindings/Python/tests/test_access_subcomponents.py
+++ b/Bindings/Python/tests/test_access_subcomponents.py
@@ -13,6 +13,8 @@ osim.Model.setDebugLevel(0)
 class TestAccessSubcomponents(unittest.TestCase):
     def test_individual_components(self):
         model = osim.Model(os.path.join(test_dir, "arm26.osim"))
+		model.finalizeFromProperties()
+		
         muscle = model.getComponent('BICshort')
         assert muscle.getName() == 'BICshort'
         # No downcasting necessary!
@@ -22,6 +24,7 @@ class TestAccessSubcomponents(unittest.TestCase):
 
     def test_component_list(self):
         model = osim.Model(os.path.join(test_dir, "arm26.osim"))
+		model.finalizeFromProperties()
 
         num_components = 0
         for comp in model.getComponentsList():
@@ -77,6 +80,7 @@ class TestAccessSubcomponents(unittest.TestCase):
 
     def test_component_filter(self):
         model = osim.Model(os.path.join(test_dir, "arm26.osim"))
+		model.finalizeFromProperties()
 
         comps = model.getMuscleList()
         comps.setFilter(osim.ComponentFilterAbsolutePathNameContainsString('BIC'))
--- a/Bindings/Python/tests/test_access_subcomponents.py
+++ b/Bindings/Python/tests/test_access_subcomponents.py
@@ -13,8 +13,8 @@ osim.Model.setDebugLevel(0)
 class TestAccessSubcomponents(unittest.TestCase):
     def test_individual_components(self):
         model = osim.Model(os.path.join(test_dir, "arm26.osim"))
-		model.finalizeFromProperties()
-		
+        model.finalizeFromProperties()
+        
         muscle = model.getComponent('BICshort')
         assert muscle.getName() == 'BICshort'
         # No downcasting necessary!
@@ -24,7 +24,7 @@ class TestAccessSubcomponents(unittest.TestCase):
 
     def test_component_list(self):
         model = osim.Model(os.path.join(test_dir, "arm26.osim"))
-		model.finalizeFromProperties()
+        model.finalizeFromProperties()
 
         num_components = 0
         for comp in model.getComponentsList():
@@ -80,7 +80,7 @@ class TestAccessSubcomponents(unittest.TestCase):
 
     def test_component_filter(self):
         model = osim.Model(os.path.join(test_dir, "arm26.osim"))
-		model.finalizeFromProperties()
+        model.finalizeFromProperties()
 
         comps = model.getMuscleList()
         comps.setFilter(osim.ComponentFilterAbsolutePathNameContainsString('BIC'))
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -131,12 +131,15 @@ void Component::addComponent(Component* subcomponent)
     while (root->hasParent()) {
         root = &(root->getParent());
     }
-
-    auto components = root->getComponentList<Component>();
-    for (auto& c : components) {
-        if (subcomponent == &c) {
-            OPENSIM_THROW( ComponentAlreadyPartOfOwnershipTree,
-                subcomponent->getName(), getName());
+    // if the root has no immediate subcomponents do not bother
+    // checking if the subcomponent is in the ownership tree
+    if ((root->getNumImmediateSubcomponents() > 0)) {
+        auto components = root->getComponentList<Component>();
+        for (auto& c : components) {
+            if (subcomponent == &c) {
+                OPENSIM_THROW(ComponentAlreadyPartOfOwnershipTree,
+                    subcomponent->getName(), getName());
+            }
         }
     }
 
@@ -1526,19 +1529,34 @@ void Component::dumpConnections() const {
 
 
 void Component::initComponentTreeTraversal(const Component &root) const {
-    // Going down the tree, node is followed by all its
-    // children in order, last child's successor is the parent's successor.
+    // Going down the tree, this node is followed by all its children.
+    // The last child's successor (next) is the parent's successor.
 
     const size_t nmsc = _memberSubcomponents.size();
     const size_t npsc = _propertySubcomponents.size();
     const size_t nasc = _adoptedSubcomponents.size();
 
-    // If this component has no parent and has no subcomponents
-    // this is an orphan component and likely failed to call 
-    // finalizeFromProperties on the root OR made and orphaned
-    // clone of the Component.
-    if (!(this->hasParent()) && !(nmsc + npsc + nasc) ){
+    // If this isn't the root component and it has no parent, then
+    // this is an orphan component and we likely failed to call 
+    // finalizeFromProperties on the root OR this is a clone that
+    // has not been added to the root (in which case would have a parent).
+    if ((this != &root) && !this->hasParent() ) {
         OPENSIM_THROW(ComponentIsAnOrphan, getName(), getConcreteClassName());
+    }
+
+    if ((this == &root) && !(nmsc + npsc + nasc)) {
+        ComponentIsRootWithNoSubcomponents ex(__FILE__, __LINE__, __func__,
+            getName(), getConcreteClassName());
+        if (getConcreteClassName() == "Model") {
+            // If we have a Model that is has no subcomponents we know this is 
+            // an issue of not calling finalizeFromProperties on the Model
+            throw ex;
+        }
+        else// Albeit strange for a Component to be root and have no subcomponents
+            // it is possible, and testComponents creates and interrogates
+            // components before they are added to a Model and many of these
+            // may have no subcomponents. So, just issue a warning in this case.
+            std::cout << "WARNING: " << ex.what() << std::endl;
     }
 
     const Component* last = nullptr;
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -1528,9 +1528,22 @@ void Component::dumpConnections() const {
 void Component::initComponentTreeTraversal(const Component &root) const {
     // Going down the tree, node is followed by all its
     // children in order, last child's successor is the parent's successor.
+
+    size_t nmsc = _memberSubcomponents.size();
+    size_t npsc = _propertySubcomponents.size();
+    size_t nasc = _adoptedSubcomponents.size();
+
+    // If this component has no parent and has no subcomponents
+    // this is an orphan component and likely failed to call 
+    // finalizeFromProperties on the root OR made and orphaned
+    // clone of the Component.
+    if (!(this->hasParent()) && !(nmsc + npsc + nasc) ){
+        OPENSIM_THROW(ComponentIsAnOrphan, getName(), getConcreteClassName());
+    }
+
     const Component* last = nullptr;
-    for (unsigned int i = 0; i < _memberSubcomponents.size(); i++) {
-        if (i == _memberSubcomponents.size() - 1) {
+    for (unsigned int i = 0; i < nmsc; i++) {
+        if (i == nmsc - 1) {
             _memberSubcomponents[i]->_nextComponent = _nextComponent.get();
             last = _memberSubcomponents[i].get();
         }
@@ -1540,7 +1553,7 @@ void Component::initComponentTreeTraversal(const Component &root) const {
             last = _memberSubcomponents[i + 1].get();
         }
     }
-    if (size_t npsc = _propertySubcomponents.size()) {
+    if (npsc) {
         if (last)
             last->_nextComponent = _propertySubcomponents[0].get();
 
@@ -1557,7 +1570,7 @@ void Component::initComponentTreeTraversal(const Component &root) const {
             }
         }
     }
-    if (size_t nasc = _adoptedSubcomponents.size()) {
+    if (nasc) {
         if (last)
             last->_nextComponent = _adoptedSubcomponents[0].get();
 
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -1536,27 +1536,20 @@ void Component::initComponentTreeTraversal(const Component &root) const {
     const size_t npsc = _propertySubcomponents.size();
     const size_t nasc = _adoptedSubcomponents.size();
 
-    // If this isn't the root component and it has no parent, then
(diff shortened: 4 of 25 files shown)


Bindings/Python/tests/test_access_subcomponents.py
```diff
@@ -13,6 +13,8 @@
 class TestAccessSubcomponents(unittest.TestCase):
     def test_individual_components(self):
         model = osim.Model(os.path.join(test_dir, "arm26.osim"))
+		model.finalizeFromProperties()
```
tkuchida: Tabs -> spaces (also on L17, 27, 83).

OpenSim/Common/Component.cpp
```diff
@@ -1526,11 +1529,39 @@ void Component::dumpConnections() const {
...
+    // finalizeFromProperties on the root OR this is a clone that
+    // has not been added to the root (in which case would have a parent).
+    if ((this != &root) && !this->hasParent() ) {
+        OPENSIM_THROW(ComponentIsAnOrphan, getName(), getConcreteClassName());
+    }
+
+    if ((this == &root) && !(nmsc + npsc + nasc)) {
+        ComponentIsRootWithNoSubcomponents ex(__FILE__, __LINE__, __func__,
+            getName(), getConcreteClassName());
+        if (getConcreteClassName() == "Model") {
```
tkuchida: Might we want to relax this check to catch any ModelComponent? I'm thinking about a situation like the Device in the hopper example: it's not a Model (it extends ModelComponent), but it would be root and a user might want to traverse the tree without attaching it to a model first.

OpenSim/Common/Component.cpp
```diff
@@ -1526,11 +1529,39 @@ void Component::dumpConnections() const {
...
+            // If we have a Model that is has no subcomponents we know this is 
+            // an issue of not calling finalizeFromProperties on the Model
+            throw ex;
+        }
+        else// Albeit strange for a Component to be root and have no subcomponents
```
tkuchida: I wouldn't expose this case to the API just because the test is designed that way. Perhaps you can use something like `Object::setDebugLevel()` to override the `throw` on L1553?

OpenSim/Simulation/Model/Model.h
```diff
@@ -220,18 +220,23 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Model, ModelComponent);
...
+    /** Default constructor creates a %Model containing only the Ground frame
     and a set of default properties. */
     Model();
 
     /** Constructor from an OpenSim XML model file. 
+    NOTE: The Model is read in (deserialized) from the model file, which means
+    the properties of the Model and its components are filled in from values in
+    the file. In order to evaluate the validity of the properties (e.g. Inertia
+    tensors, availability of a Mesh files, and to traverse the tree of the
+    Model's subcomponents, one must invoke Model::finalizeFromProperties(),
```
tkuchida: Missing ")".

OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp
```diff
@@ -99,7 +99,7 @@ int main()
 
         // The finalize flag is for testing purposes ONLY. This way we
         // can ignore invalid properties and focus the test on serialization.
-        Model deserializedModel("allComponents.osim", false);
+        Model deserializedModel("allComponents.osim");
```
tkuchida: Finalize flag was removed; comment should be updated.

OpenSim/Common/Component.cpp
aseth1: I agree other Components can serve as a root, but that is generally not the case. You may create an empty Device and then add to it, in which case the act of probing the initially empty Device you wouldn't want to throw this exception. A Model on the other hand should never be empty (it has Ground, for example) and so is clear in that case that `finalizeFromProperties()` was not called on the model.

OpenSim/Common/Component.cpp
aseth1: Do you mean we should not message the console by default? I can add a `getDebugLevel > 0` check and then throw or print instead.

OpenSim/Common/Component.cpp
tkuchida: :one: I'm not a fan of special-casing Model, but this is consistent with the rest of the API. :two: I don't think it would be so bad to throw an exception whenever a user requests a ComponentList where `(nmsc + npsc + nasc) < 1`—if they don't know whether there are subcomponents, it seems like they either don't know what's going on or they've made a mistake; otherwise, a try/catch block can be used explicitly. Your way is more forgiving in a sense, but IMHO it would be better to make sure users catch `ComponentIsRootWithNoSubcomponents` exceptions.

OpenSim/Common/Component.cpp
tkuchida: My thinking was that, if you want slightly different behavior for the test case, then you could set `Object::setDebugLevel()` in the test case and check it here rather than eternally special-casing certain classes. I think this comment will be resolved via L1550.

OpenSim/Common/Component.cpp
aseth1: 1. I agree it is ugly, but Model was the only case I could be certain about.
2. I definitely can do (and contemplated doing) that and I even used a `try/catch` in an previous incarnation of `testComponents`. I opted not to do that because I thought throwing an exception when accessing an empty component maybe too severe. But, you raise a good point that it indicates that the user may not realize that they are mistakenly dealing with an empty component. I am happy to make it more strict. It does make it more difficult to write generic code that takes a Component and rummages through it. That might be something the GUI does routinely, but even then it would be strange to be to have a Component that had no parent and was empty and might be worth alerting about. @aymanhab what do you think?

OpenSim/Simulation/Model/Model.h
```diff
@@ -220,18 +220,23 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Model, ModelComponent);
...
     Model();
 
     /** Constructor from an OpenSim XML model file. 
+    NOTE: The Model is read in (deserialized) from the model file, which means
+    the properties of the Model and its components are filled in from values in
+    the file. In order to evaluate the validity of the properties (e.g. Inertia
+    tensors, availability of Mesh files, and to traverse the tree of the
+    Model's subcomponents) one must invoke Model::finalizeFromProperties() 
+    first. Model::initSystem() invokes finalizeFromProperties() on its way to
+    creating the System and initializing the State.
```
chrisdembia: This comment is very good but I still see a small bit of room for confusion, since there is a difference between not having any subcomponents and those subcomponents not yet being "registered." Consider the following change:

"and to traverse the tree of the Model's subcomponents)" -> "to register all components with the Model"

and change the `ComponentIsRootWithNoSubcomponents` message as so:

"is the root but has no subcomponents. Verify that finalizeFromProperties() was called on this Component" ->
"is the root but no subcomponents are registered. Verify that subcomponents are registered by calling finalizeFromProperties() on this Component".

OpenSim/Common/Component.h
```diff
@@ -93,6 +93,39 @@ class ComponentNotFoundOnSpecifiedPath : public Exception {
...
+        size_t line,
+        const std::string& func,
+        const std::string& thisName,
+        const std::string& componentConcreteClassName) :
+        Exception(file, line, func) {
+        std::string msg = "Component '" + thisName + "' of type " +
+            componentConcreteClassName + " has no parent and is not the root.\n" +
+            "Verify that finalizeFromProperties() has been invoked on the " + 
+            "root Component or that this Component is not a clone, which has " +
+            "not been added to its parent.";
```
chrisdembia: Consider clarifying "parent" as "parent component" to avoid confusion with "parent frames" etc.

OpenSim/Common/Component.cpp
```diff
@@ -1526,11 +1529,39 @@ void Component::dumpConnections() const {
...
+    const size_t nmsc = _memberSubcomponents.size();
+    const size_t npsc = _propertySubcomponents.size();
+    const size_t nasc = _adoptedSubcomponents.size();
+
+    // If this isn't the root component and it has no parent, then
+    // this is an orphan component and we likely failed to call 
+    // finalizeFromProperties on the root OR this is a clone that
+    // has not been added to the root (in which case would have a parent).
+    if ((this != &root) && !this->hasParent() ) {
+        OPENSIM_THROW(ComponentIsAnOrphan, getName(), getConcreteClassName());
```
chrisdembia: Could you clarify somewhere why this exception is thrown? My understanding is that we generally want to flag orphaned components, and this is a convenient place to do so. Is that the case? Or is it that we cannot properly form a ComponentList for an orphaned component?

OpenSim/Common/Component.cpp
aymanhab: In the past, the GUI did not create any components on its own except for ExternalLoads that are instantiated from a File and passed along to Tools (would these count as orphans?) (also Makers and PathPoints now) In general Exceptions are not very friendly to GUI unless added to the signature of methods and passed on to SWIG so that handling them is done locally with meaningful error messages rather than a generic something wrong has happened.

aymanhab:
In general the main issue here is the exposure of a new method (finalizeFromProperties) to users. API users in the past have been confused by the need to call initSystem, now they will have to decide if they need finalizeFromProperties or if initSystem will be enough. I'd suggest for all example/user-facing code and most tests that can get away with initSystem that we would use it unless there's a very clear reason to use the finer granularity finalizeFromProperties explicitly. Not sure where is the documentation or where do we expect users to find it.

OpenSim/Simulation/Model/Model.h
aseth1: I am OK with using a word like "registered" to mean that the subcomponents are formally part of the parent and can be accessed through the parent, but we already have the requirement of "registering" Object types (`registerTypes`). Maybe another term would be better?

OpenSim/Common/Component.cpp
```diff
@@ -131,12 +131,15 @@ void Component::addComponent(Component* subcomponent)
...
     }
-
-    auto components = root->getComponentList<Component>();
-    for (auto& c : components) {
-        if (subcomponent == &c) {
-            OPENSIM_THROW( ComponentAlreadyPartOfOwnershipTree,
-                subcomponent->getName(), getName());
+    // if the root has no immediate subcomponents do not bother
+    // checking if the subcomponent is in the ownership tree
+    if ((root->getNumImmediateSubcomponents() > 0)) {
```
aseth1: This is to avoid invoking `initComponentTreeTraversal()` and getting the         `ComponentIsRootWithNoSubcomponents` related warning.

OpenSim/Common/Component.cpp
aseth1: > My understanding is that we generally want to flag orphaned components, and this is a convenient place to do so. Is that the case?

Yes, I thought this would be one place we could detect an orphan.

OpenSim/Simulation/Model/Model.h
tkuchida: > Maybe another term would be better?

_Catalog_/_Subscribe_ a Component with a Model? Update a Model's _inventory_ of Components?

OpenSim/Common/Component.cpp
aseth1: If you create the `ExternalLoads` and add them to the Model they (the `ExternalForce`s) will not be considered orphans and this true for any Components. If you create them and do not add them to the Model, but make inquiries on them (i.e. to get their subcomponents) they will throw the `ComponentIsRootWithNoSubcomponents` exception. It sounds like the Components are always added to Model. Is that right?

aymanhab: Yes, that is correct. For the case of experimental markers and forces (which are not implemented yet) I'll add them to a "shell" model as we did in the past so this should also be a non-issue. Thanks for checking @aseth1 👍

aseth1:
> I'd suggest for all example/user-facing code and most tests that can get away with initSystem that we would use it unless there's a very clear reason to use the finer granularity finalizeFromProperties explicitly. Not sure where is the documentation or where do we expect users to find it.

I agree. In the cases that we load a model and `initSystem()` to perform a simulation there is no change required. In the cases where you load a model and want to tweak it or inspect it (check how many subcomponents, or get a list of Frames, etc...) you need to call `finalizeFromProperties` first.

While on one hand we are increasing the cognitive "burden" on the user we are also making it possible to fully inspect the model and access all of its components in a generic way that we were not able to provide before.  

Removing the implicit `finalizeFromProperties` from the `Model` constructors is primarily for consistency. For example, if you cloned a model you would be required to call `finalizeFromProperties` to be able to inspect it. If it proves to be too significant a burden we could implement a custom copy Constructor for `Model` that also invokes `finalizeFromProperties` internally, but it would not be pretty. I would make that our last resort.

tkuchida:
> While on one hand we are increasing the cognitive "burden" on the user we are also making it possible to fully inspect the model and access all of its components in a generic way that we were not able to provide before.

Would the cognitive burden be reduced at all if the method were called `initModel()` to match `initSystem()`? :turtle:

Bindings/Java/tests/TestBasics.java
```diff
@@ -3,6 +3,7 @@
 class TestBasics {
     public static void testBasics() {
         Model m = new Model();
+        m.finalizeFromProperties();
```
aymanhab: Is it necessary to make this call even though the model is empty?

Bindings/Java/tests/TestEditProperties.java
```diff
@@ -7,6 +7,7 @@
   public static void main(String[] args) {
     try {
         Model model = new Model("gait10dof18musc_subject01.osim");
+        model.finalizeFromProperties();
```
aymanhab: This reflects the GUI workflow where initSystem is called right after construction, so I'd prefer to remove the call to finalizeFromProperties here to make sure the call sequence with only initSystem continues to work.

aseth1: You are right, I didn't see the `initSystem()` call in the `OpenSimContext` construction call. I might have been over zealous in my changes. I can try backing it off.

Bindings/Java/tests/TestBasics.java
aseth1: It is necessary because a few lines down we want to use `getComponentList()` (invoked by `m.getFrameList()`). We are asking a `Model` to list its subcomponents without giving it an opportunity to identify its subcomponents amongst it properties (all those Sets) and other data members. Again we were calling `finalizeFromProperties()` in the constructor before (thus hiding the fact) but the consensus was that it was not consistent with copying a model.

aseth1:
> Would the cognitive burden be reduced at all if the method were called initModel() to match initSystem()? 🐢

Some background:
 1. `finalizeFromProperties()` is a behavior of all `Component`s not just `Model`
 2. `initModel()` or `initComponent()` would do what? Would it result in a `Component` (`Model`) that was ready for computation? Or just read in its properties? Or also have its connections satisfied?
 3. a remnant of the earlier API (pre 3.2) is `Model::setup()`. It basically attempted to validate some properties and satisfy some of the interdependencies and this was abstracted to two distinct operations: `finalizeFromProperties()` and `finalizeConnections()`. It also suffered from the fact that it was called in the `Model` constructors (that is why `finalizeFromProperties()` was being called instead.  Copies were not `setup()` so we had the same inconsistency. 

I think we could have a *nicer* method name than `finalizeFromProperties` and perhaps combine it with `finalizeConnections` which would then essentially be `setup()` on `Model`, in which case `initModel()` would be better name.

tkuchida:
> `initModel()` or `initComponent()` would do what? Would it result in a `Component` (`Model`) that was ready for computation? Or just read in its properties? Or also have its connections satisfied?

Based on the number of question marks, I guess it was a bad idea. I don't expect the novice user to know or care what the difference is between a property and a connector/socket. The `finalizeFromProperties()` and `initSystem()` pair of calls in the API might look sort of like "open the panel and connect the blue wire to terminal 3" and "press _start_" to a novice user. Unfortunately, we're already stuck with the less-than-descriptive "initSystem".

> a remnant of the earlier API (pre 3.2) is `Model::setup()`. It basically attempted to validate some properties and satisfy some of the interdependencies and this was abstracted to two distinct operations: `finalizeFromProperties()` and `finalizeConnections()`. It also suffered from the fact that it was called in the `Model` constructors (that is why `finalizeFromProperties()` was being called instead).  Copies were not `setup()` so we had the same inconsistency.

I can understand the desire for granularity in the guts of the API, but MATLAB users for example may not need to know when or why a model needs to be finalized. The staging mechanism in Simbody might be a nice example to follow (the user just needs to know `realize()`); methods that _look_ similar (e.g., `initBlah()` to go with `initSystem()`) might be clean in OpenSim.

Bindings/Java/tests/TestBasics.java
aymanhab: Thanks for the clearup @aseth1

aymanhab:
This LGTM, unless we want to address @tkuchida concerns regarding naming. I see some tests calling the method finalizeFromProperties explicitly which is fine (though a comment regarding why it's necessary as described on this conversation may be useful to API users) When we have "Examples" rather than tests we should definitely add comments to explain the rationale for the call to avert the cognitive burden :roller_coaster:

aseth1:
> Based on the number of question marks, I guess it was a bad idea. 

Not at all. I was just pondering what would be most useful for `initModel()` to do? We could easily rename the remnants of `Model::setup()` and then use it more widely for `Model` instead of relying on the granular methods of `Component`.

tkuchida:
@aseth1 I don't want to hold up this PR. @aymanhab's comment inspired me to put on my "novice user" hat :tophat: and it occurred to me that telling the user to call `finalizeFromProperties()` is exposing a detail of the software design that doesn't appear to benefit the user-facing API. (Ideally, a question like "Why is it necessary to call finalizeFromProperties here?" would have an answer that's meaningful for the user, not something about calling virtual methods from a constructor or whatever.) At this point, it's probably most realistic to use the beta testing phase to find the wrinkles in the user experience and iron them out. :shirt:

aseth1:
> At this point, it's probably most realistic to use the beta testing phase to find the wrinkles in the user experience and iron them out.

I agree.

aseth1:
> "Why is it necessary to call finalizeFromProperties here?"

Good question. Loading a Model (or any Component) involves reading in its properties from a file. These properties include the key parameters (their values) as well as other parts of the Model (which are components). You need to call `finalizeFromProperties` to validate parameters in some cases (masses are >0, inertias are physical, mesh files are readable, ...) as well as to identify which properties are in fact its (sub)components (like Frames, Bodies, Joints, Constraints, ...) that form the Model. If you fail to `finalizeFromProperties` it will not be possible for the Model to know which of its properties are subcomponents.

tkuchida:
Yes, that explains why the operations performed by `finalizeFromProperties` are necessary, but not why OpenSim doesn't handle all that internally (e.g., calling `finalizeFromProperties` or performing its operations whenever necessary). Ideally, the novice user :tophat: shouldn't need to be concerned with taking care of OpenSim's internals. I think a user-centered design perspective might liken this situation to forcing the user to flip a switch to turn on the light in the fridge. But for now, maybe 4.0 users should just be happy that there's any light switch at all. :bulb:

aseth1:
@aymanhab @tkuchida I agree that it isn't ideal for the user to switch on the light when opening the fridge. My first priority is to make sure users can achieve the goals of constructing models and studies with an API and secondarily to make it easy to do that, but not at the cost of introducing inconsistency. It is not trivial and potentially error prone to implement a custom copy constructor, which is necessary to hide `finalizeFromProperties()` from users. We can revisit that when we establish that that the remainder of the API is stable. For the time being can we merge this PR?

tkuchida:
> My first priority is to make sure users can achieve the goals of constructing models and studies with an API and secondarily to make it easy to do that

Agreed. I've approved :white_check_mark: the PR; it can be merged IMHO.

PR #1513, opened 2017-02-10
Fix some CMC tests in Debug

chrisdembia:
Two of the CMC tests were failing in debug because the minimum step size for the integrator was set to an invalid value of 0. This PR fixes those two tests.

There is still one more CMC test that is failing in debug mode due to timing out; I have left that as ignored for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -147,7 +147,7 @@ before_install:
   ## On OSX, we know that testInitState fails; exclude it.
   - if [ "$TRAVIS_OS_NAME" = "osx" ]; then export TESTS_TO_EXCLUDE="$TESTS_TO_EXCLUDE|testInitState"; fi
   ## If we are building in debug, there are some tests to ignore.
-  - if [ "$BTYPE" = "Debug" ]; then export TESTS_TO_EXCLUDE="$TESTS_TO_EXCLUDE|testCMC|testOptimizationExample|testWrapping|testContactGeometry"; fi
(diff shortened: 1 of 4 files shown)


chrisdembia:
> Perhaps just set minimum_integrator_step_size to 1e-12?

Sounds good to me; will do.

chrisdembia:
Tests are passing. Looking for another reviewer.

aseth1:
It may avoid confusion if the default value in `AbstactTool` matched what is in the setup files as well. 
Currently,
`_minDT = 1.0e-8;`

Maybe change that to `1.0e-12` as well?

Otherwise LGTM and  :+1: for finding and sorting out the issue. Thanks!

chrisdembia:
Changing the default values will change people's results. I'd prefer to change the value in the CMC test setup files to 1e-8; is that fine?

tkuchida:
> I'd prefer to change the value in the CMC test setup files to 1e-8; is that fine?

:+1: (assuming the tests pass). I should have suggested to set `minimum_integrator_step_size` to _something small_, not to 1e-12 specifically.

aseth1:
> I'd prefer to change the value in the CMC test setup files to 1e-8; is that fine?

Yes. I think that would be consistent with what was being used by default.

chrisdembia:
I will re-ignore `testOptimizationExample*` on travis because it has been timing out sporadically on travis. This PR will still provide the benefit of no longer ignoring some of the CMC tests.

chrisdembia:
Thanks for the initial reviews. This is ready for review/merge.

chrisdembia:
Thank you @tkuchida

Issue #1488, opened 2017-01-24
Copying properties from an Object to another Object fails


aymanhab:
This is related to the discussion in PR #1460 a straightforward implementation that cycles through Properties and copy values if they match name and class fails for GeometryPath

jenhicks:
@aseth1 will investigate source of bug

aseth1:
I was looking into and was able to distill it to the `Property` assignment:
```cpp
        {
            Property<double>* a = Property<double>::TypeHelper::create("a", true);
            a->setValue(0.123456789);
            Property<double>* b = Property<double>::TypeHelper::create("b", true);
            b->setValue(10.0);

            *b = *a;

            cout << "b = " << b->toString() << endl;
            ASSERT(*a == *b);
        }
```
The ASSERT fails because Property `b` is not assigned to the property value(s) of `a` (although the property name is updated). 
Thanks @klshrinidhi for taking a look.

klshrinidhi:
I will get back soon on what is going on.

jenhicks:
Will merge this PR and then revisit script in #1460

jenhicks:
@aymanhab Would you like to test out the script again before the API beta or can I move this to the backlog?

aymanhab:
I'd move it to backlog. To set my environment to master in order to write/test the new script doesn't seem practical at the moment. Thank you.

jenhicks:
Related #1542

PR #1519, opened 2017-02-15
Disallow copy assignment & move assignment in Property | Add assign()

klshrinidhi:
Addressing #1488.

Because abstract class's copy assignment and move assignment operators cannot work polymorphically  through virtual function mechanism, disallow these operations.

### DIFF
--- a/OpenSim/Common/AbstractProperty.h
+++ b/OpenSim/Common/AbstractProperty.h
@@ -196,6 +196,7 @@ public:
     @returns The index assigned to this value in the list. **/
     template <class T> int appendValue(const T& value);
 
+    /** Assign (copy) property *that* to this object. */
     virtual void assign(const AbstractProperty& that) = 0;
     /**@}**/
 
--- a/OpenSim/Common/AbstractProperty.h
+++ b/OpenSim/Common/AbstractProperty.h
@@ -119,7 +119,7 @@ public:
         if (getComment() != other.getComment()) return false;
         if (getMinListSize() != other.getMinListSize()) return false;
         if (getMaxListSize() != other.getMaxListSize()) return false;
-     
+
         if (size() != other.size()) return false;
         // Note: we're delegating comparison of the "use default" flags
         // because only the new Property system copies them correctly, so
@@ -195,6 +195,8 @@ public:
     list isn't already of maximum size. 
     @returns The index assigned to this value in the list. **/
     template <class T> int appendValue(const T& value);
+
+    virtual void assign(const AbstractProperty& that) = 0;
     /**@}**/
 
     /** This method sets the "use default" flag for this property and the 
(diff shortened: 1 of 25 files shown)


aymanhab:
Thanks @klshrinidhi for investigating. I definitely agree that if methods/operators are not working they should be removed, however the built in copying seems to work for atomic types and properties other than ObjectProperty<>, was that just luck? Would it make sense to move the "disallowing" to the ObjectProperty if these are the only unsupported types of Properties? Or better yet, can we use template specialization to use "Clone" underneath for ObjectProperty? Thanks.

klshrinidhi:
> ... however the built in copying seems to work for atomic types and properties other than ObjectProperty<>, was that just luck? ...

Ajay reproduced the problem with property of `double`. You can see the snippet he posted in #1488.

I think it conceptually does not make sense to allow copy assignment and move assignment operators in abstract classes in general. To make the following work, 
```
Property<double>* a = ...;
Property<double>* b = ...;
*a = *b;
```
we will need to know the most derived type of `*a` in the assignment operator (templates may will not help here if I am correct) so that we can cast it to that type and use *in-place new* to replace the object that resides at the address `a`.

aseth1:
@klshrinidhi I can merge the test case I used to distill the problem. It includes Object and simple property assignments. Are you OK with that? You can than refine/strengthen as you see fit. I think it would be good to see the problem that was fixed with this PR clearly resolved and may better address @aymanhab's concerns.

klshrinidhi:
Sounds good. Go ahead.

aseth1:
The test fails because the default assignment operator on the `SimpleProperty` are ill-formed because the base has no default (even when casting to the `SimpleProperty`). I am unwinding the change. I think the correct behavior is to enable the default assignment but the utility used to copy properties has to correctly cast to the `SimpleProperty<T>` or `ObjectProperty<T>`. Unless you can think of something more clever.

aseth1:
@klshrinidhi I added test cases to show the issue with property copying. For Object copying the complete PropertyTable is deep copied, which is perfectly fine for Objects of the same type. In @aymanhab's `copyPropertiesFromObject` code we are trying to copy individual properties that are common across two Objects which may not be the same concrete type, so that is why we cannot copy the complete table. We can delete the property instances in the table of the `to` object, clone them from `from` and insert them in `to` to mimic the Object assignment operator . I think that will work and we might have to forgo using overloaded assignment on Property, unless you have a way of making it work.

klshrinidhi:
Got it. Will get back soon.

klshrinidhi:
I have added a method `AbstractProperty::assign()` to assign one AbstractProperty to another.
```
AbstractProperty* a = .....
AbstractProperty* b = .....

a->assign(*b); // Should work for both simple and object property.
```

aseth1:
@aymanhab I believe the `operator=` on the abstract base class should be protected, which is now the case. You cannot accidentally and incorrectly copy assign with `=`. The virtual `assign()` should now do the trick.

A future implementation of your `copyPropertiesFromObject()` will use `assign()`. The other issue is that the method has to be recursive, since the flags like `_valueIsDefault` on the top-level may not hold true for the properties of objects that are contained as this object's properties. That can be tackled as a subsequent PR.

aymanhab:
Sounds great. Thanks @klshrinidhi and @aseth1 I'll review and give feedback but the plan will do the job indeed.  Thank you.

klshrinidhi:
:two:  :heavy_multiplication_x:  :+1: :heavy_plus_sign: :three: :heavy_multiplication_x: :white_check_mark: = merge

Issue #1503, opened 2017-02-02
Component::printSubcomponentInfo(), printSocketInfo(), printInputInfo(), printOutputInfo()


chrisdembia:

tkuchida:
There might be some useful code in ExampleHopperDevice/helperMethods.h:
- [`showSubcomponentInfo()`](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Examples/ExampleHopperDevice/helperMethods.h#L37)
- [`showAllOutputs()`](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Examples/ExampleHopperDevice/helperMethods.h#L49)

chrisdembia:
I should have filled out the issue better...yes, I plan on using those functions, which have very nice output.

aseth1:
My preference *print* instead of *dump* and also like having the option to print to a user specified stream. Any other opinions?

chrisdembia:
@klshrinidhi please look at Component.h in the tgcs2017 branch as a starting point.

chrisdembia:
In the tgcs2017 branch, I already started moving the methods that @tkuchida mentions.

tkuchida:
> Any other opinions?

Recursive vs. non-recursive? Show types?

PR #1521, opened 2017-02-21
Add Component::printSubcomponentInfo()

klshrinidhi:
Addressing part of #1503.

Following is an example output:
```
 Class name and absolute path name for descendants of 'World' that are of type Component:
 
        [TheWorld]  /World
             [Sub]        /internalSub
             [Foo]        /Foo
             [Bar]        /Bar
             [Foo]        /Foo2
        [TheWorld]        /InternalWorld
             [Sub]                      /internalSub
             [Foo]                      /Foo
             [Bar]                      /Bar
             [Foo]                      /Foo2
        [TheWorld]        /World3
             [Sub]               /internalSub
             [Foo]               /Foo
             [Bar]               /Bar
             [Foo]               /Foo2
     [CompoundFoo]               /BigFoo
             [Foo]                      /Foo1
             [Foo]                      /Foo2
             [Bar]               /bar2
```

### DIFF
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -1507,47 +1507,6 @@ void Component::dumpConnections() const {
 void Component::printSubcomponentInfo() const {
     printSubcomponentInfo<Component>();
 }
-    
-template<typename C>
-void Component::printSubcomponentInfo() const {
-    std::string className = SimTK::NiceTypeName<C>::namestr();
-    const std::size_t colonPos = className.rfind(":");
-    if (colonPos != std::string::npos)
-        className = className.substr(colonPos+1, className.length()-colonPos);
-
-    std::cout << "Class name and absolute path name for descendants of '"
-              << getName() << "' that are of type " << className << ":\n"
-              << std::endl;
-
-    ComponentList<const C> compList = getComponentList<C>();
-
-    // Step through compList once to find the longest concrete class name.
-    unsigned maxlen = 0;
-    for (const C& thisComp : compList) {
-        auto len = thisComp.getConcreteClassName().length();
-        maxlen = std::max(maxlen, static_cast<unsigned>(len));
-    }
-    maxlen += 4; //padding
-
-    std::cout << std::setw(maxlen + 2) << "[" + getConcreteClassName() + "]"
-              << "  " << getAbsolutePathName() << std::endl;
-    auto prevPath = getAbsolutePathName();
-    // Step through compList again to print.
-    for (const C& thisComp : compList) {
-        const std::string thisClass = thisComp.getConcreteClassName();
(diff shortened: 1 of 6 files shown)


OpenSim/Common/Component.h
```diff
@@ -1434,14 +1434,18 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
     // End of Model Component State Accessors.
     //@} 
 
-    /** @name Dump debugging information to the console */
+    /** @name Print debugging information to the console */
```
chrisdembia: +1 for changing "dump". Based on a conversation I had with @aseth1, it might be nice to no longer think of these as for debugging. Could we change this to "Print information about this component and its subcomponents to the console."?

OpenSim/Common/Component.cpp
```diff
@@ -1524,6 +1504,50 @@ void Component::dumpConnections() const {
     }
 }
 
+void Component::printSubcomponentInfo() const {
+    printSubcomponentInfo<Component>();
+}
+    
+template<typename C>
+void Component::printSubcomponentInfo() const {
```
chrisdembia: As a note, putting this definition in the cpp means that users will not be able to invoke the templated version. IMO it might be nice to expose the templated version, but I don't have a strong opinion. If we do expose the templated version, it would be great if the python or java tests could make sure the non-templated version works.

klshrinidhi:
I have fixed the indent size to 4 spaces for each level. Using variable indent size may have too much whitespace for long names and high depth. The new output is as follows:
```
 Class name and absolute path name for descendants of 'World' that are of type Component:
 
        [TheWorld]  /World
             [Sub]      /internalSub
             [Foo]      /Foo
             [Bar]      /Bar
             [Foo]      /Foo2
        [TheWorld]      /InternalWorld
             [Sub]          /internalSub
             [Foo]          /Foo
             [Bar]          /Bar
             [Foo]          /Foo2
        [TheWorld]      /World3
             [Sub]          /internalSub
             [Foo]          /Foo
             [Bar]          /Bar
             [Foo]          /Foo2
     [CompoundFoo]          /BigFoo
             [Foo]              /Foo1
             [Foo]              /Foo2
             [Bar]          /bar2
```

OpenSim/Common/Component.h
klshrinidhi: Done.

OpenSim/Common/Component.cpp
klshrinidhi: Done.
I have moved the definition to header.

klshrinidhi:
I have also changed the implementation to replace use of `std::setw` with `std::string` to improve readability.

klshrinidhi:
Tests pass. Merging.

PR #1523, opened 2017-02-23
Rename Component::getParent to getOwner().

chrisdembia:
This PR renames `Component::getParent()` to `getOwner()` (similar for `hasParent()` and `setParent()`).


OpenSim/Common/Component.cpp
```diff
@@ -606,27 +606,28 @@ int Component::getNumStateVariables() const
...
-            "Has no parent Component assigned.\n" +
-            "Make sure the component was added to the Model (or its parent).";
+    if (!hasOwner()) {
+        std::string msg = "Component '" + getName() + "'::getOwner(). " +
+            "Has no owner assigned.\n" +
+            "Make sure the component was added to the Model " +
+            "(or another component).";
         throw Exception(msg);
     }
     return _parent.getRef();
```
tkuchida: `_parent` -> `_owner`?

OpenSim/Common/Component.cpp
```diff
@@ -606,27 +606,28 @@ int Component::getNumStateVariables() const
...
 }
 
-bool Component::hasParent() const
+bool Component::hasOwner() const
 {
     return !_parent.empty();
 }
 
-void Component::setParent(const Component& parent)
+void Component::setOwner(const Component& parent)
```
tkuchida: `const Component& parent` -> `const Component& owner`?

OpenSim/Common/Component.cpp
```diff
@@ -1519,7 +1520,7 @@ void Component::initComponentTreeTraversal(const Component &root) const {
     const size_t npsc = _propertySubcomponents.size();
     const size_t nasc = _adoptedSubcomponents.size();
 
-    if (!hasParent()) {
+    if (!hasOwner()) {
         // If this isn't the root component and it has no parent, then
```
tkuchida: parent -> owner

OpenSim/Common/Component.h
```diff
@@ -2195,19 +2195,21 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
     const StateVariable* findStateVariable(const std::string& name) const;
 #endif
 
-    /** Access the parent of this Component.
-        An exception is thrown if the Component has no parent.
-        @see hasParent() */
-    const Component& getParent() const;
+    /** Access the owner (parent component) of this Component.
```
tkuchida: Might be confusing to leave this "owner = parent" idea floating around.

OpenSim/Common/Component.h
```diff
@@ -2195,19 +2195,21 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
+        A component may not have an owner if it:
+        (1) is the root component, or
+        (2) has not been added to another component */
+    bool hasOwner() const;
 
 protected:
-    /** %Set this Component's reference to its parent Component */
-    void setParent(const Component& parent);
+    /** %Set this Component's reference to its owning Component */
+    void setOwner(const Component& parent);
```
tkuchida: `const Component& parent` -> `const Component& owner`?

OpenSim/Simulation/Model/Geometry.cpp
```diff
@@ -217,18 +217,18 @@ void Mesh::extendFinalizeFromProperties() {
 
     if (!isObjectUpToDateWithProperties()) {
         const Component* rootModel = nullptr;
-        if (!hasParent()) {
+        if (!hasOwner()) {
             std::cout << "Mesh " << get_mesh_file() << " not connected to model..ignoring\n";
             return;   // Orphan Mesh not part of a model yet
         }
-        const Component* parent = &getParent();
+        const Component* parent = &getOwner();
```
tkuchida: `const Component* parent` -> `const Component* owner`?

OpenSim/Simulation/Wrap/PathWrap.cpp
```diff
@@ -83,7 +83,7 @@ void PathWrap::extendConnectToModel(Model& model)
 {
     Super::extendConnectToModel(model);
 
-    _path = dynamic_cast<const GeometryPath*>(&getParent());
+    _path = dynamic_cast<const GeometryPath*>(&getOwner());
     std::string msg = "PathWrap '" + getName()
         + "' must have a GeometryPath as its parent.";
```
tkuchida: "as its parent" -> "as its owner"

chrisdembia:
Thank you for reviewing, @tkuchida . I had to fix an issue related to the fact that `GeometryPath` has a `getOwner()`. I chose to remove `const Object& GeometryPath::getOwner()`, since I think `const Component& Component::getOwner()` removes the need for the `GeometryPath` method. The GUI does use `GeometryPath::getOwner()` in a number of places, but it seems like only minimal changes would be needed, since `Component` is derived from `Object`.

@aymanhab

aymanhab:
If the new method on Component has the same semantics and signature as the method used for GeometryPath then there should be no adverse effects on GUI. 

A different concern that may have been discussed somewhere else is that there are three lists maintained by component (members, adopted, property components) what is the expected return value when calling "getOwner" on members of these lists, and does ownership imply memory management (allocation of resources, deletion, etc.)?

chrisdembia:
> If the new method on Component has the same semantics and signature as the method used for GeometryPath then there should be no adverse effects on GUI.

I think so, but I don't think I know enough Java to say for sure. For example, if you have a function `Derived my_func()`, where `Derived` is derived from `Base`, can you do an assignment like `Base b = my_func()` in Java? My guess is yes.


> A different concern that may have been discussed somewhere else is that there are three lists maintained by component (members, adopted, property components) what is the expected return value when calling "getOwner" on members of these lists, and does ownership imply memory management (allocation of resources, deletion, etc.)?

For all 3 lists, the "owner" does manage the memory.

aymanhab:
Thanks @chrisdembia for the clearup. 

I don't remember why we're replacing an easy to explain graph concept (a tree that has root, nodes with single-parent, children, descendents, etc.) with a more subtle programming concept (of ownership, managing resources) which has more connotation to programmers than users, but that's beyond the scope of this PR, so please feel free to merge.

chrisdembia:
> I don't remember why we're replacing an easy to explain graph concept (a tree that has root, nodes with single-parent, children, descendents, etc.) with a more subtle programming concept (of ownership, managing resources) which has more connotation to programmers than users, but that's beyond the scope of this PR, so please feel free to merge.

That is a very good point. The immediate issue was that some components have methods like `getParentFrame()` and `getParent()`, which is very confusing. I really do think that comprehensively adopting the language of a tree would help explain the API. For example, I've argued that `getComponent()` could be renamed `getChild()`, in which case we would want to keep `getParent()`. Do you have alternate suggestions for how to fix the issue of confusing "parent component" with "parent frame"/"parent body"?

We could rename the function to `getParentComponent()`? or `getParentInTree()`? But there's also the multibody tree...ahhhh! 😩

tkuchida:
> Do you have alternate suggestions for how to fix the issue of confusing "parent component" with "parent frame"/"parent body"?

We could adopt the MapleSim convention of "inboard" and "outboard" frames (see example [here](http://www.maplesoft.com/support/help/MapleSim/view.aspx?path=componentLibrary%2fmultibody%2fjointsAndMotions%2fPlanar)).

aseth1:
IMHO *owner* is the correct term since we use it describe the ownership tree. I don't think the

>  graph concept (a tree that has root, nodes with single-parent, children, descendents, etc.) 

is more or less well understood by our users. Although you may like using descendants I think it is less than ideal since it is ambiguous with inheritance. Is `GeometryPath` a descendant of `Muscle`? Seems weird to say that. I am more inclined to say that `Muscle` is a descendant of `Actuator`. When talking about the collection of components owned by a Component, why not refer to what we have been calling them all along *subcomponents*? A Component owns its subcomponents.

I like *inboard* and *outboard* and it came up in discussions for `Joint` before. The only issue is that we used to have `parent_body` and `child_body` properties of the `Joint` and then we generalized to `PhyscialFrame`s and renamed `parent_frame` and `child_frame` sockets. I think that is an easier transition than `parent_body` -> `inboard` but if we want to make that change, now would be the time.

aymanhab:
Thanks @aseth1 for chiming in. I just don't think that inheritance hierarchy is near the top of the list in users' minds when building models, and we shouldn't frame it as such anyway. The concept is what "kinds of" blah exist which is user/modeler-centered. 
If we have a consensus on inboard/outboard terminology change for frames then let's do it there instead of the broader change but it's your call since I don't know what else is involved across the API.

chrisdembia:
Perhaps we can make a final decision in the dev team meeting?

jenhicks:
We'll go with getOwner and document as advanced.

aymanhab:
@chrisdembia Feel free to merge once conflicts are resolved. Thanks.

chrisdembia:
I have updated the documentation. This is ready for review again.

tkuchida:
All changes since my :white_check_mark: look good. Only potential issue is still calling a Component with no *owner* an "orphan", but I don't think this causes any confusion; merging.

chrisdembia:
Thank you.

Issue #578, opened 2015-08-11
Update testMomentArms with additions from opensim32


jenhicks:
https://github.com/opensim-org/opensim32/pull/73

jenhicks:
Can you update testMomentArms @aseth1 ?

tkuchida:
Fixed by #1524.

PR #1524, opened 2017-02-24
Add moment arm test for a joint comprised of a plurality of bodies

tkuchida:
Addresses #578. Uses the same CompoundJoint as [testJoints.cpp](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/SimbodyEngine/Test/testJoints.cpp#L154). I can move the class to a common location if a suitable one exists. The only candidate I'm aware of is [auxiliaryTestFunctions.h](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Auxiliary/auxiliaryTestFunctions.h), but it doesn't seem desirable to add `#include` overhead there for Joint.

### DIFF
--- a/OpenSim/Simulation/SimbodyEngine/Test/testJoints.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/Test/testJoints.cpp
@@ -70,7 +70,7 @@
 #include <OpenSim/Common/FunctionAdapter.h>
 #include <OpenSim/Auxiliary/auxiliaryTestFunctions.h>
 
-#include <simbody/internal/MobilizedBody_BuiltIns.h>
+#include <OpenSim/Simulation/Test/SimulationComponentsForTesting.h>
 
 using namespace OpenSim;
 using namespace std;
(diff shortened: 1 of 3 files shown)


tkuchida:
> I think it would be nice to avoid the duplicate header file.

Which .h file is duplicated?

tkuchida:
> I think it would be nice to avoid the duplicate header file.

There aren't any duplicate .h files introduced in this PR and I'm not suggesting to create any. I agree: duplicate .h files would be bad. I'll follow the example of ComponentsForTesting.h and move CompoundJoint to a common location.

chrisdembia:
Oops; I meant duplicate class.

tkuchida:
> I think it would be nice to avoid the duplicate header file.

Moved definition of CompoundJoint class to a separate file.

tkuchida:
[Timeout on Travis in Debug mode](https://travis-ci.org/opensim-org/opensim-core/builds/205986818).

PR #1525, opened 2017-02-26
Update col_label renaming for opensimTimeSeriesTableToMatlab

chrisdembia:
I have updated the way that column labels get reformatted by @jimmyDunne's opensimTimeSeriesToMatlab.m script. Since the script was made, the format for output paths has changed to include a `|` character; this PR accounts for the `|` character. 

This PR also uses the full output path rather than only the leaf, since multiple columns could have the same leaf.

I forgot to skip the CI tests, but they are irrelevant for this PR (the sandbox is not tested).

I made these changes during the development of the TGCS exercise.

@jimmyDunne

### DIFF
--- a/OpenSim/Sandbox/MatlabScripts/opensimTimeSeriesTableToMatlab.m
+++ b/OpenSim/Sandbox/MatlabScripts/opensimTimeSeriesTableToMatlab.m
@@ -26,13 +26,14 @@ for iCol = 0 : nCol -1
 
     col_label  = char(table.getColumnLabels.get(iCol));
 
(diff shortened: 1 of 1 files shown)


jimmyDunne:
Where did you get this script from? This is current the set of utilities that I was working on. The table writing utilities is at #1437

chrisdembia:
The file is already in the master branch.

chrisdembia:
Oh I see; I should be using the newer version.

chrisdembia:
@jimmyDunne I am closing this since I had been updating an older file.

PR #1526, opened 2017-02-26
In bindings, use takeKeyHit() instead of waitForKeyHit(). Update Simbody commit

chrisdembia:
This PR follows up on #1506. Initially, I had wrapped SimTK's `waitForKeyHit()` for use in MATLAB. However, if I killed the simbody-visualizer before hitting any keys, then MATLAB would hang forever at `waitForKeyHit()`, and Ctrl+C would not interrupt the function call; the only way to recover was to kill MATLAB.

So, it is necessary to wrap `takeKeyHit()` instead. By not wrapping `waitForKeyHit()` we help other users avoid the same issues I ran into.

This PR also updates the commit for the Simbody dependency to include a bug fix wherein the visualizer would cause MATLAB to crash (https://github.com/simbody/simbody/pull/548).

@aymanhab

### DIFF
--- a/Bindings/simbody.i
+++ b/Bindings/simbody.i
@@ -246,6 +246,11 @@ namespace SimTK {
 %include <simbody/internal/Visualizer_InputListener.h>
 // The following is necessary because the BackgroundType enum cannot be used
 // from MATLAB.
+// The new version of takeKeyHit() allows returning the value rather than using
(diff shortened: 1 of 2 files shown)


chrisdembia:
Thank you @klshrinidhi

chrisdembia:
@nickbianco and I have both successfully used this change locally.

chrisdembia:
Thank you.

PR #1527, opened 2017-02-26
Improve access to LoadOpenSimLibrary() in bindings.

### LINKED ISSUE #669: In Java/MATLAB, when Model.LoadOpenSimPlugin fails to load a proper library, there is no error
@aseth1 ran into this issue.

Relatedly, there should be a better error when XML parsing hits an unrecognized/unregistered type.
Also I think we should move Model.LoadOpenSimPlugin to be a global method (maybe that's not possible in Java?)

chrisdembia:
This PR improves the ability to load OpenSim plugin libraries in MATLAB, and introduces this ability in Python.

There are two functions for loading libraries: `LoadOpenSimLibrary` which does not take the library extension, and `LoadOpenSimLibraryExact` which does take the library extension. Both are available in the bindings:

MATLAB:
```
import org.opensim.modeling.*;
opensimCommon.LoadOpenSimLibrary('osimCommon');
opensimCommon.LoadOpenSimLibraryExact('osimCommon.dll');
```

Python:
```
import opensim
opensim.LoadOpenSimLibrary('osimCommon')
opensim.LoadOpenSimLibraryExact('osimCommon.dll')
```

I made sure that an error message is generated if the plugin library is not loaded. I also added doxygen documentation describing how to use these functions in the bindings.

I changed the default behavior of `LoadOpenSimLibrary()` to be verbose.

I also renamed the Java modules to not contain 'Model' in their names (e.g., `opensimModelCommon` -> `opensimCommon`). I hope this is okay, but if not, I can undo this change. I feel that these new names are clearer for MATLAB users (e.g., there is no Model in the Common library), even though the names are more similar to those used in the GUI codebase and could cause confusion there.

Fixes #669.

@aymanhab

### DIFF
--- a/Bindings/Java/swig/java_actuators.i
+++ b/Bindings/Java/swig/java_actuators.i
@@ -1,5 +1,5 @@
-%module(directors="1") opensimModel
-%module opensimModel
+%module(directors="1") opensimActuatorsAnalysesTools
+%module opensimActuatorsAnalysesTools
 
 #pragma SWIG nowarn=822,451,503,516,325,401
 
--- a/Bindings/Java/swig/java_common.i
+++ b/Bindings/Java/swig/java_common.i
@@ -1,5 +1,5 @@
-%module(directors="1") opensimModelCommon
-%module opensimModelCommon
+%module(directors="1") opensimCommon
+%module opensimCommon
 
 #pragma SWIG nowarn=822,451,503,516,325,401
 
--- a/Bindings/Java/swig/java_simulation.i
+++ b/Bindings/Java/swig/java_simulation.i
@@ -1,5 +1,5 @@
-%module(directors="1") opensimModelSimulation
-%module opensimModelSimulation
+%module(directors="1") opensimSimulation
+%module opensimSimulation
 
 #pragma SWIG nowarn=822,451,503,516,325,401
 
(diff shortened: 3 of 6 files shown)


aymanhab:
The Doxygen comments recommend using "LoadOpenSimLibraryExact"  over the older variant "LoadOpenSimLibrary", what happens to "dependent" libraries? My guess they'll load based on PATH or DYLD_PATH even if exact path is provided. Probably not a major issue since most plugins are made of a single dynamic library anyway but good to check/document.

chrisdembia:
Thank you @aymanhab for reviewing. 



> (not sure if user code/examples would be an issue or if any published code uses the old module names).

Since the names were changed recently (when the java bindings were split into 3 modules), I don't think we have used these module names in MATLAB examples/documentation.

> what happens to "dependent" libraries? 

Good question! Certainly, someone's plugin could depend on some other C++ library, like an optimizer library. I'm not sure how this would work. But I don't think the two variants (`LoadOpenSimLibrary` and `LoadOpenSimLibraryExact`) differ in how they handle this scenario, and so I think this issue is unrelated to this PR. I can add vague documentation like "Make sure that you make available any other libraries your plugin depends on."

chrisdembia:
@aymanhab I have added the doxygen blurb we discussed.

aymanhab:
Thanks @chrisdembia

chrisdembia:
Thanks @aymanhab

PR #1529, opened 2017-02-26
Split and shorten testCMCGait10dof18musc.

chrisdembia:
This PR duplicates the effort in #1191 ; oops. 

I split testCMCGait10dof18musc into two tests, one for each of the two muscle models. I also shortened the duration of the CMC run, but not so much that the tolerances need to change.

We should either merge this one or #1191, but not both. Also, if shortening CMC final time is controversial because of issues related to the tolerance, I would prefer to undo that change and merge only the split into two test cases.

### DIFF
--- a/Applications/CMC/test/gait10dof18musc_Setup_CMC.xml
+++ b/Applications/CMC/test/gait10dof18musc_Setup_CMC.xml
@@ -8,13 +8,13 @@
 		<!--List of xml files used to construct an force set for the model.-->
 		<force_set_files> gait10dof_Reserve_Actuators.xml</force_set_files>
 		<!--Directory used for writing results.-->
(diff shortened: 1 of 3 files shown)


aseth1:
I prefer not splitting off these tests by type because of the code duplication and the maintenance challenges it imposes in keeping the "parts" in sync. It also doesn't shorten the total test time, it does thwart timeouts, that I agree. I prefer shortening the actual test as you've done. If that doesn't pass in debug we can address shortening even further and justify adjusting the tolerances.

chrisdembia:
The shorter tests do pass in debug. Do you want me to undo the splitting?

aseth1:
> The shorter tests do pass in debug. Do you want me to undo the splitting?

I would prefer that, yes. If you don't mind doing that.

PR #1536, opened 2017-02-28
added Matlab/Utilities dir, transferred table files

jimmyDunne:


chrisdembia:
Do you want to distribute the test file with the distribution? I think the test file should go into the `Bindings/Java/Matlab/tests` directory. We could also set it up so that CMake will run that test; by editing

jimmyDunne:
I moved the tests. Since the test is in a different folder, I will have to change the paths to input files. 

Will create an issue for myself so that the PR is not held up.

PR #1539, opened 2017-03-01
Avoid debug assert in Component::printSubcomponentInfo.

chrisdembia:
This PR is part of fixing failures in debug builds.

Previously, `printSubcomponentInfo()` used `std::mismatch()`, which, in
C++11, expects that the second container is at least as long as the
first one. Since this is not true for this usage of `std::mismatch()`, I
replaced the code with the use of `ComponentPath`.

### DIFF
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -1504,21 +1504,15 @@ public:
         std::cout << std::string(maxlen-getConcreteClassName().length(), ' ')
                   << "[" + getConcreteClassName() + "]"
                   << "  " << getAbsolutePathName() << std::endl;
-        auto prevPath = getAbsolutePathName();
+
         // Step through compList again to print.
         for (const C& thisComp : compList) {
             const std::string thisClass = thisComp.getConcreteClassName();
             std::cout << std::string(maxlen-thisClass.length(), ' ') << "["
                       << thisClass << "]  ";
-            auto path = thisComp.getAbsolutePathName();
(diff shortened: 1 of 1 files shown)


klshrinidhi:
The behavior is undefined only if we reach the end of second range while there are still elements in the first range. Because of the nature of inputs (paths) to `mismatch`, this condition is never reached. I did not build in Debug. Can you post the message here ?

chrisdembia:
> Because of the nature of inputs (paths) to mismatch, this condition is never reached.

This condition is reached because subsequent paths can be shorter than earlier paths, as the iterator walks the tree in depth-first order.

The issue only arose with MSVC. The message was something like "Expression: string iterator not dereferencable." Perhaps the implementation of `std::mismatch` in gcc / clang handles the situation differently, and just returns the mismatch as occurring when the second iterator gives null.

klshrinidhi:
Okay. What I meant is that we don't reach the end of input paths without finding a mismatch. If I am correct, there will always be a *successful* mismatch between any two paths.

Anyway, go ahead with merge.
:+1:

chrisdembia:
Oh I see. Since we are supposed to have unique names at a given level in the tree, this should be correct. However, I found that some of the muscle wrapping components still do not have unique names. :/ So we have things like this:

```
/A/B/C
/A/B/C/D
/A/B/C      <--- std::mismatch gives an error when this is the second range.
/A/B/C/D
```

klshrinidhi:
Okay. Sounds good. I made wrong assumptions about the input.
:+1:

PR #1540, opened 2017-03-01
Shorten testCMCGait10dof18musc.

chrisdembia:
This PR replaces #1529 and only shortens testCMCGait10dof18musc enough so that it does not time out.


chrisdembia:
Thank you.

PR #1543, opened 2017-03-01
Remove unused Reporter Components doxygen page.

chrisdembia:
We already have a page for the group of Reporters. We don't need an additional
"page" for reporters. The page was empty anyway.


tkuchida:
doxygenreportercomponents

chrisdembia:
Thank you.

PR #1546, opened 2017-03-01
Reset Component::_propertySubcomponents after copy.

chrisdembia:
In PR #1511 @aseth1 did a good job of throwing exceptions if the user had not called `finalizeFromProperties()` after a copy. However, I found a situation that fell through his checks. This PR addresses that situation.

The situation was that after copying, the member variable `Component::_propertySubcomponents`, of type `SimTK::Array_<SimTK::ReferencePtr<Component>>`, would have non-zero size but all the reference pointers inside it were null. Therefore, the check that [`!(nmsc + npsc + nasc)`](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Component.cpp#L1595) returned was false even though we expected it to be true.

The fix was to reset the entire `SimTK::Array_` upon copy, so that it has no entries, thereby causing `!(nmsc + npsc + nasc)` to be true after copying, as expected.

I started this PR in my quest to fix test failures in Debug. Without this fix, `testContactGeometry` fails on Windows in Debug. I have verified locally that this PR fixes `testContactGeometry` on Windows in Debug. I have also added a test to make sure that an exception is thrown if `finalizeFromProperties()` is not invoked after copying a component that has a property subcomponent.

### DIFF
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -2686,7 +2686,8 @@ private:
     SimTK::ResetOnCopy<SimTK::MeasureIndex> _simTKcomponentIndex;
 
     // list of subcomponents that are contained in this Component's properties
-    SimTK::Array_<SimTK::ReferencePtr<Component> >  _propertySubcomponents;
+    SimTK::ResetOnCopy<SimTK::Array_<SimTK::ReferencePtr<Component>>>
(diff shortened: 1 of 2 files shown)


aseth1:
Nice sleuthing 🕵️. Thanks for the find and the fix! :+1:

Issue #1365, opened 2016-10-27
Excitation is not clamped (in many cases)


carmichaelong:
From what I can see, excitation is finally clamped when `FirstOrderMuscleActivationDynamics::calcActivationDerivative()` is called. This can cause issues with things like the `Umberger2010MuscleMetabolicsProbe` since it needs excitations for calculations. A user can set the excitation of a muscle to be outside of the [0,1] range, thinking that it will be clamped at some point, which can lead to issues with the probe.

At what point should excitation be clamped?

tkuchida:
**A word of caution:** `MuscleActivationDynamics` and the classes derived therefrom (`ZerothOrderMuscleActivationDynamics` and `FirstOrderMuscleActivationDynamics`) aren't being tested AFAIK and may be out of date. I originally intended to replace the existing activation dynamics code (e.g., `MuscleFirstOrderActivationDynamicModel` and the pieces sprinkled throughout the muscles) with these, which would make it easy to change the activation model without creating an entirely new muscle (see #366)—though there are differing opinions about what should and shouldn't be configurable (see the discussion in #1266).

**To answer your question:** Classes derived from `MuscleActivationDynamics` were designed to require an `ExcitationGetter` (see [this comment](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Actuators/MuscleActivationDynamics.h#L58)). The `ExcitationGetter::getExcitation()` method should clamp before returning.

jenhicks:
@carmichaelong Is this still an issue for you? Is there a bug or is the behavior just not as expected?

carmichaelong:
I think the best way to describe this is that the behavior may be unexpected for other classes that look up muscle excitations during a simulation, which could be seen as a bug for these other classes. To give a concrete example:

1) Someone can set a muscle's control value to anything (even outside of [0,1]), assuming that it will be clamped at some point. This number is stored as is though (so if you call `getControl()` on an actuator that was set to -0.5, you'll get back -0.5).
2) When calculating activation, some methods do clamp the excitation internally (see [here](https://github.com/opensim-org/opensim-core/blob/dc17ff08e0b40311c984f978b1462f7e31e2756a/OpenSim/Actuators/MuscleFirstOrderActivationDynamicModel.cpp#L82) for example). In this case, our value of -0.5 will be clamped up to `get_minimum_activation()`. Note that the control value is still left unchanged, but does not affect calculations for determining muscle force.
[...]

PR #1548, opened 2017-03-02
Remove excitation clamping from muscles

aseth1:
Addresses issues raised in #1365 by eliminating the internal (hidden) clamping of muscle excitation by the model. 

**Summary of changes:**

0. removed internal clamping of the excitation by the ActivationDynamicModel
1. a consequence of removing the internal clamp is that the suppliers of excitation must respect the minimum acceptable excitation. Since the internal clamping was set to the minimum activation the min control property for muscles with activation dynamics now defaults to the minimum activation value.
2. in model files where min and max controls were exposed as 0.0 and 1.0 these lines were removed to use the default values
3. `ControllerExample` uses min (and max) control to set controls instead of 0.0 (and 1.0)
4. the optimization example was failing to find the same optimal solution mainly because solutions diverged at the point where the controls fell bellow the minimum (which were being clamped internally). Perturbing the initial guess (*0.01 (the min) ->0.02*)  improved the optimal solution (speed 4.6->4.9) but it was then taking over 400 evaluations to converge (~5min). This was evaluated with and without clamping in place to make sure clamping was not significantly affecting the solution. 
5. to improve the run-time performance of the optimization example, I loosened the integration accuracy (from 1e-7) to 1e-5 to improve objective function evaluation time,   but it also unexpectedly reduced the number of function evaluations (which I assumed would stay about the same) to <100 evaluations. The test now takes about 30s to run in release. I hope these improvements extend to  CI.
6. `testOptimizationExample` was updated to reflect the higher hand velocity and changes to controls. The standard/reference controls are from the optimal solution with clamping in place.
7.  `testOptimizationExample` 's regression test for the resulting forward simulation (using solution controls)  was also removed since it is likely for the optimal control solution to be slightly different on different platforms resulting in different simulations and it is redundant with `testForward`'s `testArm26` case as a regression for forward simulation with the same model.


OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -55,6 +55,8 @@ void Millard2012EquilibriumMuscle::constructProperties()
     constructProperty_ForceVelocityCurve(ForceVelocityCurve());
     constructProperty_FiberForceLengthCurve(FiberForceLengthCurve());
     constructProperty_TendonForceLengthCurve(TendonForceLengthCurve());
+
+    setMinControl(get_minimum_activation());
```
tkuchida: :warning: The `minimum_activation` property can be modified in `extendFinalizeFromProperties()`. I think `setMinControl();` needs to be there (e.g., at L123). Perhaps something like
```cpp
setMinControl( std::max(get_minimum_activation(), get_min_control()) );
```
in case the user provided a non-default value for `min_control`?

OpenSim/Examples/ControllerExample/tugOfWar_model_ThelenOnly.osim
```diff
@@ -281,10 +281,6 @@
 				<Thelen2003Muscle name="muscle1">
 					<!--Flag indicating whether the force is disabled or not. Disabled means that the force is not active in subsequent dynamics realizations.-->
 					<isDisabled>false</isDisabled>
-					<!--Minimum allowed value for control signal. Used primarily when solving for control values.-->
```
tkuchida: Do any tests cover the cases where (i) `0 < min_control < 0.01` or (ii) `0.01 < min_control < 1` to ensure the user-supplied `min_control` isn't just being ignored/overwritten?

OpenSim/Simulation/Model/Muscle.cpp
```diff
@@ -47,7 +47,7 @@ Muscle::Muscle()
 {
     constructProperties();
     // override the value of default _minControl, _maxControl
-    setMinControl(0.0);
+
```
tkuchida: Stale comment needs refreshing.

aseth1:
> I'm not sure that a user-specified value for min_control will be respected if setMinControl(get_minimum_activation()); is called from constructProperties()

Yes, user settings either in the model file or made by calling `setMinControl()` overwrite the default min activation. That's why I had to remove their `min`/`max control` 0.0 and 1.0 property settings from the model file (so it would not overwrite with 0.0 as the min).

tkuchida:
> Yes, user settings either in the model file or made by calling `setMinControl()` overwrite the default min activation.

Then couldn't a user set `min_control` to a value that will cause the muscle model to kerplode? I thought there would need to be a check in `extendFinalizeFromProperties()` to avoid that, e.g., taking the maximum of the user's request and the minimum allowable.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
aseth1: This only sets the default value for `min_control`. Deserialization populates the Object properties and overwrites the property values for those specified in the model (XML) file.

OpenSim/Examples/ControllerExample/tugOfWar_model_ThelenOnly.osim
aseth1: I can add these. The one issue I see is that if you change the `minimum_activation` in the model file, the `min_control` will still be the default `minimum_activation`.  Therefore, the user has to update both as it is now.

aseth1:
> Then couldn't a user set min_control to a value that will cause the muscle model to kerplode? I thought there would need to be a check in extendFinalizeFromProperties() to avoid that, e.g., taking the maximum of the user's request and the minimum allowable.

You are right, I didn't address it. There are a couple of options: 1) reset the `min_control` to an acceptable value as you suggest or 2) throw an Exception that an invalid property value was provided. I prefer the Exception over silently changing a user setting that is mistaken. Which do you prefer?

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
tkuchida: IIUC, the `min_control` and `minimum_activation` properties could both be set during deserialization. It seems that the following constraints would need to be enforced post-deserialization (i.e., in `extendFinalizeFromProperties()`:

1. `LB <= minimum_activation <= 1`, where `LB` is 0 or 0.01 depending on Millardian flags
2. Let `userPropVal = get_min_control()`. Then `std::max( userPropVal, minimum_activation ) <= min_control <= 1`

(1) is enforced at L107 and L116, but (2) is not.

tkuchida:
> Which do you prefer?

Ah, yes I agree: throwing in the event of incompatible properties sounds best.

OpenSim/Examples/ControllerExample/tugOfWar_model_ThelenOnly.osim
tkuchida: The desired user interaction (and behind-the-scenes behavior) isn't so clear. This scenario is also relatively unlikely to occur, so I'd suggest adding this to the list of tests we will add post-beta to avoiding holding up this PR.

chrisdembia:
> Ah, yes I agree: throwing in the event of incompatible properties sounds best.

Seconded; throw an exception! If this causes issues for users, then we can be more lenient.

carmichaelong:
> Since this arose out of an issue that @carmichaelong had, he may be interested in seeing if these changes work well with his predictive simulation code.

We've been using OpenSim 3.3, so I can't check directly. I wouldn't worry about it, though, because it would be very easy to clamp on the controller side.

tkuchida:
:+1: Changes look great. @chrisdembia: please merge unless you have comments on the commits made after you approved. :duck:

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
tkuchida: Addressed by 951f397.

aseth1:
Thank you @tkuchida and @chrisdembia. @chrisdembia let me know if you have any additional comments.

OpenSim/Simulation/Model/Muscle.cpp
```diff
@@ -115,6 +139,10 @@ void Muscle::constructProperties()
     constructProperty_max_contraction_velocity(10.0);
     constructProperty_ignore_tendon_compliance(false);
     constructProperty_ignore_activation_dynamics(false);
+
+    // By default the min and max controls on muscle are 0.0 and 1.0
+    setMinControl(0.0);
+    setMaxControl(1.0);
```
chrisdembia: This already happens in Muscle's constructor. Either remove the calls from the constructor or from here?

OpenSim/Simulation/Model/Muscle.cpp
```diff
@@ -96,6 +93,33 @@ void Muscle::updateFromXMLNode(SimTK::Xml::Element& aNode, int versionNumber)
...
+                aNode.element_begin("min_control");
+            double minControl = 0;
+            if (minControlElt != aNode.element_end()) {
+                minControlElt->getValueAs<double>(minControl);
+
+                // remove the min_control property in XML since it is a result
+                // of a mistake with the previous versions not updating
+                // the min_control to reflect the min_activation. Removing it
+                // allows the Muscle to use the default that is specified by the
+                // derived concrete Muscle
```
chrisdembia: Could you clarify the comment to say that the property is only removed if it has the value 0?

OpenSim/Simulation/Model/Muscle.cpp
aseth1: They were already removed [here](https://github.com/opensim-org/opensim-core/pull/1548/files#diff-45af1cee4e076a4870ab4f15d30b7d6eL49).

tkuchida:
:fireworks:

PR #1549, opened 2017-03-02
Avoid compiling stale .java files when building newer commits.

### LINKED ISSUE #1547: CMake should clear generated Java source files before rebuilding Java bindings
For those building the java bindings, the following happens often:

1. Build the Java bindings (this puts a bunch of swig-generated .java files in the build tree, at Bindings/Java/src).
2. Update opensim-core to a newer commit.
3. Rebuild the Java bindings (this creates new .java files).
4. Get confusing errors saying that the java classes could not be compiled.

I think this is because there are leftover .java files from the previous build (1) that should no longer exist under the new build (3). For example, this would happen if a class is renamed or removed.

We should have cmake delete all .java files in Bindings/Java/src before generating new \*.cxx and \*.h files.

chrisdembia:
This fixes #1547. Java source files are deleted before they are regenerated,
avoiding the possibility of compiling stale java files from previous builds.

### DIFF
--- a/Bindings/Java/CMakeLists.txt
+++ b/Bindings/Java/CMakeLists.txt
@@ -63,9 +63,25 @@ add_subdirectory(OpenSimJNI)
 # ---------------------
 find_package(Java 1.7 REQUIRED)
 
+# To avoid compiling stale .java files, delete any existing files and copy
+# over the new files from the module-specific directories (see
+# OpenSimJNI/CMakeLists.txt).
 add_custom_command(
     OUTPUT "${SWIG_JAVA_JAR_BUILD_OUTPUT_PATH}"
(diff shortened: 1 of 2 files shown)


chrisdembia:
I tested this locally and the builds passed. This could be a quick review. Any takers?

jimmyDunne:
I can test!

chrisdembia:
Woohoo! Thank you @jimmyDunne.

jimmyDunne:
I was able to make everything (inc Java/Matlab) without error. Since usually, I have to delete my build directory everytime I build a different branch, I would say this was a success!

Anything else I can test?

chrisdembia:
Great. That sounds good to me.

PR #1559, opened 2017-03-08
Use _WIN32 instead of WIN32 macro.

chrisdembia:
I recently tried out Visual Studio 2017, which comes with an updated C++ compiler (VC++ v141). the only build error I got was that `WIN32` was not always defined when we expected it to be. However, we should not rely on `WIN32`; the correct macro to use is `_WIN32`.

References:

- https://msdn.microsoft.com/en-us/library/b0084kay.aspx
- http://stackoverflow.com/questions/662084/whats-the-difference-between-the-win32-and-win32-defines-in-c

### DIFF
--- a/OpenSim/Actuators/osimActuatorsDLL.cpp
+++ b/OpenSim/Actuators/osimActuatorsDLL.cpp
@@ -52,7 +52,7 @@ static void Plugin_Detach()
 //
 // The code below handles both Windows and Linux library entrypoints
 //
-#if defined(WIN32)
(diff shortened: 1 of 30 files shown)


chrisdembia:
This should be a quick review. The travis-ci/pr build passed, so we can ignore the failure on travis-ci/push. Also, this is a windows-specific change, and the appveyor builds pass.

klshrinidhi:
:+1:

chrisdembia:
Thank you, kind reviewers..

Issue #1498, opened 2017-02-02
Remove dependency on OPENSIM_HOME


aymanhab:
ModelVisualizer and Geometry use OPENSIM_HOME and a hard coded set of directories to search for mesh files, and it may have been used to locate other resources. Windows machines with older installations will have this set incorrectly and the environment variables mechanism is not cross platform. Would be good to abstract the search paths into a class that maintain a list of directories with the ability to add to the list os that the GUI can decide where to pull these resources and same for other programs/visualizer.

chrisdembia:
+1; I think we have to add this to the 4.0 list.

jenhicks:
Maintain a list of directories to search for geometry files and GUIs and Visualizers will be customizers of the list.

jenhicks:
Use Carmichael's Path class
@aymanhab can point to places where hard coded path is being used

aymanhab:
Geometry.cpp has a call to ModelVisualizer::findGeometryFile that populates passed in search paths using hard coded list of options described in doxygen, there's no API to modify these search paths so that a client (e.g. the GUI) can add directories to search. This is a problem because the actual search paths have to be found on client machine after installation or could be user specified. Either abstract searchPath into a class and allow it to be passed in, or give option to provide additional search paths.

aymanhab:
Issue related to Geometry location was fixed in #1560, separate issue is open for locating visualizer executable #1572

PR #1560, opened 2017-03-08
Add ability to add search paths in ModelVisualizer

klshrinidhi:
Addressing #1498.

As per conversation with @aymanhab, this PR does not remove `OPENSIM_HOME`. This PR adds logic to search user added paths before falling back to previous behavior.

### DIFF
--- a/OpenSim/Simulation/Model/ModelVisualizer.cpp
+++ b/OpenSim/Simulation/Model/ModelVisualizer.cpp
@@ -206,7 +206,7 @@ findGeometryFile(const Model& aModel,
 // Initialize the static variable.
 SimTK::Array_<std::string> ModelVisualizer::dirsToSearch{};
 
-void ModelVisualizer::addDirToSearchGeometryFile(const std::string& dir) {
+void ModelVisualizer::addDirToGeometrySearchPaths(const std::string& dir) {
     // Make sure to add trailing path-separator if one is not present.
     if(dir.back() == Pathname::getPathSeparator().back())
         dirsToSearch.push_back(dir);
--- a/OpenSim/Simulation/Model/ModelVisualizer.cpp
+++ b/OpenSim/Simulation/Model/ModelVisualizer.cpp
@@ -139,6 +139,8 @@ void ModelVisualizer::show(const SimTK::State& state) const {
 
 // See if we can find the given file. The rules are
 //  - if it is an absolute pathname, we only get one shot, else:
+//  - search the user added paths in dirToSearch in reverse chronological order
+//    i.e. latest path added is searched first.
 //  - define "modelDir" to be the absolute pathname of the 
 //      directory from which we read in the .osim model, if we did,
 //      otherwise modelDir="." (current directory).
--- a/OpenSim/Simulation/Model/ModelVisualizer.cpp
+++ b/OpenSim/Simulation/Model/ModelVisualizer.cpp
@@ -202,9 +202,9 @@ findGeometryFile(const Model& aModel,
 }
 
 // Initialize the static variable.
-std::vector<std::string> ModelVisualizer::dirsToSearch{};
+SimTK::Array_<std::string> ModelVisualizer::dirsToSearch{};
 
-void ModelVisualizer::addDirToSearch(const std::string& dir) {
+void ModelVisualizer::addDirToSearchGeometryFile(const std::string& dir) {
     // Make sure to add trailing path-separator if one is not present.
     if(dir.back() == Pathname::getPathSeparator().back())
         dirsToSearch.push_back(dir);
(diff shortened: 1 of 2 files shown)


aymanhab:
Thanks for the quick turnaround @klshrinidhi Can we change the name of the method to make it clear what the directory is searched for (e.g. Mesh File(s)). I also see that you used std::vector<> instead of SimTK::Array_ which is used elsewhere in the file, is there an advantage in this use case? Thanks,

klshrinidhi:
* Changed the name of the function to `addDirToSearchGeometryFile`.
* Changed `std::vector` to `SimTK::Array_`.

OpenSim/Simulation/Model/ModelVisualizer.cpp
```diff
@@ -194,6 +201,17 @@ findGeometryFile(const Model& aModel,
...
+// Initialize the static variable.
+SimTK::Array_<std::string> ModelVisualizer::dirsToSearch{};
+
+void ModelVisualizer::addDirToSearchGeometryFile(const std::string& dir) {
+    // Make sure to add trailing path-separator if one is not present.
+    if(dir.back() == Pathname::getPathSeparator().back())
+        dirsToSearch.push_back(dir);
+    else
+        dirsToSearch.push_back(dir + "/");
+}
```
chrisdembia: Would OpenSim::Path handle that?

OpenSim/Simulation/Model/ModelVisualizer.cpp
```diff
@@ -160,7 +160,14 @@ findGeometryFile(const Model& aModel,
...
         foundIt = Pathname::fileExists(attempts.back());
-    } else {  
+    } else {
+        for(auto dir = dirsToSearch.crbegin();
+            dir != dirsToSearch.crend();
+            ++dir) {
+            attempts.push_back(*dir + geoFile);
+            if(Pathname::fileExists(attempts.back()))
+                return true;
+        }
```
chrisdembia: The comment above (lines 140-147) should be updated.

OpenSim/Simulation/Model/ModelVisualizer.h
```diff
@@ -194,6 +197,11 @@ class OSIMSIMULATION_API ModelVisualizer {
                             const std::string&          geoFile,
                             bool&                       isAbsolute,
                             SimTK::Array_<std::string>& attempts);
```
chrisdembia: It might be nice to add to the doxygen description that the last entry of `attempts` is the correct path of the file.

aymanhab: Agreed 👍

OpenSim/Simulation/Model/ModelVisualizer.cpp
klshrinidhi: Changed it to use `Pathname::getPathSeparator()`.

klshrinidhi:
For now, I think it is better to not use `OpenSim::Path` for now. It can be another PR later to replace all the path related operations with `OpenSim::Path` internally. Right now, the goal is to add the ability to append search paths.

chrisdembia:
> For now, I think it is better to not use OpenSim::Path for now. It can be another PR later to replace all the path related operations with OpenSim::Path internally. Right now, the goal is to add the ability to append search paths.

Fine with me; it makes sense to keep the code consistent, and use `getPathSeparator()`.

aymanhab:
I agree as well since it would make sense then to use Arrays of Paths (that you can test for permissions, whether files or folders, and other capabilities that are part of a Path but not String). Let's address documentation concerns then we can merge when tests pass.

OpenSim/Simulation/Model/ModelVisualizer.h
klshrinidhi: Done. 
Added comment to `\param`.

OpenSim/Simulation/Model/ModelVisualizer.cpp
klshrinidhi: Done.

chrisdembia:
LGTM, though we still have to wait for tests to pass. My last comment is that in the future, we may learn that we want the ability to remove paths from the search as well. Otherwise, users need to restart the GUI (not a big deal).

aymanhab:
True @chrisdembia and in earlier discussion we were considering a reset that clears the list. Practically though when users modify these search paths they want to put their preferred paths at the front more than remove anything, so unless it gets out of hand with excessively long list of directories we should be fine.

aymanhab:
Per discussion with Chris & Ajay we want to follow the search order adopted by OpenSim 3.3 GUI, which is: 1. Absolute path if provided
2. Same directory of the model
3. Geometry folder under the model
4. added search paths (assume latest dir to be added is at the top of this list )
5. OPENSIM_HOME/Geometry if OPENSIM_HOME is set
We also need to make sure documentation is updated accordingly to reflect this policy.

Thanks

klshrinidhi:
Okay. I have updated the search order to match the order in your comment.

aymanhab:
LGTM, thanks @klshrinidhi and @chrisdembia

Issue #1558, opened 2017-03-06
Create a method that returns simple, informative data type


tkuchida:
**Example:** If given `"SimTK::Vec<2,SimTK::Vec<3,double,1>,1>"`, returns `"SpatialVec"`.

See [issue 73 in GUI repo](https://github.com/opensim-org/opensim-gui/issues/73) and [Component::printOutputInfo()](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Component.cpp#L1552).

jenhicks:
@chrisdembia has a proposal for a few line solution. Will see if this works, otherwise postpone.

tkuchida:
@klshrinidhi: see [Object_GetClassName](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Object.h#L1081) and getClassName().

chrisdembia:
@tkuchida would you say this could be closed?

tkuchida:
> @tkuchida would you say this could be closed?

Yes (fixed by #1562).

PR #1562, opened 2017-03-10
Replace hardcoded map with OpenSim::Object_GetClassName<T>

klshrinidhi:
Addressing #1558.

This PR adds more specializations to `Object_GetClassname<T>` so it can be used more generally. Also removes use of hardcoded map in `Output<T>::getTypeName()` as it is not necessary anymore.

### DIFF
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -1552,16 +1552,6 @@ void Component::printSubcomponentInfo() const {
 void Component::printOutputInfo(const bool includeDescendants) const {
     using ValueType = std::pair<std::string, SimTK::ClonePtr<AbstractOutput>>;
 
-    static const std::unordered_map<std::string, std::string>
-        typeAliases{{"SimTK::Vec<2,double,1>", "Vec2"},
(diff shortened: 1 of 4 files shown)


OpenSim/Common/Component.cpp
```diff
@@ -1552,51 +1552,22 @@ void Component::printSubcomponentInfo() const {
...
-            if(typeAliases.find(name) != typeAliases.end()) {
-                len += typeAliases.at(name).length();
-                printingAlias = true;
-            }
-
-            maxlen = std::max(maxlen, len);
-        }
+        size_t maxlen{};
+        for(const auto& output : outputs)
+            maxlen = std::max(maxlen, output.second->getTypeName().length());
```
tkuchida: What happened to the alias-printing stuff that was deleted here? The description of the PR doesn't indicate that you've changed any functionality, so was the `if(typeAliases.find(name) != typeAliases.end())` always `false` for some reason?

chrisdembia:
Please merge when tests pass.

klshrinidhi:
Merging. Appveyor failure is due to `curl RPC`. Travis failure is due to timeout on a test. Not related to this PR.

Issue #1537, opened 2017-02-28
Update Matlab Table reader test script


jimmyDunne:
addresses issue created in #1536 

Need to update paths to files in test script.

PR #1566, opened 2017-03-10
Updated Test for Matlab Table conversion script

jimmyDunne:


Bindings/Java/Matlab/tests/testing_osimTableStruct.m
```diff
@@ -1,105 +1,76 @@
...
+labels.add('marker1');labels.add('marker2');
+labels.add('marker3');labels.add('marker4');
+table.setColumnLabels(labels);
+
+for i = 1 : 10
+    elem = Vec3(randi(10,1),randi(10,1),randi(10,1));
+    elems = StdVectorVec3();
+    elems.add(elem); elems.add(elem); elems.add(elem); elems.add(elem);
+    row = RowVectorOfVec3(elems);
+    table.appendRow(0.1,row);
```
chrisdembia: I would have thought using the same time for multiple rows gives an exception.

chrisdembia: I see that you change the times next.

chrisdembia:
It is great to no longer depend on the file. 

I have pushed a commit that adds the test to CMake, so that it is tested automatically when running ctest.

I also turned the test into a script (no longer a function); I hope that's okay. This was necessary to get the ctest to work.

LGTM.


Fixes #1537.

Issue #611, opened 2015-09-10
User option to reverse a Joint should be eliminated


aseth1:
Model is now capable of deciding what topology is best by using the `SimTK::MultibodyGraphMaker` which also takes into consideration closed Loops and other Constraints. Only the Model needs the ability to set this flag. The `reverse` property of  the Joint and its accessors should now be removed.

jenhicks:
Need to remove the property for the joint reverse.

jenhicks:
Add a test and throw an assertion when reading an old model that has a reverse tag.

jenhicks:
OR give a warning

jenhicks:
See discussion in #1424

tkuchida:
Related to #1406

jenhicks:
@tkuchida Will do some initial exploration to see how long this task will take. @aseth1 is ok with making the change now and resolving backward compatibility issues later.

tkuchida:
Fixed by #1567 and #1577

Issue #384, opened 2015-03-06
OpenSim 4.0 should provide a comprehensive list of API changes


aseth1:
and provide the user with the updated method or official call sequence that now stands in its place.
These include:
1. Virtual Component interface methods (e.g. `finalizeFromProperties()`, `connect()`, `addToSystem()`, `initStateFromProperties()`, `setPropertiesFromState()`, `computeStateVariableDerivatives`,  were renamed to `extend<ComponentMethod>` for the purpose of derived classes extending those behaviors. `Super::extend<ComponentMethod>` is required at the beginning of any extension implementation.
2. `finalizeFromProperties()` (and companion extension method `extendFinalizeFromProperties()`) was added to the Component interface to support building, loading internal data structures once from properties (e.g. deserialization) and not being forced to (re)build/deserialize them during connect() and or addToSystem().
3. `Body& Model::getGroundBody()` was replaced by `const Ground& Model::getGround() const` and `Ground& Model::updGround()`.

...

chrisdembia:
+1

chrisdembia:
related but different: #239

jimmyDunne:
change log is here 
https://github.com/opensim-org/opensim-core/compare/changelog

jimmyDunne:
This issue will be used to draft the changes and once we have solid list before publishing to the change log

idhamari:
It would be more nice if the changed added to the API documentation, something like: "this function replaced getGround ()" .

This helps when search for an old function to get the new one.

PR #1567, opened 2017-03-11
Remove user-facing "reverse" property from Joint - part 1 of 2

tkuchida:
See #611.

### In this PR
- Removed `reverse` flag from Joint constructors
- Replaced Joint's `reverse` property with a protected member variable (`isReversed`)

### To be addressed in part 2
- Backwards compatibility via `updateFromXMLNode()`
- Remove `Joint::getIsReversed()`
- Update and add to testJoints.cpp


OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -529,21 +495,28 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
...
+    //==========================================================================
     // DATA
-    //=========================================================================
+    //==========================================================================
+protected:
+    // Specifies the direction of the Joint in the multibody tree: parent->child
+    // (isReversed is false; default) or child->parent (isReversed is true). The
+    // Joint's transform and coordinates maintain a parent->child sense even if
+    // the joint is reversed.
+    bool isReversed = false;
```
chrisdembia: What about putting this description with the public accessor `isReversed()`? Might be good to also include a note about the history (`Before OpenSim 4.0, the user could set this. In 4.0 and later, this is determined automatically and cannot be set by the user`). I think the reason you put the description here is that this is supposed to be somewhat hidden. Perhaps the accessor's description can say "Advanced."

tkuchida: > do you think this deserves mention in the changelog?

We haven't been consistent with updating CHANGELOG; we will need to work on it prior to releasing 4.0 (see #384). But yes, I will add to CHANGELOG.

> What about putting this description with the public accessor `isReversed()`? Might be good to also include a note about the history ("Before OpenSim 4.0, the user could set this. In 4.0 and later, this is determined automatically and cannot be set by the user"). I think the reason you put the description here is that this is supposed to be somewhat hidden. Perhaps the accessor's description can say "Advanced."

I am going to remove `getIsReversed()` in part 2 when I improve the tests (`getIsReversed()` is used only by testJoints). The fact that joints can be reversed internally should be *completely* hidden from the user. I will add your note ("Before OpenSim 4.0, …") to CHANGELOG.

chrisdembia: > I am going to remove getIsReversed() in part 2 when I improve the tests (getIsReversed() is used only by testJoints). The fact that joints can be reversed internally should be completely hidden from the user. I will add your note ("Before OpenSim 4.0, …") to CHANGELOG.

Okay sounds good.

> We haven't been consistent with updating CHANGELOG

I know...I would like to create a [pull request template](https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/) that reminds us to update the CHANGELOG, but I can't get over the fact that we would have to add another file or directory to the root of the project.

chrisdembia: Okay I am going to create a pull request template; I found a way to not add another entry to the root of the repo.

tkuchida: Addressed feedback:
- Added "Remove `Joint::getIsReversed()`" to "To be addressed in part 2" list
- Added note to CHANGELOG

aseth1: > I am going to remove getIsReversed() in part 2 when I improve the tests (getIsReversed() is used only by testJoints). The fact that joints can be reversed internally should be completely hidden from the user.

I completely agree!

aseth1:
Thanks @tkuchida for cleaning this up. Merging ...

tkuchida:
Thanks for revierging.

Issue #978, opened 2016-05-16
Update reference to file containing buildinfo 


aymanhab:
GUI uses the file buildingo.txt to show Build Info (under Help/About details) and Matlab configuration script does something similar to check for compatibility. Both need to be updated to the new file name in PR #975 as well as documentation on confluence and the change log.

chrisdembia:
I have already updated the Matlab configuration script in #975, though I did not test the change.

aymanhab:
The filename is not actually used by the GUI, the call to get the name of the compiler used has been fixed in PR #1473 however it's not clear if that's enough or if 32/64 bit distinction needs to be made some other way.

aymanhab:
I (Ayman) Will check which function used by GUI for Abot dialog and update needs, @chrisdembia fixed configure opensim to use latest functions

aymanhab:
@aseth1 and @aymanhab will review and test

aymanhab:
@chrisdembia The method called by the GUI  (to show in the About Dialog is opensimModelCommonJNI.GetCompilerVersion() and it returns blank as of API alpha).

chrisdembia:
Hmm...I am able to use the function in MATLAB:

```
>> import org.opensim.modeling.*
No completions found.
>> opensimCommon.GetCompilerVersion()
ans =
Mac OS X :/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++
>>
```

(note, `opensimModelCommonJNI` should now be `opensimCommon`, or maybe `opensimCommonJNI`)

 It could be a Windows-specific issue, since I can't think of what we've done since the alpha to fix this. @carmichaelong @nickbianco since you have the 4.0 MATLAB bindings installed for Windows, could you check?

carmichaelong:
Tested this, with some caveats: 1) I tested on tgcs2017 branch. 2) Unsure what exactly I'm testing for.

I tried the following lines, none of which worked, but had different error messages:
```
>> import org.opensim.modeling.*
>> opensimCommon.GetCompilerVersion()
Undefined variable "opensimCommon" or class "opensimCommon.GetCompilerVersion".

>> opensimModelCommonJNI.getCompilerVersion()
No appropriate method, property, or field getCompilerVersion for class
org.opensim.modeling.opensimModelCommonJNI.

>> opensimCommonJNI.getCompilerVersion()
[...]

Issue #701, opened 2015-11-12
Can loadLibrary("osimJavaJNI") use more information to avoid the need for setting PATH / DYLD_LIBRARY_PATH?


chrisdembia:
See [here](https://github.com/opensim-org/opensim-core/blob/master/Bindings/Java/swig/javaWrapOpenSim.i#L267).

aymanhab:
Not sure what you mean by that @chrisdembia Embedding a hard coded absolute path in compiled code seems less flexible. Using environment variables is usually not recommended (vs. System properties) but I'm open to suggestions now that we can run on non-windows machines..

chrisdembia:
Perhaps a hard-coded relative path that is platform-dependent would work. It could be part of a search; PATH would override the setting. We could use similar logic as that used to find simbody-visualizer in a platform-specific way.

aymanhab:
@chrisdembia is the goal to avoid setting PATH, DYLD_LIBRARY_PATH or both and on what platform(s)? The reason I'm asking is that this's probably not the only scenario where we need to set these PATHs, so some big picture context helps. In general the fewer things we need to set the better, but I'd prefer a PATH that we can view/check/troubleshoot over one hidden in the code unless bullet proof.

chrisdembia:
On Ubuntu and OSX, we will almost certainly be able to avoid having users set *any* environment variables (perhaps except for PATH to access command line executables). I'm not sure about Windows. We can achieve this on OSX by using RPATH, and on Ubuntu by installing into system-wide directories.

On Windows, it's somewhat common for installers to modify the system/user PATH globally, but this is not common on OSX, and is only common on Ubuntu for non-GUI applications.

The ability to troubleshoot is a good point that I had not thought of...I think hardcoding a relative path might reduce confusion if done properly, since right now, if someone tries to open the GUI but has set the PATH to refer to different libraries (say, open 3.3 GUI with PATH set to 3.2 libraries), you might get confusing error messages.

I don't have an exactly plan yet; perhaps it will require a few iterations to get it right.

AFAIK, these are the reasons for setting environment variables (PATH, (DY)LD_LIBRARY_PATH):

* **Accessing the command line executables from a terminal or cmd window.**
  * _Ubuntu_: executables are on the PATH since they will be in /usr/bin or /usr/local/bin. The libraries that these execuables need are also already locatable since they will be in /usr/lib or /usr/local/lib.
[...]

PR #1568, opened 2017-03-11
Enhance configureOpenSim.m, update CHANGELOG.

chrisdembia:
This PR addresses part of #978, and finishes some enhancements to configureOpenSim.m mentioned in #1044 (also related: #701). Some of these changes were motivated by @aseth1 who reported that the script was not working for him, perhaps because he had previous versions of OpenSim configured with MATLAB.

A primary concern of this PR is that this script will work for users who previously had OpenSim 3.3 configured with MATLAB.

- Mention renaming of buildinfo.txt in changelog (#978).
- Running configureOpenSim now puts @jimmyDunne's utilities on the MATLAB path.
- Remove the need for administrator privileges for users with MATLAB 2012b or later who do not have previous installations of OpenSim.
- Gives verbose messages to help users troubleshoot issues (e.g., if they need administrator privileges).
- No longer adds OpenSim's bin dir to the MATLAB path; this was unnecessary.
- Removed the dependency on the OPENSIM_HOME environment variable, and no longer asks the user for the OpenSim install dir.
- Tell Windows users to update their Windows PATH.
- Handle the situation where users are running MATLAB in a headless mode (e.g., ssh to a cluster), where they would not be able to see msgbox prompts.
- Instead of erasing previous OpenSim installations from classpath.txt, etc., only comment out old entries.

I did extensive testing on macOS and Windows:
- I ensured that previous versions of OpenSim configured with MATLAB are properly commented out.
- I ensured that this script gives proper messages if administrator privileges are needed.
- I ensured the check for 64-bit vs 32-bit still works.
- I ensured that the behavior for pre-2012b MATLAB versions works.

I am confident that the script works as intended.

I did not test on Linux, but I believe the number of people using MATLAB with OpenSim on Linux is fairly small. Since the script works for macOS, it is likely to work on Linux as well.


chrisdembia:
It would be great if we could merge this. I did extensive testing and we can revisit if users run into issues.

chrisdembia:
@aseth1 I think you mentioned you could test this locally.

chrisdembia:
@apoorvar found an issue with the script; the script uses `strip()` which was introduced in MATLAB 2016b. I will find an alternative function to use.

apoorvar:
In the 'searchForOpenSimString' function, I changed the following (in bold):
    [Cnrows, **~**] = cellfun(@size, C**, 'UniformOutput', false**);
    for i = 1:Cnrows**{1}**
        if isempty(strfind(lower(C{1}{i}), 'opensim')) || ...
                **any(regexpi(C{1}{i}, '^\s*#') == 1)**

My file is attached. It seems to work ok on my version of Matlab (R2015b), but I'm not sure what/if anything was introduced in 2015 and/or was deprecated in later versions.

[configureOpenSim.txt](https://github.com/opensim-org/opensim-core/files/866702/configureOpenSim.txt)

chrisdembia:
Thank you @apoorvar; I have incorporated your changes. 

@tkuchida , I would like the configureOpenSim function to work with 2007b. If I prepare a Windows binary for you, could you test the function? If so, do you have 32-bit or 64-bit MATLAB?

tkuchida:
> I would like the configureOpenSim function to work with 2007b. If I prepare a Windows binary for you, could you test the function?

Sure, but I will need to install 2007b first.

chrisdembia:
Don't install a different MATLAB! Did you say you have 2007a, then?

tkuchida:
Yes, I have _a_ but I can download :bee: if we need it for testing.

chrisdembia:
@tkuchida thank you for troubleshooting this with me. I have decided to go with supporting 2008a and later for now. I was able to obtain MATLAB 2008a for Windows and I was able to successfully use OpenSim bindings, configured using this PR.

We can update the script as necessary.

I think this is ready to be merged; the CI builds do not test the configureOpenSim script and can be ignored.

Suggested reviewers: @tkuchida @aseth1 .

chrisdembia:
Thank you @tkuchida

chrisdembia:
@aseth1 could you review this? The script worked for you, and it also worked with 2008a on Windows. We can make changes in the future if we run into problems.

carmichaelong:
This worked for me with 2014b on a Windows 64-bit build. I agree that we can make more changes in the future and that it's important to get this in for now, so I will go ahead and merge.

chrisdembia:
Wooho!! Thank you @carmichaelong :)

Issue #1554, opened 2017-03-06
MacOS RPATH for osimJavaJNI (and maybe other osim libraries) is absolute, but should be relative


chrisdembia:
The RPATH for the osimJavaJNI is an absolute path, but should be a relative path in order make the (GUI) installation relocatable.

Relative RPATHs for the python dylibs is maybe not a good idea, since these shared libraries are usually copied into the python packages directory, therefore rendering a relative path incorrect. Perhaps the step of installing the python package on macOS should include editing the RPATH using `install_name_tool`.

chrisdembia:
While #1573 largely addresses this issue, it does not handle the case where WITH_BTK is ON. So even if #1573 is merged, we should not yet close this issue.

jenhicks:
Maybe fixed, @aymanhab will test

chrisdembia:
Given that the fact that the GUI worked on Macs at ISB, I think we can close this issue.

PR #1573, opened 2017-03-12
On macOS, use relative RPATHs to make the installation relocatable.

chrisdembia:
This PR changes the *run path list* that we put into OpenSim's executables and libraries on macOS, so that they look for their library dependencies using relative paths instead of absolute paths. For example, `libosimJavaJNI.dylib` now looks in `@loader_path/` for the libraries it depends on (e.g., `libosimTools.dylib`). Before, `libosimJavaJNI.dylib` would look for its library dependencies using an absolute directory, and this made the installation not relocatable (and not distributable to other computers).

I tested this locally and believe that the macOS GUI should now be able to start without any path-finding issues, so long as OpenSim was not built with BTK.

There are two issues this PR does not address:
- Copying BTK libraries into the installation on non-Windows systems (necessary on non-Windows for distributing). I avoided this for now because @klshrinidhi is working on copying the BTK installation into OpenSim's installation.
- Making the python package redistributable. The RPATH used for python is still absolute. A relative RPATH is not sufficient to make the python package redistributable, because users will install their python package into a path like `/usr/local/lib/python2.7/site-packages/opensim`, and so the relative PATH will be wrong. One solution is to copy all library dependencies (OpenSim, Simbody, BTK) into the python package. I will have a subsequent PR for this.

This PR partially addresses https://github.com/opensim-org/opensim-gui/issues/108.

### DIFF
--- a/cmake/OpenSimMacros.cmake
+++ b/cmake/OpenSimMacros.cmake
@@ -87,6 +87,11 @@ function(OpenSimAddLibrary)
     # at the top level in OpenSim/bin/*.dll (Windows) or OpenSim/lib/*.so
     # (Linux) or OpemSim/lib/*.dylib (Mac). Windows .lib files, and Linux/Mac
     # .a static archives are only needed at link time so go in sdk/lib.
+    if(WIN32)
+        set(OSIMADDLIB_LIBRARY_DESTINATION sdk/lib)
+    else()
+        set(OSIMADDLIB_LIBRARY_DESTINATION lib)
+    endif()
     install(TARGETS ${OSIMADDLIB_LIBRARY_NAME}
         EXPORT OpenSimTargets
         RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
--- a/cmake/OpenSimMacros.cmake
+++ b/cmake/OpenSimMacros.cmake
@@ -87,11 +87,6 @@ function(OpenSimAddLibrary)
     # at the top level in OpenSim/bin/*.dll (Windows) or OpenSim/lib/*.so
     # (Linux) or OpemSim/lib/*.dylib (Mac). Windows .lib files, and Linux/Mac
     # .a static archives are only needed at link time so go in sdk/lib.
-    if(WIN32)
-        set(OSIMADDLIB_LIBRARY_DESTINATION sdk/lib)
-    else()
-        set(OSIMADDLIB_LIBRARY_DESTINATION lib)
-    endif()
     install(TARGETS ${OSIMADDLIB_LIBRARY_NAME}
         EXPORT OpenSimTargets
         RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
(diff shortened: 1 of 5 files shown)


chrisdembia:
This addresses #1554.

cmake/OpenSimMacros.cmake
```diff
@@ -87,11 +87,6 @@ function(OpenSimAddLibrary)
     # at the top level in OpenSim/bin/*.dll (Windows) or OpenSim/lib/*.so
     # (Linux) or OpemSim/lib/*.dylib (Mac). Windows .lib files, and Linux/Mac
     # .a static archives are only needed at link time so go in sdk/lib.
-    if(WIN32)
-        set(OSIMADDLIB_LIBRARY_DESTINATION sdk/lib)
-    else()
-        set(OSIMADDLIB_LIBRARY_DESTINATION lib)
-    endif()
     install(TARGETS ${OSIMADDLIB_LIBRARY_NAME}
```
aymanhab: Why was that block removed, @chrisdembia ? I assumed the whole CMake changes will be all guarded by APPLE, or is that to account for the new layout?

chrisdembia: This was some cleanup; these variables are no longer used, now that the layout of the installation is fully specified in the root CMakeLists.txt. This change is unrelated to the PR. If you'd like, I can back this off.

chrisdembia:
> I'm ok with these changes as long as they have been tested locally (since there's no other way to test if they work).

Yes; I tested locally that I could rename the installation directory and that both `opensim-cmd` and a Java program would still work (as long as WITH_BTK is false).

> I assume this change doesn't require cmake version change/upgrade.

Correct; these features have been in CMake for a few versions now, and definitely since 3.2.3 (our minimum required).

aymanhab:
Thanks @chrisdembia let's back the cleanup off and later fold it into the layout changes we agreed upon. At that point we'll probably need significant cleanup anyway. Thanks.

chrisdembia:
> Thanks @chrisdembia let's back the cleanup off and later fold it into the layout changes we agreed upon. At that point we'll probably need significant cleanup anyway. Thanks.

Done.

aymanhab:
Great, thanks @chrisdembia Would be good to get @klshrinidhi 's input and where things stand with BTK since I expect that we'll build it by default, otherwise LGTM 👍

aymanhab:
@klshrinidhi Would be great to know where things stand for btk since the solution is not complete without it. Thanks

klshrinidhi:
I talked to @chrisdembia. This can be merged unless there are pending issues to be resolved. I will get back with a PR on next steps with installation layout.

chrisdembia:
Yes; BTK will be easier to handle with @klshrinidhi's upcoming changes.

aymanhab:
Ok, thanks @klshrinidhi and @chrisdembia Merging

chrisdembia:
Thank you @aymanhab

PR #1574, opened 2017-03-13
Allow making a standalone Python package

chrisdembia:
This PR should not be merged until after #1573 is merged, as the branch `python_standalone` is built on `macOS_RPATH_relative`.

As it is, our python package is not standalone: it depends on shared libraries in OpenSim's bin or lib directory. As we prepare for distributing the GUI, it will help users for the python package to be standalone: to contain all the libraries it needs. That is, someone could install the python package into their python installation and then delete OpenSim, and the python package would still work. 

This issue is biggest on macOS (on Windows, the python package can find the DLL dependencies via the OpenSim bin dir that we add to PATH; on Linux, a proper installation would put the shared libraries in /usr/lib or /usr/local/lib).

I created a CMake option `OPENSIM_PYTHON_STANDALONE`) to control whether or not we copy OpenSim, Simbody, and BTK libraries into the python package. By default, we do not (that is, we maintain the current behavior). It is expected that packagers (e.g., @aymanhab) will set the option as necessary. Different ways of distributing the python package may require different settings for this option; it's not exactly clear to me yet how this flag should be used for Homebrew, etc.

This PR also handles changes to the RPATH on macOS so that the OpenSim/Simbody/BTK libraries are found whether or not they are copied into the installation.

Unlike #1573, this PR handles BTK. That is, BTK libraries are copied into the python package if `WITH_BTK=ON`. However, the python libraries (e.g., `_simulation.so`) end up with an RPATH to the absolute path of the BTK libraries on the build machine. This should be modified in the future so that the python package we distribute with the GUI doesn't have an RPATH for an absolute path on @aymanhab's computer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -222,6 +222,10 @@ script:
   - make -j8 install > /dev/null
 
   ## Test python wrapping.
+  # Since we use OPENSIM_COPY_SIMBODY=ON, ensure the Python bindings are not
+  # relying on the original Simbody libraries but rather the copied ones.
+  # This should make sure we have the correct RPATH for Simbody on macOS.
+  - rm -r ~/simbody
   # Go to the python wrapping package directory.
   - if [ "$WRAP" = "on" ]; then cd $OPENSIM_HOME/lib/python2.7/site-packages; fi
   # Run the python tests, verbosely.
--- a/Bindings/Python/setup.py
+++ b/Bindings/Python/setup.py
@@ -15,7 +15,9 @@ setup(name='opensim',
       url='http://opensim.stanford.edu/',
       license='Apache 2.0',
       packages=['opensim'],
-      package_data={'opensim': ['_*.*']},
+      # The last 3 entries are for if OPENSIM_PYTHON_STANDALONE is ON.
+      # The asterisk after the extension is to handle version numbers on Linux.
+      package_data={'opensim': ['_*.*', '*.dylib', '*.dll', '*.so*']},
       include_package_data=True,
       classifiers=[
           'Intended Audience :: Science/Research',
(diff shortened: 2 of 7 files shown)


.travis.yml
```diff
@@ -222,6 +222,10 @@ script:
   - make -j8 install > /dev/null
 
   ## Test python wrapping.
+  # Since we use OPENSIM_COPY_SIMBODY=ON, ensure the Python bindings are not
+  # relying on the original Simbody libraries but rather the copied ones.
+  # This should make sure we have the correct RPATH for Simbody on macOS.
+  - rm -r ~/simbody
   # Go to the python wrapping package directory.
   - if [ "$WRAP" = "on" ]; then cd $OPENSIM_HOME/lib/python2.7/site-packages; fi
```
aymanhab: Why are we still relying on OPENSIM_HOME in .travis.yml?

chrisdembia: It might have originally been used to find files for the python tests. Now, it is mostly for convenience; the tests should not depend on it.

I have opened a separate issue for that.

Bindings/Python/setup.py
```diff
@@ -15,7 +15,9 @@
       url='http://opensim.stanford.edu/',
       license='Apache 2.0',
       packages=['opensim'],
-      package_data={'opensim': ['_*.*']},
+      # The last 3 entries are for if OPENSIM_PYTHON_STANDALONE is ON.
+      # The asterisk after the extension is to handle version numbers on Linux.
```
aymanhab: astrisk? typo

chrisdembia: *Asterisk* is the correct spelling, though I usually pronounce the word more like "astrick".

chrisdembia:
Thank you for reviewing.

> It would be extremely useful if there's a place at the top of the main CMakeLists.txt file that describes the different installation layouts and what flags to set (by the "packager") rather than sifting through deeply nested CMake files to find what flag/values to use. Not necessary for this PR though

Good idea. It's definitely very confusing right now. I created #979 to track this.

klshrinidhi:
:+1: Go ahead if there are no other issues.

chrisdembia:
Thank you both. I am merging this, as it is mostly changes to CMake; therefore, the original author may merge the PR.

Issue #1575, opened 2017-03-13
Incrementing XMLDocument::LatestVersion causes testStaticOptimization failure


tkuchida:
### To reproduce
- In `XMLDocument.cpp`, change `LatestVersion` from 30513 to 30514
- Run testStaticOptimization

### Expected result
Test passes

### Actual result
<pre>
Done, with failure(s): (testArm26_Millard2012EquilibriumMuscle,testModelWithPassiveForces)
</pre>

### Additional information
#1548 was the last PR to increment `LatestVersion` (from 30512 to 30513). The `Muscle::updateFromXMLNode()` method begins with
```cpp
if ( versionNumber < XMLDocument::getLatestVersion()) {
```
so a "version 30513" file will not trigger the code in `updateFromXMLNode()` if `LatestVersion` is 30513—however, there is an `if (versionNumber <= 30513)` block in this method, so a "version 30513" file will be handled differently once `LatestVersion` has been incremented. [A search of the repo](https://github.com/opensim-org/opensim-core/search?utf8=%E2%9C%93&q=30513) reveals that there are two "version 30513" model files ([Applications/Analyze/test/arm26.osim](https://github.com/opensim-org/opensim-core/blob/master/Applications/Analyze/test/arm26.osim) and [Applications/Analyze/test/subject01_adjusted_spring.osim](https://github.com/opensim-org/opensim-core/blob/master/Applications/Analyze/test/subject01_adjusted_spring.osim)), which were modified in #1548 and are used by testStaticOptimization.

PR #1576, opened 2017-03-13
min_control should not be adjusted for "version 30513" files

tkuchida:
See #1575. According to [the comments in Muscle.cpp at L102](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/Muscle.cpp#L102), the `min_control` property is assumed to be incorrect if its value is 0 and the model version number is 30512 or less. This assumption is not made for more recent models, so the last `if` block in `Muscle::updateFromXMLNode()` should not be run on "version 30513" files.

### DIFF
--- a/OpenSim/Simulation/Model/Muscle.cpp
+++ b/OpenSim/Simulation/Model/Muscle.cpp
@@ -93,7 +93,7 @@ void Muscle::updateFromXMLNode(SimTK::Xml::Element& aNode, int versionNumber)
             }
             XMLDocument::renameChildNode(aNode, "pennation_angle", "pennation_angle_at_optimal");
         }
-        if (versionNumber <= 30513) {
+        if (versionNumber < 30513) {
             SimTK::Xml::element_iterator minControlElt =
                 aNode.element_begin("min_control");
             double minControl = 0;


chrisdembia:
LGTM. This change correctly reflects [DEVELOPING.md](https://github.com/opensim-org/opensim-core/blob/master/DEVELOPING.md), though it is not clear to me why `if (versionNumber <= 20301)` in Muscle.cpp is correct.

tkuchida:
> it is not clear to me why `if (versionNumber <= 20301)` in Muscle.cpp is correct.

Please give me a hint. Are you requesting a review of the block (L60–95) or are you requesting an investigation into whether `<=` should be `<` on L60? Regarding the latter: it depends on the logic that was used when developing "version 20301" model files. @aseth1 committed "version 30513" files with his PR #1548 and the assumptions he added in `updateFromXMLNode` should not be applied to those files.

chrisdembia:
I'm not requesting anything. I think this PR should be merged; it is the correct change.

aseth1:
I agree with the change. Version 30513 and greater should be correct so it should have been just <305013. @tkuchida thanks for finding and fixing!

tkuchida:
> though it is not clear to me why `if (versionNumber <= 20301)` in Muscle.cpp is correct.

I think it would be easier to read if that line were `if (versionNumber < 20302)` instead (though it doesn't matter logically because `versionNumber` is an integer). According to XMLDocument.cpp, version 20302 was "for Muscle's pennation_angle -> pennation_angle_at_optimal" (which apparently is an incomplete list of the changes).

PR #1577, opened 2017-03-13
Remove user-facing "reverse" property from Joint - part 2 of 2

tkuchida:
### Summary
Completes fix for #611 (see #1567 for part 1).

### Contents
- Added `Joint::updateFromXMLNode()` block that swaps parent and child frames in old model files where a Joint's "reverse" property has been set to "true".
- Added corresponding test to testJoints.
- Removed `Joint::getIsReversed()` method. Whether a Joint is reversed internally in the multibody tree (and even the fact that Joints _can_ be reversed) should be hidden from the user. This method was used only in testJoints and was not essential to the test.

### Note
testStaticOptimization will fail until #1576 has been merged.

### DIFF
--- a/OpenSim/Simulation/SimbodyEngine/Joint.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/Joint.cpp
@@ -660,30 +660,75 @@ void Joint::updateFromXMLNode(SimTK::Xml::Element& aNode, int versionNumber)
         // found and its value is "true".
         if (documentVersion < 30514) {
             auto reverseElt = aNode.element_begin("reverse");
-            auto parentElt  = aNode.element_begin(
-                                  "socket_parent_frame_connectee_name");
-            auto childElt   = aNode.element_begin(
-                                  "socket_child_frame_connectee_name");
 
             if (reverseElt != aNode.element_end()) {
                 bool swapFrames = false;
                 reverseElt->getValue().tryConvertToBool(swapFrames);
 
-                if (swapFrames
-                    && parentElt != aNode.element_end()
-                    && childElt  != aNode.element_end())
-                {
+                if (swapFrames) {
                     std::string oldParentFrameName = "";
                     std::string oldChildFrameName  = "";
 
-                    parentElt->getValueAs<std::string>(oldParentFrameName);
-                    childElt->getValueAs<std::string>(oldChildFrameName);
+                    // Find names of parent and child frames. If more than one
+                    // "parent_frame" or "child_frame" element exists, keep the
+                    // first one. The "parent_frame" and "child_frame" elements
+                    // may be listed in either order.
+                    SimTK::Xml::element_iterator connectorsNode =
+                        aNode.element_begin("connectors");
+                    SimTK::Xml::element_iterator connectorElt = connectorsNode->
+                        element_begin("Connector_PhysicalFrame_");
+                    SimTK::Xml::element_iterator connecteeNameElt;
 
-                    parentElt->setValue(oldChildFrameName);
-                    childElt->setValue(oldParentFrameName);
+                    while (connectorElt != connectorsNode->element_end())
+                    {
+                        if (connectorElt->getRequiredAttributeValue("name") ==
+                            "parent_frame" && oldParentFrameName.empty())
+                        {
+                            connecteeNameElt = connectorElt->
+                                               element_begin("connectee_name");
+                            connecteeNameElt->getValueAs<std::string>(
+                                oldParentFrameName);
+                        }
+                        else if (connectorElt->getRequiredAttributeValue("name")
+                                 == "child_frame" && oldChildFrameName.empty())
+                        {
+                            connecteeNameElt = connectorElt->
+                                               element_begin("connectee_name");
+                            connecteeNameElt->getValueAs<std::string>(
+                                oldChildFrameName);
+                        }
+                        ++connectorElt;
+                    }
+
+                    // Swap parent and child frame names. If more than one
+                    // "parent_frame" or "child_frame" element exists, assign
+                    // the same value to all such elements.
+                    connectorsNode = aNode.element_begin("connectors");
+                    connectorElt = connectorsNode->element_begin(
+                                   "Connector_PhysicalFrame_");
+
+                    while (connectorElt != connectorsNode->element_end())
+                    {
+                        if (connectorElt->getRequiredAttributeValue("name") ==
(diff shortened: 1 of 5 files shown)


OpenSim/Simulation/SimbodyEngine/Test/testJoints.cpp
```diff
@@ -2226,6 +2231,42 @@ void testAutomaticJointReversal()
...
+
+    cout << "\n==========================================================="
+         << "\n Test joint reversal set by user in old model file"
+         << "\n==========================================================="
+         << endl;
+
+    // Open model.
+    auto model = Model("double_pendulum_testReverse.osim");
+    //model.finalizeFromProperties();
+    model.initSystem();
```
tkuchida: I thought `model.finalizeFromProperties()` should work here, but I get this error:
<pre>
Loaded model double_pendulum_testReverse from file double_pendulum_testReverse.osim
PinJoint::Socket 'parent_frame' is not connected to 'ground' of type PhysicalFrame
        In file C:\Users\tkuchida\GitHub\opensim-core\OpenSim/Common/ComponentSocket.h:329
        In function 'getConnectee'
</pre>

aseth1: I can look into this. `initSystem()` calls `finalizeFromProperties()` so it should have failed in both cases.

tkuchida: The throw occurs in [Socket::getConnectee()](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/ComponentSocket.h#L329). The comment for this method begins, "Temporary access to the connectee for testing purposes." so I don't know why this method is being called at all or whether it should still exist at this point.

OpenSim/Simulation/SimbodyEngine/Test/testJoints.cpp
```diff
@@ -2226,6 +2231,42 @@ void testAutomaticJointReversal()
...
+
+    // The topology is specified as follows:
+    //     [ground] <- (pin1) <- [rod1] <- (pin2) <- [rod2]
+    // but the "reverse" element of "pin2" is "true" so, after deserialization,
+    // the topology should be the following:
+    //     [ground] <- (pin1) <- [rod1] -> (pin2) -> [rod2]
```
aseth1: If I interpret this correctly the desired "user topology" has `rod1` as the child of two joints (`pin1` and `pin2`), which is not a valid tree, so the `MultibodyTreeMaker` will reverse the `pin2` with `rod2` as the true outboard but keep the sense of `pin2` coordinate as describing `rod1` w.r.t. `rod2`.  Is that what you intended?

tkuchida: >  Is that what you intended?

Yes, but this test is ignorant of `MultibodyTreeMaker`. What you described is already tested in `testAutomaticJointReversal()`; the job of `testUserJointReversal()` is merely to ensure that the parent/child relationships are as expected after deserialization. If you want, I can add another model file where both "reverse" flags are "false", run a short simulation, and make sure that the coordinates from pin2 have opposite signs in the two models—but I believe that test would be redundant with the tests in `testAutomaticJointReversal()`.

OpenSim/Simulation/SimbodyEngine/Test/testJoints.cpp
aseth1: Sorry that comment needs to be updated, but it doesn't have any bearing on the current issue as far as I can tell. 
Does `finalizeFromProperties()` itself cause the throw or only when you try to access the Socket afterward? `finalizeFromProperties()` does not invoke `finalizeConnections()`. If you need to resolve and access connection information you need to call `finalizeConnections()`, which `initSystem()` does as well.

tkuchida: > If you need to resolve and access connection information you need to call `finalizeConnections()`, which `initSystem()` does as well.

:scream_cat: Sorry, my revised mental model was "load a model and then call `finalizeFromProperties()` and you're safe". Yes, I probably just needed `finalizeConnections()` (checking now…).

tkuchida: `finalizeConnections()` works (`pin1.getParentFrame()` was throwing the exception).

OpenSim/Simulation/SimbodyEngine/Test/testJoints.cpp
aseth1: I was just making sure I understood the test. I was interpreting the *topology* according to the `MultibodySystem` which isn't strictly valid, but from the user perspective your comment is fine and the test to verify the correct sense of the Joint is sufficient. Thanks!

tkuchida: Ah ok, yes I agree that the comment was potentially confusing. Updated.

tkuchida:
testContext failed on AppVeyor 32-bit only (64-bit passed):
<pre>
Prop reset to false
Exception: Component::markCacheVariableValid: ERR- name not found.
 for component 'geometrypath_' of type GeometryPath
</pre>
Restarted.

OpenSim/Simulation/SimbodyEngine/Joint.cpp
```diff
@@ -655,6 +655,84 @@ void Joint::updateFromXMLNode(SimTK::Xml::Element& aNode, int versionNumber)
...
+                    std::string oldChildFrameName  = "";
+
+                    // Find names of parent and child frames. If more than one
+                    // "parent_frame" or "child_frame" element exists, keep the
+                    // first one. The "parent_frame" and "child_frame" elements
+                    // may be listed in either order.
+                    SimTK::Xml::element_iterator connectorsNode =
+                        aNode.element_begin("connectors");
+                    SimTK::Xml::element_iterator connectorElt = connectorsNode->
+                        element_begin("Connector_PhysicalFrame_");
```
tkuchida: Yeah, that was a source of confusion. I think the current design might be better than duplicating the code. Better still might be splitting up the code so that the model file is completely updated one version at a time. We might also then permit updates from an old version to a version that isn't current, which could help test backwards compatibility.

chrisdembia: I added your comment to #1588.

PR #1579, opened 2017-03-13
Update sine function

aseth1:
Added a DC `offset` to the `Sine` `Function` and updated its properties.

### DIFF
--- a/OpenSim/Common/Sine.h
+++ b/OpenSim/Common/Sine.h
@@ -107,7 +107,7 @@ public:
 private:
     void constructProperties() {
         constructProperty_amplitude(1.0);
-        constructProperty_omega(2 * SimTK::Pi);
+        constructProperty_omega(1.0);
         constructProperty_phase(0.0);
         constructProperty_offset(0.0);
     }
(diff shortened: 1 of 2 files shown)


OpenSim/Common/Sine.h
```diff
@@ -135,11 +104,18 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Sine, Function);
     int getArgumentSize() const override {return 1;}
     int getMaxDerivativeOrder() const override {return 10;}
 
+private:
+    void constructProperties() {
+        constructProperty_amplitude(1.0);
+        constructProperty_omega(2 * SimTK::Pi);
```
chrisdembia: I think we should leave the default value as it was. I actually think the name for this property is not correct; usually `f = 2*pi*omega`, so this property should be called "frequency" (as there is no `2*pi` in `calcValue()`). Changing `omega`'s default to `2*pi` is even more confusing, I think. It also makes sense that default properties just give `sin(x)`. Consider that this function is not necessarily only used for functions of time; sin(2*pi*(joint_angle_in_radians))` doesn't make much sense.

This change is also backwards-incompatible (I wouldn't say that the mechanism we have for preserving defaults in a backwards compatible way, via the `defaults` node in XML files, is fully functional).

tkuchida:
Test failure is unrelated (timeout of testCMCEMGDrivenArm_Millard on clang/Debug); merging.

PR #1581, opened 2017-03-13
Expose Rotation multiply() in SWIG (wrapping)

aseth1:
Exposes C++ `operator*` as Rotation.multiply() via scripting and enables native rotation of a `Vec3` or a `RowVectorVec3` (rows of a `TimeSeriesTableVec3`).

Addresses issue raised in review of #1434. In addition to making it more convenient to use the SimTK::Rotation in MATLAB, it should improve the performance of converting large tables of Marker and Force data.

@jimmyDunne I added a one line test to see if the method was correctly exposed. Can you verify that it works for #1434?

### DIFF
--- a/Bindings/simbody.i
+++ b/Bindings/simbody.i
@@ -186,6 +186,10 @@ namespace SimTK {
     RowVector_<Vec3> multiply(const RowVector_<Vec3>& row) {
         return operator*(*$self, row);
     }
+
+    RowVector_<Vec3> multiply(const RowVectorView_<Vec3>& row) {
+        return operator*(*$self, row);
+    }
 }
 
 
@@ -225,8 +229,8 @@ namespace SimTK {
 typedef int MobilizedBodyIndex;
 
 namespace SimTK {
-%template(ArrayIndexUnsigned) ArrayIndexTraits<unsigned>; 
-%template(ArrayIndexInt) ArrayIndexTraits<int>; 
+%template(ArrayIndexUnsigned) ArrayIndexTraits<unsigned>;
+%template(ArrayIndexInt) ArrayIndexTraits<int>;
 }
 
 %include <SWIGSimTK/PolygonalMesh.h>
--- a/Bindings/simbody.i
+++ b/Bindings/simbody.i
@@ -172,12 +172,23 @@ namespace SimTK {
 %template(MatrixOfSpatialVec) Matrix_<SpatialVec>;
 }
 
-
 %include <SWIGSimTK/Rotation.h>
 namespace SimTK {
 %template(Rotation) SimTK::Rotation_<double>;
 %template(InverseRotation) SimTK::InverseRotation_<double>;
 }
+
+%extend SimTK::Rotation_<double> {
+    Vec3 multiply(const Vec3& v) {
+        return operator*(*$self, v);
+    }
+
+    RowVector_<Vec3> multiply(const RowVector_<Vec3>& row) {
+        return operator*(*$self, row);
+    }
+}
+
+
 // Transform
 %include <SWIGSimTK/Transform.h>
 namespace SimTK {
--- a/Bindings/Java/Matlab/tests/testosimTableStruct.m
+++ b/Bindings/Java/Matlab/tests/testosimTableStruct.m
@@ -10,12 +10,16 @@ labels.add('marker1');labels.add('marker2');
 labels.add('marker3');labels.add('marker4');
 table.setColumnLabels(labels);
 
+R = Rotation(pi/3, Vec3(0.1, 0.2, 0.3));
+
 for i = 1 : 10
     elem = Vec3(randi(10,1),randi(10,1),randi(10,1));
     elems = StdVectorVec3();
     elems.add(elem); elems.add(elem); elems.add(elem); elems.add(elem);
     row = RowVectorOfVec3(elems);
-    table.appendRow(0.1,row);
+    
+    rotatedRow = R.multiply(row);    
+    table.appendRow(0.1,rotatedRow);
 end
 
 % Set the indpendentColumn (Time) values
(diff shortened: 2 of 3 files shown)


aymanhab:
Before we add a new method in the interface file, did you consider using the rename facility in swig especially if multiple operators for many classes will eventually be needed.

chrisdembia:
See here for examples: https://scion.duhs.duke.edu/vespa/gamma/wiki/SwigOperatorRenaming.

chrisdembia:
http://www.swig.org/Doc2.0/SWIG.html#SWIG_rename_ignore

jimmyDunne:
I am getting a Matlab exception ``` Undefined function or variable 'multiply'. ``` when running the test. completely deleted and rebuilt the repo but still get the same error.

chrisdembia:
Do you see a "multiply" function in <build-dir>/Bindings/Java/src/org/opensim/modeling/Rotation.java?

jimmyDunne:
In opensim-build/Bindings/Java/src/org/opensim/modeling/Rotation.java

```
public Vec3 multiply(Vec3 v) {
    return new Vec3(opensimSimbodyJNI.Rotation_multiply__SWIG_0(swigCPtr, this, Vec3.getCPtr(v), v), true);
  }

  public RowVectorOfVec3 multiply(RowVectorOfVec3 row) {
    return new RowVectorOfVec3(opensimSimbodyJNI.Rotation_multiply__SWIG_1(swigCPtr, this, RowVectorOfVec3.getCPtr(row), row), true);
  }
```

In opensim-install/share/OpneSim/java/org/opensim/modelling/Rotation.java

```
 public Vec3 multiply(Vec3 v) {
    return new Vec3(opensimSimbodyJNI.Rotation_multiply__SWIG_0(swigCPtr, this, Vec3.getCPtr(v), v), true);
  }

  public RowVectorOfVec3 multiply(RowVectorOfVec3 row) {
    return new RowVectorOfVec3(opensimSimbodyJNI.Rotation_multiply__SWIG_1(swigCPtr, this, RowVectorOfVec3.getCPtr(row), row), true);
  }
```

jimmyDunne:
Ignore me. I rebuilt again and re-booted Matlab. 
I hadn't restarted Matlab in awhile so must have been using the previous libraries.

jimmyDunne:
getting a row from a `TimesSeriesTableVec3` returns a `RowVectorViewVec3` — I added the rotation of this type to the PR. 

I can now successfully rotate a 'RowVectorViewVec3' and have tested 🍾 🎆

Bindings/simbody.i
```diff
@@ -172,12 +172,27 @@ namespace SimTK {
...
+%extend SimTK::Rotation_<double> {
+    Vec3 multiply(const Vec3& v) {
+        return operator*(*$self, v);
+    }
+
+    RowVector_<Vec3> multiply(const RowVector_<Vec3>& row) {
+        return operator*(*$self, row);
+    }
+
+    RowVector_<Vec3> multiply(const RowVectorView_<Vec3>& row) {
```
jimmyDunne: Needed to add this type since table.getRow() returns RowVectorVewVec3()

aseth1:
> Before we add a new method in the interface file, did you consider using the rename facility in swig especially if multiple operators for many classes will eventually be needed.

While agree we want to avoid name collisions, the fact that we extend the `Rotation` class and add member functions and not free functions, means that the context is very specific/limited to `Rotation.multiply()` in which overloading the method is handy for dealing with common types and not having to `multiplyRowVector()`, `multiplyRowVectorView()`, etc... It doesn't seem necessary to have the fully disambiguated names for member functions. Is that not right?

aymanhab:
@aseth1 the issue is not name collision as much as maintenance. If we just tell swig to rename every operator* to multiply then that's easily documented to users and could apply to multiple classes and be one place to change with no need to duplicate methods in the .i file.  I think this can be made to apply to all template instantiations so we don't have to add a method for RowVector, RowVectorView, RowVector3, ..... etc. it depends on how many of these are needed and whether methods already exist (in which case name collision may be an issue indeed) that replicate the functionality of operators.

jimmyDunne:
Ajay and I discussed how a general case would be great, but we are under time considerations that wouldn't allow full development and testing. This PR allows me to move on with the Matlab C3D utility functions (this PR allows me to rotate mocap data) and doesn't cause any other adverse behavior. I can make an issue to address the suggestions made in this thread. @aymanhab does that sound okay?

aymanhab:
Sure @jimmyDunne no problem, you can include the links that @chrisdembia posted earlier as a guidance for a future change if needed. Thanks

aseth1:
Thanks for clarification @aymanhab . If we could use rename to enable all use of operator* that would really be better. How do you feel about merging the PR as is since it enables functionality that is needed now, and explore the use of `rename` in the near future?

aymanhab:
Agreed @aseth1 please feel free to merge when tests pass.

aseth1:
This [failure](https://travis-ci.org/opensim-org/opensim-core/jobs/211040139) seems unrelated to the changes and related to posting doxygen pages to sourceforge.

chrisdembia:
I vaguely remember looking into exposing more operators in the bindings, and my recollection was that Simbody's templating of the operators in a lot of cases was too complex for SWIG to handle.

aseth1:
Build [6678.4](https://travis-ci.org/opensim-org/opensim-core/jobs/211040146) timing out is unrelated (does not build wrapping).

aseth1:
@aymanhab @jimmyDunne the related (#1586) issue has been created.

aseth1:
Thanks @jimmyDunne!

jimmyDunne:
Thankyou!! 
tenor

PR #1583, opened 2017-03-14
Update WrapObject XML appearance options.

### LINKED ISSUE #1553: Display preference for wrap objects is not remembered
This issue is related to https://github.com/opensim-org/opensim-gui/issues/87: a model will display its wrap surfaces even if a v3.3 model file says it's not supposed to display a wrap surface. This might be because the `display_geometry` property of wrap objects is not being used in `updateFromXML()`?

chrisdembia:
This PR fixes #1553.  It introduces a test to make sure that a WrapObject's `display_preference` and `color` are maintained through the XML version updates. And I added the necessary updateFromXMLNode code to get the test to pass. The model I added to the repo consists mainly of 3 WrapObjects.

~~I marked this PR as WIP because the default color should be cyan, not white; and I have yet to handle the SurfaceProperty "representation" (e.g., wireframe, etc.).~~

This PR now handles the default color properly, and handles "representation."


chrisdembia:
Ready for review.

OpenSim/Tools/Test/testVisualization.cpp
```diff
@@ -476,7 +476,7 @@ void populate_wrapModelPrimitives(SimTK::Array_<DecorativeGeometry>& stdPrimitiv
     // that DecorativeCylinder is Y aligned while WrapCylinder is Z aligned
     cylTransform.updR().setRotationFromAngleAboutX(SimTK_PI / 2);
     stdPrimitives.push_back(
-        DecorativeCylinder(.025, .05).setBodyId(0).setColor(SimTK::Cyan)
+        DecorativeCylinder(.025, .05).setBodyId(0).setColor(Vec3(0, 0.2, 0.8))
```
aymanhab: We used Cyan in the past so that testers know what to expect if they turn visualization on, is there a reason to revert to the RGB values?

OpenSim/Tests/Wrapping/testWrapping.cpp
```diff
@@ -66,6 +67,13 @@ int main()
         std::cout << "Exception: " << e.what() << std::endl;
         failures.push_back("TestShoulderModel (multiple wrap)"); }
 
+    try{
+        testWrapObjectUpdateFromXMLNode30515();
+    } catch (const std::exception& e) {
```
aymanhab: I assume you looked at that visually to make sure WrapObjects still show with the right shape/location/scale etc. to make sure the changes to the XML representation didn't have other unintended consequences.

OpenSim/Tools/Test/testVisualization.cpp
chrisdembia: Cyan is `[0 1 1]` and the model file here has `[0 0.2 0.8]`. Since the color from the model file was not being processed previously in updateFromXMLNode(), we lost the non-default color of `[0 0.2 0.8]`.

OpenSim/Tests/Wrapping/testWrapping.cpp
chrisdembia: Oh oops...nope! I will do that.

chrisdembia:
> I assume you looked at that visually to make sure WrapObjects still show with the right shape/location/scale etc. to make sure the changes to the XML representation didn't have other unintended consequences.

I have now done this, for both testVisualization (for `test_wrapAllVis.osim`) and testWrapping. The visual check indicates that the changes in this PR are correct.

## testVisualization

As expected, the color of the bodies with color [0, 0.2, 0.8] are now properly deserialized from the XML file. The wrap surfaces are drawn in the same location.

### Before

image

### After

image


## testWrapping

### before

We had been ignoring the flags to hide the wrap surfaces, so all 3 showed up:

image

They are a little hard to see, so here is a zoomed-in version.
image

### after

After performing the new updateFromXMLNode() code, only the ellipse is visible, and it is orange; all as expected. The location of the ellipse has not changed.

image

Zooming in shows that the mesh is indeed shown as wireframe, as expected: 

image

chrisdembia:
Ready for review again. I also pushed some very minor changes to testVisualization, and I merged with master in hopes of getting some more of the CI builds to pass.

OpenSim/Simulation/Wrap/WrapObject.cpp
```diff
@@ -178,3 +178,94 @@ int WrapObject::wrapPathSegment(const SimTK::State& s,
...
+            // in the wrong place.
+            SimTK::Xml::Element appearanceNode("Appearance"); 
+            // Use the correct defaults for WrapObject's appearance, which
+            // are different from the default Appearance.
+            SimTK::Xml::Element color("color");
+            color.setValue("0 1 1"); // SimTK::Cyan
+            appearanceNode.insertNodeAfter(appearanceNode.element_end(),
+                    color);
+            SimTK::Xml::Element defaultOpacity("opacity");
+            defaultOpacity.setValue("0.5");
```
tkuchida: Not sure about this, but would it make sense to set these values by creating a default WrapObject and looking up the values of its properties (in case the default property values change in the future)?

tkuchida: :no_entry_sign: Decided this isn't a good idea.

chrisdembia:
Thank you @tkuchida. @aymanhab could you re-review when you get the chance?

aseth1:
@aymanhab this update would be good to have in the beta. If you are happy with it please merge and I can tag a Beta release.

chrisdembia:
Thank you.

aseth1:
Great! Thanks.

Issue #1179, opened 2016-08-11
MovingPathPoint derives from Station but it should not


chrisdembia:
MovingPathPoint derives from Station even though it is not fixed to a Body. Lots of things could conceivably be done with Stations that should not be done with other Points, like defining constraints.

This inheritance from Station was introduced by #946  and also came up in #990 and #1177. In #990, @aseth1 attempted to prevent misuse by setting the location property to NaN for MovingPathPoints.

tkuchida:
> prevent misuse by setting the location property to NaN

Setting the `location` property to `NaN` did not prevent misuse; see #1104.

chrisdembia:
How about "to prevent silent misuse."? It would have been *worse* if you *did* see the moving path points but they weren't moving.

tkuchida:
> It would have been *worse* if you *did* see the moving path points but they weren't moving.

Actually, it would have been easier to spot if `location` were being set to `Vec3(0)` or `Vec3(huge_number)`; I didn't immediately notice there were missing line segments in gait10dof18musc.

What is the proposal? Should MovingPathPoint derive from Point? If so, ConditionalPathPoint would be the only descendant of PathPoint—should PathPoint be absorbed into ConditionalPathPoint?

chrisdembia:
> Actually, it would have been easier to spot if location were being set to Vec3(0) or Vec3(huge_number); I didn't immediately notice there were missing line segments in gait10dof18musc.

Does the same hold true for calculations that use the location, like moment arm or path length? I think `Vec3(huge_number)` might be nearly as good of a red flag as `NaN`, but `Vec3(0)` would not be.

> What is the proposal? Should MovingPathPoint derive from Point?

Something like that. @aseth1 doesn't want special PathPoint classes at all. Just Point classes that the muscles use.

tkuchida:
> Does the same hold true for calculations that use the location, like moment arm or path length?

Good point. Perhaps `Vec3(huge_number)` then, or just throw an exception if `MovingPathPoint::get_location()` is called.

> @aseth1 doesn't want special PathPoint classes at all. Just Point/Station classes that the muscles use.

Okay, I won't go after this, then. Thanks for the heads-up.

chrisdembia:
Yeah I don't think the fix is straightforward.

jenhicks:
@aseth1 Will move PathPoint out of Station.

jenhicks:
[...]

PR #299, opened 2014-12-19
Replace FINAL_11 with final.

chrisdembia:
The changes to OpenSim/Wrapping/SWIGSimTK/common.h should be discussed. See some discussion over at simbody [here](https://github.com/simbody/simbody/pull/302/files#diff-ba43f68612864196b99e9cd542e95b73L239). This common.h file in simbody was just updated by @sherm1.

PR #1595, opened 2017-03-17
Remove PathPoint from Station derived hierachy

aseth1:
Fixes issue #1179 

### Brief summary of changes
 1. Introduced a new base `AbstractPathPoint` class from which (stationary) `PathPoint` and `MovingPathPoint` derive
 2. `PathPoint` now uses a `Station` component internally instead of deriving from `Station` (to avoid multiple inheritance)
 3. Deprecated old `PathPoint` methods of `get/setBody()` now members of `AbstractPathPoint`

### Testing I've completed
 - existing test should pass
 - *testIterators* updated to account for internal `Station` components

 - impact on the GUI (if any) from @aymanhab 

### CHANGELOG.md 
-  no need to update because old interface was preserved (but deprecated)

### DIFF
--- a/Bindings/Java/OpenSimJNI/OpenSimContext.cpp
+++ b/Bindings/Java/OpenSimJNI/OpenSimContext.cpp
@@ -240,7 +240,9 @@ bool OpenSimContext::replacePathPoint(GeometryPath& p, AbstractPathPoint& mp, Ab
 }
 
 void OpenSimContext::setLocation(PathPoint& mp, int i, double d) {
-    mp.setLocationCoord(*_configState, i, d);
+    SimTK::Vec3 loc = mp.getLocation(*_configState);
+    loc[i] = d;
+    mp.setLocation(loc);
     _configState->invalidateAll(SimTK::Stage::Position);
     _model->getMultibodySystem().realize(*_configState, SimTK::Stage::Position);
 }
--- a/Bindings/Java/OpenSimJNI/OpenSimContext.cpp
+++ b/Bindings/Java/OpenSimJNI/OpenSimContext.cpp
@@ -120,7 +120,7 @@ double OpenSimContext::getMuscleLength(Muscle& m) {
   return m.getLength(*_configState);
 }
 
-const Array<PathPoint*>& OpenSimContext::getCurrentPath(Muscle& m) {
+const Array<AbstractPathPoint*>& OpenSimContext::getCurrentPath(Muscle& m) {
   return m.getGeometryPath().getCurrentPath(*_configState);
 }
 
@@ -174,10 +174,13 @@ void OpenSimContext::setZCoordinate(MovingPathPoint& mmp, Coordinate&  newCoord)
     return;
 }
 
-void OpenSimContext::setBody(PathPoint& pathPoint, PhysicalFrame&  newBody) {
-   pathPoint.changeBodyPreserveLocation(*_configState, newBody);
-   this->recreateSystemAfterSystemExists();
-   realizeVelocity();
+void OpenSimContext::setBody(AbstractPathPoint& pathPoint, PhysicalFrame&  newBody) {
+    PathPoint* spp = dynamic_cast<PathPoint*>(&pathPoint);
+    if (spp) {
+        spp->changeBodyPreserveLocation(*_configState, newBody);
+        this->recreateSystemAfterSystemExists();
+        realizeVelocity();
+    }
 }
 
 
@@ -220,7 +223,7 @@ void OpenSimContext::setRangeMax(ConditionalPathPoint&  via, double d) {
     return;
 }
 
-bool OpenSimContext::replacePathPoint(GeometryPath& p, PathPoint& mp, PathPoint& newPoint) {
+bool OpenSimContext::replacePathPoint(GeometryPath& p, AbstractPathPoint& mp, AbstractPathPoint& newPoint) {
    bool ret= p.replacePathPoint(*_configState, &mp, &newPoint );
    recreateSystemAfterSystemExists();
    realizeVelocity();
@@ -228,11 +231,13 @@ bool OpenSimContext::replacePathPoint(GeometryPath& p, PathPoint& mp, PathPoint&
    return ret;
 }
 
-void OpenSimContext::setLocation(PathPoint& mp, int i, double d) {
-    mp.setLocation(*_configState, i, d);
-  _configState->invalidateAll(SimTK::Stage::Position);
-  _model->getMultibodySystem().realize(*_configState, SimTK::Stage::Position);
-    return;
+void OpenSimContext::setLocation(AbstractPathPoint& mp, int i, double d) {
+    PathPoint* spp = dynamic_cast<PathPoint*>(&mp);
+    if (spp) {
+        spp->setLocationCoord(*_configState, i, d);
+        _configState->invalidateAll(SimTK::Stage::Position);
+        _model->getMultibodySystem().realize(*_configState, SimTK::Stage::Position);
+    }
 }
 
 void OpenSimContext::setEndPoint(PathWrap& mw, int newEndPt) {
@@ -257,7 +262,7 @@ bool OpenSimContext::deletePathPoint(GeometryPath& p, int menuChoice) {
   return ret;
 }
 
-bool OpenSimContext::isActivePathPoint(PathPoint& mp) {
+bool OpenSimContext::isActivePathPoint(AbstractPathPoint& mp) {
   return mp.isActive(*_configState);
 };
 
--- a/Bindings/Java/OpenSimJNI/OpenSimContext.cpp
+++ b/Bindings/Java/OpenSimJNI/OpenSimContext.cpp
@@ -239,13 +239,10 @@ bool OpenSimContext::replacePathPoint(GeometryPath& p, AbstractPathPoint& mp, Ab
    return ret;
 }
 
-void OpenSimContext::setLocation(AbstractPathPoint& mp, int i, double d) {
-    PathPoint* spp = dynamic_cast<PathPoint*>(&mp);
-    if (spp) {
-        spp->setLocationCoord(*_configState, i, d);
-        _configState->invalidateAll(SimTK::Stage::Position);
-        _model->getMultibodySystem().realize(*_configState, SimTK::Stage::Position);
-    }
+void OpenSimContext::setLocation(PathPoint& mp, int i, double d) {
+    mp.setLocationCoord(*_configState, i, d);
+    _configState->invalidateAll(SimTK::Stage::Position);
+    _model->getMultibodySystem().realize(*_configState, SimTK::Stage::Position);
 }
 
 void OpenSimContext::setEndPoint(PathWrap& mw, int newEndPt) {
--- a/Bindings/Java/OpenSimJNI/OpenSimContext.cpp
+++ b/Bindings/Java/OpenSimJNI/OpenSimContext.cpp
@@ -174,10 +174,18 @@ void OpenSimContext::setZCoordinate(MovingPathPoint& mmp, Coordinate&  newCoord)
     return;
 }
 
-void OpenSimContext::setBody(AbstractPathPoint& pathPoint, PhysicalFrame&  newBody) {
+void OpenSimContext::setBody(AbstractPathPoint& pathPoint, PhysicalFrame&  newBody)
+{
     PathPoint* spp = dynamic_cast<PathPoint*>(&pathPoint);
     if (spp) {
         spp->changeBodyPreserveLocation(*_configState, newBody);
+        this->recreateSystemAfterSystemExists();
+        realizeVelocity();
+        return;
+    }
+    MovingPathPoint* mpp = dynamic_cast<MovingPathPoint*>(&pathPoint);
+    if (mpp) {
+        mpp->setParentFrame(newBody);
         this->recreateSystemAfterSystemExists();
         realizeVelocity();
     }
--- a/OpenSim/Simulation/Model/AbstractPathPoint.cpp
+++ b/OpenSim/Simulation/Model/AbstractPathPoint.cpp
@@ -8,7 +8,6 @@
  * through the Warrior Web program.                                           *
  *                                                                            *
  * Copyright (c) 2005-2017 Stanford University and the Authors                *
- * Author(s): Peter Loan                                                      *
  *                                                                            *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may    *
  * not use this file except in compliance with the License. You may obtain a  *
--- a/OpenSim/Simulation/Model/AbstractPathPoint.cpp
+++ b/OpenSim/Simulation/Model/AbstractPathPoint.cpp
@@ -24,7 +24,6 @@
 // INCLUDES
 //=============================================================================
 #include "AbstractPathPoint.h"
-#include "OpenSim/Simulation/Model/PhysicalFrame.h"
 
 //=============================================================================
 // STATICS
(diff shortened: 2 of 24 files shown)


tkuchida:
I needed the inheritance diagram to review but the CI tests are failing so myosin hasn't built the doxygen. I pulled the branch and built locally. Here it is in case other reviewers need it:

inheritance

OpenSim/Simulation/Model/GeometryPath.cpp
```diff
@@ -350,14 +350,14 @@ void GeometryPath::addInEquivalentForces(const SimTK::State& s,
             // add in the tension point forces to body forces
             if (mppo) {// moving path point location is a function of the state
                 // transform of the frame of the point to the base mobilized body
-                auto X_BF = mppo->getParentFrame().findTransformInBaseFrame();
+                auto X_BF = mppo->getBody().findTransformInBaseFrame();
```
tkuchida: Perhaps avoid use of newly deprecated methods?

OpenSim/Simulation/Model/PathPoint.h
```diff
@@ -35,88 +35,76 @@ class PhysicalFrame;
 //=============================================================================
 //=============================================================================
 /**
- * A class implementing a path point.
- *
- * @author Peter Loan
- * @version 1.0
+ * A PathPoint that is stationary with respect to parent's PhysicalFrame
```
tkuchida: This description ("A [type of] PathPoint that is stationary…") would make sense only for a class that *derives* from PathPoint. I suggest "A PathPoint" -> "A path point".

tkuchida: Should also mention that PathPoint uses a Station under the hood so that the description of the "location" property makes sense.

OpenSim/Simulation/Model/PathPoint.h
```diff
@@ -35,88 +35,76 @@ class PhysicalFrame;
...
-class OSIMSIMULATION_API PathPoint : public Station {
-OpenSim_DECLARE_CONCRETE_OBJECT(PathPoint, Station);
+class OSIMSIMULATION_API PathPoint : public AbstractPathPoint {
+    OpenSim_DECLARE_CONCRETE_OBJECT(PathPoint, AbstractPathPoint);
+public:
+//==============================================================================
+// PROPERTIES
+//==============================================================================
+    OpenSim_DECLARE_PROPERTY(location, SimTK::Vec3,
+        "The fixed location of the station expressed in its parent frame.");
```
tkuchida: The user should be told that PathPoint uses a Station.

OpenSim/Simulation/Model/PathPoint.h
```diff
@@ -35,88 +35,76 @@ class PhysicalFrame;
...
-    const double& getLocationCoord(int aXYZ) const {
-        assert(aXYZ>=0 && aXYZ<=2); return get_location()[aXYZ];
-    }
-
-    // A variant that uses basic types for use by GUI
-    void setLocationCoord(int aXYZ, double aValue) {
+    // A variant that uses basic types for use by GUI, e.g. OpenSimContext
+    /** <b>(Deprecated)</b> Old PathPoint interface */
+    DEPRECATED_14("Use setLocation() instead.")
+    void setLocationCoord(const SimTK::State&s, int aXYZ, double aValue) {
```
tkuchida: Using the [SimTK::CoordinateAxis](https://simtk.org/api_docs/simbody/3.5/classSimTK_1_1CoordinateAxis.html) class would be cleaner than `int` followed by `assert(aXYZ>=0 && aXYZ<=2)`.

OpenSim/Simulation/Model/PathPoint.h
```diff
@@ -35,88 +35,76 @@ class PhysicalFrame;
...
 
-    // get the relative velocity of the path point with respect to the body
-    // it is connected to.
-    virtual void getVelocity(const SimTK::State& s, SimTK::Vec3& velocity) {
-        velocity.setToZero();
+private: 
+    // Satisfy Point interface
+    /* Calculate the location of this PathPoint in Ground as a function of
+       the state. */
+    SimTK::Vec3 calcLocationInGround(const SimTK::State& state) const final {
```
tkuchida: Shouldn't these three methods be `getLocationIn…()`? They don't do anything more than call `get…()` methods so I don't see why they need to become `calc`.

OpenSim/Simulation/Model/PathPoint.h
```diff
@@ -125,4 +113,4 @@ OpenSim_DECLARE_CONCRETE_OBJECT(PathPoint, Station);
 
 } // end of namespace OpenSim
 
-#endif // OPENSIM_PATH_POINT_H_
+#endif // OPENSIM_STATIONARY_PATH_POINT_H_
```
tkuchida: OPENSIM_STATIONARY_PATH_POINT_H_ -> OPENSIM_PATH_POINT_H_ to match line 1.

OpenSim/Simulation/Model/PathPoint.h
chrisdembia: I agree, though perhaps an overload that takes an int would be handy for MATLAB users.

OpenSim/Simulation/Model/PathPoint.h
chrisdembia: I think they are implementing virtual functions (thus the `final`), and the virtual function assumes a calculation may be necessary. However, we decided to mark a final function as `override final`, not simply `final` (we went back and forth on this, but ultimately decided on `override final`: https://github.com/opensim-org/opensim-core/pull/299).

OpenSim/Tests/Wrapping/testWrapping.cpp
```diff
@@ -617,14 +619,14 @@ void simulateModelWithCables(const string &modelFile, double finalTime)
...
                         ObstacleInfo* obs = wrapObs[k];
                         obs->isActive = true;
                         // pp and next pp are wrap points
                         Transform X_SB = obs->X_BS.invert();
-                        PathPoint* pp_next = activePathPoints[++i]; // increment to next pp
-                        obs->P_S = X_SB*pp->getLocation();
-                        obs->Q_S = X_SB*pp_next->getLocation();
+                        AbstractPathPoint* pp_next = activePathPoints[++i]; // increment to next pp
+                        obs->P_S = X_SB*pp->getLocation(si);
+                        obs->Q_S = X_SB*pp_next->getLocation(si);
```
chrisdembia: My sense is that these changes are actually fixing bugs for the case where the active points include MovingPathPoints. Which is great.

OpenSim/Simulation/Model/GeometryPath.cpp
chrisdembia: Agreed, especially since the functions seem to do the same thing. Even if `getBody()` actually returned the base frame (body) and not the parent frame, the function call is still to get the transform in the base frame.

Bindings/Java/OpenSimJNI/OpenSimContext.cpp
```diff
@@ -220,19 +223,21 @@ void OpenSimContext::setRangeMax(ConditionalPathPoint&  via, double d) {
...
-  _configState->invalidateAll(SimTK::Stage::Position);
-  _model->getMultibodySystem().realize(*_configState, SimTK::Stage::Position);
-    return;
+void OpenSimContext::setLocation(AbstractPathPoint& mp, int i, double d) {
+    PathPoint* spp = dynamic_cast<PathPoint*>(&mp);
+    if (spp) {
+        spp->setLocationCoord(*_configState, i, d);
+        _configState->invalidateAll(SimTK::Stage::Position);
+        _model->getMultibodySystem().realize(*_configState, SimTK::Stage::Position);
+    }
```
aseth1: If we don't want "to do nothing", which was effectively what was happening before, I would prefer to only allow `OpenSim::setLocation(PathPoint ...)` and make it impossible to call with a `MovingPathPoint` altogether.

Bindings/Java/OpenSimJNI/OpenSimContext.cpp
```diff
@@ -174,10 +174,13 @@ void OpenSimContext::setZCoordinate(MovingPathPoint& mmp, Coordinate&  newCoord)
     return;
 }
 
-void OpenSimContext::setBody(PathPoint& pathPoint, PhysicalFrame&  newBody) {
-   pathPoint.changeBodyPreserveLocation(*_configState, newBody);
-   this->recreateSystemAfterSystemExists();
-   realizeVelocity();
+void OpenSimContext::setBody(AbstractPathPoint& pathPoint, PhysicalFrame&  newBody) {
+    PathPoint* spp = dynamic_cast<PathPoint*>(&pathPoint);
+    if (spp) {
```
aseth1: I will add that.

OpenSim/Simulation/Model/GeometryPath.cpp
aseth1: This was a mistake on my part. I only replaced to verify that `getBody()` still works and meant to revert and not commit the change. I took a pass through GeometryPath to update all the `Body` related calls to `ParentFrame`.

Bindings/Java/OpenSimJNI/OpenSimContext.cpp
chrisdembia: @aseth1 I have edited your comment to read `OpenSimContext::setLocation(PathPoint&...)`

aseth1:
> It's very nice that the inheritance hierarchy is cleaned up, and my naive understanding is that you may have fixed bugs too. Do we have tests that cover MovingPathPoints (that is, models with MovingPathPoints)?

Thanks @chrisdembia. There weren't any bugs specifically related to `MovingPathPoints` that I was addressing here. There are specific checks like [here](https://github.com/opensim-org/opensim-core/pull/1595/files#diff-9be6b622aa38d36047ff8d9d42d5c7a9L351) for `MovingPathPoints` which could now be removed because of these changes and we could uniformly call `getLocation(state)`. I could test and include that as part of this change- it would be another improvement to remove unnecessary `dynamic_cast`s especially in the inner loop of path calculations.

aymanhab:
@aseth1 in terms of impact on the GUI, the code base (as of 3.3) doesn't call methods on all kinds of PathPoints directly but they do through the OpenSimContext since there's one centralized state object maintained by OpenSimContext that backs the current model in the GUI. As long as OpenSimContext methods have been reimplemented there should be no significant impact.

aymanhab:
Main issue that may require large "Textual" change is that the GeometryPath now maintains an array of AbstractPathPoints, it would be ideal if these can be used as is without safeDownCast (both for GUI and Matlab) use.

OpenSim/Simulation/Model/AbstractPathPoint.cpp
```diff
@@ -0,0 +1,79 @@
...
+{
+    setParentFrame(body);
+}
+
+const std::string& AbstractPathPoint::getBodyName() const {
+    return getParentFrame().getName();
+}
+
+void AbstractPathPoint::updateFromXMLNode(SimTK::Xml::Element& aNode, int versionNumber)
+{
```
aymanhab: Not sure how this is supposed to work, it's surprising that a new class has an updateFromXMLNode implementation since the deserialization code would not instantiate such class from an older version, unless in this case it is called explicitly by the concrete subclass(es). Can you please clarify, @aseth1 Thanks

OpenSim/Simulation/Model/ConditionalPathPoint.cpp
```diff
@@ -74,7 +74,7 @@ void ConditionalPathPoint::updateFromXMLNode(SimTK::Xml::Element& node,
     }
 
     // Call base class now assuming _node has been corrected for current version
-    PathPoint::updateFromXMLNode(node, versionNumber);
+    Super::updateFromXMLNode(node, versionNumber);
```
aymanhab: Maybe this explains the updateFromXMLNode changes above?

OpenSim/Simulation/Model/PathPoint.h
aymanhab: :+1: so that the property comment makes sense.

OpenSim/Simulation/Model/PathPoint.h
aymanhab: Agreed @tkuchida :+1:, this was written before SimTK classes were exposed to Java/Matlab but it looks like this can easily be done now. Would be good to keep track of it but doesn't seem necessary/related to this PR. The OpenSimContext class will likely have to undergo some changes anyway when the MuscleEditor is refactored/rewritten and will be slimmed down significantly.

aseth1: BTW, the existing signature with the `State` already [existed](override). I would prefer that this method be removed altogether, but I left to to reduce the *havoc* caused to the GUI. If @ayamhab can use the new interface that would be ideal.

chrisdembia:
> There are specific checks like here for MovingPathPoints which could now be removed because of these changes and we could uniformly call getLocation(state). I could test and include that as part of this change- it would be another improvement to remove unnecessary dynamic_casts especially in the inner loop of path calculations.

Thanks for explaining; I think I understand. I don't think it's necessary to increase the scope of the PR.

OpenSim/Simulation/Model/PathPoint.h
aseth1: I will revise the property description. It should not be necessary to know the internal implementation of the `PathPoint` that now uses a `Station`.

OpenSim/Simulation/Wrap/WrapObject.cpp
```diff
@@ -152,10 +154,10 @@ int WrapObject::wrapPathSegment(const SimTK::State& s,
     // Convert the path points from the frames of the bodies they are attached
     // to, to the frame of the wrap object's body
     pt1 = aPoint1.getBody()
-        .findStationLocationInAnotherFrame(s, aPoint1.getLocation(), getFrame());
+        .findStationLocationInAnotherFrame(s, aPoint1.getLocation(s), getFrame());
     
     pt2 = aPoint2.getBody()
-        .findStationLocationInAnotherFrame(s, aPoint2.getLocation(), getFrame());
+        .findStationLocationInAnotherFrame(s, aPoint2.getLocation(s), getFrame());
```
aseth1: I was not addressing a bug here, but since `WrapObject` always deal with `PathWrapPoint`s at least implicitly, this likely never encounters a `MovingPathPoint`.

OpenSim/Tests/Wrapping/testWrapping.cpp
aseth1: It might be the case, but I was not targeting to fix these. I was just getting all `PathPoint`s to conform to the `getLocation(state)` interface.

OpenSim/Simulation/Model/AbstractPathPoint.cpp
aseth1: Derived classes would have to call explicitly if they override the virtual method, but since they do not implement the base implementation is invoked. I initially had the implementation on `PathPoint` but then all `MovingPathPoint`s were failing. Instead of duplicating the code i moved up the `AbstractPathPoint` and that resolved the issue for loading older model files.

aymanhab: Thanks @aseth1 Maybe a note/comment there to explain how this method ended up in the new AbstractPathPoint would help future maintainers.

aymanhab:
Much better/logical class hierarchy @aseth1 :+1:
Is there a reason for concern that the number of components has grown significantly (since every stationary pathPoint now has two entries in the component tree? as demonstrated by the component count in testIterators?)

aseth1:
> Is there a reason for concern that the number of components has grown significantly (since every stationary pathPoint now has two entries in the component tree? as demonstrated by the component count in testIterators?)

I don't believe that is reason for concern. The internal class introduces minimal overhead, if any, since methods pass straight through to `Station`. Besides using a ComponentList through every component there isn't any way to know, or reason to have to know, that `PathPoint` uses a `Station`.

aseth1:
@tkuchida @chrisdembia @aymanhab: thank you for your reviews. I have attempted to address the issues that you have raised. The PR has been updated and is ready for (hopefully easy) re-review.

aymanhab:
LGTM, just need to fix build failure if related. Thank you.

aseth1:
Thank you @tkuchida, @chrisdembia  and @aymanhab. Fixed test failure and now merging.

Issue #1599, opened 2017-03-21
Component with 1-character-long name is omitted from state variable path


chrisdembia:
If the name of a coordinate is two characters long, then getStateVariableNames() gives the correct output.

```cpp
        using SimTK::Vec3;
        Model model;
        model.setName("model");
        auto* body = new Body("body", 0.5, Vec3(0), SimTK::Inertia(0));
        model.addComponent(body);

        auto* joint = new SliderJoint();
        joint->setName("joint");
        joint->connectSocket_parent_frame(model.getGround());
        joint->connectSocket_child_frame(*body);
        auto& coord = joint->updCoordinate(SliderJoint::Coord::TranslationX);
        coord.setName("qa");              //  <---- two characters long
        model.addComponent(joint);
        model.initSystem();
        const auto svnames = model.getStateVariableNames();
        for (int i = 0; i < svnames.getSize(); ++i) {
            std::cout << svnames[i] << std::endl;
        }
```

Output:

```
joint/qa/value
joint/qa/speed
```

*However*, if the coordinate's name is 1 character long, its name is omitted from the path:

```cpp
        using SimTK::Vec3;
        Model model;
        model.setName("model");
        auto* body = new Body("body", 0.5, Vec3(0), SimTK::Inertia(0));
        model.addComponent(body);

        auto* joint = new SliderJoint();
        joint->setName("joint");
        joint->connectSocket_parent_frame(model.getGround());
        joint->connectSocket_child_frame(*body);
        auto& coord = joint->updCoordinate(SliderJoint::Coord::TranslationX);
        coord.setName("q");              //  <---- one character long
        model.addComponent(joint);
        model.initSystem();
        const auto svnames = model.getStateVariableNames();
        for (int i = 0; i < svnames.getSize(); ++i) {
            std::cout << svnames[i] << std::endl;
        }
```

Output:

```
joint/value
joint/speed
```

tkuchida:
Perhaps `if (back >= front)` was intended [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Component.cpp#L748), [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Component.cpp#L763), and [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Component.cpp#L777)?

chrisdembia:
> Perhaps if (back >= front) was intended here, here, and here?

I think you're right; nice find.

[...]

PR #1600, opened 2017-03-21
Avoid excluding single-character Component name in path

tkuchida:
Fixes issue #1599.

### Brief summary of changes
Fixed off-by-one string parsing bug.

### Testing I've completed
Tests pass locally. Also tested on example code provided in #1599.

No specific feedback sought.

### CHANGELOG.md (choose one)
No need to update because this code is part of the new 4.0 API.

### DIFF
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -745,7 +745,7 @@ Array<std::string> Component::getStateVariableNames() const
         std::string::size_type front = subCompName.find_first_not_of(" \t\r\n");
         std::string::size_type back = subCompName.find_last_not_of(" \t\r\n");
         std::string prefix = "";
-        if (back > front) // have non-whitespace name
+        if (back >= front) // have non-whitespace name
             prefix = subCompName + "/";
         for (int j = 0; j<nsubs; ++j) {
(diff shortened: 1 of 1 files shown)


tkuchida:
Timeout of testCMCEMGDrivenArm_Millard on Travis (clang/Debug).

aseth1:
Changes LGTM. Do you want to include the example code in #1599 (or similar) as a test case in `testComponentInterface`'s `testComponentPathNames` subtest?

aseth1:
@tkuchida I don't think it is necessary to add the test formally since you verified it. You're welcome to add it later for completeness if you wish, but I will merge it so we'll have your fix as part of the beta.

tkuchida:
@aseth1 Not sure what you'd like done here. Currently, testComponentInterface.cpp lives in OpenSim/Common and doesn't use Model or the osimSimulation library. The string parsing bug was in `Component::getStateVariableNames()` and AFAIK the only way it can be called without throwing a [`!hasSystem()` exception](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Component.cpp#L720) is if Model.h is #included in testComponentInterface.cpp so that `Model::initSystem()` can be called. Is there a better home for this test or am I missing something?

Issue #1591, opened 2017-03-16
Add a DataTable::setRow(), for bindings


chrisdembia:
Right now, it is not possible to use `updRow()` (?) to set a row in MATLAB. So it would be nice to have a `updRow()`. Reported by @jimmyDunne .

klshrinidhi:
Will get back on this.

tkuchida:
Is there a reasonable workaround or is this Issue on the critical path to release?

chrisdembia:
Perhaps this feature needs to be tested in MATLAB in order to close this issue. It seems to already be tested in Java; I wonder if that's enough.

aseth1:
> Perhaps this feature needs to be tested in MATLAB  ...

That's what I was looking for. In reviewing #1434 I was able to run the scripts that executed `setRowAtIndex` without any issue. My impression is that @jimmyDunne can and is using this wrapped method in MATLAB. If that is the case, @jimmyDunne please confirm and then close. Thanks!

jimmyDunne:
Hey, Guys-
Tested setRow() this morning. It is working! 🍾 

Thanks!!!

PR #1602, opened 2017-03-21
Add DataTable::setRow()

klshrinidhi:
Addressing #1591.

I have added the function on C++ side because it was less work that adding the function to every specialization in SWIG interface file. The documentation for this function makes it clear it is equivalent to calling `updRow()`.

### DIFF
--- a/OpenSim/Common/DataTable.h
+++ b/OpenSim/Common/DataTable.h
@@ -721,6 +721,56 @@ public:
         return _depData.updRow((int)std::distance(_indData.cbegin(), iter));
     }
 
+    /** Set row at index. Equivalent to
+    ```
+    updRowAtIndex(index) = depRow;
+    ```
+
+    \throws RowIndexOutOfRange If the index is out of range.                  */
+    void setRowAtIndex(size_t index, const RowVectorView& depRow) {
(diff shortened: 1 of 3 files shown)


OpenSim/Common/DataTable.h
```diff
@@ -721,6 +721,56 @@ class DataTable_ : public AbstractDataTable {
         return _depData.updRow((int)std::distance(_indData.cbegin(), iter));
     }
 
+    /** Set row at index. Equivalent to
+    ```
+    updRowAtIndex(index) = depRow;
+    ```
+
+    \throws RowIndexOutOfRange If the index is out of range.                  */
+    void setRowAtIndex(size_t index, const RowVectorView& depRow) {
```
tkuchida: Minor: I suggest adjusting the ordering of these methods. They're currently
- 00 setRowAtIndex with RowVectorView
- 11 setRow with RowVector
- 01 setRowAtIndex with RowVector
- 10 setRow with RowVectorView

I suggest either 00-01-10-11 or 00-10-01-11.

klshrinidhi: Done.

PR #1610, opened 2017-03-24
[cmake] Find consistent python libraries.

### LINKED ISSUE #1303: FInding Python version on path (Cmake)
When building the python wrapping, Cmake defaults to a standard dir (OSX; user/bin/python2.7). If that is not the version of python that is on your path, you only get exceptions once you try to `import opensim`.

Would be swell if Cmake could find the python version currently being used.

chrisdembia:
Fixes #1303

### Brief summary of changes

This change makes sure that on macOS, we find python libraries that are consistent with the found python executable.

### Testing I've completed

Built python bindings locally and ran python tests.

### CHANGELOG.md (choose one)

- no need to update because...this issue was introduced after 3.3.

### DIFF
--- a/Bindings/Python/CMakeLists.txt
+++ b/Bindings/Python/CMakeLists.txt
@@ -1,7 +1,4 @@
 
-# Find python.
-# ============
-
 # Location of the opensim python package in the build directory, for testing.
 if(MSVC OR XCODE)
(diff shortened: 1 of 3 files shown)


chrisdembia:
Thanks @klshrinidhi

PR #1611, opened 2017-03-24
Remove setting (DY)LD_LIBRARY_PATH from README.

chrisdembia:
We now use RPATH so that macOS/Linux users need not set (DY)LD_LIBRARY_PATH; we
can remove these descriptions from the README.


chrisdembia:
Thank you @klshrinidhi

PR #1620, opened 2017-03-27
[TGCS] Create script for evaluating a hop.

chrisdembia:
I created a script called EvaluateHop that takes a model, runs a simulation, and reports the peak height, metabolic cost, and device power consumption.

This script could be used during the competition; competitors could submit a model file by email, etc., and we could run their model through this script for the leader board. They can also use it locally to evaluate their model, and the HopperExample GUI could run this script when the user clicks "Simulate."

@jimmyDunne @carmichaelong @nickbianco


carmichaelong:
LGTM. Do we have any review guidelines for this branch?

chrisdembia:
Nope; not yet. After @nickbianco puts a LGTM stamp on this, then perhaps one of you could merge it?

chrisdembia:
Thank you!

PR #1625, opened 2017-03-30
opensim-cmd: Fix loading model for AbstractTool.

### LINKED ISSUE #1623: opensim-cmd run-tool does not load model for AbstractTools
If a model is specified via the XML `model_file` field, AbstractTool will load this model in the derived class constructor. Right now, `opensim-cmd` does not invoke this derived class constructor; this causes `opensim-cmd run-tool` to not load model files for AbstractTools (analyze, cmc).

chrisdembia:
Fixes #1623 

### Brief summary of changes

There is currently a bug in that, when attempting to run an AbstractTool (e.g., CMC, RRA, Analyze), the model won't be loaded even if a valid model file is specified in the XML file. This PR fixes that bug by calling the concrete constructors for the AbstractTools.

This solution fixes the bug, but is not very generalizable. I could have attempted a more general solution by moving the model-loading code to the AbstractTool. However, I wanted to limit the scope of this PR and make sure I did not introduce new bugs. We are not currently eager to introduce new tools, so I think it is fine that this solution is not very generic; I believe it is still an improvement over the separate command-line executables we had for 3.3.

### Testing I've completed

I used `opensim-cmd run-tool` for a number of setup files in the RRA, CMC, Forward, and Analyze test directories, and the tool was able to run successfully.

### CHANGELOG.md (choose one)

- no need to update because opensim-cmd did not exist in 3.3.

### DIFF
--- a/Applications/opensim-cmd/opensim-cmd_run-tool.h
+++ b/Applications/opensim-cmd/opensim-cmd_run-tool.h
@@ -73,28 +73,46 @@ int run_tool(int argc, const char** argv) {
 
     // Deserialize.
     const auto& setupFile = args["<setup-xml-file>"].asString();
-    Object* obj = Object::makeObjectFromFile(setupFile);
+    auto obj = std::unique_ptr<Object>(Object::makeObjectFromFile(setupFile));
     if (obj == nullptr) {
         throw Exception( "A problem occurred when trying to load file '" +
                 setupFile + "'.");
     }
 
     // Detect and run the tool.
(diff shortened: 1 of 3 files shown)


chrisdembia:
Thank you @tkuchida.

aseth1:
LGTM :+1:. I agree that Tools are confusing and I tried to start clean with IK and ID, which were redone with Solvers, but didn't get around to the others. I don't know why ScaleTool is its own island (deriving directly from Object). 

Merging since it is addresses the reported bug.

chrisdembia:
Thank you @aseth1

PR #1626, opened 2017-03-30
[TGCS] Major overhaul to hopper GUI 

nickbianco:
Per our conversations with Ajay and James this week, I've made the following changes to the interactive hopper (yes, I will do smaller PRs for this going forward).

## Main things
- Added in baked in muscle settings that update the hopper "vastus" based on built in parameters
- Passive and active devices can now be adjusted via sliders, which changes performance of each device based on spring stiffness (passive device), max tension (active, unique control),  gain (active, prop myo), and mass
- Recent and best max height of hopper now reported in GUI
- Majority of the hopper's mass moved the pelvis, included masses added by muscles and devices. Masses of 1 kg kept for thigh and shank links for numerical stability

## Small stuff
- HopperExample.m renamed to InteractiveHopper.m (lost the diffs on this one because I needed to use "Save As" to rename the GUI, so we shouldn't rename this again)
- RunHopperGUI.m renamed to RunInteractiveHopperSolution.m 
- Renamed muscle "activations" to "excitations" everywhere
- Renamed device "activations" to "controls" everywhere
- Changed the default muscle to a Millard2012EquilibriumMuscle
- Spring initial length set to the length of the PathActuator at hopper initial conditions (for both patella wrapped and non wrapped cases)
- Small fixes and error handling for control signal plotting


chrisdembia:
Nick, this is awesome.

(A) I ran the RunHopper_answers and RunHopperWithDevice_answers scripts, and they ran just fine. The RunHopperWithDevice script is supposed to make the hopper jump, but now that the model is heavier, it no longer hops. That's fine for now though, we can fix that in a separate PR.

(B) If I launch the GUI and first click Simulate, the model does not simulate and I get this error:

```
Undefined function or variable 'maxTension'.
Error in InteractiveHopper>simulate_Callback (line 172)
             'maxTension', maxTension, ...
Error in gui_mainfcn (line 95)
        feval(varargin{:});
Error in InteractiveHopper (line 50)
    gui_mainfcn(gui_State, varargin{:});
    Error in
    matlab.graphics.internal.figfile.FigFile/read>@(hObject,eventdata)InteractiveHopper('simulate_Callback',hObject,eventdata,guidata(hObject))
Error while evaluating UIControl Callback
```

I can get passed the error if I enable the active device.

(C) On my laptop, the InteractiveHopper GUI takes up most of my screen. I thought I had seen that the window was smaller on @nickbianco's computer . How is the size of the GUI determined? It definitely seems like it could be smaller, but I am unable to resize the window. Could we run into issues for people with smaller monitors?
image

chrisdembia:
Two more comments:

(D) I am getting some confusing behavior with "Wrap Around Patella". If I only have the active device, Wrap Around Patella seems to work correctly. But if I have both devices, then the Active device wraps around the patella even if the box is not checked.

(E) The units for Max Tension say "N-m"; I think this should be "N" now.

Bindings/Java/Matlab/Hopper_Device/BuildHopper.m
```diff
@@ -24,26 +24,29 @@
...
-defaultMuscleModel = 'Thelen2003';
+defaultMuscleModel = 'Millard2012Equilibrium';
 defaultMaxIsometricForce = 4000.0;
 defaultMillardTendonParams = [0.049 28.1 0.67 0.5 0.25];
-defaultActivation = [0.0 1.99 2.0 3.89 3.9 4.0;
+defaultExcitation = [0.0 1.99 2.0 3.89 3.9 4.0;
                      0.3 0.3  1.0 1.0  0.1 0.1];
-defaultPrintModel = false; 
+defaultPrintModel = false;
+defaultAdditionalMass = 0;
```
nickbianco: I think this makes sense. We could have built-in "slider value to mass/stiffness/maxTension/gain" functions in BuildHopper.m and BuildDevice.m (not hard to move these over). After the workshop, users might want to tweak these functions to augment the challenge for their needs.

Bindings/Java/Matlab/Hopper_Device/getUserActivation.m
```diff
@@ -4,30 +4,38 @@
 w = [0 5 0 1];
```
nickbianco: Yes, I'll move it to InteractiveHopper.

Bindings/Java/Matlab/Hopper_Device/BuildHopper.m
```diff
@@ -53,11 +56,12 @@
 %% Bodies and joints.
 % -------------------
 % Create the pelvis, thigh, and shank bodies.
-pelvisMass = 30.0; pelvisHalfLength = 0.1;
+pelvisMass = 50.0 + additionalMass;
```
nickbianco: I didn't check this. You and @carmichaelong can play around with mass values to make it behave like the original demo.

Bindings/Java/Matlab/Hopper_Device/RunInteractiveHopperSolution.m
```diff
@@ -21,65 +21,82 @@
...
+defaultDeviceControl = [0.0 1.99 2.0 3.89 3.9 4.0;
                      0.3 0.3  1.0 1.0  0.1 0.1];
+defaultActiveMass = 0;
+defaultMillardTendonParams = [0.049 28.1 0.67 0.5 0.25];
+defaultMaxIsometricForce = 4000.0;
 
 addOptional(p,'visualize',defaultVisualize)
-addOptional(p,'muscleActivation',defaultMuscleActivation)
+addOptional(p,'muscleExcitation',defaultMuscleExcitation)
+addOptional(p,'muscleMass',defaultMuscleMass)
```
nickbianco: Noted, will do in the future.

Bindings/Java/Matlab/Hopper_Device/RunInteractiveHopperSolution.m
```diff
@@ -141,13 +161,14 @@ function RunHopperGUI(varargin)
...
+
+    if patellaWrap{d} && hopper.hasComponent([deviceNames{d}])
         if strcmp(deviceNames{d},'device_passive')
-            cable = PathSpring.safeDownCast(hopper.updComponent(['device_' deviceNames{d} '/cableAtoB']));
+            % TODO: Change to downcast from PathSpring when PathSpring is
+            % working for passive device
+            cable = PathActuator.safeDownCast(hopper.updComponent([deviceNames{d} '/cableAtoB']));
         elseif strcmp(deviceNames{d},'device_active')
-            cable = PathActuator.safeDownCast(hopper.updComponent(['device_' deviceNames{d} '/cableAtoB']));
+            cable = PathActuator.safeDownCast(hopper.updComponent([deviceNames{d} '/cableAtoB']));
```
nickbianco: I will change this to reflect the patella wrapping code in RunHopperWithDevice, which works for both scripts. It was causing your patella wrapping issue above.

Bindings/Java/Matlab/Hopper_Device/BuildHopper.m
carmichaelong: Given the structure of all this, it makes sense to do all of these types of changes within this function. I know I mentioned before to @nickbianco that it would be slick to generate a model upon loading the GUI and making changes to that one model, but the flow (of building a new model each time) is fine for this example and not worth changing.

carmichaelong: Could you also add comments to this file about the different options you can add, and possibly defaults for these options? I think moving forward, I'll use this function too, so it'd be nice to know the option names.

nickbianco: I'm going to create a function called InteractiveHopperSettings, that, given a name for a device, muscle, or anything other built-in settings we want, can be called to set those values. I would leave it inside RunInteractiveHopperSolution, except that it would need to be called by InteractiveHopper to update the values shown in the GUI. If we want to cut down the number of files and come up with a better solution, we could figure out a way to keep it in RunInteractiveHopperSolution. It could be nice to keep the settings in a separate function, rather than having to go to the bottom of RunInteractiveHopperSolution.

nickbianco: In a separate PR, I'm going to add another function called InteractiveHopperDefaults which gets called at the start up of the GUI and whenever the user clicks the 'Reset to Defaults' button I've added. That would definitely be able to be left within InteractiveHopper, but, if it's nice to make it a separate file, we could do that too.

nickbianco: @carmichaelong I'll add comments to both of these files so that they're easy to use. After they're up, let me know if there's any ways to structure them better for use in optimizations.

nickbianco:
Ready for review.

Since I introduced the InteractiveHopperSettings file, I still don't compute additional mass within BuildHopper. However, using BuildHopper, BuildDevice, and InteractiveHopperSettings, you should be able set up any analysis or optimization desired outside the GUI.

@carmichaelong, they way I set it up, you should be able to use RunInteractiveHopperSolution as a basis for setting up an optimization. The main difference for your code would be pulling out the code through the building of the hopper and devices before your optimization, and add code to update the hopper and device parameters within the optimizer. This way, you would run an optimization for every muscle+device(s) condition.

carmichaelong:
This is all super cool. Just one suggestion: Are there any problems if you were to move the code to build the hopper/devices into its own function? This would involve moving everything above [this line](https://github.com/opensim-org/opensim-core/blob/51a060c4b13bf5f1dd64df80b2183f6665a18abe/Bindings/Java/Matlab/Hopper_Device/RunInteractiveHopperSolution.m#L192) to its own function. If that works, then it would be a small change that would ensure there'd be no copied code and we can ensure the optimization is playing by the same rules as the GUI.

nickbianco:
See latest commit.

I separated out the code not related to building the hopper solution in RunInteractiveHopperSolution, and renamed it to BuildInteractiveHopperSolution. I also added a "blank" file called ModifyInteractiveHopperSolution: if you set it up for an optimization, we could also use to update the hopper solution within the GUI without having to rebuild the hopper (see [this line](https://github.com/opensim-org/opensim-core/pull/1626/files#diff-bc236ce0c04e26b727c18aff1c2a219eR160)).

chrisdembia:
Nick, if you've been able to fix (B) above, then I will "approve" this PR.

carmichaelong:
Thanks for the change. For now, I'll probably build the hopper/devices on each loop (and we should continue doing this for the GUI too), and if there's time, I can look into doing updates to the model. Once Chris's request is satisfied, feel free to merge.

nickbianco:
(B) is fixed.

chrisdembia:
Awesome stuff. I'll leave it up to you if you want to delete the tgcs2017_Nick branch.

PR #1627, opened 2017-03-30
Add thelen muscle to test cmc gait10dof18mus

aseth1:
Fixes issue #1140 (third check point) and replaces #1191 

### Brief summary of changes
Discovered that `testCMCGait10dof18musc` was not testing the `Thelen2003Muscle` model. The model itself uses `Millard2012EquilibriumMuscle` muscles so the rename has no effect. I ran into this while profiling some minor changes to path computations, primarily the allocation of temporary variables in `WrapEllipsoid` (which I left in since the change is minimal), and noticed Millard related functions being called.

1. The most significant change was to reverse the rename `Thelen2003Muscle` and that the tolerances when comparing to the standard had to be loosened to accommodate for the change.
2. The coordinates are still tracked tightly but the muscle activations were only comparable to within 10%. 
3. Took the opportunity to reduce the duration of the CMC simulation since updating the tolerances anyways

### Testing I've completed

All the tests passed locally.

Verification that nothing nefarious is being introduced.

### CHANGELOG.md (choose one)

- no need to update because no user facing or API changes

### DIFF
--- a/Applications/CMC/test/testCMCGait10dof18musc.cpp
+++ b/Applications/CMC/test/testCMCGait10dof18musc.cpp
@@ -40,16 +40,7 @@ int main() {
     try { testGait10dof18musc(); }
     catch (const std::exception& e) {
         cout << e.what() << endl;
-        failures.push_back("testGait10dof18musc_Millard");
-    }
-
-    // redo with the Thelen2003Muscle
-    Object::renameType("Millard2012EquilibriumMuscle", "Thelen2003Muscle");
-
-    try { testGait10dof18musc(); }
-    catch (const std::exception& e) {
-        cout << e.what() <<endl;
-        failures.push_back("testGait10dof18musc_Thelen");
+        failures.push_back("testGait10dof18musc");
     }
 
     if (!failures.empty()) {
@@ -62,6 +53,8 @@ int main() {
     return 0;
 }
 
+// Perform regression test with standard generated from Gait10dof18musc
+// example in OpenSim 3.2
 void testGait10dof18musc() {
     cout<<"\n******************************************************************" << endl;
     cout << "*                      testGait10dof18musc                       *" << endl;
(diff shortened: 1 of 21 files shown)


tkuchida:
@aseth1 Please see failure of testCMCGait10dof18musc on Appveyor (failure on 64-bit only).

Applications/CMC/test/testCMCGait10dof18musc.cpp
```diff
@@ -80,8 +81,12 @@ void testGait10dof18musc() {
 
     int nstates = standard->getColumnLabels().size() - 1;
 
-    // angles and speeds within .6 degrees .6 degs/s; activations within 1%
-    std::vector<double> rms_tols(nstates, 0.01);
+    // angles and speeds within 0.01 rads and 0.1 rad/s;
+    // and activations to within 10%
```
tkuchida: Seems okay to increase tolerance for activations from one arbitrary value (1%) to another arbitrary value (10%) for now. In the future, I recommend splitting these tests in twain:
1. Regression test: ensure the computed values are exactly as expected (to within some reasonable but small tolerance).
2. Logical test: ensure each muscle is on/off at the expected times.

aseth1:
I could not get x64 builds to fail locally on my desktop or laptop. I did notice that the *iliopsoas_l/activation* is close to 0.1 so I modified the final time to make the separation a bit bigger, hoping to thwart the failure on Appveyor. :crossed_fingers:

tkuchida:
> I modified the final time to make the separation a bit bigger, hoping to thwart the failure on Appveyor

Seems okay for now, but I think this is a clear signal that the test is logically flawed.

aseth1:
> Seems okay for now, but I think this is a clear signal that the test is logically flawed.

The test originated as a regression for an [example](http://simtk-confluence.stanford.edu:8080/display/OpenSim/Simulation-Based+Design+to+Reduce+Metabolic+Cost) but was taking too long so was shortened. Then we wanted to make sure it worked for the Thelen muscle as well for which there was not standard.

I can revert it back to a regression (with the Millard Muscle) and ignore the Thelen muscle (since the model is exclusively Millard muscles). I could also run the a shorter time window for both Millard and Thelen in 3.3 to create two regression tests. I agree either of these would be better than the arbitrary adulterations to get the test to pass. Preferences?

tkuchida:
I don't think testCMC*.cpp should need to worry about *which* muscle is in the model; testMuscles.cpp should be responsible for ensuring that all muscles are behaving properly. For now, I would remove the Thelen test. I suggest revisiting the organization and coverage of the test suite post-4.0. :fish:

aseth1:
> For now, I would remove the Thelen test.

I agree. I reinstated the regression test with restored tolerances and removed the Thelen test altogether. Having a test that just runs with a Thelen model does seem pointless given there are other CMC regression tests with Thelen muscles. These tests should catch specific issues due to changes in the `Thelen2003Muscle` with CMC (against previous solutions) including no longer running!

tkuchida:
Cool. LGTM :+1:

tkuchida:
@klshrinidhi Travis failures are unrelated timeouts :hourglass:

aseth1:
@klshrinidhi I reran test to verify timeouts were unrelated. Ready for your :+1: or 👎 on this PR

aseth1:
@klshrinidhi please review

PR #1628, opened 2017-03-30
[travis] Use sourceforge to obtain simbody binaries and to cache opensim-core binaries

chrisdembia:
This is now ready for review. This PR uses simbody binaries from sourceforge (uploaded by simbody's travis script) rather than from bintray (which has not been working for us for about a year). This PR is necessary for OpenSim to incorporate the latest changes from Simbody, without increasing our travis build times.

This PR also improves the security of how we upload OpenSim doxygen and binaries to sourceforge.

The uploaded binaries can be obtained from here: https://sourceforge.net/projects/myosin/files/opensim-core/


chrisdembia:
Ready for review. I tested this by making sure that binaries are indeed uploaded to myosin (see link above).

chrisdembia:
Since this is working, I may merge this myself after a few days.

chrisdembia:
Thank you @klshrinidhi

PR #1635, opened 2017-04-04
[TGCS] Hopper GUI setup files

nickbianco:
## Main things
- Can now save current configuration of InteractiveHopper GUI to a MAT file in current working directory
- Saved MAT files can be loaded to retrieve GUI settings from a previous run
- All saving and loading is handled by InteractiveHopperSettings (new file, original renamed, see below)
- A run counter now tracks number of runs in current instance of GUI (also sets a default file name for settings MAT file)

## Small stuff
- InteractiveHopperSettings is now InteractiveHopperParameters: we can use this file to keep the hard-coded values for the InteractiveHopper. Therefore, most of the tuning and tweaking for the example, other than new/removed features, can be done within this file.
- Small fixes to the GUI, added and cleaned up some comments


nickbianco:
I tried to make it such that users won't need to look into them for the workshop, but could easily modify them later if they wanted to. Ideally, no changes should need to made to InteractiveHopperSettings by anyone, so we think about moving that to live within InteractiveHopper.

nickbianco:
One thing I forgot to mention: these changes break some of the default muscle excitation and device control plotting feature (i.e. no controls are plotted if you click "Simulate" first before changing anything). 
There are some plotting features that need to be added as a least significant update to the GUI, so this can be fixed as part of a subsequent PR.

carmichaelong:
These changes look good. I don't think this changes anything with the optimization interface too. As long as I can use `BuildInteractiveHopperSolution` to get the same hopper as the GUI would, then everything should be fine. Looks like this PR does not affect that use.

carmichaelong:
I've merged this in. @nickbianco feel free to delete or leave the branch as necessary.

PR #1637, opened 2017-04-05
Fix Model::initSystem() from dirtying isObjectUpToDateWithProperties()

### LINKED ISSUE #1587: isObjectUpToDateWithProperties() should be true after calling initSystem()
See discussion in #1584

aseth1:
Fixes #1587 

### Brief summary of changes
Added a condition to `testComponents` to throw when `model.isObjectUpToDateWithProperties()` after `initSystem()`. This caused `testComponents` to fail.

`Model::initializeState` was updating the model's `ControllerSet` property for the purpose of constructing the correct header for resulting controls Storage files. This is unnecessary because `ControllerSet::constructStorage()` is called when actuators are set on the `ControllerSet` at the end of `Model::extendConnectToModel()`. The unnecessary update was removed and `testComponents` passed.

Also updated `Component::get/updSystem()` to throw `ComponentHasNoSystem` if `hasSystem()` is false. Update `Model::get/updMultibodySystem()` to use `Component` accessors. 

### Testing I've completed
Added a condition to `testComponents` to throw when `model.isObjectUpToDateWithProperties()` after `initSystem()`. All tests pass locally.

### CHANGELOG.md (choose one)
- no need to update because there are not changes to the API (just a bug fix)

### DIFF
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -1398,13 +1398,7 @@ void Component::extendRealizeAcceleration(const SimTK::State& s) const
 
 const SimTK::MultibodySystem& Component::getSystem() const
 {
-    if (!hasSystem()){
-        std::string msg = getConcreteClassName()+"::getSystem() ";
-        msg += getName() + " has no reference to a System.\n";
-        msg += "Make sure you added the Component to the Model and ";
-        msg += "called Model::initSystem(). ";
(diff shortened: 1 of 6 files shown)


chrisdembia:
I've updated the original post on this PR so that #1587 will be automatically closed when this is merged.

aseth1:
Awesome. Thanks for updating and the quick reviews @tkuchida and @chrisdembia! 👏

Issue #1622, opened 2017-03-28
GCVSplineSet is not available in bindings


chrisdembia:
To facilitate using InverseDynamicsSolver in MATLAB, it would be great to wrap GCVSplineSet, which has a nice constructor that takes a Storage file.

This came up with @clnsmith's project during the March workshop.

PR #1639, opened 2017-04-05
Wrap GCVSplineSet

klshrinidhi:
Addressing issue #1622.

### DIFF
--- a/Bindings/OpenSimHeaders_common.h
+++ b/Bindings/OpenSimHeaders_common.h
@@ -32,6 +32,7 @@
 #include <OpenSim/Common/MultiplierFunction.h>
 #include <OpenSim/Common/PolynomialFunction.h>
 #include <OpenSim/Common/GCVSpline.h>
+#include <OpenSim/Common/GCVSplineSet.h>
 #include <OpenSim/Common/Sine.h>
 #include <OpenSim/Common/SmoothSegmentedFunctionFactory.h>
 #include <OpenSim/Common/SmoothSegmentedFunction.h>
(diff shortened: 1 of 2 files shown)


chrisdembia:
It would be good to test the specific constructor that we want to have working in the bindings. Also, it would be great, as @aseth1 mentioned, to add a constructor that takes a TimeSeriesTable. For my research, I've written such a function:


```cpp
GCVSplineSet createGCVSplineSet(const TimeSeriesTable& table,
                                const std::vector<std::string>& labels = {},
                                int degree = 5,
                                double errorVariance = 0.0) {
    GCVSplineSet set;
    const auto& time = table.getIndependentColumn();
    auto labelsToUse = labels;
    if (labelsToUse.empty()) labelsToUse = table.getColumnLabels();
    for (const auto& label : labelsToUse) {
        const auto& column = table.getDependentColumn(label);
        set.adoptAndAppend(new GCVSpline(degree, column.size(), time.data(),
                                         &column[0], label, errorVariance));
    }
    return set;
}
```

klshrinidhi:
* I have added constructor to take a TimeSeriesTable as per your code snippet.
* I have added a small test to make sure the constructors are callable.

chrisdembia:
Looks great. Only one comment: could you move the new doxygen block to the header file?

klshrinidhi:
I have moved all the doxygen comments to header file to be consistent.

OpenSim/Common/GCVSplineSet.h
```diff
@@ -49,46 +49,113 @@ class Storage;
...
-    //--------------------------------------------------------------------------
-    // SET AND GET
-    //--------------------------------------------------------------------------
 public:
+    /**
+     * Get the function at a specified index.
+     *
+     * @param aIndex Index of the desired function:  0 <= aIndex < getSize().
+     * @return Function at index aIndex.  If aIndex is not value NULL is 
+     *         returned.
```
chrisdembia: Maybe edit this to "If aIndex is not a valid value..."

klshrinidhi: Done.

Issue #1618, opened 2017-03-27
Component::printInputInfo() should print "<no connectees>" if there are no connectees


chrisdembia:
Currently, the output of `Component::printInputInfo()` may look like the following:

```
Inputs for component controller of type [ToyPropMyoController] along with connectee names:
    [double] activation : 
```

Since `activation` has no connectees, no connectee names are listed. It'd be neat if we could print `<no connectees>` after `activation`.

EDIT: the code *does* try to print `no connectees`, but `getNumConnectees()` seems to be giving the wrong information. I think the issue occurs for single-value Inputs, which always have 1 connectee, even if it isn't specified.

tkuchida:
Won't [this line](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Component.cpp#L1534) print "no connectees" if `input->getNumConnectees() == 0`? Are you sure there isn't a connectee whose name is null?

chrisdembia:
Yeah I made a mistake; I edited the original post.

tkuchida:
> I think the issue occurs for single-value Inputs, which always have 1 connectee, even if it isn't specified.

Perhaps add a check for `isConnected()`?
```cpp
if ( input->getNumConnectees() == 0 ||
     (input->getNumConnecteed() == 1 && !input->isConnected()) )
```

chrisdembia:
That should work.

tkuchida:
Fixed by #1640.

PR #1640, opened 2017-04-05
Fix Component::printInputInfo()

klshrinidhi:
Addressing #1618.

### DIFF
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -1532,7 +1532,7 @@ void Component::printInputInfo() const {
                                  input->getName().length(), ' ')
                   << input->getName() << " : ";
         if (input->getNumConnectees() == 0 || 
-            (input->getNumConnectees() == 1 && !input->isConnected())) {
+            (input->getNumConnectees() == 1 && input->getConnecteeName().empty())) {
             std::cout << "no connectees" << std::endl;
         } else {
             for (unsigned i = 0; i < input->getNumConnectees(); ++i) {
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -1531,7 +1531,8 @@ void Component::printInputInfo() const {
                   << std::string(maxlenInputName -
                                  input->getName().length(), ' ')
                   << input->getName() << " : ";
-        if (input->getNumConnectees() == 0) {
+        if (input->getNumConnectees() == 0 || 
+            (input->getNumConnectees() == 1 && !input.isConnected())) {
             std::cout << "no connectees" << std::endl;
         } else {
             for (unsigned i = 0; i < input->getNumConnectees(); ++i) {
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -1532,7 +1532,7 @@ void Component::printInputInfo() const {
                                  input->getName().length(), ' ')
                   << input->getName() << " : ";
         if (input->getNumConnectees() == 0 || 
-            (input->getNumConnectees() == 1 && !input.isConnected())) {
+            (input->getNumConnectees() == 1 && !input->isConnected())) {
             std::cout << "no connectees" << std::endl;
         } else {
             for (unsigned i = 0; i < input->getNumConnectees(); ++i) {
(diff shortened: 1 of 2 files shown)


OpenSim/Common/Component.cpp
```diff
@@ -1531,7 +1531,8 @@ void Component::printInputInfo() const {
                   << std::string(maxlenInputName -
                                  input->getName().length(), ' ')
                   << input->getName() << " : ";
-        if (input->getNumConnectees() == 0) {
+        if (input->getNumConnectees() == 0 || 
+            (input->getNumConnectees() == 1 && !input.isConnected())) {
```
tkuchida: This logic assumes that `input->getConnecteeName(i)` is nonempty for all `i` if `input->getNumConnectees() > 1`. Is this assumption correct?

chrisdembia: If parsing an XML file, I think that assumption is correct, because the parser should trim whitespace. That is, 

```
<input_foo_connectee_name>a b       c</input_foo_connectee_name>
```

should be interpreted as the list `["a", "b", "c"]`.

However, if setting the connectee names programmatically through the API, the assumption might not hold. I think that would be a separate bug, though (`setConnecteeName()` should validate the provided name).

chrisdembia:
@klshrinidhi did you test this?

OpenSim/Common/Component.cpp
tkuchida: Do we need something like this:
```cpp
bool connectedConnecteeFound = false;
for (unsigned i = 0u; i < input->getNumConnectees(); ++i) {
    if (input.getConnectee(i).isConnected()) {
        std::cout << input->getConnecteeName(i) << " ";
        connectedConnecteeFound = true;
    }
}
if (!connectedConnecteeFound)
    std::cout << "no connectees" << std::endl;
```

klshrinidhi:
Output looks like the following:
```
 Inputs for component controller of type [PropMyoController] along with connectee names:
     [double] activation : no connectees
```

chrisdembia:
I think the desired behavior for `printInputInfo()` is to show *desired* connections, not established connections. That's because `printInputInfo()` might be useful for debugging why a desired connection is not occurring ("Why isn't my Geometry connecting to the Transform input I specified? Let me call geom.printInputInfo(). Oh, I see, I had a typo in the connectee_name."). Therefore, I don't think `printInputInfo()` should rely on `isConnected()`.

The original issue is that single-value inputs always have 1 connectee. We have decided that it's fine for single-value inputs to *not* have a connectee (that's how Geometry's Transform input works), but we cannot use `isConnected()` to query whether or not the single-value input has a desired connectee, since a precondition on this function is that `finalizeConnections()` has been called.

So I think a satisfactory fix for now would be to check if the connectee name is an empty string:

`getNumConnectees() == 1 && input->getConnecteeName().empty()`

As I mentioned earlier, it is still possible to have multiple desired connectees with empty connectee names (via the API, not via XML), but I think that is a separate issue that does not need to be addressed here.

> It so, quotes around the connectee name would disambiguate.

I agree with this, but for consistency, we should then update all `print*Info()` methods to put names in quotes, as they currently do not. My preference is to leave this for another PR.

I also slightly prefer seeing `<no-name>` (or `'<no-name>'`) over `''`, even though the logic to code this is a little more complicated.

tkuchida:
I agree with revisiting in the future. For now, let's just do something that makes sense given the current design and clearly describe what the method does in the doxygen.

klshrinidhi:
I have changed the if-condition and edited the documentation.

OpenSim/Common/Component.h
```diff
@@ -1493,7 +1493,7 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
     void printSocketInfo() const;
 
     /** List all the inputs of this component and whether or not they are 
-    connected. Also list the connectee names for inputs that are connected. */
+    connected. Also list the (desired) connectee names for the intputs.      */
```
tkuchida: intputs -> inputs

tkuchida: Also, the user may interpret this to mean that there's a desired name for a particular connectee.

klshrinidhi: Fixed.

PR #1641, opened 2017-04-05
Copy DataTable / TimeSeriesTable in Matlab / Java

klshrinidhi:
@jimmyDunne experienced a problem in Matlab where assigning one table (DataTable/TimeSeriesTable) to another variable only copies the handle -- shallow copy of the object. Editing the copy also changes the original in this case. This PR adds a clone() method to the tables to obtain a real (deep) copy of the original table.

### DIFF
--- a/Bindings/common.i
+++ b/Bindings/common.i
@@ -180,8 +180,8 @@ namespace OpenSim {
                                         const std::vector<std::string>&);
 %ignore OpenSim::DataTable_<double, double>::flatten;
 %extend OpenSim::DataTable_ {
-    OpenSim::DataTable_<ETX, ETY> clone() const {
-        return OpenSim::DataTable_<ETX, ETY>{*$self};
+    OpenSim::DataTable_<ETX, ETY>* clone() const {
+        return new OpenSim::DataTable_<ETX, ETY>{*$self};
     }
 }
 %extend OpenSim::DataTable_<double, double> {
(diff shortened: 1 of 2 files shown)


chrisdembia:
I think having this function is a good idea. Is there a reason not to add it in C++? That would be consistent with our other classes, and allow use of `SimTK::ClonePtr` with the tables. The current solution is nice because it does not have the potential of introducing a leak.

klshrinidhi:
There is a clone() on C++ side. The function returns an AbstractDataTable. I wanted these functions to return concrete type in Python/Java. So I replace them in the bindings.

chrisdembia:
Ah okay

jimmyDunne:
can confirm this works in Matlab in Java. clone() gives a deep copy that, when edited, doesn't affect the original version. 

Thanks!!

aymanhab:
@klshrinidhi OpenSim::Object has a Clone method that returns Concrete type, we don't mess with that in Bindings except to tell SWIG that the new instance uses new memory. Is there a reason that behavior can't be duplicated in DataTables instead of having different semantics for clone when invoked on tables? Thanks,

klshrinidhi:
As per discussion with @aymanhab, I have changed the implementation of clone() in bindings to return pointer to a dynamically allocated object instead of returning by value.

Bindings/common.i
```diff
@@ -215,6 +220,11 @@ namespace OpenSim {
 }
 
 %ignore OpenSim::TimeSeriesTable_::TimeSeriesTable_(TimeSeriesTable_ &&);
+%extend OpenSim::TimeSeriesTable_ {
+    OpenSim::TimeSeriesTable_<ETY> clone() const {
+        return OpenSim::TimeSeriesTable_<ETY>{*$self};
```
chrisdembia: Is there a reason that this is not also dynamically allocated?

klshrinidhi: Fixed it.

Issue #1650, opened 2017-04-12
Required format for printing external loads files can't be met (Matlab)


jimmyDunne:
When printing out forces from a c3d, you have to rely on `STOFileAdapter()`, which prints external loads in a [non-compliant .mot format](http://simtk-confluence.stanford.edu:8080/display/OpenSim/Motion+%28.mot%29+Files). 

```
%% import java libraries
import org.opensim.modeling.*

%% Use a c3dAdapter to turn read a c3d file
adapter = C3DFileAdapter();
forces = adapter.read(filepath).tables.get('forces');

%% flatten the Vec3 table to a table of doubles. 
postfix = StdVectorString(); postfix.add('_x');postfix.add('_y');postfix.add('_z');
forces_flattened = forces_rot.flatten(postfix);
% make a sto adapter and write the forces table to file.
STOFileAdapter().write(forces_flattened,[filepath '.mot']);
```

The output file header looks like this

```
CalibrationMatrices=Corners=DataRate=2000.000000
Origins=Types=events=DataType=double
version=2
endheader
time	f1_x	f1_y	f1_z	m1_x	m1_y	m1_z	p1_x	p1_y	p1_z	f2_x	f2_y	f2_z	m2_x	m2_y	m2_z	p2_x	p2_y	p2_z
0	1.472130060195923	17.71418762207031	-2.357758045196534	61.67040961382952	742.0819913723619	-98.77109934038099	827.2036135380047	3.6909246695582e-14	602.7737421316849	0	22.82806015014648	0.04444999992847303	2.273736754432321e-13	814.478227169007	1.585923503849422	318.8885687585697	3.419024461569678e-14	558.3690683632433
```

There are a few issues with this. (i) the header information is not compliant with standard .mot file format, and (ii) the order of colomns is also incorrect (we require f1,p1,f2,p2,...fn,pn, m1,m2,...mn).

I should be able to open the .mot in the (3.3) GUI to test/visualize, which I currently can not do.

klshrinidhi:
* Something is definitely wrong with writing the header. I will try fixing that.
* About column order, it may be better if I write a method for DataTable to rearrange columns internally. The order of columns from C3DFileAdapter seems logical -- F,1 M1, P1, F2, M2, P2, ..... , Fn, Mn, Pn. So one could rearrange the columns of the forces table coming out of C3DFileAdapter using something like:
[...]

Issue #1542, opened 2017-03-01
Discrepancy in result of getValueIsDefault()


klshrinidhi:
```c++
Model arm{"arm26.osim"};
arm.finalizeFromProperties();
// 'ground' -> 'frame_geometry' returns 1 when getValueIsDefault() called.
//  'ground' -> 'frame_geometry' -> 'socket_connectee_name' returns 0 when getValueIsDefault() called.
```
The property `frame_geometry` and its nested property `socket_connectee_name` return different results. Depending on result from top-level property may lead to problems.

This was discovered when using the function [here](https://github.com/opensim-org/opensim-core/blob/4adf5bd13af96124645c5883baec2e42763db83a/OpenSim/Tools/Test/testModelCopy.cpp#L40-L68) to copy properties of components of `arm` above.

PR #1652, opened 2017-04-13
Bug fix for STOFileAdapter AND add DataTable::reorderColumns()

klshrinidhi:
Fixes issue #1650 

* Bug fix -- Make sure to write new line when writing STO files if the metadata cannot be converted to string. This fixes the problem with header reported in #1650.
* Add method `DataTable::reorderColumns()` to allow reordering of columns of a DataTable in-place.

### DIFF
--- a/Bindings/Java/tests/TestTables.java
+++ b/Bindings/Java/tests/TestTables.java
@@ -58,14 +58,6 @@ class TestTables {
         System.out.println(table);
         // Clone table.
         DataTable tableClone = table.clone();
-        // Reorder columns of the table.
-        StdVectorInt newColumnIndices = new StdVectorInt();
-        newColumnIndices.add(3); newColumnIndices.add(1);
-        newColumnIndices.add(0); newColumnIndices.add(2);
-        table.reorderColumns(newColumnIndices);
-        newColumnIndices.set(0, 2); newColumnIndices.set(1, 1);
-        newColumnIndices.set(2, 3); newColumnIndices.set(3, 0);
-        table.reorderColumns(newColumnIndices);
         // Get independent column.
         StdVectorDouble indCol = table.getIndependentColumn();
         assert indCol.get(0) == 0.1 &&
@@ -114,6 +106,18 @@ class TestTables {
                table.getDependentColumn("3").get(1) == 40 &&
                table.getDependentColumn("3").get(2) == 40;
         System.out.println(table);
+        // Reorder columns of the table.
+        StdVectorInt newColumnIndices = new StdVectorInt();
+        newColumnIndices.add(3); newColumnIndices.add(1);
+        newColumnIndices.add(0); newColumnIndices.add(2);
+        table.reorderColumns(newColumnIndices);
+        assert table.getDependentColumnAtIndex(0).get(0) == 40 &&
+               table.getDependentColumnAtIndex(1).get(0) == 10 &&
+               table.getDependentColumnAtIndex(2).get(0) == 30 &&
+               table.getDependentColumnAtIndex(3).get(0) == 10;
+        newColumnIndices.set(0, 2); newColumnIndices.set(1, 1);
+        newColumnIndices.set(2, 3); newColumnIndices.set(3, 0);
+        table.reorderColumns(newColumnIndices);
         // Assert that clone was not edited due to above operations
         // on rows/columns.
         assert tableClone.getNumRows() == 3;
--- a/Bindings/Java/tests/TestTables.java
+++ b/Bindings/Java/tests/TestTables.java
@@ -58,6 +58,14 @@ class TestTables {
         System.out.println(table);
         // Clone table.
         DataTable tableClone = table.clone();
+        // Reorder columns of the table.
+        StdVectorInt newColumnIndices = new StdVectorInt();
+        newColumnIndices.add(3); newColumnIndices.add(1);
+        newColumnIndices.add(0); newColumnIndices.add(2);
+        table.reorderColumns(newColumnIndices);
+        newColumnIndices.set(0, 2); newColumnIndices.set(1, 1);
+        newColumnIndices.set(2, 3); newColumnIndices.set(3, 0);
+        table.reorderColumns(newColumnIndices);
         // Get independent column.
         StdVectorDouble indCol = table.getIndependentColumn();
         assert indCol.get(0) == 0.1 &&
--- a/Bindings/Python/tests/test_DataTable.py
+++ b/Bindings/Python/tests/test_DataTable.py
@@ -88,6 +88,9 @@ class TestDataTable(unittest.TestCase):
                 row2[2] == row[2] and
                 row2[3] == row[3])
         print table
+        # Reorder columns of the table.
+        table.reorderColumns([3, 1, 0, 2])
+        table.reorderColumns([2, 1, 3, 0])
         # Retrieve independent column.
         assert table.getIndependentColumn() == (0.1, 0.2, 0.3)
         # Retrieve dependent columns.
--- a/Bindings/Python/tests/test_DataTable.py
+++ b/Bindings/Python/tests/test_DataTable.py
@@ -90,6 +90,22 @@ class TestDataTable(unittest.TestCase):
         print table
         # Reorder columns of the table.
         table.reorderColumns([3, 1, 0, 2])
+        col0 = table.getDependentColumnAtIndex(0)
+        col1 = table.getDependentColumnAtIndex(1)
+        col2 = table.getDependentColumnAtIndex(2)
+        col3 = table.getDependentColumnAtIndex(3)
+        assert (col0[0] == 4  and
+                col0[1] == 8  and
+                col0[2] == 16 and
+                col1[0] == 2  and
+                col1[1] == 4  and
+                col1[2] == 8  and
+                col2[0] == 1  and
+                col2[1] == 2  and
+                col2[2] == 4  and
+                col3[0] == 3  and
+                col3[1] == 6  and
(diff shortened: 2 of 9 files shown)


Bindings/Java/tests/TestTables.java
```diff
@@ -58,6 +58,14 @@ public static void test_DataTable() {
         System.out.println(table);
         // Clone table.
         DataTable tableClone = table.clone();
+        // Reorder columns of the table.
```
tkuchida: It isn't clear/obvious how you're confirming that the column reordering is working properly. For example, if the intention is to scramble in L63–65 and unscramble in L66–68, then the test will still pass if `table.reorderColumns()` does nothing (e.g., if somehow, sometime in the future, the method ends up throwing a nonfatal exception that isn't caught).

Bindings/Python/tests/test_DataTable.py
```diff
@@ -88,6 +88,9 @@ def test_DataTable(self):
                 row2[2] == row[2] and
                 row2[3] == row[3])
         print table
+        # Reorder columns of the table.
+        table.reorderColumns([3, 1, 0, 2])
+        table.reorderColumns([2, 1, 3, 0])
```
tkuchida: Not a rigorous test, as it will still pass if `table.reorderColumns()` does nothing.

OpenSim/Common/DataTable.h
```diff
@@ -997,6 +998,82 @@ class DataTable_ : public AbstractDataTable {
         _indData[rowIndex] = value;
     }
 
+    /** Reorder columns of the DataTable by providing the new order of 
+    column indices. The new order is basically a permutation of column indices
+    0 to num-columns-1. For example:
+    \code
+    // Change order of last three columns only. Leave the first two columns 
+    // as is.
+    table.reorderColumns({0, 1, 5, 3, 4});
```
tkuchida: Your example code will throw a `ColumnIndexOutOfRange` exception. The argument should be `{0, 1, 4, 2, 3}`.

OpenSim/Common/DataTable.h
```diff
@@ -997,6 +998,82 @@ class DataTable_ : public AbstractDataTable {
         _indData[rowIndex] = value;
     }
 
+    /** Reorder columns of the DataTable by providing the new order of 
+    column indices. The new order is basically a permutation of column indices
```
tkuchida: How does the new order differ from a permutation of column indices?

OpenSim/Common/DataTable.h
```diff
@@ -997,6 +998,82 @@ class DataTable_ : public AbstractDataTable {
...
+            if(c != columnIndices[c]) {
+                _depData.updCol(c) = depDataCopy.col(columnIndices[c]);
+                setColumnLabel(c, columnLabelsCopy[columnIndices[c]]);
+            }
+        }
+
+    }
+    
+    /** Reorder columns of the DataTable by providing the new order of 
+    column labels. The new order is basically a permutation of existing
```
tkuchida: How does the new order differ from a permutation of existing column labels?

OpenSim/Common/DataTable.h
```diff
@@ -997,6 +998,82 @@ class DataTable_ : public AbstractDataTable {
...
+
+    }
+    
+    /** Reorder columns of the DataTable by providing the new order of 
+    column labels. The new order is basically a permutation of existing 
+    column labels. For example:
+    \code
+    // Change order of last three columns only. Leave the first two columns 
+    // as is. Here "col0", "col1", ... are existing column-labels.
+    table.reorderColumns({"col0", "col1", "col5", "col3", "col4"});
```
tkuchida: Should be `{"col0", "col1", "col4", "col2", "col3"}`.

OpenSim/Common/DataTable.h
tkuchida: Did you mean "basically" -> "simply"?

OpenSim/Common/DataTable.h
tkuchida: "basically" -> "simply"?

tkuchida:
Also, I don't think this PR fixes everything in #1650 (see Jen's comment from Dev Team meeting).

Bindings/Java/tests/TestTables.java
klshrinidhi: Fixed the tests.

Bindings/Python/tests/test_DataTable.py
klshrinidhi: Fixed.

OpenSim/Common/DataTable.h
klshrinidhi: I mean "simply". Fixed it.

OpenSim/Common/DataTable.h
klshrinidhi: Fixed.

OpenSim/Common/DataTable.h
klshrinidhi: I mean "simply". Fixed it.

OpenSim/Common/DataTable.h
klshrinidhi: Fixed.

tkuchida:
@klshrinidhi: The description of this PR states that it "Fixes issue #1650" but there are elements of #1650 that are ignored/not addressed. Are you planning a follow-up PR to address the remaining issues?

klshrinidhi:
I am working on addressing the comments in #1650. Will be back soon.

klshrinidhi:
I have added a function to Storage to read new (version 2) STO/MOT files.

chrisdembia:
Looks pretty good. Here are some comments:

- @klshrinidhi  could you add a note to the Storage constructor that it accepts version 2 storage files generated by the STOFileAdapter?
- @jimmyDunne @aseth1 is the logic surrounding "inDegrees" the desired behavior? I think the current behavior is that "inDegrees" is determined based on the "inDegrees" metadata, and that if the inDegrees metadata is not available, then inDegrees=false. This sounds fine to me, but this is the only place where I thought there could be an issue.
- The travis-ci/pr failure looks legitimate.

jimmyDunne:
@chrisdembia I don't think I have seen a package that wrote the angles in radians, so I would assume that the default would always degrees. IMHO, if no metadata is found, the default should probably be degrees (inDegrees = true).

chrisdembia:
Okay thanks for that info, @jimmyDunne.

klshrinidhi:
* I have set default value for `inDegrees` to true when reading newer files.
* I have added comment to constructor.

Latest CI failure is timeout related.

jenhicks:
@aseth1 can you review?

OpenSim/Common/C3DFileAdapter.h
```diff
@@ -44,10 +42,13 @@ class OSIMCOMMON_API C3DFileAdapter : public FileAdapter {
     C3DFileAdapter& operator=(C3DFileAdapter&&)      = default;
     ~C3DFileAdapter()                                = default;
 
+    explicit C3DFileAdapter(const bool& useSIMMColumnOrdering);
```
aseth1: I don't think a `C3DAdapter` or any `FileAdapter` other than the one that writes SIMMMotion (.mot) files should be aware of `useSIMMColumnOrdering`.  If the goal is to produce *.mot* files that can be read in by SIMM and older versions of OpenSim then the flag should be on `STOFileAdapter::write()` to write as a *.mot*. We do not want anything within OpenSim to be dependent on any ordering specified by another package.

klshrinidhi: Okay. Sounds good. Will get back with changes.

jimmyDunne:
Is there anything I can do help speed up this merge? The c3d conversion code is stuck until this merged.

tkuchida:
@jimmyDunne IIRC this PR didn't resolve #1650 entirely, but it was an improvement so I approved a few weeks ago. I think the objective is now for this PR to completely resolve #1650, which I believe requires two further changes:
- Column reordering should occur on write (not on read) and shouldn't affect the way OpenSim stores the data internally.
- A test case should be added to confirm that #1650 has been fixed.

aseth1:
@jimmyDunne Definitely you could help @klshrinidhi by providing a test case as described above.  A MATLAB script would work. @klshrinidhi could translate to a C++ test more easily and the script could serve as (or part of) and example. Thanks!

OpenSim/Common/C3DFileAdapter.h
aseth1: @klshrinidhi are you making progress on this? If you are stuck, lets discuss today at 11.

klshrinidhi: Adding the logic of reordering columns to STOFileAdapter::write() :
* Assumes that the table being written contains forces.
* Requires to logic to find the columns by their label (and not index like the current solution) for reordering. This will need the columns to be labeled in with suffixes assumed by the function or suffixes will have to be provided to the function as arguments. This will also need more error-checking to be correct -- for example, some columns may be missing, or not labeled correctly.

aseth1: It may make more sense and require fewer assumptions and checks to have a separate export to SIMMMotionFile and then check if the Table has forces, if so, then reorder for SIMM compatibility. In any case the output *.mot* file should have the appropriate SIMM header.

jenhicks:
Related #1542

aseth1:
Closing since #1843 more directly addresses the issue. However, keeping the branch alive since having the ability to reorder columns may come in handy for other purposes like meeting some external format requirement that users may have.

Issue #1656, opened 2017-04-13
InverseKinematicsTool leaves a stale KinematicsReporter in the Model post run


aymanhab:
InverseKinematicsTool  adds a KinematicsReporter for its internal working but doesn't remove it when done. This causes problems downstream since the reporter is created on the stack, once out of scope it's a 💣 waiting to explode.

tkuchida:
:boom: likely occurring in #1632 (see posts 15 and 16).

tkuchida:
Fixed by #1657

Issue #1632, opened 2017-04-02
exit code 139 (interrupted by signal 11: SIGSEGV) python


nikorose87:
Hi Chris, 
Unfortunately, I could do scaling in the python API because I've got some errors setting the markers of a model. A piece of the code where the error is presented is shown below:

```
scalesetup = motionHenryDir+"/Henryscaledef.xml"
Newmarkers = osim.MarkerSet(model,markerspath)
mystate = model.initSystem()
model.replaceMarkerSet(mystate,Newmarkers)
```
And I got the error of the title. So, I don't know if something in the API is not working well or If I'm doing something wrong. 

Thanks and regards.

nikorose87:
Hi again

I've got another problem that generates the same error. In fact, I'm replicating the 'RunTutorialThree.py' but with a personal case of study in python 2.7. However, I got stuck doing scaling, IK and RRA because I haven't found a function that loads the results to the model, so I think that something internally makes the API break.

Here is the piece of code:
```

model = osim.Model(motionHenryDir+'/Henry-Scaled.osim')
mystate = model.initSystem()
#mystate = model.initSystem()
# Set up the visualizer to show the model and simulation
model.setUseVisualizer(True)

# Create a zero value Vec3
zeroVec3 = osim.ArrayDouble.createVec3(0)
# set gravity 0
#model.setGravity(zeroVec3)
# Create the coordinate storage object from the input .sto file
coordinateSto = osim.Storage(motionHenryDir+'/HenryResults.mot')
#model.setAuthors('Nikolay Prieto PhD')

# Get coordinate set from model, and count the number of coordinates
modelCoorSet = model.getCoordinateSet()
nCoords = modelCoorSet.getSize()

## Inverse Kinematics
iktool = osim.InverseKinematicsTool(iksetupfile)
iktool.setModel(model)
iktool.setOutputMotionFileName(IKmodel)
iktool.run()

# Initialize model
mystate = model.initSystem()
# Add a marker set to the model

#newMarkers = osim.MarkerSet(model, Modelfolder + "/gait2354_Scale_MarkerSet.xml")
#tester = model.replaceMarkerSet()

manager = osim.Manager(model)
manager.setInitialTime(0)
manager.setFinalTime(3.0)
manager.integrate(mystate)

```
Thank you.

chrisdembia:
[...]

PR #1657, opened 2017-04-13
InverseKinematicsTool adds a KinematicsReporter for internal use, remove it when done. 

aymanhab:
This PR removes the KinemaicsReporter used internally by IKTool at the end of run so that running the tool has no side-effects on the model.

Fixes issue #1656, #1632 

### Brief summary of changes
Added a line to remove the reporter after usage is complete inside the run method

### Testing I've completed
This used to crash the GUI which needs to reuse the model after IK run, with the fix the crash is gone. 

### CHANGELOG.md (choose one)

- no need to update because no API changes


tkuchida:
I'll test with the MWE in #1632…

aymanhab:
Thanks @chrisdembia and @tkuchida merging.

Issue #1634, opened 2017-04-03
WrapObjects are not attached to correct frame for visualization


aymanhab:
WrapObject's generateDecorations method places and computes the geometry relative to Ground. While this works if recreating the scene from scratch for every frame (As used in API visualizer) this construct wrong scene graph that results in wrap objects not moving with the Frames/Bodies they are attached to.
![opensim_snapshot 16](https://cloud.githubusercontent.com/assets/3972208/24633716/bdc7f640-187e-11e7-935c-dfe871187791.jpg)

jenhicks:
Need to fix bug, update documentation, update test case to catch this bug. Review other objects that use generateDecorations to make sure they don't create the same issue.

PR #1658, opened 2017-04-13
Modify generateDecorations implemenetation for wrap objects to attach to correct body/Frame

aymanhab:
In particular:

- Geometry created by WrapGeometry is now indicated as fixed.
- DecorativeGeometry is attached to the Frame to which it is connected using MobilizedBodyId from the Frame.

Fixes issue #1634

### Brief summary of changes
No API changes, Refactored implementation of generateDecorations of WrapObject subclasses

### Testing I've completed
GUI move sliders and wrapobjects move as expected
Updatied testVisualization to account for a WrapObject not in Ground. previous test had WrapObjects all in Ground to avoid numerical issues with comparing transforms.

### Looking for feedback on.
Documentation, test case.

### CHANGELOG.md (choose one)

- no need to update because no API has changed.


aymanhab:
Test failure is unrelated time-out. Ready for review.

aseth1:
The changes look reasonable to me for fixing the geometry to a `MobilizedBody`. However, seeing the same block of code repeated four times just screams to be implemented by the base `WrapObject`.

OpenSim/Simulation/Wrap/WrapCylinder.cpp
```diff
@@ -820,13 +820,20 @@ void WrapCylinder::generateDecorations(bool fixed, const ModelDisplayHints& hint
         // assumptions between DecorativeCylinder aligned with y  and
         // WrapCylinder aligned with z
         ztoy.updR().setRotationFromAngleAboutX(SimTK_PI / 2);
-        const SimTK::Transform& X_GB = getFrame().getTransformInGround(state);
-        SimTK::Transform X_GW = X_GB*getTransform()*ztoy;
+        // B: base Frame (Body or Ground)
+        // F: PhysicalFrame that this ContactGeometry is connected to
```
chrisdembia: ContactGeometry -> WrapGeometry.

aymanhab:
Thanks @aseth1 and @chrisdembia for the review. My understanding is that we're not changing the API so I only changed the implmentation. If that is not the case then I agree on refactoring the methods and moving common code to base class, can you please confirm the policy? Thanks.

chrisdembia:
Oh, right. My opinion is that adding a protected function in this situation does not violate the policy.

aseth1:
I agree with it being a protected utility for `WrapObejcts` with no change to the API. Thanks!

aymanhab:
Thanks @chrisdembia :+1: for the quick turnaround. Test failure due to timeout.

aseth1:
Confirmed failures are unrelated timeouts. Merging.

aymanhab:
Thank you @aseth1

PR #1660, opened 2017-04-14
change file name

jimmyDunne:
### Brief summary of changes
Updated the pendulum example so it doesn't rely on a model file. 

### Testing I've completed
Tested the (output) .trc file. Opens in 3.3 GUI to view. 

🏆


Bindings/Java/Matlab/examples/simpleOptimizerObjectiveFunction.m
```diff
@@ -4,35 +4,36 @@
...
+        afl.setFiberLength(s, .1);
+end
+
+%% Make sure the muscles states are in equilibrium
+osimModel.equilibrateMuscles(s);
+
+bicShort = osimModel.getMuscles().get('BICshort');
+
+%% Compute moment arm of BICshort, flip sign since the optimizer tries to minimize rather than maximize
+f = bicShort.computeMomentArm(s, elbowFlexCoord);
```
chrisdembia: I think this should be negated.

PR #1661, opened 2017-04-14
[TGCS] Use EvaluateHopper with GUI

carmichaelong:
@chrisdembia @nickbianco This PR does the following:
1) Uses EvaluateHopper within InteractiveHopper
2) Changes the mass of the hopper so that the visualization with the answers of RunHopper and RunHopperWithDevices is roughly the same as previously
3) Some clean up of unused variables

### DIFF
--- a/Bindings/Java/Matlab/Hopper_Device/EvaluateHopper.m
+++ b/Bindings/Java/Matlab/Hopper_Device/EvaluateHopper.m
@@ -24,7 +24,7 @@
 % Perform a hop with the provided model and evaluate the performance of the
 % hop.
 
-function [score, peakHeight, finalHeight] = EvaluateHopper(hopper, print)
+function [score, peakHeight, finalHeight, heightStruct] = EvaluateHopper(hopper, visualize, print)
         
 
 import org.opensim.modeling.*;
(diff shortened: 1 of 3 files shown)


Bindings/Java/Matlab/Hopper_Device/EvaluateHopper.m
```diff
@@ -24,7 +24,7 @@
 % Perform a hop with the provided model and evaluate the performance of the
 % hop.
 
-function [score, peakHeight, finalHeight] = EvaluateHopper(hopper, print)
+function [score, peakHeight, finalHeight, heightStruct] = EvaluateHopper(hopper, visualize, print)
```
chrisdembia: I think we should eventually simplify this so that it returns only the necessary quantities (`score` seems unnecessary). Fine for now, though.

carmichaelong: I agree. I left in just in case we want to add any other penalties later. In fact, if we just keep the only metric as `peakHeight`, we can just return `peakHeight` and `heightStruct` (and I can just pull the `finalHeight` out for my own use if needed). I think this can be deferred to a later commit.

nickbianco:
Looks good to me. Working with the Millard muscle only is definitely cleaner.

carmichaelong:
Merging. Thanks @chrisdembia and @nickbianco!

PR #1664, opened 2017-04-17
Copying Simbody and BTK into OpenSim install; unifying RPATH on UNIX.

### LINKED ISSUE #462: link error when building in debug
The build system has come a long way! Although building OpenSim in release mode is smooth,  building in debug (a x64 version using VS2013 Pro running on Windows 8.1) is causing a bunch of link errors ('error LNK 1104: cannot open file ...') for these libraries:

osimCommon_d.lib
osimLepton_d.lib
osimSimulation_d.lib
osimTools_d.lib

The odd thing is that these libraries exist in both the correct build (opensim-core-build\Debug) and install directories (opensim-core\bin), and the install bin folder is on the path (and I've checked this from the command line). I'm not sure what's causing these link errors - any ideas?

### LINKED ISSUE #693: Linking Simbody and OpenSim in Debug mode cmake issue
I'm new to the community here. Thanks for being so active! 

After compiling Simbody 3.6 and opensim-core from Github, both in Debug mode I tried to compile the simple arm and one muscle example on the main Github page using the SampleCMakeLists.txt as a template CMakeLists.txt file.

I got the following error message after calling `make` on the example

```
/usr/bin/ld: cannot find -lSimTKcommon
/usr/bin/ld: cannot find -lSimTKmath
/usr/bin/ld: cannot find -lSimTKsimbody
```

When I looked in <opensim-install>/cmake/OpenSimTargets-debug.cmake

I found 

```
set_target_properties(osimCommon PROPERTIES
  IMPORTED_LINK_INTERFACE_LIBRARIES_DEBUG "SimTKcommon;SimTKmath;SimTKsimbody"
  IMPORTED_LOCATION_DEBUG "${_IMPORT_PREFIX}/lib/libosimCommon_d.so"
  IMPORTED_SONAME_DEBUG "libosimCommon_d.so"
  )
```

which looked suspicious since my SimTKcommon, SimTKmath, and SimTKsimbody files all have the trailing _d from Debug mode.

After modifying this .cmake file to this in a couple places

```
"SimTKcommon_d;SimTKmath_d;SimTKsimbody_d"
```

My problems were solved and the example program compiled and ran successfully. I'm sure there's a real way to fix this or maybe I compiled something wrong, but I thought I'd post my findings since it took a while to track down. Thanks.

chrisdembia:
Fixes #693 
Fixes #462

### Brief summary of changes

- Update Simbody commit for superbuild.
- Copy BTK libraries on macOS and Linux (not only on Windows).
- Link BTK "privately" so that clients need not link to BTK themselves (BTK is only needed at run time not link time).
- Rearrange how Simbody is installed into OpenSim's installation. Install SimbodyConfig.cmake and related files.
- OpenSimConfig.cmake finds Simbody internally, and does so differently depending on OPENSIM_COPY_DEPENDENCIES.
- SampleCMakeLists.txt: no longer tell developers to use `link_directories(${OpenSim_LIB_DIR})`; it's no longer necessary.
- Document installation layout on different platforms, and the CMake flags to use when building the distribution.
- Use relative RPATHs on Linux as well as on macOS; and unify the CMake logic for handling RPATH.
- Update the python bindings RPATH to include the relative path from the python package to the OpenSim lib dir.

### Testing I've completed

- Built and ran a client project on macOS (with OPENSIM_COPY_DEPENDENCIES=on and off), Linux, and on Windows (in both Debug and Release; to ensure that #693 is fixed).
- On macOS and Linux, I ensured that python and MATLAB bindings work without having to set any environment variables, both with OPENSIM_COPY_DEPENDENCIES=on and off.
  - I did *not* test MATLAB/Python bindings on Windows, as we did not change anything that would affect this (we tested MATLAB/Python bindings on macOS and Linux to see if changes to RPATH worked).

- On Windows, SimTK DLLs are copied into two places: `bin/` and `sdk/Simbody/bin`. The files are about 6 MB total. The copies in `sdk/Simbody/bin` are necessary because SimbodyConfig.cmake complains if it can't the DLLs where it expects them.

### CHANGELOG.md (choose one)

- Updated


chrisdembia:
@aymanhab This is ready for review now.

Bindings/Java/OpenSimJNI/CMakeLists.txt
```diff
@@ -114,15 +114,18 @@ set_source_files_properties(
         PROPERTIES COMPILE_FLAGS "${COMPILE_FLAGS}")
 
 if(${OPENSIM_USE_INSTALL_RPATH})
```
aymanhab: How would that behave on windows? Are we guaranteed OPENSIM_USE_INSTALL_RPATH is false in that case?

CMakeLists.txt
```diff
@@ -89,17 +89,11 @@ from the internet, we turn this option OFF." ON)
...
-this should be set to ON to avoid using absolute paths for RPATHs to Simbody
-libraries." ON)
+option(OPENSIM_COPY_DEPENDENCIES "Copy Simbody and BTK into the
+OpenSim installation. This should be set to ON when making a relocatable
+distribution, and should be set to OFF when packaging for Homebrew, Debian,
+etc. On Linux and macOS: we use relative RPATHs when ON, and absolute RPATHs
+when OFF." ON)
 
 option(OPENSIM_PYTHON_STANDALONE "Make the Python package standalone, meaning
 the OpenSim (and Simbody and BTK) shared libraries it depends on are copied
```
aymanhab: copied into ?? I assume a folder under the python installation. Do we need to mention the PYTHON_PATH option ? IS there a preference one way or the other on the python side?

CMakeLists.txt
```diff
@@ -490,9 +486,9 @@ endif()
 
 
 ## RPATH 
-# If it is necessary to not put RPATHS in the installed binaries, set
+# If it is necessary to not put RPATHs in the installed binaries, set
 # set CMAKE_SKIP_INSTALL_RPATH to OFF.
```
aymanhab: set set -> set

CMakeLists.txt
```diff
@@ -635,93 +620,112 @@ elseif(${PLATFORM_NAME} STREQUAL Mac OR ${PLATFORM_NAME} STREQUAL Linux)
...
+OpenSimCopyDependencyDLLsForWin(Simbody "${Simbody_BIN_DIR}")
+if(WIN32)
+    add_dependencies(Copy_Simbody_DLLs Simbody_CONFIG_check)
+endif()
+
+
 # Directories for Simbody headers and libraries for building.
 # -----------------------------------------------------------
 # TODO no longer necessary; Simbody's library targets provide the 
 # include directories.
```
aymanhab: Is this comment true? and if no longer necessary why not remove it?

CMakeLists.txt
```diff
@@ -635,93 +620,112 @@ elseif(${PLATFORM_NAME} STREQUAL Mac OR ${PLATFORM_NAME} STREQUAL Linux)
...
+    
+        # lib (and cmake, pkgconfig)
+        file(RELATIVE_PATH install_simbodylibdir 
+            "${Simbody_ROOT_DIR}/" "${Simbody_LIB_DIR}/")
+        file(GLOB_RECURSE simbody_desired_lib_files
+            RELATIVE "${Simbody_LIB_DIR}"
+            "${Simbody_LIB_DIR}/*SimTK*" "${Simbody_LIB_DIR}/*Simbody*.cmake"
+            "${Simbody_LIB_DIR}/*SampleCMakeLists.txt*"
+            "${Simbody_LIB_DIR}/*simbody.pc")
+        foreach(simbody_lib_file ${simbody_desired_lib_files})
```
aymanhab: Not sure what this simbody.pc  or what happens to lapack, blas and other dlls/libs that Simbody needs. Can only be verified during testing AFAIK.

OpenSim/Common/CMakeLists.txt
```diff
@@ -13,7 +13,8 @@ endif()
 OpenSimAddLibrary(
     KIT Common
     AUTHORS "Clay_Anderson-Ayman_Habib-Peter_Loan"
-    LINKLIBS ${Simbody_LIBRARIES} ${BTK_LIBRARIES}
+    # Clients of osimCommon need not link to BTK.
+    LINKLIBS PUBLIC ${Simbody_LIBRARIES} PRIVATE ${BTK_LIBRARIES}
```
aymanhab: Are these options PRIVATE/PUBLIC supported by the minimum cmake version that we require already?

dependencies/CMakeLists.txt
```diff
@@ -147,7 +147,7 @@ AddDependency(NAME       BTK
 
 AddDependency(NAME       simbody
               URL        https://github.com/simbody/simbody.git
-              TAG        9f2123212dadcedcd680a8d8b3832d4fe40e1f8b
+              TAG        247f0cdb2916317115e50186c2f3ee0a177e02b3
```
aymanhab: Why did we need to upgrade Simbody?

CMakeLists.txt
chrisdembia: This comment is true, so long as you are using a version of Simbody from within the last month or so. I thought about removing this line, but I did not want to unnecessarily inconvenience the dev team and force everyone to rebuild a newer Simbody.

OpenSim/Common/CMakeLists.txt
chrisdembia: Yes; we require 3.2, which includes PUBLIC/PRIVATE. Here are the 3.1 docs: https://cmake.org/cmake/help/v3.1/command/target_link_libraries.html?highlight=target_link_libraries

dependencies/CMakeLists.txt
chrisdembia: This commit of Simbody includes the feature where the search path for simbody-visualizer includes the relative path from libSimTKsimbody. This PR is the first to copy over simbody-visualizer into OpenSim on UNIX, and so including this commit allowed testing that the simbody-visualizer is indeed found properly (I tested this on mac).

Bindings/Java/OpenSimJNI/CMakeLists.txt
chrisdembia: OPENSIM_USE_INSTALL_RPATH is undefined (thereby false) on Windows, always.

chrisdembia: Here is the one place where we set this variable: https://github.com/opensim-org/opensim-core/blob/cmake_install_dependencies_BTK_RPATH/CMakeLists.txt#L520

CMakeLists.txt
chrisdembia: simbody.pc is a pkg-config file, mostly for use on UNIX (alternative to using SimbodyConfig.cmake). I just looked at the [contents of this file](https://github.com/simbody/simbody/blob/master/cmake/pkgconfig/simbody.pc.in), and I just realized that it does not work if the simbody installation has been relocated. So I will remove the copying of this file into the OpenSim installation.

We expect UNIX users to have their own lapack, blas, pthread, etc. On UNIX, we only need the 3 simbody libraries.

I think my tests were sufficiently thorough (I changed the name of the install directory to assess portability). However, I did not test using the installation on other machines. 

I'm happy to address any issues that arise.

CMakeLists.txt
chrisdembia: I think setting PYTHONPATH is very atypical; the assumption is that you get your python packages from a python package manager (pip, conda, etc.), not from some third party website (e.g., simtk.org). So even telling users to cd <opensim-install>/lib/python2.7/site-packages/ and sudo python setup.py install is atypical. The best thing would be to distribute the python package separately via pypi, conda (homebrew, debian, etc.). Even then, whether we copy the libosim*.(so|dylib) files is not uniform: for pypi, we would include the shared libraries, on conda, we would probably have a "opensim-core" or "opensim" package that contains the shared libraries, and then a separate "opensim-python". I introduced this variable to help handle these different scenarios. However, perhaps we should hide it for most users.

My suggestions:

Document the PYTHONPATH option in our README.md.
Mark OPENSIM_PYTHON_STANDALONE as advanced.

chrisdembia:
@aymanhab Thank you for the review. I am thinking making the following changes to address your comments:

- Document the PYTHONPATH option in our README.md.
- Mark OPENSIM_PYTHON_STANDALONE as advanced.
- Remove the copying of `simbody.pc` into the OpenSim installation.
- set set -> set

Let me know if you have changes to this plan.

chrisdembia:
@aymanhab I have made the agreed-upon changes, except for `Document the PYTHONPATH option in our README.md.`. I learned that PYTHONPATH will not work in general, because a previously-installed opensim python package will get preference over an opensim package in PYTHONPATH: http://blog.olgabotvinnik.com/blog/2014/03/03/2014-03-03-pythonpath-is-a-liar-site-py-and-easy-install-pth-tell/

I would prefer to defer addressing the issue of the python package.

aymanhab:
Thanks much @chrisdembia LGTM :+1:

chrisdembia:
Thank you @aymanhab!

Issue #890, opened 2016-03-24
Create overview "story" to document the API/library architecture.


chrisdembia:
This could be 1-4 pages on Doxygen as "related pages"?

tkuchida:
Ideally, also include pre- and post-conditions for methods like `finalizeConnections()`, `extendFinalizeFromProperties()`, and `initSystem()`.

PR #1665, opened 2017-04-17
Fix #1584: throw exception when attempting to access the System after finalizeFromProperties

### LINKED ISSUE #1584: Cryptic exception message if finalizeFromProperties() called after initSystem()
To reproduce:
```cpp
#include <OpenSim/OpenSim.h>
using namespace SimTK;
using namespace OpenSim;

int main() {
    try {
        auto model = Model("gait10dof18musc.osim");
        auto state = model.initSystem();
        //model.finalizeFromProperties(); //frog
        RungeKuttaMersonIntegrator integrator(model.getSystem());
        Manager manager(model, integrator);
        state.setTime(0.);
        manager.integrate(state, 1.);
    } catch (const std::exception& ex) {
        std::cout << ex.what() << std::endl;
        return 1;
    }
    return 0;
}
```
When :frog: is uncommented, the following exception message is reported:
<pre>
SimTK Exception thrown at Integrator.cpp:431:
  Integrator initialization failed apparently because:
Component::isCacheVariableValid: ERR- name not found.
 for component 'geometrypath_' of type GeometryPath
</pre>
Although this example may seem contrived, it was discovered in a "real world" situation in testWrapping.cpp (`simulateModelWithMusclesNoViz()`), where `Model::printBasicInfo()` was being called after `Model::initSystem()` (and `printBasicInfo()` must now be preceded by `finalizeFromProperties()` due to #1582).

There are at least two possible resolutions:
1. Make `Model::finalizeFromProperties()` "harmless" if the model hasn't changed. In the example above, `initSystem()` had just been called, so it doesn't seem like `finalizeFromProperties()` should do much of anything. If that's not possible, then…
2. Catch the exception from Integrator and provide a more user-friendly message that indicates what went wrong and how the problem might be fixed.

aseth1:
Fixes issue #1584 

In #1584 the `Model` was treating the `System` as valid after `finalizeFromProperties()` was called leading to a cryptic message downstream and failure when attempting to access the cache of an invalid state.

### Brief summary of changes
Key change was to reset the `Component`'s reference to the system in `finalizeFromProperties()` which was already calling its own `reset()`

### Testing I've completed
Originally used code snippet from #1584 to reproduce the failure and corresponding message. 
Then generalized into its own test `testModelInterface` was added to verify the correct behavior when the system or derivatives are accessed on a `Model` but no longer has a `System`.

- is the test case adequate

### CHANGELOG.md (choose one)
- no need to update because this is a bug fix with no interface changes

### DIFF
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -149,7 +149,7 @@ public:
                          const std::string& func,
                          const Object& obj) :
         Exception(file, line, func, obj) {
-        std::string msg = "Component has not been added to a System.\n";
+        std::string msg = "Component has no underlying System.\n";
         msg += "You must call initSystem() on the top-level Component ";
         msg += "(i.e. Model) first.";
         addMessage(msg);
@@ -716,6 +716,7 @@ public:
     */
     template <class C = Component>
     C& updComponent(const std::string& name) {
+        clearObjectIsUpToDateWithProperties();
         return *const_cast<C*>(&(this->template getComponent<C>(name)));
     }
 
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -422,9 +422,9 @@ public:
     void finalizeFromProperties();
 
     /** Satisfy the Component's connections specified by its Sockets and Inputs.
-        Locate Component and their Outputs to satisfy the connections in an
-        aggregate Component, which is the root of a tree of Components, which
-        for example, forms a complete model. */
+        Locate Components and their Outputs to satisfy the connections in an
+        aggregate Component (e.g. Model), which is the root of a tree of
+        Components. */
     void finalizeConnections(Component& root);
 
     /** Disconnect/clear this Component from its aggregate component. Empties 
@@ -586,6 +586,7 @@ public:
         static_assert(std::is_base_of<Component, T>::value,
                 "Template argument must be Component or a derived class.");
         initComponentTreeTraversal(*this);
+        clearObjectIsUpToDateWithProperties();
         return ComponentList<T>(*this);
     }
 
@@ -708,7 +709,12 @@ public:
         return getComponent<Component>(pathname);
     }
 
-    /** Get a writable reference to a subcomponent.
+    /** Get a writable reference to a subcomponent. Use this method
+    * to edit the properties and connections of the subcomponent.
+    * Note: the method will mark this Component as not up-to-date with
+    * its properties and will require finalizeFromProperties() to be
+    * invoked directly or indirectly (by finalizeConnections() or 
+    * Model::initSystem())
     * @param name       the pathname of the Component of interest
     * @return Component the component of interest
     * @throws ComponentNotFoundOnSpecifiedPath if no component exists
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -711,7 +711,7 @@ public:
 
     /** Get a writable reference to a subcomponent. Use this method
     * to edit the properties and connections of the subcomponent.
-    * Note: the method will mark this Component as not up-to-date with
+    * Note: the method will mark this Component as out-of-date with
     * its properties and will require finalizeFromProperties() to be
     * invoked directly or indirectly (by finalizeConnections() or 
     * Model::initSystem())
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -149,9 +149,9 @@ public:
                          const std::string& func,
                          const Object& obj) :
         Exception(file, line, func, obj) {
-        std::string msg = "This Component has not been added to a System.\n";
-        msg += "You must call initSystem on the top-level Component ";
-        msg += "(i.e., Model) first.";
+        std::string msg = "Component has not been added to a System.\n";
+        msg += "You must call initSystem() on the top-level Component ";
+        msg += "(i.e. Model) first.";
         addMessage(msg);
     }
 };
@@ -414,13 +414,17 @@ public:
     /** Define a Component's internal data members and structure according to
         its properties. This includes its subcomponents as part of the component
         ownership tree and identifies its owner (if present) in the tree.
-        finalizeFromProperties propagates to all of the component's
-        subcomponents prior to invoking the virtual extendFinalizeFromProperties()
-        on itself.*/
+        finalizeFromProperties propagates to all of the component's subcomponents
+        prior to invoking the virtual extendFinalizeFromProperties() on itself.
+        Note, finalizeFromProperties() disassociates the Component from the
+        System it was added to (result of addToSystem()), which results from
+        calling  Model::initSystem(), for example.*/
     void finalizeFromProperties();
 
-    /** Connect this Component to its aggregate component, which is the root
-        of a tree of components.*/
+    /** Satisfy the Component's connections specified by its Sockets and Inputs.
+        Locate Component and their Outputs to satisfy the connections in an
+        aggregate Component, which is the root of a tree of Components, which
+        for example, forms a complete model. */
     void finalizeConnections(Component& root);
 
     /** Disconnect/clear this Component from its aggregate component. Empties 
@@ -2556,20 +2560,12 @@ private:
     SimTK::DefaultSystemSubsystem& updDefaultSubsystem() const
         {   return updSystem().updDefaultSubsystem(); }
 
-    void clearStateAllocations() {
-        _namedModelingOptionInfo.clear();
-        _namedStateVariableInfo.clear();
-        _namedDiscreteVariableInfo.clear();
-        _namedCacheVariableInfo.clear();    
-    }
+    // Clear all modeling options, continuous and discrete state variables,
+    // and cache variable allocated by this Component
+    void clearStateAllocations();
 
     // Reset by clearing underlying system indices.
-    void reset() {
-        _simTKcomponentIndex.invalidate();
-        clearStateAllocations();
-
-        resetSubcomponentOrder();
-    }
+    void reset();
 
 protected:
     //Derived Components must create concrete StateVariables to expose their state 
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -416,9 +416,10 @@ public:
         ownership tree and identifies its owner (if present) in the tree.
         finalizeFromProperties propagates to all of the component's subcomponents
         prior to invoking the virtual extendFinalizeFromProperties() on itself.
-        Note, finalizeFromProperties() disassociates the Component from the
-        System it was added to (result of addToSystem()), which results from
-        calling  Model::initSystem(), for example.*/
+        Note that if the Component has already been added to a System (result of
+        addToSystem(); e.g., Model::initSystem()) when finalizeFromProperties()
+        is called, then finalizeFromProperties() disassociates the component from
+        that System.*/
     void finalizeFromProperties();
 
     /** Satisfy the Component's connections specified by its Sockets and Inputs.
--- a/OpenSim/Simulation/Test/testModelInterface.cpp
+++ b/OpenSim/Simulation/Test/testModelInterface.cpp
@@ -34,7 +34,6 @@ int main() {
     LoadOpenSimLibrary("osimActuators");
 
     try {
-        // the specific model used here is not important
         Model model("arm26.osim");
         model.finalizeFromProperties();
 
@@ -67,11 +66,11 @@ int main() {
         ASSERT_THROW(ComponentHasNoSystem, model.realizeDynamics(dummy_state));
         ASSERT_THROW(ComponentHasNoSystem, 
                      model.computeStateVariableDerivatives(dummy_state));
-        // should be able to create a Manager either
+        // should not be able to create a Manager either
         ASSERT_THROW(ComponentHasNoSystem,
                      Manager manager(model));
 
-        // get a valid state
+        // get a valid System and corresponding state
         SimTK::State state = model.initSystem();
         Manager manager(model);
         state.setTime(0.0);
@@ -86,12 +85,16 @@ int main() {
         // given a stale but compatible state
         ASSERT_THROW(ComponentHasNoSystem, manager.integrate(state, 1.));
 
+        // once again, get a valid System and corresponding state
         state = model.initSystem();
-        //re-establish all the connections in the model
+        
+        // Test for the effects of calling finalizeConnections() on the model
+        // after initSystem() has been called.
+        // In this case, there are no changes to the connections to be finalized.
         model.finalizeConnections(model);
 
         // verify that finalizeConnections() does not wipe out the underlying 
-        // System iff there are no changes to the dependencies
+        // System when there are no changes to the connections
         auto& sys = model.getSystem();
 
         auto elbowInHumerus = new PhysicalOffsetFrame("elbow_in_humerus",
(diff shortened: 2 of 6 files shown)


OpenSim/Common/Component.h
```diff
@@ -414,13 +414,17 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
+        prior to invoking the virtual extendFinalizeFromProperties() on itself.
+        Note, finalizeFromProperties() disassociates the Component from the
+        System it was added to (result of addToSystem()), which results from
+        calling  Model::initSystem(), for example.*/
     void finalizeFromProperties();
 
-    /** Connect this Component to its aggregate component, which is the root
-        of a tree of components.*/
+    /** Satisfy the Component's connections specified by its Sockets and Inputs.
+        Locate Component and their Outputs to satisfy the connections in an
```
tkuchida: Component -> Components

OpenSim/Common/Component.h
```diff
@@ -414,13 +414,17 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
+        aggregate Component, which is the root of a tree of Components, which
```
tkuchida: I think the second ", which…" does more harm than good.

OpenSim/Common/Component.h
```diff
@@ -712,6 +716,7 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
     */
     template <class C = Component>
     C& updComponent(const std::string& name) {
+        clearObjectIsUpToDateWithProperties();
```
tkuchida: Does this mean that the simple act of fetching a writable reference to a Component invalidates its up-to-dateness? Maybe that's okay—though users will need to be very careful to use `getComponent()` rather than `updComponent()`: there will be situations where both will work but `updComponent()` will take longer.

aseth1: Yes, the `upd` methods should be reserved for Component (and Model) editing and not for simulation time querying. I think this makes `updComponent` (and I also made `udComponentList` comply) consistent to write access of `Set`s that are properties. For example, if you access bodies, forces, constraint via their Sets (e.g, `Model::updBodySet()`) these have the same effect of *dirtying* the up-to-date flag.

tkuchida: > if you access bodies, forces, constraint via their Sets (e.g, `Model::updBodySet()`) these have the same effect of *dirtying* the up-to-date flag.

Ah, ok. I didn't know that.


A separate potential issue:
IIRC `get...()` and `upd...()` can both work in scripting. Is it then possible for an assumption about the up-to-date flag to be incorrect in scripting?

aseth1: Yes, absolutely- that is a big hole (issue) right now.

aseth1:
> CI failures are due to timeouts only ⌛️

The debug failures were legit. Subsystem handles were having their implementations deleted by the System. The solution was to reset the unique_ptrs to the subsystem handles first (as @chrisdembia did in #585 when the System is recreated) before the System reset.

tkuchida:
> The debug failures were legit.

Crud; I must have just looked at [this one](https://travis-ci.org/opensim-org/opensim-core/jobs/222970381). The dangers of :scream: :speech_balloon: :wolf:

OpenSim/Common/Component.h
tkuchida: Opened #1666 to track.

OpenSim/Common/Component.h
```diff
@@ -704,14 +709,20 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
         return getComponent<Component>(pathname);
     }
 
-    /** Get a writable reference to a subcomponent.
+    /** Get a writable reference to a subcomponent. Use this method
+    * to edit the properties and connections of the subcomponent.
+    * Note: the method will mark this Component as not up-to-date with
```
tkuchida: "not up-to-date" -> "out-of-date"?

OpenSim/Simulation/Test/testModelInterface.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+using namespace OpenSim;
+using namespace std;
+
+
+int main() {
+    LoadOpenSimLibrary("osimActuators");
+
+    try {
+        // the specific model used here is not important
```
tkuchida: ...but it must contain at least 1 muscle and at least 2 bodies

OpenSim/Simulation/Test/testModelInterface.cpp
```diff
@@ -0,0 +1,133 @@
...
+        Model model("arm26.osim");
+        model.finalizeFromProperties();
```
tkuchida: Following what you've done below, perhaps precede L39 with
``` cpp
ASSERT(!model.isObjectUpToDateWithProperties());
```

OpenSim/Simulation/Test/testModelInterface.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+        // model should be up-to-date with its properties
+        ASSERT(model.isObjectUpToDateWithProperties());
+
+        // make an edit to a Component contained in the model's Set
+        model.updMuscles()[0].upd_min_control() = 0.02;
```
tkuchida: According to the conversation above, it sounds like
``` cpp
model.updMuscles();
```
should be sufficient here to dirty the up-to-date flag, and would be a stricter test.

OpenSim/Simulation/Test/testModelInterface.cpp
```diff
@@ -0,0 +1,133 @@
...
+        ASSERT(!model.isObjectUpToDateWithProperties());
+
+        model.finalizeFromProperties();
+        ASSERT(model.isObjectUpToDateWithProperties());
+
+        // make an edit to a Component contained in another Set
+        model.updBodySet()[1].upd_mass() = 0.05;
```
tkuchida: Again, just `model.updBodySet();`?

OpenSim/Simulation/Test/testModelInterface.cpp
```diff
@@ -0,0 +1,133 @@
...
+        ASSERT(!model.isObjectUpToDateWithProperties());
+
+        model.finalizeFromProperties();
+        ASSERT(model.isObjectUpToDateWithProperties());
+
+        // make an edit through model's ComponentList access
+        for (auto& body : model.updComponentList<Body>()) {
+            body.upd_mass_center() = SimTK::Vec3(0);
```
tkuchida: Perhaps just `break;` at L61 would be sufficient?

OpenSim/Simulation/Test/testModelInterface.cpp
```diff
@@ -0,0 +1,133 @@
...
+        }
+        ASSERT(!model.isObjectUpToDateWithProperties());
+
+        SimTK::State dummy_state;
+
+        ASSERT_THROW(ComponentHasNoSystem, model.getSystem());
+        ASSERT_THROW(ComponentHasNoSystem, model.realizeDynamics(dummy_state));
+        ASSERT_THROW(ComponentHasNoSystem, 
+                     model.computeStateVariableDerivatives(dummy_state));
+        // should be able to create a Manager either
```
tkuchida: should -> should not

OpenSim/Simulation/Test/testModelInterface.cpp
```diff
@@ -0,0 +1,133 @@
...
+        // verify that finalizeFromProperties() wipes out the underlying System
+        ASSERT_THROW(ComponentHasNoSystem, model.getSystem());
+
+        // should not be able to "trick" the manager into integrating a model
+        // given a stale but compatible state
+        ASSERT_THROW(ComponentHasNoSystem, manager.integrate(state, 1.));
+
+        state = model.initSystem();
+        //re-establish all the connections in the model
+        model.finalizeConnections(model);
```
tkuchida: Why is `model.finalizeConnections(model);` required here? Won't `model.initSystem();` on L89 already have "re-establish[ed] all the connections in the model"?

OpenSim/Simulation/Test/testModelInterface.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+        // verify that finalizeConnections() does not wipe out the underlying 
+        // System iff there are no changes to the dependencies
```
tkuchida: This logic seems to be the opposite of the logic used to dirty the up-to-date flag. Whereas just the act of touching a writable reference dirties the flag, the logic here appears to be that you can touch the connections without invalidating the downstream product (System). Seems inconsistent when wearing :tophat: (user's hat). Totally fine to address some other time (e.g., see #1668).

OpenSim/Simulation/Test/testModelInterface.cpp
tkuchida: …and one of the bodies must be named "r_humerus" (see L98). I would delete this comment.

OpenSim/Simulation/Test/testModelInterface.cpp
```diff
@@ -0,0 +1,133 @@
...
+        state = model.initSystem();
+
+        // now update the elbow and connect to the new frame
+        Joint& elbow = model.updComponent<Joint>("r_elbow");
+        elbow.connectSocket_parent_frame(*elbowInHumerus);
+
+        // satisfy the new connections in the model
+        model.finalizeConnections(model);
+
+        // now changing the connections should invalidate the System
```
tkuchida: changing -> finalizing. The connections _weren't_ changed the first time, which is precisely why the System remained intact.

OpenSim/Simulation/Test/testModelInterface.cpp
aseth1: It is not required in the same way calling `finalizeFromProperties()` after `initSystem` is not required. I will update the comment.

OpenSim/Simulation/Test/testModelInterface.cpp
aseth1: No connection was touched. If a Component is updated in anyway (to change its connection) then the System should be invalidated (that is tested below).

OpenSim/Simulation/Test/testModelInterface.cpp
aseth1: You are right that updating the model's Set will dirty the flag, but I thought the case of editing a property was more typical.

tkuchida: > I thought the case of editing a property was more typical.

I don't disagree, but if we want to ensure that the flag is dirtied just by calling `model.updMuscles();` (which may not be true in the future if the design ever changes), then I think it's safer to use the stricter test.

OpenSim/Simulation/Test/testModelInterface.cpp
aseth1: done

aseth1:
@tkuchida let me know if the recent changes addressed your comments.

OpenSim/Simulation/Test/testModelInterface.cpp
tkuchida: This comment was neither addressed nor replied to. Is there another test somewhere else that confirms `!model.isObjectUpToDateWithProperties()` immediately after loading a Model, or was the suggestion ignored because it was deemed unnecessary to perform this check?

tkuchida:
> @tkuchida let me know if the recent changes addressed your comments.

One comment was ignored. There are a few possible explanations for this but I don't have enough information to know which is the case.

OpenSim/Simulation/Test/testModelInterface.cpp
aseth1: A newly loaded model is likely to be up-to-date with its properties (and is in this case) if there haven't been any edits.  The job of `finalizeFromProperties` is to initialize internal data structures from properties, and it is necessary to call it after loading.

tkuchida: I must not be explaining my thinking clearly. The `finalizeFromProperties()` method must be called on L39 (after loading the model) because some condition that is required by the API is not met. Is it worthwhile to add an `ASSERT` before L39 to check that the condition is not met and another `ASSERT` _after_ L39 to check that the condition is met after calling `finalizeFromProperties()`? It seems like this test case is intended to check such conditions. Is there another test case elsewhere that is already testing the pre- and post-conditions of `finalizeFromProperties()`?

OpenSim/Simulation/Test/testModelInterface.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+int main() {
+    LoadOpenSimLibrary("osimActuators");
+
+    try {
+        // the specific model used here is not important
+        Model model("arm26.osim");
+        model.finalizeFromProperties();
+
+        // model should be up-to-date with its properties
```
tkuchida: Could explain why. I interpreted this comment to mean that the model should _now_ be up-to-date with its properties—as in, now that `finalizeFromProperties()` has been called. Seems like a reasonable assumption, given the name of the method.

aseth1: That is right. After `finalizeFromProperties()` the model must be up-to-date with its properties and have all its internals be consistent with its properties. Would you like me to update the comment?

tkuchida: :confused: You replied to my comment on L39 with an apparently contradictory statement. It's probably easiest to sort this out verbally :cow::speech_balloon:

aseth1: I don't see the contradiction. We cannot assume that the model is *out-of-date* with properties before we call `finalizeFromProperties()` but they must be *up-to-date* after and that is being tested.

tkuchida: What is the pre-condition for `finalizeFromProperties()`? My suggestion is to test the pre-condition for `finalizeFromProperties()` before it's called on L39. This suggestion is to maintain consistency with several of the other tests in this file, which are of the form
- `ASSERT` to test pre-condition for :turtle:
- call :turtle:
- `ASSERT` to test post-condition for :turtle:

tkuchida: Taking a step back: I think you and Chris are the only people who understand how the Component API has been designed. I see this test case as an opportunity to communicate the design to future developers—but the same can be accomplished by addressing #890 so I think this PR is :white_check_mark:.

aseth1: OK, I can add that the ComponentList access (as in counting sub components) must throw prior to calling `finalizeFroProperties()`

OpenSim/Simulation/Test/testModelInterface.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+        model.addComponent(elbowInHumerus);
+
+        // establish the new offset frame as part of the model but not
+        // used by any joints, constraints or forces
+        state = model.initSystem();
+
+        // update the elbow Joint and connect its socket to the new frame
+        Joint& elbow = model.updComponent<Joint>("r_elbow");
+        elbow.connectSocket_parent_frame(*elbowInHumerus);
```
chrisdembia: Consider documenting that this call requires recreating the system. Places where this could be added:

- https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/ComponentSocket.h#L912 and similar functions in other macros.
- https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/ComponentSocket.h#L142 and similar `connect()` functions in derived classes.

Would be good to add a similar check for `connectInput_()`; at the very least, we should add a TODO for that.

aseth1: ATM neither `connectSocket_` nor `connectInput_` themselves invalidated the System. It is the fact that both required a writable `Component` reference and `updComponent`, `updComponentList` or any `upd` of Sets that are properties causes the `Component` to be *out-of-date*.

chrisdembia:
Thanks for editing that comment; feel free to merge.

aseth1:
Thank-you @chrisdembia and @tkuchida . I created a feature request issue to track our discussions about using a call-back on `Object` upon property editing to *dirty* the owning `Model`.

Issue #1572, opened 2017-03-12
Finding the simbody-visualizer after distributing OpenSim


chrisdembia:
On Windows, the opensim-core installation contains simbody-visualizer in the bin folder (if the CMake variable OPENSIM_COPY_SIMBODY is ON). On macOS and Linux, however, simbody-visualizer is not copied into OpenSim's installation (regardless of the value of OPENSIM_COPY_SIMBODY). Therefore, the current macOS installation would not allow users to use the simbody-visualizer.

Even if the simbody-visualizer executable is present, [we somewhat rely on the OPENSIM_HOME environment variable to find simbody-visualizer](
https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/ModelVisualizer.cpp#L236
). As we have discussed, we no longer want to rely on setting this environment variable, especially on macOS. 

On Windows, simbody-visualizer.exe is on the Windows PATH after the OpenSim GUI is installed (since the GUI adds the OpenSim bin dir to the PATH), but I don't think that the SimTK::Visualizer searches the Windows PATH for simbody-visualizer.

Fixing this issue is relevant for the TGCS hopper workshop. Users could set the OPENSIM_HOME environment variable if necessary.

Currently, I don't have any ideas for automatically finding simbody-visualizer with a distributed copy of OpenSim without using environment variables.

chrisdembia:
I just chatted with @klshrinidhi about this. We can use `dladdr()` on Linux and macOS: http://stackoverflow.com/questions/1681060/library-path-when-dynamically-loaded. We can also use a manifest file that specifies the locations of libraries/executables, and tell the MATLAB/python bindings where to find the manifest file in our installation (similar to our OpenSim_buildinfo.txt).

We can work on this after @klshrinidhi finishes the project of copying Simbody and BTK into the OpenSim installation; right now, simbody-visualizer isn't copied into OpenSim's installation.

PR #1669, opened 2017-04-20
Add components of PATH to search path for visualizer.

klshrinidhi:
Fixes issue #1572

### Brief summary of changes
Add the paths in `PATH` to the search path for visualizer.

### DIFF
--- a/OpenSim/Simulation/Model/ModelVisualizer.cpp
+++ b/OpenSim/Simulation/Model/ModelVisualizer.cpp
@@ -232,7 +232,8 @@ void ModelVisualizer::createVisualizer() {
     // for the SimbodyVisualizer executable. The search will go as 
     // follows: first look in the same directory as the currently-
     // executing executable; then look in the $OPENSIM_HOME/bin 
-    // directory, then look in various default Simbody places.
+    // directory, then look at all the paths in the environment
+    // variable PATH, then look in various default Simbody places.
     Array_<String> searchPath;
     if (SimTK::Pathname::environmentVariableExists("OPENSIM_HOME")) {
         searchPath.push_back( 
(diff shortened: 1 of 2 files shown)


klshrinidhi:
Tested on Windows:
* Install OpenSim dependencies to directory `opensim_dependencies_install`.
* Install OpenSim to directory `opensim_install`.
* Rename Simbody install directory `opensim_dependencies_install\simbody` to something else.
* Add directory `opensim_install\bin` to `PATH`.
* Go to `opensim_install\sdk\python` and start python3.4.
* Run the following:
 ```
import os
import sys
sys.path.append(os.path.join(os.getcwd(), 'opensim'))
import opensim
m = opensim.Model()
m.setUseVisualizer(True)
m.initSystem()
```
The shows the visualizer window, meaning the path to the visualizer binary was found.

chrisdembia:
Awesome. Could you add doxygen documentation that describes this new behavior? Then I think this can be merged.

tkuchida:
Please also note on #1684 the 4.0-related changes that we will need to make to the ["Scripting in Python" Confluence page](http://simtk-confluence.stanford.edu:8080/display/OpenSim/Scripting+in+Python) (e.g., Python version number and the `sys.path.append` stuff). :snake:

klshrinidhi:
Added documentation to class description.

chrisdembia:
> I've added an item to #1684 re. the 4.0-related changes that we will need to make to the "Scripting in Python" Confluence page (e.g., Python version number and the sys.path.append stuff). 🐍

Thanks @tkuchida. It seems like the python bindings work with python3, which is great (I hadn't known this before); we should definitely document this. In general, the python bindings should not require any *specific* version of python (that is, python 3.4 and 3.6 should both work). So the documentation should be that the python bindings can be built from source for both python 2 and python 3 (most likely, the binary release will only be for python 2), but I think we should not need to say which specific version of python 2 or 3 is necessary (or perhaps we could say "2.7 or greater"). 

The `sys.path` call should only be necessary because @klshrinidhi chose not to install the python package system-wide; I don't think we should encourage users to use this mechanism. They should instead run `python setup.py install` from the `sdk/python` directory; [our instructions](http://simtk-confluence.stanford.edu:8080/display/OpenSim/Scripting+in+Python) already mention this.

tkuchida:
Thanks @chrisdembia. I've added your comments to #1684.

PR #1671, opened 2017-04-20
[TGCS] Fix controllers and other tweaks

carmichaelong:
@nickbianco @chrisdembia Main points of this PR:
* Rename the active/passive devices to fix problem with `PrescribedController` finding the first actuator with the correct name
* Increase the mass of the hopper to get the desired result for muscle alone (that it isn't strong enough)
* Some tweaks to parameters (mostly increasing device strengths). Depending on time, there may be follow-up PRs with further tweaks before group testing, but I think this is stable enough for now.
* Various clean up


OpenSim/ExampleComponents/HopperDevice.h
```diff
@@ -91,9 +91,9 @@ class OSIMEXAMPLECOMPONENTS_API HopperDevice : public ModelComponent {
 protected:
     // Change the color of the device's path as its tension changes.
     void extendRealizeDynamics(const SimTK::State& s) const override {
-        const auto& actuator = getComponent<PathActuator>("cableAtoB");
-        double level = fmin(1., getTension(s) / actuator.get_optimal_force());
-        actuator.getGeometryPath().setColor(s, SimTK::Vec3(0.1, level, 0.1));
```
chrisdembia: Could you add a TODO explaining why this is commented out?

Bindings/Java/Matlab/Hopper_Device/BuildDevice.m
```diff
@@ -123,21 +124,15 @@
...
+        name = 'cableAtoBpassive';
         springDissipation = 0.1; 
         relaxationTau = 5;
-        
-        if passivePatellaWrap
-            spring0 = 0.7166;
-        else
-            spring0 = 0.6329;
-        end
+        spring0 = 0;
```
chrisdembia: Does this mean that the spring is really stretched even in the default pose?

carmichaelong: Actually, this variable represents the `initialStretch` of the spring, so a setting of 0 makes it unstretched in both cases initially (and before it was stretched). Could rename to `stretch0` or `initialStretch` to make it clearer.

chrisdembia: Ahh that makes sense. Thanks for explaining.

PR #1674, opened 2017-04-20
[doxygen] Controller: actuators must have unique names.

chrisdembia:
In developing the MATLAB hopper example, we ran into a bug wherein Controllers will not be able to control multiple actuators with the same name. I talked with @aseth1 about how we should fix this, and we decided to use list sockets, when they are available. Since adding that feature will take time, I thought it would be good to at least document the current unexpected behavior.

### DIFF
--- a/OpenSim/Simulation/Control/Controller.h
+++ b/OpenSim/Simulation/Control/Controller.h
@@ -44,6 +44,13 @@ class Actuator;
  * The defining method of a Controller is its computeControls() method.
  * @see computeControls()
  *
+ * @note Version 4.0 introduced the ability to have multiple components
+ * with the same name as long as the components have unique absolute paths
+ * (i.e., they are located in different components). However, Controllers 
+ * will currently not work properly if you have multiple actuators with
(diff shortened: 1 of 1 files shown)


chrisdembia:
Thank you @tkuchida

OpenSim/Simulation/Control/Controller.h
```diff
@@ -44,6 +44,13 @@ class Actuator;
  * The defining method of a Controller is its computeControls() method.
  * @see computeControls()
  *
+ * @note Version 4.0 introduced the ability to have multiple components
+ * with the same name as long as the components have unique absolute paths
+ * (i.e., they are located in different components). However, Controllers 
+ * will currently not work properly if you have multiple actuators with
```
aseth1: Could say that `Controller` is not using the `Socket` mechanism to locate and connect to dependent Actuators, so for the time-being it doesn't support having multiple actuators with the same name.

chrisdembia:
@aseth1 Ready for review again. I used your suggested comment.

chrisdembia:
Thank you.

PR #1680, opened 2017-04-25
[cmake] Replace CMake include_directories(); update APIExamples

### LINKED ISSUE #1663: [cmake] Set include directories on OpenSim library targets
Right now, we expect client projects to use `include_directories("${OpenSim_INCLUDE_DIR}")` in their CMakeLists.txt. With modern CMake, this should no longer be necessary. Instead, we should be using `target_include_directories()` so that when a user does `target_link_libraries()`, the correct include directories are specified automatically.

See https://github.com/simbody/simbody/pull/552.

chrisdembia:
Fixes #1663.

### Brief summary of changes

- Removes the need to use `include_directories()` for targets that depend on osim libraries. In modern CMake, the include directories are propagated via `target_link_libraries()`.
- Change install location of Lepton headers (the `lepton` folder and the `Lepton.h` header are now within the OpenSim directory (e.g., `include/OpenSim/lepton/`). This is a more correct place to install Lepton headers on UNIX than `include/Vendors/...`.
- Cleaned up the CMakeLists.txt for all APIExamples. I made a change that @klshrinidhi made a while ago (but at the time, I was not comfortable requiring CMake 3.2).

### Testing I've completed

- Compiled (and ran, if possible) all APIExamples in the resulting installation directory. Some examples would not compile, but this was unrelated (#1676).

- Is it okay to require C++ API users to have CMake 3.2? 

### CHANGELOG.md (choose one)

- no need to update because...CHANGELOG already mentions CMake improvements.


chrisdembia:
The `appveyor/branch` build is failing for unrelated reasons, and this failure can be ignored.

jenhicks:
Chris thinks this is well tested. Just needs a review, if time permits.

tkuchida:
> I've been using 3.3 on Windows and 3.5 on Mac with no issues for awhile.

In case it's relevant, just yesterday I built from scratch (without bindings) using v3.8.2 on Win 8.1 and v3.4.2 on Win 7 without any issues.

chrisdembia:
Thanks @aseth1 and @tkuchida. By rule 5, I am able to merge this on my own with 1 review. I will wait a few more days to see if others have feedback, and then I will merge this.

chrisdembia:
Merging by rule 5.

Issue #1670, opened 2017-04-20
Manipulation of GeometryPath to insert/delete/changeType of PathPoints is broken


aymanhab:
OpenSimContext.addPathPoint() throws exception
java.lang.RuntimeException: PathPoint::getSystem() muscle1-P2 has no reference to a System.
Make sure you added the Component to the Model and called Model::initSystem(). 

OpenSimContext.deletePathPoint() crashes 

OpenSimContext.replacePathPoint(..)  throws exception
java.lang.RuntimeException: Failed to connect Socket 'coordinate' of type Coordinate (details: Model'leg6dof9musc'::findComponent() cannot find a nameless subcomponent.).
	In object 'default' of type ConditionalPathPoint.
	In file D:\src\opensim-core\OpenSim\Common\Component.cpp:236
	In function 'finalizeConnections'
	at org.opensim.modeling.opensimActuatorsAnalysesToolsJNI.OpenSimContext_replacePathPoint(Native Method)

aseth1:
Please supply some test cases and I will investigate. I assumed `testContext` was testing functions used by the GUI. It looks like `testContext` is not sufficient.

aymanhab:
Created branch fix_path_manipulation_from_OpenSimContext to exercise PathPoint deletion that should fail testContext test case. Will add remaining calls after this one is fixed.

aymanhab:
Thanks @aseth1 PR #1681 includes the test and fails expectedly.

PR #1681, opened 2017-04-25
Fix manipulation of PathPoint(s) invoked from OpenSimContext for Path editing

aymanhab:
So far deletion only is tested, rest will come later in this PR but investigation into this can start now and will clarify what to change in 4.0

Fixes issue #1670

### Brief summary of changes
In the past, changes to GeometryPath didn't require recreating the system under the model but rather to re-realize the state. Now that PathPoints are Components and are part of the system, edits by insertion, deletion or typechange require recreating the system.

### Testing I've completed
testContext has been made stronger to make the calls done by GUI path editing. This guarantees the calls don't crash or throw anymore.
Testing addition, removal and/or type change from GUI works with this PR in place, though MovingPathPoints have not been thoroughly  tested.

The change of type is adhoc and doesn't cover all the combination of pairs of types of PathPoints, Need more exercise from GUI.
 
### CHANGELOG.md (choose one)

- no need to update because this is internal bug-fix. 

### DIFF
--- a/Bindings/Java/OpenSimJNI/OpenSimContext.cpp
+++ b/Bindings/Java/OpenSimJNI/OpenSimContext.cpp
@@ -264,7 +264,7 @@ void OpenSimContext::addPathPoint(GeometryPath& p, int menuChoice, PhysicalFrame
 
 bool OpenSimContext::deletePathPoint(GeometryPath& p, int menuChoice) {
     bool ret = p.deletePathPoint(*_configState, menuChoice );
-  _configState->invalidateAll(SimTK::Stage::Position);
+  _configState->invalidateAll(SimTK::Stage::Position); // This maynot be enough anymore in 4.0
   _model->getMultibodySystem().realize(*_configState, SimTK::Stage::Position);
   return ret;
 }
--- a/Bindings/Java/OpenSimJNI/OpenSimContext.cpp
+++ b/Bindings/Java/OpenSimJNI/OpenSimContext.cpp
@@ -250,9 +250,10 @@ void OpenSimContext::addPathPoint(GeometryPath& p, int menuChoice, PhysicalFrame
 }
 
 bool OpenSimContext::deletePathPoint(GeometryPath& p, int menuChoice) {
-    bool ret = p.deletePathPoint(*_configState, menuChoice );
-    recreateSystemKeepStage();
-    return ret;
+    bool deletedSuccessfully = p.deletePathPoint(*_configState, menuChoice );
+    if (deletedSuccessfully) 
+        recreateSystemKeepStage();
+    return deletedSuccessfully;
 }
 
 bool OpenSimContext::isActivePathPoint(AbstractPathPoint& mp) {
--- a/Bindings/Java/OpenSimJNI/Test/testContext.cpp
+++ b/Bindings/Java/OpenSimJNI/Test/testContext.cpp
@@ -191,7 +191,7 @@ int main()
     // Exercise PathPoint operations used to edit Path in GUI
     PathPointSet& pathPoints = dTRIlong->updGeometryPath().updPathPointSet();
     AbstractPathPoint& savePoint = pathPoints.get("TRIlong-P2");
-    const PhysicalFrame& saveBody = savePoint.getBody();
+    const PhysicalFrame& saveBody = savePoint.getParentFrame();
     AbstractPathPoint* clonedPoint = savePoint.clone();
     // Delete second PathPoint from TRIlong muscle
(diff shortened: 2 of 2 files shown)


Bindings/Java/OpenSimJNI/OpenSimContext.cpp
```diff
@@ -264,7 +264,7 @@ void OpenSimContext::addPathPoint(GeometryPath& p, int menuChoice, PhysicalFrame
 
 bool OpenSimContext::deletePathPoint(GeometryPath& p, int menuChoice) {
     bool ret = p.deletePathPoint(*_configState, menuChoice );
-  _configState->invalidateAll(SimTK::Stage::Position);
+  _configState->invalidateAll(SimTK::Stage::Position); // This maynot be enough anymore in 4.0
```
aseth1: Deleting, adding, replacing, etc the `PathPoint` changes the model. The System should no longer be valid, in general. Look at your code for `replacePathPoint` that recreates the System after the edit.

aymanhab:
Ready for review.

Bindings/Java/OpenSimJNI/OpenSimContext.cpp
```diff
@@ -243,30 +233,26 @@ void OpenSimContext::setLocation(PathPoint& mp, int i, double d) {
...
+    p.updateGeometry(*_configState);
     return;
 }
 
 bool OpenSimContext::deletePathPoint(GeometryPath& p, int menuChoice) {
     bool ret = p.deletePathPoint(*_configState, menuChoice );
-  _configState->invalidateAll(SimTK::Stage::Position);
-  _model->getMultibodySystem().realize(*_configState, SimTK::Stage::Position);
-  return ret;
+    recreateSystemKeepStage();
```
tkuchida: Not sure I understand the logic here: why is `recreateSystemKeepStage()` called even if the PathPoint can't be deleted? Wouldn't the following be more efficient?
``` cpp
if (!p.canDeletePathPoint(menuChoice))
    return false;

const bool ret = p.deletePathPoint(*_configState, menuChoice);
recreateSystemKeepStage();
return ret;
```

aymanhab: Nice catch @tkuchida, it was more of blanket code transformation to replace the invalidate/realize call pair with recreateSystemKeepStage but indeed that is not necessary in this case :+1: Will make the change, thank you.

aymanhab:
Build failure is an unrelated  timeout. Thanks @tkuchida Any feedback @aseth1 ? Thanks

aymanhab:
@aseth1 any feedback to report, this is pretty straightforward code transformation to reflect that any changes to wiring or properties require recreation of system rather than invalidating stage, or if you can take a look @chrisdembia that would be great. Thanks.

Bindings/Java/OpenSimJNI/Test/testContext.cpp
```diff
@@ -187,6 +187,41 @@ int main()
...
+    assert(pathPoints.getSize() == origSize);
+    std::string pathAfterReinsertionInXML = dTRIlong->updGeometryPath().dump();
+    std::cout << pathAfterReinsertionInXML << endl;
+
+    // Test changing type to ConditionalPathPoint
+    ConditionalPathPoint* newPoint = new ConditionalPathPoint();
+    AbstractPathPoint& oldPoint = pathPoints.get(2);
+    newPoint->setCoordinate(model->getCoordinateSet().get(0));
+    newPoint->setParentFrame(oldPoint.getParentFrame());
+    context->replacePathPoint(dTRIlong->updGeometryPath(), oldPoint, *newPoint);
```
aseth1: The requirement of `replacePathPoint()` takes a reference to the oldPoint is a bit troubling. This pretty much guarantees you have a stale reference lying around. Your `delete` method takes an index, and you get the point by index above, so would it not be safer here to also take the index to be replaced?

aseth1:
The changes LGTM to me. I only have one issue above, but it was not created by this PR. It could be addressed secondarily if not convenient to change herein.

Bindings/Java/OpenSimJNI/Test/testContext.cpp
aymanhab: Thanks for the review @aseth1 
Good point regarding references (actually most OpenSim objects live as references on scripting & GUI side) I'm trying to keep this PR though limited to change in the implementation without change to the API.  I'll open one general one regarding the use of references.

aymanhab:
issue #1701 opened to address feedback on use of references.

Issue #1675, opened 2017-04-21
Skyscraper in muscle force plot with Analyze Tool


tkuchida:
### To reproduce
- download and unzip [skyscraper40.zip](https://drive.google.com/a/stanford.edu/file/d/0B1BKpngQdIZIVWRwbUFYcjE5QUU/view?usp=sharing)
- run `analyze -S Setup_Analyze_isokinetic.xml` at the command line
- in the resulting Tug_of_War_Millard_Iso_ForceReporter_forces.sto file, plot "LeftMuscle" vs. "time"

![skyscraper](https://cloud.githubusercontent.com/assets/4203505/25827932/64235194-3402-11e7-95a5-5f4ad73d78b4.png)

### Additional information
Same aberration as reported in [issue 83 in the opensim32 repo](https://github.com/opensim-org/opensim32/issues/83).

aseth1:
The `AnalyzeTool` does not perform a forward dynamics simulation. The `controls_file` is likely being applied at each instant, but since excitation only affects the activation rate, which is not being integrated, there is no way to update the activation state at the next time frame. Instead the `AnalyzeTool` is likely computing the muscle equilibrium at its default activation.

The *skyscraper* bug in 3.3 is likely due to a jump in (or bad) the equilibrium solution.

For excitation to affect the answer, switch to the `ForwardTool` and prescribe the isokinetic motion of the block. You should be able to associate the same Analyses with the `ForwardTool`.

tkuchida:
The behavior you describe in 4.0 is different from the behavior in 3.3. Has this change been documented somewhere?

aseth1:
I believe this is the same behavior since 2.4 and perhaps earlier. The `AnalyzeTool` never performed a forward simulation AFAIK.

tkuchida:
I am not claiming that the `AnalyzeTool` performs (or ever performed) forward simulations. What I am saying is that the `AnalyzeTool` now behaves differently than it did before. (The tug-of-war example used to work.) Has this change in the behavior of the `AnalyzeTool` been documented? Users who have been relying on the old behavior in their workflows will be surprised when they upgrade to 4.0.

aseth1:
It is unclear whether the `AnalyzeTool` is behaving differently or if the solution to the muscle equilibrium (and/or other muscle model) are affecting the answer. Given how controls affect activations it is unlikely that the `AnalyzeTool` does not apply controls (and title of the issue) is the cause of the discrepency.

[...]

PR #1686, opened 2017-05-01
Fix 1675 in AnalyzeTool forming states Storage with zero activation

aseth1:
Fixes issue #1675

### Brief summary of changes
When forming a `Storage` of `State`s from coordinates, unspecified states, including activation and fiber-lengths are being set to zero. The fix is to use default state values when states are not supplied by a column of the Tool's input motion or storage file when forming the complete *States Storage*.

`Model::formStatesStorage` was updated to use the default state instead of assuming zero. Use of `AnalyzeTool::createStatesStorageFromCoordinatesAndSpeeds()` was replaced by calls to `Model::formStatesStorage` instead. 

### Testing I've completed
Verified that model and setup in #1675 produces the same answers as in 3.3.

Should `AnalyzeTool::createStatesStorageFromCoordinatesAndSpeeds()` be removed by this PR or left for the time being or replaced by another PR?

### CHANGELOG.md (choose one)
- no need to update because there is no change to the API (bug fix only)


OpenSim/Simulation/Model/Model.cpp
```diff
@@ -1975,29 +1979,30 @@ void Model::formStateStorage(const Storage& originalStorage, Storage& statesStor
         }
         mapColumns[i] = fix;
         if (fix==-1){
-            cout << "Column "<< rStateNames[i] << " not found in formStateStorage, assuming 0." << endl;
+            cout << "Column "<< rStateNames[i] 
+                 << " not found in formStateStorage, assuming its default value."
```
tkuchida: Suggest `...value of " << ...` for this user-facing message.

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -1975,29 +1979,30 @@ void Model::formStateStorage(const Storage& originalStorage, Storage& statesStor
...
+                originalVec->getDataValue(mapColumns[column] - 1, assignedValue);
+            else
+                assignedValue = defaultStateValues[column];
+
+            stateVec.setDataValue(column, assignedValue);
         }
         statesStorage.append(stateVec);
     }
     rStateNames.insert(0, "time");
     statesStorage.setColumnLabels(rStateNames);
```
aymanhab: Why was the line to delete[] mapColumns removed?

OpenSim/Tools/AnalyzeTool.cpp
```diff
@@ -401,7 +401,15 @@ loadStatesFromFile(SimTK::State& s)
...
-        _statesStore = createStatesStorageFromCoordinatesAndSpeeds(*_model, *qStore, *uStore);
+        // used to use createStatesStorageFromCoordinatesAndSpeeds(*_model, *qStore, *uStore);
+        double ti = qStore->getFirstTime();
+        double tf = qStore->getLastTime();
+        uStore->addToRdStorage(*qStore, ti, tf);
+
+        delete _statesStore;
+        _statesStore = new Storage(512, "states");
+
+        _model->formStateStorage(*qStore, *_statesStore);
```
aymanhab: The method loadStatesFromFile is used by the UI of the AnalyzeTool, it's hard to see what are the side-effects of this change without integrating. Also the plotter calls AnalyzeTool.setStatesStorage() and passes in a storage created on GUI side based on inputs from the plot dialog. Do you expect the behavior of that call to change? This has to be done because the plotter varies Coordinates internally and allows users to set Activation-override values,

OpenSim/Tools/AnalyzeTool.cpp
```diff
@@ -434,8 +442,14 @@ setStatesFromMotion(const SimTK::State& s, const Storage &aMotion, bool aInDegre
...
-    _statesStore = createStatesStorageFromCoordinatesAndSpeeds(*_model, *qStore, *uStore);
+    double ti = qStore->getFirstTime();
+    double tf = qStore->getLastTime();
+    uStore->addToRdStorage(*qStore, ti, tf);
+
+    delete _statesStore;
+    _statesStore = new Storage(512, "states");
 
+    _model->formStateStorage(*qStore, *_statesStore);
     delete qStore;
```
aymanhab: The remaining reference to createStatesStorageFromCoordinatesAndSpeeds as far as I can tell is from RRATool, does that need to be changed/assessed?

OpenSim/Simulation/Model/Model.cpp
aseth1: Because `mapColumns` is now a container (`Array<int>`).

OpenSim/Tools/AnalyzeTool.cpp
aseth1: I changed the implementation of `loadStatesFromFile` to use an existing method ( `Model::formStateStorage`) internally with no change to the interface. If the GUI relies on side-effects rather than an interface then their could be issues. I hope this would improve consistency between `AnalyeTool` and other Tools that use `Model::formStateStorage`.

OpenSim/Tools/AnalyzeTool.cpp
aseth1: It should be changed for consistency as well. Why it likely doesn't matter is that the `AnalyzeTool` is used to run the old `InverseDynamics` `Analysis` to compute residuals. The model in RRA typically replaces its `ForceSet` and therefore has no muscles, in which case there are no states other than coordinates and their speeds. 

Removing the redundancy of `AnalyzeTool::createStatesStorageFromCoordinatesAndSpeeds()` is a small improvement. I can make it part of this PR and or do that separately. Separate may be better since this PR's aims to fix the bug in #1675.

aseth1:
> assuming the code is tested by our test cases (would be good to verify or beef up if needed)

I was surprised that this bug slipped through our tests that use the `AnalyzeTool` and especially `StaticOptimization`.  In my view, the tests for `StaticOptimization` do not reflect real use cases. In all the cases, for examples, the `<solve_for_equilibrium_for_auxiliary_states>` flag is  *false*. This doesn't make sense when the input to SO are coordinate values. Had the flag been *true* as one would expect to compute fiber-length states by solving the equilibrium problem first, then we would have caught this error. 

The only reason for having the option to not solve for equilibrium is to make SO use muscle fibers provided by the input states file (say from a previous costly forward simulation). So a better setup would be to automatically compute equilibrium when coordinates (motions) are supplied and not when states are provided.

aseth1:
@aymanhab How does the GUI implement the *activation override* feature in the plotter? Does it temporarily override the default_activation or does it directly set the activation in the states passed to the Analyses?

aymanhab:
@aseth1 The GUI costructs a states file because it needs to set activation and also to vary a coordinate (sometimes). Then it uses the states file route to pass these to the AnalyzeTool. This also make the plotting have no side-effects on the model in question.

aseth1:
Both @aymanhab and @tkuchida raised concerns about this and similar bugs creeping back in because the `AnalyzeTool` lacks adequate test cases. I added a test that does what I expect the `AnalyzeTool` should do and compares it with what `AnalyzeTool`  reports. It can also be used to reproduce the skyscraper bug.

OpenSim/Simulation/Model/Model.cpp
aymanhab: :+1:

Applications/Analyze/test/testAnalyzeTool.cpp
```diff
@@ -0,0 +1,191 @@
...
+// states are not provided. Note if default_activation is 1.0 it 
+// reproduces  the "skyscraper" bug issue #83 in OpenSim32 repo
+void testTugOfWar(const string& dataFileName, const double& defaultAct);
+
+
+int main()
+{
+    SimTK::Array_<std::string> failures;
+
+    try { /*testTutorialOne();*/ }
```
tkuchida: Should this test still be commented out? :cityscape:

aseth1: Darn, no. Thanks for catching that.

Applications/Analyze/test/testAnalyzeTool.cpp
```diff
@@ -1,5 +1,5 @@
 /* -------------------------------------------------------------------------- *
- *                    OpenSim:  testAnalyzeTutorialOne.cpp                    *
+ *                       OpenSim:  testAnalyzeTool.cpp                        *
```
tkuchida: :+1: Bonus marks ;)

OpenSim/Tools/AnalyzeTool.cpp
aymanhab: Agree the removal can wait, but would be good to add some documentation to the header file warning against use or the pitfalls in case we don't get around to complete removal.

aseth1:
Thanks @tkuchida and @aymanhab. Issue #1696 was added to track the `AnalyzeTool` remnant.
The CI failure is due unrelated timeouts.

PR #1689, opened 2017-05-04
A bug in JointReaction anslysis when used in combination of iterative forward simulation

mitkof6:
### Bug explanation

If JointReaction analysis is added to a model and the simulation (forward) is performed in iterative manner, by stepping the integrate function of Manager, JointReaction analysis keeps only the last time instance. 

By closer inspection of the problem [Manager.integrate](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Manager/Manager.cpp#L680) calls `initializeStorageAndAnalyses  `which calls `begin `for each analysis. In JointReaction analysis due to the [JointReaction.setupStorage](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Analyses/JointReaction.cpp#L447) called by `begin `method the storage is reset to 0. The rest of analyses do not reset their storages in the `begin `method so the problem was not observed for any other analysis. Please note that if the manager is not used in iterative manner then begin method is called one and there is not any problem.

### Brief summary of changes

The `_storeReactionLoads.reset(0);` is moved from `setupStorage `to `setNull()`

### Testing I've completed

The behavior of JointReaction have been tested on custom code which reproduces the problem. Moreover, the JointReaction test passes.

### DIFF
--- a/Applications/Analyze/test/testJointReactions.cpp
+++ b/Applications/Analyze/test/testJointReactions.cpp
@@ -28,24 +28,32 @@
 using namespace OpenSim;
 using namespace std;
 
+void testJointReactionDuringIterativeForwardSimulation();
+
 int main()
 {
     try {
         AnalyzeTool analyze("SinglePin_Setup_JointReaction.xml");
         analyze.run();
-        Storage result1("SinglePin_JointReaction_ReactionLoads.sto"), standard1("std_SinglePin_JointReaction_ReactionLoads.sto");
+        Storage result1("SinglePin_JointReaction_ReactionLoads.sto"), 
+            standard1("std_SinglePin_JointReaction_ReactionLoads.sto");
         CHECK_STORAGE_AGAINST_STANDARD(result1, standard1, 
-            std::vector<double>(standard1.getSmallestNumberOfStates(), 1e-5), __FILE__, __LINE__,
-            "SinglePin failed");
+            std::vector<double>(standard1.getSmallestNumberOfStates(), 1e-5),
+            __FILE__, __LINE__, "SinglePin failed");
         cout << "SinglePin passed" << endl;
 
         AnalyzeTool analyze2("DoublePendulum3D_Setup_JointReaction.xml");
         analyze2.run();
-        Storage result2("DoublePendulum3D_JointReaction_ReactionLoads.sto"), standard2("std_DoublePendulum3D_JointReaction_ReactionLoads.sto");
+        Storage result2("DoublePendulum3D_JointReaction_ReactionLoads.sto"), 
+            standard2("std_DoublePendulum3D_JointReaction_ReactionLoads.sto");
         CHECK_STORAGE_AGAINST_STANDARD(result2, standard2, 
-            std::vector<double>(standard2.getSmallestNumberOfStates(), 1e-5), __FILE__, __LINE__,
-            "DoublePendulum3D failed");
+            std::vector<double>(standard2.getSmallestNumberOfStates(), 1e-5), 
+            __FILE__, __LINE__, "DoublePendulum3D failed");
         cout << "DoublePendulum3D passed" << endl;
+
+        // check if joint reaction analysis behaves correctly during
+        // iterative forward simulation
+        testJointReactionDuringIterativeForwardSimulation();
     }
     catch (const Exception& e) {
         e.print(cerr);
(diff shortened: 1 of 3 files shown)


tkuchida:
Thanks for submitting this fix! I'm unfamiliar with this machinery, but it seems like it might be safer to make sure the manager calls the initialize method only when necessary. Conceptually, I think the analyses are supposed to be running `begin()` only at the beginning and `step()` on subsequent steps. Not sure whether that matters here, though. @aseth1

aseth1:
@mitkof6 Thank you for finding and proposing a fix to the issue. It would be great if you could add the case that exposed the bug as a test case. Comparing your changes to previous analyses I don't see any inconsistencies per se, but seeing your use case might bring up conflicts with other uses. For example, if the model is being simulated repeatedly as part of an optimization to mininmize reaction loads then we would want the Storage of reaction loads to be reset every iteration and not continue to add results to the previous Storage from a prior run. I think that is what @tkuchida is (correctly) concerned about.

mitkof6:
@aseth1 @tkuchida thank you for your comments. I have updated the commit with a test case.

Applications/Analyze/test/testJointReactions.cpp
```diff
@@ -54,3 +62,56 @@ int main()
...
+
+    // analysis
+    JointReaction* reaction = new JointReaction(&model);
+    model.addAnalysis(reaction);
+
+    // build model
+    //model.setUseVisualizer(true);
+    model.finalizeFromProperties();
+    model.buildSystem();
+    auto s = model.initializeState();
```
aseth1: you can replace these three lines with 
`auto s = model.initSystem();`

Applications/Analyze/test/testJointReactions.cpp
```diff
@@ -54,3 +62,56 @@ int main()
...
+
+    // iterative numerical integration  
+    for (unsigned int i = 1; i*dt <= t_end; ++i) {
+        double t0 = s.getTime();
+        double tf = i * dt;
+
+        // do something ...
+
+        manager.integrate(s, tf);
+        s.setTime(tf);
```
aseth1: This is not necessary. `s` is updated by integration.

mitkof6:
@aseth1 I have addressed your comments.

tkuchida:
@aseth1 This issue recently came up on the Forum ([topic 8157](https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=8157)). Please re-review.

chrisdembia:
I think the analyses should be consistent with each other. 

Potentially relevant: `TableReporter` has a `clearTable()`: https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Reporter.h#L193

mitkof6:
If this can't be resolved in the `begin()` method then maybe we can change the `integrate `function of the `Manager`. But I am not sure how to proceed. Note that other analysis, e.g. `KinematicsAnalysis `, do not exhibit this behavior.

chrisdembia:
It seems that MuscleAnalysis resets in `begin()`.

aseth1:
> If this can't be resolved in the begin() method then maybe we can change the integratefunction of the Manager. But I am not sure how to proceed. Note that other analysis, e.g. KinematicsAnalysis, do not exhibit this behavior.

@mitkof6 we very much appreciate your contributions. As @chrisdembia points out the `TableReporter` mechanism works in very much the way you expect with the underlying table growing until you explicitly clear it. The main difference being that the `TableReporter` will throw if the subsequent intervals are not contiguous (at least increasing in time), since the underlying `TimeSeriesTable`  is strict about that.

The best way for you to proceed is to use the the joint reaction Outputs of a `Joint` and feed it directly to a `TableReporter`. Have you tried that? A fall back plan is to append the Storage of each iteration of your integration loop to another *total* or *accumulated* `Storage` -- that will definitely work. 

We are hesitant to accept your fix because any existing user code that uses a `JointReaction` analysis in an optimization will fail because the Storage will simply grow on every call. Please let us know if you agree or not. Thank you.

mitkof6:
@aseth1 thank you for your thoroughly suggestions. 

> We are hesitant to accept your fix because any existing user code that uses a JointReaction analysis in an optimization will fail because the Storage will simply grow on every call. Please let us know if you agree or not. Thank you.

I agree that this fix may cause problems somewhere else, and possibly we should look for another approach. My personal opinion is that if I am a user of an API, I would like to have different options available and to know that each API call is doing specific concrete things, without assuming how the user will use it. Maybe it would be best if method `integrate()` does not call `begin()` for the underlying analyses, but a separate public method is provided and documented. Moreover, this method will be automatically called when the Manager is constructed to ensure that everything is initialized correctly. This gives more freedom to the user. Thank you for your time!

aseth1:
> Maybe it would be best if method integrate() does not call begin() for the underlying analyses, but a separate public method is provided and documented. 

Agreed and this would fit in with our changes to have separate `Manager::initialize()` and `intergate()` methods (PR #1909), where `initialize() would call `begin()`  and not `integrate()`.

PR #1718, opened 2017-05-17
Update simbody dependency commit to include typo fix.

chrisdembia:
### Brief summary of changes

I updated the simbody dependency commit hash to the latest simbody commit.

### CHANGELOG.md (choose one)

- no need to update because...we might end up updating the commit again before release.


chrisdembia:
GitHub is saying there is a merge conflict, but I don't think there is. I think something GitHub is being buggy right now.

chrisdembia:
Something went wrong with this PR. I think it's because I used a branch name that had been used previously.

PR #1719, opened 2017-05-17
Update simbody commit hash to include typo fix (second attempt)

chrisdembia:
I updated the simbody dependency commit hash to incorporate a typo fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,10 @@ addons:
       - george-edison55-precise-backports # cmake 3.2.3 / doxygen 1.8.3
 
     packages:
+      # Must explicitly list cmake-data, otherwise the incorrect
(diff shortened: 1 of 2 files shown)


chrisdembia:
Thanks @tkuchida

chrisdembia:
This PR also fixes an issue that was causing all Linux travis-ci builds to fail.

aymanhab:
LGTM. I'll merge to fix CI failures even though I don't totally understand the .travis file changes.

chrisdembia:
Thanks @aymanhab . I copied the fix from here: https://github.com/travis-ci/travis-ci/issues/4631

> It's worth mentioning that to get CMake to install from the george-edison precise backports you have to explicitly install cmake-data along with cmake, otherwise apt-get will try to install the wrong version of cmake-data.

Issue #1720, opened 2017-05-19
GeometryPath has addhoc method to control its color


aymanhab:
GeometryPath is used as a utility for computation and to visualize Muscles, PathSprings. It has no associated Appearance property so its behavior is inconsistent with other geometry (e.g. Opacity and Show/Hide status are not persistent). This breaks functionality that used to exist in version 3.3.

aymanhab:
PR in review

jenhicks:
Need to verify that you can show/hide/wireframe/change transparency/color of muscles

PR #1721, opened 2017-05-19
Add appearance to geometry path

aymanhab:
Fixes issue #1720 

### Brief summary of changes
Add Appearance property to GeometryPath so that its default color and other appearance related preferences are persistent. default_color which was introduced earlier but rarely used is now carried over to Appearance. 

### Testing I've completed
From the GUI I can now change default_color, opacity and hide/show status of a Muscle and see the change in xml and in visualization upon deserialization. Default colors for PathSpring and Ligament changed per team suggestions.

How the default_color should interact /play with getColor(state) which returns an RGB tuple representing activation. As of now the GUI will use color from Appearance unless running simulation or playing back results that have activations. 

### CHANGELOG.md (choose one)
This is a new functionality to support the GUI/Visualizer. We should document the new GUI/Visualizer capability in User guide.


aymanhab:
This should be pretty quick review, I just upgraded "default_color" to "Appearance" for consistency and to make hide/show/user-colors persistent with improved functionality from GUI. @chrisdembia @tkuchida @aseth1

chrisdembia:
Do we need updateFromXMLNode() code to move previously-existing default colors to the appearance?

Do all properties of the appearance have an effect (e.g., wireframe, etc.)? If not, we should mention which ones are ignored in the class description for GeometryPath.

> As of now the GUI will use color from Appearance unless running simulation or playing back results that have activations.

This sounds good to me. Should we document this behavior in the class description for GeometryPath?

aymanhab:
@chrisdembia thanks for the awesomely fast review.
default_color had no effect on GUI (it always used the red-blue range) and the comment on the property was less than useful to actual users (Used to initialize the colour cache variable)! All the Appearance properties are observed by the new visualizer, the API visualizer is not capable of some of these (e.g. wireframe) and can't handle the distinction between simulation vs. just live editing. Documentation needs to be improved. Old "VisibleObject" had a lot of unsupported properties was only used for show/hide by GUI.

aseth1:
LGTM @aymanhab. Should there be a corresponding `GeometryPath::updateFromXMLNode()` to map the conversion of old color property to an Appearance?

IMHO `GeometryPath` is closely enough related to `Geometry` to warrant adding `Appearance`, but I wouldn't want to have arbitrary components adding them. Ideally the `Geometry` used by the `Component` could have their appearances modified directly.

aymanhab:
Thanks @aseth1 and @tkuchida 
The convention we've been using is that objects that use generateDecorations (e.g. ContactMeshes, Wrap Objects) have an associated appearance so that their display can be controlled. Actual geometric objects (e.g. Sphere, Cone, etc.) have a built in Appearance. This allows the GUI to handle both rather uniformly.

aseth1:
> ...the comment on the property was less than useful to actual users (Used to initialize the colour cache variable)! All the Appearance properties are observed by the new visualizer ...

@aymanhab I agree that the comment for the `default_color` was not great but it doesn't mean it wasn't being used. I was being respected in the API Visualizer to color Ligaments and PathSprings. What happens to those now? The color specified in the Appearance is used in the API Visualizer?

aymanhab:
@aseth1 thanks for pointing this out. The getDefaultColor method had been rewired to get the color from the Appearance rather than from the default_color property. Do you think this justifies adding an updateFromXML?

aseth1:
I think it warrants having the updateFromXML for no other reason than we automatically handle upgrading user models to the latest version without loss of information. If I spent the time to color my model's ligaments/pathsprings, then I would want my model to present those colors in 4.0.  It would at least inspire some confidence that its safe to updgrade.

aymanhab:
Thanks @aseth1 will do, but if the issue is to display the model same way as in 4.0 then we have a bigger issue with muscles which was changed to display in "Gray" by default now where it used to be "red" or "blue" or "white" depending on GUI, API visualizer. Should we restore muscle's default color to the implicit "red" that was used in the past (even though it could have been assumed to indicate full activation?) @chrisdembia @tkuchida @jenhicks @jimmyDunne

jenhicks:
My vote is to use red as a default muscle color. My rationale: 
- If I wanted to grab a screenshot of my model for a talk or poster, I'd rather have the muscles be red than gray. If the default were gray, I'd have to go and change them all. (I assume you could do this with a  bulk edit in the GUI, but would still be extra work). 
- Consistent with 3.3. I don't think default of red has ever come up as an issue or point of confusion? (That I know of). 
- If people open a model in OpenSim for the first time, a model with red muscles seems more visually appealing.

aseth1:
@jenhicks @ aymanhab red seems appropriate for muscles. I don't believe the default color for the `GeometryPath` should be red, however. Each concrete class that uses a `GeometryPath` can override the default with its own color, which is currently the case for PathSprings and Ligaments, which do not appear as red by default.

jenhicks:
I agree that ligaments and PathSprings should not be red by default. Can we use their defaults from 3.3?

aymanhab:
Added updateFromXML for GeometryPath and it handles visible flag only since the colors were never stored in XML. Modified default colors for Muscle, PathSpring and Ligament per feedback, and added method in Muscle to override default color of GeometryPath which is still neutral Gray. Models now load with red muscles.

aymanhab:
Bogus test failure, all comments have been addressed. Ready for re-review.

aseth1:
> it handles visible flag only since the colors were never stored in XML

How can you be certain of that? I can understand that any `default_color` that was set was not being used by the GUI, but I am pretty sure the API visualizer was. What is the downside of transferring any `default_color` associated with the `GeometryPath` to its new `Appearance`? It will be use default  anyways if no color is specified.

aymanhab:
@aseth1 No problem, I can transfer the old default_color into the Appearance (though API visualizer would likely overwrite it based on the state anyway). The bigger problem I'm  facing now is that colors that get changed in the GUI and saved to the osim file correctly are wiped out by the new default "Red" color on deserialization. This is definitely because of the new default "Red" is set in extendFinalizeFromProperties(). Do you have a suggestion as to where should that default be set instead? I tried Muscle::constructProperties() which seems to be the correct place but that had no effect (since it was never invoked during deserialization)! I don't want to change the DefaultColor of the base GeometryPath either.

aymanhab:
Nevermind, found a solution that makes it all work as desired. Will push shortly

chrisdembia:
>  Nevermind, found a solution that makes it all work as desired.

Whew

OpenSim/Simulation/Model/Muscle.cpp
```diff
@@ -144,6 +157,7 @@ void Muscle::constructProperties()
     // By default the min and max controls on muscle are 0.0 and 1.0
     setMinControl(0.0);
     setMaxControl(1.0);
+    upd_GeometryPath().setDefaultColor(DefaultMuscleColor);
```
chrisdembia: Does this mean that the other forces with GeometryPath now use an inconsistent mechanism?

aymanhab: The argument was made that muscles should show red by default, but not other forces that use GeometryPath, that's why the color setting can't be made by the GeometryPath itself.

aymanhab:
Concerns/issues have been addressed. Muscles display properly in red for old models, colors can be changed and serialize/deserialize/display with correct colors on reloading.

OpenSim/Simulation/Model/Muscle.cpp
chrisdembia: I'm still confused. Will custom default colors for PathSprings be properly deserialized? For PathSpring, setDefaultColor is called in `extendFinalizeFromProperties()`, not in `constructProperties()`, so it seems the behavior would be different (that user-defined default colors would always be overridden as green).

OpenSim/Simulation/Model/Muscle.cpp
```diff
@@ -120,6 +121,19 @@ void Muscle::updateFromXMLNode(SimTK::Xml::Element& aNode, int versionNumber)
...
+            SimTK::Xml::element_iterator  geomPathIter = aNode.element_begin("GeometryPath");
+            if (geomPathIter != aNode.element_end()) {
+                SimTK::Xml::element_iterator  defaultColorIter = geomPathIter->element_begin("default_color");
+                if (defaultColorIter == geomPathIter->element_end()) {
+                    SimTK::Xml::Element myDefaultColorEement("default_color");
+                    myDefaultColorEement.setValue(".8 .1 .1"); // DefaultMuscleColor
+                    geomPathIter->appendNode(myDefaultColorEement);
+                }
+            }
+        }
```
chrisdembia: I don't fully understand what's happening. This causes the GeometryPath to set the appearance color to red, which I would think would be achieved by the new line in `constructProperties()`. Also, what happens for files with version>=30516? Will they still show up red by default and also comply with different colors specified by users in XML?

aymanhab: Deserialized objects don't go through the constructProperties() call as far as I could tell (by stepping through the code) that's why I use the updateFromXMLNode route. In fact if I remove this block the red color was never set on deserialization. Newer versions will serialize with a color and will not go through this block of code (as verified by testing round-trip serialize/deserialize models with arbitrary colored muscles)

OpenSim/Simulation/Model/Muscle.cpp
aymanhab: Good point @chrisdembia.  I haven't explicitly tested (or looked for models with PathSpring, ClutchSpring, Ligament and possibly other kind of Objects that use GeometryPath). I only changed the default colors for those objects proposed by reviewers. If all works for Muscles (which is a new functionality anyway) then we can apply the same for other objects as needed. I'd prefer to have models to test with before making changes. If you think I  should hold until we have such models or you can find example models I'd be happy to make changes/test.

aseth1: It looks like this is here to force Muscle to have a default color of red regardless of its `default_color` property value. Is that right, @aymanhab?  Is that really necessary? Since muscles in the past did not have a default color and now the consumers of `GeometryPath` provide reasonable defaults, why force it here? If someone want to change the `default_color` then why prevent it? I don't think old models will accidentally visualize as not Red.

aseth1:
@aymanhab the `GeometryPath::updateFromXMLNode` changes LGTM with user settings be used if defined. The only question I raise is why we are forcing the default color of `Muscle` to be *Red* and ignoring the user set value. I don't expect any existing models to have `default_color` of `Muscle`s GeometryPath` exposed, so they would draw Red by default anyways. If someone really wanted to change the default color, why prevent it?

OpenSim/Simulation/Model/Muscle.cpp
aymanhab: @aseth1 This is here so that new muscles created by the API have the default red color, muscles that are deserialized get their default color from Properties as have been tested by GUI. I'm ok with removing this line but I expect making a model from API and writing to XML will result in gray muscles.

aymanhab: Any preference @aseth1 ? (The options are: remove the line -> new Muscles show in default gray but users can set the color through the API if they care) or leave the line -> new Muscles show in Red and users can change later. Thanks.

aseth1: I don't understand why a Muscle would not appear red by default if a Ligament is green by default. It seems that the mechanism should be consistent either way.

aseth1: OK I think you are right Ayman. The  Ligament is always green because `extendFinalizeFromProperties` is setting the color to the static const value. I think your constructProperties() intercept is correct with the consumer of the `GeometryPath` having a chance to have its say for the default color as a property. The overwriting of the color in `extendFinalizeFromProperties` should be removed however so that whatever user set value of the property is used.

aseth1:
Created issue #1738 to track that mechanism to handle default color in the various consumers of GeometryPath classes are not consistent as of this PR. 
@chrisdembia if you are OK with a subsequent PR to address #1738 in the interest of getting a GUI build with correct Muscle appearance, I would go ahead and merge, if you have no other objection.

aymanhab:
Thanks much @aseth1 I totally agree that we need to make this usage consistent. Will definitely address with #1738 before 4.0

aymanhab:
Build failure due to timeout.

aymanhab:
Can we merge @chrisdembia based on the dev meeting?

OpenSim/Simulation/Model/Muscle.cpp
chrisdembia: Thank you for explaining. My understanding is that all new model files will now have the color show up explicitly in the XML file (for muscles). I find that a little unfortunate, but fine for now.

aymanhab:
Great, thanks much @chrisdembia and @aseth1 🎆

aymanhab:
@chrisdembia The only reason red needs to show in file even though default is that it's default for muscle and not the GeometryPath that it uses, while other objects e.g. Ligaments may propose a different default. No specification now means "Gray".

Issue #1700, opened 2017-05-08
Muscle equilibrium solve mishandles Newton steps to short fiber lengths


aseth1:
When the current iteration steps to a fiber length < min_fiber_length, the fiber length is *clamped* to the minimum. Subsequent step estimates a length that overshoots the solution and then next step is back to a length that is < min_fiber_length, which gets clamped to the min. From this point on the solve is in endless cycle. A max count in which the min fiber length is *achieved* (regardless of bouncing back > min) is triggered and the fiber length is incorrectly set to the min_fiber_length. This causes a jump disontinuity in the muscle's reported fiber-length and corresponding muscle force from the AnalyzeTool.

This issue results in the [skyscraper bug](https://github.com/opensim-org/opensim32/issues/83) reported by @tkuchida.

aseth1:
The crux of the issue is the ad-hoc implementation of the Newton solve.  A robust method will halve the step-size with the error increases due to a step (which is also the case above).  To test this hypothesis I halve the step-size every time in this case the algorithm converges to the correct solution (no more cycling back and forth between two bad solutions).

A positive change would be to abstract out the Newton solver baked into the muscle model and call a separate (published/tested) RootSolver implementing Newton-Raphson or other established (e.g. [Brent's](https://en.wikipedia.org/wiki/Brent%27s_method#Brent.27s_method))  method.

chrisdembia:
@humphreysb may be interested.

tkuchida:
> This issue results in the skyscraper bug reported by @tkuchida.

See #1675.

aymanhab:
PR in review

PR #1728, opened 2017-05-23
Fix muscle equilibrium solve bug #1675 and #1700

aseth1:
Fixes issue #1675 and #1700 

### Brief summary of changes
The Newton algorithm embedded in the equilibrium solve was incorrectly assessing solutions at the bounds according to an arbitrary count (not consecutive hits) of lower bound breaches and wrongly concluding that the solution was at a minimum (#1700). This was causing the discontinuity seen in the *skyskraper* bug (#1675). 

This issued was addressed first in `Millard2012EquilibriumMuscle::estimateMuscleFiberState()` method to enable `testAnlayzeTool` to pass, which originally reproduced the #1675. The fix also required iteratively halving the size of the Newton step when the equilibrium error exceeds the previous error computed, and then updating derivatives to take another Newton step.

`testMuscles` was updated to sweep over initial conditions in solving the equilibrium problem and the `Thelen2003Muscle::estimateMuscleFiberState::initMuscleState()`, as expected, failed.  It was similarly updated. Additional care was taken to update the force-velocity multiplier *fv*, which also varies with the fiber-length.

Methods to report the tendon, active and passive fiber forces, and the equilibrium error as a function of fiber-length were added so that equilibrium function being solved could be visualized and the method debugged.

Made numerous small changes to `testMuscles` to make in more streamlined. I intend to complete the changes and extend testing to other muscle models as part of another PR.

### Testing I've completed

Added the isometric muscle contraction analysis to `testAnalyzeTool`.

A more complete/rigorous condition was added to `testMuscles` where a full range of muscle-tendon lengths and default activations are being tested. The amplitude of the prescribed motion displacement was increased as well as the duration to test both shortening and lengthening to the fullest extent of the muscle. After simulating the muscle, the equilibrium of the muscle is computed at every state and over a sweep (0.01-1.0) of activation at each state.

The changes to `testMuscles` are not complete, but do restore the previous coverage and test the same conditions plus the equilibrium. The same condition are also applied to each muscle model tested (and options that only enable one condition to be tested a time have been removed). For instance the power test is applied to the `RigidTendonModel` and that exposed a bug in its pennation angle calculation, which was also fixed.

Please verify that no gaps in testing have been introduced by the changes to `testMuscles`.

I effectively removed the *system energy - work = C* test (still there but flag is permanently false) since it is essentially equivalent in coverage to the power test (the energy test being its time integral with integration inaccuracies making the comparison inexact). Are we OK with it being removed?

### CHANGELOG.md (choose one)

- no need to update because these are bug fixes and test cases


OpenSim/Actuators/Test/testMuscles.cpp
```diff
@@ -1108,10 +958,172 @@ void testDelp1990Muscle()
...
+    double tendonForce = SimTK::NaN;
+    double activeFiberForce = SimTK::NaN;
+    double passiveFiberForce = SimTK::NaN;
+    double cosphi = SimTK::NaN;
+    double flm = SimTK::NaN;
+    double a = SimTK::NaN;
+
+    double fvm = 1.5;
+
+    assert(!muscle.get_ignore_tendon_compliance());
```
tkuchida: See comment on analogous line above/below (wherever it was…).

OpenSim/Actuators/Test/testMuscles.cpp
```diff
@@ -1108,10 +958,172 @@ void testDelp1990Muscle()
...
+
+    SimTK::RowVector row(labels.size(), SimTK::NaN);
+    for (int i = 0; i < N; ++i) {
+        fiberLength += dl;
```
tkuchida: Skipping first point? See analogous comment elsewhere.

OpenSim/Actuators/Test/testMuscles.cpp
```diff
@@ -1108,10 +958,172 @@ void testDelp1990Muscle()
...
+
+    SimTK::State s = model.getWorkingState();
+    // Independently compute the active fiber force at every state
+    for (int i = 0; i < nstates; ++i) {
+        s = statesTraj[i];
+
+        activation = 0.01;
+        // test a full sweep of default activations at each state
+        for (int j = 0; j*dAct <= 1.0; ++j) {
+            activation = (!j)*activation + j*dAct;
```
tkuchida: The logic here is unclear. Some muscles have a minimum allowable activation; presumably L1076 is meant to address that constraint. This code will break if `dAct < min_allowable_activation`, which is not obvious from the implementation. I suggest something more straightforward:
```cpp
const double minActivation = muscle.getMinimumActivation();
const int nSteps = 20;
const double dAct = (1.0 - minActivation) / nSteps;
for (int j = 0; j <= nSteps; ++j) {
    activation = std::min( minActivation + j*dAct, 1.0); //avoid exceeding 1 due to roundoff
    ...
}
```

tkuchida:
> no need to update because these are bug fixes and test cases

I think we should include a note about these bug fixes in CHANGELOG since they could change (improve) results.

OpenSim/Actuators/Test/testMuscles.cpp
```diff
@@ -497,23 +473,18 @@ void simulateMuscle(
         /*double xSpeed = */modelCoordinateSet[0].getSpeedValue(si);
         // double KEsysCheck = 0.5*ballMass*xSpeed*xSpeed;
         /*double PEsys =  */model.getMultibodySystem().calcPotentialEnergy(si);
-        double jointWork = jointWorkProbe->computeProbeInputs(si)(0);
-        double ESysMinusWork = Esys 
-                                - muscWorkProbe->computeProbeInputs(si)(0)
-                                - jointWork; 
-
+        double jointWork = jointWorkProbe->getProbeOutputs(si)(0);
```
tkuchida: Based on discussions, I think this is where the energy was being compared but there's now a more rigorous test for power so I think this block can just be deleted.

OpenSim/Actuators/Test/testMuscles.cpp
tkuchida: testAnalyzeTool.cpp L251

OpenSim/Actuators/Test/testMuscles.cpp
tkuchida: testAnalyzeTool.cpp L254

OpenSim/Actuators/Test/testMuscles.cpp
```diff
@@ -388,8 +362,8 @@ void simulateMuscle(
     double length = muscle.getLength(si);
     double trueLength = startX + xSinG - anchorWidth/2;
     
-    ASSERT_EQUAL(length/trueLength, 1.0, testTolerance, __FILE__, __LINE__, 
-        "testMuscles: path failed to initialize to correct length." );
```
chrisdembia: What is the reason for removing this assert?

OpenSim/Actuators/Test/testMuscles.cpp
```diff
@@ -473,7 +450,7 @@ void simulateMuscle(
     values for K and dl/dt
 ============================================================================
 */    
-    if(testType == 1){
+    if(false){
```
chrisdembia: I'm not sure I follow; the initialization test was empty before and is also now empty?

OpenSim/Actuators/Test/testMuscles.cpp
```diff
@@ -1108,10 +958,172 @@ void testDelp1990Muscle()
...
+
+void testMuscleEquilibriumSolve(const Model& model, const Storage& statesStore)
+{
+    // Get the muscle to test
+    const Muscle& muscle = model.getMuscles()[0];
+
+    if (dynamic_cast<const Millard2012AccelerationMuscle*>(&muscle)) {
+        // Millard2012AccelerationMuscle is not an Equilibrium-based Muscle
+        return;
+    }
```
chrisdembia: I don't think this is the right place for this if-statement; it should be in `testMillard2012AccelerationMuscle()`.

aseth1:
>I don't see a strong reason to remove this test; is it slow? I understand that it doesn't add coverage for the muscle models, but it still tests the probes' integration. Are the work probes sufficiently tested elsewhere?

Yes, virtually the identical integration test is included in `testProbes.cpp`.

chrisdembia:
> Yes, virtually the identical integration test is included in testProbes.cpp.

Oh okay; then yes I think it's fine to leave this test out.

OpenSim/Actuators/Test/testMuscles.cpp
aseth1: You are right, however, `testMuscleEquilibriumSolve` is a subtest invoked within `simulateMuscle`. I could add a flag to `simulateMuscle` to provide an option to `testEqulibriumSolve` but my preference was not to add that. Would you like me to add that option instead?

chrisdembia: No that's okay; what about adding a comment in `testMillard2012AccelerationMuscle()` where it calls `simulateMuscle()` to indicate that the equilibrium solve test is not performed for this muscle?

Applications/Analyze/test/testAnalyzeTool.cpp
```diff
@@ -140,12 +146,15 @@ void testTugOfWar(const string& dataFileName, const double& defaultAct) {
...
-    double af, pf, mf, tf;
-    af= pf = mf = tf = SimTK::NaN;
+    double af, pf, mf, tf, fl, fv;
+    af= pf = mf = tf = fl = fv = SimTK::NaN;
 
     // Tolerance for muscle equilibrium solution 
     const double equilTol = muscle.getMaxIsometricForce()*SimTK::SqrtEps;
 
+    // The maximum acceptable change in force between two contiguous states
+    const double maxDelta = muscle.getMaxIsometricForce() / 2;
```
aseth1: Good point. I simply used a value that was sufficient to detect the skyscraper bug. But now knowing what are the typical transitions I will tighten up.

OpenSim/Actuators/Test/testMuscles.cpp
aseth1: I meant to restore. Thanks!

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -1488,6 +1514,10 @@ Millard2012EquilibriumMuscle::estimateMuscleFiberState(
 
     if(abs(ferr) < aSolTolerance) {  // The solution converged.
 
+        if (isFiberStateClamped(lce, dlceN)) {
+            lce = getMinimumFiberLength();
+        }
```
aseth1: Yes, there was no check that the lce returned was below the minimum.

OpenSim/Actuators/Thelen2003Muscle.cpp
```diff
@@ -762,110 +764,163 @@ Thelen2003Muscle::initMuscleState(const SimTK::State& s,
...
+        ferrPrev = ferr;
+        lcePrev = lce;
+
+        //Update the partial derivative of the force error w.r.t. lce
+        dFm_dlce = calcDFmDlce(lce, ma, fv, fiso, ofl);
+        dFmAT_dlce = calcDFmATDlce(lce, phi, cosphi, Fm, dFm_dlce, vol);
+        dFmAT_dlceAT = dFmAT_dlce*cosphi;
+
+        dFt_d_tl = calcDFseDtl(tl, fiso, tsl);
+        dFt_d_lce = calcDFseDlce(tl, lce, phi, cosphi, fiso, tsl, vol);
```
aseth1: lambda functions are now employed and the algorithms between the `Thelen2003Muscle` and `Millard2012EquilibriumMuscle` are now identical and much easier to read (with slightly different local lambdas). Thanks for recommending the change.

OpenSim/Actuators/Thelen2003Muscle.cpp
```diff
@@ -762,110 +764,163 @@ Thelen2003Muscle::initMuscleState(const SimTK::State& s,
...
+        //acceleration level knowledge, which is not available in 
+        //general.
+
+        //Stiffness of the muscle is the stiffness of the tendon and the 
+        //fiber (along the tendon) in series
+
+        //the if statement here is to handle the special case when the
+        //negative stiffness of the fiber (which could happen in this
+        // model) is equal to the positive stiffness of the tendon.
+        if (abs(dFmAT_dlceAT + dFt_d_tl) > SimTK::SignificantReal
```
aseth1: The comment refers to the distribution of the muscle-tendon velocity to the fiber and tendon. Only the Millard2012EquilibriumMuscle had the a check and separate condition for evaluating the static fiber solution. The Thelen2003Muscle is able to solve for equilibrium without imposing the static assumption under all the conditions tested. I was not able to get the  Millard2012EquilibriumMuscle to do that. It does solve reliably with the static assumption imposed, which is it was doing before. I've only removed having to equilibrium methods and added the flag in the case we want to assume zero velocity.

aseth1:
> I think we should include a note about these bug fixes in CHANGELOG since they could change (improve) results.

The CHANGELOG has been updated.

OpenSim/Actuators/Test/testMuscles.cpp
tkuchida: The doxygen for Millard2012AccelerationMuscle already includes the following :open_umbrella: cautionary message: "The Millard2012AccelerationMuscle class is experimental and has not been extensively tested in all operational conditions."

aseth1:
@tkuchida and @chrisdembia thank your for your helpful and insightful reviews. The PR has been significantly revised to handle inconsistencies and make code changes more uniform (algorithm is virtually identical) between the Thelen2003Muscle and Millard2012EquilibriumMuscle equilibrium solve. As suggested, I used lambdas to reduce code duplication and make the solver implementation easier to read. 

The changes stop short of unifying the solver for both muscle models, which I think would be much easier to do now, but would involve deriving `Millard2012EquilibriumMuscle` from `ActivationFiberLengthMuscle` or introducing a common base like `EquilibriumMuscle` to tie them together. I don't think  that is on the critical path for 4.0. I created issue #1736 to track it.

OpenSim/Actuators/Thelen2003Muscle.cpp
```diff
@@ -762,110 +759,161 @@ Thelen2003Muscle::initMuscleState(const SimTK::State& s,
...
-                dphi = getPennationModel().calcPennationAngularVelocity(
-                                           tan(phi),lce,dlce);
-                /*dlceAT = */getPennationModel().calcFiberVelocityAlongTendon(
-                                             lce,dlce,sinphi,cosphi,dphi);
+    double ferrPrev = ferr;
+    double lcePrev = lce;
+
+    double h = 1.0;
+    while( (abs(ferr) > aSolTolerance)  && (iter < aMaxIterations)) {
+        dferr_d_lce = dFmAT_dlce - dFt_d_lce;
```
tkuchida: The comment preceding the identical line in Millard2012EquilibriumMuscle.cpp (L1414) would be helpful here as well.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -1364,122 +1281,181 @@ Millard2012EquilibriumMuscle::estimateMuscleFiberState(
...
-                //dlceAT = penMdl.calcFiberVelocityAlongTendon(lce,dlce,sinphi,
-                //                                             cosphi,dphi);
+            // Compute the force error
+            ferrFunc();
+
+            if (h <= SimTK::SqrtEps) {
+                break;
             }
+            else
+                h = 0.5*h;;
```
tkuchida: Second semicolon is unnecessary.

OpenSim/Actuators/Thelen2003Muscle.cpp
```diff
@@ -1623,9 +1670,9 @@ SimTK::Vector Thelen2003Muscle::
...
 
     if(abs(ferr) < tolerance){
-        result[0] = 1.0;
-        result[1] = aFalFv/(aAct*aFal);
+        result = max(0.0, aFalFv/(aAct*aFal));
+        return result;
     }
 
-    return result;
+    OPENSIM_THROW_FRMOBJ(Exception, "Failed to calculate fiber-velocity.");
```
tkuchida: I made the following comment in the first version of this PR:
"Can any kind of suggestion be provided here? At least something like "Fiber-velocity solver failed to converge" would be more descriptive ("Failed to calculate…" could mean a lot of things)."
My comment was not addressed and no explanation was provided. I think it is still a valid comment, but obviously is not critical.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -1364,122 +1281,181 @@ Millard2012EquilibriumMuscle::estimateMuscleFiberState(
...
 
-    while(abs(ferr) > aSolTolerance
-            && iter < aMaxIterations
-            && minFiberLengthCtr < 10) {
-
-        // Update the multipliers and their partial derivatives
+    //*******************************
+    // Helper functions
+    //Update position level quantities, only if they won't go singular
+    auto positionFunc = [&] {
```
chrisdembia: Wow cool; I did not know you could omit the parentheses.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -1364,122 +1281,181 @@ Millard2012EquilibriumMuscle::estimateMuscleFiberState(
...
+            // Compute the Newton step
+            delta_lce = -h*ferrPrev / dferr_d_lce;
+            // Take a Newton Step if the step is nonzero
+            if (abs(delta_lce) > SimTK::SignificantReal)
+                lce = lcePrev + delta_lce;
+            else {
+                // We've stagnated or hit a limit; assume we are hitting local
+                // minimum and attempt to approach from the other direction.
+                lce = lcePrev - sign(delta_lce)*SimTK::SqrtEps;
+                h = 0;
```
chrisdembia: This line previously had a comment `// force a break`; I thought this comment was helpful. Was there a particular reason for removing it?

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -395,81 +395,9 @@ computeActuation(const SimTK::State& s) const
...
-
-    } catch (const std::exception& x) {
-        OPENSIM_THROW_FRMOBJ(MuscleCannotEquilibrate,
-            "Internal exception encountered.\n" + std::string{x.what()});
-    }
-}
 
 void Millard2012EquilibriumMuscle::
-computeFiberEquilibriumAtZeroVelocity(SimTK::State& s) const
+computeFiberEquilibrium(SimTK::State& s, bool useZeroVelocity) const
```
chrisdembia: Nitpicky: consider flipping the meaning of the new parameter. Because `0` evaluates to false, it may make more sense for `false` to mean "zero velocity". The name of the parameter could instead be "solveForVelocity."

OpenSim/Auxiliary/auxiliaryTestFunctions.h
```diff
@@ -28,6 +28,7 @@
 #include <OpenSim/Common/LinearFunction.h>
 #include <OpenSim/Common/PropertyObjArray.h>
 #include "OpenSim/Common/STOFileAdapter.h"
+#include "OpenSim/Simulation/Model/ActivationFiberLengthMuscle.h"
```
chrisdembia: I was thinking of adding the new function to a new auxiliary header, since this header could be used by tests in the Common library (which should not use headers in downstream libraries).

OpenSim/Actuators/Test/testMuscles.cpp
```diff
@@ -638,51 +512,23 @@ void testRigidTendonMuscle()
...
         act0, 
-        &sineWave, 
+        &motion,
         &control, 
-        IntegrationAccuracy,
-        InitializationTest,
-        InitializationTestTolerance,
         false);
-
 }
```
chrisdembia: The identical test was performed twice?

OpenSim/Actuators/Test/testMuscles.cpp
```diff
@@ -1005,9 +821,27 @@ void testMillard2012EquilibriumMuscle()
...
+    // Test extremely short (but stretched) fiber, which should still initialize.
+    {
+        auto model = Model();
+
+        const double optimalFiberLength = 0.01; //short fiber and
+        const double tendonSlackLength =  0.1; //short tendon
+        auto muscle = new Millard2012EquilibriumMuscle("muscle", 1.,
+            optimalFiberLength, tendonSlackLength, 0.);
+        muscle->addNewPathPoint("p1", model.updGround(), SimTK::Vec3(0));
+        muscle->addNewPathPoint("p2", model.updGround(), SimTK::Vec3(0, 0, 1));
```
chrisdembia: That's a large stretch!

OpenSim/Actuators/Test/testMuscles.cpp
```diff
@@ -1033,24 +867,14 @@ void testMillard2012AccelerationMuscle()
 
     Sine motion(0.1, SimTK::Pi, 0);
 
+    // Note that simulateMuscle will call testMuscleEquilibriumSolve() but it
+    // will ignore the Millard2012AccelerationMuscle because it is not an
+    // equilibrium-based muscle model.
```
chrisdembia: Thanks.

OpenSim/Actuators/Test/testMuscles.cpp
```diff
@@ -1108,10 +929,95 @@ void testDelp1990Muscle()
...
+                // Write out the muscle equilibrium error as a function of
+                // fiber lengths.
+                const Thelen2003Muscle* thelen =
+                    dynamic_cast<const Thelen2003Muscle*>(&muscle);
+                if (thelen) {
+                    thelen->printCurveToCSVFile(
+                        Thelen2003Muscle::CurveType::FiberForceVelocity, "");
+                    reportTendonAndFiberForcesAcrossFiberLengths(*thelen, s);
+                }
+                else {
```
chrisdembia: No need to change anything, but this would be cleaner as:

```cpp
if (const auto* thelen = dynamic_cast...) {
    ...
} else if (const auto* millard = dynamic_cast...) {
    ...
}
```

OpenSim/Actuators/Thelen2003Muscle.cpp
```diff
@@ -700,19 +701,18 @@ Thelen2003Muscle::initMuscleState(const SimTK::State& s,
...
-    double ophi= getPennationAngleAtOptimalFiberLength();
-    double vol = ofl * sin(ophi);
-    double fiso= getMaxIsometricForce();
-    double vmax = getMaxContractionVelocity();//getPropertyValue<double>(VmaxName);
+    const double tsl = getTendonSlackLength();
+    const double ofl = getOptimalFiberLength();
+    const double ophi= getPennationAngleAtOptimalFiberLength();
+    const double vol = ofl * sin(ophi);
+    const double fiso= getMaxIsometricForce();
+    const double vmax = getMaxContractionVelocity();
```
chrisdembia: +1 for consts.

OpenSim/Actuators/Thelen2003Muscle.cpp
chrisdembia: +1

OpenSim/Actuators/Test/testMuscles.cpp
aseth1: Not exactly. The same simulation was executed twice, but one evaluated the "CorrectnessTest" and the other did the "InitializationTest" (which was just a stub).  I got rid of these *flags* so that all conditions are tested.

Applications/Analyze/test/testAnalyzeTool.cpp
```diff
@@ -175,6 +197,10 @@ void testTugOfWar(const string& dataFileName, const double& defaultAct) {
         // get active and passive forces given the default activation
         af = muscle.getActiveFiberForceAlongTendon(s);
         pf = muscle.getPassiveFiberForceAlongTendon(s);
+        fv = muscle.getFiberVelocityAlongTendon(s);
+
+        //assert(abs(fv) < SimTK::SqrtEps);
```
aseth1: Removed, as well as unused fv.

OpenSim/Actuators/Test/testMuscles.cpp
```diff
@@ -1108,10 +929,95 @@ void testDelp1990Muscle()
...
 
+    // muscle active, passive, total muscle and tendon force quantities
+    double af, pf, mf, tf, nfl, fv;
+    af = pf = mf = tf = nfl = fv = SimTK::NaN;
+
+    // Tolerance for muscle equilibrium solution 
+    const double equilTol = muscle.getMaxIsometricForce()*SimTK::SqrtEps;
+
+    // The maximum acceptable change in force between two contiguous states
+    const double maxDelta = muscle.getMaxIsometricForce() / 2;
```
aseth1: The same line appears in two places. The one in `testAnalyzeTool` was addressed and this one was overlooked. Thank you for seeing that!

OpenSim/Actuators/Thelen2003Muscle.cpp
aseth1: Added the comment and update the remainder of comments so they correspond identically (or so I hope).

OpenSim/Actuators/Thelen2003Muscle.cpp
aseth1: Definitely valid. Fell the through the cracks, but addressed now.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -1364,122 +1281,181 @@ Millard2012EquilibriumMuscle::estimateMuscleFiberState(
...
+        while (abs(ferr) >= abs(ferrPrev)) {
+            // Compute the Newton step
+            delta_lce = -h*ferrPrev / dferr_d_lce;
+            // Take a Newton Step if the step is nonzero
+            if (abs(delta_lce) > SimTK::SignificantReal)
+                lce = lcePrev + delta_lce;
+            else {
+                // We've stagnated or hit a limit; assume we are hitting local
+                // minimum and attempt to approach from the other direction.
+                lce = lcePrev - sign(delta_lce)*SimTK::SqrtEps;
```
aseth1: It seems to work for the Thelen2003Muscle. I haven't been able to get it to fail since I added this. From plotting the ferr curve as a function of lce, I was able to see that the solution was close, but not able to converge because of big changes in curvature (from very flat to steep or vice versa), in which case approaching from the other side seems to help. Furthermore, the curve changes with different fv, so re-evaluating the derivatives and fv helps. If we are not updating fv, it doesn't hit this condition.

aseth1: I agree we should have a more rigorous fix, I just don't know what that is.

OpenSim/Actuators/Test/testMuscles.cpp
```diff
@@ -968,34 +795,23 @@ void testMillard2012EquilibriumMuscle()
...
-    simulateMuscle(muscle, 
-        x0, 
-        act0, 
-        &motion, 
-        &control, 
-        IntegrationAccuracy,
-        InitializationTest,
-        InitializationTestTolerance,
-        false);
+    //LinearFunction motion(-4.0*OptimalFiberLength0, 2.0*OptimalFiberLength0);
```
aseth1: It was added temporarily and I meant to remove it. Thanks!

CHANGELOG.md
```diff
@@ -65,6 +65,12 @@ Converting from v3.x to v4.0
     - state.setTime(0.0);
 	- manager.integrate(state, 1.0);
 
+- `Muscle::equilibrate(SimTK::State&)` has been removed from the Muscle interface in order to reduce the plurality of muscle equilibrium methods. `Actuator::computeEquilibrium(SimTK::State&)` is overridden by Muscle and invokes pure virtual `Muscle::computeInitialFiberEquilibrium(SimTK::State&)`.
```
aseth1: Fixed.

CHANGELOG.md
```diff
@@ -88,8 +94,9 @@ Bug Fixes
...
+- Fixed a bug in the equilibrium solution of Millard and Thelen muscles, where the initial activation and fiber-length values (for solving for equilibrium) were always coming from the default values. This was unnecessary, because unless specified otherwise, the state automatically contains the default values. This fixes an issue where initial states activations from a file were not respected by the Forward Tool and instead, the initial activations would revert to the model defaults. (PR #272)
 - Fixed a bug where MuscleAnalysis was producing empty moment arm files. We now avoid creating empty Moment and MomentArm storage files when `_computeMoments` is False. (PR #324)
+- Fixed bug causing the muscle equilibrium solve routine in both Thelen2003Muscle and Millard2012EquilibriumMuscle to fail to converge and erroneously return the minimum fiber length. The fix added a proper reduction is step-size when errors increase and limiting the fiber-length to its minimum. (PR #1728)
```
aseth1: Done.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
aseth1: done.

chrisdembia:
Based on discussion with @aseth1 I will try to see if any tests slowed down on my local machine before we merge this.

aseth1:
Thanks @tkuchida and @chrisdembia for the quick re-review. I addressed the remaining issues. Please let me know if you have any additional comments. The CO failures appears to debug timing out. Thanks @chrisdembia for investigating.

chrisdembia:
Test times are roughly the same. Merging.

aseth1:
Great! Thank you!

tkuchida:
:fireworks:

Issue #1698, opened 2017-05-08
C3DFileAdapter: very slow file reading to Tables


nickbianco:
Reading a c3d file's data to a Table using the C3DFileAdapter is currently very slow. I compared loading a c3d file (1 minute of walking data) using @jimmyDunne's `btk_loadc3d()` MATLAB script to the C3DFileAdapter and got the following results after 3 runs (Windows 7, 64-bit, i5 2.50 GHz): 

btk in MATLAB: 0.39 +/- 0.07 seconds
C3DFileAdapter: 671.96 +/- 20.08 seconds

If you'd like to recreate the test, I can send you the file I used.
```matlab
import org.opensim.modeling.*

file = 'slk.c3d';

% Using btk directly in MATLAB (need btk MATLAB wrapper)
tic
data = btk_loadc3d(file);
disp('btk in MATLAB: ')
toc

disp(' ')
 
% Using new C3DFileAdapter in 4.0
tic
adapter = C3DFileAdapter();
tables = adapter.read(file);
disp('C3DFileAdapter: ')
toc

```

jimmyDunne:
Need to add a benchmark to the test case. Profile each step to see where issue is coming from.

jimmyDunne:
@aseth1 will profile

chrisdembia:
I did some profiling and it seems that the issue is the resizing that `DataTable_::appendRow()` does (this takes up 66% of the time in testC3DFileAdapter). I tried to pre-allocate the matrix underneath the DataTable, but I was not successful. @aseth1 let me know if you want to chat about this.

chrisdembia:
Here's how I got the number of rows and column:

```cpp
        int marker_num_frames = marker_pts->GetFrontItem()->GetFrameNumber();
        int marker_num_columns = marker_pts->GetItemNumber();
        marker_table.updMatrix().resize(marker_num_frames, marker_num_columns);
```

aseth1:
Thanks @chrisdembia for checking and suggesting a fix. I can confirm your findings in VS2015 profiler:
**87.1 %** -> `_depData.resizeKeep(_depData.nrow() + 1, _depData.ncol());`

chrisdembia:
Ah; it seems like your profiler was more helpful. Let me know if you want to chat more.

aymanhab:
PR in review

PR #1730, opened 2017-05-25
Fix slow C3DFileAdpaterReading  issue #1698

aseth1:
Fixes issue #1698 
`DataTable::appendRow()` is slow because it is resizing the underlying `SimTK::Matrix` on each call, which requires reallocation and copying of the entire matrix.

### Brief summary of changes
Added a protected `DataTable_` constructor that takes the independent vector, data matrix and labels so that `appendRow()` can be avoided.

`TimeSeriesTable_` exposes the convenience constructor and performs the validation and cleanup should validation fail.

`C3DFileAdpater::extendRead()` allocates the full time vector and data matrix first and then updates their elements and the uses the new constructor to create the `TimeSeriesTable` avoiding

### Testing I've completed
Profiled  `testC3DFileAdapter` prior to change the bottle neck was inside `DataTable_::appendRow()`: 
**87.1 %** -> _depData.resizeKeep(_depData.nrow() + 1, _depData.ncol()); (~1600ms)

After the change, the creation and updating of the matrix and construction of the full `TimeSeriesTable` in `C3DFileAdpater::extendRead()` now only uses **3.1 %** of the total runtime (~34ms).

And `testC3DFileAdapter` passes.

### CHANGELOG.md (choose one)
- no need to update because change required an update to implementation details.

### DIFF
--- a/OpenSim/Common/C3DFileAdapter.cpp
+++ b/OpenSim/Common/C3DFileAdapter.cpp
@@ -107,9 +107,7 @@ C3DFileAdapter::extendRead(const std::string& fileName) const {
         }
 
         double time_step{1.0 / acquisition->GetPointFrequency()};
-        for(int f = 0; 
-            f < marker_pts->GetFrontItem()->GetFrameNumber();
-            ++f) {
+        for(int f = 0; f < marker_pts->GetFrontItem()->GetFrameNumber(); ++f) {
             SimTK::RowVector_<SimTK::Vec3> row{marker_pts->GetItemNumber()};
             int m{0};
             for(auto it = marker_pts->Begin();  it != marker_pts->End(); ++it) {
@@ -120,12 +118,12 @@ C3DFileAdapter::extendRead(const std::string& fileName) const {
             }
 
             marker_matrix.updRow(f) = row;
-            marker_times[f] = 0 + f * time_step; // pt->GetValues();
+            marker_times[f] = 0 + f * time_step; //TODO: 0 should be start_time
         }
 
         // Create the data
         auto& marker_table = *new 
-            TimeSeriesTableVec3 (marker_times, marker_matrix, marker_labels);
+            TimeSeriesTableVec3(marker_times, marker_matrix, marker_labels);
 
         marker_table.
             updTableMetaData().
@@ -216,8 +214,9 @@ C3DFileAdapter::extendRead(const std::string& fileName) const {
         SimTK::Matrix_<SimTK::Vec3> force_matrix(nf, labels.size());
 
         double time_step{1.0 / acquisition->GetAnalogFrequency()};
-        for(int f = 0; f < force_times.size();  ++f) {
-            SimTK::RowVector_<SimTK::Vec3>
+
+        for(int f = 0; f <  static_cast<int>(nf);  ++f) {
+            SimTK::RowVector_<SimTK::Vec3> 
                 row{fp_force_pts->GetItemNumber() * 3};
             int col{0};
             for(auto fit = fp_force_pts->Begin(),
@@ -241,7 +240,7 @@ C3DFileAdapter::extendRead(const std::string& fileName) const {
                 ++col;
             }
             force_matrix.updRow(f) = row;
-            force_times[f] = 0 + f * time_step;
+            force_times[f] = 0 + f * time_step; //TODO: 0 should be start_time
         }
 
         auto&  force_table = 
--- a/OpenSim/Common/C3DFileAdapter.cpp
+++ b/OpenSim/Common/C3DFileAdapter.cpp
@@ -69,13 +69,19 @@ C3DFileAdapter::extendRead(const std::string& fileName) const {
     reader->Update();
     auto acquisition = reader->GetOutput();
 
+    EventTable event_table{};
+    auto events = acquisition->GetEvents();
+    for (auto it = events->Begin();
+        it != events->End();
+        ++it) {
+        auto et = *it;
+        event_table.push_back({ et->GetLabel(),
+            et->GetTime(),
+            et->GetFrame(),
+            et->GetDescription() });
+    }
+
     OutputTables tables{};
-    auto& marker_table = *(new TimeSeriesTableVec3{});
-    auto&  force_table = *(new TimeSeriesTableVec3{});
-    tables.emplace(_markers, 
-                   std::shared_ptr<TimeSeriesTableVec3>(&marker_table));
-    tables.emplace(_forces, 
-                   std::shared_ptr<TimeSeriesTableVec3>(&force_table));
 
     auto marker_pts = btk::PointCollection::New();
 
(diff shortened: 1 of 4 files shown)


OpenSim/Common/TimeSeriesTable.h
```diff
@@ -141,6 +141,29 @@ class TimeSeriesTable_ : public DataTable_<double, ETY> {
     TimeSeriesTable_& operator=(const TimeSeriesTable_&) = default;
     TimeSeriesTable_& operator=(TimeSeriesTable_&&)      = default;
     ~TimeSeriesTable_()                                  = default;
+
+    /** Convenience Constructor to efficiently populate a time series table
```
chrisdembia: Constructor -> constructor.

OpenSim/Common/TimeSeriesTable.h
```diff
@@ -141,6 +141,29 @@ class TimeSeriesTable_ : public DataTable_<double, ETY> {
...
+            validateDependentsMetaData();
+            for (size_t i = 0; i < indVec.size(); ++i) {
+                validateRow(i, indVec[i], depData.row(i));
+            }
+        }
+        catch (std::exception& x) {
+            // wipe out the data loaded if any
+            _indData.clear();
+            _depData.clear();
+            removeDependentsMetaDataForKey("labels");
```
chrisdembia: Would it make sense to rethrow?

OpenSim/Common/TimeSeriesTable.h
```diff
@@ -141,6 +141,30 @@ class TimeSeriesTable_ : public DataTable_<double, ETY> {
...
+    /** Convenience Constructor to efficiently populate a time series table
+    from available data. This is primarily useful for constructing with large
+    data read in from file without having to reallocate and copy memory.*/
+    TimeSeriesTable_(const std::vector<double>& indVec,
+        const SimTK::Matrix_<ETY>& depData,
+        const std::vector<std::string>& labels) : 
+            DataTable_<double, ETY>(indVec, depData, labels) {
+        try {
+            // Perform the validation of the data a TimeSeriesTable
+            this->validateDependentsMetaData();
```
chrisdembia: I see you added a comment in the base class about validation. Could you add a little more detail, either in the doxygen description for this constructor or for the one in the base class, explaining the situation with calling virtual functions from constructors?

OpenSim/Common/DataTable.h
```diff
@@ -1158,6 +1159,23 @@ class DataTable_ : public AbstractDataTable {
...
 protected:
+    /** Convenience Constructor to efficiently populate a data table from
+    known data. This is primarily useful for reading in large data tables
+    without having to reallocate and copy memory. NOTE derived tables
+    must validate the table according to the needs of the concrete type.*/
+    DataTable_(const std::vector<ETX>& indVec,
+        const SimTK::Matrix_<ETY>& depData,
+        const std::vector<std::string>& labels) {
+        OPENSIM_THROW_IF(indVec.size() != depData.nrow(), InvalidArgument,
+            "Length of independent column does not match number of rows of dependent data.");
```
chrisdembia: 80 columns?

OpenSim/Common/C3DFileAdapter.cpp
```diff
@@ -88,45 +94,53 @@ C3DFileAdapter::extendRead(const std::string& fileName) const {
...
+            for(auto it = marker_pts->Begin();  it != marker_pts->End(); ++it) {
                 auto pt = *it;
                 row[m++] = SimTK::Vec3{pt->GetValues().coeff(f, 0),
                                        pt->GetValues().coeff(f, 1),
                                        pt->GetValues().coeff(f, 2)};
             }
-            marker_table.appendRow(0 + f * time_step, row);
+
+            marker_matrix.updRow(f) = row;
+            marker_times[f] = 0 + f * time_step; // pt->GetValues();
```
chrisdembia: Could you either provide more detail about the comment or remove it?

OpenSim/Common/C3DFileAdapter.cpp
```diff
@@ -88,45 +94,53 @@ C3DFileAdapter::extendRead(const std::string& fileName) const {
...
         }
+
+        // Create the data
+        auto& marker_table = *new 
+            TimeSeriesTableVec3 (marker_times, marker_matrix, marker_labels);
```
chrisdembia: Space between Vec3 and (

OpenSim/Common/C3DFileAdapter.cpp
```diff
@@ -174,56 +188,35 @@ C3DFileAdapter::extendRead(const std::string& fileName) const {
...
+        const size_t nf = fp_force_pts->GetFrontItem()->GetFrameNumber();
+        
+        std::vector<double> force_times(nf);
+        SimTK::Matrix_<SimTK::Vec3> force_matrix(nf, labels.size());
 
         double time_step{1.0 / acquisition->GetAnalogFrequency()};
-        for(int f = 0;
-            f < fp_force_pts->GetFrontItem()->GetFrameNumber();
-            ++f) {
+        for(int f = 0; f < force_times.size();  ++f) {
```
chrisdembia: Consider using `nf` instead of `force_times.size()`.

OpenSim/Common/C3DFileAdapter.cpp
```diff
@@ -69,13 +69,19 @@ C3DFileAdapter::extendRead(const std::string& fileName) const {
...
+    for (auto it = events->Begin();
+        it != events->End();
+        ++it) {
+        auto et = *it;
+        event_table.push_back({ et->GetLabel(),
+            et->GetTime(),
+            et->GetFrame(),
+            et->GetDescription() });
+    }
+
```
chrisdembia: Is there a particular reason for moving this?

OpenSim/Common/C3DFileAdapter.cpp
tkuchida: A straight replacement would cause a compiler warning (type mismatch). Would need to do
```cpp
for (int f = 0; f < static_cast<int>(nf); ++f)
```

OpenSim/Common/C3DFileAdapter.cpp
aseth1: I thought it read better to get the events first, which are then added to both the marker and force tables that are created in the two code blocks below.  I also thought I might be able to extract the start time for the acquisition at some point.

aseth1:
Thank you @chrisdembia and @tkuchida. Merging.

aseth1:
@nickbianco and @jimmyDunne can you check that you are getting reasonable performance now on your big C3D files? Thanks!

nickbianco:
I re-ran my test case and I'm getting reasonable performance now.

aseth1:
> I re-ran my test case and I'm getting reasonable performance now.

Curious how it compares to using BTK directly, can you post the comparison as you did before in MATLAB?

nickbianco:
Pretty much comparable with the update:

btk in MATLAB: 0.287 +/- 0.0329 seconds
C3DFileAdapter: 0.315 +/- 0.0067 seconds

Issue #427, opened 2015-05-01
Warning about nonzero inertia for massless body is emitted even if inertia is zero.


chrisdembia:
For the DynamicWalker model I just got the warning "Body 'Pelvis_massless' is massless but nonzero inertia provided. Inertia reset to zero. Otherwise provide nonzero mass" even though this body has zero inertia.

nikorose87:
I've got the same error with some models

chrisdembia:
Thank you for reporting.

tkuchida:
This issue pertains to [`Body::getInertia()`](https://github.com/opensim-org/opensim-core/blob/4446724f558fb15a4fb4fd6a2dbd2df3c20f9d6a/OpenSim/Simulation/SimbodyEngine/Body.cpp#L124), which is performing checks that should probably reside in `Body::extendFinalizeFromProperties()` (and an `OPENSIM_THROW_IF_FRMOBJ(!isObjectUpToDateWithProperties(), ...)` should be at the top of `getInertia()`).

chrisdembia:
Related PR: #225

chrisdembia:
@nikorose87 I reproduced this warning with the model you provided in #1716:

```
Body 'talus_r' is massless but nonzero inertia provided. Inertia reset to zero. Otherwise provide nonzeromass.
Body 'toes_r' is massless but nonzero inertia provided. Inertia reset to zero. Otherwise provide nonzero mass.
Body 'talus_l' is massless but nonzero inertia provided. Inertia reset to zero. Otherwise provide nonzeromass.
Body 'toes_l' is massless but nonzero inertia provided. Inertia reset to zero. Otherwise provide nonzero mass.
```

I actually think your issue here is related to your issue in #1716, related to seeing commas where periods are expected, because the model you provided in #1716 has the following for the mass and inertia fields:

```xml
                    <!--The mass of the body (kg)-->
                    <mass>0,10000000000000001</mass>
                    <!--The location (Vec3) of the mass center in the body frame.-->
                    <mass_center>0 0 0</mass_center>
                    <!--The elements of the inertia tensor (Vec6) as [Ixx Iyy Izz Ixy Ixz Iyz] measured about the mass_center and not the body origin.-->
                    <inertia>0,0012396438758032777 0,0012396438758032777 0,0012396438758032777 0 0 0</inertia>
```

nikorose87:
The problem was solved in my case just changing the locale settings to US. Thank you.

[...]

PR #225, opened 2014-10-29
[WIP] Clarify nonzero mass/inertia message and behavior.

chrisdembia:
Running this code:

``` cpp
Model model;
State state = model.initSystem();
model.printDetailedInfo(state, std::cout);
```

produces an unnecessary warning message:

```
MODEL: 

ANALYSES (0)

BODIES (1)
body[0] = ground (mass: 0)Body 'ground' is massless but nonzero inertia provided. Iner
tia reset to zero. Otherwise provide nonzero mass.
 (inertia: [~[0,0,0]  ~[0,0,0]])

ACTUATORS (0)
numStates = 0
numCoordinates = 0
numSpeeds = 0
numActuators = 0
numBodies = 1
numConstraints = 0
numProbes = 0

STATES (0)
```

[Here](https://github.com/opensim-org/opensim-core/blob/20beb87ef317c69a4a1d11d4f12d18dcfc0e68ae/OpenSim/Simulation/SimbodyEngine/Body.cpp#L177) is where the message is generated. 

This PR only emits that message if the inertia property was indeed set to nonzero, and also modifies the inertia property.

[Review on Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/225)

PR #1731, opened 2017-05-25
Improve warning about nonzero inertia.

chrisdembia:
Fixes issue #427 

Related PR: #225

### Brief summary of changes

There is currently a warning for if the user specifies zero mass for a body with a nonzero inertia, but the if-statement for this error didn't actually check if the inertia was nonzero. Now, the if-statement checks the inertia first.

### Testing I've completed

Using the python bindings, I made sure that calling `initSystem()` on a model with massless bodies no longer emits this warning.

### CHANGELOG.md (choose one)

- no need to update because...this warning was introduced after 3.3.


chrisdembia:
Thank you for reviewing.

chrisdembia:
Thank you.

PR #1741, opened 2017-06-02
Fix crash of AbstractTools when XML has empty <model_file>

### LINKED ISSUE #697: Tools crash Matlab if model_file is not specified

chrisdembia:
Fixes #697. This PR is a redo of PR #793.

### Brief summary of changes

In general, the tools *do* throw exceptions if no model file is provided. However, you instead get a crash if you have specified `force_set_files` but no `model_file`. This PR causes the AbstractTools to throw an exception more generally if `model_file` is empty. The exception message tells users to use `aLoadModel=false` if they want to load a tool setup file that does not have a model file in it (which is a reasonable desire; for example, they may want to set the model programmatically in MATLAB).

### Testing I've completed

I reproduced the error with the gait2392 pipeline (as @jimmyDunne had done) and made a test out of these XML files.

@aymanhab : is it necessary to add `SWIG_DECLARE_EXCEPTION` for `loadModel()`?

### CHANGELOG.md (choose one)

- no need to update because this was a minor bug.

### DIFF
--- a/Bindings/Python/tests/test_basics.py
+++ b/Bindings/Python/tests/test_basics.py
@@ -105,10 +105,9 @@ class TestBasics(unittest.TestCase):
                     'std_subject01_walk1_ik.mot')), [], 5, 0)
 
     def test_deserialize_tool_with_empty_model_file(self):
-        # Previously, there was a bug where loading an AbstractTool (e.g.,
-        # ForwardTool) setup file with an empty model_file but with some
-        # force_set_files would cause a segfault. Now, we make sure to throw an
-        # exception if trying to construct the tool with an empty model_file.
+        # Ensure an exception is thrown when loading an AbstractTool (e.g.,
+        # ForwardTool) setup file with an empty model_file. In particular, we
+        # want to check the case where force_set_files is not empty.
         with self.assertRaises(RuntimeError):
             rra = osim.ForwardTool(os.path.join(test_dir,
                 'gait2392_setup_forward_empty_model.xml'))
--- a/Bindings/Python/tests/test_basics.py
+++ b/Bindings/Python/tests/test_basics.py
@@ -98,5 +98,26 @@ class TestBasics(unittest.TestCase):
         controller = osim.ToyReflexController()
         
     def test_GCVSplineSet(self):
-        splineset = osim.GCVSplineSet(os.path.join(test_dir, 'std_subject01_walk1_ik.mot'))
-        splineset = osim.GCVSplineSet(osim.TimeSeriesTable(os.path.join(test_dir, 'std_subject01_walk1_ik.mot')), [], 5, 0)
+        splineset = osim.GCVSplineSet(os.path.join(test_dir,
+            'std_subject01_walk1_ik.mot'))
+        splineset = osim.GCVSplineSet(
+                osim.TimeSeriesTable(os.path.join(test_dir,
+                    'std_subject01_walk1_ik.mot')), [], 5, 0)
+
+    def test_deserialize_tool_with_empty_model_file(self):
+        # Previously, there was a bug where loading an AbstractTool (e.g.,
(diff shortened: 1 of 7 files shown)


Bindings/Python/tests/test_basics.py
```diff
@@ -98,5 +98,26 @@ def test_ToyReflexController(self):
...
-        splineset = osim.GCVSplineSet(os.path.join(test_dir, 'std_subject01_walk1_ik.mot'))
-        splineset = osim.GCVSplineSet(osim.TimeSeriesTable(os.path.join(test_dir, 'std_subject01_walk1_ik.mot')), [], 5, 0)
+        splineset = osim.GCVSplineSet(os.path.join(test_dir,
+            'std_subject01_walk1_ik.mot'))
+        splineset = osim.GCVSplineSet(
+                osim.TimeSeriesTable(os.path.join(test_dir,
+                    'std_subject01_walk1_ik.mot')), [], 5, 0)
+
+    def test_deserialize_tool_with_empty_model_file(self):
+        # Previously, there was a bug where loading an AbstractTool (e.g.,
```
tkuchida: I don't think a historical record is necessary. Perhaps simply "Ensure an exception is thrown if…".

OpenSim/Simulation/Model/AbstractTool.cpp
```diff
@@ -382,28 +382,29 @@ getAnalysisSet() const
...
-            if (rOriginalForceSet!=NULL)
-                *rOriginalForceSet = model->getForceSet();
-        } catch(...) { // Properly restore current directory if an exception is thrown
-            IO::chDir(saveWorkingDirectory);
-            throw;
-        }
-        _model = model;
-        IO::chDir(saveWorkingDirectory);
+    OPENSIM_THROW_IF_FRMOBJ(_modelFile.empty(), Exception,
+            "No model file was specified (<model_file> tag is empty) in the "
```
tkuchida: I think "element" is more precise than "tag" (the tag is "<model_file>").

aymanhab:
LGTM @chrisdembia but I'd add the SWIG_THROW. What that does (at least on Java side) is to have the method signature indicate that it throws an exception so the call site must provide context to handle it rather than a generic something wrong happened and you're in limbo.

chrisdembia:
Thank you both for the speedy review. This is ready for review again.

> LGTM @chrisdembia but I'd add the SWIG_THROW. What that does (at least on Java side) is to have the method signature indicate that it throws an exception so the call site must provide context to handle it rather than a generic something wrong happened and you're in limbo.

I could not find any information on SWIG_THROW, but I did use the SWIG_DECLARE_EXCEPTION; it sounds like this is what you prefer.

aymanhab:
Sorry @chrisdembia 😊 I meant SWIG_DECLARE_EXCEPTION indeed, but I didn't remember the name of the MACRO correctly

aymanhab:
Do you think testControllers segfault is related, @chrisdembia ?

chrisdembia:
Thanks for raising that @aymanhab. I'm looking into it; I think we should hold off on merging until I can look into it further. Locally, testControllers is passing, so I'm not sure yet what's wrong.

chrisdembia:
I restarted the build and it passed now... I have created a separate issue for the sporadic failure on OSX (#1746).

chrisdembia:
Ready for review.

chrisdembia:
Thank you.

Issue #1722, opened 2017-05-19
AbstractDataTables cannot be used with SimTK::ClonePtr() 


chrisdembia:
Using `SimTK::ClonePtr` requires that a class `T` has a function `T* clone() const`. The AbstractDataTable class and its subclasses have a `clone()`, but it returns a `std::shared_ptr<AbstractDataTable>`. While this has the beneficial feature of making memory leaks less likely, it also means that we cannot use `SimTK::ClonePtr<TimeSeriesTable>`, etc.

aymanhab:
For a large table shared_ptr seems appropriate, need to understand the problem better. Please provide usecase @chrisdembia

PR #1742, opened 2017-06-02
Fix DataTable_<[...]>::clone() for bindings with swig 3.0.12.

chrisdembia:
### Brief summary of changes

This PR fixes SWIG bindings so they can be generated without error for SWIG 3.0.12. The issue arose from what seems to be a new bug in SWIG when extending class templates with more than one template parameter.

### Testing I've completed

The Java TestTables initially failed when I started using SWIG 3.0.12, but this test passes now.

### CHANGELOG.md (choose one)

- no need to update because DataTable did not exist in a previous release.


Bindings/Python/swig/python_simulation.i
```diff
@@ -68,7 +68,6 @@ MODEL_ADOPT_HELPER(ModelComponent);
 MODEL_ADOPT_HELPER(Body);
 MODEL_ADOPT_HELPER(Probe);
 MODEL_ADOPT_HELPER(Joint);
-MODEL_ADOPT_HELPER(Frame);
```
tkuchida: Unclear why these two lines were removed. No explanation found in commit notes or PR comment.

chrisdembia: Model no longer has a FrameSet or an `addFrame()` method.

tkuchida: Ah, ok. Thanks. :dromedary_camel:

chrisdembia:
> Not clear if new syntax still works with our min version 3.0.8.

Travis and AppVeyor test with 3.0.8, and the TestTables.java test ensures the cloning works correctly.

> More importantly if a method does "new" it should be marked as such with SWIG directive otherwise it causes a memory leak since the new instance is never released.

I ran the following Python code, and the the number printed did not change, which would seem to indicate there is no leak.

```py
import opensim
t = opensim.DataTable()
r = opensim.RowVector([1, 2])
t.appendRow(0, r)
for i in range(500): t.appendRow(i, r)
import resource
while True:
    x = t.clone()
    print(resource.getrusage(resource.RUSAGE_SELF).ru_maxrss)
```

Looking at the SWIG-generated code, it seems that 

```cpp
SWIGINTERN PyObject *_wrap_DataTable_clone(PyObject *SWIGUNUSEDPARM(self), PyObject *args) {
  PyObject *resultobj = 0;
  OpenSim::DataTable_< double,double > *arg1 = (OpenSim::DataTable_< double,double > *) 0 ;
  void *argp1 = 0 ;
  int res1 = 0 ;
  std::shared_ptr< OpenSim::DataTable_< double,double > const > tempshared1 ;
  std::shared_ptr< OpenSim::DataTable_< double,double > const > *smartarg1 = 0 ;
  PyObject * obj0 = 0 ;
  OpenSim::DataTable_< double,double > *result = 0 ;

  if (!PyArg_ParseTuple(args,(char *)"O:DataTable_clone",&obj0)) SWIG_fail;
  {
    int newmem = 0;
    res1 = SWIG_ConvertPtrAndOwn(obj0, &argp1, SWIGTYPE_p_std__shared_ptrT_OpenSim__DataTable_T_double_double_t_t, 0 |  0 , &newmem);
    if (!SWIG_IsOK(res1)) {
      SWIG_exception_fail(SWIG_ArgError(res1), "in method '" "DataTable_clone" "', argument " "1"" of type '" "OpenSim::DataTable_< double,double > const *""'");
    }
    if (newmem & SWIG_CAST_NEW_MEMORY) {
      tempshared1 = *reinterpret_cast< std::shared_ptr< const OpenSim::DataTable_<double,double> > * >(argp1);
      delete reinterpret_cast< std::shared_ptr< const OpenSim::DataTable_<double,double> > * >(argp1);
      arg1 = const_cast< OpenSim::DataTable_< double,double > * >(tempshared1.get());
    } else {
      smartarg1 = reinterpret_cast< std::shared_ptr< const OpenSim::DataTable_<double,double> > * >(argp1);
      arg1 = const_cast< OpenSim::DataTable_< double,double > * >((smartarg1 ? smartarg1->get() : 0));
    }
  }
  {
    try {
      result = (OpenSim::DataTable_< double,double > *)OpenSim_DataTable__Sl_double_Sc_double_Sg__clone((OpenSim::DataTable_< double,double > const *)arg1);
    } catch (const std::exception& e) {
      std::string str("std::exception in 'OpenSim::DataTable_< double,double > * OpenSim::DataTable_< double,double >::clone() const': ");
      std::string what(e.what());
      SWIG_exception_fail(SWIG_RuntimeError, (str + what).c_str());
    }
  }
  {
    std::shared_ptr<  OpenSim::DataTable_<double,double> > *smartresult = result ? new std::shared_ptr<  OpenSim::DataTable_<double,double> >(result SWIG_NO_NULL_DELETER_SWIG_POINTER_OWN) : 0;
    resultobj = SWIG_NewPointerObj(SWIG_as_voidptr(smartresult), SWIGTYPE_p_std__shared_ptrT_OpenSim__DataTable_T_double_double_t_t, SWIG_POINTER_OWN | SWIG_POINTER_OWN);
  }
  return resultobj;
fail:
  return NULL;
}
```
I am not sure how to interpret this, but I think it means that there is no leak.

Furthermore, `x.thisown` is `True`, which I think means that SWIG will make sure the C++ object gets deleted when the python object is deleted (http://www.swig.org/Doc1.3/Python.html).

chrisdembia:
I think the `%newobject *::clone;` lines in python_preliminaries.i and java_common.i take care of this.

chrisdembia:
An alternative is to revert `clone()` on DataTables to just return raw pointers, like with Object (https://github.com/opensim-org/opensim-core/issues/1722).

aymanhab:
Thanks for the investigation @chrisdembia Looks like our ::clone is kicking in based on your investigation. Less code disturbance at this point than redoing with raw pointers. Feel free to merge. Thanks @tkuchida

chrisdembia:
Thank you @aymanhab.

PR #1743, opened 2017-06-02
AbstractTool: Do not check address from reference.

chrisdembia:
### Brief summary of changes

Recent versions of clang generate the following warning:

image

To avoid the warning, I deleted the lines that tried to check the addresses for these references.

### Testing I've completed

I ran the test suite.

Is it okay to remove these lines of code? Is there a reason to suspect the entries of the set could be null? If not, I can create a function on Set `bool isNull(int index)`.

### CHANGELOG.md (choose one)

- no need to update because this does not affect users.


chrisdembia:
Thank you!

Issue #1693, opened 2017-05-04
finalizeFromProperties() usage and exception messages


jimmyDunne:
Since it is a new procedure, I forgot that we need to call `finalizeFromProperties()` immediately after reading a model into memory (Matlab). So I spent a solid half hour trying to figure out why I was getting exceptions when trying to work with coordinates. 

Doing this in Matlab, the exception was;

``
model.getCoordinateSet().get(0)

>>Java exception occurred:
>>java.lang.RuntimeException: ArrayPtrs.get: Array
>>index out of bounds.
>>	at   org.opensim.modeling.opensimSimulationJNI.SetCoordinates_get__SWIG_0(Native
>>        Method)
>>	at org.opensim.modeling.SetCoordinates.get(SetCoordinates.java:145) 
``
I don't find this helpful. It is also not helpful that asking the coordinate set for its size returns 0— it should return an exception. 

Perhaps there can be additions to the exception messages that suggest to the User to try `finalizeFromProperties`

tkuchida:
> It is also not helpful that asking the coordinate set for its size returns 0— it should return an exception.

:+1: Perhaps `OPENSIM_THROW_IF_FRMOBJ(!isObjectUpToDateWithProperties(), ...);` is needed.

aseth1:
Will need to come up with a solution that does no require adding `OPENSIM_THROW_IF_FRMOBJ(!isObjectUpToDateWithProperties(), ...);`  on every getter!

aseth1:
I think we are seeing that the practical benefits of automatically calling `finalizeFromProperties()` upon Model construction outweighs the inconsistency of having to invoke  `finalizeFromProperties()` manually upon cloning (which is a less common usage). Once we have bugs under control, I will look into how we can get the model to be finalized after cloning so we can put this issue to rest.

tkuchida:
> Will need to come up with a solution that does no require adding `OPENSIM_THROW_IF_FRMOBJ(!isObjectUpToDateWithProperties(), ...);`  on every getter!

Of course. Conceptually, the problem is that the user can interact with the CoordinateSet before it's meaningful. It should not be possible for the user to see a CoordinateSet of size zero unless the Model really can't move; before calling `finalizeFromProperties()`, the information given to the user should be that the size of the CoordinateSet is undefined (e.g., answer "I don't know" rather than "I know that the size is zero").

I agree with your second comment: ideally, the user would not be required to call `finalizeFromProperties()` after construction.

[...]

PR #1706, opened 2017-05-12
[WIP] Doxygen users guide

chrisdembia:
I've created this PR based on discussion with @aseth1 about investigating how to use Markdown for the API guide and simultaneously incorporate the API guide into doxygen. The result looks like this:

image

It may be helpful to use the Atom text editor locally to preview how the Markdown files would look on GitHub.

Here's another project on GitHub that does something similar: https://github.com/smtrat/smtrat

PR #1748, opened 2017-06-05
Automatic invocation of Model::finalizeFromProperties 1693

aseth1:
Fixes issue #1693 and reverses PR #1511

### Brief summary of changes
- Re-enabled `Model` constructors invocation of `finalizeFromProperties()` so users do not have to explicitly call it after loading a Model, which addresses the primary issue in #1693

- Implemented a custom override of `Muscle::clone()` which now invokes `finalizeFromProperties()` before returning the clone. This addresses a consistency issue between construction vs. cloning that #1511 attempted to address by requiring  `finalizeFromProperties()` to be called explicitly.

- Removed the now unnecessary calls to `Model::finalizeFromProperties()` in examples and test cases.

- There remains an inconsistency if using the default copy constructor, where `finalizeFromProperties()` must still be invoked manually. This can be mitigated by using `clone()` in all examples instead of the copy constructor.

### Testing I've completed
- examples and all tests cases passing.

### CHANGELOG.md (choose one)
- no need to update because the change is more consistent with the expected behavior. The need to explicitly call `finalizeFromProperties()` from #1511 was not documented.

### DIFF
--- a/OpenSim/Simulation/Model/Model.cpp
+++ b/OpenSim/Simulation/Model/Model.cpp
@@ -85,6 +85,7 @@ Model::Model() : ModelComponent(),
 {
     constructProperties();
     setNull();
+    finalizeFromProperties();
 }
 //_____________________________________________________________________________
 /*
@@ -102,6 +103,7 @@ Model::Model(const string &aFileName) :
     constructProperties();
     setNull();
     updateFromXMLDocument();
+    finalizeFromProperties();
 
     _fileName = aFileName;
     cout << "Loaded model " << getName() << " from file " << getInputFileName() << endl;
--- a/OpenSim/Simulation/Model/Model.cpp
+++ b/OpenSim/Simulation/Model/Model.cpp
@@ -103,17 +103,37 @@ Model::Model(const string &aFileName) :
     constructProperties();
     setNull();
     updateFromXMLDocument();
-    finalizeFromProperties();
 
     _fileName = aFileName;
     cout << "Loaded model " << getName() << " from file " << getInputFileName() << endl;
+
+    try {
+        finalizeFromProperties();
+    }
+    catch(const InvalidPropertyValue& err) {
+        std::cout << "WARNING: Model was unable to finalizeFromProperties "
+            "due to: " << err.what() <<
+            "\nUpdate the model file and reload OR update the property and call "
+            "finalizeFromProperties() on the model." << endl;
+    }
 }
 
 Model* Model::clone() const
 {
     // Invoke default copy constructor.
     Model* clone = new Model(*this);
-    clone->finalizeFromProperties();
+
+    try {
+        clone->finalizeFromProperties();
+    }
+    catch (const InvalidPropertyValue& err) {
+        std::cout << "WARNING: clone() was unable to finalizeFromProperties "
+            "due to: " << err.what() <<
+            "\nUpdate the model and call clone() again OR update the "
+            "clone's property and call finalizeFromProperties() on it."
+            << endl;
+    }
+
     return clone;
 }
 
--- a/OpenSim/Simulation/Model/Model.cpp
+++ b/OpenSim/Simulation/Model/Model.cpp
@@ -72,7 +72,7 @@ using namespace SimTK;
 // CONSTRUCTOR(S) AND DESTRUCTOR
 //=============================================================================
 //_____________________________________________________________________________
-/**
+/*
  * Default constructor.
  */
 Model::Model() : ModelComponent(),
@@ -87,7 +87,7 @@ Model::Model() : ModelComponent(),
     setNull();
 }
 //_____________________________________________________________________________
-/**
+/*
  * Constructor from an XML file
  */
 Model::Model(const string &aFileName) :
@@ -107,8 +107,16 @@ Model::Model(const string &aFileName) :
     cout << "Loaded model " << getName() << " from file " << getInputFileName() << endl;
 }
 
+Model* Model::clone() const
+{
+    // Invoke default copy constructor.
+    Model* clone = new Model(*this);
+    clone->finalizeFromProperties();
+    return clone;
+}
+
 //_____________________________________________________________________________
-/**
+/*
  * Override default implementation by object to intercept and fix the XML node
  * underneath the model to match current version
  */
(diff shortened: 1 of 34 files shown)


OpenSim/Simulation/Model/Model.h
```diff
@@ -254,6 +256,12 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Model, ModelComponent);
      */
     void cleanup();
 
+    /** Model clone() override that invokes finalizeFromProperties() 
+        on a default copy constructed Model, prior to returning the Model. */
+    Model* clone() const override;
+    
+    const std::string& getConcreteClassName() const override
+    {   return getClassName(); }
```
chrisdembia: Consider adding a comment like "This function is usually defined via `OpenSim_DECLARE_CONCRETE_OBJECT()`, but we do not use that macro for Model so that we can define a custom `clone()`. This comment could go here or at the top of the class where the macro usually appears.

aseth1: I put it at the top of the class where the macro is typically used.

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -105,10 +106,39 @@ Model::Model(const string &aFileName) :
...
     cout << "Loaded model " << getName() << " from file " << getInputFileName() << endl;
+
+    try {
+        finalizeFromProperties();
+    }
+    catch(const InvalidPropertyValue& err) {
+        std::cout << "WARNING: Model was unable to finalizeFromProperties "
+            "due to: " << err.what() <<
+            "\nUpdate the model file and reload OR update the property and call "
+            "finalizeFromProperties() on the model." << endl;
```
tkuchida: Minor hygienic issue: inconsistent use of `std::cout` on one side and simply `endl` on the other.

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -105,10 +106,39 @@ Model::Model(const string &aFileName) :
...
+
+    try {
+        clone->finalizeFromProperties();
+    }
+    catch (const InvalidPropertyValue& err) {
+        std::cout << "WARNING: clone() was unable to finalizeFromProperties "
+            "due to: " << err.what() <<
+            "\nUpdate the model and call clone() again OR update the "
+            "clone's property and call finalizeFromProperties() on it."
+            << endl;
```
tkuchida: Same as above.

OpenSim/Simulation/Model/Model.h
```diff
@@ -129,7 +129,14 @@ method, in which case it will allocate and maintain a ModelVisualizer.
 **/
 
 class OSIMSIMULATION_API Model  : public ModelComponent {
-OpenSim_DECLARE_CONCRETE_OBJECT(Model, ModelComponent);
+// Note, a concrete object typically employs the OpenSim_DECLARE_CONCRETE_OBJECT
+// macro, but, for Model we do not want its clone() method to be auto-generated.
+// Instead, we want to override and customize it and so we employ the subset of
+// macros that OpenSim_DECLARE_CONCRETE_OBJECT calls, excluding one that
+// implements clone() and getConcreteClassName(), which are implemented below.
```
tkuchida: :+1: Very helpful comment.

OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp
```diff
@@ -85,7 +85,7 @@ int main()
             } //Ignore the validity of the property values
             // TODO this should specifically handle "InvalidPropertyValue" exceptions
             // once we have that in place.
-            catch (const std::exception&) {
+            catch (const InvalidPropertyValue&) {
```
tkuchida: The "TODO" comment can now be deleted.

tkuchida:
Seems fine, but I think there's some additional cleanup we will need to do in the documentation, Developer's Week API Guide, MATLAB/Python scripts and tests, etc. I suggest searching the repo for all instances of `finalizeFromProperties`; there shouldn't be many left.

aseth1:
> I suggest searching the repo for all instances of finalizeFromProperties; there shouldn't be many left.

I removed the unnecessary calls from Java, Python and Matlab tests. The search on GitHub did not turn up any uses in examples. As far as I could tell, `initSystem()` is correctly being used by the examples.

aseth1:
> I think there's some additional cleanup we will need to do in the documentation, Developer's Week API Guide ...

I will address this in the *Users Guide* WIP (#1706) that I still need to update.

aseth1:
@tkuchida and thanks for the quick review. I hope I addressed your comments. @chrisdembia did you have more comments to add?

tkuchida:
> I hope I addressed your comments.

:+1: Looks great :cow:

OpenSim/Simulation/Test/testModelInterface.cpp
```diff
@@ -35,14 +35,6 @@ int main() {
...
-        // finalizeFromProperties() is required to build internal ownership tree
-        // attempt to access the ComponentList will throw that the model (root)
-        // has no subcomponents
-        ASSERT_THROW(ComponentIsRootWithNoSubcomponents, 
-            model.countNumComponents());
-
-        // finalize internal data structures from its properties
-        model.finalizeFromProperties();
 
         // all subcomponents are now accounted for.
```
chrisdembia: Consider adding a note that this is because `finalizeFromProperties()` is called in the constructor.

OpenSim/Simulation/Wrap/WrapCylinder.cpp
```diff
@@ -125,7 +125,7 @@ void WrapCylinder::extendFinalizeFromProperties()
     {
         string errorMessage = "Error: radius for WrapCylinder " + getName() +
             " was either not specified, or is negative.";
-        throw Exception(errorMessage);
+        OPENSIM_THROW_FRMOBJ(InvalidPropertyValue, errorMessage);
```
chrisdembia: The `errorMessage` argument is passed as `InvalidPropertyValue`'s `propertName` parameter, so the resulting message will not be correct. I think `InvalidPropertyValue` should be updated to take a general message in addition to the property name (also, the `propertName` parameter has a typo in the name).

Lastly, I don't think the message passed to `InvalidPropertyValue` should include `Error: `. 

These comments apply to all similar changes in this PR.

aseth1:
@chrisdembia let me know if your comments have been adequately addressed.

aseth1:
CI test (`testWrapping`) in debug failed due to timeout. I am pretty sure it is unrelated.

chrisdembia:
The PR builds passed, which is more important I think. Merging.

aseth1:
Thank you!

PR #1750, opened 2017-06-07
Fix formatting in README

tkuchida:
Before:
before

After:
after


tkuchida:
Thanks for the merge. :tropical_fish:

PR #1769, opened 2017-06-18
Fix std::length_error in Component::printSubcomponentInfo()

chrisdembia:
### Brief summary of changes

Previously, `Component::printSubcomponentInfo()` gave an `std::length_error` if
called on a component with no subcomponents and whose concrete class name was
longer than 4 characters. This occurred due to attempting to create an
`std::string` of length `maxlen - getConcreteClassName().length()`, which, in
the case above, evaluated to `4 - 10` and caused an overflow.

In this PR, I introduced a separate "no subcomponents" message if there are no
subcomponents, and also handle the situation where the concrete class name of *this*
component is longer than that of all subcomponents.

Here is an example of the previous behavior:

```
>> hopper.getCoordinateSet.get('yCoord').printSubcomponentInfo()
Class name and absolute path name for descendants of 'yCoord' that are of type
Component:

Java exception occurred:
java.lang.RuntimeException: basic_string
    at org.opensim.modeling.opensimCommonJNI.Component_printSubcomponentInfo(Native Method)
    at org.opensim.modeling.Component.printSubcomponentInfo(Component.java:238)
```

Apoorva and Ajay reported this bug during the dry run for the TGCS 2017
workshop, though they reported that the call would make MATLAB crash (not
simply throw an exception).

### Testing I've completed

I ensured locally that, for the line above, no exception is thrown and that the message "Component '\<coord-name\>' has no subcomponents." is printed.

### CHANGELOG.md

- no need to update because printSubcomponentInfo() is new in 4.0


carmichaelong:
LGTM too. Looks like to me all of Travis's errors were timeouts. I'm okay with either merging as is or restarting tests if others are more inclined to do that.

aseth1:
Looks safe to merge.

Issue #1770, opened 2017-06-18
TableReporter gives error if it has 0 inputs.


chrisdembia:
The following code (in MATLAB) causes the given error:

```
>> m = Model();
>> r = TableReporter();
>> m.addComponent(r);
>> m.initSystem();
Java exception occurred:
java.io.IOException: Key = labels
	In file
        /Users/chris/repos/opensim/16/opensim-core/OpenSim/Common/AbstractDataTable.h:390
	In function 'setColumnLabels'
	at org.opensim.modeling.opensimSimulationJNI.Model_initSystem(Native Method)
	at org.opensim.modeling.Model.initSystem(Model.java:835) 
```

This is because `AbstractDataTable::setColumnLabels()` does not allow an empty list of column labels. As a consequence, TableReporters with no inputs (no calls to `addToReport()`) generate an exception message. However, I think it's fine for a TableReporter to have no inputs (in this case, nothing would be reported).

@jenhicks reported this bug during the dry run for the TGCS 2017 workshop.

tkuchida:
> I think it's fine for a TableReporter to have no inputs (in this case, nothing would be reported).

Seems like this situation would arise if the user forgot to call `addToReport()`, in which case a warning might be nice. Either solution would be better than a java.io.IOException.

tkuchida:
Fixed by #1771 (merged).

PR #1771, opened 2017-06-19
TableReporter with no connectees no longer causes exception.

chrisdembia:
Fixes issue #1770 

### Brief summary of changes

This PR allows one to have a TableReporter that has no connectees. If this is the case, a warning message is emitted.

I found a bug in Simbody, which has been fixed via https://github.com/simbody/simbody/pull/565. However, I am not relying on this fix in this PR (I used a simple workaround with `resize()`), since I do not want to force other members of the dev team to rebuild Simbody at this time.

Lastly, I updated a few of the DataTable exceptions to provide more helpful messages, especially in MATLAB.

### Testing I've completed

Implemented a test in testComponentInterface.

### CHANGELOG.md

- no need to update because TableReporter has yet to be released.


OpenSim/Common/Reporter.h
```diff
@@ -228,7 +234,14 @@ OpenSim_DECLARE_CONCRETE_OBJECT_T(TableReporter_, InputT, Reporter<InputT>);
...
             labels.push_back( input.getLabel(idx) );
         }
-        const_cast<Self*>(this)->_outputTable.setColumnLabels(labels);
+        if (!labels.empty()) {
+            const_cast<Self*>(this)->_outputTable.setColumnLabels(labels);
+        } else {
+            std::cout << "Warning: No outputs were connected to '"
+                      << this->getName() << "' of type "
+                      << getConcreteClassName()
+                      << "; did you intend to report any outputs?" << std::endl;
```
tkuchida: The last piece of the message doesn't seem to add anything. Perhaps tell the user "You can connect outputs by calling `addToReport()`"?

chrisdembia: Good idea; done.

tkuchida: Looks great :+1:

chrisdembia:
I merged with master, which should fix the travis failure.

carmichaelong:
Thanks for fixing the travis failure. In general LGTM. At first I was hesitant about throwing exceptions when printing an empty table (rather than just a warning), but 1) I think it's easy to take a wait and see approach; and 2) a change to that doesn't belong in this PR. I will merge now.

chrisdembia:
Thanks @carmichaelong. I agree that toString() shouldn't throw the EmptyTable exception. I've created a separate issue for that.

PR #1776, opened 2017-06-20
Fix the method Model.replaceMarkerSet so that it calls initSystem 

aymanhab:
Fix the method Model.replaceMarkerSet so that it calls initSystem after removing old markers and adding new markers to model's Properties. Signature needs to change so state is not const so that we don't hang on to stale/bad state downstream.

Fixes issue #<issue_number>

### CHANGELOG.md (choose one)

- updated...

### DIFF
--- a/OpenSim/Simulation/Model/Model.cpp
+++ b/OpenSim/Simulation/Model/Model.cpp
@@ -1686,7 +1686,7 @@ void Model::writeMarkerFile(const string& aFileName)
  * @param aMarkerSet The new marker set to copy.
  * @return Number of markers that were successfully added to the model.
  */
-int Model::replaceMarkerSet(const SimTK::State& s, const MarkerSet& aMarkerSet)
+int Model::replaceMarkerSet(SimTK::State& s, const MarkerSet& aMarkerSet)
 {
     int i, numAdded = 0;
 
@@ -1700,7 +1700,8 @@ int Model::replaceMarkerSet(const SimTK::State& s, const MarkerSet& aMarkerSet)
         upd_MarkerSet().adoptAndAppend(marker);
         ++numAdded;
     }
-
+    // Reconnect the model now that we added aMarkerSet to the Property
+    s = initSystem();
     cout << "Replaced marker set in model " << getName() << endl;
     return numAdded;
 }
--- a/OpenSim/Simulation/Model/Model.h
+++ b/OpenSim/Simulation/Model/Model.h
@@ -883,7 +883,7 @@ public:
     //--------------------------------------------------------------------------
     MarkerSet& updMarkerSet() { return upd_MarkerSet(); }
     const MarkerSet& getMarkerSet() const { return get_MarkerSet(); }
-    int replaceMarkerSet(const SimTK::State& s, const MarkerSet& aMarkerSet);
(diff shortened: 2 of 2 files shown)


OpenSim/Simulation/Model/Model.cpp
```diff
@@ -1686,7 +1686,7 @@ void Model::writeMarkerFile(const string& aFileName)
  * @param aMarkerSet The new marker set to copy.
  * @return Number of markers that were successfully added to the model.
  */
-int Model::replaceMarkerSet(const SimTK::State& s, const MarkerSet& aMarkerSet)
+int Model::replaceMarkerSet(SimTK::State& s, const MarkerSet& aMarkerSet)
```
aseth1: The state as an argument seems completely unnecessary for this method. I would prefer it was removed altogether. This should be considered a properties edit `initSystem()` should not be invoked in here.

aymanhab: I also noticed that state  isn't used now (before my change) don't know the history. Generally models in the GUI are backed by a State all the time so that their transformations can be extracted etc. This method is called from Context object that maintains the state of the current model and needs to keep it up to date. 
I'm happy to replace initSystem with any other call that keeps the State up to date with Property changes due to removing Markers and adding new ones. Old method is obviously either wrong (since it doesn't connect the model after the replace) or assumes another call before or after which is a bad practice anyway.

aseth1: The current workflow in the `GenericModelMaker` is to call `Model::updateMarkerSet(newMarkerSet)` which appears to be performing a union of the existing and new markers.  I don't think we should have both methods.

aseth1: > I'm happy to replace initSystem with any other call that keeps the State up to date with Property changes due to removing Markers and adding new ones.

editting should not invoke initSystem. This would be like a property editing invoking initSystem. Once the edits are complete and the Tool should invoke initSystem(). I see you had updated `Model::updateMarkerSet()` to invoke `initSystem()` and I should have caught that earlier.

aymanhab: Thanks for the review @aseth. The workflow of the GUI is different from invoking a setup file and executing it end to end because users expect to see the effects of changing markerSets immediately (e.g. in computing measurements or other places in the GUI). So the GUI expects to have a valid model with a consistent state after replacing the markerSet. If that's not recommended inside the method, I'm happy to make the call from OpenSimContext after the update/replace call, however we need to unify these two methods (update/replace) to make sure the API and GUI are executing the same workflow.

aseth1: > If that's not recommended inside the method, I'm happy to make the call from OpenSimContext

That's correct.

> we need to unify these two methods (update/replace) to make sure the API and GUI are executing the same workflow

I agree. PR #1783 makes them consistent in that they both now do not call initSystem().

aseth1: Also I can augment #1783 so that `replaceMarkerSet` no longer takes a `State`. That would further improve the consistency of the two methods.

carmichaelong:
We decided that initSystem() would be called from OpenSimContext outside the ReplaceMarkerSet method. Related #1781.

Issue #1781, opened 2017-06-20
Model::updateMarkerSet() and replaceMarkerSet() should not invoke initSystem()


aseth1:
`Model::updateMarkerSet()` and `replaceMarkerSet()` are helpers for model (properties) editing. The Tool should invoke `Model::initSystem()` when it is done editing and ready to compute.

PR #1783, opened 2017-06-20
Remove initSystem() call from Model::updateMarkerSet() #1781

aseth1:
Fixes issue #1781

### Brief summary of changes
testScale was relying on a side-effect of updateMarkerSet() which was not necessary. Simply invoking initSystem() after editing the model markers works.

### Testing I've completed
testScale (and all tests) passed locally

### CHANGELOG.md (choose one)
- no need to update because no API change


aseth1:
@aymanhab @carmichaelong we reviewed in the meeting today, but still would be good to get your ✔️ and have one of you merge it.

PR #1787, opened 2017-06-22
Move hopper example simulate() to new SimulationUtilities.h

aseth1:
Fixes the issue that the `simulate()` method in the `helperMethods.h` is broadly useful/applicable to users and testers (e.g. I found myself writing something similar) and it should be accessible.

### Brief summary of changes
Besides the move, `simulate()` was adapted to take an arbitrary final time, return the final state, and to save a states file according to the model name (no longer hopper specific). The `simulateOnce` flag/option was removed because it seemed unnecessary when one can *ESC* to avoid replay and if visualization is off it only performs a single simulation anyways.

### Testing I've completed
Locally, ran the hopper example with and without answers with no issues.

Is the name and location of the `SimulationUtilities.h` agreeable?

### CHANGELOG.md (choose one)
- no need to update because it does not change the API and the example has not been released publicly.


carmichaelong:
Looks like `exampleHopperDevice.cpp` still has a `#include helperMethods.h` at [L41](https://github.com/opensim-org/opensim-core/pull/1787/files#diff-4ac22546bfb31c7d27557a2e8413a520L41).

OpenSim/Simulation/SimulationUtilities.h
```diff
@@ -0,0 +1,103 @@
...
+        state = initialState;
+        // Set up manager and simulate.
+        SimTK::RungeKuttaMersonIntegrator integrator(model.getSystem());
+        Manager manager(model, integrator);
+        state.setTime(0.0);
+        manager.integrate(state, finalTime);
+
+        // Save the states to a storage file (if requested).
+        if (saveStatesFile) {
+            manager.getStateStorage().print(model.getName() + ".sto");
```
chrisdembia: I suggest `<model_name>_states.sto`.

OpenSim/Simulation/SimulationUtilities.h
```diff
@@ -0,0 +1,103 @@
...
+
+namespace OpenSim {
+
+/**----------------------------------------------------------------------------
+Simulate a model from an initial state and return the final state.
+If the model's useVisualizer flag is true, the user is repeatedly prompted to
+either begin simulating or quit. The provided state is not updated but the
+final state is returned at the end of the simulation, when finalTime is reached.
+Set saveStatesFile=true to save the states to a storage file as:
+"<model_name>.sto". */
```
chrisdembia: Consider adding that one reason to use this over the Manager is to easily and visualize and replay a simulation.

OpenSim/Simulation/SimulationUtilities.h
```diff
@@ -0,0 +1,103 @@
...
+            // Get a key press.
+            silo->clear(); // Ignore any previous key presses.
+            unsigned key, modifiers;
+            silo->waitForKeyHit(key, modifiers);
+            if (key == SimTK::Visualizer::InputListener::KeyEsc) { break; }
+        }
+        else if (!simulateOnce) {
+            std::cout << "Press <Enter> to begin simulating, or 'q' followed "
+                << "by <Enter> to quit . . . " << std::endl;
+            if (std::cin.get() == 'q') { break; }
```
chrisdembia: If only simulating once, I think we do not need to wait for user input to start simulating, and we can avoid the logic around 'q'.

OpenSim/Examples/ExampleHopperDevice/buildTestbedModel.cpp
```diff
@@ -1,7 +1,5 @@
-#ifndef _helperMethods_h_
-#define _helperMethods_h_
 /* -------------------------------------------------------------------------- *
- *                         OpenSim:  helperMethods.h                          *
+ *                   OpenSim:  buildTestbedModel.cpp                          *
```
tkuchida: The file used to contain several other methods and had a different flavor. It makes sense to change the name given the current contents of the file.

chrisdembia: Was there a reason why it was a header file as opposed to a cpp file like the others?

tkuchida: I'm sure someone had a reason and there was probably a discussion about it. I don't think it's important. Given the current contents of the file, I don't think it should be a header file.

OpenSim/Examples/ExampleHopperDevice/exampleHopperDevice.cpp
```diff
@@ -174,22 +174,16 @@ int main(int argc, char* argv[]) {
     // Following are arguments to this executable:
     // (1) noVisualizer -- [Optional] Do not show visualizer. Default
     //                       behavior is to show visualizer.
-    // (2) simulateOnce -- [Optional] Simulate once and exit the program.
```
tkuchida: Suggest cleaning up "Following are arguments to this executable:" list since there is now only 1 argument.

chrisdembia:
I added a comment above "It would be great to test that this works well in Java/Python." To add to this, could you test that the MATLAB example still works with this new function of the same name? I know they have different capitalizations, but I'd like to know for sure whether or not there's a confilct.

tkuchida:
:+1: My comments have been addressed.

aseth1:
> To add to this, could you test that the MATLAB example still works with this new function of the same name?

I confirmed that MATLAB example uses `Simulate.m` and not the utility function.

OpenSim/OpenSim.h
```diff
@@ -25,6 +25,8 @@
 
 #include "Common/osimCommon.h"
 #include "Simulation/osimSimulation.h"
+#include "Simulation/SimulationUtilities.h"
```
chrisdembia: I think this include should be in `Simulation/osimSimulation.h`, *not* this file.

aseth1:
> Consider adding something to the main doxygen page, or adding a note somewhere in the API Guide, or just creating a new issue for documenting this new function.

I added `simulate()` in the README example to replace the manager and integrator calls. Is that sufficient visibility?

README.md
```diff
@@ -101,6 +103,7 @@ int main() {
     // Configure the visualizer.
     model.updMatterSubsystem().setShowDefaultGeometry(true);
     Visualizer& viz = model.updVisualizer().updSimbodyVisualizer();
+    viz.setBackgroundType(viz.SolidColor);
     viz.setBackgroundColor(White);
 
     // Simulate.
```
chrisdembia: @aseth1 please update this to be consistent with `testREADME.cpp`.

aseth1:
@chrisdembia I believe the README and test are now identical.

PR #1789, opened 2017-06-22
Fix MarkerPlacer time range errors for the average pose

aseth1:
Fixes point 4 of issue [GUI 183](https://github.com/opensim-org/opensim-gui/issues/183) related to the `MarkerPlacer`'s `TimerSeriesTable` of marker data handling of  an invalid time range when computing the average of marker data.

`TimerSeriesTable` is correct in throwing `TimeOutOfRange` when the range over which the average is to be computed exceeds the range of the data. However, our users are likely accustomed to providing large time ranges as away of ensuring the full time range of their data is used. MarkePlacer no *clamps* to the valid time range of the data if the range provide exceeds the range.

### Brief summary of changes
- made the `TimeSeriesTable::TimeOutOfRange` easier to understand 
- fixed a bug in order of the arguments passed to the `TimeOutOfRange` exception
- enable`MarkePlacer` to take up the slop when users provide large time-ranges as away of capturing all the time frames in their data.

### Testing I've completed
- all tests (including testScale) still pass
- ran isb2017 scale example successfully

### CHANGELOG.md (choose one)
- no need to update because this is just a bug fix

### DIFF
--- a/OpenSim/Common/TimeSeriesTable.h
+++ b/OpenSim/Common/TimeSeriesTable.h
@@ -105,8 +105,8 @@ public:
                    const double max) :
         Exception(file, line, func) {
         std::string msg = "Time " + std::to_string(time) + 
-            " is out of time range {" + std::to_string(min) +
-            ", " + std::to_string(max) + "}";
+            " is out of time range [" + std::to_string(min) +
+            ", " + std::to_string(max) + "]";
 
         addMessage(msg);
     }
--- a/OpenSim/Common/TimeSeriesTable.h
+++ b/OpenSim/Common/TimeSeriesTable.h
@@ -104,9 +104,9 @@ public:
                    const double min,
                    const double max) :
         Exception(file, line, func) {
-        std::string msg = "min = " + std::to_string(min);
-        msg += " max = " + std::to_string(max);
-        msg += " time = " + std::to_string(time);
+        std::string msg = "Time " + std::to_string(time) + 
+            " is out of time range {" + std::to_string(min) +
+            ", " + std::to_string(max) + "}";
 
         addMessage(msg);
     }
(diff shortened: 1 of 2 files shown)


aymanhab:
LGTM, feel free to merge when the tests pass.

OpenSim/Common/TimeSeriesTable.h
```diff
@@ -343,11 +344,11 @@ class TimeSeriesTable_ : public DataTable_<double, ETY> {
         OPENSIM_THROW_IF(beginTime < timeCol.front() ||
                          beginTime > timeCol.back(),
                          TimeOutOfRange,
-                         timeCol.front(), timeCol.back(), beginTime);
+                         beginTime, timeCol.front(), timeCol.back(),);
         OPENSIM_THROW_IF(endTime < timeCol.front() ||
                          endTime > timeCol.back(),
                          TimeOutOfRange,
-                         timeCol.front(), timeCol.back(), endTime);
+                         endTime, timeCol.front(), timeCol.back());
```
chrisdembia: Is this exception being misused anywhere else?

OpenSim/Common/TimeSeriesTable.h
```diff
@@ -104,9 +104,9 @@ class TimeOutOfRange : public Exception {
                    const double min,
                    const double max) :
         Exception(file, line, func) {
-        std::string msg = "min = " + std::to_string(min);
-        msg += " max = " + std::to_string(max);
-        msg += " time = " + std::to_string(time);
+        std::string msg = "Time " + std::to_string(time) + 
+            " is out of time range {" + std::to_string(min) +
+            ", " + std::to_string(max) + "}";
```
chrisdembia: I think it is more correct to use `[min, max]`.

apoorvar: I like the square instead of curly brackets. Otherwise seems fine to me.

OpenSim/Common/TimeSeriesTable.h
aseth1: I checked and only these two were incorrect.

aseth1:
I am uncertain as to what is causing the timeouts, but restarting in some cases finishes in under 38min. These fluctuations seem to be independent of the code changes.

Issue #650, opened 2015-10-01
Share MathJax between user and developer doxygen


jenhicks:
Otherwise GUI installer for a release has two big folders with MathJax + doxygen html files.

chrisdembia:
Related: #1796

tkuchida:
Upon merge of #1796, duplication will be <5 Mb so I think this issue will then no longer be relevant.

PR #1796, opened 2017-06-24
Update MathJax, reduce bloat.

chrisdembia:
### Brief summary of changes

1. Reduce the size of MathJax that ends up in our installation. Without this PR, the html_developer dir is about 260 MB; after this PR, it's only about 105 MB. I followed MathJax's guidelines for which files can be deleted from MathJax.
2. I updated MathJax from 2.5 to 2.7, which fixes the vertical line that appears in equations (https://stackoverflow.com/questions/34277967/chrome-rendering-mathjax-equations-with-a-trailing-vertical-line).

### Testing I've completed

1. Built the doxygen on macOS and Windows and made sure the equations for the Millard muscle appeared correctly.

### CHANGELOG.md (choose one)

- no need to update because use of MathJax was introduced after 3.3.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -29,7 +29,7 @@ if(${DOXYGEN_FOUND})
 
     # There are two ways of using MathJax: from the internet, or locally.
     # We prefer using MathJax locally, but we have to download it first.
-    # We download a 50 mb MathJax zip and unzip it (150 mb).
+    # We download a 34 mb MathJax zip and unzip it (150 mb).
     # The zip can be thrown away, but the 150 mb MathJax directory must stay
     # with the Doxygen documentation.
 
@@ -40,7 +40,7 @@ if(${DOXYGEN_FOUND})
     # This block of commands is all executing during the configuring phase.
     if(${OPENSIM_DOXYGEN_USE_MATHJAX} AND NOT EXISTS "${MATHJAX_ZIP_PATH}")
         message(STATUS "Trying to download MathJax for Doxygen...")
-        file(DOWNLOAD https://github.com/mathjax/MathJax/archive/v2.5-latest.zip
+        file(DOWNLOAD https://github.com/mathjax/MathJax/archive/2.7.0.zip
             "${MATHJAX_ZIP_PATH}"
             #SHOW_PROGRESS Clutters the CMake output.
             STATUS MATHJAX_DOWNLOAD_STATUS
@@ -69,7 +69,7 @@ if(${DOXYGEN_FOUND})
 
         # See Doxyfile.in for a description of these variables.
         set(OPENSIM_DOXYGEN_USE_MATHJAX_DOXYFILE "YES")
-        set(OPENSIM_DOXYGEN_MATHJAX_RELPATH "MathJax-2.5-latest")
+        set(OPENSIM_DOXYGEN_MATHJAX_RELPATH "MathJax-2.7.0")
 
     else()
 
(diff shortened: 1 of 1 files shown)


doc/CMakeLists.txt
```diff
@@ -153,14 +153,48 @@ if(${DOXYGEN_FOUND})
...
+            set(mathjax_unzipped_dir "${html_binary_dir}/MathJax-2.7.0")
 
-            # This creates a folder html_${audience}/MathJax-2.5-latest in the
+            # This creates a folder html_${audience}/MathJax-2.7-latest in the
             # build tree.
             add_custom_command(
                 OUTPUT "${mathjax_unzipped_dir}"
                 COMMENT "Unzipping MathJax.zip for ${audience} documentation"
                 COMMAND ${CMAKE_COMMAND} -E tar -x "${MATHJAX_ZIP_PATH}" 
+                # Trim away unnecessary files; 170 MB -> under 5 MB.
```
tkuchida: The PR comment says "Without this PR, the html_developer dir is about 150MB; after this PR, it's only about 105MB." It sounds like you have reduced the size by 45 Mb, which would suggest "170 Mb -> 125 Mb" on L164. Why only "5 MB"?

doc/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ if(${DOXYGEN_FOUND})
 
     # There are two ways of using MathJax: from the internet, or locally.
     # We prefer using MathJax locally, but we have to download it first.
-    # We download a 50 mb MathJax zip and unzip it (150 mb).
+    # We download a 34 mb MathJax zip and unzip it (150 mb).
```
tkuchida: L164 says "170 MB", not 150.

chrisdembia:
> (1) Not sure it's necessary to state the size of the .zip, etc., given that these sizes will likely continue to change as new versions of MathJax are released. Might be easier to maintain and avoid discrepancies in the future if sizes are deleted.

I agree the exact numbers are not necessary, but I think conveying the importance of the bloat reduction is good, otherwise maintainers might just remove the bloat reduction. I'll replace the exact numbers with stuff like `> 100 MB`.

> (2) Another bloat-reduction strategy is to share MathJax between developer and user versions of doxygen (#650). Is this PR intended to reduce the size of the MathJax directory so much that duplicating the directory doesn't matter anymore?

Oh I forgot about that issue. To your question, I'd say yes. The `html_developer` directory is 105 MB, only 4.1 MB of which is MathJax. The ability for the two folders to work in isolation (to be decoupled/have no dependencies) is worth 2*4.1 MB, IMO.

I think we should only distribute `html_user` in the distribution. Developers can build the developer doxygen from source, and we can host it online.

doc/CMakeLists.txt
chrisdembia: I corrected the PR description.

chrisdembia:
@tkuchida thanks for the review. I've updated the PR and this is ready for review again.

tkuchida:
> only 4.1 MB of which is MathJax

Thank you for clarifying. Duplicating ~5 Mb between developer and user documentation seems fine.

chrisdembia:
I restarted one of the travis-ci/pr build to ensure that doxygen still gets uploaded to myosin correctly.

chrisdembia:
The Travis builds are taking too long. I'm merging this, and I can address any issues in future PRs. I tested these changes locally, and I don't think there will be issues.

tkuchida:
@chrisdembia I just built doxygen twice, and it started with
<pre>
1>  Unzipping MathJax.zip for developer documentation
1>  Unzipping MathJax.zip for user documentation
</pre>
both times. Is that the expected behavior? I don't recall having to wait for MathJax to unzip every time—though I haven't built doxygen locally in a while.

chrisdembia:
Is that the expected behavior? 

Unfortunately, yes. It's only noticeable now because this same target is also deleting a bunch of files, which is particularly slow on Windows. On my computer, I can run those targets in parallel (`make -j2 doxygen`). The description for the target should be updated to include "and delete unnecessary files."

It should be possible to unzip once and copy the unzipped directory into the developer and user folders, but I think it is just a little bit more complicated to achieve with CMake.

chrisdembia:
Note that subsequent rebuilds of doxygen do *not* re-unzip. The unzipping only needs to happen the first time you build doxygen. 

I created #1807 to track this issue.

chrisdembia:
Oops. I see what you mean..the unzipping *does* occur every time you build doxygen. This is not expected behavior. I'll add this to #1807.

tkuchida:
Thanks. Yes, MathJax is twice unzipping every time.

chrisdembia:
@tkuchida see my new comment in #1807 about what I think may be causing that.

Issue #1793, opened 2017-06-23
Excessive enforcement of build type consistency


aymanhab:
My understanding is that while on windows mixing release and debug libraries is prohibited, it is not the case on osx and other platforms. While trying to compile OpenSim core as debug (to troubleshoot a specific problem at the OpenSim core level) the build failed because it claimed dependencies were not built in debug configuration, is this how it's supposed to work?

aseth1:
I build *opensim_dependencies* in debug and the I have no issues.

chrisdembia:
>  is this how it's supposed to work?

Currently, this is the expected behavior. However, I agree that we could be more flexible.

I'll submit a proposed change.

chrisdembia:
@aymanhab it sounds like it's discouraged (https://stackoverflow.com/questions/11658915/mixing-debug-and-release-library-binary-bad-practice), but I would think that at least mixing release with relwithdebinfo is okay. I've submitted a PR that I think strikes a good balance

PR #1798, opened 2017-06-25
Remove fatal error for config mismatch on UNIX.

chrisdembia:
Fixes issue #1793 

### Brief summary of changes

1. Improved formatting of message about config mismatch. Also, using a non-status type of message for this message was making `ccmake` (the command-line GUI interface) harder to use.
2. Removed the fatal check on UNIX.
3. Removed unnecessary `add_dependencies()` calls, as dependencies will be inherited from osimCommon.

Since mixing config types is discouraged in general, we always print the "notice." But we only stop the build process on Windows.

### Testing I've completed

Ensured that no error occurs now on macOS when configs do not match.

### CHANGELOG.md (choose one)

- no need to update because the original check did not exist in 3.3.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,13 +15,9 @@ if(COMMAND cmake_policy)
     cmake_policy(SET CMP0005 NEW)
 endif(COMMAND cmake_policy)
 
-string(CONCAT DEP_BUILD_MSG
-  "\n-------------------------IMPORTANT----------------------------------\n"
-  "| Make sure that all OpenSim dependencies are built with the same   |\n"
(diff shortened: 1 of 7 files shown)


chrisdembia:
I do not think the travis tests must pass for this PR. They simply need to get past compiling osimCommon, which they have.

aymanhab:
LGTM @chrisdembia Thanks for the fix. Although discouraged completely in Windows, seems useful if you have a full stack in release and you want to debug one library (at your own risk).

chrisdembia:
Thanks @aymanhab . Merging by rule 5.

PR #1799, opened 2017-06-25
[TGCS] Interactive Hopper GUI parameter tuning

nickbianco:
### Brief summary of changes
- Replaced tendon stiffness parameter with optimal fiber length parameter
- Removed MillardTendonParams: we can just use the default tendon now
- Tuned the passive spring stiffness so multiple strategies are "competitive"
- Reduced device masses to increase overall hop heights

### Testing I've completed
I tested a handful of attempts using both the "all patella wrap" and "load passive spring" strategies. For basic one hop solutions, they are relatively close in max height reached. Interestingly, "The Average Joe" outperforms "The Katie Ledecky" for the "all patella wrap" strategy and vice versa for the "load passive spring" strategy (if I'm remembering correctly, might be the other way around). 

It seems like we're getting pretty close to a finalized demo! Play around with the current parameters and let me know if there's any last tweaks that could be made.


chrisdembia:
Great! I'll try this out later today. Do you want to also remove `xticks`, etc. so that the tutorial will run on R2014b? Or perhaps it makes more sense for @carmichaelong to submit that fix, because @carmichaelong already has it working?

carmichaelong:
I technically didn't get it "working". I commented it out so that the example would work, but I never replaced the lines to get the plots to have the right ticks.

nickbianco:
I'll make the `xticks` changes, since I have the branch set up already. I'll just include here with any last tweaks we make to the parameters.

chrisdembia:
> I'll make the xticks changes, since I have the branch set up already. I'll just include here with any last tweaks we make to the parameters.

Great. Could you also put a comment in the README saying "These files have been tested with MATLAB versions R2014b and later." or something like that?

nickbianco:
Will do.

nickbianco:
Ready for review.

Made the `xticks` change and tweaked Average Joe's parameters.

Bindings/Java/Matlab/Hopper_Device/BuildHopper.m
```diff
@@ -49,22 +49,25 @@
 
 p = inputParser();
 defaultMaxIsometricForce = 4000.0;
-defaultMillardTendonParams = [0.049 28.1 0.67 0.5 0.25];
+defaultOptimalFiberLength = 0.55;
+defaultTendonSlackLength = 0.25;
```
chrisdembia: I'll push a commit to this branch at some point tonight to update the comment at the top to account for the change in parameters.

nickbianco: Cool, thanks. Should we update the default parameters to match Average Joe's? Or are we cool with the GUI muscles not match the rest of the demo?

chrisdembia: Hmm... I think it's fine for the parameters to differ, and if we change the parameters, we would need to readjust other parameters to get the initial model to not jump.

chrisdembia:
I pushed changes to the help docs. Thanks, @nickbianco, for the other changes.

Issue #1794, opened 2017-06-24
Speed up travis-ci build times


chrisdembia:
We can try reintroducing `ccache`, and we can remove the Linux clang build (use gcc to build the bindings on linux).

chrisdembia:
https://github.com/opensim-org/opensim-core/tree/speedup_travis_ci_june2017edition_trusty

chrisdembia:
CCache is now working with GCC and on macOS, but not on Linux with Clang. This is because we download separate Clang binaries.

jenhicks:
@chrisdembia Can we close now with the Linux PR?

chrisdembia:
The situation is definitely better now, but we are now seeing debug builds time out (not travis' time out, but ctest's 10-minute time out for individual tests). I'll rename the issue and I think we should leave it open.

PR #1800, opened 2017-06-25
Speed up Travis-CI, June 2017 edition

chrisdembia:
Fixes issue #1794.

### Brief summary of changes

- Reinstate use of ccache to cache build products and speed up the build. A rebuild, with no code changes, now only takes 7-11 minutes.
- Remove, for now, the more advanced caching mechanism that @klshrinidhi set up. We have not been using it, and we cannot continue using bintray, so reinstating this would be a good amount of work. If we want to revisit this, we can get it from the git history.

### Testing I've completed

- Ran Travis CI. It seems like the caching for macOS is not working well. The macOS builds have not been timing out, though, so maybe we should disable caching for macOS.

Not to discuss here, but we could also consider getting rid of one of the linux builds (either the gcc or the clang release builds).

### CHANGELOG.md (choose one)

- no need to update because this is not user-facing.


aseth1:
The speedup will be welcomed, especially as we test the Tools and examples. These would take full advantage of the caching.

aseth1:
Merging since the timeout was on the push and merge passed.

PR #1801, opened 2017-06-26
Fix missing state columns warnings

aseth1:
Fixes issue raised in [GUI #183](https://github.com/opensim-org/opensim-gui/issues/183) that StaticOptimization (but generally the AnalyzeTool) is too verbose about missing state values when forming the state storage from file. In #1686 we consolidated methods used to form the states storage. When forming states storage from coordinates only, the previous method did not warn about missing/unspecified states. Therefore, this PR adds the option/flag to warn about unspecified states and it is used to suppress warnings when forming states storage from coordinates only.

### Brief summary of changes
Added the `warnUnspecifiedStates` optional flag (default is *true*) to `Model::formStateStorage()`.
Updated `AnalyzeTool` so that when calling `formStateStorage` with coordinates only, to turn `warnUnspecifiedStates` off (to *false*).

### Testing I've completed
Ran all tests and ISB tutorial and observed that warnings are no longer printed to the console.

### CHANGELOG.md (choose one)
- no need to update because it restores previous behavior of the `AnalyzeTool`
- the `Model::formStateStorage()` interface changed by adding the option, but the default value enables user code to remain unchanged and have the expected/previous warning behavior.

### DIFF
--- a/OpenSim/Simulation/Model/Model.h
+++ b/OpenSim/Simulation/Model/Model.h
@@ -404,7 +404,7 @@ public:
      */
     void formStateStorage(const Storage& originalStorage, 
                           Storage& statesStorage,
-                          bool warnUnspecifiedStates = true);
+                          bool warnUnspecifiedStates = true) const;
 
     void formQStorage(const Storage& originalStorage, Storage& qStorage);
     
--- a/OpenSim/Simulation/Model/Model.h
+++ b/OpenSim/Simulation/Model/Model.h
@@ -394,12 +394,18 @@ public:
     /** Check that the underlying computational system representing the model is valid. 
         That is, is the system ready for performing calculations. */
     bool isValidSystem() const;
+
     /**
-     * create a storage (statesStorage) that has same label order as model's states
-     * with values populated from originalStorage, 0.0 for those states unspecified
-     * in the originalStorage.
+     * Create a storage (statesStorage) that has same label order as model's states
+     * with values populated from originalStorage. Use the default state value if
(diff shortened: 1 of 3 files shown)


OpenSim/Simulation/Model/Model.h
```diff
@@ -394,12 +394,18 @@ OpenSim_OBJECT_NONTEMPLATE_DEFS(Model, ModelComponent);
...
-     * with values populated from originalStorage, 0.0 for those states unspecified
-     * in the originalStorage.
+     * Create a storage (statesStorage) that has same label order as model's states
+     * with values populated from originalStorage. Use the default state value if
+     * a state is unspecified in the originalStorage. If warnUnspecifiedStates is
+     * true then a warning is printed that includes the default value used for
+     * the state value unspecified in originalStorage.
      */
-    void formStateStorage(const Storage& originalStorage, Storage& statesStorage);
+    void formStateStorage(const Storage& originalStorage,
```
tkuchida: Not a `const` method?

aseth1: Done.

tkuchida: :+1:

aseth1:
Debug tests are timing out. In particular `testCMCWithControlConstraintsRunningModel`  is hitting the 1500s individual time limit. I don't believe it is due to these changes. I am investigating the the slowness of `testCMCWithControlConstraintsRunningModel` locally.

PR #1802, opened 2017-06-27
Fix unnecessary resampling of uniform input data

aseth1:
Fixes issue with reported start and end-times that are slightly off from their range specified in the Tool setup file.

Noticed this running ISB 2017 example running SO in both 3.3 and 4.0. SO start time is being changed to -3.04423e-13 when both the user specified start time and the experimental data start at 0.0. The problem is that during filtering the data is re-sampled according to the minimum step-size (intended to accurately sample simulation results) but this applied to evenly sampled data where you end up with a minim step size that differs from the typical/average just due to round-off error. That causes all data points to be shifted by the that round-off. 

### Brief summary of changes
Check if the data is uniformly sampled and do not re-sample the data if it is uniform. This fixes the side-effect of having user-specified start and end-times shifted due to round-off when uniform timeline is provided but re-sampling with the min step size.

### Testing I've completed
All tests passed locally.

### CHANGELOG.md (choose one)
- no need to update because because this is a minor bug fix

### DIFF
--- a/OpenSim/Common/Storage.cpp
+++ b/OpenSim/Common/Storage.cpp
@@ -2287,15 +2287,6 @@ pad(int aPadSize)
     delete[] vecs;
 }
 
-//_____________________________________________________________________________
-/**
- * Smooth spline each of the columns in the storage.  Note that as a part
- * of this operation, the storage is resampled so that the statevectors are
- * at equal spacing.
- *
- * @param aOrder Order of the spline.
- * @param aCutoffFrequency Cutoff frequency.
- */
 void Storage::
 smoothSpline(int aOrder,double aCutoffFrequency)
 {
@@ -2336,16 +2327,6 @@ smoothSpline(int aOrder,double aCutoffFrequency)
     delete[] signal;
 }
 
-
-//_____________________________________________________________________________
-/**
- * Lowpass filter each of the columns in the storage.  Note that as a part
- * of this operation, the storage is resampled so that the statevectors are
- * at equal spacing.
- *
- * @param aOrder Order of the FIR filter.
- * @param aCutoffFrequency Cutoff frequency.
- */
 void Storage::
 lowpassIIR(double aCutoffFrequency)
 {
@@ -2383,16 +2364,6 @@ lowpassIIR(double aCutoffFrequency)
     delete[] signal;
 }
 
-
-//_____________________________________________________________________________
-/**
- * Lowpass filter each of the columns in the storage.  Note that as a part
- * of this operation, the storage is resampled so that the statevectors are
- * at equal spacing.
- *
- * @param aOrder Order of the FIR filter.
- * @param aCutoffFrequency Cutoff frequency.
- */
 void Storage::
 lowpassFIR(int aOrder,double aCutoffFrequency)
 {
--- a/OpenSim/Common/Storage.cpp
+++ b/OpenSim/Common/Storage.cpp
@@ -2299,16 +2299,21 @@ pad(int aPadSize)
 void Storage::
 smoothSpline(int aOrder,double aCutoffFrequency)
 {
+    int size = getSize();
     double dtmin = getMinTimeStep();
+    double avgDt = (_storage[size - 1].getTime() - _storage[0].getTime()) / size;
 
     if(dtmin<SimTK::Zero) {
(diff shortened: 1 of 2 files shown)


OpenSim/Common/Storage.cpp
```diff
@@ -2299,16 +2299,21 @@ pad(int aPadSize)
...
 
-    // RESAMPLE
-    dtmin = resample(dtmin,aOrder);
-    int size = getSize();
+    // RESAMPLE if the sampling interval is not uniform
+    if ((avgDt - dtmin) > SimTK::Eps) {
+        dtmin = resample(dtmin, aOrder);
+        size = getSize();
+    }
+
```
chrisdembia: +1. The only way for the minimum DT and the average DT to be the same is if you have uniform DT.

aseth1:
> Consider describing this behavior in the doxygen description for the modified functions.

IMHO it was a bug that we were attempting to resample uniformly sampled data before. Besides the tiny round-off differences that are obvious when expecting precisely 0, i don't believe it changes the answers.

I am happy to add doxygen to the effect that filters resample if the data in non-uniform, but do you think it is necessary to describe that this was changed from always resampling?

chrisdembia:
> I am happy to add doxygen to the effect that filters resample if the data in non-uniform, but do you think it is necessary to describe that this was changed from always resampling?

No, not necessary.

OpenSim/Common/Storage.cpp
```diff
@@ -2287,28 +2287,24 @@ pad(int aPadSize)
...
- *
- * @param aOrder Order of the spline.
- * @param aCutoffFrequency Cutoff frequency.
- */
 void Storage::
 smoothSpline(int aOrder,double aCutoffFrequency)
 {
+    int size = getSize();
     double dtmin = getMinTimeStep();
+    double avgDt = (_storage[size - 1].getTime() - _storage[0].getTime()) / size;
```
tkuchida: I think you need to divide by `size-1` (because it takes only 19 cuts to get 20 pieces of bread from a loaf). For example, suppose the time column is [0, 0.1, 0.2, 0.3, 0.4, 0.5]. The minimum step size is 0.1 but `avgDt` will be calculated as (0.5-0)/6 = 0.083333. Not sure why this off-by-one error wasn't caught by a test.

Also see L2335 and L2372.

OpenSim/Common/Storage.cpp
```diff
@@ -2287,28 +2287,24 @@ pad(int aPadSize)
...
 
     if(dtmin<SimTK::Zero) {
```
tkuchida: Changed to `SimTK::Eps` on L2337 and L2374 but not here?

OpenSim/Common/Storage.cpp
```diff
@@ -2287,28 +2287,24 @@ pad(int aPadSize)
...
         cout<<"Storage.SmoothSpline: storage cannot be resampled."<<endl;
         return;
     }
 
-    // RESAMPLE
-    dtmin = resample(dtmin,aOrder);
-    int size = getSize();
+    // RESAMPLE if the sampling interval is not uniform
+    if ((avgDt - dtmin) > SimTK::Eps) {
```
tkuchida: I understand the logic that `dtmin` should necessarily be less than `avgDt`, but `avgDt` is the result of a calculation. If the points are spaced uniformly, I think roundoff error could cause `avgDt` to be *smaller* than `dtmin`, so I think there should be `fabs()` here despite the apparent logical inconsistency.

aseth1: If `(avgDt - dtmin)` < 0, as you point out, it could only be due to round-off and then resampling still does not occur.

tkuchida: Ah, right.

OpenSim/Common/Storage.cpp
tkuchida: Fixed by 75a39bc58526c8264dac496f3ceddcda469f092d.

OpenSim/Common/Storage.cpp
tkuchida: Fixed by c37f0aa6577165fe10f1a64c1bbcf351a3be19ee.

aseth1:
CI timing out in debug and unrelated to these changes.

PR #1809, opened 2017-06-28
[TGCS] Address remaining dry run feedback.

chrisdembia:
### Brief summary of changes

- Give a more useful error message for when users exit the visualizer without simulating once.
- Instruct users to use an appropriate variable name for the reporter, since the remaining code assumes a specific name.
- Tell users that simulating won't work after adding the device but before wiring up the controller's input.

### Testing I've completed

Ran all 3 parts of the tutorial successfully.

### CHANGELOG.md (choose one)

- no need to update because hopper example was not in 3.3.

### DIFF
--- a/Bindings/Java/Matlab/Hopper_Device/RunHopperWithDevice_answers.m
+++ b/Bindings/Java/Matlab/Hopper_Device/RunHopperWithDevice_answers.m
@@ -73,6 +73,7 @@ anchorB.connectSocket_parent_frame(shankAttach);
 % ANSWER{
 hopper.addComponent(device);
 % }
+% Note: After this step, simulating will not work until after completing Task F.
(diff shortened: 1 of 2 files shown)


nickbianco:
LGTM. The new error message and plotting changes worked fine in my test.

chrisdembia:
> LGTM. The new error message and plotting changes worked fine in my test.

Great. Thanks for testing.

carmichaelong:
Didn't even realize that notice the error message was that seemingly weird when you don't run a simulation. Just tested before and after and all LGTM.

chrisdembia:
Yeah you get an exception in `toString_impl()`. Thanks for trying the changes @carmichaelong . I'll merge this now.

Issue #1814, opened 2017-06-29
testREADME and README are out of date (with how to use Manager::integrate()).


chrisdembia:
The [test](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Tests/README/testREADME.cpp#L126) uses the new `integrate()` call, but the README itself does not.

aseth1:
Also was addressed by #1787. Closing.

PR #1818, opened 2017-07-05
Sync README.md with testREADME.cpp

tkuchida:
Fixes issue #1814

testREADME.cpp already [says the following](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Tests/README/testREADME.cpp#L25):
<pre>
Whenever you change this test:
1. Copy the new code over to the README.md, making sure to omit the
preprocessor lines (#ifdef VISUALIZE; #endif).
</pre>

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,8 +106,8 @@ int main() {
     // Simulate.
     RungeKuttaMersonIntegrator integrator(model.getSystem());
     Manager manager(model, integrator);
-    manager.setInitialTime(0); manager.setFinalTime(10.0);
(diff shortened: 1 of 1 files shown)


chrisdembia:
I should have updated that issue, but #1787 is changing those lines.

tkuchida:
> I should have updated that issue, but #1787 is changing those lines.

The diff I see in #1787 for README.md is +3 –0. The 3 added lines are unrelated:
diff

PR #1824, opened 2017-07-11
Bindings model copyctor

aymanhab:
Fixes issue (multiple opensim-gui crashes due to early garbage collection)

### Brief summary of changes
Force generation of Model copy constructor that returns an object of the right type that can be maintained and communicates ownership correctly to the Java runtime. Other users of the API may do the same in their code to create a fresh copy of a Model in Matlab or Python.
 
### Testing I've completed
Run variety of tools multiple times without crashing the GUI.

### CHANGELOG.md (choose one)

- no need to update because this is more of a bug fix.


aymanhab:
Thank you @chrisdembia 🥇

PR #1826, opened 2017-07-12
[TGCS] InteractiveHopper demo only works on MATLAB 2014b or later, README update

nickbianco:
### Brief summary of changes
- Confirmed that InteractiveHopper demo only works on MATLAB 2014b or later, since graphics handles [were switched from doubles to objects](https://www.mathworks.com/help/matlab/graphics_transition/graphics-handles-are-now-objects-not-doubles.html) in MATLAB 2014b.
- Made update to README in Hopper_Device.

Based on feedback from Colin Smith's beta test.


chrisdembia:
Thanks @nickbianco

PR #1830, opened 2017-07-14
[doxygen] Avoid truncating short descriptions

tkuchida:
Fixes line breaks like this:
doxygen


tkuchida:
> what caused the line breaks

Doxygen interprets the last period in "w.r.t." as a sentence-ending period, terminating the *brief description*. An alternative solution is to [escape the period](http://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdchardot) but I think we can spend the characters to write "with respect to" in full.

chrisdembia:
Ah right... thanks for explaining.

chrisdembia:
Travis failure is unrelated (and annoying).

PR #1834, opened 2017-07-19
Fix InverseKinematicsTool's model out of date with its properties

aseth1:
Fixes issue #[GUI 249](https://github.com/opensim-org/opensim-gui/issues/249)

### Brief summary of changes
Explicitly call `Model::finalizeFromProperties()` before `Model::printBasicInfo`.
This isn't an issue about loading a new model. The model is modified outside the Tool somehow, e.g. during plotting. And future attempts to execute the tool subsequently fail.

### Testing I've completed
Tests and tools pass/run locally. 

Needs to be verified in the GUI.

### CHANGELOG.md (choose one)

### DIFF
--- a/OpenSim/Tools/InverseKinematicsTool.cpp
+++ b/OpenSim/Tools/InverseKinematicsTool.cpp
@@ -262,6 +262,9 @@ bool InverseKinematicsTool::run()
         else
             modelFromFile = false;
 
+        // although newly loaded model will be finalized
+        // there is no gaurantee that the _model has not been edited/modified
+        _model->finalizeFromProperties();
         _model->printBasicInfo();
 
         // Do the maneuver to change then restore working directory 
--- a/Applications/IK/ik.cpp
+++ b/Applications/IK/ik.cpp
@@ -141,12 +141,6 @@ int main(int argc,char **argv)
     // CONSTRUCT
     cout<<"Constructing tool from setup file "<<setupFileName<<".\n\n";
     InverseKinematicsTool ik(setupFileName);
-    //ik.print("ik_setup_check.xml");
-
-    // PRINT MODEL INFORMATION
-    //Model& model = ik.getModel();
-    //model.finalizeFromProperties();
-    //model.printBasicInfo();
 
     // start timing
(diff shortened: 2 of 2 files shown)


OpenSim/Tools/InverseKinematicsTool.cpp
```diff
@@ -262,6 +262,9 @@ bool InverseKinematicsTool::run()
         else
             modelFromFile = false;
 
+        // although newly loaded model will be finalized
+        // there is no gaurantee that the _model has not been edited/modified
+        _model->finalizeFromProperties();
```
tkuchida: If the comment on L265 is correct, then a newly loaded model will already be finalized and it would be more logical for this `finalizeFromProperties()` to be part of the `else` block.

aymanhab: The GUI uses setModel if that makes any difference.

aseth1: True. But it is consistent with what [InverseDynamicsTool](https://github.com/opensim-org/opensim-core/blob/b4ca2ed24aa4f243871487c4007fb389e262d7d2/OpenSim/Tools/InverseDynamicsTool.cpp#L249) is doing. Also avoiding another commit/build cycle would be preferable, I think.

chrisdembia:
IMO this should be tested with the GUI before merging to ensure it actually fixes the issue. As @aymanhab mentioned, you could just drop the new DLLs into a previous release of the GUI to test.

chrisdembia:
Sorry @aseth1 I just saw your comments in the related GUI issue.

chrisdembia:
If ID is doing the same thing, then we should merge this. We should fix both IK and ID after ISB.

aymanhab:
Will merge this but would be good to know what's involved. I believe the issue with plotting is that to create the builtin curves a MuscleAnalysis gets added to the model along the way by AnalyzeTool and gets removed on exit, I haven't stepped into the code to see if that's actually what happens since I'd have to build debuggable native libraries. Just my conjecture.

Issue #1839, opened 2017-07-23
Python ImportError - No module named '_common'


pavitrakumar78:
Hi,

I have built and installed opensim with python bindings in Windows 10 with python 3.5. The build was successful and so were the tests.

```
PS E:\Software\opensim_stuff\opensim_build> ctest -C RelWithDebInfo --parallel 8
Test project E:/Software/opensim_stuff/opensim_build
      Start  1: testLepton
      Start  2: testC3DFileAdapter
      Start  3: testComponentInterface
      Start  4: testDataTable
      Start  5: testFunctionAdapter
      Start  6: testFunctions
      Start  7: testGCVSpline
      Start  8: testMarkerData
 1/90 Test  #1: testLepton ...................................   Passed    0.07 sec
 2/90 Test  #3: testComponentInterface .......................   Passed    0.20 sec
 3/90 Test  #7: testGCVSpline ................................   Passed    0.15 sec
 4/90 Test  #6: testFunctions ................................   Passed    0.16 sec
 5/90 Test  #5: testFunctionAdapter ..........................   Passed    0.17 sec
 6/90 Test  #4: testDataTable ................................   Passed    0.19 sec
 7/90 Test  #8: testMarkerData ...............................   Passed    0.16 sec
      Start  9: testNCSpline
      Start 10: testPath
      Start 11: testRootSolver
      Start 12: testSTOFileAdapter
      Start 13: testSerialization
      Start 14: testSmoothSegmentedFunctionFactory
      Start 15: testStorage
 8/90 Test #10: testPath .....................................   Passed    0.25 sec
 9/90 Test  #9: testNCSpline .................................   Passed    0.35 sec
10/90 Test #11: testRootSolver ...............................   Passed    0.34 sec
11/90 Test #13: testSerialization ............................   Passed    0.42 sec
12/90 Test #15: testStorage ..................................   Passed    0.41 sec
      Start 16: testTRCFileAdapter
      Start 17: testConstraints
      Start 18: testJoints
      Start 19: testAssemblySolver
      Start 20: testContactGeometry
13/90 Test  #2: testC3DFileAdapter ...........................   Passed    1.58 sec
14/90 Test #14: testSmoothSegmentedFunctionFactory ...........   Passed    1.31 sec
      Start 21: testForces
      Start 22: testFrames
15/90 Test #17: testConstraints ..............................   Passed    1.63 sec
16/90 Test #22: testFrames ...................................   Passed    0.54 sec
[...]

PR #1840, opened 2017-07-24
Update README.md - Add cmake command to fix import error in python (Windows 10)

pavitrakumar78:
Added an extra command in the 'Impatient Method' for Windows to fix import error in python. 

Fixes issue #1839  

### Brief summary of changes
Added extra cmake command (to Impatient Method - Windows) to fix python import error.

### Testing I've completed
Tested on Windows 10, Python 3.5
`import opensim` works without any errors.

None

### CHANGELOG.md (choose one)

-added extra cmake command with `--target` parameter

### DIFF
--- a/README.md
+++ b/README.md
@@ -899,7 +899,7 @@ make -j8
 ctest -j8
 make -j8 install
 ```
-Note: You may need to add `<FULL-DIR>/opensim_install/bin` to your PATH variable as per [these instructions](#set-environment-variables).  
+Note: You may need to add `<FULL-DIR>/opensim_install/bin` to your PATH variable as per [these instructions](#set-environment-variables-2).  
 Example: If opensim_install is in your home directory:
 
         $ echo 'export PATH=~/opensim_install/bin:$PATH' >> ~/.bashrc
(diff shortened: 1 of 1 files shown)


chrisdembia:
Thanks for submitting this. Could you put this after the `ctest` line? You should only install if the tests pass.

Also, could you add a note after the instructions saying "You may need to set your system's PATH environment variable to include the full path to the `opensim_install\bin` folder."

Also, could you either add similar lines to the other "for the impatient" instructions," or create GitHub Issues to track adding these in the future?

pavitrakumar78:
> Thanks for submitting this. Could you put this after the ctest line? You should only install if the tests pass.
> 
> Also, could you add a note after the instructions saying "You may need to set your system's PATH environment variable to include the full path to the opensim_install\bin folder."

Done! Added instructions and extra cmake command for both Windows and Ubuntu Impatient installations. 



> Also, could you either add similar lines to the other "for the impatient" instructions," or create GitHub Issues to track adding these in the future?

Not sure how I should do this for Mac OSX. For Ubuntu, I added `make -j8 install` command after `ctest -j8`.

chrisdembia:
This is fantastic @pavitrakumar78. Thanks a bunch. 

Also, I'm happy the python package is now working for you.

PR #1843, opened 2017-07-26
Fix 1650 c3 d file adapter column order

aseth1:
Fixes issue #1650 

### Brief summary of changes
Directly addresses the issue by ordering the force, cop and moment columns associated with a force-plate (analog data) in the order expected in 3.3 and earlier version of OpenSim. The order is arbitrary anyway so choosing the one that is easiest to handle makes the most sense. A simpler fix than that attempted in #1652. It will significantly simplify and speedup #1434.

### Testing I've completed
C3DFileAdapter related tests pass locally including scripting tests. 

Would be great to see how this impacts #1434, @jimmyDunne.

### CHANGELOG.md (choose one)
- no need to update because is being introduced for the first time in 4.0 and does not change any interfaces.

### DIFF
--- a/OpenSim/Common/C3DFileAdapter.cpp
+++ b/OpenSim/Common/C3DFileAdapter.cpp
@@ -197,15 +197,15 @@ C3DFileAdapter::extendRead(const std::string& fileName) const {
                 at(_unit_index.at("force"));
             units.upd().push_back(SimTK::Value<std::string>(force_unit));
 
-            labels.push_back(SimTK::Value<std::string>("m" + fp_str));
-            auto moment_unit = acquisition->GetPointUnits().
-                at(_unit_index.at("moment"));
-            units.upd().push_back(SimTK::Value<std::string>(moment_unit));
-
             labels.push_back(SimTK::Value<std::string>("p" + fp_str));
             auto position_unit = acquisition->GetPointUnits().
                 at(_unit_index.at("marker"));
             units.upd().push_back(SimTK::Value<std::string>(position_unit));
+
+            labels.push_back(SimTK::Value<std::string>("m" + fp_str));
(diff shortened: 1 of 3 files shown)


jimmyDunne:
@aseth1 This worked like a charm!! Using this build, the table was ordered force/point/moment. I was able to avoid any Matlab side rotations, which lead to large performance improvements. 

I give a 👍 to merge!

OpenSim/Common/C3DFileAdapter.cpp
```diff
@@ -230,14 +230,14 @@ C3DFileAdapter::extendRead(const std::string& fileName) const {
                                        (*fit)->GetValues().coeff(f, 1),
                                        (*fit)->GetValues().coeff(f, 2)};
                 ++col;
+                row[col] = SimTK::Vec3{ (*pit)->GetValues().coeff(f, 0),
+                    (*pit)->GetValues().coeff(f, 1),
+                    (*pit)->GetValues().coeff(f, 2) };
+                ++col;
```
chrisdembia: Please maintain the indentation format of the surrounding code.

aseth1:
@chrisdembia you can see the changes [here](https://github.com/opensim-org/opensim-core/compare/fix_1650_C3DFileAdapter_column_order?expand=1). I don't know why the PR appears not to be updated. If you attempt to [edit](https://github.com/opensim-org/opensim-core/edit/fix_1650_C3DFileAdapter_column_order/OpenSim/Common/C3DFileAdapter.cpp?pr=/opensim-org/opensim-core/pull/1843) you'll see the change has been made already.

chrisdembia:
@aseth1 the two new commits did not make it into the master branch (see here: https://github.com/opensim-org/opensim-core/commits/master/OpenSim/Common/C3DFileAdapter.cpp).

Issue #1465, opened 2017-01-05
Better examples and documentation of working with lists in Matlab and Python


jenhicks:

jenhicks:
@aseth1 Can you check out #1436 for usage there?

PR #1436, opened 2016-12-08
{Matlab} create muscle curve script

jimmyDunne:
Plots active and passive force-length curves of a model.

PR #1851, opened 2017-08-01
[WIP] Iterator Matlab Class for easier access

jimmyDunne:
### Fixes issue 
Deals with #1465 and effects #1436

### Brief summary of changes
A Matlab class for doing basic operations with OpenSim Lists. This is a utility for working with lists on Matlab, allowing similar operation as the Sets. 

### Testing I've completed
Basic operation (does it return the correct things). I would just like to show implementation before building a proper test. 

I have spent a few hours mocking this up and seeing if it is at all feasible. I would like feedback on (i) using Matlab classes instead of a function (ii) the interface, how does the implementation look (seen in testOsimList), (iii) any spec requests (additions or deletions). 

I can also make a DevProp if requested?

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/1851)


tkuchida:
An interface like this could be really helpful for MATLAB users who are unfamiliar with C++ iterators. I'm wondering whether `get()` would be slow if, e.g., used with a large model or within an optimizer. For example, if scale factors were being set on every Body in the Model, then `get()` would be performing n(n+1)/2 string comparisons and calls to `eval()` ([which is notoriously slow](http://blogs.mathworks.com/loren/2005/12/28/evading-eval/)) on each iteration. One strategy to avoid this would be to cache a map between the string and a reference (UUIDs would be useful here :wink:)—though, to avoid using out-of-date information, the struct would then need to detect changes to the Model and regenerate the map.

Bindings/Java/Matlab/Utilities/osimList.m
```diff
@@ -0,0 +1,95 @@
...
+        function obj = osimList(model,classname)
+            % Constructor for osimList class. Takes a model and a classname
+            % (string). constructs an instance of the osimList class with
+            % properties model and list
+            if nargin == 0
+                error('no inputs to constructor')
+            elseif nargin == 1
+                error('constructor takes two inputs, not one')
+            elseif nargin > 2
+                error(['2 inputs required, num2str(nargin) ' given])
```
chrisdembia: The `num2str(nargin)` is inside the string, and therefore will not be evaluated. I think you intended to place it outside the string.

Bindings/Java/Matlab/Utilities/osimList.m
```diff
@@ -0,0 +1,95 @@
+classdef osimList
+properties
+        model % a reference to an opensim model
+        list; % an opensim list
+    end
+    
+    methods 
+        function obj = osimList(model,classname)
```
chrisdembia: We could add support for arbitrary component types using your `eval()` and `getConcreteClassName()` trick. It won't always work though.

Bindings/Java/Matlab/Utilities/osimList.m
```diff
@@ -0,0 +1,95 @@
...
+        function reference = get(obj,name)
+            % return a reference for a object in the list
+            import org.opensim.modeling.*
+            model = obj.model;
+            list = obj.list;
+            li = list.begin();
+            while ~li.equals(list.end())
+                if strcmp(char(li.getName()),name)
+                    pathname = li.getAbsolutePathName();
+                    comp = model.getComponent(pathname);
```
chrisdembia: You already have the component in the list; no need to fetch it out of the model.

chrisdembia: Oh now I think I remember...it doesn't quite work. Well it should be an easy fix to the swig interface files to make that work. I think the issue is described [here](https://github.com/opensim-org/opensim-core/blob/master/Bindings/Java/Matlab/tests/testAccessSubcomponents.m#L39) ("MATLAB methods can't start with underscore muscle = muscleIter.__deref__();").

aymanhab:
My opinion is that it's a good start but really this should be done on the opensim-core side or at worst in SWIG bindings side. Forcing Matlab users to use lists and iterating is cumbersome, error-prone and counter-intuitive as far as I can tell. We should create methods to build Sets (or any other preferred container) of proper type and return them. Most scripters are using pre-made models so cached, properly-typed containers will make such an easy task easy while the iterators allow more involved model building/modification type of tasks possible.

tkuchida:
> We should create methods to build Sets (or any other preferred container) of proper type and return them.

Potentially relevant discussion in [Forum topic 8086](https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=8086). There, the suggestion was

"As of 4.0 Beta API, we're discouraging the use of BodySet, JointSet etc. as they may not provide a complete list of Bodies, Joints, ... since the model structure is potentially hierarchical, instead you'd use model.getComponent("PathName") and then safeDowncast to proper type (if you need to)."

I'm unclear about how we're expecting users to interact with the new API. These expectations may not align with what's most natural.

💡 Also note that [`Model::getNumBodies()` returns the size of the `BodySet`](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/Model.cpp#L1248), so if we're discouraging use of the `BodySet` because it might contain incomplete information, `Model::getNumBodies()` shouldn't be using it.

aymanhab:
Good point @tkuchida, my answer on the forum though was in the context of the current (what's available today) rather than what's the ideal 4.0 final API. The forum post goes to highlight the issues with the interface as it stands. Thank you.

tkuchida:
> my answer on the forum though was in the context of the current (what's available today) rather than what's the ideal 4.0 final API. The forum post goes to highlight the issues with the interface as it stands. Thank you.

Ok, thanks for clarifying. I actually encountered this issue in my own code yesterday and I was confused because I thought use of BodySet, JointSet, etc. were now discouraged so I was surprised to see them used internally by `getNumBodies()`, etc. Might be good to update those methods.

jimmyDunne:
@aseth1, if you could review and let me know. I would like to write some documentation for using lists— perhaps we could chat (Zoom)?.

Bindings/Java/Matlab/Utilities/osimList.m
jimmyDunne: Looking at the iterator methods, it is not clear then how to return a reference to a particular object.

aseth1: Why can't you return `li` since it is the Component you are querying?

chrisdembia: @aseth1 the type of `li` is something like `ComponentListIterator<Component>`. It is not the actual component and we should hide it from the user.

chrisdembia: We could add a fix to the SWIG interface as a performance enhancement sometime in the future.

aymanhab:
@jimmyDunne deref() method added, and should show up for all iterators, please let me know if that works for you.

jimmyDunne:
Thanks, Ayman! That made the Matlab implementation doable. 

This seems to work well now.

jimmyDunne:
@aseth1, should be good for a re-review 👍

Bindings/Java/Matlab/Utilities/osimList.m
```diff
@@ -69,25 +89,36 @@
...
                     break
                 end
                 li.next();
             end 
             if ~exist('reference','var')
-                error(['Component name: ' name ' is not found'])
+                error(['Component name not found: ' name])
+            end
+        end
+        function reference = getByIndex(obj,index)
```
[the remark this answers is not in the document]
tkuchida: FWIW, I would expect most MATLABers to use a `for` loop by default. To avoid inefficiency, might it be possible to keep track of the last item that was retrieved and start searching from there (essentially doing `li.next()` under the covers until `list.end()`, then starting from the beginning)? Alternatively, perhaps just documentation (with the above warning and code snippets) would be helpful under the `getByIndex()` method.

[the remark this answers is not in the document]
jimmyDunne: Thanks @tkuchida, @aseth1, and @chrisdembia for the awesome comments! This design feedback is great— I have been spending some time thinking about the philosophy of the list design and how that conflicts with the typical Matlab user for-loop strategy. The original reason I wanted to make this was to make it easier for me to get a single component from a list without having to write a while loop everytime. Perhaps it would be better to simplify this class to a set of static methods that are for single use only;

```MATLAB
% Get the number of components in the model
nComps = osimList.getNumberOfComponents(model, 'Body')
% Get all the body names (as strings)
bNames = osimList.getNames(model, 'Body') 
% I only want to work on the Pelvis, so I get it directly. 
pelvis = osimList.getComponent(model, 'Body', names{i} ) 
```
This wouldn't stop anyone from using it in a for-loop, but we could write in documentation when to use the above method and when to use the list directly (when you want to deal with most components in the list).

I am happy to jump on a call to discuss since I feel this may need discussion.

jimmyDunne:
@chrisdembia and @aseth1, once we agreed on generating a Matlab cell array of references, it was better to turn this into a simple function. Now the user inputs a model and a classname and the function returns a Matlab cell array of references. The user can then make their own decisions on how to deal with the cell array. 

I added a test (which is more of an example for now) and ran it locally (it passed)

aseth1:
> Now the user inputs a model and a classname and the function returns a Matlab cell array of references. The user can then make their own decisions on how to deal with the cell array.

👍

Bindings/Java/Matlab/Utilities/osimList2MatlabCell.m
```diff
@@ -0,0 +1,73 @@
...
+%   When iterating through a list to get/set properties on multiple 
+%   components, use the list iterator directly;
+%
+%   % Set a value for a property for all of the components of type 'Body'
+%   list = model.getBodyList();
+%   li = list.begin()
+%   while ~li.equals(list.end())
+%       li.next();
+%       li.setSomePropertyValue(val);
+%   end
```
chrisdembia: I think people can use the cell array even in this case. If you want to leave this documentation here anyway, perhaps motivate it instead by saying "Here's what this function does internally."

Bindings/Java/Matlab/tests/testOsimList2MatlabCell.m
```diff
@@ -0,0 +1,33 @@
+%% clear working space
+clear all;close all;clc;
+
+%% Import OpenSim Libraries
+import org.opensim.modeling.*
+
+%% Instantiate a model from file
+model = Model('../../../../OpenSim/Tests/shared/arm26.osim');
```
chrisdembia: Because arm26.osim is copied into the test directory, you can simply use `'arm26.osim'` here. We don't want to read files from the repository in the test case.

chrisdembia:
```suggestion
model = Model('arm26.osim');
```

Bindings/Java/Matlab/Utilities/osimList2MatlabCell.m
jimmyDunne: Thanks for the note— I left the code snippet as an example and explained what is being done internally.

PR #343, opened 2015-02-04
Create CMake option for copying Simbody headers and libraries.

chrisdembia:
This is the next change in the progress towards #195 (for allowing for the proper installation on UNIX into a system directory (e.g., /usr/local)).

This PR introduces the cmake "OPENSIM_COPY_SIMBODY" variable that decides if Simbody libraries and headers are to be installed with OpenSim, and if (only on windows) Simbody libraries should be copied to the OpenSim build directory so that OpenSim tests can be run without needing to set PATH.

1. This does not affect the default behavior on Windows.
2. I've removed the cmake related to lapack, etc., since this actually only belongs in Simbody.
3. The `BUILD_USING_NAMESPACE` variable and related lines are removed.
4. The OpenSimConfig.cmake file now depends on OPENSIM_COPY_SIMBODY. If not copying, then we add Simbody's lib dir and include dir to OpenSim's. This allows people to build an opensim-dependent project without having to explicitly include Simbody's include dir, etc. Basically, this allows for:
```
include_directories(${OpenSim_INCLUDE_DIRS})
link_directories(${OpeSim_LIB_DIR})
```
regardless of your platform. If not copying, those variables actually also contain the Simbody paths.

I checked the following things for this PR.
- With OPENSIM_COPY_SIMBODY on, make sure layout of installation is the same as before this PR.
  - [x] windows
  - [x] linux
- With OPENSIM_COPY_SIMBODY off, make sure it's possible to build, test and install OpenSim.
  - [x] windows
  - [x] linux
- With OPENSIM_COPY_SIMBODY off, make sure it's possible to build and run a client OpenSim project.
  - [x] windows
  - [x] linux
- With OPENSIM_COPY_SIMBODY on, make sure it's possible to build and run a client OpenSim project.
  - [x] windows
  - [x] linux

Issue #1804, opened 2017-06-27
Numeric properties load with many extraneous digits


tkuchida:
For example, the mass_center property of femur_r in gait1018_wArms.osim is "0 -0.17 0" in the .osim file but "0 -0.17000000000000001 0" appears in the GUI (opensim4.0Beta0627_win64.zip). Particularly undesirable for an introductory tutorial.

tkuchida:
Another example (from Confluence):
![](http://simtk-confluence.stanford.edu:8080/download/attachments/3376207/function-3.PNG?version=1&modificationDate=1349822297234&api=v2)

tkuchida:
See discussion in [opensim-gui issue 259](https://github.com/opensim-org/opensim-gui/issues/259).

jenhicks:
Is this an opensim-core issue @aymanhab

aymanhab:
Yes, AFAIK. the toString method called by the GUI ends up calling printUnformatted which prints excess digits. We could override on API side to use some formatting (e.g. %g).

aseth1:
We could do a better job handling formatting properties out to string.
`Property::toString()` calls  `Property::writeSimplePropertyToStream(outstream)` which currently invokes `SimTK::writeUnformatted(outstream, values)` which appears to write values to stream in full precision separated by spaces (if vector value).  There is no `SimTK::writeFormatted()` in general, and it appears to only add brackets and comma separate elements in the case of Arrays but does not appear control the number of sig figs, AFAICT.

Talking with @carmichaelong a potential solution is to have a global setting (user preference) to display sig figs of properties (e.g. default 8). The same preference could be used to write properties to XML or we could have separate *PropertyDisplayPrecision* and *PropertyWritePrecision* (something like that) or just leave file writing at full precision.

In any case we would need to override `Property::toString()` on new properties to format output. We see mixed behaviors in 3.3 because new properties use unformatted and deprecated properties do something along the lines of:
```cpp
string PropertyDblArray::toString() const
{
    string str = "(";
    char dbl[256];
    for(int i=0; i < _array.getSize(); i++){
        sprintf(dbl, "%g", _array[i]);
        str += (i>0?" ":"") + string(dbl);
    }
    str += ")";
    return str;
}
```
 We could make properties format like the deprecated, which is what @aymanhab seems to be advocating.

chrisdembia:
[...]

PR #1853, opened 2017-08-02
Bindings handle vec6 properties

aymanhab:
Fixes issue #343 of gui (Inertia not editable)

### Brief summary of changes
Exposed Property type Vec6 similar to Vec3 in PropertyHelper class for non templatized access

### Testing I've completed
Added test case that gets/sets Inertia Property

### CHANGELOG.md (choose one)

- no need to update because it's a bug fix and Vec6 types were never exposed to Scripting users in the past

### DIFF
--- a/Bindings/Java/tests/TestEditProperties.java
+++ b/Bindings/Java/tests/TestEditProperties.java
@@ -41,11 +41,12 @@ class TestEditProperties {
         // Now edit a Vec6 Property (inertia)
         AbstractProperty inertiaProp = bdy.getPropertyByName("inertia");
         // This block populates a stock Vec6 from a prestored inertia property
-        // 0.10423369488353 0.08831473564548 0.05870749935560 0 0 0
+        double[] inertiaFromFile = new double[]{0.10423369488353, 0.08831473564548, 0.05870749935560, 0, 0, 0};
         Vec6 oldInertia = new Vec6();
         for(int i=0; i<6; i++){
             oldInertia.set(i, PropertyHelper.getValueVec6(inertiaProp, i));
             System.out.println(oldInertia.get(i));
+            assert Math.abs(inertiaFromFile[i] - oldInertia.get(i)) < 1e-4;
         }
         // This block sets inertia to [0,1,2,3,4,5]
         for(int i=0; i<6; i++){
--- a/Bindings/Java/tests/TestEditProperties.java
+++ b/Bindings/Java/tests/TestEditProperties.java
@@ -38,23 +38,24 @@ class TestEditProperties {
                     System.exit(-1);
                 } 
         }
-		// Now edit a Vec6 Property (inertia)
-		AbstractProperty inertiaProp = bdy.getPropertyByName("inertia");
-		Vec6 oldInertia = new Vec6();
-		for(int i=0; i<6; i++){
-			oldInertia.set(i, PropertyHelper.getValueVec6(inertiaProp, i));
-			System.out.println(oldInertia.get(i));
-		}
-		for(int i=0; i<6; i++){
-			PropertyHelper.setValueVec6(i, inertiaProp, i); // Bad inertia matrix but we're just testing set/get'
-		}
-		Vec6 newInertia = new Vec6();
-		for(int i=0; i<6; i++){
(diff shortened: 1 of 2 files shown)


Bindings/Java/tests/TestEditProperties.java
```diff
@@ -38,6 +38,23 @@ public static void main(String[] args) {
...
+		for(int i=0; i<6; i++){
+			PropertyHelper.setValueVec6(i, inertiaProp, i); // Bad inertia matrix but we're just testing set/get'
+		}
+		Vec6 newInertia = new Vec6();
+		for(int i=0; i<6; i++){
+			double ret = PropertyHelper.getValueVec6(inertiaProp, i);
+			newInertia.set(i, ret);
+			System.out.println(newInertia.get(i));
+			assert newInertia.get(i)==i;
+		}
```
chrisdembia: Should be spaces, not tabs.

Bindings/Java/tests/TestEditProperties.java
```diff
@@ -38,6 +38,23 @@ public static void main(String[] args) {
...
         }
+		// Now edit a Vec6 Property (inertia)
+		AbstractProperty inertiaProp = bdy.getPropertyByName("inertia");
+		Vec6 oldInertia = new Vec6();
+		for(int i=0; i<6; i++){
+			oldInertia.set(i, PropertyHelper.getValueVec6(inertiaProp, i));
+			System.out.println(oldInertia.get(i));
+		}
+		for(int i=0; i<6; i++){
+			PropertyHelper.setValueVec6(i, inertiaProp, i); // Bad inertia matrix but we're just testing set/get'
```
chrisdembia: Consider limiting the line length to 80 characters. However, I also think it's fine that we are more lenient about this the farther the code is from the C++ source files.

Bindings/Java/tests/TestEditProperties.java
```diff
@@ -38,6 +38,23 @@ public static void main(String[] args) {
                     System.exit(-1);
                 } 
         }
+		// Now edit a Vec6 Property (inertia)
+		AbstractProperty inertiaProp = bdy.getPropertyByName("inertia");
+		Vec6 oldInertia = new Vec6();
+		for(int i=0; i<6; i++){
```
tkuchida: I don't understand what this block is testing that the two blocks below don't already test: `set()` is called on L54, `getValueVec6()` is called on L53, and this block has no effect on the `assert` on L56.

aymanhab: Thanks @tkuchida This block tests that we can read existing Vec6 properties read from file. I couldn't assert here since there could be roundoff errors.

aymanhab:
Thanks @chrisdembia Converted tabs to spaces and shortened line as suggested.

Bindings/Java/tests/TestEditProperties.java
tkuchida: `PropertyHelper.getValueVec6(inertiaProp, i)` also appears on L53; what is L45 testing that L53 isn't?

aymanhab:
I added more comments @tkuchida to explain what's tested. Please let me know if you have questions. Thanks.

tkuchida:
> I added more comments @tkuchida to explain what's tested. Please let me know if you have questions.

I disagree with the design philosophy here, but I don't think it's worth holding up the PR. I thought that perhaps the test at L45–49 was making sure `PropertyHelper.getValueVec6()` did _something_, but that method is already used on L58 (if the method stops working, the test at L56–62 will fail anyway). My guess is that the test at L45–49 requires the developer to read the text that's printed to the console and manually compare to the numbers in the comment on L44. I don't think that follows the design of other (automated) tests in opensim-core. If the concern is digits of precision, a comparison like `assert fabs(actual-target) < tol` may be appropriate.

aymanhab:
Thanks for the feedback @tkuchida I added the explicit test (keeping in mind that the numbers may change due to #1804

aymanhab:
Thanks much @tkuchida @chrisdembia Will merge when tests pass :racing_car:

aymanhab:
Unrelated sporadic test failures. Merging.

PR #1857, opened 2017-08-03
Add java marker manipulation test

aymanhab:
Fixes issue #<issue_number>

### Brief summary of changes
The added test case exercises the Markers -> Add New, Delete, saveToFile, loadFromFile
Inconsistencies & issues I had to work around that would be best resolved here in core then ported to GUI
- method MarkerSet.addMarker is broken and should be removed.
- newMarker.setParentFrame(body); works but newMarker.setFrameName("parentName") fails to connect in initSystem complaining that initSystem must be called first!
- When adding markers from file I had to call adoptAndAppend on the markerSet which is bad/incorrect memory management. I wanted to use cloneAndAppend but that throws exception since clone apparently doesn't handle sockets.

### CHANGELOG.md (choose one)

- updated...


aymanhab:
Any thoughts or feedback @chrisdembia ?

chrisdembia:
I haven't followed along on this; could we discuss in person?

Bindings/Java/tests/TestEditMarkers.java
```diff
@@ -0,0 +1,64 @@
...
+        Model model = new Model("gait10dof18musc_subject01.osim");
+        OpenSimContext context=null;
+        long t1  = System.currentTimeMillis();
+        context = new OpenSimContext(model.initSystem(), model);
+        MarkerSet markerset = model.getMarkerSet();
+        // The following code exercises the Marker->Add New  functionality
+        Vec3 offset = new Vec3(0.11, 0.22, 0.33);
+        Body body = model.getBodySet().get(0);
+        String newMarkerName = "newMarker";
+        // This exercises Marker -> Add New
```
aseth1: Do you mean the following lines 19-29 are what `Add New Marker` in the GUI does?

Bindings/Java/tests/TestEditMarkers.java
```diff
@@ -0,0 +1,64 @@
...
+import java.io.IOException;
+import static java.nio.file.StandardOpenOption.*;
+import java.nio.file.*;
+
+class TestEditMarkers {
+  public static void main(String[] args) {
+    try {
+        Model model = new Model("gait10dof18musc_subject01.osim");
+        OpenSimContext context=null;
+        long t1  = System.currentTimeMillis();
```
aseth1: This variable is unused AFAICT.

Bindings/Java/tests/TestEditMarkers.java
```diff
@@ -0,0 +1,64 @@
...
+        context.cacheModelAndState();
+        markerset.adoptAndAppend(newMarker);
+        try {
+            context.restoreStateFromCachedModel();
+        } catch (IOException ex) {
+            System.exit(1);
+        }
+        // This exercises Marker -> delete
+        context.cacheModelAndState();
+        Marker findMarker = Marker.safeDownCast(model.getComponent("newMarker"));
```
aseth1: Why do use the generic `getComponent` facility when you have the markerset in hand? Seems unnecessarily complicated.

aymanhab:
Thanks for the feedback @aseth1 I addressed your comments and added some comments in the test to clarify. Would appreciate your feedback on the items in the "Looking for feedback on.." above.

aseth1:
I think you can simply make issues for your feedback section. They will likely be addressed separately.

Bindings/Java/tests/TestEditMarkers.java
```diff
@@ -0,0 +1,65 @@
...
+        context.cacheModelAndState();
+        markerset.adoptAndAppend(newMarker);
+        try {
+            context.restoreStateFromCachedModel();
+        } catch (IOException ex) {
+            System.exit(1);
+        }
+        // This exercises Marker -> delete
+        context.cacheModelAndState();
+        // we use getComponent to make sure Marker is accessible thru model traversal
```
aseth1: thru -> through

aseth1: I don't understand why accessing the Marker through model traversal is important here. You could have a separate test for accessing Components through `getComponent`. Here it seems more important to have an efficient way to access and remove a marker. You already know it lives in the MarkerSet (that's how you remove it after all) so why not just find it in the Set.

Bindings/Java/tests/TestEditMarkers.java
```diff
@@ -0,0 +1,65 @@
...
+
+class TestEditMarkers {
+  public static void main(String[] args) {
+    try {
+        // GUI model loading
+        Model model = new Model("gait10dof18musc_subject01.osim");
+        OpenSimContext context=null;
+        context = new OpenSimContext(model.initSystem(), model);
+        MarkerSet markerset = model.getMarkerSet();
+        // The following code exercises the Marker->Add New  functionality
```
aseth1: This begs  the question why are you not invoking the Add New call back method? Seems that the test could become stale if the method in the GUI changes and the test isn't updated. Seems less error prone to invoke the method that makes these API and Context calls.

Bindings/Java/tests/TestEditMarkers.java
```diff
@@ -0,0 +1,65 @@
...
+        }
+        // This exercises Marker -> delete
+        context.cacheModelAndState();
+        // we use getComponent to make sure Marker is accessible thru model traversal
+        Marker findMarker = Marker.safeDownCast(model.getComponent("newMarker"));
+        assert (findMarker != null);
+        markerset.remove(findMarker);
+        try {
+            context.restoreStateFromCachedModel();
+        } catch (IOException ex) {
```
aseth1: Overlooked this before, but why is failing to `restoreStateFromCachedModel` an `IOException`? Are there other OpenSim::Exceptions that could be thrown in the process. Do you catch those and throw `IOException`? If so, it seems confusing/misleading to throw an IO error for state restoration.

Bindings/Java/tests/TestEditMarkers.java
aymanhab: What exactly is the "Add New" callback method? The code you see here is essentially what's invoked when the GUI user picks "Add New" context command from the "Markers" node in the navigator. If you see a better way to implement the functionality, here is the perfect place to replace my implementation and then I can port it to the GUI while keeping a test case on the core side to make sure we don't break it moving forward. I have to strip the GUI code down to raw API calls since the tests don't have access to GUI classes/functions/abstractions.

Bindings/Java/tests/TestEditMarkers.java
aymanhab: While I agree it's not completely related to this test case, the GUI uses the top level model traversal all the time to generate/update visualization/decorations so being included in the traversal right after addition has important downstream effects. Please let me know if you prefer it removed from this test case and included in another test case later.

If there was model.remove I would have used it indeed, but there isn't unfortunately 😄

Bindings/Java/tests/TestEditMarkers.java
aymanhab: We use SWIG to map all exceptions to Java.IOException because we want a "guarded" Java exception that we can handle in proper context, and this was the most reasonable.  Users never see the actual underlying type of the Java exception, just user-friendly messages.

aymanhab:
Issues #1888 #1889 has been opened to address the "Looking for feedback items" above per @aseth1

Bindings/Java/tests/TestEditMarkers.java
aseth1: > What exactly is the "Add New" callback method?

I mean exactly "what's invoked when the GUI user picks "Add New" context command from the "Markers" node in the navigator." You say it is "essentially" what is invoked, which tells me it is not exactly the same code. 

The approach here seems a bit backwards. The test should call the method you have implemented for "Add New" not a copy of that code. It does not seem optimal to test code here and hope it gets ported in the right way to the right place. Perhaps to satisfy both requirements (to be callable in the GUI and API) is to make it part of OpenSimContext? It seems to me that was one of the purposes of OpenSimContext and its tests.

Bindings/Java/tests/TestEditMarkers.java
aseth1: OK. Thanks for explaining.

Bindings/Java/tests/TestEditMarkers.java
aseth1: I definitely like the idea of having traversal tests from the GUI. Please do add that as test case.

I don't believe it needs to be exercised here simply to be exercised. Use the MarkerSet. It would be a different story if we didn't already access the MarkerSet.

aymanhab: Thanks @aseth1 will do.

Bindings/Java/tests/TestEditMarkers.java
aymanhab: The problem with using the GUI code as is here, in tests, is that it depends on multiple other libraries/classes that are part of the GUI code base (e.g. to support undo or to get/manipulate nodes from Navigator or to communicate with the Visualization). Pulling these and their dependencies into a test is a massive undertaking and may not be possible, in general, without including the netbeans platform itself, sorry if that wasn't made clear earlier.
We can discuss whether this belongs in the OpenSimContext class in person.

aymanhab:
Feedback has been addressed, and ready for re-review/merge. Thanks.

asethca:
Would be good to get a second reviewer take a look. @chrisdembia @carmichaelong?

Bindings/Java/tests/TestEditMarkers.java
```diff
@@ -0,0 +1,65 @@
+import org.opensim.modeling.*;
+import java.io.IOException;
+import static java.nio.file.StandardOpenOption.*;
+import java.nio.file.*;
+
+class TestEditMarkers {
```
chrisdembia: It would be great to have some more documentation of the test that explains that it is for GUI regression (rather than checking the functionality/correctness of the bindings in general).

Bindings/Java/tests/TestEditMarkers.java
chrisdembia: I agree that it's ideal if the test mimics what the GUI does in a maintainable way, and that moving code into the OpenSimContext could be a way to do that. But I don't know all the details, and in the end, it seems that it's up to @aymanhab to make sure this test remains useful for him. Perhaps the GUI code could have a comment saying "If you update this block of code, update the respective code in TestEditMarkers.java in opensim-core."

Bindings/Java/tests/TestEditMarkers.java
```diff
@@ -0,0 +1,65 @@
...
+        Body body = model.getBodySet().get(0);
+        String newMarkerName = "newMarker";
+        Marker newMarker = new Marker();
+        newMarker.setName(newMarkerName);
+        newMarker.set_location(offset);
+        newMarker.setParentFrameName(body.getName());
+        context.cacheModelAndState();
+        model.addMarker(newMarker);
+        try {
+            context.restoreStateFromCachedModel();
```
chrisdembia: Is this try-catch sufficient to ensure that adding the marker actually worked? Similar comment for the rest of the test.

Bindings/Java/tests/TestEditMarkers.java
```diff
@@ -0,0 +1,65 @@
...
+        } catch (IOException ex) {
+            System.exit(1);
+        }
+		// Exercise saveToFile
+        markerset.print("savedMarkers.xml");
+        // Now create a new MarkerSet from the file
+        MarkerSet newMarkerSet = new MarkerSet(model, "savedMarkers.xml");
+        Marker newMarkerRenamed = newMarkerSet.get(0);
+        String addedMarkerName = newMarkerRenamed.getName()+"_renamed";
```
chrisdembia: Just wondering; the renaming is necessary to avoid having two markers with the same name?

aymanhab: Yes @chrisdembia it is and also it exercises the Marker renaming done in the GUI so that not all markers are called "NewMarker_x".

Bindings/Java/tests/TestEditMarkers.java
aymanhab: The catch is intended for situation where trying to restore the state (e.g. Coordinate values etc.) fails. I don't expect this exception to be fired here when adding a Marker doesn't interact with the state but may happen in other contexts.

Bindings/Java/tests/TestEditMarkers.java
aymanhab: I will add the comment to the multiple GUI files that use this code. There is a tradeoff between using the Context object (which was intended for interacting with a unique state that lives on the C++ side) and plain usage of the API. If all GUI calls go through Context then we make it into a facade and it becomes too "fat" and rather hard to use/maintain. For a while we've been moving in the other direction so that we're actually exercising the API as a scripter would do.

aymanhab:
Thanks for the feedback @aseth1 The methods you mentioned are not exercised as part of Marker manipulation add/edit/write/load workflow in the GUI, instead they are exercised from the ScaleTool. I'd prefer to have the test case exercise the functionality relevant only to Marker manipulation (as you indicated earlier on this review) and to have the other calls tested within a scaleTool test case that's already in progress so that they can be removed/updated when ScaleTool is updated. Please let me know what you think. Thanks.

aymanhab:
Thanks @aseth1, @chrisdembia

PR #1861, opened 2017-08-04
[readme] Remove unnecessary line continuation

chrisdembia:
This PR removes, from the README, an unnecessary backtick, which is used for line continuation.


chrisdembia:
Thanks.

Issue #1768, opened 2017-06-15
Duplicate names for bodies cause exception


jenhicks:
@jimmyDunne commented on [Wed Jun 14 2017](https://github.com/opensim-org/opensim-gui/issues/175)

There is an exception if the model has the same name for multiple bodies or joints. There is a testing model attached. 

[subject01.osim.zip](https://github.com/opensim-org/opensim-gui/files/1075736/subject01.osim.zip)

---

@jenhicks commented on [Thu Jun 15 2017](https://github.com/opensim-org/opensim-gui/issues/175#issuecomment-308828553)

Add to the message, "As of 4.0, duplicate names are not allowed. Please update your model's osim file."

tkuchida:
I think duplicate names are permissible, but all full paths must be unique.

tkuchida:
Catch in a general component validation method; should not simply add a message when the `runtime_error` exception is thrown by MultibodyGraphMaker. See comments from @aymanhab in unmerged PR #1866.

jenhicks:
Should be addressed by #1887

aseth1:
Closed by #1887

PR #1866, opened 2017-08-06
Add message when addBody() and addJoint() encounter duplicate names

tkuchida:
Fixes issue #1768.

### Brief summary of changes
Catch `runtime_error`s thrown by Simbody's MultibodyGraphMaker (e.g., [here](https://github.com/simbody/simbody/blob/master/SimTKmath/src/MultibodyGraphMaker.cpp#L109)) and add an OpenSim-specific message.

### CHANGELOG.md (choose one)
Added note to "Converting from v3.x to v4.0" section.

### DIFF
--- a/OpenSim/Simulation/Model/Model.cpp
+++ b/OpenSim/Simulation/Model/Model.cpp
@@ -661,9 +661,15 @@ void Model::createMultibodyTree()
 
     auto bodies = getComponentList<Body>();
     for (auto& body : bodies) {
-        _multibodyTree.addBody( body.getAbsolutePathName(),
-                                body.getMass(), false,
-                                const_cast<Body*>(&body) );
+        try {
+            _multibodyTree.addBody( body.getAbsolutePathName(),
+                                    body.getMass(), false,
+                                    const_cast<Body*>(&body) );
+        } catch (const std::runtime_error& e) {
+            cout << "Duplicate Body names are not permitted. Please update "
+                 << "your model's .osim file accordingly." << std::endl;
+            throw e;
+        }
     }
 
     std::vector<SimTK::ReferencePtr<Joint>> joints;
@@ -695,14 +701,20 @@ void Model::createMultibodyTree()
             finalizeConnections(*this);
 
         // Use joints to define the underlying multibody tree
-        _multibodyTree.addJoint(name,
-            joint->getConcreteClassName(),
-            // Multibody tree builder only cares about bodies not intermediate
-            // frames that joints actually connect to.
-            joint->getParentFrame().findBaseFrame().getAbsolutePathName(),
-            joint->getChildFrame().findBaseFrame().getAbsolutePathName(),
-            false,
-            joint.get());
+        try {
+            _multibodyTree.addJoint(name,
+                joint->getConcreteClassName(),
+                // Multibody tree builder only cares about bodies not
+                // intermediate frames that joints actually connect to.
+                joint->getParentFrame().findBaseFrame().getAbsolutePathName(),
+                joint->getChildFrame().findBaseFrame().getAbsolutePathName(),
+                false,
+                joint.get());
+        } catch (const std::runtime_error& e) {
+            cout << "Duplicate Joint names are not permitted. Please update "
+                 << "your model's .osim file accordingly." << std::endl;
+            throw e;
+        }
     }
 }
 
--- a/OpenSim/Simulation/Model/Model.cpp
+++ b/OpenSim/Simulation/Model/Model.cpp
@@ -666,9 +666,8 @@ void Model::createMultibodyTree()
                                     body.getMass(), false,
                                     const_cast<Body*>(&body) );
         } catch (const std::runtime_error& e) {
-            cout << "Duplicate Body names are not permitted. Please update "
-                 << "your model's .osim file accordingly." << std::endl;
-            throw e;
+            OPENSIM_THROW_FRMOBJ(MultibodyGraphMakerFailed,
+                                 std::string(e.what()));
         }
     }
 
@@ -711,9 +710,8 @@ void Model::createMultibodyTree()
                 false,
                 joint.get());
         } catch (const std::runtime_error& e) {
-            cout << "Duplicate Joint names are not permitted. Please update "
-                 << "your model's .osim file accordingly." << std::endl;
-            throw e;
+            OPENSIM_THROW_FRMOBJ(MultibodyGraphMakerFailed,
+                                 std::string(e.what()));
         }
     }
 }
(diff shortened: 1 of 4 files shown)


OpenSim/Simulation/Model/Model.cpp
```diff
@@ -661,9 +661,15 @@ void Model::createMultibodyTree()
...
-        _multibodyTree.addBody( body.getAbsolutePathName(),
-                                body.getMass(), false,
-                                const_cast<Body*>(&body) );
+        try {
+            _multibodyTree.addBody( body.getAbsolutePathName(),
+                                    body.getMass(), false,
+                                    const_cast<Body*>(&body) );
+        } catch (const std::runtime_error& e) {
+            cout << "Duplicate Body names are not permitted. Please update "
+                 << "your model's .osim file accordingly." << std::endl;
```
chrisdembia: Could one get this message when building a model programmatically (rather than loading an .osim file)? Consider adding to the message something like "

Is it that names must not be duplicated, or that absolute path names may not be duplicated? The message makes sense for pre-4.0 models, but is not really correct for 4.0 models.

Is there a reason you chose to send a message to cout rather than throw a new exception with a more specific message?

tkuchida: > Could one get this message when building a model programmatically (rather than loading an .osim file)? Consider adding to the message something like "

I thought that would be caught elsewhere. I thought this issue had been investigated so just added a slightly modified version of the text @jenhicks requested in #1768. I will investigate thoroughly and revise as necessary.

> Is it that names must not be duplicated, or that absolute path names may not be duplicated? The message makes sense for pre-4.0 models, but is not really correct for 4.0 models.

The string being passed to Simbody's MultibodyGraphMaker is actually the absolute path name, so I agree that the message requested in #1768 is not complete. I will revise.

> Is there a reason you chose to send a message to cout rather than throw a new exception with a more specific message?

The request in #1768 was to "add to the message…". OpenSim doesn't have control over the errors that Simbody throws. In theory, Simbody could throw a `runtime_error` from `addBody()` (or one of the methods it calls) with some other message, so I think it's safest to throw the original error after adding an OpenSim-specific message. Actually, the message from OpenSim should say "Perhaps there are duplicate names…" or something like that just in case Simbody threw a `runtime_error` for some other reason. I will revise.

chrisdembia: > I thought that would be caught elsewhere. 

Ah...that could be true; I'm not sure.


> The request in #1768 was to "add to the message…".

The cout could be just fine. In other places (e.g., [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Component.cpp#L233)), we throw a new exception like so:

```
throw Exception("Can't have duplicate names (details: " + x.what() + ").");
```

tkuchida: > In other places (e.g., here), we throw a new exception like so:

I am familiar with that paradigm, but I believe it is a different situation because the original exception was thrown by OpenSim's `findAndConnect()`. In model.cppL668, the original exception was thrown by Simbody. The exception is a generic `runtime_error` and OpenSim cannot be certain (now and in the future) that the exception was thrown because a duplicate body name (absolute path name) was found.

There are several possible designs here. One option is to assume the `runtime_error` was thrown because a duplicate name was found and re-throw an OpenSim-specific exception. A second option is to search for the word "duplicate" in the string associated with Simbody's exception—proceed as in the first option if found; otherwise, add a message like "perhaps there are duplicate absolute path names" and then throw the original exception. Please let me know which you prefer, or if you prefer a third option.

chrisdembia: Thanks for explaining. I like your suggestion in your previous post: simply clarify the cout message to say "perhaps there are duplicate..." and rethrow the same exception.

tkuchida: Will do. Actually, I don't think this error-checking strategy is correct: OpenSim shouldn't be relying on Simbody to report that the body-absolute-path-name-uniqueness requirement has been violated; this check should be done in `Component::finalizeFromProperties()` or similar. Created #1876 to track.

chrisdembia: Ah good point. I agree.

tkuchida: I've revised to throw a new OpenSim exception as you suggested. The new exception displays the "ensure absolute path names are unique" message and then appends the original exception message verbatim, making no assumption about why the MultibodyGraphMaker actually complained. A problem with my original design is that the `cout` will be displayed even if some other OpenSim code catches and adequately handles the `runtime_error`.

tkuchida:
Revised to catch `runtime_error` thrown by MultibodyGraphMaker and re-throw as `OpenSim::MultibodyGraphMakerFailed` exception. Added test case and updated note in CHANGELOG. Ready for re-review.

aymanhab:
My main concern with throwing exception at the API level is if these are not caught in context and end up with mysterious "something wrong happened" or red bubble in the application corner. Does this method get called from Model's constructor or initSystem? If either we should have the signature in Java indicate that it throws for proper handling as discussed in other issues in gui repo.

aymanhab:
There's also an earlier discussion about name validator that works across component types to avoid special characters that are problematic to our internal algorithms (spaces, / etc.) while this applies only to Bodies and Joints as far as I can tell.

tkuchida:
> My main concern with throwing exception at the API level is if these are not caught in context and end up with mysterious "something wrong happened" or red bubble in the application corner. Does this method get called from Model's constructor or initSystem? If either we should have the signature in Java indicate that it throws for proper handling as discussed in other issues in gui repo.

Yes, the underlying method is called via `initSystem()`. This PR catches `runtime_error` exceptions thrown by Simbody and re-throws as `MultibodyGraphMakerFailed`.

> There's also an earlier discussion about name validator that works across component types to avoid special characters that are problematic to our internal algorithms (spaces, / etc.) while this applies only to Bodies and Joints as far as I can tell.

You are correct; duplicate names should be handled there. I will add a note to #1768.

chrisdembia:
Perhaps we could discuss in a dev meeting whether to take a short or long term approach here.

tkuchida:
> Perhaps we could discuss in a dev meeting whether to take a short or long term approach here.

I agree with @aymanhab: the exception should be caught upstream of `Model::createMultibodyTree()`, as I've noted above and in the proposed test case. I must have misinterpreted @jenhicks's comment on #1768; I thought she wanted a message added to the `runtime_error` exception thrown by Simbody.

Per @aymanhab's concern above: if we don't address this completely for 4.0, we should at least make sure the GUI properly handles the `runtime_error` exception that is currently thrown.

PR #1867, opened 2017-08-07
Properly override connectToModelAndBody in WrapDoubleCylinderObst.

### LINKED ISSUE #1864: WrapDoubleCylinderObst does not properly override connectToModelAndBody()
I think that, currently, WrapDoubleCylinderObst does not get added to a model properly, because it fails to override the correct method from the WrapObject base class, `WrapObject::connectToModelAndBody(Model& aModel, PhysicalFrame& aBody)`: In WrapDoubleCylinderObst, the second argument is `Body&` instead of `PhysicalFrame&`.

chrisdembia:
Fixes #1864

### Brief summary of changes

I changed the signature of `WrapDoubleCylinderObst::connectToModelAndBody()` so that it properly overrides the virtual function in the base class.

I did not go so far as making WrapDoubleCylinderObst work with PhysicalFrames that are not bodies; this would require more work. I simply wanted to fix the bug where WrapDoubleCylinderObst doesn't even work with bodies. 

### Testing I've completed

Ran all API tests.

### CHANGELOG.md (choose one)

- no need to update because PhysicalFrames are new to 4.0

### DIFF
--- a/OpenSim/Simulation/Wrap/WrapDoubleCylinderObst.cpp
+++ b/OpenSim/Simulation/Wrap/WrapDoubleCylinderObst.cpp
@@ -142,8 +142,12 @@ void WrapDoubleCylinderObst::setupProperties()
 *
 * @param aModel simbody model
 */
-void WrapDoubleCylinderObst::connectToModelAndBody(Model& aModel, OpenSim::Body& aBody)
+void WrapDoubleCylinderObst::connectToModelAndBody(Model& aModel,
+        PhysicalFrame& aBody)
 {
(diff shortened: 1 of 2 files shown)


chrisdembia:
No I was trying to get rid of compiler warnings. Can we merge this and consider the removal of the class as a separate issue?

aymanhab:
Changes look safe @chrisdembia feel free to merge. But I agree with @aseth1 regarding removal of the class.

chrisdembia:
Thanks @aymanhab . I created #1890, and I'm merging this PR.

PR #1872, opened 2017-08-07
Remove unused updateGeometry() in Actuator to avoid warning with clang

chrisdembia:
### Brief summary of changes

This function `Actuator::updateGeometry()` was not called anywhere in the API, and caused a "hidden
override" warning with clang (with `Muscle::updateGeometry(const
SimTK::State&)`). This function is used in two places in the GUI source code,
but both of those places are commented out.

For consistency, I also removed `WrapObject::updateGeometry()`.

All of the functions I removed had empty implementations.

### Testing I've completed

Ran API tests.

@aymanhab is it okay to remove these functions?

### CHANGELOG.md (choose one)

- no need to update because...it's unlikely users called this function, since it doesn't seem to do anything right now.


aymanhab:
Although muscle/path editing has not been completed yet I don't expect these functions to be needed, or will address if the need comes up, so I'm ok with removal. Thanks @chrisdembia

chrisdembia:
Thanks @aymanhab

chrisdembia:
Thanks @aseth1 and @aymanhab

PR #1873, opened 2017-08-07
Fix clang's "hidden override" warning for ForceSet::set.

chrisdembia:
### Brief summary of changes

Get rid of a warning from clang about `ForceSet::set()` hiding a virtual function from the base class. I fixed the issue by giving `set()` the same signature as from the base class. 

I ensured to preserve the same default value for `preserveGroups` as is used in the base class, so that users' existing code won't be affected by this change.

### Testing I've completed

Ran API tests.

### CHANGELOG.md (choose one)

- no need to update because...this is a minor change that changes the API in a very minimal way.

### DIFF
--- a/OpenSim/Simulation/Model/ForceSet.h
+++ b/OpenSim/Simulation/Model/ForceSet.h
@@ -99,7 +99,19 @@ public:
     bool append(Force &aForce);
 #endif
     bool append(ForceSet &aForceSet, bool aAllowDuplicateNames=false);
-    bool set(int aIndex, Force *aForce);
+    /** Set the force at an index.  A copy of the specified actuator is NOT
+    * made. The force previously set at the index is removed (and deleted).
+    *
+    * @internal This method overrides the method in ModelComponentSet<Force> so 
+    * that several internal variables of the set can be updated.
+    *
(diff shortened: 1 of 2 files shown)


chrisdembia:
Thanks @aseth1 .

OpenSim/Simulation/Model/ForceSet.h
```diff
@@ -99,7 +99,19 @@ OpenSim_DECLARE_CONCRETE_OBJECT(ForceSet, ModelComponentSet<Force>);
...
     bool append(ForceSet &aForceSet, bool aAllowDuplicateNames=false);
-    bool set(int aIndex, Force *aForce);
+    /** Set the force at an index.  A copy of the specified actuator is NOT
+    * made. The force previously set at the index is removed (and deleted).
+    *
+    * @internal This method overrides the method in ModelComponentSet<Force> so 
+    * that several internal variables of the set can be updated.
+    *
+    * @param aIndex Array index where the actuator is to be stored.  aIndex
+    * should be in the range 0 <= aIndex <= getSize();
```
tkuchida: `;` -> `.`

chrisdembia:
Thanks @tkuchida

PR #1874, opened 2017-08-07
Avoid "overloaded-virtual" warning in muscle curves (using SmoothSegmentedFunctions)

chrisdembia:
### Brief summary of changes

Some of the muscle curves have their own signature for `calcDerivative()`, but this hides the virtual function `calcDerivative()` from the base `Function` class. This means that calling `Function::calcDerivative()` on a function whose derived type is, for example, `ActiveForceLengthCurve`, may not do what's intended (not using the same `SimTK::Function` as is used by `calcValue()`, etc.).

This PR implements the virtual `calcDerivative()` from the base class, and defines it in a way that is consistent with the definitions for `calcValue()`, etc.

### Testing I've completed

Ran API tests.

### CHANGELOG.md (choose one)

- no need to update because...this PR has a minimal effect on the API.
- updated...

### DIFF
--- a/OpenSim/Common/SmoothSegmentedFunction.h
+++ b/OpenSim/Common/SmoothSegmentedFunction.h
@@ -122,9 +122,15 @@ namespace OpenSim {
        */
        double calcDerivative(double x, int order) const;       
 
-       
+       /// Allow the more general calcDerivative from the base class to be used.
+       // This helps avoid the -Woverloaded-virtual warning with Clang.
+       // We could have also put this `using` line in ActiveForceLengthCurve,
+       // etc., but that would be inconsistent with how the
+       // SmoothSegmentedFunction is used (e.g., calcValue() delegates to the
+       // internal `m_value`).
+       using Function_<double>::calcDerivative;
 
-     
+
        /**This will return the value of the integral of this objects curve 
        evaluated at x. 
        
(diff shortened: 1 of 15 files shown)


OpenSim/Common/SmoothSegmentedFunction.h
```diff
@@ -122,9 +122,15 @@ namespace OpenSim {
...
        double calcDerivative(double x, int order) const;       
 
-       
+       /// Allow the more general calcDerivative from the base class to be used.
+       // This helps avoid the -Woverloaded-virtual warning with Clang.
+       // We could have also put this `using` line in ActiveForceLengthCurve,
+       // etc., but that would be inconsistent with how the
+       // SmoothSegmentedFunction is used (e.g., calcValue() delegates to the
+       // internal `m_value`).
+       using Function_<double>::calcDerivative;
```
chrisdembia: Without this line, the `calcDerivative(double x, int order)` line above would cause the needed overload (the one that takes `derivComponents`) to be hidden.

aseth1: Maybe I missed something, but don't we still need an implementation of the base override that calls the local `calcDerivative(x, order)`?

chrisdembia: Yes we do; it's in the `private` section below.

aseth1: 👍

aseth1:
A **quick review** didn't tempt you @carmichaelong?

chrisdembia:
Thanks!

PR #1879, opened 2017-08-10
In FileAdapter::readFile(), support v1.0 STO files.

### LINKED ISSUE #1725: "DataType not specified" when reading version-1 STO file
Attempting to read a version=1 STO file (see attached) using `import opensim; opensim.FileAdapter.readFile('<file-name>')` (in Python) causes the following error:

```
RuntimeError: std::exception in 'OpenSim::DataAdapter::OutputTables OpenSim::FileAdapter::readFile(std::string const &)': DataType not specified in the header.
        In file /Users/chris/repos/opensim/13/opensim-core/OpenSim/Common/STOFileAdapter.cpp:62
        In function 'createSTOFileAdapterForReading'
```

This is because the function `createSTOFileAdapterForReading(const std::string& fileName)` checks for a DataType metadata field.  I think if the STO file does not have a version number (that is, its version is 1), then it is parsed as if DataType is double.

[testGait10dof18musc_netgeneralizedforces.sto.zip](https://github.com/opensim-org/opensim-core/files/1016502/testGait10dof18musc_netgeneralizedforces.sto.zip)

chrisdembia:
Fixes #1725 

### Brief summary of changes

I modified the way `FileAdapter::readFile()` determines which file adapter to use so that version-1.0 STO files are supported. Previously, only version-2.0 files were supported.

### Testing I've completed

Added a test case. I ensured that the test case failed before I introduced the fix.

### CHANGELOG.md (choose one)

- no need to update because...data adapters are new to 4.0

### DIFF
--- a/OpenSim/Common/STOFileAdapter.cpp
+++ b/OpenSim/Common/STOFileAdapter.cpp
@@ -58,8 +58,10 @@ createSTOFileAdapterForReading(const std::string& fileName) {
             }
         }
     }
-
-    OPENSIM_THROW(STODataTypeNotFound);
+    // The file does not seem to have a DataType field, and is therefore likely
+    // a version 1.0 STO file. These files only supported double as the column
+    // data type; try to read the file with type double.
+    return std::make_shared<STOFileAdapter_<double>>();
 }
 
 std::shared_ptr<DataAdapter>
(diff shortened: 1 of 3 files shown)


chrisdembia:
test failure is unrelated.

carmichaelong:
LGTM. Merging.

chrisdembia:
Thanks @carmichaelong

PR #1885, opened 2017-08-17
Speed up comparison of states in testStatesTrajectory.

chrisdembia:
### Brief summary of changes

This PR speeds up one of the test cases in testStatesTrajectory by avoiding repeated calls to `getStateVariableValue()`. The tests used to take 25 seconds, but now runs in 9 seconds.

### Testing I've completed

I ensured that the tests pass, and that inserting a mismatch between the states and the exported table causes the test to fail.

### CHANGELOG.md (choose one)

- no need to update because...this is internal.


OpenSim/Simulation/Test/testStatesTrajectory.cpp
```diff
@@ -717,18 +717,23 @@ void tableAndTrajectoryMatch(const Model& model,
...
 
+        stateValues = model.getStateVariableValues(states[itime]);
+
         // Test state values.
         for (size_t icol = 0; icol < table.getNumColumns(); ++icol) {
             const auto& stateName = colNames[icol];
 
-            const auto& valueInStates = model.getStateVariableValue(
-                    states[itime], stateName);
+            const auto stateIndex = stateNames.findIndex(stateName);
```
aseth1: If the indices aren't going to change from State to State, you could create hold on to indices either as a flat list or a map keyed by name. `Array::findIndex()` is linear in number of elements and is doing string matching.

chrisdembia: Ah good point. I'll try to make that change.

chrisdembia:
@aseth1 I've updated the PR and done what you suggested. Ready for review again.

chrisdembia:
@aseth1 could you take a quick look at this again?

Issue #1884, opened 2017-08-16
Component::setStateVariableValues() enforces constraints with every setValue()


tkuchida:
From discussion with @chrisdembia.

When setting the values of Coordinates, we want to enforce constraints only after all Coordinates have been set. Enforcing constraints with each `setValue()` is not just inefficient: it will also be mangling the newly set Coordinate values along the way, resulting in a final pose that could be far from what's expected. [This code](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Tools/CMCTool.cpp#L652) does it correctly. `Component::setStateVariableValues()` [seems to be enforcing constraints with every call to `setValue()`](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Component.cpp#L944). I think Component.cpp#L944 needs to be
```cpp
_allStateVariables[i]->setValue(state, values[i], false);
```
when setting the value of a Coordinate, with a call to project/realize after L945.

chrisdembia:
I would really like to discuss this in the next dev meeting.

chrisdembia:
Here are some options for solving this issue (discussed with @tkuchida):

1. Have `Coordinate::CoordinateStateVariable::setValue()` pass `false` for the `enforceConstraints` parameter of `Coordinate::setValue()`. We would edit the documentation of `Component::setStateVariableValues()` to say that users must call `model.assemble()`, etc. on their own to enforce the constraints.
2. Have `Component::setStateVariableValues()` take a `enforceConstraints` parameter, and enforce the constraints only after setting all state variable values. This could be achieved by giving `Component` a virtual function that `Model` would implement by calling `assemble()`. I don't really like this idea. How is a component supposed to know how to enforce constraints?
3. Permit users to have a way to set state variable values without using `Component::setStateVariableValues()`. That is, help users use `State::updY()` directly. I see 2 options:

    1. provide a method that gives the names of the state variables in the order of `State::getY()`.
    2. Implement `Component::StateVariable::getSystemYIndex()` and expose these global indices to users, perhaps via `Component::getSystemYIndices()` (similar to `Component::getStateVariableNames()`).

tkuchida:
[...]

PR #1886, opened 2017-08-17
[WIP] Avoid repeated assembly in StatesTrajectory::createFromStatesStorage()

chrisdembia:
### Brief summary of changes

This PR avoids repeatedly assembling the states (via `Coordinate::setValue()`) within `StatesTrajectory::createFromStatesStorage()`. The repeated assemblies were causing `createFromStatesStorage()` to take many minutes to run for the Rajagopal 2016 model (with 2 coordinate coupler constraints), and this PR fixes this performance issue.

### Testing I've completed

Ran ctests.

 There are multiple options for how we can avoid assembling repeatedly, which I discussed with @tkuchida this week. This is related to #1884 (this PR is a workaround for #1884).

1. Have `Coordinate::CoordinateStateVariable::setValue()` pass `false` for the `enforceConstraints` parameter of `Coordinate::setValue()`. We would edit the documentation of `Component::setStateVariableValues()` to say that users must call `model.assemble()`, etc. on their own to enforce the constraints.
2. Have `Component::setStateVariableValues()` take a `enforceConstraints` parameter, and enforce the constraints only after setting all state variable values. This could be achieved by giving `Component` a virtual function that `Model` would implement by calling `assemble()`. I don't really like this idea. How is a generic component supposed to know how to enforce constraints?
3. Permit users to have a way to set state variable values without using `Component::setStateVariableValues()`. That is, help users use `State::updY()` directly. Then users can call `assemble()` on their own. I see 2 options:

    1. provide a method that gives the names of the state variables in the order of `State::getY()`.
    2. Implement `Component::StateVariable::getSystemYIndex()` and expose these global indices to users, perhaps via `Component::getSystemYIndices()` (similar to `Component::getStateVariableNames()`).

(I've copied these options to #1884.)

In this PR, I've chosen to go with 3(i). I think we should do 3(i) regardless of how we solve #1884. I also like option 1.

I don't actually want to merge this PR yet. Rather, I want to discuss the options and to get opinions about my implementation of `createSystemYIndices()`. If we like `createSystemYIndices()`, then I would want to move it to `Model`.

### CHANGELOG.md (choose one)

- no need to update because...StatesTrajectory is new to 4.0.

### DIFF
--- a/Bindings/Python/tests/test_states_trajectory.py
+++ b/Bindings/Python/tests/test_states_trajectory.py
@@ -49,7 +49,6 @@ class TestStatesTrajectory(unittest.TestCase):
     def test_modify_states(self):
         model = osim.Model(os.path.join(test_dir,
             "gait10dof18musc_subject01.osim"))
-        model.initSystem()
 
         states = osim.StatesTrajectory.createFromStatesStorage(
                 model, self.states_sto_fname)
@@ -57,6 +56,7 @@ class TestStatesTrajectory(unittest.TestCase):
         states[0].setTime(4)
         assert states[0].getTime() == 4
 
+        model.initSystem()
         self.assertNotAlmostEqual(model.getStateVariableValue(states[2],
                 "ground_pelvis/pelvis_tilt/value"), 8)
         model.setStateVariableValue(states[2],
@@ -78,7 +78,6 @@ class TestStatesTrajectory(unittest.TestCase):
         # wrapping works.
         model = osim.Model(os.path.join(test_dir,
             "gait10dof18musc_subject01.osim"))
-        model.initSystem()
         sto = osim.Storage(self.states_sto_fname)
         states = osim.StatesTrajectory.createFromStatesStorage(
                 model, sto)
--- a/OpenSim/Simulation/StatesTrajectory.cpp
+++ b/OpenSim/Simulation/StatesTrajectory.cpp
@@ -166,6 +166,30 @@ TimeSeriesTable StatesTrajectory::exportToTable(const Model& model,
     return table;
 }
 
+/** The map provides the index of each state variable in
+SimTK::State::getY() from its each state variable path string. */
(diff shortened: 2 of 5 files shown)


aymanhab:
@chrisdembia This is the essence of the issue we ran into before ISB where states play smoothly in the GUI (since they don't assemble at least for a model with no constraints) while coordinates force assembly and lead to choppy playback. We need to coordinate so that GUI can leverage your work on this Issue.

aymanhab:
I do prefer option 3 as well as most flexible, I may have asked for it earlier (use case in the GUI you may have a motion that changes only a few state variables, being able to selectively set these would be great).

aseth1:
My preference is to not expose `systemYindices` to users. To me that is "passing-the-buck" to do the book-keeping. I am totally OK with us getting and managing indices internally so that `Component::setStateVariableValues()` is super fast by doing a an `updY()` internally with the states in order or `Component::getStateVariableNames()`. Then users can maintain a an Array (or std::vector) of state values (say in MATLAB) and update quickly without having to know anything about system state Y indices.

chrisdembia:
>  To me that is "passing-the-buck" to the user to do the book-keeping.

Makes sense, but you could also view it as a function for advanced users who are trying to get rid of all possible overhead during direct collocation. I'll do some testing to see if we can get by with simply speeding up` setStateVariableValues()`.

aseth1:
If you maintain the indices internally to reorder *stateVariableValues* `Vector` to match `state.updY()` `Vector`, then  reordering should be negligible fixed cost especially if no new allocations are required. If we expose and further facilitate accessing `State` entries directly then users are likely to do it more, and once they have code that depends on `Y` and `Z` order they are higher risk of their code breaking and lose the advantages of using OpenSim for context.

chrisdembia:
@aseth1 has an alternative fix in #1911.

chrisdembia:
I'm not deleting the branch yet becase it might still be useful.

Issue #1877, opened 2017-08-07
runtime_error exception throws red-circle "unexpected exception" in GUI


tkuchida:
The `runtime_error` thrown by Simbody's MultibodyGraphMaker is not wrapped and leads to a red-circle exception in the GUI when attempting to load a model that contains bodies/joints with non-unique absolute path names. See #1768 and #1866.

<pre>
java.lang.RuntimeException: addBody(): Duplicate body name '/subject01/tibia'
	at org.opensim.modeling.opensimSimulationJNI.Model_initSystem(Native Method)
	at org.opensim.modeling.Model.initSystem(Model.java:840)
	at org.opensim.view.pub.OpenSimDB.addModel(OpenSimDB.java:144)
	at org.opensim.view.pub.OpenSimDB.addModel(OpenSimDB.java:139)
	at org.opensim.view.FileOpenOsimModelAction$2.run(FileOpenOsimModelAction.java:100)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:756)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:76)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:726)
	at org.netbeans.core.TimableEventQueue.dispatchEvent(TimableEventQueue.java:159)
[catch] at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:201)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:105)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:101)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:93)
	at java.awt.EventDispatchThread.run(EventDispatchThread.java:82)
</pre>

[...]

Issue #1876, opened 2017-08-07
Improve checks for naming Components


tkuchida:
- For as long as OpenSim uses its own methods for parsing path strings, Component names containing characters like `/`, `.`, and `:` are prohibited.
- A Body cannot be named "Ground". (addressed in #2285).
- Components must have unique absolute path names.

These checks could be added to `{Component|Model}::finalizeFromProperties()` and provide helpful user-facing messages.

Related: #337, #1866.

aseth1:
I might separate the issue into:
1.  catch invalid form of a `ComponentPath` (e.g. flagging illegal characters) 
2.  inform the user of ambiguity due to multiple components with the same path (upon deserialization)
3.  prevent `Ground` component from being renamed
4.  flag "Ground" (and its variants) as an illegal `Body` name

tkuchida:
Perhaps we should include in :four: a check for illegal characters in the `name` property string.

Issue #1192, opened 2016-08-17
Assign a unique ID to each Component


tkuchida:
### What is a unique ID?
A [universally unique identifier (UUID)](https://en.wikipedia.org/wiki/Universally_unique_identifier) is a 128-bit value like `123e4567-e89b-12d3-a456-426655440000`. UUIDs can be generated randomly with near certainty of uniqueness using
- [Graeme Hill's CrossGuid library](https://github.com/graeme-hill/crossguid) (uses MIT license)
- [Boost's "Uuid" library](http://www.boost.org/doc/libs/1_61_0/libs/uuid/uuid.html) (but requires Boost)
- ["UuidCreate" on Windows](https://msdn.microsoft.com/en-us/library/windows/desktop/aa379205(v=vs.85).aspx) (but not cross-platform)

### What problems would be solved?
The current API relies on user-modifiable names (string properties) to form connections. This design follows most naturally from the 3.3 API, but introduces some potential concerns:
 1. Components on the same path must have unique names at all times. This requirement puts constraints on the design of the GUI and the user experience (e.g., automatically suffix top-level Component name with "_1" after copying or delay uniqueness check) and exposes the underlying API architecture/data structures to the GUI-level user.
 2. Renaming a Component invalidates connections.
 3. Components must be named. Perhaps the name of the Component is unimportant to the user (e.g., if creating and connecting mathematical operation blocks using the future GUI model builder).
 4. Component names cannot contain special characters (see #596). Currently, we (should) disallow ".", "/", "\", " "; a path-handling class could help (see #1069), but we would still need to consider serialization/deserialization of special characters.

### What is the proposal?
 1. Assign a UUID to each Component upon creation (and deserialization, if none exists).
 2. The UUID would be constant so connections would be unaffected by Component renaming. (Modifying properties like "name" and "comment" should not trigger `finalizeFromProperties()`, which should be reserved for modifications that would affect the behavior of the Model.)
 3. The UUID would be serialized, which could help users extend models in xml (e.g., by inserting the UUID to form connections rather than assembling full paths).
 4. It might be helpful for the GUI (now and for eventual graphical model building) to hold onto a unique ID tag for each Component, regardless of its name, path, or connection status.
 5. Continue supporting lookup by name and full path for backwards compatibility and usability, but map to UUID behind the scenes.

### See also
[...]

PR #1887, opened 2017-08-18
Ensure unique subcomponent names

aseth1:
Fixes issues #1768,  #1877, #1876 (pt 2.)  
Currently, a `Model` with subcomponents  that have duplicate (path) names loads fine and it is only when Model connections are being resolved that errors including run_time exceptions within Simbody are encountered. 

### Brief summary of changes
In this PR the problem of duplicate path names is addressed directly after `Model` loading/deserialization. `Component::finalizeFromProperties` now checks for duplicate named subcompomponents of the same type. If encountered a message is printed and it is renamed to be unique.

### Testing I've completed
All tests pass. I also added an `SubcomponentsWithDuplicateName` exception and verified that duplicates were being flagged correctly before commenting out the exception and replacing it with a console message. I left the commented out lines and tests that throw because we should move to strict enforcement after a grace period of using 4.0 where models will be brought into compliance.

### CHANGELOG.md (choose one)
- updated 


CHANGELOG.md
```diff
@@ -97,6 +98,8 @@ Bug Fixes
...
 - Fixed a bug where MuscleAnalysis was producing empty moment arm files. We now avoid creating empty Moment and MomentArm storage files when `_computeMoments` is False. (PR #324)
 - Fixed bug causing the muscle equilibrium solve routine in both Thelen2003Muscle and Millard2012EquilibriumMuscle to fail to converge and erroneously return the minimum fiber length. The fix added a proper reduction in step-size when errors increase and limiting the fiber-length to its minimum. (PR #1728)
+- Fixed a bug where Models with Bodies and Joints (and other component types) with the same name were loaded without error. Duplicate names Bodies were simply being ignored and only the first Body of that name in the BodySet was being used, for example, to connect a Body to its parent via its Joint, or to affix path points to its respective Body. Now, duplicate names are flagged and renamed so they are uniquely identified. (PR #1887)
```
tkuchida: "Duplicate names Bodies" -> "Duplicately named Bodies"

OpenSim/Common/Component.cpp
```diff
@@ -192,6 +192,39 @@ void Component::finalizeFromProperties()
...
+            //OPENSIM_THROW(SubcomponentsWithDuplicateName, getName(), name);
+            
+            //for now rename the duplicately named Component
+            Component* mutableSub = const_cast<Component *>(sub.get());
+            mutableSub->setName(name + "_" + std::to_string(count));
+
+            // Warn of the problem
+            std::string msg = type + " '" + getName() + "' has subcomponents " +
+                "with duplicate name '" + name + "', and was renamed as '" +
+                sub->getName() + "'.";
```
tkuchida: This message says that the *parent* of the twins was renamed when in fact it's the second twin that gets renamed.

OpenSim/Common/Component.cpp
```diff
@@ -192,6 +192,39 @@ void Component::finalizeFromProperties()
...
+            std::cout << msg << std::endl;
+
+            names.insert(type + name);
```
tkuchida: Not sure why this isn't caught by a test, but I don't think this code will work. The `else` block (starting at L208) catches instances where `type+name` already appears in the `names` set. The `type` and `name` variables aren't modified within the block, so this `insert()` call should still fail.

tkuchida: The new name (`name + "_" + std::to_string(count)`) isn't guaranteed to be unique. For example, a user could convert the model and then later do some copy-and-paste editing where they specify a name of the form `name + "_" + <integer>` that has already been used. Perhaps use a `while(1)` that increments `count` until `type + newName` can be inserted.

tkuchida: Also, if `count` is initialized to `0` before this `while(1)`, non-unique names like "p1, p2, p2, p2" will be renamed "p1, p2, p2_0, p2_1" rather than "p1, p2, p2_17, p2_18" (or however many subcomponents have already been visited).

aseth1: I agree. I like your solution. Thanks!

aseth1:
> Also, I think it is necessary to guarantee that names are unique for subcomponents of any type. The post office needs a unique address for every destination, regardless of whether it's a business, residence, P.O. box, etc. Similarly, getComponent() doesn't care about type so I think every Component needs a unique absolute path.

`getComponent()` is templated by Component type `<C>` and it is used to locate a `Socket`'s connectee by name and type.

tkuchida:
> `getComponent()` is templated by Component type `<C>` and it is used to locate a `Socket`'s connectee by name and type.

Okay. I don't understand the benefit of defining a rule that "each Component must have an absolute path that is different from the absolute path of all other Components of the same type" rather than simply "all Components must have a unique absolute path". The first rule is fine as long as Components are never requested or referred to using just their absolute path.

Also, thanks for clarifying this. We should document that, in place of `Body.getName()`, the unique key for a Component in 4.0 is `Component.getConcreteClassName() + Component.getAbsolutePathName()`.

aseth1:
> I don't understand the benefit of defining a rule that "each Component must have an absolute path that is different from the absolute path of all other Components of the same type" rather than simply "each Component must have a unique absolute path". The first rule is fine as long as Components are never requested or referred to using just their absolute path.

My response was simply to show that we do use the type to accurately locate components that satisify `Socket`s. But that doesn't that the stricter/simpler rule shouldn't b enforced. And now is a good time to enforce it if we want to.  In the past we had models with Bodies and Joints with same name (like *floor* `Joint` and a *floor* `Body`, or Amy had a model with a *toes* `Joint` and a *toes*, `Body` and the type helped in these cases.  I still think keeping the type of Sockets makes sense.

Without access to templated `getComponent<C>()` in scripting I can see where ambiguity comes in and in this case getting the *toes* `Body` may not be what the user wants (e.g. wants the `Joint`).  If we move to all immediate subcomponents being uniquely named, would you want the rename to reflect its type?
It does seem simpler to say/explain that all Components must have a unique path.

In your case, if you are always accessing Bodies (or PhysicalFrames) I don't think adding the type to your keynames is necessary.

jenhicks:
I'm trying to think through the pros and cons of both approaches (around upgrade headaches for users, conceptual simplicity, etc.). One question that comes to mind ... if every Component has a unique name, what is the point then of Component path names?  

For example, if as a user, I'm supposed to think about a component as body/floor, it then it seems a little odd that OpenSim would tell me that I'm supposed to call this component body/floor_body.

tkuchida:
I'm still not sure about this. What happens if a concrete class inherits from another concrete class and instances of these classes have the same absolute path? Example: Suppose I create a concrete class named `Primate` that inherits from `Component` and a concrete class named `Monkey` that inherits from `Primate`. Now I create a Model containing a Primate named "banana" and a Monkey named "banana". Further suppose that there's a Component somewhere in the Model with a Socket that requires a Primate, and I specify that I want to plug "model/banana" into the Socket. Which Component gets connected: the Primate or the Monkey? What would be returned by `model.getComponent<Primate>("model/banana")`? It seems like one of the outcomes of this unique-name-ifier is to make `getComponent<type>(absolutePath)` always work, but it seems like `getComponent<Primate>("model/banana")` should still throw.

aseth1:
@jenhicks sorry for the loose wording.  I definitely meant absolute path name being above (and thus immediate subcomponents must be unique). So if there was a model named *Dennis* with a BodySet containing a *toes* `Body` and a  JointSet with a *toes* `Joint` then we have the non-unique path: */Dennis/toes* based on naming alone. For `Socket`s it hasn't been an issue (yet) since the type requirement will eliminate one or both. But @tkuchida raises a good point that if I have a `Socket<Force>`  and I have a `Spring` named *foo* and an `Actuator` named *foo* too, then with respect to the `Socket<Force>` it is undefined which of these two forces your Component will connect to. The `Component::findComponent<C>()` utility currently throws an exception if multiple components of the same `C` type are found, but the faster `getComponent<C>` returns what it finds at that path location as long as it is the correct type, which exposes the problem.  So we have two choices:

1. Make getComponent<C>() check for multiple matches and throw if multiple are found (like findComponent does). Or,
2. Adopt a stricter naming convention that says all immediate subcomponents (and thus their absolute path) must be unique, as Tom was advocating. 

I am leaning towards 2. since it keeps the code simpler and is also easier to explain and is safer for scripters who don't have access to the type templated `getComponent<C>()` variants. I can enforce that in this PR if we agree.

tkuchida:
> 1. Make `getComponent<C>()` check for multiple matches and throw if multiple are found (like `findComponent` does). Or,
> 2. Adopt a stricter naming convention that says all immediate subcomponents (and thus their absolute path) must be unique, as Tom was advocating.

> I am leaning towards 2. since it keeps the code simpler and is also easier to explain and is safer for scripters who don't have access to the type templated `getComponent<C>()` variants. I can enforce that in this PR if we agree.

Regarding option 1: I thought part of the distinction between `getComponent()` and `findComponent()` was that `getComponent()` just traverses the path to the Component (if found, return; otherwise, throw). This was a topic of discussion a while ago (IIRC, the original implementation of `getComponent()` called `findComponent()` rather than throwing); I believe the outcome of that discussion was that `getComponent()` should be fast and just alert the user (i.e., throw `ComponentNotFoundOnSpecifiedPath`) rather than call `findComponent()`. I still think it makes sense for `getComponent()` to do a straightforward tree traversal, and option 1 seems to contradict that design—albeit only partially, as only the Component's siblings would need to be queried. Nevertheless, I think this contradiction disqualifies option 1.

Regarding option 2: This option seems the most obvious to me; ease of explanation and error avoidance are other benefits, as you have pointed out.

Option 3: I don't know all the reasons behind the design choice to make `absolute_path + type` unique rather than just `absolute_path`, but a potential analogy is a file system: files in the same directory can have the same name, but their `path_and_filename + extension` must be unique. That is the best argument I have come up with to support enforcing only `absolute_path + type` uniqueness; however, the analogy reveals an issue: providing only an absolute path to a `Socket<Force>` is like asking for any runnable file (analogous to `Spring` and `Actuator` in your example). If you ask a file system for a runnable file of a particular name in a particular directory, you wouldn't be surprised if you got more than one file—but the files would have unique extensions (.bat, .exe, …). Thus, if you want to allow Components of different types to have the same absolute path, I think you would need to provide `Socket`s (and `Input`s, `Output`s, etc.) with the connectee's "extension" (type) along with its absolute path. This is a valid design, but it's more complicated and will require more explanation and code. Looking up words in a dictionary is another analogy—but, again, I can't give you a definition for a word unless you provide me with its part of speech as well. :book:

aymanhab:
I do support option 2 above as proposed by @aseth1 and @tkuchida. If we're enforcing new naming rule anyway we might as well make it simple and easy to enforce. In practice most users reuse our models old/new and if these can be handled properly we'll be well covered moving forward.  If we drop using a name as ID (per #1192) then this issue becomes relevant only to deserialization while all internal book keeping would be done using ID. It may help to have this idea embedded in the API by having a getID() that returns an opaque pointer so that code written by @tkuchida would be decoupled from how  we assign IDs in the future.

chrisdembia:
I think the reason why we considered making `path_name + type` unique rather than `path_name` unique is that multiple models have the same name for joints and bodies (perhaps `pelvis` for both the body and joint), and these would then no longer load error-free.

aseth1:
@chrisdembia please see the discussion above. You are right, that is the reason for enforcing type match, but it is not strong enough for `Socket<Force>` as @tkuchida and I have described.

jenhicks:
Option #2 sounds reasonable to me as well. It seems that we're all in agreement?

aseth1:
I will go ahead and revise the PR to implement option 2. Thanks!

tkuchida:
Not sure what documentation will need to be updated as a result of this design change. The doxygen for `Component::getAbsolutePathName()` already assumes option 2: "Accessing a Component by its absolutePathName from root is guaranteed to be unique."

aseth1:
I saw that and was happy that we made the code do what the documentation promises. 😮

aseth1:
@chrisdembia please let me know if you have any comments/issues for this PR.

CHANGELOG.md
```diff
@@ -86,7 +86,8 @@ a parent Component are accessible this way. The Model's typed %Sets and `add####
 are no longer necessary to compose a Model, since any Component can now be composed of
 components. `Model` still supports `addd####()` methods and de/serialization of Sets,
 but components added via `addComponent` are NOT included in the Sets but contained
-in the Component's *components* property list. Details in PR#1014.
+in the Component's *components* property list. Details in PR#1014. **Note**, it is now
+strictly required that immediate subcomponents of the same type have unique names. For example, a Model's `BodySet` cannot contain two bodies named *tibia* because it is  ambiguous as to which *tibia* `Body` a `Joint` or any other component is referring to.
```
chrisdembia: Could this Note be moved high up under `Converting from v3.x to v4.0`? Also, isn't the requirement even stronger now? Subcomponents of must have unique names, regardless of type?

OpenSim/Common/Component.cpp
```diff
@@ -192,6 +192,48 @@ void Component::finalizeFromProperties()
 
     markPropertiesAsSubcomponents();
     componentsFinalizeFromProperties();
+
```
chrisdembia: Do we need still need

```
  if (getName().empty()) {
         setName(IO::Lowercase(getConcreteClassName()) + "_");		         
     }		     
```

above? That seems redundant now.

Could you also document why this is the right place to put this code in `finalizeFromProperties()`? At first glance, it would seem preferable to put this new code above `markPropertiesAsSubcomponents()` (but I understand that you need to do that first to access the subcomponents).

OpenSim/Common/Component.cpp
```diff
@@ -192,6 +192,48 @@ void Component::finalizeFromProperties()
...
+    std::set<std::string> names{};
+    std::string name{};
+
+    // increment the count of duplicates to use as a unique suffix
+    int count{ 0 };
+    // temp variable to hold the unique name used to rename a duplicate
+    std::string uniqueName{};
+
+    for (auto& sub : subs) {
+        name = sub->getName();
```
chrisdembia: This copy is unnecessary; `getName()` returns a `const string&`.

OpenSim/Common/Component.cpp
chrisdembia: Also, could you add a high-level comment describing what this section of code does?

OpenSim/Common/Component.cpp
```diff
@@ -192,6 +192,48 @@ void Component::finalizeFromProperties()
...
+            // In the future this should become an Exception 
+            //OPENSIM_THROW(SubcomponentsWithDuplicateName, getName(), searchName);
+            // for now, rename the duplicately named subcomponent 
+            // but first test the uniqueness of the name (the while condition)
+            uniqueName = name + "_" + std::to_string(count++);
+        }
+
+        if (count > 0) { // if a duplicate
+            // rename the the subcomponent with its verified unique name
+            Component* mutableSub = const_cast<Component *>(sub.get());
```
chrisdembia: Consider adding a `std::vector<SimTK::ReferencePtr<Component>> getImmediateSubcomponents() const;`

OpenSim/Tests/VisualizeModel/VisualizeModel.cpp
```diff
@@ -60,12 +60,8 @@ int main(int argc, char **argv)
...
         getchar(); // Keep Visualizer from dying until we inspect the visualization window..
     }
-    catch (OpenSim::Exception ex)
-    {
-        std::cout << ex.getMessage() << std::endl;
-        return 1;
-    }
-    catch (std::exception ex)
+
+    catch (const std::exception& ex)
```
chrisdembia: +1

CHANGELOG.md
aseth1: Yes. Will update.

OpenSim/Common/Component.cpp
aseth1: It is redundant in that we could have `_0`, `_1`, `_2`, ... names for unnamed components if we do not first at the class name. I prefer having the class name appear in the no name case (e.g. pathwrap_#). I can remove the now unnecessary trailing underscore, though.

aseth1: Done.

OpenSim/Common/Component.cpp
aseth1: Good find.

OpenSim/Common/Component.cpp
aseth1: I think I will do that in a future PR and also use it to avoid cycling through the three internal lists. I was reluctant to add another accessor for this one use, but now I think I would use it.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ v4.0 (in development)
 
 Converting from v3.x to v4.0
 -----------------------------
+- A significant difference between v3.3 and 4.0 is the naming of dependencies. Unique names were not enforced in 3.3, which lead to undefined behavior. In 4.0, Component pathnames must be unique. That is a Component must be unique with respect to its peers. A Model named *model* cannot have multiple subcomponents with the name *toes* either as bodies or joints, because the pathname */model/toes* will not uniquely identify the Component. However, multiple *toes* bodies can be used as long as they are not subcomponents of the same Component. For example, a *device* Component with a *toes* Body will have no issues since this *toes* Body has a unique pathname, */model/device/toes*, which in unambiguous. One could also create multi-legged model, where each leg is identical, with *hip*. *knee* and *ankle* joints and *upper*, *lower* and *foot* bodies, but each being a unique because of Leg Component that contains the leg subcomponents, is uniquely named like */model/leg1* and */model/leg4/* and thus all of their subcomponents are unique, e.g.: */model/leg1/knee* vs. */model/leg4/knee*.
```
chrisdembia: Thanks for adding this helpful documentation. Some typos:

- `*hip*. *knee* and *ankle*` -> `*hip*, *knee* and *ankle*` (period -> comma)
- `being a unique` -> `being unique`


I think the second part could be more brief ("However, a model with two `knee` components is fine if the absolute pathnames are unique, e.g., `/model/left_leg/knee` and `/model/right_leg/knee`.").

OpenSim/Tests/VisualizeModel/VisualizeModel.cpp
tkuchida: Nice catch.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ v4.0 (in development)
 
 Converting from v3.x to v4.0
 -----------------------------
+- A significant difference between v3.3 and 4.0 is the naming of dependencies. Unique names were not enforced in 3.3, which lead to undefined behavior. In 4.0, Component pathnames must be unique. That is a Component must be unique with respect to its peers. A Model named *model* cannot have multiple subcomponents with the name *toes* either as bodies or joints, because the pathname */model/toes* will not uniquely identify the Component. However, multiple *toes* bodies can be used as long as they are not subcomponents of the same Component. For example, a *device* Component with a *toes* Body will have no issues since this *toes* Body has a unique pathname, */model/device/toes*, which in unambiguous. One could also create multi-legged model, where each leg is identical, with *hip* and *knee* joints and *upper* amd *lower* bodies, but each being unique because each `Leg` Component that contains the leg subcomponents, is uniquely named like */model/leg1* and */model/leg4/* and thus all of their subcomponents are unique, e.g.: */model/leg1/knee* vs. */model/leg4/knee*.
```
tkuchida: which lead to -> which led to
in unambiguous -> is unambiguous
create multi-legged model -> create a multi-legged model
amd &ast;lower&ast; -> and &ast;lower&ast;

aseth1: Thanks @tkuchida!

aseth1:
Thanks @tkuchida and @chrisdembia. Merging.

Issue #1845, opened 2017-07-27
Visual Studio reports warnings when building ExampleMain


tkuchida:
- built and installed opensim-core (Visual Studio 14 2015 Win64)
- copied the contents of OpenSim/Examples/ExampleMain into a new directory
- ran CMake (3.8.2); completed successfully though reported the following warning:
<pre>
CMake Warning (dev) at CMakeLists.txt:14 (if):
  Policy CMP0054 is not set: Only interpret if() arguments as variables or
  keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  Quoted variables like "MSVC" will no longer be dereferenced when the policy
  is set to NEW.  Since the policy is not set the OLD behavior will be used.
This warning is for project developers.  Use -Wno-dev to suppress it.
</pre>
- opened OpenSimTugOfWar.sln and built in RelWithDebInfo mode; Output window reports success though some warnings are displayed:
<pre>
1>------ Build started: Project: ZERO_CHECK, Configuration: RelWithDebInfo x64 ------
1>  Checking Build System
1>  CMake does not need to re-run because C:/Users/tkuchida/Documents/GitHub/test-buildAgainstOpenSimCoreInstall/build/CMakeFiles/generate.stamp is up-to-date.
2>------ Build started: Project: exampleMain, Configuration: RelWithDebInfo x64 ------
2>  Building Custom Rule C:/Users/tkuchida/Documents/GitHub/test-buildAgainstOpenSimCoreInstall/CMakeLists.txt
2>  CMake does not need to re-run because C:/Users/tkuchida/Documents/GitHub/test-buildAgainstOpenSimCoreInstall/build/CMakeFiles/generate.stamp is up-to-date.
2>  TugOfWar1_CreateModel.cpp
2>c:\users\tkuchida\documents\github\opensim-core-install\sdk\include\opensim\common\Component.h(2121): warning C4068: unknown pragma
2>c:\users\tkuchida\documents\github\opensim-core-install\sdk\include\opensim\common\Component.h(2122): warning C4068: unknown pragma
2>c:\users\tkuchida\documents\github\opensim-core-install\sdk\include\opensim\common\Component.h(2125): warning C4068: unknown pragma
[...]

PR #1895, opened 2017-08-24
Avoid sprintf in PropertyDblVec.

chrisdembia:
Fixes issue #1845 

### Brief summary of changes

Avoid using `sprintf` to get rid of a warning from MSVC.


chrisdembia:
Yay; thank you.

Issue #1883, opened 2017-08-16
Add comment to Component::getAbsolutePathName() about computational expense


tkuchida:
[The doxygen for `Component::getAbsolutePathName()`](http://myosin.sourceforge.net/1879/classOpenSim_1_1Component.html#a88a3212416a4d5a9387e395e54d4d8ae) doesn't indicate that [it assembles the absolute path on-the-fly](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Component.cpp#L641). Issues:

1. `getAbsolutePathName()` is close to `getName()`; the latter is "free" so users may expect the former to be "free" as well. The cost of calling `getAbsolutePathName()` is a function of the depth of the ownership tree (the Model's ownership tree should be flattened as much as possible to optimize performance).
2. `getAbsolutePathName()` creates a `std::vector<std::string>` and prepends parent names using `insert()` until the root node is reached, which is more work than might be expected. Prepending is particularly expensive (from http://www.cplusplus.com/reference/vector/vector/insert/: "Because vectors use an array as their underlying storage, inserting elements in positions other than the vector end causes the container to relocate all the elements that were after position to their new positions.").

In the past, names were (assumed to be) unique and were "free" to retrieve so were ideal for using as unique IDs (e.g., keys into maps). The unique ID is now the absolute path name but it isn't "free" to retrieve so using it as a key will be inefficient—particularly important for users doing optimization.

I suggest adding a note to the doxygen for `Component::getAbsolutePathName()` indicating that it shouldn't be used if efficiency is critical. In the future, the absolute path name could be cached.

chrisdembia:
I think it would be nice to actually just speed it up using a linked list or something instead of an `std::vector`.

tkuchida:
> I think it would be nice to actually just speed it up using a linked list or something instead of an `std::vector`.

That would be good, too, but if the absolute path name isn't being cached, I think the doxygen should warn users that it isn't free like `getName()`. I think it's natural to just replace `getName()` with `getAbsolutePathName()` since the former is no longer guaranteed to be unique, but these methods aren't really interchangeable if efficiency is important.

carmichaelong:
`std::deque` may also be an option here.

carmichaelong:
I did testing to benchmark performance. 

**Setup**
Using `testComponentInterface.cpp`, I created a `Component` with a depth of 5 as follows:
```
    TheWorld* A = new TheWorld();
[...]

PR #1900, opened 2017-08-28
getAbsolutePathName() improvement and add getAbsolutePath()

carmichaelong:
Related to #1883

### Brief summary of changes
- Speeds up `getAbsolutePathName()` by using `string`s only
- Adds new function `Component::getAbsolutePath()` and uses this to avoid going to and from `string`s unnecessarily
- Change `testComponentInterface` to pass
- Add a check for `Path` construction from `vector` for invalid characters (puts this more in line with the `string` constructor)

### Testing I've completed
- Tests pass locally

### CHANGELOG.md (choose one)
- no need to update because... `ComponentPath` is new to 4.0


OpenSim/Common/Component.h
```diff
@@ -627,6 +627,12 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
      * and, therefore, calling this method is not "free". */
     std::string getAbsolutePathName() const;
 
+    /** Return a ComponentPath of the absolute path of this Component. 
+     * Note that this has more overhead than calling `getName()` because 
+     * it traverses up the tree to generate the absolute pathname (and 
+     * is thus a function of depth). Consider other options if this is
```
chrisdembia: `and is thus` -> `and its computational cost is thus`

chrisdembia: `other options` -> `getAbsolutePathName()`.

OpenSim/Common/Component.h
```diff
@@ -2160,7 +2166,7 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
             throw Exception(msg);
         }
 
-        ComponentPath thisAbsPath(getAbsolutePathName());
+        ComponentPath thisAbsPath = getAbsolutePath();
```
chrisdembia: Woohoo!

OpenSim/Common/Component.h
```diff
@@ -1557,7 +1563,7 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
             const std::string thisClass = thisComp.getConcreteClassName();
             std::cout << std::string(maxlen-thisClass.length(), ' ') << "["
                       << thisClass << "]  ";
-            auto path = ComponentPath(thisComp.getAbsolutePathName());
+            auto path = thisComp.getAbsolutePath();
```
chrisdembia: Woohoo!

OpenSim/Common/Path.cpp
```diff
@@ -224,6 +227,14 @@ bool Path::isLegalPathElement(const std::string& pathElement) const
     return true;
 }
 
+bool Path::isLegalPathVec(const std::vector<std::string>& pathVec) const
+{
+    for (const std::string pathElement : pathVec) {
```
chrisdembia: Avoid the copy: `const std::string&`

OpenSim/Common/Path.cpp
```diff
@@ -78,6 +78,9 @@ Path::Path(const std::vector<std::string> pathVec,
 {
     if (_path.empty()) return;
     cleanPath();
+    if (!isLegalPathVec(_path)) {
+        OPENSIM_THROW(Exception, "Invalid character used in the path");
+    }
 }
```
chrisdembia: Nice catch.

OpenSim/Common/Test/testComponentInterface.cpp
```diff
@@ -897,52 +897,15 @@ void testListSockets() {
...
-
-    foo.setName("World3/bar2/foo1");
-    fooWrtBar = foo.getRelativePathName(bar);
-    ASSERT(fooWrtBar == "foo1");
+    otherTop.setName("OtherTop");
 
-    bar.setName("LegWithConstrainedFoot/footConstraint");
-    foo.setName("LegWithConstrainedFoot/foot");
-    barWrtFoo = bar.getRelativePathName(foo);
-    ASSERT(barWrtFoo == "../footConstraint");
```
chrisdembia: I didn't look too carefully, but have some test cases been removed? If so, could you explain why?

OpenSim/Common/Component.h
carmichaelong: I forgot that the merge didn't change the comment I had written. An alternative is to use the same comment from the function above (`The absolutePath is generated ... not "free."`).

OpenSim/Common/Test/testComponentInterface.cpp
carmichaelong: These tests looked like they were redundant with other tests, so I trimmed them down. From what I could tell, they may have added 1) testing with longer names and 2) testing after `setName()` was invoked on a `Component` that already existed. If that's important to go back in (or if I missed the reason for these tests), I'm happy to do re-create these tests so that they'll work with the interface changes.

OpenSim/Common/Component.h
carmichaelong: If we stick with this wording (rather than copying from function above), then this change may be misleading since both methods have to traverse the tree (even if `getAbsolutePathName()` is still faster).

OpenSim/Common/Test/testComponentInterface.cpp
chrisdembia: Okay sounds fine.

aseth1:
> What do others think about the naming of the methods? I was thinking getAbsolutePathNameString() and getAbsolutePathName() but that's fairly verbose. That, or getAbsolutePathString() and getAbsolutePath().

I am OK with the current use of `getAbsolutePath()` that returns a `ComponentPath` and `getAbsolutePathName()` that returns a `string`. Renaming the latter to `getAbsolutePathString()` would be OK too if that is preferred.

carmichaelong:
I still have to do some clean up that @chrisdembia mentioned above. I think I'm also leaning towards making the change: `getAbsolutePathName()` -> `getAbsolutePathString()`

chrisdembia:
I'm happy to leave the naming up to the two of you.

Issue #1889, opened 2017-08-21
marker.setParentFrame() and marker.setFrameName() behave inconsistently


aymanhab:
newMarker.setParentFrame(frameObject); works when using an actual Frame object  but newMarker.setFrameName("parentName") fails to connect in initSystem complaining that initSystem must be called first! We need to either make them consistent or document why they are different and improve the error message so it's not circular.

aseth1:
This is the code for `setFrameName()`
```cpp
void Marker::setFrameName(const string& name)
{
    const auto& refFrame = getModel().getComponent<PhysicalFrame>(name);
    setParentFrame(refFrame);
}
```

aseth1:
Calling `getModel().getComponent<PhysicalFrame>()` would be an issue with an unconnected marker where `getModel()` would throw (likely asking if you forgot to call `initSystem()` first) but would not throw in `initSystem()` AFAICT.

aseth1:
I also think the redundancy is not helpful, so I am happy to get rid of `setFrameName()` which was a carry over from `setBodyName()`.

aymanhab:
On one side I'd prefer to have it removed since its behavior is different (requires some connection which is not documented anywhere AFAIK) even though it looks very similar from an API user perspective to setFrame, which is unexpected. On the other hand it seems more natural to create an object using primitive data types (e.g. strings for names) rather than having to get a reference to a valid Frame to construct/add a Marker.

What would be your preference as an API user, @chrisdembia @tkuchida ?

aseth1:
We can update `Marker::setFrameName()` to only set the underlying Socket's connectee_name (like what happens by deserializing) and establish the connection at `finalizeConnections()`. This allows the model creator to defer the validation of connections, but it also means you don't have a "wired-up" connection. It seems the creator of Marker wanted the `setFrameName()` to result in a live connection and that's why it gets the `Frame` and calls `setParentFrame()`  with it.

While setting the Component directly does not allow you to defer creation of Components and their validation (you either have a valid Component or not) it is logically reasonable expect Components you want to connect to be available. That means potentially reordering the order of operations in user-code (although I cannot think of any of our examples that require the change) to create the dependencies before connecting them. If one insists on predefining the connections prior to the creation of the Components that satisfy them (the connectee) s/he can always use `updSocket<C>("socket_name").setConnecteeName("comp_name")`. 

[...]

PR #1901, opened 2017-08-28
Fix set<Conectee>Name() methods to only set the connectee_name property

aseth1:
Fixes issue #1889

### Brief summary of changes
1. Make `setFrameName()` consistent with other convenience methods for `Socket` `connectee_name` property renaming. 
2. Rename  `setFrameName()` to  `setParentFrameName()` consistent with the *parent_frame* `Socket`.
3. When a reference to a dependency (`Component`)  is supplied, use `connectSocket` to connect to it.

### Testing I've completed
All tests pass locally. Checked that there are no more `set<Connectee>Name()` convenience methods that attempt to form a direct connection.

### CHANGELOG.md (choose one)
- no need to update because `Socket`s and `Frame`s did not exist prior to 4.0

### DIFF
--- a/OpenSim/Simulation/Model/Marker.cpp
+++ b/OpenSim/Simulation/Model/Marker.cpp
@@ -73,17 +73,9 @@ void Marker::constructProperties()
     constructProperty_fixed(false);
 }
 
-//_____________________________________________________________________________
-/**
- * Set the 'frame name' field, which is used when the marker is added to
- * an existing model.
- *
- * @param  name of frame
- */
-void Marker::setFrameName(const string& name)
+void Marker::setParentFrameName(const string& name)
 {
-    const auto& refFrame = getModel().getComponent<PhysicalFrame>(name);
-    setParentFrame(refFrame);
+    updSocket<PhysicalFrame>("parent_frame").setConnecteeName(name);
 }
 
 //_____________________________________________________________________________
--- a/OpenSim/Simulation/Model/Marker.cpp
+++ b/OpenSim/Simulation/Model/Marker.cpp
@@ -90,48 +90,27 @@ const string& Marker::getParentFrameName() const
 }
 
 
-//_____________________________________________________________________________
-/**
- * Change the PhysicalFrame that this marker is attached to. It assumes that the frame is
- * already set, so that extendConnectToModel() needs to be called to update 
- * dependent information.
- *
- * @param aFrame Reference to the PhysicalFrame.
- */
-void Marker::changeFrame(const OpenSim::PhysicalFrame& aPhysicalFrame)
+void Marker::changeFrame(const PhysicalFrame& parentFrame)
 {
-
-    if (aPhysicalFrame == getParentFrame())
+    if (parentFrame == getParentFrame())
         return;
 
-    setParentFrameName(aPhysicalFrame.getName());
-
-    extendConnectToModel(updModel());
+    setParentFrame(parentFrame);
 }
 
-//_____________________________________________________________________________
-/**
- * Change the PhysicalFrame that this marker is attached to. It assumes that the body is
- * already set, so that extendConnectToModel() needs to be called to update 
- * dependent information.
- *
- * @param s State.
- * @param aBody Reference to the PhysicalFrame.
- */
-void Marker::changeFramePreserveLocation(const SimTK::State& s, OpenSim::PhysicalFrame& aPhysicalFrame)
+void Marker::changeFramePreserveLocation(const SimTK::State& s, 
+                                         const PhysicalFrame& parentFrame)
 {
 
-    if (aPhysicalFrame == getParentFrame())
+    if (parentFrame == getParentFrame())
         return;
 
     // Preserve location means to switch bodies without changing
     // the location of the marker in the inertial reference frame.
     Vec3 newLocation;
-    newLocation = findLocationInFrame(s, aPhysicalFrame);
+    newLocation = findLocationInFrame(s, parentFrame);
     set_location(newLocation);
-
-    setParentFrameName(aPhysicalFrame.getName());
-    extendConnectToModel(aPhysicalFrame.updModel());
+    setParentFrame(parentFrame);
 }
 
 //=============================================================================
--- a/OpenSim/Simulation/Model/Marker.cpp
+++ b/OpenSim/Simulation/Model/Marker.cpp
@@ -82,16 +82,13 @@ void Marker::setParentFrameName(const string& name)
 /**
  * Get the 'frame name' field, which is used when the marker is added to
  * an existing model.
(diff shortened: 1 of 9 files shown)


OpenSim/Simulation/Model/Marker.h
```diff
@@ -57,24 +57,30 @@ class Body;
...
 
-    const std::string& getFrameName() const;
+    /** Convenience method to set the 'parent_frame' Socket's connectee_name
+        and to establish the connection during finalizeConnections(). */
+    void setParentFrameName(const std::string& aName);
+    /** Change the parent PhysicalFrame that this marker is attached to. */
+    void changeFrame(const PhysicalFrame& parentFrame);
+    /**  Change the parent PhysicalFrame that this marker is attached to, and 
+         use the state to compute the location in the new parent frame and set
+         its location. */
```
tkuchida: I think it would be helpful to explicitly include some variation of the comment that appears in the code: "*Preserve location* means that the Marker is moved to the provided PhysicalFrame without changing the location of the Marker in the inertial reference frame." It is possible to misinterpret the meaning of "preserve location", particularly if the user is unfamiliar with whether the `location` property holds coordinates in the local or Ground frame.

OpenSim/Simulation/Model/Marker.cpp
```diff
@@ -73,76 +73,44 @@ void Marker::constructProperties()
...
- *
- * @param aFrame Reference to the PhysicalFrame.
- */
-void Marker::changeFrame(const OpenSim::PhysicalFrame& aPhysicalFrame)
-{
 
-    if (aPhysicalFrame == getParentFrame())
+void Marker::changeFrame(const PhysicalFrame& parentFrame)
+{
+    if (parentFrame == getParentFrame())
```
chrisdembia: I know this logic was not introduced in this PR, but this is checking the values, not the addresses, right? Don't you want to just check the pointers?

OpenSim/Simulation/Model/Marker.cpp
```diff
@@ -73,76 +73,44 @@ void Marker::constructProperties()
...
- * @param s State.
- * @param aBody Reference to the PhysicalFrame.
- */
-void Marker::changeFramePreserveLocation(const SimTK::State& s, OpenSim::PhysicalFrame& aPhysicalFrame)
+void Marker::changeFramePreserveLocation(const SimTK::State& s, 
+                                         const PhysicalFrame& parentFrame)
 {
 
-    if (aPhysicalFrame == getParentFrame())
+    if (parentFrame == getParentFrame())
```
chrisdembia: See above.

OpenSim/Simulation/Model/Marker.cpp
```diff
@@ -73,76 +73,44 @@ void Marker::constructProperties()
...
 
     // Preserve location means to switch bodies without changing
     // the location of the marker in the inertial reference frame.
     Vec3 newLocation;
-    newLocation = findLocationInFrame(s, aPhysicalFrame);
+    newLocation = findLocationInFrame(s, parentFrame);
     set_location(newLocation);
-
-    setFrameName(aPhysicalFrame.getName());
-    extendConnectToModel(aPhysicalFrame.updModel());
```
chrisdembia: Was @aymanhab relying on this for the GUI somehow?

OpenSim/Simulation/Model/Marker.cpp
aseth1: You are right! My personal preference is to remove this method altogether. There isn't a good reason to not just call `setParentFrame()` AFAIK.

OpenSim/Simulation/Model/Marker.cpp
aseth1: This was not doing anything - `extendConnectToModel` is not overridden between ModelComponent and Marker. I believe this is just being inefficient since it delays the connection based on the name which it just gets from the reference in the first place.

OpenSim/Simulation/Model/Marker.cpp
```diff
@@ -73,76 +73,44 @@ void Marker::constructProperties()
...
-const string& Marker::getFrameName() const
-{
-    //if (_bodyNameProp.getValueIsDefault())
-    //  return NULL;
 
-    return getParentFrame().getName();
+const string& Marker::getParentFrameName() const
+{
+    return getSocket<PhysicalFrame>("parent_frame").getConnecteeName();
 }
```
aymanhab: Is there a chance that getSocket will fail or return null (since we're calling a method on it later on this line)?

OpenSim/Simulation/Model/Marker.cpp
aymanhab: As long as the call to getParentFrameName() still works after we should be fine. Will need to change the testCase that started this  in order to exercise.

OpenSim/Simulation/Model/Marker.h
```diff
@@ -57,24 +57,31 @@ class Body;
...
 public:
     Marker();
     virtual ~Marker();
+    /** Convenience method to get the 'parent_frame' Socket's connectee_name */
+    const std::string& getParentFrameName() const;
 
-    const std::string& getFrameName() const;
+    /** Convenience method to set the 'parent_frame' Socket's connectee_name
+        and to establish the connection during finalizeConnections(). */
+    void setParentFrameName(const std::string& aName);
```
aymanhab: The wording can be made clearer by saying explicitly that no connection is established by this call. You need to call finalizeConnections to establish connection or something along these lines.

aseth1: Will do. Thanks.

OpenSim/Simulation/Model/Marker.cpp
aseth1: Not really. The Socket is baked into the Component and would throw if "parent_frame" was not a valid Socket name (but it is).

OpenSim/Simulation/Model/Marker.cpp
aseth1: Which test case should I update?

aseth1:
@aymanhab looking into testScale failure.

OpenSim/Simulation/Model/Marker.cpp
aymanhab: The testcase in PR #1857 may have used some of the renamed methods or can be improved, I'll revisit once your change is merged. Thanks @aseth1

aseth1:
@aymanhab PR is ready for a quick review and all tests are passing.

aymanhab:
LGTM. Thanks @aseth1

aseth1:
Thanks @aymanhab, @chrisdembia and @tkuchida. Merging ...

Issue #1762, opened 2017-06-12
Bouncing block model does not load.


chrisdembia:
The reason seems to be that it uses a PointToPointSpring, which does not have updateFromXMLNode() code for sockets.

Related #206, https://github.com/opensim-org/opensim-gui/issues/68

tkuchida:
The model can be loaded but `finalizeConnections()` fails:
<pre>
Failed to connect Socket 'body1' of type PhysicalFrame (details: Model'toy_with_forces'::findComponent() cannot find a nameless subcomponent.).
        In object 'spring' of type PointToPointSpring.
        In file C:\Users\tkuchida\opensim-core\OpenSim\Common\Component.cpp:236
        In function 'finalizeConnections'
</pre>

tkuchida:
Fixed by #1902.

PR #1902, opened 2017-08-30
Add update from xml to point to point spring

aseth1:
Fixes issue #1762 

### Brief summary of changes
Adds `PointToPointSpring::updateFromXMLNode()` so that spring from previous versions are correctly updated upon deserialization. This a spot fix of the issue but does not address the bigger problem of finding/addressing all potential missing updates (#206) systematically.

Also modified *bouncing_block.osim* so it does not rely on binary .vtp mesh files and uses available *box* and *sphere* meshes which are not problematic in 4.0.

### Testing I've completed
Added to `testForces` cases to also deserialize and evaluate bouncing_block_30000.osim model.

### CHANGELOG.md (choose one)

### DIFF
--- a/OpenSim/Simulation/Test/testForces.cpp
+++ b/OpenSim/Simulation/Test/testForces.cpp
@@ -516,8 +516,6 @@ void testSpringMass()
 
     osimModel.addForce(&spring);
 
-    //osimModel.print("SpringMassModel.osim");
-
     // Create the force reporter
     ForceReporter* reporter = new ForceReporter(&osimModel);
     osimModel.addAnalysis(reporter);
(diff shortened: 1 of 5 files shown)


chrisdembia:
> Also modified bouncing_block.osim so it does not rely on binary .vtp mesh files and uses available box and sphere meshes which are not problematic in 4.0.

It would be great for these changes to make their way into the `opensim-models` repo.

aseth1:
> It would be great for these changes to make their way into the opensim-models repo.

Done. See [model PR 26](https://github.com/opensim-org/opensim-models/pull/26)

OpenSim/Simulation/Test/testForces.cpp
```diff
@@ -563,6 +561,14 @@ void testSpringMass()
...
 
     ASSERT(*copyOfSpring == spring);
+
+    // Verify that the PointToPointSpring is correctly deserialized from
+    // previous major version of OpenSim.
+    Model bouncer("bouncing_block_30000.osim");
+    SimTK::State& s = bouncer.initSystem();
+    bouncer.realizeAcceleration(s);
+
+    Vec3 comA = bouncer.calcMassCenterAcceleration(s);
```
tkuchida: I think this line will generate a warning because `comA` is not used. In this situation, I have seen
```cpp
/*Vec3 comA =*/ bouncer.calcMassCenterAcceleration(s);
```
elsewhere in the repo.

Issue #1780, opened 2017-06-20
Are both Model::updateMarkerSet() and replaceMarkerSet() necessary?


aseth1:
`Model::updateMarkerSet(newMarkerSet)` takes the union of the existing marker set, which may not be what users intend/expect. During scaling, I was expecting that the provided *<marker_set_file>* replaces that of the model. The comment is ambiguous.

aseth1:
At a minimum `Model::updateMarkerSet(newMarkerSet)` could be renamed to `appendMarkerSet()` with a comment clarifying in the case of collisions (marker with same name present in the Model) that the new definition replaces the existing.

aseth1:
Actions from today's meeting: 
1. eliminate `Model::replaceMarkerSet(const SimTK::State& s, const MarkerSet& aMarkerSet)` because editing the Model should not require a `State` and have a sideeffect of creation/initialization of the `System`. Also, the `ScaleTool` in the API uses `updateMarkerSet()` which causes an inconsistency with the GUI if it uses `replaceMarkerSet()`.

2. rename `Model::updateMarkerSet(MarkerSet& newMarkerSet)` to `Model::mergeMarkerSet(MarkerSet& markerSet)` and update the comment on the `ScaleTool` that this takes the union of the  Model's and provided marker sets, with the provided set taking precedence in the case of name collisions. This method is easily confused with `updMarkerSet()` , which returns a writable reference to the Model's MarkerSet. 

I wanted to raise the issue that this treatment of the `MarkerSet` is inconsistent with the treatment of a `ForceSet` by the `AbstractTool`, which uses a `_replaceForceSet` flag/property. If *true* the model's force set is replaced with the provided `ForceSet` files. When *false* the supplied ForceSets are appended to the model's.  Here, collisions are poorly handled. Duplicate names are *disallowed* by default but when a duplicate name is found in the provided set, the append just skips that force with no error/exception or warning. 

From our discussion for `MarkerSet`s, it seemed users may want to override an existing marker so it would be better for the provided marker set to take precedence over the Model's set. Then perhaps a distinct `mergeMarkerSet()` is better than trying to be consistent with the confusing `ForceSet::append(ForceSet &aForceSet, bool aAllowDuplicateNames))` behavior.

aseth1:
We should at least do 1. It looks like it is not being used by the GUI see #1903

PR #1903, opened 2017-08-31
Add test case to exercise ScaleTool as in GUI operation.

aymanhab:
Fixes issue #<issue_number>

### Brief summary of changes
Exercise ScaleTool in workflow similar to GUI. This should be beefed up to check the values of scale factors and also to exercise the workflow where ScaleTool is constructed live from scratch rather than from XML file.

### Testing I've completed
Ran test case and loaded models before and after scaling in GUI

### CHANGELOG.md (choose one)

- no need to update because this is a test case

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/1903)


aymanhab:
Ready for review/feedback

Bindings/Java/tests/CMakeLists.txt
```diff
@@ -86,6 +86,14 @@ macro(OpenSimAddJavaTest TESTNAME)
                  DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/TestVisualization")
     file(COPY "${OPENSIM_SHARED_TEST_FILES_DIR}/gait10dof18musc_subject01.osim"
                  DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/TestEditProperties")
+    file(COPY "${OPENSIM_SHARED_TEST_FILES_DIR}/gait2354_simbody.osim"
```
tkuchida: Is it possible to use model/setup files that are already in the repo (e.g., the files used by testScale.cpp), particularly since this isn't a correctness/regression test? There are already 4 files named "gait2354_simbody.osim" in the repo.

Bindings/Java/tests/TestModelScaling.java
```diff
@@ -0,0 +1,46 @@
...
+        MarkerData measurementTrial = new MarkerData("subject01_static.trc");
+        System.out.println("MarkerData for static trial loaded!");
+        double scaleFactor = context.computeMeasurementScaleFactor(scaleTool.getModelScaler(), unscaledModel, 
+                measurementTrial, measurementSet.get(0));
+        System.out.println("Measurement computed !");
+        // When we run the tool we create a copy of the unscaled model and scale in place using the
+        // calls below which exercise the "run" button
+        Model processedModel = new Model(unscaledModel);
+        processedModel.setName(scaleTool.getName());
+        OpenSimContext peocessedContext = new OpenSimContext(processedModel.initSystem(), processedModel);
```
tkuchida: I know it doesn't matter to the compiler, but there's a typo in the variable name ("peocessedContext").

Bindings/Java/tests/CMakeLists.txt
aymanhab: Thanks for the note @tkuchida I'm actually using the recommended method of having these files added to the shared testing folder, we should make sure other tests are using it from there. This may or maynot be part of this PR but good point.

chrisdembia: Yes, this is the recommended method, but ideally, one of the pre-existing gait2354 models would have been moved to the shared folder.

tkuchida: Added a comment to this effect in Issue #206.

aymanhab: Thanks @tkuchida 👏

aseth1: While we want to move to shared resources to avoid duplication, I think copying essentially the same files in testScale violates that spirit. I would prefer we use the testScale files that already exist instead of making yet another copy.

Bindings/Java/tests/TestModelScaling.java
```diff
@@ -0,0 +1,46 @@
...
+        // Live current model in GUI would be used for unscaledModel
+        Model unscaledModel = new Model("gait2354_simbody.osim");
+        // Exercise "loadsettings" from file"
+        try {
+            scaleTool = new ScaleTool("subject01_Setup_Scale.xml");
+        } catch (IOException ex) {
+            System.exit(-1);
+        }
+        System.out.println("ScaleTool created!");
+        // Populate the tabs of Scale factors and measurements
```
aseth1: This comment does not describe what is happening below.

Bindings/Java/tests/TestModelScaling.java
```diff
@@ -0,0 +1,46 @@
...
+                processedModel, "", scaleTool.getSubjectMass());
+        processedContext.processModelMarkerPlacer(scaleTool.getMarkerPlacer(), processedModel, "");
+        System.out.println("Test finished!");
+    }
+    catch (IOException ex){
+        System.exit(-1);
+    }
+    System.gc();
+  }
+}
```
aseth1: Can we remove Model::replaceMarkerSet() (see #1780) and deleteUnusedMarkers() since they appear to be unused by the GUI?

aymanhab: MarkerPlacer::processModel uses deleteUnusedMarkers(), it's not used by the GUI directly. replaceMarkerSet is not called by the GUI anymore and can be removed.

nickbianco:
Closing: stale PR.

Issue #1904, opened 2017-08-31
Confusing message on component renaming


aymanhab:
Trying to run testContext test case I got the message below in the console. Is this expected?

GeometryPath 'geometrypath' has subcomponents with duplicate name 'pathwrap_0'. The duplicate was renamed to 'pathwrap_0'.
GeometryPath 'geometrypath' has subcomponents with duplicate name 'pathwrap_1'. The duplicate was renamed to 'pathwrap_1'.

aseth1:
Closed by #1906

PR #1906, opened 2017-08-31
Fix 1904 confusing renaming message

aseth1:
Fixes issue #1904

### Brief summary of changes
In #1887 switched local `name` variable to be a const reference instead of copy, When printing the message the *name* value is already updated by the subcomponent rename. The fix was to reorder and print the message prior to rename. 

### Testing I've completed
Verified that warning message is correct.

none

### CHANGELOG.md (choose one)
- no need to update because this is internal bug fix.


aseth1:
Needs #1904 to be merged first. Sorry!

aseth1:
Thank you @chrisdembia and @aymanhab

Issue #1662, opened 2017-04-16
Activating muscles in every step of the simulation


kidzik:
Hi, I am trying to activate muscles one by one every 0.01 second in simulation. It used to work fine, when I was compiling a version from some old commit b3163b22be7bd4cccd8c073e02dd3abb52867031 . In the beta release 399c8d57a779dd5dde2916192f8b92bfc959e269 it seems to only activate muscles in the first step of the simulation and they are deactivated in all the other steps. 
If I use `setExcitation` instead, it seems that nothing happens and muscles are deactivated throughout the entire simulation.
```python
import opensim

model_path = "arm26.osim"
model = opensim.Model(model_path)
model.setUseVisualizer(True)
state = model.initSystem()
manager = opensim.Manager(model)
muscleSet = model.getMuscles()
stepsize = 0.01

for i in range(500):
    for j in range(6):
        muscleSet.get(j).setActivation(state, 1.0)
    manager.integrate(state, stepsize * (i + 1))
```
Is there anything I am missing?
I checked `arm26` and `gait10dof18musc` models.
Thanks!

chrisdembia:
The interface for `Manager::integrate()` changed. That could be it, but it looks like you're using the new interface correctly.

junja94:
I'm having the same problem with similar code.

chrisdembia:
@kidzik What do you mean in `the first step`? You mean for `i = 0`, or for the first time step the integrator takes within each call to `manager.integrate()`?

kidzik:
@chrisdembia I meant `i=0` - it basically blinks the muscles to red then and they are turned off throughout the rest of the simulation. I will check it more precisely on Monday. Thanks!

chrisdembia:
Ah okay. My guess is that the `state` is not getting updated. That is `state.getTime() == 0` even for `i > 0`.

carmichaelong:
@chrisdembia I think at least part of the `state` is being updated. I wrote a simple case directly in C++. The initialization is about the same, and my loop looks like this:
```
    for (int i = 0; i < 10; ++i)
    {
        for (int j = 0; j < muscleSet.getSize(); ++j)
        {
            muscleSet.get(j).setActivation(state, 1.0);
        }
        cout << state.getTime() << " ";
        manager.integrate(state, stepsize*(i + 1));
        arm.realizeDynamics(state);
        cout << muscleSet.get(0).getActivation(state) << endl;
    }
```
The output is:
```
0 0.652489
[...]

PR #1909, opened 2017-09-04
Change Manager interface to prevent "reinitializing"

carmichaelong:
Fixes issue #1662 

### Brief summary of changes
The focus of this PR is to prevent a user from using the same Manager for integrating a state that has been modified. To do this, I've added `Manager::initialize(state)` which mimics the initialization method of a `SimTK::TimeStepper`.

### Testing I've completed
Updated tests for change in `Manager` interface, checked that tests pass locally.

### CHANGELOG.md (choose one)
- updated to describe new interface

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,15 +57,18 @@ Converting from v3.x to v4.0
   compatibility, a joint's parent and child PhysicalFrames are swapped when
   opening a Model if the `reverse` element is set to `true`.
 - The `Manager::integrate(SimTK::State&)` call is deprecated and replaced by
-  `Manager::integrate(SimTK::State&, double)`. Here is a before-after example
-  (see the documentation in the `Manager` class for more details):
+  `Manager::integrate(double)`. You must also now call 
+  `Manager::initialize(SimTK::State&)` before integrating. Here is a 
+   before-after example (see the documentation in the `Manager` class 
+   for more details):
   - Before:
-	- manager.setInitialTime(0.0);
-	- manager.setFinalTime(1.0);
-	- manager.integrate(state);
+    - manager.setInitialTime(0.0);
+    - manager.setFinalTime(1.0);
+    - manager.integrate(state);
   - After:
     - state.setTime(0.0);
-	- manager.integrate(state, 1.0);
+    - manager.initialize(state);
+    - manager.integrate(1.0);
 
 - `Muscle::equilibrate(SimTK::State&)` has been removed from the Muscle interface in order to reduce the number and variety of muscle equilibrium methods. `Actuator::computeEquilibrium(SimTK::State&)` is overridden by Muscle and invokes pure virtual `Muscle::computeInitialFiberEquilibrium(SimTK::State&)`.
 - `Millard2012EquilibriumMuscle::computeFiberEquilibriumAtZeroVelocity(SimTK::State&)` and `computeInitialFiberEquilibrium(SimTK::State&)` were combined into a single method:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,18 +58,24 @@ Converting from v3.x to v4.0
   opening a Model if the `reverse` element is set to `true`.
 - The `Manager::integrate(SimTK::State&)` call is deprecated and replaced by
   `Manager::integrate(double)`. You must also now call 
-  `Manager::initialize(SimTK::State&)` before integrating. Here is a 
+  `Manager::initialize(SimTK::State&)` before integrating or pass the
+  initialization state into a convenience constructor. Here is a 
    before-after example (see the documentation in the `Manager` class 
    for more details):
   - Before:
+    - Manager manager(model);
     - manager.setInitialTime(0.0);
     - manager.setFinalTime(1.0);
     - manager.integrate(state);
   - After:
+    - Manager manager(model);
     - state.setTime(0.0);
     - manager.initialize(state);
     - manager.integrate(1.0);
-
+  - After (using a convenience constructor):
+    - state.setTime(0.0);
+    - Manager manager(model, state);
+    - manager.integrate(1.0);
 - `Muscle::equilibrate(SimTK::State&)` has been removed from the Muscle interface in order to reduce the number and variety of muscle equilibrium methods. `Actuator::computeEquilibrium(SimTK::State&)` is overridden by Muscle and invokes pure virtual `Muscle::computeInitialFiberEquilibrium(SimTK::State&)`.
 - `Millard2012EquilibriumMuscle::computeFiberEquilibriumAtZeroVelocity(SimTK::State&)` and `computeInitialFiberEquilibrium(SimTK::State&)` were combined into a single method:
 `Millard2012EquilibriumMuscle::computeFiberEquilibrium(SimTK::State&, bool useZeroVelocity)`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,7 @@ Bug Fixes
 - Fixed a bug where MuscleAnalysis was producing empty moment arm files. We now avoid creating empty Moment and MomentArm storage files when `_computeMoments` is False. (PR #324)
 - Fixed bug causing the muscle equilibrium solve routine in both Thelen2003Muscle and Millard2012EquilibriumMuscle to fail to converge and erroneously return the minimum fiber length. The fix added a proper reduction in step-size when errors increase and limiting the fiber-length to its minimum. (PR #1728)
 - Fixed a bug where Models with Bodies and Joints (and other component types) with the same name were loaded without error. Duplicately named Bodies were simply being ignored and only the first Body of that name in the BodySet was being used, for example, to connect a Body to its parent via its Joint, or to affix path points to its respective Body. Now, duplicate names are flagged and renamed so they are uniquely identified. (PR #1887)
+- Fixed bug and speed issue with `model.setStateVariableValues()` caused by enforcing constraints after each coordinate value was being set (PR #1911). Removing the automatic enforcement of constraints makes setting all state values much faster, but also requires calling `model.assemble()` afterwards. Enforcing constraints after setting each coordinate value individually was also incorrect, since it neglected the effect of other coordinate changes have on the current coordinate. All coordinate values must be set before enforcing constraints.
 
 
 New Classes
--- a/OpenSim/Simulation/Manager/Manager.cpp
+++ b/OpenSim/Simulation/Manager/Manager.cpp
@@ -51,7 +51,7 @@ std::string Manager::_displayName = "Simulator";
 // CONSTRUCTOR(S)
 //=============================================================================
 Manager::Manager(Model& model) 
-        : Manager(model,  true)
+        : Manager(model, true)
 {
     _defaultInteg.reset(
             new SimTK::RungeKuttaMersonIntegrator(_model->getMultibodySystem()));
@@ -59,8 +59,22 @@ Manager::Manager(Model& model)
 }
 
 Manager::Manager(Model& model, SimTK::Integrator& integ)
-        : Manager(model, true) {
+        : Manager(model, true) 
+{
     setIntegrator(integ);
+}
+
+Manager::Manager(Model& model, const SimTK::State& state)
+        : Manager(model) 
+{
+    initialize(state);
+}
+
+Manager::Manager(Model& model, const SimTK::State& state, 
+    SimTK::Integrator& integ)
+    : Manager(model, integ) 
+{
+    initialize(state);
 }
 
 // Private constructor to handle common tasks of the two constructors above.
@@ -95,8 +109,6 @@ void Manager::
 setNull()
 {
     _sessionName = "";
-    _ti = 0.0;
-    _tf = 1.0;
     _halt = false;
     _specifiedDT = false;
     _constantDT = false;
@@ -479,7 +491,7 @@ resetTimeAndDTArrays(double aTime)
  * Sets the model  and initializes other entities that depend on it
  */
 void Manager::
-setModel(Model& aModel)
+setModel(Model& model)
 {
     if(_model != nullptr){
         // May need to issue a warning here that model was already set to avoid a leak.
@@ -491,7 +503,7 @@ setModel(Model& aModel)
         OPENSIM_THROW(Exception, msg);
     }
 
-    _model = &aModel;
+    _model = &model;
 
     // STORAGE
     constructStorage();
@@ -528,54 +540,6 @@ setIntegrator(SimTK::Integrator& integrator)
     _integ = &integrator;
     // If we had been using the _defaultInteg, we no longer need it.
     _defaultInteg.reset();
-}
-
-//-----------------------------------------------------------------------------
-// INITIAL AND FINAL TIME
-//-----------------------------------------------------------------------------
-//_____________________________________________________________________________
-/**
- * Set the initial time of the simulation.
- *
- * @param aTI Initial time.
- */
-void Manager::
-setInitialTime(double aTI)
-{
-    _ti = aTI;
-}
-//_____________________________________________________________________________
-/**
- * Get the initial time of the simulation.
- *
- * @return Initial time.
- */
-double Manager::
-getInitialTime() const
-{
-    return(_ti);
-}
-//_____________________________________________________________________________
-/**
- * Set the final time of the simulation.
- *
- * @param aTF Final time.
- */
-void Manager::
-setFinalTime(double aTF)
-{
-    _tf = aTF;
-}
-//_____________________________________________________________________________
-/**
- * Get the final time of the simulation.
- *
- * @return Final time.
- */
-double Manager::
-getFinalTime() const
-{
-    return(_tf);
 }
 
 
--- a/OpenSim/Simulation/Manager/Manager.cpp
+++ b/OpenSim/Simulation/Manager/Manager.cpp
@@ -625,8 +625,7 @@ hasStateStorage() const
 // INTEGRATION
 //-----------------------------------------------------------------------------
 
-SimTK::State Manager::
-integrate(double finalTime)
+const SimTK::State& Manager::integrate(double finalTime)
 {
     int step = 0; // for AnalysisSet::step()
 
@@ -640,7 +639,7 @@ integrate(double finalTime)
     clearHalt();
 
     // CHECK SPECIFIED DT STEPPING
-    SimTK::State s = _integ->getState();
+    const SimTK::State& s = _integ->getState();
     double initialTime = s.getTime();
     if (_specifiedDT) {
         if (_tArray.getSize() <= 0) {
@@ -737,12 +736,11 @@ integrate(double finalTime)
     clearHalt();
 
     return getState();
-
 }
 
-SimTK::State Manager::getState()
+const SimTK::State& Manager::getState() const
 {
-    return _integ->getState();
+    return _timeStepper->getState();
 }
 
 //_____________________________________________________________________________
@@ -768,7 +766,7 @@ double Manager::getFixedStepSize(int tArrayStep) const {
  * 
  * @param s system state before integration
  */
-void Manager::initializeStorageAndAnalyses(SimTK::State& s)
+void Manager::initializeStorageAndAnalyses(const SimTK::State& s)
 {
     if( _writeToStorage && _performAnalyses ) { 
 
@@ -800,7 +798,7 @@ void Manager::initializeStorageAndAnalyses(SimTK::State& s)
 /**
 * set and initialize a SimTK::TimeStepper
 */
-void Manager::initialize(SimTK::State& s)
+void Manager::initialize(const SimTK::State& s)
 {
     if (!_integ) {
         throw Exception("Manager::initialize(): "
--- a/OpenSim/Simulation/Manager/Manager.cpp
+++ b/OpenSim/Simulation/Manager/Manager.cpp
@@ -594,7 +594,7 @@ const SimTK::State& Manager::integrate(double finalTime)
     int step = 0; // for AnalysisSet::step()
 
     if (_timeStepper == nullptr) {
-        throw Exception("Manager::integrate(): Manager has not be "
+        throw Exception("Manager::integrate(): Manager has not been "
             "initialized. Call Manager::initialize() first.");
     }
 
--- a/OpenSim/Simulation/Manager/Manager.cpp
+++ b/OpenSim/Simulation/Manager/Manager.cpp
@@ -50,15 +50,16 @@ std::string Manager::_displayName = "Simulator";
 //=============================================================================
 // CONSTRUCTOR(S)
 //=============================================================================
-Manager::Manager(Model& model) : Manager(model, true)
+Manager::Manager(Model& model) 
+        : Manager(model,  true)
 {
     _defaultInteg.reset(
             new SimTK::RungeKuttaMersonIntegrator(_model->getMultibodySystem()));
     _integ = *_defaultInteg;
 }
 
-Manager::Manager(Model& aModel, SimTK::Integrator& integ)
-        : Manager(aModel, true) {
+Manager::Manager(Model& model, SimTK::Integrator& integ)
+        : Manager(model, true) {
     setIntegrator(integ);
 }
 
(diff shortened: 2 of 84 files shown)


CHANGELOG.md
```diff
@@ -57,15 +57,18 @@ Converting from v3.x to v4.0
...
+   for more details):
   - Before:
 	- manager.setInitialTime(0.0);
 	- manager.setFinalTime(1.0);
 	- manager.integrate(state);
   - After:
     - state.setTime(0.0);
-	- manager.integrate(state, 1.0);
+	- manager.initialize(state);
+	- manager.integrate(1.0);
```
aseth1: It is not clear now that the *state* passed into `Manager::initialize(state)` is operated upon by `Manager::integrate(finalT)`.

Also, is one now able to update the *state* following integration and prior to subsequent integration (as in #1662) by inserting `Manager::initialize(state)` prior to `integrate`? Might indicate that issue and your fix here or in bug fixes below.

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -76,7 +76,9 @@ class OSIMSIMULATION_API Manager
     /** Simulation session name. */
     std::string _sessionName;
     /** Model for which the simulation is performed. */
-    Model *_model;
+    Model* _model;
```
aseth1: Might as well use `ReferencePtr` here.

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -76,7 +76,9 @@ class OSIMSIMULATION_API Manager
...
+    /** Keep a pointer to the State passed in so we can update it. */
+    SimTK::State* _state;
```
aseth1: `ReferencePtr` here too.

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -221,17 +223,26 @@ class OSIMSIMULATION_API Manager
     // EXECUTION
     //--------------------------------------------------------------------------
     /**
+    * Initializes the Manager by creating and initializing the underlying 
+    * SimTK::TimeStepper. This must be called before calling 
+    * Manager::integrate() Subsequent changes to the State object passed in 
+    * here will not affect the simulation. Calling this function multiple 
+    * times with the same Manager will trigger an Exception.
```
aseth1: Since you have a separate `initialize(state)` why not create a new `TimeStepper` internally instead of forcing to recreate the `Manager`? Then we could use `initialize(state)`to update the state between integrations, which would be nice (e.g. #1662). We could make a note that this can be slow and it is more efficient to use an event handler to update the state or controls periodically.

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -240,13 +251,15 @@ class OSIMSIMULATION_API Manager
...
+    * manager.initialize(state);
     * for (int i = 1; i <= n; ++i) {
-    *     manager.integrate(state, i*dTime);
+    *     manager.integrate(i*dTime);
     * }
     * @endcode
     *
     */
-    bool integrate(SimTK::State& s, double finalTime);
+    bool integrate(double finalTime);
```
aseth1: Since we hold a reference (pointer) to the *state* passed in during `initialize`, should we also have an accessor to return the updated *state*? I'm OK with not, since the user passes the state in, but we have to make it more clear that the state is being updated by `integrate` (even though it is not an input to this method).

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -56,15 +56,15 @@ class ControllerSet;
...
  * In order to prevent an inconsistency between the Integrator and TimeStepper,
- * we only create a TimeStepper once, specifically at the first call to
- * integrate(SimTK::State&, double). To ensure this, the Manager will throw 
- * an exception if setModel() or setIntegrator() is called after 
- * integrate(SimTK::State&, double) has been called at least once.
+ * we only create a TimeStepper once, specifically when we call
+ * initialize(SimTK::State&). To ensure this, the Manager will throw
+ * an exception if initialize(SimTK::State&) is called more than once. Note 
+ * that any subsequent changes to the SimTK::State passed into 
+ * initialize(SimTK::State&) will not affect the simulation.
```
aseth1: This might be too strict now that you introduced a user-required `initilalize` step and your are not creating a `TimeStepper` in `integrate` anymore. See comments below.

CHANGELOG.md
chrisdembia: Could you edit the whitespace to consistently use tabs or spaces here? For me, the list looks like the indentation levels are different.

OpenSim/Simulation/Manager/Manager.cpp
```diff
@@ -50,15 +50,16 @@ std::string Manager::_displayName = "Simulator";
...
 {
     _defaultInteg.reset(
             new SimTK::RungeKuttaMersonIntegrator(_model->getMultibodySystem()));
     _integ = *_defaultInteg;
 }
 
-Manager::Manager(Model& aModel, SimTK::Integrator& integ)
-        : Manager(aModel, true) {
+Manager::Manager(Model& model, SimTK::Integrator& integ)
+        : Manager(model, true) {
```
chrisdembia: +1

OpenSim/Simulation/Manager/Manager.cpp
```diff
@@ -625,21 +626,21 @@ hasStateStorage() const
...
-integrate(SimTK::State& s, double finalTime)
+integrate(double finalTime)
 {
     int step = 0; // for AnalysisSet::step()
 
-    if (!_integ) {
-        throw Exception("Manager::integrate(): "
-            "Integrator has not been set. Construct the Manager "
-            "with an integrator, or call Manager::setIntegrator().");
+    if (_timeStepper == NULL) {
```
chrisdembia: Nitpicky: it would be good to transition to using `nullptr` instead of `NULL`.

OpenSim/Simulation/Manager/Manager.cpp
```diff
@@ -752,7 +751,7 @@ bool Manager::
 integrate(SimTK::State& s)
```
chrisdembia: Does this deprecated function works as intended anymore? Perhaps if it's called only once? I wonder if we should remove this? Not sure.

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -240,13 +251,15 @@ class OSIMSIMULATION_API Manager
...
     * }
     * @endcode
     *
     */
-    bool integrate(SimTK::State& s, double finalTime);
+    bool integrate(double finalTime);
+    
     /** <b>(Deprecated)</b> Integrate to a specified finalTime using
         Manager::integrate(SimTK::State&, double). */
     DEPRECATED_14("Integrate to a specified finalTime using Manager::integrate(SimTK::State&, double).")
```
chrisdembia: This deprecation message should be updated.

OpenSim/Simulation/Manager/Manager.cpp
chrisdembia: Nitpicky again: I just noticed this is a a smart pointer; I think you could just do `if (_timeStepper)`. Feel free to ignore these comments though.

OpenSim/Simulation/Manager/Manager.h
chrisdembia: Suggested change in wording: "Note that any subsequent changes to the SimTK::State passed into  initialize(SimTK::State&) will not affect the simulation." -> "Note that editing the `SimTK::State` after calling `initialize(SimTK::State&)` will not affect the simulation." (I was initially confused by "subsequent").

OpenSim/Simulation/Manager/Manager.h
chrisdembia: I talked with @aseth1 about this. We thought it could be nice to do the following:

1. Make a copy of the state passed into `initialize()`
2. Change the signature of `integrate()` to return the final state. That is, user code would look like `auto finalState = manager.integrate(tf)`.

This way, it's clear where to get the final state values from. Also, I think users might find it confusing that edits to the state passed into `initialize()` have no effect after calling `initalize()`, but that this same state object gets magically updated after integrating.

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -240,13 +251,15 @@ class OSIMSIMULATION_API Manager
     * finalTime = 10.0;
     * int n = int(round(finalTime/dTime));
     * state.setTime(0.0);
+    * manager.initialize(state);
     * for (int i = 1; i <= n; ++i) {
-    *     manager.integrate(state, i*dTime);
+    *     manager.integrate(i*dTime);
     * }
```
chrisdembia: Consider adding an example here related to something @aseth1 said elsewhere in this review: "If you must update states or controls in a loop, you must recreate the manager within the loop (such discontinuous changes are considered "events" and cannot be handled during integration of the otherwise continuous system): <example> ...The proper way to handle this situation is to create a SimTK::EventHandler."

If you adopt @aseth1's suggestion of allowing repeated calls to `initialize()`, then you could change "you must recreate the manager within the loop" to "you must call `initialize()` within the loop".

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -56,15 +56,15 @@ class ControllerSet;
...
+ * that any subsequent changes to the SimTK::State passed into 
+ * initialize(SimTK::State&) will not affect the simulation.
  *
- * Since the call to integrate(SimTK::State&, double) takes the state as an 
- * argument, it is up to the caller to ensure that the state is a legal state 
- * if the same Manager is used to integrate again. Integrating a different 
- * state for some new arbitrary system has undefined behavior.
+ * Note that this interface means that you cannot "reinitialize" a Manager.
+ * If you make changes to the SimTK::State, a new Manager must be created
+ * before integrating again.
```
chrisdembia: These comments are really helpful. I think we should mention the specific situation that Lukasz ran into; see my comments on `Manager::integrate()`.

OpenSim/Simulation/Manager/Manager.h
chrisdembia: +1, but it seems like storing this will no longer be necessary (see other comments).

carmichaelong:
> 1. Make a copy of the state passed into initialize()
> 2. Change the signature of integrate() to return the final state. That is, user code would look like auto finalState = manager.integrate(tf).

I did try out something like this somewhere in the commits, and instead went with a "minimal" change interface. The fact that you guys also thought of this means that it's worth the bigger changes in interface, so I'll definitely work on this.

OpenSim/Simulation/Manager/Manager.h
carmichaelong: From testing before, this seemed to cause a problem. @chrisdembia mentioned it that it's "reinitializing" the Integrator that seemed to be a problem for some reason that we couldn't track down.

OpenSim/Simulation/Manager/Manager.h
carmichaelong: At one point when running tests, I accidentally "reinitialized" a Manager, and this led to a crash, so I added this throw. I think also by not allowing it for now, it might make it clearer that you can't change the State?

CHANGELOG.md
carmichaelong: Don't think that should be possible, so maybe more documentation about that to be clearer?

carmichaelong:
@aseth1 @chrisdembia I wrote a new test where the state gets updated and used in a new Manager a few times, and the caching issue has crept up again. I think this is major enough that this should get fixed, but I don't have enough knowledge of the caching mechanism to know how to fix it. 

See `testManager.cpp` for the new test, where I drop a ball 3 times, with different starting heights and speeds. Calculating the height and speed of the ball at the end of integration is correct if you use calls to the Coordinate, but the caching issue slips in for the Station. The information printed to console for the test is as follows (note that the station is only incorrect at the velocity level on the third call):
```
Slider: t = 1, h = -4.905, v = -9.81 | Eq: t = 1, h = -4.905, v = -9.81 | Station: t = 1, h = -4.905, v = -9.81
Slider: t = 3, h = -5.32, v = -19.12 | Eq: t = 3, h = -5.32, v = -19.12 | Station: t = 3, h = -5.32, v = -19.12
Slider: t = 6, h = -39.145, v = -29.93 | Eq: t = 6, h = -39.145, v = -29.93 | Station: t = 6, h = -39.145, v = -19.12
```

OpenSim/Simulation/Test/testManager.cpp
```diff
@@ -136,3 +141,83 @@ void testStationCalcWithManager()
...
+    std::vector<double> integFinalTimes = {1.0, 3.0, 6.0};
+    std::vector<double> initHeights = {0.0, 13.3, 6.5};
+    std::vector<double> initSpeeds = {0.0, 0.5, -0.5};
+    state.setTime(integInitTimes[0]);
+    int n = integFinalTimes.size();
+
+    for (int i = 0; i < n; ++i) {
+        // Set initial state for integration and check that it's correct
+        sliderCoord.setValue(state, initHeights[i]);
+        sliderCoord.setSpeedValue(state, initSpeeds[i]);
```
aseth1: Changing the state without an explicit realize to at least velocity doesn't seem right. I would always realize after changing the state.  Does that make a difference?

aseth1: I just tried it and the test passes locally.

aseth1:
@carmichaelong I pushed a commit to verify that the change fixes your test failure.

OpenSim/Simulation/Test/testManager.cpp
chrisdembia: Shouldn't you then get an error that the state is not realized to the appropriate stage?

carmichaelong: This seems fragile. Also, it seems doubly weird since within the `integrate()` function itself, it just always calls `realizeVelocity()`. See [this line](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Manager/Manager.cpp#L679).

aseth1: > This seems fragile.

I agree.  I also don't know why we don't get an error/exception otherwise. There might be something up with State copying. I wouldn't expect the clone to copy the cache. I agree something is still amiss with the Manager.

OpenSim/Simulation/Manager/Manager.cpp
```diff
@@ -624,22 +625,22 @@ hasStateStorage() const
...
+        throw Exception("Manager::integrate(): Manager has not be "
+            "initialized. Call Manager::initialize() first.");
     }
 
     // CLEAR ANY INTERRUPT
     // Halts must arrive during an integration.
     clearHalt();
 
     // CHECK SPECIFIED DT STEPPING
+    SimTK::State s = _integ->getState();
```
aseth1: @carmichaelong this is the line causing the issue. Here we get a local copy that we then realize to velocity, but the internal state of the integrator is unchanged. If I update this to be  `const SimTK::State&` the issue goes away and we do not need the `model.realizeVelocity()` call I added. There is another issue that we didn't get an error that the state was not adequately realized. If I invoke `myStation->getLocationInGround(initState)` we do get at the cache is does not satisfy its prerequisites.

aseth1: I also updated `Manager`:
```cpp
    const SimTK::State& integrate(double finalTime);
    const SimTK::State& getState() const;
```
Now, return const ref. This did require updating Analysi:begin() and end() to also take const SimTK::State&`,  but I believe we have now eliminated unnecessary local copies of the `State`.

I'll push these changes once I see all the tests pass locally.

aseth1:
@carmichaelong let me know what you think of the changes that were pushed.

carmichaelong:
@aseth1 The changes all make sense, thanks! I started working on this again by writing another test case, but with trying to update the state using `setExcitation()` on the `arm26.osim` model. For some reason, this didn't seem to affect anything, though using `setActivation()` did seem to be doing things. Will have to look deeper into why this is the case.

aseth1:
> trying to update the state using setExcitation() on the arm26.osim model. For some reason, this didn't seem to affect anything, though using setActivation() did seem to be doing things

For muscles with activation dynamics, excitation is a a control (discrete state variable). If it is not set on the next time step it uses its default, which is likely zero. If there is a controller then its computed controls are being added in. In other words, the controls just changes for the initial time of the integration, and thus it may not have a perceivable effect on the activation trajectory. Let me know what you find!

carmichaelong:
I call `setExcitation()` in a for loop before initializing and integrating with the `Manager`. It looks something like this:
```
    Model arm("arm26.osim");
    SimTK::State& state = arm.initSystem();
    state.setTime(0);
    const Set<Muscle> &muscleSet = arm.getMuscles();
    double stepsize = 0.1;
    for (int i = 0; i < 10; ++i)
    {
        for (int j = 0; j < muscleSet.getSize(); ++j)
        {
            muscleSet.get(j).setExcitation(state, 0.1*i);
        }

        Manager manager(arm);
        manager.initialize(state);
        cout << state.getTime() << " ";
        state = manager.integrate(stepsize*(i + 1));
        arm.realizeDynamics(state);
        cout << muscleSet.get(0).getActivation(state) << endl;
    }
```
From what I understand in your comment is that this should work since it's updating the excitation at every time step? So in this case, I should be updating the excitation to be 0, 0.1, 0.2, ..., 0.9 every 0.1 seconds. The output is as follows:
```
0 0.0129403
0.1 0.01
0.2 0.01
0.3 0.01
0.4 0.01
0.5 0.01
0.6 0.01
0.7 0.01
0.8 0.01
0.9 0.01
```

aseth1:
> cout << muscleSet.get(0).getActivation(state) << endl;

This will print the activation not excitation. At the instant you change the excitation it does not affect the activation, it needs to integrate the excitation->activation dynamics. But, at he next integration time step, excitation will return to its default value.

carmichaelong:
@aseth1 and @chrisdembia This should be ready to be looked at again (though we'll still have to wait for sourceforge to resolve its issues before travis can test).

OpenSim/Simulation/Test/testManager.cpp
```diff
@@ -136,3 +162,138 @@ void testStationCalcWithManager()
...
+    PrescribedController* controller = new PrescribedController();
+    controller->addActuator(muscleSet.get(0));
+    Constant* fn = new Constant(0);
+    controller->prescribeControlForActuator(0, fn);
+    arm.addController(controller);
+
+    SimTK::State& state = arm.initSystem();
+    state.setTime(0);
+    double stepsize = 0.01;
+    muscleSet.get(0).setActivation(state, 0.1);
```
aseth1: How about if you make set `initAct = 0.1` here and then call `setActivation(state, initAct)` in the loop before initializing the `Manager`? That would also test that you could update the activation state (like  you did with Coordinates previously).

OpenSim/Simulation/Test/testManager.cpp
```diff
@@ -136,3 +162,138 @@ void testStationCalcWithManager()
...
+        double finalAct = muscleSet.get(0).getActivation(state);
+        double finalExcitation = muscleSet.get(0).getExcitation(state);
+        cout << state.getTime() << " " << excitation;
+        cout << " " << finalAct << endl;
+        // Check if excitation is correct
+        SimTK_TEST_EQ(excitation, finalExcitation);
+        // If excitation is not set correctly, it will go back to the default
+        // after the first integration. We can check if this is happening by
+        // ensuring the activation is increasing on each integration.
+        SimTK_TEST(finalAct > initAct);
```
aseth1: In line with the suggested changes, add `initAct = finalAct` to use to initialize the next integration loop.

OpenSim/Simulation/Test/testManager.cpp
```diff
@@ -136,3 +162,138 @@ void testStationCalcWithManager()
...
+    arm.addController(controller);
+
+    SimTK::State& state = arm.initSystem();
+    state.setTime(0);
+    double stepsize = 0.01;
+    muscleSet.get(0).setActivation(state, 0.1);
+    for (int i = 0; i < 3; ++i)
+    {
+        arm.realizeDynamics(state);
+        double initAct = muscleSet.get(0).getActivation(state);
```
aseth1: 275-276 could be removed if you `setActivation(state, initAct)` here

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -240,17 +254,34 @@ class OSIMSIMULATION_API Manager
...
+    * @endcode
+    *
+    * Example: Integrate from time = 0s to time = 10s, updating the state
+    *          (e.g., the model's first coordinate value) every 2s
+    * @code
+    * dTime = 2.0;
+    * finalTime = 10.0;
+    * int n = int(round(finalTime/dTime));
+    * state.setTime(0.0);
+    * manager.initialize(state);
```
chrisdembia: This line shouldn't be here.

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -240,17 +254,34 @@ class OSIMSIMULATION_API Manager
...
     *
     */
-    bool integrate(SimTK::State& s, double finalTime);
-    /** <b>(Deprecated)</b> Integrate to a specified finalTime using
-        Manager::integrate(SimTK::State&, double). */
-    DEPRECATED_14("Integrate to a specified finalTime using Manager::integrate(SimTK::State&, double).")
-    bool integrate(SimTK::State& s);
+    const SimTK::State& integrate(double finalTime);
+
+    const SimTK::State& getState() const;
```
chrisdembia: Consider adding a description for this function.

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -189,12 +189,12 @@ class OSIMSIMULATION_API Manager
...
-    DEPRECATED_14("Integrate to a specified finalTime using Manager::integrate(SimTK::State&, double).")
+        Manager::integrate(double). */
+    DEPRECATED_14("Integrate to a specified finalTime using Manager::integrate(double).")
     void setFinalTime(double aTF);
     /** <b>(Deprecated)</b> Integrate to a specified finalTime using
-        Manager::integrate(SimTK::State&, double). */
-    DEPRECATED_14("Integrate to a specified finalTime using Manager::integrate(SimTK::State&, double).")
+        Manager::integrate(double). */
+    DEPRECATED_14("Integrate to a specified finalTime using Manager::integrate(double).")
     double getFinalTime() const;
```
chrisdembia: Does this have any affect anymore?

chrisdembia: I'm not sure the "deprecated" way of using the manager actually works anymore.

carmichaelong: You're right, these functions shouldn't really do what you expect since they just update internal variables `_ti` and `_tf`. Also, I'm pretty sure you're right that these deprecated functions and the "deprecated" way of using the manager should just be removed.

OpenSim/Simulation/Manager/Manager.cpp
```diff
@@ -624,22 +589,21 @@ hasStateStorage() const
...
+const SimTK::State& Manager::integrate(double finalTime)
 {
     int step = 0; // for AnalysisSet::step()
 
-    if (!_integ) {
-        throw Exception("Manager::integrate(): "
-            "Integrator has not been set. Construct the Manager "
-            "with an integrator, or call Manager::setIntegrator().");
+    if (_timeStepper == nullptr) {
+        throw Exception("Manager::integrate(): Manager has not be "
```
chrisdembia: be -> been

OpenSim/Simulation/Manager/Manager.cpp
```diff
@@ -733,26 +695,16 @@ integrate(SimTK::State& s, double finalTime)
         if (checkHalt()) break;
     }
     finalize(_integ->updAdvancedState());
```
chrisdembia: It would be good to create a separate issue about determining if this should use `_integ->getState()` (if such an issue does not exist yet). No need to address in this PR.

carmichaelong: Opened up an issue for this.

chrisdembia:
@aseth1 you already accepted this PR and the lastest round of changes consisted of just a typo fix. I think this could be merged immediately, but I'll let you do it since you're working on the Manager.

Issue #1744, opened 2017-06-03
Sporadic failure of testExternalLoads on AppVeyor, sporadic segfaults of testExternalLoads and testControllers on UNIX


chrisdembia:
@aymanhab raised this earlier: testExternalLoads is failing sporadically on AppVeyor:

```
51/99 Test #41: testExternalLoads ............................***Exception: Other  8.52 sec
Updating Model file from 10905 to latest format...
Loaded model Pendulum from file Pendulum.osim
PointKinematics.setBody: set body to cylinder
```

I saw this on the 64-bit build (https://ci.appveyor.com/project/opensim-org/opensim-core/build/6582/job/bvto06yaaub062lp).

tkuchida:
testExternalLoads is also segfaulting on Travis ([example](https://travis-ci.org/opensim-org/opensim-core/jobs/255831134)).

chrisdembia:
Has anyone seen `testExternalLoads` segfaulting on their local machine?

chrisdembia:
I've reproduced the sporadic segfaults on my mac. Here's a backtrace of where the segfault occurs (the line numbers are a little off b/c I added debug print outs):

```
  * frame #0: 0x0000000101b2bc45 libosimCommon_d.dylib`OpenSim::Component::getStateVariableValues(this=0x00007fff5fbfe738, state=0x00007fff5fbfef88) const at Component.cpp:902
    frame #1: 0x0000000100c69880 libosimSimulation_d.dylib`OpenSim::Manager::initializeStorageAndAnalyses(this=0x00007fff5fbfe608, s=0x00007fff5fbfef88) at Manager.cpp:797
    frame #2: 0x0000000100c68f0f libosimSimulation_d.dylib`OpenSim::Manager::integrate(this=0x00007fff5fbfe608, s=0x00007fff5fbfef88, finalTime=0) at Manager.cpp:680
    frame #3: 0x000000010000dccd testExternalLoads`testExternalLoad() at testExternalLoads.cpp:121
    frame #4: 0x000000010000d954 testExternalLoads`main at testExternalLoads.cpp:36
    frame #5: 0x00007fffa7ed5235 libdyld.dylib`start + 1
```

Here's the code:

```cpp
  SimTK::Vector Component::
      getStateVariableValues(const SimTK::State& state) const
  {
      // Must have already called initSystem.
      OPENSIM_THROW_IF_FRMOBJ(!hasSystem(), ComponentHasNoSystem);

[...]

PR #1911, opened 2017-09-06
Make setStateVariableValues() correct by not enforcing constraints on every individual setValue()

aseth1:
Fixes issue #1884 and maybe #1744

Also fixes the incorrect procedure (bug) of enforcing constraints on each `Coordinate::setValue()`  which neglects the effect of other coordinate changes on the current coordinate. For example, take two identical blocks sliding on the same axis (1dof each), and apply a constraint that requires a face on each block to be welded together and that is satisfied when both blocks are at a coordinate value of 0, respectively. If both blocks are being adjusted together they will reassemble where you put them, but if you set block 1 to +1m along its coordinate, the constraint will be violated and cause both blocks to be moved by +0.5m to minimize the coordinate changes when the constraint is enforced. Then moving block 2 to 1m will cause the assembly to split the difference with block 1 (now at 0.5m) and now both blocks will end up at 0.75 for their coordinate values and not 1, had both values been set prior to enforcement of the constraint.

### Brief summary of changes
- Implemented Option described by @chrisdembia in #1884 so that `CoordinateStateVariable::setValue()` does not enforce contraints.
- Added suggested changes to Doxygen
-  Addressed the proposed issue behind #1744 by clearing the cached list of StateVariables during `Component::baseAddToSystem()`.

### Testing I've completed
Instrumented and compared the performance of `model.setStateVariableValues()` for the assembly test model of the full body with hand and feet constrained to the floor in a push-up.  Speedup of 19x.

Are the Doxygen comments reasonable?

### CHANGELOG.md (choose one)
- updated because `model.assemble()` is required and answers before could have been incorrect and thus values may change.


OpenSim/Common/Component.h
```diff
@@ -1256,9 +1256,15 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
 
     /**
      * %Set all values of the state variables allocated by this Component.
-     * Includes state variables allocated by its subcomponents.
+     * Includes state variables allocated by its subcomponents. Note, this
+     * method simply sets the values on the input State. If other conditions
+     * must be met (such as fiber and tendon equilibrium for muscles, or to
+     * satisfy kinematic constraints for Coordinates) you must invoke the
+     * appropriate methods on Model (e.g. assemble() to satisfy constraints or
+     * equilibrateMuscles()) to satisfy these conditions starting from the
```
chrisdembia: Use the same ordering within the parentheses outside the parentheses ((1) muscles (2) coordinates).

CHANGELOG.md
```diff
@@ -101,6 +101,7 @@ Bug Fixes
...
...peed issue with `model.setStateVariableValues()` caused by enforcing constraints after each coordinate value was being set (PR #1911). Removing the automatic enforcement of constraints makes setting all state values much faster, but also requires calling `model.assemble()` afterwards. Enforcing constraints after setting each coordinate value individually was also incorrect, since it neglected the effect of other coordinate changes have on the current coordinate. For example, take two identical blocks sliding on the same axis (1dof each), and apply a constraint that requires a face on each block to be welded together and that is satisfied when both blocks are at a coordinate value of 0, respectively. If both blocks are being adjusted together they will reassemble where you put them, but if you set block 1 to +1m along its coordinate, the constraint will be violated and cause both blocks to be moved by +0.5m to minimize the coordinate changes. Then moving block 2 to 1m will cause the assembly to split the difference with block 1 (now at 0.5m) and now both blocks will end up at 0.75 for their coordinate values and not 1, had both values been set prior to enforcement of the constraint.
```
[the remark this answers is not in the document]
tkuchida: :+1: Suggest removing from "For example, …". In its place, I suggest adding a concise statement of what's correct (e.g., "All coordinate values must be set before enforcing constraints.").

OpenSim/Common/Component.h
```diff
@@ -1256,9 +1256,15 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
...
+     * State values provided by setStateVariableValues.
      *
-     * @param state   the State for which to get the value
+     * @param state   the State for which to set State values upon
```
tkuchida: Rearranged, this description is "The State to set State values upon for [sic]". Perhaps "The State whose values are set"?

OpenSim/Simulation/SimbodyEngine/Coordinate.h
```diff
@@ -132,11 +132,16 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Coordinate, ModelComponent);
...
-    void setValue(SimTK::State& s, double aValue, bool aEnforceContraints=true) const;
+    /** Set the value of the Coordinate on to the state.
+        Optional flag to enforce the constraints immediately (true by default)
+        which will adjust its value in the state to satisfy model constraints.
+        Use getValue(s) to see if/how the value was adjusted to satisfy the
+        kinematic constraints. If setting multiple Coordinate values at a time,
+        set the flag to false and then call Model::assemble(state).
+        Alternatively, use Model::setStateVariableValues() to set all coordinate
+        values and their speeds at once followed by Model::assemble(state).
+        */
```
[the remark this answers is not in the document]
tkuchida: I think it's misleading to say "which will adjust its value in the state to satisfy model constraints" because the projection will change *all* coordinates in general.

OpenSim/Simulation/SimbodyEngine/Coordinate.h
```diff
@@ -132,11 +132,16 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Coordinate, ModelComponent);
...
-    /** set the value of the Coordinate on to the state.
-        optional flag to enforce the constraints immediately, which may 
-        adjust its value in the state. Use getValue(s) to see if/how the
-        value was adjusted to satisfy the kinematic constraints. */
-    void setValue(SimTK::State& s, double aValue, bool aEnforceContraints=true) const;
+    /** Set the value of the Coordinate on to the state.
+        Optional flag to enforce the constraints immediately (true by default)
+        which will adjust its value in the state to satisfy model constraints.
+        Use getValue(s) to see if/how the value was adjusted to satisfy the
+        kinematic constraints. If setting multiple Coordinate values at a time,
```
tkuchida: It isn't clear what "If setting multiple Coordinate values at a time" means. "At a time" seems to indicate that `setValue()` can be used to set more than 1 Coordinate value, but that is not correct. Perhaps "at a time" -> "consecutively"?

OpenSim/Simulation/SimbodyEngine/Coordinate.h
```diff
@@ -132,11 +132,16 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Coordinate, ModelComponent);
...
+        set the flag to false and then call Model::assemble(state).
```
tkuchida: Perhaps "…and then call Model::assemble(state) once all Coordinate values have been set"?

chrisdembia:
> I also think the important part of this PR is fixing the bug (efficiency is secondary to correctness), which should be reflected in the PR title, but I don't think this issue leaked into the code/docs (CHANGELOG correctly mentions the bug fix first).

I'm attempting to verify locally if the bug is fixed.

chrisdembia:
> I'm attempting to verify locally if the bug is fixed.

Oh sorry, I think you were talking about the enforcing-constraints-with-each-Coordinate-setValue() bug; I meant that I'm verifying the testExternalLoads bug (#1744).

chrisdembia:
Okay I ran `testExternalLoads` for about 20 minutes using this branch and I did not get a segfault. So I think this bug is fixed!

tkuchida:
@chrisdembia The PR title is "Make setStateVariableValues() more efficient by not enforcing constraints", but efficiency is a secondary issue. The important part of this PR is avoiding constraint projections when only part of the state vector has been set. The CHANGELOG seems fine, though, so I think it's just the PR title that was written with a different philosophy.

aseth1:
@tkuchida @chrisdembia I renamed the PR to put more emphasis on the bug fix (although my initial intent was just to speed things up and then  @chrisdembia pointed out the bug in enforcing constraints on individual setValue() calls), which I agree is of bigger importance and impacts the users (and thus the entry in the CHANGELOG).  I accepted your suggestions. Thank you.

chrisdembia:
One last question: Can `Component:: _statesAssociatedSystem` be removed now?

aseth1:
> One last question: Can `Component:: _statesAssociatedSystem` be removed now?

I haven't tried to do it. I did not change the underlying mechanism of caching `StateVariables` so it seemed safe to leave the check that the same System is in place. Clearing `_allStateVariables` at `addToSystem` might eliminate the need, but I rather keep it around until I make the bigger change of using indices directly to do `state.updY()` which will remove the need for both `_statesAssociatedSystem` and `_allStateVariables`.

chrisdembia:
Okay sounds good.

aseth1:
Thanks @chrisdembia and @tkuchida!

Issue #1913, opened 2017-09-08
Joint::scale() makes assumption about parent/child frame type that is inconsistent with the definition of its parent_frame and child_frame properties


tkuchida:
[`Joint::scale()`](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/SimbodyEngine/Joint.cpp#L297) assumes the parent_frame and child_frame are of type `PhysicalOffsetFrame` but [they can actually be any `PhysicalFrame`](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/SimbodyEngine/Joint.h#L115). The convenience constructor [creates `PhysicalOffsetFrame`s](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/SimbodyEngine/Joint.cpp#L85), but any concrete `PhysicalFrame` (i.e., `PhysicalOffsetFrame`, `Body`, or `Ground`) could be used as the parent_frame or child_frame. In some situations, this inconsistency could cause `Joint::scale()` to either throw an exception or have unexpected behavior (note that `SimTK_DYNAMIC_CAST_DEBUG` becomes `static_cast` in Release mode).

tkuchida:
The same code (and assumption) appears in `TwoFrameLinker::scaleFrames()`.

aseth1:
To be fair, the `Joint` only attempts to scale frames that it owns. These are the frames created by the `Joint` to represent the parent and child frames when offsets (location, orientation) are provided, which are exclusively `PhyscialOffsetFrames`. There is a possibility of adding other `Frame`s to a `Joint` but they would have to have the same name as the name as the parent or child.

I agree that his could be cleaner and we could that simply by using `getComponent<PhysicalOffsetFrame>(parentName)` instead for doing a find and a cast.

tkuchida:
> To be fair, the `Joint` only attempts to scale frames that it owns.

It looks inconsistent to me, but I see your point.

aseth1:
I agree it is inconsistent and worth fixing. There is still a bug that if manually adding a `Body` to the `Joint` (which is unusual but possible) and then setting the parent or child `Socket` to that frame the `SimTK_DYNAMIC_CAST_DEBUG` would fail or have unexpected behavior.

tkuchida:
[...]

PR #1921, opened 2017-09-09
Allow Joints to own PhysicalOffsetFrames only, not just any PhysicalFrame

tkuchida:
Fixes issue #1913

### Brief summary of changes
Changed definition of Joint's `frames` list property and updated `Joint::scale()` method accordingly. Also improved `scale()` method to search through the ScaleSet for parent/child scale factors only when necessary.

### Testing I've completed
Ran testScale.

### CHANGELOG.md (choose one)
This PR fixes a bug that did not exist in 3.3.

### DIFF
--- a/OpenSim/Simulation/SimbodyEngine/Joint.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/Joint.cpp
@@ -266,43 +266,58 @@ bool Joint::isCoordinateUsed(const Coordinate& aCoordinate) const
 
 void Joint::scale(const ScaleSet& scaleSet)
 {
+    // Scale the parent and/or child Frame, but only if owned by the Joint. The
+    // Joint owns Frames that appear in the Joint's "frames" list property,
+    // which can occur only if the Frame is a PhysicalOffsetFrame.
+    auto findIndexInFramesList = [&](const PhysicalFrame& frame) -> int
+    {
+        const PhysicalOffsetFrame* pof =
+            dynamic_cast<const PhysicalOffsetFrame*>(&frame);
+        if (pof == nullptr)
+            return -1;
+        return getProperty_frames().findIndex(*pof);
+    };
+    const int parentIndex = findIndexInFramesList(getParentFrame());
+    const int childIndex  = findIndexInFramesList(getChildFrame());
+
+    if (parentIndex < 0 && childIndex < 0)  // No scaling to do.
+        return;
+
+    // If either parent or child is owned by the Joint, search the scaleSet once
+    // to find the associated scale factors.
+    const std::string& parentName = getParentFrame().findBaseFrame().getName();
+    const std::string& childName  =  getChildFrame().findBaseFrame().getName();
     SimTK::Vec3 parentFactors(1.0);
     SimTK::Vec3 childFactors(1.0);
+    bool seekingParent = (parentIndex >= 0);
+    bool seekingChild  = (childIndex  >= 0);
 
-    // Find the factors associated with the PhysicalFrames this Joint connects
-    const string& parentName = getParentFrame().findBaseFrame().getName();
-    const string& childName = getChildFrame().findBaseFrame().getName();
-    // Get scale factors
-    bool found_p = false;
-    bool found_b = false;
-    for (int i = 0; i < scaleSet.getSize(); i++) {
+    for (int i=0; i<scaleSet.getSize(); ++i) {
         Scale& scale = scaleSet.get(i);
-        if (!found_p && (scale.getSegmentName() == parentName)) {
+        // TODO: The scaling infrastructure must be updated now that bodies can
+        //       have the same name (provided their absolute paths are unique).
+        if (seekingParent && (scale.getSegmentName() == parentName)) {
             scale.getScaleFactors(parentFactors);
-            found_p = true;
+            seekingParent = false;
         }
-        if (!found_b && (scale.getSegmentName() == childName)) {
+        if (seekingChild && (scale.getSegmentName() == childName)) {
             scale.getScaleFactors(childFactors);
-            found_b = true;
+            seekingChild = false;
         }
-        if (found_p && found_b)
+        if (!seekingParent && !seekingChild)
             break;
     }
 
-    // if the frame is owned by this Joint scale it,
-    // otherwise let the owner of the frame decide.
-    int found = getProperty_frames().findIndex(getParentFrame());
-    if (found >= 0) {
-        PhysicalOffsetFrame& offset
-            = SimTK_DYNAMIC_CAST_DEBUG<PhysicalOffsetFrame&>(upd_frames(found));
-        offset.scale(parentFactors);
-    }
-    found = getProperty_frames().findIndex(getChildFrame());
-    if (found >= 0) {
-        PhysicalOffsetFrame& offset
-            = SimTK_DYNAMIC_CAST_DEBUG<PhysicalOffsetFrame&>(upd_frames(found));
-        offset.scale(childFactors);
-    }
+    OPENSIM_THROW_IF(seekingParent || seekingChild, Exception,
+        "Expected scale factors not found in ScaleSet.");
+
+    // Scale the PhysicalOffsetFrames owned by this Joint.
+    if (parentIndex >= 0)
+        updComponent<PhysicalOffsetFrame>(
+            upd_frames(parentIndex).getAbsolutePathName()).scale(parentFactors);
+    if (childIndex >= 0)
+        updComponent<PhysicalOffsetFrame>(
+            upd_frames(childIndex).getAbsolutePathName()).scale(childFactors);
 }
 
 const SimTK::MobilizedBodyIndex Joint::
--- a/OpenSim/Simulation/SimbodyEngine/Joint.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/Joint.cpp
@@ -267,8 +267,8 @@ bool Joint::isCoordinateUsed(const Coordinate& aCoordinate) const
 void Joint::scale(const ScaleSet& scaleSet)
 {
     // Scale the parent and/or child Frame, but only if owned by the Joint. The
-    // Joint owns Frames that appear in the Joint's "frames" list property,
-    // which can occur only if the Frame is a PhysicalOffsetFrame.
+    // Joint owns Frames that appear in the Joint's "frames" list property
+    // (which can occur only if the Frame is a PhysicalOffsetFrame).
     auto findIndexInFramesList = [&](const PhysicalFrame& frame) -> int
     {
         const PhysicalOffsetFrame* pof =
@@ -283,7 +283,7 @@ void Joint::scale(const ScaleSet& scaleSet)
     if (parentIndex < 0 && childIndex < 0)  // No scaling to do.
         return;
 
-    // If either parent or child is owned by the Joint, search the scaleSet once
+    // If parent and/or child is owned by the Joint, search the scaleSet *once*
     // to find the associated scale factors.
     const std::string& parentName = getParentFrame().findBaseFrame().getName();
     const std::string& childName  =  getChildFrame().findBaseFrame().getName();
@@ -313,11 +313,9 @@ void Joint::scale(const ScaleSet& scaleSet)
 
     // Scale the PhysicalOffsetFrames owned by this Joint.
     if (parentIndex >= 0)
-        updComponent<PhysicalOffsetFrame>(
-            upd_frames(parentIndex).getAbsolutePathName()).scale(parentFactors);
+        upd_frames(parentIndex).scale(parentFactors);
     if (childIndex >= 0)
-        updComponent<PhysicalOffsetFrame>(
-            upd_frames(childIndex).getAbsolutePathName()).scale(childFactors);
+        upd_frames(childIndex).scale(childFactors);
 }
 
 const SimTK::MobilizedBodyIndex Joint::
--- a/OpenSim/Simulation/SimbodyEngine/Joint.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/Joint.cpp
@@ -269,9 +269,8 @@ void Joint::scale(const ScaleSet& scaleSet)
     auto scaleOffsetFrameIfOwned = [&](const PhysicalFrame& frame) -> void {
         // If the frame is owned and immediate subcomponent of this Joint then
         // scale it. Otherwise, let the immediate owner of the frame decide.
-        if (auto* offset = findComponent<PhysicalOffsetFrame>(frame.getName())
-            && (&(offset->getOwner()) == this))
-        {
+        auto* offset = findComponent<PhysicalOffsetFrame>(frame.getName());
+        if (offset && (&(offset->getOwner()) == this)) {
             // find the scale factors associated with the base PhysicalFrame
             // of this offset frame
             const string& baseFrameName = offset->findBaseFrame().getName();
--- a/OpenSim/Simulation/SimbodyEngine/Joint.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/Joint.cpp
@@ -267,9 +267,10 @@ bool Joint::isCoordinateUsed(const Coordinate& aCoordinate) const
 void Joint::scale(const ScaleSet& scaleSet)
 {
     auto scaleOffsetFrameIfOwned = [&](const PhysicalFrame& frame) -> void {
-        // If the frame is owned by this Joint then scale it.
-        // Otherwise let the owner of the frame decide.
-        if (auto* offset = findComponent<PhysicalOffsetFrame>(frame.getName()))
+        // If the frame is owned and immediate subcomponent of this Joint then
+        // scale it. Otherwise, let the immediate owner of the frame decide.
+        if (auto* offset = findComponent<PhysicalOffsetFrame>(frame.getName())
+            && (&(offset->getOwner()) == this))
         {
             // find the scale factors associated with the base PhysicalFrame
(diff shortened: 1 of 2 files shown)


tkuchida:
@aseth1 This PR is WIP. Please let me know what you want to happen if a Joint owns a PhysicalOffsetFrame but that PhysicalOffsetFrame isn't used to satisfy either its parent_frame or child_frame Socket. It seems that the Joint needs to be responsible for scaling that PhysicalOffsetFrame as well, but currently I think that PhysicalOffsetFrame would go unscaled. I think it's also possible for a PhysicalOffsetFrame (or any Component) to own a PhysicalOffsetFrame via its `components` property. When does that PhysicalOffsetFrame get scaled? There are probably other cases related to cascading ownership. I'm concerned that some Components are going to slip through the cracks.

I suggest propagating scale factors down the topological tree (analogous to how `finalizeFromProperties()` propagates down the ownership tree), e.g., having `Model::scale()` call `scale()` on all bodies in its ~~BodySet~~ `getComponentList<Body>` and have each Body scale any PhysicalOffsetFrames hanging onto it (though I'm not sure a Body currently knows the PhysicalOffsetFrames for which it's the base Frame).

Perhaps we just need to state in the Release Notes that scaling has been tested only for 3.3-like Models, and worry about handling modern Models post-4.0.

Please also let me know what tests need to be added at this point. There are many scaling-related situations (with cascading PhysicalOffsetFrames, "unorthodox" ownership trees, etc.) that need to be covered.

tkuchida:
I think this PR can be reviewed: it fixes Issue #1913 as advertised and the larger design question has been raised in Issue #1923.

OpenSim/Simulation/SimbodyEngine/Joint.cpp
```diff
@@ -266,43 +266,56 @@ bool Joint::isCoordinateUsed(const Coordinate& aCoordinate) const
...
-            = SimTK_DYNAMIC_CAST_DEBUG<PhysicalOffsetFrame&>(upd_frames(found));
-        offset.scale(parentFactors);
-    }
-    found = getProperty_frames().findIndex(getChildFrame());
-    if (found >= 0) {
-        PhysicalOffsetFrame& offset
-            = SimTK_DYNAMIC_CAST_DEBUG<PhysicalOffsetFrame&>(upd_frames(found));
-        offset.scale(childFactors);
-    }
+    OPENSIM_THROW_IF(seekingParent || seekingChild, Exception,
```
aseth1: I believe the current expectation is that the `ScaleSet` defines what is to be scaled, not the existence of the `PhyscialOffsetFrame`. In other words, if there is no `ScaleFactor` for this offset's base frame then it just doesn't get scaled. If you purposefully remove scale factors for a `Body` from the `ScaleSet` would should scaling throw?

OpenSim/Simulation/SimbodyEngine/Joint.cpp
```diff
@@ -266,43 +266,56 @@ bool Joint::isCoordinateUsed(const Coordinate& aCoordinate) const
 
 void Joint::scale(const ScaleSet& scaleSet)
 {
+    // Scale the parent and/or child Frame, but only if owned by the Joint. The
```
aseth1: The previous implementation was overly complex/confusing and your fix can go a long way to simplify it. We can use `findComponent<PhysicalOffsetFrame>` to simplify further. I tested the following and it works:

```cpp
void Joint::scale(const ScaleSet& scaleSet)
{
    auto scaleOffsetFrameIfOwned = [&](const PhysicalFrame& frame) -> void {
        // If the frame is owned by this Joint then scale it.
        // Otherwise let the owner of the frame decide.
        if (auto* offset = findComponent<PhysicalOffsetFrame>(frame.getName()))
        {
            // find the scale factors associated with the base PhysicalFrame
            // of this offset frame
            const string& baseFrameName = offset->findBaseFrame().getName();

            // Get scale factors if they exist
            for (int i = 0; i < scaleSet.getSize(); i++) {
                Scale& scale = scaleSet.get(i);
                if (scale.getSegmentName() == baseFrameName) {
                    SimTK::Vec3 factors(1.0);
                    scale.getScaleFactors(factors);
                    const_cast<PhysicalOffsetFrame*>(offset)->scale(factors);
                    return;
                }
            }
        }
    };

    scaleOffsetFrameIfOwned(getParentFrame());
    scaleOffsetFrameIfOwned(getChildFrame());
}
```
Something like this might be easier to follow and it avoids the `dynamic_cast`.

aseth1:
> Perhaps we just need to state in the Release Notes that scaling has been tested only for 3.3-like Models, and worry about handling modern Models post-4.0.

That seems most reasonable to me. I the meantime we should test the conditions/cases that you need to make progress on your work. We'll learn a lot by tackling your issues. Thanks for your willingness to help!

OpenSim/Simulation/SimbodyEngine/Joint.cpp
tkuchida: > The previous implementation was overly complex/confusing and your fix can go a long way to simplify it.

The purpose of this PR is to fix a bug and avoid searching through `scaleSet` unnecessarily.

> We can use `findComponent<PhysicalOffsetFrame>` to simplify further. I tested the following and it works:

I believe that `findComponent()` would work but I disagree with using it here. I think the intent of `dynamic_cast` is different from the intent of `findComponent()` (it doesn't matter that they produce the same result). We already have references to the parent and child Frames; there is no need to find them. What we need to do is check whether these Frames are PhysicalOffsetFrames, and it appears that the convention in OpenSim is to use `dynamic_cast` to do this. A reasonable implementation of `findComponent()` could be to navigate to the provided absolute path and check whether the Component exists and is of type PhysicalOffsetFrame; if not, continue looking around for a Component of type PhysicalOffsetFrame of the provided name (e.g., in the ownership subtree below the provided absolute path). I believe that is the semantics of a "find" operation and would be the interpretation of such a line of code, but we don't want to do that here lest we accidentally find some Frame other than the parent/child.

> Something like this might be easier to follow and it avoids the `dynamic_cast`.

I think `findComponent()` is more difficult to follow because you are searching for something you already have. Please explain why `dynamic_cast` is acceptable elsewhere in the repo but not here. It would help me to understand your objective function. Are you trying to avoid using `dynamic_cast` for some reason? Are you trying to minimize the number of lines of code?

OpenSim/Simulation/SimbodyEngine/Joint.cpp
tkuchida: I think ScaleSet should contain exactly one entry for each Body (even if that entry is `Vec3(1)`). Searching through a ScaleSet and scaling by `Vec3(1)` by default if no entry is found sounds dangerous (e.g., a Body name could be misspelled and skipped over silently). ScaleSet's `finalizeFromProperties()` should ensure there is exactly one entry for each Body. If that is not currently the case, I think it should be made so. Having one entry per Body will also improve the search time (will always require `n(n+1)/2` string comparisons for `n` bodies; the worst-case cost is `n^2` otherwise).

tkuchida:
> In the meantime we should test the conditions/cases that you need to make progress on your work.

Yes, I now realize that I need to write comprehensive tests for opensim-core.

OpenSim/Simulation/SimbodyEngine/Joint.cpp
tkuchida: > and it avoids the `dynamic_cast`

Also unclear why `dynamic_cast` should be avoided here since `findComponent()` uses `dynamic_cast` anyway. If you don't want `dynamic_cast` in this method, we could move that piece of `findComponent()` (i.e., `found = dynamic_cast<const C*>(this); if (found) return found;`) to a new method like `getAs<PhysicalOffsetFrame*>(absolute_path)` and call `getAs` on line 274.

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -102,12 +103,12 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
         "List containing the generalized coordinates (q's) that parameterize "
         "this joint.");
 
-    OpenSim_DECLARE_LIST_PROPERTY(frames, PhysicalFrame,
-        "Physical frames owned by the Joint that are used to satisfy the Joint's "
-        "parent and child frame connections. For examples, PhysicalOffsetFrames "
+    OpenSim_DECLARE_LIST_PROPERTY(frames, PhysicalOffsetFrame,
+        "Physical offset frames owned by the Joint that are used to satisfy "
+        "the Joint's parent and child frame connections. PhysicalOffsetFrames "
```
tkuchida: I accidentally deleted a comment from @aseth1 here. The comment was ":+1:".

OpenSim/Simulation/SimbodyEngine/Joint.cpp
aseth1: > The purpose of this PR is to fix a bug and improve performance (by avoiding unnecessary searches through scaleSet). I agree with simplifying code in general, but I don't think the Development Team can afford the time required to make that a priority.

I provided an alternative that maintains the purpose of the PR and IMHO is also more readable. I this case both are achievable.

>I believe that findComponent() would work but I disagree with using it here. I think the intent of dynamic_cast is different from the intent of findComponent() (it doesn't matter that they currently produce the same result) ...  but that's the wrong idea: we don't want to do that here lest we accidentally find some Frame other than the parent/child.

Your current code will not find the parent/child `PhysicalOffsetFrame` that was added by `addComponent` to the `Joint`. That was part of the criticism of the previous implementation. You are trying to *find* the parent/child `PhysicalOffsetFrame` that is owned by the `Joint`, which IMHO is the purview of `findComponent`.  You could thwart the issue of finding some other `PhysicalOffsetFrame` in the `Joint`  with the same name as the child/parent frame, by verifying its immediate `owner` is the `Joint`. I agree your current use of property's `findIndex()` ensure it is an immediate subcomponent of the `Joint` but it does not allow the child/parent to have been add via `addComponent`. 

> Also unclear why dynamic_cast should be avoided here since findComponent() uses dynamic_cast anyway

That's the point. Why pollute the implementation with your own find and dynamic cast  when there is a utility method `findComponent` to do it for you.  

To me these implementation details are not worth this level of argument.  I made a suggestion to be helpful. If you strongly believe there is nothing helpful in suggestion if the other reviewer is fine with your implementation I am too. Let's move on.

OpenSim/Simulation/SimbodyEngine/Joint.cpp
aseth1: > Searching through a ScaleSet and scaling by Vec3(1) by default if no entry is found sounds like a bad design (e.g., a Body name could be misspelled and skipped over without noticing).

That is not exactly what I proposed. If a `Body` is not listed it should not be scaled - not scaled by Vec3(1).  A `ScaleFactor` with an illegitimately named `Body` should be made to throw. Your approach would require the user to know about every `Body` in the `Model` in order to scale it. If I embed bodies in a Device model that implements its own scale say based on distances between anchor points, it seems far too burdensome for the consumer of that Device to have to include its bodies in order for the scaling to work.

OpenSim/Simulation/SimbodyEngine/Joint.cpp
tkuchida: > I provided an alternative that maintains the purpose of the PR and IMHO is also more readable. I this case both are achievable.

I find `dynamic_cast` more readable because it's immediately obvious what it does; `findComponent()` could do all sorts of things (including tree traversals, incurring overhead, etc.) that I think would be best to avoid if possible.

> Your current code will not find the parent/child `PhysicalOffsetFrame` that was added by `addComponent` to the `Joint`. That was part of the criticism of the previous implementation.

The criticism of the previous implementation is stated in #1913: although it is not currently covered by a test case, `Joint::scale()` will throw an exception (or have unexpected behavior) in certain situations. This PR is not intended (or advertised) to address the larger design flaw reported in #1923. Fixing that issue is a big job and will require design work, test case development, etc. Currently, the car doesn't work; I don't think it makes sense to worry about the fact that the windshield wipers don't work since (in this analogy) they will probably need to be removed to fix the car—and the final design might not need windshield wipers. This PR is mainly trying to keep the wheels from falling off.

> You are trying to *find* the parent/child `PhysicalOffsetFrame` that is owned by the `Joint`, which IMHO is the purview of `findComponent`.  You could thwart the issue of finding some other `PhysicalOffsetFrame` in the `Joint`  with the same name as the child/parent frame, by verifying its immediate `owner` is the `Joint`. I agree your current use of property's `findIndex()` ensure it is an immediate subcomponent of the `Joint` but it does not allow the child/parent to have been add via `addComponent`.

Using `findComponent()` and confirming that the immediate owner is the Joint would indeed address the situation you mention, but what if the Joint owns a component (added via `addComponent()`) that then owns a PhysicalOffsetFrame to which the Joint connects? I think the scaling infrastructure is more than a couple tweaks away from working and requires a more rigorous approach.

tkuchida:
@aseth1 I have replaced my implementation with your code. If you would like to make further changes, please push directly to this branch.

OpenSim/Simulation/SimbodyEngine/Joint.cpp
```diff
@@ -266,43 +266,30 @@ bool Joint::isCoordinateUsed(const Coordinate& aCoordinate) const
...
-            scale.getScaleFactors(parentFactors);
-            found_p = true;
-        }
-        if (!found_b && (scale.getSegmentName() == childName)) {
-            scale.getScaleFactors(childFactors);
-            found_b = true;
+    auto scaleOffsetFrameIfOwned = [&](const PhysicalFrame& frame) -> void {
+        // If the frame is owned by this Joint then scale it.
+        // Otherwise let the owner of the frame decide.
+        if (auto* offset = findComponent<PhysicalOffsetFrame>(frame.getName()))
```
tkuchida: It doesn't seem like this line does what the comment above advertises. Doesn't there need to be a check to satisfy the condition "If the frame is owned by this Joint"?

OpenSim/Simulation/SimbodyEngine/Joint.cpp
aseth1: > I think the scaling infrastructure is more than a couple tweaks away from working and requires a more rigorous approach.

I totally agree. It would be great to work with you on it!

OpenSim/Simulation/SimbodyEngine/Joint.cpp
chrisdembia: This code will ensure that the frame is an "ownership" descendant of the Joint, but does not ensure that the frame is a direct descendant (immediate subcomponent). Which of these two meanings is meant by "frame is owned by this Joint"?

aseth1: I don't think it really matters where the frame lives inside the `Joint` as long as it is the *child* or *parent* frame. If it is some other frame it should not get scaled in this scheme.

tkuchida: > I don't think it really matters where the frame lives inside the `Joint` as long as it is the *child* or *parent* frame. If it is some other frame it should not get scaled in this scheme.

Let *F* be a Frame owned by a Joint, where *F* is neither the Joint's child nor parent frame. Where does *F* get scaled "in this scheme"?

asethca: > Let F be a Frame owned by a Joint, where F is neither the Joint's child nor parent frame. Where does F get scaled "in this scheme"?

*F* does not get scaled by the `Joint` in this scheme. It is not an attempt at a solution to scaling, which is a bigger issue. In the existing scheme (which this PR does not change), the Joint has jurisdiction over the scaling of *parent* and *child* frames that are PhysicalOffsetFrames. There is no change to that.

A better scheme may be for `Joint` to have no say in the scaling of frames whatsoever and that require the *base* frame to scale `PhysicalOffsetFrame`s attached to it, regardless of ownership. Let's work out a better scheme for post 4.0.

tkuchida: > I don't think it really matters where the frame lives inside the `Joint` as long as it is the *child* or *parent* frame. If it is some other frame it should not get scaled in this scheme.

I think there's still an issue here that will result in some PhysicalOffsetFrames being scaled 0 times, some once, and some twice. I believe you're advocating that the frames that will be scaled 0 times should be ignored and addressed post-4.0. I think these situations should be detected and either throw an exception or at least issue a warning to the user (see suggestion added to Issue #1923).

I think the following scenario will result in a PhysicalOffsetFrame being scaled twice. Let *F* be a Frame and let *J1* and *J2* be Joints. Suppose *J1* resides in one of Model's subcomponent lists and *J2* resides in *J1*'s subcomponent list. Further suppose that *F* resides in one of *J2*'s subcomponent lists (either "frames" or "components") and that *F* is the parent or child frame for both *J1* and *J2*. I think the `findComponent()` call on L272 will cause both *J1* and *J2* to scale *F* because it is a descendant of both. Also fine if you want to push that to post-4.0, but I think it's a disservice to the user if we allow them to create models with arbitrary ownership trees (and, in fact, I believe that is one of the selling points of 4.0) without being able to scale or otherwise process such models. IMHO there at least needs to be a "scalability checker" that runs through all PhysicalOffsetFrames in the model and tells the user if any frames won't get scaled.

tkuchida:
@aseth1 and I determined that it would be more work to detect and test scenarios that aren't supported than it would be to fix scaling more generally. The changes to Joint.h are still relevant; I'll move those into another PR and then close this one.

tkuchida:
Superseded by PR #1994.

PR #1922, opened 2017-09-10
StatesTrajectory::createFromStatesStorage(): provide option to assemble

### LINKED ISSUE #1920: Correct docs for StatesTrajectory::createFromStatesStorage()
Update docs for `StatesTrajectory::createFromStatesStorage()` to say that returned states are *not* assembled (or perhaps, we could create the option to assemble the states).

chrisdembia:
Fixes #1920. 

### Brief summary of changes

- PR #1911 causes `setStateVariableValues()` to no longer enforce constraints, which changed the semantics of `StatesTrajectory::createFromStatesStorage()`. This PR gives an option to users to have constraints enforced. This option is useful because, otherwise, the user must loop through each state in the trajectory and enfornce constraints.
- This PR relies on a recent bug fix in Simbody (that allows using state objects with different copies of the same model).
- The user no longer needs to call `initSystem()` before calling `createFromStatesStorage()`.

### Testing I've completed

testStatesTrajectory passes.

### CHANGELOG.md (choose one)

- no need to update because StatesTrajectory is new to 4.0.


chrisdembia:
Thanks @tkuchida

OpenSim/Simulation/Test/testStatesTrajectory.cpp
```diff
@@ -179,13 +179,14 @@ void testFromStatesStorageGivesCorrectStates() {
 
     Storage sto(statesStoFname);
 
-    // This will fail because we have not yet called initSystem() on the model.
-    SimTK_TEST_MUST_THROW_EXC(
-            StatesTrajectory::createFromStatesStorage(model, sto),
-            OpenSim::ModelHasNoSystem);
+    // It's important that we have not yet initialized the model,
+    // since `createFromStatesStorage()` should be able to work with such a
+    // model.
```
aseth1: This comment is confusing to me. You just want o verify that you can load state trajectories without invoking `model.initSystem()` yourself.

chrisdembia:
Thanks for reviewing @aseth1. I updated the comment. I also skipped CI tests, since I was only editing a comment and the tests passed previously.

PR #1925, opened 2017-09-13
matlab side class for c3d reading and operations

jimmyDunne:
### Brief summary of changes
This is part (i) of a multi-part PR for c3d conversion using OpenSim classes.  Here, we define a Matlab side class for reading c3d files and demonstrate use.

### Testing I've completed
I have included a script to demonstrate use (.../Matlab/Examples/c3dExport). I have tested with known data and confirmed the correct behavior. Thorough tests will be written once the class has been agreed on. 

Class structure (private/public access, method names, additional methods, ...).

### DIFF
--- a/Bindings/Java/Matlab/Utilities/opensimC3D.m
+++ b/Bindings/Java/Matlab/Utilities/opensimC3D.m
@@ -1,0 +1,137 @@
+classdef opensimC3D < matlab.mixin.SetGet
+% opensimC3D C3D data into OpenSim Tables
+%   Utility  Class that uses OpenSim's C3DFileAdapter() class to produce
+%   OpenSim tables. Methods include Simbody rotation.
+    properties (Access = private)
+    path
+    markers
+    forces
+    end
+    methods
+        function obj = opensimC3D(path2c3d)
+            % Class Constructor: input is an absolute path to a C3D file. 
+            
+            % verify the file path is correct
+            if exist(path2c3d, 'file') == 0
+                error('File does not exist. Check path is correct')
+            else
+                obj.path = path2c3d;
+            end
+            % load java libs
+            import org.opensim.modeling.*
+            % Use a c3dAdapter to turn read a c3d file
+            tables = C3DFileAdapter().read(path2c3d);
+            % get the marker and force data into OpenSim tables
+            obj.markers = tables.get('markers');
+            obj.forces = tables.get('forces');
+        end            
+        function rate = getRate_marker(obj)
+            % Get the capture rate used for the Marker Data
+            rate = str2double(obj.markers.getTableMetaDataAsString('DataRate'));
+        end
+        function rate = getRate_force(obj)
+            % Get the capture rate used for the Force Data
+            rate = str2double(obj.forces.getTableMetaDataAsString('DataRate'));
+        end
+        function n = getNumTrajectories(obj)
+            % Get the number of markers in the c3d file
+            n = obj.markers.getNumColumns;
+        end
+        function n = getNumForces(obj)
+            % Get the number of forceplates in the c3d file
+            n = (obj.forces.getNumColumns)/3;
+        end
+        function t = getStartTime(obj)
+            % Get the start time of the c3d file
+            t = obj.markers.getIndependentColumn().get(0);
+        end
+        function t = getEndTime(obj)
+            % Get the end time of the c3d file
+            t = obj.markers.getIndependentColumn().get(obj.markers.getNumRows - 1);
+        end
+        function name = getFileName(obj)
+            % Get the name of the c3d file
+            [filedirectory, name, extension] = fileparts(obj.path);
+        end
+        function filedirectory = getDirectory(obj)
+            % Get the directory path for the c3d file
+            [filedirectory, name, extension] = fileparts(obj.path);
+        end
+        function table = getTable_markers(obj)
+            table = obj.markers();
+        end
+        function table = getTable_forces(obj)
+            table = obj.forces();
+        end
+        function [markerStruct, forcesStruct] = getAsStruct(obj)
+            % Convert the OpenSim tables into Matlab Structures
+            markerStruct = osimTableToStruct(obj.markers);
+            forcesStruct = osimTableToStruct(obj.forces);
+            disp('Maker and force data returned as Matlab Structs')
+        end
+        function rotateData(obj,axis,value)
+            % Method for rotating marker and force data stored in OpenSim 
+            % tables
+            if ~ischar(axis) 
+               error('Axis must be either x,y or z')
+            end
+            if ~isnumeric(value)
+                error('value must be numeric (90, -90, 270, ...')
+            end
+            
+            % rotate the tables
+            markers_rot = obj.rotateTable(obj.markers, axis, value);
+            forces_rot = obj.rotateTable(obj.forces, axis, value);
+        
+            % set the internal tables to the new rotated ones
+            obj.setMarkers(markers_rot)
+            obj.setForces(forces_rot)
+        
+            disp('Marker and Force tables have been rotated')
+        end
+   end
+   
+   methods (Access = private, Hidden = true)
+        function setMarkers(obj, a)
+            % Private Method for setting the internal Marker table
+            obj.markers = a;
+        end
+        function setForces(obj, a)
+            % Private Method for setting the internal Force table
+            obj.forces = a;
+        end 
+        function table = rotateTable(obj, table, axisString, value) 
+            % Private Method for doing the table rotation operations
+            
+            import org.opensim.modeling.*
+            % set up the transform
+            if strcmp(axisString, 'x')
+                axis = CoordinateAxis(0);
+            elseif strcmp(axisString, 'y')
+                axis = CoordinateAxis(1);
+            elseif strcmp(axisString, 'z')
+                axis = CoordinateAxis(2);
+            else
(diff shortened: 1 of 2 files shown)


Bindings/Java/Matlab/examples/c3dExport.m
```diff
@@ -0,0 +1,46 @@
...
+markerTable = c3d.getTable_markers();
+forceTable = c3d.getTable_forces();
+% Get as Matlab Structures
+[markerStruct forceStruct] = c3d.getAsStruct();
+
+
+%% TODO write markers using a TRC class
+
+
+%% TODO Write a forces to file using a MOT class
```
chrisdembia: Consider moving the TODOs out of user-facing code and into GitHub issues.

Bindings/Java/Matlab/examples/c3dExport.m
```diff
@@ -0,0 +1,46 @@
+%% Example of using the Matlab-OpenSim class to read in data and do 
+% operations
+
+%% Choose a c3d file
+[filename, path] = uigetfile('*.c3d');
```
chrisdembia: Are there C3d files available in the Matlab examples dir in the installation? That is, when this GUI pops up, will it show c3d files in that directory?

Bindings/Java/Matlab/examples/c3dExport.m
```diff
@@ -0,0 +1,46 @@
...
+
+%% Rotate the data 
+c3d.rotateData('x',-90)
+
+%% Get the c3d in different forms
+% Get OpenSim tables
+markerTable = c3d.getTable_markers();
+forceTable = c3d.getTable_forces();
+% Get as Matlab Structures
+[markerStruct forceStruct] = c3d.getAsStruct();
```
chrisdembia: Consider changing method name to `getAsStructs()`, since it returns multiple structs.

Bindings/Java/Matlab/examples/c3dExport.m
```diff
@@ -0,0 +1,46 @@
...
+nForces = c3d.getNumForces();
+% Get the force data rate
+rForces = c3d.getRate_force();
+
+% Get Start and end time
+t0 = c3d.getStartTime();
+tn = c3d.getEndTime();
+
+%% Rotate the data 
+c3d.rotateData('x',-90)
```
chrisdembia: Just wondering, what is a common use case for rotating the data? I presume it usually involves 2 or 3 rotations of +/-90 degrees. If so, should users call this one method multiple times?

It would be helpful to add more documentation about `rotateData` so users know whether they want +90 or -90, for example. Perhaps spelling out a specific example with a lab coordinate system, etc. would help. I apologize if this is already in a different PR.

Bindings/Java/Matlab/Utilities/opensimC3D.m
```diff
@@ -0,0 +1,132 @@
...
+            % Get the capture rate used for the Marker Data
+            rate = str2double(obj.markers.getTableMetaDataAsString('DataRate'));
+        end
+        function rate = getRate_force(obj)
+            % Get the capture rate used for the Force Data
+            rate = str2double(obj.forces.getTableMetaDataAsString('DataRate'));
+        end
+        function n = getNumTrajectories(obj)
+            % Get the number of markers in the c3d file
+            n = obj.markers.getNumColumns;
```
chrisdembia: I think we should make sure to always use parentheses for functions, even if they take no arguments (`getNumColumns()`).

Bindings/Java/Matlab/Utilities/opensimC3D.m
```diff
@@ -0,0 +1,132 @@
+classdef opensimC3D < matlab.mixin.SetGet
+% opensimC3D C3D data into OpenSim Tables
+%   Utility  Class that uses OpenSim's C3DFileAdapter() class to produce
+%   OpenSim tables. Methods include Simbody rotation.
+    properties (Access = private)
+    path
+    markers
+    forces
```
chrisdembia: Indent the variable names?

Bindings/Java/Matlab/Utilities/opensimC3D.m
```diff
@@ -0,0 +1,132 @@
...
+                error('value must be numeric (90, -90, 270, ...')
+            end
+            
+            % rotate the tables
+            markers_rot = obj.rotateTable(obj.markers, axis, value);
+            forces_rot = obj.rotateTable(obj.forces, axis, value);
+        
+            % set the internal tables to the new rotated ones
+            obj.setMarkers(markers_rot)
+            obj.setForces(forces_rot)
```
chrisdembia: It seems likely to me that these calls are unnecessary (that `obj.rotateTable()` rotates the tables in place).

chrisdembia:
> additional methods

I don't think we need to add additional methods now. The class is already useful, and we can add methods in futuer PRs. However, one idea would be to add a method for translating the data (e.g., add 5mm to the x coordinates); is this something that commonly needs to be done?

Bindings/Java/Matlab/examples/c3dExport.m
jimmyDunne: I have never seen a c3d file where I haven't had to transform about X, it is just the mocap standard. I do agree that we could improve the inline help for this, but additional documentation may be hard since it is a Matlab script — you can't put in Doxygen so I guess the only other space would be on Confluence. 

I still feel strongly that we should ultimately move this feature (rotating Vec3 tables) to the OpenSim side. You should be able to rotate a table or send a (Vec3) table to a Rotation object to perform the operation.

jimmyDunne:
Translation of the markers and forces aren't typical but I do like the idea. Apart from visualization, do you have any ideas for when translation of the marker and force data would be useful? I am happy to add since the operation is straight forward, just not sure it would get used.

chrisdembia:
> Apart from visualization, do you have any ideas for when translation of the marker and force data would be useful?

No not really...I guess I thought it might be useful in the same way that rotating is useful, but I suppose the model just ends up wherever the marker data is. I think it's fine to leave this out for now.

Bindings/Java/Matlab/examples/c3dExport.m
chrisdembia: I think it's fine to add the documentation inside opensimC3D.m for now, next to the signature for this function. We could move the documentation around later, if necessary.

> I still feel strongly that we should ultimately move this feature (rotating Vec3 tables) to the OpenSim side. You should be able to rotate a table or send a (Vec3) table to a Rotation object to perform the operation.

:)

chrisdembia:
The changes look great. I see you renamed the file to `osimC3D` but the class inside is still `opensimC3D`. It seems like these should be consistent.

jimmyDunne:
Thanks, Chris.

Bindings/Java/Matlab/Utilities/osimC3D.m
```diff
@@ -0,0 +1,127 @@
...
+            % Get the directory path for the c3d file
+            [filedirectory, name, extension] = fileparts(obj.path);
+        end
+        function table = getTable_markers(obj)
+            table = obj.markers().clone();
+        end
+        function table = getTable_forces(obj)
+            table = obj.forces().clone();
+        end
+        function [markerStruct, forcesStruct] = getAsStructs(obj)
```
aseth1: Is this still necessary, if we add the ability to transform tables of vector data to the API? Or is this the primary to get data out for plotting, etc... in MATLAB?

Bindings/Java/Matlab/examples/c3dExport.m
aseth1: > I still feel strongly that we should ultimately move this feature (rotating Vec3 tables) to the OpenSim side. You should be able to rotate a table or send a (Vec3) table to a Rotation object to perform the operation.

Do you mean add a method like: `DataTable<Vec3>::transformColumns(const SimTK::Transform& X)` ?  Utilities could then form Transforms (e.g. Rotation about X-axis) to rotate or translate `Vec3` data uniformly. That should be straightforward to add. Please add a related issue for the feature.

jimmyDunne: >Do you mean add a method like: DataTable<Vec3>::transformColumns(const SimTK::Transform& X)
That sounds great, Ajay! It would offer a significant speed up and also make rotations a lot more simple for users and the GUI! I will start an issue

Bindings/Java/Matlab/Utilities/osimC3D.m
jimmyDunne: I added this method as a convenience for users. So no instead of this;
```
marker_table = osimC3D().getTable_markers();
marker_struct = osimTableToStruct(marker_table);
force_table = osimC3D().getTable_forces();
force_struct = osimTableToStruct(force_table);
```
you can just do this;
```
[marker_struct, forces_struct] = osimC3D.getAsStructs()
```
That was the only reason.

Bindings/Java/Matlab/examples/c3dExport.m
aseth1: :+1:

Bindings/Java/Matlab/Utilities/osimC3D.m
aseth1: Makes sense. Thanks for clarifying.

aseth1:
I am certain these changes are causing the debug failures on Travis. Merging.

Issue #1806, opened 2017-06-28
SWIG warning about redefining `clone()`


chrisdembia:
Oops...I introduced a SWIG warning in PR #1742:

```
/Users/travis/build/opensim-org/opensim-core/Bindings/common.i:199: Warning 302: %extend definition of 'clone'.
/Users/travis/build/opensim-org/opensim-core/Bindings/common.i:186: Warning 302: Identifier 'clone' redefined by %extend (ignored),
```

PR #1928, opened 2017-09-15
[swig] Avoid multiple definitions for DataTable_::clone().

chrisdembia:
Fixes issue #1806.

### Brief summary of changes

No longer define `DataTable_::clone()` multiple times.

### Testing I've completed

Added python tests and they pass.

We still need to check AppVeyor's output to make sure the warnings are gone.

### CHANGELOG.md (choose one)

- no need to update because...this is internal

### DIFF
--- a/Bindings/Python/tests/test_DataTable.py
+++ b/Bindings/Python/tests/test_DataTable.py
@@ -5,6 +5,27 @@ import os, unittest
 import opensim as osim
 
 class TestDataTable(unittest.TestCase):
+    def test_clone(self):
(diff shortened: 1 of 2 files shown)


chrisdembia:
Yes because if SWIG fixes the bug we should move back to the old definition, since it is more general.

aymanhab:
ok, thanks for the explanation @chrisdembia. Feel free to merge if you can verify the warnings have gone.

chrisdembia:
Thanks @aymanhab

Issue #1924, opened 2017-09-11
Model.formStateStorage doesn't handle inDegrees


aymanhab:
The method Model.formStateStorage assumes that the provided Storage object is in Radians, and provides no warning if that wasn't the case. Trying to use it from the GUI on a Storage built from .mot file produces wrong motion. Solution is to either document this behavior, or warn if inDegrees is true on input (with suggestion for how to fix).

jenhicks:
@aseth1 Will document and throw that the storage passed in should be in internal units (radians).

PR #1937, opened 2017-09-20
Make Model::formStateStorage throw if input is in degrees

aseth1:
Fixes issue #1924

### Brief summary of changes
Added an exception and comments to enforce and document that input to `Model::formStateStorage()` must be in internal units (i.e.  not in degrees).

### Testing I've completed
All tests pass locally, 

None. 

### CHANGELOG.md (choose one)
- no need to update because no change to the API


chrisdembia:
I'm not sure "internal" units is well-defined (for users).

chrisdembia:
What about saying more specifically that angles must be in radians and not degrees?

Issue #1854, opened 2017-08-03
Foward Tool cannot be run with PointKinematics analysis


jenhicks:
@tkuchida commented on [Sat Jul 08 2017](https://github.com/opensim-org/opensim-gui/issues/264)

**Steps to reproduce**
- open Models\Pendulum\double_pendulum.osim
- select *Forward Dynamics…* from the *Tools* menu
- in the *Analyses* tab, add a *PointKinematics* analysis
- click *Run*

**Expected behavior**
Simulation runs.

**Actual behavior**
The following dialog box appears:

![fwd](https://user-images.githubusercontent.com/4203505/27989648-1780bbe0-63f2-11e7-86ae-d87938e86051.png)

but no details are provided in the Messages window:
<pre>
Updating Model file from 30000 to latest format...
Loaded model double_pendulum from file C:\Users\tkuchida\GitHub - GUI 4.0 alpha03\opensim\Models\Pendulum\double_pendulum.osim
ModelVisualizer ignoring 'Unassigned'; only .vtp .stl and .obj files currently supported.
ModelVisualizer ignoring 'Unassigned'; only .vtp .stl and .obj files currently supported.
ModelVisualizer ignoring 'Unassigned'; only .vtp .stl and .obj files currently supported.
ModelVisualizer ignoring 'Unassigned'; only .vtp .stl and .obj files currently supported.
ModelVisualizer ignoring 'Unassigned'; only .vtp .stl and .obj files currently supported.
ModelVisualizer ignoring 'Unassigned'; only .vtp .stl and .obj files currently supported.
ModelVisualizer ignoring 'Unassigned'; only .vtp .stl and .obj files currently supported.
ModelVisualizer ignoring 'Unassigned'; only .vtp .stl and .obj files currently supported.
ModelVisualizer ignoring 'Unassigned'; only .vtp .stl and .obj files currently supported.
ModelVisualizer ignoring 'Unassigned'; only .vtp .stl and .obj files currently supported.
ModelVisualizer ignoring 'Unassigned'; only .vtp .stl and .obj files currently supported.
ModelVisualizer ignoring 'Unassigned'; only .vtp .stl and .obj files currently supported.
ModelVisualizer ignoring 'Unassigned'; only .vtp .stl and .obj files currently supported.
ModelVisualizer ignoring 'Unassigned'; only .vtp .stl and .obj files currently supported.
Running tool double_pendulum.
No external loads will be applied (external loads file not specified).
</pre>

[...]

PR #1942, opened 2017-09-26
 Update PointKinematics to use PhysicalFrame instead of Body

aseth1:
Fixes issue #1854

### Brief summary of changes
*ground* is an invalid `Body` name and could not be found in the `Model`'s `BodySet`, which gives rise to the ` ArrayPtrs.get(aName): No object with name ground` error reported in #1854.  Now use `getComponent<PhysicalFrame>()` to locate the frame to which we can attach a station for the `PointKinematics` analysis.

### Testing I've completed
- `PointKinematics` analysis runs with `ground` 
- using an illegal name it throws now with a more reasonable message:
```
Component 'arm26' could not find 'voodoo' of type PhysicalFrame. Make sure a component exists at this path and that it is of the correct type.
        In file G:\OpenSimGit\opensim-core\OpenSim/Common/Component.h:711
        In function 'getComponent'
```

### CHANGELOG.md (choose one)
- no need to update because there is not change to the API

### DIFF
--- a/OpenSim/Analyses/PointKinematics.cpp
+++ b/OpenSim/Analyses/PointKinematics.cpp
@@ -308,7 +308,7 @@ setModel(Model& aModel)
     Analysis::setModel(aModel);
 
     // Map name to index
-    _body = &aModel.updBodySet().get(_bodyName);
+    _body = &aModel.getComponent<PhysicalFrame>(_bodyName);
     if (aModel.updBodySet().contains(_relativeToBodyName))
         _relativeToBody = &aModel.updBodySet().get(_relativeToBodyName);
 
@@ -343,8 +343,7 @@ setBodyPoint(const std::string& aBody, const SimTK::Vec3& aPoint)
  *
  * @param aBody Body ID
  */
-void PointKinematics::
-setBody(Body* aBody)
+void PointKinematics::setBody(const PhysicalFrame* aBody)
 {
     // CHECK
     if (aBody==NULL) {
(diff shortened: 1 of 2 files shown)


aseth1:
@aymanhab changes are straightforward and should be a very quick review. Please let me know if you have additional comments.

aymanhab:
@aseth1 A couple minor comments/Questions:
- Would the PhysicalFrame be specified by name, fullPath or either (users now have to think about that too).
- Ideally the property comments and tags in XML would reflect the actual functionality (didn't find any comments and the element names still refer to body)
- The label in the result will still say "body" even though underneath we changed to "PhysicalFrame"
- This maybe outside the scope of this PR but any reason to keep pointers and not const references around?

aseth1:
> Would the PhysicalFrame be specified by name, fullPath or either (users now have to think about that too).

If the Body is in the `BodySet` or the components list of the `Model` just the name will suffice. Ground is also immediate subcomponent. If the user wants a `PhysicalFrame` located as a subcomponent somewhere else, then yes they have to specify its path, but I don't want that "advertised".

> Ideally the property comments and tags in XML would reflect the actual functionality (didn't find any comments and the element names still refer to body)

Ideally users will not use `PointKinematics` analysis anymore. There is no reason to "upgrade" the analysis since any `Point` can report its location, speed, and/or acceleration directly to a `Reporter`. The only reason for the fix is so that user's existing analyses still works and since `PhysicalFrames` did not exist I don't see any reason to make users aware of the change now.

> The label in the result will still say "body" even though underneath we changed to "PhysicalFrame"

Again, I have no intention of making users aware of implementation difference in `PointKinematics`. We don't want 4.0 users to start using it to get values out.

> This maybe outside the scope of this PR but any reason to keep pointers and not const references around?

It is not worth the time to cleanup code we want to deprecate, IMHO.

aymanhab:
Agree not to try to fix, as long as we expose an alternate mechanism to obtain outputs as we discussed since majority of users will continue to use Tools to access functionality.

aseth1:
Thanks @aymanhab and @chrisdembia!

Issue #1947, opened 2017-09-29
Visualizing a GeometryPath without PathPoints segfaults (crashing MATLAB)


aseth1:
Ran into this while testing the wrapping for `McKibbenActuator` and then trying to visualize it in hopes of running into issues related to #1835.

The script:

```MATLAB
import org.opensim.modeling.*;
m = Model();
m.setName('bob');

b = Body();
b.setMass(1.0)
b.setName('ball')
b.attachGeometry(Sphere(0.1))
m.addBody(b)

slider = SliderJoint('slider', m.getGround(), b);
m.addJoint(slider)

mc = McKibbenActuator();
% Commenting out one or both of the following two lines
% was causing a segafault
%mc.addNewPathPoint('mcpG', m.getGround(), Vec3());
%mc.addNewPathPoint('mcpB', b, Vec3());
mc.set_min_control(0.25);
mc.set_optimal_force(100);
m.addForce(mc);

m.setUseVisualizer(true);
s = m.initSystem();
slider.getCoordinate().setValue(s, 1.0);
m.realizeDynamics(s);

vc = m.getControls(s);
vc.set(0, 0.5);
m.setControls(s, vc);

m.realizeAcceleration(s);
m.getVisualizer().show(s);
```

Adding the PathPoints visualizes fine (in MATLAB).

aseth1:
closed by #1948

Issue #1835, opened 2017-07-19
MATLAB initSystem() with setUseVisualizer(true) causes MATLAB crash (simbody-visualizer crashes if OSX_DEPLOYMENT_VERSION > [user's OSX version])


chrisdembia:
This was experienced by one of the participants for the TGCS tutorial. I can provide more details later.

jimmyDunne:
> setUseVisualizer(true)

I have seen this behavior in my code as well. The visualizer instance won't close, and if you force close it, Matlab will crash soon after.

chrisdembia:
Ah it would be great if you could show met this when I get back.

tkuchida:
Possibly related: https://github.com/simbody/simbody/pull/554

aseth1:
@chrisdembia do you have more details to add?

apoorvar:
I'm using Matlab 2015b. My list of commands:
import org.opensim.modeling.*
m = Model();
m.setUseVisualizer(true);
m.initSystem()

At this point, I got a message saying the simbody-visualizer quit unexpectedly. My matlab crash logs are attached.

[matlabCrashLog.txt](https://github.com/opensim-org/opensim-core/files/1853869/matlabCrashLog.txt)
[matlab_crash_dump.95264-1.txt](https://github.com/opensim-org/opensim-core/files/1853870/matlab_crash_dump.95264-1.txt)

chrisdembia:
@jenhicks had the same exact issue.

It's not clear that the original issue from 2017 is the same as the one happening now. Both Jen and Apoorva were on their Macs (Jen: OSX 10.10, Matlab 2017; Apoorva: 10.11, Matlab 2015). I think Travis is building for 10.12. 

We thought the issue might be related to Matlab version, but James had no crash and is using Matlab 2015.

The crash may also be sporadic. Jen had been able to launch a blank visualizer without a crash.

chrisdembia:
@apoorvar reported the following in an email:

> I built the master branch, and ran configureOpenSim.m from that build. With that, the simple test of m.setVisualizer(true) followed m.initSystem() worked fine -- the visualizer opened with a white screen. I then ran the version of InteractiveHopper.m distributed with the GUI build from today, and everything seems to be working as it should. This seems consistent with Ajay's suspicion that the issue may be related to the simbody visualizer distributed with the GUI having been built on OSX 10.12 and me trying to use it on 10.11.

PR #1948, opened 2017-09-29
Fix invalid path segfaulting during visualization

aseth1:
Fixes issue #1947 (and one possibility for #1835)

A `GeomertyPath` expects at least two points when drawing but it is never checked. Way too late in `GeomertyPath::generateDecorations()` it attempts to get the first point in the path by index, that is not valid (no path points).

### Brief summary of changes
-  `GeomertyPath::extendFinalizeFromProperties()` now checks that we have at least two `PathPoint`s.
- removed calls to `finalizeFromProperties()` in muscle constructors.
- made reference to `PhysicalFrame` for adding path points, `const`.
- added path points to the muscle being tested in `testMuscles`.

### Testing I've completed
All tests passed locally.

Ran the script in #1947 and we now get the following error message (and no crash):
```
Error using testViz (line 21)
Java exception occurred:
java.lang.RuntimeException: Property 'PathPointSet' has an invalid value.
(details: A valid path requires at least two PathPoints.).
	In object 'geometrypath' of type GeometryPath.
	In file G:\OpenSimGit\opensim-core\OpenSim\Simulation\Model\GeometryPath.cpp:1190
	In function 'extendFinalizeFromProperties'

	at org.opensim.modeling.opensimSimulationJNI.Model_private_addForce(Native Method)

	at org.opensim.modeling.Model.private_addForce(Model.java:976)

	at org.opensim.modeling.Model.addForce(Model.java:100)
```

### CHANGELOG.md (choose one)
- no need to update because just a bug fix with not API change


aseth1:
@chrisdembia, CI failures due to invalid paths in scripting tests have been addressed.
@aymanhab, please let me know if you have any comments.

aseth1:
@aymanhab please let me know if this can go in. Thanks.

Bindings/Java/tests/TestBasics.java
```diff
@@ -32,12 +32,15 @@ public static void testBasics() {
 
     public static void testMuscleList() {
         Model model = new Model();
-
+        Ground ground = model.getGround();
         Thelen2003Muscle thelenMuscle =
```
aymanhab: Should that be a reference?

OpenSim/Actuators/Millard2012AccelerationMuscle.cpp
```diff
@@ -226,10 +226,6 @@ Millard2012AccelerationMuscle::Millard2012AccelerationMuscle()
 {    
     setNull();
     constructProperties();
-
-    // TODO: Remove this once MuscleFirstOrderActivationDynamicModel and
-    //       MuscleFixedWidthPennationModel have been made into properties.
-    finalizeFromProperties();
 }
```
aymanhab: Is this related somehow (the removal of finalizeFromProperties)?

OpenSim/Actuators/Test/testMuscles.cpp
```diff
@@ -124,23 +124,23 @@ int main()
     try { testRigidTendonMuscle();
         cout << "RigidTendonMuscle Test passed" << endl; }
     catch (const Exception& e)
-        { e.print(cerr); failures.push_back("testRigidTendonMuscle"); }
+        { e.print(cout); failures.push_back("testRigidTendonMuscle"); }
```
aymanhab: Is there a convention that we follow in all tests to use cout vs. cerr?

aseth1: I think tests should write to cout so the messages can be read directly in the console. I don't see the point of writing to cerr in test cases where the intention is to find (and see) errors.

OpenSim/Actuators/Millard2012AccelerationMuscle.cpp
aseth1: Yes.
1. Components should not call `finalizeFromProperties` in their constructors (except for Model, which we agreed that convenience was worth the price of some inconsistency). This should have been removed earlier.
2.  Keeping it in causes the (invalid) muscle (has no path points) to be finalized which throws an exception that there are no path points, which this PR added.

aymanhab:
Thanks for the explanation @aseth1 LGTM. I'd suggest mention in release notes to document this is new behavior.

aseth1:
>  I'd suggest mention in release notes to document this is new behavior.

I'll add something now to the CHANGELOG.

aseth1:
@aymanhab @chrisdembia are you OK with the CHANGELOG comment?

CHANGELOG.md
```diff
@@ -87,6 +87,22 @@ where fiber-velocity can be estimated from the state or assumed to be zero if th
...
+    Thelen2003Muscle* muscle = new Thelen2003Muscle("muscle", 1, 0.5, 0.5, 0);
+    model.addForce(muscle); // GeometryPath throws: "A valid path requires at least two PathPoints."
+    ```
+  - After (now required):
+    ```cpp
+    Model model;
+    Thelen2003Muscle* muscle = new Thelen2003Muscle("muscle", 1, 0.5, 0.5, 0);
+    // require at least two path points to have a valid muscle path
+    muscle->addNewPathPoint("p1", model.getGround(), Vec3(0.0,0.0,0.0));
+    muscle->addNewPathPoint("p2", model.getGround(), Vec3(1.0,0.0,0.0));
```
chrisdembia: It is unlikely that someone will want two path points on the same body if there's only one body. Perhaps: 

```
muscle->addNewPathPoint("p1", ...);
muscle->addNewPathPoint("p2", ...);
```

CHANGELOG.md
```diff
@@ -87,6 +87,22 @@ where fiber-velocity can be estimated from the state or assumed to be zero if th
   `preserveMassDist` arguments were swapped and the `preserveMassDist` argument
   is no longer optional. The default argument for `preserveMassDist` in OpenSim
   3.3 was `false`. (PR #1994)
+- A GeometryPath without PathPoints is considered invalid, since it does not represent a physical system. You must specify PathPoints to define a valid GeometryPath for a Muscle, Ligament, PathSpring, etc... that is added to a Model. (PR #1948)
```
chrisdembia: It would be good to keep the lines under 80 columns, to make it easier to read this file in all contexts (some editors may not do soft wraps).

chrisdembia:
LGTM.

PR #1949, opened 2017-09-29
Avoid unnecessary exception in DelimFileAdapter

chrisdembia:
### Brief summary of changes

Using a debugger with OpenSim, with a breakpoint on any thrown exceptions, is very helpful. However, when our code uses exceptions to perform routine non-error tasks, then debugging gets more difficult (the debugger stops at all these unimportant exceptions).

This PR avoids the use of an exception for simply checking if a key exists.

### Testing I've completed

- continuous integration.

### CHANGELOG.md (choose one)

- no need to update because...this is internal.


chrisdembia:
Thanks @tkuchida

chrisdembia:
Only the debug build is failing (because 3 tests are timing out). Can I merge this?

chrisdembia:
Thanks @aseth1

PR #1951, opened 2017-10-01
Fixed broken link in README for hopper example

tkuchida:
### Brief summary of changes
Replaced link.

Not sure of the best way to automatically generate a direct link to the highest-numbered folder on myosin.

### CHANGELOG.md (choose one)

- no need to update because... documentation only


chrisdembia:
I had imaged the myosin sourceforge project only being used by developers. For example, this project has public links for OpenSim binaries for the latest commit, and I don't really want users to find those. It was for this reason that I named the project "myosin" rather than "opensim," etc. Perhaps it's fine for users to know about myosin, though.

What about (a) using https://simtk.org/api_docs/opensim/api_docs/ (which will have the correct docs after we release), (b) pointing to the APIGuide in the doc folder of the repo, or (c) just removing the link from this README?

tkuchida:
(a) doesn't have documentation for the Component class so I don't think we should link there right now. (b) seems to be WIP and the deep links don't seem to work yet. Thus, I have proceeded with option (c).

PR #1955, opened 2017-10-03
Check Component name is valid

aseth1:
Fixes issue #1876 (pt. 1)

### Brief summary of changes
- use `ComponentPath`'s detection of invalid characters to test the validity of a Component's name during `finalizeFromProperties()`
- include whitespace as illegal characters 
- extend `Component::updateFromXMLNode` to remove whitespace from Components when loading old model files (pre 4.0 dev).

### Testing I've completed
- all tests pass locally
- added a test case to invoke `InvalidComponentName` exception.
- enable test that expects `Component` to throw when it has whitespace in its name.

### CHANGELOG.md (choose one)
- no need to update because this only adds error checking. 


OpenSim/Common/Component.cpp
```diff
@@ -1090,6 +1094,26 @@ void Component::setNextSubcomponentInSystem(const Component& sub) const
...
+            if (node.hasAttribute("name")) {
+                auto name = node.getRequiredAttribute("name").getValue();
+                if (name.find_first_of("\n\t ") < std::string::npos) {
+                    std::cout << getConcreteClassName() << " name '" << name
+                        << "' contains whitespace. ";
+                    name.erase(
+                        std::remove_if(name.begin(), name.end(), ::isspace),
+                        name.end() );
+                    node.setAttributeValue("name", name);
+                    std::cout << "It is was renamed '" << name << "'." << std::endl;
```
tkuchida: "It is was" -> "It was"

OpenSim/Common/Component.cpp
```diff
@@ -1090,6 +1094,26 @@ void Component::setNextSubcomponentInSystem(const Component& sub) const
...
+                }
+            }
+            else { // All Components should have a name. If none, assign one here.
```
tkuchida: I'm guessing your unique-name-ifier code will catch duplicate names and append suffixes as necessary. Perhaps state that here in a comment (otherwise, it looks like something would end up breaking downstream if there are two or more nameless components of the same concrete class).

OpenSim/Common/ComponentPath.cpp
```diff
@@ -28,7 +28,7 @@ using namespace std;
 
 // Set static member variables.
 const char ComponentPath::separator = '/';
-const std::string ComponentPath::invalidChars = "\\/*+";
+const std::string ComponentPath::invalidChars = "\\/*+ \t\n";
```
tkuchida: Is `:` still used anywhere for aliases?

aseth1: It is. I would like the name checker to also invalidate for `:,;` even though it is OK for an Output path to use `:` to denote the alias. I have not resolved precisely how to handle that yet.

OpenSim/Common/Path.h
```diff
@@ -97,6 +97,10 @@ class OSIMCOMMON_API Path {
     /// Pure virtual function that returns a string of invalid characters.
     virtual const std::string getInvalidChars() const = 0;
 
+    /// Return true if pathElement does not contain any chars from the list
+    /// of getInvalidChars()
+    bool isLegalPathElement(const std::string& pathElement) const;
```
chrisdembia: It's good to have this method.

OpenSim/Common/Test/testComponentInterface.cpp
```diff
@@ -1889,14 +1895,9 @@ void testSingleValueInputConnecteeSerialization() {
...
-        // for now. This is because the way we determine if multiple
-        // connectee names were specified is by looking for spaces, but old
-        // models used have spaces in their names.
-        // SimTK_TEST_MUST_THROW_EXC(
-        //     TheWorld world(modelFileNameMultipleValues),
-        //     OpenSim::Exception);
-        TheWorld world(modelFileNameMultipleValues);
+        SimTK_TEST_MUST_THROW_EXC(
+             TheWorld world(modelFileNameMultipleValues),
+             OpenSim::Exception);
```
chrisdembia: I'm happy that this TODO can be resolved now, but I'm not sure which exception is being caught; it can't be the desired one. The code that generates the warning is in [ComponentSocket.h](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/ComponentSocket.h#L248), and you can see a commented-out exception there. That's the exception that is supposed to be caught here. Could you also resolve the TODO in ComponentSocket.h and ensure that exception is the one that's caught here?

aseth1: @chrisdembia I beefed `ComponentSocket::checkConnecteeNameProperty() to check for invalid characters using `ComponentPath::isLegalPathElement()` and I have it throw.  It is indeed that exception that is being caught here, as intended.

chrisdembia: Okay great.

OpenSim/Common/ComponentPath.cpp
tkuchida: > I have not resolved precisely how to handle that yet.

Planned for this PR, a separate PR, noted in an Issue, not important, …?

OpenSim/Common/Component.cpp
```diff
@@ -1103,6 +1107,29 @@ void Component::setNextSubcomponentInSystem(const Component& sub) const
 void Component::updateFromXMLNode(SimTK::Xml::Element& node, int versionNumber)
 {
     if (versionNumber < XMLDocument::getLatestVersion()) {
+        if (versionNumber < 30500) {
+            // In 3.3 and earlier, spaces in names were tolerated. Spaces are
+            // no longer acceptable in Component names.
```
tkuchida: Perhaps include in CHANGELOG (this change could affect a user's downstream code and goes beyond "no need to update because this only adds error checking" IMHO).

OpenSim/Common/Test/testComponentInterface.cpp
```diff
@@ -918,6 +918,12 @@ void testListSockets() {
 
 void testComponentPathNames()
 {
+    Foo foo;
+    foo.setName("LegWithConstrainedFoot/foot");
```
tkuchida: Might want to check a name like `"a\b"` as well to make sure the character is properly escaped and captured. Could additionally add a check for `"a:b"` that doesn't currently throw, with a TODO for later.

OpenSim/Common/Component.cpp
aseth1: Good point, I'll add the stricter naming rules to the CHANGELOG. Old models will still load but components will be renamed for you!

tkuchida: > Old models will still load but components will be renamed for you!

I know, but I thought the point of CHANGELOG (in theory, at least) was for it to describe all the changes a user would need to make to his/her pipeline if upgrading from 3.3 to 4.0. A user will need to update code if it refers to Components by name and if any of those names are now disallowed.

OpenSim/Common/ComponentPath.cpp
aseth1: I created issue #2007 to punt 🏉 to a follow-up PR.

OpenSim/Common/Test/testComponentInterface.cpp
aseth1: I will have to include when addressing #2007, so will leave it for now if OK with you.

OpenSim/Common/ComponentPath.cpp
tkuchida: It's good! Fine to tackle that later.

OpenSim/Common/Component.cpp
aseth1: The following was added to the CHANGELOG:

> Component naming is more strictly enforced and names with spaces are no longer accepted. Spaces are only allowable as separators for Output or Channel names that satisfy a list Input. (PR #1955)

I think it implies that you have to update your Component names wherever you use them. Would you prefer I state that explicitly?

OpenSim/Common/Test/testComponentInterface.cpp
tkuchida: Seems like checking "`\`" would belong in this PR since all `invalidChars = "\\/*+ \t\n"` should now be caught, but ok to include later. (I expected to see a test for each of these invalid characters since it wouldn't cost many lines to test each one, e.g., using a lambda.) I've added a reminder to #2007.

OpenSim/Common/Component.cpp
chrisdembia: Spaces are allowed in any list property of type string, as delimiters between the list's entries. The value of list property XML elements is somewhat separate from what is allowed in the `name="..."` annotation.

tkuchida: > I think it implies that you have to update your Component names wherever you use them. Would you prefer I state that explicitly?

In general, I would briefly describe what was changed and then tell the user what he/she needs to do. Perhaps add something like "OpenSim will automatically rename any Components whose names contain the following characters: …"?

aseth1:
Thanks @tkuchida and @chrisdembia

PR #1956, opened 2017-10-04
[travis] Ignore more tests in debug

chrisdembia:
The following tests have been timing out in Travis' Debug build, so I think we should ignore them for now so that the green checkmark is meaningful.

- testContactGeometry
- testForward
- testCMCArm26_Millard

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/1956)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -158,7 +158,7 @@ before_install:
   # Initialize environment variable.
   - export TESTS_TO_EXCLUDE="unmatched" # This is just for the regex.
   ## If we are building in debug, there are some tests to ignore.
-  - if [ "$BTYPE" = "Debug" ]; then export TESTS_TO_EXCLUDE="$TESTS_TO_EXCLUDE|testOptimizationExample|testCMCGait10dof18musc|testCMCWithControlConstraintsRunningModel"; fi
+  - if [ "$BTYPE" = "Debug" ]; then export TESTS_TO_EXCLUDE="$TESTS_TO_EXCLUDE|testOptimizationExample|testCMCGait10dof18musc|testCMCWithControlConstraintsRunningModel|testContactGeometry|testForward|testCMCArm26_Millard"; fi
   
   ## Set compiler flags.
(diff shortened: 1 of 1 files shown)


aseth1:
I am worried that recent changes may have slowed down our tests. Was testForward an issue before?

If it is travis that is allocating us less resources (and slower performance) then I am OK with this change. Can you comment on the AppVeyor debug test times? Has there been a systematic rise in the last few weeks/months?  If not then I think it would be safe to say it is an issue with travis and we should make this change.

chrisdembia:
That's a good point. 

We do not run debug tests on AppVeyor. I think we would have to try running the tests locally on the latest branch and an older branch.

aseth1:
> We do not run debug tests on AppVeyor. I think we would have to try running the tests locally on the latest branch and an older branch.

Do you mind flagging the two commits (or branches) so a few of us can test and present timings. Since `testForward` seems to be the latest problem,lets focus on its timing in debug on various platforms. For example, I can test VS2015, you on Mac (clang), and perhaps @carmichaelong on Ubuntu (gcc)?

chrisdembia:
Sounds good. My choice of which older commit to use would be arbitrary. Perhaps we can use [this commit](https://github.com/opensim-org/opensim-core/commit/ded15c765b0f43f32642cad4bd063065b64b0457) from Aug 31: ded15c765b0f43f32642cad4bd063065b64b0457

carmichaelong:
I tested with gcc 4.9.1 on a CentOS machine and didn't see big differences between the commits for longer tests. Here's a sample of the tests:

Current commit:
- testContactGeometry: 341.12s
- testOptimizationExampleRuns: 668.88s
- testForward: 298.80s
- testCMCGait10dof18musc: 927.07s
- testCMCArm26_Millard: 252.23s

Older commit:
- testContactGeometry: 332.70s
- testOptimizationExampleRuns: 669.09s
- testForward: 285.00s
- testCMCGait10dof18musc: 921.78s
- testCMCArm26_Millard: 251.01s

aymanhab:
travis is retired

PR #1959, opened 2017-10-05
Wrap Rotation_::setRotationFromTwoAxes().

### LINKED ISSUE #1958: Expose Rotation methods in Matlab
Biomechanical frames are computed from vectors derived from experimental marker locations. I would like to use these vectors to define a `Rotation()` object in Matlab. The method that looks most likely to help me is `setRotationFromTwoAxes()`, however this seems to not be wrapped. The `Unitvector()` object also is needed and doesn't seem to be callable in Matlab.

chrisdembia:
Fixes #1958.

### Brief summary of changes

Wrapped Rotation_::setRotationFromTwoAxes().

### Testing I've completed

Added a MATLAB test case and ran it locally.

### CHANGELOG.md (choose one)

- no need to update because...this is minor

### DIFF
--- a/Bindings/Java/Matlab/tests/testSimbody.m
+++ b/Bindings/Java/Matlab/tests/testSimbody.m
@@ -43,3 +43,9 @@ for i = 0:residual.size()-1
     assert(abs(residual.get(i)) < 1e-10);
 end
 
+% Test Rotation.
+% --------------
+rot = Rotation()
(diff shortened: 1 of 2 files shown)


jimmyDunne:
Thanks, Chris. I built and tested this branch. Using a simple set of vectors I was able to test and got the correct answers. It looks great! I can merge if you are happy?

PR #1965, opened 2017-10-06
[contributing] Allow only 1 reviewer for changes to Travis/AppVeyor scripts

chrisdembia:
In the last few days I've been making lots of changes to the Travis/AppVeyor scripts. It would be great if these PRs could move a little faster. In my opinion, one review would be sufficient for these PRs. Let me know if you agree.

@aseth1 @tkuchida @aymanhab @jenhicks

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,7 +51,7 @@ When you are ready to make a PR, please adhere to the following guidelines:
 
 4. Typo fixes can be merged by any member of the Development (Dev) Team.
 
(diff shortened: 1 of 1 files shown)


aymanhab:
Totally agree @chrisdembia These PRs only affect the build system and so have same scope as CMake file changes (or even less) and only way to test is to have CI run them. 

On a related note, it would be great to have documentation somewhere of all the tools/services we use and our credentials so as we get closer to release or if @chrisdembia is unavailable others can pitch in and troubleshoot.

chrisdembia:
Great. Good idea @aymanhab.

chrisdembia:
Thanks @aseth1

PR #1968, opened 2017-10-09
Scaletool binding fixes

aymanhab:
Fixes issue # Scale GUI crash due to upstream memory management mixup (BodyScale is double deleted)

### Brief summary of changes
API calls that change ownership are indicated in SWIG interface files

### Testing I've completed
These are bug fixes regardless of GUI code, testing will happen when opensim-core and gui are integrated.

### Looking for feedback on.
Where to add corresponding code for python wrapping, then we can add a python test case

### CHANGELOG.md (choose one)

- no need to update because it's a bug fix

### DIFF
--- a/Bindings/Java/swig/java_actuators.i
+++ b/Bindings/Java/swig/java_actuators.i
@@ -60,11 +60,12 @@ SWIG_JAVABODY_PROXY(public, public, SWIGTYPE)
 %rename OpenSim::ModelScaler::addScale private_addScale;
 
 %typemap(javacode) OpenSim::ModelScaler %{
-    addScale(Scale scale){
+    public void addScale(Scale scale){
         scale.markAdopted();
         private_addScale(scale);
     }
-    addMeasurement(Measurement meas){
+
+    public void addMeasurement(Measurement meas){
         meas.markAdopted();
         private_addMeasurement(meas);
     }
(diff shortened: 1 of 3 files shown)


chrisdembia:
Are you able to test with the GUI before merging this PR? You could also add a java test case (perhaps to TestBasics.java):

```
ModelScaler ms;
Scale s;
ms.addScale(s);
```

> Where to add corresponding code for python wrapping, then we can add a python test case

Something like the following should be added [here](https://github.com/opensim-org/opensim-core/blob/master/Bindings/Python/swig/python_tools.i#L62):

```
%pythonappend OpenSim::ModelScaler::addScale %{
    aScale._markAdopted()
%}

%pythonappend OpenSim::ModelScaler::addMeasurement %{
    aMeasurement._markAdopted()
%}
```

aymanhab:
Thanks for the quick turnaround @chrisdembia I added the python side of the fix. and included a trivial Java test per your recommendation, thanks for the pointers. 
While I built the GUI with the fix and it didn't crash on Windows, it didn't crash before the fix either for me! and there's a GUI code change to go along.  So I can't guarantee that fixing this will fix all ailments of scaling but there is definitely a bug that was fixed by this PR.

aymanhab:
Thanks much @chrisdembia Will do when tests are done.

PR #1971, opened 2017-10-11
Update NOTICE

### LINKED ISSUE #1960: Update NOTICE file to reflect GUI change to Apache license
If you use the OpenSim GUI, you must acknowledge the OpenSim project where
appropriate, by citing the following paper. If you only use the OpenSim API, you
are not required to acknowledge us, but we would be extremely grateful if you
did so anyway.

chrisdembia:
Fixes #1960.

@aseth1 please commit any revisions directly to this branch.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/1971)

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -33,13 +33,24 @@ our impact, which is essential for securing funding to improve the software and
 provide support to our users (you). We would be extremely grateful if you
 acknowledge us by citing the following papers.
 
+Delp SL, Anderson FC, Arnold AS, Loan P, Habib A, John CT, Guendelman E, Thelen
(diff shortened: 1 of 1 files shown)


chrisdembia:
It would be great to make progress on this.

I'm not sure I agree with adding the moment arm paper here. I agree that people should read it but I think Confluence is a better place for sharing such resources (the moment arm paper is already listed here: https://simtk-confluence.stanford.edu/display/OpenSim/Dynamics+Theory+and+Publications). If we list the moment arm paper, we should probably list the V&V paper.

PR #1987, opened 2017-10-25
Exception message shows file name, not full path.

### LINKED ISSUE #1984: Exception messages should not show an absolute path from the developer's machine.
Right now, OpenSim exception messages show an absolute path name to a source file, and the absolute path might contain something like `/Users/chris/...`. It's not ideal to reveal this path on a user's computer. Simbody exception messages show only the file name, not the full path. We should do the same.

chrisdembia:
Fixes #1984 

### Brief summary of changes

- Adds a small function to pull off the filename from a path (I wanted to use @carmichaelong's Path class but it's abstract).
- Uses the function to only show the filename (not the full path) in the exception message.

**After**:
```
Socket 'childFoo' not connected.
     In object 'Bar' of type Bar.
     In file /Users/chris/repos/opensim/16/opensim-core/OpenSim/Common/Component.h:856
     In function 'getConnectee'
```

**Before**:
```
Socket 'childFoo' not connected.
     In Object 'Bar' of type Bar.
     Thrown at Component.h:856 in getConnectee().
```

### Testing I've completed

- Ran testComponentInterface and inspected the new exception message format.

- Do we think this format is an improvement?

### CHANGELOG.md (choose one)

- no need to update because...minor change

### DIFF
--- a/OpenSim/Common/Exception.cpp
+++ b/OpenSim/Common/Exception.cpp
@@ -66,11 +66,21 @@ exception()
 //#endif
 }
 
+namespace {
+/// Grab the last element of a file path.
+std::string findFileName(const std::string& filepath) {
+    // Based on a similar function from Simbody.
+    std::string::size_type pos = filepath.find_last_of("/\\");
+    if (pos + 1 >= filepath.size()) pos = 0;
+    return filepath.substr(pos + 1);
+}
+} // namespace
+
 Exception::Exception(const std::string& file,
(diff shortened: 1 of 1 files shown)


chrisdembia:
Haha thanks @tkuchida

chrisdembia:
This is a quick review for anyone who has a preference for clean error messages.

chrisdembia:
Thanks @aseth1

Issue #1977, opened 2017-10-12
Create an analysis tool that reports outputs


jenhicks:

aseth1:
closed by #1991

Issue #1954, opened 2017-10-02
Separate Invocation of AnalysisSet.begin()  from Manager::integrate()


aseth1:
In #1689 @mitkof6 identified an issue of the `JointReactions` `Analysis` overwriting existing analysis results (`Storage`) when invoking `Manager::integrate()` repeatedly (in a loop) where the final time is advanced iteratively. We concluded that both the scenarios: 1) where the results are reset (e.g. in evaluating an objective function during an optimization) and 2) where the results are appended to in the case of performing an integration as series of smaller intervals, are useful. 

In #1909 we introduce an explicit `Manager::initialize(state)` and could invoke `AnalysisSet.begin()` (via `Manager::initializeStorageAndAnalyses` ) instead of `integrate`. Then repeated calls to `integrate` would cause Analyses to append to existing results and not reset. In the case of optimizations, the requirement to call `initialize` first ensures that the results are reset every objective function evaluation.

PR #1991, opened 2017-10-30
Output reporting analysis

aseth1:
Fixes issue #1977
This PR doesn't create a separate `Tool`, but simply creates an `Analysis` that wraps the `TableReporter` Component. The `OutputReporter` Analysis creates and adds the `TableReporter`(s) according to the types of the `Outputs` that are listed in the property of the `Analysis`. 

### Brief summary of changes
Added `OutputReporter` as an `Analysis` that so that `Outputs` can be  reported on via the `ForwardTool` and `AnalzeTool` in the GUI. Only Outputs or type `double`, `SimTK::Vec3` and `SimTK::SpatialVec` are supported. This is intended to be a short term solution until the GUI adds support  for wiring Inputs and Outputs and for adding `Reporters` so we can deprecate Analyses.

### Testing I've completed
- added `OutputReporter` with various outputs of different types.
- verified that underlying tables are written to file.
- verified that values reported, written to file and reloaded correspond to model values

### CHANGELOG.md (choose one)
- no need to update because adds an Analysis,  but does affect existing Analyses.

### DIFF
--- a/OpenSim/Analyses/OutputReporter.cpp
+++ b/OpenSim/Analyses/OutputReporter.cpp
@@ -39,7 +39,7 @@ int OutputReporter::begin(const SimTK::State& s)
 {
     if (!proceed()) return 0;
 
-    if (_pvtModel == nullptr)
+    if (_model == nullptr)
         return -1;
 
     _pvtModel.reset(_model->clone());
@@ -51,6 +51,8 @@ int OutputReporter::begin(const SimTK::State& s)
     _pvtModel->addComponent(_tableReporterDouble.get());
     _pvtModel->addComponent(_tableReporterVec3.get());
     _pvtModel->addComponent(_tableReporterSpatialVec.get());
+
+    _minimumStageToRealize = SimTK::Stage::Time;
 
     for (int i = 0; i < getProperty_output_paths().size(); ++i) {
         auto outpath = ComponentPath(get_output_paths(i));
@@ -84,10 +86,14 @@ int OutputReporter::begin(const SimTK::State& s)
                 << out.getTypeName() << " is not supported by OutputReporter."
                 << " Consider adding a TableReporter_ to the Model." << endl;
         }
+
+        if (out.getDependsOnStage() > _minimumStageToRealize) {
+            _minimumStageToRealize = out.getDependsOnStage();
+        }
     }
 
     _pvtModel->initSystem();
-    _pvtModel->realizeReport(s);
+    report(s);
 
     return 0;
 }
@@ -98,21 +104,22 @@ int OutputReporter::step(const SimTK::State& s, int stepNumber)
         return 0;
     }
 
-    _pvtModel->realizeReport(s);
-
+    report(s);
     return 0;
 }
 
 int OutputReporter::end(const SimTK::State& s)
 {
     if (!proceed()) return 0;
-    _pvtModel->realizeReport(s);
+    report(s);
     return 0;
 }
 
 int OutputReporter::printResults(const std::string& baseName,
     const std::string& dir,  double dT, const std::string& extension)
 {
+    cout << "OutputReporter.printResults: " << endl;
+
     if (!getOn()) {
         printf("OutputReporter.printResults: Off- not printing.\n");
         return 0;
@@ -121,23 +128,33 @@ int OutputReporter::printResults(const std::string& baseName,
     OPENSIM_THROW_IF_FRMOBJ(IO::Lowercase(extension) != ".sto",
         Exception, "Only writing results to '.sto' format is supported.");
 
+    std::string prefix = dir.empty() ? "" : dir + "/";
+
     auto& tableD = _tableReporterDouble->getTable();
     if (tableD.getNumColumns()) {
         STOFileAdapter_<double>::write(tableD,
-            dir + "/"+ baseName + "_Outputs" + extension);
+            prefix + baseName + "_Outputs" + extension);
     }
 
     auto& tableV3 = _tableReporterVec3->getTable();
     if (tableV3.getNumColumns()) {
         STOFileAdapter_<SimTK::Vec3>::write(tableV3,
-            dir + "/" + baseName + "_OutputsVec3" + extension);
+            prefix + baseName + "_OutputsVec3" + extension);
     }
 
     auto& tableSV = _tableReporterSpatialVec->getTable();
     if (tableSV.getNumColumns()) {
         STOFileAdapter_<SimTK::SpatialVec>::write(tableSV,
-            dir + "/" + baseName + "_OutputsSpatialVec" + extension);
+            prefix + baseName + "_OutputsSpatialVec" + extension);
     }
 
     return 0;
+}
+
+void OutputReporter::report(const SimTK::State& s)
+{
+    _pvtModel->getMultibodySystem().realize(s, _minimumStageToRealize);
+    _tableReporterDouble->report(s);
+    _tableReporterVec3->report(s);
+    _tableReporterSpatialVec->report(s);
 }
--- a/OpenSim/Analyses/OutputReporter.cpp
+++ b/OpenSim/Analyses/OutputReporter.cpp
@@ -39,29 +39,32 @@ int OutputReporter::begin(SimTK::State& s)
 {
     if (!proceed()) return(0);
 
-    _model = _model->clone();
+    _pvtModel.reset(_model->clone());
 
     _tableReporterDouble = new TableReporter_<double>();
     _tableReporterVec3 = new TableReporter_<SimTK::Vec3>();
     _tableReporterSpatialVec = new TableReporter_<SimTK::SpatialVec>();
 
-    _model->addComponent(_tableReporterDouble.get());
-    _model->addComponent(_tableReporterVec3.get());
-    _model->addComponent(_tableReporterSpatialVec.get());
+    _pvtModel->addComponent(_tableReporterDouble.get());
+    _pvtModel->addComponent(_tableReporterVec3.get());
+    _pvtModel->addComponent(_tableReporterSpatialVec.get());
 
     for (int i = 0; i < getProperty_output_paths().size(); ++i) {
         auto outpath = ComponentPath(get_output_paths(i));
         // assume that Outputs are for the Model
-        string compName = _model->getAbsolutePathString();
+        string compName = _pvtModel->getAbsolutePathString();
         // if a a ComponentPath is specified then use the parent path
         if (outpath.getNumPathLevels()) {
             compName = outpath.getParentPathString();
         }
         auto outputName = outpath.getComponentName();
 
-        const Component* comp = _model;
+        // In keeping with assumption of Outputs for the Model (Component)
+        const Component* comp = _pvtModel.get();
+
+        // If the ComponentPath has a Component name get that Component
         if (!compName.empty())
-            comp = &_model->getComponent(compName);
+            comp = &_pvtModel->getComponent(compName);
 
         auto& out = comp->getOutput(outputName);
         if (out.getTypeName() == "double") {
@@ -80,7 +83,7 @@ int OutputReporter::begin(SimTK::State& s)
         }
     }
 
-    _model->initSystem();
+    _pvtModel->initSystem();
 
     return 0;
 }
@@ -90,10 +93,10 @@ int OutputReporter::step(const SimTK::State& s, int stepNumber)
     if (!proceed(stepNumber)) {
         return 0;
     }
-    if (_model == nullptr)
+    if (_pvtModel == nullptr)
         return -1;
 
-    _model->realizeReport(s);
+    _pvtModel->realizeReport(s);
 
     return 0;
         
--- a/OpenSim/Analyses/OutputReporter.cpp
+++ b/OpenSim/Analyses/OutputReporter.cpp
@@ -35,7 +35,7 @@ using namespace std;
 //=============================================================================
 // ANALYSIS
 //=============================================================================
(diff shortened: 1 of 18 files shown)


tkuchida:
Cool! :sunglasses:

OpenSim/Analyses/Test/testOutputReporter.cpp
```diff
@@ -0,0 +1,300 @@
...
+        ActivationFiberLengthMuscle *aflMuscle2
+            = dynamic_cast<ActivationFiberLengthMuscle *>(muscle);
+        if (aflMuscle2){
+            // Define the default states for the muscle 
+            //that has activation and fiber-length states
+            aflMuscle2->setDefaultActivation(act0);
+            aflMuscle2->setDefaultFiberLength(aflMuscle2
+                ->getOptimalFiberLength());
+        }
+    }
```
chrisdembia: You're only testing one muscle, so this if-statement seems unnecessary.

OpenSim/Analyses/Test/testOutputReporter.cpp
```diff
@@ -0,0 +1,300 @@
...
+    auto tableD = STOFileAdapter::read("testOutputReporter.sto");
+    auto tableV3 = STOFileAdapter_<SimTK::Vec3>::
+        read("testOutputReporterVec3.sto");
+    auto tableSV = STOFileAdapter_<SimTK::SpatialVec>::
+        read("testOutputReporterSpatialVec.sto");
+
+    const SimTK::Real& val_ke = tableD.getRowAtIndex(tableD.getNumRows() - 1)[0];
+    const Vec3& val_omega = tableV3.getRowAtIndex(tableV3.getNumRows() - 1)[1];
+    const SimTK::SpatialVec& val_jrf = 
+        tableSV.getRowAtIndex(tableSV.getNumRows() - 1)[1];
```
chrisdembia: No need to do it in this PR, but perhaps we should have a convenience function `getLastRow()`.

OpenSim/Analyses/OutputReporter.h
```diff
@@ -0,0 +1,102 @@
...
+//=============================================================================
+namespace OpenSim { 
+
+/**
+ * The OutputReporter Analysis is a wrapper for a TableReporter. It generates
+ * and writes TimeSeriesTables of Output values according to the names listed
+ * as properties of the Analysis. OutputReporter enables the AnalyzeTool to
+ * report on Outputs via the Analysis interface but backed by a TableReporter.
+ * The OutputReporter currently only supports Outputs of type, double, Vec3
+ * and SpatialVec.
```
chrisdembia: I think we should either rename this class to `OutputAnalysis` (which would be in line with the naming of `MuscleAnalysis`), or we should add a comment here saying "@note This class does *not* derive from Reporter.". I *strongly* prefer the first option. I am fairly certain users will be confused if we name this class  OutputReporter.

chrisdembia: Please mention exactly what files are written, and how it depends on the types of the outputs the user lists.

OpenSim/Analyses/OutputReporter.cpp
```diff
@@ -0,0 +1,129 @@
...
+        string compName = _model->getAbsolutePathString();
+        // if a a ComponentPath is specified then use the parent path
+        if (outpath.getNumPathLevels()) {
+            compName = outpath.getParentPathString();
+        }
+        auto outputName = outpath.getComponentName();
+
+        const Component* comp = _model;
+        if (!compName.empty())
+            comp = &_model->getComponent(compName);
```
chrisdembia: This is a little confusing; it would be helpful to add a comment here. I had to read this a few times to understand that you're handling the case that the output is one of the model's outputs (e.g., kinetic energy).

OpenSim/Analyses/OutputReporter.cpp
```diff
@@ -0,0 +1,129 @@
...
+using namespace std;
+
+//=============================================================================
+// ANALYSIS
+//=============================================================================
+int OutputReporter::begin(SimTK::State& s)
+{
+    if (!proceed()) return(0);
+
+    _model = _model->clone();
```
chrisdembia: Assigning back to the same variable is confusing to me. It seems that you're leaking this model. I think it would be better to use a `SimTK::ResetOnCopy<std::unique_ptr<Model>>` or a `SimTK::ClonePtr` member variable.

OpenSim/Analyses/OutputReporter.cpp
```diff
@@ -0,0 +1,129 @@
...
+    if (tableV3.getNumColumns()) {
+        STOFileAdapter_<SimTK::Vec3>::write(tableV3,
+            dir + baseName + "Vec3" + extension);
+    }
+
+    auto& tableSV = _tableReporterSpatialVec->getTable();
+    if (tableSV.getNumColumns()) {
+        STOFileAdapter_<SimTK::SpatialVec>::write(tableSV,
+            dir + baseName + "SpatialVec" + extension);
+    }
```
chrisdembia: I don't think you should assume the user requested a STO file. What if `extension` is `.csv`? Instead, you could rely on the file adapter registry by using `FileAdapter::writeFile` (I think that's the name of the function).

The documentation for this class should also specify the file format used, and whether or not it depends on the extension provided by the user.

Alternatively, you could give an error message if `extension = ".sto"`.

aseth1: I think the extension option should be removed for writing the results of  any Analysis. There are no supported cases of reporting non `.sto` files from an `Analysis`. I can add that.

OpenSim/Analyses/OutputReporter.h
aseth1: Should we then rename `ForceReporter` and `ProbeReporter`? I can rename it to `OutputAnalysis` but there is no *analysis* being performed - we just report outputs.  I agree it can be confusing if you are trying to understand the class hierarchy. But, as a user of a `Tool`, I think `OutputReporter` is clearer and corresponds to the Force and Probe reporters.

chrisdembia: > Should we then rename ForceReporter and ProbeReporter?

I wouldn't mind this, other than not wanting to rename existing classes. Alternatively, if you leave this class named as OutputReporter, shouldn't you rename MuscleAnalysis to MuscleReporter?

If a user has a CMC simulation and they want to know joint reactions during that simulation, I don't think they would be confused by the language "analyze your CMC simulation to obtain joint reactions." I'm not sure it matters to the user that the calculation of the joint reactions actually happens in the Model, not in the Analysis class.

Perhaps there's an alternative name that can please everyone. OutputAggregator, OutputCollector.

OpenSim/Analyses/Test/testOutputReporter.cpp
aseth1: Missed this in my gutting of testProbes. Gone!

OpenSim/Analyses/OutputReporter.cpp
aseth1: 👍

aseth1:
Thanks @chrisdembia for the quick and thorough review!

>@mitkof6 raised an issue about when data tables should be reset if an analysis is performed multiple times. Which of the two options did you choose? It would be good to explicitly document this choice.

I  reset at the beginning of any analysis sequence (when `begin()` is invoked). I believe we agreed in #1954 that we could hold uniformly prevent unnecessary resetting by only calling `begin()` when the integrator is initialized.  I believe these changes are consistent with that plan.

aseth1:
> It would be good to test the analysis when it's used without integration, as is done when using the analyses through the AnalyzeTool.

I will add a case to testAnalyzeTool.

OpenSim/Analyses/OutputReporter.h
```diff
@@ -0,0 +1,109 @@
...
+    int begin(SimTK::State& s) override;
+    int step(const SimTK::State& s, int setNumber) override;
+    int printResults(const std::string& baseName,
+        const std::string& dir = "",
+        double dT = -1.0,
+        const std::string& extension = ".sto") override;
+
+private:
+    // Keep a reference to the underlying TableReporter(s) for this Analysis
+    SimTK::ReferencePtr< TableReporter_<double> > _tableReporterDouble;
```
aymanhab: With paths potentially Absolute, relative or relative to top model, which ones are we using/assuming, can't see that except in doxygen comment. If users will be checkign the syntax using XMLBrowser in GUI the syntax for specification should be very clear.

OpenSim/Analyses/OutputReporter.cpp
```diff
@@ -0,0 +1,136 @@
...
+}
+
+int OutputReporter::step(const SimTK::State& s, int stepNumber)
+{
+    if (!proceed(stepNumber)) {
+        return 0;
+    }
+    if (_pvtModel == nullptr)
+        return -1;
+
```
aymanhab: Is it possible for _pvtModel to be nullptr at all?

aymanhab: Would there be an issue with realizing the state to Report using the cloned model having unexpected side effects (e.g. if model has no dynamics/masses, will that throw or should Outputs realize lazily on their own to the needed stage?)

chrisdembia:
Alright, we can use OutputReporter, but I get a nickel from each of you whenever a user gets confused!

aymanhab:
Sure, open a paypal account ;)

OpenSim/Analyses/OutputReporter.h
aseth1: Relative to the Model or absolute work. I added more comments to the property comment that appears in XML.

OpenSim/Analyses/OutputReporter.cpp
aseth1: > Is it possible for _pvtModel to be nullptr at all?

There is nothing stopping anyone from calling Analysis::step() AFAIK. Other analyses had similar check for `_model == NULL` so I figured to leave it in. It does make more sense for the check to be on `_model` at begin(), so I'll move it there.

aseth1: > Would there be an issue with realizing the state to Report using the cloned model having unexpected side effects (e.g. if model has no dynamics/masses, will that throw or should Outputs realize lazily on their own to the needed stage?)

Good point. Yes, definitely there is an issue with realizing to `Report` when `Dynamics` and `Acceleration` are invalid. I am not going to access `Output`s here myself because I want to leverage the `TableReporter` to do the actual reporting. I might be able to get the `TableReporter` to report *on demand*, however. I'll look into it. Thanks.

aseth1:
@chrisdembia @aymanhab I believe all issues raised have been addressed. Please let me know. Thanks!

aymanhab:
Thanks much for the fixes @aseth1 Can you explain at a high level what needed to be changed in the Manager/Integrator communication to support this functionality, to help me understand the scope? Thanks

aymanhab:
There are also build failures that seem related.

aseth1:
> Can you explain at a high level what needed to be changed in the Manager/Integrator communication to support this functionality, to help me understand the scope?

In short, we were incorrectly using the condition `(status != SimTK::Integrator::EndOfSimulation)` to decide when to *report* (that is call `step` on Analyses and write states to Storage).  `TimeStepper::stepTo()` returns its status which includes things like `StartOfContinuousInterval` which causes the first state to be reported twice if `Manager::initialize()` calls `begin()` on Analyses, which it should. Otherwise we get a mismatch between outputs written with the `AnalyzeTool` and a forward simulation using the `Manager` regarding the first row. This bug was detected using the `TableReporter` because it will not allow duplicate times to be appended to the table. Writing only when `TimeHasAdvanced`  (actual forward integration step) or `ReachedScheduledEvent` (e.g. and interpolated step to periodically update controls in CMC) outputs unique time-steps and fixes the issue.

aymanhab:
LGTM, thanks @aseth1 Feel free to merge if other feedback has been addressed. Do you think we need to document this change in release notes?

OpenSim/Simulation/Manager/Manager.cpp
```diff
@@ -789,13 +804,21 @@ void Manager::initialize(const SimTK::State& s)
...
         analysisSet.end(s);
     }
+    if (_writeToStorage) {
+        SimTK::Vector stateValues = _model->getStateVariableValues(s);
+        StateVector vec;
+        vec.setStates(s.getTime(), stateValues);
+        getStateStorage().append(vec);
+        if (_model->isControlled())
+            _controllerSet->storeControls(s, getStateStorage().getSize());
+    }
```
chrisdembia: This duplication is not ideal. Perhaps at least place a comment saying that this is duplicated, so that maintainers don't accidentally only update one location?

aseth1: I agree. Eliminated that duplication.

chrisdembia:
@aymanhab's question about the CHANGELOG remains.

aseth1:
> Do you think we need to document this change in release notes?

Do you mean the integrator changes? I don't think so because the interface has not changed. I consider it just a bug fix, since there was a chance of recording the same state twice (at least initially) that has now been fixed.

chrisdembia:
@aymanhab may be referring to the new Analysis feature.

aseth1:
@aymanhab @chrisdembia I added `OutputReporter` to the CHANGELOG.md.
 I am not sure why Appveyor is still testing even though I had [ci skip] in the commit message. I don't think you need to wait to merge if you are OK with documentation update.

aymanhab:
LGTM, merging per earlier test passing.

Issue #1993, opened 2017-10-31
The "standard" model against which ligament scaling is tested contains correct path points but incorrect `resting_length`s


tkuchida:
I provide a detailed description of the issue below because the resolution involves changing a "standard" test model.

This issue concerns the following files in the [Applications/Scale/test](https://github.com/opensim-org/opensim-core/tree/master/Applications/Scale/test) directory:
- [testScale.cpp](https://github.com/opensim-org/opensim-core/blob/master/Applications/Scale/test/testScale.cpp)
- [toyLigamentModel.osim](https://github.com/opensim-org/opensim-core/blob/master/Applications/Scale/test/toyLigamentModel.osim) (the original model)
- [std_toyLigamentModelScaled.osim](https://github.com/opensim-org/opensim-core/blob/master/Applications/Scale/test/std_toyLigamentModelScaled.osim) (the "standard" model)
- [toyLigamentModel_Setup_Scale.xml](https://github.com/opensim-org/opensim-core/blob/master/Applications/Scale/test/toyLigamentModel_Setup_Scale.xml) (the ScaleTool setup file)

In [testScale.cpp, line 244](https://github.com/opensim-org/opensim-core/blob/master/Applications/Scale/test/testScale.cpp#L244), the `scaleModelWithLigament()` method uses the ScaleTool (with the toyLigamentModel_Setup_Scale.xml setup file) to scale the "block" and "block1" bodies in toyLigamentModel.osim by `(2, 2, 2)`. The "standard" model (std_toyLigamentModelScaled.osim) contains the correct GeometryPath points, but the `resting_length`s are incorrect. Here are the values of the `resting_length`s in the original and "standard" models:

**Table 1**

| Ligament | `resting_length` in original model | `resting_length` in "standard" model
| --- | --- | ---
| "lateralLigament" | 0.14 | 0.14
| "MedialLigament" | 0.14 | 0.14
| "internalLigament" | 0.04 | 0.06
| "crossLigament" | 0.147 | 0.289196876423888

The Ligament's [postScale() method](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/Ligament.cpp#L224) indicates that the `resting_length` should be multiplied by `scaleFactor = {path length after scaling} / {path length before scaling}`, so the table above suggests that "lateralLigament" has not changed length; however, this is not the case.

I opened the original model and "standard" models, and used the following script to read the lengths of the ligaments in the default pose:
```
[...]

Issue #1990, opened 2017-10-26
WrapObject derives from ModelComponent but lists Component in its DECLARE_ABSTRACT_OBJECT macro


tkuchida:
See [WrapObject.h#L46](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Wrap/WrapObject.h#L46):
```cpp
class OSIMSIMULATION_API WrapObject : public ModelComponent {
OpenSim_DECLARE_ABSTRACT_OBJECT(WrapObject, Component);
```
Presumably, `Component` should be `ModelComponent` on the second line. If `Component` is correct, there should be a comment here to explain why WrapObjects are unique.

tkuchida:
@aseth1 I'll change the second line to ModelComponent for now so I can use `Super`. Please let me know if "Component" was necessary for some reason.

aseth1:
@tkuchida thanks for finding. Yes, please go ahead and fix. It definitely was not intentional on my part!

Issue #1973, opened 2017-10-11
Intermediate PhysicalOffsetFrames will cause scaling failure for 5 Components


tkuchida:
Several Components point to a "parent" Frame, which was previously a Body but can now be a PhysicalOffsetFrame (and will almost always be a PhysicalOffsetFrame for Joints). Consequently, when searching through a ScaleSet for scale factors (by name), `getParentFrame().getName()` must now be `getParentFrame().findBaseFrame().getName()`. AFAIK this issue affects 5 Components:
- [GeometryPath](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/GeometryPath.cpp#L797)s whose ParentFrame is not a Body
- (Almost) all [EllipsoidJoint](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/SimbodyEngine/EllipsoidJoint.cpp#L122)s (Joints are typically created using the 7-argument convenience constructor, which automatically creates PhysicalOffsetFrames)
- (Almost) all [CustomJoint](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp#L151)s (as above)
- Markers (via [MarkerSet](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/MarkerSet.cpp#L132)) that are mounted on PhysicalOffsetFrames
- [CoordinateCouplerConstraint](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/SimbodyEngine/CoordinateCouplerConstraint.cpp#L260)s connected to PhysicalOffsetFrames

### Tasks
- [x] Replace `getParentFrame()` with `getParentFrame().findBaseFrame()` in the 5 places linked above.

Improve coverage of testScale, which currently only checks the locations of markers mounted directly to bodies. Possible improvements:
- [x] mount some markers to PhysicalOffsetFrames
- [x] also check points on GeometryPaths
- [x] test with a model containing a GeometryPath whose ParentFrame is not a Body
- [x] test with a model containing an EllipsoidJoint, CustomJoint, and CoordinateCouplerConstraint

tkuchida:
Should be fixed by PR #1994. Add tests to testScale to confirm.

Issue #1970, opened 2017-10-11
"Assigned an invalid SimTK::MobilizedBodyIndex" Exception on initSystem() if a PhysicalOffsetFrame is owned by a foreign Joint


tkuchida:
### Background
Working on improving coverage of testScale. Plan to include a check in `Model::scale()` to throw/warn if the model is in an unsupported format (see #1923).

### Issue
The API permits arbitrary ownership trees, but a low-level Exception is thrown on `initSystem()` if a PhysicalOffsetFrame is owned by a Joint to which it does not connect. Unclear whether there is a bug somewhere or whether the user must be prohibited from building such a model.

### Minimal example
``` cpp
using namespace OpenSim;
using namespace SimTK;
const Transform tfY = Transform(Vec3(0,1,0));

// Create a model consisting of three pendula pinned to ground, where each
// PinJoint's child PhysicalOffsetFrame (POF) is added to the model with a
// different ownership structure.
Model model;
model.setName("threedulum");

// Create bodies and add to model.
OpenSim::Body* pend1 = new OpenSim::Body("pend1", 1., Vec3(0), Inertia(0));
OpenSim::Body* pend2 = new OpenSim::Body("pend2", 1., Vec3(0), Inertia(0));
OpenSim::Body* pend3 = new OpenSim::Body("pend3", 1., Vec3(0), Inertia(0));
model.addBody(pend1);
model.addBody(pend2);
model.addBody(pend3);

// Case 1: Use PinJoint's convenience constructor to create the child POF
//         automatically. The POF will be stored in PinJoint's "frames" list
//         property.
PinJoint* pin1 = new PinJoint("pin1", model.getGround(), Vec3(0), Vec3(0),
                                      *pend1, Vec3(0,1,0), Vec3(0));
model.addJoint(pin1);

// Case 2: Create the child POF manually and store it in PinJoint's "frames"
//         list property. Should be functionally equivalent to Case 1.
PinJoint* pin2 = new PinJoint();
PhysicalOffsetFrame* pof2 = new PhysicalOffsetFrame("pof2", *pend2, tfY);
pin2->setName("pin2");
pin2->connectSocket_parent_frame(model.getGround());
pin2->connectSocket_child_frame(*pof2);
model.addJoint(pin2);
pin2->updProperty_frames().adoptAndAppendValue(pof2);

// Case 3: Create the child POF manually (as in Case 2), but store it in a
//         *different* PinJoint's "frames" list property.
[...]

PR #1994, opened 2017-10-31
Update scaling to support 4.0-style models

tkuchida:
### ~TODO~ TODONE:
- [x] Rename virtual methods on ModelComponent to `extendPreScale()`, `extendScale()`, and `extendPostScale()`, which will be called by (non-virtual) `preScale()`, etc.
- [x] Avoid collision between `ModelComponent::scale()` and `Model::scale()`.
- [x] Replace `ScaleSet::getIndexBySegmentName()` with `const Vec3 ModelComponent::getScaleFactors(const ScaleSet&, const Frame&) const`, per Chris's review comment.
- [x] Update CHANGELOG.md.

Fixes issues #1993, #1990, #1973 (partially), #1923, #1913, and #896.

### Brief summary of changes
The existing scaling infrastructure wasn't updated to support 4.0-style models, leading to several issues (e.g., not all PhysicalOffsetFrames were being scaled). Rather than have GeometryPaths scale the Points they own, Joints scale the Frames they own, etc., each ModelComponent is now responsible for scaling itself. Any subclass of ModelComponent can override the virtual `extendPreScale()`, `extendScale()`, and/or `extendPostScale()` methods. There are lots of changes; many fall into one of the following two categories:
- Pushed scaling operations as high up the class hierarchy as possible (e.g., Stations are scaled rather than Markers [note that Stations weren't being scaled at all]).
- `scale()` is now called directly on all ModelComponents, which (1) allows removal of `scale()` from BodySet, ConstraintSet, etc. (which just called `scale()` on all bodies, constraints, etc. in the Set); and (2) avoids skipping any ModelComponents (e.g., because Joints no longer decide whether to scale the Frames they own).

**Notes:**
- I suggest starting by reviewing `Model::scale()`.
- See #1993 for detailed discussion of changes to std_toyLigamentModelScaled.osim.
- Added some tests to testScale.cpp; more will follow (in a subsequent PR) to complete #1973. Cases 3(a) and 3(b) won't work until #1970 is fixed.

### Testing I've completed
Added tests to testScale.cpp.

~I believe [this error on Travis](https://travis-ci.org/opensim-org/opensim-core/jobs/295229942#L4709) is occurring because `ModelComponent::scale()` is virtual and unfortunately Model is itself a ModelComponent. `Model::scale()` has a different signature than `ModelComponent::scale()` and is not overriding the virtual method. This issue must be addressed before merging, but should not be difficult once a path has been chosen. An obvious option is to rename `scale()` on either Model or ModelComponent. Of the two, I would prefer to rename the method on Model because I think users are more likely to call `scale()` on ModelComponent (e.g., for scaling bodies during model construction).~ (Addressed by removing one of the defaults from the existing `Model::scale()` method.)

### CHANGELOG.md (choose one)
~Not sure whether it's necessary to note any of these changes in CHANGELOG.md. Pre-4.0 models should scale the same as before (with the possible exception of `resting_length`s of ligaments). I wouldn't expect users to be calling methods like `BodySet::scale()` directly, and the signature of the higher-level `Model::scale()` method hasn't changed (yet). It might be necessary to note a bug fix for scaling ligaments (see #1993)—thoughts?~
- [x] Note that `SimbodyEngine::scale()` has been removed.
- [x] Note the change in the `Model::scale()` argument list.
- [x] Note the bug fix in Ligament's `resting_length` property.
- [x] Summarize the new interface for implementing `extend{PreScale|Scale|PostScale}()` methods.


tkuchida:
Updating based on discussion with @chrisdembia…

tkuchida:
Updating based on discussion with @chrisdembia…

Renaming virtual methods on ModelComponent to `extendPreScaleSelf()`, `extendScaleSelf()`, and `extendPostScaleSelf()`, which will be called by `preScaleSelf()`, etc. This rename will avoid the collision between `ModelComponent::scale()` and `Model::scale()` (causing the CI failures), and will use the existing convention for virtual methods.

chrisdembia:
> Not sure whether it's necessary to note any of these changes in CHANGELOG.md. Pre-4.0 models should scale the same as before (with the possible exception of resting_lengths of ligaments). I wouldn't expect users to be calling methods like BodySet::scale() directly, and the signature of the higher-level Model::scale() method hasn't changed (yet). It might be necessary to note a bug fix for scaling ligaments (see #1993)—thoughts?

I agree with mentioning the fix for ligaments. You could consider mentioning the addition of a general and flexible mechanism for any (custom) model component to be scaled. My understanding is that, before this PR, users could not make their own custom model components scalable.

Applications/Scale/test/testScale.cpp
```diff
@@ -322,3 +328,136 @@ bool compareStdScaleToComputed(const ScaleSet& std, const ScaleSet& comp) {
...
+        testScaling(model, s);
+    }
+
+    // Case 2: Create the child POF manually and store it in
+    //         (a) PinJoint's "frames" list property (i==0)
+    //         (b) PinJoint's "components" list property (i==1)
+    //         (c) Model's "components" list property (i==2)
+    for (int i=0; i<3; ++i)
+    {
+        cout << "- case 2(" << std::string("abc").substr(i,1) << ")" << endl;
```
chrisdembia: nice

OpenSim/Simulation/Model/ModelComponent.h
```diff
@@ -108,6 +109,38 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(ModelComponent, Component);
...
+        by any subclass of ModelComponent, but all implementations must begin
+        with a call to `Super::scale()` to execute the parent class methods
+        before the child class method. The base class implementation in
+        ModelComponent does nothing.
+        @see preScale()
+        @see postScale() */
+    virtual void scale(const SimTK::State& s, const ScaleSet& scaleSet) {};
+
+    /** Perform any computations that must occur after the Model has been
+        scaled. This method is virtual and may be implemented by any subclass of
```
chrisdembia: "after the Model has been scaled" is a little vague to me, since "postScale" really is still scaling the model. Consider rephrasing as "after ModelComponent::scale() has been invoked on all components of the model."

OpenSim/Simulation/Model/Frame.cpp
```diff
@@ -181,6 +182,31 @@ void Frame::attachGeometry(OpenSim::Geometry* geom)
...
 }
 
+void Frame::scale(const SimTK::Vec3& scaleFactors)
+{
+    for (int i = 0; i < getProperty_attached_geometry().size(); ++i) {
+        Geometry& geo = upd_attached_geometry(i);
+        geo.upd_scale_factors() = geo.get_scale_factors()
+                                  .elementwiseMultiply(scaleFactors);
+    }
+}
```
chrisdembia: Ah interesting. Geometry is a Component, not a ModelComponent, so it doesn't have its own scale() function. What happens to Geometry that is not `attached_geometry`? It seems to not get scaled. If that's the case, it would be great to document this somewhere.

OpenSim/Simulation/Model/Frame.cpp
```diff
@@ -181,6 +182,31 @@ void Frame::attachGeometry(OpenSim::Geometry* geom)
...
+    if (getProperty_attached_geometry().size() == 0)
+        return;
+
+    // Get scale factors for base frame (if an entry for the base frame exists).
+    const int idx = scaleSet.getIndexBySegmentName(findBaseFrame().getName());
+    if (idx < 0)
+        return;
+    const Vec3& scaleFactors = scaleSet[idx].getScaleFactors();
+
+    scale(scaleFactors);
```
chrisdembia: Consider renaming this function to `scaleAttachedGeometry()`. As-is, I'm confusing it with the more general "scale()" function.

OpenSim/Simulation/Model/Frame.h
```diff
@@ -329,6 +329,10 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Frame, ModelComponent);
         transform used to position the Geometry is that of this Frame. */
     void attachGeometry(OpenSim::Geometry* geom);
 
+    /** Scale any Geometry attached to this Frame. */
+    void scale(const SimTK::Vec3& scaleFactors);
```
chrisdembia: See above; renaming would also make this function self-documenting.

OpenSim/Simulation/Model/ModelComponent.h
```diff
@@ -108,6 +109,38 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(ModelComponent, Component);
...
+        current body position before scaling so that an owning Muscle can use
+        this information to update the properties of the muscle after scaling.
+        This method is virtual and may be implemented by any subclass of
+        ModelComponent, but all implementations must begin with a call to
+        `Super::preScale()` to execute the parent class methods before the child
+        class method. The base class implementation in ModelComponent does
+        nothing.
+        @see scale()
+        @see postScale() */
+    virtual void preScale(const SimTK::State& s, const ScaleSet& scaleSet) {};
```
chrisdembia: Does `preScale()` actually require a `scaleSet`? If not, we could consider removing it. I won't hold up the PR if we decide to leave this in, though.

OpenSim/Simulation/SimbodyEngine/Body.cpp
```diff
@@ -179,43 +180,39 @@ void Body::setInertia(const SimTK::Inertia& inertia)
...
+//==============================================================================
+void Body::scale(const SimTK::Vec3& scaleFactors, bool scaleMass)
 {
-    Super::scale(aScaleFactors);
-    for(int i=0; i<3; i++) {
-        upd_mass_center()[i] *= aScaleFactors[i];
-    }
+    Super::scale(scaleFactors);
+    upd_mass_center() = get_mass_center().elementwiseMultiply(scaleFactors);
+    scaleInertialProperties(scaleFactors, scaleMass);
```
chrisdembia: Does this mean that inertial properties get scaled multiple times when `Model::scale()` is called (unnecessarily)? I know that the masses are readjusted if preserveMass is true, but that's not what I'm talking about. If so, this should at least be documented.

OpenSim/Simulation/Model/Frame.cpp
tkuchida: > What happens to Geometry that is not `attached_geometry`?

Do you mean, for example, adding a Mesh to a Frame's `components` list property? Correct: that Mesh wouldn't get scaled (nor would meshes added to any other Component's `components` list). This PR does not change that behavior. There are a few strategies we could take: search for and scale Geometry residing in the `components` list, disallow Geometry in the `components` list, add documentation, address in a subsequent PR, etc. This issue isn't directly aligned with the purpose of this PR, but I can address it here if it isn't too involved. Perhaps @aseth1 has a preference.

tkuchida:
> Unrelated to this PR: the scale tool should validate the segment names for all `Scale`s to make sure they are the names of bodies, and not the names of PhysicalOffsetFrames, etc., which can't have scale factors.

Created Issue #1995.

OpenSim/Simulation/Model/Frame.cpp
tkuchida: Done. Also, this 1-argument `Frame::scale()` method would be overloading the virtual `scale()` on ModelComponent, which smells like the same problem I encountered with `Model::scale()`.

OpenSim/Simulation/Model/Frame.h
tkuchida: Good point. Done.

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -1428,87 +1428,47 @@ void Model::removeController(Controller *aController)
...
-    //    position and then scale the tendon and fiber length
-    //    properties.
+    // Call preScale() on each ModelComponent owned by the model to store
+    // GeometryPath lengths (and perform any other necessary computations).
+    for (ModelComponent& comp : updComponentList<ModelComponent>())
+        comp.preScale(s, scaleSet);
 
+    // Scale the model.
+    if (!updSimbodyEngine().scale(s, scaleSet, finalMass, preserveMassDist))
+        return false;
```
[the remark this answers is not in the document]
tkuchida: I don't know why the workload was distributed between Model and SimbodyEngine—perhaps for historical reasons? I'm happy to update this here if there's consensus about what to do. @aseth1

OpenSim/Simulation/Model/ModelComponent.h
tkuchida: I think a ModelComponent developer could, in theory, want to calculate something in `preScale()` or `postScale()` that requires knowing scale factors. The original `preScale()` method took the same arguments.

OpenSim/Simulation/Model/ModelComponent.h
tkuchida: Done. Also updated the analogous comment for `ModelComponent::preScale()`, and changed the comment "Scale the model." to "Scale the rest of the model." in `Model::scale()`.

OpenSim/Simulation/SimbodyEngine/Body.cpp
tkuchida: The `void Body::scale(const SimTK::Vec3& scaleFactors, bool scaleMass)` method is *not* called during `Model::scale()`. This version of `Body::scale()` is (sometimes) used during model construction, e.g., [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Test/testForces.cpp#L197). Seems like a reasonable enough use case to want to leave the method around.

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp
```diff
@@ -798,65 +798,51 @@ void SimbodyEngine::convertQuaternionsToDirectionCosines(double aQ1, double aQ2,
...
-    // When bodies are scaled, the properties of the model are changed.
-    // The general rule is that you MUST recreate and initialize the system 
-    // when properties of the model change. We must do that here or
-    // we will be querying a stale system (e.g. wrong body properties!).
+    // Call scale() on each ModelComponent owned by the model. Each
+    // ModelComponent is responsible for scaling itself. All scaling operations
+    // are performed here except scaling inertial properties of bodies, which is
+    // done below.
+    for (ModelComponent& comp : _model->updComponentList<ModelComponent>())
+        comp.scale(s, scaleSet);
```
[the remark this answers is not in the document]
tkuchida: Good question. I guess it makes more sense to me to use the big roller first and then paint around the trim. Practically, I think someone could implement a Device ModelComponent that computes inertia based on a bunch of other ModelComponents (motors, cuffs, etc.)—in which case you would want to compute their locations first. I guess such a calculation could also go in `Device::postScale()` (or these smaller pieces could just be bodies themselves). Mostly it's about the paint thing.

OpenSim/Simulation/Model/Frame.cpp
aymanhab: How about objects like ContactGeometry and WrapObjects, what happens to their Geometry?

tkuchida: > How about objects like ContactGeometry and WrapObjects, what happens to their Geometry?

Same thing as before: WrapObjects get scaled (please see the changes to Wrap{Cylinder|Ellipsoid|Object|Sphere|Torus}.{h|cpp}) and [ContactGeometry throws](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/ContactGeometry.cpp#L113). The purpose of this PR was not to fix every aspect of scaling for every component: I can't manage that in a single PR. If it is now important for contact geometry to be able to scale, I can work on that once this PR is merged and I've finished addressing #1973.

chrisdembia: I don't think this PR should try to fix scaling for previously-unscaled components. But it might be a good opportunity to quickly and simply document for users which components are not scaled, especially if this list is fresh in our minds right now.

OpenSim/Simulation/Model/ModelComponent.h
chrisdembia: Okay; leave it in.

OpenSim/Simulation/SimbodyEngine/Body.cpp
chrisdembia: Thanks for clarifying.

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp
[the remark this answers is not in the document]
chrisdembia: Sounds good.

OpenSim/Simulation/SimbodyEngine/EllipsoidJoint.cpp
```diff
@@ -99,40 +99,22 @@ void EllipsoidJoint::setEllipsoidRadii(const Vec3& radii)
...
-    }
+    // Get scale factors for base frame (if an entry for the base frame exists).
+    const int idx = scaleSet.getIndexBySegmentName(getParentFrame()
+                                                   .findBaseFrame().getName());
+    if (idx < 0)
+        return;
+    const Vec3& scaleFactors = scaleSet[idx].getScaleFactors();
+
+    //TODO: Need to scale transforms appropriately, given an arbitrary axis.
+    upd_radii_x_y_z() = get_radii_x_y_z().elementwiseMultiply(scaleFactors);
```
chrisdembia: If you want to get really fancy you could do `upd_radii_x_y_z().elementwiseMultiplyInPlace(scaleFactors);`

OpenSim/Simulation/SimbodyEngine/Joint.cpp
```diff
@@ -259,52 +258,6 @@ bool Joint::isCoordinateUsed(const Coordinate& aCoordinate) const
...
-            = SimTK_DYNAMIC_CAST_DEBUG<PhysicalOffsetFrame&>(upd_frames(found));
-        offset.scale(parentFactors);
-    }
-    found = getProperty_frames().findIndex(getChildFrame());
-    if (found >= 0) {
-        PhysicalOffsetFrame& offset
-            = SimTK_DYNAMIC_CAST_DEBUG<PhysicalOffsetFrame&>(upd_frames(found));
-        offset.scale(childFactors);
-    }
-}
```
chrisdembia: Woohoo!

aseth1:
> Rename virtual methods on ModelComponent to extendPreScaleSelf(), extendScaleSelf(), and extendPostScaleSelf(), which will be called by (non-virtual) preScaleSelf(), etc. This rename will avoid the collision between ModelComponent::scale() and Model::scale() (causing CI failures), and will use the existing "extend" convention for virtual methods.

I wouldn't worry about renaming at this moment. The error you linked to is also related to building the Java wrapping.  Having a non-virtual `scale` method on `Model` shouldn't be an issue if it has a different signature.

OpenSim/Simulation/Model/ModelComponent.h
```diff
@@ -108,6 +110,45 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(ModelComponent, Component);
...
+
+    /** Scale the ModelComponent. This method is virtual and may be implemented
+        by any subclass of ModelComponent, but all implementations must begin
+        with a call to `Super::scale()` to execute the parent class methods
+        before the child class method. The base class implementation in
+        ModelComponent does nothing.
+        @see preScale()
+        @see postScale() */
+    virtual void scale(const SimTK::State& s, const ScaleSet& scaleSet) {};
```
aseth1: Is it your plan that is will no longer be virtual and `extendScale()` will be virtual? Just double checking.

OpenSim/Simulation/Model/ModelComponent.cpp
```diff
@@ -127,4 +127,16 @@ void ModelComponent::updateFromXMLNode(SimTK::Xml::Element& aNode,
     Super::updateFromXMLNode(aNode, versionNumber);
 }
 
+const SimTK::Vec3 ModelComponent::
+getScaleFactors(const ScaleSet& scaleSet, const Frame& frame) const
```
aseth1: :+1:  I like this because it hides the implementation of how you get scale factors given a `Frame`. __In the future__, we may consider passing `ModelComponent` instead of `Frame` since we are moving towards the `ModelComponent` knowing how to scale itself, so it should be able to access scale factors to do that.

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -1428,87 +1428,47 @@ void Model::removeController(Controller *aController)
...
- *
- * @param aScaleSet the set of XYZ scale factors for the bodies
- * @param aFinalMass the mass that the scaled model should have
- * @param aPreserveMassDist whether or not the masses of the
- *        individual bodies should be scaled with the body scale factors.
- * @return Whether or not scaling was successful.
- */
-bool Model::scale(SimTK::State& s, const ScaleSet& aScaleSet, double aFinalMass, bool aPreserveMassDist)
+//------------------------------------------------------------------------------
+bool Model::scale(SimTK::State& s, const ScaleSet& scaleSet, double finalMass,
```
aseth1: The compile issue may be due to the fact that `finalMass` and `preserveMassDist` have default arguments so the default signature is the same as the virtual `ModelComponent::scale`.

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp
```diff
@@ -798,65 +798,51 @@ void SimbodyEngine::convertQuaternionsToDirectionCosines(double aQ1, double aQ2,
...
- * Scale the dynamics engine
- *
- * @param aScaleSet the set of XYZ scale factors for the bodies
- * @param aFinalMass the mass that the scaled model should have
- * @param aPreserveMassDist whether or not the masses of the
- *        individual bodies should be scaled with the body scale factors.
- * @return Whether or not scaling was successful.
- */
-bool SimbodyEngine::scale(SimTK::State& s, const ScaleSet& aScaleSet, double aFinalMass, bool aPreserveMassDist)
+bool SimbodyEngine::scale(SimTK::State& s, const ScaleSet& scaleSet,
```
aseth1: Up to you @tkuchida (and maybe it is in your plan for phase II) but I think the this implementation could be moved into `Model` since the added indirection to the `SimbodyEngine` provides no advantages at all.

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp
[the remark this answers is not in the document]
aseth1: :+1:

OpenSim/Simulation/Model/ModelComponent.cpp
tkuchida: That sounds good. We might also consider using something more general than ScaleSet so "pre-scale lengths" and other scaling-related cache can be stored there instead of in private member variables like `GeometryPath::_preScaleLength`. Opened #1996 to track.

tkuchida: (We wouldn't need to *pass in* a reference to the ModelComponent since this is a method on the ModelComponent class. But, yes, each ModelComponent could implement its own version of `getScaleFactors()`).

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp
tkuchida: Done. Also added note to CHANGELOG.md indicating that `SimbodyEngine::scale()` has been removed.

OpenSim/Simulation/Model/ModelComponent.h
tkuchida: Yes, that's the plan: remove default arguments from `Model::scale()`; rename current virtual methods to `extendPreScale()`, `extendScale()`, and `extendPostScale()`; and create non-virtual `preScale()`, etc. methods on ModelComponent, per discussion.

OpenSim/Simulation/Model/Frame.cpp
tkuchida: Added comments to `Frame::scale()` and ContactGeometry class.

CHANGELOG.md
```diff
@@ -176,6 +176,8 @@ programmatically in MATLAB or python.
   and may be installed in a different location.
 - macOS and Linux users should no longer need to set `LD_LIBRARY_PATH` or
   `DYLD_LIBRARY_PATH` to use OpenSim libraries.
+- The `scale()` method was removed from the `SimbodyEngine` class (the contents
+  were moved into `Model::scale()`).
```
aseth1: :+1:

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp
```diff
@@ -798,55 +798,6 @@ void SimbodyEngine::convertQuaternionsToDirectionCosines(double aQ1, double aQ2,
 //--- Private Utility Methods Below Here ---
 
 
-bool SimbodyEngine::scale(SimTK::State& s, const ScaleSet& scaleSet,
```
aseth1: 🎆 🍾

OpenSim/Simulation/Model/Model.cpp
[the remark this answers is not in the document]
tkuchida: Ajay had the same suggestion. I moved the contents of `SimbodyEngine::scale()` into `Model::scale()` and noted the removal of `SimbodyEngine::scale()` in CHANGELOG.md.

OpenSim/Simulation/SimbodyEngine/EllipsoidJoint.cpp
tkuchida: <pre>Error: class "SimTK::Vec<3, SimTK::Real, 1>" has no member "elementwiseMultiplyInPlace"</pre>
😕

OpenSim/Simulation/Model/ModelComponent.h
tkuchida: Oh, actually I don't think it's a good idea to remove default arguments from the existing `Model::scale()` because then the user would need to pass in `-1.0` (or any number less than zero) for `finalMass` to get the same behavior. I'll work out something else.

tkuchida: I'll change `bool Model::scale(State&, const ScaleSet&, double finalMass = -1.0, bool preserveMassDist = false)` to `bool Model::scale(State&, const ScaleSet&, bool preserveMassDist, double finalMass = -1.0)`. The user will now need to specify whether they want to preserve the mass distribution, but can still leave `finalMass` unspecified (which I think is better than entering `-1` for a mass). I'll make a note of this change in CHANGELOG.md.

tkuchida:
Thanks for the quick turnaround, @chrisdembia and @aseth1. All comments addressed; ready for re-review.

OpenSim/Simulation/Model/ModelComponent.cpp
```diff
@@ -127,4 +127,26 @@ void ModelComponent::updateFromXMLNode(SimTK::Xml::Element& aNode,
...
+void ModelComponent::preScale(const SimTK::State& s, const ScaleSet& scaleSet)
+{   extendPreScale(s, scaleSet); }
+
+void ModelComponent::scale(const SimTK::State& s, const ScaleSet& scaleSet)
+{   extendScale(s, scaleSet); }
+
+void ModelComponent::postScale(const SimTK::State& s, const ScaleSet& scaleSet)
+{   extendPostScale(s, scaleSet); }
+
+const SimTK::Vec3 ModelComponent::
```
aseth1: You can remove the `const` since you are returning a copy anyways. Otherwise to avoid copies, you could return a const reference. You could define  `const static SimTK::Vec3 InvalidScaleFactors{SimTK::NaN}`  as a private data member to return in the not found case (e.g.`return InvalidScaleFactors;`).  This is just a minor fine-tuning.

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -1428,87 +1428,86 @@ void Model::removeController(Controller *aController)
...
+    // GeometryPath lengths (and perform any other necessary computations).
+    for (ModelComponent& comp : updComponentList<ModelComponent>())
+        comp.preScale(s, scaleSet);
+
+    // Call scale() on each ModelComponent owned by the model. Each
+    // ModelComponent is responsible for scaling itself. All scaling operations
+    // are performed here except scaling inertial properties of bodies, which is
+    // done below.
+    for (ModelComponent& comp : updComponentList<ModelComponent>())
+        comp.scale(s, scaleSet);
```
aseth1: Nice! This is so clean and uniform now. 🥇

CHANGELOG.md
```diff
@@ -82,6 +82,10 @@ Converting from v3.x to v4.0
 where fiber-velocity can be estimated from the state or assumed to be zero if the flag is *true*.
 - `Millard2012EquilibriumMuscle::computeInitialFiberEquilibrium(SimTK::State&)` invokes `computeFiberEquilibrium()` with `useZeroVelocity = true` to maintain its previous behavior.
 - `Model::replaceMarkerSet()` was removed. (PR #1938) Please use `Model::updMarkerSet()` to edit the model's MarkerSet instead.
+- The argument list for `Model::scale()` was changed: the third and fourth
+  arguments were swapped and the `preserveMassDist` argument is no longer
+  optional. The default argument for `preserveMassDist` in OpenSim 3.3 was
+  `false`. (PR #1994)
```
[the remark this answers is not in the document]
tkuchida: Done.

OpenSim/Simulation/Model/ModelComponent.cpp
tkuchida: I removed the `const` and added a note (somewhat obvious) above the `return` statement. I think defining `InvalidScaleFactors` and returning a const reference is unnecessary for what's going on here. (I would also need to change the calling methods to use something like `areScaleFactorsValid()` because checking `scaleFactors.isNaN()` would make an assumption about the default return value that, presumably, could change. If we're concerned about performance, the next thing to improve is the string comparison part in `ModelComponent::getScaleFactors()`.

chrisdembia:
>  TODONE:

:) LGTM.

> Removed meaningless const and added note about default return.

(to be annoying) I don't think the const is quite exactly meaningless. It means that the caller could not modify the return value. For example, the following would not be possible.

```
auto x getScaleFactors();
x[0] = 1; // ERROR, x is const.
```

tkuchida:
> (to be annoying) I don't think the const is quite exactly meaningless. It means that the caller could not modify the return value. For example, the following would not be possible.

That's why I put it there; Ajay said to remove it: "You can remove the `const` since you are returning a copy anyways." @aseth1 and @chrisdembia please provide clear direction on what you would like me to do here.

chrisdembia:
@tkuchida I've approved the PR. I agree with @aseth1 about removing the const. If the caller wants const, they can just use `const auto`.

tkuchida:
@chrisdembia Perhaps you'd prefer it if I implemented Ajay's other suggestion: "Otherwise to avoid copies, you could return a const reference. You could define  `const static SimTK::Vec3 InvalidScaleFactors{SimTK::NaN}`  as a private data member to return in the not found case (e.g.`return InvalidScaleFactors;`)."

chrisdembia:
> @chrisdembia Perhaps you'd prefer...

No no no I want this to be merged.

tkuchida:
Ah, ok. I think you're referring to my commit comment then. I guess "meaningless" was incorrect. I don't know how to change that comment without amending the commit, but I think you said that commits shouldn't be amended once they have already been pushed to a public repo. Not sure how to fix what I've done, apart from being more careful about my commit comments in the future.

tkuchida:
FWIW, I put `const` there to communicate that the caller shouldn't be changing the `Vec3` that is returned; it's what the user specified. I'm always fetching the scale factors with `const Vec3 scaleFactors = getScaleFactors(...);`, but yes I'm not *enforcing* that the returned Vec3 isn't changed. I think there are higher-impact things we could work on.

aseth1:
>  I think there are higher-impact things we could work on.
 
I agree! Let's merge it :)

Issue #1729, opened 2017-05-23
Binary vtp files can't be opened


aymanhab:
Up to version 3.3, we used vtk to read mesh files (vtp) so we handled them seamlessly. The API and new visualization doesn't use the vtk native libraries and accordingly fails to load binary vtp files (only ASCII vtp files). This showed in one user model/geometry for the OpenSim 4.0 paper (Apoorva's model).

tkuchida:
3.3:
![image33x8](https://cloud.githubusercontent.com/assets/4203505/26370199/f815260c-3fab-11e7-99ed-ea487549e623.png)

4.0:
![image40](https://cloud.githubusercontent.com/assets/4203505/26370240/0d8985aa-3fac-11e7-8527-ee2c4948d197.png)

aymanhab:
Thanks @tkuchida 💯

aymanhab:
Will manually convert our vtp files to ascii and ask users to do that using Paraview for their own files.

chrisdembia:
This example might be helpful: http://www.vtk.org/Wiki/VTK/Examples/Cxx/IO/WriteVTP

PR #1997, opened 2017-11-03
Catch exception for binary vtp.

chrisdembia:
Addresses part of #1729 

### Brief summary of changes

- Catch the exception that is thrown if a VTP file is binary, and print the exception message:

```
Visualizer couldn't open hat_ribs_scap.vtp because:
SimTK Exception thrown at PolygonalMesh.cpp:411:
  Error detected by Simbody method PolygonalMesh::loadVtpFile(): Attempt to load a VTK PolyData (.vtp) file from file name '/Users/chris/repos/opensim32/1/install/Geometry/hat_ribs_scap.vtp' failed with message:
  SimTK Exception thrown at PolygonalMesh.cpp:340:
  Error detected by Simbody method PolygonalMesh::loadVtpFile(): Only format="ascii" is supported for .vtpfile DataArray elements, got format="appended" for Points DataArray.
  (Required condition 'pointData.getRequiredAttributeValue("format") == "ascii"' was not met.)

  (Required condition '!"failed"' was not met.)
```

- I made sure to add the check when visualizing, *not* in `extendFinalizeFromProperties()`, to avoid slowing down `initSystem()`.

### Testing I've completed

- Used the handy `visualizeModel` exectuable to load a model that pointed to a binary VTP file, and ensured the model initially did not visualize, and that with these changes, the model visualized and an error message was printed. 

@aymanhab can you test this in the GUI?

Do we have any advice to give to users about how to convert binary VTP files to other formats?

### CHANGELOG.md (choose one)

- no need to update because...we never supported binary VTP files in opensim-core; we just handle the error silently now.

### DIFF
--- a/OpenSim/Simulation/Model/Geometry.cpp
+++ b/OpenSim/Simulation/Model/Geometry.cpp
@@ -218,7 +218,7 @@ void Mesh::extendFinalizeFromProperties() {
     if (!isObjectUpToDateWithProperties()) {
         const Component* rootModel = nullptr;
         if (!hasOwner()) {
-            std::cout << "Mesh " << get_mesh_file() << " not connected to model..ignoring\n";
+            std::cout << "Mesh " << get_mesh_file() << " not connected to model..ignoring" << std::endl;
             return;   // Orphan Mesh not part of a model yet
         }
         const Component* owner = &getOwner();
@@ -234,7 +234,7 @@ void Mesh::extendFinalizeFromProperties() {
         }
 
         if (rootModel == nullptr) {
-            std::cout << "Mesh " << get_mesh_file() << " not connected to model..ignoring\n";
+            std::cout << "Mesh " << get_mesh_file() << " not connected to model..ignoring" << std::endl;
             return;   // Orphan Mesh not descendant of a model
         }
 
@@ -251,7 +251,7 @@ void Mesh::extendFinalizeFromProperties() {
         const string lowerExtension = SimTK::String::toLower(extension);
         if (lowerExtension != ".vtp" && lowerExtension != ".obj" && lowerExtension != ".stl") {
             std::cout << "ModelVisualizer ignoring '" << file
-                << "'; only .vtp, .stl, and .obj files currently supported.\n";
+                << "'; only .vtp, .stl, and .obj files currently supported." << std::endl;
             return;
         }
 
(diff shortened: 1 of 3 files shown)


OpenSim/Simulation/Model/Geometry.cpp
```diff
@@ -218,7 +218,7 @@ void Mesh::extendFinalizeFromProperties() {
     if (!isObjectUpToDateWithProperties()) {
         const Component* rootModel = nullptr;
         if (!hasOwner()) {
-            std::cout << "Mesh " << get_mesh_file() << " not connected to model..ignoring\n";
+            std::cout << "Mesh " << get_mesh_file() << " not connected to model..ignoring" << std::endl;
```
chrisdembia: This is to flush the stream so that users will see the messages. Otherwise, the messages might not be printed (until the stream is eventually flushed somewhere else).

aymanhab:
@chrisdembia Thanks for the quick turnaround. Can you please merge master into the branch so I can test a unified updated environment? Thanks

OpenSim/Tests/VisualizeModel/VisualizeModel.cpp
```diff
@@ -30,7 +30,8 @@
 
 //==============================================================================
 //==============================================================================
-#include <OpenSim/OpenSim.h>
+#include <OpenSim/Common/IO.h>
+#include <OpenSim/Simulation/Model/Model.h>
```
aymanhab: 👍

OpenSim/Tests/VisualizeModel/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ set(VISUALIZE_MODEL visualizeModel)
 
 add_executable(${VISUALIZE_MODEL} VisualizeModel.cpp)
 
-target_link_libraries(${VISUALIZE_MODEL} osimTools)
+target_link_libraries(${VISUALIZE_MODEL} osimAnalyses)
```
aymanhab: Not sure why osimAnalyses here, in the past I used osimTools which is the top of stack due to using OpenSim.h, now that we're using dev. environment headers we likely can get away with using osimActuators (to pull in muscles) but I don't know if you tried that. Glad you found this utility useful.

OpenSim/Simulation/Model/Geometry.cpp
```diff
@@ -298,6 +297,20 @@ void Mesh::extendFinalizeFromProperties() {
...
+            // it's expensive to repeatedly load meshes.
+            cachedMesh->getMesh();
+        } catch (const std::exception& e) {
+            std::cout << "Visualizer couldn't open "
+                << get_mesh_file() << " because:\n"
+                << e.what() << std::endl;
+            // No longer try to visualize this mesh.
+            cachedMesh.reset();
+            return;
+        }
```
aseth1: :+1:

OpenSim/Tests/VisualizeModel/CMakeLists.txt
aseth1: What would happen if you tried to visualize a Model with Analyses?

aymanhab: Analyses get ignored similar to any other plugins. The goal is to minimize dependencies if possible.

chrisdembia: I did not try depending on only `osimActuators`, but I thought it wouldn't work for the reason @aseth1 mentioned. I agree with minimizing dependencies and speeding up compile time but this is separate from the purpose of the PR and I would prefer to not try to further optimize here.

chrisdembia:
Thanks for the quick review.

@aymanhab I merged with master.

aymanhab:
Tested integrated build and works great, thanks @chrisdembia Merging since test failures are due to unrelated timeout

chrisdembia:
Great. Thanks @aymanhab

PR #1999, opened 2017-11-04
Convenience constructor for Marker

nickbianco:
### Brief summary of changes
Added a convenience constructor to the Marker class that allows the user to provide the name, PhysicalFrame, and location in PhysicalFrame of the marker at construction. 

### Testing I've completed
Made a simple test file using the constructor that compiles. Approval from @chrisdembia to move forward with PR.

Not looking for feedback.

### CHANGELOG.md (choose one)
- no need to update because changes are fairly simple, not a major change to the Marker class


OpenSim/Simulation/Model/Marker.cpp
```diff
@@ -49,6 +49,18 @@ Station()
 
 //_____________________________________________________________________________
 /**
+* Conveniece constructor.
```
tkuchida: Conveniece -> Convenience.
Also, it's unnecessary to use doxygen-style formatting (`/** ... */`) for comments in .cpp files.

chrisdembia: @nickbianco, you might know this already but the thing that makes it a doxygen comment is the two asterisks rather than 1 (`/**` vs `/*`).

nickbianco: I'll fix the typo. @chrisdembia I didn't know that, I'll remove them from the comments too while I'm at it.

nickbianco:
Pushed typo/comment formatting changes.

chrisdembia:
I'm not sure why a merge conflict appeared. It would be good to check the master branch after this PR is merged to ensure the new `Marker::extendScale()` function is in place, etc.

nickbianco:
I don't see any scaling methods in the Marker class on master.

chrisdembia:
Ah okay; you're right. I got confused.

PR #2002, opened 2017-11-06
Fix bug preventing writing non-string metadata.

chrisdembia:
### Brief summary of changes

This PR fixes a bug when writing a table that contains non-string table metadata: it doesn't get serialized. The reason is that all metadadta was assumed to be of type `std::string`.

An issue remains that when reading a table, all metadata is assumed to be `std::string`. This is harder to fix, because we do not write type information to the file.

### Testing I've completed

- Tested that this fix works with Muscollo.

### CHANGELOG.md (choose one)

- no need to update because...file adapters are new in 4.0.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2002)

### DIFF
--- a/OpenSim/Common/DelimFileAdapter.h
+++ b/OpenSim/Common/DelimFileAdapter.h
@@ -542,9 +542,10 @@ DelimFileAdapter<T>::extendWrite(const InputTables& absTables,
 
     // First line of the stream is the header.
     if (table->getTableMetaData().hasKey("header")) {
-        out_stream << table->
-                      getTableMetaData().
-                      getValueForKey("header").getValueAsString() << "\n";
+        const auto& header = table->getTableMetaData().getValueForKey("header");
+        // TODO We assume the header is of type std::string.
+        // This is usually the case, but we don't enforce this anywhere.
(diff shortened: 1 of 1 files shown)


aseth1:
CI test failure:
```
Testing reading/writing STOFileAdapter_<double>
Testing STOFileAdapter::read() and STOFileAdapter::write()
  std_subject01_walk1_ik.mot
terminate called after throwing an instance of 'OpenSim::Exception'
```

Perhaps the *type* associated with the meta data in the file header has changed (correctly) and the test case should be updated?

chrisdembia:
Yes, you're right. Sorry for the delay here.

chrisdembia:
I just looked into this. The issue is that `SimTK::Value::getValueAsString()` is not implemented. So for now, I reverted back to assuming that the header is of type `std::string`.

chrisdembia:
Chatting with @aseth1, I should add a test to ensure we can actually serialize non-string metadata.

aymanhab:
Sorry for the delay but this doesn't seem to be a user issue at the moment so aging out to reduce clutter!

Issue #1957, opened 2017-10-04
Need ScaleTool to write intermediate files to user specified directory


aymanhab:
Running Scale on the gait2354 model using provided setup files on Windows in GUI produced the following files in the OpenSim installation folder:

- subject01_simbody
- subject01_static_output
- Unassigned (file with Markers)
- subject01_scaledOnly
- subject01_scaleSet_applied.

On osx the working directory is the root folder and so is not writable. Need to

1. Decide what files are useful and need to write based on users feedback
2. A way to specify where the files are written

jenhicks:
Don't output Unassigned. 

Default of the tool: write to the same folder as set-up file. If tool is run in from the GUI, will prompt.

aymanhab:
@aseth1 will check

aseth1:
closed by #2003

tkuchida:
From @aymanhab via https://github.com/opensim-org/opensim-gui/issues/401: "We need to decide behavior if files are specified in xml or else why the failure on osx." :apple:

aymanhab:
@aseth1 Using the latest osx build with the supposed fix does not write the intermediate files to the directory of the setup file, instead it fails with 

Exception:
  SimTK Exception thrown at Xml.cpp:115:
  Error detected by Simbody method Xml::writeToFile(): Failed to write to the
  Xml file 'subject01_scaledOnly.osim' with error ' (line=1, col=1)'.
  (Required condition 'saveOK' was not met.)

  In Object '<no-name>' of type ModelScaler.
  Thrown at ModelScaler.cpp:332 in processModel().
  line= 0


can you please explain if it should have worked or if the files (included in Modesl/Gait2392_Simbody) need an update? Thanks.

aseth1:
Are you able to confirm that it isn't trying to write to the setup file directory? It isn't clear to me where it is trying to write instead.

PR #2003, opened 2017-11-06
Write scale results to folder with setupfile

aseth1:
Fixes issue #1957
There were two bugs addressed: 
1. "*Unassigned*" files were always being written out because the check used was inadequate. This included both `MarkerSet` and `ScaleSet` files from the `MarkerPlacer` and `ModelScaler`, respectively.

2. The `ModelScaler` updates the working directory to the location of the `ScaleTool` setup file (`ScaleTool` it calls it the `_pathToSubject`), but he `MarkerPlacer` receives the same path in its `processModel` method except it was not doing anything with it.

### Brief summary of changes
1. Employ `PropertyStr::isValidFileName()` to check for the validity of filenames. *Unassigned* fails this test so these files are not longer written out.
2. Make  `MarkerPlacer::processModel` change directory to the the setup file directory (like `ModelScaler` does) upon printing results.
 
### Testing I've completed
- verified that Unassigned files are no longer printed.
- verified that all `testScale` generated files are written to the directory with the setup file. 

- verification that in the GUI files are not written to the application directory

### CHANGELOG.md (choose one)

### DIFF
--- a/OpenSim/Common/PropertyStr.h
+++ b/OpenSim/Common/PropertyStr.h
@@ -95,7 +95,7 @@ public:
     void clearValue() { _value = getDefaultStr(); setValueIsDefault(true); }
     static const std::string& getDefaultStr();
 
-    bool isValidFileName() { return _value!="" && _value!=getDefaultStr(); }
+    bool isValidFileName() const { return _value!="" && _value!=getDefaultStr(); }
 
 //=============================================================================
 };  // END of class PropertyStr
--- a/OpenSim/Tools/MarkerPlacer.cpp
+++ b/OpenSim/Tools/MarkerPlacer.cpp
@@ -391,19 +391,19 @@ bool MarkerPlacer::processModel(Model* aModel,
     //_outputStorage->print("statesReporterOutputWithMarkers.sto");
 
     if(_printResultFiles) {
-        if (!_outputModelFileNameProp.getValueIsDefault())
+        if (!_outputModelFileNameProp.isValidFileName())
         {
             aModel->print(aPathToSubject + _outputModelFileName);
             cout << "Wrote model file " << _outputModelFileName << " from model " << aModel->getName() << endl;
         }
 
-        if (!_outputMarkerFileNameProp.getValueIsDefault())
+        if (!_outputMarkerFileNameProp.isValidFileName())
         {
             aModel->writeMarkerFile(aPathToSubject + _outputMarkerFileName);
             cout << "Wrote marker file " << _outputMarkerFileName << " from model " << aModel->getName() << endl;
         }
         
-        if (!_outputMotionFileNameProp.getValueIsDefault())
+        if ( _outputMotionFileNameProp.isValidFileName())
         {
             _outputStorage->print(aPathToSubject + _outputMotionFileName, 
                 "w", "File generated from solving marker data for model "+aModel->getName());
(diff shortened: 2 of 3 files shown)


OpenSim/Common/PropertyStr.h
```diff
@@ -95,7 +95,7 @@ class OSIMCOMMON_API PropertyStr : public Property_Deprecated
     void clearValue() { _value = getDefaultStr(); setValueIsDefault(true); }
     static const std::string& getDefaultStr();
 
-    bool isValidFileName() { return _value!="" && _value!=getDefaultStr(); }
+    bool isValidFileName() const { return _value!="" && _value!=getDefaultStr(); }
```
tkuchida: :+1:

OpenSim/Tools/MarkerPlacer.cpp
```diff
@@ -391,23 +392,27 @@ bool MarkerPlacer::processModel(Model* aModel,
...
     if(_printResultFiles) {
-        if (!_outputModelFileNameProp.getValueIsDefault())
-        {
+        std::string savedCwd = IO::getCwd();
+        IO::chDir(aPathToSubject);
+
+        if (_outputModelFileNameProp.isValidFileName()) {
             aModel->print(aPathToSubject + _outputModelFileName);
-            cout << "Wrote model file " << _outputModelFileName << " from model " << aModel->getName() << endl;
+            cout << "Wrote model file " << _outputModelFileName <<
```
tkuchida: Could `aModel->print()`, `aModel->writeMarkerFile()`, etc. throw exceptions (e.g., if write access is denied for the specified directory, insufficient space, etc.)? If so, perhaps any exceptions need to be caught here so that the directory can be restored (i.e., `IO::chDir(savedCwd);`), and then re-throw the original exception?

aseth1: That's a good point. I think that is a safe thing to add. I'll add that here and in `ModelScaler`.

OpenSim/Tools/MarkerPlacer.cpp
```diff
@@ -391,23 +392,34 @@ bool MarkerPlacer::processModel(Model* aModel,
...
+
+            if (_outputMotionFileNameProp.isValidFileName()) {
+                _outputStorage->print(aPathToSubject + _outputMotionFileName,
+                    "w", "File generated from solving marker data for model "
+                    + aModel->getName());
+            }
+        } // catch the exception so we can reset the working directory
+        catch (std::exception& ex) {
+            IO::chDir(savedCwd);
+            // now re-throw in context of ModelPlacer
```
tkuchida: ModelPlacer -> MarkerPlacer, or just remove the comment as I think this line is self-documenting

OpenSim/Tools/ModelScaler.cpp
```diff
@@ -311,27 +311,30 @@ bool ModelScaler::processModel(Model* aModel, const string& aPathToSubject,
...
+                        cout << "Wrote scale file " << _outputScaleFileName <<
+                        " for model " << aModel->getName() << endl;
+                }
             }
-
-            if (!_outputScaleFileNameProp.getValueIsDefault())
-            {
-                if (theScaleSet.print(_outputScaleFileName))
-                    cout << "Wrote scale file " << _outputScaleFileName << " for model " << aModel->getName() << endl;
+            catch (std::exception& ex) {
```
tkuchida: The comment in the analogous place in MarkerPlacer.cpp (`// catch the exception so we can reset the working directory`) would have the same usefulness here.

aseth1:
Thanks for the rapid review @tkuchida 💨

aseth1:
Thanks @aymanhab for the ✔️

PR #2008, opened 2017-11-08
Storage constructor can read version 2 STO files

### LINKED ISSUE #1983: Edit the constructor to Storage() so that it can read version 2 storage file.
STOFileAdapter() prepends `Version=2' in output header info when writing to file. When used by InverseDynamicsTool(), an exception is thrown because (apparently) version (2) files aren't supported. 

 ```
Storage: file=/Users/projects/xxxx/mocap-subject03/data/HS_1.mot (nr=6000 nc=19)
InverseDynamicsTool Failed: Error: File version (2) not supported. Use STOFileAdapter instead.
Java exception occurred:
java.io.IOException: InverseDynamicsTool Failed, please see messages
window for details...
	at
        org.opensim.modeling.opensimActuatorsAnalysesToolsJNI.InverseDynamicsTool_run(Native
        Method)
	at
        org.opensim.modeling.InverseDynamicsTool.run(InverseDynamicsTool.java:117) 
```
If the version row is deleted, the tool runs as expected.

### PR DESCRIPTION
**Pull request by chrisdembia**
Fixes #1983 

### Brief summary of changes

- Allow Storage to read version 2 storage files, as we agreed in a dev meeting a while ago.
- Renamed `Storage::getTimeSeriesTable()` to `Storage::exportToTable()`.

### Testing I've completed

- Added test case to `testStorage`.

@jimmyDunne can you test this locally with the inverse dynamics tool and a version 2 storage file?

### CHANGELOG.md (choose one)

- no need to update because...version 2 storage files are new in 4.0.

### DIFF
--- a/OpenSim/Common/Storage.cpp
+++ b/OpenSim/Common/Storage.cpp
@@ -1268,7 +1268,7 @@ Storage::getColumnIndicesForIdentifier(const std::string& identifier) const
     return found;
 }
 
-TimeSeriesTable Storage::getAsTimeSeriesTable() const {
+TimeSeriesTable Storage::exportToTable() const {
     TimeSeriesTable table{};
 
     table.addTableMetaData("header", getName());
--- a/OpenSim/Common/Storage.cpp
+++ b/OpenSim/Common/Storage.cpp
@@ -36,9 +36,26 @@
 #include "SimTKcommon.h"
 #include "GCVSpline.h"
 #include "StateVector.h"
+#include "STOFileAdapter.h"
+#include "TimeSeriesTable.h"
 
 using namespace OpenSim;
 using namespace std;
+
+void convertTableToStorage(const TimeSeriesTable& table, Storage& sto) {
+    sto.purge();
+    OpenSim::Array<std::string> labels("", (int)table.getNumColumns() + 1);
+    labels[0] = "time";
+    for (int i = 0; i < (int)table.getNumColumns(); ++i) {
+        labels[i + 1] = table.getColumnLabel(i);
+    }
+    sto.setColumnLabels(labels);
+    const auto& times = table.getIndependentColumn();
+    for (unsigned i_time = 0; i_time < table.getNumRows(); ++i_time) {
+        auto rowView = table.getRowAtIndex(i_time);
+        sto.append(times[i_time], SimTK::Vector(rowView.transpose()));
+    }
+}            
 
 
 //============================================================================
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by chrisdembia** [comment]
Thanks for the quick review @aseth1

File: OpenSim/Common/Storage.h [hunk 1]
```diff
@@ -191,8 +195,8 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Storage, StorageInterface);
     int getDataColumn(const std::string& columnName,double *&rData) const;
     void getDataColumn(const std::string& columnName, Array<double>& data, double startTime=0.0) override;
 
-    /** Get a TimeSeriesTable out of the Storage.                             */
-    TimeSeriesTable getAsTimeSeriesTable() const;
+    /** Convert to a TimeSeriesTable (new in 4.0).                            */
```
tkuchida: I don't think methods have been marked "new in 4.0" anywhere else. If it's necessary to draw the user's attention to this method, perhaps add a note in CHANGELOG?

chrisdembia: I was anticipating the user wondering "why would I want a TST over a Storage?" The "4.0" bit was to say "you might want a TST if you have to work with other new features in 4.0." Also, it was an attempt to subtly say that TST is the way of the future.

How about the following alternative: "Convert to a TimeSeriesTable. This may be useful if you need to use parts of the API that require a TimeSeriesTable instead of a Storage. We encourage you to use TimeSeriesTables instead of Storages if possible."

OpenSim/Common/Storage.cpp
```diff
@@ -36,10 +36,27 @@
 #include "SimTKcommon.h"
 #include "GCVSpline.h"
 #include "StateVector.h"
+#include "STOFileAdapter.h"
+#include "TimeSeriesTable.h"
 
 using namespace OpenSim;
 using namespace std;
 
+void convertTableToStorage(const TimeSeriesTable& table, Storage& sto) {
```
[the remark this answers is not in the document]
chrisdembia: Making this private static would be fine with me.

We decided in a dev meeting to not support programmatically converting a TST to a Storage, to encourage migration to TST. So I didn't want to expose this function in the API.  I suppose I put this function only in the CPP file to further enforce that this is just a hidden utility.

As for why I pass in the storage rather than making this a member function: I think free functions (or equivalently, static member function) are easier to understand; you don't have to worry about how the function is affecting the state of the object so much. Also, I wanted to write a function that would be easy to move to the API if we decided to expose it (though perhaps making a static private function would be even easier to move into the API).

OpenSim/Common/Storage.h
tkuchida: Ah, ok. I suggest adding the full explanation, as "new in 4.0" didn't convey that information to me. If you use the text in your second paragraph, I suggest also explaining why "We encourage you to use TimeSeriesTables instead of Storages if possible" so we're not just telling the user to blindly do what we tell them to. I think the statement in your first paragraph ("you might want a TST if you have to work with other new features in 4.0") would be helpful.

OpenSim/Common/Storage.cpp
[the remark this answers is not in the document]
tkuchida: Ok, thanks for the explanation. If the Dev Team doesn't want to support TST->Storage conversion, then I can see wanting to keep the method out of the header.

chrisdembia:
@tkuchida I improved the doxygen comment for `exportToTable()`.

tkuchida:
> @tkuchida I improved the doxygen comment for `exportToTable()`.

:+1: Looks great.

PR #2013, opened 2017-11-09
Add ContactGeometry::updateFromXMLNode() for appearance.

### LINKED ISSUE #2010: ContactGeometry updateFromXMLNode does not handle display_preference or color
[ContactGeometry used to have `display_preference` and `color` properties](https://github.com/opensim-org/opensim-core/blob/d579acb06f5e7c2a6938ff3273cf8a4664190de0/OpenSim/Simulation/Model/ContactGeometry.h), but these are now part of `Appearance`. The old properties are not copied over to the new properties, and so `display_preference` and `color` are ignored.

I can try fixing this.

chrisdembia:
Fixes #2010 

### Brief summary of changes

- Converted the old `color` and `display_preference` properties of ContactGeometry into `Appearance`.

### Testing I've completed

- Updated DeMers' drop landing model (with some of my own modifications) and ensured it visualized as expected.

Here is the model in the 3.3 GUI:
screen shot 2017-11-08 at 8 31 52 pm

Here is the model in 4.0 before this PR (left) and after this PR (right).
screen shot 2017-11-08 at 8 24 27 pm

Before this PR, the meshes showed up as wireframe instead of solid, and the color was ignored.

The one remaining issue is that the `<visible>` flag is ignored by the 4.0 GUI; I've opened a separate issue for that: https://github.com/opensim-org/opensim-gui/issues/438

You can ignore the fact that the contact half space is visualized differently; that's a separate issue.

### CHANGELOG.md (choose one)

- no need to update because...this is a bug fix for a bug introduced after 3.3.

### DIFF
--- a/OpenSim/Simulation/Model/ContactGeometry.cpp
+++ b/OpenSim/Simulation/Model/ContactGeometry.cpp
@@ -133,6 +133,61 @@ void ContactGeometry::updateFromXMLNode(SimTK::Xml::Element& node,
             XMLDocument::addConnector(node, "Connector_PhysicalFrame_",
                     "frame", body_name);
         }
+
+        if (versionNumber < 30507) {
+            // display_preference and color were replaced with Appearance in PR
+            // #1122, at which point the latest XML version number was 30507;
+            // however, the corresponding updateFromXMLNode code below was
+            // added a while later.
+            // https://github.com/opensim-org/opensim-core/pull/1122
+            
+            SimTK::Xml::Element appearanceNode("Appearance");
+
+            // Move color inside Appearance.
+            SimTK::Xml::element_iterator colorIter =
+                        node.element_begin("color");
+            bool addAppearanceNode = false;
+            if (colorIter != node.element_end()) {
+                appearanceNode.insertNodeAfter(appearanceNode.element_end(),
+                        node.removeNode(colorIter));
+                addAppearanceNode = true;
+            } else {
+                // If the user didn't set a color but set one of the other
+                // Appearance properties, then we must set color explicitly so
+                // that the Appearance's default of white is not used.
+                SimTK::Xml::Element color("color");
+                color.setValue("0 1 1"); // SimTK::Cyan
+                appearanceNode.insertNodeAfter(appearanceNode.element_end(),
+                        color);
+            }
+
+            // Move <display_preference> to
+            // <Appearance><SurfaceProperties><representation>
+            SimTK::Xml::element_iterator reprIter =
+                    node.element_begin("display_preference");
+            if (reprIter != node.element_end()) {
+                if (reprIter->getValue() == "0") {
+                    SimTK::Xml::Element visible("visible");
+                    visible.setValue("false");
+                    appearanceNode.insertNodeAfter(appearanceNode.element_end(),
+                            visible);
+                } else {
+                    reprIter->setElementTag("representation");
+                    if (reprIter->getValue() == "4") {
+                        // Enum changed to go 0-3 instead of 0-4
+                        reprIter->setValue("3");
+                    }
+                    SimTK::Xml::Element surfProp("SurfaceProperties");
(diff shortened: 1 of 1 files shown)


OpenSim/Simulation/Model/ContactGeometry.cpp
```diff
@@ -133,6 +133,61 @@ void ContactGeometry::updateFromXMLNode(SimTK::Xml::Element& node,
...
+            if (reprIter != node.element_end()) {
+                if (reprIter->getValue() == "0") {
+                    SimTK::Xml::Element visible("visible");
+                    visible.setValue("false");
+                    appearanceNode.insertNodeAfter(appearanceNode.element_end(),
+                            visible);
+                } else {
+                    reprIter->setElementTag("representation");
+                    if (reprIter->getValue() == "4") {
+                        // Enum changed to go 0-3 instead of 0-4
```
aseth1: Perhaps you can explain why the list 0-3 now instead of 0-4?

Not for this PR, but I really don't like the use of integers here. They are virtually meaningless to the user. We should use strings that reflect the enums used in the code: Hide, DrawPoints, DrawWireframe, DrawSurface.

OpenSim/Simulation/Model/ContactGeometry.cpp
```diff
@@ -133,6 +133,61 @@ void ContactGeometry::updateFromXMLNode(SimTK::Xml::Element& node,
...
+                        reprIter->setValue("3");
+                    }
+                    SimTK::Xml::Element surfProp("SurfaceProperties");
+                    surfProp.insertNodeAfter(surfProp.element_end(),
+                            node.removeNode(reprIter));
+                    appearanceNode.insertNodeAfter(appearanceNode.element_end(),
+                            surfProp);
+                }
+                addAppearanceNode = true;
```
aseth1: I'm not sure I understand the use of this bool.  AppearanceNode should be added even in the defaults case, or not? In L160 you add the default color, but how will this take effect if the appearanceNode is not added?

OpenSim/Simulation/Model/ContactGeometry.cpp
chrisdembia: I copied this comment from a similar section in PhysicalFrame.cpp. I looked into this more and the old property, "display_preference," could be `0:Hide 1:Wire 3:Flat 4:Shaded` ([see here](https://github.com/opensim-org/opensim32/blob/OpenSim33/OpenSim/Common/VisibleObject.cpp#L220)). The new property, "representation," can be `1:Points, 2:Wire, 3:Shaded` ([see here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/Appearance.h#L62)). I'm not sure what happens if `representation` is set to 0.

chrisdembia: I'm not sure what course of action to take without fixing larger issues.

OpenSim/Simulation/Model/ContactGeometry.cpp
chrisdembia: The default appearance of contact geometry is not simply the default properties of the Appearance class. Therefore, without the bool and the associated logic, if someone altered only the "display_preference" but left the color as default, then the color would not end up as Cyan (which it should). So in L160, I set the color in case we determine later on that the user edited some other properties of the Appearance.

If the user didn't edit any of these properties, there is no need to add the appearance node; we should just use the default property value.

[WrapObject](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Wrap/WrapObject.cpp#L185) has similar logic.

aseth1: Thanks for the clarification @chrisdembia.

chrisdembia:
Thanks @aseth1

chrisdembia:
Need 1 more review.

chrisdembia:
Thank you @aymanhab

PR #2021, opened 2017-11-11
Improve test coverage for scaling

tkuchida:
Fixes issue #1973

### Brief summary of changes
Some ModelComponents implement `extendScale()` but were not being tested. This PR enhances tests for scaling ModelComponents attached to PhysicalOffsetFrames (rather than bodies directly), and adds tests for EllipsoidJoints, CustomJoints, and CoordinateCouplerConstraints.

### Testing I've completed
Ran tests locally.

### CHANGELOG.md (choose one)
- no need to update because... added test cases only.


Applications/Scale/test/testScale.cpp
```diff
@@ -460,4 +470,329 @@ void scalePhysicalOffsetFrames()
...
+            SimTK::SignificantReal, __FILE__, __LINE__,
+            "SliderJoint has incorrect Coordinate value before scaling.");
+
+        slider1->getCoordinate().setValue(s, 1., true);
+        ASSERT_EQUAL(slider2->getCoordinate().getValue(s), intercept + slope,
+            SimTK::SignificantReal, __FILE__, __LINE__,
+            "SliderJoint has incorrect Coordinate value before scaling.");
+
+        // Scale body2 uniformly.
+        const double uniformFactor = 3.14159;
```
chrisdembia: I think this is fine here, but why not use a more random number?

Applications/Scale/test/testScale.cpp
```diff
@@ -460,4 +470,329 @@ void scalePhysicalOffsetFrames()
...
+        // Attach bodies to Ground with slider joints.
+        SliderJoint* slider1 =
+            new SliderJoint("slider1", model->getGround(), *body1);
+        slider1->updCoordinate().setName("indepCoord");
+        model->addJoint(slider1);
+
+        SliderJoint* slider2 =
+            new SliderJoint("slider2", *body2, model->getGround());
+        slider2->updCoordinate().setName("depCoord");
+        model->addJoint(slider2);
```
chrisdembia: I was confused when I saw this because I thought the test might be similar to the tests above (making sure the scaling behavior differs depending on the parent frame). Perhaps construct this joint as `SliderJoint(..., *body1, *body2)`?

Applications/Scale/test/testScale.cpp
tkuchida: What random number would you like me to use?

Applications/Scale/test/testScale.cpp
tkuchida: Ok, sure.

Applications/Scale/test/testScale.cpp
tkuchida: I've replaced with 1.234, which I used on L351 and presumably didn't cause confusion. Let me know if you want me to use a seeded random number generator.

tkuchida: I prefer typing the numbers in rather than using a random number generator (even if it's seeded) to make debugging easier. Introducing a random number generator also increases the complexity of the test code (potentially introducing issues related only to the random number generator) and reduces readability. I typically use the decimal expansion of pi, the golden ratio, etc. for "random" (arbitrary) numbers, but I recognize that others may not pick up on this and could think that these arbitrary numbers were chosen purposefully. I have therefore replaced the arbitrary numbers in these tests with 1.1, 2.22, 3.333, 0.123, 0.456, 0.789, 1.234, 5.678, 123.456, and 1.23456. These numbers were chosen to convey that they are arbitrary while maintaining the property that they are not multiples of each other (which are often bad numbers to use in test cases). Let me know if these numbers don't align with your definition of "a more random number".

chrisdembia: I could have been clearer; I did not mean to use a random number generator function. Your recent change is exactly in line with what I was thinking.

Applications/Scale/test/testScale.cpp
```diff
@@ -460,4 +470,328 @@ void scalePhysicalOffsetFrames()
...
+                    "The location of point " + std::to_string(i)
+                    + " does not match (" + msg + ")\n  PathActuator1: "
+                    + p1.toString() + "\n  PathActuator2: " + p2.toString());
+            }
+        };
+
+        testPathPointLoc(s, "before scaling");
+
+        // Scale the model.
+        model->scale(s, scaleSet, false);
```
aseth1: Minor point, but you might want to add a comment that indicates that `Model::scale()` updates the properties of the same model and re-inits the System and updates the the input state.  In the future, I hope that scaling just returns a new model and leaves it to the consumer of the scaled model to `initSystem()` or not, especially if it plans to perform more edits.  If/when that happens the lambda will break (which is OK) and a comment here will shed light as to why (i.e. `act1` and `act2` will still be pointing to the old model).

tkuchida: I was unaware you were planning this API change. I'll add the comment here, but I think there will be a lot of (uncommented) code to change other than this lambda.

aseth1: True! I was just surprised that `act1` and `act2` were not stale on the call to `testPathPointLoc` after scaling. I am not planning to make any API changes for 4.0.

tkuchida: `act1` and `act2` are defined inside the lambda. How would they be stale on repeated calls to the function?

tkuchida: I see what you mean: I'll fix that!

tkuchida: Scaling doesn't destroy the Model, so the pointers to the two PathActuators remain valid. Not sure I understand why `scale()` would create a new model (it seems like it should be the responsibility of the caller to create a duplicate if it wants to retain an unscaled version). Regardless, I've updated the test to :fish: the PathActuators out of the model for the sake of the future developer. Bonus marks if you caught the two puns :wink:

aseth1:
Looks great! :+1:

tkuchida:
Thanks, @aseth1 and @chrisdembia 😎

PR #2033, opened 2017-12-05
Reorder component properties in Model

### LINKED ISSUE #1763: XML Ground appears as first element in Model, above <publications>, etc.
The Ground property now appears at the top of XML files (unless the Model also has subcomponents in <components>), above other elements like <publications>. I think it makes more sense to put the Ground property next to the BodySet, etc.

This was originally raised in #776.

chrisdembia:
Fixes #1763.

### Brief summary of changes

This PR introduces a more logical ordering of component properties in Model. With this PR, the XML tags `<gravity>`, etc. will come before `<Ground>`.

### Testing I've completed

- None; relying on CI.

- Do you like this ordering?

### CHANGELOG.md (choose one)

- no need to update because...this does not alter the API.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -101,7 +101,7 @@ before_install:
   ## Avoid bug on OSX (https://github.com/travis-ci/travis-ci/issues/6307)
   # This prevents the build from failing with:
   # `/Users/travis/build.sh: line 109: shell_session_update: command not found`
-  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then rvm get head; fi
+  # - if [ "$TRAVIS_OS_NAME" = "osx" ]; then rvm get head; fi
   
   # Avoid "Homebrew must be run under Ruby 2.3!"
   # https://github.com/PowerShell/PowerShell/issues/5062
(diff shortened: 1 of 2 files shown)


chrisdembia:
Ready for review.

.travis.yml
```diff
@@ -101,7 +101,7 @@ before_install:
   ## Avoid bug on OSX (https://github.com/travis-ci/travis-ci/issues/6307)
   # This prevents the build from failing with:
   # `/Users/travis/build.sh: line 109: shell_session_update: command not found`
-  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then rvm get head; fi
+  # - if [ "$TRAVIS_OS_NAME" = "osx" ]; then rvm get head; fi
```
aseth1: Travis failure seems unrelated since the linux build fails to find Doxygen, but can you confirm?

chrisdembia: Yes the failure is unrelated; see #2039. The test doesn't get far along enough to actually test the code, though.

chrisdembia:
Thank you @aseth1

PR #2034, opened 2017-12-11
Replace MSVC-specific _DEBUG with NDEBUG.

chrisdembia:
This PR fixes a bug that caused tests to fail on Linux in Debug: some tests were unable to load the `osimActuators` library in Debug. This was caused by use of a non-existent macro `_DEBUG` (this macro is only defined on Windows).


chrisdembia:
> Does not look wrong to me.

👍

chrisdembia:
Could I interest anyone else in reviewing this? It should be a very quick review!

chrisdembia:
Thank you!

PR #2036, opened 2017-12-12
Improve error message when osimJavaJNI not found.

### LINKED ISSUE #1555: Make osimJavaJNI loadLibrary message platform-dependent.
The Java exception/error for failing to find the OpenSim shared libraries is tailored toward Windows (it mentions the PATH environment variable, but not LD_LIBRARY_PATH or DYLD_LIBRARY_PATH) (https://github.com/opensim-org/opensim-core/blob/master/Bindings/Java/swig/java_common.i#L189). It would be great to update this error message to be more accurate for platforms like macOS and linux.

chrisdembia:
Fixes #1555

### Brief summary of changes

- Improves the "cannot find osimJavaJNI" message for Mac and Linux.
- Improves support for Matlab 2017*.

### Testing I've completed

- Ensured the osimJavaJNI error message appeared properly on Linux via Matlab and via Java tests.


aymanhab:
I like the change, thanks @chrisdembia Some follow up questions:
- Name of the library is different per platform so osimJavaJNI.dll on windows is libosimJavaJNI.x on different platforms.
- There're other reasons the library fails to load (example, dependent libraries not found) we may want to change the language to allow for that, for others this only maps to "some random Java error".  Would it make sense to include a link/url on the error message to help troubleshoot?

tkuchida:
Here's a use case from the Forum: https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=8491

chrisdembia:
> Name of the library is different per platform so osimJavaJNI.dll on windows is libosimJavaJNI.x on different platforms.

I'm considering removing any specific library names, since the detailed exception message will tell us the library name if relevant. However, if we leave it in, **osimJavaJNI** should be enough to indicate what file we're looking for. I would prefer not to make this an exact cross-platform name, because there might be debug variants (`_d`) or versions numbers (`libosimJavaJNI.so.1`) in the future.

> There're other reasons the library fails to load (example, dependent libraries not found) we may want to change the language to allow for that, for others this only maps to "some random Java error".

How about this: 

```java
if (windows) {
    libDesc = "bin";
} else {
    libDesc = "lib";
}

msg = "Failed to load one or more dynamic libraries for OpenSim.\n" + e +
"\nMake sure OpenSim's " + libDesc + " directory is on your " + envVar + ".";

if (System.getProperties().getProperty('java.home').lowercase() contains "matlab") {
    msg +=  "\nSee https://simtk-confluence.stanford.edu/display/OpenSim/Scripting+with+Matlab";
}
```

(I put the detailed exception message before the "Make sure...".)

Is there a good link to give if the user is getting this message in the GUI rather than in MATLAB?

Ideally, the link would be a "Help" button in the dialog box, not text of the message. @aymanhab is that possible?

I would appreciate any suggestions for improving the content or ordering of content for this error message.

aymanhab:
I like this better already since it avoids the narrow/misleading specification of the issue as missing JNI library. The dialog pops up during the loading of the jar file so at this point we don't much context to rely on. We can customize the dialog more instead of using the built in JOptionPane but that feels like an overkill and not worth it to me (as we need to pull in more libraries/packages to support it while in truth we're dead in the water if that dialog shows up at all ). The link to confluence is perfectly fine and then it's a living document for troubleshooting that we can beef up as we go.

chrisdembia:
I've updated the PR based. I've tested on Linux only.

pure Java:
image

In Matlab:
image

Getting the URL to be clickable would be a lot more work. The message is also printed to the Matlab command window, and the URL should be easy to click/copy from there.

chrisdembia:
Ready for review.

chrisdembia:
Ready for review.

chrisdembia:
@aymanhab I created a diff locally; hopefully that suffices:

screen shot 2018-01-30 at 2 48 19 pm

If you click the image, it should be large enough to read.

I will test on Mac and Windows and get back to you.

chrisdembia:
### Mac

Running the Matlab_TugOfWar CTest, I get the following console output:

```
98: Failed to load one or more dynamic libraries for OpenSim.
98: java.lang.UnsatisfiedLinkError: no osimJavaJNI in java.library.path
98: See https://simtk-confluence.stanford.edu/display/OpenSim/Scripting+with+Matlab
```

Trying to create a Model in the MATLAB GUI gives the following pop-up:

screen shot 2018-01-30 at 3 31 01 pm

and also produces the same text in the console as above:

screen shot 2018-01-30 at 3 31 53 pm


### Windows

Here's Windows:

screen shot 2018-01-30 at 4 35 16 pm



So, the updated error message works properly on Linux, Mac, and Windows.

chrisdembia:
It would be great to get another reviewer for this. As I mentioned earlier, I'm happy to walk the reviewer through these changes.

chrisdembia:
If no one else wants to review this, perhaps we could discuss if @aymanhab's sole review is sufficient for the changes in this PR.

aseth1:
LGTM

chrisdembia:
Thank you @aseth1.

PR #2040, opened 2017-12-14
Suppress SWIG warnings from "using" declaration.

chrisdembia:
SWIG does not support C++11 usage of inherited constructors ("using").
This PR suppresses SWIG warnings about "using".

I ensured that SWIG no longer generates warnings related to the edited code.


OpenSim/Common/Exception.h
```diff
@@ -276,7 +276,9 @@ class KeyNotFound : public Exception {
 
 class IOError : public Exception {
 public:
+#ifndef SWIG
```
aseth1: According to swig [docs](http://www.swig.org/Doc3.0/CPlusPlus11.html#CPlusPlus11_object_construction_improvement) this shouldn't be necessary. SWIG should at least parse correctly.

chrisdembia: Ah I was able to remove the changes to this file. Consistent with the documentation, there is no warning on this line.

However, there is a warning in `TRCFileAdapter.h:32`:

```
/Users/chris/repos/opensim/13/opensim-core/OpenSim/Common/TRCFileAdapter.h:32: Wa
rning 315: Nothing known about 'IOError::IOError'.  
```

OpenSim/Simulation/Model/PhysicalOffsetFrame.h
```diff
@@ -56,7 +56,9 @@ class OSIMSIMULATION_API PhysicalOffsetFrame : public OffsetFrame<PhysicalFrame>
     // CONSTRUCTION
     //--------------------------------------------------------------------------
     /** Constructors are defined by the OffsetFrame base class */
+#ifndef SWIG    
     using OffsetFrame<PhysicalFrame>::OffsetFrame;
+#endif
```
chrisdembia: This change is to avoid the following warning:

```
/Users/chris/repos/opensim/13/opensim-core/OpenSim/Simulation/Model/PhysicalOffse
tFrame.h:59: Warning 315: Nothing known about 'OpenSim::OffsetFrame< OpenSim::Phy
sicalFrame >::OffsetFrame'.   
```

We already wrap `OpenSim::OffsetFrame< OpenSim::PhysicalFrame >`, so apparently that wasn't sufficient for avoiding this warning. Hence the `ifndef`.

aseth1: OK. Thanks for the check. Likely future releases of SWIG will remedy the issue. We can attempt to remove when that happens.

chrisdembia:
@aseth1 I removed an unnecessary `#ifndef SWIG` and provided the warnings that are generated without this PR.

aseth1:
Feel free to merge when CI is complete.

chrisdembia:
Great; thank you.

chrisdembia:
Merging. The travis-ci failure is unrelated.

Issue #2037, opened 2017-12-12
configureOpenSim.m: allow user to specify OpenSim install location


chrisdembia:
When configureOpenSim.m is installed in a user's documents folder, it will be necessary for this script to ask the user for the OpenSim install location (rather than assuming the script is within an OpenSim installation).

PR #2045, opened 2017-12-21
configureOpenSim: prompt for install dir.

chrisdembia:
Fixes issue #2037 

### Brief summary of changes

- The `configureOpenSim.m` function now will prompt users for the OpenSim install directory (restoring the initial behavior of the function) if it seems that the `configureOpenSim.m` file is not within an OpenSim installation.
- We added this change because we want the GUI to copy the Matlab scripts to the user's Documents folder, and without this PR, the `configureOpenSim.m` script will not work from the user's Documents folder.

### Testing I've completed

- I tested on both Mac and Windows, and ensured I got the proper error message if I selected an incorrect folder.

### CHANGELOG.md (choose one)

- no need to update because...this restores old behavior.

### DIFF
--- a/Bindings/Java/Matlab/configureOpenSim.m.in
+++ b/Bindings/Java/Matlab/configureOpenSim.m.in
@@ -19,8 +19,10 @@ function configureOpenSim()
 %    Depending on your operating system, this function may require
 %    administrator (or sudo) privileges to succeed.
 %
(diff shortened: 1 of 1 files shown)


aseth1:
I was able to get the prompt and error message when trying it on my Mac. I am going to go ahead and merge.

Issue #2050, opened 2018-01-11
Old model file does not open in 4.0, but opens in 3.3 and opens after converting/saving in 3.3


tkuchida:
@tkuchida commented on [Thu Jan 11 2018](https://github.com/opensim-org/opensim-gui/issues/470)

One of Sam's model files (version 10600) does not open in 4.0. Messages window:
<pre>
Updating Model file from 10600 to latest format...
Model name '3D Gait Model with Simple Arms' contains whitespace. It was renamed '3DGaitModelwithSimpleArms'.
Got 0 object values for Property function but the minimum is 1. Continuing anyway.
Got 0 object values for Property function but the minimum is 1. Continuing anyway.
[35 more instances of the same error.]
Loaded model 3DGaitModelwithSimpleArms from file [path]\FullBodyModel_SimpleArms_Hamner2010_HPLMarkers_subject05_v191.osim
</pre>
Nothing appears in the visualizer window despite the Messages window reporting that the model was loaded. The Navigator panel is also empty. Red-circle error message:
<pre>
java.lang.RuntimeException: SimTK Exception thrown at TransformAxis.cpp:89:
  Internal bug detected: CustomJoint (ground_pelvis) r1 axis has no function but has coordinates.
  (Assertion 'coordNames.size() == 0' failed).
  Please file an Issue at https://github.com/simbody/simbody/issues.
  Include the above information and anything else needed to reproduce the problem.
	at org.opensim.modeling.opensimSimulationJNI.Model_initSystem(Native Method)
	at org.opensim.modeling.Model.initSystem(Model.java:840)
	at org.opensim.view.pub.OpenSimDB.addModel(OpenSimDB.java:146)
	at org.opensim.view.pub.OpenSimDB.addModel(OpenSimDB.java:141)
	at org.opensim.view.FileOpenOsimModelAction$2.run(FileOpenOsimModelAction.java:98)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:756)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:80)
[...]

PR #2053, opened 2018-01-12
Delay TransformAxis check for Function when Coordinates present

aseth1:
Fixes issue #2050

A check that we added in 4.0  was causing models with a `SpatialTransform` `TransformAxis` without an associated `Function` to fail . This was causing the failure  to load in #2050.

### Brief summary of changes
In 92d3de2d017b26f236e9c292e7a6980ba2440740 we added a check that a `TransformAxis` with `Coordinate`s must have an associated `Function`. This check was happening before the `SpatialTransform` had the opportunity to supply functions in the case that functions where unspecified, which was the case in model version 10600 and earlier. It loads in 3.3 because the check was not there and the functions are correctly set afterwards. Fix was simply to delay the check until after the `SpatialTransform` has a chance to add them. 

### Testing I've completed
Verified that model in #2050 loads.

### CHANGELOG.md (choose one)
- no need to update because this is a big fix.

### DIFF
--- a/OpenSim/Simulation/SimbodyEngine/SpatialTransform.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/SpatialTransform.cpp
@@ -105,9 +105,6 @@ void SpatialTransform::connectToJoint(CustomJoint& owningJoint)
     for(int i=0; i < NumTransformAxes; ++i) {
         TransformAxis& transform = updTransformAxis(i);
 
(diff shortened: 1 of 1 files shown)


tkuchida:
Travis failed on OSX with unrelated "ERROR: cannot verify prdownloads.sourceforge.net's certificate"; merging.

Issue #2052, opened 2018-01-11
InverseKinematicsTool marker error calculation: vector size is incorrect.


chrisdembia:
In InverseKinematicsTool, if the number of model markers is less than the number of experimental markers, then we incorrectly read past the end of an array:

https://github.com/opensim-org/opensim-core/blob/4199ecd2f211a5063716390f73b62aff8e0cce8c/OpenSim/Tools/InverseKinematicsTool.cpp#L315-L334

Here, `nm` is the number of experimental markers, but ` ikSolver.computeCurrentSquaredMarkerErrors(squaredMarkerErrors);` resizes `squaredMarkerErrors` to be the number of model markers. The next few lines then index into `squaredMarkerErrors` as if its length were the number of experimental markers.

This could potentially cause us to report much smaller marker errors than is correct, because the RMS error is computed by dividing by the (incorrectly) larger number of experimental markers.

aseth1:
I am not sure this assessment is correct. 
In `InverseKinematicsSolver::computeCurrentSquaredMarkerErrors()` the `squaredMarkerErrors` vector is resized here:
```cpp
markerErrors.resize(_markerAssemblyCondition->getNumMarkers());
```
and the `markerAssemblyCondition` is not the model but the internal assembly condition which only includes markers that were supplied by the `MarkersReference`.

aseth1:
Talking with @chrisdembia the issue is that `_markerAssemblyCondition->getNumMarkers()` is the size of the intersection of experimental and model markers and `markersReference.getNumRefs()` includes all the experimental markers, regardless if they are used or not. Reporting should only be based on the intersection and not on all the markers provided.

aseth1:
Closed by #2054

PR #2054, opened 2018-01-17
Fix IKTool reporting to use size of markerErrors returned from IKSolver

aseth1:
Fixes issue #2052 

### Brief summary of changes
Use the size of the vector of marker errors returned by the `InverseKinematicsSolver` and do not assume it is the number of references in the `MarkersReference`. In fact, the number is not necessarily fixed, since a marker is ignored if its reference value is NaN.

### Testing I've completed
Changes have no impact on existing tests. They still pass.

Would like @chrisdembia to apply fix on the case he reported.

### CHANGELOG.md (choose one)
- no need to update because this is an internal bug fix

### DIFF
--- a/OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
+++ b/OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
@@ -561,7 +561,7 @@ void testNumberOfMarkersMismatch()
 
         auto namesIter = usedMarkerNames.begin();
         for (int j = 0; j < nme; ++j) {
-            auto& markerName = ikSolver.getMarkerNameForIndex(j);
+            const auto& markerName = ikSolver.getMarkerNameForIndex(j);
             cout << " " << markerName << " error = " << markerErrors[j];
 
             SimTK_ASSERT_ALWAYS( *namesIter++ != "unused",
--- a/OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
+++ b/OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
@@ -51,9 +51,10 @@ Model* constructPendulumWithMarkers();
 // randomly select the noise to be added at each frame.
 TimeSeriesTable_<SimTK::Vec3>
 generateMarkerDataFromModelAndStates(const Model& model,
-                                     const StatesTrajectory& states,
-                                     double noiseRadius = 0, 
-                                     bool constantOffset = false);
+                        const StatesTrajectory& states,
+                        const SimTK::RowVector_<SimTK::Vec3>& biases,
+                        double noiseRadius = 0, 
+                        bool constantOffset = false);
 
 // Verify that accuracy improves the number of decimals points to which
 // the solver solution (coordinates) can be trusted as it is tightened.
@@ -65,9 +66,16 @@ void testUpdateMarkerWeights();
 // weights and marker error is being reduced as its weighting increases.
 void testTrackWithUpdateMarkerWeights();
 
+// Verify that solver does not confuse/mismanage markers when reference
+// has more markers than the model.
+// the solver solution (coordinates) can be trusted as it is tightened.
+void testNumberOfMarkersMismatch();
+
 int main()
 {
     SimTK::Array_<std::string> failures;
+
+    testNumberOfMarkersMismatch();
 
     try { testMarkersReference(); }
     catch (const std::exception& e) {
@@ -199,8 +207,9 @@ void testAccuracy()
     StatesTrajectory states;
     states.append(state);
 
+    SimTK::RowVector_<SimTK::Vec3> biases(3, SimTK::Vec3(0));
     MarkersReference
-        markersRef(generateMarkerDataFromModelAndStates(*pendulum, states));
+        markersRef(generateMarkerDataFromModelAndStates(*pendulum, states, biases));
     markersRef.setDefaultWeight(1.0);
 
     // Reset the initial coordinate value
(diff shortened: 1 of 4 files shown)


OpenSim/Tools/InverseKinematicsTool.cpp
```diff
@@ -329,6 +329,10 @@ bool InverseKinematicsTool::run()
                 int worst = -1;
 
                 ikSolver.computeCurrentSquaredMarkerErrors(squaredMarkerErrors);
+                // if a marker is not present in the model or its reference value 
+                // is NaN it is not included. Therefore, size of the errors can
+                // vary from frame to frame.
```
tkuchida: Perhaps an Issue should be opened about this? If `squaredMarkerErrors` changes in size, then it seems like the caller would have trouble monitoring the error of a particular marker (its index in the vector will be changing). Furthermore, I'm not sure `getMarkerNameForIndex()` below (on L353) will report the correct marker name if a marker with a lower index has dropped out. I would expect `computeCurrentSquaredMarkerErrors()` to always return a vector of the same (max) size; the caller would then step through (as on L336–342) and ignore any `NaN`s.

tkuchida:
> Changes have no impact on existing tests. They still pass.

Are tests missing?

aseth1:
@tkuchida Yes, I think we were missing some fundamental tests for how order and missing data are handled by the  `InverseKinematicsSolver`. I added a test case to address it at the Solver level. It doesn't directly affect this issue, but now I can definitively say that the number of markers (in the assembly condition) does not change from Frame to frame. In the case of NaN it is not included in the solve but its marker error is reported as 0. @chrisdembia is correct that incorrect number is being used when computing mean error since the number of markers should be decreased by 1 for very NaN marker.

I will fix that next in the Tool. Thanks for the review so far.

chrisdembia:
@aseth1 I pushed a commit with a fix for clang. I then tested this change in my code and the issue is gone.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
```diff
@@ -456,6 +474,114 @@ void testTrackWithUpdateMarkerWeights()
...
+            const auto& markerName = ikSolver.getMarkerNameForIndex(j);
+            cout << " " << markerName << " error = " << markerErrors[j];
+
+            SimTK_ASSERT_ALWAYS( *namesIter++ != "unused",
+                "InverseKinematicsSolver failed to ignore "
+                "unused marker reference (observation).");
+            
+            if (markerName == "m0") {// marker we biased shoulder the error
+                SimTK_ASSERT_ALWAYS(abs(markerErrors[j]-err) <= tol,
+                    "InverseKinematicsSolver mangled marker oder.");
```
tkuchida: oder -> order (also L577)

aseth1:
The issue related to zero marker error being reported when reference is NaN resulting in incorrect error stats was added in #2058.

aseth1:
@tkuchida and @chrisdembia thanks for the review. I decided to add a method to return the number of markers in use by the solver. This number is fixed during IK (repeated calls to track) and I am now using it in the InverseKinematicsTool. Let me know if you are OK with the change. Thanks!

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
```diff
@@ -456,6 +474,120 @@ void testTrackWithUpdateMarkerWeights()
...
 
+
+void testNumberOfMarkersMismatch()
+{
+    cout << 
+        "\ntestInverseKinematicsSolver::testNumberOfMarkersMismatch()"
+        << endl;
+
+    std::unique_ptr<Model> pendulum{ constructPendulumWithMarkers() };
+    Coordinate& coord = pendulum->getCoordinateSet()[0];
```
chrisdembia: Should this be `const Coordinate&`?

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
```diff
@@ -456,6 +474,120 @@ void testTrackWithUpdateMarkerWeights()
...
+    for (unsigned i = 0; i < markersRef.getNumFrames(); ++i) {
+        state.updTime() = i*dt;
+        ikSolver.track(state);
+
+        //get the marker errors
+        ikSolver.computeCurrentMarkerErrors(markerErrors);
+
+        int nme = markerErrors.size();
+
+        SimTK_ASSERT_ALWAYS(nme = nm,
```
chrisdembia: This is an assignment, not an equality comparison, and will thus return the value of `nm`.

OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp
```diff
@@ -456,6 +474,120 @@ void testTrackWithUpdateMarkerWeights()
...
+        auto namesIter = usedMarkerNames.begin();
+        for (int j = 0; j < nme; ++j) {
+            const auto& markerName = ikSolver.getMarkerNameForIndex(j);
+            cout << " " << markerName << " error = " << markerErrors[j];
+
+            SimTK_ASSERT_ALWAYS( *namesIter++ != "unused",
+                "InverseKinematicsSolver failed to ignore "
+                "unused marker reference (observation).");
+            
+            if (markerName == "m0") {// marker we biased shoulder the error
```
chrisdembia: Could you rephrase "marker we biased shoulder the error"? I don't understand what this means.

aseth1:
> To be clear, this does not fix #2058, correct?

Correct.

aseth1:
@chrisdembia thank you for pointing out those errors. They should be fixed now.

aseth1:
Thanks @chrisdembia!

tkuchida:
> Let me know if you are OK with the change.

:+1: Travis/osx is sitting at "waiting to be queued" (≈ "to be deprecated"); all other tests :white_check_mark:

aseth1:
Thank you @tkuchida

PR #2055, opened 2018-01-17
Fix wrapobject appearance

aymanhab:
Fixes issue https://github.com/opensim-org/opensim-gui/issues/482 reported by beta tester

### Brief summary of changes
updateFromXML has been modified to always create an Appearance node for old WrapObjects because WrapObject Appearance defaults  (cyan, half transparent) are different from a default Appearance object (white, opaque).

### Testing I've completed
Tests pass and GUI built with this change do in fact keep Appearance correctly post (de)serialization.

modifying updateFromXML 

### CHANGELOG.md (choose one)

- no need to update because bug fix

### DIFF
--- a/OpenSim/Simulation/Wrap/WrapObject.cpp
+++ b/OpenSim/Simulation/Wrap/WrapObject.cpp
@@ -203,7 +203,7 @@ void WrapObject::updateFromXMLNode(SimTK::Xml::Element& node,
             appearanceNode.insertNodeAfter(appearanceNode.element_end(),
                     defaultSurfProp);
             SimTK::Xml::Element rep("representation");
-            rep.setValue("2"); // VisualRepresentation::DrawSurface
+            rep.setValue("3"); // VisualRepresentation::DrawSurface
             defaultSurfProp.insertNodeAfter(defaultSurfProp.element_end(), rep);
 
             // color.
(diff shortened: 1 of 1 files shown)


chrisdembia:
FYI I updated the PR description to include a link to https://github.com/opensim-org/opensim-gui/issues/482

aymanhab:
Test failures unrelated. Ready for review.

aymanhab:
Basically the change in this PR is to modify the old updateFromXML code so that an Appearance node is always added to WrapObject. This is necessary since the default Appearance object has different defaults (White, opaque, visible) from those assumed by WrapObject (Cyan, half-transparent-visible).

OpenSim/Simulation/Wrap/WrapObject.cpp
```diff
@@ -241,29 +237,25 @@ void WrapObject::updateFromXMLNode(SimTK::Xml::Element& node,
...
                 }
                 node.removeNode(visObjIter);
             }
             if (visibleNode.getValue() != "") {
                 appearanceNode.insertNodeAfter(appearanceNode.element_end(),
                         visibleNode);
-                appearanceModified = true;
             }
-
+            // Regardless, we'll add Appearance since default for WrapObject is different from
```
aseth1: Regardless of what?
Minor, comment exceeds 80cols.

aymanhab:
Thanks for the quick review @aseth1 I updated the comment to clarify.

aymanhab:
I'll close this since I reproduced the issue with a 4.0 wrap object, so it's not model conversion issue. Thanks for the review anyway.

Issue #2058, opened 2018-01-23
InverseKinematicsTool error stats do not properly account for NaN marker reference 


aseth1:
In the case that a marker reference provides NaN the resulting error is zero, which enable the optimization to run smoothly. However, when computing error statistics we should reduce N of marker errors by 1 for each marker that is  NaN at that instant in time.

PR #2059, opened 2018-01-24
Computing marker RMS uses the number of non-NaN values

aseth1:
Fixes issue #2058

### Brief summary of changes
Added methods to access the marker reference values from the `InverseKinematicsSolver`.
Using this to check if any reference value is NaN when accounting for it when reporting marker RMS errors. 

### Testing I've completed
Added conditions to testInverseKinematicsSolver to check that NaN reference values are being returned and correspond to the correct marker.

There isn't an explicit test for running the IKTool with with experimental data with NaNs.

 @jimmyDunne , @chrisdembia , @apoorvar perhaps you have a model and data handy for me to use as a test case?

### CHANGELOG.md (choose one)
- no need to update because this a bug fix

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2059)

### DIFF
--- a/OpenSim/Tools/InverseKinematicsTool.cpp
+++ b/OpenSim/Tools/InverseKinematicsTool.cpp
@@ -341,7 +341,7 @@ bool InverseKinematicsTool::run()
 
                 ikSolver.getMarkerReferenceValues(markerRefValues);
                 for (const Vec3& val : markerRefValues) {
-                    val.isNaN() ? ignore_cnt++ : 0;
+                    if(val.isNaN()) ignore_cnt++;
                 }
 
                 ikSolver.computeCurrentSquaredMarkerErrors(squaredMarkerErrors);
--- a/OpenSim/Tools/InverseKinematicsTool.cpp
+++ b/OpenSim/Tools/InverseKinematicsTool.cpp
@@ -332,10 +332,17 @@ bool InverseKinematicsTool::run()
             ikSolver.track(s);
             
             if(_reportErrors){
-                Array<double> markerErrors(0.0, 3);
+                Array_<Vec3> markerRefValues(nm);
+                Array<double> markerErrors(0.0, nm);
                 double totalSquaredMarkerError = 0.0;
                 double maxSquaredMarkerError = 0.0;
                 int worst = -1;
+                int ignore_cnt = 0;
+
+                ikSolver.getMarkerReferenceValues(markerRefValues);
+                for (const Vec3& val : markerRefValues) {
+                    val.isNaN() ? ignore_cnt++ : 0;
+                }
 
                 ikSolver.computeCurrentSquaredMarkerErrors(squaredMarkerErrors);
                 for(int j=0; j<nm; ++j){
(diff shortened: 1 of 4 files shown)


chrisdembia:
@aseth1 I have a model and a TRC file for the case that the model has fewer markers than the TRC file. I do not have data that has NaNs.

[Archive.zip](https://github.com/opensim-org/opensim-core/files/1662096/Archive.zip)

OpenSim/Tools/InverseKinematicsTool.cpp
```diff
@@ -332,10 +332,17 @@ bool InverseKinematicsTool::run()
...
+                Array_<Vec3> markerRefValues(nm);
+                Array<double> markerErrors(0.0, nm);
                 double totalSquaredMarkerError = 0.0;
                 double maxSquaredMarkerError = 0.0;
                 int worst = -1;
+                int ignore_cnt = 0;
+
+                ikSolver.getMarkerReferenceValues(markerRefValues);
+                for (const Vec3& val : markerRefValues) {
+                    val.isNaN() ? ignore_cnt++ : 0;
```
chrisdembia: I think it makes more sense to write this as `if (val.isNaN()) ++ignore_cnt;` because the value of the `? :` expression is not used.

chrisdembia: Does this count skip over references for markers that do not have a corresponding model marker?

chrisdembia (commented):
I like the improved RMS reporting in this PR. I made one small comment. I think it's fine that there is not an explicit test for the updated RMS calculation, but it would be good to just manually inspect that the RMS is larger with this change. Doing so is difficult, because introducing NaNs into the TRC file also changes the IK solution. It seems that verifying this fix requires using the setup you have in the IKSolver test, where a marker is not in the plane of the multibody system.

OpenSim/Tools/InverseKinematicsTool.cpp
aseth1: > Does this count skip over references for markers that do not have a corresponding model marker? 

Markers without a corresponding model marker are not part of the reference values returned because they are not added to the solver to begin with.

chrisdembia: Okay thanks for clarifying.

chrisdembia (approved):

aseth1:
@jimmyDunne do you have a trial with NaNs handy? We need to verify that the RMS error reported increases when markers drop out (instead of getting more accurate, which was wrong).

jimmyDunne:
Attached is an edited trc file and corresponding model. The trc file is the one distributed with gait2392.

[Archive.zip](https://github.com/opensim-org/opensim-core/files/1700466/Archive.zip)

jimmyDunne:
@aseth1, it seems like some conflicts would need to be fixed but, is this still something we should try to get in?

aseth1:
@jimmyDunne I thought this got merged ages ago! I can resolve the conflicts if you are willing to verify on a real example. The test case included makes sure that NaNs are ignored and the number of markers is reduced at that instant, so that the mean error is computed over the correct number. I am very confident it is correct, but would be great to verify.

nickbianco:
Closing: stale PR.

PR #2060, opened 2018-01-25
[WIP] MATLAB muscle analysis example

carmichaelong:
This provides two functions, one that takes runs muscle analysis by taking in strings as arguments, and another that uses the GUI in MATLAB to populate a muscle list and allows the user to choose muscles via the GUI. There's definitely some redundant code between the functions. These files were made over a year ago, so it may not work with the current build.

@aseth1 let me know how you want to proceed with this.


chrisdembia:
@aseth1 did you test these files?

PR #2064, opened 2018-01-29
[travis] Fix sourceforge certificate error

chrisdembia:
This PR changes Travis CI to use the xcode8.3 image (macOS 10.12) to avoid sourceforge certificate error that is causing Mac builds to fail.


chrisdembia:
Thanks.

PR #2065, opened 2018-01-29
Register slope of -1 in CustomJoint as rotational.

### LINKED ISSUE #2062: Model and kinematics of normal walking from 3.3 visualize with incorrect hip kinematics in 4.0
I tried visualizing kinematics of normal walking using Apoorva Rajagopal's model (both originally from 3.3) in 4.0, and I get the following (both in the GUI and with the API in simbody-visualizer):

pastedgraphic-2
pastedgraphic-1

In the 3.3 GUI, this motion looks like normal walking. However, in 4.0, the left hip (only the left hip) seems to undergo incorrect hip abduction/rotation. Ajay says "my guess is that there is a bug in the assignment of the coordinate values from the data from file."

chrisdembia:
Fixes #2062.

### Brief summary of changes

- This PR fixes a bug that caused some rotational DOFs in a CustomJoint to not be detected as rotational.

### Testing I've completed

- Added a test case in testJoints.

### CHANGELOG.md (choose one)

- no need to update because... this fixes a bug that was introduced post-3.3.

### DIFF
--- a/OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp
@@ -171,20 +171,23 @@ void CustomJoint::constructCoordinates()
                 if (spatialTransform[j].hasFunction()) {
                     lf = dynamic_cast<const LinearFunction*>(
                             &spatialTransform[j].get_function() );
-                    // if displacement on axis is linear (w/ slope of 1) w.r.t.
-                    // the coordinate value, we have a pure rotation/translation
-                    if (lf && lf->getSlope() == 1.0) {
+                    // if displacement on axis is linear (w/ slope of +/-1)
+                    // w.r.t. the coordinate value, we have a pure
+                    // rotation/translation
+                    if (lf && (lf->getSlope() == 1.0 || lf->getSlope() == -1.0))
+                    {
                         // coordinate is pure axis displacement
                         if (j < 3)
                             // coordinate about rotational axis 
                             mt = Coordinate::MotionType::Rotational;
                         else // otherwise translational axis 
                             mt = Coordinate::MotionType::Translational;
-                    }
-                    else { // scaled (slope !=1) or nonlinear relationship means
-                           // not a pure rotational or translational for this axis.
-                           // designate as Coupled unless already defined as
-                           // pure Rotational or Translational about another axis
+                    } else {
+                        // scaled (slope != +/-1) or nonlinear relationship
+                        // means not a pure rotational or translational for
+                        // this axis. designate as Coupled unless already
+                        // defined as pure Rotational or Translational about
+                        // another axis
                         mt = (mt == Coordinate::MotionType::Undefined ?
                             Coordinate::MotionType::Coupled : mt);
                     }
--- a/OpenSim/Simulation/SimbodyEngine/Test/testJoints.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/Test/testJoints.cpp
@@ -2501,69 +2501,112 @@ void testUniversalJointAccessors()
 
 void testMotionTypesForCustomJointCoordinates()
 {
-    Model osimModel;
+    {
+        Model osimModel;
 
-    //OpenSim bodies
-    const Ground& ground = osimModel.getGround();
-    //OpenSim thigh
-    auto osim_thigh = new OpenSim::Body ("thigh", femurMass.getMass(),
-        femurMass.getMassCenter(), femurMass.getInertia());
+        //OpenSim bodies
+        const Ground& ground = osimModel.getGround();
+        //OpenSim thigh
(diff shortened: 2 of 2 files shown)


OpenSim/Simulation/SimbodyEngine/Test/testJoints.cpp
```diff
@@ -2501,69 +2501,112 @@ void testUniversalJointAccessors()
...
-    ASSERT( mt == Coordinate::MotionType::Coupled, __FILE__, __LINE__,
-        "Coordinate `" + coordName + "' failed to register as MotionType::Coupled");
-
-    // hip_tz is the fourth coordinate, which is pure translational along Z
-    coordName = hip->getCoordinate(3).getName();
-    mt = hip->getCoordinate(3).getMotionType();
-    ASSERT( mt == Coordinate::MotionType::Translational, __FILE__, __LINE__,
-        "Coordinate `" + coordName + 
-        "' failed to register as MotionType::Translational");
+    {
```
chrisdembia: Although it looks like I edited every line of this function, all I did was indent the existing lines (see the diff without changes in whitespace: https://github.com/opensim-org/opensim-core/pull/2065/files?w=1).

OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp
```diff
@@ -171,20 +171,23 @@ void CustomJoint::constructCoordinates()
                 if (spatialTransform[j].hasFunction()) {
                     lf = dynamic_cast<const LinearFunction*>(
                             &spatialTransform[j].get_function() );
-                    // if displacement on axis is linear (w/ slope of 1) w.r.t.
-                    // the coordinate value, we have a pure rotation/translation
-                    if (lf && lf->getSlope() == 1.0) {
+                    // if displacement on axis is linear (w/ slope of +/-1)
+                    // w.r.t. the coordinate value, we have a pure
+                    // rotation/translation
+                    if (lf && (lf->getSlope() == 1.0 || lf->getSlope() == -1.0))
```
tkuchida: I think this `if` statement should be
```cpp
if (lf && ( std::abs(lf->getSlope() - 1.) < epsilon || ... ))
```
for suitable `epsilon > 0`.

chrisdembia: I'm not sure about this, but I think the exact comparison is okay for a number that can be represented exactly in a floating point representation.

tkuchida: That is technically correct in a closed problem, but here you are relying on the deserializer/wrapping to correctly handle precision. If `lf->getSlope()` can come directly from the deserializer, then perhaps `isEqual()` should be used instead:

https://github.com/opensim-org/opensim-core/blob/71036894bd59422d0ecfb715046317d12e3cd2ac/OpenSim/Common/Property.cpp#L39-L53

which uses an eps of 1e-7 for some reason. Otherwise, if the slope were set to 1e-8 programmatically, a different branch would be taken if the model were used immediately vs. if it were serialized/deserialized first.

tkuchida: > we should also then update the value to be precisely 1 or -1.

Can we put that in CustomJoint's `extendFinalizeFromProperties()`? I think that would avoid treating the joint differently depending on whether the model has been deserialized, and then the equality test on L177 seems ok.

chrisdembia: Is there a reason to believe that deserialization would read a user-specified `1` or `-1` as anything but `double(1.0)` or `double(-1.0)`?

aseth1: I haven't experienced it, but I think there are models with values written like `1.00000000001` AFAICR, which likely occurred during serialization. But perhaps these weren't whole numbers.

tkuchida: I think the best strategy is to go ahead with this PR. I will create a test case that demonstrates the problem.

chrisdembia: @aseth1 I have seen that too, with numbers that cannot be represented exactly as floating point (e.g., 0.1).

tkuchida: Punted :football: to Issue #2066.

aseth1: @chrisdembia @tkuchida I am OK to merge because this fixes the larger issue of a reflected pure rotational coordinate. We can keep #2066 to remind us that serialization can be an issue here.

chrisdembia:
Thanks @aseth1 and @tkuchida for reviewing.

PR #2075, opened 2018-02-05
Make ModelComponent::InvalidScaleFactors protected.

### LINKED ISSUE #2074: Hide ModelComponent::InvalidScaleFactors from SWIG
The `ModelComponent::InvalidScaleFactors` member variable may unnecessarily clutter the SWIG interface, and since it seems unlikely that users need to access this member variable directly, we could consider either hiding it from SWIG or making it protected instead of public.

See https://github.com/opensim-org/opensim-gui/pull/503 for the initial discussion.

chrisdembia:
Fixes #2074

Related: #2006, https://github.com/opensim-org/opensim-gui/pull/503

### Brief summary of changes

- This PR makes `getScaleFactors()` and `InvalidScaleFactors` protected, as @tkuchida suggested in #2006.

### Testing I've completed

- Built Java wrapping and ensured that `InvalidScaleFactors` did not appear.

### DIFF
--- a/OpenSim/Simulation/Model/ModelComponent.h
+++ b/OpenSim/Simulation/Model/ModelComponent.h
@@ -138,6 +138,7 @@ public:
         @see extendPostScale() */
     void postScale(const SimTK::State& s, const ScaleSet& scaleSet);
 
+protected:
     /** Get the scale factors corresponding to the base OpenSim::Body of the
         specified Frame. Returns ModelComponent::InvalidScaleFactors if the
         ScaleSet does not contain scale factors for the base Body. */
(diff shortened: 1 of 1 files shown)


aymanhab:
May I ask why this belongs to ModelComponent altogether?

tkuchida:
> May I ask why this belongs to ModelComponent altogether?

Added in #2006 so a const reference could be returned by `getScaleFactors()` rather than a copy of a `Vec3` (see [`ModelComponent::getScaleFactors()`](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/ModelComponent.cpp#L143)). Open to suggestions for an alternative design :ballot_box:

chrisdembia:
Thanks.

Issue #2028, opened 2017-11-28
Lacking time data on Inverse Kinematics (IK) tool


hkimpara:
Inverse Kinematics (IK) tool is sometimes missing to pick up time from marker data.  
One example, IK analysis for data (0.10 to 0.15 [s]) creates IK outputs for 0.10 to 0.14 [s].  The last one was missing.
I assume this may relate with some numerical bugs.

Here is the example model and associated files.
[https://drive.google.com/open?id=1s0JMDU7eabVBnM2LXl6DkI9EnJtepJ_f](url)

Thank you, Hide

aseth1:
@hkimpara thank you fore reporting this issue. We are unable to access your archive on google drive. Please attach the files here.

hkimpara:
[Setup_IK_R_Leg_link_pedal_002.zip](https://github.com/opensim-org/opensim-core/files/1684260/Setup_IK_R_Leg_link_pedal_002.zip)
I tried to make access to the Google drive again.
Could you try to access following link?
[https://drive.google.com/open?id=1s0JMDU7eabVBnM2LXl6DkI9EnJtepJ_f](url)
Or could you use attached zip file?

aseth1:
The attached zipped works.

tkuchida:
Fixed by #2077

PR #2077, opened 2018-02-07
Fix InverseKinematicsTool to respect user time range

aseth1:
Fixes issue #2028

### Brief summary of changes
The last frame was intermittently being missed due to round-off error. The condition for finding time frames was updated to consider +/- round off error.

### Testing I've completed
Tested on case provided by user in #2028. It now gets the last time-frame at 0.15s.
Added an assert if user final time is not achieved. 

### CHANGELOG.md (choose one)

### DIFF
--- a/OpenSim/Common/TimeSeriesTable.h
+++ b/OpenSim/Common/TimeSeriesTable.h
@@ -246,6 +246,46 @@ public:
         *this = std::move(*table);
     }
 
+    /** Get index of row whose time is nearest/closest to the given value.
+
+    \param time Value to search for.
+    \param restrictToTimeRange  When true -- Exception is thrown if the given
+                                value is out-of-range of the time column.
+                                When false -- If the given value is less than or
+                                equal to the first value in the time column, the
+                                row returned is the first row. If the given value
+                                is greater than or equal to the last value in the
+                                time column, the row returned is the last row.
+                                This operation only returns existing rows and
+                                does not perform any interpolation. Defaults to
+                                'true'.
+
+    \throws TimeOutOfRange If the given value is out-of-range of time column.
+    \throws EmptyTable If the table is empty.                                 */
+    size_t getNearestRowIndexForTime(const double time,
+                                     const bool restrictToTimeRange = true) const {
+        using DT = DataTable_<double, ETY>;
+        const auto& timeCol = DT::getIndependentColumn();
+        OPENSIM_THROW_IF(timeCol.size() == 0,
+            EmptyTable);
+        const SimTK::Real eps = SimTK::SignificantReal;
+        OPENSIM_THROW_IF(restrictToTimeRange &&
+            ((time < timeCol.front() - eps) ||
+            (time > timeCol.back() + eps)),
+            TimeOutOfRange,
+            time, timeCol.front(), timeCol.back());
+
+        auto iter = std::lower_bound(timeCol.begin(), timeCol.end(), time);
+        if (iter == timeCol.end())
+            return timeCol.size() - 1;
+        if (iter == timeCol.begin())
+            return 0;
+        if ((*iter - time) <= (time - *std::prev(iter)))
+            return std::distance(timeCol.begin(), iter);
+        else
+            return std::distance(timeCol.begin(), std::prev(iter));
+    }
+
     /** Get row whose time column is nearest/closest to the given value. 
 
     \param time Value to search for. 
@@ -266,26 +306,8 @@ public:
     getNearestRow(const double& time,
                   const bool restrictToTimeRange = true) const {
         using DT = DataTable_<double, ETY>;
-        const auto& timeCol = DT::getIndependentColumn();
-        OPENSIM_THROW_IF(timeCol.size() == 0,
-                         EmptyTable);
-        const SimTK::Real eps = SimTK::SignificantReal;
-        OPENSIM_THROW_IF(restrictToTimeRange &&
-                         ( (time < timeCol.front()-eps) ||
-                           (time > timeCol.back()+eps) ),
-                         TimeOutOfRange,
-                         time, timeCol.front(), timeCol.back());
-
-        auto iter = std::lower_bound(timeCol.begin(), timeCol.end(), time);
-        if(iter == timeCol.end())
-            return DT::getRowAtIndex(timeCol.size() - 1);
-        if(iter == timeCol.begin())
-            return DT::getRowAtIndex(0);
-        if((*iter - time) <= (time - *std::prev(iter)))
-            return DT::getRowAtIndex(std::distance(timeCol.begin(), iter));
-        else
-            return DT::getRowAtIndex(std::distance(timeCol.begin(),
-                                                   std::prev(iter)));
+        return DT::getRowAtIndex( 
+            getNearestRowIndexForTime(time, restrictToTimeRange) );
     }
 
     /** Get writable reference to row whose time column is nearest/closest to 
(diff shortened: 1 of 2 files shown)


OpenSim/Tools/InverseKinematicsTool.cpp
```diff
@@ -312,7 +312,13 @@ bool InverseKinematicsTool::run()
 
         const clock_t start = clock();
         double dt = 1.0/markersReference.getSamplingFrequency();
-        int Nframes = int((final_time-start_time)/dt)+1;
+        int Nframes = int((final_time-start_time)/dt + SimTK::SignificantReal)+1;
```
tkuchida: Looks right, but how do you know only one `SimTK::SignificantReal` is necessary here? What is the precision of `markersReference.getSamplingFrequency()`? Are `start_time` and `final_time` both passed though the serializer (with precision 1e-6ish)? To avoid any chance of precision/roundoff issues, I usually do something like
```cpp
int Nframes = floor((final_time - start_time)/dt + 0.5) + 1;
```

chrisdembia: @aseth1:

@tkuchida and I are chatting about this. If the IKTool's output is supposed to correspond to frames in the TRC file, then perhaps we should be counting the number of frames within the user-specified time range and avoid any truncation or precision issues:

```cpp
int NFrames = 0;
const auto& times = markersReference.getMarkerTable().getIndependentColumn();
int irow = 0;
while (times[irow] < start_time) {
    ++irow;
    if (irow > times.size()) throw an exception.
}
const int istart = irow;
while (times[irow] <= final_time) {
    ++irow;
    ++NFrames;
}
```

Then the loop below could set time like `s.updTime() = times[istart + i]`.

This alternative should be more consistent with the fact that `MarkersReference` does not interpolate.

aseth1: I waffled on doing that, but I agree. For IK where we are essentially processing discrete frames (and not interpolating) incrementing time is somewhat dubious. Your solution is also independent of the sampling rate and I didn't like relying on it. This is more flexible and enables varying sampling rates. I'll update.

Thanks @chrisdembia and @tkuchida!

aseth1:
@chrisdembia @tkuchida I updated to use indices of the marker data rows by reusing the capability of `TimeSeriesTable` to locate its nearest row by time. I made it a separate method to return just the index, which `TimeSeriesTable` reuses. I figure users might find it useful as well.

tkuchida:
> I'm fine with the asterisk behavior (thinking about my own use cases, processing a frame past my specified final time might be nice), but I am interested in hearing what @tkuchida thinks, and I think the behavior should be documented.

Either `*` or `~` behavior seems fine, but `*` seems best. If a user desires a particular start/end time point, he/she can interpolate in a post-processor if IKTool returns `*`; if `~` are returned, the tool would need to be run a second time with a larger time range. I agree with documenting.

OpenSim/Common/TimeSeriesTable.h
```diff
@@ -250,15 +250,15 @@ class TimeSeriesTable_ : public DataTable_<double, ETY> {
 
     \param time Value to search for.
     \param restrictToTimeRange  When true -- Exception is thrown if the given
-                                value is out-of-range of the time column.
+                                value is out-of-range of the time column. A value
+                                within SimTK::SignifcantReal of a time column
+                                bound is considered to be equal to the bound.
```
aseth1: @tkuchida and @chrisdembia, updated to include the fact that some slop is used to evaluate equality with the time bounds.

OpenSim/Common/TimeSeriesTable.h
```diff
@@ -250,15 +250,15 @@ class TimeSeriesTable_ : public DataTable_<double, ETY> {
...
-                                row returned is the first row. If the given value
-                                is greater than or equal to the last value in the
-                                time column, the row returned is the last row.
-                                This operation only returns existing rows and
-                                does not perform any interpolation. Defaults to
-                                'true'.
+                                index returned is of the first row. If the given
+                                value is greater than or equal to the last value
+                                in the time column, the index of the last row is
+                                returned. Defaults to 'true'.
```
aseth1: @chrisdembia updated to make it clear that index is returned (not row).

OpenSim/Tools/InverseKinematicsTool.cpp
```diff
@@ -144,57 +144,78 @@ void InverseKinematicsTool::setNull()
...
 
-    const double defaultTimeRange[] = {-std::numeric_limits<SimTK::Real>::infinity(), std::numeric_limits<SimTK::Real>::infinity()};
-    _timeRangeProp.setComment("Time range over which the inverse kinematics problem is solved.");
+    const double defaultTimeRange[] =
+        {-std::numeric_limits<SimTK::Real>::infinity(),
+          std::numeric_limits<SimTK::Real>::infinity()};
+    _timeRangeProp.setComment(
+        "The desired time range over which inverse kinematics is solved. "
+        "The closest start and final times from the provided observations "
+        "are used to specify the actual time range to be processed.");
```
aseth1: @tkuchida I updated the Tool comment to reflect that the specified time range is "desired" and how the actual time range is determined.

aseth1:
CI failure on Travis due to: `sh: connect to host web.sourceforge.net port 22: Connection refused` but all test cases passed otherwise.

OpenSim/Common/TimeSeriesTable.h
```diff
@@ -247,6 +246,46 @@ class TimeSeriesTable_ : public DataTable_<double, ETY> {
...
+            ((time < timeCol.front() - eps) ||
+            (time > timeCol.back() + eps)),
+            TimeOutOfRange,
+            time, timeCol.front(), timeCol.back());
+
+        auto iter = std::lower_bound(timeCol.begin(), timeCol.end(), time);
+        if (iter == timeCol.end())
+            return timeCol.size() - 1;
+        if (iter == timeCol.begin())
+            return 0;
```
tkuchida: Not sure, but might generate a warning about implicit type conversion; perhaps `return 0u;` or `static_cast` to `size_t`?

OpenSim/Tools/InverseKinematicsTool.cpp
```diff
@@ -144,57 +144,78 @@ void InverseKinematicsTool::setNull()
  */
 void InverseKinematicsTool::setupProperties()
 {
-    _modelFileNameProp.setComment("Name of the .osim file used to construct a model.");
+    _modelFileNameProp.setComment(
+        "Name of the model (.osim) to use for inverse kinematics.");
```
tkuchida: I think you meant "Name of the model *file* (.osim)…" (to distinguish from the Model's `name` property).

OpenSim/Tools/InverseKinematicsTool.cpp
```diff
@@ -144,57 +144,78 @@ void InverseKinematicsTool::setNull()
...
+        "Markers and coordinates to be considered (tasks) and their weightings. "
+        "The sum of weighted-squared task errors composes the cost function.");
     _ikTaskSetProp.setName("IKTaskSet");
     _propertySet.append(&_ikTaskSetProp);
 
-    _markerFileNameProp.setComment("TRC file (.trc) containing the time history of observations of marker positions.");
+    _markerFileNameProp.setComment(
+        "TRC file (.trc) containing the time history of observations of marker "
+        "positions obtained during a motion capture experiment. Markers in this "
+        "file that have a corresponidng task and model marker are included.");
```
tkuchida: corresponidng -> corresponding

OpenSim/Tools/InverseKinematicsTool.cpp
```diff
@@ -144,57 +144,78 @@ void InverseKinematicsTool::setNull()
...
-    _outputMotionFileNameProp.setComment("Name of the motion file (.mot) to which the results should be written.");
+    _outputMotionFileNameProp.setComment(
+        "Name of the resulting inverse kinematics motion (.mot) file.");
     _outputMotionFileNameProp.setName("output_motion_file");
     _propertySet.append(&_outputMotionFileNameProp);
 
-    _reportMarkerLocationsProp.setComment("Flag indicating whether or not to report model marker locations in ground.");
+    _reportMarkerLocationsProp.setComment(
+        "Flag indicating whether or not to report model marker locations. "
+        "Note, model marker location are expresssed in Ground.");
```
tkuchida: location -> locations
expresssed :snake: -> expressed

OpenSim/Common/TimeSeriesTable.h
tkuchida: Ok. I see `return 0;` and automatically think "return with flag indicating that execution was successful" but yes I guess `0` will be interpreted as `0u` with no type conversion.

aseth1:
Thanks @tkuchida and @chrisdembia

PR #2083, opened 2018-02-10
Fix error cases with Component::findStateVariable().

chrisdembia:
Fixes #2082 

### Brief summary of changes

- This PR fixes the two bugs mentioned in #2082. Invalid state variable paths were silently being accepted, and sometimes the wrong state variable value could be returned.

### Testing I've completed

- Added a test case that failed without this PR. With the bug fix, the tests now pass.

### Looking for feedback on...

- Is the use of ComponentPath okay?

### CHANGELOG.md (choose one)

- no need to update because...this bug was introduced after 3.3.

PR #2084, opened 2018-02-10
traversePathToComponent(): implement using ComponentPath to improve speed and fix bugs

chrisdembia:
Currently, `Component::traversePathToComponent()` makes repeated calls to `getComponentList()` even as it is already traversing the tree of components. This seemed inefficient, so I attempted to reimplement this function by leveraging `ComponentPath`. The test case runs 2x faster now (and much of the test time involves just building the model).

However, some tests now fail, because inputs may now fail to find the output they're supposed to connect to. This is because of another bug, related to #1609 , not a bug in `traversePathToComponent()`.

We can consider merging this once issues surrounding #1609 are fixed.

Note that the new test case I wrote passes even with the current implementation of `traversePathToComponent()`, so this PR is more about speed and a clean implementation rather than about fixing bugs.

Also, this PR should not be merged until #2083 is merged.


OpenSim/Common/Component.cpp
```diff
@@ -737,47 +737,24 @@ const Component::StateVariable* Component::
...
     const StateVariable* found = nullptr;
-    const Component* comp = traversePathToComponent<Component>(prefix);
-
-    if (comp) {
-        found = comp->findStateVariable(varName);
-    }
-
-    // Path not given or could not find it along given path name
-    // Now try complete search.
-    if (!found) {
```
aseth1: You removed this backstop in the case that direct traversal could not locate the Compoent at the same time as modifying the traversal algorithm. While I agree with this change, it is difficult now to know if failures occur due to traversal bugs or because the backstop has been removed   where we were relying on it to find state variables without a complete or valid path. In the past we had no strict rule or a way to enforce it, but state names were unique in our models so we were probably relying on the find in old models/examples.

We should separate the two changes: traversal algorithm speedup (here) and then removing the find step, in which case we should rename this method (`findStateVariable`) or have a separate `getStateVariable(path)`  method.

OpenSim/Common/Component.h
```diff
@@ -695,7 +695,7 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
     bool hasComponent(const std::string& pathname) const {
         static_assert(std::is_base_of<Component, C>::value, 
             "Template parameter 'C' must be derived from Component.");
-        const C* comp = this->template traversePathToComponent<C>(pathname);
+        const C* comp = this->template traversePathToComponent<C>({pathname});
```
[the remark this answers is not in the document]
aseth1: Oh I see this is for implicit conversion from pathname string to to `ComponentPath`. I did not see that you changed  the method signature.

OpenSim/Common/Component.cpp
chrisdembia: @aseth1 that change was actually part of https://github.com/opensim-org/opensim-core/pull/2083, which has already been merged, and in whicih `findStateVariable()` was renamed to `traverseToStateVariable()`. I've updated this PR. Sorry for the confusion. The tests will still fail for this PR, though, because socket connectees can't be found. 

Let me know if you want me to make any changes to how state variables are found (e.g., creating a `getStateVariable()`).

chrisdembia:
@aseth1 thanks for the review. See my response to your [collapsed comment](https://github.com/opensim-org/opensim-core/pull/2084/files/ca839600e4ca966166187f76e43b877c1706cfa9#diff-c36551dbeb71daa4f26cfe5c4474d409).

aseth1:
> @aseth1 that change was actually part of #2083, which has already been merged, and in whicih findStateVariable() was renamed to traverseToStateVariable(). I've updated this PR. Sorry for the confusion. The tests will still fail for this PR, though, because socket connectees can't be found.

Right. And not having the find there did not cause failures, which is good :+1:. I cannot quite see how this PR changes the the finding of connectees. Why would `findAndConnect` now fail in some cases? (Assuming that is the issue with the current CI failures.) 

>Let me know if you want me to make any changes to how state variables are found (e.g., creating a getStateVariable()).

We might want to rename `traverseToStateVariable(pathNameToSv)` to `getStateVariable(pathNameToSv)`. It might be handy to have `findStateVariables(pattern)` that returns a list of ` StateVariable`s that match. But not for this PR.

chrisdembia:
> Why would findAndConnect now fail in some cases? 

I'm not sure; I probably changed the logic of the function. This PR is still WIP, but I was not planning to work on it until issues surrounding #1609 are fixed.

> (Assuming that is the issue with the current CI failures.)

Right now, CI is failing because Sourceforge is down again.

aseth1:
>I'm not sure; I probably changed the logic of the function. This PR is still WIP, but I was not planning to work on it until issues surrounding #1609 are fixed.

#1609 is about warning whenever a Component is not found and I don't quite see why that is relevant to the work on this PR. I think this PR should focus on the traversal  improvements. Then it would no longer be WIP and could get merged.

chrisdembia:
@aseth1 I've updated this PR to restore the workaround that allows the connectee path for an input to be incorrect. This workaround mirrors the workaround that already exists for sockets: 

https://github.com/opensim-org/opensim-core/blob/8dac161ec8bb76164155ba658c8410c2ce9983c2/OpenSim/Common/Component.h#L3003-L3019

It seems that the previous `traversePathToComponent()` implementation had provided this workaround for us.

Also, the previous `traversePathToComponent()` had a bug where `foo.getComponent("foo/bar")` would be equivalent to `foo.getComponent("bar")` (assuming `foo` is named `"foo"`); as a result, I had to fix 2 subtests.

This PR is ready for review again. 

Fixes #2119.

aseth1:
@chrisdembia Clang and cl build errors seem legitimate as well as test of `exampleHopperDeviceAnswers ` on built with gcc.

chrisdembia:
I'm working on fixing the tests.

aseth1:
> I'm working on fixing the tests.

Great. Sorry if I missed that note earlier.

chrisdembia:
> Sorry if I missed that note earlier.

Nope I hadn't said anything.

Bindings/Java/Matlab/Hopper_Device/RunHopperWithDevice_answers.m
```diff
@@ -36,11 +36,18 @@
 %% Connect the device to the hopper.
 % ----------------------------------
 
+% TODO: Add the device to the hopper model.
+% [Step 2, Task A]
+% ANSWER{
+hopper.addComponent(device);
```
chrisdembia: Note: I had to add the device to the hopper before connecting the sockets because, otherwise `connectSocket_*()` generates invalid relative paths. The invalid paths cause `findComponent()` to look up the component by name only, and since there are two components with the name `deviceAttachmentPoint`, `findComponent()` throws an exception.

As has been discussed elsewhere, `connectSocket()` should not attempt to generate a relative path; it should simply store a pointer to the connectee (or we should use UUIDs).

aseth1: I agree that we should only hang onto the reference when we connect. We should compute the path when it is time to serialize.

chrisdembia: > I agree that we should only hang onto the reference when we connect. We should compute the path when it is time to serialize.

To clarify, this is not for this PR, correct?

chrisdembia:
The CI failure is unrelated (similar to #2097).

chrisdembia:
Reedy for review.

Bindings/Java/Matlab/Hopper_Device/RunHopperWithDevice_answers.m
aseth1: Correct!

OpenSim/Common/Test/testComponentInterface.cpp
```diff
@@ -1047,6 +1048,103 @@ void testComponentPathNames()
     top.connect();
 }
 
+void testTraversePathToComponent() {
+    class Base : public Component {
+        OpenSim_DECLARE_ABSTRACT_OBJECT(Base, Component);
+    public:
+        // Allow use of this protected function for testing purposes:
+        using Component::traversePathToComponent;
```
aseth1: Not sure why this would be necessary, since `Component::getComponent<C>` delegates to `traversePathToComponent()` and the not found cases throw a specify exception that you could `ASSERT_THROW` instead of testing for `nullptr`. That seems a bit cleaner to me.

chrisdembia: Done.

chrisdembia:
The CI tests have passed (the one failure is unrelated).

aseth1:
@aymanhab @carmichaelong we need another review on this. Thanks!

OpenSim/Common/Component.h
```diff
@@ -721,6 +721,10 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
      */
     template <class C = Component>
     const C& getComponent(const std::string& pathname) const {
+        return getComponent<C>(ComponentPath(pathname));
+    }
+    template <class C = Component>
```
aymanhab: Is there a chance the constructor/conversion from pathname to ComponentPath fails? If so what's returned? Is there an InvalidComponentPath? If this is wired to user input at some point we need to account for these variations, maybe done already elsewhere.

OpenSim/Common/Component.h
```diff
@@ -2285,70 +2294,50 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
 #endif
 
     template<class C>
-    const C* traversePathToComponent(const std::string& path) const
+    const C* traversePathToComponent(ComponentPath path) const
     {
+        // Get rid of all the ".."'s that are not at the front of the path.
+        path.cleanPath();
+
```
aymanhab: This method takes a ComponentPath (passed by value) and modifies it. Would user expect path to change here? Can the method take const reference, make a local copy to change so that a user of this function doesn't see unexpected side-effects? or am I misreading the implementation.

OpenSim/Common/Path.h
```diff
@@ -101,6 +101,12 @@ class OSIMCOMMON_API Path {
     /// of getInvalidChars()
     bool isLegalPathElement(const std::string& pathElement) const;
 
+    /// Cleans up a path. This includes removing "." and resolving ".." if
+    /// possible (i.e. it will not remove leading ".." but otherwise will
+    /// remove the previous pathElement from _path. This method also checks
+    /// if a path is invalid due to an absolute path starting with "..".
+    void cleanPath();
```
aymanhab: The name cleanPath is a bit ambiguous. Maybe ok while the method is private or protected but in public interface I'd lean to more descriptive name (e.g trim.Something, or resolve or Canonicalize etc.)

OpenSim/Common/Component.h
chrisdembia: @aymanhab making a copy of a const reference achieves the same thing here as pass by value. The user's path is not modified.

aymanhab: Good point @chrisdembia though the signature would be different/more explicit about it.

OpenSim/Common/Component.h
chrisdembia: > Is there a chance the constructor/conversion from pathname to ComponentPath fails?

@aymanhab yes, an `OpenSim::Exception` can be thrown but the current code base still creates a `ComponentPath` from the provided `std::string`; it's just delayed until `traversePathToComponent()`.

> If so what's returned?

The function doesn't return.

> Is there an InvalidComponentPath?

Nope; but something like `ComponentNotFoundOnSpecifiedPath` might get thrown.

> If this is wired to user input at some point we need to account for these variations, maybe done already elsewhere.

By "wired to user input," do you mean in the GUI? I'm not sure what you mean.

aymanhab: @aseth1 I mean if this is expected to be called directly by users through C++ API or Specified in XML for example, then this code should be ready to handle misformed "ComponentPath" String.

chrisdembia: An exception will be thrown if the pathname has invalid characters or is ill-formed, or if a component does not exist at the specified path. Is there another way that invalid user input should be handled?

OpenSim/Common/Path.h
chrisdembia: I chatted with @carmichaelong on this and we settled on the name `trimDotAndDotDotElements()`.

OpenSim/Common/Component.h
aymanhab: Thanks @chrisdembia an exception is perfectly reasonable, and I'm assuming whoever calls this somehow knows that it may throw and can handle it. Thanks for the clearup.

aymanhab:
Thanks @chrisdembia LGTM. Feel free to merge when tests pass.

chrisdembia:
Thanks @aymanhab and @aseth1

PR #2086, opened 2018-02-14
Update CustomActuatorExample

chrisdembia:
This PR updates the CustomActuatorExample to use Sockets. I updated this PR in advance of updating the related Confluence example, in preparation for the user feedback session.

I fixed the display geometry for this example, which was not displaying properly originally.

Before:
screen shot 2018-02-13 at 8 59 07 pm

After:
screen shot 2018-02-13 at 9 00 14 pm

I will hold off on updating the Confluence example until this is reviewed.


OpenSim/Examples/CustomActuatorExample/PistonActuator.h
```diff
@@ -121,38 +112,33 @@ OpenSim_DECLARE_CONCRETE_OBJECT(PistonActuator, ScalarActuator);
...
-    double getStress( const SimTK::State& s ) const override;
+    /** This is the force generated by this actuator divided by its optimal
+     * force. */
+    double getStress(const SimTK::State& s) const override;
     //--------------------------------------------------------------------------
-    // APPLICATION
+    // FORCE INTERFACE
     //--------------------------------------------------------------------------
+    /** Apply the actuator force to frameA and frameB. */
     virtual void computeForce(const SimTK::State& s,
```
aseth1: We should remove the virtual keyword. This `PistonActuator` is not defining the virtual method it is overriding the virtual method  of the base `Force` so `override` keyword is sufficient.

OpenSim/Examples/CustomActuatorExample/ControllableSpring.h
```diff
@@ -51,113 +45,102 @@ OpenSim_DECLARE_CONCRETE_OBJECT(ControllableSpring, PistonActuator);
...
+            pointB_inGround = frameB.findStationLocationInGround(s, pointB);
+        }
+
+        // Find the direction along which the actuator applies its force.
+        SimTK::Vec3 r = pointA_inGround - pointB_inGround;
+        SimTK::UnitVec3 direction(r);
+        double length = r.norm();
+
+        /* find the stiffness.  computeActuation is defined in PistonActuator and
+         * just returns the product of the actuator's control and it's
```
carmichaelong: it's -> its

OpenSim/Examples/CustomActuatorExample/PistonActuator.h
```diff
@@ -121,38 +112,33 @@ OpenSim_DECLARE_CONCRETE_OBJECT(PistonActuator, ScalarActuator);
...
+    /** The force applied by this actuator is its control signal multiplied
+     * by this optimal force. */
     void setOptimalForce(double aOptimalForce);
     double getOptimalForce() const override;
-    // STRESS
-#ifndef SWIG
-    double getStress( const SimTK::State& s ) const override;
+    /** This is the force generated by this actuator divided by its optimal
+     * force. */
+    double getStress(const SimTK::State& s) const override;
```
carmichaelong: It looks like this function uses an absolute value in it, and the comments in this class clearly state what the sign means for the force computed.  I'm not certain if a user would expect this to only return positive values?

carmichaelong:
Just a couple minor comments for clarity (they could be ignored and it would be fine). Otherwise, LGTM.

OpenSim/Examples/CustomActuatorExample/ControllableSpring.h
chrisdembia: Fixed.

OpenSim/Examples/CustomActuatorExample/PistonActuator.h
chrisdembia: It seems like the `fabs()` is unnecessary. However, `PathActuator` also has an `fabs()`:

https://github.com/opensim-org/opensim-core/blob/f721ab592a60b88fde6eb059c3d73d21400879d7/OpenSim/Simulation/Model/PathActuator.cpp#L123-L131

So I modified the comment to mention that there's an absolute value.

chrisdembia:
Thanks for reviewing @carmichaelong . I've tried to address your comments. I also deleted duplicate code files in the `OutputReference` folder. There is no need to put code in the `OutputReference`, because there is no "fill-in-the-black" portion to this example. The code files in the `OutputReference` were out of date but were clearly intended to simply be duplicates of the original source files.

chrisdembia:
This PR is ready for review; the `travis-ci/push` failure can be ignored.

chrisdembia:
Thanksk @aseth1

Issue #1697, opened 2017-05-08
Test and update examples for 4.0 release


jimmyDunne:
## API Examples (from #1297)
- [x] Script to build a template actuators (see PR #541 (Shrinidhi)
- [ ] Convert dynamic walker example (C++ @chrisdembia,  MATLAB @jimmyDunne)
- [x] Convert hopper example (Chris)
- [ ] Convert between Millard and Thelen muscle models  (Ayman)
- [x] Pendulum with markers; output marker locations (James)
- [x] Plot muscle curves (James)
- [x] Show how to use the DataTable's `flatten()` and pack() in scripting (James)
- [ ] Muscle analysis (@aseth1 and @carmichaelong)
- [ ] Point and body kinematics; also report Center of Mass (@aseth1 using OutputReporter)
- [ ] States trajectory for custom post-hoc analysis (maybe trapeziodal rule for integration, Chris) 
- [ ] Update old matlab and python examples (James)
- [x] Dump model's connections, inputs and outputs (done by @tkuchida added to Component)

## API Examples (not from #1297)
- [ ] Working with lists (Matlab/Python) #1465 
- [ ] Sample code for dealing with enums in Matlab #1456 

## API Examples (Confluence) 
- [ ] [Performing a Simulation ](https://simtk-confluence.stanford.edu/display/OpenSim/__40__Performing+a+Simulation)
- [ ] [Creating your own Analysis](https://simtk-confluence.stanford.edu/display/OpenSim/__40__Creating+Your+Own+Analysis)
- [ ] [Creating a Controller](https://simtk-confluence.stanford.edu/display/OpenSim/__40__Creating+a+Controller)
- [ ] [Creating an Optimization](https://simtk-confluence.stanford.edu/display/OpenSim/__40__Creating+an+Optimization)
- [ ] [Creating a Customized Actuator](https://simtk-confluence.stanford.edu/display/OpenSim/__40__Creating+a+Customized+Actuator) @chrisdembia
- [ ] [Creating a Cusomized Muslce Model](https://simtk-confluence.stanford.edu/display/OpenSim/__40__Creating+a+Customized+Muscle+Model)
- [ ] [Getting Started as a Developer](https://simtk-confluence.stanford.edu/display/OpenSim/__40__Getting+Started+as+a+Developer) @chrisdembia https://github.com/opensim-org/opensim-core/issues/2072

## Application Examples (Confluence) 
Introductory
[...]

PR #2091, opened 2018-02-20
Add Python example using_StatesTrajectory.py

chrisdembia:
Part of #1697.

### Brief summary of changes

- This PR adds a Python example of using the StatesTrajectory class to compute joint reaction loads.

The content of the example was decided in discussion with @tkuchida.

### Testing I've completed

- Ran the python example and inspected the output.

### CHANGELOG.md (choose one)

- no need to update because...minor change


chrisdembia:
Thanks @tkuchida .

The CI failure can be ignored; see #2092.

> Only ssuggesstion iss to remove "using_" from the filename, which doessn't sseem to add anything.

I am definitely not satisfied with the filename (I should have listed that under the "Looking for feedback on..." section of the PR description). I didn't want users to think this file holds *the* StatesTrajectory class/utility. I've renamed the file but I'm still interested in other filename suggestions. `analyze_simulation_with_StatesTrajectory.py`?

tkuchida:
> I'm still interested in other filename suggestions

`posthoc_StatesTrajectory_example.py`?

chrisdembia:
Hmm what about `StatesTrajectory_example.py`? Users for whom English is not a first language may not be familiar with "posthoc."

tkuchida:
> Users for whom English is not a first language may not be familiar with "posthoc."

I don't think "posthoc" is too obscure (https://en.wikipedia.org/wiki/Post_hoc_analysis) -- nor would it be the most challenging use of the language (e.g., `equilibrate()`).

Users may not know what a StatesTrajectory is: they have simulation results and want to run an analysis on those data. Ideally, the user would know to look at this example based on its filename.

chrisdembia:
You made a strong case. I went with your suggestion.

Bindings/Python/examples/posthoc_StatesTrajectory_example.py
```diff
@@ -0,0 +1,81 @@
...
+
+joint = osim.PinJoint('joint',
+        model.getGround(), osim.Vec3(0), osim.Vec3(0), 
+        body, osim.Vec3(0, 1.0, 0), osim.Vec3(0))
+joint.updCoordinate().setName('q')
+model.addComponent(joint)
+
+# This reporter will collect states throughout the simulation at intervals of
+# 0.05 seconds.
+rep = osim.StatesTrajectoryReporter()
```
aseth1: minor, but I would prefer `reporter` instead of `rep` especially when you use variable names like `model`, `body`, `joint`, etc...

Bindings/Python/examples/posthoc_StatesTrajectory_example.py
```diff
@@ -0,0 +1,81 @@
...
+
+for itime in range(statesTraj.getSize()):
+    state = statesTraj[itime]
+    time = state.getTime()
+    q = model.getStateVariableValue(state, 'joint/q/value')
+
+    # Calculating joint reactions requires realizing to Acceleration.
+    model.realizeAcceleration(state)
+    # This returns a SpatialVec, which is two Vec3's ([0]: moment, [1]: force).
+    reaction = joint.calcReactionOnParentExpressedInGround(state)
```
aseth1: Might also be a nice opportunity to use the joint's `Output` for the reaction force here.

chrisdembia: My understanding is the outputs are to be used with reporters (or for wiring inputs and outputs between components, in general). If accessing the quantity directly, there's no need to use the output and I can use the underlying method directly. Also, the concrete outputs are harder to access via scripting.

My preference would be to add the following (commented-out code):

```python
# The reaction can also be accessed as an output:
# abstractOutput = joint.getOutput("reaction_on_parent")
# output = osim.OutputSpatialVec.safeDownCast(abstractOutput)
# outputValue = output.getValue(s)
```

chrisdembia:
Ready for review again.

chrisdembia:
@aseth1 could you look at this PR again quickly?

jenhicks:
@chrisdembia Ajay is out of town for ~two weeks. Since Ajay's comment was "Might also be a nice opportunity ..." I think you could go ahead and merge this.

chrisdembia:
Okay thanks.

PR #2113, opened 2018-03-01
Fix crash in ExternalForce when force_expressed_in_body is unspecified.

### LINKED ISSUE #2110: Running RRA with opensim-cmd gives "Bus error: 10"
Running RRA with Subject 2 of Sam Hamner's data (https://simtk.org/frs/download_confirm.php/file/2512/RunningSimulation_simTK.zip?group_id=516) using opensim-cmd gives "Bus error: 10" after loading the model. Note: a few other things with the setup file need to be updated before one gets this error.

Seems related to memory management: https://stackoverflow.com/questions/8716714/bus-error-10-error

Seen on Rachel Jackson's Mac at UEFS.

chrisdembia:
Fixes #2110

### Brief summary of changes

This PR fixes an issue with `ExternalForce` where, if the `force_expressed_in_body`  property is not set, the `_forceExpressedInBody` pointer is set to `Model`, causing a crash (because `Model` is not of type `PhysicalFrame`, and CMC was calling `getTransformInGround()` on `Model`). The cause was this line:

```cpp
_forceExpressedInBody = static_cast<const PhysicalFrame*>(&_model->getComponent(forceExpressedInBodyName));
```

The call `getComponent(forceExpressedInBodyName)`, with `forceExpressedInBodName==“”`, returned a pointer to the model, and `static_cast` doesn’t check type.

In this PR, I removed the static cast and added a test case to ensure that `force_expressed_in_body` is `”ground”` by default, as the documentation states. I made various related changes as well.

I wasn’t able to reproduce the exact same “Bus error: 10” error in #2110, but the crash I fixed occurred at the same part of RRA as #2110, and until other evidence surfaces, I think it’s safe to say this PR fixes #2110.
### Testing I've completed

- Added a test case.
- Ran RRA on Sam Hamner's data (see #2110) and the crash I was getting no longer occurred.

### CHANGELOG.md (choose one)

- no need to update because...minor fix.

### DIFF
--- a/OpenSim/Simulation/Model/ExternalForce.cpp
+++ b/OpenSim/Simulation/Model/ExternalForce.cpp
@@ -88,38 +88,17 @@ ExternalForce::ExternalForce(SimTK::Xml::Element& node) : Super(node)
 {
     setNull();
     constructProperties();
-    updateFromXMLNode(node);
+    updateFromXMLNode(node, -1);
 }
 
 void ExternalForce::setNull()
 {
     setAuthors("Ajay Seth");
-    _dataSource = NULL;
-    _appliedToBody = NULL;
-    _forceExpressedInBody = NULL;
-    _pointExpressedInBody = NULL; 
+    _dataSource = nullptr;
+    _appliedToBody = nullptr;
+    _forceExpressedInBody = nullptr;
+    _pointExpressedInBody = nullptr; 
 }
-
-
-//-----------------------------------------------------------------------------
-// UPDATE FROM XML NODE
-//-----------------------------------------------------------------------------
-//_____________________________________________________________________________
-/**
- * Update this object based on its XML node.
- */
-void ExternalForce::
-updateFromXMLNode(SimTK::Xml::Element& aNode, int versionNumber)
-{
-    // Base class
-    Force::updateFromXMLNode(aNode, versionNumber);
-
-    if( getProperty_force_identifier().empty() 
-        && getProperty_torque_identifier().empty()){
-        throw Exception("ExternalForce:: no force or torque identified.");
-    }
-}   
-
 
 /**
  * Connect properties to local pointers.
@@ -127,12 +106,12 @@ updateFromXMLNode(SimTK::Xml::Element& aNode, int versionNumber)
 void ExternalForce::constructProperties()
 {
     constructProperty_applied_to_body("");
-    constructProperty_force_expressed_in_body("");
-    constructProperty_point_expressed_in_body("");
-    constructProperty_force_identifier("");
-    constructProperty_point_identifier("");
-    constructProperty_torque_identifier("");
-    constructProperty_data_source_name("");
+    constructProperty_force_expressed_in_body("ground");
+    constructProperty_point_expressed_in_body("ground");
+    constructProperty_force_identifier();
+    constructProperty_point_identifier();
+    constructProperty_torque_identifier();
+    constructProperty_data_source_name();
 }
 
 void ExternalForce::setDataSource(const Storage &dataSource)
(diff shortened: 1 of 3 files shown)


OpenSim/Simulation/Model/ExternalForce.cpp
```diff
@@ -88,51 +88,30 @@ ExternalForce::ExternalForce(SimTK::Xml::Element& node) : Super(node)
...
 {
     constructProperty_applied_to_body("");
-    constructProperty_force_expressed_in_body("");
-    constructProperty_point_expressed_in_body("");
-    constructProperty_force_identifier("");
-    constructProperty_point_identifier("");
-    constructProperty_torque_identifier("");
-    constructProperty_data_source_name("");
+    constructProperty_force_expressed_in_body("ground");
+    constructProperty_point_expressed_in_body("ground");
```
aseth1: These should be updated to use a `Socket<PhysicalFrame>`

OpenSim/Simulation/Model/ExternalForce.cpp
```diff
@@ -160,27 +147,35 @@ void ExternalForce::extendConnectToModel(Model& model)
...
+            _appliedToBody = &_model->getComponent<PhysicalFrame>(
+                    appliedToBodyName);
+        if (_model->hasComponent<PhysicalFrame>(forceExpressedInBodyName))
+            _forceExpressedInBody = &_model->getComponent<PhysicalFrame>(
+                    forceExpressedInBodyName);
+
+        if (_specifiesPoint && _model->hasComponent<PhysicalFrame>(
+                    get_point_expressed_in_body()))
+            _pointExpressedInBody = &_model->getComponent<PhysicalFrame>(
+                    get_point_expressed_in_body());
```
aseth1: This finding and connecting to dependent Components (e.g. `PhysicalFrame`s) is precisely what `Socket`s were intended for. If updating we might as well make it a proper Component.

OpenSim/Tools/Test/testExternalLoads.cpp
```diff
@@ -266,3 +262,59 @@ void testExternalLoad()
     // kinematics should match to within integ accuracy
     ASSERT_EQUAL(0.0, norm_err, integ_accuracy);
 }
+
+// Ensure the default values for the ExternalForce properties work as expected.
+void testExternalLoadDefaultProperties() {
```
aseth1: 👍

chrisdembia:
> This finding and connecting to dependent Components (e.g. PhysicalFrames) is precisely what Sockets were intended for. 

@aseth1 I know, and I thought about updating this class to use Sockets; doing so would really reduce the amount of error-prone code in this class. However, I decided against this; I think a larger fix like that would have made sense a few months ago, but not with our current time scale. I don't have time for the larger fix, and revamping the class is likely to introduce new bugs. There are larger issues like the fact that the ExternalLoads XML file would contain relative paths to frames in the Model, which would be confusing to users.* This PR fixes the bug, and provides a test case to help with a future update to this class.

* Maybe we could address this using "private sockets," but that hasn't been tried or documented yet.

aseth1:
@chrisdembia I agree that this fixes the issue and that the using Sockets would be better. I think the latter could be done. I don't think the path names to the Frames were the forces are applied are of too much concern considering those were changes in the `Model` and in it would make it more consistent.  

For now lets get this PR in and make an issue to update `ExternalForce` .

chrisdembia:
@aseth1 I've created issue #2115.

chrisdembia:
Thanks @aseth1

OpenSim/Tools/Test/testExternalLoads.cpp
```diff
@@ -266,3 +262,59 @@ void testExternalLoad()
...
+    // occur when initializing ExternalForce.
+    model.initSystem();
+
+    // ExternalForce throws an exception if it can't find the applied_to_body.
+    xf->set_applied_to_body("");
+    SimTK_TEST_MUST_THROW_EXC(model.initSystem(), OpenSim::Exception);
+    xf->set_applied_to_body(pendBodyName);
+
+    // If force_expressed_in_body can't be found, it's set to ground; no error.
+    xf->set_force_expressed_in_body("nonexistant");
```
tkuchida: nonexistant -> nonexistent

chrisdembia:
@tkuchida I've addressed your comment.

tkuchida:
Great; LGTM. Can be merged IMO. :elephant:

aseth1:
I am going to merge. I don't think we need to rebuild and test for this text change.

chrisdembia:
Thanks @aseth1

Issue #2108, opened 2018-02-26
CMC Results for Strength Of Simulation (leg69) differ from 3.3


apoorvar:
Simulating leg69 in swing using CMC per tutorial instructions works smoothly (and quickly) in 3.3, but not in 4.0. In 4.0, CMC runs if we added reserve actuators. However, the simulation was still really slow (about an order of magnitude longer in 4.0 compared to 3.3).

@jryong  tested on Windows and I ran on OSX and we had similar results.

([https://simtk-confluence.stanford.edu/display/OpenSim/__40__Part+I%3A+Leg+Muscle+Force+Estimation+in+Swing])

tkuchida:
Also see #2089. Presumably, the failure was encountered in #2089 because we need to add reserve actuators but this isn't specified in the instructions.

apoorvar:
You're right, adding reserve actuators does allow the simulation to run in 4.0. However, these reserve actuators were not necessary for the simulation to successfully run in 3.3, so there seems to be something else going on too.

We checked that integration tolerances were the same b/w our attempts in 3.3 and 4.0.

aseth1:
@apoorvar I was able to run on Mac without reserves but I had to lock the pelvis translations. I also got the CMC to perform 10x slower than identical run on 3.3 with dramatically different results.

Oddly running the gait2354_simbody example CMC results in CMC being a bit faster in 4.0. So my  guess is that there is something specific to the model. I will investigate.

tkuchida:
Fixed by #2114. Potential performance issue noted for further investigation in #2129.

Issue #2089, opened 2018-02-15
CMC fails for "Strength of Simulation" example


tkuchida:
See "Strength of Simulation" example, Part I, Section D, first unordered list. Messages window:
<pre>
Updating Model file from 20303 to latest format...
Loaded model leg6dof9musc from file C:\Models\Leg6Dof9Musc\leg6dof9musc.osim
Running tool leg6dof9musc.
No external loads will be applied (external loads file not specified).

 taskSet size = 3

CMC::Warning: CMC cannot compute controls for muscles with muscle controls < 0.02.
The minimum control limit for muscle 'bifemlh_r' has been reset to 0.02.
CMC::Warning: CMC cannot compute controls for muscles with muscle controls < 0.02.
The minimum control limit for muscle 'bifemsh_r' has been reset to 0.02.
CMC::Warning: CMC cannot compute controls for muscles with muscle controls < 0.02.
The minimum control limit for muscle 'glut_max2_r' has been reset to 0.02.
CMC::Warning: CMC cannot compute controls for muscles with muscle controls < 0.02.
The minimum control limit for muscle 'psoas_r' has been reset to 0.02.
CMC::Warning: CMC cannot compute controls for muscles with muscle controls < 0.02.
The minimum control limit for muscle 'rect_fem_r' has been reset to 0.02.
CMC::Warning: CMC cannot compute controls for muscles with muscle controls < 0.02.
The minimum control limit for muscle 'vas_int_r' has been reset to 0.02.
CMC::Warning: CMC cannot compute controls for muscles with muscle controls < 0.02.
The minimum control limit for muscle 'med_gas_r' has been reset to 0.02.
CMC::Warning: CMC cannot compute controls for muscles with muscle controls < 0.02.
The minimum control limit for muscle 'soleus_r' has been reset to 0.02.
CMC::Warning: CMC cannot compute controls for muscles with muscle controls < 0.02.
The minimum control limit for muscle 'tib_ant_r' has been reset to 0.02.

WARN- a desired points file was not specified.

Loading desired kinematics from file C:\Models\Leg6Dof9Musc\Swing\leg69_IK_swing.mot ...
Storage: file=C:\Models\Leg6Dof9Musc\Swing\leg69_IK_swing.mot (nr=31 nc=33)

Low-pass filtering desired kinematics with a cutoff frequency of 6...

MODEL: leg6dof9musc

numStates = 32
numCoordinates = 7
numSpeeds = 7
numActuators = 9
[...]

PR #2114, opened 2018-03-04
Fix CMC slowdown due to reassembly of model within CMCActuatorSubsystem  

aseth1:
Fixes issues: #2108, #2089

### Brief summary of changes
Using `Coordinate::setValue()` in `CMCActuatorSubsystem` ensures consistency of coordinate values in the underlying state, but it was unnecessarily triggering an `assemble()` of the model to satisfy constraints. The assemble() to enforce/satisfy constraints is unnecessary since desired (target) kinematics have already been through `SimbodyEngine::formCompleteStorages()`.  The fix to speed things up is to just `projectQ()` just in case coordinates are perturbed slightly from the desired while tracking.

It was also a potential bug in 3.3 not to do a projection of the kinematics during tracking which could lead to constraint violations and in the case of the patella this could lead to different moment-arms for the quadriceps.

### Testing I've completed
Reran CMC from leg69 during swing with no reserve actuators. This fix (4.0) runs CMC in 27s and 3.3 completed in 32s. The results are also comparable, but not identical:

cmc_40_vs_33

@tkuchida can you verify that running the setup file and model for leg6dof9musc that you used in 4.0 with this branch produces similar answers to 3.3 and in similar (or better) time?

@apoorvar your model has couple constraints for the patella as well. Could you check the performance and answers from CMC on your model resulting from this fix?

### CHANGELOG.md (choose one)
- no need to update because this is a  bug fix.

### DIFF
--- a/OpenSim/Simulation/Model/AbstractTool.cpp
+++ b/OpenSim/Simulation/Model/AbstractTool.cpp
@@ -617,11 +617,11 @@ bool AbstractTool::createExternalLoads( const string& aExternalLoadsFileName, Mo
         // Form complete storage so that the kinematics match the state labels/ordering
         Storage *qStore=NULL;
         Storage *uStore=NULL;
-        aModel.getSimbodyEngine().formCompleteStorages(s, *loadKinematicsForPointTransformation,qStore,uStore);
+        aModel.getSimbodyEngine().formCompleteStorages(s, 
+            *loadKinematicsForPointTransformation,
+            qStore,
+            uStore);
         // qStore should be in radians
-        if (qStore->isInDegrees()){
-            aModel.getSimbodyEngine().convertDegreesToRadians(*qStore);
-        }
         _externalLoads.transformPointsExpressedInGroundToAppliedBodies(*qStore, _ti, _tf);
         delete qStore;
         delete uStore;
--- a/OpenSim/Simulation/Model/AbstractTool.cpp
+++ b/OpenSim/Simulation/Model/AbstractTool.cpp
@@ -621,7 +621,7 @@ bool AbstractTool::createExternalLoads( const string& aExternalLoadsFileName, Mo
             *loadKinematicsForPointTransformation,
             qStore,
             uStore);
-        // qStore should be in radians
+       
         _externalLoads.transformPointsExpressedInGroundToAppliedBodies(*qStore, _ti, _tf);
         delete qStore;
         delete uStore;
--- a/OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp
+++ b/OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp
@@ -148,11 +148,16 @@ void CMCActuatorSubsystemRep::setSpeedTrajectories(FunctionSet *aSet) {
     // controls
     const CoordinateSet& coords = _model->getCoordinateSet();
     for (int i = 0; i < nq; ++i) {
-        // the last argument to setValue is a bool to enforce kinematic constraints
-        // or not. It is being set to true when we set the last coordinate value.
-        coords[i].setValue(mutableCompState, _qWork[i] + _qCorrections[i], i==(nq-1));
+        // the last argument to setValue, a bool to enforce constraints,
+        // is false since values come from a _qSet of splined desired
+        // kinematics formed from formCompleteStorages, which enforces
+        // model constraints.
+        coords[i].setValue(mutableCompState, _qWork[i] + _qCorrections[i], false);
         coords[i].setSpeedValue(mutableCompState, _uWork[i] + _uCorrections[i]);
     }
+    // projectQ is sufficient to satisfy constraints perturbed by _qCorrections
+    _model->getMultibodySystem().projectQ(mutableCompState,
+        getModel()->get_assembly_accuracy());
 
      /* copy  muscle states computed from the actuator system to the muscle states
         for the complete system  then compute forces*/
--- a/OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp
+++ b/OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp
@@ -155,19 +155,21 @@ void CMCActuatorSubsystemRep::setSpeedTrajectories(FunctionSet *aSet) {
         coords[i].setValue(mutableCompState, _qWork[i] + _qCorrections[i], false);
         coords[i].setSpeedValue(mutableCompState, _uWork[i] + _uCorrections[i]);
     }
-    // projectQ is sufficient to satisfy constraints perturbed by _qCorrections
+    // project() to satisfy constraints perturbed by _q/_uCorrections
     _model->getMultibodySystem().projectQ(mutableCompState,
-        getModel()->get_assembly_accuracy()/10);
+        getModel()->get_assembly_accuracy() / 10);
+    _model->getMultibodySystem().projectU(mutableCompState,
+        getModel()->get_assembly_accuracy() / 10);
 
-     /* copy  muscle states computed from the actuator system to the muscle states
-        for the complete system  then compute forces*/
+    /* copy  muscle states computed from the actuator system to the muscle states
+       for the complete system  then compute forces*/
     mutableCompState.updZ() = s.getZ();
     mutableCompState.updTime() = t;
 
-     _model->getMultibodySystem().realize(_completeState, SimTK::Stage::Acceleration);
+    _model->getMultibodySystem().realize(_completeState, SimTK::Stage::Acceleration);
 
-     /* copy 1st derivatives of muscle states from complete system to actuator system */ 
-     s.updZDot() = _completeState.getZDot();
+    /* copy 1st derivatives of muscle states from complete system to actuator system */ 
+    s.updZDot() = _completeState.getZDot();
 
 /*
     cout << "_qWork=" << _qWork << endl;
--- a/OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp
+++ b/OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp
@@ -157,7 +157,7 @@ void CMCActuatorSubsystemRep::setSpeedTrajectories(FunctionSet *aSet) {
     }
     // projectQ is sufficient to satisfy constraints perturbed by _qCorrections
     _model->getMultibodySystem().projectQ(mutableCompState,
-        getModel()->get_assembly_accuracy());
+        getModel()->get_assembly_accuracy()/10);
 
      /* copy  muscle states computed from the actuator system to the muscle states
         for the complete system  then compute forces*/
--- a/OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp
@@ -897,8 +897,7 @@ formCompleteStorages( const SimTK::State& s, const OpenSim::Storage &aQIn,
     
     delete qStore;
     
-    // Compute storage object for simulation
-    // Need to set column labels before converting rad->deg
+    // Set column labels before returning
     rQComplete->setColumnLabels(columnLabels);
     rUComplete->setColumnLabels(speedLabels);
 }
--- a/OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp
@@ -801,23 +801,6 @@ void SimbodyEngine::convertQuaternionsToDirectionCosines(double aQ1, double aQ2,
 //=============================================================================
 // CONFIGURATION
 //=============================================================================
-/**
- * From a potentially partial specification of the generalized coordinates,
- * form a complete storage of the generalized coordinates (q's) and
- * generalized speeds (u's).
- *
- * @param aQIn Storage containing the q's or a subset of the q's.  Rotational
- * q's should be in degrees.
- * @param rQComplete Storage containing all the q's.  If q's were not
- * in aQIn, the values are set to 0.0.  When a q is constrained, its value
- * is altered to be consistent with the constraint.  The caller is responsible
- * for deleting the memory associated with this storage.
- * @param rUComplete Storage containing all the u's.  The generalized speeds
- * are obtained by spline fitting the q's and differentiating the splines.
- * When a u is constrained, its value is altered to be consistent with the
- * constraint.  The caller is responsible for deleting the memory
- * associated with this storage.
- */
 void SimbodyEngine::
 formCompleteStorages( const SimTK::State& s, const OpenSim::Storage &aQIn,
     OpenSim::Storage *&rQComplete,OpenSim::Storage *&rUComplete) const
@@ -883,7 +866,6 @@ formCompleteStorages( const SimTK::State& s, const OpenSim::Storage &aQIn,
     if (aQIn.isInDegrees())
         convertDegreesToRadians(*qStore);
 
-
     // Compute generalized speeds
     GCVSplineSet tempQset(5,qStore);
     std::unique_ptr<Storage> uStore{tempQset.constructStorage(1)};
@@ -919,9 +901,6 @@ formCompleteStorages( const SimTK::State& s, const OpenSim::Storage &aQIn,
     // Need to set column labels before converting rad->deg
     rQComplete->setColumnLabels(columnLabels);
     rUComplete->setColumnLabels(speedLabels);
-    // Convert back to degrees
-    convertRadiansToDegrees(*rQComplete);
-    convertRadiansToDegrees(*rUComplete);
 }
 
 //=============================================================================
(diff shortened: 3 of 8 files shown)


OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp
```diff
@@ -148,11 +148,16 @@ void CMCActuatorSubsystemRep::setSpeedTrajectories(FunctionSet *aSet) {
...
+        // the last argument to setValue, a bool to enforce constraints,
+        // is false since values come from a _qSet of splined desired
+        // kinematics formed from formCompleteStorages, which enforces
+        // model constraints.
+        coords[i].setValue(mutableCompState, _qWork[i] + _qCorrections[i], false);
         coords[i].setSpeedValue(mutableCompState, _uWork[i] + _uCorrections[i]);
     }
+    // projectQ is sufficient to satisfy constraints perturbed by _qCorrections
+    _model->getMultibodySystem().projectQ(mutableCompState,
+        getModel()->get_assembly_accuracy());
```
chrisdembia: I suggest either updating the documentation for the `assembly_accuracy` property or renaming the property to `constraint_accuracy`.

aseth1: The `assembly_accuracy` property name and description is more or less correct. I am using the `assembly_accuracy` as the accuracy for satisfying constraint equations using `projectQ()` instead of setting it arbitrarily. If we consider the constraint solution should be more stringent to achieve the desired coordinate accuracy, then perhaps I should update this to be:  `getModel()->get_assembly_accuracy()/10` since this is how the `SimTK::Assembler` sets the error tolerance on constraints from the user provided accuracy.

chrisdembia: > The assembly_accuracy property name and description is more or less correct. I am using the assembly_accuracy as the accuracy for satisfying constraint equations using projectQ() instead of setting it arbitrarily.

Sounds good. Developers might not realize this accuracy property is used outside of the AssemblySolver, so perhaps a non-doxygen comment could be helpful. But this is a minor point. I'll approve the PR.

If, in the future, we find we need more fine-grained control, perhaps we can add a property to CMCTool for the `projectQ()` accuarcy.

apoorvar:
I downloaded a clean version of the model and simulation data from simtk.org, but was unable to re-run the CMC (command line) simulations for walking or running. I don't think the issue is specific to this change, though, as I also tried running using the current version of master with no success. In all cases, the optimizer failed in the initial 30-ms equilibration.

The error message was:
"SimTK Exception thrown at InteriorPointOptimizer.cpp:264:
  Optimizer failed: Ipopt: Restoration failed (status -2)"

aseth1:
@apoorvar before you dive in, can you make sure all the tests run successfully in your environment. Thanks!

apoorvar:
Yes, all tests passed

tkuchida:
Here is my test log:
- upzip and run latest AppVeyor artifact (OpenSim-de42e7ae-2018-03-10.zip)
- follow instructions for "Strength of Simulation" example, Part I, Section D, first unordered list (as specified in #2089); failed:
<pre>
================================================================
================================================================
Using CMC to track the specified kinematics
Integrating from 0.146667 to 0.616667
Start time = Tue Mar 13 17:40:00 2018
================================================================
CMC.computeControls:  t = 0.146667
SimTK Exception thrown at InteriorPointOptimizer.cpp:264:
  Optimizer failed: Ipopt: Infeasible problem detected (status 2)
OPTIMIZATION FAILED...


CMC.computeControls: ERROR- Optimizer could not find a solution.
Unable to find a feasible solution at time = 0.146667.
Model cannot generate the forces necessary to achieve the target acceleration.
Possible issues: 1. not all model degrees-of-freedom are actuated, 
2. there are tracking tasks for locked coordinates, and/or
3. there are unnecessary control constraints on reserve/residual actuators.
</pre>
- pulled and built `fix_cmc_slowdown_for_models_with_constraints` branch
- ran setup file generated above in build directory; failed:
<pre>
================================================================
================================================================
Using CMC to track the specified kinematics
Integrating from 0.146667 to 0.616667
Start time = Tue Mar 13 17:46:03 2018
================================================================
CMC.computeControls:  t = 0.146667
SimTK Exception thrown at InteriorPointOptimizer.cpp:264:
  Optimizer failed: Ipopt: Infeasible problem detected (status 2)
OPTIMIZATION FAILED...


CMC.computeControls: ERROR- Optimizer could not find a solution.
Unable to find a feasible solution at time = 0.146667.
Model cannot generate the forces necessary to achieve the target acceleration.
Possible issues: 1. not all model degrees-of-freedom are actuated,
2. there are tracking tasks for locked coordinates, and/or
3. there are unnecessary control constraints on reserve/residual actuators.
</pre>

Windows 8.1

aseth1:
@tkuchida thanks for checking. 

How long did you observe for CMC to compute initial states in these cases? In #2089 it reported 28s. 
With the change, I had noticed computing initial states dropped to around 5-6s. I am wondering if at least we are seeing a speedup. It is possible the two issues (slowdown vs not running at all) are not directly related.

tkuchida:
> How long did you observe for CMC to compute initial states in these cases? In #2089 it reported 28s.
With the change, I had noticed computing initial states dropped to around 5-6s. I am wondering if at least we are seeing a speedup. It is possible the two issues (slowdown vs not running at all) are not directly related.

Execution times cannot be compared unless the tests are run on the same hardware.

### Timing results:
Specs: Windows 8.1 Enterprise, 64-bit / Intel Core i7-4930K @ 3.4 GHz / 32 Gb RAM

Using GUI (AppVeyor artifact OpenSim-de42e7ae-2018-03-10.zip):
<pre>
================================================================
Start time   = Thu Mar 15 12:53:20 2018
Finish time  = Thu Mar 15 12:53:44 2018
Elapsed time = 24 seconds.
================================================================
</pre>

Using `fix_cmc_slowdown_for_models_with_constraints` branch:
<pre>
================================================================
Start time   = Thu Mar 15 12:55:17 2018
Finish time  = Thu Mar 15 12:55:22 2018
Elapsed time = 5 seconds.
================================================================
</pre>

Using current `master` branch:
<pre>
================================================================
Start time   = Thu Mar 15 13:08:52 2018
Finish time  = Thu Mar 15 13:08:59 2018
Elapsed time = 7 seconds.
================================================================
</pre>

aseth1:
That's interesting that maser shows the same relative speedup without the fix.  The speedup indicates that the fix is in place. I will have to check why it also fixed the running of CMC on leg6musc9dof for me.
Thanks @tkuchida for the help verifying.

aseth1:
With OpenSim 403b104-2018-03-21 build, I am unable to run CMC to completion when having the pelvis_tilt locked.  If unlock pelis_tilt I can run to completion without reserves. 
screen shot 2018-03-22 at 9 20 57 pm

@aymanhab does this build have this fix? The initialization time is more in line with this fix.

aymanhab:
@aseth1 The build I made yesterday was to include the GUI fix to visualization and the Storage flatten fix. I believe I pulled latest on master on Both first. If you're doubtful I'd download the latest appveyor build from ci today and it sure has all the fixes including today's.

aseth1:
@aymanhab this PR has not been merged, but I am not seeing the slowdown that this PR addresses. I was also unable to run CMC to completion for any condition, and now it works for several cases (pelvis_tilt unlocked).

Most likely, there are a combination of issues, but this PR likely only addresses the slowdown issue but not the failures reported in #2108 and  #2089.

aseth1:
Turns out the problem in #2108 has nothing to do with CMC and everything to do with converting degrees to radians, back to degrees and then back to radians again. In particular for this model, the `knee_patella_r` coordinate is not a rotational coordinate. The automatic system for determining the coordinate's motion type is correct. When the knee_angle is rotated 120degrees the patella only undergoes about a tenth of that rotation w.r.t. the tibia. It is correctly identified as a coupled coordinate. In 3.3 the coordinate is manually labeled rotational which means it was treated exactly as the knee angle. 

This in itself was not the problem, since `formCompleteStorage` correctly computes the couple coordinate value when it forms valid states by enforcing constraints, etc.... since  `knee_patella_r` is never supplied by the user or IK.  The problem is that `SimbodyEngine::formCompleteStorage()` after doing all this forming, assumes the output (complete) storages should be in degrees. This is immediately requires each Tool to convert the complete storages back to radians as valid states. The fix here is simply to return the complete storages in radians and be done. 

The conversion to degrees is likely legacy and/or is needed by the GUI. @aymanhab, if you want to present coordinates in degrees after `formCompleteStorage` for users, I recommend doing the conversion only at presentation time, in the same way you are converting radians to degrees for the coordinate sliders. Maintaining values internally in degrees is going to be error prone.

aseth1:
Update results using the same setup as in 3.3 (w/ pelvis_tilt locked):
image
@tkuchida please let me know if this resolves the issue #2108 for you as well. Thanks!

tkuchida:
> please let me know if this resolves the issue #2108 for you as well.

Did you mean #2089, or did you want #2108 tested on Windows?

aseth1:
> Did you mean #2089, or did you want #2108 tested on Windows?

Yes, #2089 to make sure it works and #2108 to verify that you get the same results as above. It is not a perfect match, but I believe it is close enough.

aseth1:
@chrisdembia @aymanhab this PR is ready for re-review.

OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp
```diff
@@ -148,21 +148,28 @@ void CMCActuatorSubsystemRep::setSpeedTrajectories(FunctionSet *aSet) {
...
-     /* copy  muscle states computed from the actuator system to the muscle states
-        for the complete system  then compute forces*/
+    // project() to satisfy constraints perturbed by _q/_uCorrections
+    _model->getMultibodySystem().projectQ(mutableCompState,
+        getModel()->get_assembly_accuracy() / 10);
+    _model->getMultibodySystem().projectU(mutableCompState,
+        getModel()->get_assembly_accuracy() / 10);
+
+    /* copy  muscle states computed from the actuator system to the muscle states
+       for the complete system  then compute forces*/
```
aymanhab: Is this different from assemble? If so, is it a facility that needs to be exposed at a higher level? Ideally developers of similar tools don't need to go to this low level multibody system calls if the're necessary.

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.h
```diff
@@ -120,8 +120,26 @@ OpenSim_DECLARE_CONCRETE_OBJECT(SimbodyEngine, Object);
...
+    * @param rQComplete Storage containing all the q's.  If q's were not
+    * in aQIn, the values are set to 0.0.  When a q is constrained, its value
+    * is altered to be consistent with the constraint.  The caller is responsible
+    * for deleting the memory associated with this storage. Units are radians.
+    * @param rUComplete Storage containing all the u's.  The generalized speeds
+    * are obtained by spline fitting the q's and differentiating the splines.
+    * When a u is constrained, its value is altered to be consistent with the
+    * constraint.  The caller is responsible for deleting the memory
+    * associated with this storage. Units are radians/s.
+    */
```
aymanhab: Can we explicitly state which storages are in radians, if all then just state that. Did we search all the codebase to make sure the change in the implementation of the functionality doesn't break anything else.

OpenSim/Simulation/Model/AbstractTool.cpp
```diff
@@ -617,11 +617,11 @@ bool AbstractTool::createExternalLoads( const string& aExternalLoadsFileName, Mo
         // Form complete storage so that the kinematics match the state labels/ordering
         Storage *qStore=NULL;
         Storage *uStore=NULL;
-        aModel.getSimbodyEngine().formCompleteStorages(s, *loadKinematicsForPointTransformation,qStore,uStore);
+        aModel.getSimbodyEngine().formCompleteStorages(s, 
+            *loadKinematicsForPointTransformation,
+            qStore,
+            uStore);
```
tkuchida: With the removal of `if (qStore->isInDegrees())`, it looks like a flag is now being ignored. The comment `qStore should be in radians` implies that the units of `qStore` must be checked ("should be"). Something like the following may be appropriate:
```cpp
// qStore is always in radians
if (qStore->isInDegrees()) { qStore->setIsInDegrees(false); }
```
…but that doesn't seem like something the consumer should be doing. Instead, perhaps throw an exception if `qStore->isInDegrees()` (i.e., `qStore` has unexpected property)?

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp
```diff
@@ -919,9 +901,6 @@ formCompleteStorages( const SimTK::State& s, const OpenSim::Storage &aQIn,
     // Need to set column labels before converting rad->deg
```
tkuchida: This comment doesn't make sense anymore because the `convertRadiansToDegrees()` lines have been removed.

tkuchida:
### Test results:
- This PR fixes #2089 (i.e., CMC no longer fails for that example).
- Results are still slightly different (as reported in #2108; see comparison plots, below), but likely acceptable provided there is a reasonable explanation for why the results should differ.
- The title of the PR ("Fix CMC slowdown") suggests an improvement in computation time, but CMC now takes twice as long on my machine vs. 3.3:

| Version | Computing initial states | Tracking specified kinematics |
| --- | --- | --- |
| 3.3 release | 3 seconds | 22 seconds |
| PR 2114 | 7 seconds | 44 seconds |

**Comparison plots**

OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp
aseth1: Projection is not an assembly. It does not need to be exposed at a higher level. This is similar to what the integrator does to satisfy kinematic constraints as time advances. Here, the kinematic state is being slightly altered by` _qCorrections` and `_uCorrections` and the projection is ensuring that the update state still satisfies constraints.

The `CMCActuatorSubsystem` does not integrate the multibody dynamics, just muscle dynamics, and by not doing the projection it can potentially use kinematic states (which it holds constant) that are not valid in estimating muscle controls. When the actual system is integrated, the constraints are satisfied during integration so we do not see constraint, but it is potentially using an internal state that is not accurate. This likely explains why the solutions are: 1) different and 2) slower (but not as slow as when we were doing an assembly).

OpenSim/Simulation/SimbodyEngine/SimbodyEngine.h
aseth1: It explicitly states that the outputs, rQComplete and r UComplete are in radians and radians/s.  The input doesn't matter as long as the `Storage` is `inDegrees` if contains angles in degrees. That has not changed. The only change was to the output Storages which were always returning degrees.

OpenSim/Simulation/Model/AbstractTool.cpp
aseth1: I should have removed the comment. `qStore` is in radians since `formCompleteStorages` by definition now only returns rotational coordinates in radians and radians/s.

I will remove it now.

aseth1:
> This PR fixes #2089 (i.e., CMC no longer fails for that example).

Great! Thanks for the thorough comparison @tkuchida. 

>Results are still slightly different (as reported in #2108; see comparison plot for muscle activations, below), but likely acceptable provided there is a reasonable explanation for why the results should differ.

My understanding from @apoorvar was that the errors were larger than the initial differences presented in this PR.  They now seem inline with previous results but not identical. There are many potentially reasons for differences of this magnitude, but most likely the results differ because we introduced an additional step of projecting kinematic states onto the constraint manifold when being set on the `CMCActuatorSubsystem`. This is the system that integrates the muscle dynamics (over the CMC time window) when solving for controls. We were not doing that before, which means that this system when computing controls could have been using states that violated kinematic constraints. When the the full (Model) system is integrated forward, the constraints are satisfied so this would have been impossible to notice until we changed how states were being set inside `CMCActuatorSubsystem`. 

>The title of the PR ("Fix CMC slowdown") suggests an improvement in computation time, but CMC now takes twice as long on my machine vs. 3.3:

The title references the slowdown due to the reassembly of the model that was taking place. I believe that was causing the bigger slowdown that @apoorvar experienced. I agree we shouldn't make things slower than 3.3 but I believe that is now another issue. The fact of the matter is that `CMCActuatorSubsystem` had a bug in 3.3, in that it could use a state that violated kinematic constraints. 

I opened issue #2129 to track the issue of the resulting slowdown.

tkuchida:
> I opened issue #2129 to track the issue of the resulting slowdown.

Sounds good, @aseth1; this PR is already a great improvement. :lizard:

apoorvar:
I was able to successfully run a CMC simulation of one gait cycle of walking using my model and this fix. The results were comparable to 3.3.

Computation time was really slow, though there's a good chance it's because my VM was running slowly. It took 34 minutes to complete the simulation using this 4.0 build, but also took 20+ minutes using the 3.3 GUI.

Issue #524, opened 2015-07-01
Enforce minimum compiler versions via CMake.


chrisdembia:
To help our users (who skip the README.md) from running into mysterious errors, our CMakeLists should check that the user is using a compiler version that meets our requirements.

http://stackoverflow.com/questions/14933172/how-can-i-add-a-minimum-compiler-version-requisite

PR #2116, opened 2018-03-05
[WIP] Add analysis clone assign cases

aseth1:
Targets conjectured cause of issue [GUI #524](https://github.com/opensim-org/opensim-gui/issues/524)

### Brief summary of changes
Added cases for cloning and assigning each concrete `Analysis` registered in OpenSim.

### Testing I've completed
Both C++ and Java tests pass on Windows 10, but the behavior in the GUI is intermittent on Windows and fatal on Mac, so we expect Travis CI to fail.

@aymanhab how does this differ from the code you use when editting Analyses? We can beef up the Java test to mirror your call sequence until we force the failure.

### CHANGELOG.md (choose one)
- no need to update because this is for a bug fix.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2116)


nickbianco:
Stale PR, closing.

Issue #1578, opened 2017-03-13
Add/remove columns from a DataTable


chrisdembia:
In my current research project, I find myself wanting to remove or add columns from a DataTable. For example, I might load a csv file of all states, and then want to create a table with only joint angles, and so I would want to load the csv file and remove all columns that aren't joint angles. Or, create a new table and add the columns that are the joint angles.

aymanhab:
@chrisdembia I find myself with similar needs in the GUI. Ideally there would be some sort of annotation maintained with the datatable that says what's what (e.g. Columns 1-5 Coordinates, 7-20 Muscle states etc.) in cases where we create the files that are read into datatables at least this should be possible I think.

chrisdembia:
We could have `StatesTrajectory::exportTable()` add those annotations.

klshrinidhi:
As per discussion with Ajay & James, I will add `addColumn` function to `DataTable`.

chrisdembia:
Awesome :) Would `appendColumn()` be more consistent with `appendRow()`, or is there a difference between the two that we would want to signal by using `add` instead?

PR #2117, opened 2018-03-09
Allow building TimeSeriesTables by column from scratch

chrisdembia:
While we can now append columns to a table (#1578), we cannot currently append columns to an empty table. This PR adds the ability to create a table with no columns and append columns to it.

I added this functionality for use in Muscollo.

### Brief summary of changes

- Added a constructor to `TimeSeriesTable` that takes only an independents column, to allow creating a table with no dependent columns.
- I did *not* add a `setIndependentColumn()` function because I did not want to handle the logic of changing the number of rows in a table.
- I removed the restriction that data tables must have at least 1 column.

### Testing I've completed

- Added tests for creating tables without any columns.
- Ensured that empty tables and tables with 0 dependent columns can be read/written.

### CHANGELOG.md (choose one)

- no need to update because...data tables are new in 4.0

### DIFF
--- a/OpenSim/Common/DataTable.h
+++ b/OpenSim/Common/DataTable.h
@@ -646,7 +646,7 @@ public:
     void appendRow(const ETX& indRow, const RowVectorView& depRow) {
         validateRow(_indData.size(), indRow, depRow);
 
-        try {
+        if (_dependentsMetaData.hasKey("labels")) {
             auto& labels =
                     _dependentsMetaData.getValueArrayForKey("labels");
             OPENSIM_THROW_IF(static_cast<unsigned>(depRow.ncol()) !=
@@ -654,8 +654,6 @@ public:
                              IncorrectNumColumns,
                              labels.size(),
                              static_cast<size_t>(depRow.ncol()));
-        } catch(KeyNotFound&) {
-            // No "labels". So no operation.
         }
 
         _indData.push_back(indRow);
--- a/OpenSim/Common/DataTable.h
+++ b/OpenSim/Common/DataTable.h
@@ -646,20 +646,21 @@ public:
     void appendRow(const ETX& indRow, const RowVectorView& depRow) {
         validateRow(_indData.size(), indRow, depRow);
 
+        try {
+            auto& labels =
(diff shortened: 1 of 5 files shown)


OpenSim/Common/Test/testDataTable.cpp
```diff
@@ -701,5 +697,60 @@ int main() {
         std::cout << tableSVec << std::endl;
     }
 
+    {
```
tkuchida: Perhaps add a comment to state what this block is testing (or a `std::cout` as has been done with the other blocks in this file)?

chrisdembia: Done.

chrisdembia:
@tkuchida I've addressed your comment.

tkuchida:
Thanks; LGTM :+1:

OpenSim/Common/DataTable.h
```diff
@@ -646,20 +646,21 @@ class DataTable_ : public AbstractDataTable {
...
         validateRow(_indData.size(), indRow, depRow);
 
+        try {
+            auto& labels =
+                    _dependentsMetaData.getValueArrayForKey("labels");
+            OPENSIM_THROW_IF(static_cast<unsigned>(depRow.ncol()) !=
+                             labels.size(),
+                             IncorrectNumColumns,
+                             labels.size(),
+                             static_cast<size_t>(depRow.ncol()));
```
aseth1: This try catch block could be avoided by checking `hasKey()`. I realize this is how it was before, so could revise in another PR.

chrisdembia: I actually really want this change, so I've added it to this PR.

chrisdembia:
@aseth1 ready for review again.

chrisdembia:
The `travis-ci/push` failure can be ignored.

OpenSim/Common/DataTable.h
aseth1: 👍 Nice!

chrisdembia:
Thank you.

Issue #540, opened 2015-07-17
[travis] Download swig from github instead of sourceforge


chrisdembia:
Thanks @aymanhab

chrisdembia:
Fixed by #544.

PR #2118, opened 2018-03-12
Make clamping of Coordinates invoked only if enforceConstraints is true

aymanhab:
Fixes issue #540 of gui
https://github.com/opensim-org/opensim-gui/issues/540
### Brief summary of changes
Moved the block to enforce Coordinate clamp range so it is executed only when enforceConstraints is true
### Testing I've completed
GUI playback doesn't enforce clamping on bouncing block model of GUI issue above
If documentation needs update?

### CHANGELOG.md (choose one)

### DIFF
--- a/OpenSim/Simulation/SimbodyEngine/Coordinate.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/Coordinate.cpp
@@ -317,7 +317,7 @@ double Coordinate::getValue(const SimTK::State& s) const
  */
 void Coordinate::setValue(SimTK::State& s, double aValue , bool enforceConstraints) const
 {
-    // If the coordinate is clamped, pull aValue into range, only if enforceConstraints
+    // If enforceConstraints is true and coordinate is clamped, clamp aValue into range
     if (enforceConstraints && getClamped(s)) {
         if (aValue < get_range(0))
             aValue = get_range(0);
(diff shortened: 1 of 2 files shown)


chrisdembia:
Does the comment in the header need to be updated?

aseth1:
The comment in the header:
```cpp
    /** Set the value of the Coordinate on to the state.
        Optional flag to enforce the constraints immediately (true by default),
        which can adjust all coordinate values in the state to satisfy model
        constraints. Use getValue(s) to see if/how the value was adjusted to
        satisfy the kinematic constraints. If setting multiple Coordinate values
        consecutively, e.g. in a loop, set the flag to false and then call
        Model::assemble(state) once all Coordinate values have been set.
        Alternatively, use Model::setStateVariableValues() to set all coordinate
        values and their speeds at once followed by Model::assemble(state).
        */
```
Did not mention clamping. The expectation from reading this is that the value should NOT change if enforceConsrtaints is false, which is now the case. Otherwise, we can update to describe that clamping adjust values if enforceConstraints is true. Preference @chrisdembia ?

chrisdembia:
Thanks @aseth1; I think it would be good for the doxygen comment to mention that the clamping only occurs if `enforceConstraints == true`. Other than that, LGTM.

aymanhab:
@aseth1 Please let me know if I need to update the comments in the header based on @chrisdembia review. Thanks

chrisdembia:
I edited the doxygen comment.

aseth1:
Since just a comment update, I am merging.

aymanhab:
Thanks @aseth1 and @chrisdembia :+1:

PR #2122, opened 2018-03-13
Storage constructor accepts version 2 files without nColumns and nRows in the header

### LINKED ISSUE #2120: Storage constructor requires nRows and nColumns in header even for version 2 STO files
@aseth1 said version 2 storage files do not need to have `nRows` or `nColumns` in the header. While the `Storage` constructor can read version 2 storage files, it still requires these fields to be in the header. We should remove the need to find these fields for a version 2 storage file.

Relevant code:

https://github.com/opensim-org/opensim-core/blob/8dac161ec8bb76164155ba658c8410c2ce9983c2/OpenSim/Common/Storage.cpp#L145-L156

chrisdembia:
Fixes #2120.

### Brief summary of changes

- The Storage constructor can now read version 2 files without nColumns and nRows in the header.
- I fixed another issue where "version=1" would appear in the header for version 2 storage files.
- I removed a duplicate message about failing to parse the header.

### Testing I've completed

- Added a test case and inspected the written STO files.

### CHANGELOG.md (choose one)

- no need to update because...version 2 storage files are new to 4.0

### DIFF
--- a/OpenSim/Common/Storage.cpp
+++ b/OpenSim/Common/Storage.cpp
@@ -139,20 +139,33 @@ Storage::Storage(const string &aFileName, bool readHeadersOnly) :
 
     // OPEN FILE
     std::unique_ptr<ifstream> fp{IO::OpenInputFile(aFileName)};
-    if(fp==NULL) throw Exception("Storage: ERROR- failed to open file " + aFileName, __FILE__,__LINE__);
+    OPENSIM_THROW_IF(fp == nullptr, Exception,
+            "Storage: Failed to open file " + aFileName);
 
     int nr=0,nc=0;
-    if (!parseHeaders(*fp, nr, nc)) throw Exception("Storage: ERROR- failed to parse headers of file " + aFileName, __FILE__,__LINE__);
-    cout << "Storage: file=" << aFileName << " (nr=" << nr << " nc=" << nc << ")" << endl;
-    // Motion files from SIMM are in degrees
-    if (_fileVersion < 1 && (0 == aFileName.compare (aFileName.length() - 4, 4, ".mot"))) _inDegrees = true;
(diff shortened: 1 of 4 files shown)


OpenSim/Common/Storage.cpp
```diff
@@ -1294,7 +1307,6 @@ TimeSeriesTable Storage::exportToTable() const {
     TimeSeriesTable table{};
 
     table.addTableMetaData("header", getName());
-    table.addTableMetaData("version", std::to_string(LatestVersion));
     table.addTableMetaData("inDegrees", std::string{_inDegrees ? "yes" : "no"});
```
aymanhab: Not sure why version was removed from meta data, doesn't really matter as long as when we eventually write a file it has the correct version number (so it can be checked if format changes again in the future), can you verify? Thanks

chrisdembia: This line was causing STO files to contain the following in the header:

```
version=1
...
version=2
```

There is no need to store the version number of the STO file in the TimeSeriesTable, because a TimeSeriesTable is not versioned.

chrisdembia:
I fixed the tests and this is ready for review again. The Travis failure is unrelated (cannot upload doxygen to myosin).

OpenSim/Common/Storage.cpp
aymanhab: Thanks for the clarification @chrisdembia, just making sure that if/when the table is eventually written to a file then it would have the correct version number since the table is not the only client of these files. If that holds then LGTM and feel free to merge.

chrisdembia: > just making sure that if/when the table is eventually written to a file then it would have the correct version number since the table is not the only client of these files.

Yes, this is the case

chrisdembia:
Thanks @aymanhab

PR #2123, opened 2018-03-16
Reduce number of digits printed for floats when using Property's toString() with new property system

carmichaelong:
Fixes issue #1804 

### Brief summary of changes
Add a new method (`toStringForDisplay()`) to control the number of digits displayed for types that contain `float`s. I also noticed in 3.3 that properties of type `Vector`, such as `mass_center` used to print with parentheses around the numbers whereas this is no longer true in 4.0. This should restore those parentheses.

### Testing I've completed
Compiles, but no tests exist for this.

- Should there be an API test? If so, where and what? Or should there just be some part of a test that prints out these numbers for visual inspection? I haven't built the GUI before so I don't think that would be the best way for me to test, though it would be good if someone that has that working could test that.
- General structure. For instance, should this new method just be in `Property.h` rather than a virtual method in `AbstractProperty.h`?

### CHANGELOG.md (choose one)
No need to update as this should restore previous behavior.


aseth1:
Some quick tests would be that the string output of properties compares to `std::toString()` value for numbers like 0.1, 9.80655, 0.66666666666666666666 -> 0.666667 (for num precision = 6).

chrisdembia:
@aseth1 you mean `std::to_string()`, correct?

aseth1:
Whoops, yes "std::to_string". Thanks @chrisdembia

carmichaelong:
Any suggestions to where / which test to add these to? Or just have a new test file?

OpenSim/Common/AbstractProperty.h
```diff
@@ -104,6 +104,14 @@ class OSIMCOMMON_API AbstractProperty {
     // TODO: replace this with something more reasonable
```
chrisdembia: @carmichaelong do you think it makes sense to now remove this TODO?

What do you think about adding the `precision` option to `toString()`, with a default argument of 6 or 7? One issue there is that the deprecated properties would ignore the precision, but that is already the case with this implementation.

OpenSim/Common/AbstractProperty.h
```diff
@@ -104,6 +104,14 @@ class OSIMCOMMON_API AbstractProperty {
...
     virtual std::string toString() const = 0;
 
+    /** For relatively simple types, return the current value of this property
+    in a string suitable for displaying to a user in the GUI. Objects just
+    return something like "(Object)". This differs from `toString()` as it has
+    an optional argument, `precision`, for controlling the number of digits
+    printed to string. **/
```
chrisdembia: Consider adding that if this function is not overridden in a derived class (e.g., deprecated properties), this function uses `toString()` and the precision argument is ignored.

OpenSim/Common/Property.h
```diff
@@ -876,6 +882,14 @@ class SimpleProperty : public Property<T> {
         SimTK::writeUnformatted(o, values);
     }
 
+    // This is the default implementation. Specializations for types that
+    // contain floats (e.g., double, Vec3, Vec6, Vector, Transform) are
+    // provided below to control the number of digits shown.
+    void writeSimplePropertyToStreamForDisplay(std::ostream& o, 
+                                               const int precision) const {
+        writeSimplePropertyToStream(o);
```
chrisdembia: Similar to above: the comment should say that by default, the precision is ignored.

OpenSim/Common/Property.h
```diff
@@ -938,6 +952,88 @@ readSimplePropertyFromStream(std::istream& in)
...
 
+// We provide specializations for all types that contain floats to manage
+// the number of digits displayed. We rely on the default setting of how
+// `stringstream`s handle floats, where `precision` is the number of digits
+// printed, and the `stream` automatically determines whether float or
+// exponential notation is used.
+template <> inline void SimpleProperty<double>::
+writeSimplePropertyToStreamForDisplay(std::ostream& o, const int precision) const
+{
+    o << std::setprecision(precision);
```
chrisdembia: I wonder if you could do something like the following:

```cpp
o << std::setprecision(precision);
SimTK::writeUnformatted(o, values);
// restore precision
```

OpenSim/Common/Property.h
```diff
@@ -938,6 +952,88 @@ readSimplePropertyFromStream(std::istream& in)
...
+    for (int i = 0; i < values.size(); ++i) {
+        if (i != 0) o << " ";
+        o << "(" << values[i][0] << " ";
+        o << values[i][1] << " ";
+        o << values[i][2] << " ";
+        o << values[i][3] << " ";
+        o << values[i][4] << " ";
+        o << values[i][5] << ")";
+    }
+}
```
chrisdembia: (from discussion with @aseth1) You might be able to use one implementation for both `Vec3` and `Vec6` (actually, for `Vec<N>`). See the function template specialization of `writeUnformatted()` for `Vec<N>`.

@aseth1 said there is at least one `Vec2` property; so right now, `Vec2` would be displayed in full precision.

OpenSim/Common/Property.h
```diff
@@ -938,6 +952,88 @@ readSimplePropertyFromStream(std::istream& in)
...
+    // Convert array of Transform objects to an array of Vec6 objects.
+    SimTK::Array_<SimTK::Vec6> rotTrans;
+    for (int i = 0; i<values.size(); ++i) {
+        SimTK::Vec6 X6;
+        SimTK::Vec3& angles = X6.updSubVec<3>(0);
+        SimTK::Vec3& pos = X6.updSubVec<3>(3);
+        angles = values[i].R().convertRotationToBodyFixedXYZ();
+        pos = values[i].p();
+        rotTrans.push_back(X6);
+    }
```
chrisdembia: Consider creating a helper function `convertTransformToVec6()` (probably operating on Arrays), to avoid duplicating this code.

aseth1:
Since properties exist for Serialization, creating a new block in `testSerialization` makes sense to me. We can worry about gutting some of the other conditions in there another time.

aymanhab:
Gui never uses properties directly since they are templatized. Only use propertyhelper class to set, get values.

chrisdembia:
@aymanhab thanks. How do you convert from the user's string input to a double, for example? I don't see anything for that [here](https://github.com/opensim-org/opensim-core/blob/94e2271f79fe486da614b7dab199093b580e4f2d/Bindings/Java/OpenSimJNI/OpenSimContext.h).

aymanhab:
Double.parseDouble in java for space separated vectors of doubles or get as a double directly if primitive.

aymanhab:
Note that the same java code worked fine for old style properties in 3.3 so likely not a suspect.

chrisdembia:
Ah thanks; I wonder how parentheses are handled.

aymanhab:
@chrisdembia The class ArrayDouble (which was used to pass around old Vectors) has a method fromString() that was injected through SWIG that handles setting the values from a String and takes care of the parentheses.

chrisdembia:
Ah okay. It seems that we should have an `AbstractProperty::fromString()` to avoid spreading similar functionality over different parts of the code base, but we could leave this change to another time.

OpenSim/Common/Property.h
```diff
@@ -938,6 +952,88 @@ readSimplePropertyFromStream(std::istream& in)
...
+// printed, and the `stream` automatically determines whether float or
+// exponential notation is used.
+template <> inline void SimpleProperty<double>::
+writeSimplePropertyToStreamForDisplay(std::ostream& o, const int precision) const
+{
+    o << std::setprecision(precision);
+    for (int i = 0; i < values.size(); ++i) {
+        if (i != 0) o << " ";
+        o << values[i];
+    }
```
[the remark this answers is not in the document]
carmichaelong: Interesting thought. Right now, the stream is only used temporarily (i.e., created and destroyed within `toStringForDisplay()`) before outputting to a string. Is there a use case you had in mind that might break this down the road?

OpenSim/Common/Property.h
```diff
@@ -938,6 +952,88 @@ readSimplePropertyFromStream(std::istream& in)
...
+}
+
+template <> inline void SimpleProperty<SimTK::Vec3>::
+writeSimplePropertyToStreamForDisplay(std::ostream& o, const int precision) const
+{
+    o << std::setprecision(precision);
+    for (int i = 0; i < values.size(); ++i) {
+        if (i != 0) o << " ";
+        o << "(" << values[i][0] << " ";
```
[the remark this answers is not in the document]
carmichaelong: Yes, it will. So, for a `OneValueProperty` of a `Vec3`, such as `mass_center`, we'll get something like `(0 0 0)` (and this falls in line with 3.3), but then if we have a property that is a list property of many `Vec3`s, we'll get something like `(0 0 0) (1 1 1) (2 2 2)`. Or at least that's the intention here.

OpenSim/Common/Property.h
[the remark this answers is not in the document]
chrisdembia: Ah good point. I had wrongly assumed `o` would be `std::cout`.

>  Is there a use case you had in mind that might break this down the road?

Nope. You can ignore this comment. If we expect `std::cout` to be passed in, then we can revisit.

OpenSim/Common/AbstractProperty.h
carmichaelong: I originally had it like that, but then all of the signatures of the deprecated properties would have to change. I figured this was better than doing that, but I'm happy to do that.

OpenSim/Common/Property.h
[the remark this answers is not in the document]
chrisdembia: It seems to me that you would get `((0 0 0) (1 1 1) (2 2 2))`, unless I'm missing something.

OpenSim/Common/Property.h
carmichaelong: Oh that's interesting.. hadn't thought of that. If that even works, is it still useful to have all of these specializations on the OpenSim side?

OpenSim/Common/Property.h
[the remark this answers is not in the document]
carmichaelong: Oh oops, what you wrote is what I meant to write.

OpenSim/Common/Property.h
carmichaelong: Nevermind, if I understand what's going on on the Simbody side, numbers are converted to `String` first using Simbody's `String` constructor before being passed into the stream, so precision should have no effect. This is how it guarantees that sufficient digits are printed for reproducibility.

chrisdembia: oh, okay. Nevermind then.

OpenSim/Common/Property.h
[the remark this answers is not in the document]
chrisdembia: Okay, as long as that's the expected behavior.

OpenSim/Common/AbstractProperty.h
chrisdembia: Let's see what @aseth1 thinks.

OpenSim/Common/AbstractProperty.h
chrisdembia: I just read about how default arguments work with virtual functions (https://stackoverflow.com/questions/3533589/can-virtual-functions-have-default-parameters). Seems to be a bit of a mess. I'm wondering if it would be best to put the default argument in the base class, and to not provide a default argument in any of the derived classes, because I think people are more likely to call the method from the base class than the derived class.

aseth1:
If the GUI is using a surrogate containers (like DoubleArray) to contain "numbers" (and *"never uses properties"*) then it does not seem like the formatting should be the job of properties. It seems inconsistent to use toString to fill a text box in this case since the numerical values are available. The numerical value should be written to the field and use the [numerical formatters in Swing](https://stackoverflow.com/questions/5498109/question-about-jtextfield-and-text-format-with-numbers).  

Then the user can directly control the display in the GUI for the format/precision they desire and it does affect the values in the API where accuracy (full precision) should be of main concern. Also keeps property behavior simple.

OpenSim/Common/Property.h
carmichaelong: After discussion with @aseth1 and @chrisdembia, the plan above isn't allowed in C++. Will make these functions a free function in the OpenSim namespace, which then will be in parallel with Simbody's `writeUnformatted()` function.

carmichaelong:
@chrisdembia @aseth1 This PR should be ready for another round of reviews.

OpenSim/Common/Property.h
```diff
@@ -694,6 +696,101 @@ Property<T>::getTypeName() const {
...
+template <class T> inline void
+writeSimplePropertyToStreamForDisplay(std::ostream& o,
+    T& v, const int precision)
+{
+    SimTK::writeUnformatted(o, v);
+}
+
+inline void
+writeSimplePropertyToStreamForDisplay(std::ostream& o,
+    const double v, const int precision)
```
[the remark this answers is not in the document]
carmichaelong: When I looked into `SimTK::writeUnformatted()` a bit more, it seemed to be that it's actually a bunch of overloading with some templates to reduce code duplication. I chose to limit the templatization here rather than do all of the full blown stuff `writeUnformatted` does since OpenSim doesn't need all of those types. Might be more future-proof though if more types are covered?

OpenSim/Common/Test/testSerialization.cpp
```diff
@@ -81,8 +81,98 @@ static void dumpObj(const Object& obj, int nSpaces) {
     }
 }
 
+static void testPropertyOutputHelper(const double val, const std::string ans)
```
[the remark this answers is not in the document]
carmichaelong: No reason except that I forgot. The latest commit should add pass by reference for types that are non-basic (I couldn't remember what our style is for passing by reference on basic types?)

carmichaelong:
@chrisdembia All your comments should be incorporated or have a response in the discussion above. I'll have to wait for the builds to double check that I worked the doxygen correctly, but I'd appreciate you double checking the output/code for the doxygen stuff since I'm not comfortable with it yet.

OpenSim/Common/Test/testSerialization.cpp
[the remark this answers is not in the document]
chrisdembia: > I couldn't remember what our style is for passing by reference on basic types?

I don't think we have an agreed-upon style for this.

OpenSim/Common/Property.h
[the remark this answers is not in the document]
chrisdembia: Oh okay; this is fine. I don't think we need to cover all types.

OpenSim/Common/Property.h
```diff
@@ -693,17 +693,16 @@ Property<T>::getTypeName() const {
     return TypeHelper::getTypeName();
 }
 
-// Hide SimpleProperty and ObjectProperty from Doxygen; users don't need
-// to know about these.
-/** @cond **/
-
```
chrisdembia: Oh oops; I didn't realize this was all hidden. In that case, you could leave these classes as hidden and add the comment about the precision back in the base class.

Sorry about that!

chrisdembia:
@carmichaelong I added one more comment (oops). Once that's addressed, I'll approve the PR.

carmichaelong:
@chrisdembia I pushed some of those comments up to AbstractProperty since they are useful for users, and I reverted the change to which methods are hidden. Hopefully I interpreted your suggestions currently and incorporated the correct changes in the last commit?

carmichaelong:
Thanks @chrisdembia. BTW, were you also able to test this with the GUI?

OpenSim/Common/Test/testSerialization.cpp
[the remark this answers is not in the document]
aseth1: I generally like `const T&` for const input arguments. For simple types it rarely matters but gets us in the habit of using references so when it does matter we don't have to fix it.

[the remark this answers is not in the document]
tkuchida: IIUC you don't gain much, if anything, by doing `const double& val` since the address is as long as the data itself—but I like using `const` wherever possible and I think it should be `const std::string& ans` (call by reference for string) on L84.

aseth1:
I just tested this on a local build.
Locally I can edit and see no round-off being added:
image
vs current release:
image

It appears this issue is fixed! Thanks @carmichaelong

carmichaelong:
Hm, something looks a little off for `mass` though? Doesn't look like that particular property has the right number of digits?

aseth1:
> Hm, something looks a little off for mass though? Doesn't look like that particular property has the right number of digits?

That's interesting. I was only checking the values I edited. But is should format all to the default precision, correct?

aseth1:
@carmichaelong I hope you don't mind I made all `const` arguments `const` refs since I had to build it locally anyways.

aseth1:
It looks like `Property<double>` was being formatted differently all along:
(Left is local with this PRs fix and Right is current beta release)
image

aseth1:
Merging but need to investigate the issue of `Property_<double>` being independently formatted.

Issue #609, opened 2015-09-10
Update cmake version for windows vs 2015.


chrisdembia:

chrisdembia:
Fixed by #612.

PR #2125, opened 2018-03-20
Name stateStorage used by AnalyzeTool so it doesn't end up named UNKNOWN

aymanhab:
Fixes issue # opensim-gui/issues/594

### Brief summary of changes
AnalyzeTool internally creates a stateStorage that is loaded by the GUI when user runs the tool. The storage is never given a name. Added a call to name it "states".

### Testing I've completed
Followed instructions in issue and the loaded motion after AnalyzeTool run is now called "states"

Is there a better/more meaningful name? Variety of tools and analyses produce "states".

### CHANGELOG.md (choose one)

- no need to update because, the name of the displayed Storage  is  likely never documented.

### DIFF
--- a/OpenSim/Tools/AnalyzeTool.cpp
+++ b/OpenSim/Tools/AnalyzeTool.cpp
@@ -371,7 +371,7 @@ loadStatesFromFile(SimTK::State& s)
         cout<<"\nLoading states from file "<<_statesFileName<<"."<<endl;
         Storage temp(_statesFileName);
         _statesStore = new Storage();
-        _statesStore->setName("standardized_"+temp.getName()); // Name appears in GUI
+        _statesStore->setName("states"); // Name appears in GUI
         _model->formStateStorage(temp, *_statesStore, true);
     } else {
         if(!_coordinatesFileNameProp.isValidFileName()) 
--- a/OpenSim/Tools/AnalyzeTool.cpp
+++ b/OpenSim/Tools/AnalyzeTool.cpp
@@ -371,6 +371,7 @@ loadStatesFromFile(SimTK::State& s)
         cout<<"\nLoading states from file "<<_statesFileName<<"."<<endl;
         Storage temp(_statesFileName);
         _statesStore = new Storage();
+        _statesStore->setName("states"); // Name appears in GUI
         _model->formStateStorage(temp, *_statesStore, true);
     } else {
         if(!_coordinatesFileNameProp.isValidFileName()) 
--- a/OpenSim/Tools/AnalyzeTool.cpp
+++ b/OpenSim/Tools/AnalyzeTool.cpp
@@ -371,7 +371,7 @@ loadStatesFromFile(SimTK::State& s)
         cout<<"\nLoading states from file "<<_statesFileName<<"."<<endl;
         Storage temp(_statesFileName);
         _statesStore = new Storage();
-        _statesStore->setName("states"); // Name appears in GUI
+        _statesStore->setName("standardized_"+temp.getName()); // Name appears in GUI
         _model->formStateStorage(temp, *_statesStore, true);
     } else {
         if(!_coordinatesFileNameProp.isValidFileName()) 


tkuchida:
> Is there a better/more meaningful name?

Could you use the string at the top of the file being read (`_statesFileName`)?

tkuchida:
> could be misleading since the states we're referring to are not the input states (that the user may have renamed to something more meaningful to him/her) but rather the full states that we assembled on our own.

My point was that, because states are being read from a file, we have some information about where they came from and therefore should be able to use a more descriptive name than the generic "states". I don't know what it means to "assemble" states, but if they differ from the original states, then I think it would be sensible to use a name of the form `<"modified_"|"adjusted_"|"assembled_"><original name>`. I don't know what the prefix string should be because I don't know what "assemble" means in this context (though I don't think you mean a kinematic assembly). I'm also not sure the user needs to know that there's a distinction between the states they provided and whatever OpenSim is doing under the hood. :car:

aymanhab:
@tkuchida By assembled above I mean we added values for unspecified states (based on defaults) and reordered the columns to match the state layout.
Does the user need to know these details? probably not but he/she probably needs to know this altered and is not the states specified as input in case these get saved and used downstream.

tkuchida:
`"standardizedStates_" + <original name>`?

aymanhab:
Will wait for another review before making the one line change in this PR.

aymanhab:
I changed the name in line with @tkuchida suggestion. Need 2 reviews for this one line change, should be quick.

chrisdembia:
Is this redundant with https://github.com/opensim-org/opensim-gui/issues/609?

aymanhab:
@chrisdembia yes in some sense, this inspired @tkuchida to open it, but this limited scope issue was encountered by @jenhicks preparing for the workshop so it's higher priority, much limited scale.

chrisdembia:
Just chatted with @aseth1; neither of us are too excited about `standardized_`; perhaps it would be better to discuss this tomorrow.

tkuchida:
> neither of us are too excited about `standardized_`

`added_values_for_unspecified_states_based_on_defaults_and_reordered_columns_to_match_state_layout_<original name>` is too long :wink:. I don't understand the philosophy of how all this is supposed to work; the name should be chosen by someone who does.

aseth1:
> `added_values_for_unspecified_states_based_on_defaults_and_reordered_columns_to_match_state_layout_<original name>`

is correct, but points to the fact that we are exposing our internals to the user. This is necessary to only for the Tool/Analysis to do its calculations not for the user to play the motion. If we want to load their input motion/states to the Navigator as a courtesy (not sure why we need to do this) then IMHO it should be the that source file loaded and unaltered.

tkuchida:
> If we want to load their input motion/states to the Navigator as a courtesy (not sure why we need to do this) then IMHO it should be the same source file specified by the user and unaltered.

I think I understand. Yes, I agree: the user doesn't need to know that the Analyze Tool needs to "form state storage". The user-specified file could be loaded and shown in the Navigator as if the user had loaded it via *File* -> *Load Motion…*, but I don't think that's desirable since repeated runs of the Analyze Tool (e.g., when adjusting parameters of the analyses) would pollute the Navigator with the same Results/states/Motion.

aymanhab:
I think the rationale for displaying the motions during Tools run was  to confirm to the user what "trajectory", and what muscle states were used by the Tool.  That said, we don't have to load the altered state in the Navigator, and we could show the original state instead. 

Keeping in mind that most motions displayed are results from tools and that most of them call them just "states", I'd lean to call these "states" as well which is better than "UNKNOWN" until we settle on a more elaborate naming convention per issue #609

jenhicks:
Hmmm ... I still get an "UNKNOWN" if I run a muscle analysis.

jenhicks:
Ignore the comment I just deleted ... I was running on my Mac. (I was able to run an Analysis with a pre-made set-up file and add a muscle analysis to it without issue)

aymanhab:
Per discussion with @aseth1 and @jenhicks We'll use "states" for the build nitended for Demo/Workshop as a placeholder until we decide if we should show states altogether when using the AnalyzeTool and if so what name to use.

aymanhab:
@aseth1 and @jenhicks Please let me know if I should merge this. Thanks

jenhicks:
I'm OK with using "states" for now.

aymanhab:
Thanks @chrisdembia

PR #2127, opened 2018-03-21
Retrofit storage to read flattened tables

aseth1:
Fixes issue [GUI #588](https://github.com/opensim-org/opensim-gui/issues/588)

### Brief summary of changes
Update `Storage` constructor that takes a filename to use a `FileAdapter` under the covers, since it is equipped to read all file types we expect to be used with OpenSim.  Logic for keeping the handling of `.mot` and `.sto` files (that are version 0 or 1) through the existing loading code was added as to not break any existing assumptions about old files. 

### Testing I've completed
Updated and added test cases to `testStorage` to exercise the different file types and includes results from the `OutputReporter`. All the tests pass with regressions that typically load `Storage` data that are in the older file versions.

This issue was related to plotting results from `OutputReporter`, so I would like verification that the GUI can now load Storage(s) that support non-scalar types and plot its column data.

### CHANGELOG.md (choose one)
- Did not update since fix should not effect existing behavior/documentation. But, we should now be able to load and plot marker, forces and other data that are in `.trc`, `.c3d`  as well as `.sto` with different element types, which may be a nice feature to document.


OpenSim/Common/Storage.cpp
```diff
@@ -42,20 +42,60 @@
...
+    else if (auto tst = dynamic_cast<const TimeSeriesTable_<SimTK::Vec<10>>*>(table))
+        out = tst->flatten();
+    else if (auto tst = dynamic_cast<const TimeSeriesTable_<SimTK::Vec<11>>*>(table))
+        out = tst->flatten();
+    else if (auto tst = dynamic_cast<const TimeSeriesTable_<SimTK::Vec<12>>*>(table))
+        out = tst->flatten();
+    else if (auto tst = dynamic_cast<const TimeSeriesTable_<SimTK::UnitVec3>*>(table))
+        out = tst->flatten({ "_x", "_y", "_z" });
+    else if (auto tst = dynamic_cast<const TimeSeriesTable_<SimTK::Quaternion>*>(table))
+        out = tst->flatten();
```
aymanhab: How would this scale if a new type of TimeSeriesTable_<x> is introduced?

aseth1: It doesn't scale automatically. We would have to add it here. Given how slow we have been to add new types (at the moment we do not even have Outputs that exercise the types that are handled) I don't think we will be needing to support new types soon . If we need more flexibility we may have to revisit, but I think it is unlikely. It is more likely that `Storage` will be deprecated by then.

aseth1:
>  the result files from Reporter Vec3 can be plotted in GUI build with this change.

Woohoo!

OpenSim/Common/Storage.cpp
```diff
@@ -123,44 +163,74 @@ Storage::Storage(int aCapacity,const string &aName) :
...
+        try {
+            // Try using FileAdpater to read all file types
+            OPENSIM_THROW_IF(readHeadersOnly, Exception,
+                "Cannot read headers only if not a STO file or its "
+                "version is greater than 1.");
+            FileAdapter::OutputTables tables = FileAdapter::readFile(fileName);
+            if (tables.size() > 1) {
+                cout << "Storage: cannot read data files with multiple tables. "
+                    << "Only the first table '" << tables.begin()->first << "' will "
+                    << "loaded as Storage." << endl;
```
chrisdembia: Typo ("will loaded").

OpenSim/Common/Storage.cpp
```diff
@@ -123,44 +163,74 @@ Storage::Storage(int aCapacity,const string &aName) :
...
+            convertTableToStorage(tables.begin()->second.get(), *this);
+            return;
+        }
+        catch (const std::exception& x) {
+            cout << "Storage: FileAdpater failed to read data file.\n"
+                << x.what() << endl;
+            if (isStoFile) 
+                cout << "Reverting to use conventional Storage reader." << endl;
+            else 
+                throw x;
```
chrisdembia: Can you add comments here about the specific scenarios you are expecting? For example, why what type of storage file can be read using the conventional reader that can't be read with the adapters?

OpenSim/Common/Storage.cpp
```diff
@@ -3144,14 +3213,15 @@ bool Storage::parseHeaders(std::ifstream& aStream, int& rNumRows, int& rNumColum
         }
         firstLine=false;
     }
+
+    if (_fileVersion < 1) {
```
chrisdembia: What's the rationale for using `1` instead of `LatestVersion`?

OpenSim/Common/Test/testStorage.cpp
```diff
@@ -27,77 +27,113 @@
...
+        Storage stVersion2("testStorage_version2_short_header.sto");
+        SimTK_TEST_EQ(table.getMatrix(),
+            stVersion2.exportToTable().getMatrix());
+    }
+}
+
+int main() {
+    SimTK_START_TEST("testStorage");
+
+        // Verify the loading scalar Outputs (2) from ..sto into a Storage
```
chrisdembia: It's not clear to me what `(2)` means.

OpenSim/Common/Storage.cpp
```diff
@@ -123,44 +163,74 @@ Storage::Storage(int aCapacity,const string &aName) :
...
 //_____________________________________________________________________________
-/**
+/*
  * Construct an Storage instance from file.
- * This constructor is far from bullet proof.
  *
- * @param aFileName Name of the file from which the Storage is to be
+ * @param fileName Name of the file from which the Storage is to be
  * constructed.
  */
```
chrisdembia: The doxygen comment should say that Storage can now read non-STO files (e.g., C3D), and the doxygen block should be moved to the header.

OpenSim/Common/Storage.cpp
aseth1: I do not anticipate the `FileAdapter` to fail on new files nor do I have any specific expectations that using the conventional reader will help. I only figure if it's a `.sto` file it might be worth trying anyways, in the unexpected case the `FileAdapter` fails.

OpenSim/Common/Storage.cpp
aseth1: Latest version is 2. Only 1 or earlier goes through the conventional pathway (not through the `FileAdpater`).

OpenSim/Common/Test/testStorage.cpp
aseth1: That was a note to expect two Outputs. I will clarify.

OpenSim/Common/Storage.cpp
aseth1: I will add comments to the effect that this ability exists but point readers to using the `FileAdapter` and tables for reading and using the data.

aymanhab:
@chrisdembia can we get your blessing? Thanks

aseth1:
@chrisdembia I updated the comments per your comments. Please feel free to tweak the wording directly.

OpenSim/Common/Storage.h
```diff
@@ -112,10 +112,21 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Storage, StorageInterface);
...
+    elements are flattened to scalar values with more columns.
+    @see DataTable_::faltten()
+    Note, this capability was introduced to support plotting OutputReporter results
+    in the GUI and is not recommended for API users. In addition to only supporting
+    scalar elements, Storage does not preserve the metadata that can be contained by
+    version 2 STO files and read in from C3D and TRC files via their respective
+    FileAdapters. In the case of data files with multiple tables (like C3D files)
+    only the first table read is converted into a Storage.
+    Please use FileAdapter (STOFileAdpater, C3DFileAdapter, ...) and TimeSeriesTable
+    instead, whenever possible. */
```
chrisdembia: I like this documentation.

aymanhab:
Thanks @chrisdembia for the review and @aseth1 for the fix.

PR #2132, opened 2018-03-26
If user asks for no warning on missing states do not warn on count either

aymanhab:
Fixes issue #<issue_number>

### Brief summary of changes
Move warning inside the  warn check

### Testing I've completed
Loaded motions in GUI that are not states and got no warning

### CHANGELOG.md (choose one)

- no need to update because it's internal use


aymanhab:
Unlikely that build failure is related.

aymanhab:
Thanks @tkuchida @aseth1

PR #2134, opened 2018-03-27
Fix bug in traversePathToComponent() for InteractiveHopper

chrisdembia:
### Brief summary of changes

- This PR is an urgent fix for the InteractiveHopper example that we hope to test during the OpenSim meeting on March 27 (tomorrow). Hopefully we can review it early in the morning so we can have a build ready before lunch.
- The `addComponent` line must now come before connecting sockets (I fixed this in #2084 in multiple places but neglected to fix it for the interactive hopper files).
- I fixed the path we were providing to `updComponent()`.
- I fixed a bug where `updComponent("/foo")` would return the root model, even if it wasn't called "foo".

### Testing I've completed

- I ensured that the InteractiveHopper Matlab GUI works as expected when adding a device.

### DIFF
--- a/Bindings/Java/Matlab/Hopper_Device/BuildInteractiveHopperSolution.m
+++ b/Bindings/Java/Matlab/Hopper_Device/BuildInteractiveHopperSolution.m
@@ -164,6 +164,9 @@ for d = 1:length(devices)
         device.printSubcomponentInfo();
     end
     
+    % Add the device to the hopper model.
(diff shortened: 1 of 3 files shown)


chrisdembia:
Thanks @tkuchida

chrisdembia:
The tests pass; whew.

chrisdembia:
Thanks.

PR #2156, opened 2018-04-24
[Dynamic Walking] Script to optimize jump height of hopper

nickbianco:
### Brief summary of changes
This PR adds a short script to optimize the jump height of the hopper model, using pre-existing hopper functions and the Matlab `fmincon` optimizer. 

Some additional minor changes include:
-  A more succinct variable name for whether or not to print model component info.
- Error checking in `BuildInteractiveHopperSolution.m` so a user can't specify control bounds that would lie outside the control window of the GUI.
- Support for reporting Vec3's (workshop participants may want to do this).
- Ignore DAT files
- In `osimTableToStruct`, replace periods with underscores in column names. (Contact force components from a ForceReporter include periods in column names. Workshop participants may also be interested in reporting forces.)

### Testing I've completed
With `fmincon` and an initial guess of 0.25 for each variable (which have bounds [0,1]), I get a solution around 1.44 meters. This isn't as high as it could be, but the optimization is super simple and the solution looks reasonable. 

Should we add anything else or try to improve the solution? (This could probably be addressed in a later PR after the trial run of the workshop.)

### DIFF
--- a/Bindings/Java/Matlab/Hopper_Device/.gitignore
+++ b/Bindings/Java/Matlab/Hopper_Device/.gitignore
@@ -3,3 +3,4 @@ hopper_results.csv
 *.asv
 *.mat
 *.sto
+*.dat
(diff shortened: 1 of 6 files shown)


chrisdembia:
Looks great. +1 for the thorough comments.

PR #529, opened 2015-07-13
Fix more deprecated warnings

sherm1:
Avoid a few more harmless deprecated warnings for `ReferencePtr` user.

PR #2159, opened 2018-04-25
Test editing sockets that specify an invalid topology and reverting in OpenSimContext

aymanhab:
Fixes issue #<issue_number>
Replicates issue opensim-gui#529 with entirely C++/API code to fix/test

If this is not the correct method to edit sockets, please update code or provide feedback.

### CHANGELOG.md (choose one)

- updated...


aseth1:
Updated this PR to address [GUI #529] https://github.com/opensim-org/opensim-gui/issues/529 and make OpenSimContext test case pass.

### Brief summary of changes
1. Implemented `Joint::extendConnectToModel()` to explicitly test that parent and child frames of the `Joint` do not share the same base `PhysicalFrame`. The `SimTK::MultibodyGraphMaker` attempts to nreak the loop, but this a degenerate case. A `PhysicalFrame` connected to itself via a `Joint` in invalid.
2. Added `JointCannotJoinTheSamePhysicalFrame` `Exception` to `Joint` to throw in the above case.
3. Added cases to `testModelInterface` for two different ways you can get connect a `PhysicalFrame` to itself inadvertently by incorrectly using or defining an (or chain of) `OffsetFrame`(s) with a common base frame.

### Testing I've completed
All tests pass including specific cases above.

No doubt these are just some ways of defining a degenerate `Model` and Multibody tree, and it would be good to have more test cases. These could come as additional PRs.

### CHANGELOG.md (choose one)
- no need to update because this is a bug fix with additional test cases. No API change.

Bindings/Java/OpenSimJNI/Test/testContext.cpp
```diff
@@ -222,6 +222,19 @@ int main()
     std::string pathAfterTypeChangeToViaInXML = dTRIlong->updGeometryPath().dump();
     std::cout << pathAfterTypeChangeToViaInXML << endl;
  
+    // edit sockets
+    context->cacheModelAndState();
+    Joint& shoulder = model->updComponent<Joint>("r_shoulder");
+    AbstractSocket& socket = shoulder.updSocket("child_frame");
+    socket.setConnecteeName("ground"); // This create a loop and throws excption at initSystem
```
tkuchida: create -> creates
excption -> exception

Bindings/Java/OpenSimJNI/Test/testContext.cpp
```diff
@@ -222,6 +222,19 @@ int main()
...
+    try {
+        context->restoreStateFromCachedModel();
+    }
+    catch (...) {
+        // undo the change
```
tkuchida: Not something like this to undo the change?
```cpp
const std::string originalConnecteeName = socket.getConnecteeName();
socket.setConnecteeName("ground");
...
socket.setConnecteeName(originalConnecteeName);
```

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -794,7 +782,7 @@ void Model::extendConnectToModel(Model &model)
             Body* child = static_cast<Body*>(mob.getOutboardBodyRef());
             cout << "Body '" << child->getName() << "' not connected by a Joint.\n"
                 << "A FreeJoint will be added to connect it to ground." << endl;
-            Body* ground = static_cast<Body*>(mob.getInboardBodyRef());
+            Ground* ground = static_cast<Ground*>(mob.getInboardBodyRef());
```
tkuchida: Does this change make L788 redundant?

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -33,6 +33,24 @@ namespace OpenSim {
...
+        size_t line,
+        const std::string& func,
+        const std::string& thisName,
+        const std::string& parentName,
+        const std::string& childName,
+        const std::string& baseName) :
+        Exception(file, line, func) {
+        std::string msg = "Joint '" + thisName + "' attempting to connect '" +
+            parentName + "' to '" + childName +
+            "' on the same base PhyscialFrame '" + baseName + "'.";
```
tkuchida: Physcial -> Physical

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -33,6 +33,24 @@ namespace OpenSim {
...
+public:
+    JointCannotJoinTheSamePhysicalFrame(const std::string& file,
+        size_t line,
+        const std::string& func,
+        const std::string& thisName,
+        const std::string& parentName,
+        const std::string& childName,
+        const std::string& baseName) :
+        Exception(file, line, func) {
+        std::string msg = "Joint '" + thisName + "' attempting to connect '" +
```
tkuchida: Suggest:
attempting to -> cannot
on the same... -> because they have the same...

OpenSim/Simulation/Test/testModelInterface.cpp
```diff
@@ -133,11 +134,68 @@ int main() {
...
+                  model.initSystem() );
+
+    // restore the previous frame connected to the shoulder
+    shoulder.connectSocket_child_frame(shoulderOnHumerus);
+    model.initSystem();
+
+    // now do the same but connect the elbow from humerus to humerus
+    auto elbowInHumerus = new PhysicalOffsetFrame("elbow_in_humerus",
+        model.getComponent<Body>("r_humerus"),
+        SimTK::Transform(SimTK::Vec3(0, -0.33, 0.0)));
```
tkuchida: 0 and 0.0?

Bindings/Java/OpenSimJNI/Test/testContext.cpp
aseth1: :+1: Much nicer! I'll update.

OpenSim/Simulation/Model/Model.cpp
aseth1: L788 verifies that the *ground* that was used to build the graph is also the `Ground` of the the `Model`. It likely isn't necessary to check anymore. Is your preference to remove?

aseth1: That is correct.

aseth1:
> do you need to wrap the new JointCannotJoinTheSamePhysicalFrame Exception so the GUI doesn't kerplode 🔴 if it's thrown?

Good question. I think Exception thrown by `OpenSimContext` methods are handled.  @aymanhab?

aymanhab:
OpenSim::Exception subclasses are mapped by swig to JavaIOExceptions and so are handled more gracefully. Lower level exceptions and SimTK Exceptions that are not caught and rethrown would show as red stop sign in corner of app, or possibly crash Matlab.

OpenSim/Simulation/SimbodyEngine/Joint.cpp
```diff
@@ -270,6 +270,24 @@ void Joint::setChildMobilizedBodyIndex(const SimTK::MobilizedBodyIndex index) co
...
 
+void Joint::extendConnectToModel(Model& model) 
+{
+    Super::extendConnectToModel(model);
+
+    auto& parent = updSocket<PhysicalFrame>("parent_frame").getConnectee();
+    auto& child = updSocket<PhysicalFrame>("child_frame").getConnectee();
+
+    // make sure that parent and child frames are connected to the Model
+    // so we can traverse to base frame
```
chrisdembia: The comment says "connected to the Model" but we are not invoking `connectToModel()` on the frames. Would it be more accurate to say 

`finalizeConnections()` invokes `extendConnect()` which calls `extendConnectToModel()`; this doesn't seem like the right place to make these calls. Could you end up in an infinite recursion if the parent or child frame owns the joint? Doesn't this mean `finalizeConnections()` will be called multiple times for these frames (once in the normal call sequence across the hierarchy, and once for each joint for which it is a child or parent frame)?

Bindings/Java/OpenSimJNI/OpenSimContext.cpp
```diff
@@ -46,9 +46,10 @@
 
 namespace OpenSim {
 
-OpenSimContext::OpenSimContext( SimTK::State* s, Model* model ) :
-    _configState(s),
-  _model(model) {}
+OpenSimContext::OpenSimContext( SimTK::State* s, Model* model ) {
+    _configState.reset(s);
+    _model.reset(model);
+}
```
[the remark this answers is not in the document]
aseth1: I had compilation error, otherwise I would not have changed.  I will have to go back to see the error.

Bindings/Java/OpenSimJNI/Test/testContext.cpp
```diff
@@ -222,6 +222,21 @@ int main()
...
+    context->cacheModelAndState();
+    Joint& shoulder = model->updComponent<Joint>("r_shoulder");
+    AbstractSocket& socket = shoulder.updSocket("child_frame");
+    const std::string originalConnecteeName = socket.getConnecteeName();
+    // Create a loop f length 0 that should throw an Exception during initSystem
+    try {
+    socket.setConnecteeName("ground"); 
+        context->restoreStateFromCachedModel();
+    }
+    catch (...) {
```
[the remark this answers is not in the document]
aseth1: I agree in principle, however the `OpenSimContext` is created and edited by @aymanhab to replicate the behavior/conditions of the GUI. I believe this is how the GUI handles exceptions.

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -33,6 +33,25 @@ namespace OpenSim {
...
+        size_t line,
+        const std::string& func,
+        const std::string& thisName,
+        const std::string& parentName,
+        const std::string& childName,
+        const std::string& baseName) :
+        Exception(file, line, func) {
+        std::string msg = "Joint '" + thisName + "' cannot connect '" +
+            parentName + "' to '" + childName +
+            "' because they have the same base PhysicalFrame '" +
```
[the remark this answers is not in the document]
tkuchida: "they" -> "both frames" is potentially misleading. "Joint A cannot connect B to C because both frames have the same base PhysicalFrame" implies that both B and C have the same base PhysicalFrame as some unspecified third thing. You could say "...because these two frames have..." if the objective is to insert the word "frames" so it's clear what the Joint is trying to connect.

Bindings/Java/OpenSimJNI/OpenSimContext.cpp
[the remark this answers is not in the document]
chrisdembia: Okay; ignore my comment then.

Bindings/Java/OpenSimJNI/Test/testContext.cpp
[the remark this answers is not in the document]
chrisdembia: Okay sounds good.

aseth1:
> @aseth1, is there a reason you placed the check in Joint rather than in Model::createMultibodyTree()? I wonder if placing the check in createMultibodyTree() would avoid the possibility of recursion (even if placing the check in Joint feels more appropriate).

My preference is to have Joint check for these Joint related conditions. You are right that if  the parent or child frame owns the joint, it will get stuck in infinite recursion. I updated the PR so that the Joint check is specific to duplicate frames for parent and child and moved the check for the same base frame to `Model::createMultibodyTree()`.

aseth1:
@chrisdembia @tkuchida PR is ready for re-review 🔁

OpenSim/Simulation/Model/Model.h
```diff
@@ -105,6 +105,27 @@ class PhysicalOffsetFramesFormLoop : public Exception {
...
+        const std::string& func,
+        const std::string& thisName,
+        const std::string& parentName,
+        const std::string& childName,
+        const std::string& baseName) :
+        Exception(file, line, func) {
+        std::string msg = "Joint '" + thisName + 
+            "' cannot connect parent frame '" +
+            parentName + "' to child frame '" + childName + "'.\n"
+            "'Parent and child frames have the same base frame.'"
```
tkuchida: I think there are a couple extra punctuation marks (beginning and end), but feel free to ignore (or push with `[ci skip]`):
`"'Parent and child frames have the same base frame.'"` -> `"Parent and child frames have the same base frame '"`

PR #2161, opened 2018-05-01
Prevent repeated throws from subcomponents when invalid properties are set

### LINKED ISSUE #2106: Exception thrown when optimal fiber length is zero (in GUI)
When you (accidently) set optimal fiber length to zero in the GUI you get an (expected) exception. 

However, when you try to correct the error, you can't. Exceptions keep getting thrown and you can't change to a non-zero value

screen shot 2018-02-26 at 12 13 05 pm

tkuchida:
Fixes #2106.

### Brief summary of changes
`Thelen2003Muscle` and `Millard2012EquilibriumMuscle` now call `finalizeFromProperties()` on their subcomponents. If the subcomponent throws an Exception, the Muscle catches it and sets default properties in the subcomponent before re-throwing, which prevents the subcomponent from throwing again when its `finalizeFromProperties()` method is called directly by the MCI.

### Testing I've completed
Added tests to testMuscles; pass locally.

### CHANGELOG.md (choose one)

- no need to update because... fixes bug in new MCI.
- updated...

### DIFF
--- a/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
+++ b/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
@@ -156,6 +156,12 @@ void Millard2012EquilibriumMuscle::extendFinalizeFromProperties()
     penMdl.set_optimal_fiber_length(getOptimalFiberLength());
     penMdl.set_pennation_angle_at_optimal(getPennationAngleAtOptimalFiberLength());
     penMdl.set_maximum_pennation_angle(get_maximum_pennation_angle());
+    try {
+        penMdl.finalizeFromProperties();
+    } catch (const SimTK::Exception::Base& ex) {
+        penMdl.setDefaultProperties(); //avoid throwing again
+        OPENSIM_THROW(Exception, ex.getMessage());
+    }
 
     // Set properties of activation dynamics model subcomponent. Values of
     // activation_time_constant, deactivation_time_constant, and
@@ -167,14 +173,18 @@ void Millard2012EquilibriumMuscle::extendFinalizeFromProperties()
         actMdl.set_activation_time_constant(get_activation_time_constant());
         actMdl.set_deactivation_time_constant(get_deactivation_time_constant());
         actMdl.set_minimum_activation(get_minimum_activation());
+        try {
+            actMdl.finalizeFromProperties();
+        } catch (const SimTK::Exception::Base& ex) {
+            actMdl.setDefaultProperties(); //avoid throwing again
+            OPENSIM_THROW(Exception, ex.getMessage());
+        }
     }
 
     // Compute and store values that are used for clamping the fiber length.
     const double minActiveFiberLength = falCurve.getMinActiveFiberLength()
                                         * getOptimalFiberLength();
-    // Must update the pennation model's internal data members before requesting
-    // the minimum fiber length.
-    penMdl.finalizeFromProperties();
+    // The pennation model's internal data members are updated above.
     const double minPennatedFiberLength = penMdl.getMinimumFiberLength();
     m_minimumFiberLength = max(SimTK::SignificantReal,
         max(minActiveFiberLength, minPennatedFiberLength));
--- a/OpenSim/Actuators/MuscleFirstOrderActivationDynamicModel.cpp
+++ b/OpenSim/Actuators/MuscleFirstOrderActivationDynamicModel.cpp
@@ -65,6 +65,13 @@ void MuscleFirstOrderActivationDynamicModel::constructProperties()
     constructProperty_minimum_activation(0.01);
 }
 
+void MuscleFirstOrderActivationDynamicModel::setDefaultProperties()
+{
+    set_activation_time_constant(0.010);
+    set_deactivation_time_constant(0.040);
+    set_minimum_activation(0.01);
+}
+
 //==============================================================================
 // SERVICES
 //==============================================================================
--- a/OpenSim/Actuators/MuscleFirstOrderActivationDynamicModel.h
+++ b/OpenSim/Actuators/MuscleFirstOrderActivationDynamicModel.h
@@ -118,6 +118,11 @@ protected:
 private:
     void setNull();
     void constructProperties();
+    void setDefaultProperties();
+
+    // These classes are friends because they call setDefaultProperties().
+    friend class Thelen2003Muscle;
+    friend class Millard2012EquilibriumMuscle;
 
 };
 
--- a/OpenSim/Actuators/MuscleFixedWidthPennationModel.cpp
+++ b/OpenSim/Actuators/MuscleFixedWidthPennationModel.cpp
@@ -42,6 +42,13 @@ void MuscleFixedWidthPennationModel::constructProperties()
     constructProperty_maximum_pennation_angle(acos(0.1));
 }
 
+void MuscleFixedWidthPennationModel::setDefaultProperties()
+{
+    set_optimal_fiber_length(1.0);
+    set_pennation_angle_at_optimal(0.0);
+    set_maximum_pennation_angle(acos(0.1));
+}
+
 MuscleFixedWidthPennationModel::MuscleFixedWidthPennationModel()
 {
     setNull();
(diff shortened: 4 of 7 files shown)


tkuchida:
AppVeyor failures due to `testCommandLineInterface` and Travis failures because it can't find :nose:.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -156,6 +156,12 @@ void Millard2012EquilibriumMuscle::extendFinalizeFromProperties()
     penMdl.set_optimal_fiber_length(getOptimalFiberLength());
     penMdl.set_pennation_angle_at_optimal(getPennationAngleAtOptimalFiberLength());
     penMdl.set_maximum_pennation_angle(get_maximum_pennation_angle());
+    try {
+        penMdl.finalizeFromProperties();
+    } catch (const SimTK::Exception::Base& ex) {
+        penMdl.setDefaultProperties(); //avoid throwing again
```
aseth1: Ideally, we would restore the pendMdl to its previous (working) state and not all the way back to the default values. Assuming that the pendMdl is initialized with default values the effect would be the same until the user define other valid values and then entered an invalid value for one property, which would then reset all the property values.

I liked your idea yesterday of keeping a clone of the subcomponent (pendMdl) and the restoring using assignment which will copy property values, i.e. `pendMdl = *clonePendMdl;`

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -156,6 +156,12 @@ void Millard2012EquilibriumMuscle::extendFinalizeFromProperties()
     penMdl.set_optimal_fiber_length(getOptimalFiberLength());
     penMdl.set_pennation_angle_at_optimal(getPennationAngleAtOptimalFiberLength());
     penMdl.set_maximum_pennation_angle(get_maximum_pennation_angle());
+    try {
+        penMdl.finalizeFromProperties();
+    } catch (const SimTK::Exception::Base& ex) {
```
aseth1: ideally would only trap just the `InvalidPropertyValue` exception.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -156,6 +156,12 @@ void Millard2012EquilibriumMuscle::extendFinalizeFromProperties()
...
+        penMdl.setDefaultProperties(); //avoid throwing again
+        OPENSIM_THROW(Exception, ex.getMessage());
```
aseth1: Also you can rethrow the existing exception with just a `throw;` without recreating an exception with the same message.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
tkuchida: > Ideally, we would restore the pendMdl to its previous (working) state and not all the way back to the default values. Assuming that the pendMdl is initialized with default values the effect would be the same until the user define other valid values and then entered an invalid value for one property, which would then reset all the property values.

`Model::finalizeFromProperties()` won't complete successfully until the Muscle's `optimal_fiber_length` property has been set to a valid quantity. The underlying pennation model subcomponent is a slave of the Muscle and won't be used for anything until a valid optimal fiber length has been set. Thus, the state of the subcomponent is irrelevant and here I'm simply setting its properties to something "safe" to prevent it from throwing. I'll redo it the other way.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
tkuchida: The pennation model throws `SimTK::Exception::ErrorCheck` and `SimTK::Exception::ValueOutOfRange`, so catching `InvalidPropertyValue` won't fix #2106.

tkuchida:
I'll just start over rather than trying to undo all my changes.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
aseth1: Good point. Thanks for clarifying. You can just add a comment to that effect without redoing.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -156,6 +156,12 @@ void Millard2012EquilibriumMuscle::extendFinalizeFromProperties()
     penMdl.set_optimal_fiber_length(getOptimalFiberLength());
     penMdl.set_pennation_angle_at_optimal(getPennationAngleAtOptimalFiberLength());
     penMdl.set_maximum_pennation_angle(get_maximum_pennation_angle());
+    try {
```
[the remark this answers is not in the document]
tkuchida: I will write a longer story here and create an Issue regarding the MCI design.

OpenSim/Actuators/MuscleFirstOrderActivationDynamicModel.cpp
```diff
@@ -65,6 +65,13 @@ void MuscleFirstOrderActivationDynamicModel::constructProperties()
     constructProperty_minimum_activation(0.01);
 }
 
+void MuscleFirstOrderActivationDynamicModel::setDefaultProperties()
+{
+    set_activation_time_constant(0.010);
+    set_deactivation_time_constant(0.040);
+    set_minimum_activation(0.01);
```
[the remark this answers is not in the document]
tkuchida: Yes, now irr:elephant:.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
aseth1: Disregard this comment now.

aseth1:
@tkuchida I don't think you needed to close the PR. Your clarifications were helpful.

tkuchida:
> I don't think you needed to close the PR. Your clarifications were helpful.

It's better to avoid introducing ~~`setDefaultProperties()`~~ `setPropertiesToDefault()` methods and it's faster to just start over.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
tkuchida: I'll catch internally, but the code is a bit ugly: I need to pass the name of the property when throwing `InvalidPropertyValue`, so each property needs to be in a separate try/catch.

aseth1:
> A cleaner long-term solution would be to introduce extendFinalizeFromPropertiesAfterSubcomponents() or something similar (similar toextendAddToSystemAfterSubcomponents()). I don't expect such a large change at this point in the development cycle, but I think it should be mentioned in comments that the current solution is not a great long-term solution, and that an Issue should be created for the long-term solution.

While agree that this would be cleaner I disagree with the notion that this solution is simply a temporary hack/workaround. It is a legitimate/correct approach to the general problem of handling your components without exposing them to the outside. An equally viable/clean solution is to expose the internal components and remove the properties on the muscle. 

I think too much discussion based on opinion in the documentation/issues will in fact be more confusing  and counter productive. Adding a bit more description for why the check is necessary seems sufficient to me. I really do not want to see an MCI issue on this.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
aseth1: Then don't do it. It would be cleaner/better to throw `InvalidPropertyValue` in the `MuscleFixedWidthPennationModel::extendFinalizeFromProperties()` (and in all subcomponents in general) regarding invalid property values.

chrisdembia:
>  It is a legitimate/correct approach to the general problem of handling your components without exposing them to the outside. 

I don't think it's desirable that we call `extendFinalizeFromProperties()` on a subcomponent multiple times for one call of `finalizeFromProperties()` on the model, especially because, in the first call, the property values in the subcomponent are not the ones that will actually be used.

> An equally viable/clean solution is to expose the internal components and remove the properties on the muscle.

Agreed. I think the issue in this case is that some of the properties are in base classes of the muscle.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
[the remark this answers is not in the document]
tkuchida: Created #2162.

Issue #2162, opened 2018-05-02
Component::finalizeFromProperties() is called on Muscle subcomponents twice


tkuchida:
Both `Thelen2003Muscle` and `Millard2012EquilibriumMuscle` have two subcomponents: `MuscleFixedWidthPennationModel` and `MuscleFirstOrderActivationDynamicModel`. The properties of these subcomponents correspond to properties of the Muscle, and the Muscle sets these properties to match its own properties in `${MUSCLE_CLASS}::extendFinalizeFromProperties()`. The Muscle then calls `finalizeFromProperties()` on its subcomponents to check for invalid properties (and compute internal data members); if invalid properties are encountered, the Muscle must restore valid properties in the subcomponent before re-throwing the subcomponent's exception because, otherwise, the subcomponent will throw again when its `finalizeFromProperties()` method is called directly. A cleaner approach (and one that would avoid calling `finalizeFromProperties()` on subcomponents twice) would be to avoid duplicating properties in the Muscle and its subcomponents.

PR #2163, opened 2018-05-02
Prevent repeated throws from subcomponents when invalid Muscle properties are set

### LINKED ISSUE #2106: Exception thrown when optimal fiber length is zero (in GUI)

tkuchida:
Fixes #2106. Redo of PR #2161 using different design.

### Brief summary of changes
Affects `Thelen2003Muscle` and `Millard2012EquilibriumMuscle`, each of which has two subcomponents: `MuscleFixedWidthPennationModel` and `MuscleFirstOrderActivationDynamicModel`.
- Muscles propagate their properties down to their subcomponents and call `finalizeFromProperties()` on each subcomponent; if an exception is thrown, the Muscle restores the subcomponent's properties to their previous (though now out-of-sync) values and re-throws the exception. Underlying issue noted in #2162 for post-4.0.
- Replaced property checks in subcomponents with `OPENSIM_THROW_IF_FRMOBJ` macro that throws only `InvalidPropertyValue` so that the Muscles can catch this exception (rather than potentially catching and misinterpreting a SimTK exception caught with `SimTK::Exception::Base`). Updated corresponding tests.
- Added informative messages when throwing `InvalidPropertyValue` exception.
- Fixed inconsistency in Millard2012EquilibriumMuscle property checking.

### Testing I've completed
Added tests to testMuscles. All tests pass locally.

### CHANGELOG.md (choose one)

- no need to update because... fixes MCI-related bug.
- updated...

### DIFF
--- a/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
+++ b/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
@@ -164,7 +164,7 @@ void Millard2012EquilibriumMuscle::extendFinalizeFromProperties()
     penMdl.set_maximum_pennation_angle(get_maximum_pennation_angle());
     try {
         penMdl.finalizeFromProperties();
-    } catch (const InvalidPropertyValue& ex) {
+    } catch (const InvalidPropertyValue&) {
         penMdl = *penMdlCopy;
         throw;
     }
@@ -180,7 +180,7 @@ void Millard2012EquilibriumMuscle::extendFinalizeFromProperties()
         actMdl.set_minimum_activation(get_minimum_activation());
         try {
             actMdl.finalizeFromProperties();
-        } catch (const InvalidPropertyValue& ex) {
+        } catch (const InvalidPropertyValue&) {
             actMdl = *actMdlCopy;
             throw;
         }
--- a/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
+++ b/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
@@ -125,9 +125,9 @@ void Millard2012EquilibriumMuscle::extendFinalizeFromProperties()
             InvalidPropertyValue, getProperty_minimum_activation().getName(),
             "Minimum activation cannot be less than zero");
 
-        OPENSIM_THROW_IF_FRMOBJ(getMinControl() < 0.0,
+        OPENSIM_THROW_IF_FRMOBJ(getMinControl() < get_minimum_activation(),
             InvalidPropertyValue, getProperty_min_control().getName(),
-            "Minimum control cannot be less than zero");
+            "Minimum control cannot be less than minimum activation");
 
         set_minimum_activation(clamp(0, get_minimum_activation(), 1));
         falCurve.setMinValue(0.0);
--- a/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
+++ b/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
@@ -147,34 +147,44 @@ void Millard2012EquilibriumMuscle::extendFinalizeFromProperties()
     fpeCurve.ensureCurveUpToDate();
     fseCurve.ensureCurveUpToDate();
 
-    // Set properties of pennation model subcomponent. Values of
-    // optimal_fiber_length, pennation_angle_at_optimal, and
-    // maximum_pennation_angle are checked by
-    // MuscleFixedWidthPennationModel::extendFinalizeFromProperties().
+    // Propagate properties down to pennation model subcomponent. If any of the
+    // new property values are invalid, restore the subcomponent's current
+    // property values (to avoid throwing again when the subcomponent's
+    // extendFinalizeFromProperties() method is called directly) and then
+    // re-throw the exception thrown by the subcomponent.
     auto& penMdl =
         updMemberSubcomponent<MuscleFixedWidthPennationModel>(penMdlIdx);
+    MuscleFixedWidthPennationModel* penMdlCopy = penMdl.clone();
     penMdl.set_optimal_fiber_length(getOptimalFiberLength());
     penMdl.set_pennation_angle_at_optimal(getPennationAngleAtOptimalFiberLength());
     penMdl.set_maximum_pennation_angle(get_maximum_pennation_angle());
+    try {
+        penMdl.finalizeFromProperties();
+    } catch (const InvalidPropertyValue& ex) {
+        penMdl = *penMdlCopy;
+        throw(ex);
+    }
 
-    // Set properties of activation dynamics model subcomponent. Values of
-    // activation_time_constant, deactivation_time_constant, and
-    // minimum_activation are checked by
-    // MuscleFirstOrderActivationDynamicModel::extendFinalizeFromProperties().
+    // Propagate properties down to activation dynamics model subcomponent.
+    // Handle invalid properties as above for pennation model.
     if (!get_ignore_activation_dynamics()) {
         auto& actMdl =
             updMemberSubcomponent<MuscleFirstOrderActivationDynamicModel>(actMdlIdx);
+        MuscleFirstOrderActivationDynamicModel* actMdlCopy = actMdl.clone();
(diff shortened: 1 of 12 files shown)


tkuchida:
Tests pass on Travis; the :x: is the familiar error on osx (`/usr/bin/python: No module named nose`) :nose:

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -147,34 +147,44 @@ void Millard2012EquilibriumMuscle::extendFinalizeFromProperties()
...
         updMemberSubcomponent<MuscleFixedWidthPennationModel>(penMdlIdx);
+    MuscleFixedWidthPennationModel* penMdlCopy = penMdl.clone();
     penMdl.set_optimal_fiber_length(getOptimalFiberLength());
     penMdl.set_pennation_angle_at_optimal(getPennationAngleAtOptimalFiberLength());
     penMdl.set_maximum_pennation_angle(get_maximum_pennation_angle());
+    try {
+        penMdl.finalizeFromProperties();
+    } catch (const InvalidPropertyValue& ex) {
+        penMdl = *penMdlCopy;
+        throw(ex);
```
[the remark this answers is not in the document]
tkuchida: Ah ok, good point. I'll just `throw;` :baseball:

OpenSim/Actuators/MuscleFirstOrderActivationDynamicModel.cpp
```diff
@@ -97,15 +97,17 @@ void MuscleFirstOrderActivationDynamicModel::extendFinalizeFromProperties()
...
-        "MuscleFirstOrderActivationDynamicModel::extendFinalizeFromProperties",
-        "%s: Deactivation time constant must be greater than zero",
-        getName().c_str());
-    SimTK_VALUECHECK_ALWAYS(0.0, get_minimum_activation(),
-        1.0-SimTK::SignificantReal, "minimum_activation",
-        errorLocation.c_str());
+    OPENSIM_THROW_IF_FRMOBJ(
+        get_activation_time_constant() < SimTK::SignificantReal,
+        InvalidPropertyValue,
+        getProperty_activation_time_constant().getName());
```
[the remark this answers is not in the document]
tkuchida: Fixed here and elsewhere.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -147,34 +151,44 @@ void Millard2012EquilibriumMuscle::extendFinalizeFromProperties()
...
-    // maximum_pennation_angle are checked by
-    // MuscleFixedWidthPennationModel::extendFinalizeFromProperties().
+    // Propagate properties down to pennation model subcomponent. If any of the
+    // new property values are invalid, restore the subcomponent's current
+    // property values (to avoid throwing again when the subcomponent's
+    // extendFinalizeFromProperties() method is called directly) and then
+    // re-throw the exception thrown by the subcomponent.
     auto& penMdl =
         updMemberSubcomponent<MuscleFixedWidthPennationModel>(penMdlIdx);
+    MuscleFixedWidthPennationModel* penMdlCopy = penMdl.clone();
```
chrisdembia: This memory is leaked (ditto for `actMdlCopy`). Consider:

```cpp
std::unique_ptr<MuscleFixedWidthPennationModel> penMdlCopy(penMdl.clone());
```

(no need to change the copy assignment below; it will work as-is).

An alternative is to use the copy constructor:

```cpp
MuscleFixedWidthPennationModel penMdlCopy(penMdl);
```

Side note: The pattern of cloning and restoring is fine here because the specific subcomponent doesn't have sockets or add anything to the system; I'm not sure this pattern would work for a subcomponent with state variables.

OpenSim/Actuators/Thelen2003Muscle.cpp
```diff
@@ -102,24 +102,46 @@ void Thelen2003Muscle::extendFinalizeFromProperties()
...
+    } catch (const InvalidPropertyValue&) {
+        pennMdl = *pennMdlCopy;
+        throw;
+    }
 
+    // Propagate properties down to activation dynamics model subcomponent.
+    // Handle invalid properties as above for pennation model.
     auto& actMdl =
         updMemberSubcomponent<MuscleFirstOrderActivationDynamicModel>(actMdlIdx);
+    MuscleFirstOrderActivationDynamicModel* actMdlCopy = actMdl.clone();
```
chrisdembia: See comment above about the memory leak.

OpenSim/Simulation/Wrap/WrapCylinder.cpp
```diff
@@ -118,13 +118,11 @@ void WrapCylinder::extendFinalizeFromProperties()
...
-        string errorMessage = getProperty_radius().getName()
-            + " is not specified or is negative.";
-        OPENSIM_THROW_FRMOBJ(InvalidPropertyValue,
-            getProperty_radius().getName(), errorMessage);
-    }
+    OPENSIM_THROW_IF_FRMOBJ(
+        get_radius() < 0,
+        InvalidPropertyValue,
+        getProperty_radius().getName(),
+        "Radius must be specified and cannot be less than zero");
```
chrisdembia: Just curious: What led you to update these error checks? It doesn't look like you changed the constructor to `InvalidPropertyValue` or anything like that.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
aseth1: Good catch @chrisdembia. I think the pattern for any subcomponent would work, since it is a mechanism of caching and restoring its properties and not its state. I have slight preference for the copy constructor.

OpenSim/Simulation/Wrap/WrapCylinder.cpp
aseth1: I appreciate the additions/correction (goes beyond the call of duty for this PR) as part of the sweep to add meaningful messages. I think it is fine leave it in here- please don't spend the time to separate at this point.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
tkuchida: Thanks @chrisdembia. I guess I misinterpreted Ajay's instructions. I'll change to copy constructor.

OpenSim/Simulation/Wrap/WrapCylinder.cpp
chrisdembia: Of course; I did not mean that these changes should be separated. Ignore the question.

tkuchida: Ajay wanted the exceptions in the `MuscleFixedWidthPennationModel` and `MuscleRebellionActivationDynamicModel` `extendFinalizeFromProperties()` methods changed to `InvalidPropertyValue`, so I changed them by following examples elsewhere in the repo. I then got more comments saying that my use of `InvalidPropertyValue` was poor, so I updated all uses of `InvalidPropertyValue` in the repo so that doesn't happen again.

aseth1: Thanks @tkuchida!

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
tkuchida: @chrisdembia & @aseth1 - Have you ever considered removing `clone()` from the public interface? One could argue that I should have known better, but I wouldn't have eaten the poison if the API hadn't left it in the kitchen---and I don't use a poison detector in my workflow. User-centered design principle: error prevention.

tkuchida:
Thanks @chrisdembia.
@jimmyDunne Would you please add a task to verify in the GUI? :car: Thanks!

PR #2165, opened 2018-05-04
[travis] No longer use nose to run Python tests

chrisdembia:
### Brief summary of changes

- This PR removes `nose` from Travis. We were using the nose library to run python tests but we have had trouble downloading nose on Mac and we can easily do without nose.

### Testing I've completed

- Ran Travis CI tests.

### CHANGELOG.md (choose one)

- no need to update because...CI does not need to documented in the changelog.

Thanks @tkuchida for starting this fix.


tkuchida:
:heart_eyes:

PR #2166, opened 2018-05-04
Fix correction controller with GUI use case

aseth1:
Fixes issue #<52164>

### Brief summary of changes
GUI use case was replicate in testForward. Repeated `CorrectionController::extendConnectToModel()` calls were incorrectly adding in *corrector* actuators. The reason being they were expecting actuators in the model `ForceSet`, but actuators but `CorrectionController` owns the actuators it adds and includes them as subcomponents so that deleting the `CorrectionController` will take its actuators with it. The fix was to use `model.getComponentList<CoordinateActuator> to also find the subcomponents of `CorrectionController` should they exist (e.g. previous calls to `connectToModel`).

Also snuck in changes to eliminate unnecessary dynamic_cast in `computeControls` and cleaned  up `testForward`.

### Testing I've completed
Included a test case that emulates the GUI call sequence, which initially reproduced the bug. 
Verified that it produces identical results to the command line solution.

### CHANGELOG.md (choose one)
- no need to update because this is a bug fix, with no API changes.

### DIFF
--- a/Applications/Forward/test/subject01_Setup_Forward_Controller.xml
+++ b/Applications/Forward/test/subject01_Setup_Forward_Controller.xml
@@ -60,8 +60,8 @@
 					    indicates the controller will controll all the acuators in the model-->
 					<actuator_list> </actuator_list>
 					<!--Flag (true or false) indicating whether or not the controller is
-					    enabled (ON) should-->
-					<enable_controller> true </enable_controller>
+					    disabled.-->
+					<isDisabled> false </isDisabled>
 					<!--XML file containing the controls for the controlSet.-->
 					<controls_file> subject01_walk1_controls.xml </controls_file>
 				</ControlSetController>
@@ -70,12 +70,12 @@
 					    indicates the controller will controll all the acuators in the model-->
 					<actuator_list> </actuator_list>
 					<!--Flag (true or false) indicating whether or not the controller is
-					    enabled (ON) should-->
-					<enable_controller> true </enable_controller>
+					    disabled.-->
+					<isDisabled> false </isDisabled>
 					<!--Gain for position errors-->
-					<kp>     25.00000000 </kp>
+					<kp>    16.00000000 </kp>
 					<!--Gain for velocity errors-->
-					<kv>      10.00000000 </kv>
+					<kv>     8.00000000 </kv>
 				</CorrectionController>
 			</objects>
 			<groups/>
--- a/Applications/Forward/test/testForward.cpp
+++ b/Applications/Forward/test/testForward.cpp
@@ -39,7 +39,7 @@ void testPendulumExternalLoadWithPointInGround(); // test application of externa
 void testArm26();       // now add computation of controls and generation of muscle forces
 void testGait2354();    // controlled muscles and ground reactions forces 
 void testGait2354WithController(); // included additional controller
-
+void testGait2354WithControllerGUI(); // implements steps GUI takes to provide a model
 int main() {
     Object::renameType("Thelen2003Muscle", "Thelen2003Muscle_Deprecated");
 
@@ -79,6 +79,15 @@ int main() {
         cout << "\ngait2354 with correction controller test PASSED " << endl; }
     catch (const std::exception& e)
         { cout << e.what() <<endl; failures.push_back("testGait2354WithController"); }  
+
+    try {
+        testGait2354WithControllerGUI();
+        cout << "\nGUI run gait2354 with correction controller test PASSED " << endl;
+    }
+    catch (const std::exception& e) {
+        cout << e.what() << endl;
+        failures.push_back("testGait2354WithControllerGUI");
+    }
 
     if (!failures.empty()) {
         cout << "Done, with failure(s): " << failures << endl;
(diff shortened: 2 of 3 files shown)


OpenSim/Tools/CorrectionController.cpp
```diff
@@ -257,16 +255,26 @@ void CorrectionController::extendConnectToModel(Model& model)
...
-                  <<  cs.get(i).getName()+"_corrector" << "  added " 
-                  << std::endl;
-        std::string name = cs.get(i).getName()+"_corrector";
-        CoordinateActuator *actuator = NULL;
-        if(_model->getForceSet().contains(name)){
-            actuator = (CoordinateActuator *)&_model->getForceSet().get(name);
+        const Coordinate& coord = cs[i];
+        std::string name = coord.getName() + "_corrector";
+
+        CoordinateActuator* actuator = nullptr;
```
tkuchida: The code on L279–289 would use `actuator` defined on L278, but reusing the variable name makes the code difficult to read. Can you just use a `bool` (flag) here and change `if(!actuator)` on L273 to `if (!actuatorFound)`?

OpenSim/Tools/CorrectionController.cpp
```diff
@@ -257,16 +255,26 @@ void CorrectionController::extendConnectToModel(Model& model)
...
+        for (auto& ca : actuators) {
+            if (ca.getName() == name) {
+                actuator = &ca;
+                break;
+            }
         }
-        else{
+
+        if(!actuator) {
+            std::cout << " CorrectionController::extendConnectToModel(): "
```
tkuchida: This comment seems to be premature, and if something throws on L278–289, it would seem like the Controller had been added successfully. I suggest either changing "...added" to "adding...", or (perhaps better) move this comment down, after `addActuator(*actuator);`.

OpenSim/Tools/CorrectionController.cpp
```diff
@@ -257,16 +255,26 @@ void CorrectionController::extendConnectToModel(Model& model)
...
+                << coord.getName() + "_corrector" << "  added "
```
tkuchida: Should replace `coord.getName() + "_corrector"` with `name` so the code is easier to follow, to prevent the comment going out of sync with the code, and because `name` is used on L280 anyway.

OpenSim/Tools/CorrectionController.cpp
aseth1: Good point.

OpenSim/Tools/CorrectionController.cpp
aseth1: will do.

OpenSim/Tools/CorrectionController.cpp
aseth1: I think I introduced a bug! The reason for the pointer is that if the actuator is found in the model it should be added to the working set of the Controller in L289, which should be outside the if/else block. Just happens that the test case does not have any correctors as part of the model (they are introduced and owned by the Controller itself).

tkuchida: Ah ok, cool. Perhaps we should add a test case, or create an Issue to do it later?

aseth1: I made issue #2168 to track this. Thanks!

aseth1:
@aymanhab this PR is waiting another reviewer if you would like to include it in the next build of OpenSim

Applications/Forward/test/subject01_Setup_Forward_Controller.xml
```diff
@@ -60,8 +60,8 @@
 					    indicates the controller will controll all the acuators in the model-->
 					<actuator_list> </actuator_list>
 					<!--Flag (true or false) indicating whether or not the controller is
-					    enabled (ON) should-->
-					<enable_controller> true </enable_controller>
+					    disabled.-->
+					<isDisabled> false </isDisabled>
```
aymanhab: @aseth1 Was this change made by hand or by saving the file? How come there's no change in version number? I'd expect the xml tags to be consistent with the XMLDocument version number otherwise something is amiss. Conversions need to be built into the code so that user files continue to function in a predictable manner.

OpenSim/Tools/CorrectionController.cpp
```diff
@@ -275,16 +280,22 @@ void CorrectionController::extendConnectToModel(Model& model)
...
-        actuator->setOptimalForce(1.0);
+            std::cout << " CorrectionController::extendConnectToModel(): "
+                << name << " added " << std::endl;
+
+            actuator->setOptimalForce(1.0);
+        }
         
-        updActuators().adoptAndAppend(actuator);
+        // Add to the Controller's list of Actuators (no ownership).
+        addActuator(*actuator);
```
aymanhab: Is it true that there's no more ownership or is the ownership now maintained through the subcomponents list? some clarifying comment would help

aseth1: `Controller::addActuator` does not take ownership and that was definitely not changed here. I added the comment for clarity. The `Model` generally owns the actuators being controlled. In this case, the `CorrectionController` owns the actuators it creates and includes them as its subcomponents, yes.

Applications/Forward/test/testForward.cpp
```diff
@@ -249,3 +223,37 @@ void testGait2354WithController() {
...
+
+    // For good measure we'll make sure we still get the identical results
+    Storage results(resultsDir+"/subject01_states.sto");
+    //Storage standard("std_subject01_walk1_states.sto");
+    Storage standard("ResultsCorrectionController/subject01_states.sto");
+
+    int nstates = forward.getModel().getNumStateVariables();
+    int nq = forward.getModel().getNumCoordinates();
+    std::vector<double> rms_tols(2 * nstates, SimTK::SqrtEps);
+
```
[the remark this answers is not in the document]
aseth1: L241-244
```cpp
+    forward.updateModelForces(*model, "");
+    forward.setModel(*model);
+
+    model->initSystem();
```

Not sure it is worth adding another function for three lines.  I would also argue that if the `model.initSystem()` is required here, after setting the model on the Tool, then the logic/flow is is error prone. If you need access to the Tool's model, you should access it through the Tool's `getModel()` and get a reference to the same model. Otherwise you have two models, one in the GUI ForwardToolModel (Worker) and one within the `ForwardTool` itself and their state will be mismatched.

Applications/Forward/test/subject01_Setup_Forward_Controller.xml
aymanhab: ok, thanks for the clarification.

Applications/Forward/test/testForward.cpp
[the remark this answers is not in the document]
aymanhab: @aseth1 While it may help the GUI stay consistent with tools, the point about making one function to use here is driven by repeated user requests to run tools on live models. Not critical to this PR tho.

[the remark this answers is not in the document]
aseth1: Maybe we can discuss this further. If it is primarily through the API that they want to run Tools on a live model, I don't think that is necessarily the best approach, but it should be possible if the GUI can do it.

aseth1:
Thanks @aymanhab and @tkuchida.

PR #2173, opened 2018-05-08
Minor correction to exception message

tkuchida:
Leftover from #2159


aseth1:
We don't have to wait to merge this.

Issue #2038, opened 2017-12-12
configureOpenSim.m: on Windows, detect if OpenSim (or previous installs) are on PATH


chrisdembia:
One of the most difficult parts of using setting up Matlab on Windows is that users must manually set their PATH (https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=8491). The configureOpenSim script could be more intelligent: it could detect if OpenSim is already on the PATH (in which case no warning is necessary). It could also detect if 3.3 or other copies of OpenSim are on the path, and warn users that they should remove these.

PR #2174, opened 2018-05-09
No longer require users to set PATH on Windows for Matlab.

chrisdembia:
Partly addresses #701 

### Brief summary of changes

- This PR removes the need for users of the Matlab interface to edit their Windows PATH environment variable, which is the most painful part of using the Matlab interface (especially if the user has multiple OpenSim installations).
- On Windows, there is now a text file `osimJavaJNI_dependencies.txt` that sits in `sdk/Java/` (next to `org-opensim-modeling.jar`) that contains a list of libraries (specifically, relative paths to DLLs in `bin/`) that `org-opensim-modeling.jar` loads before loading `osimJavaJNI`.
- If anything goes wrong with the new mechanism, we fall back on the old Windows PATH mechanism.
- This new mechanism depends on Java 1.7 (for `readAllLines()`). [MATLAB 2013b is the oldest version to come with Java 1.7](https://www.mathworks.com/support/sysreq/previous_releases.html).

Note: Users of the Python interface will still need to edit their Windows PATH.

One flaw in this scheme is that the list of libraries we need to write to `osimJavaJNI_dependencies.txt` could change, and we will have to maintain this list.

TODO: Update Confluence documentation that tells users they must update PATH for Matlab.

### Testing I've completed

- I ensured that users can have `C:\OpenSim 3.3\bin` on their PATH (and have OpenSim 3.3 installed) and Matlab will still properly find 4.0 bindings.
- I ensured the new mechanism works for `WITH_BTK=on` and `WITH_BTK=off`.
- I ensured the new mechanism works in a fresh Windows 10 environment.
- I ensured that if the `osimJavaJNI_dependencies.txt` mechanism has errors (the file is missing, it has bogus entries, it is missing library dependencies), but the Windows PATH contains the correct 4.0 bin directory, Matlab can still load OpenSim.

- Is this mechanism worthwhile to include, given that it is a little fragile (maintaining this new file)?
- Does this mechanism introduce any security issues we care about? The code will load any DLL listed in `osimJavaJNI_dependencies.txt`, including malicious code from...malicious people.
- Can we expect most users to be using MATLAB 2013b or later?

### CHANGELOG.md (choose one)

- updated.

Related: #2038.


chrisdembia:
I accidentally merged the dynamicwalking2018 branch into this PR...oops. Don't merge until I remove those changes.

chrisdembia:
Okay I fixed this PR; it no longer includes changes from `dynamicwalking2018`. The PR can be reviewed (and merged) now.

chrisdembia:
Based on discussion with @aymanhab, I will test this with Matlab 2013.

chrisdembia:
Thanks for testing @jimmyDunne

chrisdembia:
I discovered an issue in testing with Matlab 2013b; the Visual C++ runtime libraries are not in the correct dependency order. Do not merge this yet.

chrisdembia:
I've  updated this PR to work with older versions of Matlab. I tested that this PR works as intended with Matlab 2013b. I also tested that, with Matlab 2013a, users still must set the Path (because this version uses Java 1.6).

The issue with Matlab 2013b was that the Visual C++ runtimes we needed were not installed (by Matlab), so we were relying on the runtimes we provide in the OpenSim bin folder. But the order of these libraries in `osimJavaJNI_dependencies.txt` was not correct. I reordered the libraries, but this is done in somewhat of a fragile way (moving `vcruntime*.dll` to the front of the list); I don't know if the libraries will continue to remain in the correct order as we move to newer versions of Visual Studio.

It seems like we shouldn't be relying on "local deployment" of the Visual C++ runtimes anyway (https://docs.microsoft.com/en-us/cpp/ide/deployment-in-visual-cpp). I think @aymanhab has already expressed preference for using the Visual C++ Redistributable.

chrisdembia:
Ready for review again.

aseth1:
@aymanhab please review when you get a chance. It would be great to get this in and used locally before DW 2018.

chrisdembia:
I'm not sure we should merge this. I am running into issues when using the `simbody-visualizer` through Matlab.

chrisdembia:
I have no plan to continue working on this now. I'm closing this PR.

PR #2176, opened 2018-05-09
Avoid some SWIGTYPE_p_... types

chrisdembia:
Partially addresses https://github.com/opensim-org/opensim-gui/issues/722

### Brief summary of changes

- This PR prevents some unused SWIG types from being generated (e.g., `SWIGTYPE_p_std__functionT_void_fOpenSim__Component_const_p_SimTK__State_const_R_std__string_const_R_SimTK__VecT_2_SimTK__VecT_3_double_1_t_1_t_RF_t`).

### Testing I've completed

- Built Java bindings locally and ran all tests.


OpenSim/Common/ComponentOutput.h
```diff
@@ -266,6 +270,7 @@ class Output : public AbstractOutput {
      @endcode
      */
     const ChannelMap& getChannels() const { return _channels; }
+#endif
```
aymanhab: Is there loss of functionality here for scripting users?

OpenSim/Common/ComponentSocket.h
```diff
@@ -807,6 +808,7 @@ class Input : public AbstractInput {
     const ChannelList& getChannels() const {
         return _connectees;
     }
+#endif
```
aymanhab: Same question, as I'm not sure if ChannelList get mapped to an actual humanly usable class or not, if not then LGTM

aymanhab:
If this results in only removable of useless/unused SWIG types then absolutely go for it, we probably need one pass through the Java Bindings, using netbeans refactoring tools to find out where these classes are used and avoid generating them. Good start @chrisdembia

chrisdembia:
> If this results in only removable of useless/unused SWIG types then absolutely go for it

Correct; I only "removed" methods that were already useless because they involved `SWIGTYPE_p_...` types.

chrisdembia:
@aymanhab let me know if you need anything else on this PR to get your ✅

chrisdembia:
Thanks @aymanhab

Issue #2133, opened 2018-03-26
testStorage should pass even when WITH_BTK is OFF


chrisdembia:
After #2127 was merged, `testStorage` requires `WITH_BTK` to be `ON` in order to pass. To fix this, we should use the `WITH_BTK` macro within `testStorage` to avoid reading C3D if `WITH_BTK` is `OFF`.

Reported by @tkuchida.

tkuchida:
Fixed by #2177.

PR #2177, opened 2018-05-09
Do not test c3d reading if WITH_BTK=off.

chrisdembia:
### Brief summary of changes

- Do not test reading a c3d file if `WITH_BTK` is off. Without this change, `testStorage` fails if `WITH_BTK=off`.

### Testing I've completed

- Locally built and ran `testStorage` with both `WITH_BTK` as `off` and `on`, and ensured that when `WITH_BTK=on`, the c3d file is tested.

### CHANGELOG.md (choose one)

- no need to update because this only affects a test case.


tkuchida:
Fixes #2133.

PR #2180, opened 2018-05-09
OptimizeHopper overhaul: triple hop with global optimization

nickbianco:
### Brief summary of changes
So I don't really know when to quit with this hopper stuff. Here's an overhaul of `OptimizeHopper()` that lets the user choose between two problems:
- A one hop problem with a MATLAB local optimizer
- A three hop problem with a MATLAB global optimizer

I've also added some plotting so users can visualize the optimized controls and the jump trajectory.

### Testing I've completed
Both problems run and solve consistently. I've bound the global optimizer to ten generations, but it seems to pretty much find the global optimum by then (>1.75!).

The file is a little more involved now, but I tried to comment everything and make it as clear as possible. Let me know what isn't clear.


nickbianco:
I removed the simulations of initial guesses, mostly because they already looked pretty good. I like having them in there before each optimization though. Thoughts?

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
```diff
@@ -25,81 +26,184 @@
...
+
+% ANALYZE OPTIMIZED SOLUTION
+% Construct the actuator controls from solution 'x' and interpolate. See 
+% CONSTRUCTCONTROLS below for details.
+[deviceControl, muscleExcitation] = constructControls(x, problem);
+time = 0:0.01:5;
+deviceControlInterp    = interp1(deviceControl(1,:), ... 
+                                 deviceControl(2,:), time);
+muscleExcitationInterp = interp1(muscleExcitation(1,:), ... 
+                                 muscleExcitation(2,:), time);
```
chrisdembia: Is the interpolation necessary? The control is piecewise linear.

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
```diff
@@ -25,81 +26,184 @@
...
+        t1 = x;
+        deviceControl    = [on(0)  off(t1)];
+        muscleExcitation = [off(0) on(t1)];
+        
+    % To coordinate the sequence of hops in this problem, the spring must
+    % be loaded and released (with the activated muscle) three times. The
+    % first variable, x(1), is the wait time between when the spring is 
+    % loaded and released for each hop. The x(1) value can be used for all 
+    % three hops since we also optimize the wait time between each hop with 
+    % the other two variables, x(2) and x(3). The active device is turned
```
chrisdembia: I had to read this a few times to understand. How about "The x(1) value is the duration for which the spring is loaded before each hop. The variables x(2) and x(3) are  the wait times to start loading the spring again for the 2nd and 3rd hops, respectively."

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
```diff
@@ -25,81 +26,184 @@
...
+        options = optimset('TolFun', 1e-2, ...
+                           'MaxIter', 10000, ...
+                           'Display', 'iter'); 
+        [x, f] = fminbnd(@(x) objective(x,problem),lb,ub,options);
+                     
+    % Optimize the height of a sequence of three hops. The optimization
+    % variables decide to switch the actuators on and off to load and 
+    % release the passive device with the activated muscle (see 
+    % CONSTRUCTCONTROLS below). This problem is subject to falling into
+    % undesired local minima (e.g. solutions with less than three hops) and
```
chrisdembia: In my view: the problem *has* local minima. Being subject to falling into the local minima is an attribute of the solver. Consider "We use a global optimizer (genetic algorithm from MATLAB's Global Optimization Toolbox) to avoid falling into local minima (e.g., solutions with less than three hops)."

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
```diff
@@ -25,81 +26,184 @@
...
+switch problem
+    
+    % Optimize the height of one hop. The single optimization variable 
+    % determines when to simultaneously release a loaded passive device and 
+    % activate the hopper vastus muscle (see CONSTRUCTCONTROLS below). This
+    % simple univariate problem uses FMINBND from the MATLAB Optimization 
+    % Toolbox. 
+    case 'one_hop'
+        lb = 0;
+        ub = 1;
```
chrisdembia: Do you have space to spell these out? `lowerBound`, `upperBound`?

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
```diff
@@ -25,81 +26,184 @@
...
+    % variables decide to switch the actuators on and off to load and 
+    % release the passive device with the activated muscle (see 
+    % CONSTRUCTCONTROLS below). This problem is subject to falling into
+    % undesired local minima (e.g. solutions with less than three hops) and
+    % requires an optimizer from the MATLAB Global Optimization Toolbox.
+    % Here we use the genetic algorithm optimizer, GA.
+    case 'three_hops'
+        numVars = 3; 
+        lb = 0.25*ones(numVars, 1);
+        ub = 0.75*ones(numVars, 1);
```
chrisdembia: Does this mean the hops happen within the first second? Consider putting units in comments for `lb` and `ub`.

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
```diff
@@ -25,81 +26,184 @@
...
+
+% Construct the actuator controls. See CONSTRUCTCONTROLS below for details.
+[deviceControl, muscleExcitation] = constructControls(x, problem);
           
-% Build the hopper. This is the same function called after a solution is
-% is created in the InteractiveHopper GUI. The default muscle is "The
-% Average Joe".
+% Build the hopper with the passive device wrapped and the active device
+% unwrapped (behind the knee). This is the same function called after a
+% solution is created in the InteractiveHopper GUI. Here we use 'The Katie
```
chrisdembia: Consider "This is the same function that the The InteractiveHopper GUI calls when you click Simulate." (mainly to avoid saying "solution" for a guess/"iterate").

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
```diff
@@ -25,81 +26,184 @@
...
+
+% Anonymous functions to conveniently construct 'bang-bang' control signals 
+% for both the active device and the muscle. 
+% 
+% Switch an actuator ON at time 't'. 
+on  =@(t) [t  t+0.01; 
+           0  1.0];
+% Switch an actuator OFF at time 't'.
+off =@(t) [t   t+0.01; 
+           1.0 0];
```
chrisdembia: Very clean.

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
```diff
@@ -25,81 +26,184 @@
...
             'printModelInfo', false);
 end
 
-function f = objective(x)
-
-hopper = buildHopperFromSolution(x);
-[peakHeight, ~] = EvaluateHopper(hopper, false, true);
+function [deviceControl, muscleExcitation] = constructControls(x, problem)
+% CONSTRUCTCONTROLS
+%   Construct controls from the optimization optimization variables. Each
```
chrisdembia: > optimization ~optimization~

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
nickbianco: No it's not. Not sure why I included it...

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
nickbianco: I like this rewording. Mine definitely wasn't clear on second read.

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
nickbianco: Good points. I'll incorporate this phrasing too.

carmichaelong: nitpicky: "less" --> "fewer"

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
carmichaelong: Perhaps a clear description of the design variables (i.e., the vector `x`) above could help here, since these map directly to them.

nickbianco: The units are in seconds. It means the first hop happens with the first second, and then the others happen based on additions of the three variables as described in `constructControls`.

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
nickbianco: Good idea.

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
```diff
@@ -25,81 +26,184 @@
...
+        
+        deviceControl    = [on(0)  off(t1) on(t2)  off(t3) on(t4)  off(t5)];
+        muscleExcitation = [off(0) on(t1)  off(t2) on(t3)  off(t4) on(t5)];    
+end
 
-% Negate peak height for minimization
-f = -peakHeight;
+% For both muscle and device, impose a zero-order hold on the control value
+% to the end of the jump cycle.
+deviceControl    = [deviceControl(1,:) 5;
```
carmichaelong: I think this is a relic of something I did before, but is there a good place to have a variable such as `finalTime` rather than hard-coding in the `5` in a few places? There's a good chance there isn't, and hence why I was lazy and never did it :-/

nickbianco: No, I did this. I don't think there's a need for a variable, but I could clarify that the jump cycle is [0 5] somewhere in the comments.

nickbianco:
I've decided to not doing anything live. I think I'll go through the snippets you suggested instead, and say something briefly about the difference in the problems and optimizers used. Since we won't be showing an initial guess simulation, we could show the one hop and three hop solutions side by side. Or maybe described the first problem, show the result, then do the same for the second problem.

I don't have a solid idea how to improve handling arguments to `BuildInteractiveHopperSolution`, but a better name could be `BuildCustomHopper`.  Also, something like `constructHopper` instead of `buildHopperFromSolution` , which would be more in line with `constructControls`, could work.

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
chrisdembia: The 5 seconds is hard coded in multiple files.

chrisdembia:
I like those suggested file names.

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
nickbianco: Right, like in the input checking for `BuildInteractiveHopperSolution`. In that case I'll definitely leave it hard coded.

nickbianco:
Thanks for the helpful feedback. Ready for review.

PR #2188, opened 2018-05-12
[WIP] Throw Exception if results directory cannot be created or file cannot be written

### LINKED ISSUE #2181: IO Exception isn't written to messages window
@jimmyDunne commented on [Wed May 09 2018](https://github.com/opensim-org/opensim-gui/issues/725)

### Steps to reproduce
Run IK in the GUI, specify an Output path that contains a non-existent folder/dir. 

### Expected result
IK output file is not written, exception message appears in the Messages Window:

```
IO.OpenFile(const string&,const string&): failed to open ./ResultsIK/subject01_run_IK.mot
Running tool run.
```
### Actual result
No file, no exception. 

### Environment and GUI version
W10. OpenSim-800bfe33-2018-05-09-win64.exe

---

@tkuchida commented on [Wed May 09 2018](https://github.com/opensim-org/opensim-gui/issues/725#issuecomment-387868611)

Some students ran into this bug yesterday (via Scale Tool in 3.3). There was a :red_circle: error but I agree: a message in the Messages Window would have been helpful. Further, "failed to open" sounds like an input file is missing; "could not write" would be more appropriate here.

---

@aymanhab commented on [Wed May 09 2018](https://github.com/opensim-org/opensim-gui/issues/725#issuecomment-387919557)

The fundamental issue is that we allow a low level error to bubble up unchecked resulting in either subtle message or incorrect one. If the std exception is caught by core and thrown as OpenSim::Exception with meaningful message (which can be more meaningful since we know the context better than OpenFile). This should be fixed on core and will automatically be displayed as an ErrorDialog in GUI.

tkuchida:
Fixes #2181.

### Brief summary of changes
- Catch [`errno == ENOENT`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/errno-doserrno-sys-errlist-and-sys-nerr) if results directory cannot be created.
- Throw Exception from Storage if file cannot be opened for writing.

### Testing I've completed
Ran example setup file that attempted to write results to `C:\Program Files`:
<pre>
InverseKinematicsTool Failed: Storage: Failed to open file 'C:\Program Files/ik_marker_errors.sto' for writing.
Verify that the destination directory is writable.
        Thrown at Storage.cpp:2746 in print().
</pre>

### CHANGELOG.md (choose one)

- no need to update because... bug fix.
- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2188)

### DIFF
--- a/OpenSim/Actuators/Test/testMuscles.cpp
+++ b/OpenSim/Actuators/Test/testMuscles.cpp
@@ -429,8 +429,9 @@ void simulateMuscle(
 
     //An analysis only writes to a dir that exists, so create here.
     if(printResults){
-        IO::makeDir("testMuscleResults");
-        muscleAnalysis->printResults(actuatorType, "testMuscleResults");
+        const std::string dirName = "testMuscleResults";
+        OPENSIM_THROW_IF(errno == ENOENT, UnableToCreateDirectory, dirName);
+        muscleAnalysis->printResults(actuatorType, dirName);
     }
 
     double muscleWork = muscWorkProbe->getProbeOutputs(si)(0);
--- a/OpenSim/Common/FileAdapter.h
+++ b/OpenSim/Common/FileAdapter.h
@@ -70,6 +70,19 @@ public:
     }
 };
 
+class UnableToCreateDirectory : public IOError {
+public:
+    UnableToCreateDirectory(const std::string& file,
+                            size_t line,
+                            const std::string& func,
+                            const std::string& directory) :
+        IOError(file, line, func) {
+        std::string msg = "Unable to create directory '" + directory + "'.";
+
+        addMessage(msg);
+    }
+};
+
 class FileExtensionNotFound : public InvalidArgument {
 public:
     FileExtensionNotFound(const std::string& file,
--- a/OpenSim/Common/Storage.cpp
+++ b/OpenSim/Common/Storage.cpp
@@ -2741,7 +2741,9 @@ print(const string &aFileName,const string &aMode, const string& aComment) const
 {
     // OPEN THE FILE
     FILE *fp = IO::OpenFile(aFileName,aMode);
-    if(fp==NULL) return(false);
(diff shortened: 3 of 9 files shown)


chrisdembia (commented):
Would it make sense to put the check `OPENSIM_THROW_IF(errno == ENOENT, UnableToCreateDirectory, aDir);` inside `IO::makeDir()`?

tkuchida:
> Would it make sense to put the check OPENSIM_THROW_IF(errno == ENOENT, UnableToCreateDirectory, aDir); inside IO::makeDir()?

Yes, that is another option. I thought I would get in trouble editing IO.cpp so I literally flipped a coin. Bigger issue is that tests pass locally on Windows yet fail even on AppVeyor, so I need to investigate.

chrisdembia:
The line number in this exception from the AppVeyor log is suspicious:

```cpp
Exception:
4466  Unable to create directory 'twoMusclesOnBlock_ResultsCMC'.
4467  Thrown at CMCTool.cpp:831 in run().
4468  line= 1978645904
```

I don't think any of our source code files have that many lines.

tkuchida:
@aseth1 Do you want these exceptions thrown from IO.cpp instead, or are there scenarios in which a consumer would want to notify the user but proceed anyway?

tkuchida:
> Would it make sense to put the check OPENSIM_THROW_IF(errno == ENOENT, UnableToCreateDirectory, aDir); inside IO::makeDir()?

It depends what you mean by "would it make sense". The current interface of `IO::makeDir()` is the following:
https://github.com/opensim-org/opensim-core/blob/671076b4764d0d9d9ffc0ddc1d73cbe81da41df5/OpenSim/Common/IO.cpp#L467-L470
so it seems like the problem is that the callers are ignoring the return value/error condition, not that `makeDir()` has been implemented incorrectly.

nickbianco:
Closing: stale PR.

PR #2189, opened 2018-05-12
[WIP] OptimizeHopper: update optimized solution plots

nickbianco:
### Brief summary of changes
I updated the solution plots so they're more informative and interesting. 

The implementation is as concise as I could make it on a first pass, but it's not very clean. I put the plotting in a new subfunction `plotSolution()` to address this, but feedback would be helpful.

### DIFF
--- a/Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
+++ b/Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
@@ -55,35 +55,15 @@ switch problem
         lowerBound = 0.25*ones(numVars, 1); % seconds
         upperBound = 0.75*ones(numVars, 1); % seconds
         options = optimoptions('ga','FunctionTolerance', 1e-2, ...
-                                    'MaxGenerations', 10, ...
+                                    'MaxGenerations', 1, ...
                                     'Display', 'iter'); 
         [x, f] = ga(@(x) objective(x,problem),numVars,[],[],[],[], ...
                                    lowerBound,upperBound,[],options);                 
 end
 
-% ANALYZE OPTIMIZED SOLUTION
-% Construct the actuator controls from solution 'x'. See 
-% CONSTRUCTCONTROLS below for details.
-[deviceControl, muscleExcitation] = constructControls(x, problem);
-                             
-% Plot optimized controls.
-figure(1);
-plot(deviceControl(1,:), deviceControl(2,:), 'b-', 'linewidth', 2)
-hold on
-plot(muscleExcitation(1,:), muscleExcitation(2,:), 'r-', 'linewidth', 2)
-xlabel('time (s)')
-ylabel('control value')
 
-% Evaluate optimized solution. See CONSTRUCTHOPPER and OBJECTIVE 
-% below for details.
-hopper = constructHopper(x, problem);
-[~, heightStruct] = EvaluateHopper(hopper, true, true);
-
-% Plot jump height.
-figure(2);
-plot(heightStruct.time, heightStruct.height, 'b-', 'linewidth', 2)
-xlabel('time (s)')
-ylabel('height')
+% PLOT OPTIMIZED SOLUTION
+plotSolution(x, problem);
 
 end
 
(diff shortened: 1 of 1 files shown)


Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
```diff
@@ -55,35 +55,15 @@
         lowerBound = 0.25*ones(numVars, 1); % seconds
         upperBound = 0.75*ones(numVars, 1); % seconds
         options = optimoptions('ga','FunctionTolerance', 1e-2, ...
-                                    'MaxGenerations', 10, ...
+                                    'MaxGenerations', 1, ...
```
chrisdembia: You can still get 3 nice hops with 1 generation?

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
```diff
@@ -201,4 +182,57 @@
...
+subplot(2,1,1);
+plot(deviceControl(1,:), deviceControl(2,:), 'b-', 'linewidth', 2)
+hold on
+plot(muscleExcitation(1,:), muscleExcitation(2,:), 'r-', 'linewidth', 2)
+ylim([-0.10 1.10]); xlim([0 5]); xticks(0:1:5);
+switch problem
+    case 'one_hop'
+        text_x = [x/2 t];
+        text_y = [1.05 -0.05];
+        labels = {'<--x(1)-->', 't1'};
```
chrisdembia: I wonder if the `<--x(1)-->` will consistently fill up the correct amount of space on different computers or for different figure sizes; perhaps there are cases where the `<...>` extend beyond the desired interval.

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
```diff
@@ -201,4 +182,57 @@
...
+    case 'three_hops'
+        text_x = [(x([1 2 1 3 1])/2 + [0 t([1 2 3 4])]) t];
+        text_y = [1.05*ones(1,5) -0.05*ones(1,5)];
```
chrisdembia: Nice!

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
```diff
@@ -201,4 +182,57 @@
...
+        x_labels = arrayfun(@(x) sprintf('<--x(%i)-->', x), [1 2 1 3 1], ...
+            'UniformOutput', false);
```
chrisdembia: I did not know about `arrayfun()` but you should be able to get "iteration" using `sprintf()` on its own. I believe that `sprintf('%i ', [1, 2, 3])` will print `1 2 3` (it should repeat the string pattern).

nickbianco: `sprintf()` will do the iteration, but returns a string. I need a cell array of strings for the `text()` argument, which `arrayfun(@func, 'UniformOutput', false)` returns. 

The other option to get a cell array would be `strsplit(sprintf('<--x(%i)-->,', [1 2 1 3 1]), ',')`. It gives an extra empty string in the last cell, but that could be ignored when creating the final label array:
`labels = [x_labels(1:end-1) t_labels(1:end-1)]`. This method would also avoid the line continuation that `arrayfun()` needs for the `UniformOutput` argument.

Bindings/Java/Matlab/Hopper_Device/OptimizeHopper.m
nickbianco: I ran it a few times with one generation, and I'm still getting solid triple jumps (close to 1.75 if not above). So we could probably get away with one generation.

chrisdembia:
I ran this locally and the plots look nice. I like the ⭐️ at the peak.

chrisdembia:
You've marked the PR as `[WIP]` so I won't merge it; feel free to merge it yourself when you're ready.

chrisdembia:
@nickbianco do you plan on adding more to this PR?

Issue #2076, opened 2018-02-06
Python/Matlab examples for using the tools


mitkof6:
I find it a little bit difficult to setup the AnalyzeTool without the .xml file (through the API). It would be great if examples are provided (especially for Python and Matlab users) that demonstrate the simulation pipelines (e.g. gait2354, SO, IK, ID, JRA, etc.). Something like this: [IK](https://github.com/opensim-org/opensim-core/blob/4446724f558fb15a4fb4fd6a2dbd2df3c20f9d6a/Bindings/Java/Matlab/examples/setupAndRunIKBatchExample.m).

For example, it isn't very intuitive how to append reserve actuators and how to add different analysis. The doxygen is a little bit vague and do not explain the meaning of some functions and how to use them.

chrisdembia:
Such examples would be great. @jimmyDunne worked on a utility script for appending reserve actuators. 

#2060 is related.

mitkof6:
@chrisdembia it seems that a lot of documentation is located in the .cpp files, it would be nice if these can be accessed in the doxygen.

chrisdembia:
Yes we've discussed that a lot. Over time a lot of the documentation in the cpp files has become less than correct, and we think that incorrect documentation is worse than no documentation. We definitely want to move the comments from the cpp files to the headers, but they documentation should be corrected in the process.

mitkof6:
Continuing our discussion, here are some comments for using the AnalyzeTool. This is how I managed to initiate the SO (v3.3) after a lot of effort:


    string tempExternalLoadsXML = parameters.resultsDir +
        parameters.subjectName + "_external_loads.xml";
    // prepare external loads
    ExternalLoads externalLoads(parameters.model,
                                parameters.groundReactionXMLTemplate);
    // update template
    externalLoads.setExternalLoadsModelKinematicsFileName(
        parameters.inverseKinematicsMotion);
    externalLoads.setDataFileName(parameters.groundReactionForces);
    externalLoads.setLowpassCutoffFrequencyForLoadKinematics(6);
    externalLoads.print(tempExternalLoadsXML);

    // add reserve actuators
    ForceSet forceSet(parameters.model, parameters.reserveActuators);
    for (int i = 0; i < forceSet.getSize(); i++) {
        parameters.model.updForceSet().append(forceSet.get(i));
    }

    // add static optimization
    Storage motion(parameters.inverseKinematicsMotion);
    auto staticOptimization = new StaticOptimization();
[...]

PR #2046, opened 2017-12-21
Kinematics reporter used internally by InverseKinematicsTool should n…

aymanhab:
…ot be deleted (will go out of scope)

Fixes issue https://github.com/opensim-org/opensim-gui/issues/421

### Brief summary of changes
Analysis created on the stack shouldn't be deleted as it will be deleted again when goes out of scope.
### Testing I've completed
Ran Scale followed by IK or IK standalone in the GUI

### CHANGELOG.md (choose one)

- no need to update because...
- updated...

The Doxygen for this PR can be viewed at http://myosin.sourceforge.net/?C=N;O=D; click the folder whose name is this PR's number.

PR #2191, opened 2018-05-15
Solve bugs in IK and JR tools

mitkof6:
Fixes issue #<issue_number>

### Brief summary of changes

Fix two bugs presented in the tools.  

Bug 1: When InverseKinematicsTool finishes the Kinematics analysis is deleted, which caused a segmentation fault. The problem was that the Kinematics analysis was allocated on the stack. I am surprised that this was working in the previous versions. It is possible that I detected this error because I am working on archlinux which fetches the newest versions of packages (e.g. clang, etc.).

Bug 2: When JointReaction is performed and the muscle forces are applied I was getting segmentation fault in python (#2076). I tried to investigate this error through c++ so I can detect the cause, and it turn out that it was not a python problem. The analyze tool wasn't working either both on Linux and Windows. It turns out that the cast 

`const ScalarActuator* act = dynamic_cast<const ScalarActuator*>(&actuatorSet[actuatorIndex]);`

in the following lines failed because we had 

`const Set<Actuator>* actuatorSet = &_model->getActuators();`

and when I changed to 

` const auto& actuatorSet = _model->getActuators();`

In theory it is the same but for some reason I was getting the error. Now I am able to perform the JR analysis. 

### Testing I've completed

I have scripts where I have automated the execution of IK, ID, SO, JRA. They worked with v3.3 but when I migrated to v4.0 I had problems with IK and JRA. 

### CHANGELOG.md (choose one)

- updated...

### DIFF
--- a/OpenSim/Tools/InverseKinematicsTool.cpp
+++ b/OpenSim/Tools/InverseKinematicsTool.cpp
@@ -274,10 +274,10 @@ bool InverseKinematicsTool::run()
         IO::chDir(directoryOfSetupFile);
 
         // Define reporter for output
-        Kinematics kinematicsReporter;
-        kinematicsReporter.setRecordAccelerations(false);
-        kinematicsReporter.setInDegrees(true);
-        _model->addAnalysis(&kinematicsReporter);
+        Kinematics* kinematicsReporter = new Kinematics();
+        kinematicsReporter->setRecordAccelerations(false);
+        kinematicsReporter->setInDegrees(true);
+        _model->addAnalysis(kinematicsReporter);
 
         cout<<"Running tool "<<getName()<<".\n";
 
@@ -304,7 +304,7 @@ bool InverseKinematicsTool::run()
         ikSolver.setAccuracy(_accuracy);
         s.updTime() = start_time;
         ikSolver.assemble(s);
-        kinematicsReporter.begin(s);
+        kinematicsReporter->begin(s);
 
         const clock_t start = clock();
         double dt = 1.0/markersReference.getSamplingFrequency();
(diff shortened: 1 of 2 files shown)


OpenSim/Tools/InverseKinematicsTool.cpp
```diff
@@ -362,17 +362,17 @@ bool InverseKinematicsTool::run()
...
 
         // Do the maneuver to change then restore working directory 
         // so that output files are saved to same folder as setup file.
         if (_outputMotionFileName!= "" && _outputMotionFileName!="Unassigned"){
-            kinematicsReporter.getPositionStorage()->print(_outputMotionFileName);
+            kinematicsReporter->getPositionStorage()->print(_outputMotionFileName);
         }
         // Once done, remove the analysis we added
-        _model->removeAnalysis(&kinematicsReporter);
+        _model->removeAnalysis(kinematicsReporter);
```
chrisdembia: Are you sure this memory isn't being leaked?

chrisdembia:
> @mitkof6 I made a similar fix last year on the 4.0 code base. Please verify against 4.0

@aymanhab this PR is on 4.0.

@mitkof6 , this is the PR that @aymanhab is referring to: https://github.com/opensim-org/opensim-core/pull/2046/files This is likely the source of the merge conflict.

OpenSim/Tools/InverseKinematicsTool.cpp
mitkof6: Yes because the problem was that the delete operator was called on kinematicsReporter

https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/ArrayPtrs.h#L752

mitkof6:
@chrisdembia and @aymanhab in any case the fix in the JointReaction is also important since it is not working. Can you please verify this on your working version of OpenSim.

chrisdembia:
This PR no longer fixes any bug; it simply uses a reference instead of a pointer. Still, this is cleaner and I'll merge this.

PR #2194, opened 2018-05-18
Revert "Avoid some SWIGTYPE_p_... types"

chrisdembia:
Reverts opensim-org/opensim-core#2176 to fix https://github.com/opensim-org/opensim-gui/issues/763

@aymanhab before merging this, it would be good to know if https://github.com/opensim-org/opensim-gui/issues/763 could be fixed by updating the `.java` files in the GUI repo.


aymanhab:
@chrisdembia I verified that the changes in ComponentList are responsible by undoing the change only in ComponentList.h and being able to build, launch GUI and load models. Likely reason is that while cbegin, cend are not used directly, they may be called by some other instantiations of the template ComponentsList. This is platform specific since the issue doesn't reproduce on windows. Merging.

Issue #2025, opened 2017-11-23
c3d reading of moments is incorrect. 


jimmyDunne:
C3DFileAdapter() uses btk to read a c3d file and produce an OpenSim table with (apparently) forces, COP and moments. Typically, you would want the ground reaction forces, moments, and COP but you (btk) can output the forces, wrenches, and point of wrench application (PWA). If you want to compute the free moment, you would need the forceplate forces, moments, and COP— not the wrenches. 

Unfortunately, we seem to be using the wrenches in our output table. 

``` C++
// Get ground reaction wrenches for the force platform.
        auto ground_reaction_wrench_filter = 
            btk::GroundReactionWrenchFilter::New();
        ground_reaction_wrench_filter->SetInput(*platform);
        auto wrench_collection = ground_reaction_wrench_filter->GetOutput();
        ground_reaction_wrench_filter->Update();
        
        for(auto wrench = wrench_collection->Begin();
            wrench != wrench_collection->End(); 
            ++wrench) {
            // Forces time series.
            fp_force_pts->InsertItem((*wrench)->GetForce());
            // Moment time series.
            fp_moment_pts->InsertItem((*wrench)->GetMoment());
            // Position time series.
            fp_position_pts->InsertItem((*wrench)->GetPosition());
```

I don't think this is the correct thing to do. Users would expect the forceplate moments and COP not the wrenches and PWA. 

I would like to;
i) Change the standard output table to use forceplate COP and forceplate moments
ii) Add check to automatically convert mm and Nmm to m and Nm
iii) Compute forceplate free moment from forceplate data and include in output table
iv) Add a method to output a flattened table (save users on scripting side from flattening)

If possible, I would like someone to double check my interpretation of what we are currently doing in [C3DFileAdapter](https://github.com/opensim-org/opensim-core/blob/307eb6ecf3f91efd73e5b23f28fb1cb74d779e24/OpenSim/Common/C3DFileAdapter.cpp) and offer feedback on my proposed steps to improve this class. Perhaps @aseth1 or @tkuchida if you have time?

aseth1:
@jimmyDunne 
In i) do you mean the force-plate moments that result in the equivalent wrench at the COP, or the moment readings from the force-plate (usually about a corner of the force-plate? I think we should always provide the same effective wrench and make sure that if we provide include the COP that the moments are about that point. 

[...]

PR #2198, opened 2018-05-30
Express c3d forces at optional location

aseth1:
Fixes issue #2025

### Brief summary of changes
Added an optional location to `C3DFileAdapter::read()` to express force-plate measurements (force and moment) either at the fixed *Origin* of the force-plate or at the center-of-pressure, *COP*, or what BTK  calls the point of wrench application, *PWA*.  

Changes were introduced to [BTKCore](https://github.com/opensim-org/BTKCore/pull/3) and exposed through the `C3DFileAdapter`.

### Testing I've completed

This PR tests that that the markers and forces read in match the data and interpretation of Mokka for markers and our existing MATLAB scripts for reading in C3D files, which form the standard (*std_*) files used for comparison as *.trc* for marker data and *.sto* for forces. Tests include real C3D data from walking with missing markers.

### CHANGELOG.md (choose one)
- no need to update because addresses a bug in C3DFileAdapter.  Data handling and Tables and Adapters should be added to the CHANGELOG, but that should be a separate PR.


OpenSim/Auxiliary/auxiliaryTestFunctions.h
```diff
@@ -43,6 +43,9 @@ void ASSERT_EQUAL(T expected,
                   std::string message = "") {
     if (found < expected - tolerance || found > expected + tolerance)
         throw OpenSim::Exception(message, file, line);
+    // A variable is not equivalent to itself only if it is NaN
+    else if( (found != found)  && (expected != expected) )
```
tkuchida: Three questions here:
1. Why is this `&&` rather than `||` (i.e., throw if *either* `found` *or* `expected` is NaN)? I don't think `ASSERT_EQUAL` should report that `5` is equal to `NaN`.
2. Why not use `SimTK::isNaN()` rather than the somewhat less readable `if (a != a)`?
3. I would have expected this condition to be caught before doing anything else:
```cpp
{
    if ( SimTK::isNaN(found) || SimTK::isNaN(expected) )
        throw ...
    // inputs are valid; proceed with algorithm
    ...
}
```

OpenSim/Auxiliary/auxiliaryTestFunctions.h
```diff
@@ -51,6 +54,8 @@ void ASSERT_EQUAL(const SimTK::Vec<M, ELT, STRIDE>& vecA,
                   int line = -1,
                   const std::string& message = "") {
     try {
+        if (vecA.isNaN() && vecB.isNaN())
```
tkuchida: Not `||`? Same comment for L72.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
```diff
@@ -64,12 +69,33 @@ void ASSERT_EQUAL(const SimTK::Vec<M, ELT, STRIDE>& vecA,
...
+                    std::string file = "",
+                    int line = -1,
+                    std::string message = "") {
+
+    if (vecA.size() != vecB.size()) {
+        throw OpenSim::Exception(message, file, line);
+    }
+    else {
+        for (size_t i = 0; i < vecA.size(); ++i) {
+            if (vecA[i] < vecB[i] - tolerance || vecA[i] > vecB[i] + tolerance) {
```
tkuchida: Should there be a check here to ensure neither `vecA[i]` nor `vecB[i]` is NaN?

OpenSim/Common/C3DFileAdapter.h
```diff
@@ -37,6 +37,12 @@ class OSIMCOMMON_API C3DFileAdapter : public FileAdapter {
     typedef std::vector<Event>                         EventTable; 
     typedef std::map<std::string, std::shared_ptr<TimeSeriesTableVec3>> Tables;
 
+    enum class ForceLocation {
```
tkuchida: How are these enums accessed through Python/MATLAB? Perhaps provide example MATLAB code (see, e.g., [PlanarJoint::Coord](http://myosin.sourceforge.net/2198/classOpenSim_1_1PlanarJoint.html#aebd231f18a083296d612564a9c60f462)).

OpenSim/Common/C3DFileAdapter.cpp
```diff
@@ -108,13 +112,18 @@ C3DFileAdapter::extendRead(const std::string& fileName) const {
...
+            SimTK::RowVector_<SimTK::Vec3> row{ marker_pts->GetItemNumber(), 
+                                                SimTK::Vec3(SimTK::NaN) };
             int m{0};
             for(auto it = marker_pts->Begin();  it != marker_pts->End(); ++it) {
                 auto pt = *it;
-                row[m++] = SimTK::Vec3{pt->GetValues().coeff(f, 0),
-                                       pt->GetValues().coeff(f, 1),
-                                       pt->GetValues().coeff(f, 2)};
+                if (!( pt->GetValues().row(f).isZero() ||  //not precisely zero
+                    (pt->GetResiduals().coeff(f) == -1))) {//nor is residual -1
```
tkuchida: This condition is confusing, which is probably why you wrote the comment. I suggest removing the comment and replacing `if (!( a || b ))` with `if (!a && !b)` which is equivalent and self-commenting. Might be useful to include a comment that explains what's going on here. What does a residual of -1 mean? Perhaps comment before L115: "Initialize table with SimTK::NaN and fill in only the quantities that are valid."

OpenSim/Common/Test/testC3DFileAdapter.cpp
```diff
@@ -31,75 +31,132 @@
...
+                      table2.getIndependentColumn(), tolerance,
+                       __FILE__, __LINE__,
+                         "Independent columns are not equivalent.");
     } catch (const OpenSim::KeyNotFound&) {}
 
     const auto& matrix1 = table1.getMatrix();
     const auto& matrix2 = table2.getMatrix();
-    for(int r = 0; r < matrix1.nrow(); ++r)
+
+    assert(nFramesToSkip > 0);
```
tkuchida: Why not use `unsigned`?

OpenSim/Common/Test/testC3DFileAdapter.cpp
```diff
@@ -31,75 +31,132 @@
...
+    for(int r = 0; r < matrix1.nrow(); r=r+ nFramesToSkip)
         for(int c = 0; c < matrix1.ncol(); ++c) {
             auto elt1 = matrix1.getElt(r, c); 
             auto elt2 = matrix2.getElt(r, c);
-            ASSERT_EQUAL(elt1[0], elt2[0], 1e-6);
-            ASSERT_EQUAL(elt1[1], elt2[1], 1e-6);
-            ASSERT_EQUAL(elt1[2], elt2[2], 1e-6);
+
+            ASSERT_EQUAL(elt1, elt2, tolerance, __FILE__, __LINE__,
+                "Elements at row, " + std::to_string(r) + " col, " +
```
tkuchida: The message will say "Elements at row, 1 col, 2 failed...".

OpenSim/Common/Test/testC3DFileAdapter.cpp
```diff
@@ -31,75 +31,132 @@
...
 
-template<typename> class shrik;
-
-void compare_tables(const OpenSim::TimeSeriesTableVec3& table1,
-                    const OpenSim::TimeSeriesTableVec3& table2) {
+template<typename ETY = SimTK::Real>
+void compare_tables(const OpenSim::TimeSeriesTable_<ETY>& table1,
+            const OpenSim::TimeSeriesTable_<ETY>& table2,
+            const double tolerance = SimTK::SignificantReal,
+            const int nFramesToSkip = 1) {
```
tkuchida: This method is called `compare_tables` but skips frames, so two tables may differ even if they pass through `compare_tables`. Before this method returns, I suggest `cout`ing the number of frames that were checked (e.g., "10 of 100 rows were tested and are equal").

OpenSim/Common/Test/testC3DFileAdapter.cpp
```diff
@@ -31,75 +31,132 @@
...
+        << 1.e3*(std::clock() - t0) / CLOCKS_PER_SEC << "ms" << endl;
+
+
+    // Compare C3DFileAdapter read-in and written marker data
+    compare_tables<SimTK::Vec3>(markers, *marker_table);
+    // Compare C3DFileAdapter written marker data to standard
+    // Note std exported from Mokka with only 5 decimal places 
+    compare_tables<SimTK::Vec3>(markers, std_markers, 1e-4, 10);
+
+    cout << "\tMarkers " << marker_file << " equivalent to standard" << endl;
```
tkuchida: You cannot draw this conclusion because only 10% of the rows are being tested.

OpenSim/Common/Test/testC3DFileAdapter.cpp
```diff
@@ -31,75 +31,132 @@
...
+    auto std_forces = sto_adapter.read("std_" + forces_file);
+    // Compare C3DFileAdapter read-in and written forces data
+    compare_tables<SimTK::Vec3>(forces.pack<SimTK::Vec3>(), 
+                                *force_table,
+                                SimTK::SqrtEps, 100);
+    // Compare C3DFileAdapter written forces data to standard
+    // Note std generated using MATLAB C3D processing scripts 
+    compare_tables(forces, std_forces, SimTK::SqrtEps, 100);
+
+    cout << forces_file << " equivalent to std_" << forces_file << endl;
```
tkuchida: Testing only 1% of the rows. Same comment for L156.

aseth1:
> I don't think we should be adding huge data files to the repo. We could be bloating the size of the repo with no added benefit. You should only need a few lines of the data files for these tests.

I don't disagree with this sentiment. @chrisdembia do you propose: 
1.  removing the *walking#.c3d* files from the *shared* test folder and replace them with shorter files and keeping the test case intact?
2. keep the  *walking#.c3d* files but down-sample the resulting tables and writing out them out .trc and .sto files for comparison to similarly down-sampled standard?

The *walking#.c3d* files are real data collection files of human gait with different force-plate configurations, so there is value in testing that we can load and interpret these with out issues. They also have markers come in and out of view and are somewhat typical of the size of mocap files. Perhaps option 2, then?

OpenSim/Common/C3DFileAdapter.h
```diff
@@ -45,14 +51,23 @@ class OSIMCOMMON_API C3DFileAdapter : public FileAdapter {
...
+    }
     
     /** Read in a C3D file into separate markers and forces tables of type
         TimeSeriesTableVec3. The markers table has each column labeled by its
         corresponding marker name. For the forces table, the data are grouped
         by sensor (force-plate #) in force, point and moment order, with the
         respective *f#*, *p#* and *m#* column labels. */
     static
-    Tables read(const std::string& fileName);
+    Tables read(const std::string& fileName,
```
tkuchida: Perhaps add a comment to the doxygen indicating the scenarios in which one might want to use each force location?

aseth1:
@chrisdembia @tkuchida which of the two options above,to reduce repo bloat, do you prefer? In both cases, the option to skip frames can be removed with smaller data sets.

aseth1:
Talking with @jimmyDunne he makes a good point that realistic C3D data files (and their size) should be kept and that the test should include a minimum load time for these files so that slowdowns ( like #1698) are not inadvertently introduced.  I will add that. I'll also move to implementing option 2. above and reduce the size of the standards rather than down sampling upon comparison.

OpenSim/Common/C3DFileAdapter.h
```diff
@@ -37,6 +37,41 @@ class OSIMCOMMON_API C3DFileAdapter : public FileAdapter {
...
+
+        <b>Java example</b>
+        \code{.java}
+        tables = C3DFileAdapter.read("myData.c3d",
+                    C3DFileAdapter.ForceLocation.CenterOfPressure);
+        \endcode
+
+        <b>MATLAB example</b>
+        \code{.m}
+        tables = C3DFileAdapter.read("myData.c3d", 1);
```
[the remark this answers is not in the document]
aseth1: Sorry, I meant to add a scripting friendly extension to the SWIG wrapping (*common.i*). I added it and I just verified that it works both in MATLAB and Python.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
```diff
@@ -33,24 +33,61 @@
...
                   T tolerance, 
                   std::string file = "", 
                   int line = -1, 
                   std::string message = "") {
+    // if both values are NaN treat them as being equivalent for the
+    // sake of comparing experimental data and results where NaNs are
+    // possible
+    if(SimTK::isNaN(found) && SimTK::isNaN(expected))
+        return;
+    if (found < expected - tolerance || found > expected + tolerance)
```
tkuchida: How does NaN compare to Real? Perhaps provide a comment here, or include a case that explicitly throws if *either* `found` *or* `expected` is `NaN`.

OpenSim/Common/Test/testC3DFileAdapter.cpp
```diff
@@ -31,75 +31,163 @@
...
+                std::to_string(c) + " failed to have matching value.");
         }
 }
 
+template<typename ETY = SimTK::Real>
+void downsample_table(OpenSim::TimeSeriesTable_<ETY>& table,
+    const unsigned int increment) {
+    for (size_t r = table.getNumRows() - 2; r > 0; --r) {
+        if (r%increment)
+            table.removeRowAtIndex(r);
```
tkuchida: This looks like it might be inefficient. Does the table resize/reorganize in memory when rows are removed? I expected a new table to be created. Also, because you're stepping through backwards, it seems like you could end up with strange gaps at the beginning (e.g., rows 1, 2, 5, 8, 11, ...).

tkuchida: A table with 10 rows will end up with rows 0, 1, 2, 4, 5, 7, 8, 9, which is an unexpected result after "downsampling" by 3.

aseth1: > This looks like it might be inefficient. Does the table resize/reorganize in memory when rows are removed? I expected a new table to be created.

Efficiency isn't of any real concern here. Down sampling added no significant time to the test, and the downstream time savings are much more significant. In any case, removing in place does not reallocate.

> A table with 10 rows will end up with rows 0, 1, 2, 4, 5, 7, 8, 9, which is an unexpected result after "downsampling" by 3. 

It removes whenever their is a remainder, so in this case t looks like [0, 3, 6, 9] when the increment is 3.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
aseth1: A NaN to real will always fail. In fact a NaN with anything fails including another NaN. That is why I have to check of both are NaN if I want the comparison to pass in that case (which we do).

OpenSim/Common/Test/testC3DFileAdapter.cpp
tkuchida: Oh ok, my bad. Thanks for clarifying.

tkuchida: I still think this algorithm produces an unexpected result when the number of rows in the table is not 1 greater than a multiple of `increment`.

| Number of rows in original table | Rows retained when `increment` is 3 | Expected |
| --- | --- | --- |
| 4 | 0, 3 | 0, 3 |
| 5 | 0, 3, 4 | 0, 3 |
| 6 | 0, 3, 5 | 0, 3 |
| 7 | 0, 3, 6 | 0, 3, 6 |

OpenSim/Common/C3DFileAdapter.cpp
```diff
@@ -108,13 +112,21 @@ C3DFileAdapter::extendRead(const std::string& fileName) const {
...
             int m{0};
             for(auto it = marker_pts->Begin();  it != marker_pts->End(); ++it) {
                 auto pt = *it;
-                row[m++] = SimTK::Vec3{pt->GetValues().coeff(f, 0),
-                                       pt->GetValues().coeff(f, 1),
-                                       pt->GetValues().coeff(f, 2)};
+                // BTK reads empty values as zero, but sets a "residual" value
+                // to -1 and it is how it knows to export these values as 
+                // blank, instead of 0,  when exporting to .trc
+                if (!pt->GetValues().row(f).isZero() &&    //not precisely zero
```
tkuchida: Why is it necessary to include `!pt->GetValues().row(f).isZero()` in the condition? Shouldn't the NaNs (initialized at L115–116) be overwritten with zeros if the zeros are real data (i.e., if "residual" is not -1)? It looks like (0,0,0) becomes NaNs in `row` regardless of the value of the "residual" flag, which seems to contradict the comment.

OpenSim/Common/C3DFileAdapter.h
```diff
@@ -45,14 +80,27 @@ class OSIMCOMMON_API C3DFileAdapter : public FileAdapter {
...
     /** Read in a C3D file into separate markers and forces tables of type
         TimeSeriesTableVec3. The markers table has each column labeled by its
         corresponding marker name. For the forces table, the data are grouped
         by sensor (force-plate #) in force, point and moment order, with the
-        respective *f#*, *p#* and *m#* column labels. */
+        respective *f#*, *p#* and *m#* column labels. C3DFileAdpater provides
+        options for expressing the force-plate measurements either as the
+        net force and moments expressed at the ForcePlateOrigin, the 
+        CentereOfPressure, or the PointOfWrenceApplication (PWA) as defined by
+        Shimba 1984*/
```
tkuchida: Has the complete reference to "Shimba 1984" been provided in the doxygen somewhere? I can't find it in this file. You probably mean "An estimation of center of gravity from force platform data" which is the first hit on Google Scholar, but it might be good to be explicit here since many users may be unfamiliar with wrenches. :wrench:

OpenSim/Common/C3DFileAdapter.h
```diff
@@ -45,14 +80,27 @@ class OSIMCOMMON_API C3DFileAdapter : public FileAdapter {
...
     
     /** Read in a C3D file into separate markers and forces tables of type
         TimeSeriesTableVec3. The markers table has each column labeled by its
         corresponding marker name. For the forces table, the data are grouped
         by sensor (force-plate #) in force, point and moment order, with the
-        respective *f#*, *p#* and *m#* column labels. */
+        respective *f#*, *p#* and *m#* column labels. C3DFileAdpater provides
+        options for expressing the force-plate measurements either as the
+        net force and moments expressed at the ForcePlateOrigin, the 
+        CentereOfPressure, or the PointOfWrenceApplication (PWA) as defined by
```
tkuchida: PointOfWrenceApplication -> PointOfWrenchApplication

OpenSim/Common/Test/testC3DFileAdapter.cpp
```diff
@@ -31,75 +31,163 @@
...
+    }
+}
+
 
 void test(const std::string filename) {
     using namespace OpenSim;
+    using namespace std;
+
+    // The walking C3D files included in this test should not take more
+    // than 40ms on most hardware. We make the max time 1000ms to account
```
tkuchida: Comment "1000ms" disagrees with code on L81.

OpenSim/Simulation/Test/testFrames.cpp
```diff
@@ -450,7 +450,7 @@ void testFilterByFrameType()
         std::cout << "frame[" << ++i << "] is " << component.getName()
             << " of type " << typeid(component).name() << std::endl;
     }
-    ASSERT_EQUAL(8, i, 0, __FILE__, __LINE__,
+    ASSERT_EQUAL(8, i, __FILE__, __LINE__,
```
tkuchida: If `i != 8`, the message will say "failed to find 6 PhysicalFrames".

OpenSim/Simulation/Test/testFrames.cpp
```diff
@@ -471,7 +471,7 @@ void testFilterByFrameType()
         std::cout << "frame[" << ++i << "] is " << component.getName()
             << " of type " << typeid(component).name() << std::endl;
     }
-    ASSERT_EQUAL(5, i, 0, __FILE__, __LINE__,
+    ASSERT_EQUAL(5, i, __FILE__, __LINE__,
```
tkuchida: If `i != 5`, the message will say "failed to find the 3 PhyscicalOffsetFrame".

OpenSim/Common/Test/testC3DFileAdapter.cpp
aseth1: I intentionally left the start and end rows. This method is not available or intended for public use. It is just a quick way to speed up the test and reduce repo bloat with standard data files. It may be useful to have a down-sampling utility for users, but this is not it. That's why it only exists in this test.

aseth1:
>Presumably, the consumer is responsible for handling NaNs and shouldn't be relying on the ASSERT_EQUAL to catch them

I wouldn't put it that way. If the consumer wants `ASSERT _EQUAL` to throw on a `NaN` then they simply don't give it a `NaN` as the expected value. If the expected value is NaN, then the caller should have some rational for that.  For example, if it is experimental data with missing values, in which case they may not have values to be compared. I thought this would be more convenient than having the consumer have to prune NaNs from their data before comparing, which in itself could lead to more bugs/issues.

tkuchida:
> I wouldn't put it that way. If the consumer wants ASSERT _EQUAL to throw on a NaN then they simply don't give it a NaN as the expected value. If the expected value is NaN, then the caller should have some rational for that. For example, if it is experimental data with missing values, in which case they may not have values to be compared. I thought this would be more convenient than having the consumer have to prune NaNs from their data before comparing, which in itself could lead to more bugs/issues.

Seems fine, and I think either way (throwing or not throwing if NaN is encountered) would be acceptable. One interpretation of `ASSERT_EQUAL(a, b, tol)` is that the user wants to perform an arithmetic comparison between two numbers, `a` and `b`; it's an undefined request if `a` and/or `b` isn't a number. However, the `ASSERT_EQUAL` methods are templatized and can also be interpreted as a check that the *objects* are equal/identical, in which case saying that NaN and NaN are equal makes perfect sense.

OpenSim/Common/C3DFileAdapter.cpp
```diff
@@ -108,13 +112,22 @@ C3DFileAdapter::extendRead(const std::string& fileName) const {
...
             for(auto it = marker_pts->Begin();  it != marker_pts->End(); ++it) {
                 auto pt = *it;
-                row[m++] = SimTK::Vec3{pt->GetValues().coeff(f, 0),
-                                       pt->GetValues().coeff(f, 1),
-                                       pt->GetValues().coeff(f, 2)};
+                // BTK reads empty values as zero, but sets a "residual" value
+                // to -1 and it is how it knows to export these values as 
+                // blank, instead of 0,  when exporting to .trc
+                // Read in value if it is not zero or residual is not -1
+                if (!pt->GetValues().row(f).isZero() ||    //not precisely zero
```
tkuchida: I'm still missing why the first condition is necessary. Wouldn't `if (pt->GetResiduals().coeff(f) != -1)` be sufficient? Does BTK set the "residual" flag only in some cases, or is the "residual" flag sometimes wrong? If the data from BTK is untrustworthy, then perhaps there should even be a check here that warns the user if the value and residual are contradictory (i.e., if `!pt->GetValues().row(f).isZero()` yet `pt->GetResiduals().coeff(f) == -1`)?

aseth1: The BTK code explicitly interprets the 0 and the residual to -1 as blank entries. I don't understand why I should not ONLY check for this condition to replace with a NaN. You may be right that the residual in only set to -1 when no value is encountered, but I do not know that for sure. I am 100% certain about the case I am testing for.

tkuchida: > I don't understand why I should not ONLY check for this condition to replace with a NaN.

I'm not a BTK expert so I'm going by the comment, which seems to imply that the signal is simply whether `residual == -1`. The first half of the condition will accept nonzero values *regardless* of the state of the `residual` flag. I don't know whether nonzero values should be trusted if `residual == -1`.

> I am 100% certain about the case I am testing for.

Ok.

aseth1:
@chrisdembia please let me know if your concerns have been addressed.

jimmyDunne:
@chrisdembia FYI, I hadn't finished reviewing this.

aseth1:
> FYI, I hadn't finished reviewing this.

Please still do test in MATLAB. We can always fix/change in follow-up PR. Thanks @jimmyDunne

Issue #2201, opened 2018-06-05
ScaleTool() exception when first time colomn isn't 0


jimmyDunne:
Issue reported on the [forum](https://simtk.org/plugins/phpBB/posting.php?mode=reply&f=91&t=9141&sid=23ef0764eff9fda1dba687f9cce0475b), which I confirmed locally. 

If the time column ~= 0, the ScaleTool returns an exception. This issue is not seen in 3.3.  
```
java.lang.RuntimeException: Time 0.000000 is out of time range [0.517000, 0.517000]
	Thrown at TimeSeriesTable.h:304 in getNearestRowIndexForTime().
	at org.opensim.modeling.opensimActuatorsAnalysesToolsJNI.OpenSimContext_processModelMarkerPlacer__SWIG_0(Native Method)
	at org.opensim.modeling.OpenSimContext.processModelMarkerPlacer(OpenSimContext.java:301)
	at org.opensim.tracking.ScaleToolModel$ScaleToolWorker.construct(ScaleToolModel.java:347)
	at org.opensim.swingui.SwingWorker$2.run(SwingWorker.java:130)
[catch] at java.lang.Thread.run(Thread.java:745)
```

All of our static marker trc files start from time zero so we don't test for this.

A testing folder can be found in Dropbox/OpenSim Beta Testing Files/4.0 Testing/scaling_time_zero

PR #2202, opened 2018-06-05
Specify correct time for IK of the static pose

aseth1:
Fixes issue #2201

### Brief summary of changes
As of 4.0 time coherence is enforced by `TimeSeriesTable` and we throw if the time requested is out of range.  In #2201 the range of data to compute the average frame was 1 to 1.01s but IK was being solved for time = 0, and the `TimeSeriesTable` or marker data correctly throws. 

### Testing I've completed
Example in #2021 runs correctly and tests pass locally.

### CHANGELOG.md (choose one)
- no need to update because. this is a bug fix.

### DIFF
--- a/OpenSim/Tools/MarkerPlacer.cpp
+++ b/OpenSim/Tools/MarkerPlacer.cpp
@@ -261,7 +261,7 @@ bool MarkerPlacer::processModel(Model* aModel,
                                                    _timeRange[1]);
     for(size_t r = staticPoseTable.getNumRows() - 1; r > 0; --r)
         staticPoseTable.removeRowAtIndex(r);
-    staticPoseTable.setRow(_timeRange[0], avgRow);
+    staticPoseTable.appendRow(_timeRange[0], avgRow);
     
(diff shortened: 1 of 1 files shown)


OpenSim/Tools/MarkerPlacer.cpp
```diff
@@ -261,7 +261,7 @@ bool MarkerPlacer::processModel(Model* aModel,
                                                    _timeRange[1]);
     for(size_t r = staticPoseTable.getNumRows() - 1; r > 0; --r)
         staticPoseTable.removeRowAtIndex(r);
-    staticPoseTable.updRowAtIndex(0) = avgRow;
+    staticPoseTable.appendRow(_timeRange[0], avgRow);
```
chrisdembia: I think the table will have 2 rows, but I think it's supposed to have only 1 row. The for-loop removes all but the first row, and then you append a row.

aseth1:
Any further comments or concerns @chrisdembia @jimmyDunne?

PR #2203, opened 2018-06-08
Do not search OPENSIM_HOME to find simbody-visualizer.

chrisdembia:
### Brief summary of changes

In 3.3, the OpenSim installer set the OPENSIM_HOME environment variable
and this was used on Windows to find the simbody-visualizer executable.

In 4.0, OpenSim's installer no longer sets the OPENSIM_HOME environment
variable, which means that this environment variable may point to a
pre-existing OpenSim installation. If OpenSim searches OPENSIM_HOME for
simbody-visualizer, then it's likely that OpenSim 4.0 will find an
incompatible simbody-visualizer from 3.3.

To avoid the above issue, this PR causes OpenSim to no longer search OPENSIM_HOME for simbody-visualizer.

### Testing I've completed

This change was used in the Dynamic Walking competition.

### CHANGELOG.md (choose one)

- updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -211,6 +211,8 @@ programmatically in MATLAB or python.
   its properties are updated during scaling. (PR #1994)
 - The source code for the "From the Ground Up: Building a Passive Dynamic
   Walker Example" was added to this repository.
+- OpenSim no longer looks for the simbody-visualizer using the environment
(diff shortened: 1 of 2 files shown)


chrisdembia:
Thank you @aseth1 and @tkuchida

Issue #766, opened 2016-02-16
Making BTK optional when building OpenSim 


aymanhab:
As of now getting BTK and building it is streamlined, nevertheless it adds a dependency that some users may not need or want (e.g. to follow the API examples, some tutorials or build your own controller or optimization). With the potential of supporting other formats in the future (e.g. Microsoft Kinect) it would be good to not require building all interfaces that are mutually exclusive.

klshrinidhi:
I will try to get this done soon.

chrisdembia:
What should happen when trying to load a C3D file when BTK is not integrated? Should the C3D adapter still be compiled and registered in the file registry, or no? Would it be possible to generate a useful error message that basically says "Your OpenSim needs to have been compiled with BTK to read C3D files"?

klshrinidhi:
@chrisdembia 
Right now BTK is *required*. So OpenSim will not even finish *configuring* without BTK.

I think it is good to keep BTK optional. Will start a PR about it soon.

chrisdembia:
I know; I'm talking about what should happen when it's optional and not included (I'm talking about a runtime error).

klshrinidhi:
Got it.
In that case, since there will be no registered adapter for "c3d" extension, 
* an exception will be thrown when using the generic `FileAdapter` class, and
* compiler will error when attempting to use nonexistent C3DFileAdapter class.

aymanhab:
@chrisdembia Assuming there's a runtime registry of readers that handle specific file extensions then you should get a message like ".c3d" is an unsupported extension, and probably a link or page that describes what formats are supported and how, which would help users add support for new formats as needed.

chrisdembia:
An alternative is to still provide the C3DFileAdapter class even if OpenSim was not compiled with BTK, because all that users need to do is drop in a BTK shared library and it'll work. Right? This would be similar to what Simbody does for CFSQP.

klshrinidhi:
>An alternative is to still provide the C3DFileAdapter class even if OpenSim was not compiled with BTK, because all that users need to do is drop in a BTK shared library and it'll work. Right? This would be similar to what Simbody does for CFSQP.

C3DFileAdapter class needs BTK binaries to compile. It will be more work to get to dynamically check and load BTK binaries and enable C3D functionality.

chrisdembia:
> and probably a link or page that describes what formats are supported and how

I like this idea.

PR #2209, opened 2018-06-26
Expose Object::getDocument() as a const public method.

aseth1:
Necessary to fix [GUI issue #766](https://github.com/opensim-org/opensim-gui/issues/766) 

### Brief summary of changes
Allow `Object` to provide the underling `XMLDocument` and subsequently its version.

### Testing I've completed
Existing tests pass. 

### CHANGELOG.md (choose one)
- no need to update because method was only used internally

### DIFF
--- a/OpenSim/Common/Object.cpp
+++ b/OpenSim/Common/Object.cpp
@@ -1241,10 +1241,15 @@ updateDefaultObjectsXMLNode(SimTK::Xml::Element& aParent)
  *
  * @return Document's filename for this object.
  */
-string Object::
-getDocumentFileName() const
+string Object::getDocumentFileName() const
 {
     return _document ? _document->getFileName() : "";
+}
+
+
+int Object::getDocumentFileVersion() const
+{ 
+    return _document ? _document->getDocumentVersion() : -1;
 }
 
 
--- a/OpenSim/Common/Object.h
+++ b/OpenSim/Common/Object.h
@@ -563,13 +563,14 @@ protected:
     lost here. **/
     void setDocument(XMLDocument* doc) {_document=doc;}
 
-    /** Get a const pointer to the document (if any) associated with this
-    object. **/
-    const XMLDocument* getDocument() const {return _document;}
     /** Get a writable pointer to the document (if any) associated with this
     object. **/
     XMLDocument* updDocument() {return _document;}
 public:
+    /** Get a const pointer to the document (if any) associated with this
+    object. **/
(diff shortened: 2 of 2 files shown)


chrisdembia:
I prefer (2), but I also agree that we should not be sprinkling an explicit number like `30516` throughout the code base.

I think the GUI warning could be generated whenever `Object::getDocumentFileVersion() < XMLDocument::getLatestVersion()`, thereby avoiding a check with a specific version number.

aymanhab:
Thanks for the feedback @chrisdembia, the issue is that we want a prominent dialog in the case when transitioning from pre 4.0 to 4.0 format but to keep it quiet for incremental changes post 4.0, so a generic test may not provide this granularity.

chrisdembia:
I'm not convinced that we should make upgrading to 4.0 a special case. If we do decide to make this a special case, then I think (3) is the correct solution.

aseth1:
@aymanhab why not implement whatever method makes most sense for this case in the `OpenSimContext` now that you can access the document version. I never intended exposing `XMLDocument` or `SimTK::XML` via wrapping. I also don't see the rationale for API having two methods to provide the same information.

aymanhab:
@aseth1 OpenSimContext class is dedicated to methods that interact wit the state backing the Model in the GUI so the method doesn't fit. It seems reasonable to add the DocumentVersion handling in the same class that handles DocumentName but if you prefer not to, we can inject the method during the SWIG binding. The drawback is that it will be hard to document and I believe users who are doing XML processing on Models will need access and documentation anyway. Let me know what you think.

aymanhab:
@aseth1 Let me know when done with your changes and I'll make the SWIG interface changes so that I can test them. Thanks

aseth1:
@aymanhab I  reverted the change and no longer expose `XMLDocument`. From our discussion it seems you don't need access to the `XMLDocument`.  I don't like having both pathways (through XMLDocument or from `Object`), but since this seems more convenient, I went ahead and added  `Object::getDocumentFileVersion()` instead.

aseth1:
@aymanhab you can ignore the failure. I don't know why this one instance is loading C3Ds very slowly.  Since the other tests pass there isn't a systematic slowdown to the C3D reading and this change certainly does not affect that process.

aymanhab:
Can we get another quick review to merge this? @chrisdembia @tkuchida Thx

aymanhab:
Thanks @tkuchida :+1:

PR #2212, opened 2018-07-10
Fix ContactHalfSpace decoration to lie within the contact geometry.

### LINKED ISSUE #2211: Height of ContactHalfSpace geometry is incorrect.
The current placement of geometry for ContactHalfSpace is misleading. The actual halfspace is located through the middle of the brick geometry that is displayed:

screen shot 2018-07-09 at 4 37 53 pm

For `SimTK::DecorativeBrick`, [Simbody's doxygen](https://simbody.github.io/simbody-3.6-doxygen/api/classSimTK_1_1DecorativeBrick.html) says "This defines a rectangular solid centered at the origin and aligned with the local frame axes." The implementation of `ContactHalfSpace::generateDecorations()` contains the following:

https://github.com/opensim-org/opensim-core/blob/801f16943814e4b4dacbb5a35cf1aa8d9ce2e19f/OpenSim/Simulation/Model/ContactHalfSpace.cpp#L77-L86

I think the geometry should be shifted by `-0.0025` in the `x` direction.

chrisdembia:
Fixes #2211

### Brief summary of changes

- This PR fixes the transform for `ContactHalfSpace` so that the half-space's decorative geometry does not protrode outside of the contact geometry (half in, half out).

### Testing I've completed

- I temporarily turned the visualizer on in `testContactGeometry` and ensured that the entire `DecorativeBrick` for the `ContactHalfSpace` appeared below ground.
- I ran and fixed `testVisualization` (this test initially failed after I changed `ContactHalfSpace.cpp`).

### CHANGELOG.md (choose one)

- no need to update because...our current way of displaying `ContactHalfSpace` was not in any previous release.

### DIFF
--- a/OpenSim/Simulation/Model/ContactHalfSpace.cpp
+++ b/OpenSim/Simulation/Model/ContactHalfSpace.cpp
@@ -77,8 +77,14 @@ void ContactHalfSpace::generateDecorations(bool fixed, const ModelDisplayHints& 
     const auto& X_BF = getFrame().findTransformInBaseFrame();
     const auto& X_FP = getTransform();
     const auto X_BP = X_BF * X_FP;
-    geometry.push_back(SimTK::DecorativeBrick(Vec3{.005, 0.5, 0.5})
-        .setTransform(X_BP).setScale(1)
+    const double brickHalfThickness = 0.005;
+    geometry.push_back(
+        SimTK::DecorativeBrick(Vec3{brickHalfThickness, 0.5, 0.5})
+        // The brick is centered on the origin. To ensure the decorative
+        // geometry is within the contact geomtry, we must offset by half
+        // the thickness of the brick.
+        .setTransform(X_BP * Transform(Vec3(brickHalfThickness, 0, 0)))
+        .setScale(1)
         .setRepresentation(get_Appearance().get_representation())
         .setBodyId(getFrame().getMobilizedBodyIndex())
         .setColor(get_Appearance().get_color())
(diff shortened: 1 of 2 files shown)


chrisdembia:
> Would this need adjustment based on whether
contact is on top/bottom

No.

The change I made is correct regardless of the orientation of the contact half-plane.

chrisdembia:
Thanks @aymanhab

OpenSim/Simulation/Model/ContactHalfSpace.cpp
```diff
@@ -77,8 +77,14 @@ void ContactHalfSpace::generateDecorations(bool fixed, const ModelDisplayHints&
     const auto& X_BF = getFrame().findTransformInBaseFrame();
     const auto& X_FP = getTransform();
     const auto X_BP = X_BF * X_FP;
-    geometry.push_back(SimTK::DecorativeBrick(Vec3{.005, 0.5, 0.5})
-        .setTransform(X_BP).setScale(1)
+    const double brickHalfThickness = 0.005;
```
aseth1: The issue here is that we are representing a plane with a brick. I think it is misleading to present a plane with a block with thickness. While I agree using the top surface of the brick is better (than the middle) in reality we shouldn't be presenting any thickness. Ideally we could draw planes. For the time being we could use a thickness of 1mm instead of 1cm?

chrisdembia: I've changed the thickness to 1mm. Here's a comparison:


1cm:

screen shot 2018-07-13 at 11 35 00 am

1mm:

screen shot 2018-07-13 at 11 34 15 am

aseth1: :+1: Looks more plane like to me.

chrisdembia: Ah right. Updated.

tkuchida:
@aseth1's only suggestion was implemented and the AppVeyor failure has already been noted in #2217.

chrisdembia:
Thanks @tkuchida

Issue #2217, opened 2018-07-16
testC3DFileAdapter failing on AppVeyor: "Unable to load 'walking2.c3d' within 100.000000ms."


tkuchida:
<pre>
C3DFileAdapter 'walking2.c3d' loaded in 167ms
testC3DFileAdapter FAILED: Unable to load 'walking2.c3d' within 100.000000ms.
</pre>
https://ci.appveyor.com/project/opensim-org/opensim-core/branch/master/job/9cqyblh302drdukf#L245

See testC3DFileAdapter.cpp#L78-L81:
https://github.com/opensim-org/opensim-core/blob/5a59cd79fc43ef289af0f170c3bff27443be55f7/OpenSim/Common/Test/testC3DFileAdapter.cpp#L78-L81

From PR #2198. @aseth1

aseth1:
I agree this is annoying especially since we had no issues with it failing initially.  The purpose of this condition was to avoid slowdowns creeping back in without us knowing. If we need to make this condition  200ms or longer to pass on CI I can do that, but then the test may be less effective. The  original C3D reading code was taking several seconds, so perhaps even 200ms would still be better than nothing. I am open to suggestions.

tkuchida:
It would be nice if this failure didn't cause ![](https://user-images.githubusercontent.com/4203505/42780960-9be3a852-88f9-11e8-838b-5ad66498cd23.png) to appear on the home page. The test condition is an absolute metric (within 100ms); perhaps it should be relative (e.g., within some multiple of reading another file) since we can't predict latency on the CI computers---or on the computer of a user who is building from source?

aseth1:
> perhaps it should be relative (e.g., within some multiple of reading another file) since we can't predict latency on the CI computers

Great idea! It'll try to swing this.

aymanhab:
The issue reappeared on AppVeyor, the timing test fails in one configuration and not the other! Obviously we can't control how ci servers prioritize jobs. I'm hesitant to double the time in general but maybe we can double it on appveyor machines only since these seem to be much slower. Other thoughts @chrisdembia @tkuchida ?

tkuchida:
@aymanhab The conversation continues in #2218 and #2221. I have the same suggestion now as on Jul 18, 2018: "Perhaps the wall clock speed test should be skipped on CI?"

aymanhab:
[...]

PR #2218, opened 2018-07-16
Update C3D reading timing test so that it is relative to platform specific timing

aseth1:
Fixes issue #2217 

Test fails sporadically since CI machines can vary widely in terms of their read/write performance.

### Brief summary of changes
 This PR instruments the read time for a simple Storage file and scales the maximum read time accordingly.

### Testing I've completed
All testC3DFileAdapter cases pass locally.

### CHANGELOG.md (choose one)
- no need to update because this is an update to test case

PR #2221, opened 2018-07-17
Make C3D read time test relative

aseth1:
Fixes issue #2217 

### Brief summary of changes
Update to #2218 

### Testing I've completed
All tests pass locally.

N/A

### CHANGELOG.md (choose one)
- no need to update

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2221)

### DIFF
--- a/OpenSim/Common/Test/testC3DFileAdapter.cpp
+++ b/OpenSim/Common/Test/testC3DFileAdapter.cpp
@@ -198,16 +198,16 @@ int main() {
     using namespace std::chrono;
 
     steady_clock::time_point startTime = steady_clock::now();
-    Storage("std_walking2_grfs.sto");
+    Storage("test.sto");
     steady_clock::time_point t2 = steady_clock::now();
 
     double loadTime = duration_cast<duration<double>>(t2 - startTime).count();
 
-    // The walking C3D files included in this test should not take more
-    // than 10x the time to load the 'std_walking2_grfs.sto' Storage on most
-    // hardware. We make the max time 50x to account for other potential
-    // slowdowns on CI machines.
-    MaximumLoadTimeInS = 50 * loadTime;
+    // The walking C3D files included in this test are benchmarked against the
+    // load time of a standard 'test.sto' Storage that is 1KB in size. The C3D
+    // are ~1000x larger than the storage. We make the max time 1200x to account
+    // for other potential slowdowns on CI machines.
+    MaximumLoadTimeInS = 1000 * loadTime;
 
     cout << "Platform specific Maximum load time is " 
         << MaximumLoadTimeInS << "s" << endl;
--- a/OpenSim/Common/Test/testC3DFileAdapter.cpp
+++ b/OpenSim/Common/Test/testC3DFileAdapter.cpp
@@ -76,15 +76,16 @@ void test(const std::string filename) {
     using namespace std;
 
     std::clock_t startTime = std::clock();
-    Storage("test.sto");
+    Storage("std_walking2_grfs.sto");
     double loadTime = 1.e3*(std::clock() - startTime) / CLOCKS_PER_SEC;
 
     // The walking C3D files included in this test should not take more
-    // than 20x the time to load the 'test.sto' Storage on most hardware.
-    // We make the max time 50x to account for other potential slowdowns
-    // on CI machines.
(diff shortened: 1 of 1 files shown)


chrisdembia:
```

464The following tests FAILED:
465	  2 - testC3DFileAdapter (Failed)
466Errors while running CTest
467Command exited with code 8
```

:(

aseth1:
It is very suspicious that the `MaximumLoadTimeInMS = 100ms` in the second case. That means the `loadTime` for the standard file was 2ms, which is 2x faster than my machine. In the first case, 'walking2.c3d' loaded in 401ms and passed, which indicates a `loadTime` > 8ms (2x slower than my machine). What could be creating such large variability in `loadTime`?

chrisdembia:
> What could be creating such large variability in loadTime?

I don't know, but it seems like this approach won't work.

aseth1:
The issue seems to stem from `std::clock` not necessarily being monotonic and depends on what other demands are on the core, see [here](https://en.cppreference.com/w/cpp/chrono/c/clock). The correct class to use for comparing timing then, is perhaps[ `std::chrono::steady_clock`](https://en.cppreference.com/w/cpp/chrono/c/clock)?

chrisdembia:
Did you try running the `appveyor/pr` test multiple times?

tkuchida:
> I believe the combination of using std::chrono::steady_clock and using a size-based rationale for scaling the load time of the benchmark file makes this test more robust across different platforms. However, it is also far less stringent, at least locally, where the MaximumLoadTimeInS is close to 2s and loading the C3D files typically takes less than 50ms.

The CPU could be doing other things when loading "test.sto" and/or "walking*.c3d", so I'm not sure wall clock time is the best metric. It also isn't obvious that it would take 1000× longer to load a file that is 1000× larger---wouldn't it depend on, e.g., how fragmented the file is (seek time)? It's difficult to think of a solution without knowing why the CI machine is slow. Perhaps the wall clock speed test should be skipped on CI?

aseth1:
>  I'm not sure wall clock time is the best metric

[`std::chrono::steady_clock`](https://en.cppreference.com/w/cpp/chrono/steady_clock) is not wall clock time. It should be OK at measuring intervals.

> It also isn't obvious that it would take 1000× longer to load a file that is 1000× larger---wouldn't it depend on, e.g., how fragmented the file is (seek time)?

Yes it's possible, but the test files are pretty small. I am hoping that because the benchmark `test.sto` file is tiny it easily fits within one allocation block. Of course, *file open* time could be the dominant time in this case so that the limit reflects the operation of opening the file 1000 times and has nothing to do with time to read data. This seems to be the case since I can read the 1000x sized file typically within 20x the time to load the benchmark file.

> It's difficult to think of a solution without knowing why the CI machine is slow.

Slow machines would be fine. The paradox I was running into was that CI would occasionally read the benchmark file at blazing speeds (faster than my desktop) such that the relative max load time was unrealistically fast and the test fails.

> Perhaps the wall clock speed test should be skipped on CI?

I could remove the test condition altogether and just have the test report the time. We would then have to manually monitor this time to make sure the load times don't creep or jump up again.

tkuchida:
> Yes it's possible, but the test files are pretty small. I am hoping that because the benchmark test.sto file is tiny it easily fits within one allocation block. Of course, file open time could be the dominant time in this case so that the limit reflects the operation of opening the file 1000 times and has nothing to do with time to read data. This seems to be the case since I can read the 1000x sized file typically within 20x the time to load the benchmark file.

My point was that I wouldn't expect a file to take 1000x longer to open if it's 1000x larger, not that files will necessarily be fragmented.

> Slow machines would be fine. The paradox I was running into was that CI would occasionally read the benchmark file at blazing speeds (faster than my desktop) such that the relative max load time was unrealistically fast and the test fails.

Yes, I recognize that the absolute speed of the machine is irrelevant: the issue is that files sometimes take more/less time to read than expected, based on linearly extrapolating the read time of a small file.

> I could remove the test condition altogether and just have the test report the time. We would then have to manually monitor this time to make sure the load times don't creep or jump up again.

That seems reasonable. It's also possible to skip tests in the .yml files.

jimmyDunne:
@aseth1 is this still necessary? With the performance improvements with dealing with tables this seems unnecessary. Can I close and delete?

nickbianco:
Stale PR, closing.

PR #2227, opened 2018-07-23
Call initSystem after adding WrapObject to a GeometryPath.

aymanhab:
Fixes https://github.com/opensim-org/opensim-gui/issues/845

### Brief summary of changes
Path edits require a call to initSystem which was not done for this specific call (while done for other Path editing support functionality. Added the call.

### Testing I've completed
I was able to remove/add-back PathWrap to BRA muscle in Arm26 model and got no exception and Path rendering updated accordingly.

### CHANGELOG.md (choose one)

- no need to update because it is a bugfix

### DIFF
--- a/Bindings/Java/OpenSimJNI/OpenSimContext.cpp
+++ b/Bindings/Java/OpenSimJNI/OpenSimContext.cpp
@@ -292,6 +292,9 @@ void OpenSimContext::setStartPoint(PathWrap& mw, int newStartPt) {
 
 void OpenSimContext::addPathWrap(GeometryPath& p, WrapObject& awo) {
     p.addPathWrap( awo );
+    // Adding WrapObject to a GeometryPath requires initSystem similar to other Path edit operations
+    recreateSystemKeepStage();
+    p.updateGeometry(*_configState);
     return;
 }
 


aymanhab:
@aseth1 this is the fix to call init System when a wrapObject is added to a GeometryPath. It brings this function in line with the rest of the functions in OpenSimContext that handle GeometryPath editing.

Bindings/Java/OpenSimJNI/OpenSimContext.cpp
```diff
@@ -292,6 +292,9 @@ void OpenSimContext::setStartPoint(PathWrap& mw, int newStartPt) {
 
 void OpenSimContext::addPathWrap(GeometryPath& p, WrapObject& awo) {
     p.addPathWrap( awo );
+    // Adding WrapObject to a GeometryPath requires initSystem similar to other Path edit operations
+    recreateSystemKeepStage();
```
aseth1: This seems inline with other model editing changes where you need to see the change take effect on a live system.

aymanhab:
Thanks @aseth1  for the quick turnaround :+1:
Can we get another reviewer for this quick one: @chrisdembia

chrisdembia:
The `appveyor/ci` failure is unrelated. Merging.

aymanhab:
Awesome, thanks @chrisdembia and @aseth1

Issue #2230, opened 2018-07-25
ExpressionBasedBushingForce generateDecorations assumes Dynamics stage realization


aymanhab:
ExpressionBasedBushingForce::generateDecorations assumes Forces are computed and use that to generate visuals. This is a bug that needs fixing by either realizing to desired stage or checking that stage is realized as needed.

PR #2235, opened 2018-08-01
Generate decorations does not realize to stage

aseth1:
Fixes issue #2230 and related to https://github.com/opensim-org/opensim-gui/issues/427

### Brief summary of changes
Removed calls to `realizeDynamics()` within `generateDecorations()`. Documented the additional rule for implementing `generateDecorations` for custom user components that there must be a check that the required stage has been achieved before trying to draw stage dependent quantities beyond `Position`

### Testing I've completed
Build the GUI can could load the model. It does not display the forces. Verified that the forces are still drawn in the API visualizer during forward simulation.

Does anyone foresee any unnecessary limitation by imposing this "no realizing" rule on user overrides of `generateDecorations`? I think it is much safer and avoids side-effects like attempting to compute forces during IK.

### CHANGELOG.md (choose one)
- no need to update because because this is a bug fix.

### DIFF
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -480,6 +480,12 @@ public:
     obtain ground- and body-fixed geometry (with \a fixed=\c true), and then 
     once per frame (with \a fixed=\c false) to generate on-the-fly geometry such
     as rubber band lines, force arrows, labels, or debugging aids.
+
+    Please note, that the state passed to generateDecorations is only guaranteed
+    to be realized to Stage::Position. If your component can visualize quantities
+    realized and Velocity, Dynamics or Acceleration stages, then you must check
+    that the required stage has been realized before using requesting realized
+    values.
   
     If you override this method, be sure to invoke the base class method first, 
     using code like this:
@@ -493,6 +499,14 @@ public:
         // invoke parent class method
         Super::generateDecorations(fixed,hints,state,appendToThis); 
         // ... your code goes here
+        // can render velocity dependent quanities if stage is Velocity or higher
+        if(state.getSystemStage() >= Stage::Velocity) {
+            // draw velocity vector for model COM
+        }
+        // can render computed forces if stage is Dynamics or higher
+        if(state.getSystemStage() >= Stage::Dynamics) {
+            // change the length of a force arrow based on the force in N
+        }
     }
     @endcode
 
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -481,7 +481,7 @@ public:
     once per frame (with \a fixed=\c false) to generate on-the-fly geometry such
     as rubber band lines, force arrows, labels, or debugging aids.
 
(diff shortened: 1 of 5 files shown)


OpenSim/Common/Component.h
```diff
@@ -480,6 +480,14 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Component, Object);
     obtain ground- and body-fixed geometry (with \a fixed=\c true), and then 
     once per frame (with \a fixed=\c false) to generate on-the-fly geometry such
     as rubber band lines, force arrows, labels, or debugging aids.
+
+    Please note, that the state passed to generateDecorations is only guaranteed
+    to be realized to Stage::Position. If your component can visualize quantities
```
chrisdembia: How is the `Position` guarantee enforced?

chrisdembia:
Could you describe testing you did related to muscle path coloring?

chrisdembia:
Please further document the `Position` guarantee if you think it's necessary. It's not clear to me that the `Position` guarantee exists. Other than this, this LGTM.

I'm comfortable with merging this despite CI failures (the failures seem unrelated).

OpenSim/Common/Component.h
aseth1: *Guarantee* is perhaps too strong. Just during `Model::initSystem()` we realize to `Position` so any use immediate after getting the state will be guaranteed. 

This is of course not the case if copy the state or do other things.  There is an expectation that state is at `Position` stage since all frames, bodies and points are being drawn with out explicit realization to `Position`.

From the point of view or our Tools and what they do, IK states are realized to `Position`, AnalyzeTool to `Velocity` and `ForwardTool` uses an Integrator that realizes the current state to `Acceleration`. In the case of a the `ForwardTool` it computes the next state and that state I believe is only realized to `Time`.

Should I soften the language? We can also make the drawing code realize to  `Position` explicitly.

chrisdembia: I think you could say that there is a precondition that the state is realized to `Position`. No need for explicitly realizing.

aseth1:
> The only concern is if there's some other code on our side or users that was relying on generateDeocrations realizing to Dynamics that will break now, better now than later.

I agree  that is a concern. On our side, I don't think there is anything that will break further than this PR addresses, but we don't know about user code. The overriding of `generateDecorations` is likely to be limited at the moment, so now is the time change to have minimal impact.

OpenSim/Common/Component.h
aseth1: Done.

chrisdembia:
Thanks @aseth1. I committed a typo fix. LGTM. Feel free to merge when tests pass.

aseth1:
I believe the Mac build failures are unrelated to this change. If that's correct @chrisdembia do you mind merging?

chrisdembia:
I've merged but because of the Travis-CI failure, the change will not yet propagate to the GUI builds.

PR #2237, opened 2018-08-02
{WIP} Update C3D Matlab Class

jimmyDunne:
### Brief summary of changes
Primary osimC3D Matlab class updates to include
- ForceLocation specification
- TRC and Mot writing

Secondarily, update osimTableToStruct.m to
- recognize more illegal chars for Matlab Struct() 

### Todo
- add tests for written trc and mot files can be read by the API

### DIFF
--- a/Bindings/Java/Matlab/Utilities/osimC3D.m
+++ b/Bindings/Java/Matlab/Utilities/osimC3D.m
@@ -1,3 +1,25 @@
+% ----------------------------------------------------------------------- %
+% The OpenSim API is a toolkit for musculoskeletal modeling and           %
+% simulation. See http://opensim.stanford.edu and the NOTICE file         %
+% for more information. OpenSim is developed at Stanford University       %
+% and supported by the US National Institutes of Health (U54 GM072970,    %
+% R24 HD065690) and by DARPA through the Warrior Web program.             %
+%                                                                         %
+% Copyright (c) 2005-2018 Stanford University and the Authors             %
+% Author(s): James Dunne                                                  %
+%                                                                         %
+% Licensed under the Apache License, Version 2.0 (the "License");         %
+% you may not use this file except in compliance with the License.        %
+% You may obtain a copy of the License at                                 %
+% http://www.apache.org/licenses/LICENSE-2.0.                             %
+%                                                                         %
+% Unless required by applicable law or agreed to in writing, software     %
+% distributed under the License is distributed on an "AS IS" BASIS,       %
+% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or         %
+% implied. See the License for the specific language governing            %
+% permissions and limitations under the License.                          %
+% ----------------------------------------------------------------------- %
+
 classdef osimC3D < matlab.mixin.SetGet
 % osimC3D(filepath, ForceLocation)
 %   C3D data to OpenSim Tables.
@@ -27,6 +49,9 @@ classdef osimC3D < matlab.mixin.SetGet
                 error('File does not exist. Check path is correct')
             else
                 [path, name, ext] = fileparts(path2c3d);
+                if isempty(path)
+                    error('Input path must be full path to file (C:/data/Walking.mot)')
+                end
                 obj.path = path;
                 obj.name = name;
             end
--- a/Bindings/Java/Matlab/Utilities/osimC3D.m
+++ b/Bindings/Java/Matlab/Utilities/osimC3D.m
@@ -5,7 +5,8 @@ classdef osimC3D < matlab.mixin.SetGet
 %   Inputs:
 %   filepath           Full path to the location of the C3D file
 %   ForceLocation      Integer value for representation of force from plate
-%                      0 = forceplate orgin, 1 = Forceplate surface, 2 = COP
+%                      0 = forceplate orgin, 1 = COP, 2 = Point Of Wrench
+%                      Application
     properties (Access = private)
         path
         name
--- a/Bindings/Java/Matlab/Utilities/osimC3D.m
+++ b/Bindings/Java/Matlab/Utilities/osimC3D.m
@@ -5,7 +5,7 @@ classdef osimC3D < matlab.mixin.SetGet
 %   Inputs:
 %   filepath           Full path to the location of the C3D file
 %   ForceLocation      Integer value for representation of force from plate
-%                      0 = forceplate orgin, 1 = surface, 2 = COP
+%                      0 = forceplate orgin, 1 = Forceplate surface, 2 = COP
     properties (Access = private)
         path
         name
@@ -124,9 +124,9 @@ classdef osimC3D < matlab.mixin.SetGet
         end
         function writeTRC(obj,varargin)
             % Write marker data to trc file.
-            % osimC3d.writeTRC()                       Write to current dir using c3d name
-            % osimC3d.writeTRC('Walking.trc')          Write to current dir with input name
-            % osimC3d.writeTRC('C:\data\Walking.trc')  Write to input location from full input path
+            % osimC3d.writeTRC()                       Write to dir of input c3d.
+            % osimC3d.writeTRC('Walking.trc')          Write to dir of input c3d with defined file name.
+            % osimC3d.writeTRC('C:/data/Walking.trc')  Write to defined path input path.
 
             % Compute an output path to use for writing to file
             outputPath = generateOutputPath(obj,varargin,'.trc');
@@ -134,14 +134,14 @@ classdef osimC3D < matlab.mixin.SetGet
 
             import org.opensim.modeling.*
             % Write to file
-            TRCFileAdapter().write( obj.getTable_markers, outputPath)
+            TRCFileAdapter().write( obj.markers, outputPath)
             disp(['Marker file written to ' outputPath]);
         end
         function writeMOT(obj,varargin)
-         % Write forces data to mot file
-         % osimC3d.writeMOT()                       Write to current dir using c3d name
-         % osimC3d.writeMOT('Walking.mot')          Write to current dir with input name
-         % osimC3d.writeMOT('C:\data\Walking.mot')  Write to input location from full input path
+        % Write force data to trc file.
+        % osimC3d.writeMOT()                       Write to dir of input c3d.
+        % osimC3d.writeMOT('Walking.mot')          Write to dir of input c3d with defined file name.
+        % osimC3d.writeMOT('C:/data/Walking.mot')  Write to defined path input path.
 
          % Compute an output path to use for writing to file
          outputPath = generateOutputPath(obj,varargin,'.mot');
@@ -151,9 +151,12 @@ classdef osimC3D < matlab.mixin.SetGet
          forces = obj.getTable_forces();
          % Get the column labels
          labels = forces.getColumnLabels();
-         % make a copy
+         % Make a copy
          updlabels = labels; 
           
+         % Labels from C3DFileAdapter are f1, p1, m1, f2,...
+         % We edit them to be consistent with requirements of viewing 
+         % forces in the GUI (ground_force_vx, ground_force_px,...)
          for i = 0 : labels.size() - 1
             % Get the label as a string
             label = char(labels.get(i));
@@ -243,7 +246,8 @@ classdef osimC3D < matlab.mixin.SetGet
             
             % Validate the output filename
             if size(path,2) > 1
-                % More than 1 input to the method, only takes one (filepath)
+                % Path object should be of size == 1, any larger and user
+                % input multiple variables into function. 
                 error([ num2str(size(path,2)) ' inputs, expecting zero or one'])
             end
         
@@ -255,12 +259,10 @@ classdef osimC3D < matlab.mixin.SetGet
             else
             
                 if ~ischar(path{1})
-                   % Path must be a string  
                    error('Input must be a sting of characters')
                 end
 
                 if isempty(strfind(path{1}, ext))
-                   % Path must have an input path
                    error(['Input must be a path to a ' ext ' file']);
                 end
             
--- a/Bindings/Java/Matlab/Utilities/osimC3D.m
+++ b/Bindings/Java/Matlab/Utilities/osimC3D.m
@@ -8,6 +8,7 @@ classdef osimC3D < matlab.mixin.SetGet
 %                      0 = forceplate orgin, 1 = surface, 2 = COP
     properties (Access = private)
         path
+        name
         markers
         forces
         ForceLocation
@@ -24,7 +25,9 @@ classdef osimC3D < matlab.mixin.SetGet
             if exist(path2c3d, 'file') == 0
                 error('File does not exist. Check path is correct')
             else
-                obj.path = path2c3d;
+                [path, name, ext] = fileparts(path2c3d);
+                obj.path = path;
+                obj.name = name;
             end
             % Verify the ForceLocation input is correct
             if  ~isnumeric(ForceLocation) || ~ismember(ForceLocation,[0:2])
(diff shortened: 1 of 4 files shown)


Bindings/Java/Matlab/Utilities/osimC3D.m
```diff
@@ -1,30 +1,67 @@
...
-% osimC3D C3D data into OpenSim Tables
-%   Utility  Class that uses OpenSim's C3DFileAdapter() class to produce
-%   OpenSim tables. Methods include Simbody rotation.
+% osimC3D(filepath, ForceLocation)
+%   C3D data to OpenSim Tables.
+%   OpenSim Utility Class
+%   Inputs:
+%   filepath           Full path to the location of the C3D file
+%   ForceLocation      Integer value for representation of force from plate
+%                      0 = forceplate orgin, 1 = surface, 2 = COP
```
chrisdembia: What do you mean by `surface`; does `surface` mean "PointOfWrenchApplication"? Based on the code below, isn't COP denoted by `1`?

Bindings/Java/Matlab/Utilities/osimC3D.m
```diff
@@ -123,5 +237,43 @@ function rotateTable(obj, table, axisString, value)
...
         end
+        function outputPath = generateOutputPath(obj,path, ext)
+            % Function to generate an output path from no, partial, or fully
+            % defined user path. 
+            
+            % Validate the output filename
+            if size(path,2) > 1
+                % More than 1 input to the method, only takes one (filepath)
+                error([ num2str(size(path,2)) ' inputs, expecting zero or one'])
+            end
```
chrisdembia: This warning might belong in `writeTRC()`, which takes zero or one argument. This function (`generateOutputPath()`) takes 2 arguments.

Bindings/Java/Matlab/Utilities/osimTableToStruct.m
```diff
@@ -77,14 +77,16 @@
...
-    % Remove an initial slash.
-    col_label = regexprep(col_label, '^/', '');
-    % Replace '/' and '|' if they are present.
-    col_label = strrep(col_label,'/', '_');
-    col_label = strrep(col_label,'|', '_');
+    illegalChars = [{' '} {'*'} {'.'} {'/'} {'-'} {'|'}];
+    for i = 1 : length(illegalChars)
+        if ~isempty(strfind(col_label, illegalChars{i}))
+           col_label = strrep(col_label,illegalChars{i}, '_'); 
+           disp(['Illegal Column label. ' disp_label ' changed to ' col_label ]);
```
chrisdembia: If there are multiple different illegal characters, the user will get this warning multiple times. For example, a column label `hi*/` would generate:

```
Illegal Column Label. hi*/ changed to hi_/
Illegal Column Label. hi_/ changed to hi__
```

Ideally the user would only see the warning once:

```
Illegal Column Label. hi*/ changed to hi__
```

You could use a boolean in the for-loop to track if any of the illegal characters are present, and check that boolean outside of the loop to determine if you should print a message.

Bindings/Java/Matlab/Utilities/osimC3D.m
```diff
@@ -123,5 +237,43 @@ function rotateTable(obj, table, axisString, value)
...
+               name = obj.getName();
+            else
+            
+                if ~ischar(path{1})
+                   % Path must be a string  
+                   error('Input must be a sting of characters')
+                end
+
+                if isempty(strfind(path{1}, ext))
+                   % Path must have an input path
```
chrisdembia: You could remove this comment; it's redundant with the message in the error. Also, the message in the error is clearer than this comment :)

Bindings/Java/Matlab/Utilities/osimC3D.m
```diff
@@ -81,11 +120,86 @@ function rotateData(obj,axis,value)
...
+
+            import org.opensim.modeling.*
+            % Write to file
+            TRCFileAdapter().write( obj.getTable_markers, outputPath)
+            disp(['Marker file written to ' outputPath]);
+        end
+        function writeMOT(obj,varargin)
+         % Write forces data to mot file
+         % osimC3d.writeMOT()                       Write to current dir using c3d name
+         % osimC3d.writeMOT('Walking.mot')          Write to current dir with input name
```
chrisdembia: I'm not sure the invocations above write to the current dir. I think these invocations write the TRC file next to the C3D file (`obj.getPath()`), wherever it is.

Bindings/Java/Matlab/Utilities/osimC3D.m
```diff
@@ -81,11 +120,86 @@ function rotateData(obj,axis,value)
...
+            if ~isempty(strfind(label,'f'))
+                label = strrep(label,'f', 'ground_force_');
+                label = [label '_v'];
+            elseif ~isempty(strfind(label,'p'))
+                label = strrep(label,'p', 'ground_force_');
+                label = [label '_p'];
+            elseif ~isempty(strfind(label,'m'))
+                label = strrep(label,'m', 'ground_moment_');
+                label = [label '_m'];
+            end
```
chrisdembia: I'm having trouble following this part. What do the labels look like beforehand? Do they look something like `f1`, `f2`, `f3`?

Bindings/Java/Matlab/Utilities/osimC3D.m
```diff
@@ -81,11 +120,86 @@ function rotateData(obj,axis,value)
...
+            import org.opensim.modeling.*
+            % Write to file
+            TRCFileAdapter().write( obj.getTable_markers, outputPath)
+            disp(['Marker file written to ' outputPath]);
+        end
+        function writeMOT(obj,varargin)
+         % Write forces data to mot file
+         % osimC3d.writeMOT()                       Write to current dir using c3d name
+         % osimC3d.writeMOT('Walking.mot')          Write to current dir with input name
+         % osimC3d.writeMOT('C:\data\Walking.mot')  Write to input location from full input path
```
chrisdembia: Can you use `/` instead of `\` to help promote cross-platform code among our users? `/` works on both Mac and Windows.

Bindings/Java/Matlab/Utilities/osimC3D.m
chrisdembia: I think you put this method in `generateOutputPath()` to avoid duplicating the code in `writeTRC()` and `writeMOT()`. Perhaps you can just update the comments, then. Or make `generateOutputPath()` a private method.

Bindings/Java/Matlab/Utilities/osimC3D.m
```diff
@@ -81,11 +120,86 @@ function rotateData(obj,axis,value)
...
+            % osimC3d.writeTRC('Walking.trc')          Write to current dir with input name
+            % osimC3d.writeTRC('C:\data\Walking.trc')  Write to input location from full input path
+
+            % Compute an output path to use for writing to file
+            outputPath = generateOutputPath(obj,varargin,'.trc');
+
+
+            import org.opensim.modeling.*
+            % Write to file
+            TRCFileAdapter().write( obj.getTable_markers, outputPath)
```
chrisdembia: You don't need to make a clone of the markers table here. You could use `obj.markers` instead of `obj.getTable_markers`. Also, it is good practice to use parentheses when invoking methods, even if there are no arguments (`obj.getTable_markers()`).

Bindings/Java/Matlab/Utilities/osimC3D.m
```diff
@@ -81,11 +120,86 @@ function rotateData(obj,axis,value)
...
+         % osimC3d.writeMOT()                       Write to current dir using c3d name
+         % osimC3d.writeMOT('Walking.mot')          Write to current dir with input name
+         % osimC3d.writeMOT('C:\data\Walking.mot')  Write to input location from full input path
+
+         % Compute an output path to use for writing to file
+         outputPath = generateOutputPath(obj,varargin,'.mot');
+         
+         import org.opensim.modeling.*
+         % Get the forces table
+         forces = obj.getTable_forces();
```
chrisdembia: As opposed to my comment in `writeTRC()`, it's correct that this is a copy (clone), because you modify the table later on.

chrisdembia:
Looking good, @jimmyDunne . I left a bunch of minor comments.

Bindings/Java/Matlab/Utilities/osimC3D.m
jimmyDunne: Thanks for point that out, it makes sense. I think in an earlier version I was doing something to the marker table but that got pushed to the C++ side but I kept using a copy.

Bindings/Java/Matlab/Utilities/osimC3D.m
jimmyDunne: I added a comment, let me know if it makes sense or not :/

Bindings/Java/Matlab/Utilities/osimTableToStruct.m
```diff
@@ -77,14 +77,19 @@
...
+    illegalChars = [{' '} {'*'} {'.'} {'/'} {'-'} {'|'}];
+    for i = 1 : length(illegalChars)
+        if ~isempty(strfind(col_label, illegalChars{i}))
+           col_label = strrep(col_label,illegalChars{i}, '_'); 
+        end
+    end
+    % If the label has changed, inform the User.
+    if ~strcmp(disp_label,col_label)
+        disp(['Illegal Column label. ' disp_label ' changed to ' col_label ]); 
+    end
```
jimmyDunne: Instead of a boolean, I just check to see if the label has changed after the for loop— does that sound okay?

Bindings/Java/Matlab/Utilities/osimC3D.m
chrisdembia: Yes that makes sense.

Bindings/Java/Matlab/Utilities/osimTableToStruct.m
chrisdembia: Ah yeah that works.

Bindings/Java/Matlab/tests/testOsimC3D.m
```diff
@@ -8,21 +8,137 @@
...
+%% Test Rotations
+for i = 1 : 4     
+    % Rotate Data 180 degrees around X and check that Y and Z data are
+    % mirrored
+    c3d.rotateData('x',-90);c3d.rotateData('x',-90);
+    [markerRef180X, forceRef180X] = c3d.getAsStructs;
+    % For each marker, Y and Z data should be mirrored when rotating 180
+    % about x. 
+    for u = 1 : nMlabels
+        % Get the sum of the differences between original and rotated data
```
chrisdembia: Remove "the differences between"? You're doing a sum.

Bindings/Java/Matlab/tests/testOsimC3D.m
```diff
@@ -8,21 +8,137 @@
...
+        p = nansum(markerRef.(mlabels{u}) + markerRef180X.(mlabels{u}));
+        % Difference should be at or near zero. Round to 8 decimal places
```
chrisdembia: Difference -> Sum

chrisdembia:
I've left some more comments.

@jimmyDunne is this still WIP?

chrisdembia:
This PR fixes https://github.com/opensim-org/opensim-gui/issues/919

jimmyDunne:
Thanks, @chrisdembia!

@nickbianco, should be good to test now!

Bindings/Java/Matlab/Utilities/osimC3D.m
```diff
@@ -301,7 +304,24 @@ function rotateTable(obj, table, axisString, value)
...
             outputPath = fullfile(filepath, [name ext]);
+        end
+        function table_flat = mm2m(obj,table_flat)
             
+            nForces = table_flat.getNumColumns();
+            nRows  = table_flat.getNumRows();
+            labels = table_flat.getColumnLabels();
+            
+            for i = 0 : nForces - 1
+                if ~contains(char(labels.get(i)),'v')
```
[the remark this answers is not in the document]
chrisdembia: Which types of columns are you converting? Are you converting moments from `N-mm` to `N-m`?

nickbianco:
The TRC and MOT writing changes are working for me. For two different C3D files, I was able to go from the C3D file itself to a sync'd visualization in the GUI in only a couple of minutes. 

The only issue I ran into was that for larger files (60+ seconds worth of motion data), the script got hung up [here](https://github.com/opensim-org/opensim-core/blob/81e1c285b5ef82ddecb3b34e22df634f6c9802f8/Bindings/Java/Matlab/Utilities/osimC3D.m#L210) where the TimeSeriesTableVec3 of the forces gets flattened. @jimmyDunne have you tested larger files, and if so, how long do they take to flatten the forces? In the meantime, I'll time the larger file I tried to see how long it actually takes.

Rachel is testing the changes now too, I'll report back when she is done.

nickbianco:
For the file I mentioned, it took ~18.9 minutes to write an MOT file for 60 seconds worth of data.

aseth1:
> For the file I mentioned, it took ~18.9 minutes to write an MOT file for 60 seconds worth of data.

It is appending rows one by one, which resizes (and moves data). I had a fix for data reading that should be the same solution of preallocating the size and just updating the contents. Do you want to take a look @nickbianco?

chrisdembia:
I opened #2242 for the performance issue. I don't think we need to hold up the Matlab `osimC3D` file for the performance issue.

Bindings/Java/Matlab/Utilities/osimC3D.m
```diff
@@ -223,8 +223,11 @@ function writeMOT(obj,varargin)
           forces_flat.addTableMetaDataString('nColumns',num2str(forces_flat.getNumColumns()+1))
           forces_flat.addTableMetaDataString('nRows',num2str(forces_flat.getNumRows()));
 
+          % Convert mm to m
+          forces_flat_m  = obj.mm2m(forces_flat);
```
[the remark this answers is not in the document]
jimmyDunne: We don't. It is most typical that it is stored in mm, but there does not seem to be metadata in the table to suggest if it or isn't. I think for now I will record this in documentation and, if users encounter issues, add an optional input to convert or not.

Bindings/Java/Matlab/Utilities/osimC3D.m
```diff
@@ -301,7 +304,24 @@ function rotateTable(obj, table, axisString, value)
...
+            nRows  = table_flat.getNumRows();
+            labels = table_flat.getColumnLabels();
+            
+            for i = 0 : nForces - 1
+                if ~contains(char(labels.get(i)),'v')
+                    for u = 0 : nRows - 1
+                        % Get the table value
+                        c = table_flat.getDependentColumnAtIndex(i).get(u);
+                        % set the table value
+                        table_flat.getDependentColumnAtIndex(i).set(u,c/1000);
```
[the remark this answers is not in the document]
jimmyDunne: scalarDivideInPlace() isn't a method for Column vectors so I wasn't able to use it. Would be great if it was!

jimmyDunne:
Thanks for feedback, Chris. It has been great!

> took ~18.9 minutes to write an MOT file for 60 seconds 

Ouch, that is rough. Almost all of the data I have/use is only a few seconds long so I don't see the performance issues. A few seconds (3-4sec) of data is converted almost instantaneously for me., so the scale-up doesn't seem correct. Weird. 

Anyway, I have tested the outputs in IK and ID against a range of my own C3Ds going back over the last 10 years and it seems to working great. The one worry I have is that I don't have any tests using Kistler type forceplates. I will make a note of the limitations in the documentation and if it becomes a problem I can source some files for testing.

chrisdembia:
> The one worry I have is that I don't have any tests using Kistler type forceplates. I will make a note of the limitations in the documentation and if it becomes a problem I can source some files for testing.

I agree; this is just something to document. You could also create a GitHub issue for seeking Kistler forceplate data.

tkuchida:
> You could also create a GitHub issue for seeking Kistler forceplate data.

Could also ask on the Forum.

jimmyDunne:
From Tom's suggestion, I posted on the forum. I can start a new issue for the Kistler plates and if we get community members donating C3Ds with Kistler plates, I can add them to testing in the next sprint.

PR #2239, opened 2018-08-03
Model::formStateStorage() return value: are all coordinate values specified?

chrisdembia:
This is a first step towards addressing https://github.com/opensim-org/opensim-gui/issues/86

### Brief summary of changes

`Model::formStateStorage()` has a return value that denotes if all coordinate values are provided. This will allow the GUI to determine if needs to call `Model::assemble()` for visualization.

### Testing I've completed

I added a test case that ensures that "removing" a coordinate value column causes the function to return false.

Are there more robust or faster ways to check if the unspecified states are coordinates?

### CHANGELOG.md (choose one)

- no need to update because...this is a minor change

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2239)

### DIFF
--- a/OpenSim/Simulation/Model/Model.cpp
+++ b/OpenSim/Simulation/Model/Model.cpp
@@ -1952,11 +1952,6 @@ bool Model::formStateStorage(const Storage& originalStorage,
     rStateNames.insert(0, "time");
     statesStorage.setColumnLabels(rStateNames);
     
-    /*const auto coords = getCoordinatesInMultibodyTreeOrder();
-    for (const auto& coord : coords) {
-        if
-    }*/
-    
     // https://stackoverflow.com/questions/874134/find-if-string-ends-with-another-string-in-c
     auto string_ends_with =
             [](const std::string& str, const std::string& ending) {
@@ -1964,8 +1959,9 @@ bool Model::formStateStorage(const Storage& originalStorage,
                 return std::equal(ending.rbegin(), ending.rend(), str.rbegin());
             };
     for (int i = 0; i < mapColumns.size(); ++i) {
-        if (mapColumns[i] == -1 && string_ends_with(rStateNames[i], "/value")) {
-            std::cout << "DEBUG " << i << " " << mapColumns[i] << " " << rStateNames[i] << std::endl;
+        // Must add 1 because, at this point, rStateNames includes "time".
+        if (mapColumns[i] == -1
+                && string_ends_with(rStateNames[i+1], "/value")) {
             return false;
         }
     }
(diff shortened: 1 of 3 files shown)


aymanhab:
@chrisdembia in the past (3.3 if we want to restore exact behavior) the check was for all states not just coordinates, so we can use the fast Vector form to set state variables. When the GUI runs FD it uses statesReporter/fullStates so that muscles are colored correctly. Do we want check for Coordinates only or FullState?

chrisdembia:
Ah, @aseth1 and I discussed this yesterday. IK actually does report the values of dependent coordinates. And there is no need to call `assemble()` if all coordinates are provided, meaning that for an IK solution, we could avoid `assemble()`. 

If the function returned `isFullState` instead, then that means you'd `assemble()` even for an IK solution, which would be unnecessary (since all coordinates are provided).

We also discussed using an enum or a bitmask. We could possibly return both `hasAllCoordinates` and `hasFullState` (in an `std::pair` or similar, or an enum).

aymanhab:
Thanks @chrisdembia I wouldn't make this more involved with return types because I think eventually we have to store the description with the data itself.

aymanhab (commented):
LGTM, one question regarding dependent coordinates, are these reported by IK or FD?

chrisdembia:
> one question regarding dependent coordinates, are these reported by IK or FD?

I just checked: both IK and FD report dependent coordinates.

aymanhab (approved):

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -1951,6 +1951,31 @@ void Model::formStateStorage(const Storage& originalStorage,
     }
     rStateNames.insert(0, "time");
     statesStorage.setColumnLabels(rStateNames);
+    
+    
+    // Determine the return value.
+    // ---------------------------
+    // Get coordinate state variable names.
+    const auto coords = getCoordinatesInMultibodyTreeOrder();
+    // Coordinate value names must be unique.
```
tkuchida: Suggest "names" -> "full paths" to avoid confusion

tkuchida (approved):
:+1: (one minor suggestion). Also, the name of the method doesn't suggest that the return value would depend specifically on dependent coordinates (I would expect `formStateStorage` to return `false` if, e.g., the entire operation were unsuccessful). However, I agree with Ayman's comment about addressing in a more rigorous way post-release.

chrisdembia:
@aseth1 this might be useful.

nickbianco:
Closing: stale PR.

Issue #2041, opened 2017-12-16
More Integrator options in Python/MATLAB


carmichaelong:
The biggest feedback from the NIPS challenge was to have more control over the Integrator, and, especially the accuracy or tolerance. Some possible options (could do multiple):
1) Add new methods to `Manager` to add some simple options. For instance, a new `setIntegrator()` method that takes a `string` to specify the type of integrator, and `setAccuracy()` to specify the accuracy of the internal integrator.
2) Wrap the base `Integrator` class.
3) Wrap the individual `Integrator`s.

chrisdembia:
+1

1 is probably the easiest interface for most users but results in code duplication.

aseth1:
Another option is to add an `OpenSim::Integrator` component with sub-classes that wrap the SimTK::Integrators. Then we could support different integrator types with different parameters (properties) without forcing the Manager to contain these properties and/or enumerate Integrator types. This would fit with our move towards composing studies and analyses.

chrisdembia:
Discussion with @aseth1 @carmichaelong @nickbianco:

- Remove the ability to pass an integrator into the Manager.
- Use an enum to specify an integrator type (use strings in Python/Matlab).
- Document that the string version should be used in scripting.
- Add `Manager::setIntegratorAccuracy()`.
- Do not wrap any SimTK classes.

chrisdembia:
I noticed that I get different behavior between the following two pieces of code:

A:
```cpp
        Manager manager(model);
        manager.setIntegratorAccuracy(1e-5);
        manager.initialize(state);
        state = manager.integrate(1.0);
```

B:
```cpp
        Manager manager(model, state);
        manager.setIntegratorAccuracy(1e-5);
        state = manager.integrate(1.0);
```

I think what's happening is that in (B), the integrator accuracy I specify is not being used.

It seems important to document that `initialize()` must be called after setting integrator settings.

carmichaelong:
Hm, I think that this deserves a closer look than just documentation. That behavior does not seem great. Have you tested if B is equivalent to:
```
        Manager manager(model);
        manager.initialize(state);
        manager.setIntegratorAccuracy(1e-5);
        state = manager.integrate(1.0);
```

chrisdembia:
> Have you tested if B is equivalent to:

I just tried this and yes, B is equivalent.

I don't see anything in Simbody's doxygen that says that changes to integrator settings must occur before calling `initialize()`, so perhaps you're right about a closer look.

[...]

PR #2241, opened 2018-08-04
[WIP] Add more integrator options into Manager

carmichaelong:
Fixes issue #2041 

### Brief summary of changes
So far:
- Removes option of passing in an Integrator to the Manager
- Adds minimal functions needed to cover OpenSim's use of the Manager

### Testing I've completed
- All tests still pass

- What extra test cases should be used?
- Thoughts on implementations?

### CHANGELOG.md (choose one)
- TODO: Add Manager usage changes to changelog

TODO:
- Documentation/Examples
- Add fixed step option?

@chrisdembia @aseth1 Although I don't think the PR is ready for merging, any feedback at this stage before going any further I think would be useful.

### DIFF
--- a/OpenSim/Simulation/Manager/Manager.cpp
+++ b/OpenSim/Simulation/Manager/Manager.cpp
@@ -503,6 +503,12 @@ setModel(Model& model)
   */
 void Manager::setIntegrator(Integrator integMethod)
 {
+    if (_timeStepper) {
+        std::string msg = "Cannot set a new integrator on this Manager";
+        msg += "after Manager::integrate() has been called at least once.";
+        OPENSIM_THROW(Exception, msg);
+    }
+
     auto& sys = _model->getMultibodySystem();
     switch (integMethod) {
         case Integrator::CPodes:
@@ -540,8 +546,38 @@ void Manager::setIntegrator(Integrator integMethod)
         case Integrator::Verlet:
             _integ = new SimTK::VerletIntegrator(sys);
             break;
+
+        default:
+            std::string msg = "Integrator method not recognized.";
+            OPENSIM_THROW(Exception, msg);
     }
-    
+}
+
+void Manager::setIntegrator(const std::string& integMethod)
+{
+    if (integMethod == "CPodes")
+        setIntegrator(Integrator::CPodes);
+    else if (integMethod == "ExplicitEuler")
+        setIntegrator(Integrator::ExplicitEuler);
+    else if (integMethod == "RungeKutta2")
+        setIntegrator(Integrator::RungeKutta2);
+    else if (integMethod == "RungeKutta3")
+        setIntegrator(Integrator::RungeKutta3);
+    else if (integMethod == "RungeKuttaFeldberg")
+        setIntegrator(Integrator::RungeKuttaFeldberg);
+    else if (integMethod == "RungeKuttaMerson")
+        setIntegrator(Integrator::RungeKuttaMerson);
+    else if (integMethod == "SemiExplicitEuler2")
+        setIntegrator(Integrator::SemiExplcitEuler2);
+    else if (integMethod == "Verlet")
+        setIntegrator(Integrator::Verlet);
+    else
+    {
+        std::string msg = "Integrator method '";
+        msg += integMethod;
+        msg += "' not recognized";
+        OPENSIM_THROW(Exception, msg);
+    }
 }
 
 /**
--- a/OpenSim/Simulation/Manager/Manager.cpp
+++ b/OpenSim/Simulation/Manager/Manager.cpp
@@ -50,26 +50,11 @@ std::string Manager::_displayName = "Simulator";
 //=============================================================================
 Manager::Manager(Model& model) : Manager(model, true)
 {
-    _defaultInteg.reset(
-            new SimTK::RungeKuttaMersonIntegrator(_model->getMultibodySystem()));
-    _integ = *_defaultInteg;
-}
-
-Manager::Manager(Model& model, SimTK::Integrator& integ)
-        : Manager(model, true) 
-{
-    setIntegrator(integ);
+    _integ = new SimTK::RungeKuttaMersonIntegrator(_model->getMultibodySystem());
 }
 
 Manager::Manager(Model& model, const SimTK::State& state)
         : Manager(model) 
-{
-    initialize(state);
-}
-
-Manager::Manager(Model& model, const SimTK::State& state, 
-    SimTK::Integrator& integ)
-    : Manager(model, integ) 
 {
     initialize(state);
 }
@@ -514,6 +499,52 @@ setModel(Model& model)
 //-----------------------------------------------------------------------------
 //_____________________________________________________________________________
 /**
+  * Set the integrator.
+  */
+void Manager::setIntegrator(Integrator integMethod)
+{
+    auto& sys = _model->getMultibodySystem();
+    switch (integMethod) {
+        case Integrator::CPodes:
+            _integ = new SimTK::CPodesIntegrator(sys);
+            break;
+
+        case Integrator::ExplicitEuler:
+            _integ = new SimTK::ExplicitEulerIntegrator(sys);
+            break;
+
+        case Integrator::RungeKutta2:
+            _integ = new SimTK::RungeKutta2Integrator(sys);
+            break;
+
+        case Integrator::RungeKutta3:
+            _integ = new SimTK::RungeKutta3Integrator(sys);
+            break;
+
+        case Integrator::RungeKuttaFeldberg:
+            _integ = new SimTK::RungeKuttaFeldbergIntegrator(sys);
+            break;
+
+        case Integrator::RungeKuttaMerson:
+            _integ = new SimTK::RungeKuttaMersonIntegrator(sys);
+            break;
+
+        //case Integrator::SemiExplicitEuler:
+        //    _integ = new SimTK::SemiExplicitEulerIntegrator(sys, stepSize);
+        //    break;
+
+        case Integrator::SemiExplcitEuler2:
+            _integ = new SimTK::SemiExplicitEuler2Integrator(sys);
+            break;
+
+        case Integrator::Verlet:
+            _integ = new SimTK::VerletIntegrator(sys);
+            break;
+    }
+    
+}
+
+/**
  * Get the integrator.
  */
 SimTK::Integrator& Manager::
@@ -521,24 +552,41 @@ getIntegrator() const
 {
     return *_integ;
 }
+
 /**
- * Set the integrator.
- */
-void Manager::
-setIntegrator(SimTK::Integrator& integrator) 
-{   
-    if (_integ.get() == &integrator) return;
-    if (_timeStepper) {
-        std::string msg = "Cannot set a new integrator on this Manager";
-        msg += "after Manager::integrate() has been called at least once.";
+  * Set the Integrator's accuracy. 
+  */
+void Manager::setAccuracy(double accuracy)
+{
+    if (!_integ->methodHasErrorControl()) {
+        std::string msg = "Integrator method ";
+        msg += _integ->getMethodName();
+        msg += " does not support error control.";
         OPENSIM_THROW(Exception, msg);
     }
 
-    _integ = &integrator;
-    // If we had been using the _defaultInteg, we no longer need it.
-    _defaultInteg.reset();
+    _integ->setAccuracy(accuracy);
 }
 
+void Manager::setMinimumStepSize(double hmin)
+{
+    _integ->setMinimumStepSize(hmin);
+}
+
+void Manager::setMaximumStepSize(double hmax)
+{
+    _integ->setMaximumStepSize(hmax);
+}
+
+//void Manager::setFixedStepSize(double stepSize)
+//{
+//    _integ->setFixedStepSize(stepSize);
+//}
+
+void Manager::setInternalStepLimit(int nSteps)
+{
+    _integ->setInternalStepLimit(nSteps);
+}
 
 //=============================================================================
 // EXECUTION
--- a/OpenSim/Simulation/Manager/Manager.h
+++ b/OpenSim/Simulation/Manager/Manager.h
@@ -179,6 +179,7 @@ public:
         Verlet
     };
     void setIntegrator(Integrator integMethod);
+    void setIntegrator(const std::string& integMethod);
     SimTK::Integrator& getIntegrator() const;
     void setAccuracy(double accuracy);
     void setMinimumStepSize(double hmin);
--- a/OpenSim/Simulation/Manager/Manager.h
+++ b/OpenSim/Simulation/Manager/Manager.h
@@ -86,7 +86,7 @@ private:
 
     // The integrator that is used when using the model-only constructor.
     // This is allocated only if necessary.
-    std::unique_ptr<SimTK::Integrator> _defaultInteg;
+    //std::unique_ptr<SimTK::Integrator> _defaultInteg;
 
     /** TimeStepper */
     std::unique_ptr<SimTK::TimeStepper> _timeStepper;
@@ -130,9 +130,6 @@ private:
 // METHODS
 //=============================================================================
 public:
-    /** This constructor cannot be used in MATLAB/Python, since the
-     * SimTK::Integrator%s are not exposed in those languages. */
-    Manager(Model& model, SimTK::Integrator& integ);
     /** Constructor that takes a model only and internally uses a
      * SimTK::RungeKuttaMersonIntegrator with default settings (accuracy,
      * constraint tolerance, etc.). MATLAB/Python users must use this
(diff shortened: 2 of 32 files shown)


OpenSim/Simulation/Manager/Manager.h
```diff
@@ -173,11 +167,25 @@ class OSIMSIMULATION_API Manager
     { _writeToStorage =  writeToStorage; }
 
     // Integrator
+    enum class Integrator {
```
chrisdembia: Consider naming this `IntegratorType`.

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -130,19 +130,13 @@ class OSIMSIMULATION_API Manager
 // METHODS
 //=============================================================================
 public:
-    /** This constructor cannot be used in MATLAB/Python, since the
-     * SimTK::Integrator%s are not exposed in those languages. */
-    Manager(Model& model, SimTK::Integrator& integ);
     /** Constructor that takes a model only and internally uses a
      * SimTK::RungeKuttaMersonIntegrator with default settings (accuracy,
      * constraint tolerance, etc.). MATLAB/Python users must use this
      * constructor. */
```
chrisdembia: "MATLAB/Python users must use this constructor" is no longer true; the other constructor can also be used from scripting.

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -173,11 +167,25 @@ class OSIMSIMULATION_API Manager
...
+        CPodes,
+        ExplicitEuler,
+        RungeKutta2,
+        RungeKutta3,
+        RungeKuttaFeldberg,
+        RungeKuttaMerson,
+        //SemiExplicitEuler, no error ctrl, requires fixed stepSize arg on construction
+        SemiExplcitEuler2,
+        Verlet
+    };
```
chrisdembia: It would be nice to add doxygen documentation that links to each of these integrators in Simbody. I think the doxygen output will look nice if you do the following:

```
    ...
    /** For details, see SimTK::CPodesIntegrator. */
    CPodes,
    ...
```

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -173,11 +167,25 @@ class OSIMSIMULATION_API Manager
...
+    void setIntegrator(const std::string& integMethod);
     SimTK::Integrator& getIntegrator() const;
-    /** %Set the integrator. The Manager does *not* take ownership of the
-     * passed-in integrator.
-     */
-    void setIntegrator(SimTK::Integrator&);
+    void setAccuracy(double accuracy);
+    void setMinimumStepSize(double hmin);
+    void setMaximumStepSize(double hmax);
+    void setInternalStepLimit(int nSteps);
```
chrisdembia: For all of these, perhaps link to the Simbody documentation (e.g., `/** See SimTK::Integrator::setAccuracy(). */`).

If you built and installed Simbody's doxygen before building OpenSim's, the links should go to your local copy of Simbody's doxygen.

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -173,11 +167,25 @@ class OSIMSIMULATION_API Manager
...
+        SemiExplcitEuler2,
+        Verlet
+    };
+    void setIntegrator(Integrator integMethod);
+    void setIntegrator(const std::string& integMethod);
     SimTK::Integrator& getIntegrator() const;
-    /** %Set the integrator. The Manager does *not* take ownership of the
-     * passed-in integrator.
-     */
-    void setIntegrator(SimTK::Integrator&);
```
chrisdembia: I think it would improve clarity to prefix these methods with "Integrator" (e.g., `setIntegratorAccuracy()`).

OpenSim/Simulation/Manager/Manager.cpp
```diff
@@ -513,6 +498,88 @@ setModel(Model& model)
...
+        setIntegrator(Integrator::RungeKuttaFeldberg);
+    else if (integMethod == "RungeKuttaMerson")
+        setIntegrator(Integrator::RungeKuttaMerson);
+    else if (integMethod == "SemiExplicitEuler2")
+        setIntegrator(Integrator::SemiExplcitEuler2);
+    else if (integMethod == "Verlet")
+        setIntegrator(Integrator::Verlet);
+    else
+    {
+        std::string msg = "Integrator method '";
```
chrisdembia: Consider changing "method" to "type".

carmichaelong:
@chrisdembia Thanks for the quick feedback.  I've made changes based on your comments. In the end, I went with "method" instead of "type" because it seemed like it aligned better with Simbody's terminology and function names, such as `getMethodName()`.

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -167,25 +168,64 @@ class OSIMSIMULATION_API Manager
...
+
+        /** For details, see SimTK::VerletIntegrator. */
         Verlet
     };
-    void setIntegrator(Integrator integMethod);
+
+    /** Sets the type of integrator used via IntegratorMethod enum. The 
+      * integrator will be set to its default options. For MATLAB/Python
+      * users, see `setIntegrator(const std::string&)` instead.*/
+    void setIntegrator(IntegratorMethod integMethod);
```
chrisdembia: What do you think about naming this method `setIntegratorMethod()`? Right now, there is a `getIntegrator()` that returns the actual integrator.

carmichaelong: I think that'd be a fine change. If you're leaning that way I'd be happy to do so.

carmichaelong:
@chrisdembia Pushed some changes as requested and updated the changelog. With those, I've removed the [WIP] tag.

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -173,11 +168,64 @@ class OSIMSIMULATION_API Manager
...
+    };
+
+    /** Sets the integrator method used via IntegratorMethod enum. The 
+      * integrator will be set to its default options. For MATLAB/Python
+      * users, see `setIntegrator(const std::string&)` instead.*/
+    void setIntegratorMethod(IntegratorMethod integMethod);
+    /** Sets the integrator method used via a string. The integrator 
+      * will be set to its default options. For C++ users, see 
+      * `setIntegrator(IntegratorMethod)` instead. */
+    void setIntegratorMethod(const std::string& integMethod);
```
chrisdembia: What do you think about adding a Matlab or Python example to make it crystal clear that the string should contain the exact same text as the enum?

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -299,7 +347,7 @@ class OSIMSIMULATION_API Manager
 
     // Helper functions during initialization of integration
     void initializeStorageAndAnalyses(const SimTK::State& s);
-    void initializeTimeStepper(const SimTK::State& s);
+    //void initializeTimeStepper(const SimTK::State& s);
```
chrisdembia: Remove?

chrisdembia:
@carmichaelong made a good point: `_integ` should be a `unique_ptr`

carmichaelong:
@chrisdembia The last push should address the rest of your suggestions.

OpenSim/Simulation/Manager/Manager.cpp
```diff
@@ -50,7 +50,7 @@ std::string Manager::_displayName = "Simulator";
 //=============================================================================
 Manager::Manager(Model& model) : Manager(model, true)
 {
-    _integ = new SimTK::RungeKuttaMersonIntegrator(_model->getMultibodySystem());
+    _integ = make_unique<SimTK::RungeKuttaMersonIntegrator>(_model->getMultibodySystem());
```
chrisdembia: The function template `std::make_unique` is C++14, and we are still only on C++11. Could you use `_integ.reset(new ...)` instead?

OpenSim/Simulation/Manager/Manager.cpp
```diff
@@ -568,7 +568,7 @@ void Manager::setIntegratorMethod(const std::string& integMethod)
     else if (integMethod == "RungeKuttaMerson")
         setIntegratorMethod(IntegratorMethod::RungeKuttaMerson);
     else if (integMethod == "SemiExplicitEuler2")
-        setIntegratorMethod(IntegratorMethod::SemiExplcitEuler2);
+        setIntegratorMethod(IntegratorMethod::SemiExplicitEuler2);
```
chrisdembia: Nice catch!!

OpenSim/Simulation/Manager/Manager.cpp
carmichaelong: Absolutely. Travis also strongly agrees with you already...

carmichaelong:
@chrisdembia I had to make a somewhat substantial change, removing `CPodes` from the possible integrators. With the `unique_ptr` change, `testManager` was failing on CI (and stochastically failing locally) on the `setIntegratorMethod()` call specifically right AFTER `CPodes` was the current integrator. Debugging locally, it looked to be an issue with `CPodes` destructor when assigning a new integration method. I have not dug deeper than this yet, but I think that shouldn't get in the way of this PR.

chrisdembia:
LGTM!

OpenSim/Simulation/Manager/Manager.cpp
```diff
@@ -513,6 +498,88 @@ setModel(Model& model)
...
+            _integ.reset(new SimTK::VerletIntegrator(sys));
+            break;
+
+        default:
+            std::string msg = "Integrator method not recognized.";
+            OPENSIM_THROW(Exception, msg);
+    }
+}
+
+void Manager::setIntegratorMethod(const std::string& integMethod)
```
aseth1: I know you really want to make it easy for the scripting users to specify the integrator, but having two ways to do it- using and enum and a string is not healthy code. My strong preference is to only have the enumerator interface. This does require enabling MATLAB users to pass an integer value. For example, see how we handle[ `ForceLocation`](https://simtk.org/api_docs/opensim/api_docs/classOpenSim_1_1C3DFileAdapter.html#ac2eb67cb6bac08281775988cd46fa604)  for the `C3DFileAdapter`. Users have to look up the documentation to know what strings to type, so its not like the string interface is any easier for scripting users.

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -173,11 +161,71 @@ class OSIMSIMULATION_API Manager
...
+      * The integrator will be set to its default options. For C++ users, see 
+      * `setIntegrator(IntegratorMethod)` instead. 
+      *
+      * Example:
+      * @code
+      * manager = Manager(model);
+      * manager.setIntegratorMethod('SemiExplicitEuler2');
+      * @endcode 
+      */
+    void setIntegratorMethod(const std::string& integMethod);
```
aseth1: See comment above.

OpenSim/Simulation/SimbodyEngine/Test/testJoints.cpp
```diff
@@ -372,9 +372,9 @@ void integrateOpenSimModel(Model *osimModel, SimTK::State &osim_state)
 
     // SETUP OpenSim SIMULATION Manager
     osimModel->getMultibodySystem().realize(osim_state, Stage::Velocity);
-    RungeKuttaFeldbergIntegrator integrator(osimModel->getMultibodySystem() );
-    integrator.setAccuracy(integ_accuracy);
-    Manager manager(*osimModel, integrator);
+    Manager manager(*osimModel);
+    manager.setIntegratorMethod(Manager::IntegratorMethod::RungeKuttaFeldberg);
```
aseth1: :+1: Cool!

OpenSim/Simulation/Manager/Manager.cpp
carmichaelong: We could add just the `enum` case for now with examples and then monitor how many/few problems people have. Thanks for the link to the `ForceLocation` documentation. I'll use that as a guide.

chrisdembia: Bah, I suppose we should be more consistent but I really do prefer the string interface to the integer. Yes, in both cases, the writer of the code must look at the documentation, but it would be nice if the reader didn't need to do so.

chrisdembia:
@carmichaelong if a user makes the following calls:

```cpp
Manager man(model);
man.setIntegrator(...);
man.initialize(state);
man.setIntegrator(...);
```

is it necessary to call `man.initialize()` again? What happens if the user does *not* call `initialize()` again? Could you document this requirement/behavior?

Could you also convey in the documentation that calling `setIntegrator()` with the same integrator method as was already specified still causes the integrator to be reset to the default options?

carmichaelong:
@chrisdembia Good points. I'm making a push soon that covers what you said. Actually, `setIntegratorMethod()` will thrown an exception if it's called after a call to `initialize()` since `_timeStepper` will not be null. I've updated the Exception message to reflect this.

carmichaelong:
I made changes as per @aseth1's request. Perhaps looking at the doxygen gives us a better idea of what this looks like to users if string version isn't included.

I've also made some documentation changes and tests to address @chrisdembia's comments.

OpenSim/Simulation/Test/testManager.cpp
```diff
@@ -525,3 +500,32 @@ void testIntegratorInterface()
...
+    // ok to switch models
+    manager.setModel(arm2);
+
+    // can't integrate if not initialized
+    ASSERT_THROW(Exception, manager.integrate(1.0));
+    manager.initialize(s2);
+
+    // can't change model or integrator method now
+    ASSERT_THROW(Exception, manager.setModel(arm1));
+    ASSERT_THROW(Exception, manager.setIntegratorMethod(Manager::IntegratorMethod::ExplicitEuler));
```
chrisdembia: 👍

chrisdembia:
@carmichaelong I inspected the doxygen comment here: http://myosin.sourceforge.net/2241/classOpenSim_1_1Manager.html#a1da7ce502bdc284a7ed8a8df3f877c34

You can't actually use the enum in Matlab. This is the reason why `ForceLocation` shows using integers and has the table mapping the enum to integers:

image

carmichaelong:
Oh that makes way more sense. I definitely misread how that worked :-/

CHANGELOG.md
```diff
@@ -77,6 +77,15 @@ Converting from v3.x to v4.0
     - state.setTime(0.0);
     - Manager manager(model, state);
     - manager.integrate(1.0);
+- `Manager::setIntegrator(SimTK::Integrator)` has been removed and replaced by
+  `Manager::setIntegratorMethod(IntegratorMethod)` which uses an enum and
+  `Manager::setIntegratorMethod(std::string)` which uses a string. The latter
```
asethca: The string option has been removed, correct?

OpenSim/Simulation/Manager/Manager.h
```diff
@@ -173,11 +161,70 @@ class OSIMSIMULATION_API Manager
...
+      import opensim
+      manager = opensim.Manager(model)
+      manager.setIntegratorMethod(opensim.Manager.IntegratorMethod_SemiExplicitEuler2)
+      \endcode
+
+      <b>MATLAB example</b>
+      \code{.m}
+      import org.opensim.modeling.*
+      manager = Manager(model);
+      manager.setIntegratorMethod(5);
```
asethca: :+1:

CHANGELOG.md
carmichaelong: Yes, good catch. Should be addressed in latest commit.

clnsmith:
I'm guessing that I am one of the only people this effects negatively and in general it is definitely nice to have access to more integrators through the scripting interfaces. Don't want to be the annoying exception here, so I'm happy to try work arounds.

Is there a way to still use the CPODES integrator with the manager (in C++)? or an example of how to perform integration without the manager?

We always used the CPODES integrator for forward simulation of joint mechanics involving cartilage contact and ligaments because it ends up being a stiff problem. If I remember correctly, in general implicit integrators tend to work better for simulations with contact. In previous testing I found substantial speed up with CPODES compared to the RungeKuttaMerson. I know I tried the SemiExplicitEuler at one point, which was really fast, but quite sloppy if I remember right.

carmichaelong:
Whoops, didn't mean to break things for you. In fact, we definitely wanted CPODES in, but the problem was that it (likely) has a memory leak somewhere, so it was removed just because it broke our tests.

> so I'm happy to try work arounds.

Awesome, cause I'd say that's probably the easiest:
1) Add in another enum in [Manager.h](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Manager/Manager.h#L176)
2) Uncomment some lines in `setIntegratorMethod` in [Manager.cpp](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Manager/Manager.cpp#L514)
3) If you want to get the more advanced options, you can then use the `getIntegrator()` method, which will give you a pointer to the actual integrator (so really, it should be called updIntegrator).

> Is there a way to still use the CPODES integrator with the manager (in C++)? or an example of how to perform integration without the manager?

Absolutely. We thought this could be a good way to still use their own integrators. Examples are unfortunately littered throughout tests, so they're hard to find. Essentially, you'll use the `SimTK::TimeStepper` directly, rather than using it through a thin layer in the `Manager`. Look up "TimeStepper" in the repository. One such example is [here](https://github.com/opensim-org/opensim-core/blob/b526b3bd69169238147dc8311515b0545df7e119/OpenSim/Simulation/SimbodyEngine/Test/testConstraints.cpp#L197).

> I know I tried the SemiExplicitEuler at one point, which was really fast, but quite sloppy if I remember right.

I think this is also true. Could also try SemiExplicitEuler2. Probably worth a try regardless just for testing, but hopefully one of the options above should get you back to CPODES.

clnsmith:
Got it working with the TimeStepper, thanks for the help!

Issue #2240, opened 2018-08-03
Users should be warned when 4.0 assigns a different MotionType to a Coordinate


aseth1:
In 3.3  a `Coordinate`'s `MotionType` was user settable. The potential for labeling things Rotational was always present. When outputting .mot files we convert rotations (angles) into degrees. In the case of the model in https://github.com/opensim-org/opensim-gui/issues/617 that means all non-pelvis coordinates were specified to be `Rotational`. 

In 4.0 we eliminated the potential for `MotionType` inconsistencies by using the `Joint` type to determine the `MotionType`s of its coordinates.  For a `CustomJoint`,  a coordinate is considered `Rotational` (or conversely`Translational`) if and only if it is a pure rotation or translation along one of the corresponding six spatial transform axes, which means the relationship is 1 to 1 (e.g. linear function of slope +/-1). If the transform axis motion is any other function (besides that or Constant) it is interpreted as *Coupled* where  the motion about that axis is coupled to the generalized coordinate and it is not a physical measure of spatial motion (not an angle or displacement in Cartesian axes) of the joint's child frame.

The `SpatialTransform` for the *tib_pat_r* `Joint`  in the leg69 in https://github.com/opensim-org/opensim-gui/issues/617 with a *knee_angle_pat_r* coordinate has the rotation of the patella body as a spline function of *knee_angle_pat_r* and thus *knee_angle_pat_r* is not a `Rotational` coordinate.  Unfortunately, it was specified to be Rotational in 3.3 and in earlier model files. Now we have an inconsistency in the .mot files produced in 3.3 that has the coupled *knee_angle_pat_r* coordinate value converted to degrees in 3.3 but 4.0 does not interpret it as `Rotational` so it isn't converted back to internal units. This leads to the display issues in https://github.com/opensim-org/opensim-gui/issues/617.

aseth1:
@jenhicks asked:
> ... if labeling the MotionType as rotational was incorrect in 3.3, why is that you get the same answers in both 3.3 and 4.0 as long as you run IK and SO in the same software versions? (i.e., what are the specific cases where labeling a couple motion as rotation gives incorrect answers?)

There only cases, that I can think of, were we will get incorrect answers is when we do not correctly remove unit conversions,  because Simbody's coordinates and generalize forces are self-consistent - their product is always work.  In general, we can multiply all coordinate values by any arbitrary factor as long as we remember to undo that factor when applying it to the state.

[...]

Issue #2088, opened 2018-02-15
Static Optimization results for "Strength of Simulation" example differ from 3.3


tkuchida:
See "Strength of Simulation" example, Part I, Section C, first unordered list. Minimal example:
- open leg6dof9musc.osim
- lock pelvis_tilt, pelvis_tx, and pelvis_ty coordinates
- run Static (Motion = leg69_IK_swing.mot, Filter = 6 Hz)

**leg6dof9musc_StaticOptimization_activation.sto from 3.3:**
![](https://user-images.githubusercontent.com/4203505/36516040-1827b6aa-1731-11e8-900c-a4f8505c3526.png)

**leg6dof9musc_StaticOptimization_activation.sto from 4.0:**
![](https://user-images.githubusercontent.com/4203505/36516045-1df31764-1731-11e8-8f69-f7488392fd1f.png)

(AppVeyor artifact OpenSim-2fd23ef3-2018-02-20.zip)

tkuchida:
Possibly related to #2035

tkuchida:
Update: activations greater than 1 were caused by interpolation in Excel. Main reproducibility issue persists.

chrisdembia:
Verified on Mac  with OpenSim 2fd23ef-2018-02-20:

<img width="871" alt="screen shot 2018-02-21 at 8 16 40 pm" src="https://user-images.githubusercontent.com/846001/36520142-26b2b874-1744-11e8-8b2f-f627c59682b5.png">

tkuchida:
Not a model-specific issue. Tested with arm26 model on OpenSim-2416f17a-2018-02-25.zip ("OpenSim 4.0 User Experience and Feedback Session" version):

**3.3**
![](https://user-images.githubusercontent.com/4203505/36686210-96d9a3c6-1ada-11e8-80f3-5790b3faddba.png)

**4.0**
![](https://user-images.githubusercontent.com/4203505/36686222-9da32ba0-1ada-11e8-9365-ee703a653984.png)

tkuchida:
**Solution found, for arm26 model**
The `min_control` property for each muscle (Thelen2003Muscle) has a value of 0 in the .osim file but 4.0 changes these values to 0.01 when it loads the model, *without telling the user*. The user is unable to change these values back to 0:

![](https://user-images.githubusercontent.com/4203505/36687079-537732da-1add-11e8-9b01-94a77462aa14.png)

[...]

PR #2243, opened 2018-08-09
Warn when motion type is inconsistent in old models

aseth1:
Fixes issue #2240 and related to #2088 

### Brief summary of changes
Detect when an inconsistency in `Coordinate::MotionType` occurs while converting from 3.3 and earlier to 4.0 model.

Added stub of utility method to correct old motion (.mot) given the previous model and its related (list of) motion files.

### Testing I've completed
Able to warn (only in messages) for problematic models like *leg6dof9musc.osim* and *Rajagopal2015.osim*.

1. Is the warning message sufficient? @aymanhab can you show in a pop-up window? 
2. Where should the utility method live? I put it temporarily in *SimulationUtilities.h*
3. Volunteers to flesh out the file "fixer" for old kinematics files (in degrees).
4. Can we make it easier for the user  to run the utility through the GUI?

### CHANGELOG.md (choose one)
- need to add details for the PR(s) that made`MotionType` obsolete, describe how and why an inconsistency can arise, and point to utility to do something about it. 


OpenSim/Simulation/Model/Model.cpp
```diff
@@ -394,6 +394,9 @@ SimTK::State& Model::initializeState() {
     if (!hasSystem()) 
         throw Exception("Model::initializeState(): call buildSystem() first.");
 
+    std::string warn = getWarningMesssageForMotionTypeInconsistency();
+    cout << warn << endl;
```
chrisdembia: In the case that there is no warning, this will still print a newline.

OpenSim/Simulation/SimulationUtilities.h
```diff
@@ -103,9 +103,55 @@ inline SimTK::State simulate(Model& model,
 
     return state;
 }
-
 /// @}
 
+/** Moving from OpenSim 3.3 (and earlier) to OpenSim 4.0, MotionTypes for
+    Coordinates are now determined strictly by the coordinates' owning Joint.
+    In older models, the MotionType, particularly for CustomJoints, were user-
+    specified. That entailed in some cases, incorrectly labeling a Coordinate
```
chrisdembia: Can remove `,`.

OpenSim/Simulation/SimulationUtilities.h
```diff
@@ -103,9 +103,55 @@ inline SimTK::State simulate(Model& model,
...
+    as being Rotational, for example, when it is in fact Coupled. Some models
+    that exhibit this issue are the Rajagopal 2015, le6dof9musc models, where
```
chrisdembia: `Rajagopal 2015 and leg6dof9musc models` (`le6` -> `leg6`).

OpenSim/Simulation/SimulationUtilities.h
```diff
@@ -103,9 +103,55 @@ inline SimTK::State simulate(Model& model,
...
+    std::vector<Coordinate> problemCoords;
+    auto coordinates = model.getComponentList<Coordinate>();
+    for (auto& coord : coordinates) {
+        const Coordinate::MotionType oldMotionType =
+            coord.getOldUserSpecifiedMotionType();
+        const Coordinate::MotionType motionType = coord.getMotionType();
+
+        if ((oldMotionType != Coordinate::MotionType::Undefined) &&
+            (oldMotionType != motionType)) {
+            problemCoords.push_back(coord);
```
chrisdembia: This makes a copy. If I am going to implement this function, I can take care of that change.

OpenSim/Simulation/SimulationUtilities.h
```diff
@@ -103,9 +103,55 @@ inline SimTK::State simulate(Model& model,
...
+    the patella Coordinate was user-specified to be Rotational, but the angle
+    of the patella about a Z-axis of the patella body, is a spline function
+    (e.g. coupled function) of the knee_patella Coordinate. Thus, knee_patella
+    no longer represents a Cartesian rotation and is no longer classified as
+    Rotational. Use this utility to remove any unit conversions from Coordinates
+    that were incorrectly labeled as Rotational in the past. For these Coordinates
+    it will undo the incorrect radians to degrees conversion. */
+inline void updateKinematicsFilesForUpdatedModel(const Model& model, 
+                const std::vector<std::string>& filePaths, std::string suffix="")
+{
```
chrisdembia: Might be good to check if the provided model is older than `30514`.

If the XML document version is `>=30514` or the XML document version doesn't exist, then this function won't correct old columns, and users will be confused when their data still looks wrong. In this condition, we should either throw an exception or give a warning saying that this function won't really do anything.

OpenSim/Simulation/SimulationUtilities.h
```diff
@@ -103,9 +103,55 @@ inline SimTK::State simulate(Model& model,
...
+            problemCoords.push_back(coord);
+        }
+    }
+
+    if (problemCoords.size() == 0)
+        return; 
+
+    // Cycle through the data files 
+    for (auto filePath : filePaths) {
+        Storage data(filePath);
```
chrisdembia: What should we do if the file includes columns for speeds? Should those be fixed/converted also? I do not think the following code multiplies the speed coordinates (despite the comment regarding `_u`):

https://github.com/opensim-org/opensim-core/blob/c461d0290a9413e4552b32dc6a1f55d08fb15980/OpenSim/Simulation/SimbodyEngine/SimbodyEngine.cpp#L927-L943

OpenSim/Simulation/SimulationUtilities.h
```diff
@@ -103,9 +103,55 @@ inline SimTK::State simulate(Model& model,
...
+    specified. That entailed in some cases, incorrectly labeling a Coordinate
+    as being Rotational, for example, when it is in fact Coupled. Some models
+    that exhibit this issue are the Rajagopal 2015, le6dof9musc models, where
+    the patella Coordinate was user-specified to be Rotational, but the angle
+    of the patella about a Z-axis of the patella body, is a spline function
+    (e.g. coupled function) of the knee_patella Coordinate. Thus, knee_patella
+    no longer represents a Cartesian rotation and is no longer classified as
+    Rotational. Use this utility to remove any unit conversions from Coordinates
+    that were incorrectly labeled as Rotational in the past. For these Coordinates
+    it will undo the incorrect radians to degrees conversion. */
```
chrisdembia: I think the low-level explanation is important to provide, but I think we should lead with the user-facing consequences. Perhaps "This function can be used to upgrade MOT files generated with versions 3.3 and earlier in which some data columns are associated with coordinates that were incorrectly marked as Rotational (rather than Coupled). Specific instances of issue are the patella coordinate in the Rajagopal 2015 and leg6dof9musc models. In these cases, the patella will visualize incorrectly in the GUI when replaying the kinematics from the MOT file, and static optimization results may be blah blah blah."

OpenSim/Simulation/SimbodyEngine/Coordinate.cpp
```diff
@@ -677,3 +677,36 @@ void Coordinate::SpeedStateVariable::
...
+//=============================================================================
+void Coordinate::updateFromXMLNode(SimTK::Xml::Element& aNode,
+    int versionNumber)
+{
+    if (versionNumber < XMLDocument::getLatestVersion()) {
+        if (versionNumber < 30514) {
+            SimTK::Xml::element_iterator iter = aNode.element_begin("motion_type");
+            if (iter != aNode.element_end()) {
+                SimTK::Xml::Element motionTypeElm = 
+                    SimTK::Xml::Element::getAs(aNode.removeNode(iter));
```
chrisdembia: Is whitespace automatically trimmed?

OpenSim/Simulation/SimulationUtilities.h
```diff
@@ -103,9 +103,55 @@ inline SimTK::State simulate(Model& model,
...
+        return; 
+
+    // Cycle through the data files 
+    for (auto filePath : filePaths) {
+        Storage data(filePath);
+
+        // Cycle the inconsistent Coordinates
+        for (auto coord : problemCoords) {
+            // Get the corresponding column of data and if in degrees
+            // undo the radians to degrees conversion on that column.
```
chrisdembia: You could use `Storage::getStateIndex()` here to avoid duplicating the conversion between 4.0-style and 3.3-style state variable names/paths.

OpenSim/Simulation/SimulationUtilities.h
aseth1: If we didn't do the unit conversion (when we should have) we shouldn't attempt to undo it.

chrisdembia: We should make sure to have test cases (e.g., running static optimization) that ensures we undo the conversion correctly.

aymanhab:
@aseth1 I don't think a bug-fix/conversion belong permenantly in the GUI that maybe out for years to come with only a small fraction of users ever. We used to have a "convert files" utility to migrate from older version to most recent, we can try to leverage that, otherwise a command line program would be most appropriate IMHO.

aseth1:
> I don't think a bug-fix/conversion belong permenantly in the GUI that maybe out for years to come with only a small fraction of users ever. We used to have a "convert files" utility to migrate from older version to most recent, we can try to leverage that, otherwise a command line program would be most appropriate IMHO.

To which code, comment or review request are you referring to? This one "4. Can we make it easier for the user to run the utility through the GUI?"

I really don't see the issue with the GUI providing a few hooks so the user doesn't have to run the command-line utility. Are we really thinking about making transitions easy for the user by making them run command line utilities? The plan is to have a command-line utility to do the actual task, so invoking it for the user through the GUI seems almost trivial and is easier and more contextual for the user. Nice to do the conversion when you get the warning instead of ignoring it and being bitten way downstream after they have forgotten the issue.

As for being there for year to come, it doesn't seem an issue to me because if the user is getting the warning because it is an OLD file being converted to 4.0. Why should that be irrelevant even many years down the road? It really has nothing to do with time, only what model and what kinematics .mot files you might have affected. If these conditions don't hold, they won't be warned and no execution of the utility will be required.

aymanhab:
Regarding using pop-up/dialog, a while ago I created a member variable (Model::_validationLog) for handling scenarios where something goes really wrong with model conversion and users may scroll by without seeing/comprehending it. This is not used by the GUI code at the moment but seems like a perfect fit for this, so you can append a short message to this String, and the GUI can check if the string is non-empty then we can display a dialog. May help with renaming as well if initialization fails after a rename.

aymanhab:
@aseth1 yes my earlier comment regarding not belonging to the GUI was referring to "4. Can we make it easier for the user to run the utility through the GUI?" and I see a big problem with having the GUI add special options/menus/dialogs to handle individual bugs, doesn't scale, makes for maintenance nightmare and shakes confidence of new users that upgrades are problematic. If the utility can be run from the Scripting shell then that's reasonable. If this can be incorporated in a generic "upgrade" functionality that may include other functionality that could be acceptable and in line with major software upgrades seen elsewhere.

jenhicks:
A behavior that I have seen in other programs (office suite, adobe suite) .... when an incompatible/old file is detected you get a warning and also get asked if you would like the file to be automatically updated. This may be what Ajay has in mind, though I'm not sure how easy it would be to implement.

CHANGELOG.md
```diff
@@ -57,6 +57,7 @@ Converting from v3.x to v4.0
...
+- The `MotionType` of a `Coordinate` is now fully determined by the Joint. The user cannot set the `MotionType` for a `Coordinate`. There are instances such as in the *leg6dof9musc* and *Rajagopal2015* models, where a `Coordinate` was assigned an incorrect type. Specifically, when a coordinate of a `CustomJoint` is not a measure of Cartesian angle. In 4.0 the coordinate is correctly marked as `Coupled` since a function couples the coordinate value to the angular displacement of the patella in Cartesian space. **NOTE**, this causes issues (e.g.  opensim-org/opensim-gui#617, #2088) when using kinematics files generated in 3.3 (or earlier) where `Rotational` coordinates have been converted to degrees. Because OpenSim 4.0 does not recognize the coordinate's `MotionType` to be `Rotational` it will not convert it back to radians internally. For motion files generated prior to 4.0 where the file has `inDegrees=yes`, please use the following conversion utility: `updateKinematicsFilesForUpdatedModel()`. When loading a pre-4.0 model, OpenSim will warn users if any changes in `MotionType` when updating an existing model to OpenSim 4.0.
```
carmichaelong: > "... incorrect type. Specifically, when a coordinate of a `CustomJoint` is not a measure of Cartesian angle."

Consider: "... incorrect type (e.g., when a coordinate of a `CustomJoint` is not a measure of Cartesian angle)."

> "In 4.0 the coordinate is correctly marked ...."

In 4.0, the coordinate is correctly marked ...

> "OpenSim will warn users if any changes in `MotionType`"

if -> of

OpenSim/Simulation/SimbodyEngine/Coordinate.h
```diff
@@ -221,6 +221,11 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Coordinate, ModelComponent);
     SimTK::MobilizedBodyIndex getBodyIndex() const { return _bodyIndex; };
     /**@}**/
 
+    /* For internal consistency checking. Returns the user-specified MotionType
+       serialized with older model files (prior to 4.0) if one is provided, other-
```
carmichaelong: maybe "older model files (prior to 4.0)" -> "pre-4.0 model files", so that this comment doesn't get stale?

OpenSim/Simulation/SimulationUtilities.h
```diff
@@ -103,9 +103,70 @@ inline SimTK::State simulate(Model& model,
...
 /// @}
 
+/** This function can be used to upgrade MOT files generated with versions
+    3.3 and earlier in which some data columns are associated with coordinates
+    that were incorrectly marked as Rotational (rather than Coupled). Specific
+    instances of the issue are the patella coordinate in the Rajagopal 2015 and
+    leg6dof9musc models. In these cases, the patella will visualize incorrectly
+    in the GUI when replaying the kinematics from the MOT file, and static 
+    optimization results.
+    Moving from OpenSim 3.3 (and earlier) to OpenSim 4.0, MotionTypes for
```
carmichaelong: > Moving from OpenSim 3.3 (and earlier) to OpenSim 4.0,

Beginning in OpenSim 4.0,

OpenSim/Simulation/SimulationUtilities.h
```diff
@@ -103,9 +103,70 @@ inline SimTK::State simulate(Model& model,
...
+    Coordinates are now determined strictly by the coordinates' owning Joint.
```
carmichaelong: Should the second "coordinates'" be capitalized?? I guess this depends on if you meant the class or coordinates more generically.

CHANGELOG.md
chrisdembia: It would be good to hardwrap these lines so they appear nicely in a text editor.

chrisdembia:
I pushed some changes to the test case. There are still some more tests I'd like to add (see the TODOs).

aseth1:
@chrisdembia let me know how those TODOs are coming along and/or if you want tackle them together.

chrisdembia:
I'm working on them; will update the PR soon.

chrisdembia:
Ready for review.

I added a python test to ensure the new function is available from scripting.

chrisdembia:
@aseth1 and @carmichaelong can you review?

OpenSim/Simulation/SimulationUtilities.cpp
```diff
@@ -0,0 +1,174 @@
...
+}
+/// @}
+
+std::unique_ptr<Storage>
+OpenSim::updatePre40KinematicsStorageFor40MotionType(const Model& pre40Model,
+        const Storage &kinematics)
+{
+    // There is no issue if the kinematics are in internal values (i.e. not
+    // converted to degrees)
+    if(!kinematics.isInDegrees()) return nullptr;
```
chrisdembia: @aseth1 I changed the logic here; please verify this is what you intended. You originally had:

```cpp

    if(!kinematics.isInDegrees())
        if (pre40Model.getDocumentFileVersion() >= 30415) {
            throw Exception("updateKinematicsStorageForUpdatedModel has no updates "
                "to make because the model '" + pre40Model.getName() + "'is up-to-date.\n"
                "If input motion files were generated with this model version, "
                "nothing further must be done. Otherwise, provide the original model "
                "file used to generate the motion files and try again.");
        }
```

aseth1: :+1: yes returning makes most sense!

OpenSim/Simulation/Test/testSimulationUtilities.cpp
```diff
@@ -0,0 +1,151 @@
...
+void testUpdatePre40KinematicsFor40MotionType() {
+    
+    // The model and motion files for this test are from the opensim-models
+    // repository. This PR is related to issues #2240 and #2088.
+    
+    Model model("testSimulationUtilities_leg6dof9musc_20303.osim");
+    
+    // Ensure the model file has an inconsistent motion type
+    // (that it wasn't accidentally updated in the repository).
+    SimTK_TEST(!model.getWarningMesssageForMotionTypeInconsistency().empty());
```
aseth1: :+1:

OpenSim/Simulation/Test/testSimulationUtilities.cpp
```diff
@@ -0,0 +1,151 @@
...
+                "updatePre40KinematicsStorageFor40MotionType() altered columms incorrectly.");
+    }
+    
+    // Test updatePre40KinematicsFilesFor40MotionType().
+    const std::string updatedKinematicsFile =
+            "testSimulationUtilities_leg69_IK_stance_pre4_updated.mot";
+    if (IO::FileExists(updatedKinematicsFile)) {
+        std::remove(updatedKinematicsFile.c_str());
+    }
+    updatePre40KinematicsFilesFor40MotionType(model, {origKinematicsFile});
```
aseth1: Minor, but do you think this test should exercise multiple files?

OpenSim/Simulation/Test/testSimulationUtilities.cpp
```diff
@@ -0,0 +1,151 @@
...
+    if (IO::FileExists(updatedModelFile)) {
+        std::remove(updatedModelFile.c_str());
+    }
+    {
+        // TODO check that we get the exception if a model has no document version.
+        model.print(updatedModelFile);
+        Model updatedModel(updatedModelFile);
+        SimTK_TEST(updatedModel.getWarningMesssageForMotionTypeInconsistency().empty());
+        
+        SimTK_TEST_MUST_THROW_EXC(
```
aseth1: What is the exception you are expecting here? Method "has no updates to make"? Perhaps state that in a comment. 
I realize now the previous comment is a TODO but I was expecting the exception to be related to no document version.

aseth1:
> There's a test in testManager that exercises simulate(). With the creation of testSimulationUtilities, that test should probably be moved over at some point in the future.

That sounds appropriate to me.

chrisdembia:
@aseth1 @carmichaelong I've updated the PR to address your comments.



> It's unclear to me what should and shouldn't go into this. Everything seems reasonable now, but I'm not sure down the road what should or shouldn't go into there.

I'm thinking that this file holds free functions that don't have a better place to live.

aseth1:
@carmichaelong feel free to merge if your comments have been addressed.

PR #2263, opened 2018-08-16
Remove member variables/properties representing quadrants

aymanhab:
AnalyticGeometry  doesn't use quadrants now as they're not implemented by underlying DecorativeGeometry, and we don't want them sticking around in the API.

Fixes issue opensim-gui#346 follow up

### Brief summary of changes
Unused Properties/member variables were removed.

### Testing I've completed
Nothing to test since they were unused, just compiled and ran tests.

### CHANGELOG.md (choose one)

- no need to update because these classes are new in 4.0

### DIFF
--- a/OpenSim/Simulation/Model/Geometry.h
+++ b/OpenSim/Simulation/Model/Geometry.h
@@ -293,51 +293,17 @@ private:
 
 
 /**
- * Utility class used to abstract analytic geometry. This will need to be 
- * revisited when wrapping is re-done to handle quadrants or analytic shapes
- * that were supported in earlier releases before 4.0. 
+ * Utility class used to abstract analytic geometry (surfaces of revolution)
  *
- * TODO: using start/end angle may be a better choice.
+ * TODO: using start/end angle to indicate partial surfaces.
  */
 class OSIMSIMULATION_API AnalyticGeometry : public Geometry
 {    
     OpenSim_DECLARE_ABSTRACT_OBJECT(AnalyticGeometry, Geometry);
-    // Amended with a quadrants array to support pieces of analytic geometry (for wrapping)
-    OpenSim_DECLARE_LIST_PROPERTY(quadrants, std::string,
-        "Quadrants to use: combination of +X -X +Y -Y +Z -Z space separated."); 
-private:
-    bool                    _bounds[6];     //-X, +X, -Y, +Y, -Z, +Z
-    bool                    _piece;
+
 public:
-    AnalyticGeometry():
-        _piece(false)
-    {
-        constructProperties();
-        for(int i=0; i<6; i++) _bounds[i]=true;
-    }
+    AnalyticGeometry() {}
     virtual ~AnalyticGeometry() {}
-    void setQuadrants(const bool quadrants[6])
-    {
-        _piece=false;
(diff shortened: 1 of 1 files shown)


OpenSim/Simulation/Model/Geometry.h
```diff
@@ -293,51 +293,17 @@ class OSIMSIMULATION_API Arrow : public Geometry
...
 /**
- * Utility class used to abstract analytic geometry. This will need to be 
- * revisited when wrapping is re-done to handle quadrants or analytic shapes
- * that were supported in earlier releases before 4.0. 
+ * Utility class used to abstract analytic geometry (surfaces of revolution)
  *
- * TODO: using start/end angle may be a better choice.
+ * TODO: using start/end angle to indicate partial surfaces.
  */
 class OSIMSIMULATION_API AnalyticGeometry : public Geometry
```
aseth1: Would appear that `AnalyticGeometry` is no longer necessary.

aseth1: There is nothing left.

aymanhab: While the class is empty, it still serves a purpose to group types of geometry that have common features allowing uniform handling on the GUI/visualization side. 
All and only subtypes of AnalyticGeometry are built into the threejs library and performing operations on them requires different handling from other geometry primitives making the grouping useful.

aseth1: 1.  What are the common defining features of `AnalyticGeometry` vs. `Geometry`?
2.  What makes an `Ellipsoid` "analytic" and a `Brick` not?
3. It seems this idea of `Analytic` was tied to using/drawing partial geometry. Is  that right?

aymanhab: From abstraction point of view, they are surfaces of revolution. Because they are a well defined abstraction, the partial geometry rendering falls naturally in place, no surprise. 

From practical point of view, these types have built in implementation in threejs optimized for smooth rendering (compute derivatives, normals, tesselation etc.) so modifying their attributes after creation does not work, they have to be treated differently and recreated on edit. Not true for Mesh or Brick that can change attributes without recreation.

Hope this answers 1-3

OpenSim/Simulation/Model/Geometry.h
```diff
@@ -293,51 +293,19 @@ class OSIMSIMULATION_API Arrow : public Geometry
...
- * Utility class used to abstract analytic geometry. This will need to be 
- * revisited when wrapping is re-done to handle quadrants or analytic shapes
- * that were supported in earlier releases before 4.0. 
- *
- * TODO: using start/end angle may be a better choice.
+ * Abstract class for analytical geometry (e.g. surfaces of revolution) whose
+ * rendering is optimized by the graphics library (e.g. threejs). Unlike other
+ * geometry, property edits require a recreation of the AnalyticGeometry on
+ * the renderer and not simple updates. AnalyticGeometry is the base class for
+ * Sphere, Cylinder, Cone, Ellipsoid and Torus geometry.
```
aseth1: I went ahead and clarified the description myself.

aymanhab:
Thanks much @aseth1 and @chrisdembia 👍

PR #2264, opened 2018-08-16
Popup warning message when MotionType between <=3.3 to 4.0 is inconsistent

aseth1:
Final part of fixes to issue #2240 

### Brief summary of changes
Wired the message generated in #2243 to the validation log used by https://github.com/opensim-org/opensim-gui/pull/953. Added method to append to the log. 

### Testing I've completed
Loaded leg6dof9musc with a build of https://github.com/opensim-org/opensim-gui/pull/953 to produce the following error message:
image

Is the message comprehensible?

### CHANGELOG.md (choose one)
- MotionType handling was added to the CHANGELOG in #2240

### DIFF
--- a/OpenSim/Simulation/Model/Model.cpp
+++ b/OpenSim/Simulation/Model/Model.cpp
@@ -655,19 +655,13 @@ std::string Model::getWarningMesssageForMotionTypeInconsistency() const
     if (message.size()) {
         message = "\nModel '" + getName() + "' has inconsistencies:\n" + message;
         message += 
-            "You must update any motion files you generated in version 3.3 or earlier. You can:\n"
-            "  (1) Run the updateKinematicsFilesForUpdatedModel() utility (in the scripting shell) OR\n"
-            "  (2) Re-run the inverse kinematics tool with the updated model in 4.0\n"
-            "In versions 3.3 and earlier, we allowed some Coupled coordinates to be incorrectly\n"
-            "labeled as Rotational. Coordinate values for Coupled coordinates previously (pre-4.0)\n"
-            "labeled Rotational, can be assigned values that are in degrees if a coordinate. This\n"
-            "will lead to incorrect motion when playing back a pre-4.0 motion file (.mot or .sto in\n"
-            "degrees) and produces erroneous velocity and acceleration estimates that\n"
-            "yield incorrect inverse-dynamics and static-optimization results. You can\n"
-            "apply the utility:\n"
-            "\tupdateKinematicsFilesForUpdatedModel(model_file, list_of_MOT_files)\n"
-            "to update motion files generated prior to 4.0 and undo the conversion of\n"
-            "Coupled coordinates to degrees so that they are consistent with the 4.0 model.";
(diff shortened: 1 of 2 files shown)


aymanhab:
@aseth1 The message is clear to me but would be quite intimidating to most users to digest in one shot (as you would expect from a dialog in an application rather than a page full of text), and if they hit ok they're dead! 
I'd lean to shorten the message with a stern warning about what could be wrong suggest rerunning the tools and to add a link to a  "living" confluence page that describes what users need to do, and why and what tools to use to fix the problem if they have old result files.

jenhicks:
I suggest you replace text starting with "Consequently ..." with something like the following

**You must update any motion files you generated in version 3.3 or earlier**. You can:

1) Run the utility .... (in the scripting shell) OR
2) Re-run the inverse kinematics tool with your model in 4.0

In versions 3.3 and earlier, we allowed some Coupled coordinates to be incorrectly labeled as "Rotational" degrees of freedom .... [Then add in the explanatory/background text]

aseth1:
@chrisdembia the message was updated based on feedback above. Please test with https://github.com/opensim-org/opensim-gui/pull/953 to see if the message is readable. Thanks!

chrisdembia:
Here's the latest message:

image

chrisdembia:
The name of the method is incorrect; the correct name is `updatePre40KinematicsFilesFor40MotionType()`.


Here is my suggested revision:

            "You must update any motion files you generated in version 3.3 or earlier. You can:\n"
            "  (1) Run the updatePre40KinematicsFilesFor40MotionType() utility (in the scripting shell) OR\n"
            "  (2) Re-run the inverse kinematics tool with the updated model in 4.0\n"
            "In versions prior to 4.0, we allowed some Coupled coordinates to be incorrectly\n"
            "labeled as Rotational. This\n"
            "will lead to incorrect motion when playing back a pre-4.0 motion file (.mot or .sto in\n"
            "degrees) and produces inverse-dynamics and static-optimization results."

aseth1:
@chrisdembia please go ahead and update. Thanks!

chrisdembia:
@aseth1 I've updated the message (and made some additional tweaks beyond my comment above).

aymanhab:
I like the details but don't think "motionType" should appear at the top of user-facing messages. I'd use something like "was labeled as Rotational but was found to be Coupled based on joint definition" this would avoid the impression that the decision was arbitrary on our side.

chrisdembia:
@aymanhab I incorporated your feedback. Is it possible to have the GUI wrap the text for us so we don't need to put in so many newlines explicitly?

aymanhab:
@chrisdembia This off-the-shelf warning message does not have much formatting support.  The html formatted message with scrollbar had more since it's html but wasn't appealing to the team.

chrisdembia:
Okay well then this PR LGTM.

aymanhab:
LGTM as well @chrisdembia 
For this to show up in the gui though we need to merge https://github.com/opensim-org/opensim-gui/pull/953

PR #2268, opened 2018-08-23
Fix GeometryPath method to place new PathPoints 

aymanhab:
Fixes issue #0 obvious bug!

### Brief summary of changes
Method GeometryPath::placePathPoint was broken by earlier change in 4.0 that replaced passing a reference to the new point's location with passing a copy that gets thrown away on return. This PR restores old functionality.
### Testing I've completed
Built GUI with this PR and new pathpoints are  placed consistent with 3.3 (though not exact) where inserting at end points extends the geometry path out a little and inserting between points performs linear interpolation.

### CHANGELOG.md (choose one)

- no need to update because this is a bugfix

### DIFF
--- a/OpenSim/Simulation/Model/GeometryPath.cpp
+++ b/OpenSim/Simulation/Model/GeometryPath.cpp
@@ -492,10 +492,12 @@ addPathPoint(const SimTK::State& s, int aIndex, const PhysicalFrame& frame)
 {
     PathPoint* newPoint = new PathPoint();
     newPoint->setParentFrame(frame);
-    Vec3 location = newPoint->getLocation(s); // 0 by construction
-    // location is now a COPY not a reference so the computed location is LOST unless setLocation is invoked
-    placeNewPathPoint(s, location, aIndex, frame);
-    newPoint->setLocation(location); // 
+    Vec3 newLocation(0.0); 
+    // placeNewPathPoint takes a newLocation reference and changes it 
+	// based on the index in the path (interpolate or extend, in ground frame)
+    placeNewPathPoint(s, newLocation, aIndex, frame);
+	// Now set computed new location into the newPoint
+    newPoint->setLocation(newLocation);
     upd_PathPointSet().insert(aIndex, newPoint);
 
     // Rename the path points starting at this new one.
--- a/OpenSim/Simulation/Model/GeometryPath.cpp
+++ b/OpenSim/Simulation/Model/GeometryPath.cpp
@@ -492,8 +492,10 @@ addPathPoint(const SimTK::State& s, int aIndex, const PhysicalFrame& frame)
 {
     PathPoint* newPoint = new PathPoint();
     newPoint->setParentFrame(frame);
-    Vec3 location = newPoint->getLocation(s);
+    Vec3 location = newPoint->getLocation(s); // 0 by construction
+    // location is now a COPY not a reference so the computed location is LOST unless setLocation is invoked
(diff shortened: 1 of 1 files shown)


OpenSim/Simulation/Model/GeometryPath.cpp
```diff
@@ -492,8 +492,10 @@ addPathPoint(const SimTK::State& s, int aIndex, const PhysicalFrame& frame)
 {
     PathPoint* newPoint = new PathPoint();
     newPoint->setParentFrame(frame);
-    Vec3 location = newPoint->getLocation(s);
+    Vec3 location = newPoint->getLocation(s); // 0 by construction
```
aseth1: I don't understand this comment. Where would you expect the location to be if none was set?

OpenSim/Simulation/Model/GeometryPath.cpp
```diff
@@ -492,8 +492,10 @@ addPathPoint(const SimTK::State& s, int aIndex, const PhysicalFrame& frame)
...
+    // location is now a COPY not a reference so the computed location is LOST unless setLocation is invoked
```
aseth1: Also do not understand the relevance of  this comment. `location` is a `Vec3` above not a `Vec3&`!  What is lost?
The passing  in of a reference you expect to get overwritten in `placeNewPathPoint` is what is unexpected/wrong in this scheme and that should be documented here.

OpenSim/Simulation/Model/GeometryPath.cpp
aseth1: You do not even need to as newPoint for its location since you know you are going to have to set it correctly.  It would be more readable if this was just `Vec3 newLocation;`

OpenSim/Simulation/Model/GeometryPath.cpp
aymanhab: @aseth1 The comments were for myself trying to understand what went wrong when the change was made to getLocation(s). I agree the current call sequence doesn't make sense, I should refer you to the developer who made the change in the first place 😃 

I'll update anyway per your feedback. Thank you.

OpenSim/Simulation/Model/GeometryPath.cpp
```diff
@@ -492,10 +492,12 @@ addPathPoint(const SimTK::State& s, int aIndex, const PhysicalFrame& frame)
 {
     PathPoint* newPoint = new PathPoint();
     newPoint->setParentFrame(frame);
-    Vec3 location = newPoint->getLocation(s); // 0 by construction
-    // location is now a COPY not a reference so the computed location is LOST unless setLocation is invoked
-    placeNewPathPoint(s, location, aIndex, frame);
-    newPoint->setLocation(location); // 
+    Vec3 newLocation(0.0); 
+    // placeNewPathPoint takes a newLocation reference and changes it 
+	// based on the index in the path (interpolate or extend, in ground frame)
```
aseth1: perhaps something like:

Note: placeNewPathPoint() returns a location by reference. It computes a new location according to the index where the new path point will be inserted along the path (among the other path points).

chrisdembia: Also, convert tabs to spaces

aymanhab:
comment updated, tabs removed.

chrisdembia:
It seems that @aseth1's concerns were addressed. I'll merge.

aymanhab:
Great, thanks @chrisdembia

Issue #2271, opened 2018-08-24
StatesTrajectory::createFromStatesStorage(): assembling a model with user set default coordinate values produces the wrong trajectory


nickbianco:
This was observed while trying to visualize results from a direct collocation problem after converting the solution to a StatesTrajectory:

- If constraints were enforced, I could only visualize the correct solution if I commented out the lines that set the default values in my programmatically created model.
- If default coordinate values were set, I could only visualize the correct solution if I set the `enforceConstraints` argument to false. 

The option to assemble a model to enforce constraints was added in #1922.

chrisdembia:
Nice detective work. I don't understand why explicitly setting the default value causes a change in behavior.

chrisdembia:
I'm happy to look at this with you today.

chrisdembia:
Thanks @aymanhab for the reminder but @nickbianco is using the simbody visualizer.

chrisdembia:
@nickbianco checked and it's not that calling `setDefaultValue()` causes the issue; it's specifically when the default value is not 0.

chrisdembia:
Perhaps `enforceConstraints` should be `false` by default. This would be consistent with the discussion in https://github.com/opensim-org/opensim-gui/issues/86

We should make this change before releasing 4.0, because it would be backwards-incompatible.

nickbianco:
>@nickbianco checked and it's not that calling setDefaultValue() causes the issue; it's specifically when the default value is not 0.

In addition, the states that get created in the StateTrajectory's are always "locked" to a constant value at every time point. It seems like `assemble` is trying really hard to obey the constraint, to the point where it's not obeying the current state coordinate values.

PR #2274, opened 2018-08-24
Do not enforceConstraints by default.

chrisdembia:
Related to #2271 

### Brief summary of changes

As-is, `StatesTrajectory::createFromStatesStorage()` will enforce constraints via assembly, by default. However, based on discussion in https://github.com/opensim-org/opensim-gui/issues/86, it seems that we do not want to call `assemble()` by default.

### Testing I've completed

Ran test suite

Should we make this change?

### CHANGELOG.md (choose one)

- no need to update because...this function has not been released yet.

### DIFF
--- a/OpenSim/Simulation/StatesTrajectory.h
+++ b/OpenSim/Simulation/StatesTrajectory.h
@@ -438,9 +438,11 @@ public:
      *      columns in the Storage that do not correspond to continuous state
      *      variables in the Model. If true, such columns of the Storage are
      *      ignored.
-     * @param enforceConstraints Modify state variable values to satisfy
-     *      kinematic constraints (by calling Model::assemble()). Be careful
-     *      with this option; enforcing constraints can drastically alter the
+     * @param assemble Modify state variable values to satisfy
+     *      kinematic constraints (by calling Model::assemble()).
+     *      Use this option if the provided states are incomplete (for example,
+     *      the values for dependent coordinates are unspecified).
+     *      Caution: enforcing constraints can drastically alter the
      *      provided states if they do not already obey the constraints.
      *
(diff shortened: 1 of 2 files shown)


OpenSim/Simulation/StatesTrajectory.h
```diff
@@ -439,7 +439,9 @@ class OSIMSIMULATION_API StatesTrajectory {
      *      variables in the Model. If true, such columns of the Storage are
      *      ignored.
      * @param enforceConstraints Modify state variable values to satisfy
-     *      kinematic constraints (by calling Model::assemble()).
+     *      kinematic constraints (by calling Model::assemble()). Be careful
+     *      with this option; enforcing constraints can drastically alter the
+     *      provided states if they do not already obey the constraints.
```
aseth1: The comment is helpful, but perhaps the flag should be `performAssembly` instead. Then the comment would say perform the assembly if the provided states are incomplete, e.g. do not provide values for dependent coordinates of constraints.

chrisdembia:
@aseth1 I've addressed your feedback. Ready for review.

chrisdembia:
Thanks @aseth1 . Looking for another reviewer.

aymanhab:
@chrisdembia The tricky part to explain is that results of forward simulation do not meet assembly tolerance and so could also be changed. Not sure if that needs mention, otherwise LGTM

chrisdembia:
Thanks @aymanhab I've addressed your comment.

aymanhab:
Perfect, thanks @chrisdembia

chrisdembia:
Thanks @aymanhab. The tests passed previously and I only updated documentation. Merging.

Issue #2244, opened 2018-08-09
DataTable_::flatten() takes too long for reasonably-sized data files


chrisdembia:
In #2237 @nickbianco said:

> The only issue I ran into was that for larger files (60+ seconds worth of motion data), the script got hung up [here](https://github.com/opensim-org/opensim-core/blob/81e1c285b5ef82ddecb3b34e22df634f6c9802f8/Bindings/Java/Matlab/Utilities/osimC3D.m#L210) where the TimeSeriesTableVec3 of the forces gets flattened. @jimmyDunne have you tested larger files, and if so, how long do they take to flatten the forces? In the meantime, I'll time the larger file I tried to see how long it actually takes.

> For the file I mentioned, it took ~18.9 minutes to write an MOT file for 60 seconds worth of data.

@aseth1:

> It is appending rows one by one, which resizes (and moves data). I had a fix for data reading that should be the same solution of preallocating the size and just updating the contents. Do you want to take a look @nickbianco?

This is the line @aseth1 is referring to: https://github.com/opensim-org/opensim-core/blob/c461d0290a9413e4552b32dc6a1f55d08fb15980/OpenSim/Common/DataTable.h#L389

jenhicks:
@nickbianco Will file a new issue about scaling all the values of the matrix.

nickbianco:
#2280 created for the DataTable fix.
#2281 created to capture the matrix scaling for unit conversion issue.

chrisdembia:
I checked locally and `c3dExport` worked fine and I was able to preview the resulting GRF MOT file in the GUI. `c3dExport` is still fairly slow (35 seconds for [opensim-core/OpenSim/Tests/shared/walking5.c3d](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Tests/shared/walking5.c3d)). Here's some profiling results:

![image.png](https://images.zenhubusercontent.com/549329337c6ccc6277a6ab20/46aa1df7-cb0d-448b-8d79-751718a8a15b)

@nickbianco is right that we should speed up `convertMillimeters2Meters` but it would also be good to speed up `osimTableToStruct`.

nickbianco:
[...]

PR #2280, opened 2018-09-05
DataTable constructor fixes for slow pack and flatten operations

nickbianco:
Fixes issue #2244 

### Brief summary of changes
- `DataTable` constructors for flatten and pack operations now modify the underlying `SimTK::Matrix` directly rather than using `appendRow()`, which resized the `SimTK::Matrix` upon each call, causing performance issues.

### Testing I've completed
- Ran `testDataTable` and `testC3DFileAdapter` locally, which passed
- Used the Matlab `osimC3D` utility to write GRFs to a MOT file. The flatten operation in Matlab, which used to take ~15 minutes, now takes <1 second.
- AppVeyor and windows Travis builds pass.

- At the time of writing, the Mac Travis build wasn't passing for some reason. Retriggered it since it passed on the last commit, which was the same except for one line of unused code. 

### CHANGELOG.md (choose one)
- no need to update because changes are for performance only, nothing user-facing

### DIFF
--- a/OpenSim/Common/DataTable.h
+++ b/OpenSim/Common/DataTable.h
@@ -1402,7 +1402,7 @@ protected:
     template<typename ELT>
     static
     SimTK::RowVector_<double> splitElement(const ELT& elt) {
-        SimTK::RowVector_<double> result;
+        SimTK::RowVector_<double> result(numComponentsPerElement_impl(elt));
         splitElementAndAppend(result.begin(), result.end(), elt);
         return result;
     }
--- a/OpenSim/Common/DataTable.h
+++ b/OpenSim/Common/DataTable.h
@@ -224,25 +224,20 @@ public:
         setColumnLabels(thisLabels);
 
         // Construct matrix for this table from that table.
-        SimTK::Matrix_<ETY> depData((int)that.getNumRows(),
+        _depData.resize((int)that.getNumRows(),
             (int)that.getNumColumns() * numComponentsPerElement());
+        SimTK::RowVector_<ETY> 
+        thisRow((int)that.getNumColumns() * numComponentsPerElement());
         for(unsigned r = 0; r < that.getNumRows(); ++r) {
             const auto& thatRow = that.getRowAtIndex(r);
-            std::vector<ETY> thisRow{};
-            for(unsigned c = 0; c < that.getNumColumns(); ++c)           
-                splitElementAndPushBack(thisRow, thatRow[c]);
-            // This loop could be replaced with
-            //      depData.updRow(r) = thisRow;
-            // if splitElementAndPushBack supported SimTK::RowVector_<ETY>
-            for(unsigned comp = 0; 
-                comp < that.getNumColumns() * numComponentsPerElement();
-                ++comp) {
-                depData.set(r, comp, thisRow.at(comp));
-            }
+            for (unsigned c = 0; c < that.getNumColumns(); ++c)
+                splitElementAndAppend(
+                    _depData.updRow(r).begin() + c*numComponentsPerElement(), 
+                    _depData.updRow(r).end(),
+                    thatRow[c]);
         }
 
         _indData = that.getIndependentColumn();
-        _depData = depData;
     }
 
     /** Construct this DataTable from a DataTable_<double, double>. This is the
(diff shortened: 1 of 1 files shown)


OpenSim/Common/DataTable.h
```diff
@@ -375,19 +380,19 @@ class DataTable_ : public AbstractDataTable {
...
         for(unsigned r = 0; r < that.getNumRows(); ++r) {
-            const auto& thatInd = that.getIndependentColumn().at(r);
             auto thatRow = that.getRowAtIndex(r).getAsRowVector();
-            std::vector<ETY> thisRow{};
             for(unsigned c = 0;
-                c < that.getNumColumns();
-                c += numComponentsPerElement()) {
-                thisRow.push_back(makeElement(thatRow.begin() + c,
-                                              thatRow.end()));
+                c < that.getNumColumns() / numComponentsPerElement(); ++c) {
```
chrisdembia: Does it make sense to have a local variable for the new number of columns?

nickbianco:
@aseth1, @chrisdembia want to give this a look?

OpenSim/Common/DataTable.h
```diff
@@ -1350,26 +1355,42 @@ class DataTable_ : public AbstractDataTable {
...
-    void splitElementAndPushBack(std::vector<double>& row,
-                                 const SimTK::Vec<N>& elem) {
-        for(unsigned i = 0; i < N; ++i)
-            row.push_back(elem[i]);
+    void splitElementAndAppend(Iter begin, Iter end, 
+                               const SimTK::Vec<N>& elem) {
+        for(unsigned i = 0; i < N; ++i) {
+            OPENSIM_THROW_IF(begin == end,
+                InvalidArgument,
+                "Iterators do not produce enough elements."
```
chrisdembia: I think you need a space after the period (here and below).

OpenSim/Common/DataTable.h
```diff
@@ -1350,26 +1355,42 @@ class DataTable_ : public AbstractDataTable {
...
-            for(unsigned j = 0; j < N; ++j)
-                row.push_back(elem[i][j]);
+    void splitElementAndAppend(Iter begin, Iter end, 
+                               const SimTK::Vec<M, SimTK::Vec<N>>& elem) {
+        for(unsigned i = 0; i < M; ++i) {
+            for(unsigned j = 0; j < N; ++j) {
+                OPENSIM_THROW_IF(begin == end,
+                    InvalidArgument,
+                    "Iterators do not produce enough elements."
+                    "Expected: " + std::to_string(M * N) +
```
chrisdembia: Add space; `" Expected: "`.

OpenSim/Common/DataTable.h
```diff
@@ -1350,26 +1355,42 @@ class DataTable_ : public AbstractDataTable {
...
     // Split element into constituent components and append the components to
     // the given vector. For example Vec3 has 3 components.
-    template<int N>
+    template<int N, typename Iter>
     static
-    void splitElementAndPushBack(std::vector<double>& row,
-                                 const SimTK::Vec<N>& elem) {
-        for(unsigned i = 0; i < N; ++i)
-            row.push_back(elem[i]);
+    void splitElementAndAppend(Iter begin, Iter end,
```
chrisdembia: I wouldn't say this function is appending. What about `splitElementAndCopy()`?

OpenSim/Common/DataTable.h
```diff
@@ -1350,26 +1355,42 @@ class DataTable_ : public AbstractDataTable {
 
     // Split element into constituent components and append the components to
     // the given vector. For example Vec3 has 3 components.
```
chrisdembia: Can you add to the documentation that this function will write `N` elements starting from `begin` and does not necessarily write all the way to `end`?

OpenSim/Common/DataTable.h
nickbianco: Since `_depData` has already been resized, I could use `this->getNumColumns()`.

OpenSim/Common/DataTable.h
nickbianco: I like that better, good suggestion.

OpenSim/Common/DataTable.h
chrisdembia: +1

OpenSim/Common/DataTable.h
nickbianco: Actually, I don't think I need this space if I add after the period above.

nickbianco:
Thanks @chrisdembia. I addressed your comments.

OpenSim/Common/DataTable.h
```diff
@@ -1348,28 +1352,48 @@ class DataTable_ : public AbstractDataTable {
...
-                                 const SimTK::Vec<M, SimTK::Vec<N>>& elem) {
-        for(unsigned i = 0; i < M; ++i)
-            for(unsigned j = 0; j < N; ++j)
-                row.push_back(elem[i][j]);
+    void splitElementAndCopy(Iter begin, Iter end, 
+                               const SimTK::Vec<M, SimTK::Vec<N>>& elem) {
+        for(unsigned i = 0; i < M; ++i) {
+            for(unsigned j = 0; j < N; ++j) {
+                OPENSIM_THROW_IF(begin == end,
+                    InvalidArgument,
```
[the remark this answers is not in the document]
nickbianco: What do you think the correct exception would be then?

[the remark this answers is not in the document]
chrisdembia: I think `InvalidArgument` is fine but you could just use the base class exception.

nickbianco:
After looking at OpenSim's Exception classes, I decided to just change `InvalidArgument` to `Exception` wherever an exception was being thrown due to an iterator running out of room. 

Ready for review.

OpenSim/Common/DataTable.h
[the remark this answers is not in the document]
aseth1: What about `IndexOutOfRange`? This can be interpreted as the table being too small or the element being to big, which I think is fine.

Issue #2121, opened 2018-03-13
Inform user/GUI that a Component has been renamed or properties have changed


aseth1:
[GUI #483](https://github.com/opensim-org/opensim-gui/issues/483) notes concern for changes to models upon loading that lead to load failures. Informing the user of changes can help them adapt/fix the model so they can load it OpenSim.

From discussion there:

> In the API we can change warnings to Exceptions, but GUI will have catch and throw up a dialog with the message. If there are many renames (e.g. dozens to hundreds of PathPoints) this could be very agitating. Would need away to "ignore all". 
> 
> Another option, is to add/return a flag that informs us if `finalizeFromProperties` has changed any property values, and in that case the GUI would show a single pop-up with a message to check the messages console for the list of the components that were renamed.

I am leaning towards the second option. @aymanhab @tkuchida @chrisdembia?

chrisdembia:
The second option seems like a bit of a hack to me. `finalizeFromProperties()` might edit properties other than the name and thus the pop-up could be generated for a false flag.

The load failure that currently occurs seems to be the exception that is thrown if a socket cannot find its connectee. Could we add a note in that message saying "this may be because the non-unique component names were changed; see the Messages window"?

aseth1:
@chrisdembia I was thinking the second option would be more general than just names. @tkuchida raised concerns with any values being updated by `finalizeFromProperties()` not just naming. For example, illegal mass properties get massaged by Body and it might be good to alert the user of those changes.

I also like your more direct fix for the unable to connect issue.

chrisdembia:
> @chrisdembia I was thinking the second option would be more general than just names.

Ah okay, if the GUI pop-up is equally general, then that could be nice. Would you be able to use `isObjectUpToDateWithProperties()` to automatically determine with return value, so you do not rely on component developers to manually specify the return value? That is, ideally `extendFinalizeFromProperties()` continues to return `void`.

aseth1:
I don't think that `isObjectUpToDateWithProperties()` can help us here since a user edit prior to `finalizeFromProperties` will still make the object dirty.  One proposal @tkuchida discussed with me at UEFS is to clone the properties in the base method, invoke the derived extend and the check if properties were edited. Perhaps we could even generically present which property was changed and from old to new value.

Issue #2024, opened 2017-11-20
SoccerKickingModel fails to open


tkuchida:
@aymanhab commented on [Wed Nov 08 2017](https://github.com/opensim-org/opensim-models/issues/41)

Multiple warnings due to space in model name (likely common), then failure/abort open due to Error
<img width="542" alt="screen shot 2017-11-08 at 9 58 18 am" src="https://user-images.githubusercontent.com/3972208/32565326-9863af5e-c46b-11e7-8cf2-bb5f8a9dcc29.png">

---

@tkuchida commented on [Wed Nov 08 2017](https://github.com/opensim-org/opensim-models/issues/41#issuecomment-342937348)

Possibly related to [1970 on opensim-core](https://github.com/opensim-org/opensim-core/issues/1970).

aseth1:
I am doubtful that the error is due to a space in a name.  It is complaining about a an invalid index, which means it is connected to the Body but that Body is does not have a valid underlying (system) index yet. Could be an issue with the MultibodyTree that was generated. I'll take a look.

aseth1:
There appears to be an issue with the `SimTK::MultibodyGraphMaker` when the root body is massless. This is the resulting multibody graph for a model with zero mass for the root, *swivel*, `Body` connected to *ground*.  Everything beyond the *femur_r* `Body` was removed for brevity.
```
MULTIBODY GRAPH
---------------

5 BODIES:
 0  0: /testMasslessRoot/ground mass=inf mob=-1 master=-1
  jointsAsParent=[ 0 3]   jointsAsChild=[]        slaves=[]
 1  1: /testMasslessRoot/swivel mass=0 mob=0 master=-1
  jointsAsParent=[ 1]     jointsAsChild=[ 0]      slaves=[]
 2  2: /testMasslessRoot/pelvis mass=10.4349 mob=1 master=-1
  jointsAsParent=[ 2]     jointsAsChild=[ 1]      slaves=[]
 3  1: /testMasslessRoot/femur_r mass=8.54927 mob=2 master=-1
  jointsAsParent=[]       jointsAsChild=[ 2 3]    slaves=[ 4]
 4  3: #/testMasslessRoot/femur_r_slave_1 mass=nan mob=3 master=3
  jointsAsParent=[]       jointsAsChild=[ 2]      slaves=[]

4 JOINTS:
[...]

Issue #2276, opened 2018-08-27
Remove the model name from absolute component paths.


chrisdembia:
In discussion with @aymanhab and @carmichaelong 

- [x] Matlab HopperDevice (fix submitted)
- [x] C++ exampleHopperDevice
- [x] Matlab KneeReflex(_answers)
- [x] wiringInputsAndOutputsWithTableReporter
- [x] OpenSimCreateTugOfWarModel (fix submitted)
- [x] prescribeMotionInModel
- [x] setupAndRunAnalyzeBatchExample
- [x] setupAndRunIKBatchExample
- [x] strengthScaler

chrisdembia:
For example: my hard drive is called `Macintosh HD`, but paths are `/Users/chris/...`, not `/Macintosh HD/Users/chris`. The first slash `/` indicates "root" and the hard drive name is not necessary.

chrisdembia:
Try running Matlab examples (Dynamic Walker, Hopper Device).

chrisdembia:
I'm making a PR for a fix for the Matlab HopperDevice example.

chrisdembia:
The OutputReporter accepts a connectee path `/com_position` but the correct path is `/|com_position` (this is the column label from the output file of the OutputReporter).

chrisdembia:
The OutputReporter does not use the proper syntax for input connectee paths. The output should be separated by a vertical line, `|`.

jenhicks:
I can confirm this is the case. I also specified "bodyset/hand_r/position" which resulted in a file with the column labels bodyset/hand_r|position.

Why are we using the vertical lines and slashes? (I'm sure this has been discussed previously, but I don't recall)

jenhicks:
(And it is weird to have a column label that leads with / | )

chrisdembia:
The vertical line allows us to distinguish an output name from component names. We could use a slash but then it's not clear if the element after the last slash is supposed to be a component or an output.

chrisdembia:
Perhaps we can discuss this in tomorrow's stand-up.

aymanhab:
@chrisdembia Should I try to verify this in latest GUI build or are there loose ends in Path specification?

chrisdembia:
There will be updates to OutputReporter but otherwise this is ready to verify.

scrum2a:
James is working on updating Dynamic Walking.

scrum2a:
Test prescribeMotionInModel.

[...]

PR #2285, opened 2018-09-10
Make ModelComponentSet a ModelComponent

aseth1:
Addresses a significant cause of the issues in #2024, #2121 and https://github.com/opensim-org/opensim-gui/issues/483 that stem from the strictly enforced requirement that all Components have unique pathnames. This fix allows us to use old model files with a `Body`, `Joint` and even a `Marker` sharing the same name as long as they live in different `ModelComponentSet`s. For example, what used to require renaming for having a *toes* ` Joint` and *toes* `Body` (which also broke the `Socket` connection to the *toes* Body because it was renamed) are now both unique as */model/jointset/toes* and */model/bodyset/toes*. As a consequence many older model will have fewer renaming and connection issues.

As part of the serialization fix, it makes Ground's name unchangable, which also addresses  #1876 (3.).

### Brief summary of changes
1. Made a `Set<ObjectTyp, DerivedType>` mixin so that the Set interface would not have to be duplicated for Sets of objects and Sets of Components.

2. `Object` Sets default to` Set<T, Object>` which is exactly what it was before. Thus, FunctionSet, ScaleSet, etc... are unchanged.

3. `Component` Sets and `ModelComponent` Sets now derive from ` Set<T, Component>` and ` Set<T, ModelComponent>` respectively.

4. Gutted `ModelComponentSet` and  derived implementations since they no longer need custom setup and model connection implementations since that machinery is provided by `Component` and `ModelComponent` interfaces.

5. Removed a KLUDGE in `Property` writing that would change the name of the `Component` (`Object`) during printing (which is supposed to be const!) that caused `Set` names to be overwritten and resulted in-memory and serialized->deserialized models to differ. Now `Object` XML writing (node updating) is responsive to the Property being *isUnnamedProperty* and logic  was added to decide when it is appropriate to use the Property name in place of the `Object` name. 

6. `ExternalLoads` and its convoluted handling by the Tools was reworked so that is just added to the `Model` like any `ModelComponent` and it is configured by overriding the extension of the Component interface. For example, Tools no longer have to manually add the individual `ExternalForces` to the `Model`. Much more could be done to separate `ExternalLoads` from Tools, but as it stands these changes do not effect Tool interface or existing setup files.

7.  The bulk of the remaining changes are updates to component access to include the intermediate `ModelComponentSet`s in their pathnames and similarly for finding accessing state variable values by name. In many instance, asking the Component for its stateVariableNames() was a lot simpler.

### Testing I've completed
~~More than 80% of the tests pass from `Object` serialization, `Component` interface up to most of the Tools, but the following fail:~~

~~Java bindings fail to build due to:~~

UPDATE 9/20/2018 Merged master into branch. All tests pass locally.

1. Changes to `Set` and its impact on Java and Python bindings. @chrisdembia @aymanhab 
2. Any other potential fallout from removing the KLUDGE to Object/Property serialization? @aymanhab

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2285)


aymanhab:
@aseth1 this is a massive PR to review. Will it be possible to separate the bugfix (item 5 on your list) into a separate PR/branch so it can be separately reviewed? By itself this shouldn't break any test cases or require wrapping changes. Thank you.

aseth1:
> Will it be possible to separate the bugfix (item 5 on your list) into a separate PR/branch so it can be separately reviewed? 

No, I tried to do that. I need the fix for serialization and other tests to pass. My suggestion is to focus on the changes to Object and properties only, at this time. The "massive" number of changes are mostly repetitive changes to pathnames or their formation.

aymanhab:
Generally speaking, using the  "using Super::*" to inherit constructors ; The SWIG documentation says "This is parsed correctly, but the additional constructors are not currently added to the derived proxy class in the target language"

OpenSim/Simulation/Model/Ground.cpp
```diff
@@ -32,6 +32,8 @@
 using namespace std;
 using namespace OpenSim;
 
+#define GroundNameString "ground"
```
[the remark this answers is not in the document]
chrisdembia: This would preferably be a variable, not a macro.

OpenSim/Simulation/Model/Ground.cpp
```diff
@@ -41,10 +43,22 @@ using namespace OpenSim;
...
+    Super::extendFinalizeFromProperties();
+    // Ground is always named "ground"
+    if (getName() != GroundNameString) {
+        std::string msg = getConcreteClassName() + " '" + getName() + "' ";
+        setName(GroundNameString);
+        msg += "was renamed and is being reset to '" + getName() + "'.";
+        std::cout << msg << std::endl;
+    }
+}
+
```
chrisdembia: Record that this fixes one of the other issues on GitHub.

chrisdembia:
@aseth1 I pushed some fixes for Clang.

chrisdembia:
@aseth1 please review the SWIG warnings; they will guide you toward how to fix the wrapping.

aymanhab:
As of now PathPointSet does not wrap as a Set (likely swig wrapping issue), same for WrapObjectSet which may explain most of the 100 errors trying to compile the GUI. Way too early to know the side-effects.

aseth1:
@aymanhab all core tests are passing locally. Have you tried building the GUI against this branch. The only interface change is that `ModelComponentSet` constructors no longer have a constructor that takes a `Model`.  I don't think this should be a major issue. If you are just using the Set as container you don't need a model. If you want to augment a Model by including the ModelComponentSet, you need only add it to the model via `Model::addModelComponent()` and there should be no reason to manually invoke `connectToModel()`. Please give i a try. I am sure we can work through the changes together tomorrow.

aseth1:
> 1. Given that this PR makes path names longer, should we address #2276 along with this issue?

I think we can keep them separate and we should get both done.

> 2. This is why we had to create a SWIG macro for the Joint constructors:

👍 

>  There's a lot of if (...) else (prepend "./<type>set/") when finding dependencies without sockets. Could this be handled in updateFromXMLNode() instead? I think we really should avoid all these new if-statements if possible.

I agree this is ugly but I also wanted to invest minimum time on these dependencies because they should all be `Socket`s. That can be a subsequent PR  (even post 4.0  since functionally nothing should change except for better/consistent error handling).

>    What is the reason for the use of the initial period in the paths, e.g., ./bodyset/...?
When I see a path that begins  with "./" it indicates to me a relative path starting where I am. I am happy to remove if it causes confusion.

OpenSim/Analyses/InverseDynamics.cpp
```diff
@@ -344,10 +344,11 @@ record(const SimTK::State& s)
         const Coordinate& coord = *coordinates[_accelerationIndices[i]];
         int ind = _statesStore->getStateIndex(coord.getSpeedName(), 0);
         if (ind < 0){
-            string fullname = coord.getJoint().getName() + "/" + coord.getSpeedName();
+            // get the full coordinate speed state variable path name
+            string fullname = coord.getStateVariableNames()[1];
```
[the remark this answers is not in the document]
aseth1: This is more reliable than before. The state variable names are always listed with the Component's added states first. Only changing the order of coordinate *value* and *speed* would break this.

OpenSim/Common/Object.h
```diff
@@ -542,9 +547,7 @@ class OSIMCOMMON_API Object
 
     This flag is cleared automatically but if you want to clear it manually
     for testing or debugging, see clearObjectIsUpToDateWithProperties(). **/
```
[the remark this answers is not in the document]
aseth1: Restored `setObjectIsUpToDateWithProperties()` (so no there is no renaming or tree traversal when it is invoked)

OpenSim/Common/Object.cpp
```diff
@@ -1053,8 +1052,15 @@ updateXMLNode(SimTK::Xml::Element& aParent) const
     
     // GENERATE XML NODE for object
     SimTK::Xml::Element myObjectElement(getConcreteClassName());
-    if (!getName().empty())
+    
+    // if property is provided and it is not of unnamed type, use the property name
+    if(prop && !prop->isUnnamedProperty()) {
+            myObjectElement.setAttributeValue("name", prop->getName());
```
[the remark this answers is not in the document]
aseth1: Yes, and prop is only supplied in the case of a single-object property:
https://github.com/opensim-org/opensim-core/blob/54d7be61f25e8e4be4f8525afb3499dc76b6d055/OpenSim/Common/AbstractProperty.cpp#L205-L226

I can add another condition here to double check.

OpenSim/Common/Component.cpp
```diff
@@ -157,6 +157,8 @@ void Component::finalizeFromProperties()
 {
     reset();
 
+    setObjectIsUpToDateWithProperties();
```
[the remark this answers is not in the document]
aseth1: I separated the name updating part to `Object::makeObjectNamesConsistentWithProperties()` and now it is called separately and only once.

OpenSim/Simulation/Model/ControllerSet.h
```diff
@@ -86,7 +70,7 @@ OpenSim_DECLARE_CONCRETE_OBJECT(ControllerSet, ModelComponentSet<Controller>);
 
 private:
 
-    std::unique_ptr<Storage> _controlStore;
+    SimTK::ClonePtr<Storage> _controlStore;
```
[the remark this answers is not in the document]
aseth1: > If someone makes a copy of the model, should the control storage stay around? I'm not sure but I think not.

I agree in principle, but the _controlStorage is a result of the model's controllers, and not an input like other storages  like input kinematics and ground reaction forces. If for some reason you cloned the model before asking to write results, you would lose all the controls computed during a simulation (e.g. CMC, RRA). Since the storage is reset when you rerun a simulation, I didn't see the harm in keeping it around.  This could be a separate issues as well.

OpenSim/Simulation/Model/Ground.cpp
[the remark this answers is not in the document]
aseth1: 👍

OpenSim/Common/Object.cpp
```diff
@@ -1552,16 +1558,44 @@ makeObjectFromFile(const std::string &aFileName)
...
+    // If a single object property, assign the object's name as the property
+    // since renaming will cause the name to be inconsistent with what is 
+    // serialized (property name). We enforce consistency while the object is
+    // still editable and not during serialization when the model is const.
+    for (int i = 0; i < getNumProperties(); ++i) {
+        auto& prop = updPropertyByIndex(i);
+        // check if property is of type Object
+        if (prop.isObjectProperty()) {
+            // a property is a list so cycle through its contents
+            for (int j = 0; j < prop.size(); ++j) {
```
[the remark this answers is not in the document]
aseth1: Yes, because we are processing all single-object properties under this Object.

OpenSim/Common/Object.cpp
```diff
@@ -1552,16 +1558,44 @@ makeObjectFromFile(const std::string &aFileName)
...
+                Object& obj = prop.updValueAsObject(j);
+                if (!prop.isUnnamedProperty() && prop.isOneObjectProperty()) {
+                    obj.setName(prop.getName());
+                }
+                obj.setObjectIsUpToDateWithProperties();
```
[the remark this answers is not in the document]
aseth1: I agree, I've separated the updating of the object name from the setting of the flag up-to-date flag.

OpenSim/Simulation/Model/ElasticFoundationForce.cpp
```diff
@@ -63,10 +63,17 @@ void ElasticFoundationForce::extendAddToSystem(SimTK::MultibodySystem& system) c
...
-
+            // TODO: Dependency of ElasticFoundationForce on ContactGeometry
+            // should be handled by Sockets.
+            const ContactGeometry* contactGeom = nullptr;
+            if (getModel().hasComponent<ContactGeometry>(params.getGeometry()[j]))
+                contactGeom = &getModel().getComponent<ContactGeometry>(
+                    params.getGeometry()[j]);
+            else
+                contactGeom = &getModel().getComponent<ContactGeometry>(
+                    "./contactgeometryset/" + params.getGeometry()[j]);
```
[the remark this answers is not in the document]
aseth1: This is calling the model so it is effectively the absolute path.

OpenSim/Simulation/Model/Ground.cpp
aseth1: Yep this is point in 3. in  #1876

OpenSim/Analyses/InverseDynamics.cpp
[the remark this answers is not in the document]
chrisdembia: Okay thanks for explaining.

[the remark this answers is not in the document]
chrisdembia: This still might not be efficient if there are subcomponents, because the array of state variable names is not cached.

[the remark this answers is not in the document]
aseth1: > This still might not be efficient if there are subcomponents, because the array of state variable names is not cached.

Possibly, but this Analysis is no longer used to perform InverseDynamics. It is only here for the mass adjustment step of RRA. The whole name lookup every step is kind of absurd to begin with, but not worth fixing a virtually unused class.

aymanhab:
As promised here are GUI compilation errors as of now:
- Gui\opensim\view\src\org\opensim\view\experimentaldata\EditMotionObjectsPanel.java:431: error: no suitable constructor found for ExternalLoads(Model,String)
                ExternalLoads exLoads = new ExternalLoads(modelLocalCopy, fileName);
    constructor ExternalLoads.ExternalLoads(long,boolean) is not applicable
      (argument mismatch; Model cannot be converted to long)
    constructor ExternalLoads.ExternalLoads(String,boolean) is not applicable
      (argument mismatch; Model cannot be converted to String)
- Gui\opensim\view\src\org\opensim\view\experimentaldata\EditMotionObjectsPanel.java:433: error: cannot find symbol
                for(int i=0; i<exLoads.getSize(); i++){
  symbol:   method getSize()
  location: variable exLoads of type ExternalLoads
- Gui\opensim\view\src\org\opensim\view\experimentaldata\EditMotionObjectsPanel.java:434: error: cannot find symbol
                    ExternalForce ef = exLoads.get(i);
  symbol:   method get(int)
  location: variable exLoads of type ExternalLoads
- Gui\opensim\view\src\org\opensim\view\experimentaldata\EditMotionObjectsPanel.java:495: error: cannot find symbol
            exLoad.cloneAndAppend(extForce);
  symbol:   method cloneAndAppend(ExternalForce)
  location: variable exLoad of type ExternalLoads
- Gui\opensim\view\src\org\opensim\view\experimentaldata\EditMotionObjectsPanel.java:502: error: cannot find symbol
    exLoad.print(fullFilename);
  symbol:   method print(String)
  location: variable exLoad of type ExternalLoads
- Gui\opensim\view\src\org\opensim\view\experimentaldata\ForceListModel.java:53: error: cannot find symbol
        for (int i=0; i<fSet.getSize(); i++){
  symbol:   method getSize()
  location: variable fSet of type ExternalLoads
- Gui\opensim\view\src\org\opensim\view\experimentaldata\ForceListModel.java:54: error: cannot find symbol
            Force f= fSet.get(i);
  symbol:   method get(int)
  location: variable fSet of type ExternalLoads
- Gui\opensim\view\src\org\opensim\view\markerEditor\MarkersLoadFromFileAction.java:81: error: no suitable constructor found for MarkerSet(Model,String)
            newMarkerSet = new MarkerSet(model, fileName);
    constructor MarkerSet.MarkerSet(long,boolean) is not applicable
      (argument mismatch; Model cannot be converted to long)
    constructor MarkerSet.MarkerSet(String,boolean) is not applicable
      (argument mismatch; Model cannot be converted to String)
Note: Some input files additionally use or override a deprecated API.
Note: Some input files additionally use unchecked or unsafe operations.
Note: Some messages have been simplified; recompile with -Xdiags:verbose to get full output
8 errors

aseth1:
> As promised here are GUI compilation errors as of now:

As expected, since passing a `Model` to a `ModelComponent` constructor is not supported. You have to add the `ModelComponentSet` to the model of interest like any other `Component`.

You need to construct with set without the model, first. Then you have two options:
1. Simply call. `model.addModelComponent(modelComponentSet);`
2. Update the set that lives with the model you would were passing in, e.g.:
`model.updMarkerSet() = loadedMarkerSet()`.

aymanhab:
Thanks @aseth1 
- Using the upd syntax doesn't work in languages other than C++, neither does the = operator, so we're left with addModelComponent or addComponent, any reason to choose one over the other? What would be the "path" after addition? Other places in the code assumed all along that markers live under top level MarkerSet, would that need to change?
- The removed constructor didn't require an initSystem after the call, we'll see if this will be problematic here or not after the new addComponent call.
- The old constructor used to throw exception if reading from file failed, we need to restore that. Let me know if I need to do that (likely for all Sets).

While these issues are pending, I took a stab at them, made a branch and built the GUI to see the new names/paths. Some things worked out of the box while others broke (some visualization of muscles, some plots etc.)

Let's converge on the suggested solutions to the issues above then we'll have a build that we can all evaluate/assess.

Thanks for your help @aseth1

Applications/Forward/test/pendulum_no_gravity.osim
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <OpenSimDocument Version="20302">
-	<Model name="pendulum_no_gravity">
+	<Model name="pendulum">
 		<credits> Model authors names.. </credits>
```
aymanhab: Any reason to change the name here? Is it diffed in the test case?

Bindings/Java/Matlab/Hopper_Device/RunHopperWithDevice_answers.m
```diff
@@ -61,9 +61,9 @@
 anchorA = WeldJoint.safeDownCast(device.updComponent('anchorA'));
 anchorB = WeldJoint.safeDownCast(device.updComponent('anchorB'));
 thighAttach = PhysicalFrame.safeDownCast(...
-        hopper.getComponent('thigh/deviceAttach'));
+        hopper.getComponent('bodyset/thigh/deviceAttach'));
```
aymanhab: In a model created programmatically, is it "required" now to have bodyset in Pathname? Does it depend on how the Body is added to the model?

Bindings/Java/Matlab/Hopper_Device/RunHopperWithDevice_answers.m
```diff
@@ -119,9 +119,9 @@
 reporter.setName('hopper_device_results');
 reporter.set_report_time_interval(0.2); % seconds.
 reporter.addToReport(...
-    hopper.getComponent('slider/yCoord').getOutput('value'), 'height');
+    hopper.getComponent('jointset/slider/yCoord').getOutput('value'), 'height');
 reporter.addToReport(...
-    hopper.getComponent('vastus').getOutput('activation'), 'vastus_activation');
+    hopper.getComponent('forceset/vastus').getOutput('activation'), 'vastus_activation');
```
aymanhab: Same question for all sets, is that just a side effect of using an old model file that introduces the sets or is the layout now required? Thanks

Bindings/Python/tests/test_access_subcomponents.py
```diff
@@ -13,11 +13,11 @@
 class TestAccessSubcomponents(unittest.TestCase):
     def test_individual_components(self):
         model = osim.Model(os.path.join(test_dir, "arm26.osim"))
-        muscle = model.getComponent('BICshort')
+        muscle = model.getComponent('forceset/BICshort')
         assert muscle.getName() == 'BICshort'
         # No downcasting necessary!
         muscle.get_max_isometric_force() # Method on Muscle.
-        muscle = model.updComponent('BICshort')
+        muscle = model.updComponent('forceset/BICshort')
```
aymanhab: Is it true that "ForceSet" is a component now and contributes to component path but MuscleSet isn't (because it doesn't own) is the implication that all Force names need to be unique regardless of type?

OpenSim/Actuators/Test/testActuators.cpp
```diff
@@ -522,6 +449,82 @@ void testClutchedPathSpring()
...
+        double theoretical = (pressure / (4 * pow(num_turns, 2) * SimTK::Pi)) * (3 * pow(pos(0), 2) - pow(B, 2));
+
+        ASSERT_EQUAL(applied, theoretical, 10.0);
+    }
+
+
+    std::cout << " ******** Test McKibbenActuator time = ********" <<
+        1.e3*(std::clock() - startTime) / CLOCKS_PER_SEC << "ms\n" << endl;
+}
+
```
aymanhab: github is not showing this side by side, was there material change other than moving text around?

OpenSim/Analyses/JointReaction.cpp
```diff
@@ -310,8 +310,16 @@ void JointReaction::setupReactionList()
...
-                currentKey.expressedInFrame = &_model->getComponent<Frame>(expressedIn);
+                currentKey.expressedInFrame =
+                    &_model->getComponent<Frame>(expressedIn);
+            else if (_model->getBodySet().hasComponent<Frame>(expressedIn))
+                currentKey.expressedInFrame =
+                    &_model->getBodySet().getComponent<Frame>(expressedIn);
+            else if (_model->getJointSet().hasComponent<Frame>(expressedIn))
+                currentKey.expressedInFrame =
+                    &_model->getJointSet().getComponent<Frame>(expressedIn);
             else {
```
aymanhab: Why was this necessary? Was it a bug fixed along the way?

Bindings/Java/Matlab/Hopper_Device/RunHopperWithDevice_answers.m
aseth1: Not required. Just consequence of previous format and if using `Model::add##ComponentType##()`

Bindings/Python/tests/test_access_subcomponents.py
aseth1: Yes, `ForceSet` is a Component and owns the `Force`s it contains. That ownership has not changed. Yes you cannot have two forces of the same name regardless of their subtype.

OpenSim/Actuators/Test/testActuators.cpp
aseth1: no change.

aseth1:
> @aseth1 GUI branch is make_set_as_component you may need to manually delete a stale file that fails to compile, then you should be good to go.

Thanks. Building now.

aymanhab:
GUI Branch has been updated and builds now. Sorry for the delay @aseth1

chrisdembia:
Please let me know when this is ready for review again.

aseth1:
@aymanhab GUI built for me too. I will give it a whirl.

aseth1:
@chrisdembia I've addressed most of your feedback by making changes except for updating paths in `updateFromXMLNode` (your point 3.) for the reason that these connections should be managed by Sockets. That can be a subsequent change.

Bindings/Java/Matlab/Hopper_Device/RunHopperWithDevice_answers.m
aymanhab: Thanks @aseth1 Do you think it would be confusing to users to have different paths depending on what call is used to add (e.g. addMarker, addComponent, addModelComponent)?

OpenSim/Actuators/TorqueActuator.cpp
```diff
@@ -191,13 +191,16 @@ void TorqueActuator::extendConnectToModel(Model& model)
...
-    // corresponding body objects.
-    _bodyA =
-        static_cast<const PhysicalFrame*>(&getModel().getComponent(get_bodyA()));
-    _bodyB =
-        static_cast<const PhysicalFrame*>(&getModel().getComponent(get_bodyB()));
-
+    // TODO: Replace this custom lookup with Sockets
+    if(getModel().hasComponent<PhysicalFrame>(get_bodyA()))
+        _bodyA = &getModel().getComponent<PhysicalFrame>(get_bodyA());
+    else
```
aymanhab: The sequence of hasComponent<> then getComponent<> now appears in many places, it likely does the work twice, can these be combined for mentainance and efficiency?

aymanhab: Also I'm not sure if the first hasComponent<> would succeed even if the Body is under  bodyset, or does it search only immediate children?

OpenSim/Analyses/JointReaction.cpp
aymanhab: Thanks @aseth1 Would it make sense to move this sequence to one function to locate Frame then so that this logic doesn't need to be repeated elsewhere and so it can be updated if Frames can be located in other places?

OpenSim/Common/Component.cpp
```diff
@@ -780,53 +782,8 @@ Array<std::string> Component::getStateVariableNames() const
...
-        const std::string& subCompName = _adoptedSubcomponents[i]->getName();
-        std::string::size_type front = subCompName.find_first_not_of(" \t\r\n");
-        std::string::size_type back = subCompName.find_last_not_of(" \t\r\n");
-        std::string prefix = "";
-        if (back >= front) // have non-whitespace name
-            prefix = subCompName + "/";
-        for (int j = 0; j<nsubs; ++j) {
-            names.append(prefix + subnames[j]);
-        }
-    }
```
aymanhab: Can you explain how the new block (that used to be commented out) accomplishes the function of the removed one? It sure looks cleaner.

OpenSim/Common/Object.cpp
```diff
@@ -1017,44 +1016,52 @@ updateDefaultObjectsFromXMLNode()
...
+        XMLNode::RemoveChildren(_refNode); // Shouldn't have any children in the reference document*/
         }
         return;
     }
     
     // GENERATE XML NODE for object
     SimTK::Xml::Element myObjectElement(getConcreteClassName());
-    if (!getName().empty())
+    
+    // if property is provided and it is not of unnamed type, use the property name
```
aymanhab: Does this handle Deprecated properties as well since these were implicated in the previous hack?

OpenSim/Common/Set.h
```diff
@@ -100,22 +93,21 @@ Set() :
  * @param aFileName             Name of the file.
  * @param aUpdateFromXMLNode    Whether to update from XML.
  */
-Set(const std::string &aFileName, bool aUpdateFromXMLNode = true) :
-    Object(aFileName),
+Set(const std::string &aFileName, bool aUpdateFromXMLNode = true) : Super(aFileName),
```
aymanhab: We need to annotate this with SWIG_DECLARE_EXCEPTION so that file loading failures are caught in proper context.

OpenSim/Examples/ExampleHopperDevice/exampleHopperDevice.cpp
```diff
@@ -212,14 +212,15 @@ void run(bool showVisualizer, double finalTime)
     {
         // Build the hopper.
         auto hopper = buildHopper(showVisualizer);
+        hopper.finalizeFromProperties();
```
aymanhab: Why was this added? It would be great if the internal steps of the construction/initializaton are not exposed and users only need to know initSystem.

OpenSim/Simulation/Model/AbstractTool.cpp
```diff
@@ -553,22 +553,31 @@ printResults(const string &aBaseName,const string &aDir,double aDT,
...
     // as relative paths
     std::string savedCwd = IO::getCwd();
     IO::chDir(IO::getParentDirectory(aExternalLoadsFileName));
     // Create external forces
+    ExternalLoads* externalLoads = nullptr;
     try {
-        _externalLoads = ExternalLoads(aModel, aExternalLoadsFileName);
+        externalLoads = new ExternalLoads(aExternalLoadsFileName, true);
+        copyModel.addModelComponent(externalLoads);
     }
```
aymanhab: This will put ExternalForce(s) in the miscModelComponents bag, does that have side effects downstream?

aseth1:
I was able to load the arm26.osim model, I got a crash when navigating. I did a clean build and then when I try to load the mode I get:

```java.lang.NullPointerException
	at org.opensim.threejs.ModelVisualizationJson.createComputedPathPoints(ModelVisualizationJson.java:1194)
	at org.opensim.threejs.ModelVisualizationJson.createJsonForGeometryPath(ModelVisualizationJson.java:1614)
	at org.opensim.threejs.ModelVisualizationJson.processDecorationsForComponent(ModelVisualizationJson.java:500)
	at org.opensim.threejs.ModelVisualizationJson.createJsonForModel(ModelVisualizationJson.java:460)
	at org.opensim.threejs.ModelVisualizationJson.<init>(ModelVisualizationJson.java:420)
	at org.opensim.view.pub.ViewDB.getJsonForModel(ViewDB.java:1857)
	at org.opensim.view.pub.ViewDB.update(ViewDB.java:492)
	at java.util.Observable.notifyObservers(Observable.java:159)
	at org.opensim.view.pub.OpenSimDB.addModel(OpenSimDB.java:155)
	at org.opensim.view.pub.OpenSimDB.addModel(OpenSimDB.java:143)
	at org.opensim.view.FileOpenOsimModelAction$2.run(FileOpenOsimModelAction.java:100)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:756)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:80)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:726)
	at org.netbeans.core.TimableEventQueue.dispatchEvent(TimableEventQueue.java:159)
[catch] at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:201)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:105)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:101)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:93)
	at java.awt.EventDispatchThread.run(EventDispatchThread.java:82)
```
The model is loaded in the navigator but no model appears. It appears to be related to muscle path points. To verify, I can load the *bouncing_block.osim* file

OpenSim/Actuators/TorqueActuator.cpp
aseth1: > The sequence of hasComponent<> then getComponent<> now appears in many places, it likely does the work twice, can these be combined for mentainance and efficiency?

The code is not duplicated. They both share the same worker method, `traversePathToComponent()`. I am inclined not to expose  `traversePathToComponent()` unless we can show that the current interface has a performance issue.

Applications/Forward/test/pendulum_no_gravity.osim
aseth1: The model name appears in the column labels and we are trying to compare two results one with gravity and one without with gravity replaced by an external load.

Bindings/Java/Matlab/Hopper_Device/RunHopperWithDevice_answers.m
aseth1: Yes if you add a body to the Model's `BodySet` you have to include *bodyset* in the path. to retrieve it. That was the point of the change to distinguish components added to different Sets by recognizing the Set as part of the pathname.

Bindings/Java/Matlab/Hopper_Device/RunHopperWithDevice_answers.m
aseth1: It is a bit late to be asking this question. We agreed to maintain the Sets primarily to support navigation in the GUI. The only reason to have these Sets is for the GUI and the only reason `Model::add##ComponentType##()` are necesssary is to populate these sets so the added components appear in the GUI.

If you created Sets on the fly using `getComponentList<C>` then we wouldn't need them on the API side and the change to make Sets ModelComponents so they appear on the pathname would not be necessary. That being said, the transition from 3.3 to 4.0 would be more difficult with non unique names.

OpenSim/Common/Component.cpp
aseth1: The iterator traverses the three separately maintained private list of subcomponents. Teh comment was added before we had the iterator working efficiently and should have been updated long ago. I noticed it while making this change and decided to finish it.

OpenSim/Common/Object.cpp
aseth1: Yes.

OpenSim/Common/Set.h
aseth1: Good catch! Please feel free to add it. One good thing is that it is now centralized instead distributed amongst the concrete objects. Could this not be pushed even further down to Object itself? We could modify the exception to:
1. be more specific (e.g. FileNotFound), and 
2. include the concrete class name so the message is more meaningful.

OpenSim/Examples/ExampleHopperDevice/exampleHopperDevice.cpp
aseth1: Looks like it was added while I was debugging. (See `exampleHopperDevice_answers.cpp` below was not modified). I will remove.

OpenSim/Simulation/Model/AbstractTool.cpp
aseth1: I would not call it a side-effect, but one might assume that `ExternalForces` live in  the model's `ForceSet`, which was what IAA was doing. The fix is simple and we need to move away from assuming the organization of the Model and rely on/use  `getComponenList<C>`to obtain all components of certain type.

OpenSim/Common/Set.h
aseth1: I'll add this now.

OpenSim/Simulation/Model/AbstractTool.cpp
```diff
@@ -553,22 +553,31 @@ printResults(const string &aBaseName,const string &aDir,double aDT,
...
+bool AbstractTool::createExternalLoads( const string& aExternalLoadsFileName,
+                                        Model& aModel, const Storage *loadKinematics)
 {
     if(aExternalLoadsFileName==""||aExternalLoadsFileName=="Unassigned") {
         cout<<"No external loads will be applied (external loads file not specified)."<<endl;
         return false;
     }
 
+    Model copyModel = aModel;
+    // speedup realize position calculations by removing all force elements
```
aymanhab: I'm not sure I understand the downstream effects of this change, instead of the ExternalLoads added to passed in Model as suggested by the signature, now a copy of the Model is made and the  passed in Model is unchanged, what's the rationale?

OpenSim/Simulation/Model/AbstractTool.cpp
```diff
@@ -597,48 +604,48 @@ bool AbstractTool::createExternalLoads( const string& aExternalLoadsFileName, Mo
...
-    
-    // Add external loads to the set of all model forces
-    for(int i=0; i<_externalLoads.getSize(); ++i){
-        aModel.addForce(&_externalLoads[i]);
     }
 
+    //Now add the ExternalLoads (transformed or not) to the Model to be analyzed
+    ExternalLoads* exLoadsClone = externalLoads->clone();
+    aModel.addModelComponent(exLoadsClone);
+
```
aymanhab: NVM, now added to the passed in aModel but in the ModelComponents bin!

OpenSim/Simulation/Model/BodySet.cpp
```diff
@@ -22,71 +22,4 @@
  * -------------------------------------------------------------------------- */
 
 #include "BodySet.h"
-#include <OpenSim/Common/ScaleSet.h>
```
aymanhab: Should this file be removed?

OpenSim/Analyses/JointReaction.cpp
aseth1: I don't think it is worth making more changes to this Analysis. There is more control using Outputs, which Joint has for reaction forces are part of. The change preserves the functionality of the `JointReaction` Analysis.

OpenSim/Simulation/Model/AbstractTool.cpp
aseth1: I separate to things that were going on with the passed in model.
1. It was being used to compute the transformation of the force into the body in which it was being applied.
2. It was being augmented with ExternalLoads.

Doing both these operations on the same Model means you have to remove the first ExternalLoads that was added for the purpose of computing transformed points.

Since we only really care about getting the passed in model back with the correct ExternalLoads, it seems safer and more straightforward to use a copy to be the workhorse for doing the transformation. In that case, we might as strip it down to its bare bones to do the transformation as quickly as possible. Forces are completely unnecessary for the realization of position transforms.

Once we are happy with the ExternalLoads we then add it the model we care about that has remained untouched.

OpenSim/Simulation/Model/AbstractTool.cpp
aseth1: NVM? We have a ModleComponentSet that lives on the model. We can get rid of it. It just means that Components (e.g. the ExternalLoads) added via addComponent() do not appear in the GUI.

OpenSim/Simulation/Model/BodySet.cpp
aseth1: Yes, I think all the `<C>Set` files can safely be removed now.

OpenSim/Simulation/Model/ComponentSet.h
```diff
@@ -43,15 +42,10 @@ OpenSim_DECLARE_CONCRETE_OBJECT(ComponentSet,
                                 ModelComponentSet<ModelComponent>);
 
 public:
-    ComponentSet() : Super() {}
+    /** Use Super's constructors. @see ModelComponentSet */
```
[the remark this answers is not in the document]
aseth1: Its a bad name and I believe predates `Component` having a list of subcomponents. We considered `MiscellaneousComponentSet` and `MiscellaneousModelComponentSet`. `OtherModelComponents` or `UncategorizedModelComponentSet` could also work. Preferences?

Bindings/Java/Matlab/Hopper_Device/RunHopperWithDevice_answers.m
aymanhab: @aseth1 For the record, the GUI does NOT use the Sets in the model directly, instead it uses iterators to populate a cached "Set" and that's what's used to back the navigator.

aseth1: Awesome. So we can remove all the Sets from `Model` and not have to support the multiple `add<ModelComponent>()` methods.  Sweet. That would simplify the `Model` interface a lot. If this is the case that makes this PR rather unnecessary, right?

OpenSim/Simulation/Model/ElasticFoundationForce.cpp
```diff
@@ -63,10 +63,17 @@ void ElasticFoundationForce::extendAddToSystem(SimTK::MultibodySystem& system) c
...
+
+            const ContactGeometry& geom = *contactGeom;
             // B: base Frame (Body or Ground)
```
aymanhab: If there was a typo or problem upstream wouldn't this blow up dereferencing nullptr?

aymanhab: or did we guarantee earlier that contactGeom cannot be nullptr?

OpenSim/Simulation/Model/ExpressionBasedPointToPointForce.cpp
```diff
@@ -117,10 +117,17 @@ void ExpressionBasedPointToPointForce::extendConnectToModel(Model& model)
...
-    _body1 =
-        static_cast<const PhysicalFrame*>(&getModel().getComponent(body1Name));
-    _body2 =
-        static_cast<const PhysicalFrame*>(&getModel().getComponent(body2Name));
+    if(getModel().hasComponent(body1Name))
+        _body1 = &(getModel().getComponent<PhysicalFrame>(body1Name));
+    else
+        _body1 = &(getModel().getComponent<PhysicalFrame>(
+            "./bodyset/" + body1Name));
+
```
aymanhab: At connect stage, are we guaranteed that either of the two pathways will succeed and we're not left with nullptr?

OpenSim/Simulation/Model/ForceSet.cpp
```diff
@@ -313,11 +209,11 @@ Set<Actuator>& ForceSet::updActuators()
...
 void ForceSet::updateActuators()
 {
+    _actuators.setMemoryOwner(false);
     _actuators.setSize(0);
-    for (int i = 0; i < getSize(); ++i)
-    {
+    for (int i = 0; i < getSize(); ++i) {
         Actuator* act = dynamic_cast<Actuator*>(&get(i));
-        if (act != NULL)  _actuators.adoptAndAppend(act);
+        if (act)  _actuators.adoptAndAppend(act);
```
aymanhab: I think some compilers don't like using pointer as a boolean?

OpenSim/Simulation/Model/ForceSet.cpp
```diff
@@ -328,11 +224,14 @@ void ForceSet::updateActuators()
  */
 const Set<Muscle>& ForceSet::getMuscles() const
 {
+    if (_muscles.getSize() == 0)
+        const_cast<ForceSet*>(this)->updateMuscles();
     return _muscles;
```
aymanhab: Why is this necessary now? aren't muscles kept in sync forces all along?

OpenSim/Simulation/SimbodyEngine/Coordinate.h
```diff
@@ -330,7 +330,7 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Coordinate, ModelComponent);
     SimTK::ReferencePtr<const Joint> _joint;
 
     /* User set MotionType from versions of OpenSim that predate 4.0 */
-    MotionType _userSpecifiedMotionTypePriorTo40;
+    MotionType _userSpecifiedMotionTypePriorTo40{ Undefined };
```
aymanhab: @aseth1 is this related to the undefined motion type reported by Colin earlier? does it fix it, or try to address it?

OpenSim/Tools/StateTrackingTask.h
```diff
@@ -83,14 +83,15 @@ class OSIMTOOLS_API StateTrackingTask : public TrackingTask
...
         if(dix != std::string::npos){
             std::string varName = getName();
             varName.replace(dix, 1, "/");
-            val = _model->getStateVariableValue(s, varName);
+            val = forceSet.getStateVariableValue(s, varName);
         }
         else{
-            val = _model->getStateVariableValue(s, getName());
+            val = forceSet.getStateVariableValue(s, getName());
         }
```
aymanhab: Why is this being limited from model to forceSet? Some context?

Bindings/Java/Matlab/Hopper_Device/RunHopperWithDevice_answers.m
aseth1: If we removed `Model`'s `Set`s then we would have to deal with renaming in the case of duplicates, yes, but this PR wouldn't make make sense since we would not have `BodySet`, `ForceSet`, .... As it stands, we still have a mix of old (`addBody`, `addJoint`, ...) and new `addComponent()`, which as you point out is confusing.

OpenSim/Simulation/Model/ElasticFoundationForce.cpp
aseth1: The behavior is identical to before the PR. We do not dereference a nullptr because `getComponent()` throws if the component is not returned.

OpenSim/Simulation/Model/ExpressionBasedPointToPointForce.cpp
aseth1: No `getComponent` throws if no component.

OpenSim/Simulation/Model/ForceSet.cpp
aseth1: Wouldn't it cause a compiler warning if that if that were true? In that case we have to do a sweep of the entire codebase, because I think there are several other instances.

OpenSim/Simulation/SimbodyEngine/Coordinate.h
aseth1: I thought it might have been but it was not the cause. Since we have `Undefined` as a value, I think it is cleaner to initialize with it.

OpenSim/Common/Object.h
```diff
@@ -554,6 +556,13 @@ class OSIMCOMMON_API Object
         _objectIsUpToDate = false;
     }
 
+    /** To be up-to-date with properties we must make sure the name of an object,
+    which can be changed independent of the property name, remains consistent
+    with any restrictions specified by the Property. For example, unnamed
+    properties should not have a name. Furthermore, named properties whose object
+    has no name, should have the property name. **/
+    void makeObjectNamesConsistentWithProperties();
```
chrisdembia: I think we should document that this function is recursive (it visits each node of the ownership tree).

OpenSim/Common/Component.cpp
```diff
@@ -157,6 +157,9 @@ void Component::finalizeFromProperties()
 {
     reset();
 
+    // last opportunity to modify Object names based on properties
+    makeObjectNamesConsistentWithProperties();
```
chrisdembia: It seems to me that we are running this function multiple times per object (note that `componentsFinalizeFromProperties()` descends the ownership tree).

OpenSim/Common/Object.h
```diff
@@ -518,9 +518,13 @@ class OSIMCOMMON_API Object
...
-        are ever generated and the parent node is not used. 
-    **/
-    virtual void updateXMLNode(SimTK::Xml::Element& parent) const;
+        are ever generated and the parent node is not used.
+    @param      prop (optional)
+        The pointer to the property that contains this object. If it is
+        present, check if the property is unnamed and if NOT, use the property
+        name as its name when updating the XML node. **/
+    void updateXMLNode(SimTK::Xml::Element& parent,
+                       const AbstractProperty* prop=nullptr) const;
```
chrisdembia: I forget if I already asked this...is `prop` always passed? If so, I think we should make it a required argument. Otherwise, we could get different deserializations based on whether or not this parameter is passed.

OpenSim/Simulation/Model/AbstractTool.cpp
```diff
@@ -553,22 +553,31 @@ printResults(const string &aBaseName,const string &aDir,double aDT,
...
-bool AbstractTool::createExternalLoads( const string& aExternalLoadsFileName, Model& aModel, const Storage *loadKinematics)
+bool AbstractTool::createExternalLoads( const string& aExternalLoadsFileName,
+                                        Model& aModel, const Storage *loadKinematics)
 {
     if(aExternalLoadsFileName==""||aExternalLoadsFileName=="Unassigned") {
         cout<<"No external loads will be applied (external loads file not specified)."<<endl;
         return false;
     }
 
+    Model copyModel = aModel;
```
chrisdembia: I prefer `modelCopy`, used elsewhere throughout the codebase. `copyModel` sounds like an action, better suited for a function name.

aymanhab:
@aseth1 Where are we keeping track of the issues with the GUI built with this PR and corresponding GUI branch, I have a few items to add and results of some investigations I made that need discussion. Thanks

OpenSim/Simulation/Model/AbstractTool.cpp
```diff
@@ -597,48 +604,48 @@ bool AbstractTool::createExternalLoads( const string& aExternalLoadsFileName, Mo
...
-    for(int i=0; i<_externalLoads.getSize(); ++i){
-        aModel.addForce(&_externalLoads[i]);
     }
 
+    //Now add the ExternalLoads (transformed or not) to the Model to be analyzed
+    ExternalLoads* exLoadsClone = externalLoads->clone();
+    aModel.addModelComponent(exLoadsClone);
+
+    // copy over created external loads to the external loads owned by the tool
+    _externalLoads = *externalLoads;
```
chrisdembia: I'm confused why this is necessary. Why is there an `ExternalLoads` in both the tool and the model?

chrisdembia: If this is how the code was before, then it's fine I suppose.

OpenSim/Simulation/Model/ExpressionBasedPointToPointForce.cpp
chrisdembia: `getComponent()` throws an exception if it can't find the component.

OpenSim/Simulation/Model/ExternalLoads.cpp
```diff
@@ -132,6 +116,7 @@ void ExternalLoads::copyData(const ExternalLoads &aAbsExternalLoads)
     _dataFileName = aAbsExternalLoads._dataFileName;
     _externalLoadsModelKinematicsFileName = aAbsExternalLoads._externalLoadsModelKinematicsFileName;
     _lowpassCutoffFrequencyForLoadKinematics = aAbsExternalLoads._lowpassCutoffFrequencyForLoadKinematics;
+    _storages = aAbsExternalLoads._storages;
```
chrisdembia: Because you are using `shared_ptr`, this is a shallow copy.

OpenSim/Simulation/Model/ExternalLoads.h
```diff
@@ -84,7 +84,7 @@ OpenSim_DECLARE_CONCRETE_OBJECT(ExternalLoads,
     /* If point of applications for external forces must be re-expressed
        then build new storages to be assigned to the individual ExternalForces
        with the transformed point data. Hang-on to them so we can delete them. */
-    ArrayPtrs<Storage> _storages;
+    std::vector<std::shared_ptr<Storage>> _storages;
```
chrisdembia: I think this should be `SimTK::ClonePtr` or `SimTK::ResetOnCopy`.

OpenSim/Simulation/Model/ForceSet.cpp
chrisdembia: We do this extensively throughout the code base.

OpenSim/Simulation/Model/ForceSet.cpp
chrisdembia: I agree, we should avoid this. The sets should be kept in sync when they are modified, not when they are accessed. We could add a virtual `extendAppend...()` function if necessary to allow derived classes to do postprocessing.

scrum2a:
@aymanhab suggests making the `GeometryPath` `PathPointSet` an Object set, not a Model Component Set.

OpenSim/Simulation/Model/ExternalLoads.h
aseth1: I used `shared_ptr` because I didn't want to re-specify the data source after copying or pay the price of copying potentially large data sets.  It this case it is the `ExternalLoads` itself that is creating the new data source based on transforming the point of force application. This should be an operation done outside of `ExternalLoads` and the data source reset. It could then be easily reassigned to a clone, but as it stands now, the clone is added to the `Model` and it would require more widespread changes that I thought would be going even further beyond the scope of this PR. This could be addressed as part of a more general `ExternalLoads` refactor. As it stands, I don't think there is an issue of leaking.

aseth1:
___
*[OpenSim/Tools/StateTrackingTask.h, line 86 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2285#-LNlHsDsDF2qKSRY1_eP:-LNlpBXF3qPLLUzqfWo-:bgqiuxh) ([raw file](https://github.com/opensim-org/opensim-core/blob/60ed6894d1d20b79580d7cfeccf2a18c667c0b25/OpenSim/Tools/StateTrackingTask.h#L86)):*

Do you know for sure that only force states are tracked? Seems like we should have `updateFromXMLNode()` that changes `.` to `/` rather than doing this in code.
</blockquote></details>

The only example is muscle activation tracking. I also did this to prevent trying to track joint angles and speeds with states tracking. If the states tracking worked in general (not clear it works for activation), I would be more inclined to open it up, but I think that requires more testing and development. I don't feel too strongly about it either and could revert to use any state variable, I just don't think it will work.

OpenSim/Common/Object.cpp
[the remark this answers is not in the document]
chrisdembia: If possible, I think it would be better to always supply the property and check if it's single-object within this function.

aseth1:
> have you assessed if the runtime of Model::initSystem() has changed?

```cpp
void measureInitSystemInALoop()
{
    Model model("arm26.osim");
    // Test initSystem in a loop
    for (int i = 0; i < 100; ++i) {
        model.initSystem();
        model.invalidateSystem();
    }
}

void measureFinalizePropertiesInALoop()
{
    Model model("arm26.osim");
    // Test initSystem in a loop
    for (int i = 0; i < 100; ++i) {
        model.finalizeFromProperties();
    }
}

void measureCloneInitSystemInALoop()
{
    Model model("arm26.osim");
    // Test initSystem in a loop
    Model* cloneModel = nullptr;
    for (int i = 0; i < 100; ++i) {
        cloneModel = model.clone();
        cloneModel->initSystem();
        delete cloneModel;
    }
}

```
**master:**
 ```
measureInitSystemInALoop done. real/CPU ms: 1738.2 / 1734.4
measureFinalizePropertiesInALoop done. real/CPU ms: 587.5 / 593.8
measureCloneInitSystemInALoop done. real/CPU ms: 2474.8 / 2484.4
```

**PR:**
```
measureInitSystemInALoop done. real/CPU ms: 1784.4 / 1781.3
measureFinalizePropertiesInALoop done. real/CPU ms: 590.2 / 593.8
measureCloneInitSystemInALoop done. real/CPU ms: 2816.3 / 2812.5
```

We have definitely made `initSystem()` slower and I would have chalked that up to  `finalizeFromProperties()` but does not seem to be the case.

**Comment by aseth1** [comment]
___
*[OpenSim/Simulation/Model/ExternalLoads.h, line 87 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2285#-LNnPVZ51H6tgZqYKbK5:-LNqrig85NZEhPn2SBXV:b3bw2ks) ([raw file](https://github.com/opensim-org/opensim-core/blob/60ed6894d1d20b79580d7cfeccf2a18c667c0b25/OpenSim/Simulation/Model/ExternalLoads.h#L87)):*
I agree that it isn't ideal, but it isn't completely unexpected that a clone of `ExternalLoads` would refer to the same data. The main issue is:
 
> I didn't want to re-specify the data source after copying or pay the price of copying potentially large data sets. In this case it is the ExternalLoads itself that is creating the new data source based on transforming the point of force application. This should be an operation done outside of ExternalLoads and the data source reset. It could then be easily reassigned to a clone, but as it stands now, the clone is added to the Model and it would require more widespread changes that I thought would be going even further beyond the scope of this PR.

**Comment by aseth1** [comment]
___
*[OpenSim/Simulation/Model/ForceSet.cpp, line 228 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2285#-LNnQ0f34kPkh-PtI1Z3:-LNqtc-dB0uVrZ05pCFr:bmco4sn) ([raw file](https://github.com/opensim-org/opensim-core/blob/60ed6894d1d20b79580d7cfeccf2a18c667c0b25/OpenSim/Simulation/Model/ForceSet.cpp#L228)):*

I commented on GitHub that updating should occur when the set is modified, not accessed. I know that you didn't introduce the problem, but it's not great to make this change in a const function.
</blockquote></details>

I will try undoing it, but your approach of updating upon edit of the model is not straightforward since there are many ways to edit a `Set` (`append, set(i), remove, ...`) which means code duplication and potentially missing one, which makes the approach potentially more error prone.

File: OpenSim/Common/Component.cpp (hunk 34 above)
Comment (aseth1) [reply OpenSim/Common/Component.cpp]: Yes, but not all Objects are Components. A couple cases that fail if we don't traverse all objects are use of `Function` and `SpatialTransform`.

File: OpenSim/Common/Component.cpp [hunk 40]
```diff
@@ -765,9 +768,8 @@ Array<std::string> Component::getStateVariableNames() const
     // Must have already called initSystem.
     OPENSIM_THROW_IF_FRMOBJ(!hasSystem(), ComponentHasNoSystem);
 
-    Array<std::string> names = getStateVariablesNamesAddedByComponent();
+    Array<std::string> stateNames = getStateVariablesNamesAddedByComponent();
```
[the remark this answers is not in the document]
aseth1: :+1:

OpenSim/Common/Object.h
aseth1: No it is not always passed- it is optional.  You previously asked to make it clear that it was optional. It is used exclusively to replace the KLUDGE in `AbstractProperty` that was renaming a `const Object`.

aseth1:
___
*[OpenSim/Tests/Components/testComponents.cpp, line 184 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2285#-LNnK18vC177K3T7RCtT:-LNrSGhvDh5HBrG69CR0:b-9byrlx) ([raw file](https://github.com/opensim-org/opensim-core/blob/60ed6894d1d20b79580d7cfeccf2a18c667c0b25/OpenSim/Tests/Components/testComponents.cpp#L184)):*

Is it true that the more specific reason for this call is to correct object names?
</blockquote></details>

Not exactly. `finalizeFromProperties` validates properties and the naming of `Object` properties is part of that validation task.

OpenSim/Common/Component.cpp
chrisdembia: Ah okay. Well we really should find a way to avoid these repeated calls. Can you have only Model invoke this function in its `extendFinalizeFromProperties()`?

OpenSim/Common/Object.h
chrisdembia: Thanks for clarifying. It's been difficult for me to keep track.

aseth1:
@aymanhab 
> The following tests FAILED:
>	 64 - testIterators (Failed)
>	 92 - python_tests (Failed)

aymanhab:
> @aymanhab
> 
> > The following tests FAILED:
> > 64 - testIterators (Failed)
> > 92 - python_tests (Failed)

I'll have to build this environment debuggable to find out why. Likely reverting PathPointSet to Set of Objects rather than ModelComponentSet messed up the count.

aymanhab:
@aseth1 The test has 5 subtests, it fails on the first one on count mismatch. If it's clear to you how to fix the numbers please advise, expected numbers are in the code indeed, actual numbers now are unknown AFAICT.

aseth1:
@aymanhab you have to make a similar fix to `test_component_interface.py`

aymanhab:
I fixed testIterators already, will now do  test_component_interface.py

aymanhab:
@chrisdembia does the branch builds automatically on commit? I committed a fix to the test cases (testIterators and python_tests) but don't see the build log.

aseth1:
@chrisdembia @aymanhab please approve if your comments have been adequately addressed.

chrisdembia:
I just pushed the fix to `testExternalLoads`. It seems that `testCMCGait10dof18musc` might be failing now.

aseth1:
@chrisdembia your changes LGTM and all test are passing. Thanks!

aseth1:
@aymanhab all tests are passing and issues should have been addressed.

aymanhab:
@aseth1 I'll take another pass thru my comments/responses since I didn't use reviewable.
Would be good to update the list of broken items on https://github.com/opensim-org/opensim-gui/issues/1008 so we can get a good picture of where we stand.

Bindings/Java/Matlab/Hopper_Device/RunHopperWithDevice_answers.m
aymanhab: This should be documented somewhere since it is not obvious that addBody(body) and addComponent(body) and addModelComponent(body) would result in 3 different layouts with different pathnames.

Bindings/Java/Matlab/Hopper_Device/RunHopperWithDevice_answers.m
```diff
@@ -78,7 +78,7 @@
 % (Done for you) Configure the device to wrap over the patella.
 if hopper.hasComponent('device_active') || hopper.hasComponent('device_passive') 
     cable = PathActuator.safeDownCast(device.updComponent('cableAtoBactive'));
-    patellaPath = 'thigh/patellaFrame/patella';
+    patellaPath = 'bodyset/thigh/patellaFrame/wrapobjectset/patella';
     wrapObject = WrapCylinder.safeDownCast(hopper.updComponent(patellaPath));
```
aymanhab: This maybe late but wrapobjectset feels just an extraneous level, similar to pathpointset (that I reverted to regular rather than ModelComponentSet), should we try that in a separate issue, or is there a use case where it's needed?

Bindings/simulation.i
```diff
@@ -330,3 +335,16 @@ EXPOSE_JOINT_CONSTRUCTORS_HELPER(PlanarJoint);
...
+EXPOSE_SET_CONSTRUCTORS_HELPER(JointSet);
+EXPOSE_SET_CONSTRUCTORS_HELPER(ConstraintSet);
+EXPOSE_SET_CONSTRUCTORS_HELPER(ForcesSet);
+EXPOSE_SET_CONSTRUCTORS_HELPER(ControllerSet);
+EXPOSE_SET_CONSTRUCTORS_HELPER(ContactGeometrySet);
+EXPOSE_SET_CONSTRUCTORS_HELPER(PathPointSet);
+EXPOSE_SET_CONSTRUCTORS_HELPER(ProbeSet);
+EXPOSE_SET_CONSTRUCTORS_HELPER(MarkerSet);
+EXPOSE_SET_CONSTRUCTORS_HELPER(WrapObjectSet);
+EXPOSE_SET_CONSTRUCTORS_HELPER(CoordinateSet);
```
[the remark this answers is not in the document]
aymanhab: Agreed @chrisdembia 👍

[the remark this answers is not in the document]
aseth1: I am of the opposite opinion. Seeing the list gives me a quick view of what IS wrapped. If they are scattered you could easily add a new set and forget. Here you see all the other Sets being handled. If you both feel it is a necessary change, I'll do it.

Bindings/Java/Matlab/Hopper_Device/RunHopperWithDevice_answers.m
aseth1: `Model` has documentation in the section for adding components to the model. I have added some addition info to indicate the pathnames are different.
```cpp
    /** @name Adding components to the Model
     * Model takes ownership of the ModelComponents and adds them to specialized
     * (typed) Sets within the model. These sets can be useful for uniquely
     * identifying components that share the same name but are of different types
     * since they will live in different sets. For example, using addBody(toesBody)
     * and addJoint(toesJoint) will have unique paths "/bodyset/toes" and 
     * "/jointset/toes", respectively even if they share the same name, "toes".
     * Model will maintain Components added using these methods in separate %Sets
     * of the corresponding type and they will serialize as part of type specific
     * %Sets.
     * @note these are legacy methods and remain as a convenience alternative to
     * using Component::addComponent().
     * In contrast, components added using addComponent() are not stored in the
     * model's %Sets but live in a flat components list (which is also serialized).
     * Component provides access via getComponentList<%SpecificType> or 
     * getComponent<%SpecificType> to get any subcomponent, including those that
     * are contained in Model's %Sets. In this case, either a Body or a Joint has
     * the pathname "/toes" but both cannot share the same name and will throw
     * SubcomponentsWithDuplicateName.
     * Future versions of OpenSim are likely to deprecate the use of Sets and
     * these methods, because they cannot support new types without modifying the
     * API (for more add####() methods), whereas getComponentList<%SpecificType>()
     * and getComponent<%SpecificType> are more general: they do not have these
     * limitations and are applicable for any Component not just Model.
     */
```

OpenSim/Simulation/Model/AbstractTool.cpp
aymanhab: @aseth1 does that explain or relate to the RRA adjusted model problems you mention in https://github.com/opensim-org/opensim-gui/issues/1008

OpenSim/Simulation/Model/CoordinateSet.cpp
```diff
@@ -67,7 +38,6 @@ CoordinateSet::CoordinateSet(const CoordinateSet& aCoordinateSet):
  */
 void CoordinateSet::populate(Model& model)
 {
-    setModel(model);
     // Aggregate Coordinates owned by the Joint's into a single CoordinateSet
```
aymanhab: What invokese setModel now, or is it not needed anymore?

aseth1: `setModel()` is not part of the `ModelComponent` interface.

OpenSim/Simulation/Model/Ground.cpp
aymanhab: Is component name consistency/checking done across the board in extendFinalizeFromProperties() is this documented?

OpenSim/Simulation/Model/Station.h
```diff
@@ -26,11 +26,11 @@
 
 // INCLUDE
 #include "OpenSim/Simulation/Model/Point.h"
+#include "PhysicalFrame.h"
```
aymanhab: Any reason PhysicalFrame.h include is needed now?

OpenSim/Simulation/Model/TwoFrameLinker.h
```diff
@@ -599,13 +598,16 @@ void TwoFrameLinker<C, F>::updateFromXMLNode(SimTK::Xml::Element& aNode,
                 XMLDocument::addPhysicalOffsetFrame30505(aNode,
                         frame1_connectee_name,
                         frame1Name, locationInFrame1, orientationInFrame1);
+            }
+            // Prior to 4.0, all bodies lived in the Model's BodySet and
```
aymanhab: If this is true for all pre 4.0 models I'd change the test to remove reference to version 30505 which is not supported and refer to 30516 instead

aseth1: I'm OK with the 30505 reference since this is when the change was made. This is not user-facing code so there's no reason for them to know about it. Changing it now is a bit like rewriting history and it has no impact on users or model files generated in 4.0.

OpenSim/Simulation/Model/Station.h
aseth1: Not sure, why this was touched. I'll revert and see.

aseth1: Done.

aseth1:
> One thing to note regarding the addX methods of Model is that these methods are modified in SWIG .i files to handle ownership, not clear if these changes need to be duplicated for the variety of sets or if there are signature changes

The semantics, signature, and scripting use-cases have for the `Model::add###()` methods have not changed in this PR.  If there were issues before then they remain. I don't believe this PR is the correct place to address testing of ownership semantics for scripting (via SWIG).  Having said that, I do believe thy are being tested in *TestModelBuilding.java*

OpenSim/Simulation/Model/TwoFrameLinker.h
aymanhab: @aseth1 by modifying this block you're rewriting history already. The typical use of these version upgrades is that you don't modify the old function (as done here) but instead write a new one to perform the incremental change to previous to current version that's why I'm suggesting you use 30516 which we agreed to maintain so future developers looking into this code don't miss it as an unsupported version.

aymanhab:
Thanks @aseth1 it's only a relevant issue if the signatures changed since then the memory management change is broken by this PR, if the signatures didn't then no need to discuss this issue further.

OpenSim/Simulation/Model/Station.h
aseth1: gcc and clang correctly flag forward declaration as insufficient to define the `Socket<PhysicalFrame>`. Not sure why visual studio lets it through.

aymanhab:
Only a handful of comments/open conversations, once resolved we can merge.

OpenSim/Simulation/Model/TwoFrameLinker.h
aseth1: I don't agree with the statement that the modifications here are changing history, but since it doesn't matter anyways I will make the change to get this PR through.

aseth1:
> Only a handful of comments/open conversations, once resolved we can merge.

Great. I believe the only issue outstanding is the renaming of  `XMLDocument::addPhysicalOffsetFrame30505()`. See above.

OpenSim/Simulation/Model/TwoFrameLinker.h
aseth1: As compromise suggest by @chrisdembia I'll rename `XMLDocument::addPhysicalOffsetFrame30505` to  `XMLDocument::addPhysicalOffsetFrame30505_30517()` since 30505 introduced the `PhysicalOffsetFrame` to embody the transform and 30517 to to reflect the update to the pathname.

aseth1:
@aymanhab are there any issues remaining to be discussed? Would it be easier to discuss over zoom?

aseth1:
Thank you @aymanhab and @chrisdembia. 🎆 🍾 
 I know it was a tough one to review.

PR #2286, opened 2018-09-11
Wrapping object library

pariterre:
Fixes issue {No number, I did not open an issue for that particular problem}
Problem description: wrapping object, apart from inherited parameters, could not be simply modified via swig API for a loaded model. This problem splits in two parts:
1) Some of the wrapping object have legacy parameters
2) The swig interface returns a WrapObject instead of a specific instance (e.g. WrapEllipsoid)

### Brief summary of changes
All the wrapping objects were move to an external library (based on what was already done for the Actuators). 
All properties for all wrapping objects uses the OpenSim_DECLARE_PROPERTY macro allowing simple interface for modifications
Constructor with a WrapObject pointer has been added to all objects so Python (or other swig interface), can construct a new wrapping object from a previous one.
Python interface files (.i) were update accordingly.

### Testing I've completed
testActuators
testWrapping

Some more tests if needed

### CHANGELOG.md (choose one)
- updated the first section with this sentence:
    - Properties of wrapping objects are now defined using the OpenSim_DECLARE_PROPERTY macro
- updated the python section with this sentence:
    - Wrapping objects can be modified


aymanhab:
@pariterre Can you please explain what specific problem you had with wrapping objects and why this need to be solved by separating the WrapObjects into a separate library? WrapObject and its descendents already use DECLARE_PROPERTY macro AFAIK. Thank you.

pariterre:
Hello!
I am actually considering to step back on the separate library. It made sense to me in the first place since wrapping object are different materials that bones or markers. But it seems that this creates more problem than it solves. As you said, most of the wrapping objects use the DECLARE_PROPERTY, but not all of them, so I think the best way is to just modify these (instead of moving the whole thing..). 

The problem I was actually solving is the fact that I could not dynamically change a wrapping object dimension via the Python API, more specifically the WrapEllipsoid. The reason is that the return of the `get` from the `get_WrapObjectSet` method is a pointer to WrapObject which is not able to access specific methods of an inherited class. So, my idea was to add a constructor for each wrapping object that accepts a WrapObject*, dynamically cast it to the proper wrapping object then copies it. 

Like that, in python, it is possible to copy the wrapping object, modify it and then insert it back like that:  
`e = osim.WrapEllipsoid(osim_model.get_BodySet().get(0).get_WrapObjectSet().get(0))
e.set_dimensions(0, osim.Vec3(1, 2, 3))
osim_model.get_BodySet().get(0).get_WrapObjectSet().set(0, e, True)`

chrisdembia:
@pariterre you can use `osim.WrapEllispoid.safeDownCast(osim_model.get_BodySet().get(0).get_WrapObjectSet().get(0))`.

pariterre:
Yep!
I'll clean this properly with the new macro without moving to a new library :)

PR #2291, opened 2018-09-13
Test case to reproduce crash editing joint parent frame

aymanhab:
Fixes issue # https://github.com/opensim-org/opensim-gui/issues/994

### Brief summary of changes
This is just a test case on the core side to exercise and reproduce the crash, not a fix

### CHANGELOG.md (choose one)

- no need to update because test case

### DIFF
--- a/Bindings/Java/OpenSimJNI/Test/testContext.cpp
+++ b/Bindings/Java/OpenSimJNI/Test/testContext.cpp
@@ -238,6 +238,19 @@ int main()
         socket.setConnecteeName(originalConnecteeName);
         context->restoreStateFromCachedModel();
     }
+    AbstractSocket& psocket = shoulder.updSocket("parent_frame");
+    const std::string poriginalConnecteeName = psocket.getConnecteeName();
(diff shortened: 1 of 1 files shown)


Bindings/Java/OpenSimJNI/Test/testContext.cpp
```diff
@@ -238,6 +238,19 @@ int main()
...
+    AbstractSocket& psocket = shoulder.updSocket("parent_frame");
+    const std::string poriginalConnecteeName = psocket.getConnecteeName();
+    try {
+        // create an invalid model
+        psocket.setConnecteeName("r_ulna_radius_hand"); 
+        context->restoreStateFromCachedModel();
+    }
+    catch (...) {
+        // undo the change
+        socket.setConnecteeName(originalConnecteeName);
```
aseth1: The "fix" here is not restoring `psocket` but `socket` from above.

aymanhab:
Folded into PR #2305 closing

Issue #2293, opened 2018-09-13
Rename socket_parent_frame_connectee_name -> socket_parent_frame


jenhicks:

scrum2a:
@chrisdembia is traveling so might not be able to finish this in the sprint.

aymanhab:
@chrisdembia if not working on it can you point me to the branch if you did any work on this? Thanks

chrisdembia:
https://github.com/opensim-org/opensim-core/tree/remove_connectee_name_from_tag

carmichaelong:
Tested on W10 OpenSim 4.0-2018-09-27-d1326ea7. Chose a (pseudo)random subset of models from the distributed Models folder and saved them with the GUI. Checked by looking at the file output. Subset of results are below:

bouncingblock:
![socket-bouncingblock](https://user-images.githubusercontent.com/7455790/46174939-3adc5500-c25f-11e8-9165-bf5d272b10b7.JPG)

gait2354:
![socket-gait2354](https://user-images.githubusercontent.com/7455790/46174950-3f087280-c25f-11e8-8724-a9d04a835168.JPG)

leg39:
![socket-leg30](https://user-images.githubusercontent.com/7455790/46174967-43349000-c25f-11e8-9b48-ffd965c29ad0.JPG)

wrist:
![socket-wrist](https://user-images.githubusercontent.com/7455790/46174971-46c81700-c25f-11e8-8786-7e3e5b62c5e6.JPG)

aymanhab:
Any reason this hasn't been closed @chrisdembia ?

PR #2301, opened 2018-09-22
Remove connectee name suffix

aymanhab:
Fixes issue #2293 

### Brief summary of changes
Change XML tag for Socket's connectee_name to remove the trailing 'connectee_name'

### Testing I've completed
Loaded, saved model, contents were same but tags changed as expected. Ran test suite.

### CHANGELOG.md (choose one)

- updated...

### DIFF
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -28,6 +28,7 @@
 #include "XMLDocument.h"
 #include <unordered_map>
 #include <set>
+#include <regex>
 
 using namespace SimTK;
 
@@ -1137,6 +1138,21 @@ void Component::updateFromXMLNode(SimTK::Xml::Element& node, int versionNumber)
                 }
             }
             
+        }
+        if (versionNumber == 30516) {
+            // Rename xml tags for socket_*_connectee_name to socket_*
+            for (auto iter = node.element_begin();
+                iter != node.element_end();
+                ++iter) {
+                auto tagname = iter->getElementTag();
+                if (std::regex_match(tagname, std::regex("(socket_)(.*)(_connectee_name)"))) {
+                    auto pos = tagname.find("_connectee_name");
+                    if (pos != std::string::npos) {
+                        tagname.replace(pos, 15, "");
+                        iter->setElementTag(tagname);
+                    }
+                }
+            }
         }
     }
     Super::updateFromXMLNode(node, versionNumber);
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -1141,14 +1141,22 @@ void Component::updateFromXMLNode(SimTK::Xml::Element& node, int versionNumber)
         }
         if (versionNumber <= 30516) {
             // Rename xml tags for socket_*_connectee_name to socket_*
+            std::string connecteeNameString = "_connectee_name";
             for (auto iter = node.element_begin();
                 iter != node.element_end();
                 ++iter) {
                 auto tagname = iter->getElementTag();
-                if (std::regex_match(tagname, std::regex("(socket_)(.*)(_connectee_name)"))) {
-                    auto pos = tagname.find("_connectee_name");
+                if (std::regex_match(tagname, std::regex("(socket_|input_)(.*)(_connectee_name)"))) {
+                    auto pos = tagname.find(connecteeNameString);
                     if (pos != std::string::npos) {
-                        tagname.replace(pos, 15, "");
+                        tagname.replace(pos, connecteeNameString.length(), "");
+                        iter->setElementTag(tagname);
+                    }
+                }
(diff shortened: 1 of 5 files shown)


aymanhab:
This PR is ready for review and is very limited. 
It removes connectee_name suffix from socket tag, leaves it in for input (because there are two different tags one for singular, the other for plural and inputs are not used much in XML AFAIK).

OpenSim/Common/Component.cpp
```diff
@@ -1138,6 +1139,21 @@ void Component::updateFromXMLNode(SimTK::Xml::Element& node, int versionNumber)
             }
             
         }
+        if (versionNumber <= 30516) {
+            // Rename xml tags for socket_*_connectee_name to socket_*
+            for (auto iter = node.element_begin();
+                iter != node.element_end();
+                ++iter) {
+                auto tagname = iter->getElementTag();
+                if (std::regex_match(tagname, std::regex("(socket_)(.*)(_connectee_name)"))) {
```
aseth1: Neat!

OpenSim/Common/XMLDocument.cpp
```diff
@@ -64,7 +64,8 @@ using namespace std;
 // 30514 for removing "reverse" property from Joint
 // 30515 for WrapObject color, display_preference, VisibleObject -> Appearance
 // 30516 for GeometryPath default_color -> Appearance
-const int XMLDocument::LatestVersion = 30516;
+// 30517 for removal of Socket _connectee_name from XML markup for Socket
```
aseth1: It sounds a bit like the Socket was removed from XML, but the XML tag name was just shortened.

aymanhab:
Thanks @aseth1 documentation was updated as suggested.

OpenSim/Common/Component.cpp
```diff
@@ -1138,6 +1139,21 @@ void Component::updateFromXMLNode(SimTK::Xml::Element& node, int versionNumber)
...
+                    auto pos = tagname.find("_connectee_name");
+                    if (pos != std::string::npos) {
+                        tagname.replace(pos, 15, "");
```
chrisdembia: It would be preferable to create an `std::string` variable for `"connectee_name"` and to use `size()` here instead of hardcoding 15.

OpenSim/Common/ComponentSocket.h
```diff
@@ -884,15 +884,15 @@ class Input : public AbstractInput {
     /** @{                                                               */ \
     /** comment                                                          */ \
     /** In an XML file, you can set this Socket's connectee name         */ \
-    /** via the <b>\<socket_##cname##_connectee_name\></b> element.      */ \
+    /** via the <b>\<socket_##cname\></b> element.      */ \
```
chrisdembia: Can you realign the end of this line (here and below)?

OpenSim/Common/Component.cpp
aseth1: :+1: good call!

OpenSim/Common/ComponentSocket.h
aymanhab: Done

OpenSim/Common/Component.cpp
aymanhab: Done

aymanhab: @chrisdembia The problem with inputs as I pointed out earlier is that there are two different tags e.g.
'<input_inputs_connectee_names>' or '<input_transform_connectee_name>' should both be removed losing the fact that one of them is plural?

aymanhab:
Changes were made for inputs as well, feedback addressed. Ready for review again @chrisdembia

OpenSim/Common/Component.cpp
chrisdembia: Yes.

OpenSim/Common/ComponentSocket.h
```diff
@@ -960,13 +959,13 @@ class Input : public AbstractInput {
...
     /** In an XML file, you can set this socket's connectee name         */ \
-    /** via the <b>\<socket_##cname##_connectee_name\></b> element.      */ \
+    /** via the <b>\<socket_##cname\></b> element.                       */ \
     /** See AbstractSocket for more information.                         */ \
     /** @see connectsocket_##cname##()                                   */ \
     OpenSim_DOXYGEN_Q_PROPERTY(T, cname)                                    \
     /** @}                                                               */ \
     /** @cond                                                            */ \
-    PropertyIndex PropertyIndex_socket_##cname##_connectee_name {           \
+    PropertyIndex PropertyIndex_socket_##cname {           \
```
chrisdembia: Can you fix the whitespace here and on 894?

aymanhab: Done

aymanhab:
Thanks @aseth1 @chrisdembia 
Will merge after tests pass.

Issue #2292, opened 2018-09-13
Default to absolute pathnames when you serialize


jenhicks:

aymanhab:
@chrisdembia are you working on this or you want me take over?

chrisdembia:
I am not working on this. Thank you for checking.

aseth1:
I forgot why this was on the critical path for 4.0. While I don't think this will break existing models, I do think it makes it more difficult to preserve subassemblies (branch of subcomponents) and to make copies since they are no longer self contained. They will now all refer to components outside the assembly. 

For example  suppose I create a device that controls a joint and you add it to a Joint. The device contains may contain several subcomponents but all it needs to know is the parent joint. So its `Socket_parent = ".."` . If you now copy the device and add it to every joint, you don't need to updated the socket connectee_name. If we make this update, we will then be forced to update each connectee_name one by one and find the absolute path to each joint. Worse, is that inside the assembly it may be wired up by relative paths and these are now going to also be made absolute and will have to be edited if the Device is added elsewhere.

So why is this change needed? I apologize that I cannot remember the rationale.

jenhicks:
If I recall the discussion correctly, we thought that the confusion created by using relative paths (particularly to users who are new to the concept of paths for components) would be greater than the pain created by the case you mentioned above.

I think we discussed still supporting relative paths, so that in the case mentioned above you could define the subassembly using relative paths. The absolute pathnames would just show up when you serialized (presumably after everything was all hooked up).

Once subassemblies become more common and users have a better sense of paths, I think we could change the serialization to relative paths (in a future version).

aseth1:
Thanks Jen, that's helpful. The problem in the solution is that we are storing the absolute pathname (instead of relative) when the socket is Connected. So even it was specified relatively it immediately gets replaced with absolute pathname. So if you create a leg and wire it up, when you copy it and attempt to connect  the copy, its subcomponents will contain absolute paths which try to connect to the existing leg by virtue of the absolute pathnames.

I do think long relative pathnames can be confusing, especially when they traverse through root. Perhaps we could detect those cases and replace them with absolute pathname.

chrisdembia:
I agree with @jenhicks's recollection.

aymanhab:
[...]

PR #2302, opened 2018-09-24
Use absolute pathname when serializing connections

aymanhab:
Fixes issue #2292 

### Brief summary of changes
One place where connections are serialized, replace call to getRelativePathName() with getAbsolutePath()
### Testing I've completed
Loaded models, saved in GUI and got Absolute paths inside sockets.

Should we do the removal of model name by changing the implementation of getAbsolutePath(), or write a variant e.g. getAbsolutePathToTop() or pass an optional argument to getAbsolutePath() this depends on how widely used are these across the API.

### CHANGELOG.md 
- no need to update because this is new for 4.0

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2302)

### DIFF
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -719,7 +719,8 @@ ComponentPath Component::getAbsolutePath() const
 
     while (up && up->hasOwner()) {
         up = &up->getOwner();
-        pathVec.insert(pathVec.begin(), up->getName());
+        if (up->hasOwner()) // Prepend parent only of not top level issue #2276 
+            pathVec.insert(pathVec.begin(), up->getName());
     }
 
     return ComponentPath(pathVec, true);
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -702,7 +702,8 @@ std::string Component::getAbsolutePathString() const
 
     while (up && up->hasOwner()) {
         up = &up->getOwner();
-        absPathName.insert(0, "/" + up->getName());
+        if (up->hasOwner()) // Prepend parent only of not top level issue #2276 
+            absPathName.insert(0, "/" + up->getName());
     }
 
     return absPathName;
--- a/OpenSim/Common/ComponentSocket.h
+++ b/OpenSim/Common/ComponentSocket.h
@@ -355,7 +355,7 @@ public:
         else if(objPathName == ownerPathName)
             setConnecteeName(objPathName);
         else { // otherwise store the absolute path name to the object
-            std::string absPathName = objT->getAbsolutePath().toString();
+            std::string absPathName = objT->getAbsolutePathString();
             setConnecteeName(absPathName);
         }
     }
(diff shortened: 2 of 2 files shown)


aymanhab:
Thoughts on "feedback items" @aseth1 @carmichaelong @chrisdembia

aseth1:
> Should we do the removal of model name by changing the implementation of getAbsolutePath(), or write a variant e.g. getAbsolutePathToTop() or pass an optional argument to getAbsolutePath() this depends on how widely used are these across the API.

I don't see a compelling reason to have have two variants for `getAbsolutePath()` (i.e. adding `getAbsolutePathToTop()`). One can aways call `model.getName()`.  A use case where having the model name in the pathname is perhaps useful when comparing outputs of two similar models.  In the case of outputs, we might want to store the model name in table metadata going forward. I don't think this a new problem, however. If you run a tool twice, having only changed a parameter and the model name, the output files would have identical names and column labels.

jenhicks:
@aseth1 I responded to your comment in the issue, summarizing the discussion we all had (as I recall it).

chrisdembia:
> Should we do the removal of model name by changing the implementation of getAbsolutePath(), or write a variant e.g. getAbsolutePathToTop() or pass an optional argument to getAbsolutePath() this depends on how widely used are these across the API.

I do not think we should have two types of absolute paths (with and without the model name). I think that will be a nightmare to manage, because there is no way to disambiguate an absolute path name "with model name" from an absolute path name "without model name".

aseth1:
Given the that the reason to use absolute pathnames is to avoid confusion for users inspecting dependencies/Sockets in XML, I don't think that warrants using absolute pathnames everywhere.

We can make a relatively simple decision to use absolute pathnames when the absolute pathname has fewer elements that the relative pathname. For example, a Force that is dependent on a Body, will typically have a relative path like:  *../../bodyset/##bodyname##* and it is clearly more direct to have the absolute pathname in this case: */bodyset/##bodyname##*, which has 2 path elements vs the 4 of the relative pathname. But this in not going to be true for a Device and its parts. If the Device has a force (say a spring) that connects two OffsetFrames (which are internal to the Device) the its Sockets will have relative connectee pathnames like:  *../anchor_1* and *../anchor_2* which tell you that the anchor's are part of the Device. As absolute pathnames they look like: */jointset/knee_r/device/anchor_1/* and  */jointset/knee_r/device/anchor_2/*. This doesn't seem horrible, but imagine if the device was composed of a dozen components instead of just three. 

Furthermore, if you clone the device and add it to say the left knee you now have to edit every socket in the device since all the paths are absolute and will no longer be valid. That it going to be a headache for model building. The relative paths within the device are would remain valid no matter where the device is ultimately connected, so you can clone and add the device to get unique instances without having to rename the connections.

I propose using the absolute pathname ONLY when the relative pathname has more path elements. I think this a good compromise between having readable path names without *../../..* everywhere while keeping assemblies self-contained.

@opensim-org/dev-team  please let me  know your thoughts.

chrisdembia:
@aseth1 I copied your last comment to #2292, because most of our discussion is already in that issue.

chrisdembia:
Thanks @aymanhab for making this PR. I copied your changes into #2310 so we can close this PR.

PR #2310, opened 2018-10-06
Delay connectee name

chrisdembia:
This is a WIP PR to delay setting the connectee name property until `finalizeConnections()`.

`testComponentInterface()` does not yet pass. I'm posting this in case @aymanhab or @aseth1 want to take an early look.

I'll add more details later. The main issue is that for copies to work, one must call `finalizeConnections()` first.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2310)


Applications/Analyze/test/arm26.osim
```diff
@@ -683,7 +683,7 @@
 					<frames>
 						<PhysicalOffsetFrame name="ground_offset">
 							<!--Path to a Component that satisfies the Socket 'parent' of type C (description: The parent frame to this frame.).-->
-							<socket_parent_connectee_name>ground</socket_parent_connectee_name>
+							<socket_parent_connectee_name>../../../ground</socket_parent_connectee_name>
 							<!--Translational offset of this frame's origin from the parent frame's origin, expressed in the parent frame.-->
```
aymanhab: @chrisdembia is this the end game/format or an intermediate, or older format used for testing, can you clarify? Thanks

aymanhab: Also if the change is supposed to handle relative and absolute path, wouldn't it be backward compatible and not require any file changes? Sorry if you answer that later in the PR, it just hit me as I started looking at the diffs. Thank you

Bindings/Java/tests/TestEditMarkers.java
```diff
@@ -18,7 +18,7 @@ public static void main(String[] args) {
         Marker newMarker = new Marker();
         newMarker.setName(newMarkerName);
         newMarker.set_location(offset);
-        newMarker.setParentFrameName(body.getName());
+        newMarker.setParentFrame(body);
         context.cacheModelAndState();
```
aymanhab: @chrisdembia If setting names is not supported anymore, this may have wide ranging effect since there are places in the  GUI that work with names/Strings and will need to change to use Components of specific type instead.

aymanhab: I think the modelBuildingUtilties work with Strings/names and primitive types exclusively to avoid having users manipulating paths or Objects. Are these integrated into testing? @carmichaelong can you comment?

aseth1:
___
*[Bindings/Java/tests/TestEditMarkers.java, line 22 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2310#-LP7-gT6O33XbUuZ09tR-r3-22:-LPCUM62Bk5zXIoS2xNl:b-jc6mcz) ([raw file](https://github.com/opensim-org/opensim-core/blob/4dc76044a6689e45856f677c2a195e5f9ea48076/Bindings/Java/tests/TestEditMarkers.java#L22)):*

I think the modelBuildingUtilties work with Strings/names and primitive types exclusively to avoid having users manipulating paths or Objects. Are these integrated into testing? @carmichaelong can you comment?
</blockquote></details>

@aymanhab I don't think these changes exclude setting connections by name- that essentially what has to happen when you deserialize. The issue I was raising, which seems like a discussion for later, is whether we should have two ways of setting connections via the API. I am leaning towards just using references and not names.

@chrisdembia this change was to test that setting by reference works, not because setting by name does not, is that right?

aseth1:
___
*[README.md, line 102 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2310#-LPCACUu30GVqRcfYm3i:-LPCjryn2nO6yc44TWfs:bqas6yp) ([raw file](https://github.com/opensim-org/opensim-core/blob/4dc76044a6689e45856f677c2a195e5f9ea48076/README.md#L102)):*

I can use an absolute path here instead. @aseth1 are you okay with that?

I know the use of `*` and `&` confuses our users, but it sounds like we want to push for using object references instead of strings to make the connections.

The way we add components to a model right now is far from ideal. Ideally, we would do `model.add<CoordinateActuator>(constructorArgs...)`, which would return a pointer and clean up our memory management. This is what they are doing with Drake. 
</blockquote></details>

:+1:  I like `model.add<CoordinateActuator>(constructorArgs...)` a lot and we could use default construction for user defined component  types as well.  We had a bit of a discussion awhile back to do this, but it seemed that SWIG wrapping/scripting use would still be an issue.

aseth1:
Sure. I'll take a look.

aseth1:
___
*[README.md, line 102 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2310#-LPCACUu30GVqRcfYm3i:-LPCvYj3CNWEWxZ5pkUD:bd0dv23) ([raw file](https://github.com/opensim-org/opensim-core/blob/4dc76044a6689e45856f677c2a195e5f9ea48076/README.md#L102)):*

@aseth1 what about using an absolute path for the connectee path here instead of using a reference, as @aymanhab is suggesting?
</blockquote></details>

I prefer to use the reference even in the README. I think it is easier for users to use Objects than to have to know pathnames.

aseth1:
I am unable to build in Debug on Windows:
```
1>
4>docopt_s.lib(docopt.obj) : error LNK2038: mismatch detected for '_ITERATOR_DEBUG_LEVEL': value '0' doesn't match value '2' in opensim-cmd.obj
4>     1>
4>docopt_s.lib(docopt.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MD_DynamicRelease' doesn't match value 'MDd_DynamicDebug' in opensim-cmd.obj
4>     1>
4>LINK : warning LNK4098: defaultlib 'MSVCRT' conflicts with use of other libs; use /NODEFAULTLIB:library
4>     1>
4>G:\OpenSimGit\build_core_64\Debug\opensim-cmd.exe : fatal error LNK1319: 2 mismatches detected
```

In Release I am getting when running `testUpdateFile()`:
```
Exception thrown at 0x00007FFB8E6FA388 in testCommandLineInterface.exe: Microsoft C++ exception: std::runtime_error at memory location 0x000000D47E58F840.
```

Bindings/Java/OpenSimJNI/OpenSimContext.h
```diff
@@ -221,7 +221,13 @@ OpenSim_DECLARE_CONCRETE_OBJECT(OpenSimContext, Object);
 
     void cacheModelAndState();
     void restoreStateFromCachedModel()  SWIG_DECLARE_EXCEPTION;
-    void setSocketConnecteeName(AbstractSocket& socket, const std::string& newValue )   SWIG_DECLARE_EXCEPTION;
+    void setSocketConnecteePath(AbstractSocket& socket,
+            const std::string& newValue)   SWIG_DECLARE_EXCEPTION;
+    // Deprecated.
+    void setSocketConnecteeName(AbstractSocket& socket,
```
aymanhab: No need to keep this , will update the GUI code base. it's used in only a couple places.

Bindings/Python/tests/test_sockets_inputs_outputs.py
```diff
@@ -255,16 +255,18 @@ def print_model():
             reporter = osim.TableReporterVec3()
             reporter.setName('reporter')
             reporter.set_report_time_interval(0.1)
+
+            model.addComponent(reporter)
+
```
aymanhab: @chrisdembia is it documented that reporters need to be added to the model before their inputs are wired? Is that necessary?

OpenSim/Common/Object.cpp
```diff
@@ -236,23 +236,55 @@ void Object::setNull()
...
         const AbstractProperty& myProp    = getPropertyByIndex(px);
         const AbstractProperty& otherProp = other.getPropertyByIndex(px);
 
-        if (!myProp.equals(otherProp))
+        if (!myProp.equals(otherProp)) {
+            printDiff("property '" + myProp.getName() + "'",
+                      myProp.toString(), otherProp.toString());
             return false;
+        }
     }
```
aymanhab: 👍 for returning informative message when diffs are found

OpenSim/Simulation/Model/AbstractPathPoint.cpp
```diff
@@ -75,10 +75,14 @@ void AbstractPathPoint::updateFromXMLNode(SimTK::Xml::Element& aNode,
...
-                // relative path.
-                if (!bodyName.empty()) bodyName = "../../../" + bodyName;
+                // (model, muscle, geometry path), and Bodies are
+                // necessarily 1 levels deep; here we create the correct
+                // relative path (accounting for sets being components).
+                if (bodyName == "ground") {
+                    bodyName = "../../../../" + bodyName;
+                } else if (!bodyName.empty()) {
+                    bodyName = "../../../../bodyset/" + bodyName;
+                }
```
aymanhab: @chrisdembia was that leaked over from an earlier change? Without this PR I can till open the tug-of-war model at 3.3 format and pathpoints on ground are properly attached, any clues?

OpenSim/Simulation/Model/ContactGeometry.cpp
```diff
@@ -127,9 +127,13 @@ void ContactGeometry::updateFromXMLNode(SimTK::Xml::Element& node,
...
-            // prepend "../" to get the correct relative path.
-            if (!body_name.empty()) body_name = "../" + body_name;
+            // (model, contact_geom), and Bodies are necessarily 1 level deep.
+            // Here we create the correct relative path (accounting for sets
+            // being components).
+            if (body_name == "ground")
+                body_name = "../../" + body_name;
+            else if (!body_name.empty())
+                body_name = "../../bodyset/" + body_name;
             XMLDocument::addConnector(node, "Connector_PhysicalFrame_",
```
aymanhab: @chrisdembia Can we move this into an XML utility so that next round of changes to handle AbsolutePath changes one place rather than all over the code base?

OpenSim/Simulation/Model/Model.h
```diff
@@ -266,6 +266,14 @@ OpenSim_OBJECT_NONTEMPLATE_DEFS(Model, ModelComponent);
...
     explicit Model(const std::string& filename) SWIG_DECLARE_EXCEPTION;
 
+    /** Satisfy all connections (Sockets and Inputs) in the model, using this
+     * model as the root Component. This is a convenience form of
+     * Component::finalizeConnections() that uses this model as root.
+     */
+    void finalizeConnections() { finalizeConnections(*this); }
+    // Allow overloading.
+   using Component::finalizeConnections;
+
```
aymanhab: @chrisdembia Can you clarify why is this necessary if Model is already a Component?

OpenSim/Simulation/SimbodyEngine/Joint.h
```diff
@@ -225,6 +225,11 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Joint, ModelComponent);
     // Utility
     bool isCoordinateUsed(const Coordinate& aCoordinate) const;
 
+    /** Add a frame to the *frames* property in this Joint. The frame is
+     * adopted, and should have been dynamically allocated.
+     * Use this function instead of append_frames(). */
+    void addFrame(PhysicalOffsetFrame* frame);
+
```
aymanhab: @chrisdembia  Was this necessary for this PR?

chrisdembia: Yes. There was coding that was using `append_frames()` to copy components into the `Joint`. and this was causing connections to fail because the connection was established with the original component outside the model, not with the component in the model.

OpenSim/Simulation/Model/Model.h
chrisdembia: Overloading a virtual function causes the base class overloads to be hidden, unless you use `using ...`. https://stackoverflow.com/questions/6727087/c-virtual-function-being-hidden

Bindings/Python/tests/test_sockets_inputs_outputs.py
chrisdembia: It was necessary early on in the PR but I later removed that restriction (that's why there's `Input::_registeredChannels` now. I can try reverting this change.

OpenSim/Simulation/Test/testModelInterface.cpp
```diff
@@ -193,9 +193,9 @@ void testModelTopologyErrors()
 
 
     Model degenerate;
-    auto frame1 = new PhysicalOffsetFrame("frame1", model.getGround(),
+    auto frame1 = new PhysicalOffsetFrame("frame1", degenerate.getGround(),
         SimTK::Transform(SimTK::Vec3(0, -0.1, 0)));
-    auto frame2 = new PhysicalOffsetFrame("frame2", model.getGround(),
+    auto frame2 = new PhysicalOffsetFrame("frame2", degenerate.getGround(),
         SimTK::Transform(SimTK::Vec3(0, 0.2, 0)));
```
aymanhab: @chrisdembia isn't this changing the test? Is this related to the PR?

OpenSim/Tests/Components/testComponents.cpp
```diff
@@ -248,9 +248,7 @@ void testComponent(const Component& instanceToTest)
                 const Object* depRawPtr = dependency.get();
                 addObjectAsComponentToModel(dependency.release(), model);
 
-                // Connect the socket. This should come after adding the
-                // dependency to the model, otherwise the "connectee name"
-                // relative path will be incorrect.
+                // Connect the socket.
```
aymanhab: Is the old comment out of date now?

OpenSim/Tools/Test/visualize_contacts.osim
```diff
@@ -185,7 +185,7 @@
           <connectors>
             <Connector_PhysicalFrame_ name="frame">
               <!--Name of the component this Connector should be connected to.-->
-              <connectee_name>model_/ground</connectee_name>
+              <connectee_name>../../ground</connectee_name>
             </Connector_PhysicalFrame_>
```
aymanhab: I assume this is a 4.0 early format, true @chrisdembia

Bindings/Python/tests/test_sockets_inputs_outputs.py
aymanhab: No need to, I mainly wanted to know if the documentation needed to match the new behavior.

Bindings/Java/tests/TestEditMarkers.java
chrisdembia: Nice catch @aymanhab. ModelBuilder.py does use strings, but there's at least one location that must be updated as a result of sets being components: https://github.com/opensim-org/opensim-gui/blob/f9ad640d243aacdf825a8c2345fb7fa5c150ce48/Gui/opensim/Scripts/ModelBuilder.py#L32

That is not related to this PR; it is related to the previous PR about sets.

OpenSim/Simulation/Model/ContactGeometry.cpp
chrisdembia: Done.

chrisdembia:
@aymanhab for updating ModelBuilder, I've created a new issue: https://github.com/opensim-org/opensim-gui/issues/new

@aseth1 can you re-review? GitHub says you're requesting changes.

chrisdembia:
Ready for re-review.

chrisdembia:
Thanks @aymanhab

chrisdembia:
Thanks @aseth1. As discussed, do not merge this yet.

aymanhab:
@chrisdembia Can we merge master into this branch so that we have final set of bindings to work with?

chrisdembia:
@aymanhab suggested that we review what XML files look like now that we tweaked connectee paths, since the whole point of embarking on these changes was to improve the XML format. So I've attached an updated Arm26 model. Please share whether you think this format is good or if you think we should make more changes.

```xml
<WeldJoint name="offset">
	<!--Path to a Component that satisfies the Socket 'parent_frame' of type PhysicalFrame (description: The parent frame for the joint.).-->
	<socket_parent_frame>ground_offset</socket_parent_frame>
	<!--Path to a Component that satisfies the Socket 'child_frame' of type PhysicalFrame (description: The child frame for the joint.).-->
	<socket_child_frame>base_offset</socket_child_frame>
	<!--Physical offset frames owned by the Joint that are typically used to satisfy the owning Joint's parent and child frame connections (sockets). PhysicalOffsetFrames are often used to describe the fixed transformation from a Body's origin to another location of interest on the Body (e.g., the joint center). When the joint is deleted, so are the PhysicalOffsetFrame components in this list.-->
	<frames>
		<PhysicalOffsetFrame name="ground_offset">
			<!--The geometry used to display the axes of this Frame.-->
			<FrameGeometry name="frame_geometry">
				<!--Path to a Component that satisfies the Socket 'frame' of type Frame.-->
				<socket_frame>..</socket_frame>
				<!--Scale factors in X, Y, Z directions respectively.-->
				<scale_factors>0.20000000000000001 0.20000000000000001 0.20000000000000001</scale_factors>
			</FrameGeometry>
			<!--Path to a Component that satisfies the Socket 'parent' of type C (description: The parent frame to this frame.).-->
			<socket_parent>/ground</socket_parent>
			<!--Translational offset (in meters) of this frame's origin from the parent frame's origin, expressed in the parent frame.-->
			<translation>0 0.80000000000000004 0</translation>
			<!--Orientation offset (in radians) of this frame in its parent frame, expressed as a frame-fixed x-y-z rotation sequence.-->
			<orientation>0 0 0</orientation>
		</PhysicalOffsetFrame>
		<PhysicalOffsetFrame name="base_offset">
			<!--The geometry used to display the axes of this Frame.-->
			<FrameGeometry name="frame_geometry">
				<!--Path to a Component that satisfies the Socket 'frame' of type Frame.-->
				<socket_frame>..</socket_frame>
				<!--Scale factors in X, Y, Z directions respectively.-->
				<scale_factors>0.20000000000000001 0.20000000000000001 0.20000000000000001</scale_factors>
			</FrameGeometry>
			<!--Path to a Component that satisfies the Socket 'parent' of type C (description: The parent frame to this frame.).-->
			<socket_parent>/bodyset/base</socket_parent>
			<!--Translational offset (in meters) of this frame's origin from the parent frame's origin, expressed in the parent frame.-->
			<translation>0 0 0</translation>
			<!--Orientation offset (in radians) of this frame in its parent frame, expressed as a frame-fixed x-y-z rotation sequence.-->
			<orientation>0 0 0</orientation>
		</PhysicalOffsetFrame>
	</frames>
</WeldJoint>
```

```xml
<PathPointSet>
	<objects>
		<PathPoint name="TRIlong-P1">
			<!--Path to a Component that satisfies the Socket 'parent_frame' of type PhysicalFrame (description: The frame in which this path point is defined.).-->
			<socket_parent_frame>/bodyset/base</socket_parent_frame>
			<!--The fixed location of the path point expressed in its parent frame.-->
			<location>-0.053650000000000003 -0.013729999999999999 0.14723</location>
		</PathPoint>
...
	</objects>
	<groups />
</PathPointSet>
```

aseth1:
The format looks good to me.  IMHO we shouldn't expose attributes in XML that users cannot change. For example, AFAICT it doesn't matter if you remove or edit `FrameGeometry`. 
```XML
				<Body name="base">
					<!--The geometry used to display the axes of this Frame.-->
					<FrameGeometry name="frame_geometry">
						<!--Path to a Component that satisfies the Socket 'frame' of type Frame.-->
						<socket_frame>..</socket_frame>
						<!--Scale factors in X, Y, Z directions respectively.-->
						<scale_factors>0.20000000000000001 0.20000000000000001 0.20000000000000001</scale_factors>
					</FrameGeometry>
...
```

The `FrameGeometry` is always created for a `Frame` and I think we want that. Changing its properties also has no effect. Perhaps FrameGeometry could be private member subcomponent of Frame and not exposed in XML and does not take up XML real-estate.

aymanhab:
@chrisdembia I'm a little surprised to see the ".." in FrameGeometry/socket_frame, could be misunderstanding on my part. Can you clarify the expectation of when any ../* would appear in XML?

I wouldn't make major changes at this point to remove FrameGeometry. Users can't change the type but can change scale and appearance/visibility which doesn't show in XML by default.

chrisdembia:
@aymanhab good point. Geometry::setFrame() sets the connectee path to be a relative path. I can remove this inconsistent handling of connecting if you would like. 

I started looking into hiding frame_geometry from XML and I think I could finish the change on Monday but I do think we should maybe hold off on this change.

aseth1:
IMHO ".." is preferable to an absolute path in the direct parent case. Why would we want a long path name here instead, especially if we support relative paths?

aymanhab:
I'm ok with either, it was mainly to understand what scenarios allow for leading ".." since earlier discussion suggested these will be systematically replaced with absolute paths.

aymanhab:
I think we agreed to merge this branch so we can get closer to final layout of Pathname. Anything else to add before the merge, @chrisdembia @aseth1 ?

aseth1:
I think we agreed to merge this branch so we can get closer to final layout of Pathname. Anything else to add before the merge, @chrisdembia @aseth1 ?

:+1:

chrisdembia:
Is this PR good to go?

chrisdembia:
Thanks @aymanhab and @aseth1 for the reviews. I'm happy to help with GUI testing and any fallout from this PR.

PR #2275, opened 2018-08-25
API guide updates

nickbianco:
This PR updates the API guide per discussion in #2254 and today's stand-up meeting.

It is still work-in-progress, but I'm creating the PR now in anyone wants to see the updates in the Doxygen build. I'll add reviewers once changes have been finalized.

### Brief summary of changes  API guide 
- Updated/added figures
- Moved over content from Confluence relevant to API structure
- Fixed markdown commands that weren't being parsed correctly.

### Looking for feedback on...
Overall layout, things to add/remove, general feedback, etc.

### CHANGELOG.md 
Should I update the change log?

The Doxygen for this PR can be viewed at http://myosin.sourceforge.net/?C=N;O=D; click the folder whose name is this PR's number.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2275)

PR #2311, opened 2018-10-09
Remove 'core' as reference to SimTK libraries in the API guide

nickbianco:
Small change to the API guide that didn't make it into #2275. This removes the term "core" as a reference to the underlying SimTK libraries.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2311)


nickbianco:
@aseth1 here's that last change that didn't make it into the original PR. (I did that comma in since "low-level" and "efficient" both are adjectives for "math and algebra libraries").

PR #2312, opened 2018-10-09
Make two PhysicalOffsetFrames when upgrading a Joint

carmichaelong:
Fixes issue #2294

### Brief summary of changes
- Removed `if` statements that prevented a PhysicalOffsetFrame to be made when it would share the same frame location and orientation as the parent body.
- Updated tests to reflect increased number of Components.

### Testing I've completed
All tests pass except for `testVisualization`. @aymanhab 

Should this be mentioned in the CHANGELOG?

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2312)

### DIFF
--- a/Bindings/Python/tests/test_sockets_inputs_outputs.py
+++ b/Bindings/Python/tests/test_sockets_inputs_outputs.py
@@ -41,14 +41,14 @@ class TestSockets(unittest.TestCase):
 
         # Check that the connectees point to the correct objects.
         assert (shoulder.getConnectee("child_frame").this ==
-                model.getBodySet().get("r_humerus").this)
+                shoulder.getComponent("r_humerus_offset").this)
 
         assert (
             type(shoulder.getSocket("child_frame").getConnecteeAsObject())
             == osim.OpenSimObject)
         # In Python, we are able to get the concrete type from this method.
         # by using a SWIG typemap(out).
-        assert type(shoulder.getConnectee("child_frame")) == osim.Body
+        assert type(shoulder.getConnectee("child_frame")) == osim.PhysicalOffsetFrame
 
     def test_iterate_sockets(self):
         model = osim.Model(os.path.join(test_dir, "arm26.osim"))


OpenSim/Simulation/Test/testFrames.cpp
```diff
@@ -441,7 +441,7 @@ void testFilterByFrameType()
             << " of type " << typeid(component).name() << std::endl;
     }
 
-    ASSERT_EQUAL(9, i, __FILE__, __LINE__,
+    ASSERT_EQUAL(11, i, __FILE__, __LINE__,
```
tkuchida: Suggest changing `9` to `11` in the comment (L445) as well (to avoid mismatch between condition and message). Analogous comment applies to L453 and L474.

chrisdembia:
I remember chatting about adding offset frames as properties or internal subcomponents to the Joint. Did we decide against that? I just don't remember.

aseth1:
> I remember chatting about adding offset frames as properties or internal subcomponents to the Joint. Did we decide against that? I just don't remember.

`Joint` already has a list property of *frames* of type  `PhyscialOffsetFrame`  but it does not appear like we can edit this list property in the GUI (@aymanhab  can comment).  If these were editable it could provide sufficient flexibility to utilize OffsetFrames when updating old model files (as in this PR) but also allow users to connect to directly to other `PhysicalFrame`s without extra offset frames. If we always want to have parent and child offset frames associated with the Joint (even if they are identity) then simply having two properties for `PhysicalOffsetFrame` also makes sense. Would call them *parent_offset* and *child_offset*? In that case we should remove the *frames* list property.

One thing that may be confusing, is that even the `Joint`'s *child_offset*  as a `PhysicalOffsetFrame` will have a `Socket<PhysicalFrame>` named *parent_frame*.

tkuchida:
> If we always want to have parent and child offset frames associated with the Joint (even if they are identity)

Yes, I advocate for applying the same logic to the topology regardless of the particular property values (i.e. location of the Joint). Currently, the user must follow a different sequence of operations (in the GUI, in XML, in scripting, in C++) to relocate a Joint depending on whether it was originally coincident with the Body's origin, which is not ideal. Turn the key clockwise to unlock the door, but counterclockwise if the car is red :car:.

aymanhab:
The tests pass except for test_sockets_inputs_outputs.py (Line 44) because the child_frame is not "r_humerus" itself but the offset frame inside the joint. Test needs updating. Suggestions? @aseth1 @carmichaelong or @chrisdembia

aymanhab:
All tests pass now, this is ready for review.

PR #2315, opened 2018-10-16
Fix bug in InducedAccelerationSolver

chrisdembia:
This PR fixes what I think is a small bug in InducedAccelerationSolver.


tkuchida:
All tests passed before and after this change. Perhaps create an Issue to improve test case coverage?

chrisdembia:
> 👍 🥇 I don't know how we were computing the correct total acceleration with that mistake. How did you find it?

I was skimming the code during our discussion about IAA the other day.

chrisdembia:
@tkuchida I created an issue.

aseth1:
Updated *testInducedAccelerations* case that calls the solver to include a condition for the "total" acceleration.

PR #2326, opened 2018-10-23
Remove model name from absolute path, and use absolute path (sometimes) for connectee path

chrisdembia:
Fixes issue #2276, #2292.

### Brief summary of changes

- Remove the model name from absolute paths.
- Use absolute paths for connectee paths (if there are ".."'s in the relative path).

The base branch for this PR is `delay_connectee_name` so that reviewers can start reviewing this PR while #2310 is still going through review. Once we merge #2310, I will change the base branch for this PR to `master`.

### Testing I've completed

- CTest

### CHANGELOG.md (choose one)

- no need to update because...these are post-4.0 changes.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2326)


chrisdembia:
Based on discussion in the daily sprint today, we would like to review and merge this PR before merging #2310 so GUI testing can be done only once on both of these changes.

OpenSim/Common/Component.h
```diff
@@ -3050,11 +3052,12 @@ void Socket<C>::finalizeConnection(const Component& root) {
             "root component. Did you intend to add '" +
             rootOfConnectee.getName() + "' to '" + myRoot.getName() + "'?");
 
-        std::string objPathName = connectee->getAbsolutePathString();
-        std::string ownerPathName = getOwner().getAbsolutePathString();
-        // otherwise store the relative path name to the object
-        std::string relPathName = connectee->getRelativePathName(getOwner());
-        updConnecteePathProp().setValue(0, relPathName);
+        std::string connecteePath = connectee->getRelativePathName(getOwner());
+        // If the relative path start with ".." the use an absolute path
```
carmichaelong:
```suggestion
        // If the relative path starts with ".." then use an absolute path
```

OpenSim/Common/Component.h
```diff
@@ -3050,11 +3052,12 @@ void Socket<C>::finalizeConnection(const Component& root) {
...
+        // instead.
+        if (connecteePath.size() >= 2 && connecteePath.substr(0, 2) == "..")
```
carmichaelong: I doubt this would really be an issue, but there shouldn't be a way to form a "correct" path such that it's something like
`..someCharsBeforeTheFirstSlash/foo/bar`, right?

OpenSim/Common/Component.h
```diff
@@ -3138,11 +3141,15 @@ void Input<T>::finalizeConnection(const Component& root) {
...
             // Update the connectee path as
-            // <RelOwnerPath>/<Output><:Channel><(annotation)>
-            ComponentPath path(chan->getOutput().getOwner().getRelativePathName(
-                    getOwner()));
-
-            auto pathStr = composeConnecteePath(path.toString(),
+            // <OwnerPath>/<Output><:Channel><(annotation)>
+            const auto& outputOwner = chan->getOutput().getOwner();
+            std::string path = outputOwner.getRelativePathName(getOwner());
+            // If the relative path start with ".." the use an absolute path
```
carmichaelong:
```suggestion
            // If the relative path starts with ".." then use an absolute path
```

OpenSim/Common/Test/testComponentInterface.cpp
```diff
@@ -690,6 +690,11 @@ void testMisc() {
 
     cout << "Adding world3 to theWorld" << endl;
     theWorld.add(world3.clone());
+    // We now use absolute paths for the connectee name, and adding world3 to
+    // theWorld makes the previously-stored absolute path incorrect.
+    // TODO add() should update absolute connectee paths; no longer necessary.
```
carmichaelong: What does the "no longer necessary" refer to here?

OpenSim/Common/Component.h
chrisdembia: Good question. Based on these lines:

https://github.com/opensim-org/opensim-core/blob/083fcc4efcbf47839bdc1906fcfa634df14fe148/OpenSim/Common/Component.cpp#L178-L180

...I think it would be possible to have such a path and it might mostly function without errors. Perhaps we should add to the path validation that if a path element is not "." or ".." then it cannot contain dots.

In this case, an "absolute path" will be used (e.g., `/bar/foo/..someChars/biz/faz`) and the connectee should still be found.

I suppose the issue is not so different from that for filesystems: it's valid to have `.` in a filename.

OpenSim/Common/Test/testComponentInterface.cpp
chrisdembia: Ah thanks for the catch. I can remove these changes now that connectee paths are updated in `add...()`.

OpenSim/Common/Component.h
carmichaelong: Maybe the way to be sure then is to use the `Path` class to check if the first element of the `Path` is exactly `..`? I guess it just depends on how much we want this to be like a filesystem and/or accept `.` in a path. Separately, I don't think this is critical for this PR.

chrisdembia: I've updated the PR to use the `Path` class. Let me know what you think.

carmichaelong: LGTM

chrisdembia: Thanks.

chrisdembia:
@aseth1 awaiting your review. Note that this PR is based on the `delay_connectee_name` branch.

Issue #2324, opened 2018-10-22
XML files for external loads are not readable


aymanhab:
PR https://github.com/opensim-org/opensim-core/pull/2314 changes the layout of ExternalLoads.xml file causing old files not to be readable. This can be checked by creating a new file and diffing against older file.  Must fix for 4.0 since it affects all our and user workflows that use external loads.

aymanhab:
In case it helps, the latest build from master has the same issue (without the rollback of PR #2314) trying to edit an old ExternalLoads.xml file.

aseth1:
I am using the build OpenSim 4.0-2018-10-17-a1ed8256 from [here ](https://ci.appveyor.com/project/opensim-org/opensim-gui/builds/19587105/artifacts) that we tested during the sprint review.

When running InverseDynamics from the GUI, I do see the error in trying to edit (clicking the pencil) and external loads XML file:
![image](https://user-images.githubusercontent.com/6494122/47325453-acb67d00-d618-11e8-9fd8-420e6d72dd53.png)
However, the Tool runs and produces identical results with 3.3.

I then updated the file using *Tools->Convert Files...* and I still get the same error. Therefore, I don't believe it is an updateFromXML issue but some particular expectation by the GUI about the file. Here is the diff:
![image](https://user-images.githubusercontent.com/6494122/47325577-60b80800-d619-11e8-907b-783928e04e95.png)

Again the Tool runs and produces the same results.

I was able to edit the *subject01_walk1_grf.xml* in an earlier build OpenSim_128d4fbd-2018-08-06 and save it as a new ExternalLoads file. That file was editable in the latest build. The diff between these files is:
![image](https://user-images.githubusercontent.com/6494122/47325759-416daa80-d61a-11e8-831e-cd4b98e4c5f9.png)

Consequently, the source of the problem is the path to the files. If we used the absolute path there doesn't seem to be an issue editing it in the GUI. The irony is that the Tool runs and can find the file locally. I do not believe this is an `updateFromXML` issue.

aseth1:
I was also able to make an `ExternalLoads` from scratch without issues and produce the same results from testID case with the *subject01.osim* model.

aseth1:
[...]

PR #2332, opened 2018-10-25
ExternalLoads itself finds datasource local to external loads (XML) file

aseth1:
Fixes issue #2324

### Brief summary of changes
- made `ExternalLoads` "self-sufficient" in terms of loading its own data source. If the data source cannot be located directly, the file is tested in the in the same location of the `ExternalLoads` document directory, if it has one.
- removed change of directory maneuvers from `AbstractTool` and `DynamicsTool` `createExternalLoads()` for loading the external loads data source.

### Testing I've completed
- all API tests pass locally
- build GUI (master) on top and was able to edit existing external loads (xml) file and perform steps in https://github.com/opensim-org/opensim-gui/issues/1035

### CHANGELOG.md (choose one)

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2332)

### DIFF
--- a/OpenSim/Simulation/Model/ExternalLoads.cpp
+++ b/OpenSim/Simulation/Model/ExternalLoads.cpp
@@ -190,9 +190,11 @@ void ExternalLoads::extendConnectToModel(Model& aModel)
             catch (const std::exception &ex) {
                 cout << "Error: failed to construct ExternalLoads from file "
                     << _dataFileName << endl;
-                if (getDocument()) IO::chDir(savedCwd);
+                if (getDocument())
+                    IO::chDir(savedCwd);
                 throw(ex);
             }
+            IO::chDir(savedCwd);
         }
 
         for (int i = 0; i < getSize(); ++i)
--- a/OpenSim/Simulation/Model/ExternalLoads.cpp
+++ b/OpenSim/Simulation/Model/ExternalLoads.cpp
@@ -134,7 +134,8 @@ void ExternalLoads::setupSerializedMembers()
     _propertySet.append(&_dataFileNameProp);
 
     _externalLoadsModelKinematicsFileName="";
-    comment =   "Optional motion file (.mot) or storage file (.sto) containing the model kinematics "
+    comment =   "The option is deprecated and unnecessary to apply external loads. " 
+                "A motion file (.mot) or storage file (.sto) containing the model kinematics "
                 "used to transform a point expressed in ground to the body of force application."
                 "If the point is not expressed in ground, the point is not transformed";
     _externalLoadsModelKinematicsFileNameProp.setComment(comment);
--- a/OpenSim/Simulation/Model/ExternalLoads.cpp
+++ b/OpenSim/Simulation/Model/ExternalLoads.cpp
@@ -176,8 +176,24 @@ void ExternalLoads::extendConnectToModel(Model& aModel)
     // BASE CLASS
     Super::extendConnectToModel(aModel);
 
+    Storage *forceData = nullptr;
     if (_dataFileName.length() > 0) {
-        Storage *forceData = new Storage(_dataFileName);
+        if(IO::FileExists(_dataFileName))
+            forceData = new Storage(_dataFileName);
+        else if(getDocument()) { // ExternalLoads constructed from file
+            // then change working directory the ExternalLoads location
+            std::string savedCwd = IO::getCwd();
+            IO::chDir(IO::getParentDirectory(getDocumentFileName()));
+            try {
+                forceData = new Storage(_dataFileName);
+            }
+            catch (const std::exception &ex) {
+                cout << "Error: failed to construct ExternalLoads from file "
+                    << _dataFileName << endl;
+                if (getDocument()) IO::chDir(savedCwd);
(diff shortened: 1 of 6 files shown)


OpenSim/Simulation/Model/ExternalLoads.cpp
```diff
@@ -176,8 +176,24 @@ void ExternalLoads::extendConnectToModel(Model& aModel)
...
+            // then change working directory the ExternalLoads location
+            std::string savedCwd = IO::getCwd();
+            IO::chDir(IO::getParentDirectory(getDocumentFileName()));
+            try {
+                forceData = new Storage(_dataFileName);
+            }
+            catch (const std::exception &ex) {
+                cout << "Error: failed to construct ExternalLoads from file "
+                    << _dataFileName << endl;
+                if (getDocument()) IO::chDir(savedCwd);
```
chrisdembia: You already know that `getDocument()` is true; you can remove the check here.

OpenSim/Simulation/Model/ExternalLoads.cpp
```diff
@@ -176,8 +176,24 @@ void ExternalLoads::extendConnectToModel(Model& aModel)
...
+                throw(ex);
+            }
```
chrisdembia: You should add `IO::chDir(savedCwd)` here.

aseth1:
___
*[OpenSim/Simulation/Model/AbstractTool.cpp, line 589 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2332#-LPj3iRDCfiIvsPkaD_P:-LPlRu4C3GgmvoAfKe8O:b2osfio) ([raw file](https://github.com/opensim-org/opensim-core/blob/0c15b7ac7594a1de4f5150f17c5011028ace1e00/OpenSim/Simulation/Model/AbstractTool.cpp#L589)):*

Good point!
</blockquote></details>

I agree. But it doesn't seem to be an issue in my testing- it was able to find the kinematics just fine. The better solution is to push the transformation operation into ExternalLoads OR (which I prefer) is to remove that "option" since it is more of a waste of time than providing any different functionality. It is such a source of confusion t hat it should have been removed years ago.

aseth1:
___
*[OpenSim/Simulation/Model/AbstractTool.cpp, line 589 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2332#-LPj3iRDCfiIvsPkaD_P:-LPlXD1m9xtFEuoakDe6:b-sf5u3) ([raw file](https://github.com/opensim-org/opensim-core/blob/0c15b7ac7594a1de4f5150f17c5011028ace1e00/OpenSim/Simulation/Model/AbstractTool.cpp#L589)):*

I agree. But it doesn't seem to be an issue in my testing- it was able to find the kinematics just fine. The better solution is to push the transformation operation into ExternalLoads OR (which I prefer) is to remove that "option" since it is more of a waste of time than providing any different functionality. It is such a source of confusion t hat it should have been removed years ago.
</blockquote></details>

OK I was not switch the directory back! Thanks for pointing it out. I will try to push all file related processing into ExternalLoads. Let's discuss now if we want to keep the option.

chrisdembia:
> I'm fine with any change that minimizes code disturbance at this point.

In the daily scrum today, @aseth1 was pushing for removing the kinematics option. @carmichaelong saw arguments for both sides and suggested we discuss on Monday. @aseth1 said he could prepare both changes.

My preference is to keep the kinematics option but I am fine with removing it if we are willing to consider adding it back if removing the option substantially prevents users from moving to 4.0.

aseth1:
___
*[OpenSim/Simulation/Model/ExternalLoads.cpp, line 195 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2332#-LPhF8G1UkxukIj776tV-r1-195:-LPly_s85T09sLzFpRXH:b-rfhxnm) ([raw file](https://github.com/opensim-org/opensim-core/blob/0c15b7ac7594a1de4f5150f17c5011028ace1e00/OpenSim/Simulation/Model/ExternalLoads.cpp#L195)):*

agreed
</blockquote></details>

As you guys probably guessed, the kinematics file is now not found in the GUI. I am exploring pushing it into ExternalLoads, but I don't think it is trivial. A band-aid solution is to let the Tool find it by putting back the chDir. If we plan to remove it, this seem easiest, although ugly.

aseth1:
@aymanhab @chrisdembia thank you for pointing out the missing change of directory back to the saved working directory. PR was updated, but this caused GUI testing to fail, in that now the kinematics for external loads (also as pointed out during the review) is now not found.

As a band-aid solution, I have the `AbtractTool` and `DynamicsTool` do the change of director *maneuver* as before. The Tool then run successfully (as before), but the issue is that the kinematics for external loads is not found by the `EditExternalLoadsPanel` in the GUI and it shows up red. The panel has its own code for checking the files, which seems redundant. Perhaps that could be updated separately to ask the Tool for its ExternalLoads/files.

As it is now the user will see a red field, but it doesn't prevent the Tool from running:
image

aseth1:
@aymanhab @chrisdembia, PR is ready for re-review.

PR #2338, opened 2018-10-29
Make Frame frame_geometry a member subcomponent.

chrisdembia:
This PR is in response to @aseth1 's feedback in #2310 that the `FrameGeometry` that comes with Frames should not be exposed.

This should not be merged until after #2310 is merged.


aymanhab:
@chrisdembia The reason this and some other geometry components were exposed in the first place is to allow modification of display options e.g. show/hide or scale and to make these changes persistent. As of now some of these options may not work for FrameGeometry but to remove it completely from XML and then try to find a place to include  a separate Appearance object (similar to how we do for Wrap Objects)  seems a bit of an incomplete solution. Sorry I didn't explain in my earlier feedback. 

If we can get the XML to be hidden because all the contents are the same for all FrameGeometry then that would reduce the clutter without affecting functionality though I don't know if that's possible or important at this point.

chrisdembia:
@aymanhab I understand, and I agree we want to be able to expose certain properties of the `FrameGeometry`. I was thinking what you were saying: add in properties for the specific things we want to be able to customize. I figure it will be easier to add in such customization later rather than expose `frame_geometry` now and require writing `updateFromXMLNode()` code later.

I am fine either way. It's up to the group if we want to merge this.

Issue #2339, opened 2018-10-29
Support result files from earlier Beta versions of 4.0


aymanhab:
Per discussion we agreed to support result files generated in 4.0 Betas. To support this we'll move some duplicate functionality from GUI responsibility to API.

- bool motionAssociationPossible(Model modelForMotion, Storage storage) will detect that the storage contains at least one column corresponding to a coordinate from the model 
-

PR #2341, opened 2018-10-30
testStorageGetStateIndexBackwardsCompatibility()

chrisdembia:
This PR is the start to fixing #2339

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2341)


aseth1:
@chrisdembia I pushed a small fix to `Storage::getStateIndex` to handle the intermediate 4.0 versions of Component path-naming when finding associated columns.

aseth1:
> Is this ready for review?

Yes.

I am also building the GUI against to verify that I can load legacy versions of states files. Will report soon.

aseth1:
Using CMC results from ISB 2017 jumper solution I no longer get `motionAssociationPossible` to return false (and the cannot associate message) but now the motion is not applied- model is motionless, no muscle color changes, but timeline appears in motion player.

aseth1:
Looks like the culprit was `Model::formStateStorage()` was out of sink with the changes so made it to use `Storage::getStateIndex()` as well.

chrisdembia:
LGTM.

chrisdembia:
Multiple tests failed.

chrisdembia:
I'm working on a fix. I think it's an off-by-1 error.

chrisdembia:
I pushed a fix.

aseth1:
Thanks @chrisdembia! So strange that all the tests passed locally. Anyhow LGTM!

chrisdembia:
I’m confused about whether we need another reviewer since we worked on this together. Anyone interested?

aseth1:
@aymanhab @chrisdembia I can confirm in the GUI that I can play states results (e.g. from CMC and Forward) in the GUI (master) and these changes.

aymanhab:
Thanks @aseth1 and @chrisdembia, I'll start the review shortly

aseth1:
___
*[OpenSim/Simulation/Model/Model.cpp, line 1962 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/2341#-LQFf6zP466scD46SK_e:-LQFhj3g4Gc5GFE5YmI5:b-wimuyq) ([raw file](https://github.com/opensim-org/opensim-core/blob/609321376207b9ff7f2c5888707ff72eb9d4bb61/OpenSim/Simulation/Model/Model.cpp#L1962)):*

The block in getStateIndex is quite different from what was removed here, what are the side effects?
</blockquote></details>

The were not identical but were having the same outcomes. First step to towards having unified conversion code is to centralize it.

aseth1:
___
*[OpenSim/Common/ComponentSocket.h, line 158 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/2341#-LQFdGlu87tjTbNFNgXG:-LQFiGNY8JOBNJ8Gom_y:b-fw4sh0) ([raw file](https://github.com/opensim-org/opensim-core/blob/609321376207b9ff7f2c5888707ff72eb9d4bb61/OpenSim/Common/ComponentSocket.h#L158)):*

Is "partial" accurate, any middle segment of Path? Trailing only?
</blockquote></details>

This is merged code from a previously accepted PR

aseth1:
___
*[OpenSim/Simulation/Model/ExternalLoads.cpp, line 205 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/2341#-LQF_4InB8QXrgIGK-EF:-LQFia5F5JV67SNkjlXC:bvehkac) ([raw file](https://github.com/opensim-org/opensim-core/blob/609321376207b9ff7f2c5888707ff72eb9d4bb61/OpenSim/Simulation/Model/ExternalLoads.cpp#L205)):*

Was that from a different PR? Doesn't seem related to search
</blockquote></details>

This change was also not made in the PR. Sorry for the merge issues.

aseth1:
___
*[OpenSim/Common/Storage.cpp, line 542 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/2341#-LQFdc3x54bPRhht5nv0:-LQFijTs0LvDNDKOafLa:b-1l9j1z) ([raw file](https://github.com/opensim-org/opensim-core/blob/609321376207b9ff7f2c5888707ff72eb9d4bb61/OpenSim/Common/Storage.cpp#L542)):*

The contents of the change look fine, what I'm having problem with is having this code embedded into an innocent looking function inside the Storage class.  It would be best to separate this conversion into a preprocessing step (similar to updateFromXML) then users can convert files and see the output. It would also be more general to handle future changes  to labels and/or format in one place.
If we start having document-version as part of meta data we'd also know if/when to execute this preprocessing. 
</blockquote></details>

I agree. The problem is that we do not have a data-version number so it is unclear what association rule should be applied. We could attempt to detect the data-version and do at load time. Would you like that done in this PR?

aymanhab:
Thanks for the feedback @aseth1 and @chrisdembia I'm ok with having the document version done in a separate PR but strongly recommend that we put the version number in meta data before releasing.

Since we're creating more of these files (due to data import) and the format will likely keep changing, the problem will get exponentially worse. At least both us and users can tell if the data files come from previous betas (with no document version number in meta data) or the official release of 4.0 going forward.

chrisdembia:
> strongly recommend that we put the version number in meta data before releasing.

Then I think we should add this change in this PR.

aseth1:
> Then I think we should add this change in this PR.

We should understand what we are adding first. To be clear this is not a file or Storage version. It is specifically a data (column) labeling version (which IS different the Storage version). Second, it seems pointless to me to add the versioning without making some use of it. We can add methods to try to detect the version when none is provided. That would mean we are already at least on version 3 (0, prior to 4.0; 1, 4.0 relative paths; 2.  4.0 absolute with model (with sets)).  Third we should use the versioning so that the labels are converted to the latest form before any index by labels is attempted.

I think this should be a separate issue and put in the next sprint with some pre-planning.

aymanhab:
@aseth1 As I explained above, it's just fine to have it as a separate PR. The rationale to have it here would be to actually use it. 
We already know that  storage files used up to OpenSim version 3.3 had a header that contained (version=2) so we can even be more specific and fold this fix in with less guessing, but also be ready for changes in the future. As an added benefit we can troubleshoot files created by users over the last few years easier.

Issue #2346, opened 2018-10-31
osimTableFromStruct() incorrectly index's values when new values are added


jimmyDunne:
Discovered during the virtual workshop. 

If you add values to an existing Struct, the index numbers get screwed up and an error is returned when trying to convert Struct to OpenSim table.

PR #2347, opened 2018-10-31
osimTableFromStruct() Index fix

jimmyDunne:
Fixes issue #2346

### Brief summary of changes
Revamped how the method works. Rather than trying to store the time index to use later, we store and remove the time field from the structure and do all remaining operations on the remaining (consistent) Matlab structure.

### Testing I've completed
[X] Updated Function to remove 'time' field and work on remaining struct
[X] Local test with known data
[X] Update Automated test

This should have been caught in the tests, so having a better, more thorough test, is important. I will tag reviewers when ready.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2347)

### DIFF
--- a/Bindings/Java/Matlab/Utilities/osimTableFromStruct.m
+++ b/Bindings/Java/Matlab/Utilities/osimTableFromStruct.m
@@ -1,3 +1,9 @@
+%% Convert Matlab Struct to OpenSim time Series Table
+%  Input is a Maltab stucture where data.label = nX1 or nX3 array
+%  eg s.LASI = [2 3 4; 4 5 6, ...
+%  One of the structures values MUST be a time vector and called 'time'
+%  Output is an OpenSim TimesSeriesTable
+
 % ----------------------------------------------------------------------- %
 % The OpenSim API is a toolkit for musculoskeletal modeling and           %
 % simulation. See http://opensim.stanford.edu and the NOTICE file         %
@@ -5,7 +11,7 @@
 % and supported by the US National Institutes of Health (U54 GM072970,    %
 % R24 HD065690) and by DARPA through the Warrior Web program.             %
 %                                                                         %
-% Copyright (c) 2005-2016 Stanford University and the Authors             %
+% Copyright (c) 2005-2018 Stanford University and the Authors             %
 % Author(s): James Dunne                                                  %
 %                                                                         %
 % Licensed under the Apache License, Version 2.0 (the "License");         %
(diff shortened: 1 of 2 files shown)


Bindings/Java/Matlab/Utilities/osimTableFromStruct.m
```diff
@@ -130,35 +119,21 @@
         % append the RowVectorofVec3's to the opensim table
         osimtable.appendRow(iRow-1, row);
```
chrisdembia: I don't see `row` defined here. Did you mean to convert `StdVectorVec3` into `row`?

Bindings/Java/Matlab/Utilities/osimTableFromStruct.m
```diff
@@ -20,108 +26,91 @@
...
-        for iCol = indexArray
+        for iCol = 1 : nfields 
             % get the data from the input structure
-            rowdata = structdata.(labelnames{iCol+1})(iRow,:);
+            elemtdata = s.(labels{iCol})(iRow,:);
             % make a vec3 element from the rowdata
-            elem = Vec3(rowdata(1), rowdata(2), rowdata(3));
+            elem = Vec3(elemtdata(1), elemtdata(2), elemtdata(3));
             % append the Vec3 element to the vec3 vector
             elems.add(elem);
```
chrisdembia: As-is, I think `elem` will keep growing with each row. Perhaps copy what you did earlier: set the size of `StdVectorVec3` then index into `elems`, rather than using `add()`.

Bindings/Java/Matlab/Utilities/osimTableFromStruct.m
```diff
@@ -130,35 +119,21 @@
...
+for i = 1 : nRows 
+      DataTabletime.set(i-1, time(i));
+end
+  
+%% Type change the DataTable to a TimeSeriesTable 
+if nCols == 1
+   timeseriesosimtable = TimeSeriesTable(osimtable);    
 else
-    timeseriesosimtable = TimeSeriesTable(osimtable);
+    timeseriesosimtable = TimeSeriesTableVec3(osimtable);
```
chrisdembia: Is there a reason you can't use a `TimeSeriesTable` to start with?

Bindings/Java/Matlab/Utilities/osimTableFromStruct.m
```diff
@@ -20,108 +26,91 @@
...
+tIndex = find(cellfun(@(s) contains('time', s),lower(labels)));
+% Check to see if the time exists
+if isempty(tIndex)
+    error('"time" field required, none found')
 end
-
+% Get the time vector
+time = s.time();
+% Remove time from the struct
+s = rmfield(s,'time');
```
chrisdembia: I'm not sure I understand how the previous code exhibited a bug.

Bindings/Java/Matlab/Utilities/osimTableFromStruct.m
jimmyDunne: row is defined on line 118.

Bindings/Java/Matlab/Utilities/osimTableFromStruct.m
jimmyDunne: It worked fine as long as the 'time' field was the last field in the struct. If you added more fields, I did a poor job of dealing with indexes. Also, the code was so poorly written that it took me ages to figure out exactly where the index issue was occurring, so that is a sign I should improve it. Taking the time out of the Structure here just makes everything easier.

Bindings/Java/Matlab/Utilities/osimTableFromStruct.m
jimmyDunne: This is a great suggestion, thanks.

Bindings/Java/Matlab/Utilities/osimTableFromStruct.m
jimmyDunne: I initially want to also support DataTable() type but I agree that this is not needed.

jimmyDunne:
Thanks, @chrisdembia - I made changes based on your great feedback! It is a lot cleaner now ;)

jimmyDunne:
Hahaha! Yeah, I was like 'I am pretty sure I made something previously that did this for me??'

Thanks for the Review, Chris :)

PR #2352, opened 2018-11-01
Fix OutputReporter output_paths syntax

### LINKED ISSUE #2349: OutputReporter does not parse input connectee paths correctly.
OutputReporter accepts `/com_position` and `/bodyset/femur_r/position` when these should be `/|com_position` and `/bodyset/femur_r|position`.

The fix should be straightforward; OutputReporter should use AbstractInput::parseConnecteePath().

chrisdembia:
Fixes #2349

### Brief summary of changes

- Updates the parsing in `OutputReporter` to be consistent with the syntax for input connectee paths. Instead of `/jointset/joint/coord/value`, one must now use `/jointset/joint/coord|value` or `jointset/joint/coord|value`.

### Testing I've completed

- Ran ctest.

- The syntax for input connectee paths.

### CHANGELOG.md (choose one)

- no need to update because...this class is new in 4.0.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2352)

### DIFF
--- a/Applications/Analyze/test/Tug_of_War_Setup_Analyze.xml
+++ b/Applications/Analyze/test/Tug_of_War_Setup_Analyze.xml
@@ -44,7 +44,13 @@
 				</ForceReporter>
 				<OutputReporter name="OutputReporter">
 					<!--The names of Outputs to be reported. To select specific Component Outputs, provide its path name. For example, 'slider/tx/value' is an Output for the value of a Coordinate 'tx' belonging to the Joint 'slider'.-->
-					<output_paths>forceset/LeftMuscle/fiber_length forceset/LeftMuscle/tendon_force com_position forceset/LeftMuscle/active_force_length_multiplier jointset/ground_block/reaction_on_child</output_paths>
+					<output_paths>
+						forceset/LeftMuscle|fiber_length
+						forceset/LeftMuscle|tendon_force
+						|com_position
+						forceset/LeftMuscle|active_force_length_multiplier
+						jointset/ground_block|reaction_on_child
+					</output_paths>
 					<!--Flag (true or false) specifying whether on. True by default.-->
 					<on>true</on>
 					<!--Start time.-->
--- a/OpenSim/Analyses/OutputReporter.cpp
+++ b/OpenSim/Analyses/OutputReporter.cpp
@@ -46,10 +46,13 @@ int OutputReporter::begin(const SimTK::State& s)
 
     _tableReporterDouble = new TableReporter_<double>();
     _tableReporterDouble->setName("ReporterDouble");
+    auto& inputDouble = _tableReporterDouble->updInput("inputs");
     _tableReporterVec3 = new TableReporter_<SimTK::Vec3>();
     _tableReporterVec3->setName("ReporterVec3");
+    auto& inputVec3 = _tableReporterVec3->updInput("inputs");
     _tableReporterSpatialVec = new TableReporter_<SimTK::SpatialVec>();
     _tableReporterSpatialVec->setName("ReporterSpatialVec");
+    auto& inputSpatialVec = _tableReporterSpatialVec->updInput("inputs");
 
     _pvtModel->addComponent(_tableReporterDouble.get());
     _pvtModel->addComponent(_tableReporterVec3.get());
@@ -58,31 +61,30 @@ int OutputReporter::begin(const SimTK::State& s)
     _minimumStageToRealize = SimTK::Stage::Time;
 
     for (int i = 0; i < getProperty_output_paths().size(); ++i) {
-        auto outpath = ComponentPath(get_output_paths(i));
-        // assume that Outputs are for the Model
-        string compName = _pvtModel->getAbsolutePathString();
-        // if a a ComponentPath is specified then use the parent path
-        if (outpath.getNumPathLevels()) {
-            compName = outpath.getParentPathString();
+        std::string componentPath;
+        std::string outputName;
+        std::string channelName;
+        std::string alias;
+        AbstractInput::parseConnecteePath(get_output_paths(i),
+                                          componentPath, outputName,
+                                          channelName, alias);
+        // The user supplied a path relative to the model, but that won't be a
+        // valid relative path from the reporters. Form an absolue path.
+        if (componentPath.empty() || componentPath[0] != '/') {
+            componentPath = "/" + componentPath;
         }
-        auto outputName = outpath.getComponentName();
-
-        // In keeping with assumption of Outputs for the Model (Component)
-        const Component* comp = _pvtModel.get();
-
-        // If the ComponentPath has a Component name get that Component
-        if (!compName.empty())
-            comp = &_pvtModel->getComponent(compName);
-
-        auto& out = comp->getOutput(outputName);
+        const auto path = AbstractInput::composeConnecteePath(
+                componentPath, outputName, channelName, alias);
+        const auto& comp = _pvtModel->getComponent(componentPath);
+        auto& out = comp.getOutput(outputName);
         if (out.getTypeName() == "double") {
-            _tableReporterDouble->addToReport(out);
+            inputDouble.appendConnecteePath(path);
         }
         else if (out.getTypeName() == "Vec3") {
-            _tableReporterVec3->addToReport(out);
+            inputVec3.appendConnecteePath(path);
         }
         else if (out.getTypeName() == "SpatialVec") {
-            _tableReporterSpatialVec->addToReport(out);
+            inputSpatialVec.appendConnecteePath(path);
         }
         else {
             cout << "Output '" << out.getPathName() << "' of type "
--- a/OpenSim/Analyses/OutputReporter.h
+++ b/OpenSim/Analyses/OutputReporter.h
@@ -54,6 +54,11 @@ namespace OpenSim {
  *   - `<results-file-name>Vec3.sto`, and
  *   - `<results-file-name>SpatialVec.sto`.
  *
+ * Output paths can be absolute (e.g., `/joint/slider/tx|value`) or relative
+ * to the model (by leaving off the first slash; `joint/slider/tx|value`).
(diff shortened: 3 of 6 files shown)


jenhicks:
That seems a little better :-) ... we can at least teach people that | is how you specific outputs.

chrisdembia:
Thanks @jenhicks.

Still looking for reviews on this PR.

chrisdembia:
Build failures are unrelated but I restarted the builds.

Applications/Analyze/test/Tug_of_War_Setup_Analyze.xml
```diff
@@ -44,7 +44,13 @@
 				</ForceReporter>
 				<OutputReporter name="OutputReporter">
 					<!--The names of Outputs to be reported. To select specific Component Outputs, provide its path name. For example, 'slider/tx/value' is an Output for the value of a Coordinate 'tx' belonging to the Joint 'slider'.-->
-					<output_paths>forceset/LeftMuscle/fiber_length forceset/LeftMuscle/tendon_force com_position forceset/LeftMuscle/active_force_length_multiplier jointset/ground_block/reaction_on_child</output_paths>
+					<output_paths>
+						forceset/LeftMuscle|fiber_length
+						forceset/LeftMuscle|tendon_force
+						|com_position
```
tkuchida: Is the leading vertical bar necessary here? Looks strange and seems inconsistent with "forceset/..." above which has no prefix. IMHO a more intuitive interface would be to prefix with a keyword in this case (e.g., "model|com_position") or just drop the vertical bar if there's no path.

OpenSim/Analyses/OutputReporter.cpp
```diff
@@ -58,31 +61,30 @@ int OutputReporter::begin(const SimTK::State& s)
...
-            comp = &_pvtModel->getComponent(compName);
-
-        auto& out = comp->getOutput(outputName);
+        const auto path = AbstractInput::composeConnecteePath(
+                componentPath, outputName, channelName, alias);
+        const auto& comp = _pvtModel->getComponent(componentPath);
+        auto& out = comp.getOutput(outputName);
         if (out.getTypeName() == "double") {
-            _tableReporterDouble->addToReport(out);
+            inputDouble.appendConnecteePath(path);
```
tkuchida: Might look cleaner as a `switch`.

OpenSim/Analyses/OutputReporter.h
```diff
@@ -63,8 +68,9 @@ OpenSim_DECLARE_CONCRETE_OBJECT(OutputReporter, Analysis);
 public:
 OpenSim_DECLARE_LIST_PROPERTY(output_paths, std::string,
     "The names of Outputs to be reported. To select specific Component Outputs"
-    ", provide its path name. For example, 'slider/tx/value' is an Output for "
-    "the value of a Coordinate 'tx' belonging to the Joint 'slider'.");
+    ", provide its path name. For example, '/joinset/slider/tx|value' is an "
```
tkuchida: joinset -> jointset

OpenSim/Analyses/Test/testOutputReporter.cpp
```diff
@@ -183,15 +183,17 @@ void simulateMuscle(
     // 2. OUTPUTREPORTER SETUP: create and add the OutputReporter
     //==========================================================================
     OutputReporter* outputReporter = new OutputReporter(&model);
-    outputReporter->append_output_paths("kinetic_energy");
-    outputReporter->append_output_paths("jointset/slider/tx/value");
-    outputReporter->append_output_paths("bodyset/ball/linear_velocity");
-    outputReporter->append_output_paths("bodyset/ball/angular_acceleration");
-    outputReporter->append_output_paths("bodyset/ball/acceleration");
-    outputReporter->append_output_paths("jointset/slider/reaction_on_child");
+    outputReporter->append_output_paths("/|kinetic_energy");
```
tkuchida: `/|k...` -> `k...`, or is this deliberate?

OpenSim/Analyses/Test/testOutputReporter.cpp
```diff
@@ -183,15 +183,17 @@ void simulateMuscle(
...
+    outputReporter->append_output_paths("/jointset/slider/tx|value");
+    outputReporter->append_output_paths("/bodyset/ball|linear_velocity");
+    outputReporter->append_output_paths("/bodyset/ball|angular_acceleration");
+    // Paths can also be relative to the model.
+    outputReporter->append_output_paths("bodyset/ball|acceleration");
+    outputReporter->append_output_paths("jointset/slider|reaction_on_child");
+    outputReporter->append_output_paths("|com_position");
```
tkuchida: Seems like the vertical bar should not be necessary.

chrisdembia:
The vertical bar allows us to distinguish between a component and an output. As it is now, `com_position` would specify a component and cause an error (because such a component doesn't exist and an output in that component is not specified). But we can introduce some ambiguity if the cleaner syntax is worth it.

OpenSim/Analyses/Test/testOutputReporter.cpp
chrisdembia: Deliberate.

chrisdembia: `/` is the path for the `Model`, and `kinetic_energy` is one of its outputs.

OpenSim/Analyses/OutputReporter.cpp
chrisdembia: Can't use strings in switch statements; https://stackoverflow.com/questions/650162/why-the-switch-statement-cannot-be-applied-on-strings

tkuchida: Ah ok, sorry, wrong language.

tkuchida:
> The vertical bar allows us to distinguish between a component and an output. As it is now, com_position would specify a component and cause an error (because such a component doesn't exist and an output in that component is not specified). But we can introduce some ambiguity if the cleaner syntax is worth it.

If I'm specifying an Output (e.g., in `output_paths`), then a string without slashes or bars should be interpreted as an Output, not a Body or some other Component. Seems intuitive, like a filesystem. (Granted, files in a filesystem have an extension, which could have been done for Outputs to specify their type: `forceset/LeftMuscle|fiber_length.dbl`.) Seems like a design change that would need discussion elsewhere.

OpenSim/Analyses/Test/testOutputReporter.cpp
chrisdembia: The advantage is that absolute paths are shorter and you can rename your model without needing to rename every absolute connectee path. We also had an inconsistency with state variable paths (didn't include the model name) and absolute component paths.

chrisdembia:
The extension idea is neat.

That sounds reasonable. Should we accept `/jointset/joint/coord/value` instead of `/jointset/joint/coord|value`?

OpenSim/Analyses/Test/testOutputReporter.cpp
tkuchida: I think it would be worth adding a comment in the doxygen. Users may edit in XML because they aren't comfortable programming. Something like `/|blah` isn't obvious and might send them to the documentation—and (hopefully) [this page](https://simtk.org/api_docs/opensim/api_docs/classOpenSim_1_1OutputReporter.html#a507dbb97cb33e0e858cfa521e655cf4e) where they would find only this: `"The names of Outputs to be reported. To select specific Component Outputs" ", provide its path name. For example, 'slider/tx/value' is an Output for " "the value of a Coordinate 'tx' belonging to the Joint 'slider'."`

tkuchida:
> Should we accept /jointset/joint/coord/value instead of /jointset/joint/coord|value?

Yes, I think so. If I'm entering a string like `[delim]a[delim]b[delim]c[delim]d` in `output_paths`, I think it's safe to assume the intent and `d` should be interpreted as an Output; otherwise, the string isn't a path to an Output as required by `output_paths`. If `d` actually is a Component, then the user has made an error and will get an appropriate message. I don't think forcing the user to type `|` rather than `/` is useful. IMHO paths should be parsed fine regardless of which delimiters are used (`/why\not|accept:this`)—provided these characters are prohibited in names. Note, in particular, potential confusion between forward and backward slashes.

tkuchida:
> I am definitely opposed to accepting any delimiter type since that exposes us to all sorts of future parsing nightmares if/when we decide to use some of these characters for other purposes.

(Out of curiosity, what will the difference be between `/` and `\` in the future?)

Is it possible to name a Component "value" (or the name of any other Output registered for that Component)? If so, it might not be safe to interpret the `value` in `/jointset/joint/coord/value` as an Output after all…

chrisdembia:
> Is it possible to name a Component "value" (or the name of any other Output registered for that Component)?

Yes

> (Out of curiosity, what will the difference be between `/` and `\` in the future?)

I hope we never see `\` anywhere in component paths (as delimiters or otherwise), since this is usually an escape character.

chrisdembia:
> If so, it might not be safe to interpret the value in /jointset/joint/coord/value as an Output after all…

As long as we are clear in documentation about what choice we make in the event of ambiguity (e.g., the last element is always treated as an output), this could be okay.

My preference is to have a strict format in the initial release and add leniency if we find it necessary.

tkuchida:
> I hope we never see \ anywhere in component paths (as delimiters or otherwise), since this is usually an escape character.

I would expect users to make this mistake.


> My preference is to have a strict format in the initial release and add leniency if we find it necessary.

Ok, sounds good. Hopefully the GUI will eventually make all this moot anyway.

chrisdembia:
Looking for more reviewers.

carmichaelong:
It looks like there was only a typo fixed in the comments between my last approval and now, so I'm still in the approve camp.

chrisdembia:
The `travis-ci/pr` failures are unrelated. The tests pass but the process of uploading files to Travis-CI isn't working.

chrisdembia:
@aseth1 could you merge this?

Issue #2351, opened 2018-11-01
Output data versionioning for identifying column labeling conventions


aseth1:
With the introduction of unique Component paths, column labels have become more informative but also inconsistent across OpenSim versions. This resulted in issues like #2339.  While #2341 addresses the issue temporarily, the consensus is that `Storage::getStateIndex()` is not the place to be doing adhoc translations. Instead we should convert labels of "old" files to the new labeling and have a simple/direct `Storage::getDataColumnIndex()` method.

Implementation steps:
1. define versions numbers: 0, prior to 3.3; 1, 4.0 beta version without sets; current 4.0 (with sets + without model name) 
2. add a data (label) versioning number as meta data to .sto
3. detect the data version at load time if a version is not provided
4. provide a version update scheme from previous versions to the current version
5. update labels upon loading? (or do we only want to know versioning for association with model states?)
...

tkuchida:
Regarding 4: Assuming it makes sense in this situation, perhaps try out the design we had discussed for improving `updateFromXMLNode`—namely, designing the method to update from version 0 to 1, then from 1 to 2, etc.

aymanhab:
@chrisdembia Did this get discussed at Planning or Standup? Current code returns latestVersion=1 !!

aseth1:
@aymanhab we have not discussed or addressed this issue. You are confusing the Storage version with the data column versioning being proposed above. Yes they could be related, but lets not confuse the two issues by expecting the Storage version to be magically correct for data column version that has not been defined yet.

aymanhab:
@aseth1 The reason the two are related as you point out is that the parsing has changed. The main point I want to address before 4.0 is having some indication in the data files regarding what version of the software produced them because otherwise it will be troubleshooting nightmare with files from 3.3 or older, Beta and official 4.0.

aseth1:
> because otherwise it will be troubleshooting nightmare with files from 3.3 or older, Beta and official 4.0.

We are already handling these cases. I agree we could handle it better and versioning will help as indicated in this issue. My point is that you are attempting to use the Storage version to be the versioning about labels which it was not designed to be. We can potentially use it but that is muddying the issue as described above.

I don't see this as critical for releasing 4.0.

chrisdembia:
Use opensim-core version, not XML document version.

chrisdembia:
[...]

PR #2364, opened 2018-11-08
Put OpenSim version in Storage metadata

chrisdembia:
Fixes part of issue #2351

### Brief summary of changes

- Refactor code for getting OpenSim's version (this code should be in .cpp files anyway, not in a header, where the behavior of the code depends on using the correct compiler definitions).
- Use OpenSim::GetVersion() in the DelimFileAdapter when writing a file.

### Testing I've completed

- All tests passed; ensured the OpenSim version is present in STO files we write using the STOFileAdapter.
- Made sure the version info produced by opensim-cmd and the GUI (About dialog) is still correct.

### CHANGELOG.md (choose one)

- no need to update because...minor changes

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2364)

### DIFF
--- a/OpenSim/Analyses/About.cpp
+++ b/OpenSim/Analyses/About.cpp
@@ -70,7 +70,7 @@ using namespace std;
 
 extern "C" {
 
-void opensim_version_Analyses(int* major, int* minor, int* build) {
+void opensim_version_analyses(int* major, int* minor, int* build) {
     static const char* l = "OPENSIM library="   GET_LIBRARY_STRING;
     static const char* t = "OPENSIM type="      GET_TYPE_STRING;
     static const char* d = "OPENSIM debug="     GET_DEBUG_STRING;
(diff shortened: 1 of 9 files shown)


chrisdembia:
Looking for reviews.

aymanhab:
One quclarifying question @chrisdembia, considering where the changes were made, I don't think the version will show on Storage files svaed by tools/analyses, correct? these haven't changed (maybe?)

chrisdembia:
> One clarifying question @chrisdembia, considering where the changes were made, I don't think the version will show on Storage files svaed by tools/analyses, correct? these haven't changed (maybe?)

Correct. The Storage class prints version 1 storage files and I did not want to alter the format printed by the Storage class.

aymanhab:
Thanks for confirming @chrisdembia 👍

Issue #2362, opened 2018-11-07
Marker Data: error about duplicate column labels when there are no duplicates


chrisdembia:
https://docs.google.com/spreadsheets/d/1E9Em79S5Y52hfAePmK-XcvTcPYwFWQBSchYemuWYTfE/edit#gid=1519816561

Lauren emailed Chris the TRC file that causes the issue.

chrisdembia:
![image](https://user-images.githubusercontent.com/846001/48163544-cef4ff80-e294-11e8-93e9-b106517845fb.png)


[Static1_VM.trc.txt](https://github.com/opensim-org/opensim-core/files/2559354/Static1_VM.trc.txt)

aseth1:
Do we have that data file?

aymanhab:
I think this is a pilot error based on the names of the files, the static
pose created by Scaling contains Markers even if they don't show up. Trying
to associate data with same labels could be problematic

chrisdembia:
@rwj1 , could you provide more details about the steps that caused the duplicate column labels issue?

aseth1:
The error is reproducible:
![image](https://user-images.githubusercontent.com/6494122/48169965-5f8a0a80-e2aa-11e8-8ae5-f3d46b804826.png)

@ayamnhab is correct that resultant static pose motion includes flattened marker (_tx, _ty, _tz) columns. But the problem is not pilot error: 1.) the user does not know about the experimental marker data in the pose motion file;  2) it is not displayed nor do markers show up as elements in the motion file (like individual markers appear when you associate a .trc file) in the navigator. It totally makes sense to want to associate experimental markers to compare the scaled static pose model to the experimental markers and visualize the fit and we should encourage our users to do that. 

Other weird behavior is that the user specified a static pose of 13.5 sec but the resultant motion is from 0.0. to 0.0 and the motion player bar is enable.

aymanhab:
@chrisdembia @aseth1 Some background info:

1. This has been the behavior for ages since the static pose is just the average frame used for IK.
2. Markers don't show up per earlier changes before 3.3 and because we now handle all 'motions' using formStateStorage which does not handle non-coordinates (per agreement earlier in 4.0)
3. Because the motion is just a pose with meaningless/fixed time, trying to associate a motion with it even if successful will be quite useless.
4. We suggested since 3.3 that users wanting to see the static pose markers would preview the marker data and superimpose on their own.

[...]

PR #2368, opened 2018-11-10
Do not put experimental markers in static pose file.

chrisdembia:
Fixes https://github.com/opensim-org/opensim-core/issues/2362

### Brief summary of changes

- No longer add static pose marker data to static pose motion file.

### Testing I've completed

- Built the GUI and ensured the static pose output did not contain markers and that I could successfully association the static TRC file.

image

- no need to update because...minor bugfix.


chrisdembia:
Need 1 more review.

PR #2377, opened 2018-11-15
[WIP] Add check in ModelVisualizer to only add visualizer paths if correct OpenSim version

### LINKED ISSUE #2363: ModelVisualizer: the wrong SimbodyVisualizer found on search path
In `ModelVisualizer::createVisualizer()`, a search path is created including paths in the PATH environment variable by default. When the `SimTK::Visualizer` is allocated, it will find the SimbodyVisualizer executable in the OpenSim install highest in the search path. If working with 4.0 in C++, but you have an 3.3 install higher in the path (for example, to run simulations with an older research pipeline, as I do) it will find the SimbodyVisualizer executable in the 3.3 and throw an error since it is incompatible with the visualizer protocol used in 4.0:
image

After discussing with @chrisdembia, one possible fix would be to look for the `OpenSim_buildinfo.txt` file in each path added to the `SimTK::Visualizer` search path and verify that it is the correct version of OpenSim and exclude it if otherwise.

chrisdembia:
I'm creating this PR for @nickbianco .

We should test this change a little bit more before merging. @nickbianco and I are using this change in Muscollo, so we should be able to get in some good testing there.

Fixes #2363.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2377)

### DIFF
--- a/OpenSim/Simulation/CMakeLists.txt
+++ b/OpenSim/Simulation/CMakeLists.txt
@@ -25,3 +25,5 @@ OpenSimAddLibrary(
     TESTDIRS SimbodyEngine Test
     INCLUDEDIRS ${SIMULATION_SUBDIRS}
     )
+
+set_source_files_properties(Model/ModelVisualizer.cpp PROPERTIES COMPILE_FLAGS -DOPENSIM_INSTALL_SYSCONFDIR=${CMAKE_INSTALL_SYSCONFDIR})
--- a/OpenSim/Simulation/CMakeLists.txt
+++ b/OpenSim/Simulation/CMakeLists.txt
@@ -26,4 +26,6 @@ OpenSimAddLibrary(
     INCLUDEDIRS ${SIMULATION_SUBDIRS}
     )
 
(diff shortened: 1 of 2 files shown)


OpenSim/Simulation/CMakeLists.txt
```diff
@@ -25,3 +25,5 @@ OpenSimAddLibrary(
     TESTDIRS SimbodyEngine Test
     INCLUDEDIRS ${SIMULATION_SUBDIRS}
     )
+
+set_source_files_properties(Model/ModelVisualizer.cpp PROPERTIES COMPILE_FLAGS -DOPENSIM_INSTALL_SYSCONFDIR=${CMAKE_INSTALL_SYSCONFDIR})
```
chrisdembia:
```suggestion
set_source_files_properties(Model/ModelVisualizer.cpp
    PROPERTIES COMPILE_FLAGS
    -DOPENSIM_INSTALL_SYSCONFDIR=${CMAKE_INSTALL_SYSCONFDIR})
```

nickbianco:
Closing since this is very outdated.

Issue #2394, opened 2018-12-05
FileAdapter()


jimmyDunne:
[User report](https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=9835&p=27032) poor performance of the STOFileApater() when trying to read sto files that are very large (>100mb). These large sto files are generated from OpenSim simulations that are very long (~500 sec).

tkuchida:
"Python" is tagged in the Forum post. Same issue in C++ or only via wrapping?

aseth1:
It is likely the same memory allocation/move issue we saw for C3DFileAdapter (#1698, #1730) because STOFileAdapter also appends a row one by one to the underlying matrix.

PR #2399, opened 2018-12-18
Speed up DelimFileAdapter

carmichaelong:
Fixes issue #2394 

### Brief summary of changes
Instead of appending rows one by one, which creates unnecessary overhead, `std::vectors` are first populated because they can deal with growing storage better.

### Testing I've completed
Used a 50 MB file from @aseth1 that had ~80000 rows. This previously took ~10-11 min to `read()`. This now takes 2-3 seconds.

- Are the comments sufficient?
- Where should this change be documented? CHANGELOG or elsewhere?

### DIFF
--- a/OpenSim/Common/DelimFileAdapter.h
+++ b/OpenSim/Common/DelimFileAdapter.h
@@ -319,7 +319,7 @@ DelimFileAdapter<T>::extendRead(const std::string& fileName) const {
                      FileIsEmpty,
                      fileName);
 
-    auto table = std::make_shared<TimeSeriesTable_<T>>();
+    //auto table = std::make_shared<TimeSeriesTable_<T>>();
 
     size_t line_num{};
     // All the lines until "endheader" is header.
@@ -327,6 +327,7 @@ DelimFileAdapter<T>::extendRead(const std::string& fileName) const {
     std::regex keyvalue{R"((.*)=(.*))"};
     std::string header{};
     std::string line{};
+    ValueArrayDictionary keyValuePairs;
     while(std::getline(in_stream, line)) {
         ++line_num;
 
@@ -361,7 +362,8 @@ DelimFileAdapter<T>::extendRead(const std::string& fileName) const {
                     // Discard OpenSim version number. Version number is added
                     // during writing.
                 } else {
-                    table->updTableMetaData().setValueForKey(key, value);
+                    //table->updTableMetaData().setValueForKey(key, value);
+                    keyValuePairs.setValueForKey(key, value);
                 }
                 continue;
             }
@@ -372,7 +374,8 @@ DelimFileAdapter<T>::extendRead(const std::string& fileName) const {
         else
             header += "\n" + line;
     }
-    table->updTableMetaData().setValueForKey("header", header);
+    //table->updTableMetaData().setValueForKey("header", header);
+    keyValuePairs.setValueForKey("header", header);
 
     // Callable to get the next line in form of vector of tokens.
     auto nextLine = [&] {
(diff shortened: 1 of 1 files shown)


OpenSim/Common/DelimFileAdapter.h
```diff
@@ -400,41 +403,51 @@ DelimFileAdapter<T>::extendRead(const std::string& fileName) const {
...
-        value_array.upd().push_back(SimTK::Value<std::string>{cl});
-    typename TimeSeriesTable_<T>::DependentsMetaData dep_metadata{};
-    dep_metadata.setValueArrayForKey("labels", value_array);
-    table->setDependentsMetaData(dep_metadata);
 
     // Read the rows one at a time and fill up the time column container and
     // the data container.
+    std::vector<double> timeVec;
+    std::vector<SimTK::RowVector_<T>> depDataVec;
+    int initCapacity = 2500;
```
chrisdembia: Could you add a comment for how you chose this number? I would lean towards not reserving memory initially, especially if it doesn't have a big impact on the performance test you ran.

OpenSim/Common/DelimFileAdapter.h
```diff
@@ -372,7 +374,8 @@ DelimFileAdapter<T>::extendRead(const std::string& fileName) const {
         else
             header += "\n" + line;
     }
-    table->updTableMetaData().setValueForKey("header", header);
+    //table->updTableMetaData().setValueForKey("header", header);
```
chrisdembia: I think it's fine to delete the old lines.

OpenSim/Common/DelimFileAdapter.h
```diff
@@ -400,41 +403,51 @@ DelimFileAdapter<T>::extendRead(const std::string& fileName) const {
...
+    // Create and fill up the matrix of data
+    int nrow = static_cast<int>(depDataVec.size());
+    int ncol = static_cast<int>(column_labels.size());
+    SimTK::Matrix_<T> matrix(nrow, ncol);
+    for (int iRow = 0; iRow < static_cast<int>(depDataVec.size()); ++iRow) {
+        matrix.updRow(iRow) = depDataVec[iRow];
+    }
+
+    // Create the table and update other metadata from above
+    auto& table = *new TimeSeriesTable_<T>(timeVec, matrix, column_labels);
```
chrisdembia: Please ignore this comment: Ideally, we would `std::move` the matrix into the table, but this would require changing the signature of the table constructor to take a `T&&`   or `T` intead of a `const T&`.

carmichaelong:
@chrisdembia @aseth1 
I've made another push that changes the PR more than the original scope. I wanted to note some larger changes that may seem out of scope given previous comments:
- `DelimFileAdapater.h` now only fills up a `SimTK::Matrix_` and manages the size directly rather than using a `std::vector`. This seemed to shave off another 0.5s or so (~15% faster).
- Moved a free function out of `auxiliaryTestFunctions.h` that was only being used by `MarkerData.cpp`. This meant that fewer tests had to be built every time I made changes to `DelimFileAdapter.h`. Further improvements could be had by going through tests and replacing some `#include <OpenSim.h>` lines.
- I made a change to prevent possible leaks in both `DelimFileAdapter` and `C3DFileAdapter` as per @chrisdembia's suggestions in `DelimFileAdapter`. I thought making them similar was good, but can roll back the `C3DFileAdapter` change if needed.

OpenSim/Common/DelimFileAdapter.h
```diff
@@ -361,7 +360,8 @@ DelimFileAdapter<T>::extendRead(const std::string& fileName) const {
                     // Discard OpenSim version number. Version number is added
                     // during writing.
                 } else {
-                    table->updTableMetaData().setValueForKey(key, value);
+                    //table->updTableMetaData().setValueForKey(key, value);
```
chrisdembia: Could remove this line.

OpenSim/Common/DelimFileAdapter.h
```diff
@@ -400,39 +399,60 @@ DelimFileAdapter<T>::extendRead(const std::string& fileName) const {
...
+            column_labels.size(),
+            static_cast<size_t>(row_vector.size()));
+        
+        matrix.updRow(curRow-1) = std::move(row_vector);
 
         row = nextLine();
     }
 
+    // Resize the matrix down to the correct number of rows
+    matrix.resizeKeep(curRow, ncol);
```
tkuchida: Perhaps fine for this purpose, but it may be safer to hold onto the memory that has been allocated. If the strategy of L423 is used elsewhere (i.e., double capacity when container is full), then repeatedly adding one (e.g., in a loop) would result in the following operations for a container of size *n*: double container size to *2n*, add the row, reduce container size to *n+1*, double container size, etc. I think [Issue 401 in the simbody repo](https://github.com/simbody/simbody/issues/401) is relevant. If this is indeed what would happen, perhaps add a comment in the doxygen telling users to manually preallocate (e.g., set elements to `NaN`) rather than repeatedly appending rows.

carmichaelong: > it may be safer to hold onto the memory that has been allocated

I agree, but I think because of the limitations of the `Matrix_` class as reported in issue 401, there is no separation between the `size` and `capacity` of the container. Perhaps I could add a comment to change this this logic if issue 401 is fixed (possibly by just implementing this logic in simbody)?

> repeatedly adding one (e.g., in a loop) would result in the following operations for a container of size n: double container size to 2n, add the row, reduce container size to n+1, double container size, etc.

I think this indeed would happen, however, this function is limited in scope to reading in a file. So, hopefully, a user doesn't 1) add a line, 2) then write out, 3) then read in the file, in a loop? Once it's loaded in, there's still the problem that adding any rows to a TimeSeriesTable will be slow (issue 401), but I don't think that's specific to this function. Definitely let me know if I'm missing something in the interpretation here.

carmichaelong: > I was asking about whether the allocation strategy used here is used elsewhere

Ah, I see now. AFAIK, it's not used elsewhere.

tkuchida: Ok, sounds good. Thanks- the new comments in f83a9f6 are very helpful. :+1:

chrisdembia:
Feel free to merge when tests pass.

OpenSim/Common/DelimFileAdapter.h
```diff
@@ -400,39 +399,62 @@ DelimFileAdapter<T>::extendRead(const std::string& fileName) const {
...
+    // a 50 MB file with ~80000 lines.
+    std::vector<double> timeVec;
+    int initCapacity = 100;
+    int ncol = static_cast<int>(column_labels.size());
+    timeVec.reserve(initCapacity);
+    SimTK::Matrix_<T> matrix(initCapacity, ncol);
+    
+    // Initialize current row and capacity
+    int curCapacity = initCapacity;
+    int curRow = 0;
```
aseth1: Minor, but *curRow* is not the current row but the  size, that is why  you have to set the `curRow-1`th row in the Matrix.  I found that a bit confusing.

OpenSim/Auxiliary/auxiliaryTestFunctions.h
```diff
@@ -333,26 +332,6 @@ inline bool revertToVersionNumber1(const std::string& filenameOld,
     return changedVersion;
 }
 
-// Add number of rows (nRows) and number of columns (nColumns) to the header of
-// the STO file. Note that nColumns will include time, so it will be number of
-// columns in the matrix plus 1 (for time).
-inline void addNumRowsNumColumns(const std::string& filenameOld,
```
aseth1: I think this was used to create `.sto` files in the earlier format so the the `Storage` class could read it. I just want to double check that his is no longer necessary because `Storage` now uses the `STOFileAdapter` internally, correct?

Issue #2400, opened 2018-12-19
LogManager: Make cerr & cout redirect optional


helmesjo:
Currently, the logging utility [automatically redirects all output](https://github.com/opensim-org/opensim-core/blob/70de9e7ae9f93bc7056d75779cd8532ba5cb1eff/OpenSim/Common/LogManager.cpp#L105) from `std::cerr` & `std::out` to the custom buffers, but also to files `err.log` & `out.log` respectively. 
There is no convenient way to disable this. In our case, we do not want the extra log files magically popping up in our current working directory.

Could this be changed from the hard-coded 1 to a preprocessor define?

chrisdembia:
This would be great. As a workaround, I think you can disable this programmatically in C++ via the following:

```cpp
    std::cout.rdbuf(LogManager::cout.rdbuf());
    std::cerr.rdbuf(LogManager::cerr.rdbuf());
```

helmesjo:
Unfortunately the files will be created non the less... We'll probably use a custom patch until this has been fixed.

Thanks anyways!

PR #2403, opened 2019-01-14
Make output redirect optional (out.log & err.log)

helmesjo:
Fixes issue #2400

### Brief summary of changes
Added a cmake option `OPENSIM_REDIRECT_OUTPUT` (& a define of same name added to target `osimCommon` if option == ON) to make it optional whether output (cout/cerr) should be redirect to out.log & err.log.

### Testing I've completed
Built with `OPENSIM_REDIRECT_OUTPUT=ON` & `OPENSIM_REDIRECT_OUTPUT=OFF`, using `static_assert` to verify if assertion hit.

### CHANGELOG.md (choose one)

- no need to update because the change is irrelevant to most users (and defaults to current behavior).

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2403)

### DIFF
--- a/OpenSim/Common/LogManager.cpp
+++ b/OpenSim/Common/LogManager.cpp
@@ -107,10 +107,11 @@ LogManager::LogManager()
     std::cout.rdbuf(&out);
     std::cerr.rdbuf(&err);
 
-    // Redirect output to both the terminal and an output file
+    // Redirect output to the terminal
     out.addLogCallback(new StreamLogCallback(&cout,false));
     err.addLogCallback(new StreamLogCallback(&cerr,false));
 
+    // Optional: Redirect output to file
 #if OPENSIM_LOG_TO_FILE
     out.addLogCallback(new StreamLogCallback("out.log"));
     err.addLogCallback(new StreamLogCallback("err.log"));
--- a/OpenSim/Common/LogManager.cpp
+++ b/OpenSim/Common/LogManager.cpp
@@ -102,7 +102,7 @@ sync()
 LogManager::LogManager()
 {
     // Seems to be causing crashes in the GUI... maybe a multithreading issue.
-#if 1
+#if OPENSIM_REDIRECT_OUTPUT
     // Change the underlying streambuf for the standard cout/cerr to our custom buffers
     std::cout.rdbuf(&out);
     std::cerr.rdbuf(&err);
--- a/OpenSim/Common/LogManager.cpp
+++ b/OpenSim/Common/LogManager.cpp
@@ -102,15 +102,17 @@ sync()
 LogManager::LogManager()
 {
     // Seems to be causing crashes in the GUI... maybe a multithreading issue.
-#if OPENSIM_REDIRECT_OUTPUT
+
     // Change the underlying streambuf for the standard cout/cerr to our custom buffers
     std::cout.rdbuf(&out);
     std::cerr.rdbuf(&err);
 
-    // Example setup: redirect output to both the terminal and an output file
+    // Redirect output to both the terminal and an output file
     out.addLogCallback(new StreamLogCallback(&cout,false));
-    out.addLogCallback(new StreamLogCallback("out.log"));
     err.addLogCallback(new StreamLogCallback(&cerr,false));
+
+#if OPENSIM_LOG_TO_FILE
+    out.addLogCallback(new StreamLogCallback("out.log"));
     err.addLogCallback(new StreamLogCallback("err.log"));
 #endif
 }
--- a/OpenSim/Common/CMakeLists.txt
+++ b/OpenSim/Common/CMakeLists.txt
@@ -20,12 +20,13 @@ OpenSimAddLibrary(
     TESTDIRS "Test"
     )
 
-option(OPENSIM_REDIRECT_OUTPUT
+option(OPENSIM_LOG_TO_FILE
     "Redirect output to both the terminal and an output file."
     ON)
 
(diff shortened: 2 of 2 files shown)


chrisdembia:
Thank you for contributing this.

> using `static_assert` to verify if assertion hit.

What do you mean by this? Did you verify that you still get terminal output if your option is OFF?

I would prefer that the option is handled at run-time, and could be set programmatically. But this is better than nothing.

Do you think the CMake `option` should live in the root CMakeLists.txt?

helmesjo:
> What do you mean by this? Did you verify that you still get terminal output if your option is OFF?

I basically just inserted a `static_assert(false, "fail");` [here](https://github.com/opensim-org/opensim-core/blob/670e16f6d8cf37df86443a338a7b0d58e3e5974c/OpenSim/Common/LogManager.cpp#L106) and verified it was not hit when compiling with cmake option `OPENSIM_REDIRECT_OUTPUT=OFF` (and that it was with `OPENSIM_REDIRECT_OUTPUT=ON`).

With the changes above I (very naively) just continued on the original code and skip the entire "redirect logic". So `std::cout` will (of course) work and output to terminal, but internal calls to [`LogManager::log(...)`](https://github.com/opensim-org/opensim-core/blob/1130711c12606517f0adce971036ebf230d150a0/OpenSim/Common/LogManager.cpp#L52) will not. Also, listeners [won't receive logs from `std::cout` or `std::cin`](https://github.com/opensim-org/opensim-core/blob/670e16f6d8cf37df86443a338a7b0d58e3e5974c/OpenSim/Common/LogManager.cpp#L93).

Perhaps this PR should be changed to only enable/disable redirect to file? Something like:

```
    // Change the underlying streambuf for the standard cout/cerr to our custom buffers
    std::cout.rdbuf(&out);
    std::cerr.rdbuf(&err);

    // Redirect output to the terminal
    out.addLogCallback(new StreamLogCallback(&cout,false));
    err.addLogCallback(new StreamLogCallback(&cerr,false));

#if OPENSIM_LOG_TO_FILE
    // Redirect output to file
    out.addLogCallback(new StreamLogCallback("out.log"));
    err.addLogCallback(new StreamLogCallback("err.log"));
#endif
```

This way it will not affect the current (internal) behavior. Actually this might be wiser, given that there is [more stuff](https://github.com/opensim-org/opensim-core/blob/1130711c12606517f0adce971036ebf230d150a0/OpenSim/Common/LogManager.cpp#L30-L31) going on in this class.

> I would prefer that the option is handled at run-time, and could be set programmatically. But this is better than nothing.

Yeah, I agree, but that would of course up the complexity of this change rather drastically. I guess this could be considered a good start (in our case we just want to get rid of the log-files that appear "magically").

> Do you think the CMake `option` should live in the root CMakeLists.txt?

Personally I'm against putting all options in the root, since then it's so disconnected from where it is actually used (and when you reorganize files the option might get lost all together). But consistency is more important than personal preference of course...

chrisdembia:
I agree about your proposed change (OPENSIM_LOG_TO_FILE). Please test to ensure that the output is still sent to the terminal.

Thanks for explaining the placement of ‘option’. It’s fine where it is.

helmesjo:
> I agree about your proposed change (OPENSIM_LOG_TO_FILE). Please test to ensure that the output is still sent to the terminal.

Great, I'll make this change (and confirm output is sent to terminal)!

chrisdembia:
Thank you, Fred.

helmesjo:
Option renamed to `OPENSIM_LOG_TO_FILE` and it now only makes redirection to file optional (redirection to terminal stays as before), and I confirmed output was still sent to the terminal with both `OPENSIM_LOG_TO_FILE=ON` & `OPENSIM_LOG_TO_FILE=OFF`.

`LogManager::getInstance()->out << "Test\n"; // Prints to terminal`
`LogManager::getInstance()->cerr << "Test\n"; // Prints to terminal`

**Note:** 
It appears that redirection to file does not (and did not) work. Files are created but remain empty.
Adding breakpoint in `StreamLogCallback::log`. Perhaps another issue should be created for that...

chrisdembia:
What if you use `std::cout` and `std::cerr`? Is output still sent to the terminal in both cases? We never use `LogManager::getInstance()->out` in the code base.

helmesjo:
Sorry for late response!

Yes, I just confirmed that both `std::cout` & `std::cerr` send output to terminal with `OPENSIM_LOG_TO_FILE=ON`.

helmesjo:
What is the status here? :)

chrisdembia:
@aymanhab could you review this?

OpenSim/Common/LogManager.cpp
```diff
@@ -102,15 +102,17 @@ sync()
...
     // Seems to be causing crashes in the GUI... maybe a multithreading issue.
-#if 1
+
     // Change the underlying streambuf for the standard cout/cerr to our custom buffers
     std::cout.rdbuf(&out);
     std::cerr.rdbuf(&err);
 
-    // Example setup: redirect output to both the terminal and an output file
+    // Redirect output to both the terminal and an output file
     out.addLogCallback(new StreamLogCallback(&cout,false));
```
aymanhab: The wording in the comment 'both' is a bit misleading since it's conditional now.

helmesjo:
Done!

helmesjo:
Could this be merged & closed? :)

tkuchida:
Travis failures are unrelated; merging.

Issue #500, opened 2015-06-17
Use travis to put doxygen somewhere online to help review of PR's


chrisdembia:
@aseth1 mentioned that this might be possible. He suggested using the travis build number to create a directory on some website, and we can regularly prune the site so only the last 5 or so builds are online.

He suggested making a `travis` user that can log into the website.

jenhicks:
Fixed by PR #629

PR #2404, opened 2019-01-21
Static Optimisation using muscle equilibrium

pariterre:
Fixes issue #500

### Brief summary of changes
Static optimisation (without linearization) now solves for muscle equilibrium when calculating the acceleration. So instead of solving the "realize acceleration" via a an infinite tendon stiffness assumption, it now solves for the proper equilibrium. 

### Testing I've completed
None, this breaks the previous values that were considered as reference. 
In that sense, I will need your opinion on what I've done. 

### CHANGELOG.md (choose one)

- added the following mention sentence into the "Other changes" section: 
- Static optimization now solves for muscle equilibrium. 

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2404)


pariterre:
Up?

aymanhab:
@aseth1 Can you comment on whether this functionality needs to be supported by mainstream OpenSim rather than as a plugin? 
@pariterre can you say whether this would change previous results of staticOptimization and if so how?

pariterre:
@aymanhab 
Thanks for you comment! It would definitely change previous results as it computes for muscle equilibrium unlike the previous implementation. 

The idea is to linearize at the point of previous activation instead of full activation. My implementation doesn't fully mimic what happens if you don't linearize at all (that is commenting the line that defines USE_LINEAR_CONSTRAINT_MATRIX), but is way closer than the actual linearization. For instance, for the arm26, the actual implementation shows differences up to 30% while this new linearization method show differences up to 2%. 

As an outside project, I've also coded a way to use the non-linearize method by approaching the constraint jacobian by series of cublic spline. It therefore constructs the Jacobian at the beginning of the optimization and then approximate it using the derivative of the spline. Using this method, I get almost no difference with the non-linearized method. However, I takes about twice the time as the linearized method (which is still fast). I did not intend to add this method to OpenSim though, unless you guys think that it would make sense too..

Thanks!

pariterre:
Up?

pariterre:
@aymanhab 
I fast-forwarded the PR to fit the master
thanks :)

chrisdembia:
@jimmyDunne I'm happy to help review this after discussing with the dev team what we want to do.

jenhicks:
@pariterre Thank you for making this (and other) pull requests! It is great to have contributions from the community. Since this is a pretty large addition to the code, could we set-up a time to talk via Skype about how best to incorporate this work? You can email us at opensim@stanford.edu and we can arrange a time to chat. Thanks again!

OpenSim/Analyses/StaticOptimization.cpp
```diff
@@ -589,8 +587,18 @@ int StaticOptimization::begin(const SimTK::State& s )
...
 
+        // Set initial guess to lowest activation
         _parameters.resize(_modelWorkingCopy->getNumControls());
-        _parameters = 0;
+        const Set<Actuator>& fs = _modelWorkingCopy->getActuators();
+        for(int i=0,j=0;i<fs.getSize();i++) {
+            ScalarActuator* act = dynamic_cast<ScalarActuator*>(&fs.get(i));
+            if (act) {
+                if (act->getMinControl() != -INFINITY)
+                    _parameters[j++] = act->getMaxControl();
```
jenhicks: Do you mean act ->get**Min**Control()?

pariterre: Actually, the error is in the if. Unfortunately, contrary to what is previously done, the initial guess has on the first frame has to be maxControl. The reason is that I linearize between 0 (actually MinControl) and value from previous frame. If we set the previous frame to be minimal, it is a singularity. Setting it to max solves to the same value as the previous linear version : the problem being quadratic when p=2 and therefore the solver finds a good optimal solution even though I initialize at max.

pariterre:
Up? :)

nickbianco:
Closing: stale PR.

PR #2405, opened 2019-01-22
Python Static Optimization

pariterre:
New feature

### Brief summary of changes
This is a Python implementation of the Static Optimization. One can use this example as template to perform their own static optimization in Python. 

### Testing I've completed
This gives similar results as the implementation of the Static Optimization that uses muscle equilibrium

### CHANGELOG.md (choose one)

- updated the python interface section with the following:
- Added an example of Static optimization performed in Python 3.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2405)


aymanhab:
Thanks for this valuable contribution. Will review and respond on the PR. :+1

pariterre:
I was finishing to implemente an improved linear approximation of the constraints in the StaticOptimization of the C++ code, when I noticed that sometimes, for no apparent reason, the muscles equilibrium failed. The work-around (in C++ code) is basically to modify ever so slightly the values of the activations (+0.1%) hoping this solves the equilibrium issues. 

However, since it throws an error if it fails, right now it is impossible to do the same trick in Python (right now, the C++ throw by-passes the try-catch of python, crashing anyway). Is there a way to "try-catch" this error in python? If it is not possible, I don't think it is useful in any way to implement a full in-depth example of the Static Optimization in Python as the one proposed in this PR since the chances of success are very slim!

pariterre:
Or, if I am allowed, I would add the exceptions in the swig files so we can catch them :)

nickbianco:
Closing: stale PR.

PR #2412, opened 2019-02-12
Add Orientation tracking to InverseKinematicsSolver

aseth1:
Fixes issue https://github.com/opensim-org/sprints/issues/11

### Brief summary of changes
1. Add the `OrientatationsReference` class to associate orientation data (e.g. from IMUs) with a PhysicalFrame in the model and assign relative weightings (similar to `MarkersReference`)
2. Add support for `TimeSeriesTable_<Rotation>` and building tables from `Quaternion` storage/csv data files.
3. Make `InverseKinematicsSolver` operate on references that are const.
4. Provide test cases that Orientation-based InverseKinematics is working correctly based on known IK solution and computed `Body` orientations.

### Testing I've completed
1. `testOrientationsReference` verifies that the sensor is tracked and that weightings have the expected behavior.
2. `testNumberOfOrientationsMismatch` verifies that order of IMUs and NaN values do not cause failures.
3.  In `testInverseKinematicsSolverWithOrientations` verifies that orientation-based IK runs and animates based on orientations computed from the model and previous (marker-based) IK solution.
4.  `testInverseKinematicsSolverWithEulerAnglesFromFile` reruns IK but with orientations stored as a storage file of Euler angles (internally computes Rotations) and verifies that resultant joint angles are within  0.1 degrees RMS.

- any inefficiencies or anticipated problems with the `InverseKinematicsSolver` interface?
- test coverage.

### CHANGELOG.md (choose one)
- to be updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2412)


aseth1:
@aymanhab @chrisdembia this PR is ready for review. Remaining build failures are unrelated.

aseth1:
___
*[OpenSim/Simulation/OrientationsReference.h, line 111 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2412#-LYZ2sLlBhGdNl1ZRS8u:-LZpWTdw1wd0lUN4bE_m:b1m0i47) ([raw file](https://github.com/opensim-org/opensim-core/blob/7edc3a5f717f99b5b5afd1eb439346dc4ce7f126/OpenSim/Simulation/OrientationsReference.h#L111)):*

Is this tested?
</blockquote></details>

No. I will remove with the intention of either removing the file loading task (including file of Euler angles) from `OrientationsReference` or having a single utility to read data from multiple sources, since we can distinguish them internally. The [`InverseKinematicsStudy`](https://github.com/opensim-org/BananaPhone/pull/17) already has a more efficient quaternion loading capability.

aseth1:
___
*[Applications/IK/test/testIK.cpp, line 351 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2412#-LZsAzBI-7OvMWOBRmfC:-LZunpI_2SxPpsGBMfru:b-zspad) ([raw file](https://github.com/opensim-org/opensim-core/blob/f27934c1971e02a7b2fe8583d78bf12b3315fa09/Applications/IK/test/testIK.cpp#L351)):*

This is somewhat fragile. It would be preferable to check for column labels containing "pelvis".
</blockquote></details>

I agree. I fixed that and also realized that my RMSE calculation was wrong. I needed to take the mean of the squared norm across time points not the number of coordinates. Both have been fixed.

aseth1:
Thanks @chrisdembia for your thorough review!

OpenSim/Common/Object.cpp
```diff
@@ -1647,19 +1647,17 @@ void Object::updateFromXMLDocument()
     IO::chDir(saveWorkingDirectory);
 }
 
-std::string Object::dump(bool dumpName) {
+std::string Object::dump() const {
     SimTK::String outString;
```
aymanhab: While this change could be reasonable by itself, folding it into this PR is unnecessary and will be confusing. Also breaks backward compatibility.

aseth1:
___
*[OpenSim/Common/Object.cpp, line 1651 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2412#-LYUJIhnuFJ3jbQYGLlB-r3-1651:-L_9EnXLDkrV4wXVPyql:b-1iks4k) ([raw file](https://github.com/opensim-org/opensim-core/blob/26494c27351102089b7322ea88031932fad6b8f0/OpenSim/Common/Object.cpp#L1651)):*

While this change could be reasonable by itself, folding it into this PR is unnecessary and will be confusing. Also breaks backward compatibility.
</blockquote></details>

I agree that ideally this change would not be present. However it was necessary to debug and is part of the testing. It was simply wrong to have dump change the state of the Object. If you insist, I can cut this out into another PR, merge that one first and then this PR.

aseth1:
___
*[Applications/IK/test/testIK.cpp, line 351 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2412#-LZsAzBI-7OvMWOBRmfC:-L_96FHk6O9bo9GmYBm0:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/f27934c1971e02a7b2fe8583d78bf12b3315fa09/Applications/IK/test/testIK.cpp#L351)):*

I agree. I fixed that and also realized that my RMSE calculation was wrong. I needed to take the mean of the squared norm across time points not the number of coordinates. Both have been fixed.
</blockquote></details>

Done.

aseth1:
___
*[CHANGELOG.md, line 195 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2412#-L_93lvS3mUIhcbqO1la:-L_9iE3hDSZPbYfVOver:bkzui1w) ([raw file](https://github.com/opensim-org/opensim-core/blob/26494c27351102089b7322ea88031932fad6b8f0/CHANGELOG.md#L195)):*

If we change the signature of the Object::dump method, we should include that as it  potentially breaks client code
</blockquote></details>

Added not to CHANGELOG.MD

aseth1:
___
*[Applications/IK/test/testIK.cpp, line 224 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2412#-L_94QNk6lxUfDK2naBL:-L_9iKvc8FKYh5g5ZVRR:b-vozqbo) ([raw file](https://github.com/opensim-org/opensim-core/blob/26494c27351102089b7322ea88031932fad6b8f0/Applications/IK/test/testIK.cpp#L224)):*

The following two methods are used in the test but can be of a more general use, suggest moving them somewhere else for re-use.
</blockquote></details>

I was thinking the same thing when I wrote them, but I couldn't think of a good place to move them, yet. I'm happy to move them when they're needed or we can plan to have a general set of OpenSense or other appropriately name set of Utilities.

aseth1:
Thanks @aymanhab and @chrisdembia. Merging ...

PR #2437, opened 2019-04-18
Opensense command line tool

aymanhab:
Fixes issue #<issue_number>

### Brief summary of changes
Migrate OpenSense support from private repo/branch into opensim-core
Changes from branch/earlier version of the code that need review:

1. Moved files from standalone executable to a folder OpenSense under OpenSim/Simulation. This was a compromise to group relevant new code without the overhead in build/distribution.
2. Made OpenSenseUtilities a class and removed the OpenSense namespace. free functions in a name space are not easily supported cross languages/platforms and seems unjustified at the moment.
3. Fixed some  compilation errors that should be clear from commits
4. Replaced std::stuple with SimTK::Array_<int> across interface as not supported by SWIG and a new thing to learn for users. Need to discuss default/iinitialization
5. Moved methods inside OpenSenseUtilities class convertQuaternionsToRotations. calibrateModelFromOrientations.
6. Made preview method of InverseKinematicsStudy private, removed custom registerTypes
7. Included proper headers in SWIG bindings so functionality would be available through Matlab/Python.

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2437)


aymanhab:
@chrisdembia This is mainly a heads up, I'm still working on the test but could use another pair of 👀

aymanhab:
Build failure unrelated to this PR. Feel free to start reviewing @aseth1 @chrisdembia

aymanhab:
@jimmyDunne You can checkout the branch and build in your own environment to start.  The test case testOpensense should give you an idea what the functionality is as was exercised in the demo (1. calibrate model, 2. solve IK with orientations). The data files under Applications/opensense/test are a good place to look into. If you prefer to see visualization, change the hard-coded value at InverseKinematicsStudy.cpp:line 236 from false to true and recompile and the test will show the animation then hang waiting for you to close it.

jimmyDunne:
I was able to run this in Matlab (code below as an example). It all seemed to work as expected— a calibrated model was created and IK was performed. 

```
%% Instantiate an OpenSenseUtilities object
opensenseUtility = OpenSenseUtilities();
% Generate a calibrated model
opensenseUtility.calibrateModelFromOrientations(modelpath, imupath);

%% Run IK using the calibrated model and the IMU Orientations
% Setup the tool
ik = InverseKinematicsStudy(ikSetupPath);
% Run IK
ik.run()
```
One request— if we could make the visualization an option rather than hardcoded? I did make the change in InverseKinematicsStudy.cpp, but the rebuild took half an hour on my machine. I think it would be useful to have it as an option on InverseKinematicsStudy()

aseth1:
___
*[Applications/opensense/opensense.cpp, line 226 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LdQiAeo7yB7Y8x6HeJc:-LdQiAep3QhprYzkO4kP:beavl3a) ([raw file](https://github.com/opensim-org/opensim-core/blob/4433c86f5fbeda9fa8b9fd72d26a80f7229e226c/Applications/opensense/opensense.cpp#L226)):*
> ```cpp
>     aOStream << "-PropertyInfo, -PI                  Print help information for properties in setup files.\n";
>     aOStream << "-ReadX, -RX  directory settingsFile.xml   Parse Xsens exported files from directory using settingsFile.xml.\n";
>     aOStream << "-ReadA, -RA  datafile.csv settingsFile.xml   Parse APDM exported files from directory using settingsFile.xml.\n";
> ```

Comment should be updated to reflect that APDM data is a from a single csv file.

aseth1:
___
*[Applications/opensense/test/testOpensense.cpp, line 47 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LdQkx9C7wQATsNcPP4g:-LdQkx9D5z82kRLWB0ih:bpfu16i) ([raw file](https://github.com/opensim-org/opensim-core/blob/4433c86f5fbeda9fa8b9fd72d26a80f7229e226c/Applications/opensense/test/testOpensense.cpp#L47)):*
> ```cpp
>     InverseKinematicsStudy ik("setup_IMUtracking.xml");
>     // RUN
>     ik.run(); 
> ```

Should the test also check that the IMU-based tracking is within tolerance of Marker-based?

aymanhab:
Thanks James, update and you will see the option added to run method.

aymanhab:
@chrisdembia Any idea while a model file created by a test case opens on windows and some travis builds but fail to open on others? Thanks

Bindings/Java/tests/TestXsensDataReader.java
```diff
@@ -0,0 +1,32 @@
...
+        XsensDataReaderSettings settings = new XsensDataReaderSettings();
+        ExperimentalSensor  nextSensor = new ExperimentalSensor("000_00B421AF",
+            "shank");
+        settings.append_ExperimentalSensors(nextSensor);
+        settings.set_trial_prefix(0, "MT_012005D6_031-");
+
+        // Test costruct XsensDataReader from a XsensDataReaderSettings object 
+        XsensDataReader xsensDataReader = new XsensDataReader(settings);
+        // Make sure types returned by the xsensDataReader are usable in Java
+        StdMapStringAbstractDataTable tables = xsensDataReader.extendRead("");
```
aseth1: `extendRead` should not be part of the Reader's public interface.

OpenSim/Common/APDMDataReader.h
```diff
@@ -0,0 +1,110 @@
...
+    list of tables depending on the contents of the file read. 
+    - One table for rotations, 
+    - one for LinearAccelerations
+    - one for MagneticHeading data, 
+    - one for AngularVelocity data. 
+    - Barometer and Temperature data is ignored for now
+     
+     @see IMUDataReader class for utilities to extract/access specific table(s)
+    */
+    DataAdapter::OutputTables extendRead(const std::string& fileName) const override;
```
aseth1: This should extend the public `read` of the `DataAdapter` it should be called internally, not through the public interface.

OpenSim/Common/Test/testAPDMDataReader.cpp
```diff
@@ -0,0 +1,106 @@
...
+        // Programmatically add items to name mapping, write to xml
+        for (int index = 0; index < imu_names.size(); ++index) {
+            ExperimentalSensor  nextSensor(names_in_experiment[index], imu_names[index]);
+            readerSettings.append_ExperimentalSensors(nextSensor);
+        }
+        readerSettings.print("apdm_reader.xml");
+        // read xml we wrote into a new APDMDataReader to readTrial
+        APDMDataReaderSettings reconstructFromXML("apdm_reader.xml");
+        APDMDataReader reader(reconstructFromXML);
+        DataAdapter::OutputTables tables = reader.extendRead("imuData01csv.csv");
```
aseth1: This should be `reader.read("...")`

Applications/opensense/opensense.cpp
```diff
@@ -231,7 +231,7 @@ TimeSeriesTable_<SimTK::Quaternion> readRotationsFromXSensFiles(const std::strin
 {
     XsensDataReaderSettings readerSettings(readerSetupFile);
     XsensDataReader reader(readerSettings);
-    DataAdapter::OutputTables tables = reader.extendRead(directory);
```
aseth1: I think the initial design was OK but incomplete. The base `DataAdapter` should provide the public interface `read(const string& sourceIdenitifier)` and delegate to the virtual method implemented by concrete readers. That design enables us to put common operations (checks) in the base `read` method while delegating details to each individual reader. I would not simply rename `extendRead` here.

Applications/opensense/opensense.cpp
```diff
@@ -0,0 +1,499 @@
...
+                    std::string settingsFile{ argv[i + 2] };
+                    TimeSeriesTable_<SimTK::Quaternion> rotationsTable =
+                        readRotationsFromAPDMFile(dataFile, settingsFile);
+                    cout << "Done." << endl;
+                    return 0;
+                }
+                else if ((option == "-Transform") || (option == "-T")) {
+                    std::string markerFile{ argv[i + 1] };
+                    TimeSeriesTable_<SimTK::Quaternion> quaternions =
+                        createOrientationsFileFromMarkers(markerFile);
```
aseth1: Should be added to `OpenSenseUtilities` and the writing of the quaternions to file should NOT be baked into the method. Instead the writing to file is a convenience we can add here.

aseth1:
___
*[Applications/opensense/opensense.cpp, line 226 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LdQih0uDkKL4fZNzxk1:-LduA0LjD3lGTcYJkyFC:b40xne3) ([raw file](https://github.com/opensim-org/opensim-core/blob/4433c86f5fbeda9fa8b9fd72d26a80f7229e226c/Applications/opensense/opensense.cpp#L226)):*

Can you provide the text, @aseth1 ? Thanks
</blockquote></details>

`
-Transform <markerFileWithIMUframes>.trc    Transform experimental marker locations that define axes of IMUs, or the plates upon which they are rigidly affixed, into the orientations of the IMUs expressed in the motion capture (markers) lab frame. The orientations over the trial are written out as quaternions to  <markerFileWithIMUframes>_quaternions.sto. Markers on IMUs (or plates) are assumed to be labeled '<base>_IMU_O',  '<base>_IMU_X',  '<base>_IMU_Y', '<base>_IMU_D', where <base> is the label of the IMU, and O, X, Y and D denote the origin, x -axis direction, y-axis direction and diagonal, respectively. The resulting quaternions file as <base>_IMU as its column labels.
`

`
-AddIMUs  <modelFile>.osim  <markerFileWithIMUframes>.trc  Add IMUs to the provided model based on marker data. Marker data with the naming convention of '<base>_IMU_O',  '<base>_IMU_X',  '<base>_IMU_Y', '<base>_IMU_D' are used to compute the location and orientation the IMU frame which is then affixed to the same "base" segment (frame) to which the markers are attached.  Before the IMU frames are attached to the model, the model is posed according to marker-based IK.
`

`
-Calibrate modelCalibrationPoseFile.osim <calibrationPoseOrientationsFile>.sto  Calibrate the <modelCalibrationPoseFile>.osim model by registering IMU frames whose orientations  in the sensor world frame are specified in <calibrationPoseOrientationsFile>.sto. and assuming the model's default pose is the calibration pose. The resultant model with IMU frames registered is written to file as calibrated_<modelCalibrationPoseFile>.osim.
`

aymanhab:
@aseth1 I believe most issues from your feedback have been addressed. One minor complication is that the name "read" for base class public interface has been overused by the earlier DataAdapter subclasses, which caused compilation failures on gcc, clang. I renamed the new public method to readSource to get around that but I'm open to better names as I'd prefer not to break existing code that uses "FileAdapter::read"

aymanhab:
@chrisdembia would you be able to review this PR? Thanks

aseth1:
@aymanhab here is data for marker-based IK and should be the corresponding orientations data from the IMUs for subject 03 for the hip joint center trial.
[markerBasedIKS03.zip](https://github.com/opensim-org/opensim-core/files/3149961/markerBasedIKS03.zip)

OpenSim/Common/APDMDataReader.cpp
```diff
@@ -67,21 +67,25 @@ APDMDataReader::extendRead(const std::string& fileName) const {
     bool newFormat = false;
     if (tokens[0] == "Format=7") {
         newFormat = true;
-        std::vector<std::string> labels;
+        dataRate = 128; // Will fix on reading first 2 rows
```
aseth1: Why is this initialized to 128 instead of NaN or some other way to detect an error?

OpenSim/Common/APDMDataReader.cpp
```diff
@@ -67,21 +67,25 @@ APDMDataReader::extendRead(const std::string& fileName) const {
...
         // Header Line 1:Format=7, [I1,,,$IMU1,,,,,,,,,,,]*
         // Header Line 2: Time,[Accelerometer,,,Gyroscope,,,Magnetometer,,,Barometer,Orientation,,,]*
         // Header Line 3: ,[X,Y,Z,X,Y,Z,X,Y,Z,,S,X,Y,Z]*
+        std::vector<std::string> mtLabels;
```
aseth1: i don't understand the reason for the name change. What does "mt" signify?

OpenSim/Common/APDMDataReader.cpp
```diff
@@ -119,8 +123,8 @@ APDMDataReader::extendRead(const std::string& fileName) const {
     bool foundMagneticHeadingData = magIndex.size()>0;
     bool foundAngularVelocityData = gyroIndex.size()>0;
 
-    // If no Orientation data is available or dataRate can't be deduced we'll abort completely
-    OPENSIM_THROW_IF((orientationsIndex.size() == 0 || SimTK::isNaN(dataRate)),
+    // If no Orientation data is available we'll abort
```
aseth1: It should be legitimate IMU file that does not have orientation data but has linear acceleration, angular velocity and/or magnetic heading

OpenSim/Common/APDMDataReader.cpp
aymanhab: @aseth1 the old format had an explicit datarate line, new format has none but provide timestamp as a longlong integer. We'll need to compute data rate from actual timestamps of the first two rows of data, so this part is WIP.

aseth1:
___
*[OpenSim/Common/XsensDataReader.h, line 72 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LeELCLuBJPaAmAQ3Y2O:-LeSn6Ag9UdTHucR6Lh2:b-ch809y) ([raw file](https://github.com/opensim-org/opensim-core/blob/2371de2d8cc4d7a5bb8e21e7577282d4e6aea0dc/OpenSim/Common/XsensDataReader.h#L72)):*

Pure virtual DataAdapter::extendRead dates back to 2016 so I need to implement it , this has been the crux of all readers and so this is followed in the new readers.  If you have ideas to change the design that you described as incomplete earlier please advise or just do it and I'll review.
</blockquote></details>

I agree extendRead needs to be implemented-- that is the interface. I do not agree with the new virtual `readSource` that pollutes the interface and violates the previous design of base public `read()` invokes hidden `virtual extendRead()` implemented by concrete readers.

OpenSim/Common/APDMDataReader.cpp
aymanhab: mtLabels stands for EmptyLabels. Old format had suffixes for various quanities e.g. Lumbar/Acceleration, new one has Lumber on one line and Acceleration on a different line, using an array with one entry that is a n empty string allows us to reuse the function to locate the column number corresponding to a specific sensor. If it makes it more clear I can rename it to emptyLabels.

OpenSim/Common/APDMDataReader.cpp
aymanhab: @aseth1 In the two examples format we have we always had orientations while other data was missing, in fact orientation is needed for downstream usage AFAIK. If you prefer we don't abort here and abort downstream on demand let me know.

aseth1:
___
*[Bindings/Java/tests/TestXsensDataReader.java, line 17 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LcmQY6Iub7tc_0GgdDn-r5-17:-LeSoUce6TcLQ916hHXR:b-f1vqdp) ([raw file](https://github.com/opensim-org/opensim-core/blob/b4779a2669673626b97a0a97600e22fc92d321ab/Bindings/Java/tests/TestXsensDataReader.java#L17)):*

The code does not compile on gcc and clang because there's a static read method with same signature, you can't have a virtual and static method with same signature. Changing the static method name will break existing code. Sorry if that wasn't clear.
</blockquote></details>

The static reads were for convenience. It would not violate the design if the concrete class had a static `readFile` that invoked `extendRead()`. The point is that we should be able to use the base interface to `read()` ` Tables` from any `const std::string& source_identifier`.

aseth1:
___
*[OpenSim/Common/APDMDataReader.cpp, line 70 at r8](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LcmQY6Iub7tc_0GgdDo-r8-70:-LeSpGLV9NQHJ7vaM-O1:b-wh5gdu) ([raw file](https://github.com/opensim-org/opensim-core/blob/489f61507bdea6706cd53b0cb2374391b8fefec3/OpenSim/Common/APDMDataReader.cpp#L70)):*

@aseth1 the old format had an explicit datarate line, new format has none but provide timestamp as a longlong integer. We'll need to compute data rate from actual timestamps of the first two rows of data, so this part is WIP.
</blockquote></details>

OK, I would still initialize to NaN or -1 if integer to indicate that it is uninitialized.

aseth1:
___
*[OpenSim/Common/APDMDataReader.cpp, line 74 at r8](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LcmQY6Iub7tc_0GgdDo-r8-74:-LeSpYGhCUfi8jP_WR0Z:bogpf3h) ([raw file](https://github.com/opensim-org/opensim-core/blob/489f61507bdea6706cd53b0cb2374391b8fefec3/OpenSim/Common/APDMDataReader.cpp#L74)):*

mtLabels stands for EmptyLabels. Old format had suffixes for various quanities e.g. Lumbar/Acceleration, new one has Lumber on one line and Acceleration on a different line, using an array with one entry that is a n empty string allows us to reuse the function to locate the column number corresponding to a specific sensor. If it makes it more clear I can rename it to emptyLabels.
</blockquote></details>

Haha, OK, yeah "empty" would be more obvious. But I don't understand how it helps you locate the IMU column. Shouldn't the array of IMU labels be sufficient to find the IMY column? Shouldn't you have a second array of sensor Type?

aseth1:
___
*[OpenSim/Common/APDMDataReader.cpp, line 126 at r8](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LcmQY6Iub7tc_0GgdDo-r8-126:-LeSrmcz9KXAQOnLW4gT:bh6g6jj) ([raw file](https://github.com/opensim-org/opensim-core/blob/489f61507bdea6706cd53b0cb2374391b8fefec3/OpenSim/Common/APDMDataReader.cpp#L126)):*

@aseth1 In the two examples format we have we always had orientations while other data was missing, in fact orientation is needed for downstream usage AFAIK. If you prefer we don't abort here and abort downstream on demand let me know.
</blockquote></details>

We definitely want to support NOT reading orientations. These come from onboard data fusion. We will want to give users the ability to fuse their own sensor data to estimate orientations using their own or provided algorithms.

aseth1:
___
*[OpenSim/Common/APDMDataReader.cpp, line 118 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LdQmVp69oCZf34HrP84:-LeSs7KT9v6d5FruQf2f:b-awbjv0) ([raw file](https://github.com/opensim-org/opensim-core/blob/4433c86f5fbeda9fa8b9fd72d26a80f7229e226c/OpenSim/Common/APDMDataReader.cpp#L118)):*

No, the issue is with initialization. We have constructors for a single row/vector that can initialize to NaN (efficiently I assume) then we don't need to do anything for unpopulated entries. if we initialize the whole matrix to NaN it will be less efficient and time consuming (since some entries could end up being discarded and the cost will grow with each doubling). If there's an efficient way to initialize a row to NaN without copying then please advise.
</blockquote></details>

Initialization seems like conjecture to me. Have you timed how long it takes to initialize a matrix of NaNs versus repeatedly creating RowVectors and then copying those values into the matrix? I would think allocating memory in a loop eats more time.

aseth1:
___
*[OpenSim/Tests/shared/IMUData/APDM/apdm_format2_short.csv, line 1 at r8](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LeStBp_1MdvnQD3z3dE:-LeStBp_1MdvnQD3z3dF:b-6jgtbr) ([raw file](https://github.com/opensim-org/opensim-core/blob/489f61507bdea6706cd53b0cb2374391b8fefec3/OpenSim/Tests/shared/IMUData/APDM/apdm_format2_short.csv#L1)):*
> ```CSV
> Format=7,5718,,,Lumbar,,,,,,,,,,,5778,,,Right Ankle,,,,,,,,,,,5811,,,Left Ankle,,,,,,,,,,,6570,,,Left Foot,,,,,,,,,,,6575,,,Chest,,,,,,,,,,,6833,,,Right Foot,,,,,,,,,,
> ```

Maybe relabel the file "APDM_Format7..."?

aseth1:
___
*[OpenSim/Common/APDMDataReader.h, line 107 at r8](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LeStlMk-2dj3HiNIbT_:-LeStlMl6hy-n9wJgJYz:b-tqi7nd) ([raw file](https://github.com/opensim-org/opensim-core/blob/489f61507bdea6706cd53b0cb2374391b8fefec3/OpenSim/Common/APDMDataReader.h#L107)):*
> ```objc
>         std::vector<std::string> search_labels,
>         const std::string& sensorName,
>         std::vector<int>& indices, bool newFormat=false) const;
> ```

Considering the format as binary is kind of short-sighted given they are already at 7!

OpenSim/Common/APDMDataReader.h
```diff
@@ -62,6 +62,7 @@ class OSIMCOMMON_API APDMDataReader : public IMUDataReader {
     // Header associated with Time
     static const std::string TimeLabel;
 
+protected:
```
aseth1: 👍

Applications/opensense/opensense.cpp
```diff
@@ -251,7 +251,7 @@ TimeSeriesTable_<SimTK::Quaternion> readRotationsFromXSensFiles(const std::strin
 {
     XsensDataReaderSettings readerSettings(readerSetupFile);
     XsensDataReader reader(readerSettings);
-    DataAdapter::OutputTables tables = reader.readSource(directory);
+    DataAdapter::OutputTables tables = reader.read(directory);
```
aseth1: 👍

OpenSim/Common/APDMDataReader.h
aseth1: Looks good.

OpenSim/Common/DataAdapter.h
```diff
@@ -119,9 +119,11 @@ class OSIMCOMMON_API DataAdapter {
                              const DataAdapter& adapter);
 
     /** Public interface to read data from a dataSourceSpecification, typically a file or folder */
-    virtual DataAdapter::OutputTables readSource(const std::string& dataSourceSpecification) const = 0;
+    DataAdapter::OutputTables read(const std::string& dataSourceSpecification) const {
+        return extendRead(dataSourceSpecification);
```
aseth1: 👍

aymanhab:
Thanks for the feedback @aseth1 this looks cleaner indeed 👍  but there are compilation errors due to the following:
1. On MSVC:  testAnalyzeTool.cpp uses the static method call  syntax STOFileAdapter_<double>::read I can fix this to use non-static call convention but it's a change to the public interface. We may as well rename these 'convenience' methods and that will take care of the second problem below.
2. Bindings: SWIG fails on e.g. CSVFileAdapter, C3DFileAdapter because overriding method (read) is static (all methods are virtual in java);

What's your preference @aseth1 to resolve build failures?

aseth1:
I think it was a mistake for the static read methods to have the same signature as the public adapter interface. I think it would be OK for the convenience methods to be renamed on the concrete class to `readFile()`,  for example.

OpenSim/Common/Test/testTRCFileAdapter.cpp
```diff
@@ -145,7 +145,7 @@ int main() {
         std::cout << "  " << filename << std::endl;
         TRCFileAdapter trcfileadapter{};
         try {
-            auto table = trcfileadapter.read(filename);
+            auto table = trcfileadapter.readFile(filename);
```
aseth1: `read` should work here since this is an instance method.

aymanhab: Not quite, since read returns a map of tables while the convenience method returns a single table that can be directly written to a file.

OpenSim/Common/APDMDataReader.cpp
```diff
@@ -67,17 +67,17 @@ APDMDataReader::extendRead(const std::string& fileName) const {
     bool newFormat = false;
     if (tokens[0] == "Format=7") {
         newFormat = true;
-        dataRate = 128; // Will fix on reading first 2 rows
+        dataRate = 128; // Will fix after reading computing it from time column
```
aseth1: Quick question @aymanhab does APDM not report data rate because they do not guarantee a fixed sampling rate? Given that each sample is time stamped, it is possible that the sampling rate is not perfectly uniform. In that case, reporting a data rate based on the first two samples won't be right.

aseth1:
___
*[OpenSim/Common/APDMDataReader.h, line 107 at r8](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LeStlMk-2dj3HiNIbT_:-LexJtMM9uR4hUZ7H0e0:bamskjn) ([raw file](https://github.com/opensim-org/opensim-core/blob/489f61507bdea6706cd53b0cb2374391b8fefec3/OpenSim/Common/APDMDataReader.h#L107)):*

Considering the format as binary is kind of short-sighted given they are already at 7!
</blockquote></details>

Should we create an issue that APDM format identifier should be an enum instead of bool?

aseth1:
___
*[OpenSim/Common/APDMDataReader.cpp, line 118 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LdQmVp69oCZf34HrP84:-LexKPRICdQdXswknwyC:b-30z1ih) ([raw file](https://github.com/opensim-org/opensim-core/blob/4433c86f5fbeda9fa8b9fd72d26a80f7229e226c/OpenSim/Common/APDMDataReader.cpp#L118)):*

Initialization seems like conjecture to me. Have you timed how long it takes to initialize a matrix of NaNs versus repeatedly creating RowVectors and then copying those values into the matrix? I would think allocating memory in a loop eats more time.
</blockquote></details>

We should profile readers. This can be in another issue.

OpenSim/Common/APDMDataReader.cpp
aymanhab: The two APDM formats handle this differently, in the old one they reported dataRate and we assumed it's fixed by sensor and calculated time, the later format has no reference to data rate and the data is timestamped. I'm leaning to remove the dataRate reference altogether and just report the times relative to starttime with no units attached. Can't find documentation anywhere, I'm in touch with their person to try to get more info.

aseth1:
___
*[OpenSim/Common/FileAdapter.h, line 194 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LeEQa6mCRt3xSwN8rSQ:-LexL6NW-QFCRIs87K94:bpoy3ui) ([raw file](https://github.com/opensim-org/opensim-core/blob/2371de2d8cc4d7a5bb8e21e7577282d4e6aea0dc/OpenSim/Common/FileAdapter.h#L194)):*

Does not belong here.
</blockquote></details>

Concrete readers can have their own convenience methods to read a file, I don't think the method to read tables should be here since it is identical to the DataAdapter interface. Can we remove altogether?

aymanhab:
> _[OpenSim/Common/APDMDataReader.h, line 107 at r8](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LeStlMk-2dj3HiNIbT_:-LexJtMM9uR4hUZ7H0e0:bamskjn) ([raw file](https://github.com/opensim-org/opensim-core/blob/489f61507bdea6706cd53b0cb2374391b8fefec3/OpenSim/Common/APDMDataReader.h#L107)):_
> 
> Previously, aseth1 (Ajay Seth) wrote…
> Should we create an issue that APDM format identifier should be an enum instead of bool?

Yes we should create the issue, though we have no clue if they will stick with the latest format=?? method or do something else. Remember the real data lives in their binary encoded files so they may not care about the csv format much.

aseth1:
___
*[OpenSim/Common/Test/testTRCFileAdapter.cpp, line 148 at r10](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LeTfGdPv8t62F9epbCj-r10-148:-LexMLwA8wfH0oJEHQHq:b32wtd) ([raw file](https://github.com/opensim-org/opensim-core/blob/d0d74bc7b0b28dbf1fc3bf2c9eb009bfa3a75305/OpenSim/Common/Test/testTRCFileAdapter.cpp#L148)):*

Not quite, since read returns a map of tables while the convenience method returns a single table that can be directly written to a file. 
</blockquote></details>

ok

aseth1:
___
*[OpenSim/Simulation/OpenSense/OpenSenseUtilities.h, line 48 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LdQy12y9SCPDEBVcGbh:-LexOrVv6jVU6_Q_x68e:b-frdeoc) ([raw file](https://github.com/opensim-org/opensim-core/blob/4433c86f5fbeda9fa8b9fd72d26a80f7229e226c/OpenSim/Simulation/OpenSense/OpenSenseUtilities.h#L48)):*

The assumptions are: 1) the model default pose is the same as the pose used to collect calibration data, and 2) that IMUs are labeled `<body(or physical frame)_in_model>_imu` so the underlying `PhysicalFrame` in the model can be identified and a corresponding Offset frame attached.
</blockquote></details>

@aymanhab do you want to add these assumptions in the comments? I agree they are helpful. Would be good to expose these to command line users as well.

aseth1:
___
*[OpenSim/Tests/shared/IMUData/APDM/apdm_format2_short.csv, line 1 at r8](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LeStBp_1MdvnQD3z3dE:-LexPXhO0q6EV5NrAjmp:b-88m0w6) ([raw file](https://github.com/opensim-org/opensim-core/blob/489f61507bdea6706cd53b0cb2374391b8fefec3/OpenSim/Tests/shared/IMUData/APDM/apdm_format2_short.csv#L1)):*

Maybe relabel the file "APDM_Format7..."?
</blockquote></details>

Might be nice to rename the file to be consistent with the actual APDM format number inside the file.

aseth1:
___
*[OpenSim/Common/APDMDataReader.cpp, line 70 at r13](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LcmQY6Iub7tc_0GgdDo-r13-70:-LexQ-c26pmEt_A2hT6y:b-mg66kz) ([raw file](https://github.com/opensim-org/opensim-core/blob/ea5fd0c9d914ea79f2f229616c780f9f35e450ea/OpenSim/Common/APDMDataReader.cpp#L70)):*
> I'm leaning to remove the dataRate reference altogether and just report the times relative to starttime with no units attached.

I agree with that since the times are stamped. I imagine they are still in seconds. Would be good to know what units of time if not second!

This can be another issue to be addressed in the next rev.

aseth1:
___
*[OpenSim/Common/FileAdapter.h, line 194 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/2437#-LeEQa6mCRt3xSwN8rSQ:-LexQobS4xW55saq6CL8:b-68r17r) ([raw file](https://github.com/opensim-org/opensim-core/blob/2371de2d8cc4d7a5bb8e21e7577282d4e6aea0dc/OpenSim/Common/FileAdapter.h#L194)):*

readSource is obsolete and was removed altogether already.
</blockquote></details>

True. But what about the `readFile`? It seems redundant with read() since it doesn't provide any other convenience (not a single table, etc.).

aseth1:
Yes, but we already support two version: unnumbered and 7. If the put out another one like 8 or 10, at least we can choose to accommodate (if we wish). Otherwise we're stuck with these two versions, which is fine for now. Let's create the issue and decide what to do later.

aymanhab:
FileAdapter::readFile also serves as a Factory to create adapter based on file extension so that users don't need to know the specific Adapter to use (it does it based on file extension). If you prefer to remove it, we can do that after. I created an issue for APDM format, will create the rest and post here. Merging.

PR #2441, opened 2019-05-03
Opensense select base imu

aseth1:
### Brief summary of changes

Enable the user to select the base body and its heading axis for performing a heading correction during calibration and tracking.

### Testing I've completed

1. verified that I could reproduce the original calibration model and IK motion files identically
2. verified that an invalid IMU throws and an empty name performs no correction
3. verified that `torso_imu` produces similar walking results but that adduction angles are better with `pelvis_imu`

It is nice to to have the flexibility to do calibration and tracking separately, but the cost is that it is now easy for the user introduce an inconsistency between the two steps.  Currently, calibration takes command-line arguments to perform heading correction but tracking uses an opensense setup file with options as properties, and therefore it is easy to have a mismatch between the two. If the heading correction on the experimental orientation data is not done both at calibration and at tracking and are not applied identically, the results will be wrong.  Ideas on how to thwart the inconsistency or make the user aware of it, are welcomed.

### CHANGELOG.md (choose one)
- no need to update because this is a local fix to opensense. Opensense itself should be documented.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2441)


jenhicks:
Thanks for the update @aseth1 . We separated the calibration from the tracking step so that users could do their own calibration. Could we set-up tracking some way so that you either specify the calibrated model or you use OpenSim's calibration process? (And thus still allow flexibility but also keep various settings in the same file to help prevent these inconsistencies.)

aymanhab:
My thoughts:
- Would be great to have the ability to serialize/reconstruct a PIPELINE from XML file(s) but this is not currently supported and doesn't fit the mental picture of what we typically write to XML (there's no OpenSense pipeline abstraction though we definitely should consider creating one in the future).
- The data import part (done once) is fundamentally different from the iterative calibrate/solve/calibrate/solve which is more similar to our typical model scaling workflow.
- Calibration and Tracking seem to belong together indeed so if we leave only these it should be workable.
- While command line is the one and only interface now we should be planning for the long term where users are invoking this from either XML/GUI or Matlab.

aseth1:
Thanks @aymanhab.

I agree we can leave the pure data reading out, since that typically would only happen once. The -Read option should still have specify the concrete IMU reader IMHO so that we can scale to different readers, without changing the interface. Since all readers will derive from the base IMUReader and comply with its interface, we don't need to hardcode the options.

I wouldn't say command line is one and only interface. At the demo most were interested in using via MATLAB and Patrick is already using Python. For the most part, it is easier to construct the pipeline you need via scripting.  For command line users the order and maintaining consistency between multiple calls with different options can be confusing. I think if OpenSense had XML settings for DataTransformation, Calibration, and Tracking it  would make things easier. Eventually, a GUI would simply expose the XML settings like it does for every other Tool.

aseth1:
@jenhicks @aymanhab Hannah most definitely and perhaps Patrick could use the flexibility to change the base sensor and its heading direction. Perhaps we could review as is and merge and then we can address how best to construct the settings for OpenSense in order that transformations are applied consistently in calibration and tracking steps.

aymanhab:
@aseth1 If this is needed to actually make use of the opensense tools then indeed let's merge it to the branch so that we can make a build and they can get going.

jenhicks:
Agreed ... sounds good to merge to the branch.

aseth1:
@aymanhab can you review. Feel free to merge upon approval.

aseth1:
___
*[Applications/opensense/opensense.cpp, line 131 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2441#-LeXbGTw5ql6L_stfLEH:-LeYQLNkDexa1uD_twJL:be2ratw) ([raw file](https://github.com/opensim-org/opensim-core/blob/a26214256ef30f023be9c8506e86ba722f553e89/Applications/opensense/opensense.cpp#L131)):*

Looks fins for the short term, longer term though the order of parameters on the command line becomes important and could be error prone
</blockquote></details>

I totally agree. Command arguments aren't the best way to configure complex options. As proposed/discussed I would like to see the DataTransform and Calibration settings be part of an OpenSense settings XML file.

aseth1:
___
*[OpenSim/Simulation/OpenSense/InverseKinematicsStudy.h, line 79 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2441#-LeXbsxe0I0f_TUnXs1n:-LeYSs-REGu7kNcdQK0o:bl3bjyu) ([raw file](https://github.com/opensim-org/opensim-core/blob/a26214256ef30f023be9c8506e86ba722f553e89/OpenSim/Simulation/OpenSense/InverseKinematicsStudy.h#L79)):*

Done.
</blockquote></details>

@aymanhab let me know if this is more meaningful.  If so feel free to merge.

Issue #2446, opened 2019-05-15
Enable calibrateModelFromOrientations() output model to be written to a file of user's choosing


jimmyDunne:
Calibrate models are printed using the input name and an appended string ('calibrated'). It would be great if there was some method to change the output model name to a custom string since a user may need to print multiple calibrated models, from multiple input orientation files, to the current directory.

aseth1:
@aymanhab @jimmyDunne I was thinking that `OpenSenseUtilities::calibrateModelFromOrientations()` should simply return the calibrated model and should not write to file as part of the API. You can the always print the model to any file you like in scripting, etc... I prefer to minimize file reading/writing as part of the OpenSenseUtilities API.

I will maintain the command line behavior to write to `calibrated_<given_model_filenmae>.osim`

aymanhab:
Agreed, this also allows for the GUI, eventually, to create a calibrated model for inspection without polluting the file system. For command line of opensense though, it would be good to allow users to pass in an output file name so the calibration can be done without programming/coding and to reduce the risk of model overwrite. Thoughts?

aseth1:
> For command line of opensense though, it would be good to allow users to pass in an output file name so the calibration can be done without programming/coding and to reduce the risk of model overwrite. Thoughts?

I like that, but another command line argument is going to be excessive. For now, I say we leave it as is and create calibration settings (XML) that would include all the options and allow users to maintain/reuse them.

PR #2452, opened 2019-05-16
Output model from OpenSense calibration

aseth1:
Fixes issue #2446

### Brief summary of changes
Instead of automatically writing the calibrated model to file, the calibrated model is now the output of the utility method. The caller/user is then free to use the model directly or write it to file as they choose.

### Testing I've completed
Updated testopensense.cpp accordingly.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2452)


jimmyDunne:
Confirmed on Matlab that a Model is returned. Wasn't able to test on the command line but I don't think it has anything to do with these changes.

aseth1:
Thanks @jimmyDunne and @aymanhab!

PR #2453, opened 2019-05-16
Restore build options appveyor

aymanhab:
Fixes issue #0
Restore publishing only on master branch and build Doxygen when specified in appveyor configuration, both were unintentionally modified by PR #2437
This only affects appveyor build, we can safely merge if Artifact can be found/published

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2453)


aymanhab:
Thanks @aseth1 much appreciated.

PR #2454, opened 2019-05-16
Replace line counting to find labels by looking for PacketCounter label 

aymanhab:
Fixes issue #2457

### Brief summary of changes
Instead of looking for a fixed line number, parse headers looking for a line starting with "PacketCounter". This allows for extra comment lines in header as reported by user testing of XsensReader

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2454)


aymanhab:
This PR makes the XSensReader more flexible to handle extra comments (e.g. // Options) in header. Test case updated accordingly

aymanhab:
@jimmyDunne @aseth1 for a quick review. Thanks

OpenSim/Common/XsensDataReader.cpp
```diff
@@ -62,13 +63,20 @@ XsensDataReader::extendRead(const std::string& folderName) const {
                     dataRate = std::stod(tokens[3]);
                 }
             }
-            if (j == labels_line_number) { // Find indices for PacketCounter, Acc_{X,Y,Z}, Gyr_{X,Y,Z}, Mag_{X,Y,Z} on line 5
+            if (!labelsFound) { // Find indices for PacketCounter, Acc_{X,Y,Z}, Gyr_{X,Y,Z}, Mag_{X,Y,Z} on line 5
```
aseth1: The `labelsFound` and `packetCounterIndex != -1` seem redundant. I don't see how this `if` is doing anything since `labelsFound` must be false to increment the `for` loop. I would expect the condition to be something like `while(packetCounterIndex < 0)` and a comment about incrementing the line number until the packet counter is found.

aymanhab: While I think labelsFound is more readable, and scales better to other methods of detecting labels than checking a index==-1, I made the suggested change anyway since it's rather local  in the function implementation.

OpenSim/Common/XsensDataReader.cpp
```diff
@@ -63,15 +62,14 @@ XsensDataReader::extendRead(const std::string& folderName) const {
                     dataRate = std::stod(tokens[3]);
                 }
             }
-            if (!labelsFound) { // Find indices for PacketCounter, Acc_{X,Y,Z}, Gyr_{X,Y,Z}, Mag_{X,Y,Z} on line 5
+            if (packetCounterIndex == -1) { // Find indices for PacketCounter, Acc_{X,Y,Z}, Gyr_{X,Y,Z}, Mag_{X,Y,Z} on line 5
```
aseth1: Can `packetCounterIndex == -1` ever be false given the `for` condition?

aymanhab:
Changes made, thanks for the review @aseth1

OpenSim/Common/XsensDataReader.cpp
aymanhab: No, good point 👍

aymanhab:
Thanks @aseth1

Issue #2447, opened 2019-05-15
Updated base Heading Axis in calibrateModelFromOrientations()


jimmyDunne:
The base IMU rotation in calibrateModelFromOrientations() (may) assume that a 180 degree rotation is an error with the IMU and not an intentional location of the subject in the OpenSim frame. It would be great to remove this assumption and rather have more flexibility in providing a base heading axis for the base IMU.

jimmyDunne:
fixed in #2459

Issue #2456, opened 2019-05-17
InverseKinematicsStudy() results being written over themselves


jimmyDunne:
Currently, InverseKinematicsStudy() does some [automatic editing of the orientations file name-- based on the location of '_'-- to generate an output filename for the orientation results. ](https://github.com/opensim-org/opensim-core/blob/7e4f6ab1e3346f9e822a8a63e008ac16ec25b16d/OpenSim/Simulation/OpenSense/InverseKinematicsStudy.cpp#L190-#L198). 

I have my input orientations files named like this;
```
MT_012005D6_009-quaternions_Walking.sto
MT_012005D6_009-quaternions_RHJCSwinger.sto
MT_012005D6_009-quaternions_LHJCSwinger.sto
```
When I perform an InverseKinematicsStudy() with each of the above input files, the results get written over the top of one another as;
```
IK_MT_012005D6_009-quaternions.mot
```
To solve, we could either (a) don't do any editing of the input string and rather just prepend 'IK_', or (ii) have a property for an output string that the user can set.

aseth1:
I am running into this issue to in my testing of the calibration with different headings. I will fix it as part of m PR since I need it now.

jimmyDunne:
Sounds good. Happy to test :)

aymanhab:
Should be fixed, folded into PR #2459

Issue #2455, opened 2019-05-17
InverseKinematicsStudy, setModel(), & set_model_file_name()


jimmyDunne:
(Matlab, Mac)

`InverseKinematicsStudy` method `setModel()` would be useful since `OpenSimUtilities.calibrateModelFromOrientations()` returns a calibrated model. A user wouldn't have to write a calibrated model to file as an intermediate step. 

However, if I give setModel() a model, but don't have any value for set_model_file_name(), i get an exception

```
>> ik.setModel(obj.calibratedModel)
>> ik.set_model_file_name('')
>> ik.run()
Java exception occurred:
java.lang.RuntimeException: Object: Cannot construct from empty filename. No filename specified.
	Thrown at Object.cpp:100 in Object().
	at org.opensim.modeling.opensimSimulationJNI.InverseKinematicsStudy_run__SWIG_1(Native Method)
	at org.opensim.modeling.InverseKinematicsStudy.run(InverseKinematicsStudy.java:381)
```

It would be great if a user could use either a Model or a ModelPath.

jimmyDunne:
I think that this is possibly a more general issue than with OpenSense since the other tools have similar issues. Perhaps down the road we can look at the interface and see how it could be more flexible across scripting and command line functions. For now, I will ignore the setModel() method and just use intermediate models files.

PR #2459, opened 2019-05-18
Allow for 180 heading correction

aseth1:
Fixes issue #2447, #2455, #2456

### Brief summary of changes
Initial heading correction implementation was assuming that a correction >90 degrees was due to the user inadvertently specifying the negative of the IMU axis (as the heading direction) and so we were doing another 180 degree correction. Of course in reality we do have instances where the pelvis is actually oriented ~180 from the OpenSim +X direction and so the model was being oriented backwards compared to the data.  The fix was to remove that assumption and trust the correct axis is specified as forward (#2447).

If the model is set on the Study, it takes precedence over the model file specified in the properties (#2455).

The output filename is now set to be the input filename prepended with "ik_" without any attempt to shorten, which was causing the output file to no longer be unique (#2456). For every unique input file you will get a similarly unique output file.

### Testing I've completed
Add a test where two calibration poses are known to have their pelvis headings opposite (180 degrees) from one another. The resulting calibrated model should be relatively the same (besides small offsets in IMU registration) and we test both models on a dynamic (hip joint center) trial. The resulting kinematics are compared and passes if the solutions are with 10degrees of each other.

- since we are doing tracking twice and comparing solutions I removed the previous tracking "test" that simply ran tracking with no comparison.
- we could add a regression test that is more stringent to flag us if the model kinematics change from the current result

### CHANGELOG.md (choose one)

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2459)


aseth1:
@jimmyDunne please give it a whirl!

jimmyDunne:
I tested this and all the changes that were made seem to work great on my machine. The coordinate heading issue is resolved and I testing using multiple, different,  initial calibration headings with equal success. The model setting is great and I can now use setModel() without specifiying a path to a model. Finally, the IK output is now just using a prepended 'IK_'.

aymanhab:
@aseth1 Can you look into the test failures before merging? Thanks

aseth1:
@aymanhab I incorrectly assumed that vector of tolerances limits the number of comparisons (i.e. 23 for the lower extremity) but that isn't the case. The tolerances must be the same size as the number of columns being compared. I made that change to the test.

aseth1:
Thanks @aymanhab and @jimmyDunne !

Issue #2465, opened 2019-05-21
OpenSim::TimeSeriesTable_<SimTK::Rotation> is not SWIG wrapped


aseth1:
Access to to resulting `TimeSeriesTable` of Rotations (e.g. using `OpenSenseUtilies`) is opaque in MATLAB.

aymanhab:
While exposing TimeSeriesTable <Rotation> is straightforward, the base class DataTable_<double, Rotation> is harder to wrap because all previous examples were templatized with subclasses of SimTK::Vec. Do you know if that was a limitation on the design of the DataTable_ template, @aseth1 ? 
We get away with it in C++ because methods like averageRow and clone are never generated or invoked but SWIG generates these regardless.

aseth1:
> Do you know if that was a limitation on the design of the DataTable_ template, @aseth1 ?

We use `TimeSeriesTable_<SimTK::Rotation>` everywhere in OpenSense, so I don't think there is any particular issue here, other than creating the specialized `TimeSeriesTableRotation` in [common.i](https://github.com/opensim-org/opensim-core/blob/0af1649e48b88e0a561a73976b8dea4ce56e4809/Bindings/common.i#L123-L134) as was done for the other `TimeSeriesTable_` types.

aymanhab:
Working on this now since otherwise some methods are not available to scripting users

aseth1:
I took the liberty to add these in to common.i  and added the specialization for `DataTable::makeElement_helper` to handle a SimTK::Mat<M,N>. Want me to pus that for you to continue and test @aymanhab?

aseth1:
@aymanhab it is looking like this will be important for scripting users who want to use `TimeSeriesTableRotation` to access individual rotations for comparison (Hannah) as well as to setup the `InverseKinematicsSolver` with an `OrientationsReference` initialized from the table of rotations (current limitation for Patrick).

aymanhab:
I believe this has been fixed by PR #2467 Please let me know otherwise

jimmyDunne:
Waiting to be added to release notes

PR #2467, opened 2019-05-21
SWIG wrap TimeSeriesTableRotation

aseth1:
Fixes issue #2465 

### Brief summary of changes
- Added bindings for TimeSeriesTable_<SimTK::Mat33> as TimeSeriesTableMat33 and TimeSeriesTable_<SimTK::Rotation> as TimeSeriesTableRotation.
- specialized `DataTable::makeElement_helper()` to handle elements of type SimTK::Mat<M,N>.

### Testing I've completed
- compiled and tests passed locally

@aymanhab please feel free to take this PR over. I hope it will be straightforward.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2467)


aymanhab:
Thanks for getting this started @aseth1 I believe some of the following types will also need to be exposed as they are used to get data in/out of the table (where ETY is SimTK::Rotation):
``
SimTK::RowVector_<ETY,  SimTK::RowVectorView_<ETY>, SimTK::Vector_<ETY>, SimTK::VectorView_<ETY>, SimTK::Matrix_<ETY> SimTK::MatrixView_<ETY>
``
Was it necessary to add the specialization for Mat33, or just for potential future use?

aseth1:
I was using `TimeSeriesTableSpatialVec` as an example, and I did not see any additional special handling

> Was it necessary to add the specialization for Mat33, or just for potential future use?

Not necessary, just useful and a potential intermediary if `Rotation` has other challenges.

aymanhab:
As of now:
- DataTableRotation is wrapped but getRowAtIndex returns SWIGTYPE_p_SimTK__RowVectorView_T_SimTK__Rotation_T_double_t_t getRowAtIndex
- Inheritance is wrong since TimeSeriesTableRotation does not subclass DataTableRotation

Bindings/simbody.i
```diff
@@ -175,7 +175,20 @@ namespace SimTK {
...
 %template(Rotation) SimTK::Rotation_<double>;
-%template(InverseRotation) SimTK::InverseRotation_<double>;
+
+// TimeSeriesTableMat33 works perfectly as expected, while TimeSeriesTableRotation doesn't
+%template(MatrixBaseMat33)    SimTK::MatrixBase<Mat33>;
+%template(MatrixMat33)        SimTK::Matrix_<Mat33>;
+%template(RowVectorBaseMat33) SimTK::RowVectorBase<Mat33>;
+%template(RowVectorViewMat33) SimTK::RowVectorView_<Mat33>;
+%template(RowVectorMat33)     SimTK::RowVector_<Mat33>;
+
```
aymanhab: TimeSeriesTableMat33 now works perfectly and rows can be obtained and used, but doing the same thing for Rotation fails due to the method elementwiseAssign. @aseth1 do you know if this method is supposed to exist for Rotation? To see the error, uncomment the lines below and build

aymanhab: Latest version, swig out elementwiseAssign and now the swig generated code compiles and DataTable_<Rotation> acts properly. TimeSeriesTable_<Rotation> however doesn't have the correct class hierarchy likely because it was encountered y swig before both TimeSeriesTable and Rotation were known. Investigating..

aymanhab:
Some update as I continue to fight with SWIG to get proper wrapping and realizing that many other classes along the chain are improperly wrapped:
- DataTableMat33, TimeSeriesTableMat33 (are properly wrapped and usable)
- DataTableQuaternion: Correct hierarchy but bad return types/unusable
- TimeSeriesTableQuaternion: Correct hierarchy but bad return types/unusable
- DataTableRotation:  Correct hierarchy but bad return types/unusable
- TimeSeriesTableRotation: Incorrect hierarchy and bad return types/unusable

If a code snippet to demonstrate actual use case can be provided, that would be extremely useful in deciding where to focus work/suggest workaround. While Rotation is usefully more strict, if the use case doesn't require it, we may get away with using Mat33 for return types to let testing proceed.
Let me know your thoughts @aseth1 and @jimmyDunne since you're most familiar with what users are actually doing in the API

aymanhab:
TimeSeriesTableRotation now works as expected and Rows can be obtained an dereferenced.

aymanhab:
All java wrapping now works exactly as expected. Build filing on python wrapping/travis seems likely unrelated but will try on windows.

aymanhab:
Python compilation errors boil down to utility classes e.g. MatrixView_  expecting being able to perform +=, -= on Quaternion type which does not support these operations.

aymanhab:
@chrisdembia Any thoughts on this issue since you worked more on the python wrapping? My plan is to get SWIG to ignore the methods that invoke these -=, += methods but that may lead to some methods not available for some specializations. Is there a better way to do this? Thoughts @aseth1 ?

aymanhab:
OrientationReference also need to be wrapped and be accessible from Python.

aymanhab:
@sherm1 I wonder what are your thoughts on this issue:
- We recently introduced Vector_<SimTK::Rotation>, Vector_<SimTK::Quaternion> across the OpenSim interface.
- The SWIG generated code fails to compile because it does not find implementation of +=, -= of these SimTK types.
- C++ likely never invokes these methods so the issue is hidden (until someone invokes these methods).
- I have been trying to convince SWIG not to wrap these methods but it seems there could be a solution on SimTK side. 
Your thoughts? Thanks much.

sherm1:
Hi, Ayman -- thanks for thinking of me! The += and -= operations are not defined for Rotation and Quaternion because those operations don't make sense for those objects. If the underlying Matrix33 and Vec4 objects are extracted, the usual operations could be applied but the result can't be a Rotation or a Quaternion.

So if you are thinking we could define those in C++ I don't think that is the right solution because it would imply that they were meaningful operations. Is that what you had in mind?

aseth1:
@sherm1, perhaps @aymanhab is suggesting that the template class be specialized in these cases to throw, for example. Then the specialization could be wrapped and used outside of C++.

sherm1:
That seems like it would convert an (early) compile-time error into a (late) runtime error. Generally, that's not a good thing to do. I'm not clear why we would want to do that in this case?

aymanhab:
Thanks for responding @sherm1 Indeed I was thinking the same thing as @aseth1 that we add an implementation that throws since the methods are assumed by SimTK::Vector.
While I agree that compile time error is better than runtime, it feels like a 💣 waiting to explode if a user creates a SimTK::Vector<Rotation>, there's the possibility that some compiler may fail to resolve the operator properly with wrong results which are likely worse than an exception.

sherm1:
I realize from your response Ayman that I must not have a grasp on the problem. I do not understand why there would ever be a circumstance in which any code would try to add Rotations together. Can you explain?

aymanhab:
Thanks @sherm1 for chiming in on this.

- The source of the issue is that SWIG generates code for ALL the methods of Vector_ template because it can't anticipate what will be called at runtime.
- The corollary is that SWIG generates code for Vector_<Rotation>+= (or scalarPlusEq method since it ignores operators) same for Vector_<Quaternion>, this code fails to compile.
- We can try to fix this by injecting code into the SWIG wrapping since we don't want to annotate SimTK headers with SWIG macros, the other option is to actually implement these methods and throw on SimTK side.

Please let me know if that clarifies the issue and whether you would consider handling this on the SimTK side at all.

sherm1:
Thanks, but that still leaves my mystified as to how it affects Rotation and Quaternion, which are not the same objects as a Vector_. Where do they fit in to this discussion?

aymanhab:
@sherrm1 Here's the error log compiling SWIG output 
`4>c:\install\simbody361\include\simtkcommon\internal\vec.h(788): error C2677: binary '-=': no global operator found which takes type 'const EE' (or there is no acceptable conversion)
4>        with
4>        [
4>            EE=SimTK::Quaternion_<SimTK::Real>
4>        ]
4>c:\install\simbody361\include\simtkcommon\internal\vec.h(777): note: see reference to function template instantiation 'SimTK::Vec<4,SimTK::Real,1> &SimTK::Vec<4,SimTK::Real,1>::scalarMinusEq<EE>(const EE &)' being compiled
4>        with
4>        [
4>            EE=SimTK::Quaternion_<SimTK::Real>
4>        ]
....
`
These errors occur we now expose the Quaternion class and instantiate the templates `SimTK::Vector_<Quaternion_<double>>,  SimTK::MatrixView_<Quaternion_<double>, ...` to support making matrices of Quaternions. same issue with Rotation. Build log [here]( https://travis-ci.org/opensim-org/opensim-core/jobs/539358639)

sherm1:
Since Quaternion and Rotation are not scalar types (and not "composite numerical types"), they aren't suitable for use as elements of a Vector or Matrix. Someone could do that with the underlying storage Vec4 or Mat33, but not with the Quaternion and Rotation types because they impose a lot of constraints that scalar types can't (like no addition or scalar multiply is allowed for these objects).

sherm1:
P.S. Rotation and Quaternion can be put in other types of containers, like std::vector or SimTK::Array, but Vector and Matrix are not just containers and shouldn't be used that way.

aymanhab:
Thanks very much @sherm1 much appreciated. That explains why it felt like trying to fit a square peg  into a round hole. Will open an issue to address the misuse. Thanks again 👍 🥇

aymanhab:
Indeed that was the issue since these brought up constructors and operators
that failed to build. I'll add the note on the swig interface files.

Issue #2470, opened 2019-05-22
Weird Patella Coordinates results when viewed through OpenSim GUI


jimmyDunne:
After running through OpenSense Pipeline and not seeing an issue through the visualizer, if I open the calibrated model (Rajagopal model) and load the output motion, the patellas go crazy. In discussion with @aseth1, this may be caused by some assumptions about the coordinate type during the IK study.

aseth1:
I'll tackle this ASAP.

aseth1:
Can this be closed by #2473?

jimmyDunne:
Yes thank you! Issue is fixed now

PR #2473, opened 2019-05-22
InverseKinematicsStudy only converts purely Rotational coordinates to degrees.

aseth1:
Fixes issue  #2470

### Brief summary of changes
- Explicitly check for Rotational coordinates (instead of not Translational)

### Testing I've completed
- reran opensense IK and loaded calibrated model and resultant motion in OpenSim GUI where patellas moved as expected

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2473)

### DIFF
--- a/OpenSim/Simulation/OpenSense/InverseKinematicsStudy.cpp
+++ b/OpenSim/Simulation/OpenSense/InverseKinematicsStudy.cpp
@@ -113,18 +113,21 @@ runInverseKinematicsWithOrientationsFromFile(Model& model,
     TableReporter* ikReporter = new TableReporter();
     ikReporter->setName("ik_reporter");
     auto coordinates = model.updComponentList<Coordinate>();
-    std::vector<std::string> transCoordNames;
+    std::vector<std::string> rotCoordNames;
 
     // Hookup reporter inputs to the individual coordinate outputs
-    // and keep track of coordinates that are translations
+    // and keep track of coordinates that are rotational
     for (auto& coord : coordinates) {
         ikReporter->updInput("inputs").connect(
             coord.getOutput("value"), coord.getName());
-        if (coord.getMotionType() == Coordinate::Translational) {
-            transCoordNames.push_back(coord.getName());
-            coord.set_locked(true);
+        if (coord.getMotionType() == Coordinate::Rotational) {
+            rotCoordNames.push_back(coord.getName());
+        }
+        else if(coord.getMotionType() == Coordinate::Translational) {
+            coord.setDefaultLocked(true);
         }
     }
+
     model.addComponent(ikReporter);
 
     TimeSeriesTable_<SimTK::Quaternion> quatTable =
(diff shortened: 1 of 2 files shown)


OpenSim/Simulation/OpenSense/InverseKinematicsStudy.cpp
```diff
@@ -113,18 +113,21 @@ runInverseKinematicsWithOrientationsFromFile(Model& model,
...
             coord.getOutput("value"), coord.getName());
-        if (coord.getMotionType() == Coordinate::Translational) {
-            transCoordNames.push_back(coord.getName());
-            coord.set_locked(true);
+        if (coord.getMotionType() == Coordinate::Rotational) {
+            rotCoordNames.push_back(coord.getName());
+        }
+        else if(coord.getMotionType() == Coordinate::Translational) {
+            coord.setDefaultLocked(true);
         }
```
aymanhab: Is there difference in behavior between set_locked and setDefaultLocked?

OpenSim/Simulation/OpenSense/InverseKinematicsStudy.cpp
```diff
@@ -195,11 +198,11 @@ runInverseKinematicsWithOrientationsFromFile(Model& model,
...
+    // but only for rotational coordinates
     for (size_t i = 0; i < report.getNumColumns(); ++i) {
-        auto it = find( transCoordNames.begin(), transCoordNames.end(),
+        auto it = find(rotCoordNames.begin(), rotCoordNames.end(),
                         report.getColumnLabel(i) );
-        if (it == transCoordNames.end()) { // not found = not translational
+        if (it != rotCoordNames.end()) { // found a rotational coordinate
             auto repVec = report.updDependentColumnAtIndex(i);
             repVec *= SimTK::Real(SimTK_RTD);
         }
```
aymanhab: Any reason not to use the method convertRadianToDegrees e to have these checks and conversions done in one pace?

OpenSim/Simulation/OpenSense/InverseKinematicsStudy.cpp
aseth1: No difference. `setDefaultLocked` is more explicit about the default state of the coordinate is locked.

OpenSim/Simulation/OpenSense/InverseKinematicsStudy.cpp
aseth1: I didn't want to convert to a `Storage`, but I see that Shrinidhi added a `TimeSeriesTable` version. I'll try that.

aseth1: I remember now,  conversion fails if the "inDegrees" key is not specified. I've fixed that now too.

jimmyDunne:
I built this branch and re-ran the tracking, but I still seem to be getting the same issue. I have attached the model and the output kinematics. 

[opensense_patella.zip](https://github.com/opensim-org/opensim-core/files/3209344/opensense_patella.zip)

aseth1:
@aymanhab thanks for the suggestion. Looks a lot cleaner and nice to centralize the conversion too. Should be good to go.

aseth1:
@jimmyDunne thanks for reviewing and testing.

PR #2484, opened 2019-06-04
Add Python scripts

jimmyDunne:
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2484)


aseth1:
Looks. Relatively minor comments for consistency.

jimmyDunne:
> I'm being picky, but would prefer that the comments and names referred to the contents of the data and not the sensor type, especially since the method names also correspond: E.g.
> 
> orientations
> linearAccelerations
> magneticNorthHeadings
> angularVelocities

You should also chang them in the C++ test since that is where I got them from.

aseth1:
> You should also change them in the C++ test since that is where I got them from.

I made the same requests for the readers when I reviewed them. The `get###Table()` methods on the readers were updated accordingly but I guess the the tests were not. The only grace is that the C++ tests aren't user facing, but I agree the inconsistency is confusing.

aseth1:
Built and tested locally using test files from SimTK [here ](https://simtk.org/frs/download_confirm.php/file/5774/OpenSenseExampleFiles.zip?group_id=1715). All ran perfectly. Just had to rename imuTrackingModel.osim -> Rajagopal_2015.osim. Setting up Python went smoothly too-- no issues!

aseth1:
test failure on travis-ci/push looks unrelated (testGait10dof18musc).
@aymanhab or @jenhicks can you do a quick review and merge if OK. Thanks!

aseth1:
Great. Thanks @aymanhab.

Issue #2435, opened 2019-04-18
Error during IAA


jimmyDunne:
User reported bug [here](https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=10219&p=0&start=0&view=&sid=e53d8d4f2297969a1f06d33909213a9c)

If report_constraint_reactions == true, during IAA, then an error is returned;

![image](https://user-images.githubusercontent.com/5296507/56376876-0b7b3e80-61d7-11e9-93e1-2988fafa3400.png)

Testing files are attached. 

[InducedAccelerations.zip](https://github.com/opensim-org/opensim-core/files/3095387/InducedAccelerations.zip)

aseth1:
Fixed by #2489

PR #2489, opened 2019-06-10
Fix IAA exception when reporting constraint forces

aseth1:
Fixes issue #2435 

### Brief summary of changes
Identify constraint's `PhysicalFrame` instead of `Body`. 

### Testing I've completed
- user supplied case works
- enabled reporting of constraint reactions for IAA test case and it produces GRF like totals:
image

- none

### CHANGELOG.md (choose one)
- no need to update because this is a simple bug fix

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2489)

### DIFF
--- a/Applications/Analyze/test/subject02_Setup_IAA_02_232.xml
+++ b/Applications/Analyze/test/subject02_Setup_IAA_02_232.xml
@@ -75,7 +75,7 @@
 					<compute_potentials_only> false </compute_potentials_only>
 					<!--Report individual contributions to constraint reactions in addition to
 					    accelerations.-->
-					<report_constraint_reactions> false </report_constraint_reactions>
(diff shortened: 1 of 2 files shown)


aymanhab:
@aseth1 The changes look reasonable but I was surprised to see the xml for the test case changed with no corresponding change in results. Is there a separate test case with the flag set to false? Thanks

aseth1:
> I was surprised to see the xml for the test case changed with no corresponding change in results. Is there a separate test case with the flag set to false? 

The flag only controls if the constraint reaction forces are written out or not. This was a reporting bug not a computation. The IAA results remain identical.

aseth1:
Thanks @aymanhab and @jimmyDunne!

PR #2490, opened 2019-06-12
No need to build SIMM translator anymore

aymanhab:
Fixes issue #0

### Brief summary of changes
no need to build SIMM translators anymore since not used by API or GUI
### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2490)


aymanhab:
@chrisdembia Mainly CMake change to not build/wrap/distribute a library that we do not support in 4.0, also removing the obsolete source !

Issue #2421, opened 2019-03-07
C3DFileAdapter should not throw if either forces or markers data is not present


aseth1:
From @jimmyDunne 

>C3D files without any force data. I am getting an exception when I try to read them with the C3DFileAdapter().
```
Java exception occurred:
java.lang.RuntimeException: map::at:  key not found
at
        org.opensim.modeling.opensimCommonJNI.C3DFileAdapter_read__SWIG_1(Native
        Method)
at
        org.opensim.modeling.C3DFileAdapter.read(C3DFileAdapter.java:66)
 
I have attached a zip with four C3Ds from the same session— three are kinematic calibration trials that don't have forces and one walking trial with both markers and forces (which works).
```

jimmyDunne:
@jimmyDunne to test. May have been fixed already. 

May need to change read to readFile(), I will check and set up an issue to fix. If so, will have to update Matlab and Python scripts as well.

aymanhab:
@jimmyDunne Can you attach the c3d file with the missing data (described in the issue description) so we can test a fix? Thanks

jimmyDunne:
[Archive.zip](https://github.com/opensim-org/opensim-core/files/3288280/Archive.zip)

I have attached a zip with four C3Ds from the same session— three are kinematic calibration trials that don't have forces and one walking trial with both markers and forces (which works).

aymanhab:
One reason we use readFile rather than the generic read is that the read method of C3DFileAdapter takes a second argument

 ` C3DFileAdapter::readFile(filename,
        C3DFileAdapter::ForceLocation::OriginOfForcePlate);`

tkuchida:
> One reason we use readFile rather than the generic read is that the read method of C3DFileAdapter takes a second argument

I found this part of the interface confusing. I would have expected only one method to read a C3D file. Why not override `readFile()` in the C3DFileAdapter class to throw so that people know to use the 2-argument `read()` instead? (The second argument of `read()` has a default anyway.) A potential danger is that, if a user knows only about `readFile()` and the data doesn't make sense, it isn't obvious that there's a separate `read()` method where you can specify how the forces are expressed.

jimmyDunne:
Below are the behaviors as I see right now;

**Case 1 - C3D with only Marker**
If I use readFile(), I get an exception;
```
>> c3d = C3DFileAdapter().readFile('Static.c3d',0)
Java exception occurred:
java.lang.RuntimeException: map::at:  key not found
[...]

PR #2495, opened 2019-06-17
If data missing from c3d file, return an empty table 

aymanhab:
rather than throw exception downstream

Fixes issue #2421

### Brief summary of changes
Branch of the code that reads data updated to return empty Markers/Forces tables

### Testing I've completed
Stepped through the files in provided dataset and empty tables were created and returned indeed.

### CHANGELOG.md (choose one)

- not sure if this needs documentation somewhere since the old behavior was basically a bug. May need mention in changelog.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2495)

### DIFF
--- a/OpenSim/Common/C3DFileAdapter.cpp
+++ b/OpenSim/Common/C3DFileAdapter.cpp
@@ -155,7 +155,13 @@ C3DFileAdapter::extendRead(const std::string& fileName) const {
 
         tables.emplace(_markers, marker_table);
     }
-
+    else { // insert empty table
+        std::vector<double> emptyTimes;
+        std::vector<std::string> emptyLabels;
+        SimTK::Matrix_<SimTK::Vec3> noData;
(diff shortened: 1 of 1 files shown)


aymanhab:
@jimmyDunne There's inherent danger of having the scripting interface drift from the C++ interface, users will likely get confused by the omission. Doxygen will still have it. Exclusion from wrapping feels like a band-aid. Can you outline the common use case so we can fix the interface? The method read is in the base class so we can't hide it here, however we can override it and throw pointing the user to use readFile, would that be a better option for you?

jimmyDunne:
Thanks for the clarification. Makes total sense. 

I think we can just highlight the correct usage in documentation. I updated the confluence page and I will put in a PR that uses readFile() over read() whenever the C3DFileAdapter() is used.

jimmyDunne:
@aymanhab  will check the test then Merge.

aymanhab:
Failure due to timing which is flaky. Merging

Issue #2395, opened 2018-12-05
No exception when loading outdated model


jimmyDunne:
[Reported on the User forum](https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=9841&p=27038&start=0&view=)

A user loaded a model in 4.0 and the pelvis coordinates where all screwy. I got the user to email me the model so I could figure out the issue. It turns out the model is a much older version and so 4.0 was messing up the update. The user could fix the issue by loading and saving the model in 3.3. The model version is 10600 🤣 

![image](https://user-images.githubusercontent.com/5296507/49528393-e0c0c700-f881-11e8-9ae3-5e821c180bc3.png)

I think the correct behavior should be an exception when trying to load this model in 4.0. I have attached the offending model for diagnoses and testing. 

[gait2392_arms.osim.zip](https://github.com/opensim-org/opensim-core/files/2649618/gait2392_arms.osim.zip)

PR #2498, opened 2019-06-17
Throw exception on model files that are before version 1.8.1

aymanhab:
Fixes issue #2395

### Brief summary of changes
Throw exception when model file format predates version 1.8.1 (document version 10901) since we don't handle the upgrade for these model files.

### Testing I've completed
Loaded model in issue report and Exception was thrown.

The text/language of the exception message.

### CHANGELOG.md (choose one)

- no need to update because bugfix?

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2498)

### DIFF
--- a/OpenSim/Simulation/Model/Model.cpp
+++ b/OpenSim/Simulation/Model/Model.cpp
@@ -105,7 +105,7 @@ Model::Model(const string &aFileName) :
     updateFromXMLDocument();
     // Check is done below because only Model files have migration issues, version is not available until 
     // updateFromXMLDocument is called. Fixes core issue #2395
-    OPENSIM_THROW_IF(getDocument()->getDocumentVersion() < 30000,
+    OPENSIM_THROW_IF(getDocument()->getDocumentVersion() < 20302,
         Exception,
         "Model file " + aFileName + " is using unsupported file format"
         ". Please open model and save it in OpenSim version 3.3 to upgrade.");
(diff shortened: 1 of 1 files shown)


aseth1:
AFAIK we load models that are 2.4 without issues. Many test cases still have 2.4 models. So perhaps you can throw if < 2.4?

aymanhab:
Thanks @aseth1 I changed the number to 20302 but left the message at 3.3 because earlier versions will be unlikely to be runnable going forward.

aymanhab:
@jimmyDunne Please let us know if you're aware of any 2.4 models that are problematic in 4.0. Thank you

jimmyDunne:
I downloaded and installed 2.4 on a Windows VM so I could get the 2.4 models. I was able to open all of the models in 4.0 without an issue. 

image

I do remember there were some issues with specific models (Dynamic Walker) but that was because of unrelated reasons.

jimmyDunne:
Waiting on the tests.

aymanhab:
@aseth1 was correct, there are models from 1.8,1 in the test suite. Changing test version number to 10901 make these tests pass and also captures the offending model in issue #2395

aymanhab:
@aseth1 Please let me know if this can be merged now.
@jimmyDunne If you think this needs to go into changelog let me know.

aymanhab:
Approved verbally by @jimmyDunne

jimmyDunne:
@jimmyDunne will add to the changelog what minimum version 4.0 supports.

Issue #2443, opened 2019-05-07
IKTool ignores IKTaskSet and tracks all markers in the model


jimmyDunne:
Forum post [here](https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=10372&p=28752&start=0&view=)

Two issues that are most likely related;
(1) IKTool ignores <apply>false</apply> in IKTaskSet()
(2) If model and trc contain matching markers, but the IKTaskSet doesn't, markers are tracked anyway. So you can intentionally remove a marker from the IKTaskSet and the marker will still be tracked. The only way to stop the tracking seems to be to remove the marker from the model. 

I have attached a model, trc, and IKTool setup file for testing;
[IK_bug_test.zip](https://github.com/opensim-org/opensim-core/files/3152699/IK_bug_test.zip)

PR #2500, opened 2019-06-18
Fix to IK's MarkersReference to include markers only specified in tasks/weights

aseth1:
Fixes issue #2443 

Changed the behavior of the `InverseKinematicsTool` to ignore markers not specified as IKTasks in the tool's IKTaskSet.

### Brief summary of changes
1. `InverseKinematicsTool` correctly created a `MarkerWeightSet` with only Marker tasks that were applied (`getApplied() == true`) and now the `MarkersReference` constructor only creates reference signals for those markers that are specified in the given *markerWeightSet*. If no *markerWeightSet* is given, then all Markers are included with the same default weight.
2. The `MarkersReference`'s underly table is modified to remove columns for markers with a corresponding `MarkerWeight`.
3. `DataTable_` interface was expanded to include methods to remove columns by index and name.
4. `testIK::checkMarkersReferenceConsistencyFromTool()` condition updated to verify markers not in the TaskSet are excluded (no longer included with default weight).

### Testing I've completed
1. IK problem in #2443 runs as expected with no significant movement since only tracking a shoulder/stationary marker. Verified that reported errors are only for the one marker.
2. Updated testIK case to verify that removing tasks also reduces the number of markers to be tracked in the `MarkersReference`.

- Is the behavior with a `MarkerWeightSet` (only including markers specified in the Set) and without (use all markers) what users would expect? 

### CHANGELOG.md (choose one)
- update required because the default behavior of IK and results will change when given tasks which are a subset of the model and experimental markers.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2500)


OpenSim/Common/DataTable.h
```diff
@@ -948,6 +948,45 @@ class DataTable_ : public AbstractDataTable {
...
+        auto labels = getColumnLabels();
+
+        assert(labels.size() == _depData.ncol());
+
+        // shift columns unless we're already at the last column
+        for (size_t c = index; c < getNumColumns()-1; ++c) {
+            _depData.updCol((int)c) = _depData.col((int)(c + 1));
+            labels[c] = labels[c + 1];
+        }
+
```
aymanhab: In general, do users care that removing a column could changes the order of other columns? If not it would be more efficient to exchange the column/label to be removed with the last column/label rather than repeated shifting.

OpenSim/Simulation/MarkersReference.h
```diff
@@ -105,13 +105,17 @@ class OSIMSIMULATION_API MarkersReference : public Reference_<SimTK::Vec3> {
     //--------------------------------------------------------------------------
     MarkersReference();
 
-    /** Convenience load markers from a file */
+    /** Convenience load markers from a file. See below. */
     MarkersReference(const std::string& markerFileName,
+                     const Set<MarkerWeight>* markerWeightSet = nullptr,
                      Units modelUnits = Units(Units::Meters));
```
aymanhab: I understand using a pointer is to allow a default value for nullptr, but is that an actual use case in our code base? If not I would suggest passing a const reference to `Set<MarkerWeight>` instead

OpenSim/Simulation/MarkersReference.cpp
```diff
@@ -37,24 +37,27 @@ MarkersReference::MarkersReference() :
...
         upd_marker_weights() = *markerWeightSet;
-    populateFromMarkerData(markerTable, units.getAbbreviation());
+    populateFromMarkerData(_markerTable, markerWeightSet, units.getAbbreviation());
 }
 
 void MarkersReference::loadMarkersFile(const std::string markerFile,
-                                       Units modelUnits) {
+                            const Set<MarkerWeight>* markerWeightSet,
+                            Units modelUnits) {
     auto fileExt = FileAdapter::findExtension(markerFile);
```
aymanhab: From a user perspective, it would be confusing for a method called loadMarkersFile  (that is supposed to take a file and create a table) to also take `Set<MarkerWeight>*`, can we refactor/rename it so the loading is just simple reading then operations are performed on result? Maybe also rename method populateFromMarkerData to something like populateFromDataAndWeights since Markers is implicit in the class name.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ v4.1
...
 - Added `OrientationsReference` as the frame orientation analog to the location of experimental markers. Enables experimentally measured orientations from wearable sensors (e.g. from IMUs) to be tracked by reference frames in the model. A correspondence between the experimental (IMU frame) orientation column label and that of the virtual frame on the `Model` is expected. The `InverseKinematicsSolver` was extended to simultaneously track the `OrientationsReference` if provided. (PR #2412)
 - Removed the undocumented `bool dumpName` argument from `Object::dump()` and made the method `const` so it can be safely called on `const` objects. (PR #2412)
+- `MarkersReference` convenience constructors were updated to take a const pointer to a `MarkerWeightSet` as its second argument. If a `Set` is provided, then only the markers listed are used as reference signals. That means `InverseKinematicsTool` no longer tracks all experimental markers even those not in the `MarkerWeightSet`. One can track all experimental markers (that have a corresponding model marker) by providing a nullptr or pointer to an empty `Set`, in which case all markers are assigned the default weight (typically 1.0).
```
aymanhab: I'd also note somewhere that the interface for MarkersReference has changed.

OpenSim/Simulation/MarkersReference.cpp
```diff
@@ -100,9 +103,26 @@ populateFromMarkerData(const TimeSeriesTable_<SimTK::Vec3>& markerTable,
         _markerTable.addTableMetaData("Units", units);
     }
     
+    
+    const auto& allMarkerNamesInFile = markerTable.getColumnLabels();
+
+    // If user specifies a MarkerWeightSet only track markers that it specifies
+    // therefore remove extraneous markers and their data from the Reference
```
aymanhab: Can we just change the label by prepending  'do-not-track-me-' and leave the rest of the data table alone?

OpenSim/Simulation/MarkersReference.h
aseth1: Yes the InverseKinematicsStudy (does not use IKTasks) uses the default arguments and only provides the filename to the constructor.
Also, the interface was minimally changed and made to be consistent with the other convenience constructor that takes a `TimeSeriesTable` (instead of file name) and already used a pointer to the `MarkerWeightSet`. I agree a const reference is a bit cleaner, but then it requires the caller to create an empty one.  What's your preference?

OpenSim/Common/DataTable.h
aseth1: We encourage users to use column names, but yes I think they would be annoyed if the order changed. I agree swapping for last column would be more efficient, but it will not appear to the user that a column is simply being removed.

OpenSim/Simulation/MarkersReference.cpp
aseth1: If a user wants to read a Table of Marker data they should NOT use a `MarkersReference` for that. We could rename `loadMarkersFile` to `initializeFromMarkerFile` but the method serves the `MarkersReference`-- it isn't a `DataAdapter`. I can refactor to make this method private and force Tools to use the constructor. Is that your preference?

CHANGELOG.md
aseth1: The text currently describes changes to the `MarkersReference`. What more of the interface change would you like to see. There is no reason for API users to `MarkersReference::loadMarkersFromFile` which is a convenience for `InverseKinematicsTool`.

OpenSim/Simulation/MarkersReference.cpp
aseth1: Why is that preferable? It then forces more logic downstream when the number of columns in the underlying container is different from the number of signals. Removal seems cleaner/hassle free to me.

OpenSim/Common/DataTable.h
aymanhab: Thanks @aseth1 I would just add a note to the documentation that this operation can be expensive (potentially copying all columns of the table if first data column is removed).

OpenSim/Simulation/MarkersReference.h
aymanhab: Considering where we are, I'd leave it as is to minimize change especially if not used in InverseKinematicsStudy.  It just stood out looking at the MarkerReference implementation the mixed use of references and pointers which is the price we pay.

OpenSim/Simulation/MarkersReference.cpp
aymanhab: Your call @aseth1 I think renaming would make the most sense because it's in the public interface.

CHANGELOG.md
aymanhab: Let's see where the API ends up then update accordingly, I wouldn't detail passing a pointer and nullptr since this is very c++ specific. and rather low level, I would describe the high level old vs new behavior, and reference the bug that triggered this as reported by @jimmyDunne

OpenSim/Simulation/MarkersReference.cpp
aymanhab: The cleanest solution, in my opinion, is to leave the data tables alone and build the logic downstream to filter/mask/weight. If we're going to mess with the Table (under the premise that it's a local copy) then the least expensive option (changing label to anything or even blank) would make more sense, but if there's much logic downstream to force tracking each and every column, I'd leave it as is to minimize change.

OpenSim/Common/DataTable.h
aseth1: My tests show that it only takes <75ms to delete 24 columns out of 25 with 50000 rows. Not sure that warrants a note.

OpenSim/Simulation/MarkersReference.h
aseth1: I went ahead and changed it for consistency. All arguments are const references.

OpenSim/Simulation/MarkersReference.cpp
aseth1: OK I renamed to `initializeFromMarkersFile` and left it in the public interface although it is only used by `InverseKinematicsTool`.

OpenSim/Simulation/MarkersReference.cpp
aseth1: I don't really follow your rationale. This is the downstream logic.  We are not "messing with the Table"-- the interface takes a const reference to a Table. The fact `MarkersReference` maintains the reference signals internally as a Table is not exposed anywhere to the user. This could be transformed to a Set of splines. This is all implementation detail the user never sees and at this stage trimming the data down to exactly what you need as the Reference signals seems perfectly reasonable/simple.

aseth1:
@jimmyDunne please let me know if this addresses #2443 as you would expect. Thanks!

OpenSim/Simulation/MarkersReference.cpp
aymanhab: 👍

OpenSim/Simulation/MarkersReference.cpp
aymanhab: I was advocating for not trimming (which I assumed is more expensive/error prone)  the table just changing the labels but sounds like this didn't turn out to be a major issue in practice, so we'll leave it as is.

aseth1:
Thanks @aymanhab and @jimmyDunne!

Issue #2463, opened 2019-04-10
Enable more user options/settings for transforming data


jenhicks:
- [ ] IMU world to OpenSim ground transformation and Heading correction are part of a single conversion step from experimental quaternions (of rotations in sensor world) to rotation matrices expressed in OpenSim Ground fame. We should abstract out the expected sensor world definition (e.g. Z-up and X-magnetic north) 

- [x] Enable user to identify the base frame (e.g. the pelvis) and the sensor axis that represents its forward direction, which we use to realign the IMU data with the heading (forward direction) of the base frame of the model.

aseth1:
We need to abstract out the builtin transform from IMU World to OpenSim ground because marker triads already transformed to OpenSim ground reference frame provide orientations in ground, so the additional transform is now an error. This is impacting Hannah, @malborno and team from producing results.

@jimmyDunne @aymanhab I will bump this up into the current sprint backlog.

aseth1:
@aymanhab this change somehow affected `TimeSeriesTableRotation` such that calling `table. getMatrix.getElt()` no longer works. It seems as though `SimTK::Matrix_<Rotation>` is no longer wrapped.

aymanhab:
@aseth1 Indeed you can't get the Matrix of Rotations directly per Sherm's feedback 
https://github.com/opensim-org/opensim-core/pull/2467#issuecomment-498876581, 
but there's a workaround. Take a look at the testcase Java_TestXsensDataReader and let me know if you can work around it.

aymanhab:
We resolved the issue with access to matrix elements (also used in java test case).  @malborno can you comment if this is useful and should be merged in or not yet, and if not what the hold-up, thanks.

PR #2501, opened 2019-06-18
Specify/modify the underlying IMU World to OpenSim ground rotation for OpenSense

aseth1:
Fixes issue #2463

### Brief summary of changes
- Added another optional argument to `OpenSense::convertQuaternionsToRotations` which is the assumed sensor world to opensim transform.  It has a default that works for most IMUs with Z-axis up (gravity) and X is the magnetic north direction.

### Testing I've completed
- test produce the same result as before with the default argument.

- testing with marker-based orientations

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2501)

### DIFF
--- a/OpenSim/Simulation/OpenSense/OpenSenseUtilities.cpp
+++ b/OpenSim/Simulation/OpenSense/OpenSenseUtilities.cpp
@@ -42,14 +42,6 @@ TimeSeriesTable_<SimTK::Rotation> OpenSenseUtilities::
     // OpenSim ground reference frame with Y up and X forward.
     SimTK::Rotation R_XG = sensorToOpenSim;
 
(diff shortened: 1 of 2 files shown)


aseth1:
@jimmyDunne and @malborno, Hannah is testing this fix. The code should be straightforward to review. Instead of just assuming the IMU world to OpenSim ground rotation, you can optionally set it as the final argument to the `convertQuaternionsToRotations ` utility function.

aymanhab:
@aseth1 Any update on the testing? Thanks

aseth1:
I believe the utility works as expected. The Rotation is applied. The issue is that we cannot find the correct Rotation that enables direct comparison if IMU vs Marker-based, but I think that is a separate issue.

aymanhab:
Thanks @aseth1

Issue #2418, opened 2019-02-27
set_GeometryPath() is setting incorrect socket_parent_frame properties


jimmyDunne:
Bug [seen on the forum](https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=10117&p=0&start=0&view=&sid=2e38ed32d9c097bc1f6a8e3d6c27e673) and confirmed by me. 

Constructing a PathSpring Geometry Path from a Muscle Geometry Path object using set_GeometryPath() is causing an incorrect socket_parent_frame, which results in an error on loading the model or initSystem()

``` Matlab
% OpenSim: addPathSpring 
import org.opensim.modeling.*

% Get a handle to the current model and create a new copy 
model = 'gait2392_simbody.osim';

% Set geometry path for the path spring to match the gastrocnemius muscle
gastroc = model.getMuscles().get('med_gas_r');    

pathSpring = PathSpring('BiarticularSpringDamper', 0.4, 1000, 0.01);
pathSpring.set_GeometryPath(gastroc.getGeometryPath());

%% Add the spring to the model
model.addForce(pathSpring);

```

The expected `socket_parent_frame` property for the `pathSpring` should be;

``` Matlab
>>> pathSpring.getGeometryPath().getPathPointSet().get(0).getPropertyByName('socket_parent_frame')
/bodyset/femur_r
```

Instead the result is 

``` Matlab
>>> pathSpring.getGeometryPath().getPathPointSet().get(0).getPropertyByName('socket_parent_frame')
/forceset/BiarticularSpringDamper/bodyset/femur_r
```

Then, calling `model.initSystem()` results in the following exception
``` Matlab
Java exception occurred:
java.io.IOException: Failed to connect Socket 'parent_frame' of type PhysicalFrame
(details: Component '3DGaitModel2392' could not find
'/forceset/BiarticularSpringDamper/bodyset/femur_r' of type PhysicalFrame. Make sure a
component exists at this path and that it is of the correct type.
	Thrown at Component.h:764 in getComponent().).
	In Object 'med_gas_r-P1' of type PathPoint.
	Thrown at Component.cpp:312 in finalizeConnections().
	at org.opensim.modeling.opensimSimulationJNI.Model_initSystem(Native Method)
	at org.opensim.modeling.Model.initSystem(Model.java:848)
```

I have tried a few work around but havent solved it— any suggestions?

jimmyDunne:
[...]

Issue #2388, opened 2018-11-20
Adding cloned component to a new model creates incorrect connectee path after prepend


nickbianco:
`Component::prependComponentPathToConnecteePath()` assumes that the absolute path of the provided subcomponent is always valid, which is not the case for components added to a model that we're cloned from a different model.

aymanhab:
@nickbianco you may want to check the build artifact of PR #2504 to see if the issue has been fixed. Thank you.

nickbianco:
This one slipped by me. Confirmed that the fix works for DeGrooteFregly2016Muscle in Moco, where we originally encountered it.

PR #2504, opened 2019-06-20
Fix set geometrypath issue 2418

aymanhab:
Fixes issue #2418

### Brief summary of changes
Do not prepend Componentpath to socket connectees if path is already valid

### Testing I've completed
Ran all test cases and tested script in issue and produced correct socket connectees

Not sure which test case to beef up to handle this change since it's generic and it can't be reproduced using a single component.

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2504)


aymanhab:
As of now this is PR is a place holder for discussion regarding the current behavior, no fix to merge yet.

aymanhab:
@chrisdembia The ComponentPath  for the PathSpring sockets gets corrupted during call to prependComponentPathToConnecteePath(*force); which you introduced as part of  commit 7b182f5eab5341af24b98e776d33912a5718bae7 (delay_connectee_name). The code in question is triggered only if "connecteepath.isAbsolute()" is true.  Is there any documentation as to how this flag is supposed to be maintained?

chrisdembia:
There is definitely a bug in the prepend code. For example, https://github.com/opensim-org/opensim-core/issues/2388

> Is there any documentation as to how this flag is supposed to be maintained?

I'm not sure I understand the question.

aymanhab:
Thanks @chrisdembia Bug #2388 is a duplicate of the same bug I'm looking at #2418, and I'm trying to understand what are the assumptions that the method prependComponentPathToConnecteePath is making about the added component. If you don't recall that's fine, I will keep digging. Thank you

aymanhab:
I believe the logic in the function prependComponentPathToConnecteePath should apply iff connections are within the added component/subcomponents (in addition to Absolute being true) and not otherwise.  I don't know how to check that connections are within the component so I check that the connection is not available from the Root instead. This fixes the bug and all test cases pass except for testComponentInterface which I'm investigating

aymanhab:
@chrisdembia and @aseth1 when you have a chance please let me know if you see something wrong with the change in Component.cpp. I'm debugging testComponentInterface to see why it fails and if the assumptions in it were justified.

OpenSim/Common/Component.cpp
```diff
@@ -163,12 +163,15 @@ void Component::addComponent(Component* subcomponent)
...
     for (auto& comp : subcomponent.updComponentList()) {
         for (auto& it : comp._socketsTable) {
-            it.second->prependComponentPathToConnecteePath(compPath);
+            if (!root.hasComponent(it.second->getConnecteePath()))
+                it.second->prependComponentPathToConnecteePath(compPath);
         }
         for (auto& it : comp._inputsTable) {
-            it.second->prependComponentPathToConnecteePath(compPath);
+            if (!root.hasComponent(it.second->getConnecteePath()))
+                it.second->prependComponentPathToConnecteePath(compPath);
```
chrisdembia: The connectees of inputs are not components.

aymanhab: Thanks @chrisdembia I undid the change for inputs, now all test cases pass.  Thank you :+1

aymanhab:
@jimmyDunne When you have a chance you can test the artifact to verify that the issue has been fixed.

aymanhab:
@chrisdembia this ended up being a two liner defensive fix, so should be easy to review when you have a chance.
@jimmyDunne Please let me know if you can test the fix using the build artifact. Thank you.

jimmyDunne:
I built locally and tested using the code in #2418. I didn't get the exception and the PropertyByName() returns the correct path. So LGTM!

aymanhab:
Thanks @jimmyDunne and @chrisdembia 👍

Issue #2438, opened 2019-04-19
Muscles that are disabled still produce activations in Static Optimization


jimmyDunne:
If you have muscles enabled == 0, activations are still produced for the disabled muscles that are colored and are also written to file. Below, the TRIlat and TRImed are disabled during the playback of a Static Optimization analysis. 

![arm25_SO](https://user-images.githubusercontent.com/5296507/56435006-e5bc6b00-62a4-11e9-8644-207e1932e5c9.gif)

Once the simulation is complete, the output activation files contain information for the muscles that should not have been part of the simulation. Below the TRIlat & TRImed (Green) produce activation.

![image](https://user-images.githubusercontent.com/5296507/56434728-d1c43980-62a3-11e9-9bca-bcd3612eb3b1.png)

There is no force information for the TRIlat and TRImed muscles written to file. Having any activations generated for disabled muscles is confusing, both in the visualization and in the output files. This will be fixed when disabled muscles do not produce any activations. 

A package with files need to replicate the issue is posted below;
[disabled_Muscle_SO.zip](https://github.com/opensim-org/opensim-core/files/3098811/disabled_Muscle_SO.zip)

jenhicks:
Thanks for reporting @jimmyDunne . I was able to reproduce this issue. Two comments/questions:

- I believe this is an issue only with Static Optimization. Forward and CMC seem to respect when muscles are disabled. (I ran CMC to track the same motion and the disabled muscles were not activated. I ran a forward simulation using the SO controls and the disabled muscles were not activated.)

- There is a second potential GUI issue. @aymanhab : disabled muscles seem to show up as always red when you visualize a states file, which is confusing (since red usually means activated). Is there already an issue for this or have we already discussed previously?

aymanhab:
@jenhicks As far as I know, the muscles provide the color (unfortunate in my opinion but true), when playing back motion or during simulation this is based on the state. The GUI does not manufacture colors on its own. We can change the behavior of the muscles so that they show some shade of gray if not activated, but I wouldn't call this strictly a GUI issue. I was advocating earlier in 4.0 that we shouldn't use any colors in the red-blue spectrum as default color to avoid confusion but we collectively decided not to change how models looked from 3.3 to 4.0. I don't believe there's an issue so feel free to open a new one.

aymanhab:
[...]

PR #2509, opened 2019-06-25
Handle disabled actuators so

aymanhab:
Fixes issue #2438
### Brief summary of changes
Make sure disabled actuators are ignored in Static Optimization. This is done by making the local copy of the model not contain any disabled actuators. Test case was beefed up accordingly.

### Testing I've completed
Ran SO with model that has disabled muscles and got no result for these disabled actuators in activations or forces. All existing test cases passed.

Verification of correct behavior (though hard to verify correct result since no canonical example). 

### CHANGELOG.md (choose one)
- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2509)


OpenSim/Analyses/StaticOptimization.cpp
```diff
@@ -553,7 +562,7 @@ int StaticOptimization::begin(const SimTK::State& s )
         sWorkingCopy.setTime(s.getTime());
         sWorkingCopy.setQ(s.getQ());
         sWorkingCopy.setU(s.getU());
-        sWorkingCopy.setZ(s.getZ());
+        // No need to copy Z as Actuators could be disabled/removed from _modelWorkingCopy
```
tkuchida: Does this mean that there will be different results across the entire Model if just one Actuator is disabled? Why not copy over the parts of `Z` that are still part of the model?

aymanhab: Thanks for the feedback @tkuchida If you actually look at the record method at the heart of SO you'll see that Z is never used/copied between states so it seems unnecessary here as well. Considering that Z contains only muscle states which are never used by SO. This felt totally reasonable. The fact that test cases pass suggests that these Zs were not actually used. On a side note I'm not aware of a method to generically copy Z  state variable values unless we cycle through components, know which ones have them and copy :pois

aymanhab: As to whether one disabled actuator changes results, I think yes and I also think that is the correct behavior. Having disabled actuators with min control and bounds in the model where the disabled flag is handled inconsistently would result in unpredictable results as shown in the reported issue.

tkuchida: Thanks, @aymanhab. Presumably, someone could create their own Actuator that has some important entry in Z, in which case Static won't work---but that probably isn't the only thing that wouldn't work. If the rationale for removing this line is as you said, I suggest changing the comment to something like "Z not copied because it isn't used by record()".

aymanhab:
To the general questions:
- Indeed we can reuse a model file but then the conversion of flag which changed names etc. may introduce uncertainty polluting the test, but let me know your preference. Good point.
- Earlier investigation by @jenhicks suggested that CMC and forward are handling the flag correctly so I didn't investigate further.

OpenSim/Analyses/StaticOptimization.cpp
tkuchida: > As to whether one disabled actuator changes results, I think yes and I also think that is the correct behavior.

What I meant was that the *mere* act of disabling a muscle (e.g., disabling a muscle that is attached to Ground at both ends) shouldn't change the results.

aymanhab: @tkuchida There's no change in computation whatsoever in the properly handled case with no disabled actuators. The behavior with disabled actuators was incorrect so no guarantee of no-change to this incorrect behavior 😃 The documentation should say something like: results were incorrect when some actuators were disabled and are now totally ignored"

aymanhab:
@jimmyDunne if you can test the artifact to verify the fix it would be great. Latest commit is a change in comment only FYI.

aseth1:
I agree @tkuchida that adding another model file and a setup file makes the testing environment more complicated. IMHO it is better to get the model from the Tool, disable the muscles by name and then have the test confirm that the (named) muscles do not produce activation and force results.

OpenSim/Analyses/StaticOptimization.cpp
tkuchida: > The documentation should say something like: results were incorrect when some actuators were disabled and are now totally ignored"

Ok, that should probably be mentioned in the changelog.

aymanhab:
Thanks for the feedback @tkuchida and @aseth1 the proposed changes are done.

jimmyDunne:
I built locally and ran through Matlab. If I use a model with disabled muscles, muscle activations and forces aren't written to file— which solved the initial issue. LGTM.

aymanhab:
Thanks @tkuchida much appreciated :+1

aymanhab:
@jimmyDunne, not sure if this needs more reviewers or can be merged. Your call until we have an updated policy in place.

OpenSim/Analyses/StaticOptimization.cpp
```diff
@@ -553,7 +562,7 @@ int StaticOptimization::begin(const SimTK::State& s )
         sWorkingCopy.setTime(s.getTime());
         sWorkingCopy.setQ(s.getQ());
         sWorkingCopy.setU(s.getU());
-        sWorkingCopy.setZ(s.getZ());
+        // No need to copy Zs to be consistent with record method below
```
chrisdembia: I don't understand why Zs are no longer copied, even for enabled muscles?

tkuchida: See thread here:

chrisdembia: Thanks @tkuchida . I like your Paint job.

Issue #2487, opened 2019-06-08
"NumMarkers" metadata isn't updated when dependent column appended to TimeSeriesTable


tkuchida:
### Python example

```
import opensim as osim
# read existing TRC file
markers = osim.TRCFileAdapter.read('origfile.trc')
# create and append new dependent column
newcol = osim.VectorOfVec3()
newcol.resize(markers.getNumRows())
newcol.setToZero()
markers.appendColumn('foo', newcol)
# write to file
osim.TRCFileAdapter.write(markers, 'newfile.trc')
# read from file
test = osim.TRCFileAdapter.read('newfile.trc')
```

### Expected result
File can be written and read back.

### Actual result
`IncorrectNumColumnLabels` exception (thrown from [here](https://github.com/opensim-org/opensim-core/blob/7e4f6ab1e3346f9e822a8a63e008ac16ec25b16d/OpenSim/Common/TRCFileAdapter.cpp#L121)).

### Workaround
It looks like the "NumMarkers" metadata key isn't being updated from its old value when `appendColumn()` is called, which is surprising since this key is created automatically if it doesn't exist (thus the user isn't aware of his/her responsibility to update it). Inserting
```
markers.removeTableMetaDataKey('NumMarkers')
```
before writing to file fixes the problem since the "NumMarkers" key will then be re-created automatically with the correct number of columns.

### Version
OpenSim 4.0 public release

jimmyDunne:
TRCFileAdapter.write() should update the meta based on the data.

jenhicks:
@jimmyDunne Need to add to changelog (under bug fixes)

PR #2510, opened 2019-06-26
TRCFileAdapter writes number of table columns when writing NumMarkers header data

aseth1:
Fixes issue #2487 

### Brief summary of changes
Updated `extendWrite` to always use the actual number of rows and columns in the table to set NumFrames and NumMarkers respectively in the header data.

### Testing I've completed
Ran similar test as posted in issue and NumMarkers was updated correctly.
`testTRCFileAdapter` failed initially because it correctly found the inconsistency in the test file with NumFrames not actually being the number of frames. NumFrames was updated in the test file.

### CHANGELOG.md (choose one)

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2510)


aseth1:
Thanks @tkuchida and @jimmyDunne!

Issue #2512, opened 2019-06-27
Function Based Bushing does not apply rotational damping


jimmyDunne:
{Reported during workshop}

aymanhab:
Thanks @jimmyDunne, Was there a sample model provided to work with?

jimmyDunne:
I edited an existing model.

jimmyDunne:
No, not that I am aware of.

aymanhab:
fixed by PR #2514

PR #2514, opened 2019-06-27
fix computation of damping force in function based bushing 2512

aymanhab:
Fixes issue #2512
### Brief summary of changes
Implemented the method extendFinalizeFromProperties on FunctionBasedBushing and used it to populate _dampingMatrix as was done in very early version of the code.

### Testing I've completed
Stepped through testForces and got non-zero damping force, also ran test suite. Beefed up testForces test case to exercise damping and verify that energy goes down,

If test is now adequate.

### CHANGELOG.md (choose one)
- Updated

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2514)


OpenSim/Simulation/Model/FunctionBasedBushingForce.cpp
```diff
@@ -186,6 +186,7 @@ SimTK::Vec6 FunctionBasedBushingForce::
     calcDampingForce(const SimTK::State& s) const
 {
     Vec6 dqdot = computeDeflectionRate(s);
+    // This _dampingMatrix is initialized to zero and never updated!
```
aymanhab: @aseth1 Where should this _dampingMatrix be computed/updated?  Is finalizeFromProperties() the correct place? Thanks.

aymanhab:
@tkuchida  with so many changes along the way I compared against v3.3 which was more monolithic, not using the Component interface/layout so likely this was lost in the translation to the new Component interface. Test case didn't capture it because it sets damping to 0. anyway

aymanhab:
@tkuchida Not sure that was what you have in mind, so appreciate another look over the test. I guess the trend makes sense the rest is rather unknown. Thank you :+1

OpenSim/Simulation/Test/testForces.cpp
```diff
@@ -848,12 +848,35 @@ void testFunctionBasedBushingForce()
...
+    double lastEnergy = 1E20;   // Large
+    for (int i = 1; i <= nsteps; ++i) {
+        osim_state2 = manager2.integrate(dt*i);
+        osimModel.getMultibodySystem().realize(osim_state2, Stage::Acceleration);
+        double newEnergy = osimModel.calcKineticEnergy(osim_state2) + osimModel.calcPotentialEnergy(osim_state2);
+        ASSERT(newEnergy < lastEnergy);
+        lastEnergy = newEnergy;
+    }
+
+    osimModel.disownAllComponents();
```
tkuchida: This comment from 4 years ago is relevant (and note that my code was just going into the sandbox): https://github.com/opensim-org/opensim-core/pull/663#discussion_r41592947

OpenSim/Simulation/Test/testForces.cpp
```diff
@@ -1096,13 +1142,36 @@ void testExpressionBasedBushingForceRotational()
...
     }
 
-    osimModel.disownAllComponents();
-
     manager.getStateStorage().print("expression_based_bushing_rotational_model_states.sto");
 
     // Save the forces
     reporter->getForceStorage().print("expression_based_bushing_rotational_model_forces.mot");
 
+    // Now add damping to the bushing force and make sure energy is monotonically decreasing
```
tkuchida: Could dump this code into a lambda since it's repeated 3 times.

OpenSim/Simulation/Test/testForces.cpp
aymanhab: True @tkuchida, though I didn't introduce this line I just moved it to reuse the model. I can add a comment stating this is bad if you think it would be helpful. We could use heap allocated objects and the line would not be needed but the code will look more ugly.

OpenSim/Simulation/Test/testForces.cpp
aymanhab: True 👍

OpenSim/Simulation/Test/testForces.cpp
tkuchida: I don't have a strong opinion (though others may). There's already an Issue about disowning the `disownAllComponents()` method (#148) which seems fine to me.

aymanhab: I added a comment as  suggested

OpenSim/Simulation/Test/testForces.cpp
aymanhab: Due to scope issues (function has to be called outside block) I used a separate utility function to avoid duplication. Please let me know what you think.

tkuchida: LGTM 🐴

aymanhab:
Thanks @tkuchida much appreciated

PR #2516, opened 2019-06-27
Update Moco branch with lenient search for simbody-visualizer.

chrisdembia:
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2516)


aymanhab:
@chrisdembia this branch is 400+ commits behind is there a plan for syncing these up so that users of moco don't need to shuffle Paths to get it working with an official OpenSim release?

chrisdembia:
Moco bundles a copy of opensim-core. But yes it would be great to sync up.

PR #2525, opened 2019-07-13
Use python3 as default on ci

aymanhab:
Fixes issue #0
Make Python 3 the default on CI

### Brief summary of changes
Change CMake option to default PYTHON_OPENSIM_VERSION set to 3.
Fix ci files accordingly.

### Testing I've completed
Built locally and tests pass on windows

### CHANGELOG.md (choose one)

- Will update once we agree to merge

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2525)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -269,7 +269,7 @@ script:
   # Go to the python wrapping package directory.
   - if [ "$WRAP" = "on" ]; then cd $OPENSIM_HOME/sdk/Python; fi
   # Run the python tests, verbosely.
-  - if [ "$WRAP" = "on" ]; then /usr/bin/python3 -m unittest discover --start-directory opensim/tests --verbose; fi
+  - if [ "$WRAP" = "on" ]; then python3 -m unittest discover --start-directory opensim/tests --verbose; fi
     
   ## Set up ssh for sourceforge.
   - if [[ "$DEPLOY" = "yes" || ("$DOXY" = "on" && "$TRAVIS_OS_NAME" = "linux" && "$TRAVIS_PULL_REQUEST" != "false") ]]; then PREP_SOURCEFORGE_SSH=0; else PREP_SOURCEFORGE_SSH=1; fi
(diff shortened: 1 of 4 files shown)


aymanhab:
Thanks @chrisdembia for your earlier work on python3 the switch was rather smooth, except that running tests on mac osx fails due to finding the wrong python. Any ideas? Thanks

.travis.yml
```diff
@@ -266,7 +269,7 @@ script:
   # Go to the python wrapping package directory.
   - if [ "$WRAP" = "on" ]; then cd $OPENSIM_HOME/sdk/Python; fi
   # Run the python tests, verbosely.
-  - if [ "$WRAP" = "on" ]; then /usr/bin/python -m unittest discover --start-directory opensim/tests --verbose; fi
+  - if [ "$WRAP" = "on" ]; then /usr/bin/python3 -m unittest discover --start-directory opensim/tests --verbose; fi
```
chrisdembia: This probably works only on Linux. We can probably change this to just `python3` now (remove `/usr/bin`, which was there for Mac).

appveyor.yml
```diff
@@ -40,8 +40,8 @@ branches:
   - /^(?i:devprop).*$/
 
 init:
-  # Note: python 2.7 32bit is already on the path. We want v2.7 64bit,
-  # so we must add v2.7 64bit earlier on the PATH so that CMake finds it when
+  # Note: python 3.7 32bit is already on the path. We want v3.7 64bit,
+  # so we must add v3.7 64bit earlier on the PATH so that CMake finds it when
```
chrisdembia: I don't know if this statement is true. You could just remove the comment or say "Avoid using other version of python that might already be on the PATH."

aymanhab: Done

.travis.yml
aymanhab: Done. Will see if that fixes the osx build without breaking the linux build. Thanks @chrisdembia

aymanhab:
Thanks @chrisdembia all tests now pass. Your call if another reviewer is needed ( :thinking)

aymanhab:
@jimmyDunne please let me know if you can test this. 
@chrisdembia any reason to hold this off? Testing on CI uses python3 now. Thanks

jenhicks:
Need to update the documentation (will need to time this update with the release).

jimmyDunne:
I was able to build against 3.7 locally without issue. 
image

Issue #2526, opened 2019-07-13
Add finalizeConnections() to multiple C++ Example files and throw exceptions on bad models


vinaym815:
Please add finalizeConnections() command to the the list of C++ examples that follows, as deserializing the saved model without it causes issue.

1) OpenSim/Examples/ExampleMain/OutputReference/TugOfWar_Complete.cpp   (line 289)
2) OpenSim/Examples/checkEnvironment/checkEnvironment.cpp  (line198)
3) OpenSim/Examples/CustomActuatorExample/toyLeg_example.cpp  (line 220)
4) OpenSim/Examples/MuscleExample/mainFatigue.cpp  (line 243)

tkuchida:
The lines of code you have referred to are calling `Model::print()`. If these lines don't work, it likely means that you are trying to save the file to a directory where you do not have write access.

vinaym815:
The printing works fine, but when I try to load the saved model in a different script and initialize it I face the  issue.

tkuchida:
Ah, thanks for clarifying. Yes, it seems that it is necessary to call `Model::finalizeConnections()` before serializing, as in [this example](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Test/testForces.cpp#L627). A potential source of confusion is that `Model::initSystem()` calls the `finalize...()` methods under the hood, so an explicit call to `finalizeConnections()` is not necessary if `initSystem()` has been called (which is true in most situations).

**Devs:** Perhaps `Object::print()` should call `finalizeConnections()` automatically (or at least throw an informative exception for the user to call `finalizeConnections()` if it can't be called from `Object::print() const`).

jimmyDunne:
Thanks, @vinaym815. Would you be willing to make the changes and submit a PR? We are always keen to get community contributions to improve the code 😉

aymanhab:
I'd vote to have print call the method internally unless there's strong reason against it. This has the benefit of not exposing and requiring users to understand the rather complex internals of how models are wired which may undergo changes in the future.

tkuchida:
@aymanhab Not sure that would be permitted/desirable since `Model::print()` is const, but I agree that it would be better if the user didn't need to worry about the internal details—at least in typical use cases.

aymanhab:
[...]

PR #2527, opened 2019-07-14
Added finalizeConnections() before the serialization of model to different examples

vinaym815:
Fixes issue #2526 

I have added finalizeConnections() to different C++ API examples before the model is serialized in them.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2527)

### DIFF
--- a/OpenSim/Examples/CustomActuatorExample/toyLeg_example.cpp
+++ b/OpenSim/Examples/CustomActuatorExample/toyLeg_example.cpp
@@ -217,9 +217,9 @@ int main()
         //Examine the model
         osimModel.printDetailedInfo(si, std::cout);
 
-	// To print (serialize) the latest connections of the model, it is
-	// necessary to finalizeConnections() first.`
-	osimModel.finalizeConnections();
+        // To print (serialize) the latest connections of the model, it is
+        // necessary to finalizeConnections() first.`
+        osimModel.finalizeConnections();
         // Save the model
         osimModel.print("toyLeg.osim");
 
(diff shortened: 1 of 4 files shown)


aymanhab:
@vinaym815 Please replace tabs with spaces as they causes ci to fail.

vinaym815:
I think these changes shall suffice till changes to model.print() functions are not made.

aymanhab:
I restarted the builds that were failing for unrelated reasons.

aymanhab:
@aseth1 on a related note,  I ran into some other code that was working around a similar issue with finalizeFromProperties by calling print on a clone of the model 
https://github.com/opensim-org/opensim-core/blob/840bfac5d28a70a16335f441950bf669407031e2/OpenSim/Simulation/Test/testContactGeometry.cpp#L174

Do you think clone should call finalizeConnections() as well?

aseth1:
@aymanhab we used clone to make sure that the non-const `print()` wasn't changing the model. Now that `print()` is const, there isn't a good reason to clone first IMHO. We can remove unnecessary clones. Note, that cloning is not the issue. If the original model topology changed (components added/removed) and you want those changes to be reflected in properties of Sockets, you must call finalizeConnections. In this case, the connections should be finalized before the `clone()`. The subsequent `print()` does not need to finalize to serialize since its properties are identical to the original.

aymanhab:
The PR can be merged when the tests pass  but I'd leave this issue open until we move the discussion there and/or open issues as needed. Thanks for your contribution @vinaym815

aymanhab:
Tests actually pass but uploading to sourceforge is broken (likely because of other admin/access issues). Merging

jimmyDunne:
Thanks, @vinaym815 for contributing! Much appreciated :)

vinaym815:
I just checked, their are other c++ examples with same command missing. Will add the same to them and submit a pull request.

PR #2529, opened 2019-07-18
Make warnBeforePrint throw exception unless raised debug level.

aymanhab:
Fixes issue #<issue_number>

### Brief summary of changes
Make warnBeforePrint method throw an exception if it detects missing connections rather than quietly produce corrupt model files. 

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2529)

### DIFF
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -1795,6 +1795,7 @@ void Component::warnBeforePrint() const {
                 << "::print(): The following connections are not finalized "
                    "and will not appear in the resulting XML file. "
                    "Call finalizeConnections() before print().\n"
+                   "To ignore, please call Object::setDebugLevel(1) first.\n"
                 << message << std::endl;
     }
 }
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -1791,12 +1791,14 @@ void Component::warnBeforePrint() const {
         }
     }
     if (!message.empty()) {
-        std::cout << "Warning in " << getConcreteClassName()
+        std::stringstream buffer;
+        buffer << "Warning in " << getConcreteClassName()
                 << "::print(): The following connections are not finalized "
                    "and will not appear in the resulting XML file. "
                    "Call finalizeConnections() before print().\n"
                    "To ignore, please call Object::setDebugLevel(1) first.\n"
                 << message << std::endl;
+        OPENSIM_THROW_FRMOBJ(Exception, buffer.str());
     }
 }
 
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -1796,7 +1796,8 @@ void Component::warnBeforePrint() const {
                 << "::print(): The following connections are not finalized "
                    "and will not appear in the resulting XML file. "
                    "Call finalizeConnections() before print().\n"
-                   "To ignore, please call Object::setDebugLevel(1) first.\n"
+                   "To ignore, set the debug level to at least 1 "
+                   "(e.g, by calling Object::setDebugLevel(1)) first.\n"
                 << message << std::endl;
         OPENSIM_THROW_FRMOBJ(Exception, buffer.str());
     }
--- a/OpenSim/Common/Object.cpp
+++ b/OpenSim/Common/Object.cpp
@@ -1367,9 +1367,15 @@ setAllPropertiesUseDefault(bool aUseDefault)
 bool Object::
 print(const string &aFileName) const
 {
-    try {
+    // Default to strict exception to avoid creating bad files
+    // but for debugging allow users to be more lenient.
+    if (_debugLevel >= 1) { 
+        try {
+            warnBeforePrint();
+        } catch (...) {}
+    }
+    else
         warnBeforePrint();
-    } catch (...) {}
     // Temporarily change current directory so that inlined files are written to correct relative directory
     std::string savedCwd = IO::getCwd();
     IO::chDir(IO::getParentDirectory(aFileName));
--- a/OpenSim/Common/Object.h
+++ b/OpenSim/Common/Object.h
@@ -815,7 +815,7 @@ private:
      * printing is aborted.
      * Derived classes can use this as an opportunity to issue warnings to users.
      */
-    virtual void warnBeforePrint(bool debug) const {}
+    virtual void warnBeforePrint() const {}
 
 //==============================================================================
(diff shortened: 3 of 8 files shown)


aymanhab:
These test cases all fail now:
	 23 - testForces (Exit code 0xc0000374)
	 38 - testActuators (Failed)
	 41 - testMuscles (Failed)
	 52 - testAddComponents (Failed)
	 73 - testBuildDynamicWalker (Failed)

OpenSim/Common/Object.cpp
```diff
@@ -1367,9 +1367,15 @@ setAllPropertiesUseDefault(bool aUseDefault)
 bool Object::
 print(const string &aFileName) const
 {
-    try {
+    // Default to strict exception to avoid creating bad files
+    // but for debugging allow users to be more lenient.
```
chrisdembia: Could you document this in the doxygen description for `print()`?

aymanhab:
@chrisdembia, thanks for the thoughtful feedback. I'm totally open to switching back the default after we fix the examples/test-cases, that's why I left it as WIP. Other options:
- Add an argument to the method to either throw or print.
- Suggest dump() instead of print, it's already there and can pinpoint the problem and do away with the if/else in print.
- Add a different method e.g. 'rawPrint' that doesn't perform the check.

Any thoughts or preferences?

Longer term I'd be happy to chat to understand the issue better and try to come up with solutions. Thank you :+1

aymanhab:
@jenhicks and @jimmyDunne Just a heads up that I'm working on this now as I think it's quite important that our test cases/examples produce correct model files.

aymanhab:
Build failures are related to compiler upgrade, but all test cases pass now with strict exception mode.

OpenSim/Common/Object.cpp
aymanhab: Done

aymanhab:
@chrisdembia build failures are unrelated. I opted not to modify the print signature per earlier discussions along #2526 and to use debug level which seems to be used rather sparingly. Please let me know what you think. Thank you.

OpenSim/Common/Object.h
```diff
@@ -808,11 +810,11 @@ class OSIMCOMMON_API Object
...
     void updateDefaultObjectsXMLNode(SimTK::Xml::Element& aParent);
 
-    /** This is invoked at the start of print(). Derived classes can use this
-     * as an opportunity to issue warnings to users.
-     * Any exception thrown in this function is ignored, as exceptions would
-     * prevent the user from printing the object, which could be useful for
-     * debugging. */
+    /** This is invoked at the start of print() if second argument to print is true
+     * printing is allowed to proceed even if the resulting file is corrupt, otherwise
+     * printing is aborted.
```
chrisdembia: There is no second argument to `print()`.

aymanhab: Nice catch @chrisdembia will update.

aymanhab: Updatd.

aymanhab:
Waiting for review when possible @chrisdembia Thank you

jenhicks:
@chrisdembia Can you review again to make sure all feedback has been addressed?

OpenSim/Common/Object.h
```diff
@@ -596,7 +596,9 @@ class OSIMCOMMON_API Object
     /** Write this %Object into an XML file of the given name; conventionally
     the suffix to use is ".osim". This is useful for writing out a Model that
     has been created programmatically, and also very useful for testing and
-    debugging. **/
+    debugging. If object has invalid connections, then printing is aborted.
+    You can override this behavior by calling Object::setDebugLevel(1) prior
+    to printing. **/
```
carmichaelong: I like below how it's precise that it's `if _debugLevel is greater than 0`, but I see why you want to give the full call here. Perhaps to do both: `You can override this behavior by setting the debug level to at least 1 (e.g., Object::setDebugLevel(1)) ...`

OpenSim/Common/Object.h
carmichaelong: Should this be `_debugLevel is at least 1`?

Also, looks like this should be two sentences: `This is invoked at the start of print(). If _debugLevel is greater than 0 then ...`.

OpenSim/Common/Component.cpp
```diff
@@ -1791,11 +1791,14 @@ void Component::warnBeforePrint() const {
         }
     }
     if (!message.empty()) {
-        std::cout << "Warning in " << getConcreteClassName()
+        std::stringstream buffer;
+        buffer << "Warning in " << getConcreteClassName()
                 << "::print(): The following connections are not finalized "
                    "and will not appear in the resulting XML file. "
                    "Call finalizeConnections() before print().\n"
+                   "To ignore, please call Object::setDebugLevel(1) first.\n"
```
carmichaelong: Same comment as in `Object.h`. Could be something like `To ignore, please set _debugLevel to at least 1 (e.g, by calling Object::setDebugLevel(1))`

aymanhab: I prefer not to show private variable names to API users in case we change how this is implemented in the future, also non C++ users have no choice but call setDebugLevel, so I'd prefer to keep user facing message at this level.

aymanhab:
Thanks for the quick turnaround @carmichaelong much appreciated, updated accordingly.

OpenSim/Common/Component.cpp
carmichaelong: Thanks for the clarification, this makes sense. I think it's fine as is, or a middle ground could be then just to have a similar message as the one above where it doesn't show the private variable name:
`To ignore, please set the debug level to at least 1 (e.g, by calling Object::setDebugLevel(1)) first.`

carmichaelong:
@aymanhab Just one response to your comment that I agreed with. I think this PR is fine as is too.

aymanhab:
Thanks @carmichaelong 💯 made the change you suggested. Build failures are unrelated.  It would be great iff you can approve and merge, or let me know and I'll merge.

carmichaelong:
I think I'll defer merging since there was some discussion between you and @chrisdembia from before, and it's unclear to me whether or not that has been sufficiently addressed (or if they were just ideas for the longer term?)?

chrisdembia:
Sorry for the delay.

aymanhab:
Thanks @chrisdembia and @carmichaelong  much appreciated. Merging.

PR #2530, opened 2019-07-19
Add function approximation to GeometryPath

chrisdembia:
### Brief summary of changes

This PR adds to GeometryPath the ability to approximate the path's length, lengthening speed, and moment arms using any Function for the path's length.

For OpenSim to properly support function-approximated muscle paths, OpenSim needs multivariate Functions. For now, @antoinefalisse added a multivariate function to Moco, but it is not ready for the opensim-core repository. Hopefully, we can add this feature gradually. Adding N-dimensional functions isn't as straightforward as I had hoped; we might need a third-party dependency.

The code for employing the function approximation to compute length, lengthening speed, and moment arms comes from @antoinefalisse.

For now, the GeometryPath connects to Coordinates using a list of strings rather than sockets. This is because we do not yet support list sockets.

### Testing I've completed

I added test cases to fit a polynomial to the path point-based functions and tested various error cases. 

Naming of the new properties in OpenSim.

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2530)


aseth1:
I like this feature a lot and have been supportive of adding it to OpenSim. It is necessary for speedups for SCONE and MOCO. I do not, however, think the proposed implementation is ideal. IMHO it would be much less confusing and error prone to make `GeometryPath` abstract and have two concrete `GeometryPath` types: something like `PointBasedPath` (or `LineSegmentPath`, ...) and `FunctionBasedPath` (or `SplinedPath`). The job of `GeometryPath` (or suitable abstract class name) is to provide path length and velocity given a state and it would implement corresponding Outputs given the public base class methods `getLength()` , `getLengtheningSpeed()` and effective moment-arm (which all already exist).  The two concrete classes would have very different implementations and separate. 

In the current solution it is unclear which function should be used and what happens to the other call when the flag is set.  It also seems error prone and difficult to debug, if for example, the function is completely different (even opposite sign)  with respect to the current point-based path.  It seems if the fundamental behavior and values can be utterly different it shouldn't belong to the same class even less the same instance.

Instead, we would have a utility/converter that takes a `PointBasedPath`  and converts it a `FunctionBasedPath` and that could be performed  automatically by MOCO or Scone to simplify the model and speedup computation.

I totally agree with starting simple, and having a `FunctionBasedPath` or `SplinedPath` that only handles a single coordinate and then working our way up to a more general Multivariate function as the need arises.

chrisdembia:
Tabling this issue for now. Please do not delete the branch.

PR #2531, opened 2019-07-22
[WIP] [travis] Updates for Ubuntu 16.04

aymanhab:
[skip appveyor]

Fixes issue #<issue_number>
Build failure with clang

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2531)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,10 +49,10 @@ env:
     - OPENSIM_DEPENDENCIES_INSTALL_DIR=~/opensim_dependencies-install
     - SWIG_VER=3.0.8
     - PATH="$PATH:$TRAVIS_BUILD_DIR/.github"
-    - USE_CCACHE=0
-    - CCACHE_COMPRESS=0
+    - USE_CCACHE=1
+    - CCACHE_COMPRESS=1
     # For Clang to work with ccache.
-    - CCACHE_CPP2=0
+    - CCACHE_CPP2=1
 
 cache: ccache
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,15 +59,6 @@ cache: ccache
 addons:
   # Dependencies on linux.
   apt:
-    sources:
-      # For gcc >= 4.8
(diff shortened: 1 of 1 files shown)


aymanhab:
@chrisdembia build on osx with latest clang worked but other linux builds broke, was there a reason for the change on linux?

chrisdembia:
Yes, the Linux build is failing otherwise.

```
../../../libosimCommon.so: undefined reference to `SimTK::canonicalizeTypeName(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&&)'
/home/travis/opensim_dependencies-install/BTK/lib/btk-0.4dev/libBTKIO.so: undefined reference to `std::ios_base::failure::failure(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Seems like we should patch BTK.

jimmyDunne:
Another reason  for us to start really looking at switching out btk for [ezc3d](https://github.com/pyomeca/ezc3d)

aymanhab:
@chrisdembia the compilation error here is different from what we had before the upgrade, and is in a branch of BTK that is neither Little_Endian or Big_Endian, do you know the correct architecture for the linux build machines? Thanks.

chrisdembia:
Don't you think if we add the static_casts that the compiler is suggesting that then the compiler errors will go away? I'm thinking that there is no branch of BTK that works with this recent of a compiler.

aymanhab:
@chrisdembia Indeed I can sprinkle static casts to fix compilation errors but according to 
https://github.com/BurntSushi/byteorder/issues/106
 travis supports Big_Endian and Little_Endian byte order, so the fact that the branch for neither is being compiled suggests a problem upstream with compilation flags, but with no support behind btk we'll have to guess.

Issue #2499, opened 2019-06-17
Using read() or readFile()


jimmyDunne:
Differentiate the use of read() and readFile() across the FileAdapaters(). 

We should be able to answer the question "Why do we have multiple methods that do similiar things?"

tkuchida:
IMHO, it seems like there should be only one method. If C3DFileAdapter wants to implement its own read method, it should either override the method from the parent with its own implementation or override the method from the parent to just throw an exception that points to its own method. A potential danger in the current interface is that a user might be using one of the read methods and not know about the other one where you can specify how the forces are expressed.

jenhicks:
Proposal: Make convenience constructors (readFile() ) protected, and instead have users create an instance of reader, set parameters, and then use generic read function.

reader = C3DFileAdapter()
reader.setCOPlocation(...)
tableMap = reader.read()
forces = reader.getTable(tableMap, "Forces")

Changes to interface should not have a huge effect on users as long as we update matlab utility scripts. @tkuchida  how does this sound to you? How big of an effect would this on your existing code?

tkuchida:
LGTM, just so long as there's only one read-related method in the public interface. :book:

aymanhab:
Thanks for the feedback @tkuchida Some investigation shows the following items that should be good to discuss at some point:

1.  The factory feature of FileAdapter::readFile is used in STOFileAdapter_<>::readFile so that the return type of the table is dictated and so that the concrete type of Adapter is hidden. 
2. Same issue for TimeSeriesTable constructor from file, and for Storage constructor from file.
3. In case one Table is returned, the generic interface returns a map from name to a table, we'll need to make-up a name that we didn't have to introduce in the past.

I'll proceed with the assumption that this is the intended behavior per earlier discussion and we can discuss in person or along the review. I will not remove readFile function itself for now, just its usage, in case we want to return it to life under a different name or scope.

chrisdembia:
As I mentioned in the dev meeting, these two functions actually have two different uses. `read()` is for use by users, and `readFile()` is for use by tools. Users should not really ever need `readFile()` unless they do not want to assume the type of file they are reading.

I think the issue is mostly just one of naming. `readFile()` should be renamed to something like `readGeneric()` or `readAnyFile()` or `readFactory()`.

aymanhab:
[...]

PR #2535, opened 2019-07-25
Eliminate readfile

aymanhab:
Fixes issue #2499

### Brief summary of changes
Change client code to use the generic read method rather than readFile. In most cases (when one Table is populated from the file) the Adapters are not exposed altogether, read method is used internally.

### Testing I've completed
Ran test suite, client code in python, Matlab was tested and ok'd by @jimmyDunne 

### CHANGELOG.md (choose one)
- updated by adding a line item

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2535)


jimmyDunne:
For TRCFileAdapter(), read() is returning a map and not the concrete type. I am assuming that read should be returning the concrete type in this case. 

```
>>> trc_filename = 'arm26_elbow_flex.trc';
>>> trc = TRCFileAdapter();
>>> trcTable = trc.read(trc_filename);
>>> whos trcTable

 Name          Size            Bytes  Class                                                 Attributes
  trcTable      1x1                    org.opensim.modeling.StdMapStringAbstractDataTable  

```

It seems that the STOFileAdapter() doesn't have access to either read method at the moment
```
>>> sto_filename = 'Running_GRF_cop_nans.mot';
>>> sto = STOFileAdapter();
>>> stoTable = sto.read(sto_filename)
Undefined function 'read' for input arguments of type 'org.opensim.modeling.STOFileAdapter'.

```

aymanhab:
James,

Thanks for testing.

The new paradigm is to not use the adapter explicitly altogether. Just
create a table of the proper concrete type from the constructor that takes
the filename.

I would recommend against having the users see or use the adapter for
either trc or sto files. The diffs in the PR for java, python should be a
good example.

Ayman

OpenSim/Common/C3DFileAdapter.cpp
```diff
@@ -42,7 +42,7 @@ C3DFileAdapter*
 C3DFileAdapter::clone() const {
     return new C3DFileAdapter{*this};
 }
-
+#if 0
```
carmichaelong: Looks like the PR is in good enough shape that all the code with `#if 0 ...` could be removed at this point.

OpenSim/Common/Test/testTRCFileAdapter.cpp
```diff
@@ -155,12 +155,12 @@ int main() {
...
     }
 
-    std::cout << "Testing FileAdapter::readFile() and FileAdapter::writeFile()"
+    std::cout << "Testing FileAdapter::read() and FileAdapter::writeFile()"
               << std::endl;
     for(const auto& filename : filenames) {
         std::cout << "  " << filename << std::endl;
         try {
-            auto table = FileAdapter::readFile(filename).at("markers");
+            auto table = FileAdapter::createAdapterBasedOnExtension(filename)->read(filename).at("markers");
```
carmichaelong: Are there any other use cases for `createAdapterBasedOnExtension()`? This test is aimed at testing `read()`, but if there are other use cases it would be good to add them somewhere.

OpenSim/Common/C3DFileAdapter.h
```diff
@@ -40,8 +40,8 @@ class OSIMCOMMON_API C3DFileAdapter : public FileAdapter {
     /** Enumerated list of locations in which read in forces are expressed.
         %Measurement from force plates can be expressed by the C3DFileAdapter
         either at the OriginOfForcePlate (the default), CenterOfPressure, or
-        the PointOfWrenchApplication. It is an optional argument to 
-        C3DFileAdapter::readFile().
+        the PointOfWrenchApplication. You need to call  setLocationForForceExpression
+        before invoking C3DFileAdapter::read().
```
carmichaelong: When I first read this, I thought it was a requirement to call `setLocationForForceExpression`, but it's to change location if you don't want `OriginOfForcePlate`. Perhaps it could say `If you want to change how the measurement from force plates are expressed, call setLocationForForceExpression(ForceLocation)` before invoking `C3DFileAdapter::read().`.

OpenSim/Common/C3DFileAdapter.h
```diff
@@ -91,15 +97,18 @@ class OSIMCOMMON_API C3DFileAdapter : public FileAdapter {
     ~C3DFileAdapter()                                = default;
 
     C3DFileAdapter* clone() const override;
-
+    /**  C3DFileAdpater provides options for expressing the force-plate 
+        measurements either as the net force and moments expressed at the 
+        ForcePlateOrigin, the CentereOfPressure, or the
```
carmichaelong: `CentereOfPressure` --> `CenterOfPressure`

OpenSim/Common/C3DFileAdapter.h
```diff
@@ -91,15 +97,18 @@ class OSIMCOMMON_API C3DFileAdapter : public FileAdapter {
     ~C3DFileAdapter()                                = default;
 
     C3DFileAdapter* clone() const override;
-
+    /**  C3DFileAdpater provides options for expressing the force-plate 
+        measurements either as the net force and moments expressed at the 
+        ForcePlateOrigin, the CentereOfPressure, or the 
+        PointOfWrenchApplication  This function sets the option. */
```
carmichaelong: `PointOfWrenchApplication. This function sets the option.`

OpenSim/Common/FileAdapter.h
```diff
@@ -175,7 +175,7 @@ class TableMissingHeader : public Exception {
 };
 
 /** FileAdapter is a DataAdapter that reads and writes files with methods
-readFile and writeFile respectively.                                          */
+read and writeFile respectively.                                          */
```
carmichaelong: Since `read()` isn't defined here, it may be confusing to users that they won't find the function prototype here. Could be useful to spend more words in the documentation here to explain how this works (e.g., read() is inherited from `DataAdapter`, child classes implement `extendRead()`, etc.). The suggestions are just rough ideas, no need to implement all of them necessarily. Also, if you think doxygen will be sufficient at picking this up, no need to add extra information either.

OpenSim/Common/C3DFileAdapter.h
aymanhab: Thanks, excuse my French 😃

OpenSim/Common/Test/testTRCFileAdapter.cpp
aymanhab: For TRCFileAdapter there are no other use cases, you always create a table of concrete type from the file and TRCFileAdapter is used internally, I just didn't want to weaken the test.

OpenSim/Common/C3DFileAdapter.cpp
aymanhab: Done

OpenSim/Common/C3DFileAdapter.h
aymanhab: Clarified

OpenSim/Common/FileAdapter.h
aymanhab: Clarified.

aymanhab:
@carmichaelong I updated the comments based on your feedback, feel free to re-review and mark resolved conversations as s such to reduce clutter. 
@chrisdembia I responded to your feedback on reviewable, please let me know what you think.

Thanks for the quick turnaround guys 👍

carmichaelong:
Thanks @aymanhab for addressing my comments. Reviewing again, I had one more suggestion. It might be good to strengthen the documentation for `DataAdapter::read()` since it seems that we want users to avoid using this and use the constructors as mentioned above:

> The new paradigm is to not use the adapter explicitly altogether. Just
create a table of the proper concrete type from the constructor that takes
the filename.

After this, LGTM pending addressing review from @chrisdembia.

aymanhab:
Your interpretation is correct @chrisdembia I'm sure it would be helpful to users but don't know if it belongs in change log, thoughts @jenhicks @carmichaelong?

jenhicks:
I think it would be OK to put this in the changelog.

carmichaelong:
I agree that this would be good to document in the change log.

jenhicks:
We should update the matlab utility files and perhaps create another example.

aymanhab:
This addresses issue #2499

aymanhab:
@chrisdembia Appreciate if you take a look and respond if latest changes (mainly to documentation) are satisfactory. changelog has also been updated accordingly. Thank you,

aymanhab:
> > Reviewed 1 of 1 files at r6, 2 of 2 files at r7, 1 of 1 files at r8.
> > _[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2535)_ status: all files reviewed, 7 unresolved discussions (waiting on @aseth1, @aymanhab, @carmichaelong, and @jimmyDunne)
> 
> _[OpenSim/Common/C3DFileAdapter.h, line 129 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2535#-LmVM4iO5lcvwenSmtrI:-LmgFj0q-jjXEFFBAHfM:b-r8ssrp) ([raw file](https://github.com/opensim-org/opensim-core/blob/c92ebc4e2fc938a964f2bc3dd1cb13dd59dd4fa6/OpenSim/Common/C3DFileAdapter.h#L129)):_
> 
> Previously, aymanhab (Ayman Habib) wrote…
> `StatesTrajectory::exportToTable()`, `Storage::exportToTable()`.

Thanks @chrisdembia Convenience accessors were renamed accordingly, doxygen added.

aymanhab:
@chrisdembia Please let me know if you have any further questions or concerns. The failures were due to renaming of accessors in bindings which are now updated. Thank you

aymanhab:
Thanks much @chrisdembia @carmichaelong and @jimmyDunne for the feedback and review. Merging.

PR #2539, opened 2019-08-15
Weaken StatesTrajectory::isCompatibleWith()

chrisdembia:
The current implementation of StatesTrajectory::isCompatibleWith() does not properly handle models with (unused) quaternion slots. This PR removes checks that would incorrectly fail for models and states that are actually compatible.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2539)

### DIFF
--- a/OpenSim/Simulation/StatesTrajectory.cpp
+++ b/OpenSim/Simulation/StatesTrajectory.cpp
@@ -98,12 +98,9 @@ bool StatesTrajectory::isCompatibleWith(const Model& model) const {
     // need to check if the first one is compatible with the model.
     const auto& state0 = get(0);
 
-    if (model.getNumStateVariables() != state0.getNY()) {
-        return false;
-    }
-    if (model.getNumCoordinates() != state0.getNQ()) {
-        return false;
-    }
+    // We only check the number of speeds because OpenSim does not count
+    // quaternion slots, while the SimTK State contains quaternion slots even if
+    // quaternions are not used.
     if (model.getNumSpeeds() != state0.getNU()) {
         return false;
     }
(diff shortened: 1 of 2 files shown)


chrisdembia:
This is needed for Moco.

chrisdembia:
Thanks @tkuchida

jenhicks:
@chrisdembia will check whether there is a test and write one, if not.

chrisdembia:
@aymanhab we have a test case: https://github.com/opensim-org/opensim-core/blob/fc3977bf24d607c0e3e4d68aeb2abc81eb1fb577/OpenSim/Simulation/Test/testStatesTrajectory.cpp#L635-L661

aymanhab:
Great, thanks @chrisdembia LGTM.

chrisdembia:
Thanks @aymanhab

tkuchida:
@chrisdembia Is there an easy way to work around this issue in Python for 4.0 users?

chrisdembia:
I don't think so :(. Are you using `StatesTrajectory.createFromStatesStorage()`?

tkuchida:
Ok, thanks. I'm getting the error at `exportToTable()`:
```
RuntimeError: std::exception in 'OpenSim::TimeSeriesTable
    OpenSim::StatesTrajectory::exportToTable(OpenSim::Model const &) const':
    The provided model 'myModel' is not compatible with the StatesTrajectory.
Thrown at statestrajectory.cpp:130 in exportToTable().
```
The StatesTrajectory was built by appending rows from a StatesTrajectoryReporter. No errors or warnings when appending, just [this apparent false positive](https://github.com/opensim-org/opensim-core/blob/4.0/OpenSim/Simulation/StatesTrajectory.cpp#L129) when trying to export the table. The issue originates from [these comparisons](https://github.com/opensim-org/opensim-core/blob/4.0/OpenSim/Simulation/StatesTrajectory.cpp#L101-L109) but it isn't apparent to the user why the numbers shouldn't match. No worries… there are several hacks that will work.

chrisdembia:
Yeah, I think I was overeager with those checks.

PR #2540, opened 2019-08-15
Add destructor to Model.

chrisdembia:
### Brief summary of changes

This PR fixes an issue of dangling pointers to `SimTK::Subsystems` by ensuring that subsystems are destroyed in the correct order.

### Testing I've completed

Ran all CI tests, and we've used this change in Moco all year.

### CHANGELOG.md (choose one)

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2540)


chrisdembia:
Needed for Moco.

PR #2546, opened 2019-08-26
Upgrade osimC3D and corresponding test case 

aymanhab:
to use the latest interfaces of DataAdapters

Fixes issue #0

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2546)


jimmyDunne:
Tested and works great. Tests pass locally.

PR #2548, opened 2019-08-27
Try out GitHub Actions

chrisdembia:
This PR is an attempt to try out GitHub Actions. To actually try this out, I think this needs to be merged into the master branch. @aymanhab could you review/merge?

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2548)

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -4,7 +4,7 @@ name: continuous-integration
 # syntax https://help.github.com/en/articles/workflow-syntax-for-github-actions
 on:
   schedule:
-    - cron: '0 2 * * *'
+  - cron: '0 2 * * *'
     
 jobs:
   windows:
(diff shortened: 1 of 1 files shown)


chrisdembia:
There is no local testing.

I don't know if we should merge this yet. Perhaps we could discuss.

This CI has some shortcomings as is, but those shortcomings are resolvable I think.

chrisdembia:
@aymanhab would you be interested in getting the Windows CI to work? I can't figure out the syntax. I think GitHub Actions uses CMD.exe (batch) to run the commands but I keep getting syntax errors. We need to find examples of a GitHub Actions Workflow written for Windows.

Then the main remaining task is uploading artifacts to Sourceforge. I can find time to add this eventually. Ideally, we would use GitHub to host the artifacts but I don't know if that's possible yet.

aymanhab:
Sure, @chrisdembia I'll look into the Windows build.

aymanhab:
Windows and Mac now pass tests, while linux fails python tests. Did we ever try these? shared libraries are not loading in python environment.

chrisdembia:
Neat! That's exciting. I cannot figure out what's wrong with the Python tests on Linux. It looks like the Travis and GitHub commands are very similar.

chrisdembia:
It would be great to build bindings on Windows.

aymanhab:
@chrisdembia Should we skip install on linux for now to get the green check mark and continue trying/testing on a separate PR? The additional environments provided by the new actions are worth keeping.

aymanhab:
@chrisdembia I think this can be merged as it builds on more environments and keeps us up-to-date with no downside, please let me know if you disagree. Thank you.

chrisdembia:
I suppose that's fine but I don't want to waste computational resources and create clutter in the PR status. We can try it out for a bit and re-evaluate soon.

aymanhab:
The resource usage point is definitely valid, maybe we can switch the build to nightly instead of on push. It is just good to have this live so we can tweak it to do more of the matrix and eventually use it to upload/publish.

chrisdembia:
Nightly is a good idea. We can also merge the PR and turn off the builds in Settings. Up to you.

aymanhab:
I made the change to nightly and will merge unless I hear otherwise from you @chrisdembia We should find out tomorrow how that looks or if it ran.

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,128 @@
...
+    - cron: '0 2 * * *'
+    
+jobs:
+  windows:
+    name: Windows
+
+    runs-on: windows-2016
+
+    steps:
+     - uses: actions/checkout@v1
```
chrisdembia: Indentation is inconsistent throughout the file. (line 6 is indented 2 spaces, line 90 is indented 0 spaces from the previous line).

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,128 @@
+name: continuous-integration
+
+# TODO: build master branch after PRs are merged.
+on:
+  schedule:
+    - cron: '0 2 * * *'
```
chrisdembia: Could you document the syntax?

aymanhab: The syntax for the whole file and specific lines was linked to. Good idea.

.github/workflows/continuous_integration.yml
aymanhab: Last edits I made were in web interface of github so using spaces only. I just followed the examples on github for indentation, if incorrect the script will not run.

PR #2551, opened 2019-08-29
Integrate spdlog

aymanhab:
Fixes issue #0
### Brief summary of changes
Integrate spdlog into the build/installation process. 

### Testing I've completed
Local and  ci builds are running, added one use case in Object.cpp in prep for a full roll-out.

Build mechanism uses dependencies and SUPERBUILD exclusively rather than complicated CMake structures/commands. Would be good to discuss this vs how Simbody, BTK are integrated and how to scale.

### CHANGELOG.md (choose one)

- no need to update unless we expose to users.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -749,6 +749,15 @@ if(${OPENSIM_COPY_DEPENDENCIES})
                 DESTINATION "${CMAKE_INSTALL_LIBDIR}")
     endif()
 
+	#SPDLOG
+	#---
+	set(SPDLOG_HOME "${OPENSIM_DEPENDENCIES_DIR}/spdlog")
+    if(WIN32)
+         install(DIRECTORY "${SPDLOG_HOME}/include"
+                DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
+
+		include_directories("${SPDLOG_HOME}/include")
+	endif()
 endif()
 
 if(BUILD_PYTHON_WRAPPING AND OPENSIM_PYTHON_STANDALONE)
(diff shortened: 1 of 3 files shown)


chrisdembia:
@noahggordon

chrisdembia:
@aymanhab I've updated this PR. I've set up the infrastructure so that we could replace the old mechanism the GUI used for populating the Messages window. Could you review? Before merging, we would need to move all uses of `std::cout` to the new logging functions.

aymanhab:
Thanks much for working on this @chrisdembia Will likely wait on merging until after 4.1 since it's still a work in progress and it adds a new library to the build/install mechanism, but I'll review anyway.

chrisdembia:
That’s fine with me. I would like the code to be reviewed soon but it’s fine if it’s not merged for a few months.

chrisdembia:
Based on conversation with Ayman , I will create a feature branch, split into multiple PRs, fix the MATLAB logging issue, and add a command line option.

aymanhab:
Skimming through the changes, looks very good. The issue I have is the mix of changes to build system, new abstractions and changes to our old code to use the new abstractions, if we create a "feature" branch for logging, first PR with only CMake changes to it, that would be great, 2nd PR into the branch would introduce new abstractions, bindings, third PR would use variety of these new abstractions/options that would help move this along fast. Very excited about this. Thanks @chrisdembia 🥇

carmichaelong:
Making a note here (from #2685) that logging was lost from IK. Does not necessarily need to be addressed in this PR.

PR #2552, opened 2019-08-30
Fix GCC 8 warnings about deprecated throw() and incomplete type.

chrisdembia:
This PR fixes some warnings generated by GCC-8 about throw() being deprecated and Component being an incomplete type in ComponentSocket.h.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2552)

### DIFF
--- a/OpenSim/Common/Component.h
+++ b/OpenSim/Common/Component.h
@@ -24,19 +24,19 @@
  * -------------------------------------------------------------------------- */
 
 /** @file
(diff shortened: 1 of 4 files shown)


chrisdembia:
I'm copying what we've already done for other functions in `Socket`.

The code is templated so it must be in a header, but the code is using functions in `Component` so the code must come after the definition of the `Component` class.

PR #2556, opened 2019-09-05
Update osimC3D and test

jimmyDunne:
#### osimC3D.m 
expose convertMillimeters2Meters(). This was previously performed internally only when writing the forces to file during and should be user-controlled. 

#### testOsimC3D.m 
Updated test so that it runs in significantly less time (120 seconds -> 57 seconds on my local machine). 
- I compare OpenSim table values, rather than Matlab Structures, since the conversion of tables to structs is expensive. 
- I only test rotations about 1 axis. 
- Added test for convertMillimeters2Meters() (not previously tested).

#### c3dExport.m 
included improved use of convertMillimeters2Meters() in example script

### Testing I've completed
Running ctest locally (passed) and checking the outputs on various C3D files (locally). 

### CHANGELOG.md (choose one)
- updated since users will have to change their code to explicitly use convertMillimeters2Meters() before writing to file.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2556)

### DIFF
--- a/Bindings/Java/Matlab/Utilities/osimC3D.m
+++ b/Bindings/Java/Matlab/Utilities/osimC3D.m
@@ -38,7 +38,7 @@ classdef osimC3D < matlab.mixin.SetGet
         ForceLocation
     end
     methods
-        function obj = osimC3D(path2c3d, ForceLocation)
+        function self = osimC3D(path2c3d, ForceLocation)
             % Class Constructor: input is an absolute path to a C3D file.
 
             % Verify the correct number of inputs
@@ -53,8 +53,8 @@ classdef osimC3D < matlab.mixin.SetGet
                 if isempty(path)
                     error('Input path must be full path to file (C:/data/Walking.mot)')
                 end
-                obj.path = path;
-                obj.name = name;
+                self.path = path;
+                self.name = name;
             end
             % Verify the ForceLocation input is correct
             if  ~isnumeric(ForceLocation) || ~ismember(ForceLocation,[0:2])
@@ -67,8 +67,8 @@ classdef osimC3D < matlab.mixin.SetGet
 			c3dAdapter.setLocationForForceExpression(ForceLocation);
             tables = c3dAdapter.read(path2c3d);
             % Set the marker and force data into OpenSim tables
-            obj.markers = c3dAdapter.getMarkersTable(tables);
-            obj.forces = c3dAdapter.getForcesTable(tables);
+            self.markers = c3dAdapter.getMarkersTable(tables);
+            self.forces = c3dAdapter.getForcesTable(tables);
             % Set the force location specifier in case someone wants to
             % check.
             switch(ForceLocation)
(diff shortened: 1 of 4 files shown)


Bindings/Java/Matlab/Utilities/osimC3D.m
```diff
@@ -150,6 +150,32 @@ function rotateData(obj,axis,value)
             obj.rotateTable(obj.forces, axis, value);
             disp('Marker and Force tables have been rotated')
         end
+        function convertMillimeters2Meters(obj)
```
aymanhab: Can we rename obj to osimC3Dobject for clarity?

Bindings/Java/Matlab/Utilities/osimC3D.m
```diff
@@ -150,6 +150,32 @@ function rotateData(obj,axis,value)
...
+            import org.opensim.modeling.*
+            
+            nRows  = obj.forces.getNumRows();
+            labels = obj.forces.getColumnLabels();
+            
+            for i = 0 : obj.forces.getNumColumns - 1
+                % All force columns will have the 'f' prefix while point
+                % and moment columns will have 'p' and 'm' prefixes,
+                % respectively. 
+                if ~contains(char(labels.get(i)),'f')
```
aymanhab: Maybe unrelated to this PR but is the 'f' prefix an arbitrary convention we made or is it enforced by C3D?

Bindings/Java/Matlab/Utilities/osimC3D.m
jimmyDunne: As far as I am aware, it can only be obj (obj is itself).

Bindings/Java/Matlab/Utilities/osimC3D.m
jimmyDunne: Yep, forces are labelled _fn_, where n is the force number. We use _p_ for COP and _m_ for Moment

Bindings/Java/Matlab/Utilities/osimC3D.m
aymanhab: So on line 40 we can't name the object as c3dObject for example?? I guess then users are accustomed to confusion! My bad

jimmyDunne: I checked out some documentation from a few different sources, and you are correct, I can change this into a custom name. However, all examples I had ever seen used _obj_ , with some of the examples I found today using _self_. So I am hesitant to change it since someone new to OOP in Matlab may nor realize the mapping.

Bindings/Java/Matlab/Utilities/osimC3D.m
```diff
@@ -150,6 +150,32 @@ function rotateData(obj,axis,value)
...
+                    for u = 0 : nRows - 1
+                        % Get the Vec3
+                        vec3 = obj.forces.getDependentColumnAtIndex(i).get(u);
+                        % Divide the Vec3 by 1000
+                        vec3_new = Vec3(vec3.get(0)/1000,vec3.get(1)/1000,vec3.get(2)/1000);
+                        % set the table value
```
aymanhab: This is excessively expensive to divide a Vec3 by a double, I''m looking into exposing the equivalent of operator for +=,-=, *=, /= to Java/Matlab so avoid the inefficiency/overhead but looks fine considering what we have now.

Bindings/Java/Matlab/Utilities/osimC3D.m
```diff
@@ -170,10 +196,6 @@ function writeMOT(obj,varargin)
         % osimC3d.writeMOT()                       Write to dir of input c3d.
         % osimC3d.writeMOT('Walking.mot')          Write to dir of input c3d with defined file name.
         % osimC3d.writeMOT('C:/data/Walking.mot')  Write to defined path input path.
-        %
```
aymanhab: Can we remove these comments unless you use for debugging, I was worried about sneaking absolute paths into user facing code.

Bindings/Java/Matlab/tests/testOsimC3D.m
```diff
@@ -34,76 +34,92 @@
...
+randomRows = randsample(forces.getNumRows(),101)-1;
+for i =  0 : forces_m.getNumColumns() - 1
+    
+    col =  forces_rotated.getDependentColumnAtIndex(i);
+    col_m = forces_m.getDependentColumnAtIndex(i);
+    label = forces_m.getColumnLabel(i);
+    
+    for u = 1 : length(randomRows)
+        if contains(char(forces_m.getColumnLabels().get(i)),'f')
+            % Get the values
```
aymanhab: Should this at least be start with 'f' rather than contains 'f'? Not convinced that making up a convention while reading data is the right thing to do, but seems unrelated.

Bindings/Java/Matlab/Utilities/osimC3D.m
aymanhab: Also can we verify the units instead of assuming them to avoid potentially corrupting data or are the units not built into the format?

aymanhab: I would prefer 'self' which means something special over a generic 'obj' that can easily get lost, but you're call. Thanks for investigating.

Bindings/Java/Matlab/Utilities/osimC3D.m
jimmyDunne: They are actually help statements for the function that a user can call;

``` java
>> help osimC3D/writeMot
  Write force data to mot file.
  osimC3d.writeMOT()                       Write to dir of input c3d.
  osimC3d.writeMOT('Walking.mot')          Write to dir of input c3d with defined file name.
  osimC3d.writeMOT('C:/data/Walking.mot')  Write to defined path input path.
  
  This function assumes point and torque data are in mm and Nmm and
  converts them to m and Nm. If your C3D is already in M and Nm,
  comment out the internal function convertMillimeters2Meters()
```
In this case, you can use either of the three methods (empty, filename, or full path) to write the output.

Bindings/Java/Matlab/tests/testOsimC3D.m
jimmyDunne: I preferred _contains()_  in this case because it will return a 1 or 0. If I use something like _strfind()_ it will return either 1 or [], which means I have to do something like ~isempty(strfind()).  Since we (_C3DFileAdapter()_) are the ones that define the column names, we are guaranteed of not having a case when there may be a rogue _f_ in the column name (like if someone named their column COP_f for some reason).

aymanhab: ok, if there's nothing like startsWith() that returns logical.

Bindings/Java/Matlab/Utilities/osimC3D.m
jimmyDunne: I agree, _self_ feels like it would be a better differentiator than _obj_. I had always seen _obj_ in examples so went with that. I will update to _self_ here since it doesn't break anything (I tested it locally already).

jimmyDunne: > Also can we verify the units instead of assuming them to avoid potentially corrupting data or are the units not built into the format?

Agreed, but there isn't information on units in the Forces table metadata.  There is probably (almost certainly) a way of getting it from _C3DFileAdapter()_ and including it in the metadata.

The reasons for exposing _convertMillimeters2Meters()_ was actually because of this. I had been converting everyones data by default at write time, which is probably fine in most cases (Vicon stores in mm) but it could cause havoc on if not.

Bindings/Java/Matlab/tests/testOsimC3D.m
jimmyDunne: Oh, SNAP! hahaha, I hadn't seen that! Thanks! Perfect I will change to startsWith()

Bindings/Java/Matlab/Utilities/osimC3D.m
aymanhab: PR #2558 should allow you to replace this with `vec3.scalarDivideEq(1000)` in a fraction of the time but don't hold this PR off for it.

jenhicks:
@aymanhab will take a pass through the latest. Need to update here and elsewhere (maybe) to leverage new methods for vector multiplication/division by a scalar.

aymanhab:
Thanks @jimmyDunne I merged this and #2558, should be minor change to upgrade after. I didn't delete the branch in case you want to reuse it.

PR #2558, opened 2019-09-05
Expose methods to operate on vectors through bindings 

aymanhab:
Fixes issue #0

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2558)

### DIFF
--- a/Bindings/Java/tests/TestVectors.java
+++ b/Bindings/Java/tests/TestVectors.java
@@ -1,0 +1,24 @@
+import org.opensim.modeling.*;
+
+class TestVectors {
+    public static void TestVectors() {
+        Vec3 m = new Vec3(1, 2, 3);
+        System.out.println(m.toString());
+        m.scalarEq(5.);
+        System.out.println(m.toString());
+	m.scalarPlusEq(3.);
+	Vec3 n = new Vec3(8);
+        System.out.println(n.toString());
+	n.scalarTimesEq(1000.0);
+	n.scalarDivideEq(1000.0);
+	// Check m & n are same = {8, 8, 8}
+	for (int i=0; i<3; i++)
+	   assert(m.get(i) == n.get(i));
+    }
+  public static void main(String[] args) {
+      TestVectors();
+      
+      System.out.println("Test finished!");
+      // TODO to cause test to fail: System.exit(-1);
+  }
+}
--- a/Bindings/Java/tests/TestVectors.java
+++ b/Bindings/Java/tests/TestVectors.java
@@ -4,16 +4,22 @@ class TestVectors {
     public static void TestVectors() {
         Vec3 m = new Vec3(1, 2, 3);
         System.out.println(m.toString());
+        m.scalarPlusEq(3.);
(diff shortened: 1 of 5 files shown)


aymanhab:
@chrisdembia @tkuchida and @jimmyDunne This was inspired by seeing code that jumps through hoops to divide a Vec3 by a constant in Matlab. Don't know if we should enable the methods for all Vec<x> types and also what other methods would be helpful. Feedback is appreciated. This should reduce the over head significantly.

aymanhab:
This should also work for python @chrisdembia @nickbianco unless you found a better way to expose operators already.

tkuchida:
> This should also work for python @chrisdembia @nickbianco unless you found a better way to expose operators already.

In Python, you can do the following (see [Stack Overflow](https://stackoverflow.com/questions/50599045/python-replacing-a-function-within-a-class-of-a-module)):

```
import opensim as osim

def myVec3Add(self,v):
    newvec = osim.Vec3(self[0]+v[0], self[1]+v[1], self[2]+v[2])
    return newvec

osim.Vec3.__add__ = myVec3Add

a = osim.Vec3(1,2,3)
b = osim.Vec3(4,5,6)
c = a+b
```

chrisdembia:
Nice!

Bindings/Java/tests/TestVectors.java
```diff
@@ -0,0 +1,24 @@
+import org.opensim.modeling.*;
+
+class TestVectors {
+    public static void TestVectors() {
+        Vec3 m = new Vec3(1, 2, 3);
+        System.out.println(m.toString());
+        m.scalarEq(5.);
+        System.out.println(m.toString());
+	m.scalarPlusEq(3.);
```
chrisdembia: Fix spacing?

Bindings/Java/tests/TestVectors.java
```diff
@@ -0,0 +1,24 @@
...
+	Vec3 n = new Vec3(8);
+        System.out.println(n.toString());
+	n.scalarTimesEq(1000.0);
+	n.scalarDivideEq(1000.0);
+	// Check m & n are same = {8, 8, 8}
+	for (int i=0; i<3; i++)
+	   assert(m.get(i) == n.get(i));
```
chrisdembia: This test would pass if both functions do nothing.

aymanhab: This is just place holder to make sure the methods are available. Will turn into real test based on answers to questions above. Thanks for the prompt review @chrisdembia

Bindings/Java/tests/TestVectors.java
aymanhab: Done

aymanhab:
@chrisdembia I addressed your feedback, thanks for the prompt review. More feedback on the issues in my first comment would be much appreciated as I think this is a safe change that can have significant positive impact to include if merged. Thank you

aymanhab:
Thanks @chrisdembia

PR #2572, opened 2019-09-23
Fix Clang warning about implicitly deleted assignment operator.

chrisdembia:
### Brief summary of changes

This PR gets rid of a warning with recent versions of Clang about an implicitly deleted assignment operator (because the base class has `const double` members that cannot be assigned).

### CHANGELOG.md (choose one)

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2572)


chrisdembia:
Thanks @aymanhab . Can I merge?

aymanhab:
Sure, thank you @chrisdembia

PR #2576, opened 2019-09-26
Fix scale factors for Marker geometry so it doesn't show inside-out

aymanhab:
Fixes issue https://github.com/opensim-org/opensim-gui/issues/1107

### Brief summary of changes
GenerateDecorations for Marker was not setting scaleFactors so they default to -1,-1,-1 which makes them show as inside-out. This PR fixes the scale factors to 1.0 so shading is correct.

### Testing I've completed
Loaded model with a marker in GUI, pictures below

### CHANGELOG.md (choose one)

- Will update once agreed to include on 4.1

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2576)

### DIFF
--- a/OpenSim/Simulation/Model/Marker.cpp
+++ b/OpenSim/Simulation/Model/Marker.cpp
@@ -176,6 +176,7 @@ void Marker::generateDecorations(bool fixed, const ModelDisplayHints& hints, con
     appendToThis.push_back(
         SimTK::DecorativeSphere(.01).setBodyId(frame.getMobilizedBodyIndex())
         .setColor(color).setOpacity(1.0)
-        .setTransform(get_location()));
+        .setTransform(get_location())
+        .setScaleFactors(Vec3(1)));
     
 }


aymanhab:
opensim_snapshot_marker41

aymanhab:
@tkuchida an easy  start on the list in https://github.com/opensim-org/opensim-gui/issues/1137

jenhicks:
Please be sure to add a note to the changelog 😄

tkuchida:
Perhaps scale factors should default to `NaN` rather than `-1`?

aymanhab:
Thanks for the quick turnaround @tkuchida I can make the change to default scale factor but that would be more far reaching than the markers. The comment in SimTK::DecorativeGeometry suggests that -ve values of scale factors mean "use default" so I prefer not to interfere with this scheme.
If we go this route I would likely not use NaNs which would lead to unpredictable visuals (as opposed to intentionally blow up computation fast) but I don't think it is necessary here.

aymanhab:
@tkuchida On another note, do the colors/material-properties/shading match what you used for the book (I understand you had different lighting, but that will be addressed separately). Thank you

tkuchida:
> The comment in SimTK::DecorativeGeometry suggests that -ve values of scale factors mean "use default" so I prefer not to interfere with this scheme.

I'm unfamiliar with the implementation details, but if the parent class demands its children to set the scale factors, then I would have expected the default values to be nonsense as opposed to a value that could potentially be interpreted as meaningful (as was the case with `-1`).

> do the colors/material-properties/shading match what you used for the book (I understand you had different lighting, but that will be addressed separately).

No, David drew prettier spheres over the GUI's spheres. A nice upgrade would be if the model markers and experimental markers were the same in all respects except for their color. See https://github.com/opensim-org/opensim-gui/issues/509#issuecomment-389361074.

aymanhab:
opensim_snapshot_marker2
Materials are now verbatim same between model and experimental markers.

aymanhab:
@tkuchida We use DecorativeGeometry subclasses as is from the Simbody library, the use cases in our code are slightly different from simbody because they have 2D objects, text etc. where 3d-scaling doesn't apply,  we can still use their classes as primitives on our side if we set scale properly. Not sure I understand your concern here.

tkuchida:
> Not sure I understand your concern here.

The default value of `-1` was interpreted as meaningful when it shouldn't have been (which resulted in inside-out spheres). It seems like a design flaw that perhaps should be corrected further upstream—but, as I said, I haven't investigated the implementation details so it's your call ☎️

aymanhab:
Thanks @tkuchida for the feedback 🥇  I'd lean to merge it as is rather than change the design to handle "default" scaling which made sense in the simbody visualizer context, but not in OpenSim context where objects always live in 3D space and have no built in scaling. This may change when we include text or annotations and we can tackle it then. I asked @chrisdembia for another review and to chime in.

aymanhab:
@chrisdembia can you review this 1 liner and give us feedback, thank you.

chrisdembia:
Woohoo!

aymanhab:
Thanks @chrisdembia much appreciated. 
Thanks @tkuchida for the quick turnaround I opened issue #2583 to make sure the discussion doesn't get lost.

PR #2376, opened 2018-11-14
Update Tug-of-War Scripts

jimmyDunne:
Fixes issue #2372

### Brief summary of changes
- Changed order of front matter so Help is useful (doesn't just show the license)
- Added inertia to the model
- Made scripts look pretty 

### Testing I've completed
- Ran scripts locally and worked fine. The visualizer doesn't display the model geometry, but that is a separate issue

### Looking for feedback on...
- How does my script look?

### CHANGELOG.md (choose one)
- no need to update because it fixes 🐛

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2376)

PR #2577, opened 2019-09-27
Include addJoint() method to Tug-0f-War example script

jimmyDunne:
Reported in issue #2376 

### Brief summary of changes
Added addJoint method for adding a joint

### Testing I've completed
Locally, script runs, builds model, model is able to simulate. 
 
Any issues?

### CHANGELOG.md (choose one)
- no need to update because is an improvement for clarity.

### DIFF
--- a/Bindings/Java/Matlab/examples/OpenSimCreateTugOfWarModel.m
+++ b/Bindings/Java/Matlab/examples/OpenSimCreateTugOfWarModel.m
@@ -65,8 +65,9 @@ for i=0:2, blockToGround.upd_coordinates(i).setRange(angleRange); end
 for i=3:5, blockToGround.upd_coordinates(i).setRange(positionRange); end
 
 
-% Add the block body to the model
-model.addBody(block)
+% Add the block body and joint to the model
+model.addBody(block);
+model.addJoint(blockToGround);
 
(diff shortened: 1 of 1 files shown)


jimmyDunne:
@clnsmith, can you have a quick look?

jimmyDunne:
Yes, it worked because a freejoint was being made under the covers. I agree with @clnsmith that this is not good for an example and we should include the addJoint() method since that is how we want people to build models.

aymanhab:
Thanks for the clear up @jimmyDunne it's surprising that this was working considering that the coordinates modified earlier in the script would be ignored if a joint was created under the cover, an obvious bug. Can we change the anchors to use a "Brick" object instead of a Mesh so that even if the path for mesh files is not set the user will get some visual feedback.

jimmyDunne:
@ayman, I changed the geometry to use Brick. I had to add the use of offset frames to position some the anchor bricks in the Ground body. The below picture shows the original (left) and the updated (right)

image

aymanhab:
Awesome, thanks very much @jimmyDunne 🥇 I think this is more instructive since it shows the Frames and the built-in Geometry. If you agree, feel free to merge once tests have passed. Thank you 👍

Issue #2559, opened 2019-09-05
Moving opensim-moco into opensim-core


chrisdembia:
The opensim-moco contains a lot of code that belongs better in opensim-core. Here's a list:

- [x] ModelProcessor #2712
- [x] TableProcessor #2713
- [x] analyze() #2718
- [x] visualize() #2717
- [x] DeGrooteFregly2016Muscle #2714
- [x] ModelFactory #2715
- [x] Stopwatch
- [x] Catch2 testing framework #2716
- [x] format()
- [x] SmoothSphereHalfspaceForce
- [x] ActivationCoordinateActuator
- [x] requiring numpy for python bindings.
- [x] Combine Moco's doxygen with opensim-core's.
- [ ] What to do with `build_on_*` scripts?
- [x] Should opensim-models still be a submodule?
- [ ] Merge CHANGELOG and README?
- [x] Update `opensim-cmd` to support running .omoco files.

@nickbianco feel free to add to this list.

Proposal in Google Doc: https://docs.google.com/document/d/10SiVJGqMJxiXtiDdI4Ozpxf5HTKHaixX_ynymYuJwk4/edit?usp=sharing

aymanhab:
Moving DeGroote muscle to opensim-core is high priority so that users don't have to juggle muscle models to use/visualize results from Moco in GUI. What is involved @chrisdembia @nickbianco ?

jenhicks:
@chrisdembia Is this issue up-to-date? I think one key task we talked about was formulating a plan/rough timeline for integration. Let us know how we can help. I chatted with Scott a couple weeks ago and he agreed that the integration sounded like a good plan (in terms of how we spend development resources/time). @aymanhab

chrisdembia:
I've started working on requiring NumPy for the Python bindings.

chrisdembia:
I created a branch in this repo `moco` into which I've merged the `opensim-moco` repository. I roughly followed these steps to perform the merge: https://stackoverflow.com/questions/1425892/how-do-you-merge-two-git-repositories

The merge had the conflicts listed below; I accepted the `opensim-core` version in these conflicts. These conflicts still need to be resolved through subsequent PRs.

warning: Cannot merge binary files: .github/.deploy_myosin_sourceforge_rsa.enc (HEAD vs. moco/master)
dependencies/CMakeLists.txt
cmake/CMakeLists.txt
README.md
CONTRIBUTING.md
CMakeLists.txt
CHANGELOG.md
[...]

PR #2580, opened 2019-09-27
Add function visualize().

chrisdembia:
Fixes part of issue #2559 

### Brief summary of changes

- This PR adds a function for visualizing an existing motion from the API.

### Testing I've completed

- We use this heavily in Moco, including during workshops.

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2580)


marnunez:
Is or will this be included for Python/Matlab wrapping?

chrisdembia:
> Is or will this be included for Python/Matlab wrapping?

@marnunez I haven't tested that yet but that is the goal.

chrisdembia:
@aymanhab I only included the changes for CommonUtilities for now. I'll make a separate PR for `visualize()`. Ready for review.

chrisdembia:
Bah, `put_time` requires gcc > 5.0. I'm not sure what to do. Should we update the Travis machines?

aymanhab:
I see some versions of gcc from 2015 https://gcc.gnu.org/releases.html if it comes as default now on linux, osx then definitely, if not maybe we can find a replacement call that works on earlier versions

chrisdembia:
> maybe we can find a replacement call that works on earlier versions

Thanks @aymanhab. I'm trying that.

chrisdembia:
@aymanhab ready for review.

chrisdembia:
CI has passed. I mainly want to use this in the visualize() function to set the title of the visualizer. Matlab/Python surely have their own convenient ways to create a date/time string.

PR #2592, opened 2019-10-11
Splice TimeSeriesTable

aymanhab:
Fixes issue #0

### Brief summary of changes
Introduce methods that crop a TimeSeriesTable to a specified time range

### Testing I've completed
I updated the test case testTRCFileAdapter to test the new methods and used my development environment to successfully chop IMU data files for use in OpenSense

- Implementation feedback, 
- Crop in-place vs return copy, 
- Potential additional functions to split based on events, or passed-in-lambdas to enable detecting static-poses, turns, gait-cycle etc.

### CHANGELOG.md (choose one)

- no need to update yet because still work in progress

### DIFF
--- a/OpenSim/Common/Test/testTRCFileAdapter.cpp
+++ b/OpenSim/Common/Test/testTRCFileAdapter.cpp
@@ -190,6 +190,17 @@ int main() {
     if (failed)
         return 1;
 
+    TimeSeriesTable_<SimTK::Vec3> table(tmpfile);
+    double cropToTime = 0.1;
+    table.crop(0.0, cropToTime);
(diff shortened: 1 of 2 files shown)


aymanhab:
Thanks for the feedback @chrisdembia ready for another pass.

aymanhab:
The reason I added the eps is that it's already baked into the underlying NearestRowIndex, otherwise we're not self consistent. It feels uneasy comparing doubles that have potential roundoff errors for practical purposes but it's  hypothetical, so we can remove until needed.

chrisdembia:
Okay I support consistency but I do wonder if it's necessary. LGTM.

PR #2595, opened 2019-10-14
Use CoordinateDirection rather than CoordinateAxis 

aymanhab:
To allow for consistent handling in data conversion and calibration

Fixes issue #0

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2595)

### DIFF
--- a/OpenSim/Simulation/OpenSense/OpenSenseUtilities.cpp
+++ b/OpenSim/Simulation/OpenSense/OpenSenseUtilities.cpp
@@ -123,7 +123,7 @@ Model OpenSenseUtilities::calibrateModelFromOrientations(
     const string& modelCalibrationPoseFile,
     const string& calibrationOrientationsFile,
     const std::string& baseImuName,
-    const SimTK::CoordinateAxis& baseHeadingAxis,
+    const SimTK::CoordinateDirection& baseHeadingDirection,
     bool visualizeCalibratedModel)
 {
     Model model(modelCalibrationPoseFile);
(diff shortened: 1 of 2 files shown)


aymanhab:
@jimmyDunne I believe this is what we discussed today, please let me know if it solves the problem. Thank you

jimmyDunne:
Built and tested behavior on Matlab. Updates work. I will update the Matlab example in a separate PR.

jenhicks:
That's great! Does that mean that it is applying the heading correction as expected? Does it help at all with the IK solution for Hannah's data?

jimmyDunne:
I think we will have to also add the same CoordinateDirection() to the InverseKinematicsStudy() since the same axis heading has to be applied to in InverseKineaticsStudy as in the Calibration.

aymanhab:
@jimmyDunne If we transform the data earlier using the heading correction, why does the InverseKinemaicsStudy need to know about it?

jimmyDunne:
I think because IKStudy is not assumed to use the same data as in the calibration. The Calibration step also doesn't return the rotated data, just a 'calbrated' model.

jenhicks:
Looking through the code, I believe @jimmyDunne is right. The IKstudy does its own heading correction ... it looks like by parsing a string with the axis name, which can only be positive. I'm not sure why the data isn't corrected once and then used to calibrate the model and run IK. It certainly seems error prone and replicates code. Are there advantages?
https://github.com/opensim-org/opensim-core/blob/654e20e3f092eaf18aba3ca904cc502e20e7bb68/OpenSim/Simulation/OpenSense/InverseKinematicsStudy.cpp#L135

jimmyDunne:
@aymanhab, I merged some recent updates into this PR so I could use the Rotation(Quaternion) constructor without having to switch branches and rebuilding

aymanhab:
Merging per @jimmyDunne review, while refactoring.

PR #2596, opened 2019-10-15
Add Log.h, LogSink.h.

chrisdembia:
# Brief summary of changes

This PR adds infrastructure for logging with spdlog, including adding a JavaLogSink to log OpenSim messages through Java., and a command-line option for controlling the log level.

### Testing I've completed

Ran ctests.

Naming of classes

### CHANGELOG.md (choose one)

- no need to update because...it's part of a larger feature

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2596)

### DIFF
--- a/Applications/opensim-cmd/opensim-cmd.cpp
+++ b/Applications/opensim-cmd/opensim-cmd.cpp
@@ -113,12 +113,12 @@ int main(int argc, const char** argv) {
             if (!success) return EXIT_FAILURE;
         }
     }
-
+/*
     if (args["--log-level"]) {
         Logger::setLevelString(args["--log-level"].asString());
         Logger::info("Logger level is set to {}.", Logger::getLevelString());
     }
-
+*/
     // Did the user provide a valid command?
     // -------------------------------------
     if (!args["<command>"]) {
--- a/Applications/opensim-cmd/opensim-cmd.cpp
+++ b/Applications/opensim-cmd/opensim-cmd.cpp
@@ -34,15 +34,18 @@
 #include <OpenSim/OpenSim.h>
 #include <OpenSim/version.h>
 
-static const char HELP[] = 
+static const char HELP[] =
 R"(OpenSim: musculoskeletal modeling and simulation.
 
 Usage:
-  opensim-cmd [--library=<path>]... <command> [<args>...]
+  opensim-cmd [--log-level=<level>] [--library=<path>]... <command> [<args>...]
   opensim-cmd -h | --help
   opensim-cmd -V | --version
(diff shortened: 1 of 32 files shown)


aymanhab:
Looks great @chrisdembia 
On the naming, do you think Log should be changed to Logger, it feels like the Log would be the name of the product of this class not the class itself but maybe others could chime in as well.

chrisdembia:
@aymanhab you make a great point. I went with `Log` because it's shorter, and we will have commands like `Logger::info()` appearing everywhere throughout the code. 

What do you think about `info()` being a free function in the OpenSim namespace? That way, we can simply use `info("blah")`? If so, then I will gladly change `Log` to `Logger`.

aymanhab:
@chrisdembia I agree with the free function (or a macro) so we don't have widespread changes once done with this round.

chrisdembia:
Well it seems that the MSVC regex library is struggling to handle the updated documentation strings. I don't have an MSVC environment setup. @aymanhab would you be willing to help debug this?

aymanhab:
Sure, will do.

chrisdembia:
Thank you 🙇

aymanhab:
@chrisdembia will try back in the office since my laptop has VS2019 which fails much earlier due to docopt build errors.

chrisdembia:
Okay sounds good.

aymanhab:
@chrisdembia Building locally with my VS 2017 shows no failures at all! Note that the reported failures on appveyor are in tests (that do not exercise the new code AFAICT, and after compilation/linking is done i.e. they are runtime errors). This is likely msvc version/patch dependent but I'll keep investigating.

chrisdembia:
Ah I think it has to do with the amount of RAM available to the process?

aymanhab:
@chrisdembia I realized that the failures only happen for cases that invoke opensim-cmd from CMake. I undid the change to opensim-cmd and the tests continue to fail but I still think these are related to the changes in opensim-cmd.

chrisdembia:
@aymanhab we could roll back the new command-line flag. As-is, users cannot control the debug level for the command-line tools.

aymanhab:
@chrisdembia I attempted that earlier and somehow it didn't help, maybe I missed one place. Feel free to do/try it.

chrisdembia:
@aymanhab okay I'm trying it now.

aymanhab:
Latest test failures are unrelated so this can be merged to the branch, LGTM.

chrisdembia:
Thanks @aymanhab

PR #2608, opened 2019-10-31
Include angular direction to heading correction calculation. 

jimmyDunne:
## Fixes issue 
Fixes bug in computing heading correction when the subject imu heading requires a negative rotation to align with the model. 

### Brief summary of changes
Added a check that uses the cross-product between the model X-axis and the (user-defined) heading axis of the IMU that determines either a positive or negative rotation of the IMU data. 

### Testing I've completed
I have used two example data sets, locally, to check that the correct sign is computed. However, we should have a more robust testing framework, which will be included in a planned, future, refactoring of the methods. 

- Is this a robust method of determining the direction of rotation?
- Is this efficient enough or safe enough from bugs?
- Is there a better method that I am missing?
- Should I add test now, rather than waiting?

### CHANGELOG.md (choose one)

- no need to update because it fixes a bug in unreleased code.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2608)


aymanhab:
Thanks @jimmyDunne, the test case testOpenSense is failing at the moment, so maybe it exercises a scenario that hasn't been taken into consideration or maybe the standard was incorrect, please take a look when you have a chance. Thank you. 🎃 👻

jimmyDunne:
Will do! Thanks, Ayman.

jimmyDunne:
## What are the latest commits?
The test was initially failing because we were comparing an internal Model with a version of the model that was written to file and read back in. However, we were not actually doing that, we were reading in a model that existed and was not getting re-written each time. I deleted the old model (since the model needs to be written to file each time) and updated the _assert_. 

I also fixed the order of the cross product, which was incorrect. 

## What to do next?
The test passes locally, and checking the output in the OpenSim GUI I get the desired kinematics. I still have a problem with this though since I should be testing rotations in all four quadrants, which this does not do. I think a more robust test would be to take the initially forward-facing data, rotating it about Y so that there are four versions (one in each quadrant), and performing the correction heading on each one. 

@aymanhab is working on a parallel refactoring that would affect the tests so I am inclined to close this PR without merging and rather beef up the tests there. Thoughts?

jimmyDunne:
Great! I will work on generating the data and the logic for a more robust test in Matlab first and then will port it across to CPP. 

With regard to testing a known model— my thought would be to compare the orientation of the OffSet frames that we attach, only. If we are comparing the instantiated models then this test will fail and have to be updated everytime there is a change in the xml format, which seems unnecessary.

PR #2610, opened 2019-11-04
Refactor opensense utils

aymanhab:
Fixes issue #0

### Brief summary of changes
Break the methods used to calibrate a model and run InverseKinematicsStudy into smaller methods with clear responsibilities for troubleshooting and maintenance purposes. A separate class ModelCalibrator is now responsible for the calibration process.

### Testing I've completed
Ran test cases, more done by @jimmyDunne 

- API changes, consistency with other tools in OpenSim.
- Class name (ModelCalibrator)
- Should we have output model file available in XML?

### CHANGELOG.md (choose one)

- no need to update because new functionality since 4.0

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2610)


aymanhab:
As it stands, I removed the heading correction out of IKStudy and I'm still getting the test case to pass, which makes sense since it seems this should be done as part of calibration (analog MarkerPlacement) but not on every IK solution. My plan is to remove the hardcoded IMU-to-OpenSim transform, make it into a Property of IKStudy and to remove the heading info from IKStudy and into a separate ModelCalibrationFromIMU object to be used by the Calibration, or explore removing heading correction altogether and adjusting pelvis rotation instead.

OpenSim/Simulation/OpenSense/OpenSenseUtilities.h
```diff
@@ -56,9 +63,20 @@ namespace OpenSim {
                 const SimTK::Array_<int>& startEnd = { 0, 1 },
```
jimmyDunne: SimTK::Array_<int> translates to SimTKArrayInt (in Matlab) and has no set or upd methods that are scripting accessible, so I would like to suggest;
(1) the input data type change to something useable in scripting (ArrayDouble?) 
(2) the default end time is the full size of the input table

aymanhab:
@jimmyDunne The refactoring is mostly done at this point, would be good to have you take a look at the interface from Matlab and update accordingly or let me know if you have questions or concerns.
I need to make sure the documentation is up to date and to beef up the test case. Thank you

OpenSim/Simulation/OpenSense/ModelCalibrator.h
```diff
@@ -0,0 +1,89 @@
...
+            "Name/path to a .sto file of sensor frame orientations as "
+            "quaternions to be used for calibration.");
+
+public:
+    virtual ~ModelCalibrator();
+    ModelCalibrator();
+    ModelCalibrator(const std::string& setupFile);
+    bool run(bool visualizeResults = false);
+    void setModel(Model& aModel) { _model = &aModel; };
+    Model& getModel() const { return *_model; }
```
jimmyDunne: I had doubts on if this returned the original model or the calibrated version at the end. Not sure if we could do something here to ensure the User that the returned model is the calibrated version. Also, I called .getModel() before run(), to test if it returned anything, and I got segFault that crashed Matlab (WHOOPS ;). 

Perhaps we could change the method name to .getCalibratedModel() to convey that this is indeed the calibrated model. And, if called before run(), an exception occurs. Just a thought.

OpenSim/Simulation/OpenSense/ModelCalibrator.cpp
```diff
@@ -0,0 +1,234 @@
...
+        directionOnIMU = SimTK::CoordinateDirection(SimTK::YAxis, direction);
+    else if (back == 'z')
+        directionOnIMU = SimTK::CoordinateDirection(SimTK::ZAxis, direction);
+    else { // Throw, invalid specification
+    
+    }
+
+    // Compute rotation matrix so that (e.g. "pelvis_imu"+ SimTK::ZAxis) lines up
+    // with model forward (+X)
+    SimTK::Rotation headingRotation =
```
jimmyDunne: Could we return the angularDifference from OpenSenseUtilities::computeHeadingCorrection and make a get() method from the model calibrator?

When debugging the orientations, it is helpful to know what the angular difference is but it is printed to the console above all of the imu frame orientations so you have to scroll up and try and find it. It would be nice if the value could be gettable (.getAngluarDifference() ) or printable (.printAngluarDifferenceToConsol() ).

jimmyDunne:
I am getting great results running this through Matlab with the OpenSense example data. The ModelCalibrator() and the updated IKStudy() worked great. I will go on to test some more of the functionality before we meet, but these changes are looking great so far!

OpenSim/Simulation/OpenSense/ModelCalibrator.cpp
aymanhab: Good ideas @jimmyDunne I was debating whether computeHeadingCorrection would be better buried inside the ModelCalibrator as the only client or keep at OpenSenseUtilities for potential use by other calibration schemes. If we move it inside ModelCalibrator then it can be maintained as a part of the ModelCalibrator object, retrieving the angle and/or the Transform. Any thoughts?

OpenSim/Simulation/OpenSense/OpenSenseUtilities.cpp
```diff
@@ -360,4 +218,59 @@ OpenSenseUtilities::createOrientationsFileFromMarkers(const std::string& markers
...
+        Model& model,
+            OpenSim::TimeSeriesTable_<SimTK::Quaternion_<double>>&
+                    quaternionsTable,
+            const std::string& baseImuName,
+            const SimTK::CoordinateDirection baseHeadingDirection) 
+{
+    SimTK::Rotation rotation;
+ 
+    // if a base imu is specified, perform heading correction, otherwise skip
+    if (!baseImuName.empty()) {
```
jimmyDunne: Since the other functionality has been removed from this function, this is unnecessary. Perhaps now, if you use the function, you should have to give a valid baseImuName or it should throw an exception?

OpenSim/Simulation/OpenSense/ModelCalibrator.cpp
jimmyDunne: From chatting offline about this, it feels like the thing to do would be to keep computeHeadingCorrection() in OpenSenseUtilities so other people could use it if needed. The function could return the angle and the Rotation matrix could be constructed in ModelCalibrator().

aymanhab:
@chrisdembia would love to get your feedback on the interface and if you see it consistent with other tools.

OpenSim/Simulation/OpenSense/ModelCalibrator.h
aymanhab: Done

OpenSim/Simulation/OpenSense/ModelCalibrator.cpp
aymanhab: @jimmyDunne to strike a balance between returning a Rotation matrix vs returning a double with the implicit assumption that it is represents rotation about Y, I modified the method to return  a Vec3 of rotations that the caller can convert to a Rotation. Please let me know what you think.

aymanhab:
Thanks for the quick turnaround @chrisdembia there's probably some context that you need to know to refine the scope of this PR, we can discuss in person today or by email/chat. Thank you

aymanhab:
Per meeting/discussion we agreed to rename ModelCalibrator to IMUPlacer (to draw the parallel to MarkerPlacer and to rename InverseKinemaicsStudy to a new name similar to InverseKinemaicsTool but that includes IMU in it for now with the plan to merge the new Tool with existing InverseKinemaicsTool in the future.

aymanhab:
To continue the refactoring we agreed upon at the dev meeting, do we have suggestions for a new name for InverseKinematicsStudy. Has to include IMU_ prefix and be similar to InverseKinematicsTool, thoughts on IMU_InverseKinematicsTool? @chrisdembia @jimmyDunne @jenhicks

chrisdembia:
I like IMU_InverseKinematicsTool or IMU_KinematicsTool. I’m fine with the underscore but I have a slight preference for no underscore.

aymanhab:
Thanks @chrisdembia Would you then prefer IMUPlacer over IMU_Placer?

chrisdembia:
Yes, but I’m curious what others think.

jimmyDunne:
I would prefer IMUPlacer only because it seems more consistent with what we have done. Was the reason for the underscore because 'IMU' is all capitilized?

aymanhab:
> I would prefer IMUPlacer only because it seems more consistent with what we have done. Was the reason for the underscore because 'IMU' is all capitilized?

Yes, that was the concern of @jenhicks if I remember correctly

jimmyDunne:
# Vote (add your reaction)
👀  for imuPlacer and  imuInverseKinematicsTool 
🚀   for IMUPlacer and  IMUInverseKinematicsTool 
❤️   for ImuPlacer and  ImuInverseKinematicsTool 
👍   for IMU_Placer and  IMU_InverseKinematicsTool

aymanhab:
Ready for review based on in-person discussion/agreement.

aymanhab:
> # Vote (add your reaction)
> 👀 for imuPlacer and imuInverseKinematicsTool
> 🚀 for IMUPlacer and IMUInverseKinematicsTool
> ❤️ for ImuPlacer and ImuInverseKinematicsTool
> 👍 for IMU_Placer and IMU_InverseKinematicsTool

Renaming applied, backward compatibility exercised by test case that uses InverseKinematicsStudy. Do we want to keep it this way even if it never appeared in public releases of OpenSim? @jimmyDunne @jenhicks

jenhicks:
I like IMUPlacer and IMUInverseKinematicsTool.

aymanhab:
Great, renaming done. Should we support old example files with
InverseKinematicsStudy (did it already to reuse in testOpensense0 or
convert to use all new classes? I think any files we distribute should use
the new names as we continue to build these up so less luggage to carry.

jimmyDunne:
If you could include the fix for respecting user-specified times in IMUInverseKinematicsStudy, that would be great. Description of the issue is [here](https://github.com/opensim-org/opensim-core/pull/2621#discussion_r348156572).

aymanhab:
@jimmyDunne the changes for trim were merged using a separate PR, I made the changes to respect time bounds on this branch and removed the function to getTimRangeInUse so that interfaces/users are working with time rather than indices. Appreciate your feedback/review.

jimmyDunne:
I built the branch locally and updated Matlab code to run according to the new class names. Everything seemed to run well, without any exceptions or errors in expected results. Start and end times are respected and for the Tool. 

One thing was that trying to perform repeated calls of IMUInverseKinematicsTool().run() resulted in an exception; 

```
Java exception occurred:
java.lang.RuntimeException: Attempting to update reporter with rows having invalid timestamps. Hint: If running simulation in a loop, use clearTable() to clear table at the end of each loop.

Timestamp at row 41 with value 0.000000 is less-than/equal to timestamp at row 40 with value 1.000000
	Thrown at TimeSeriesTable.h:486 in validateRow().
	Thrown at Reporter.h:224 in implementReport().
	at org.opensim.modeling.opensimSimulationJNI.IMUInverseKinematicsTool_run__SWIG_0(Native Method)
	at org.opensim.modeling.IMUInverseKinematicsTool.run(IMUInverseKinematicsTool.java:341)
```
The exception indicates to run .clearTable() but that method is for the tableReporter and not available for the IMUInverseKinematicsTool. So, to run the tool repeatedly, I have to reinstantiate each time. Not that big of a deal but I wonder if we could clear the table inside of the tool after each run() is called since the tool prints the results to file and you can't get access to the tableReporter through the tool object anyway?

aymanhab:
Thanks @jimmyDunne  The Reporter is added every time you call run so there're probably multiple ones we can clearTable but also need to make sure we reuse the reporter if there is one. I'm working on that now.

aymanhab:
@chrisdembia Did you have a chance to review yet? Also is there a version of findComponent that returns non const ptr (similar to updComponent vs. getComponent)? Thanks.

aymanhab:
Ready for another review as we try to make a build for internal testers by tomorrow, thanks for the quick turnaround and feedback @chrisdembia @jimmyDunne

jimmyDunne:
I built locally and am using the classes in Matlab without issue 👍

aymanhab:
Ready for another pass @chrisdembia Thank you

aymanhab:
Thanks @chrisdembia Feedback has been addressed, appreciate another look when you have a chance.

aymanhab:
Thanks Chris, can you resolve all other trails, discussions so I know which
two documentation issues you are referring to. Thank you

aymanhab:
Thanks @chrisdembia will address.
@jimmyDunne I'll be fixing documentation before merging to address the feedback from @chrisdembia  but you can finalize the code you have for tomorrow using the latest branch. Thank you.

aymanhab:
Thanks @chrisdembia for the great feedback, please let me know if we can merge.

aymanhab:
Thanks @chrisdembia for the thorough review and thanks @jimmyDunne for testing.

aymanhab:
Will delete branch once builds are done and artifacts available in case the branch is needed in the interim

PR #2619, opened 2019-11-15
MSVC: Treat warnings as errors

chrisdembia:
### Brief summary of changes

This PR causes MSVC to treat warnings as errors and checks an additional warning about signed/unsigned integers for consistency with warnings generated by Clang.

I'm making this change to help with a similar change in Moco.

### Testing I've completed

Ran all ctests.

### CHANGELOG.md (choose one)

- no need to update because...related to build system.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2619)


aymanhab (commented):
@chrisdembia Most of these are due to mixing int and size_t, can we use size_t instead of casting whenever possible? Thanks

chrisdembia:
A while ago, I was trying to figure out the best way to handle the different integer types in loops, and my first thought was to use `size_t`. But that just moved the warning elsewhere; we can't avoid the cast because we use `int` throughout Simbody's numeric types, etc. For example, we would need to use `simbodyVec[(int)index)]` in many places.

So, using `int` is consistent with Simbody (e.g., https://github.com/simbody/simbody/blob/c6f465bc7de0ca2adaecdbffac6076a173f5a347/Simbody/src/MultibodySystemRep.h#L393).

I also referred to the C++ Core Guidelines: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es107-dont-use-unsigned-for-subscripts-prefer-gslindex.

These guidelines suggest using a different type from a third-party library `gsl::index`. This is the explanation they give:

> The built-in array uses signed subscripts. The standard-library containers use unsigned subscripts. Thus, no perfect and fully compatible solution is possible (unless and until the standard-library containers change to use signed subscripts someday in the future). Given the known problems with unsigned and signed/unsigned mixtures, better stick to (signed) integers of a sufficient size, which is guaranteed by gsl::index.

They suggest sticking to signed integers. I know that `int` may not be of sufficient size but I'm not sure that's a practical issue for us, considering that Simbody uses `int` everywhere anyway.

.gitignore
```diff
@@ -28,3 +29,5 @@ dependencies/*
 # MAC File System files
 .DS_Store
 
+/CMakeSettings.json
+
```
aymanhab: Are these changes related?

aymanhab (commented):
> Reviewed 15 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2619)* status: all files reviewed, 1 unresolved discussion (waiting on @chrisdembia)

nickbianco:
Closing: stale PR.

PR #2621, opened 2019-11-18
{WIP} Update time range for IKStudy analysis

jimmyDunne:
### Fixes issue 
User input time is not respected by the IK Study. This is most likely caused due to the times being defined directly from the orientationsTable and (seemingly) not from the user-specified times. 

```
...
auto& times = oRefs.getTimes();

s0.updTime() = times[0];
    ikSolver.assemble(s0);
    if (visualizeResults) {
        model.getVisualizer().show(s0);
        model.getVisualizer().getSimbodyVisualizer().setShowSimTime(true);
    }
    for (auto time : times) {
        s0.updTime() = time;
...
```

### Brief summary of changes
I added a line that trim's the orientationTable based on user specified times. This seemed to be the quickest/easiest way for me to get the user-specified times working locally. 

### Testing I've completed
Run locally (through Matlab) on my own files. 

I doubt this is the most robust way to solve this bug, but was the easiest for me to code. Happy to improve :)
### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2621)

### DIFF
--- a/OpenSim/Simulation/OpenSense/InverseKinematicsStudy.cpp
+++ b/OpenSim/Simulation/OpenSense/InverseKinematicsStudy.cpp
@@ -225,8 +225,8 @@ SimTK::Array_<int> InverseKinematicsStudy::getTimeRangeInUse(
     int nt = static_cast<int>(times.size());
     int startIx = 0;
     int endIx = nt-1;
-    int startTime = times[startIx];
-    int endTime   = times[endIx];
+    auto startTime = times[startIx];
+    auto endTime   = times[endIx];
     
     if (get_time_range(0)>get_time_range(1)){
       throw std::invalid_argument( "Start time greater than End time" );
--- a/OpenSim/Simulation/OpenSense/InverseKinematicsStudy.cpp
+++ b/OpenSim/Simulation/OpenSense/InverseKinematicsStudy.cpp
@@ -143,7 +143,10 @@ runInverseKinematicsWithOrientationsFromFile(Model& model,
 
     TimeSeriesTable_<SimTK::Rotation> orientationsData =
         OpenSenseUtilities::convertQuaternionsToRotations(quatTable);
-
+    
+    // Trim orientation data based on User input times 
+    orientationsData.trim(get_time_range(0),get_time_range(1));
+        
     OrientationsReference oRefs(orientationsData);
     MarkersReference mRefs{};
 
--- a/OpenSim/Simulation/OpenSense/InverseKinematicsStudy.cpp
+++ b/OpenSim/Simulation/OpenSense/InverseKinematicsStudy.cpp
@@ -140,12 +140,14 @@ runInverseKinematicsWithOrientationsFromFile(Model& model,
     OpenSenseUtilities::rotateOrientationTable(quatTable, sensorToOpenSim);
 
     auto startEnd = getTimeRangeInUse(quatTable.getIndependentColumn());
-
+    
     TimeSeriesTable_<SimTK::Rotation> orientationsData =
         OpenSenseUtilities::convertQuaternionsToRotations(quatTable);
-    
-    // Trim orientation data based on User input times 
-    orientationsData.trim(get_time_range(0),get_time_range(1));
+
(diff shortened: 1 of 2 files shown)


OpenSim/Simulation/OpenSense/InverseKinematicsStudy.cpp
```diff
@@ -143,7 +143,10 @@ runInverseKinematicsWithOrientationsFromFile(Model& model,
 
     TimeSeriesTable_<SimTK::Rotation> orientationsData =
         OpenSenseUtilities::convertQuaternionsToRotations(quatTable);
-
+    
+    // Trim orientation data based on User input times 
+    orientationsData.trim(get_time_range(0),get_time_range(1));
```
tkuchida: May want to check that `orientationsData` is non-empty after trim. It doesn't look like [`trim()`](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/TimeSeriesTable.h#L440) guarantees this as a post condition. Also, `trim()` calls [`trimToIndices()`](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/TimeSeriesTable.h#L492), which notes that "Side effect may include that headers/metaData may be left stale." Not sure whether any of the code below assumes the metadata is correct.

aymanhab: Thanks @tkuchida I agree that having an empty table is likely due to error so I'd leave this change here as is and modify the trim method to throw on empty table so the check is not repeated elsewhere, I'll make the change to this branch to be part of the PR. Typically meta data does not contain info that depends on number of rows and this is the case here.

aymanhab: Ready for review @tkuchida Build failures are due to refactoring on the destination branch so unrelated to this PR

jimmyDunne:
I am still working on this. There are errors if you don't set a time. Currently working on something so will update in ~hour

jimmyDunne:
I had issues when no user specifed time was set. This is because the default time would be `Infinity`. 

Instead, I updated the function `getTimeRangeInUse` to validate and set the time ranges that would be used in the `table.trim()` method. I (think) I have tested for all conditions (Matlab). If the user doesn't set any time then the start and end times are used. 

This change will break the function `previewExperimentalData` at L99, since the `for` loop is expecting indices and not start and end time values. I figure this is okay since we won't (shouldn't) be using the function anyway (it only deals with markers whose use will be removed from this class). 

@aymanhab, can you give me feedback on anything I can change?

OpenSim/Simulation/OpenSense/InverseKinematicsStudy.cpp
tkuchida: Perhaps add test cases to exercise the failure modes and ensure these methods remain well-behaved in the future? I don't know where the tests are for OpenSense's InverseKinematicsStudy, though, as there's no Test directory under OpenSim/Simulation/OpenSense 🤔

aymanhab: These are lumped into the testOpensense test case, as we refactor we should break this into more modular tests.

aymanhab:
@jimmyDunne This feels like replicating the trim functionality somewhere else. The trim functionality is implemented at the TimeSeriesTable level and should be available for Marker data as well. If that doesn't work now please advise and we'll refine it. I'd prefer that over repeating the functionality inside a 'preview...' method

aymanhab:
@jimmyDunne not sure why you closed this instead of using trim but the branch has the improvement to trim that @tkuchida suggested so I'd like to include it at some point. Thank you

jimmyDunne:
I couldn't get Xcode or Clion working so I am trying to write the CPP code in a basic text editor and so any type of debugging or type fixes are taking an exceedingly long time. I have to move on to some other work and don't have the time right now to address the required feedback.

aymanhab:
No worries @jimmyDunne I'll take care of the changes. Thank you

PR #2427, opened 2019-03-28
Added Smith2016Ligament, started implementing tests.

clnsmith:
Fixes issue #<issue_number>

### Brief summary of changes
Implemented Smith2016Ligament as a new force component. Tests, Doxygen, and SWIG wrapping are still a work in progress. 

### CHANGELOG.md (choose one)

- no need to update because...
- updated...

The Doxygen for this PR can be viewed at http://myosin.sourceforge.net/?C=N;O=D; click the folder whose name is this PR's number.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2427)

PR #2632, opened 2019-12-05
Blankevoort1991Ligament

clnsmith:
This pull request introduces the Blankevoort1991Ligament, a widely used parameterization of ligaments in multibody modeling.

It intends to replace the orginal pull request for the Smith2016Ligament here:
https://github.com/opensim-org/opensim-core/pull/2427

I have refactored and tried to address all of the orginal comments on the Smith2016Ligament. 

Currently this is failing the testComponent test because the added ligament has no path geometry.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2632)


clnsmith:
No seeing I missed a few small comments on the smith2016ligament that are relevant here. Will finish up tomorrow, but feel free to start reviewing.

chrisdembia:
Thanks @clnsmith ! 

I'm going to create a PR into your branch to clean up formatting.

chrisdembia:
@clnsmith I made a few comments; I hope to do a more in-depth review later.

Could you update the CHANGELOG?

I've opened a PR in your fork to touch up formatting. https://github.com/clnsmith/opensim-core/pull/1

clnsmith:
> > _[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2632)_ status: 0 of 8 files reviewed, 7 unresolved discussions (waiting on @chrisdembia and @clnsmith)
> 
> _[OpenSim/Simulation/Model/Blankevoort1991Ligament.cpp, line 121 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2632#-LvMkhrF1NMmNgIpIUjE:-LvMsiFe-7sjzIkh4Zo9:b-73kw6s) ([raw file](https://github.com/opensim-org/opensim-core/blob/ef14adefc2637712bbe41e2b99f573d511c63895/OpenSim/Simulation/Model/Blankevoort1991Ligament.cpp#L121)):_
> 
> Previously, chrisdembia (Christopher Dembia) wrote…
> > @aymanhab , is this standard practice for setting default colors? I want to make sure this doesn't lead to unexpected behavior.
> 
> Yes @chrisdembia That's how it's done for the Ligament class for the default (not in simulation mode). There's a different method computePathColor(state) that allows components to decide what color to show based on the passed in state.

Is there a way to set the parameter being displayed and the limits for color mapping for all ligaments in the model at will interact nicely with the gui? Most of the dynamic ligament parameters are not normalized from 0 to 1 ( ie strain will be 0-5% for walking, maybe 0-20% for injury simulations, force would be 0 to 200 N or so). So it would be nice to let the user set the parameter visualized (to any of the outputs?) and color map limits. I could make this a property of the Blankevoort1991Ligament class, but then it would need to be changed for all ligaments in the model everytime you want to change activities etc.

clnsmith:
> _[OpenSim/Tests/Components/testComponents.cpp, line 112 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2632#-LvM_u2I81gmu5WHLsZf:-LvM_u2I81gmu5WHLsZg:bqbzfre) ([raw file](https://github.com/opensim-org/opensim-core/blob/ef14adefc2637712bbe41e2b99f573d511c63895/OpenSim/Tests/Components/testComponents.cpp#L112)):_
> 
> > Currently this is failing the testComponent test because the added ligament has no path geometry.
> 
> You can add path geometry here.

Here is the failure I am getting. Any suggestions on how to fix? Looks like it maybe in PathGeometry class?
image

clnsmith:
I just pushed a new commit that should address all of the comments.

> OpenSim/Simulation/Model/Blankevoort1991Ligament.h, line 199 at r4 (raw file):
> 
>         SimTK::Stage::Position);
>     OpenSim_DECLARE_OUTPUT(lengthening_rate, double, getLengtheningRate,
>         SimTK::Stage::Velocity);
> You could omit these last two outputs, because they duplicate outputs in the GeometryPath. If you find it important to keep the outputs, it would be good to name this getLengtheningSpeed() for consistency with the function in GeometryPath.

I left the length and lengthening_rate outputs because for the visualization code I have, I loop over the ligament outputs and it is nice to have these parameters included in the list that can be visualized.

I agree it is strange to rename the lengthening_speed output from GeometeryPath to lengthening_rate here. I decided to do it here to make it consistent with the strain_rate output naming. I could change the strain_rate to strain_speed to make everything consistent, but to me it sounds very strange to say strain_speed. 

I'm not especially stubborn about this point though, so I'm open to whatever to most consistent with the code base. Thoughts?

tkuchida:
> I'm fine with leaving the redundant outputs but have a slight preference for lengthening_speed, even if it's inconsistent with strain_rate.

FWIW, `lengthening_speed` sounds natural to me and doesn't create any dissonance when composed with `strain_rate` 🎵 (bonus marks if you spotted all 5 puns)

chrisdembia:
> FWIW, `lengthening_speed` sounds natural to me and doesn't create any dissonance when composed with `strain_rate` 🎵

I agree with @tkuchida .

> (bonus marks if you spotted all 5 puns)

sounds, natural, dissonance, composed, and the emoji?

tkuchida:
@chrisdembia 80% (the emoji doesn't count)

chrisdembia:
@clnsmith I made a commit to fix a merge conflict. You'll have to pull before you can push.

jenhicks:
FWIW is the name of a song? Kermit's favorite version: https://www.youtube.com/watch?v=R6Oq0aB0LQw

carmichaelong:
+1 for `lengthening_speed`. I like consistency with `GeometryPath`.

Also does strain_rate = dissonance count as another or is that one in the same?

tkuchida:
The answer I was looking for was "me", a homophone of "mi", but I'll also accept @jenhicks's amphibious answer :frog:

clnsmith:
Well I don't think I can come up with a punny enough response to rebuttal @tkuchida so I changed to lengthening_speed 😃 

I updated everything to be in terms of strain, which is much clearer than the mix of strain and length I had before. This is also better inline with the orginal Blankevoort 1991 description. 

> OpenSim/Simulation/Test/testForces.cpp, line 2048 at r4 (raw file):
> 
> Previously, chrisdembia (Christopher Dembia) wrote…
> If this is a pully, shouldn't the force be half the weight?

 This is different from the PathSpring test where both ends of the spring are attached to the block. Here one end of the spring is attached to ground, the ligament loops over a pulley and attaches to a block. So the tension in the ligament is equal to the weight of the block. Equivalent to the One Wheel case here: https://www.explainthatstuff.com/pulleys.html

clnsmith:
> This is nitpicky but I prefer to word these message as "Expected ..." with an optional "..., but [instead]...". That makes it slight more clear what the expected behavior is. You can ignore this comment if you'd like.

These comments are helpful for me to know the standard practices. 

This commit should address all comments!

chrisdembia:
The CI builds are failing for unrelated reasons.

clnsmith:
great! Do I need to approve anything? Or now this is waiting on others?

chrisdembia:
You don’t have to do anything. Ayman raised a point about the tests potentially taking more time but I can look into that.

jenhicks:
@clnsmith We are working on finalizing the 4.1 release. Would it be helpful for the workshop to include this PR in the release? To do that, we'd need a model that uses the ligament so that we could test whether it loads in the GUI. Have you tested models with the ligament and other tools (CMC, static, etc.)?

clnsmith:
Hi Jen,

Would be great to get it into the 4.1 release. For the workshop, I am planning to release a plugin as well for the contact components, so it is not a big deal if the ligament isn't in 4.1, as I can include it in this plugin as well. 

Attached is a model for visualization purposes (femur and tibia with 3 MCL ligaments). If you need a full body model, you can use this one (although it includes contact components you may need to remove):

https://github.com/clnsmith/opensim-jam/tree/master/models/lenhart2015

I have tested the ligament in models with the ScaleTool, and InverseKinematicsTool but nothing else. 
[Blankevoort1991Ligament_demo_osim.zip](https://github.com/opensim-org/opensim-core/files/3991354/Blankevoort1991Ligament_demo_osim.zip)

chrisdembia:
@jenhicks if the team agrees, we can use continuous integration to create a GUI for a separate "camsknee" branch, alleviating the need for a plugin, which would be a more difficult experience for workshop attendees.

clnsmith:
@chrisdembia that sounds great. Should a make a separate camsknee PR for this?

Also, I'm color blind, so if you disagree with the default color of the Blankevoort1991Ligament I don't have any issues with changing it!

chrisdembia:
> @chrisdembia that sounds great. Should a make a separate camsknee PR for this?

I'll try to handle this but if I'm delaying, please pester me!

> Also, I'm color blind, so if you disagree with the default color of the Blankevoort1991Ligament I don't have any issues with changing it!

The color looks good, and it certainly looks better than (0, 1, 0).

chrisdembia:
Ayman made a good point about creating a separate "camsknee" GUI. Our review process for the GUI involves substantial manual testing. If we create a separate "fast-track" "camsknee" GUI, then we would not be doing the manual testing on your end, and @clnsmith , you might need to do some of your own manual GUI testing.

Ayman gave the following examples of testing that should be done before putting the GUI in the hands of workshop attendees:

- Editing properties
- How the new ligament/component works with tools (ID, SO, FD, CMC)
- How the new ligament/components works with visualization (path editing, selection, color changes, color by state during playback of simulation).

chrisdembia:
> I'll try to handle this but if I'm delaying, please pester me!

I've started the process of using GitHub Actions to automate the build of opensim-core's `camsknee` branch. Progress can be viewed here: https://github.com/opensim-org/opensim-core/actions

clnsmith:
> Ayman made a good point about creating a separate "camsknee" GUI. Our review process for the GUI involves substantial manual testing. If we create a separate "fast-track" "camsknee" GUI, then we would not be doing the manual testing on your end, and @clnsmith , you might need to do some of your own manual GUI testing.
> 
> Ayman gave the following examples of testing that should be done before putting the GUI in the hands of workshop attendees:
> 
> * Editing properties
> * How the new ligament/component works with tools (ID, SO, FD, CMC)
> * How the new ligament/components works with visualization (path editing, selection, color changes, color by state during playback of simulation).

No problem, I can take care of the manual testing. I'll get some of my students to test it as well.

chrisdembia:
@aymanhab could we move this forward now that 4.1 is released?

aymanhab:
@chrisdembia Once merge conflicts are resolved feel free to merge, assuming your feedback/concerns were addressed by @clnsmith If you think a second reviewer is needed let me know.

chrisdembia:
@clnsmith do you have some time to resolve the merge conflict? If not, I can get to it eventually.

chrisdembia:
I fixed the merge conflict and also fixed an issue that arose when running testForces in debug.

chrisdembia:
I'm merging this. The travis-ci/pr tests fail only because the PR is coming from a fork.

chrisdembia:
Thanks @clnsmith! 

We should come up with a plan for testing this in the GUI.

PR #2634, opened 2019-12-05
Fix bug with getFormattedDateTime().

chrisdembia:
### Brief summary of changes

Fixed a bug with getFormattedDateTime() where we used an incorrect string length, causing bogus characters to appear in the string.

### Testing I've completed

Tested locally in Moco.

### CHANGELOG.md (choose one)

- no need to update because...this function hasn't been released yet

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2634)

### DIFF
--- a/OpenSim/Common/CommonUtilities.cpp
+++ b/OpenSim/Common/CommonUtilities.cpp
@@ -53,7 +53,7 @@ std::string OpenSim::getFormattedDateTime(
     }
 
     std::stringstream ss;
-    ss << std::string(formatted.get(), size);
+    ss << formatted.get();
 
     if (appendMicroseconds) {
         // Get number of microseconds since last second.
--- a/OpenSim/Common/Test/testComponentInterface.cpp
+++ b/OpenSim/Common/Test/testComponentInterface.cpp
@@ -25,6 +25,7 @@
 #include <OpenSim/Common/Reporter.h>
 #include <OpenSim/Common/TableSource.h>
 #include <OpenSim/Common/STOFileAdapter.h>
+#include <OpenSim/Common/CommonUtilities.h>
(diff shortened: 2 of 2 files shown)


aymanhab:
Also do you think we should hold 4.1 Beta for this fix?

chrisdembia:
In opensim-core, I did not test with the first argument as `true`, and the bug only appears when the first argument is `true`. But in Moco, we use `true`.

No I don't think we need to hold the 4.1 beta.

chrisdembia:
> LGTM, Would it have helped if we had a test case that gets time and formats it with appendMicroseconds on/off and checks that one string is a substring of the other?

Good idea. I added a test.

chrisdembia:
Ready for review again.

aymanhab:
Thanks @chrisdembia I'm thinking this test may still fail if the seconds digit flips between the two calls, it would be safer to cache the time and format it in two different ways but I don't think the interface supports that. We can merge and keep an eye on the test to see if this scenario ever happens in practice, or drop the test here, your call.

chrisdembia:
> Thanks @chrisdembia I'm thinking this test may still fail if the seconds digit flips between the two calls, it would be safer to cache the time and format it in two different ways but I don't think the interface supports that. We can merge and keep an eye on the test to see if this scenario ever happens in practice, or drop the test here, your call.

For this reason, I chose for the date in the test to include only the year ("%Y"), so there's only one time a year---particularly inactive for development---when this would be an issue. Let me know if you think this is not sufficient.

aymanhab:
Thanks for the explanation, sorry I didn't check the actual format string closely. Agreed it's unlikely to fail more than once a year 😃

chrisdembia:
Thanks for merging :)

PR #2636, opened 2019-12-05
Build opensense executable always

aymanhab:
Regardless of the value of BUILD_INDIVIDUA…L_APPS CMake variable, build opensense (other executables particularly id conflict with system level commands on non-windows machines)

Fixes issue #0

### Brief summary of changes
Remove conditional from CMakeLists.txt

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2636)

### DIFF
--- a/Applications/opensense/CMakeLists.txt
+++ b/Applications/opensense/CMakeLists.txt
@@ -1,6 +1,4 @@
-if(OPENSIM_BUILD_INDIVIDUAL_APPS)
-    OpenSimAddApplication(NAME opensense)
-endif()
+OpenSimAddApplication(NAME opensense)
 
 if(BUILD_TESTING)
     subdirs(test)


aymanhab:
Quick review @chrisdembia when you have a chance. Thank you.

I think long term we should remove this option altogether and rename id executable to something else (e.g. idtool) instead of the overkill in CMake to make sure we don't install an id application. Thoughts?

chrisdembia:
> I think long term we should remove this option altogether and rename id executable to something else (e.g. idtool) instead of the overkill in CMake to make sure we don't install an id application. Thoughts?

It may not be a surprise, given that I introduced the overkill, that I disagree. I think it's clearer for users if we use a single command-line tool; this also avoids duplicating code. I tested this new executable rigorously (testCommandLineInterface). If we are going to make users change the name of the `id` command in their scripts anyway, we might as well make them use `opensim-cmd` instead of `idtool`. I think the agreed-upon plan was to deprecate the older individual tools (these executables already print a deprecation message); we can remove the overkill when we remove the individual tools.

If we rename `id`, then we should rename all the tools for consistency. The executables should have `opensim` in the name, because UNIX systems place executables from all programs in the same `bin` directory.

This PR LGTM.

aymanhab:
Thanks @chrisdembia 

I'd agree with any solution that doesn't attempt to preserve self-inflicted bad names, opensim-cmd is a perfectly fine solution, less luggage to carry in the build system.

PR #2657, opened 2019-12-24
Add opensim-install-command-line.sh.

chrisdembia:
### Brief summary of changes

This PR adds a shell script on Mac and Linux systems to aid with making OpenSim's command-line tools accessible in the Terminal from any directory.

### Testing I've completed

Locally, I installed opensim-core and ran the script multiple times. I ensured that I get a password prompt and that I get prompts if the symlinks already exist.

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2657)

### DIFF
--- a/Applications/CMakeLists.txt
+++ b/Applications/CMakeLists.txt
@@ -12,3 +12,8 @@ add_subdirectory(opensense)
 add_subdirectory(versionUpdate)
 
 add_subdirectory(opensim-cmd)
+
+if(UNIX)
+    install(PROGRAMS opensim-install-command-line.sh DESTINATION
+        ${CMAKE_INSTALL_BINDIR})
+endif()
--- a/Applications/opensim-install-command-line.sh
+++ b/Applications/opensim-install-command-line.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+# This script makes the OpenSim command-line tools accessible from any
+# directory (assuming /usr/local/bin is on the PATH, by being listed in the file
+# /etc/paths).
+# This script should be run from the directory containing opensim-cmd and
+# opensense.
+
+if [ ! -f "$(pwd)/opensim-cmd" ]; then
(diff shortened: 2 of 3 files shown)


aymanhab:
PR looks good but I don't want us inventing a new mechanism different from what other packages do, especially for rather corner use case, can we explore options first.

chrisdembia:
CMake GUI:

The CMake GUI menu has an item "How To Install for Command Line Use", which brings up this dialog with instructions:

<img width="631" alt="image" src="https://user-images.githubusercontent.com/846001/71546588-72581680-2966-11ea-8fb3-8378a53b96d5.png">

This is similar to our instructions on Confluence, with the flaw that the first instruction does not work for all shells (only for bash, zsh, but not for fish).

Intellij IDEs create shell scripts in `/usr/local/bin`:

<img width="552" alt="Screen Shot 2019-12-28 at 11 34 03 AM" src="https://user-images.githubusercontent.com/846001/71546555-0b3a6200-2966-11ea-950b-a3a4c2ac890d.png">

The contents of one of these shell scripts is the following:

```
#!/bin/sh                                                                                                                                                                                                         
# Generated by JetBrains Toolbox 1.16.6207 at Sat Dec 28 11:15:43 2019                                                                                                                                            
                                                                                                                                                                                                                  
open -a "/Users/chris/Library/Application Support/JetBrains/Toolbox/apps/CLion/ch-1/193.5662.56/CLion.app/Contents/MacOS/clion" "$@" 
```
This script works for macOS, but we would need a different script for Ubuntu/Linux.

Based on the above, I'm not sure there is a standard process. If you want more examples, could you look?

It's true that we have few Mac users, but we in the lab use Macs and so we should consider which workflow we personally want.

chrisdembia:
Thank you!

PR #2658, opened 2019-12-29
initial commit of Smith2018ContactMesh and Smith2018ArticularContactForce

clnsmith:
### Brief summary of changes
This PR introduces a new ContactGeometry and Force component to represent articular contact using triangulated surfaces meshes based on the method introduced in this paper:

Smith, C. R., Won Choi, K., Negrut, D., & Thelen, D. G. (2018). Efficient computation of cartilage contact pressures within dynamic simulations of movement. Computer Methods in Biomechanics and Biomedical Engineering: Imaging & Visualization, 6(5), 491-498.

https://www.tandfonline.com/doi/abs/10.1080/21681163.2016.1172346

### Testing I've completed
Testing is a work in progress. Currently I added a test to testForces that builds a model using these components and performs a simple forward simulation. I will add more testing soon to make sure results are correct. 

I tried to add to testComponents, but without a mesh file (.stl etc) the Smith2018ContactMesh fails.

Overall layout of components etc. This is a pretty good sized chunk of code, so it may take a while to review. 

### CHANGELOG.md (choose one)
- updated to introduce new components

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2658)


chrisdembia:
AppVeyor (Windows testing) generates warnings like the following:

```
C:\projects\opensim-core\OpenSim\Simulation\Model\Smith2018ArticularContactForce.cpp(349): warning C4244: 'argument': conversion from 'SimTK::Real' to 'int', possible loss of data
```

Could you fix these issues? Once the tests pass, we should create a separate PR to merge these components into the `camsknee` branch without review.

In parallel, we can review these changes for being merged into the master branch.

chrisdembia:
The Travis-CI build fails due to the use of tabs instead of spaces.

clnsmith:
I just pushed a commit that should address both of these issues.

Apparently visual studio edit->advanced->untabify selected lines doesn't actually remove all tabs.

chrisdembia:
I quickly skimmed this PR. I think it will take a while to review.

clnsmith:
> I quickly skimmed this PR. I think it will take a while to review.

Yeah agreed. Feel free to just make some high level comments to begin with. I will keep cleaning up the code as I address these. 

The only thing that I think is important to try to finalize before the workshop would be the user interface (ie properties and component types) so the representation in the .osim model file is stable.

chrisdembia (changes requested):
I have some initial disorganized comments.

I have not finished thinking about the user interface.

I made a PR into your fork: https://github.com/clnsmith/opensim-core/pull/3

> Reviewed 4 of 12 files at r1, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-:-LxahjSQ9vlaVgeD2sN6:b2rsubl)* status: 5 of 12 files reviewed, 36 unresolved discussions (waiting on @clnsmith)

___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 40 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxYUWYW7wtrtMkX5YZY:-LxYUWYX-sGHS7CrmUzQ:b-4efwv0) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L40)):*
> ```objc
>  interpenetrate and the local overlap depth (proximity) is calculated for 
>  each triangle. The contact pressure on each triangle face is then calculated 
>  based on the overlap depth (see below).
> ```

Do you think it would make sense to bring over some figures from the paper (if the copyright permits it)?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 40 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxacBeA9wQF7BJq6SsO:-LxacBeA9wQF7BJq6SsP:b-3lavhy) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L40)):*
> ```objc
>  interpenetrate and the local overlap depth (proximity) is calculated for 
>  each triangle. The contact pressure on each triangle face is then calculated 
>  based on the overlap depth (see below).
> ```

I love how detailed this documentation is.

Most users will want to know "should I use this contact model or a different contact model?" and "how do I use this contact model?". I think the documentation could be organized a bit more to answer these two questions before delving into the implementation.

One question I have is what makes this contact model specific to articulations versus foot-ground contact? Might be good to define "articular", too. Can this model be used only for articulations?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 49 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxYUqvk8ylmbGjK5Do_:-LxYUqvk8ylmbGjK5Doa:b-rby0ef) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L49)):*
> ```objc
>  method to efficiently detect contact between triangular meshes using object 
>  Oriented Bounding Boxes (OBB, a common approach in computer graphics) and 
>  several additional speed ups that take advantage of the constrained nature of 
> ```

What are the constraints? Does this mean the surfaces *must* intersect?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 68 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxYjtzW73cAm-_x0h-W:-LxYjtzW73cAm-_x0h-X:b-os4qt5) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L68)):*
> ```objc
>  generally small. Thus, after reposing the meshes, (i.e. realizePosition) each 
>  triangle in the casting mesh is tested against the contacting target triangle 
>  from the previous pose. Additional speed up is gained by casting the normal 
> ```

Is the case of "no previous pose" handled gracefully?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 128 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxaYy-PBbs3WBh0bbAf:-LxaYy-Q8Ae9kXOLw0bc:b-cf7y8j) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L128)):*
> ```objc
>  to accomodate variable properties. The use_lumped_contact_model property 
>  controls whether the constant property or variable property formulation is 
>  used.
> ```

This paragraph is a bit hard for me to follow.
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 137 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxaY4JE1j4rOQzc8moe:-LxaY4JE1j4rOQzc8mof:bahzbcr) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L137)):*
> ```objc
>  \f[
>    P_\mathrm{casting} = f(E,\nu,h,d_\mathrm{casting})\:
>    (linear or non-linear formulation above)
> ```

This is not formatted well when generating the HTML documentation:


___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 167 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Lxa_ceh9wCLnrcHqO1P:-Lxa_ceh9wCLnrcHqO1Q:b-6bjptc) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L167)):*
> ```objc
>  system of equations is solved using a numerical solver. 
> 
>  The min_proximity and max_proximity properties limit the search region for a 
> ```

These are distances in meters along the ray?

How does "proximity" differ from "penetration depth" or "overlap depth"?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 172 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Lxa_S2R66L34aa1I1nY:-Lxa_S2R66L34aa1I1nZ:b-llga8z) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L172)):*
> ```objc
>  from being found. This is particularly important for situations with highly
>  curved meshes where a ray cast from the casting_mesh may intersect the 
>  target_mesh multiple times, or at infeasible contact locations. An example
> ```

Can you explain? This description hints that, if the proximity properties are not set properly, one could get invalid results. That does not seem desirable.
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 176 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Lxa_iqHAhVwSMFCiLnw:-Lxa_iqHAhVwSMFCiLnx:b-w886uo) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L176)):*
> ```objc
>  cartilage mesh can interesect the backsides of the femoral condyles, or the 
>  the sides of the trochealar groove multiple times. The min_proximity can be 
>  set to a negative value if you would like distance maps for the out of 
> ```

What is a distance map?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 182 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Lxa_nWPCrRgP5O_ghJ3:-Lxa_nWPCrRgP5O_ghJ4:b-ba69aa) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L182)):*
> ```objc
>  mapped.
> 
>  This component has some outputs such as tri_proximity, tri_pressure, and 
> ```

Would it be possible to use the full word `triangle` everywhere?

Relatedly, should these be named `mesh_proximity`, etc.?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 191 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxabBKj85ZLuELMPPQo:-LxabBKj85ZLuELMPPQp:b-js4966) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L191)):*
> ```objc
>  summary metrics in a subspace of the mesh. The entries in the 
>  1 x 6 SimTK::Vector correspond to the subset of 
>  triangles whose center is located in the half space [+x, -x, +y, -y, +z, -z].
> ```

I don't completely follow how these subspaces are defined.
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 197 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxaaEUN0NS7Cv3pNwWi:-LxaaEUN0NS7Cv3pNwWj:b-qjdln6) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L197)):*
> ```objc
>  medial-lateral axis, points medially, and the origin is located between the 
>  femoral condyles, then the regional outputs corresponding to +z and -z will
>  enable comparisons of the loading in the medial and lateral compartments. 
> ```

These outputs seem difficult to post-process. What do you think about including a Matlab or Python example for post-processing these outputs to create a 3-D visualization? Would this be a lot of work?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 241 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxaXuXX7nEqg5C70CuX:-LxaXuXY8AvIKM67YoaU:bescmlg) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L241)):*
> ```objc
>     OpenSim_DECLARE_PROPERTY(elastic_foundation_formulation, std::string,
>         "Formulation for depth-pressure relationship: "
>         "'linear' or 'nonlinear'")
> ```

These comments should mention the default value.
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 260 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxaaYDAE9XF_teo5CWz:-LxaaYDAE9XF_teo5CX-:b-recmwb) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L260)):*
> ```objc
>         "Target mesh for collision detection.")
>     OpenSim_DECLARE_SOCKET(casting_mesh, Smith2018ContactMesh,
>         "Ray casting mesh for collision detection.")
> ```

The use of sockets allows a single mesh to be used with multiple forces, which is great. 
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 267 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Lxack4F6bjV5RnnOZbQ:-Lxack4F6bjV5RnnOZbR:b-o152gx) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L267)):*
> ```objc
>     //=========================================================================
>     //number of colliding triangles
>     OpenSim_DECLARE_OUTPUT(target_total_n_contacting_tri, int,
> ```

Consider "num" instead of "n".
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 273 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxacteY8BOHuwgg80Q1:-LxacteY8BOHuwgg80Q2:b-anzks3) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L273)):*
> ```objc
> 
>     //tri proximity
>     OpenSim_DECLARE_OUTPUT(target_tri_proximity, SimTK::Vector,
> ```

What is the length of this Vector (and all other vectors)? Does it change with time/states? How does a user associate an element of this vector with the triangles in the mesh?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 424 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxacgMN7n2zgbyrOYhe:-LxacgMN7n2zgbyrOYhf:b-pvh7w1) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L424)):*
> ```objc
> 
>     //number of contacting triangles
>     int getTargetNContactingTri(const SimTK::State& state) const {
> ```

Consider "Num" instead of "N"
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 581 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxaeUpD417OgXBizBdh:-LxaeUpD417OgXBizBdi:bk6lvec) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L581)):*
> ```objc
>     }
> 
>     SimTK::Vector_<SimTK::Vec3> getTargetRegionalCenterOfProximity(
> ```

In what frame are these locations expressed?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 614 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxaeH7O3H4HuyLC8juy:-LxaeH7O3H4HuyLC8juz:b1fc3k5) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L614)):*
> ```objc
> 
>     //contact force
>     SimTK::Vec3 getTargetContactForce(const SimTK::State& state) const {
> ```

In what frame are the forces expressed? To what point are they applied? 
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 656 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Lxae7iw7Pc4KouAqcpI:-Lxae7iw7Pc4KouAqcpJ:b-4ictdb) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L656)):*
> ```objc
> 
>     OpenSim::Array<double> getRecordValues(const SimTK::State& s) const;
>     OpenSim::Array<std::string> getRecordLabels() const;
> ```

Could you document here what the labels are?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 664 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxaeeqN5-gUQjNEAVb_:-LxaeeqN5-gUQjNEAVba:bssom9r) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L664)):*
> ```objc
>     void computeForce(const SimTK::State& state,
>         SimTK::Vector_<SimTK::SpatialVec>& bodyForces,
>         SimTK::Vector& generalizedForces) const override;
> ```

This function and the one above could be public, I think?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 697 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxJR-JI7ObTacABkagj:-LxJR-JI7ObTacABkagk:b-tovkli) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L697)):*
> ```objc
>     void constructProperties();
> 
>     static void calcNonlinearPressureResid(
> ```

Consider avoiding abbreviations; "Resid" -> "Residual"
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 704 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxJQvnR95pgZLxvdKiT:-LxJQvnR95pgZLxvdKiU:b5i72tw) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L704)):*
> ```objc
>     //Member Variables
>     //=========================================================================
>     struct nonlinearContactParams {
> ```

Names of structs should start with a capital letter.
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 734 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxadI4nBp28O--oftdT:-LxadI4nBp28O--oftdU:b-2k2a61) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L734)):*
> ```objc
> //              Smith2018ArticularContactForce :: CONTACT PARAMETERS
> class OSIMSIMULATION_API Smith2018ArticularContactForce::ContactParameters :
> ```

Would you ever want different contact parameters for the same contact mesh? That is, if I have meshes A, B, and C, with contact force X between A-B and contact force Y between A-C, would I want A in X to have different contact parameters than A in Y?

Perhaps the contact parameters could be part of the Smith2016ContactMesh. It might make sense to use a different class name, in that case. I'm not sure.

We are not constrained to using the same class organization as for the other contact models.
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 574 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxafNfG-MiNxHgkTljW:-LxafNfG-MiNxHgkTljX:boayzal) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L574)):*
> ```cpp
>             int ipvt[1];
>             double qtf[1];
>             double wa1[1], wa2[1], wa3[1], wa4[1];
> ```

It would be nice to create a helper function in this class so all this "noise" is removed from this function.
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 801 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Lxafdnl6sCuQY8OU0-R:-Lxafdnl6sCuQY8OU0-S:b-soyau8) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L801)):*
> ```cpp
>     force(0) = normal(0) * pressure * area;
>     force(1) = normal(1) * pressure * area;
>     force(2) = normal(2) * pressure * area;
> ```

No need to do this element-wise; `force = normal * pressure * area;` should work.
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 814 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Lxag0NJ6Zx6vM5_ozK7:-Lxag0NJ6Zx6vM5_ozK8:bfrcuxe) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L814)):*
> ```cpp
>     force(0) = normal(0) * pressure * area;
>     force(1) = normal(1) * pressure * area;
>     force(2) = normal(2) * pressure * area;
> ```

No need to vectorize.
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 818 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxafwQbEnfTvhPMJTXu:-LxafwQbEnfTvhPMJTXv:b-xmbcc7) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L818)):*
> ```cpp
>     moment(0) = force(2) * center(1) - force(1) * center(2);
>     moment(1) = force(0) * center(2) - force(2) * center(0);
>     moment(2) = force(1) * center(0) - force(0) * center(1);
> ```

Simbody has a function `cross()`, also accessible with the `%` operator.
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 107 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxJRFxg6vn14uZ9dtUT:-LxJRFxhEtHg2J3UkIQv:b-i8g1kn) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L107)):*
> ```objc
>         "Maximum thickness threshold for elastic layer [m] when calculating "
>         "cartilage thickness for each triangle.")
>     OpenSim_DECLARE_PROPERTY(scale_factors,SimTK::Vec3,
> ```

Add a space after the comma.
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 149 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxJRMtUAi8Px373SX2v:-LxJRMtUAi8Px373SX2w:b-wbznxh) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L149)):*
> ```objc
> 
>     int getNumFaces() const {
>         return _mesh.getNumFaces();}
> ```

This placement of the brackets is not conventional in our codebase. One option is to place the closing bracket on a separate line.
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 191 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxJRUVfCbdsUMn5nA_d:-LxJRUVfCbdsUMn5nA_e:b3qmkgj) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L191)):*
> ```objc
>         return _vertex_locations;}
> 
>     const OBBTreeNode& getObbTreeNode() const {
> ```

For consistency, the name of the function should be `OBBTreeNode`.
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 264 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxJRlAi752ZVnaCQjPX:-LxJRlAi752ZVnaCQjPY:b-463w78) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L264)):*
> ```objc
> 
> public:
>     class OBBTreeNode {
> ```

Does this class need to be public?
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 266 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxJR_nLD_LTTDjjJUAF:-LxJR_nLD_LTTDjjJUAG:b-cnl3d2) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L266)):*
> ```objc
>     class OBBTreeNode {
>         public:
>             OBBTreeNode() : _child1(NULL), _child2(NULL), _numTriangles(0) {
> ```

With C++11, we use `nullptr` instead of `NULL`.
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.cpp, line 283 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxagqWVE7r6wxTR35JU:-LxagqWVE7r6wxTR35JV:bagg107) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ContactMesh.cpp#L283)):*
> ```cpp
>         _tri_center(i)(0) = (v1(0) + v2(0) + v3(0)) / 3.0;
>         _tri_center(i)(1) = (v1(1) + v2(1) + v3(1)) / 3.0;
>         _tri_center(i)(2) = (v1(2) + v2(2) + v3(2)) / 3.0;
> ```

No need to do this element-wise.
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.cpp, line 292 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxaguKP6GmZ_H6KQQK5:-LxaguKP6GmZ_H6KQQK6:bablbjr) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ContactMesh.cpp#L292)):*
> ```cpp
> 
>         double mag = sqrt(
>             cross[0] * cross[0] + cross[1] * cross[1] + cross[2] * cross[2]);
> ```

Use `.norm()`.
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.cpp, line 390 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LxJU_6p4YOPa_qCc7tI:-LxJU_6p4YOPa_qCc7tJ:b-oryy5f) ([raw file](https://github.com/opensim-org/opensim-core/blob/b284515759b338dccc6fed71e7859cb8891edcfd/OpenSim/Simulation/Model/Smith2018ContactMesh.cpp#L390)):*
> ```cpp
> }
> 
> void Smith2018ContactMesh::computeVariableCartilageThickness() {
> ```

Is it necessary for this contact model to be specific to cartilage or joint contact? Or can this handle generic contact?

chrisdembia (changes requested):
Hey Colin, I looked through `Smith2018ContactMesh` a little bit. I think these two components might be our most complex components!

The next time we chat synchronously, perhaps we could discuss (a) whether `Smith2018ContactMesh` should derive from `ContactGeometry` and if `Smith2018ContactMesh` should have its own scale factors.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-:-Ly6n1Pb9FJ8TOuqbXJK:b-peia4m)* status: 5 of 12 files reviewed, 57 unresolved discussions (waiting on @clnsmith)

___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 268 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6kU9609nQE7Ko5GxX:-Ly6kU9609nQE7Ko5GxY:b-9rl4p4) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L268)):*
> ```objc
>     // OUTPUTS
>     //=========================================================================
>     //number of colliding triangles
> ```

Please always place a space after `//`.
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 511 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6kqSb8akS1eMXCe53:-Ly6kqSb8akS1eMXCe54:b-5tk156) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L511)):*
> ```cpp
>                 tri_pressure(i) = K * tri_proximity(i) / h;
>                 tri_energy(i) = 0.5 * tri_area(i) * K *
>                     pow(tri_proximity(i), 2) / h;                
> ```

Use `SimTK::square()` instead.
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 537 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6kszY9Nt4nwrTqBlI:-Ly6kszY9Nt4nwrTqBlJ:b-x4yjfs) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L537)):*
> ```cpp
>             double depthC = kT / (kT + kC)*tri_proximity(i);
> 
>             double energyC = 0.5 * tri_area(i) * kC * pow(depthC, 2);
> ```

Use `SimTK::square()`.
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 634 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6kvwN7_DZ8Exn__3c:-Ly6kvwN7_DZ8Exn__3d:btyept9) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L634)):*
> ```cpp
> * @param flag2 A status flag
> * @param ptr Pointer to data structure containing values for h1, h2, k1, k2, dc
> */
> ```

Doxygen comments should go in the header.
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 1015 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6Y5O73pwAO88eCZrA:-Ly6Y5O73pwAO88eCZrB:bodgp0w) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L1015)):*
> ```cpp
>     stats.center_of_proximity(0) = COPrx_x;
>     stats.center_of_proximity(1) = COPrx_y;
>     stats.center_of_proximity(2) = COPrx_z;
> ```

No need to do this elementwise.
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 1018 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6l3sACjmXkJtIoBuD:-Ly6l3sACjmXkJtIoBuE:b-9l9mdz) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L1018)):*
> ```cpp
> 
>     //Center of Pressure
>     SimTK::Vector Num = tri_pressure.elementwiseMultiply(tri_area);
> ```

Only the names of classes/structs should start with an upper-case letter.
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 1036 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6YLSi348J1f1ccWw6:-Ly6YLSi348J1f1ccWw7:bodgp0w) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L1036)):*
> ```cpp
>     stats.center_of_pressure(0) = COPx;
>     stats.center_of_pressure(1) = COPy;
>     stats.center_of_pressure(2) = COPz;
> ```

No need to do this elementwise.
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 1155 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6YQkE4fHyORXLQc7p:-Ly6YQkE4fHyORXLQc7q:b-ck7msz) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L1155)):*
> ```cpp
>     constructProperty_poissons_ratio(0.0);
>     constructProperty_thickness(0.0);
>     constructProperty_use_variable_thickness(true);
> ```

These defaults should be functional rather than 0. If you want to force users to always provide a value, you could use NaN as the default and check in `extendFinalizeFromProperties()` that the user provided a value, but I prefer providing reasonable defaults.
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 34 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6mCc06Ww_jDgvKgd5:-Ly6mCc06Ww_jDgvKgd6:b-idcn3y) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L34)):*
> ```objc
> //                       Smith2018ArticularContact
> ```

In general, this class description could be organized better so that the information relevant to all users come first. For example, most users will likely inherit meshes from others. So details about how to create a mesh could go farther down, and it would be nice to have a subheading such as "Choosing the coarseness of the mesh".
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 41 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6lKJu8UwBoyV97mzG:-Ly6lKJu8UwBoyV97mzH:b-ifzdjr) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L41)):*
> ```objc
> only the contact surface. The mesh does not need to be closed (ie water tight) 
> and a smaller number of triangles in the mesh will lead to faster 
> collision dection performance. The normal vectors of the mesh should be 
> ```

typo "dection" 
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 43 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6lQWFA3ZWs0379I1f:-Ly6lQWFA3ZWs0379I1g:b5nfy2v) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L43)):*
> ```objc
> collision dection performance. The normal vectors of the mesh should be 
> pointing outwards from the articular surface towards the opposing contact 
> mesh. Misdirected triangle normals is a common issue when constructing new 
> ```

What happens if the objects are facing the opposite direction from what's intended? For example, what happens if the tibia is above the femur? This specific scenario is not likely, but it's possible in general.
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 47 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6lXS_8cnBM1NJRx3M:-Ly6lXS_8cnBM1NJRx3N:b-i7u69l) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L47)):*
> ```objc
> the triangle areas and normals, and the derivatives of these outputs with 
> respect to joint coordinates are commonly calculated in integrators and 
> optimizers, the best performance will be achieved with a smooth mesh whos 
> ```

whos -> whose
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 53 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6ltcX7l04gq8h_n9C:-Ly6ltcX7l04gq8h_n9D:b-814hsi) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L53)):*
> ```objc
> number of triangles in the mesh. For details on a convergence study based on 
> triangle area see [1]. Note that the GPU implementation described in the paper 
> are not implemented here.
> ```

are not -> is not
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 58 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6m0uY5ug6HKj_Irs0:-Ly6m0uY5ug6HKj_Irs1:b-6the3e) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L58)):*
> ```objc
> Force component to be in contact with another Smith2018ContactMesh. The 
> collision detection algorithm is described in the Smith2018ArticularContact
> doxygen. The Smith2018ContactMesh stores all geometric mesh data and also
> ```

"doxygen" -> "class description"
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 71 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6m_r5-1jj0bDWSd3B:-Ly6m_r5-1jj0bDWSd3C:b-vpgy8k) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L71)):*
> ```objc
> mesh_back_file mesh. If the calculated thickness is outside the range defined
> by the min_thickness and max_thickness properties then it is set to the 
> respective bound. 
> ```

Consider rewording as "We clip the thickness so it remains within the range defined by min_thickness and max_thickness."
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 81 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6mrIRBLKIAQLTGGUk:-Ly6mrIRBLKIAQLTGGUl:b17biam) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L81)):*
> ```objc
> scale_frame socket for both the femur and tibia Smith2018ContactMeshes would 
> be connected to the femur frame to ensure both meshes are scaled by the femur 
> scale factors.
> ```

I'm not sure this type of scaling should be done as part of the Scale Tool. It would be helpful to understand the use case.

What happens if the model is scaled multiple times? Does the mesh keep increasing?
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 86 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6ioAzBc9CY1sTf0Cs:-Ly6ioAzBc9CY1sTf0Ct:byxmcat) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L86)):*
> ```objc
> 
> 
> class OSIMSIMULATION_API Smith2018ContactMesh : public ContactGeometry {
> ```

Do you use facilities provided by the ContactGeometry class?
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 116 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6jUNF34u5j2jsdgHW:-Ly6jUNF34u5j2jsdgHX:bc6cd56) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L116)):*
> ```objc
>     OpenSim_DECLARE_SOCKET(scale_frame, PhysicalFrame,
>         "When using the ScaleTool, the scale factors from this frame will be "
>         "used to scale the mesh.")
> ```

Could you use the frame to which the ContactGeometry is attached?
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 140 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6jilq8QOLL7ue2GnM:-Ly6jilq8QOLL7ue2GnN:b-b6hpt4) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L140)):*
> ```objc
>     //TODO This function must be overriden for because this component is 
>     //derived from OpenSim::ContactGeometry. Is this the right way to do 
>     // nothing?
> ```

Your implementation here seems to indicate that you shouldn't derive from ContactGeometry.
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 211 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6jCkJ1NX3ZA0vktA0:-Ly6jCkJ1NX3ZA0vktA1:b-89txi7) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L211)):*
> ```objc
> 
>     const PhysicalFrame& getMeshFrame() const {
>         return getComponent<PhysicalOffsetFrame>("mesh_frame");
> ```

How do you ensure this subcomponent exists?
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.cpp, line 139 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Ly6k-hL0jb_gL_JpkLs:-Ly6k-hL0jb_gL_JpkLt:bh48668) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ContactMesh.cpp#L139)):*
> ```cpp
>     mesh_frame->set_orientation(get_orientation());
> 
>     adoptSubcomponent(mesh_frame);
> ```

It would be good to make sure that calling `initSystem()` on the model multiple times doesn't cause multiple subcomponents to be added.

carmichaelong (changes requested):
I just finished up a first pass of the code. Thanks for all the hard work on getting this ready. For such a huge class, it was mostly easy to follow along. Happy to chat about steps moving forward, but I think this your goal of a stable interface by the workshop is a good timeline.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2658)* status: 5 of 12 files reviewed, 75 unresolved discussions (waiting on @clnsmith)

___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 66 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LyQooTc9IKOv0Z1AJwY:-LyQooTc9IKOv0Z1AJwZ:bksn8c8) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L66)):*
> ```objc
> The major speed-up in the algorithm leverages the fact that changes in joint
> coordinates and thus articular contact patterns between time steps are
> generally small. Thus, after reposing the meshes, (i.e. realizePosition()) each
> ```

Is it guaranteed that "time steps are small" or should users only use this in cases when time steps are on? (e.g., is there anything stopping users from using this contact in a forward simulation with very loose tolerances or with fixed time steps?)
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 87 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LyQpKfg9PB7nsQiMWI2:-LyQpKfg9PB7nsQiMWI3:b-xd7r1z) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L87)):*
> ```objc
> casting_mesh and target_mesh are switched, the simulation results will NOT be
> exactly the same. For best performance, the casting_mesh should be set to the
> mesh that contains the smaller number of triangles.
> ```

Comments like this that explain some of the implications for users are great!
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 156 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LyQsSro3zDcKsrcvKcw:-LyQsSro3zDcKsrcvKcx:bxr6t21) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L156)):*
> ```objc
> force equilibrium, assuming that the force applied to a pair of contacting
> triangles is equal and opposite. This formulation further assumes that the
> triangles in contact have the same area. The fourth equation states that the
> ```

Does this mean that when using this class, one should make sure that triangles in contact have the same area always? How sensitive is the formulation when this assumption is not followed?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 236 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LyR-ukn4ticlc3viD0I:-LyR-ukn4ticlc3viD0J:b-msc5jz) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L236)):*
> ```objc
>     // PROPERTIES
>     //=========================================================================
>     OpenSim_DECLARE_PROPERTY(min_proximity, double, "The minimum overlap "
> ```

What happens if a user sets min_proximity > max_proximity?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 405 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LyQxUs7-0vDhG1oUlNv:-LyQxUs7-0vDhG1oUlNw:bk8q6tt) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L405)):*
> ```objc
>     Smith2018ArticularContactForce();
> 
>     Smith2018ArticularContactForce(const std::string& name,
> ```

When should a user use this constructor or another constructor? e.g., does this particular one use the parameters of the mesh passed in while the one below overrides the mesh with the passed in contact parameters?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.h, line 745 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LyR-TFw6q7n7UCF_3ko:-LyR-TFw6q7n7UCF_3kp:b5z4pjg) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.h#L745)):*
> ```objc
>     // PROPERTIES
>     //=========================================================================
>         OpenSim_DECLARE_PROPERTY(elastic_modulus, double, 
> ```

What are the default values?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 135 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LyR1JAi2IVt19Uc-3rW:-LyR1JAi2IVt19Uc-3rX:b1iit3l) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L135)):*
> ```cpp
>     Super::extendAddToSystem(system);
> 
>     Smith2018ArticularContactForce::ContactParameters target_mesh_params = 
> ```

Are these two parameters used in this function?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 221 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LyR324b3EaOEZOX86q9:-LyR324b3EaOEZOX86qA:bmv26m) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L221)):*
> ```cpp
>         casting_mesh_def_vec3, Stage::Dynamics);
> 
>     addCacheVariable<double>("target.contact_area",0, Stage::Dynamics);
> ```

Add a space between "," and "0"
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 433 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LyRcMZ01oRtU45go2ro:-LyRcMZ01oRtU45go2rp:b-5zghaq) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L433)):*
> ```cpp
>     else {
>         cache_mesh_name = "target";
>         tContactParams = get_casting_mesh_contact_params();
> ```

The variable names look flipped here, but I'm assuming this is to account for some flip in which is the cached mesh? Maybe some comments here could help for maintainability?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 515 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LySijeW53ueZlpgJ8ZV:-LySijeW53ueZlpgJ8ZW:b48n7hl) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L515)):*
> ```cpp
>             }
> 
>             if (get_elastic_foundation_formulation() == "nonlinear") {
> ```

Is there a specific reason to use continue's here rather than an if/else if statement? What happens if get_elastic_foundation_formulation returns something other than "linear" or "nonlinear"?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 561 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LyRd3DB1AY0aNYrVvm-:-LyRd3DB1AY0aNYrVvm0:b3pf5ib) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L561)):*
> ```cpp
> 
>             //solution params
>             double ftol = 1e-4, xtol = 1e-4, gtol = 0.0;
> ```

Should a user be able to adjust these parameters for their own model?
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 713 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LyRfWjz3ZcIpHhYRKIJ:-LyRfWjz3ZcIpHhYRKIK:b-7ix0j2) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L713)):*
> ```cpp
> 
>     auto stats = computeContactStats(casting_mesh, casting_tri_proximity,
>         casting_tri_pressure,casting_faces);
> ```

space after comma
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 854 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LyRj86U-LgeZmrm17c4:-LyRj86U-LgeZmrm17c5:b9wa443) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L854)):*
> ```cpp
>     for (int i = 0; i < 6; ++i) {
>         auto stats = computeContactStats(casting_mesh, casting_tri_proximity,
>         casting_tri_pressure, casting_region_tri_ind[i]);
> ```

extra indentation to improve readability
___
*[OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp, line 1062 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LySKB885pxS2bzOiNvo:-LySKB885pxS2bzOiNvp:bf7fs63) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ArticularContactForce.cpp#L1062)):*
> ```cpp
>     OpenSim::Array<std::string> labels("");
> 
>     labels.append(getName() + ".casting.contact_area");    
> ```

Would be great to add this list in the comments/doxygen of the .h file
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 55 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LyjmHVE2Vo8zfPhR-Cy:-LyjmHVE2Vo8zfPhR-Cz:b-cp97zi) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L55)):*
> ```objc
> are not implemented here.
> 
> This component can only be used with the Smith2018ArticularContact 
> ```

How is this enforced? If a user gives an ill-posed model (e.g., expecting this to contact another type of Force) what feedback will a user get?
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.h, line 125 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-Lyjnf9gDnvjUMYPq3lf:-Lyjnf9hEfE1Z7XarBAh:b-eomhbu) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ContactMesh.h#L125)):*
> ```objc
>     Smith2018ContactMesh();
> 
>     Smith2018ContactMesh(const std::string& name,
> ```

Like in the other class, which constructors should I use for what cases? What assumptions/defaults are used when using one constructor over the other?
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.cpp, line 229 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LyjoYi7EZ6ji8S25Esx:-LyjoYi7EZ6ji8S25Esy:b-htz4m7) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ContactMesh.cpp#L229)):*
> ```cpp
>     _mesh_is_cached = true;
> 
>     // Initialize Reused Variables
> ```

Looks like this section is no longer needed?
___
*[OpenSim/Simulation/Model/Smith2018ContactMesh.cpp, line 785 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2658#-LyjpbTM2nyg7Y3JqPmo:-LyjpbTM2nyg7Y3JqPmp:b-ywvsbo) ([raw file](https://github.com/opensim-org/opensim-core/blob/446e6a4ea2176b1052abe3eefbd42aee1e98d41b/OpenSim/Simulation/Model/Smith2018ContactMesh.cpp#L785)):*
> ```cpp
>     //this should be adjusted for precision, a=0 when e1 and h are pependicular
> 
>     if (a > -0.00000001 && a < 0.00000001) {
> ```

Could make this a property with a default value for now to avoid hard-coding a constant? Would still keep the door open for some automated precision down the road with another optional boolean? (not sure what's best here to avoid the magic numbers)

nickbianco:
Closing: stale PR.

PR #2672, opened 2020-02-12
Fix Python example code

tkuchida:
Reported on [the Forum](https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=11502).

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2672)

### DIFF
--- a/OpenSim/Common/C3DFileAdapter.h
+++ b/OpenSim/Common/C3DFileAdapter.h
@@ -73,9 +73,8 @@ public:
         <b>Python example</b>
         \code{.py}
         import opensim
-        C3DFileAdapter c3dFileAdapter;
-        c3dFileAdapter.setLocationForForceExpression(opensim.C3DFileAdapter.ForceLocation_CenterOfPressure);
-        tables = c3dFileAdapter.read("myData.c3d")
+        c3dFileAdapter = opensim.C3DFileAdapter()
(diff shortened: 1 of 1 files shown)


OpenSim/Common/C3DFileAdapter.h
```diff
@@ -73,9 +73,8 @@ class OSIMCOMMON_API C3DFileAdapter : public FileAdapter {
         <b>Python example</b>
         \code{.py}
         import opensim
-        C3DFileAdapter c3dFileAdapter;
-        c3dFileAdapter.setLocationForForceExpression(opensim.C3DFileAdapter.ForceLocation_CenterOfPressure);
-        tables = c3dFileAdapter.read("myData.c3d")
+        c3dFileAdapter = opensim.C3DFileAdapter()
+        tables = c3dFileAdapter.read('myData.c3d', opensim.C3DFileAdapter.ForceLocation_CenterOfPressure)
         \endcode
```
aymanhab: @tkuchida Thanks for submitting this. It maybe true that the opensim. qualifier is needed in python 3.7, but the signature of the read method has changed and it doesn't take two arguments as proposed, instead it uses the generic read as was done correctly in the old code. Can you check if the proposed code segment works with OpenSim 4.1? Thanks

tkuchida: Sorry, I didn't realize this code is for Python 3.7.

Issue #2674, opened 2020-02-13
TRCFileAdapter is very slow reading large files


aymanhab:
This was traced to using appendRow on every line of the data file. A similar fix was done for other readers and need to be ported to TRCFileAdapter

aymanhab:
Working on a fix to help @malborno and Johanna

jenhicks:
@aymanhab Is this issue Done?

aymanhab:
Yes, and can be closed.

PR #2675, opened 2020-02-14
Block resize trcfileadapter

aymanhab:
Fixes issue #2674

### Brief summary of changes
Port fix from other adapters/readers to pre-allocate and resize in blocks rather than use appendRow

### Testing I've completed
Ran test suite

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2675)

### DIFF
--- a/OpenSim/Common/TRCFileAdapter.cpp
+++ b/OpenSim/Common/TRCFileAdapter.cpp
@@ -167,7 +167,8 @@ TRCFileAdapter::extendRead(const std::string& fileName) const {
     }
     
     const size_t expected{ column_labels.size() * 3 + 2 };
-    // Will read into in memory matrix, then create table from matrix
+    // Will first store data in a SimTK::Matrix to avoid expensive calls 
+    // to the table's appendRow() which reallocates and copies the whole table.
     int rowNumber = 0;
     int last_size = 1024; 
     SimTK::Matrix_<SimTK::Vec3> markerData{last_size, static_cast<int>(num_markers_expected)};
@@ -202,7 +203,7 @@ TRCFileAdapter::extendRead(const std::string& fileName) const {
         // Column 1 is time.
         times[rowNumber] = std::stod(row.at(1));
         rowNumber++;
-        if (std::remainder(rowNumber, last_size) == 0) {
+        if (rowNumber== last_size) {
             // resize all Data/Matrices, double the size  while keeping data
             int newSize = last_size * 2;
             times.resize(newSize);
@@ -224,11 +225,7 @@ TRCFileAdapter::extendRead(const std::string& fileName) const {
     auto table = std::make_shared<TimeSeriesTableVec3>(
             times, markerData, labels);
     table->updTableMetaData() = metaData;
-    /*
-    TimeSeriesTableVec3::DependentsMetaData dep_metadata{};
-    dep_metadata.setValueArrayForKey("labels", value_array);
-    table->setDependentsMetaData(dep_metadata);
-    */
+
     OutputTables output_tables{};
     output_tables.emplace(_markers, table);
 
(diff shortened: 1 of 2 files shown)


aymanhab:
With the fix, parsing is in seconds instead of minutes but the file still takes 3 hours to solve IK!

OpenSim/Tools/InverseKinematicsTool.cpp
```diff
@@ -362,7 +362,8 @@ bool InverseKinematicsTool::run()
         for (int i = start_ix; i <= final_ix; ++i) {
             s.updTime() = times[i];
             ikSolver.track(s);
-            
+            if (std::remainder(i - start_ix, 1000) == 0 && i != start_ix) // show progress line every 1000 frames
```
chrisdembia: This line is longer than 80 characters.

OpenSim/Tools/InverseKinematicsTool.cpp
```diff
@@ -362,7 +362,8 @@ bool InverseKinematicsTool::run()
...
+                cout << "Solved " << i - start_ix << " frames.." << std::endl;
```
chrisdembia: Please use either 1 periods or 3 periods.

OpenSim/Common/TRCFileAdapter.cpp
```diff
@@ -194,22 +198,37 @@ TRCFileAdapter::extendRead(const std::string& fileName) const {
...
     for(const auto& cl : column_labels)
-        value_array.upd().push_back(SimTK::Value<std::string>{cl});
+            labels.push_back(SimTK::Value<std::string>{cl});
+    auto table = std::make_shared<TimeSeriesTableVec3>(
+            times, markerData, labels);
+    table->updTableMetaData() = metaData;
+    /*
     TimeSeriesTableVec3::DependentsMetaData dep_metadata{};
     dep_metadata.setValueArrayForKey("labels", value_array);
     table->setDependentsMetaData(dep_metadata);
```
chrisdembia: Why is this commented out now?

chrisdembia: Seems to me that this could be deleted.

OpenSim/Common/TRCFileAdapter.cpp
```diff
@@ -169,9 +167,15 @@ TRCFileAdapter::extendRead(const std::string& fileName) const {
     }
     
     const size_t expected{ column_labels.size() * 3 + 2 };
+    // Will read into in memory matrix, then create table from matrix
```
chrisdembia: "Will read into in memory matrix" was a little confusing to me. Consider "Will first store data in a SimTK::Matrix to avoid expensive calls to the table's appendRow()."

OpenSim/Common/TRCFileAdapter.cpp
```diff
@@ -194,22 +198,37 @@ TRCFileAdapter::extendRead(const std::string& fileName) const {
...
             ++ind;
         }
-
+        markerData[rowNumber] = row_vector;
         // Column 1 is time.
-        table->appendRow(std::stod(row.at(1)), std::move(row_vector));
-
+        times[rowNumber] = std::stod(row.at(1));
+        rowNumber++;
+        if (std::remainder(rowNumber, last_size) == 0) {
```
chrisdembia: Could this use `rowNumber == last_size` instead? The remainder would be 0 if `rowNumber` is 50 and `last_size` is 10.

aymanhab: By construction rowNumber will always be less than last_size but I can change the test to == if it makes it more clear.

aymanhab:
All changes made @chrisdembia can we move this since it's holding the imu group? Thank you

OpenSim/Common/TRCFileAdapter.cpp
chrisdembia: Yes, I think it is more clear. By using `remainder`, a maintainer might think that `rowNumber` could go greater.

aymanhab:
Thanks @chrisdembia much appreciated.

Issue #2686, opened 2020-03-08
Thelen2003Muscle equilibrium solve failure


aseth1:
@chrisdembia reported during a developer meeting that Scott Uhlrich encountered this failure only for specific  model and state values affecting a single muscle. The failure can be reproduced but it is not clear why the algorithm fails to find an equilibrium since it does not appear to be a singularity in fiber or tendon force production. The  Newton solve fails to find appropriate step size and fails due to exceeding the iteration limit. A test case is being added.

jenhicks:
@aseth1 @mjhmilla OK to close?

mjhmilla:
Actually I think this one should remain open. Here's why:

Problem: Several muscle models have the possibility of failing to initialize correctly.

Solution Steps:
1. [Done] Build a general purpose initialization test that can catch all of the various ways a model can fail to initialize. This has been done and the code is currently sitting in the branch feature_equilibrium.
2. [Open] Fix the initialization methods for the models that have initialization problems. 

I have a straight forward fix (remove all Newton methods and replace with bisection + initialize the model so that the tendon is static). I simply haven't gotten to this yet. If there is a date when you'd like to see part 2 completed please let me know.

jenhicks:
Closing as the (important) remaining issues are covered elsewhere.

Issue #1736, opened 2017-06-01
Unify the equilibrium solver for Thelen2003 and Millard2012Equilibrium Muscles


aseth1:
Both `Millard2012EquilibriumMuscle` and `Thelen2003Muscle` implement very similar muscle force equilibrium solvers using Newton's method. The differences were further diminished by PR #1728, which patched both implementations to address #1700.

The challenge is mainly one of defining an interface to resolve the different naming conventions since the two solvers use different local methods to evaluate their respective muscle curves and to compute the partial derivatives of muscle and tendon force.

tkuchida:
Related: #366

PR #2687, opened 2020-03-08
Muscle equilibrium bug thelen2003

aseth1:
Test case for issue #2686

### Brief summary of changes
A test case was added that is a stripped down (one muscle) model and conditions that reproduce the muscle equilibrium solve failure encountered in #2686.
 
### Testing I've completed
This test case fails as the activation approaches the activation value that caused the original failure.

@mjhmilla please use this test case to investigate and verify that #2684 is the correct solution

### CHANGELOG.md (choose one)
- no need to update because this is a test case for existing functionality

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2687)

### DIFF
--- a/OpenSim/Actuators/Test/testMuscles.cpp
+++ b/OpenSim/Actuators/Test/testMuscles.cpp
@@ -100,6 +100,7 @@ void testSchutte1993Muscle();
 void testDelp1990Muscle();
 
 void testMuscleEquilibriumSolve(const Model& model, const Storage& statesStore);
+void testThelenMuscleEquilibriumInGaitModel();
 
 int main()
 {
@@ -121,6 +122,13 @@ int main()
     catch (const Exception& e)
         { e.print(cerr); failures.push_back("testDelp1990Muscle"); }
 */
+
+	try { testThelenMuscleEquilibriumInGaitModel();
+        cout << "testThelenMuscleEquilibriumInGaitModel  passed" << endl; }
+	catch (const Exception& e) {
+        e.print(cout);
+        failures.push_back("testThelenMuscleEquilibriumInGaitModel"); }
+
     try { testRigidTendonMuscle();
         cout << "RigidTendonMuscle Test passed" << endl; }
     catch (const Exception& e)
@@ -1113,5 +1121,64 @@ void testMuscleEquilibriumSolve(const Model& model, const Storage& statesStore)
                 muscle.getConcreteClassName() + 
                 " failed to solve for muscle (fiber) and tendon equilibrium. ");
         }
+    }
+}
+
+/** Test case for failure in the equilibrium solve in Thelen2003Muscle.
+    Model and conditions provided by Scott Uhlrich when initializing the
+    Lerner et al. knee model during gait. The model has been stripped down to
+    the single muscle (tib_post_r) where the failure occurs when activation
+    approaches ~0.8 in specific configuration (angles and velocities). The
+    failure does not appear to be due to any approaching singularity in afl or
+    fv multipliers.*/
+void testThelenMuscleEquilibriumInGaitModel()
+{
+    Model model("lerner_one_muscle.osim");
+    const CoordinateSet& coords = model.getCoordinateSet();
+    const Muscle& muscle = model.getMuscles().get("tib_post_r");
+
+    SimTK::State& state = model.initSystem();
+
+    coords.get("subtalar_angle_r").setValue(state, 0.2581);
+    coords.get("ankle_angle_r").setValue(state, 0.2207);
+    coords.get("subtalar_angle_r").setSpeedValue(state, -1.0344);
+    coords.get("ankle_angle_r").setSpeedValue(state, -2.3226);
+
+	double fm = SimTK::NaN, ft = SimTK::NaN;
+
+    for (int i = 1; i < 10; ++i) {
+        cout << i << endl;
+        muscle.setActivation(state, 0.01 * i); // 0.0798916);
+        model.realizeVelocity(state);
+
+        cout << "Input: act=" << muscle.getActivation(state)
+             << " fiberlen=" << muscle.getFiberLength(state)
+             << " afl =" << muscle.getActiveForceLengthMultiplier(state)
+             << " pfl =" << muscle.getPassiveForceMultiplier(state)
+             << " fibervel=" << muscle.getFiberVelocity(state)
+             << " fv =" << muscle.getForceVelocityMultiplier(state)
+             << " fm =" << muscle.getFiberForceAlongTendon(state)
+             << " ft =" << muscle.getTendonForce(state) << endl;
+
+        model.equilibrateMuscles(state);
+
+        fm = muscle.getFiberForceAlongTendon(state);
+        ft = muscle.getTendonForce(state);
+
+        cout << "After: act=" << muscle.getActivation(state)
+             << " fiberlen=" << muscle.getFiberLength(state)
+             << " afl =" << muscle.getActiveForceLengthMultiplier(state)
+             << " pfl =" << muscle.getPassiveForceMultiplier(state)
+             << " fibervel=" << muscle.getFiberVelocity(state)
+             << " fv =" << muscle.getForceVelocityMultiplier(state)
+             << " fm =" << fm
+             << " ft =" << ft << endl;
+
+		// equilibrium demands tendon and muscle fiber forces are equivalent
+        ASSERT_EQUAL<double>(ft, fm, SimTK::SqrtEps, __FILE__, __LINE__,
+                "testThelenMuscleEquilibriumInGaitModel(): " +
+                        muscle.getConcreteClassName() +
+                        " failed to solve for muscle (fiber) and tendon "
+                        "equilibrium. ");
     }
 }
--- a/OpenSim/Actuators/Test/testMuscles.cpp
+++ b/OpenSim/Actuators/Test/testMuscles.cpp
@@ -123,12 +123,6 @@ int main()
         { e.print(cerr); failures.push_back("testDelp1990Muscle"); }
 */
 
-	try { testThelenMuscleEquilibriumInGaitModel();
-        cout << "testThelenMuscleEquilibriumInGaitModel  passed" << endl; }
-	catch (const Exception& e) {
-        e.print(cout);
-        failures.push_back("testThelenMuscleEquilibriumInGaitModel"); }
-
     try { testRigidTendonMuscle();
         cout << "RigidTendonMuscle Test passed" << endl; }
     catch (const Exception& e)
@@ -138,6 +132,12 @@ int main()
         cout << "Thelen2003Muscle Test passed" << endl; }
     catch (const Exception& e)
         { e.print(cout); failures.push_back("testThelen2003Muscle"); }
+
+    try { testThelenMuscleEquilibriumInGaitModel();
+        cout << "testThelenMuscleEquilibriumInGaitModel  passed" << endl; }
+    catch (const Exception& e) {
+        e.print(cout);
+        failures.push_back("testThelenMuscleEquilibriumInGaitModel"); }
 
     try { testMillard2012EquilibriumMuscle();
         cout << "Millard2012EquilibriumMuscle Test passed" << endl; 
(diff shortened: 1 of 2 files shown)


OpenSim/Actuators/Test/testMuscles.cpp
```diff
@@ -1115,3 +1123,62 @@ void testMuscleEquilibriumSolve(const Model& model, const Storage& statesStore)
         }
     }
 }
+
+/** Test case for failure in the equilibrium solve in Thelen2003Muscle.
+    Model and conditions provided by Scott Uhlrich when initializing the
+    Lerner et al. knee model during gait. The model has been stripped down to
+    the single muscle (tib_post_r) where the failure occurs when activation
+    approaches ~0.8 in specific configuration (angles and velocities). The
```
chrisdembia: Approaches 0.08

chrisdembia (approved):

OpenSim/Actuators/Test/testMuscles.cpp
aseth1: Thank you for correcting the typo.

aseth1:
Thank you @chrisdembia for the review and approval.

mjhmilla:
@aseth1 Thank you very much for the test case! Here is the problem: the curves that are used in the Thelen2003Muscle are not even C1 continuous, and a Newton method is used to initialize the model (and it requires C2 continuous curves). The test case you've given me (that blows up) happens to have its solution near lceN = 1: for lceN < 1 the passive-force-length curve has a gradient of zero, for lceN >= 1 the passive-force-length-curve has a non-zero gradient. So the Newton method just jumps from branch-to-branch near lceN=1 and never converges. There might be another curve that blows up - I'd have to take another careful look at the equations again.

There are a number of ways of dealing with this and I've proposed a few below in order of increasing complexity. In either case a bunch of code is going to change in initMuscleState. Please @chrisdembia read this and give me your thoughts. I am not going to code a single line until we're both happy with the plan. 
 
If you'd prefer to skype I'd be happy to do so! Just give me a date/time/skype id (over email).

----------------------------------------
1. Add a bisection method to get close to the solution. Try a standard Newton method for 5-10 iterations: if the solution is better take it, otherwise return the result from the bisection method (this is what Sherm does).

Pro  : easily maintained, each iteration is pretty fast, and its stable. I have a working prototype.
Con : many iterations are required to get to a tolerance of 1e-8 N. This is however a ridiculous tolerance. We can, as an example get to a tolerance of  0.015 N with 25 bisection iterations (I've got a prototype version) on a muscle with a maximum-isometric-force of 1500 N.

2. Add a bisection method to get close to the solution, use a branch-aware Newton method to polish it.

Pro : should be able to maintain the solution tolerance that we currently enjoy, along with the speed.
Con: This should work fine but there could be surprises due to the fiber-tendon velocity distribution heuristic that I have to use. And this will take time. And be harder to maintain.

----------------------------------------

In either case I recommend 

1. Removing the line-searching feature of the Newton method that is currently in there: it is not necessary with the good initialization provided by the bisection method. It also has its own failure cases: this is a heuristic that I never should have put in the code.

2. Keep the fix I added to keep fv > fvMin: if fv ends up going to zero the gradient goes to zero. Since the choice of fv is a part of the fiber-tendon velocity distribution heuristic I can really do this without any bad impacts.  I know you rejected this earlier, but this is a problem waiting to happen.

In either case both fixes require changing the code as much as I did on the pull request that was rejected earlier. I don't see a way around this: using just the Newton method is not an option with C1 curves. I was too inexperienced when I wrote the Thelen2003Muscle implementation to realize it at the time.

mjhmilla:
@aseth1 Thank you very much for the test case! Here is the problem: the curves that are used in the Thelen2003Muscle are not even C1 continuous, and a Newton method is used to initialize the model (and it requires C2 continuous curves). The test case you've given me (that blows up) happens to have its solution near lceN = 1: for lceN < 1 the passive-force-length curve has a gradient of zero, for lceN >= 1 the passive-force-length-curve has a non-zero gradient. So the Newton method just jumps from branch-to-branch near lceN=1 and never converges. There might be another curve that blows up - I'd have to take another careful look at the equations again.

There are a number of ways of dealing with this and I've proposed a few below in order of increasing complexity. In either case a bunch of code is going to change in initMuscleState. Please @chrisdembia read this and give me your thoughts. I am not going to code a single line until we're both happy with the plan. 
 
If you'd prefer to skype I'd be happy to do so! Just give me a date/time/skype id (over email).

----------------------------------------
1. Add a bisection method to get close to the solution. Try a standard Newton method for 5-10 iterations: if the solution is better take it, otherwise return the result from the bisection method (this is what Sherm does).

Pro  : easily maintained, each iteration is pretty fast, and its stable. I have a working prototype.
Con : many iterations are required to get to a tolerance of 1e-8 N. This is however a ridiculous tolerance. We can, as an example get to a tolerance of  0.015 N with 25 bisection iterations (I've got a prototype version) on a muscle with a maximum-isometric-force of 1500 N.

2. Add a bisection method to get close to the solution, use a branch-aware Newton method to polish it.

Pro : should be able to maintain the solution tolerance that we currently enjoy, along with the speed.
Con: This should work fine but there could be surprises due to the fiber-tendon velocity distribution heuristic that I have to use. And this will take time. And be harder to maintain.

----------------------------------------

In either case I recommend 

1. Removing the line-searching feature of the Newton method that is currently in there: it is not necessary with the good initialization provided by the bisection method. It also has its own failure cases: this is a heuristic that I never should have put in the code.

2. Keep the fix I added to keep fv > fvMin: if fv ends up going to zero the gradient goes to zero. Since the choice of fv is a part of the fiber-tendon velocity distribution heuristic I can really do this without any bad impacts.  I know you rejected this earlier, but this is a problem waiting to happen.

In either case both fixes require changing the code as much as I did on the pull request that was rejected earlier. I don't see a way around this: using just the Newton method is not an option with C1 curves. I was too inexperienced when I wrote the Thelen2003Muscle implementation to realize it at the time.

mjhmilla:
To follow up on some loose ends:

1. Curve gradient check: the tendon-force-length curve could be an issue. 
2. Minimal update: at a bare minimum putting a bisection-method ahead of the (existing) Newton method and returning the best solution will fix the current problem.
3. Questions: expectations, notes, & the future

**1. Curve gradient check:**
- The initialization method depends on C2 continuous derivatives w.r.t. length, which means the active-force-length, passive-force-length, and tendon-force-length curve. The active-force-length curve is a Gaussian: its fine. The passive-force-length curve has the problem that we've already discussed: a C1 discontinuity at lceN=1. 

The tendon-force length cure has a discontinuity at ltN=1. At ltN = 1 the derivative of this curve jumps between 0 (ltN < 0) and a finite value. If the root of the initialization is small, say around 10-100x the size of the solution tolerance, then this could happen. 

The fiber-tendon velocity distribution heuristic depends on lengths only, and so the requirement on the force-velocity curve is that it be C0 continuous. Which it is.

Moving forward there are 2 curves that can cause a gradient-based root-solver to switch branches: the passive-force length curve and the force velocity curve.

**2. Minimal update**

Both @aseth1 and @chrisdembia expressed hesitation, to put it mildly, at pulling in code that fixes problems that have not yet been reported yet. As such the minimum update (given Ajay's very helpful test case) is the following:

- Insert a bisection method ahead of the Newton method and let it run for ~15 iterations. Save the best solution.
- Let the Newton method run (using the bisection method's result as a starting point) until the tolerance is met or its fixed iterations have expired
- Return the best solution.

In the worst case the user will get a solution that meets a tolerance of fiso/2^15. This isn't as good as the current 1e-8, but its still quite nice. The C1 discontinuity of the tendon curve and the issue when the force-velocity curve goes to zero won't be an issue: the bisection method is immune to these problems. I'm happy to implement this if this is what you want: you guys are after all the ones maintaining the code and working with the users. I'm embarrassed that some of my dumber ideas will continue to exist in the code (the line-search + fv can go to zero), but I can live with my embarrassment.

**3. Questions: expectations, notes, & the future**

Expectations: If I see something that could be a problem its clear that I cannot just submit code to fix it. So if I want to fix an issue that I see from looking through the code do I first have to add a test case to show that it fails before I can fix it? Is that enough?

Notes: Before Ajay's test case gets pulled into master the CMake has to be updated to copy the osim file over to the build directory: on my machine I had to do this manually, which isn't going to work for the average user. The alternative is that I hand craft a test case that has an initialization right at the sweet spot of lceN = 1.

The future: Would an implicit muscle model be useful? It should be pretty straight forward to make a version of the 2012-equilibrium model that is implicit: all of the curves need a slight modification, the pennation model as well, and one or more implicit functions need to be added. It would be nice if this were associated with an OCP run in Mocco benchmarking the solution against a problem done with an explicit model. Is there interest in this?  Is anyone working on this?

mjhmilla:
I wrote way too much earlier - a shorter version is below. I am not issuing another pull request until either @chrisdembia, @carmichaelong  or @aseth responds and approves the idea of the minimal fix, or asks for something else: I do not have a lot of time these days to work on code that gets thrown out.

**Problem**

The initialization method is using Newton's method and the Thelen2003Muscle has C1 and C2 disconinuties: the test case Ajay made happens to have a its solution at lceN=1 and so that C1 discontinuity is causing the Newton method to jump from branch to branch.

**Minimal Fix**

I can add a bisection method ahead of the Newton method and return the better of the two solutions. In the worst case the user will get an initialization with a tolerance of fiso/2^15, in the best case the Newton method will polish it up to 1e-8 N or better.

**Expectations: fixing bugs**

There are two other problems I noticed while I was looking at the initialization method: if fv goes to zero the Newton method fails, and there is a failure case with weird Newton + line-search method that is in there. I see now that I cannot just fix things, so, in the future if I see something that really should be fixed should I build a test case first?

In principle the minimal fix will cover these problems: in the worse case the rougher bisection-tolerance solution will be returned when a much better Newton-tolerance could have been. That's not so bad.

**Notes: test case **

Before Ajay's test case gets pulled into master the CMake has to be updated to copy the osim file over to the build directory: on my machine I had to do this manually, which isn't going to work for the average user. The alternative is that I hand craft a test case that has an initialization right at the sweet spot of lceN = 1.

**Related Future Work **

Is there interest in an implicit version of the 2012 Equilibrium model? If yes, this is something I could start to work on. It should require some modest changes to the curves (so that the gradients don't go to zero) and one additional function.

chrisdembia:
The dev team discussed what to do about these changes and we would be a lot more comfortable with larger changes to the equilibrium algorithm if there's a more thorough test across activation levels and MTU lengths. We have started prototyping such a test for the DeGrooteFregly2016Muscle in Moco:

https://github.com/opensim-org/opensim-moco/blob/4e77683b6c8e1dbaa5937bc994722f3de92e3f05/Moco/Tests/testDeGrooteFregly2016Muscle.cpp#L1176-L1247

Such a test should be straightforward to implement and can reduce uncertainty when changing the algorithm. I'm happy to work on the test with you.
> I see now that I cannot just fix things, so, in the future if I see something that really should be fixed should I build a test case first?

We have a pull request review process outlined [here](https://github.com/opensim-org/opensim-core/blob/master/CONTRIBUTING.md#making-a-pull-request-pr). It would be great to build a test case first and discuss the proposed changes with the development team before opening a pull request. Naturally, this depends on the scope of the changes. 


> Related Future Work

That's a great idea! I created a separate issue #2711 to discuss this.

mjhmilla:
Thanks for the response - I'll set aside some time this week to work on:
- a more generic version of the initialization test to go in testMuscles
- the minimal fix to the Thelen2003Muscle code

chrisdembia:
Sounds great :)

nickbianco:
@mjhmilla any plans to continue working on this, or should we close this PR?

mjhmilla:
@nickbianco I somehow completely missed this. If it is still a problem I can take a look at it next week. Is it still a problem?

mjhmilla:
I've just read #2686. As long as the issue of improving initialization remains open then this specific issue can close. @jenhicks mentioned that the initialization issue mentioned in #2686 is covered in other places.

So maybe it is ok to close it?

nickbianco:
@mjhmilla, not a major problem, I was just sorted through old PRs and seeing if any could be closed. We still have #1736 open, but it's not exactly related to bug. Either way, I think it's fine to close, especially if there's no plan to continue work on this in the near future.

nickbianco:
Closing: stale PR.

Issue #2689, opened 2020-03-11
Induced Accelerations can not be run when external loads are present


tnamayeshi:
When running "Induced Accelerations" analysis with "External Loads", OpenSim returns:

> External force expressedInBody ground not found.
> 
> Exception:
>   ArrayPtrs.get: Array index out of bounds.

As I investigated in the source code, this happens because OpenSim search for force body "appliedTo" and "expressedIn" in model body set and if the force applied to or expressed in "ground" it will fail to find it!

Error source is in "**applyContactConstraintAccordingToExternalForces**" function in "**InducedAccelerations.cpp**" file!

aseth1:
@tnamayeshi thank you for reporting this issue. Can you please indicate what case you are trying run and what you are trying to achieve? The logic of InducedAccelerations is intended to replace the applied external (ground reaction) force at the point of application with a constraint. If the applied point is in Ground then the external force has has no affect on the model and should be ignored. I am not sure that is the case, but perhaps you could supply your use case so we can check. Thank you.

PR #2690, opened 2020-03-11
Fixed a critical bug in InducedAccelerations which prevents analysis …

tnamayeshi:
…to run

Fixes issue #2689

### Brief summary of changes
Added extra check for external force if it's expressed in or applied to "ground" and transformed the force to respective frame.

### Testing I've completed
I successfully compiled the changes and tested for Induced Accelerations with External Loads analysis.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2690)

### DIFF
--- a/OpenSim/Analyses/InducedAccelerations.cpp
+++ b/OpenSim/Analyses/InducedAccelerations.cpp
@@ -951,10 +951,12 @@ Array<bool> InducedAccelerations::applyContactConstraintAccordingToExternalForce
 {
     Array<bool> constraintOn(false, _constraintSet.getSize());
     double t = s.getTime();
-
+    const OpenSim::Ground& ground = _model->getGround();
+	
     for(int i=0; i<_externalForces.getSize(); i++){
-        ExternalForce *exf = _externalForces[i];
-        SimTK::Vec3 point, force, gpoint;
+
+        ExternalForce* exf = _externalForces[i];
+        SimTK::Vec3 point, force;
 
         force = exf->getForceAtTime(t);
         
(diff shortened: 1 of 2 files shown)


aymanhab:
Thanks for contributing @tnamayeshi While forces can be expressed in "Ground" indeed, applying force to Ground makes little sense to me and probably should be flagged as a user error instead. Requesting review from @aseth1 as more familiar with the context/changes.

aseth1:
Thank you @tnamayeshi for bringing this issue to our attention and we very much appreciate your attempt to fix it. We definitely want you to contribute to OpenSim.  In this case, however @aymanhab is correct that if the external force is applied to Ground, then it has no effect on the model and there is no representative constraint applied. I don't believe this is a software error and I don't think it should be flagged as an error, since you should be able to use the same external loads file from a forward dynamics/CMC simulation for consistency. Did you simulate your model with an ExternalLoads applied to Ground? Were you satisfied with that simulation? If so, then it is consistent to ignore the external force and not apply a representative constraint during induced accelerations.

tnamayeshi:
@aseth1 @aymanhab  thank you very much for your comments and review. First of all, it seems that you ignored the main issue. Instead, you are noticing to another not important/never occurring one! The main issue is where the external force is "expressed" in ground not "applied" to ground! If the force is applied to ground, as you said it has no effect on the model and it will do nothing! Therefore, the model can give an error or not! Does not matter!!
However, when the force is expressed in ground, which is a valid case and that is how it actually is in the case of GRF data (i.e. force plates often gives the values of GRF expressed in ground), OpenSim should not return an error. Because it is how it should be...GRF has to be expressed in ground and is expressed in ground! Therefore, it needs fixing.

I never had a simulation with external loads applied to the ground! In fact if somebody does that, I think opensim must show a warning that those forces will be ignored! Because applying external forces to the ground is meaningless. However, I think a warning is the right way not stopping the whole simulation and returning an error because of an unhandled exception!
I can modify my code to do that for external loads "applied to the ground"! But I think it is only valid to remove the error for external loads "expressed in ground". That is a valid and critical issue which makes all previous "Induced Acceleration Analysis" results invalid, because they couldn't include GRF in their analysis!

aseth1:
@tnamayeshi in #2689  I wrote:

> I am not sure that is the case, but perhaps you could supply your use case so we can check. Thank you.

It very well may be a bug and we appreciate you finding it. It would be great if the case you found was added as a test case to reveal both the bug and serve as verification of the fix.

In hindsight, a bug could have been introduced in 4.0 when we added reference Frames. `Ground` is now simply a `PhysicalFrame` not a `Body` (which has CoM, mass and inertia  properties) and likely the code was not updated because the test case did not fail. My guess is that the test case has the GRF and CoP expressed in `Ground`, but uses IK kinematics to transform the point into the foot frame, which makes the point expressed in the same Body it is applied to and does not enter the code block that requires the transformation from Ground to the foot Frame. If this is the case, the easiest fix is to replace code related to:

```.cpp
    const Body &appliedToBody = _model->getBodySet().get(appliedToBodyIndex);
    const Body &expressedInBody = _model->getBodySet().get(expressedInBodyIndex);
```
with,
```.cpp
                const PhysicalFrame* appliedToBody =
                        _model->findComponent<PhysicalFrame>(exf->getAppliedToBodyName());
                const PhysicalFrame* expressedInBody =
                        _model->findComponent<PhysicalFrame>(exf->getPointExpressedInBodyName());

                OPENSIM_THROW_IF_FRMOBJ(appliedToBody == nullptr, Exception,
                        "ExternalForce's appliedToBody " +
                                exf->getAppliedToBodyName() + " not found.");

                OPENSIM_THROW_IF_FRMOBJ(appliedToBody == nullptr, Exception,
                        "ExternalForce's pointExpressedInBodyName " +
                                exf->getPointExpressedInBodyName() + " not found.");
```

Ideally, `ExternalForce` would use `Socket`s for its dependencies and the reference to the `PhysicalFrame` accessed directly without searching.

tnamayeshi:
I tried your code and it simply throws an error :

`Expected stage to be at least Position in getCacheEntry _but_ current stage was Model`

 **Stack trace:**
`>	SimTKcommon.dll!SimTK::CacheEntryInfo::throwHelpfulOutOfDateMessage(const SimTK::StateImpl & stateImpl, const char * funcName) Line 964	C++
 	[Inline Frame] SimTKcommon.dll!SimTK::StateImpl::getCacheEntry(const std::pair<SimTK::SubsystemIndex,SimTK::CacheEntryIndex> &) Line 2015	C++
 	SimTKcommon.dll!SimTK::State::getCacheEntry(SimTK::SubsystemIndex subsys, SimTK::CacheEntryIndex index) Line 2480	C++
 	osimSimulation.dll!OpenSim::Frame::getTransformInGround(const SimTK::State & s) Line 95	C++
 	[Inline Frame] osimSimulation.dll!OpenSim::Frame::findTransformBetween(const SimTK::State &) Line 218	C++
 	osimSimulation.dll!OpenSim::Frame::findStationLocationInAnotherFrame(const SimTK::State & state, const SimTK::Vec<3,double,1> & station_F, const OpenSim::Frame & otherFrame) Line 239	C++
 	osimAnalyses.dll!OpenSim::InducedAccelerations::applyContactConstraintAccordingToExternalForces(SimTK::State & s) Line 978	C++
 	osimAnalyses.dll!OpenSim::InducedAccelerations::record(const SimTK::State & s) Line 531	C++
 	osimSimulation.dll!OpenSim::AnalysisSet::begin(const SimTK::State & s) Line 202	C++
 	osimTools.dll!OpenSim::AnalyzeTool::run(SimTK::State & s, OpenSim::Model & aModel, int iInitial, int iFinal, const OpenSim::Storage & aStatesStore, bool aSolveForEquilibrium) Line 656	C++
 	osimTools.dll!OpenSim::AnalyzeTool::run(bool plotting) Line 583	C++
 	osimJavaJNI.dll!Java_org_opensim_modeling_opensimActuatorsAnalysesToolsJNI_AnalyzeTool_1run_1_1SWIG_10(JNIEnv_ * jenv, _jclass * jcls, __int64 jarg1, _jobject * jarg1_) Line 49688	C++
 	[External Code]	
`


But my code works fine without any errors!

aseth1:
According to the stack trace the following line was either omitted:
```cpp
_model->getMultibodySystem().realize(s, SimTK::Stage::Velocity);
```
or not called prior to calling `findStationLocationInAnotherFrame()`.

tnamayeshi:
OK, with the line you provided, it is not giving any errors as expected.
I created another pull request (#2691) in master and applied the bug fix

PR #2695, opened 2020-03-26
Connecting Ezc3d to OpenSim

pariterre:
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2695)


carmichaelong:
> The implementation of extendRead() is a little hard to follow with all the macros now.
 
I assume you all the #ifdef's? I skimmed over the code and I think my vote would be to go for a little more code duplication to remove most of these.

pariterre:
The tests pass, but I suspect it is because neither BTK nor ezc3d is actually tested...

chrisdembia:
> The tests pass, but I suspect it is because neither BTK nor ezc3d is actually tested... @chrisdembia is that expected?

Yes, though I want to get @aymanhab's opinion on if we should change this.

The last thing to do for this PR is to improve the clarity of `extendRead()` by reducing the number of `ifdef`s. What do you think about this, @pariterre ?

pariterre:
@chrisdembia I think it makes sense, I wasn't sure if it was something you guys wanted to do, or if you prefer I do it. I have no opinion about that: I mean, I will happily do it as I think it is a matter of few minutes, but I am not sure to what extend you want to have clarified it and therefore if you guys was preferring to do it yourselves

aymanhab:
Thanks much @pariterre and @chrisdembia for putting this together, I'll review it today.

aymanhab:
Thanks for the feedback I'll try to build on my windows machine and report back.

pariterre:
@aymanhab @carmichaelong @chrisdembia 
I made a lot of tests during the previous two days in order to improve the speed of this simple linear algebra library. The conclusion is that what slows down the code is not the calculation as I expected, but the allocation and deallocation of the matrices. In Eigen, they are using a highly optimized way that I am nowhere near a good programmer enough (nor that I have the resources) to be able to reproduce. I tried to connect ezc3d to Eigen as backend to store the data, and the reading and analysis of one of my test file (28Mo) is reduced from 530ms to 300ms.

So my question to you: do you guys prefer to accelerate the reading of the files by connecting Eigen as backend or would you prefer to not rely on an external library but being a little slower?

chrisdembia:
My opinion is that the current implementation is working and we proceed with it until speed poses an issue for users.

aymanhab:
Thanks for the investigation @pariterre I agree with @chrisdembia not to spend more time optimizing now, the reason I brought this up to begin with was that the test case time-limit was increased 5x and it wasn't clear how that would scale to larger files.

chrisdembia:
2x, only.

aymanhab:
Trying to build on windows with "None" or "BTK" C3D_PARSER succeeds, however picking ezc3d option enables ezc3d_DIR cmake variable, however setting this to {dependencies_install/ezc3d/Lib/cmake) produces an error as attached.
Capturecmakeezc3d
Note that the folder actually contains ezc3dConfig.cmake 
I can investigate further if you give me pointer what to look for and where. Thank you

CMakeLists.txt
```diff
@@ -584,24 +584,48 @@ set(OPENSIM_DEPENDENCIES_DIR
...
-set(WITH_BTK
-    OFF
-    CACHE
-    BOOL
-    "Compile OpenSim with BTK ? BTK provides C3D reading.")
-
-# If compiling with BTK, find and use it.
-if(WITH_BTK)
-    find_package(BTK 
+set(C3D_PARSER None CACHE STRING
```
chrisdembia, line 587: @pariterre can you change this to something like `OPENSIM_C3D_PARSER` to be consistent with some other CMake options/settings?

chrisdembia: I plan to make a PR to do this renaming.

chrisdembia:
@aymanhab I'm testing out Windows.

chrisdembia:
@aymanhab I opened https://github.com/pariterre/opensim-core/pull/4 to fix the issues on Windows. 

Perhaps @pariterre can merge https://github.com/pariterre/opensim-core/pull/4 and then @aymanhab can test on Windows again?

aymanhab:
Thanks @chrisdembia and @pariterre I'll test again on windows

aymanhab:
Thanks again, the build worked smoothly no issues with CMake, the test case for C3D failed for timing
>Test failed due to exception: Unable to load 'walking2.c3d' within 400000000ns.
We can adjust the timing on the test for now just to make sure dev team doesn't run into it. 

CI runs without timing (and default to BTK) so should be ok on that front.

aymanhab:
Full timing info here:
1>Starting test testC3DFileAdapter ...
1>  test ...
1>	C3DFileAdapter 'walking2.c3d' loaded in 651 millisecond(s)
1>	Wrote 'walking2_markers.trc' in 162 millisecond(s)
1>	Wrote'walking2_grfs.sto' in 57 millisecond(s)
1>	Read'walking2_markers.trc' and its standard in 52 millisecond(s)
1>	Markers walking2_markers.trc equivalent to standard.
1>	Forces walking2_grfs.sto equivalent to standard.
1>	C3DFileAdapter 'walking2.c3d' read with forces at COP in 574 millisecond(s)
1>  test done. real/CPU ms: 13461.9 / 6687.5
1>Test failed due to exception: Unable to load 'walking2.c3d' within 400000000ns.

chrisdembia:
What should we do to move this PR forward, @aymanhab ?

aymanhab:
@chrisdembia I'd take out the timing part of the test since it doesn't seem reliable anyway, then we can merge.

chrisdembia:
I merged this PR into a feature branch, and will now fix testing in the feature branch.

pariterre:
Thanks a lot for the follow up!
That was a pleasure to work on this :)

chrisdembia:
It was a pleasure working with you in this. Thank you for responding so quickly. 

The work is not quite done. We must still merge into the master branch.

Issue #2717, opened 2020-04-08
Move Moco's visualize() free function to opensim-core


chrisdembia:
Moco has a very handy visualize() free function for animating a trajectory using the Simbody Visualizer.

Related: https://github.com/opensim-org/opensim-core/pull/2590

aymanhab:
@chrisdembia I'm thinking of promoting this function to a utility program (that can be invoked thru opensim-cmd) that includes visualization of marker data and IMU data. Thoughts?

chrisdembia:
That would be fantastic!

jenhicks:
From @chrisdembia and @nickbianco : could call it "playback()" or "animate()".

PR #2722, opened 2020-04-09
Visualization utils

aymanhab:
Fixes issue #2717 

### Brief summary of changes
Moved an incarnation of Moco method to SimulationUtilities, renamed method to visualizeModelMotion to be specific in anticipation of additional methods. 
 
Naming, 
I also plan to move visualization of data to DataVisualizer class for mocap and imu data. All will be accessible through extension of opensim-cmd
### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2722)


aymanhab:
@chrisdembia @carmichaelong @nickbianco thoughts on the naming and plan in description

chrisdembia:
Fantastic! What additional methods do you anticipate? This could help us give feedback on naming. Perhaps we can discuss in today's standup meeting.

aymanhab:
Definitely would be good to discuss in meeting. As of now we have a separate utility program to visualizeModel, this could be moved there and the utility can be removed altogether.

aymanhab:
@chrisdembia Also found another incarnation of the method that takes TimeSeriesTable, which signature did you find more convenient/used in Moco?

chrisdembia:
> @chrisdembia Also found another incarnation of the method that takes TimeSeriesTable, which signature did you find more convenient/used in Moco?

We use both.

aseth1:
@aymanhab @chrisdembia it would be great if there was a proposal to visualize "Experimental" data (using representative components) with a Model that is being simulated. For example, how would one use the `DataVisualizer` (I would call it DataPreviewer) with the IKSolver to visualize IK motion with both model (with its Markers) AND  `ExperimentalMarker's? It would also be great to run a forward simulation and visualize external forces applied to the feet or hands. If we can visualize model(s) and experimental data together, then previewing becomes trivial since that is just the case where there is no model.

chrisdembia:
Is this still a work in progress?

aymanhab:
@chrisdembia i wanted to move the methods into DataVisualizer class (to parallel ModelVisualizer) rather than in global namespace. As of now this is just a refactoring of existing code, we can get this reviewed and merged before introducing new functionality.

aymanhab:
This document summarizes the status and naming issues we discussed
https://docs.google.com/document/d/104b6NpkI7c9DwiYzuDKArkmd6g8gSu-XlMi3R6eVmNU/edit?usp=sharing

aymanhab:
Ready for review, particularly feedback on method names, before adding new functionality. Thanks @chrisdembia @nickbianco @carmichaelong for the feedback on this PR and earlier on the proposal.

aymanhab:
Thanks @chrisdembia

PR #2725, opened 2020-04-10
Make code consistent with comment 

aymanhab:
Utility claims visualizer will be displayed (to test installation integrity) but UseVisualizer was not set

Fixes issue #0

### Brief summary of changes
Uncommented line that made the code and user message inconsistent

### Testing I've completed
Ran utility in dev environment and a visualizer window was opened.

Should we open an issue and include in change log?
### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2725)


aymanhab:
@chrisdembia this PR is not part of the test suite, however all platform tests fail because of simbody, any clues? Thanks

aymanhab:
Thanks @carmichaelong

carmichaelong:
Oh weird, I thought at first the failures were random since this was such a small change. Sorry, I should have been more careful there. Looks like consistent failure with `testExampleModelRuns `

chrisdembia:
I think we need to disable that line if the test is run as part of the test suite. We use the environment variable `OPENSIM_USE_VISUALIZER` in for the Matlab tests to handle this; we could also use this for `testExampleModelRuns`.

https://github.com/opensim-org/opensim-core/blob/048d13753a6d1f153e5481984a4576bfd65be218/Bindings/Java/Matlab/tests/CMakeLists.txt#L53-L54

OpenSim/Examples/checkEnvironment/checkEnvironment.cpp
```diff
@@ -204,7 +204,7 @@ int main()
         // PERFORM A SIMULATION //
         /////////////////////////
         
-        //osimModel.setUseVisualizer(true);
+        osimModel.setUseVisualizer(true);
```
chrisdembia, line 207: Next, use the environment variable in the code here.

aymanhab:
@chrisdembia the error below indicates failure before launching the visualizer, is that the expected behavior? I expect seg-fault or timeout. 

simbody-visualizer: error while loading shared libraries: libSimTKsimbody.so.3.8: cannot open shared object file: No such file or directory

chrisdembia:
This is probably not the expected behavior. I don't know the reason for this error message.

aymanhab:
Thanks for the feedback @chrisdembia, ready for quick review. 
We need to test an artifact built off master after merge to check visualization functionality.

chrisdembia:
My understanding is it's an example that helps users test if OpenSim is installed properly.

nickbianco:
Ah okay, like a self-test.

aymanhab:
Thanks @carmichaelong @nickbianco and @chrisdembia 
It is a utility that we distribute so that users can test if the installation was functional. We include it in tests out of abundance of caution to make sure the code compiles and runs but the actual visualization needs to be checked manually in an installed version.

PR #2726, opened 2020-04-11
Document adding dependencies.

chrisdembia:
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2726)


chrisdembia:
I just pushed an update to this documentation.

chrisdembia:
Ready for review.

I've requested reviews from @aymanhab, @carmichaelong, and @nickbianco because I imagine all of you are curious about the procedure for adding dependencies and could help point out any parts of this documentation that are too vague.

To some extent, the instructions are intentionally vague; each dependency has its own needs, so it's difficult to be too specific.

I also fixed an issue where ezc3d/BTK binaries were being copied even when they shouldn't be. The lines I deleted are duplicated further down in `CMakeLists.txt`. I tested on my Mac that the lines I deleted are not necessary.

aymanhab:
LGTM @chrisdembia I just had one request for clarification regarding python standalone, you can ignore if it mostly works in your experience. Thanks for the nice job documenting the process 🥇

aymanhab:
Thanks much LGTM

PR #2727, opened 2020-04-11
Disable timing test with ezc3d.

chrisdembia:
### Brief summary of changes

Remove the C3D timing test for ezc3d.

By default, superbuild no longer builds C3D libraries.

### Testing I've completed

Ran testC3DFileAdapter with BTK and with ezc3d and ensured that the timing test only occurs with BTK.

### CHANGELOG.md (choose one)

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2727)

### DIFF
--- a/dependencies/CMakeLists.txt
+++ b/dependencies/CMakeLists.txt
@@ -56,12 +56,13 @@ endfunction()
 
 # Add a dependency. Arguments:
 #   NAME       -- (Required) Name of the project.
+#   DEFAULT    -- (Required) Default value for SUPERBUILD_${NAME} variable.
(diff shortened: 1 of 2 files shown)


dependencies/CMakeLists.txt
```diff
@@ -56,12 +56,13 @@ endfunction()
 
 # Add a dependency. Arguments:
 #   NAME       -- (Required) Name of the project.
+#   DEFAULT    -- (Required) Default value for SUPERBUILD_${NAME} variable.
 #   URL        -- (Required) git repository to download the sources from.
```
aymanhab, line 60: 👍

PR #2729, opened 2020-04-11
Test muscles initialization

mjhmilla:
Fixes issue #2686 

### Brief summary of changes

- Function runMuscleInitializationSweep has been added: it tests the initialization method for a Muscle object across a dense mesh across activation, path length, and path speed. 
- The settings for runMuscleInitializationSweep are currently globally set and are defined in the struct InitializationTestSettings.
- Because the numerical problems of initialization are related to the tendon slack length the function runMuscleInitializationSweep is applied across the tendon slack lengths listed in NormTendonLengths.
- I have added detailed documentation for runMuscleInitializationSweep and InitializationTestSettings but not with doxygen braces since this is not API code.

### Testing I've completed

This test has been applied to the muscle models that are actively maintained: Thelen200Muscle, Millard2012EquilibriumMuscle, Millard2012AccelerationMuscle. They all fail the test but for different reasons:

- Thelen2003Muscle: fails due to the bad mix of a Newton method + curves that are not C2 continuous.
- Millard2012EquilibriumMuscle: fails because it always returns a solution with a fiber velocity of zero. This is happening because of a flag that is permanently set to false (line 431 in Millard2012EquilbriumMuscleh). However the problem is deeper: even when the computeFIberEquilibrium method is set to compute fiber velocity the method fails.
- Millard2012AccelerationMuscle: fails because the method tries to evaluate a fiber length that is shorter than the minimum length (defined by the pennation model in this case). The initialization method is simply not respecting the lower bound.

On two aspects from one or all of: @chrisdembia, @carmichaelong , and  @aseth1 :

1. Is this code acceptable? If not please tell me what needs some attention.
2. Planned fixes for the Muscle models:

- Remove the Newton method that is currently used in all of these models.
- Option 1: Slow & simple: use a bisection method to get to the final tolerance. This will take about 30 iterations for a fiso*1e-8 solution.
- Option 2: Faster: use bisection to get to an initial tolerance of 1e-3*fiso (10 iterations). Try Newton for a fixed number of iterations. If this improves, continue to tolerance (6 iterations). If it does not, bring the solution to tolerance using bisection (for another 20 iterations).
- Ensure that the lower bound on fiber length is respected, and the fiber velocity is less than the maximum value (exceeding this causes the gradient to go to zero).

### CHANGELOG.md (choose one)

- I haven't updated this: should I?

The Doxygen for this PR cannot be viewed: this is test code, not API code. I have documented the new function extensively but it is not enclosed with doxygen's /**

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2729)

### DIFF
--- a/OpenSim/Actuators/Test/testMuscles.cpp
+++ b/OpenSim/Actuators/Test/testMuscles.cpp
@@ -50,11 +50,11 @@ using namespace std;
 //==============================================================================
 static const double IntegrationAccuracy         = 1e-5;
 static const double SimulationTestTolerance     = 1e-4;
-static const double InitializationTestTolerance = 1e-6;
+static const double EquilibriumTestTolerance = 1e-6;
 static const double CorrectnessTestTolerance    = 1e-6;
 
 static const int SimulationTest         = 0;
-static const int InitializationTest     = 1;
+static const int EquilibriumTest     = 1;
 static const int CorrectnessTest        = 2;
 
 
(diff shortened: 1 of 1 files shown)


chrisdembia:
I've merged this branch into the `feature_equilibrium` branch. @mjhmilla, you can create subsequent PRs that target `feature_equilibrium`.

Issue #2710, opened 2020-04-07
Report IMU tracking errors when running InverseKinematics on IMU data


aymanhab:

jenhicks:
@aymanhab This was something that came up as useful info to help with the IMU project (with Mazen, Hannah).

aymanhab:
Thanks @jenhicks for the heads up, PR #2733 has a fix.

aymanhab:
@malborno Not sure if you can try PR #2733 to see if reporting is helpful or can be improved. No GUI build yet but can trigger one if needed, also maybe helpful to Hannah.

PR #2733, opened 2020-04-14
Report orientationErrors from orientation sensors tracking

aymanhab:
Fixes issue #2710 
### Brief summary of changes
Report errors maintained already by InverseKinematicsSolver to the user based on report_errors property.
### CHANGELOG.md (choose one)

- updated...

### DIFF
--- a/OpenSim/Tools/IMUInverseKinematicsTool.cpp
+++ b/OpenSim/Tools/IMUInverseKinematicsTool.cpp
@@ -164,9 +164,27 @@ void IMUInverseKinematicsTool::runInverseKinematicsWithOrientationsFromFile(
     ikSolver.setAccuracy(accuracy);
 
     auto& times = oRefs.getTimes();
-
+    Storage* modelOrientationErrors =
+            get_report_errors() ? new Storage(times.size(), "ModelOrientationErrors")
+                                : nullptr;
     s0.updTime() = times[0];
     ikSolver.assemble(s0);
+    // Create place holder for orientation errors, populate based on user pref.
+    // according to report_errors property
+    int nos = ikSolver.getNumOrientationSensorsInUse();
+    SimTK::Array_<double> orientationErrors(nos, 0.0);
+
+    if (get_report_errors()) { 
+        Array<string> labels;
+        labels.append("time");
+        for (int i = 0; i < nos; ++i) {
+            labels.append(ikSolver.getOrientationSensorNameForIndex(i));
+        }
+        modelOrientationErrors->setColumnLabels(labels);
+        ikSolver.computeCurrentOrientationErrors(orientationErrors);
+        modelOrientationErrors->append(
+                s0.getTime(), nos, &orientationErrors[0]);
+    }
     if (visualizeResults) {
         model.getVisualizer().show(s0);
         model.getVisualizer().getSimbodyVisualizer().setShowSimTime(true);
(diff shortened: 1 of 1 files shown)


OpenSim/Tools/IMUInverseKinematicsTool.cpp
```diff
@@ -164,16 +164,39 @@ void IMUInverseKinematicsTool::runInverseKinematicsWithOrientationsFromFile(
     ikSolver.setAccuracy(accuracy);
 
     auto& times = oRefs.getTimes();
-
+    Storage* modelOrientationErrors =
```
aseth1, line 167: Would be great to see this implemented using a `TableReporter`. IMHO we should be committed to the removal of Storage or at least not to propagate its usage.

aymanhab: Thanks for the feedback @aseth1 👍 
The reason we can't use TableReporter here (or in older IKTool) is that Reporter requires "Outputs" to connect to its inputs. None of these quantities are outputs AFAICT. We could however use a TimeSeriesTable instead of Storage object and then write to storage file thus avoiding the use of Storage class.

aymanhab: Markers error reporting has been there for a while though works differently by reporting only worst marker, are you proposing changing the behavior? or additional reporting? @aseth1

OpenSim/Tools/IMUInverseKinematicsTool.cpp
```diff
@@ -200,7 +222,10 @@ void IMUInverseKinematicsTool::runInverseKinematicsWithOrientationsFromFile(
 
     std::cout << "Wrote IK with IMU tracking results to: '" <<
         outputFile << "'." << std::endl;
-
+    if (get_report_errors()) {
+        STOFileAdapter_<double>::write(*modelOrientationErrors,
+                get_results_directory() + "/" + "orientationErrors.sto");
```
chrisdembia, line 227: I think the user-given name of the tool should be included in the filename.

aymanhab: Done, thanks for the feedback @chrisdembia

OpenSim/Tools/IMUInverseKinematicsTool.cpp
aseth1: Sorry for my brevity:
1. I  did mean using a `TableReporter` and to use `Outputs`.  The `Component` abstraction is not only for `ModelComponent`s but for Solvers and Operators and other calculators that use System resources. It would be straightforward to make `Solver` derive from `Component`. `InverseKinematicsSolver` would have `ListOutput` with a `Channel` for every error. These would include Coordrinate, Marker (distance) and Orientation (angle) errors. Then, we would not be restricted to accessing errors in files only, but using Solvers in the context of broader pipeline. That was always the intent.
2. I mean actual reporting of  Marker errors as quantities, not just the text of the worst marker in the Messages window (current behavior). Many users (including myself) have scripts that diff the marker data in a trc file against marker locations reported by IK in .sto so they can compute distances and use them to refine scaling and IK. With Outputs, it would be easier for users could to select just the Markers/Sensors/Coordinates they care about.

Since the functionality has not yet been exposed and widely used, I think now would be the best time to make the upgrade.

aymanhab: Thanks for the feedback @aseth1, since this is a major conceptual change that needs further discussion (of whether Component is the best abstraction to derive Solvers from)which is way beyond the scope of this PR and the fact that local users are waiting on this for ongoing research. I opened issue #2740 to continue the discussion while not blocking ongoing research.

aymanhab:
@aseth1 it would be great to hear your thoughts/feedback on this PR by Thursday 4/23, since there's ongoing research that needs it rather quickly. If you need more time also please let us know. Thank you.

aseth1:
@aymanhab OK. Please also note that as AFAIK, the behavior of `get_report_errors()` is different between `IMUInverseKinematicsTool`  (reports errors to file) and `InverseKinematicsTool` (prints summary message to console) . IMHO it is a mistake to have two separate classes for the same underlying Solver and it is made worse by having different behaviors.

aymanhab:
Thanks @aseth1 and @chrisdembia for the review.

PR #2741, opened 2020-04-16
Replace BTK with EZC3D on ci

aymanhab:
Fixes issue #<issue_number>

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2741)


appveyor.yml
```diff
@@ -105,7 +105,7 @@ build_script:
   # Configure. # TODO -DBUILD_SIMM_TRANSLATOR=ON
   # Set the CXXFLAGS environment variable to turn warnings into errors.
   # Skip timing test section included by default
-  - cmake -E env CXXFLAGS="/WX -DSKIP_TIMING" cmake %OPENSIM_SOURCE_DIR% -G"%CMAKE_GENERATOR%" -T"%CMAKE_TOOLSET%" -DSIMBODY_HOME=C:\simbody%NUGET_PACKAGE_ID_SUFFIX% -DOPENSIM_DEPENDENCIES_DIR=%OPENSIM_DEPENDENCIES_INSTALL_DIR% -DCMAKE_INSTALL_PREFIX=%OPENSIM_INSTALL_DIR% -DBUILD_JAVA_WRAPPING=ON -DBUILD_PYTHON_WRAPPING=ON -DWITH_BTK:BOOL=ON
+  - cmake -E env CXXFLAGS="/WX -DSKIP_TIMING" cmake %OPENSIM_SOURCE_DIR% -G"%CMAKE_GENERATOR%" -T"%CMAKE_TOOLSET%" -DSIMBODY_HOME=C:\simbody%NUGET_PACKAGE_ID_SUFFIX% -DOPENSIM_DEPENDENCIES_DIR=%OPENSIM_DEPENDENCIES_INSTALL_DIR% -DCMAKE_INSTALL_PREFIX=%OPENSIM_INSTALL_DIR% -DBUILD_JAVA_WRAPPING=ON -DBUILD_PYTHON_WRAPPING=ON -DWITH_EZC3D:BOOL=ON
```
chrisdembia, line 108: This is the incorrect flag; should be `-DOPENSIM_C3D_PARSER=ezc3d`. Same for the Travis file.

aymanhab:
@pariterre I'm trying to make EZC3D the default option on our ci machines but the tests are failing for Python and Java (they just compare the sizes of the tables against a standard) did you run these locally when you were testing? Ideally we don't need to modify the tests and they should pass using either BTK or EZC3D if we're getting the same data. Thank you

pariterre:
Dear @aymanhab 
The tests were passing locally when I wrapped up the implementation. Maybe I changed something in the lasts commit and didn't realize it changed something important. I'll have a look this morning.

Sorry about that!

pariterre:
Hello again!
I just realized that the failing results are from the Python binding. I did not test the Python binding. So I compiled this morning with BUILD_PYTHON_WRAPPING set to ON. However I get the following error 
```
import opensim
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/pariterre/Programmation/miniconda3/envs/opensim-dev/lib/python3.8/site-packages/opensim/__init__.py", line 1, in <module>
    from .simbody import *
  File "/home/pariterre/Programmation/miniconda3/envs/opensim-dev/lib/python3.8/site-packages/opensim/simbody.py", line 13, in <module>
    from . import _simbody
ImportError: cannot import name '_simbody' from partially initialized module 'opensim' (most likely due to a circular import) (/home/pariterre/Programmation/miniconda3/envs/opensim-dev/lib/python3.8/site-packages/opensim/__init__.py)
```
whether C3D parsers are ON or OFF. Am I doing something wrong in the compilation process?

aymanhab:
Thanks @pariterre. Indeed opensim-core without wrapping passes all tests but the Java and Python tests fail.  The issue with python you observed is likely related to your python environment since the ci machines python runs successfully but fails. Here are the code snippets that fail:
https://github.com/opensim-org/opensim-core/blob/064b38252d55ce38394d25e2a5668bb0d3452a09/Bindings/Java/tests/TestC3DFileAdapter.java#L6
https://github.com/opensim-org/opensim-core/blob/064b38252d55ce38394d25e2a5668bb0d3452a09/Bindings/Python/tests/test_DataAdapter.py#L37
Not sure which lines fail but i can go deeper if you think these should pass as is. Thank you

pariterre:
@aymanhab 
In my understanding, it should not fail. When I augmented the C3DFileAdapted I've put great caution in returning exactly the same as what BTK was returning. Something may have slipped of course. 

I wanted to test by using the BTK backend to see if it produces the same error, but I did not manage to solve the importing problem in Python...

aymanhab:
Here's the block that fails @pariterre 
`        StdVectorMatrix fpCorners =
            forceTable.getTableMetaDataVectorMatrix("Corners");
        assert fpCorners.size() == 2;
        Matrix corner0 = fpCorners.get(0);
        assert corner0.nrow() == 3; 
        assert corner0.ncol() == 4;`
Note that the C++ test doesn't check this meta data so no surprise it passes while the java test fails.

pariterre:
I PRed (#2750 ) a fix for that!

Issue #36, opened 2014-06-21
Add message logging/handling service


aseth1:
OpenSim and its Component builders need a way to provide debug and other component messages in a systematic way. Users should be able to choose the level of messaging from non to the most fine Optimizer details. I propose that Component could maintain the messages with an assigned priority/verbosity level which the Component builder can specify using `addMessage(string msg, level)` like methods we have for Output, StateVariable, etc...  One of the advantages of this approach is that it would make it easy to have selective verbosity- for example having only Muscles running with full verbosity to debug muscle related issues.

- [x] Common (@chrisdembia )
  - [x] Reporter (@chrisdembia )
- [x] Simulation (@aymanhab )
- [x] Actuators
- [x] Analyses (@nickbianco)
- [x] Tools (@nickbianco )
- [x] Applications (@carmichaelong )
- [x] Change logging level with command-line flag. (@nickbianco )
- [x] Adding/removing custom sinks.
- [x] Test logging in Matlab on Windows. (@aymanhab )
- [x] Test logging in Matlab on Mac. (@aymanhab )
- [x] Disable/remove old logging infrastructure (LogManager, etc.)
- [x] Test logging in GUI on Mac (@chrisdembia )
- [ ] User documentation for how to change the log level in CONTRIBUTING.md (@chrisdembia )
  - [ ] Create private Confluence page.
- [x] Find and replace usages of cerr, cout, and printf. (@chrisdembia )

sherm1:
Nice that the easylogging++ guy is responsive! I filed an issue there to see if he'll fix his use of illegal identifier names (standard prohibits leading underscores): easylogging/easyloggingpp#239.

mkhan3189:
Legal identifiers are successfully implemented from [Easylogging++ v9.79](https://github.com/easylogging/easyloggingpp/releases/tag/v9.79)

sherm1:
> Legal identifiers are successfully implemented from Easylogging++ v9.79

Thanks, Majid!

chrisdembia:
What do y'all think about renaming the output files?

```
out.log -> opensim_output.log
err.log -> opensim_error.log
```

tkuchida:
>What do y'all think about renaming the stream output files?

:+1:

chrisdembia:
Another 1-header logging library: https://github.com/emilk/loguru.

Getting component-level logging seems like the most challenging part of all of this; I'd advocate getting just *some* logging working, then we can worry later about improving it.

[...]

PR #2742, opened 2020-04-18
Logging actuators

chrisdembia:
Related to #36 

### Brief summary of changes

- Logging to a file is on by default.
- Add command-line flag to set logging level when using `opensim-cmd`.
- Introduced `log_cout()` for logging messages in `print`-like functions. Such messages are printed regardless of the logging level, and go to all registered sinks, including the file sink. Callers of `print`-like functions can use `Logger::shouldLog()` to determine if the `print`-like function should be called.
- Update logging in the Common library, the Model class, and other files.

### Testing I've completed

- Tested `addFileSink()`, `removeFileSink()` in Python.
- Tested running the `opensim-cmd` tool with different `--log` levels.

What do you think about the logic for handling `print`-like functions?

### CHANGELOG.md (choose one)

- no need to update because...part of a feature

### DIFF
--- a/Applications/opensim-cmd/opensim-cmd.cpp
+++ b/Applications/opensim-cmd/opensim-cmd.cpp
@@ -38,7 +38,7 @@ static const char HELP[] =
 R"(OpenSim: musculoskeletal modeling and simulation.
 
 Usage:
-  opensim-cmd [--library=<path>]... <command> [<args>...]
+  opensim-cmd [--library=<path>]... [--log=<level>] <command> [<args>...]
   opensim-cmd -h | --help
   opensim-cmd -V | --version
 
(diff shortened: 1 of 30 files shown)


aymanhab:
Thanks @chrisdembia no need for me to review this one.

chrisdembia:
The command-line tool is failing on Windows: 

```
regex_error(error_stack): There was insufficient memory to determine whether the regular expression could match the specified character sequence.
```

I've ran into this issue before. I'm not sure yet how to fix it.

aymanhab:
Can we separate the cmd-line changes to a separate PR while we investigate?

chrisdembia:
Thanks @nickbianco

PR #2743, opened 2020-04-20
Remove DevelopmentProposals/

adamkewley:
This PR removes `./DevelopmentProposals` from the repo.

From what I can see in the logs, it was last used in 2016 and contains now-stale suggestions? Dev proposals, like this one, are typically done via Issues and PRs for GitHub-based projects.

If this is a bad call just feel free to reject this PR - I'm just beginning to spin up on the OpenSim codebase and I figured the easiest way to do that would be to submit PRs and see what the current maintainers think.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2743)


chrisdembia:
Thanks @adamkewley! We'll try to discuss this in our developer meeting tomorrow.

chrisdembia:
Hey @adamkewley. We think this folder could still be useful, especially for big projects from outside contributors. But the two proposals `componentConnectInterface` and `reporterColumnLabels` are stale/inactive. Could you create a `./DevelopmentProposals/archive` folder and move these old proposals into the archive? It would also be great if you could add a note to `./DevelopmentProposals/readme.md` saying that proposals should be moved into the archive if they are implemented or are inactive.

adamkewley:
@chrisdembia updated the branch (forcibly, because the intermediate commit isn't historically useful) to have a `./DevelopmentProposals/archive` directory which contains the stale proposals. I also updated the `./DevelopmentProposals/readme.md` with a note about archiving (not so high-detail).

chrisdembia:
Neat!

We do not typically force-push because someone might have tried your branch locally, but I think it's alright in this situation.

PR #2744, opened 2020-04-22
WIP: Removed docopt dependency (passes testCommandLineInterface)

adamkewley:
This is a work in progress - code still needs refactoring + cleaning up, but the general idea is in there. 

There is no expectation that this work will be merged: I'm playing around with the OpenSim codebase whenever I get a spare hour. I just wanted something concrete to work on so that I could understand the build, test, dependencies, style, etc. of OpenSim, rather than because this is something I feel passionately about.

This changeset removes `docopt` from the codebase. `docopt` is a fine CLI argument parser but I feel that there's a tradeoff between pulling it in vs. home-rolling this. Integrating, documenting, and dependency-managing `docopt` might actually end up having a bigger cognitive overhead than just writing a few basic argument parsing functions (assuming the replacements aren't buggy, of course ;))

Notes:

- Merged the various subcommand headers into a single `opensim-cmd.cpp` file.
- Only used by one part of the platform (opensim-cmd)
- Removes overhead of having to maintain github links, docs, download the dependency, etc.
- Removes overhead of having to understand how docopt parses help text strings
- Allows for customization (of behavior, when things happen, etc.)

- Drawback: increases code complexity (because parsing is done in-tree). Custom
  parser avoid any heavy abstractions in order to make it simple enough to
  understand

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2744)


chrisdembia:
I'm open to removing the `docopt` dependency, especially because it can cause errors with MSVC's std regex library (running out of memory?). 

However, I would prefer that all command-line parsing occur outside of the logic of the command-line tools. That is, perhaps we could flesh out more of our own simple command-line parsing library. While I trust an individual to write bug-free command-line parsing, the command-line tool becomes more difficult to extend by future maintainers/developers if the command-line parsing is done somewhat "manually."

Another alternative is to find a command-line parsing library that is simpler than docopt.cpp.

I like having the subcommands in separate files, as I think that encourages expanding the command-line tool to become more capable over time.

adamkewley:
@chrisdembia yeah, sounds reasonable.

For alternatives, I've had some success with something like [CLI11](https://github.com/CLIUtils/CLI11). It's a header-only, which means it can up compile-times, but it's alright for smaller CLIs. I think the issue I've had with many of the parsers is that they can all become a little bit sucky once subcommands and customizations come into play: I couldn't easily refactor the code to have the exact same user-facing behavior as docopt (I had to change tests, and part of this little refactoring session was to see if I could refactor it with no test changes).

I could try and refactor the CLI parsing into something more reusable as I tinker with this branch, but I guess that'd only be a value-add if I also investigated other CLI interfaces (there's a bunch of deprecated ones, but some still seem to be live). 

As for splitting the commands into separate files, I'll go change that on the next go-round. it was just an opinion edit: I personally prefer larger source files with internal methods that are `static`or in an anomynous `namespace` because it makes it easier to refactor them without having to worry about potential impacts on other dependants. The only change from `master` I would make, though, is to have the (separate) implementations in `.cpp` files, rather than headers.

adamkewley:
I've moved the command-line parsing and subcommand code into their own files just to see how that looks. This still needs a bit more work (to ensure the help printing is fine etc.) - I'm just playing with the structure at the moment.

adamkewley:
Should I go and finalize this work or kill it off? What's the verdict on dropping the dependency?

chrisdembia:
I'm still open to this but I would like even less "manual" command-line parsing code amongst the logic of the subcommands. The hope is to avoid any friction / bug-prone parsing that might deter future developers from adding new subcommands. This means, for example, avoiding checks like `if (argc > 1)` to determine if a given argument were provided.

adamkewley:
On the fence about that kind of programming, tbh, @chrisdembia . I mean, I sympathise with  exactly what you mean--because I've been through more CLI parsers and languages than I'm willing to admit to--and I'm probably going to regret writing this, but what exactly are the OpenSim devs trying to protect themselves against?

```c++
// some primitive `while` loop that skips over named args

if (argc == 0) {
    // insufficient unnamed args
}
```

vs.

```c++
auto parser = some_library_the_dev_needs_to_study();
parser.add_arg(...);
auto some_return_value_you_need_to_study = parser.parse(argc, argv);

if (argc == 0) {
    // whoops: this particular library takes argc and argv by reference
}
```

I think I used to prefer the latter--it "protected" me from having to understand basic CLI parsing--but, over time (and, granted, this may be a fleeting feeling I have) I'm gradually beginning to lean towards the former because even a relatively inexperienced dev that only knows the basic C++ control structures can probably get through the raw source code, guess what's going on, and modify it.

Not bothered either way on this particular PR: it was a throwaway. However, I'm pointing this out as a general thing to keep in mind, especially when considering the architecture of other parts of OpenSim, which might require a newer dev (e.g. me) to study a *very* large hierarchy of libraries and abstractions to understand what's actually going on at runtime.

adamkewley:
Dropping this - clearly, there is no interest in reducing dependency sprawl and I am too busy to kick this into shape

chrisdembia:
>  there is no interest in reducing dependency sprawl

I don't think this is generally true.

adamkewley:
Look into the trend of dependency usage over time in opensim, plus any
scheduled changes to opensim, and square that with my observation.

Clearly, there's close to no interest in reducing dependency sprawl. If
there was, it'd be actually worked on. We'd be refactoring out unused and
redundant classes, removing libs where possible, etc. This isn't a problem
right now--theres other fish to fry--but it sounds like you think the
opposite is true and that things are being reduced down. Fine. I'm wrong.
Please educate me. Where is that happening?

chrisdembia:
@adamkewley this is what we are going for https://xkcd.com/2347/ 😉

PR #2747, opened 2020-04-23
Remove previous logging system

chrisdembia:
Part of #36 

### Brief summary of changes

Removed LogManager, LogCallback, and related classes in the Java bindings (previously replaced with JavaLogSink.java that is part of the java bindings).

### Testing I've completed

Ran CTests locally.

Is this the right time to remove the old logging classes? 

### CHANGELOG.md (choose one)

- no need to update because...part of a feature

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2747)


aymanhab:
Thanks @chrisdembia I will review the code but the critical issue is whether the GUI behavior has changed or can be reproduced, in particular messages that used to show in the messages window.

chrisdembia:
I understand; that is why I mentioned the JavaLogSink class; see https://github.com/opensim-org/opensim-core/blob/feature_logging/Bindings/Java/JavaLogSink.java.

aymanhab:
Thanks for the merge @chrisdembia I'll pull, build, test and review.

chrisdembia:
Thanks @aymanhab

aymanhab:
@chrisdembia ci builds are failing because we don't superbuild spdlog and don't pass in configuration to opensim-core, which of this is preferred? Thanks

chrisdembia:
Thanks @aymanhab . I realize now that I introduced a superbuild "DEFAULT" setting that was set to OFF (by default) for spdlog. No need to update the CI scripts.

aymanhab:
@chrisdembia I'm trying to get this branch to work with the GUI and I'm getting messages like

[2020-04-24 12:34:39.702] [info] Loaded model Wrist_Model from file WristModel\wrist.osim

in stdout, but they are not piped into the message window. I'm likely missing something in the wire-up. 
I added a subclass of JavaLogSink to the Logger, but the method `public void sinkImpl(final String msg)` in my subclass is never invoked, is that expected? Thank you

chrisdembia:
Bah! No that's not expected. You used `addSink`?

aymanhab:
Yes I did addSink(), is there a way to list the sinks?

chrisdembia:
> Yes I did addSink(), is there a way to list the sinks?

Not in Java.

@aymanhab I was confident that this would work in the GUI because of this test I added to opensim-core:

https://github.com/opensim-org/opensim-core/blob/fa74b53aa000c6b903fe3233006ebca5439c0d15/Bindings/Java/tests/TestBasics.java#L104-L125

Can you identify an issue with this test?

aymanhab:
The test looks fine, I'm building with debug info to step into the calls and callbacks when made from the GUI. Will keep you posted.

chrisdembia:
Okay cool. stdcout should not contain the date/time; that should only be in opensim.log. Was the message you showed earlier from the log file?

aymanhab:
The line was from the ide/debugger console which is typically what you see when using "println" in Java or std::cout in C++. There were no output files (ala out.log or err.log) should we expect one?

chrisdembia:
There should be a file `opensim.log`.

aymanhab:
Found the file, thanks @chrisdembia On related note, the test was not enabled (missing call) in TestBasics.java I enabled it and the test failed. Will keep digging

chrisdembia:
🤦‍♂️ sorry! I'm happy to look into it too.

aymanhab:
No worries, I'm still investigating, will let you know if I need help. Thank you

aymanhab:
@chrisdembia Wanted to get your thoughts on whether the last commit  is the right behavior before we merge, as I'm starting to work on the simulation library. In particular, the fact that messages that used to show in messages window would be log_cout instead of log_info. Alternatively,  if you can change the test case to pass with log_info, or suggest log-level then let me know. Thank you

chrisdembia:
Using `log_cout()` here is not correct. But it's useful to know the test passes with this change! I don't have any ideas for a fix yet. Maybe I need to investigate further.

chrisdembia:
@aymanhab the test passes for me on Mac when using `log_info()`. This may be a platform-dependent issue.

aymanhab:
@chrisdembia I reverted to the branch and still see the failure locally even though ci passes. Let's merge into the feature_logging branch for while I continue investigating and working on the simulation library. At least we have hope now that it could work from the GUI.

chrisdembia:
Weird!

PR #2750, opened 2020-04-25
Fixed a transpose

pariterre:
I cannot locally test for Python, but it seems that the difference with BTK was a transpose in the Matrix. Here is the fix for that!

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2750)


pariterre:
@aymanhab 
I am not sure what fails, but it seems that the tests pass :)

aymanhab:
Thanks @pariterre much appreciated. The failures on ci are unrelated and due to using a fork rather than a branch so will merge. Thank you

Issue #2723, opened 2020-04-09
Visualize mocap and imu data in API


aymanhab:
There's no utility as of now to allow users to inspect mocap data or IMU data in the API, through Matlab or Python making debugging quite hard or next to impossible.

aseth1:
@aymanhab this would be great to add to 4.2. We've had a previewer for marker data [here](https://github.com/opensim-org/opensim-core/blob/2fc6dc1d23734f3b5f8d58c3b3833871e756c675/OpenSim/Sandbox/futureMarkerPreview.cpp) and that has been integrated in OpenSense with `ExperimentalMarker` and was part of [`IMUInverseKinematicsTool`](https://github.com/opensim-org/opensim-core/blob/1e22b11c22feb6c29546f1b0148f4af331a95a41/OpenSim/Tools/IMUInverseKinematicsTool.cpp#L43-L92).  There is an analog to the Marker for IMU called [ExperimentalFrame ](https://github.com/opensim-org/opensim-core/search?q=ExperimentalFrame&unscoped_q=ExperimentalFrame )but unlike `ExperimentalMarker`s, visualizing them without some sort of way (Model) to position them in space makes it difficult to interpret the coincident rotating frames.

The main idea is that previewing is simply a case of `Component` visualization that provides convenience methods to add `ExperimentalMarker`s (Sensors, etc...) to an empty `Model` and they are visualized as the model is "simulated". Ideally we would have an `Environment`/`World`  top-level `Component`  where we can add Models as well as Experimental Markers, Sensors, External Forces, ...  to be simulated/visualized together. A "Data Previewer" in this case simply creates and adds the corresponding components to the Environment from Experimental Data (as Table columns) and wires them up to the data table.

PR #2755, opened 2020-04-29
Visualize orientations, add entry point utility

aymanhab:
Fixes issue #2717, #2723 

### Brief summary of changes
Added command line utility visualize that allow users to exercise visualization of Models, Motions and Data.

### Testing I've completed
visualize Model, and Model with Motion, Orientations all work as expected. MarkerData (ported code) in progress.

Name of utility, command-line help, visual feedback.

### CHANGELOG.md (choose one)
Will update once functionality is agreed upon.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2755)


aymanhab:
Thanks for the feedback @chrisdembia and @carmichaelong Ready for another review

aymanhab:
ready for review

aymanhab:
@johanping is interested in this functionality

aymanhab:
Thanks @carmichaelong and @chrisdembia for the review. Summary of issues to finish:
1. Keep separate utility or merge into opensim-cmd (thoughts @carmichaelong)? Would be nice to leverage some common utilities and less pollution of namespace, at the expense of mixing rather unrelated/loosely-related functionality and potential maintenance issues since not a self contained functionality.
2. What visualization controls would be useful (e.g. real-time, single-step, trim-time, show meshes?)
3. Any other outstanding issues?

aymanhab:
@carmichaelong regarding "press any character to exit", what's the desired behavior? Ideally any character in visualizer window would close it but none of the simbodyVisualizer methods advertised to shutdown/setShutdownWhenDestructed actually closes the visualizer window.

carmichaelong:
It's a good question. I think for me, the more important thing is to make the message correct for whatever the behavior is. I know for visualization while using MATLAB, we usually tell users to hit Escape twice when in the visualizer window (though this is especially important so that MATLAB isn't confused, which is less of an issue with command line).

Is it possible for hitting ESC twice AND for just closing the visualizer window to both be good ways to return the command prompt / terminal to a usable state?

aymanhab:
@carmichaelong The two ESC are actually used as follows:1st supposed to close the window but actually it doesn't, it just disconnects it, then the 2nd is the equivalent of widow close. Sure we can do that.

johanping:
Hi all - great work on this! 

My only comments are as follows:
- this has been helpful for visualizing the "raw" IMU and markerplate orientations so that I can see if they "make sense" to begin with
- I try to pause with spacebar (in the visualizer) and it seems like it only pauses for 3 seconds then advances a frame, not a huge deal but don't know if it needs to do this?
- I also think it could be useful to have a way to rotate these data the same way you can rotate marker data with "preview experimental data" in the GUI, (I think we talked about this) but as I see it more I'm definitely finding that it would be beneficial

Thanks everyone! :)

aymanhab:
@carmichaelong I updated the message to indicate Esc. to exit, and fixed behavior of pause to not advance time. Ready for re-review.

aymanhab:
@johanping I have implemented the rotation of data option on command line but it is implemented as rotating the data followed by the existing visualization. Would it make sense to build this in as opposed to having the user pre-rotate the data? In the GUI context it definitely helps but I'm undecided on this use-case. Thoughts?

carmichaelong:
@aymanhab Just tried it again and it's headed in the right direction. Some issues I saw this time:
- The `-VM` option has a different way of ending visualization (hitting any button)
- The `-VS` option does not give any instructions on how to end the visualization, or how to pause.
- The `-VD` option gives instructions on how to end visualization, but not on how to pause.
- Unpausing with `-VD` takes a second or two before it works, while with `-VS` it is much snappier.

Separately, I don't think I feel confident in knowing the true pros/cons of merging with `opensim-cmd`, so take my opinion with a grain of salt. I feel like it could be worth trying to merge to get a feel for the overhead, and revert if it's clearly too much upkeep.

aymanhab:
@carmichaelong thanks for the feedback. The VM and VS options were ported as is from earlier incarnations so each has its own ad-hoc mechanism, I can fix VM since I wrote it, VS option (aka Moco::visualize) I didn't change because it is a refactor of what's in use by @chrisdembia and @nickbianco 
I'd settle on functionality when all the behavior and command line options are self contained in one utility before I attempt merging with opensim-cmd.
I'll look into the snappiness issue that you brought up, likely related to the sleep time set when pause is issued.

johanping:
Awesome! I think that it would make sense to build it in if that’s not too much trouble - otherwise, it could be presented easily as just having the user visualize, check orientations, then close the visualizer and set rotations? I don’t think it’s worth doing if a huge hassle, but it would be more similar to the rotating of marker data in preview experimental data if you could rotate it after visualizing

aymanhab:
@carmichaelong thanks for the feedback, I addressed your comments except for -VS option which has its own controls that were tuned by @chrisdembia and @nickbianco 
@johanping I added a command line option to transform the orientation data, I didn't document it in usage since it is experimental for now, I can give you details separately.

carmichaelong:
Thanks @aymanhab. I tested that the visualizer changes did work.

Just one more comment for some of the command prompt feedback to a user with incorrect inputs. Currently, if a necessary flag is not even passed, the error message may be confusing to a user:

For `-VM`:
```
opensim-visualize.exe -VM std_calibrated_subject07.osim

Loaded library osimActuators
Object: Cannot construct from empty filename. No filename specified.
        Thrown at Object.cpp:100 in Object().
```

For `-VS`:
```
opensim-visualize.exe -VS -M std_calibrated_subject07.osim

IO.OpenInputFile(const string&,openmode mode): failed to open
Storage: Failed to open file ''. Verify that the file exists at the specified location.
        Thrown at Storage.cpp:181 in Storage().
```

For `-VD`:
```
opensim-visualize.exe -VD -M std_calibrated_subject07.osim

Input file '' does not have an extension.
```

Seems like it might be good for `-VM` to check if `-M` exists (and `-VS` to check `-S` and  `-VD` to check `-D`), or else the user might be confused since they didn't actually pass in a file to start with.

chrisdembia:
> I'd settle on functionality when all the behavior and command line options are self contained in one utility before I attempt merging with opensim-cmd.

@aymanhab I'm willing to attempt the merge now, if you're interested.

chrisdembia:
I would probably feel better about a new command-line tool if we merged `opensense` into `opensim-cmd`, which you seemed open to in the past.

aymanhab:
Thanks for testing, and for the offer to help.

If you start a PR off this branch it would be great so we can size what's
involved and we can work on it together.

chrisdembia:
Neat! 
image

johanping:
This is awesome - you guys rock!

chrisdembia:
@aymanhab , the playback speed for `showOrientationData()` is slower than real-time. I think the playback speed should be real-time.

aymanhab:
One issue I ran into with using realtime in the visualizer is that it assumed time starts at 0.0 which caused unexpected behaviors as it is not always the case for experimental data. Maybe there are ways to workaround that. Primarily this is intended for debugging so realtime is not an issue AFAICT.

chrisdembia:
Ah interesting.

We should at least document what real-time factor is being used, if any.

aymanhab:
Folded into PR #2777 that was just merged.

PR #2769, opened 2020-05-13
logging for applications

carmichaelong:
Part of #36

### Brief summary of changes
Use spdlog in Applications folder.

### Testing I've completed
Ran some of the executables without arguments, saw a different behavior in the help section with the `[warning]` tag:
```
[2020-05-13 10:25:30.305] [warning]
    THIS EXECUTABLE IS DEPRECATED AND WILL BE REMOVED IN A FUTURE RELEASE.
```
The `opensim.log` file was empty.

- What other testing would be good to do?
- There is heavy usage of `PrintUsage()` and `Object::PrintPropertyInfo()`. Should I deal with those too? If so, in a new PR?

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2769)

### DIFF
--- a/Applications/Analyze/analyze.cpp
+++ b/Applications/Analyze/analyze.cpp
@@ -58,7 +58,7 @@ int main(int argc,char **argv)
       analyze -S SetupFileName -> opensim-cmd run-tool SetupFileName
       analyze -PS              -> opensim-cmd print-xml analyze
     )";
-    std::cout << deprecationNotice << std::endl;
+    log_warn(deprecationNotice);
 
     // PARSE COMMAND LINE
     int i;
@@ -88,7 +88,8 @@ int main(int argc,char **argv)
             Object::setSerializeAllDefaults(true);
             tool->print("default_Setup_Analyze.xml");
             Object::setSerializeAllDefaults(false);
-            cout << "Created file default_Setup_Analyze.xml with default setup" << endl;
+            log_info("Created file default_Setup_Analyze.xml with default "
+                     "setup");
             return(0);
 
         // IDENTIFY SETUP FILE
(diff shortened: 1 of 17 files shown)


aymanhab:
@carmichaelong any reason not to merge this PR to feature_logging?

carmichaelong:
I needed to make some changes to the tests after approval. Looks like they all pass now, but would be good if someone double checks the changes to the tests before merging.

carmichaelong:
___
*[Applications/opensim-cmd/test/testCommandLineInterface.cpp, line 118 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2769#-M7dAPtH6HigIivxLSFv:-M7dM8H78Kin1s_dVdOc:bj8i1fu) ([raw file](https://github.com/opensim-org/opensim-core/blob/74730534a4295ecf644d94e43190b22afa3e9c97/Applications/opensim-cmd/test/testCommandLineInterface.cpp#L118)):*

I'd change the parameter name to substring to avoid confusion with the member, same for StartsWith class, though not user facing. Maybe more robust to have a method that removes logging prefix (within []) and leave the checks as is if not much work.
</blockquote></details>

I started off on that route, but some checks are multiple lines, so can't remove the start cleanly each time. Another route could be to just convert lots of this to regex?

aymanhab:
That's fine @carmichaelong if it's too messy, feel free to merge as is once other comments are addressed.

carmichaelong:
Thanks @aymanhab and @chrisdembia for feedback. I will definitely not add more `regex`'s then.

I think I'll leave the `ContainsSubstring` as is since it's a bit easier than removing `[]` sections, and I think would end up to a similar check.

@aymanhab I just made a push to make the parameters and member variables different in the two helper functions you mentioned. Let me know if this addresses your comment, or if I misinterpreted it I'm happy to make another change.

aymanhab:
Thanks @carmichaelong LGTM.

Issue #2334, opened 2018-10-26
Remove the option for using kinematics for ExternalLoads


aseth1:
**What does the "Kinematics for external loads" option actually do and why should anyone care?**

When specifying external loads the force(s) and point(s) of application are typically specified in the lab (ground) reference frame, but applied to points on the model (e.g. the foot). The option provides a preprocessing step that re-expresses the force and point (originally in ground) into the *applied-to* body (`Frame`)  frame of reference. In the case that the motion being analyzed (e.g. `InverseDynamicsTool`, `StaticOptimization` and any other `Analysis` of the `AnalyzeTool`  involving applied `ExternalLoads`) the answers will be identical **IF** the kinematics being analyzed and the kinematics for external loads (for transforming the force into the body) are identical.  In the cases that kinematics are not identical (choosing wrong file, down sampling, filtering, etc...) the transformation introduces inconsistencies. 

So why do it? The ONLY reason to register forces using input kinematics instead of just applying them in the lab frame is to "mitigate or reduce" drift of the RRA and particular CMC solutions which are tracking those kinematics.  If the model kinematics are deviating from the input kinematics the conjecture is that large forces (e.g at the foot) applied in ground reference frame will no longer be on the model foot and it will hasten the divergence of the tracking solution. While the first part is true, if tracking errors are large the foot can be out of position and the applied force (in ground) are way off the foot. The question is whether registering the forces to the foot improves the answers, and that is definitely, NO! If the solution is tracking, the differences are insignificant. If CMC is diverging, then not registering the forces to the foot typically diverges a few CMC time-windows earlier than with registering, but in both cases tracking fails. You can buy a few more time steps before solution diverge irrecoverably by using this transformation "trick". It is not, however, a better answer. If your RRA/CMC tracking is diverging and you are squeezing out a few more time steps, you still have a poor solution and warrants further refinement by increasing tracking weights or having reserve actuators or accepting larger residuals.

**Pros and Cons to removing this option:**

Some of the reasons to remove it:
1. confusing for users to comprehend 
2. slows down tools like ID, SO, etc.. because of loading and filtering motion data twice
3. more often than not produces inconsistent results
4. is error prone
5. intended for one obscure use-case
6. because of recent improvements to `ExternalLoads` handling (#2332) making the change now eliminates the complexity/additional implementation required to preserve it

and reasons to keep it:
1. preserving users existing workflows
2. prevent other code/GUI changes close to release

[...]

PR #2770, opened 2020-05-13
Remove kinematics_for_external_loads property 

aymanhab:
Remove properties from from ExternalLoads object and remove usage in AbstractTool and DynamicsTool. 

Fixes issue https://github.com/opensim-org/opensim-gui/issues/348

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2770)


chrisdembia:
https://github.com/opensim-org/opensim-core/issues/2334

aymanhab:
Thanks @chrisdembia This PR predates the change to logging so log_ functions were not available then. Will update accordingly. Generally we don't keep track if a warning is issued while upgrading and i don't think we will because the upgrade is per object not per file/document.

chrisdembia:
@aymanhab I don't know why this keeps happening but the diff contains changes that are already in master.

aymanhab:
@chrisdembia the two warnings are on two different control paths  of if/else so they can't both be true AFAIK 😄 

I'll look into the diffs issue.

aymanhab:
Running IAA in 3.3 without kinematics for external loads gives identical results to branch with the fix as expected. I can post plots but visibly the totals are indistinguishable which is rather expected if we didn't introduce a bug elsewhere.

aymanhab:
Based on feedback and diffs above this seems ready to go (separate GUI PR will remove it from the application), main issue that I'd like feedback on is whether we upgrade file version number (corresponding to XML schema change). It's included here already but has the side effect that new files/models saved moving forward will not be backward compatible with 4.0 or 4.1, this is needed so that saved external loads files don't have the removed properties but comes with this  inconvenience. Thoughts @carmichaelong @jenhicks @nickbianco 
We can skip this and rely on the warning each time the file is opened and on documentation instead.

aymanhab:
@jenhicks @carmichaelong @aseth1 Any feedback on this so we can move on? Thank you

jenhicks:
My initial inclination is to keep backwards compatibility, but I am open to other thoughts.

aymanhab:
Thanks @jenhicks I tend to agree, particularly that ExternalLoads are saved to separate files typically so losing backward compatibility for model files feels like an overreach. I will undo the version change unless I hear otherwise from reviewers.

aymanhab:
@carmichaelong one more look and will merge if no objections when you have a chance, thanks

aymanhab:
Approved in dev meeting by @carmichaelong, Merging.

Issue #2771, opened 2020-05-14
Failed tests and ModuleNotFoundError


annshorn:
Hi, there is another problem that I faced. I went through rules and on the step:
`ctest -j8`

I got this:

```
95% tests passed, 5 tests failed out of 106

Total Test time (real) = 252.03 sec

The following tests FAILED:
	 76 - testStaticOptimization (Failed)
	 94 - python_tests (SEGFAULT)
	 95 - python_examples (SEGFAULT)
	102 - Java_TestTables (Failed)
	106 - Java_TestXsensDataReader (Failed)
Errors while running CTest
```

My main goal is the opportunity to import OpenSim into Jupiter notebook. And after doing installation still can't import (ModuleNotFoundError: No module named 'opensim')

Is it because of not all tests passes?

aymanhab:
test failures have no downstream effect, so unrelated but may indicate problems with your build/environment. Likely you didn't build C3D support, make sure you choose value other than None for OPENSIM_C3D_PARSER is using latest master branch, or turn on WITH_BTK if using 4.1 code base.  This likely will take care of the java test failures. Python failures likely due to using unsupported python, our continuous integration builds and tests against standard python 3.7 from python.org. If successful in importing into Jupiter notebook please open a PR to document the steps. Thank you

annshorn:
In the "CMakeCache.txt" in opensim_build changed 

```
//Compile OpenSim with a C3D parser ? ezc3d or BTK provide C3D
// reading.
OPENSIM_C3D_PARSER:STRING=None
```

to

`OPENSIM_C3D_PARSER:STRING=WITH_BTK`

It gives:
```
96% tests passed, 4 tests failed out of 106

Total Test time (real) = 252.13 sec
The following tests FAILED:
	 94 - python_tests (SEGFAULT)
	 95 - python_examples (SEGFAULT)
	102 - Java_TestTables (Failed)
	106 - Java_TestXsensDataReader (Failed)
Errors while running CTest
```

About python actually I don't understand: I use python 3.7 from anaconda - is it not supported?

chrisdembia:
Do you mean `OPENSIM_C3D_PARSER:STRING=BTK` (no `WITH_`)?

Python 3.7 from Anaconda is supported if you are building from source (as you are). The issue may be that you are trying to import the python libraries using a version of python different from the one that CMake found and used to build OpenSim's python library.

aymanhab:
[...]

PR #2773, opened 2020-05-19
Fix linker flags for building Python package for conda.

chrisdembia:
Fixes issue #2771 

### Brief summary of changes

This PR brings over a fix I saw elsewhere on GitHub for getting the python package to work with Anaconda's python on Mac.

The diff in this PR provides more background.

### Testing I've completed

On Mac, I built the Python package with Anaconda and ran the python CTest tests.

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2773)

### DIFF
--- a/Bindings/Python/CMakeLists.txt
+++ b/Bindings/Python/CMakeLists.txt
@@ -133,8 +133,28 @@ macro(OpenSimAddPythonModule)
             "${PYTHON_INCLUDE_PATH}"
             "${PYTHON_NUMPY_INCLUDE_DIR}")
     
-    target_link_libraries(${_libname}
-        osimTools osimExampleComponents ${PYTHON_LIBRARIES})
+    # On macOS, the python interpreter might link to the python library 
+    # statically, in which case we do not want to link dynamically to the
+    # python library. This situation occurs with Anaconda's python.
+    # In this situation, we must add a linker flag to mark all undefined
+    # symbols as having to be looked up at runtime, and we ignore
+    # PYTHON_LIBRARIES.
+    # https://github.com/opensim-org/opensim-core/issues/2771
+    # https://stackoverflow.com/questions/25421479/clang-and-undefined-symbols-when-building-a-library
+    # https://github.com/lisitsyn/shogun/commit/961f6109c2a8abab4941bcdde1a19dfe70c440f1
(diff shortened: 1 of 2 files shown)


aymanhab:
Nice investigative work @chrisdembia 👍 The question in my mind is whether it's worth it to complicate the build system and future maintenance to support this specific use-case, when it sounds like many other environment including miniconda work out of the box. Do you know if this is Mac specific, and if Anaconda works out of the box on windows? Thank you

chrisdembia:
> Do you know if this is Mac specific, ...

I think this issue could occur on any UNIXy system where the Python interpreter statically links to the Python library, because the PYTHON_LIBRARIES CMake variable assumes that the Python interpreter *dynamically* links to the Python library. From my experience and from what I've read on GitHub, this occurs with Macs and Anaconda Python, but the fix I implemented does not assume Mac or Anaconda; it attempts to get closer to the root problem.

The fix I implemented is copied from the links I put in the CMakeLists.txt file. It is likely that there are better fixes. The python interpreter itself provides some C/C++ flags that should be used when building C/C++ extensions. I briefly looked to see if we could leverage those flags to replace PYTHON_LIBRARIES and avoid adding a special case, but nothing stood out to me. I could investigate more deeply, but other projects have already chosen this fix.

I think using Anaconda Python on Mac is a substantial use case that we should support because Conda (which uses Anaconda Python) will be the easiest way for us to package the OpenSim Python package. In fact, I ran into this same bug myself when trying to package Moco for Conda over the last few days.

> ...and if Anaconda works out of the box on windows?
I am also attempting to package Moco for Conda for Windows but am running into unrelated issues, so I do not know yet. I do not know yet if Anaconda's Python interpreter links statically or dynamically to the Python library, or if doing so would present a similar issue on Windows. My guess is not.

> (including miniconda?)

I do not know enough about the differences between Miniconda and Anaconda, but my guess is the issue is the same with Miniconda.

> whether it's worth it to complicate the build system and future maintenance

I've tried to give enough background links so that if the issue evolves in the future, it'll be easy to see how other projects are adapting.

aymanhab:
Thanks @chrisdembia for the clarification of the scope. Merging.

Issue #2761, opened 2020-05-05
BodyKinematics file comments are not self-consistent


carmichaelong:
All 3 files from `BodyKinematics` analysis have inconsistencies in the notes:

From `pos` file:

> Positions
> version=1
> nRows=1277
> nColumns=88
> inDegrees=no
> 
> This file contains the kinematics (positions and orientations,
> velocities and angular velocities, or accelerations and angular accelerations)
> of the centers of mass of the body segments in model Jumper.
> 
> Body segment orientations are described using body-fixed X-Y-Z Euler angles.
> 
> Angular velocities and accelerations are given about the body-local axes.
> 
> Units are S.I. units (seconds, meters, Newtons, ...)
> Angles are in degrees.

`inDegrees=no` and `Angles are in degrees` are not self-consistent.

Something else to check is if #2208 is related or not.

[Jumper_BodyKinematics.zip](https://github.com/opensim-org/opensim-core/files/4583782/Jumper_BodyKinematics.zip)

PR #2775, opened 2020-05-21
Address mismatch in Analyses about degrees vs radians

carmichaelong:
Fixes issue #2761 

### Brief summary of changes
Description for Analyses are set during construction, which often called `getInDegrees()`, however users can then change the flag afterwards (before `Storage` prints out the header)

### Testing I've completed
None (aside from the standard current tests). No current tests seem to exercise the Analyses rigorously, but unsure what to do with that when we are moving away from Analyses?

- Any test that would make you feel confident about this PR?
- The description now does not depend on `getInDegrees()` but does still assume a header will be written. Is this okay for now?

### CHANGELOG.md (choose one)
Not updated, but happy to do so if reviewers think it's big enough to put in.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2775)

### DIFF
--- a/OpenSim/Analyses/Actuation.cpp
+++ b/OpenSim/Analyses/Actuation.cpp
@@ -234,9 +234,9 @@ constructDescription()
     strcat(descrip, "from the model.\n");
 
     strcat(descrip, "\nUnits are S.I. units (second, meters, Newtons, ...)");
-    strcat(descrip, "\nIf header above indicates 'inDegrees=yes' above, ");
-    strcat(descrip, "then angles are in degrees, ");
-    strcat(descrip, "otherwise angles are in radians.");
+    strcat(descrip, "\nIf the header above contains a line with ");
+    strcat(descrip, "'inDegrees', this indicates whether rotational values ");
+    strcat(descrip, "are in degrees (yes) or radians (no).");
     strcat(descrip, "\n\n");
 
     setDescription(descrip);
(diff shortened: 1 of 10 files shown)


chrisdembia:
This LGTM. I don't think any testing is necessary.

In my opinion, I'm not sure the added header describing the meaning of the "inDegrees=yes" is necessary. Some files do not contain angles, so mentioning angles could be confusing. Perhaps adding "(if any)" would help clarify. Feel free to ignore these comments, though.

carmichaelong:
I agree with all the points you bring up. If we keep it, I think I'd add the "(if any)" part. One reason why I might lean towards keeping it is that many (if not all) of the descriptions include `Units are S.I. units (second, meters, Newtons, ...)` which then may be misleading to users who now may expect radians. Thoughts?

chrisdembia:
> many (if not all) of the descriptions include `Units are S.I. units (second, meters, Newtons, ...)` which then may be misleading to users who now may expect radians.

OH! Good point. I agree.

carmichaelong:
I couldn't figure out a simple way to drop in the "(if any)" cleanly, so I rewrote the line all together. I hope this is concise and also is a little more clear. One issue I had with the last version was that the logic implied: if `inDegrees=yes`, then it's degrees, otherwise it's radians. However, that would mean if the header line didn't exist, then it's also definitely radians.

PR #2784, opened 2020-05-29
CMake error if SWIG version is too great.

chrisdembia:
### Brief summary of changes

opensim-core does not yet support SWIG 4, and multiple people have tried building from source using SWIG 4 and run into a Java compiler error. While we figure out a fix, this PR alerts builders that they cannot use SWIG 4.

### Testing I've completed

Ensured that if CMake finds SWIG 4.0.1, then CMake generates this new error.

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2784)


chrisdembia:
Dimitar pointed out that this only affects Java, not Python. So I could add BUILD_JAVA_WRAPPING as part of the condition that generates the error.

aymanhab:
Thanks for fixing this @chrisdembia Yes I'd make it Java specific, and change the text to be less severe (too great -> not supported yet ;) 👍

chrisdembia:
Ready for review.

Issue #2780, opened 2020-05-26
Enable specifying weights for Orientation tracking in IK


aymanhab:
As of now the orientation weights are available internally but are not exposed to users in XML. Needed for ongoing research.

PR #2786, opened 2020-06-01
Expose ik orientation weights

aymanhab:
Fixes issue #2780 
### Brief summary of changes
Exposed existing class OrientationWeight, and include a set of OrientationWeights in IMUInvereseKinematicsTool. Did cleanup to remove old style properties removing two boilerplate cpp files along the way.

### Testing I've completed
Ran test suite so far, looking into beefing up testopensense to verify the wiring, testInverseKinematicsSolver already tests usage of the weights at the solver level.

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2786)


aymanhab:
@carmichaelong any thoughts or feedback on the XML changes or what would be a good test case that is not a regression against whatever results we get now? Thanks

aymanhab:
If we can get this moving to give Johanna a build by monday, would be awesome. I can also make a special GUI build if needed. Test case can be beefed up but would be good to have some more testing on the field.

aymanhab:
@chrisdembia can you take a look as a second reviewer when you have a chance? Thank you

OpenSim/Tools/IMUInverseKinematicsTool.h
```diff
@@ -83,7 +83,8 @@ class OSIMTOOLS_API IMUInverseKinematicsTool
             " Default to (0, 0, 0).");
     OpenSim_DECLARE_PROPERTY(orientation_weights, OrientationWeightSet,
             "Set of orientation weights identified by orientation name with "
-            "weight being a positive scalar, default to 1 to track all IMU data.");
+            "weight being a positive scalar, default to 1 to track all IMU "
+            "orientations.");
```
chrisdembia, line 87: I think this could be a little clearer:

```
... If not provided, all IMU orientations are tracked with weight 1.0.
```

aymanhab:
Thanks @chrisdembia 👍

Issue #2789, opened 2020-06-04
Millard2012EquilibriumMuscle::extendFinalizeFromProperties(): construction of ForceVelocityInverseCurve


mjhmilla:
This bug comes thanks to a post by Saurabh Kothari to the forum here:

https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=11970&p=34029&start=0&view=

Summary:
1. In Millard2012EquilibriumMuscle.cpp in extendFinalizeFromProperties the parameters conSlopeAtVmax and eccSlopeAtVmax are set to be 0.1 if they happen to be less than 0.1.

2. However the parameters eccSlopeNearVmax and conSlopeNearVmax are not updated.

3. The bug happens when eccSlopeNearVmax <= 0.1 or concSlopeNearVmax <= 0.1. In this case eccSlopeNearVmax < eccSlopeAtVmax and/or concSlopeNearVmax < concSlopeAtVmax. This is not allowed because the resulting Bezier curve will not be suitable for this section of a force velocity curve.

Suggested fix:

If there are good reasons for keeping the if statement on lines 138-140 then also update eccSlopeNearVmax (and concSlopeNearVmax) such that it meets the conditions imposed by the ForceVelocityInverseCurve constructor. I can imagine that there are good reasons: such as a flood of users who pass in invalid parameters and don't know how to fix the resulting error. 

If the fix turns out to be quick and easy I'll submit something today. Here's hoping.

PR #2790, opened 2020-06-04
Fix for issue #2789 which comes from a SimTK forum post from Saurabh …

mjhmilla:
…Kothari on May 22 2020. Independently testActuators was failing in debug mode because a pendulum on a joint was being simulated with a zero inertia: this test passes with a non-zero inertia.

Fixes issue #2789 

### Brief summary of changes

Error description:

In Millard2012EquilibriumMuscle.cpp in extendFinalizeFromProperties the parameters conSlopeAtVmax and eccSlopeAtVmax are set to have a lower bound of 0.1. However the parameters eccSlopeNearVmax and conSlopeNearVmax are not updated and these parameters need to be greater than eccSlopeAtVmax and conSlopeAtVmax respectively. The bug happens when eccSlopeNearVmax <= 0.1 or concSlopeNearVmax <= 0.1: after conSlopeAtVmax and eccSlopeAtVMax are updated we end up with: eccSlopeNearVmax < eccSlopeAtVmax and/or concSlopeNearVmax < concSlopeAtVmax. This is not allowed because the resulting Bezier curve will not be suitable for this section of a force velocity curve.

Fix description: 

The 'if' statement near line 138 of extendFinalizeFromProperties in Millard2012EquilibriumMuscle.cpp has been updated:

The lower bounds used for eccSlopeAtVmax for the ForceVelocityInverseCurve have been changed from 0.1 to be instead 0.5*eccSlopeNearVmax. This change has also been applied to conSlopeAtVmax. This change: 

1. Ensures that the ensuing call to ForceVelocityInverseCurve succeeds. 
2. That eccSlopeAtVmax (and conSlopeAtVmax) is non-zero so that the inverse curve can be constructed.
3. That the curve between eccSlopeNearVmax and concSlopeNearVmax is unchanged so that the ForceVelocityCurveInverse really is the inverse of the ForceVelocityCurve between this range.

### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2790)


carmichaelong (commented):
Thanks, @mjhmilla! I just went through the code and some of the context from the issue and forum post. Looks like a nice surgical fix. Some notes below:
- Would be great to add a note about this change in `CHANGELOG.md`, especially since this might slightly affect numbers between in the "near" regions.
- You made a good point in the forum post that this if statement has been making a "secret" change. Perhaps we can use the new logging functionality (I'd suggest `log_info()`) to notify a change is being made.
- It looks like there's a merge conflict (I'm guessing because the test case updated here has the change in master already). I'm happy to resolve the merge conflict if that's helpful for you.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2790#-:-MHxQZw6D-ImAYWws48H:b-1b311h)* status: 0 of 2 files reviewed, all discussions resolved

carmichaelong (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2790#-:-MK3MZOj8Shyc4fc2hFi:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved

PR #2796, opened 2020-06-09
Fix SWIG warning when building Moco bindings.

chrisdembia:
### Brief summary of changes

This PR fixes a SWIG warning I saw when building Moco's python bindings:

```
.../opensim-moco/opensim-core/OpenSim/Common/LogSink.h:54: Warning 520: Derived class 'StringLogSink' of 'OpenSim::LogSink' is not similarly marked as a smart pointer.
```

### Testing I've completed

Built Moco's bindings with this fix, and the warning went away.

### CHANGELOG.md (choose one)

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2796)

### DIFF
--- a/Bindings/common.i
+++ b/Bindings/common.i
@@ -12,6 +12,7 @@
 %include <OpenSim/Common/Exception.h>
 
 %shared_ptr(OpenSim::LogSink);
+%shared_ptr(OpenSim::StringLogSink);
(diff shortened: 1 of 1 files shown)


chrisdembia:
I think the source code should avoid `#ifndef SWIG` as much as possible.

Thanks @aymanhab

PR #2797, opened 2020-06-09
Handle situation where logfile can't be created.

aymanhab:
Fixes issue #0

### Brief summary of changes
Handle the case of logfile that can't be opened.

### Testing I've completed
None
What should be the behavior on failure? Before this PR, GUI fails to launch, we can have GUI add its own sink in a different folder that GUI decides on (e.g. Resources folder which we know is writable, other folder common to Netbeans, /AppData, ... or under home directory though I don't like sprinkling files all over the file system)
### CHANGELOG.md (choose one)

- updated...

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2797)

### DIFF
--- a/OpenSim/Common/Logger.cpp
+++ b/OpenSim/Common/Logger.cpp
@@ -145,7 +145,7 @@ bool Logger::shouldLog(Level level) {
 
 void Logger::addFileSink(const std::string& filepath) {
     if (m_filesink) {
-        warn("Already logging to file '{}'; file sink not added. Call "
+        warn("Already logging to file '{}'; log file not added. Call "
              "removeFileSink() first.", m_filesink->filename());
         return;
     }
--- a/OpenSim/Common/Logger.cpp
+++ b/OpenSim/Common/Logger.cpp
@@ -149,19 +149,18 @@ void Logger::addFileSink(const std::string& filepath) {
              "removeFileSink() first.", m_filesink->filename());
         return;
     }
-    // verify if file can be opened at the specified path
-    std::ifstream ifs(filepath);
-
-    if (ifs.is_open()) {
-        ifs.close();
-    } else {
-        // show message:
-         warn("Can't open file '{}' for writing. Log file will not be created. "
-             "Check that you have write permissions to the specified path.", 
-             filepath);
+    // check if file can be opened at the specified path if not return meaningful
+    // warning rather than bubble the exception up.
+    try {
+        m_filesink =
+                std::make_shared<spdlog::sinks::basic_file_sink_mt>(filepath);
+    }
+    catch (...) {
+        warn("Can't open file '{}' for writing. Log file will not be created. "
+             "Check that you have write permissions to the specified path.",
+                filepath);
         return;
     }
-    m_filesink = std::make_shared<spdlog::sinks::basic_file_sink_mt>(filepath);
     addSinkInternal(m_filesink);
 }
 
--- a/OpenSim/Common/Logger.cpp
+++ b/OpenSim/Common/Logger.cpp
@@ -156,7 +156,8 @@ void Logger::addFileSink(const std::string& filepath) {
         ifs.close();
     } else {
         // show message:
-         warn("Can't open file {} for writing. Log file will not be created.", 
+         warn("Can't open file '{}' for writing. Log file will not be created. "
+             "Check that you have write permissions to the specified path.", 
              filepath);
         return;
     }
--- a/OpenSim/Common/Logger.cpp
+++ b/OpenSim/Common/Logger.cpp
@@ -44,7 +44,6 @@ Logger::Logger() {
     m_default_logger->set_pattern("[%l] %v");
     m_cout_logger->set_level(spdlog::level::info);
     m_cout_logger->set_pattern("%v");
-    addFileSink();
     // This ensures log files are updated regularly, instead of only when the
     // program shuts down.
     spdlog::flush_on(spdlog::level::info);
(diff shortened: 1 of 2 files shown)


aymanhab:
@chrisdembia this is a strawman attempt mainly to see if the GUI would launch on Mac with this rather defensive and surgical change. To address in general we could bubble an exception but we still need to answer the question of what to do while keeping the exception handling localized.

chrisdembia:
We could rely on this code for determining where to write the log file: https://github.com/sago007/PlatformFolders/blob/master/sago/platform_folders.cpp

OpenSim/Common/Logger.cpp
```diff
@@ -150,6 +150,16 @@ void Logger::addFileSink(const std::string& filepath) {
...
         return;
     }
+    // verify if file can be opened at the specified path
+    std::ifstream ifs(filepath);
+
+    if (ifs.is_open()) {
+        ifs.close();
+    } else {
+        // show message:
+        std::cout << "Can't open file " << filepath << "Log will not be created." << std::endl;
```
chrisdembia, line 160: Should use the logger: `log_warn(...)`.

We could add a function like `Logger::fileSinkExists()` or `Logger::hasFileSink()` that the GUI could invoke to determine if the GUI should log to some user data directory.

I think it's also fine for there to be no logging file when using the GUI—the content is in the Messages window, anyway.

aymanhab: @chrisdembia The function addFileSink() is invoked from inside the constructor of Logger so it's not clear if any other logging is possible at this point, including log_cout or log_warn 😕 

I think we can make the GUI add its own FileSink as needed but no logging from GUI for now (if the file can't be created) seems less catastrophic than the current behavior.

chrisdembia: That's a good point but `addFileSink()` could be invoked by a user after the constructor has completed, in which case we would want `log_warn()` to be used.

aymanhab:
@chrisdembia since we're using the singleton pattern to create/access the Logger, I moved the function outside the constructor and updated to use the logging syntax.

aymanhab:
> > Reviewed 2 of 2 files at r2.
> > _[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2797)_ status: all files reviewed, 1 unresolved discussion (waiting on @aymanhab)
> 
> _[OpenSim/Common/Logger.cpp, line 159 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2797#-M9ZT6NZ8BCB6CPIF4FF:-M9ZT6NZ8BCB6CPIF4FG:b-yh61di) ([raw file](https://github.com/opensim-org/opensim-core/blob/aad14db3924cf7af55b52d6745eeb0aad073f79a/OpenSim/Common/Logger.cpp#L159)):_
> 
> > ```c++
> >     } else {
> >         // show message:
> >          warn("Can't open file {} for writing. Log will not be created.", 
> > ```
> 
> Consider being a little more specific: `File sink will not be created.`

@chrisdembia I think from a user perspective "Sink" makes no sense while log does. Thoughts?

aymanhab:
> We could rely on this code for determining where to write the log file: https://github.com/sago007/PlatformFolders/blob/master/sago/platform_folders.cpp

Thanks fr finding this @chrisdembia 
There's similar logic in the application 
https://github.com/opensim-org/opensim-gui/blob/5aaa56d214331f9730384133631cea4e45cf6737/Gui/opensim/utils/src/org/opensim/utils/TheApp.java#L189
which seems like better suited for Platform specific code than low level API. 

It's also possible that users run API programs from different folders to keep separate log files and if we force a common location at the API we'd work against them, while an application can set its own convention.

chrisdembia:
> @chrisdembia I think from a user perspective "Sink" makes no sense while log does. Thoughts?

Perhaps "Log file" then. "Log" on its own could mean that messages will not be written to the console.

aymanhab:
Done, thanks @chrisdembia

OpenSim/Common/Logger.cpp
```diff
@@ -150,6 +149,17 @@ void Logger::addFileSink(const std::string& filepath) {
...
         return;
     }
+    // verify if file can be opened at the specified path
+    std::ifstream ifs(filepath);
+
+    if (ifs.is_open()) {
+        ifs.close();
+    } else {
+        // show message:
+         warn("Can't open file {} for writing. Log file will not be created.",
```
chrisdembia, line 159: Can you make the style of this warning consistent with the one above?

aymanhab: Not sure what specifically you want changed, I took a stab at it. As a one liner, feel free to change as well as you see fit. Thank you.

OpenSim/Common/Logger.cpp
```diff
@@ -150,6 +149,18 @@ void Logger::addFileSink(const std::string& filepath) {
              "removeFileSink() first.", m_filesink->filename());
```
chrisdembia, line 149: @aymanhab thanks for adding the quotes below. You mentioned that "file sink" is not clear to users; perhaps the usage of "file sink" in this warning should be changed to "log file" for consistency.

aymanhab:
Thanks @chrisdembia

aymanhab:
@chrisdembia this rather innocent looking change is causing the test case testCommandLineInterface to fail because of failing to open a log file. Any clues about what the test case assumes? Debugging the test case is problematic because it uses "system" methods that the debugger can't step through 😨

chrisdembia:
I'll take a look.

chrisdembia:
@aymanhab I don't have that issue locally; do you encounter this issue locally?

chrisdembia:
@aymanhab I think the issue is multiple processes (tests running in parallel) trying to open the same file. Instead of trying to open the file ourselves, I suggest we catch an exception thrown by `std::make_shared<spdlog::sinks::basic_file_sink_mt>(filepath);`

aymanhab:
@chrisdembia I don't have the issue locally, and I don't understand why this behavior has changed to begin with but I can add a catch as you suggested

aymanhab:
This seems to have taken care of the issue somehow, thanks for the suggestion @chrisdembia

chrisdembia:
I think `ifs.good()` was false because the file was already open, rather than because of a lack of write privileges.

PR #2802, opened 2020-06-14
On Windows with MSVC, avoid generating SWIG files for Python twice

chrisdembia:
### Brief summary of changes

Fix issues with CMake that causes SWIG to run multiple times when generating the Python bindings.

### Testing I've completed

Locally, I ensured that building the `PythonBindings` target only causes SWIG to be run once (for each module).

### CHANGELOG.md (choose one)

### DIFF
--- a/cmake/OpenSimMacros.cmake
+++ b/cmake/OpenSimMacros.cmake
@@ -515,9 +515,9 @@ macro(OpenSimFindSwigFileDependencies OSIMSWIGDEP_RETURNVAL
     # Clean up the output, since it's in the form of a makefile
     # (and we just want a list of file paths).
     if(${_successfully_got_dependencies} EQUAL 0) # return code 0 is success.
-        # '^.*:' matches the first line of the makefile (the output file path).
+        # '^.*cxx:' matches the first line of the makefile (the output file path).
         # '\\\\' matches a single \ (escape for CMake, and escape for regex).
-        string(REGEX REPLACE "(^.*:|\\\\\n)" "" ${OSIMSWIGDEP_RETURNVAL}
+        string(REGEX REPLACE "(^.*cxx:|\\\\\n)" "" ${OSIMSWIGDEP_RETURNVAL}
             ${_dependencies_makefile})
         # Replace spaces with semicolons to create a list of file paths.
         separate_arguments(${OSIMSWIGDEP_RETURNVAL})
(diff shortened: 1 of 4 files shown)


cmake/OpenSimMacros.cmake
```diff
@@ -515,9 +515,9 @@ macro(OpenSimFindSwigFileDependencies OSIMSWIGDEP_RETURNVAL
     # Clean up the output, since it's in the form of a makefile
     # (and we just want a list of file paths).
     if(${_successfully_got_dependencies} EQUAL 0) # return code 0 is success.
-        # '^.*:' matches the first line of the makefile (the output file path).
+        # '^.*cxx:' matches the first line of the makefile (the output file path).
         # '\\\\' matches a single \ (escape for CMake, and escape for regex).
-        string(REGEX REPLACE "(^.*:|\\\\\n)" "" ${OSIMSWIGDEP_RETURNVAL}
```
chrisdembia, line 520: We use SWIG to generate a Makefile listing the file dependencies for running SWIG. The first line is the target file, and contains `cxx:`. The expression containing `:` is used to remove the first line of the Makefile. This is fine on UNIX. On Windows, each subsequent line contains a `:` (e.g., `C:\Users\...`). Adding the `cxx:` fixes how the regex converts the Makefile into a list of file dependencies.

chrisdembia:
This PR reduces the duration of the Windows GitHub Actions CI test by 20 minutes by avoiding rebuilding the Python/Java bindings.

There are a few different issues here:
1. There was a bug in the CMake macro we use to determine the file dependencies for our SWIG interface files.
2. The CMake custom commands/targets for the Python bindings caused Visual Studio to invoke SWIG twice; I fixed this by tweaking how we use custom commands to copy files into the Python package.
3. After I made those two changes, Visual Studio 2017 no longer built the Python bindings twice. But, Visual Studio 2019 still *does* build the bindings twice (if building the `install` target is done separately). So I updated the GitHub Actions script to `install` when building the project the first time. This saves 20 minutes during the `install` that is wasted by rebuilding the Python/Java bindings.

Ready for review.

aymanhab:
@chrisdembia on related question, can we leverage the core build for github action of opensim-core, instead of duplicating the fixes for GUI build? Maybe caching there will make it less relevant but will continue to be duplicated.

chrisdembia:
> @chrisdembia on related question, can we leverage the core build for github action of opensim-core, instead of duplicating the fixes for GUI build? Maybe caching there will make it less relevant but will continue to be duplicated.

Yes, we could create a new "action" in a repo like `github.com/opensim-org/build-opensim-core` that builds opensim-core, and we could employ the action using something like:

```
steps:
     - uses: opensim-org/build-opensim-core@v1
```

I'm not sure this is worthwhile yet.

PR #2804, opened 2020-06-16
Increase Clang/GCC warnings

chrisdembia:
### Brief summary of changes

Enable more warnings with GCC/Clang, and turn warnings into errors.

### Testing I've completed

Ran tests locally.

### CHANGELOG.md (choose one)
 
---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2804)


OpenSim/Common/APDMDataReader.cpp
```diff
@@ -213,36 +213,36 @@ APDMDataReader::extendRead(const std::string& fileName) const {
...
-                throw Exception{ "Expected labels for sensor " + sensorName + 
-                    " were not found." };
             }
+        } else { // first label found but the remaining didn't. Throw
+            throw Exception{"Expected labels for sensor " + sensorName +
+                            " were not found."};
         }
-    return; // not found
     }
 }
```
chrisdembia, line 246: @aymanhab can you look over these changes carefully?

aymanhab: @chrisdembia I will but the PR has 87 files all over the place, can we piecemeal it in a meaningful fashion? I also worry that being on the bleeding edge of a specific compiler/version may result in compilation being less robust for other environments (users or ci with slightly different version/upgrade). We can make sure the code produces no warnings to test but not build it into CMakeLists that are distributed or live on master. Thoughts?

chrisdembia:
Ready for review.

OpenSim/Common/APDMDataReader.cpp
chrisdembia: My comment here was regarding this specific file; the majority of the other changes are much simpler.

Would you be okay with piecemealing through Reviewable? That allows you or other reviewers to review a few files at a time without needing to split the PR.

My preference is to keep the settings I put on this branch: generate more warnings, and present these warnings as errors. This promotes the best hygiene for our codebase, and makes our codebase more robust to the latest compilers.

On the master branch, perhaps we can leave `-Wextra -Wall` but remove `-Werror`.

chrisdembia: Would you be willing to try leaving `-Werror` on the master branch for a while, and remove it if it turns out to cause headaches?

aymanhab: Thanks @chrisdembia, will use Reviewable as you suggested to split the review.
Again I feel warnings as errors is more drastic considering the possible variety of platforms, compilers/upgrades etc. We wouldn't know if it's mere headache or a nightmare for users until release, so definitely release/tag without -Werror but we can leave it for a bit on master to test against ci. We'll never know who else is using it and need to remember to revert the flag before public release.

chrisdembia: Thanks @aymanhab. I've removed `-Werror` for now.

chrisdembia:
Finally, the CI passes. Ready for review.

chrisdembia:
If there are ways to make the review easier, perhaps by splitting files between @carmichaelong, @aymanhab, and @nickbianco, that would be great.

aymanhab:
LGTM, thanks @chrisdembia 
I made a suggestion regarding making -Werror permanent in github action to keep the code clean of the issues you addressed while leaving it off CMakeLists to be lenient on current/future compiler versions at user/client side but this could be done separately if you agree.

chrisdembia:
Thanks for the prompt review @aymanhab.

> I made a suggestion regarding making -Werror permanent in github action to keep the code clean of the issues you addressed while leaving it off CMakeLists to be lenient on current/future compiler versions at user/client side but this could be done separately if you agree.

Yes, `-Werror` is used in GitHub Actions, even on master branch (that's why `continuous_integration.yml` no longer appears in the diff).

chrisdembia:
@aymanhab because you gave a LGTM, I'm merging this.

We can continue to discuss `-Werror` on CI.

PR #2806, opened 2020-06-17
Added _enabledControllers cache to OpenSim::Model

adamkewley:
This PR is to try and optimize certain (mostly, controller-heavy) simulations, such as the FD sim for `ToyDropLanding`. In those simulations, profilers (`valgrind`/`perf`) reported that ~8 % of CPU time was spent calling `getComponentList`.

The reason that `getComponentList` can have this kind of overhead is because its implementation involves `dynamic_cast`ing (downcasting) every element in a container to determine whether the element is a `Controller` or not (the container's elements are `Component`s). The `computeControls` method can be called many times in a simulation, so this iterate-and-downcast can happen *a lot*.

The proposed fix for this is to cache a `std::vector<Controller const&>` of enabled `Controller`s in the `Model` (`Model::_enabledControllers`) just after the model's connection graph is fully established. At this point in `Model`'s lifecycle, `Controller`s and other subcomponents shouldn't change, so the cache should remain valid during the remainder of the simulation. The cache is used by `computeControls` at runtime.

Overall, this caching change makes *some* simulations 5-10 % faster. Usually the simulations containing a bunch of muscles. Not a huge performance improvement--something I'm desperately searching for in OpenSim ;)--but not a terrible performance improvement for a relatively small change (albeit, a change that increases the complexity of `Model`). 

Overall stats from my current measurement technique, which effectively just runs some FD/IK simulations (the suite is work in progress, and should probably be ran on different hardware, etc. temporary repo [here](https://github.com/adamkewley/throwaway-osim-perf)).

```
source activate && ./comparison
```

repeats = 64
|                  Test Name | master [secs] | σ [secs] | branch [secs] | σ [secs] | Speedup |
| -------------------------- | ------------- | -------- | ------------- | -------- | ------- |
|                   Gait2354 |          0.27 |     0.00 |          0.27 |     0.00 |    0.99 |
|             ToyDropLanding |         15.41 |     0.05 |         14.43 |     0.06 |    1.07 |
|   ToyDropLanding_nomuscles |          0.45 |     0.00 |          0.45 |     0.00 |    1.00 |
|            passive_dynamic |          5.10 |     0.02 |          5.07 |     0.01 |    1.01 |
| passive_dynamic_noanalysis |          3.20 |     0.01 |          3.22 |     0.01 |    1.00 |
|                      Arm26 |          0.31 |     0.00 |          0.29 |     0.00 |    1.07 |
|             RajagopalModel |         15.88 |     0.06 |         15.10 |     0.06 |    1.05 |

---

Commit note:

- Caches all enabled OpenSim::Controllers in a model after
  the model graph is finalized (pre-computation)

- At compute time, used by Model::computeControllers as a
  faster substitute to using getComponentList<T>

- This is purely a performance modification. Benchmarks
  (e.g. on ToyDropLanding, from the OpenSim Tutorials)
  indicate a 5-10 % speedup with this change

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2806)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ v4.2
 - Fix a segfault that occurs when using OpenSim's Python Package with Anaconda's Python on a Mac.
 - Expose PropertyHelper class to python bindings to allow editing of objects using the properties interface (useful for editing objects defined in plugins) in python (consistent with Java/Matlab).
 - Whitespace is trimmed when reading table metadata for STO, MOT, and CSV files.
+- Minor performance improvements (5-10 %) for controller-heavy models (PR #2806)
+- `Controller::isEnabled` will now only return whether the particular controller is enabled
+  - Previously, it would return `false` if its parent `Model`'s `Model::getAllControllersEnabled` returned `false`
+  - The previous behavior would mean that `Controller::setEnabled(true); return Controller::isEnabled();` could return `false`
 
 
 v4.1
--- a/OpenSim/Simulation/Control/Controller.cpp
+++ b/OpenSim/Simulation/Control/Controller.cpp
@@ -105,11 +105,7 @@ void Controller::updateFromXMLNode(SimTK::Xml::Element& node,
  */
 bool Controller::isEnabled() const
 {
-    if( getModel().getAllControllersEnabled() ) {
-       return( get_enabled() );
-    } else {
-       return( false );
-    }
+    return get_enabled();
 }
 //_____________________________________________________________________________
 /**
--- a/OpenSim/Simulation/Model/Model.cpp
+++ b/OpenSim/Simulation/Model/Model.cpp
@@ -1925,12 +1925,11 @@ const Vector& Model::getControls(const SimTK::State &s) const
 /** Compute the controls the model */
 void Model::computeControls(const SimTK::State& s, SimTK::Vector &controls) const
 {
-    if (not this->getAllControllersEnabled()) {
+    if (!this->getAllControllersEnabled()) {
         return;
     }
 
-    for (const auto& controllerRefWrapper : this->_enabledControllers) {
-        const Controller& controller = controllerRefWrapper.get();
+    for (const Controller& controller : this->_enabledControllers) {
         controller.computeControls(s, controls);
     }
 }
--- a/OpenSim/Simulation/Model/Model.cpp
+++ b/OpenSim/Simulation/Model/Model.cpp
@@ -974,6 +974,15 @@ void Model::extendConnectToModel(Model &model)
 
     // TODO: Get rid of the SimbodyEngine
     updSimbodyEngine().connectSimbodyEngineToModel(*this);
+
+    // Now that the model is fully connected, cache controllers
+    // locally to reduce the runtime cost of Model::computeControls()
+    this->_enabledControllers.clear();
+    for (const Controller& controller : getComponentList<Controller>()) {
+        if (controller.isEnabled()) {
+            this->_enabledControllers.emplace_back(controller);
+        }
+    }
 }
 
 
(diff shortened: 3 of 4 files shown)


adamkewley:
So there's CI failures due to not being able to upload to sourceforge or something (ignoring this), but the errors with the Windows build are because MSVC (effectively) thinks this is invalid C++:

```c++
if (not x) {
    // bleh
}
```

Because it thinks `not` is an identifier, rather than a keyword. `not` is a standards-compliant C++ keyword, so it's unusual for the compiler to reject it (MSFT has it's reasons, I imagine). The solution is to either remove the `not`s (eugh) or add `/permissive-` to MSVC's build flags:

https://docs.microsoft.com/en-us/cpp/build/reference/permissive-standards-conformance?view=vs-2019

I'll try and add it into `.travis` etc. when I get a second.

adamkewley:
Great review @chrisdembia . I'll make the changes tomorrow and push them up. They're all reasonable. 

For the `not` thing, it's purely something I've picked up from the C++ codebases I worked on previously. I became a fan because it can make for nice-to-read when combined in some awkward flow combined:

```c++
if (user.logged_in() and user.is_admin()) {
    if (not user.can_delete(other_user)) {
        return ErrorReponse{"ruh roh"};
    }
}
```

I didn't really think about the implicit `std::reference_wrapper<T>` -> `T&` conversion also - good catch.

adamkewley:
This looks like it's good to go, is the main outstanding concern the removal of the enablement logic?

chrisdembia:
> is the main outstanding concern the removal of the enablement logic?

Yes, can you address that concern?

The Travis CI tests are failing.

The CHANGELOG should also be updated, especially if the behavior of that function is changing.

adamkewley:
The `CHANGELOG.md` has been updated. 

The reason those travis tests are failing is because of a sourceforge encryption key, not the build or the tests. According to the logs, everything is building + passing. The failure is happening during some `ssh` into sourceforge (presumably, to deploy binaries?).

aseth1:
>is the main outstanding concern the removal of the enablement logic?

Looks to me that the logic was not removed but put in Model to decide whether to override the individual Controllers and disable all Controllers at once. This is effectively what was happening before, but the individual Controller had to look up into it Model, which was confusing. Now it is centralized and the Model is the final arbiter right before the System is created. IMHO the fix is preferred.

chrisdembia:
Hey @aseth1; I was wondering if other parts of the code base were depending on `Controller::isEnabled()` or if `Model::computeControls()` contains the only call site.

chrisdembia:
> The reason those travis tests are failing is because of a sourceforge encryption key, not the build or the tests. According to the logs, everything is building + passing. The failure is happening during some `ssh` into sourceforge (presumably, to deploy binaries?).

Ah...this is because the PR is from a fork.

adamkewley:
So, we're good to go?

chrisdembia:
> So, we're good to go?

No; I clarified my question above:

> Hey @aseth1; I was wondering if other parts of the code base were depending on Controller::isEnabled() or if Model::computeControls() contains the only call site.

I've gone through the code myself and found two other uses of `Controller::isEnabled()`:

https://github.com/opensim-org/opensim-core/blob/c00135abd92d4d7ac155d1e604dfa1612493ea3c/OpenSim/Simulation/Model/ControllerSet.cpp#L81-L91

https://github.com/opensim-org/opensim-core/blob/c00135abd92d4d7ac155d1e604dfa1612493ea3c/OpenSim/Simulation/Model/ControllerSet.cpp#L97-L102

The behavior at these two locations is now different. I suggest the `Model::getAllControllersEnabled()` be employed here.

aseth1:
@chrisdembia thanks for looking into the other places where `Controller::getEnabled()` is used. IMHO, what we had was incorrect, because `Model::_allControllersEnabled` is not a property and can be used (by design I believe) to override the computing of controls by controllers at System execution time.  Your suggestion would not enable a Controller to be "toggled" during computations since it would never be a part of the connected Model. Now a Controller is wired up only if it is individually enabled. At compute time, the Model can still override. I think this is more clear and less error prone behavior.

chrisdembia:
Ah, thanks for explaining. Yes, we want to set the desired states for `TrackingControllers` regardless of `getAllControllersEnabled()`.

chrisdembia:
Thank you, @adamkewley and @aseth1 !

Issue #2781, opened 2020-05-26
Model Calibration before IK fails for some configurations


aymanhab:
Reported by @malborno

aseth1:
Would be nice to document what the configuration dependent failure is. Can you describe the issue?

aymanhab:
@aseth1 In the example provided by @malborno that fails calibration, the subject is facing a random direction.  The IMU-z axis is aligned with subject Pelvis-x (users can populate this info based on IMU placement) but heading correction was trying to align IMU-z with Ground-x (which is hard to define for users in random configurations since pelvis rotation between ground and pelvis is arbitrary). The result was wrong heading_correction. 

I'll beef up the test case, to include this scenario/data-files so you can see what changed and why or try it for yourself. Please let me know your thoughts.

aymanhab:
More details regarding the workflow used above by @malborno :

- Marker data was used to scale and pose the model
- Resulting pose had non-zero pelvis rotation
- Pose resulting from previous step is set as a default-pose which is used to place/register the IMUs to the model. 
- Because of non-zero pelvis rotation, ground-x and pelvis-x are not aligned. 
- The previous mechanism could have worked by guessing which IMU-axis is aligned with Ground-x and redoing calibration but that's error prone.

@jenhicks @carmichaelong Please let me know if I missed any details that could help clarify the confusion based on our dev meeting discussion. Thank you.

jenhicks:
Thanks @aymanhab for the additional details. @aseth1 please let us know your thoughts or if you have additional questions.

RE you comment on the PR, my understanding is that if you specify a pelvis rotation of zero in the default pose and the subject begins walking forward from the calibration pose, you will still get meaningful pelvis rotation, etc. values. @aymanhab have you run some tests to verify that this is indeed the case?

aymanhab:
@jenhicks That's exactly what's tested in the testOpenSense test case, running IK on a calibrated model with 0 pelvis rotation, results downstream will see no change. I'm beefing up the test case to include calibration with non-zero pelvis rotation.

aseth1:
[...]

PR #2810, opened 2020-06-21
Fix heading correction

aymanhab:
Fixes issue #2781 
### Brief summary of changes
Compute heading correction wrt base segment X rather than Ground X. Also fix condition to change sign of correction angle.

### Testing I've completed
Ran tests, including on test data provided by @malborno 

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2810)


aymanhab:
@malborno Do you mind using a shortened versions of your data files to beef up the test case for calibration? Assuming calibration works well with the suggested fix of course.

aymanhab:
@carmichaelong would love to hear your feedback since you're most familiar with this functionality. Thanks

carmichaelong:
@aymanhab Absolutely. I was planning on looking at this more closely this afternoon. Quickly skimming over the contents, I agree with you that it'd be really nice to grab a shortened version of the files you were testing with. Seems like even just a few lines would be useful for "real-world" tests.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ v4.2
 - Fix a segfault that occurs when using OpenSim's Python Package with Anaconda's Python on a Mac.
 - Expose PropertyHelper class to python bindings to allow editing of objects using the properties interface (useful for editing objects defined in plugins) in python (consistent with Java/Matlab).
 - Whitespace is trimmed when reading table metadata for STO, MOT, and CSV files.
+- Fix OpenSense calibration algorithm to handle models facing arbitrary direction, and handle angle sign correctly in all scenarios.
```
aseth1, line 17: The point of the heading correction was to align the gait model so that begins facing the positive X-direction (that is where the pelvis faces when pelvis_rotation is zero)  That is to enable the experimental IMU to be facing any direction, but when represented in OpenSim it will start walking in the X-direction. Is it now going to face the direction determined by the initial pelvis_rotation? What does that mean? How do I now compare pelvis_rotations between subjects?

aymanhab:
Feedback addressed @carmichaelong Thank you

aymanhab:
Updated Changelog, let me know @carmichaelong Thank you

jenhicks:
@antoinefalisse This is the PR I was mentioning today.

aymanhab:
Merging based on review/approval and successful usage by local users. Thanks for the feedback @carmichaelong @aseth1

PR #2826, opened 2020-07-06
Remove temporary logging from StatesTrajectory.

chrisdembia:
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2826)


chrisdembia:
Ah...right! There's probably very good usages for `log_debug()` in this function but I don't think this is one of them.

Thanks for reviewing.

PR #2837, opened 2020-07-16
Add (and implement) `CacheVariable<T>` support

adamkewley:
- Added `CacheVariable<T>` support:
  - Adds strongly-typed cache variable support
  - Now returned from `Component::addCacheVariable` (previously: returned `void`)
  - The returned `CacheVariable<T>` can be used instead of a string name for the cache variable:
    - previously: `isCacheVariableValid(state, "velocity");`
    - now: `isCacheVariableValid(state, velocity);  // decltype(velocity) == CacheVariable<T>`
  - Makes cache variables clearer in a class's signature:
    - e.g: `class SomeComponent : public Component { mutable CacheVariable<double> velocity; };`
  - Adds type-safety to cache variables:
```c++
Vec3 v{};

// previously:
addCacheVariable<Vec3>("velocity", v, Stage::Velocity); 
getCacheVariable<double>(state, "velocity");  // runtime error

// now:
CacheVariable<Vec3> cv = addCacheVariable("velocity", v, Stage::Velocity);
getCacheVariable<double>(state, cv);  // compile-time error
```
  - Non-breaking change. Can be used interchangably with the existing API:

```c++
CacheVariable<double> cv = addCacheVariable("velocity", 0.0, Stage::Velocity);
getCacheVariableValue<double>(state, "velocity");  // valid
```

  - Enables runtime performance enhancements, because the internals of 
    `CacheVariable<T>` can be used to elilde runtime lookups and casts

    - Lookup ellison has already been implemented and yields a ~5 % performance
      increase in ToyDropLanding

    - Further optimizations, such as elilding runtime `dynamic_cast`s have
      not yet been implemented, but can be done in the future in a non-breaking
      way with this API

  - The long-term intention is to phase-out support for storing and 
    looking up cache variables using a string

- internal: a `Component`s cache variables are now stored in an 
            `std::unordered_map` (previously: `std::map`)

- `addCacheVariable` now takes its value prototype argument by value, 
  rather than by const reference:

  - The implementation was copy-constructing into the final location anyway
  - This enables using rvalues when calling `addCacheVariable`:

```c++
// previously: compiler error
addCacheVariable<Vec3>("vec", Vec3{}, Stage::Velocity);
```

- `addCacheVariable` now takes its key argument by value (`std::string`), rather
  than by const reference (`const std::string&`):

  - The implementation requires a value, this just prevents needing to redundantly
    construct a temporary at the callsite, and enables rvalue construction

- `setCacheVariableValue` now accepts the new value as a value, rather than a reference
  to a value (same reasoning as above)

- `setCacheVariableValue` now returns a const reference to the value after setting it

  - Combined with the above, allows single-construction of the new value:

```c++
// previously:
setCacheVariableValue(state, cv, someExpensiveToCopyObject);  // copy occurs
setCacheVariableValue(state, cv, SomeExpensiveClass{});  // temporary constructed, then copied
SomeExpensiveClass& v = getCacheVariableValue(state, cv);  // additional lookup + cast

// now:
SomeExpensiveClass& v = setCacheVariableValue(state, cv, SomeExpensiveClass{});  // arg rvalue-constructed (cheap), no lookups
```

- `addCacheVariable` no longer allows double-assignment of the same key

  - This prevents weird double-key runtime bugs and enables `CacheVariable<T>`
    optimizations (specifically: `dynamic_cast` elilding) in the future:

```c++
// previously:
addCacheVariable<double>("k", 0.0, Stage::Velocity);
addCacheVariable<int>("k", 0, Stage::Velocity);  // no error
getCacheVariable<double>(state, "k");  // runtime error

// now:
addCacheVariable<double>("k", 0.0, Stage::Velocity);
addCacheVariable<int>("k", 0, Stage::Velocity);  // exception with explanation
```

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2837)


adamkewley:
One open question I need to figure out is whether it should actually be renamed to `NamedCacheVariable<T>`, to make room for a future with no string lookups, which would use a `CacheVariable<T>`. This would enable both methods to be used at the same time (if string lookups are desired longer-term).

chrisdembia:
Neat! I'm halfway through my review.

adamkewley:
@chrisdembia thanks for the feedback. If I come off as negative/disagreeable in some comments then please don't take them *too* seriously - I'm just trying to be direct.

I believe I've addressed each thing you asked about in the latest couple of commits. I am still in disagreement about `OPENSIM_THROW` (rather than `throw OPENSIM_EXCEPTION`) but arguing about it is verging on extreme bikeshedding, so I changed the raw `throw`s for the macro and I've kicked the suggestion of *constructing* the throwable in another PR (this: https://github.com/opensim-org/opensim-core/pull/2841) so that this PR isn't left dangling for ages on a tiny thing.

I *believe* the only change I've omitted is splitting the error messages that make the line >80 chars. My reason for keeping that as-is is because, imho, error messages should just be allowed to sprawl a bit so that text searches over the codebase are more likely to succeed (I've worked on larger projects where finding the error message can actually be a challenge, so I'm probably more scared of this non-issue than reasonable). I can change it if you want, though - it's no big.

tkuchida:
> imho, error messages should just be allowed to sprawl a bit so that text searches over the codebase are more likely to succeed (I've worked on larger projects where finding the error message can actually be a challenge, so I'm probably more scared of this non-issue than reasonable).

FWIW, I've had minor difficulty tracking down an error message because of this.

tkuchida:
@chrisdembia Re. error messages:

> I had never thought about making an exception for message strings. I like the idea, but I also think that questions of style should be handled separately from the objective of this PR. Perhaps leave this for now; I'll chat with others on the dev team.

Another possible solution is to assign each error message a unique number/code; also might help someone search the Forum.

adamkewley:
So I *think* that should be everything addressed now. If anyone else has comments to make then feel free. Otherwise, this should be good to merge?

adamkewley:
Rough stats (32 repeats) between this PR and `master`:

|                   Gait2354 |       0.26 |     0.00 |       0.26 |     0.00 |    1.01 |
|             ToyDropLanding |      13.74 |     0.01 |      11.89 |     0.01 |    1.16 |
|   ToyDropLanding_nomuscles |       0.43 |     0.00 |       0.43 |     0.00 |    1.00 |
|            passive_dynamic |       4.84 |     0.00 |       4.76 |     0.00 |    1.02 |
| passive_dynamic_noanalysis |       3.05 |     0.00 |       3.00 |     0.00 |    1.02 |
|                      Arm26 |       0.28 |     0.00 |       0.26 |     0.00 |    1.06 |
|             RajagopalModel |      14.27 |     0.01 |      13.32 |     0.41 |    1.07 |

I don't have measurements to figure out how much of a boost elilding the `dynamic_cast`s will be. It'd be interesting to valgrind this branch to see where (e.g.) ToyDropLanding now spends its time

adamkewley:
So removing the `sort` step for the keys causes almost all tests to pass apart from `testMomentArmSolver`, which segfaults.

Previous attempts to switch to `unordered_map` caused several tests to segfault and many other tests to fail, so something has already improved things (probably, that the `ResetOnCopy` is ensuring that the map initialization always happens in a *clean* map in a deterministic order) 

Looking into why `testMomentArmSolver` is special, here is the backtrace:

```
#0  0x00007ffff6793e98 in SimTK::ClonePtr<SimTK::AbstractValue>::reset (this=0x5555558cd0a0) at /home/adam/Desktop/opensim-core/dependencies/simbody/SimTKcommon/./include/SimTKcommon/internal/ClonePtr.h:251
#1  SimTK::ClonePtr<SimTK::AbstractValue>::~ClonePtr (this=0x5555558cd0a0, __in_chrg=<optimized out>) at /home/adam/Desktop/opensim-core/dependencies/simbody/SimTKcommon/./include/SimTKcommon/internal/ClonePtr.h:191
#2  SimTK::CacheEntryInfo::~CacheEntryInfo (this=0x5555558cd050, __in_chrg=<optimized out>) at /home/adam/Desktop/opensim-core/dependencies/simbody/SimTKcommon/Simulation/include/SimTKcommon/internal/StateImpl.h:259
#3  SimTK::Array_<SimTK::CacheEntryInfo, unsigned int>::destruct (e=0x5555558cdfa0, b=0x5555558cd050) at /home/adam/Desktop/opensim-core/dependencies/simbody/SimTKcommon/./include/SimTKcommon/internal/Array.h:3234
#4  SimTK::Array_<SimTK::CacheEntryInfo, unsigned int>::clear (this=0x5555558caea8) at /home/adam/Desktop/opensim-core/dependencies/simbody/SimTKcommon/./include/SimTKcommon/internal/Array.h:2601
#5  SimTK::Array_<SimTK::CacheEntryInfo, unsigned int>::deallocate (this=0x5555558caea8) at /home/adam/Desktop/opensim-core/dependencies/simbody/SimTKcommon/./include/SimTKcommon/internal/Array.h:1755
#6  SimTK::Array_<SimTK::CacheEntryInfo, unsigned int>::~Array_ (this=0x5555558caea8, __in_chrg=<optimized out>) at /home/adam/Desktop/opensim-core/dependencies/simbody/SimTKcommon/./include/SimTKcommon/internal/Array.h:1742
#7  SimTK::PerSubsystemInfo::~PerSubsystemInfo (this=0x5555558cad40, __in_chrg=<optimized out>) at /home/adam/Desktop/opensim-core/dependencies/simbody/SimTKcommon/Simulation/include/SimTKcommon/internal/StateImpl.h:619
#8  0x00007ffff679458c in SimTK::Array_<SimTK::PerSubsystemInfo, unsigned int>::destruct (e=0x5555558cbbf0, b=0x5555558cb030) at /home/adam/Desktop/opensim-core/dependencies/simbody/SimTKcommon/./include/SimTKcommon/internal/Array.h:3233
#9  SimTK::Array_<SimTK::PerSubsystemInfo, unsigned int>::clear (this=0x5555558c85c0) at /home/adam/Desktop/opensim-core/dependencies/simbody/SimTKcommon/./include/SimTKcommon/internal/Array.h:2601
#10 SimTK::Array_<SimTK::PerSubsystemInfo, unsigned int>::deallocate (this=0x5555558c85c0) at /home/adam/Desktop/opensim-core/dependencies/simbody/SimTKcommon/./include/SimTKcommon/internal/Array.h:1755
#11 SimTK::Array_<SimTK::PerSubsystemInfo, unsigned int>::~Array_ (this=0x5555558c85c0, __in_chrg=<optimized out>) at /home/adam/Desktop/opensim-core/dependencies/simbody/SimTKcommon/./include/SimTKcommon/internal/Array.h:1742
#12 SimTK::StateImpl::~StateImpl (this=0x5555558c85c0, __in_chrg=<optimized out>) at /home/adam/Desktop/opensim-core/dependencies/simbody/SimTKcommon/Simulation/include/SimTKcommon/internal/StateImpl.h:952
#13 0x00007ffff678a771 in SimTK::State::~State (this=<optimized out>, __in_chrg=<optimized out>) at /home/adam/Desktop/opensim-core/dependencies/simbody/SimTKcommon/Simulation/src/State.cpp:53
#14 0x000055555557497e in OpenSim::MomentArmSolver::~MomentArmSolver (this=0x7fffffffc460, __in_chrg=<optimized out>) at /home/adam/Desktop/opensim-core/opensim-dependencies-install/simbody/include/simbody/SimTKcommon/internal/MatrixHelper.h:92
#15 testMomentArmDefinitionForModel (filename=..., coordName=..., muscleName=..., rom=..., mass=<optimized out>, errorMessage="testMomentArmsAcrossCompoundJoint: FAILED") at /home/adam/Desktop/opensim-core/OpenSim/Simulation/Test/testMomentArms.cpp:286
#16 0x000055555557570e in testMomentArmsAcrossCompoundJoint () at /usr/include/c++/8/ext/new_allocator.h:79
#17 0x000055555556f35f in main () at /home/adam/Desktop/opensim-core/OpenSim/Simulation/Test/testMomentArms.cpp:67
```
One immediate observation is that constructing a `MomentArmSolver` involves copying the `SimTK::State` locally as a `mutable SimTK::State _stateCopy`, so that later member methods can use the local (mutable) state copy, rather than the caller-provided state:

- ctor: https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/MomentArmSolver.cpp#L38
- usage: https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/MomentArmSolver.cpp#L61

Which is an example of the kind of code that is technically illegal (these methods should ideally use one state, no re-init, no re-realizing, etc.), but it's very easy to do. Interchanging SimTK::States is the most common source of segfaults when I worked on cache variables.

The bug I need to find is that the layout of the caller-provided `State` mismatches the local copy of the state, which causes a cache variable deep inside the copy to be corrupted (e.g. by `nullptr`ing it, double-freeing it, etc.), which causes the copy's destruction to later segfault (e.g. because the `ClonePtr` is trying to `delete` a dead pointer).

Quick debug printing was insightful. I added this into `Component.cpp` to see what was going on:

```c++
        std::cerr << static_cast<const void*>(this) << ":" << std::endl;
        for (const std::string& k : keys) {
          std::cerr << "    " << k << std::endl;
        }
```

Which prints the insertion order of cache variable entries. Most of them are small, or have no entries, but these two printouts were interesting:

```
// some allocs

0x5593f0d27460:
    fiber_length_deriv
    activation_deriv
    speed
    dynamicsInfo
    actuation
    lengthInfo
    velInfo
    potentialEnergyInfo

// some more allocs

0x5593f0d27460:
    fiber_length_deriv
    activation_deriv
    potentialEnergyInfo
    velInfo
    lengthInfo
    speed
    dynamicsInfo
    actuation
```

Because it's the same model (address) that is being initialized twice (probably by a second call to an init method somewhere). Which indicates a new BEWARE source of bugs:

- Reinitializing the same `Component` may change the order in which cache variables are allocated in SimTK

And a new observation about map iteration order in general:

- Input keys are the same
- And they are inserted in the same order
- **And** the initial state of the map is the same, **including internal state***

That is to say, calling `map.clear();` does not put a `map` into the same state as when it is default/value initialized. A quick C++ test demonstrates that, yeah, that's true:

```c++
#include <unordered_map>
#include <random>
#include <iostream>

int main(int argc, char** argv) {
        using v_t = unsigned;
        constexpr size_t num_entries = 10000;

        std::uniform_real_distribution<v_t> dist{};
        std::default_random_engine e(std::chrono::system_clock::now().time_since_epoch().count());
        auto generate = [&]() { return dist(e); };

        std::vector<v_t> keys_in_order;
        for (size_t i = 0; i < num_entries; ++i) {
            keys_in_order.push_back(generate());
        }

        // populate from a default-initialized instance
        std::unordered_map<v_t, v_t> um;
        for (v_t v : keys_in_order) {
            um[v] = {};
        }

        // save the keyorder of a default-populated map
        std::vector<v_t> default_init_keyorder;
        for (auto p : um) {
            default_init_keyorder.push_back(p.first);
        }

        // logically, this resets the map to the above state?
        um.clear();

        // populate from the '.clear()'ed state
        for (v_t v : keys_in_order) {
            um[v] = {};
        }

        // save the keyorder of the `.clear()`ed map
        std::vector<v_t> reset_keyorder;
        for (auto p : um) {
            reset_keyorder.push_back(p.first);
        }

        // fails
        assert(default_init_keyorder == reset_keyorder);

        return 0;
}
```

So the above behavior sets the stage for a bug:
- Two--*logically* identical, from a cachevariable PoV--`Component`s produce incompatible `SimTK::State`s
- Those two states are 'live' in the test (a state is saved somewhere, and then a new state is created later)

I don't know exactly *how* this particular segfault happens. I'm guessing that the state copy does something fancy internally that causes a little bit of aliasing, followed by double-destruction, or something like that. No idea, will take a while to dig. From the perspective of this PR, though, it's clear that the BEWARE needs to be updated (with the above) and that the key sorting needs to remain because it seems painfully easy to cause those types of faults in downstream code (e.g. by copying states around).

adamkewley:
@chrisdembia I don't really have time to do a full deep-dive into the particular sources of segfaults etc. here (and in all the other places this pattern is used). Hopefully, just clarifying why the BEWARE section is important is enough to move this forward.

Essentially, before changing from a `std::map` people were reassuring me that "it shouldn't matter" and "yeah, we never interchange states" etc. but I've seen enough examples--in tree, even--that it's not so clear-cut, and that `SimTK::State`s are potentially interchanged a bunch, or models are reinitialized, etc.

Ultimately, the switch to `unordered_map` is a perf change, to prevent the string lookups from doing rbtree searches (profilers indicated that the CPU was spending a bunch of time fetching tree nodes), but changing the datastructure kicked up some bugs. The bugs might be in (what I call) "downstream" code, which makes incorrect assumptions, but my goal is to ensure that I can integrate this change without having to change any existing behavior

adamkewley:
W.r.t reuse, the issue is whether it's ok to use different states
initialized from one model, created by multiple calls to init methods,
rather than copying a state and using it with a one-time initialized model.

The bug in the tests is that there are two different states created from
the same model being initialized twice. I'm not sure where that falls on
the correctness spectrum

chrisdembia:
> The bug in the tests is that there are two different states created from
the same model being initialized twice. I'm not sure where that falls on
the correctness spectrum

Ah thanks for clarifying. That's a bit of a gray area I suppose, and I'm fine with that not being an acceptable use.

chrisdembia:
@adamkewley can you clean up some doxygen warnings that resulted from this PR?

```
/Users/chris/nmbl/moco/merge/3/opensim-core/OpenSim/Common/Component.h:1597: warning: argument 'name' of command @param is not found in the argument list of OpenSim::Component::setCacheVariableValue(const SimTK::State &state, const std::string &k, T value) const
/Users/chris/nmbl/moco/merge/3/opensim-core/OpenSim/Common/Component.h:1611: warning: The following parameter of OpenSim::Component::setCacheVariableValue(const SimTK::State &state, const std::string &k, T value) const is not documented:
  parameter 'k'
```

adamkewley:
Sure,

I also just noticed, when setting up a new toolchain etc on my laptop, that
there's a bug that makes this break when building in pure debug mode (I
usually use RelWithDebugInfo).

Apparently, the simtk indices have a debug-only assertion that checks that
both sides of an equality expression are not invalid. The patch contains
lines like 'index != InvalidIndex', which now causes a termination in debug
mode.

I'm away at the moment but I'll try and hotfix that bug ASAP because it
will break anyone who uses debug mode on the master branch

chrisdembia:
Ah. I think you can use `SimTK::CacheEntryIndex::isValid()`:

https://github.com/simbody/simbody/blob/320bbe9f39f5d24bc06ff30bc882caedee2d0654/SimTKcommon/include/SimTKcommon/internal/common.h#L454

PR #2839, opened 2020-07-17
WIP: Exception-safe directiory switching

adamkewley:
This is just a refactoring idea I'm playing around with on the side. 

Effectively, this PR aims to implement an RAII wrapper for directory switching. The motivation for doing this is to simplify some of the code sections that currently rely on changing the working directory of the caller temporarily (e.g. because XML files can contain relative paths).

Here is a simplified example of a common code pattern in OpenSim (see: [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/AbstractTool.cpp#L413) for a direct example).

```c++
std::string saveWorkingDirectory = IO::getCwd();
std::string directoryOfSetupFile = IO::getParentDirectory(aToolSetupFileName);

IO::chDir(directoryOfSetupFile);  // the calling process now switches

// do stuff

IO::chDir(saveWorkingDirectory);
```

This pattern occurs in *many* places in OpenSim. There are ~85 uses of `chDir`, most of them are part of patterns like this. In principle, it's fine, but it misses out some subtleties, such as when `// do stuff` can throw:

```c++
std::string saveWorkingDirectory = IO::getCwd();
std::string directoryOfSetupFile = IO::getParentDirectory(aToolSetupFileName);

IO::chDir(directoryOfSetupFile);

// bug: the process will still be dir-switched when unwinding
doSomethingPotentiallyExceptionThrowing(); 

// correct
try {
  doSomethingPotentiallyExceptionThrowing(); 
} catch (...) {
  IO::chDir(saveWorkingDirectory);
  throw;
}

IO::chDir(saveWorkingDirectory);
```

Or when the directory switch is conditional (see [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/Model/AbstractTool.cpp#L873)):

```c++
std::string wd;
if (cond) {
  wd = IO::getCwd();
  IO::chDir(IO::getParentDirectory(someOtherDir));
}

// bug if cond != true
IO::chDir(wd);

 // correct
if (cond) {
  IO::chDir(wd);
}
```

The solution to these problems is to lean on C++'s strong RAII guarantees and lifetime semantics. This PR introduces a new class, currently called `IO::CwdChanger`, which temporarily changes the working directory of the calling process for its lifetime. This behavior is maintained even when unwinding, so the exceptional case becomes this:

```c++
IO::CwdChanger cwdChanger{IO::getParentDirectory(aToolSetupFileName)};
doSomethingPotentiallyExceptionThrowing(); 
// IO::CwdChanger::~CwdChanger automatically switches dir back, even when throwing
```

The class also has a noop case, which makes the class not change the directory on construction or destruction. This might seem dumb, but is useful for the conditional case:

```c++
IO::CwdChanger cwdChanger = cond ? IO::CwdChanger{path} : IO::CwdChanger::noop();
// now the cwd is *either* path, or unchanged
// the destructor implicitly changes back to the original dir 
```

This PR may also include some quality-of-life refactors in the code areas around the usage points.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2839)

### DIFF
--- a/Bindings/common.i
+++ b/Bindings/common.i
@@ -36,8 +36,8 @@
 %include <OpenSim/Common/Storage.h>
 %template(ArrayStorage) OpenSim::ArrayPtrs<OpenSim::Storage>;
 %include <OpenSim/Common/Units.h>
-%include <OpenSim/Common/IO.h>
 %ignore OpenSim::IO::CwdChanger;
+%include <OpenSim/Common/IO.h>
 %include <OpenSim/Common/Function.h>
 
 %template(SetFunctions) OpenSim::Set<OpenSim::Function, OpenSim::Object>;
--- a/Bindings/common.i
+++ b/Bindings/common.i
@@ -37,6 +37,7 @@
 %template(ArrayStorage) OpenSim::ArrayPtrs<OpenSim::Storage>;
 %include <OpenSim/Common/Units.h>
 %include <OpenSim/Common/IO.h>
+%ignore OpenSim::IO::CwdChanger;
 %include <OpenSim/Common/Function.h>
 
 %template(SetFunctions) OpenSim::Set<OpenSim::Function, OpenSim::Object>;
--- a/OpenSim/Common/IO.cpp
+++ b/OpenSim/Common/IO.cpp
@@ -690,3 +690,48 @@ void IO::eraseEmptyElements(std::vector<std::string>& list)
             ++it;
     }
 }
+
+IO::CwdChanger::CwdChanger() {
+}
+
+IO::CwdChanger::CwdChanger(const std::string& newDir) :
+    _existingDir{getCwd()} {
+
+    chDir(newDir);
+}
+
+IO::CwdChanger IO::CwdChanger::noop() {
+    return CwdChanger{};
+}
+
+IO::CwdChanger IO::CwdChanger::changeTo(const std::string& newDir) {
+    return CwdChanger{newDir};
+}
+
+IO::CwdChanger IO::CwdChanger::changeToParentOf(const std::string& path) {
+    return changeTo(getParentDirectory(path));
+}
+
+// a custom move constructor is necessary because the default move
+// constructor, which effectively calls `this->_existingDir{std::move(tmp._existingDir)}`
+// is specified to leave `tmp._existingDir` in an unspecified state:
+//
+// from: https://en.cppreference.com/w/cpp/string/basic_string/basic_string
+//
+// > Move constructor. Constructs the string with the contents of
+// > other using move semantics. other is left in valid, but
+// > unspecified state
+//
+// `~CwdChanger` requires that `tmp._existingDir.empty() == true`; otherwise,
+// destruction of the temporary will cause a directory change.
+IO::CwdChanger::CwdChanger(IO::CwdChanger&& tmp) :
+    _existingDir{} {
+
+    std::swap(this->_existingDir, tmp._existingDir);
+}
+
+IO::CwdChanger::~CwdChanger() noexcept {
+    if (!_existingDir.empty()) {
+        chDir(_existingDir);
+    }
+}
--- a/OpenSim/Common/IO.h
+++ b/OpenSim/Common/IO.h
@@ -144,17 +144,30 @@ public:
             return CwdChanger{};
         }
         CwdChanger() {
-            // noop ctor: enables conditional directory switching:
+            // default (noop) ctor that enables conditional directory switching:
             //    CwdChanger c = changeDir ? CwdChanger{"new/dir"} : CwdChanger{};
         }
         CwdChanger(const std::string& newDir) : _existingDir{IO::getCwd()} {
             chDir(newDir);
         }
         CwdChanger(const CwdChanger&) = delete;
-        CwdChanger(CwdChanger&& tmp) {
-	    std::swap(_existingDir, tmp._existingDir);
-	    tmp._existingDir.clear();
+
+        // a custom move constructor is necessary because the default move
+        // constructor, which effectively calls `this->_existingDir{std::move(tmp._existingDir)}`
+        // is specified to leave `tmp._existingDir` in an unspecified state:
+        //
+        // from: https://en.cppreference.com/w/cpp/string/basic_string/basic_string
+        //
+        // > Move constructor. Constructs the string with the contents of
+        // > other using move semantics. other is left in valid, but
+        // > unspecified state
+        //
+        // `~CwdChanger` requires that `tmp._existingDir.empty() == true`; otherwise,
+        // destruction of the temporary will cause a directory change.
+        CwdChanger(CwdChanger&& tmp) : _existingDir{} {
+            std::swap(this->_existingDir, tmp._existingDir);
         }
+
         CwdChanger& operator=(const CwdChanger&) = delete;
         CwdChanger& operator=(CwdChanger&&) = delete;
         ~CwdChanger() noexcept {
--- a/OpenSim/Common/IO.h
+++ b/OpenSim/Common/IO.h
@@ -134,6 +134,7 @@ public:
             const std::string& string, const std::string& ending);
     static void eraseEmptyElements(std::vector<std::string>& list);
 
+#ifndef SWIG
     /**
      * A class that:
      *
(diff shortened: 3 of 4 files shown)


adamkewley:
@chrisdembia where's the similar RAII class, just so I can either integrate the existing class or keep using this?

chrisdembia:
Oh sorry, the class is not for changing directories, it's for deleting files.

https://github.com/opensim-org/opensim-core/blob/8a4171e248890a0fcdbb40ec80814ab90d465245/OpenSim/Common/CommonUtilities.h#L52-L69

I just meant that this other class relies on the same RAII pattern.

Side rant: I still don't understand the name RAII. I like the name "Scope-Bound Resource Management" from https://stackoverflow.com/questions/2321511/what-is-meant-by-resource-acquisition-is-initialization-raii

adamkewley:
Ah, I see, thanks :)

RAII is a bit of a tongue twister, and I prefer your name :).

It's an interesting ideology when applied to the extreme though (e.g. even the acquisition of stack space is "resource allocation" and should be "initialized"). I worked with some C++ devs who argued that doing anything in a C++ constructor is a code-smell because it implies that a class's members (a resource) are in an uninitialized state at one point in the application:

```c++
struct File {
    int fd;
    size_t filesize;
    File(const std::string& path) {
        // impure: this->fd is in an undefined state
        int fd = ::open(path.c_str(), ...);
        if (fd == -1) {
            throw std::runtime_error{"could not open file"};
        }
        this->fd = fd;
        // impure: this->filesize is in an undefined state
        // ::lseek(...); etc.
        if (lseek_failed) {
            throw std::runtime_error{"you now have a fd leak"};
        }
    }
    // dtor etc.
};

// "correct"
struct Fd {
    int fd;
    Fd(const std::string& path) : fd{::open(path.c_str())} {
        if (this->fd == -1) {
            throw std::runtime_error{"could not open file"};
        }
    }
    // dtor etc.
};

struct File {
    Fd fd;
    size_t filesize;
    File(const std::string& path) : fd{path}, filesize{get_filesize_or_throw(fd)} {}
};

**Comment by adamkewley** [comment]
I'm not planning to touch this until Friday (I have a new thing - "refactoring Fridays") so I'll leave this here to perculate until then.

If you guys think it'd be a good refactor to roll out codebase-wide, I can do that. At the moment, this PR just contains the proposed implementation and an example rollout in `Object.cpp`. A full rollout will hit ~5-15 other implementation files, and I'll likely refactor things near any changes (cleanups, etc.).

**Comment by adamkewley** [comment]
> Could you explain why the default copy constructor would not suffice?

For that, the *default* ctor would copy the saved (existing) working directory into the copy target, no immediate side-effect, but the class's destructor would then cause the destruction of either the copy or the original to switch to the existing working dir. This *might* be what the copyer intended, but might not be, so I figured it's best to just `delete` copy construction until there's a use-case that requires it (e.g. like there's a use-case for default construction).

This is also a muscle-memory side-effect from writing the destructor. Whenever I write one, I immediately just `delete` the other methods until I need them because of RO5:

https://en.cppreference.com/w/cpp/language/rule_of_three

And I just add them back in whenever they're needed. Imho, no need to write the code until there's a use-case.

**Comment by adamkewley** [comment]
I'll clean up the other issues. I recently changed to a macbook air (closer to what users have, perf measurements, SCONE build, etc.) and it's too slow for CLion, and emacs (my Linux default) is a bit iffy with the mac keybinds, so I'm also midway through configuring vim >,< - thank you for spotting these because it isn't showing it (yet)

**Comment by adamkewley** [comment]
> For my understanding, why not delete the move constructor?

In that case, it's because C++11 does not have copy-ellison (/move-ellison) guranteees (C++17 does), so trying to compile the following:

```c++
struct C {
    C() = default;

    C(const C&) = delete;
    C& operator=(const C&) = delete;
    C& operator=(C&&) = delete;

    C(C&&) = delete;  // must exist in C++11
};

int main() {
    C c = true ? C{} : C{};
    return 0;
}
```

Results in (C++11):
 
```bash
:!c++ -std=c++11 copy-ellison.cpp -o ce
copy-ellison.cpp:12:18: error: call to deleted constructor of 'C'
    C c = true ? C{} : C{};
                 ^~~
copy-ellison.cpp:8:5: note: 'C' has been explicitly marked deleted here
    C(C&&) = delete;  // must exist in C++11
    ^
1 error generated.

shell returned
```

And works in C++17:

```c++
c++ -std=c++17 copy-ellison.cpp -o ce
// exits silently with exit code == 0
```

And this is significant when direct initialization of `IO::CwdChanger` (e.g. `IO::CwdChanger{"path"}`) is not preferred, such as:

```c++
// conditionally directly initializes an instance, followed by rvalue-constructing `c`
// with it
IO::CwdChanger c = cond ? IO::CwdChanger{"/"} : IO::CwdChanger::noop();
```

If you want some, uh, "fun" I'd recommend reading:

https://en.cppreference.com/w/cpp/language/initialization

Which, if you dig through several related concepts, will show that (in C++11):

- `C c = C{};` is *copy initialization* (here: https://en.cppreference.com/w/cpp/language/copy_initialization), but there's this cute point buried on that page:

> but, because

adamkewley:
(fat fingers)

adamkewley:
@chrisdembia what's the feeling about using a static method rather than directly constructing it? I'm thinking that it could be more readable:

```c++
IO::CwdChanger cwd = IO::CwdChanger::changeTo("some-dir");
IO::CwdChanger cwd = cond ? IO::CwdChanger::changeTo("other-dir") : IO::CwdChanger::noop();
```

And I've added a convenience method, `changeToParentOf` for the most common pattern, which seems to be changing into the parent directory of some file (see diffs).

As for the naming, I see where you're coming from but there's a few things to keep in mind about using a very literal name for this. 

First is that the resulting name, with namspacing, is going to result in really long lines in the conditional case (ignoring the static methods for a second):

```c++
IO::WorkingDirectoryChanger cwd = cond ? IO::WorkingDirectoryChanger::noop() : IO::WorkingDirectoryChanger{"path"};
```

Which is kind of annoying, but nothing major. The other thing is that the existing code already introduces the concept of a "cwd" for all cwd-related methods (e.g. `curCwd` `changeCwd`, etc. etc.), so the introduction of a class called `CwdChanger` in the same namespace + file as all the other cwd-related functions probably probably introduces a minimal amount of new information and gives the impression that the class is in some way related to the functions (it is - it's effectively a lifetime wrapper for them).

chrisdembia:
> what's the feeling about using a static method rather than directly constructing it?

This is great!

> As for the naming, I see where you're coming from but there's a few things to keep in mind about using a very literal name for this.

Yes, these are good points. Given that "Cwd" is already used, I'm fine with your chosen class name.

adamkewley:
Still looking into this - seems broken when generating SWIG bindings because SWIG doesn't like move-only types (?!).

For the doxygen formatting - haven't looked into it.

adamkewley:
Sure, I can change it to that - it's only a 2-liner.

To play devil's advocate, I think the drawback of putting the exclusion in `common.i` is that it's non-obvious to people who aren't keeping SWIG in-mind when looking through the header files (e.g. the devs, like me, who are mostly just hacking on the C++ backend).

So if I'm reading (e.g.) some class def in a header and I see:

```c++
struct C {
void f1();
#ifndef SWIG
void f2();
#endif
void f2();
}
```

It's cosmetically a bit ugly--and the C++ puritans will throw a fit at the macro-use--but, even without knowing what SWIG is or how it builds etc., I'd probably guess that there's something called SWIG and `f2` will not be declared when using it. I'd probably `grep` for `SWIG` in-tree, or google it, to see what's going on. This is especially important when, on the bindee's side, something like `c.f2()` doesn't compile. In that instance, the first place I would look for a bug is the header file, because I'd assume a definition is missing somewhere, and I'd immediately see the `ifndef` and go fix/change whatever's needed.

adamkewley:
All looking good to go, with the change, should we go ahead and merge this?

adamkewley:
Documentation seems fine. This is good to go?

chrisdembia:
@adamkewley we typically use Reviewable to keep track of which comments are in progress versus resolved.

Thanks for adding this class; it's so much cleaner than what we were doing before.

Issue #118, opened 2014-07-22
Relative paths in XML files: make use of Simbody's new Pathname functions, perhaps with a Pathname property.


chrisdembia:
People have reported various issues with using relative paths in XML files. The only issue I have run into is the following:
1. A setup file `setup.xml` specifies an external loads file `../extloads.xml`. This can be successfully specified with a path relative to `setup.xml`.
2. The external loads file then specifies a path to a GRF file `data/grf_data.mot` relative to the external loads file.

The issue is that `data/grf_data.mot` will be resolved relative to `setup.xml`, _not_ the external loads file. That's because, to resolve relative paths, OpenSim changes directory to the setup file (well, the directory containing the setup file) and then tries to open files.

jimmyDunne:
Is this issue on the agenda for 3.2.1?

sherm1:
> Is this issue on the agenda for 3.2.1?

Assuming we're going to use Simbody 3.6 with OpenSim 3.2.1 (per earlier discussion), this could use @carmichaelong 's "specified working directory" routine designed for this purpose and under discussion in PR simbody/simbody#307.

chrisdembia:
In order to fix this issue, we have to come up with cases where we know relative paths are failing. I know of one specific case, but @jimmyDunne has mentioned other distinct cases that I haven't yet encountered.

aymanhab:
@jimmyDunne Most other issues for 3.2.1 are surgical fixes and require no across the board  testing of the API/GUI/Scripting environments. Switching to using a Path property that we haven't even created or tried under the variety of scenarios that it needs to address seems risky to me. Also the GUI and scripting need to handle this new property type. It's a great long term project but I feel it's too much for 3.2.1.

aymanhab:
@jimmyDunne I'd say however that I can look at the reported bug(s) related to relative paths and if a surgical fix can be made do it, otherwise document behavior and/or improve error message(s). That seems like more reasonable scope.

chrisdembia:
I agree with @aymanhab. I am *very* excited for improved path handling in OpenSim, but even if OpenSim uses Simbody 3.6, we'd be rushed to try to get OpenSim to use it in time for the 3.2.1 release.

jenhicks:
Create a class FilePath or Path that can generate useful strings of filenames. Knows working directory. Knows original relative path name.

jenhicks:
@chrisdembia filed a specific bug on the SimTK bug tracker

[...]

PR #2844, opened 2020-07-23
Optimize component path splitting in OpenSim::Component::traverseToState

adamkewley:
So, I'm looking for fairly-isolated performance bottlenecks, the kind that do not require changing a bunch of fundamental APIs (e.g. like the recent CacheVar changes). Running my usual ToyDropLanding sim post-CacheVariable change, revealed a fairly isolated ~8 % bottleneck here:

graph

Which indicates that the `ComponentPath` constructor,  is sucking up ~7.5 % CPU (the remainder is string allocations, rbtree lookups, and the `Path`'s virtual dtor). The reason the constructor is so slow is because it:

- Copies things into the base class, such as invalid chars (1 %), which do not change at runtime
- Slices up each element of the path into a new `std::string` object, and allocates them into a `std::vector` (~3-4 %)
- Reorganizes the path elements to absolute-ize them (e.g. `a/b/../c` -> `a/c`), which involves iterating over the vector and performing string comparisons on each element (~1 %)
- Searches for illegal characters in each element with another search over each element (~1.5 %)

Ultimately, this is all to satisfy a step-by-step traversal of a component tree with the aim of locating a state variable (in `Component::traverseToStateVariable`). Key obersvations:

- `Component::getStateVariableValue` (10.5 % of perf) is called 3964729 times
- `ComponentPath::ComponentPath` (7.5 % perf) is called 3965134 times
  - `Component::traverseToStateVariable` is recursive when the path contains multiple elements
  - The minor difference in these call numbers indicates that most state variable lookups are direct, rather than requiring actual traversal, so it *should* be possible to optimize for the common case of not needing to split the string at all
- Whatever this PR proposes needs to handle relative paths at *some* level, but not necessarily in the proposed `split` function. There could be a separate `abspath` function that handles that.
- Whatever this PR proposes needs to handle the invalid elements, which are currently handled by throwing an exception. The advantage of not moving things into a `std::vector` is that comparisons can be done on one string target. The advantage of knowing all invalid characters at compile-time is that the comparison can be loop-unrolled into `n` char comparisons (effectively `memchr`, once the compiler hits it)

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2844)


chrisdembia:
Can you mark this PR as WIP or as a Draft?

adamkewley:
Even this half-hearted and fairly ugly-looking attempt seems to yield a measurable speedup without having to perform loads of repeats etc:

|                   Gait2354 |       0.25 |     0.00 |       0.25 |     0.00 |    1.00 |
|             ToyDropLanding |      13.67 |     0.00 |      12.94 |     0.01 |    1.06 |
|   ToyDropLanding_nomuscles |       0.43 |     0.00 |       0.43 |     0.00 |    0.99 |
|            passive_dynamic |       4.81 |     0.00 |       4.82 |     0.00 |    1.00 |
| passive_dynamic_noanalysis |       3.05 |     0.00 |       3.06 |     0.00 |    1.00 |
|                      Arm26 |       0.28 |     0.00 |       0.27 |     0.00 |    1.04 |
|             RajagopalModel |      14.23 |     0.01 |      13.48 |     0.00 |    1.06 |

(this is pre-CacheVariable merge, probably a larger percentage in the post-merge comparison)

So one thing to do is to see if the `ComponentPath` class can be updated to internally hold its data in a single `std::string`, rather than a `std::vector`, because I suspect (and the profiles + perf improvements confirm) that there's some perf losses from constructing temporaries that heap-allocate substrings.

My (fairly ugly atm) attempt to get around this is to write a pure function that rearranges the string in-place with no auxiliary storage. This lets callers `normalize` a path string without any allocation overhead.

adamkewley:
So the latest iteration of this work essentially:

- Removes `OpenSim::Path` altogether, because it doesn't seem to be used anywhere (is some external codebase reliant on this base class)?
- Rewrites `OpenSim::ComponentPath`'s implementation to use a single `std::string` internally for all operations, such that constructing a `ComponentPath` is guaranteed to, at most, involve one allocation of a string (in many cases, SSO will make it involve zero allocations for strings <16 or so characters)
- Adds more tests to `ComponentPath`, because it has a much trickier implementation now (which must juggle a single string internally)
- Does not remove any of the existing tests for `ComponentPath`, only cleans them up (this refactor was performed before the implementation change)
- Does not break the public interface of `ComponentPath` - bar the fact that it can no longer be upcasted to a `Path` (because it doesn't exist)

There is *one* "breaking" change that I found in the Component interface (see diff), which is that the new implementation guarantess that a string will not end with a slash unless it was "/", this broke one existing tests, which seems to expect `../../../` to be a valid output.

adamkewley:
In terms of what this brings to the table, it's effectively a perf improvement that affects models that tend to traverse the component tree using paths at runtime:

comparing 8a4171e (an older `master` commit build I have lying around - pre CacheVariable merge) to a174085 (this PR, also based on an older master) seems to show a speedup (8 repeats on a Linux Ryzen1600 desktop):


|             ToyDropLanding |      14.41 |     0.05 |      13.54 |     0.01 |    1.06 |
|   ToyDropLanding_nomuscles |       0.45 |     0.00 |       0.45 |     0.00 |    0.99 |
|            passive_dynamic |       5.07 |     0.00 |       4.95 |     0.00 |    1.02 |
| passive_dynamic_noanalysis |       3.24 |     0.00 |       3.23 |     0.00 |    1.00 |
|                      Arm26 |       0.29 |     0.00 |       0.28 |     0.00 |    1.03 |
|             RajagopalModel |      16.73 |     1.31 |      14.99 |     0.03 |    1.12 |

I'll need to rebuild with the latest master to see what that speedup looks like when combined with the cache variable changes

adamkewley:
After rebuilding and taking a measurement with a bunch less noise (more repeats, leave the computer alone, etc.) I'm getting data that now has an alright stddev:

|             ToyDropLanding |      12.04 |     0.01 |      11.20 |     0.00 |    1.07 |
|            passive_dynamic |       4.79 |     0.00 |       4.62 |     0.00 |    1.04 |
| passive_dynamic_noanalysis |       3.01 |     0.00 |       3.01 |     0.00 |    1.00 |
|                      Arm26 |       0.27 |     0.00 |       0.26 |     0.00 |    1.04 |
|             RajagopalModel |      13.40 |     0.01 |      13.15 |     0.01 |    1.02 |

So this change will give a modest perf. increase in acouple of the sims, but nothing major, but I anticipate some futher perf. increases in the caller side by changing how this class is used (esp. tree traversal with it).

The main thing I want to explore is whether this can even be merged at all, so (sorry for dragging people into this) ignoring me needing to clean up the implementation file a little bit:

- This change is going to remove the `Path` class so that the implementation can use a concrete `PathComponent` class, which has no virtual overhead and produces fewer allocations. I would like to integrate further perf. increases on the caller side (e.g. not continually counting the number of components in a path as a `for` loop condition), but I want this PR to specifically focus on delivering a drop-in replacement for `PathComponent` that has all the same behaviors (apart from that one test case, above).

- @aymanhab will removing the `Path` class cause an issue in the GUI? There's the copied-in bindings to it in opensim-gui but I can't see any classes in the GUI using the bindings to the `OpenSim::Path` class
- @tgeijten are you using the `OpenSim::Path` class in SCONE? A quick `grep` says no (`xo/xo/utility/file_resource_cache.h` is the closest-looking thing, but seems unrelated)
- @chrisdembia same deal with you + Moco

Assuming those core projects aren't using it, I'm *guessing* we can remove it?

chrisdembia:
@adamkewley I would expect that these changes provide a speedup when initializing a model with `initSystem()`, but would not speed up a simulation (forward integration), but it seems that `ComponentPath` is being exercised even during a simulation.

It seems we should avoid `traverseToStateVariable()` being called during simulations, altogether. Most of the time, I think a component seeks the values of its own state variables, and so much of this searching logic is not necessary.

adamkewley:
@chrisdembia you are right, but my top-level goal in this pr is to implement a fairly isolated and easy-to-test change in an isolated part of the codebase with an intention to, afterwards, roll out client-side fixes. My question, in that context, is "do we need the Path virtual base class or can I scrap it?". 

I anticipate the implementation can be cleaned up a little now/later because this change mostly only affects one cpp file and has a bunch of tests.

What I mean is, what you're proposing is ultimately correct, but requires changing the runtime behavior of a core opensim class, such that it pre-perfrorms the lookups. I'm more hesitant to do that right now because that is a harder-to-test, slower to develop (recompiling on changing component's layout sucks) change that likely is also going to need to:

- potentially handle it on the caller side for the state variable API, much like how cache variables have been handled (this is a planned piece of work, but larger)
- handle the intermediate state between when things are added to the model, and internal references are strings, and when the model is initialized, and internal references are direct pointers
- do the above without making component's data bloat even worse by (eg by having to add a string-to-reference lookup table internally)
- handle what happens when a component is copied, making those references stale, such that the resulting runtime error is obvious to debug
- fix any unknown use-cases that rely on runtime tree traversal or path construction
- fix any other unknowns (because opensim is so large, and has downstream projects, this could be the actual pita)

Like, I get what you're suggesting, but the thing to keep in mind is that even implementing something as trivial-looking as the cache variable change took several weeks and iterations in order to implement it in a way that yields the desired speedup without breaking any existing--sometimes buggy--downstream code. 

If I could remove the constraint of "there's other people working with opensim and they might be doing things in technically incorrect but currently working way" then I'd make much more dramatic architectural changes than this :)

aymanhab:
Thanks for checking @adamkewley The GUI code base does not use the Path class anywhere.

adamkewley:
Great, so this is good to go?

chrisdembia:
@adamkewley we discussed this just now in our developer meeting. @aymanhab reminded us that we created the Path class for file paths because we were planning to create a `OpenSim::Property<Path>` class to better de/serialize file paths as part of OpenSim XML files (rather than using `OpenSim::Property<std::string>` as we do now). There are many examples of properties that are file paths, and using a user-defined `Path` type could handle system-dependent paths and resolve relative paths (avoiding the need for all the directory-changing code that you've seen). See https://github.com/opensim-org/opensim-core/issues/118

adamkewley:
The fact is that this class is *currently* exclusively used for component paths, not file paths. If there is some long-term ambition to change the properties of models that contain file paths to something like a `Property<FilePath>`, then what--*exactly*--is this change *actually* doing to prevent that hypothetical future change?

Do you feel that you *must* keep the `Path` abstraction hanging around just incase? Do you feel that porting some of this code (the majority of which is OpenSim-independent string manipulation) to a `FilePath` class later is going to be a large amount of work? Is the proposed change *really* imminent, given that `Path` was last edited in 2016 and has not been used anywhere since?

Like, really, this is a change that:

- Reduces the *active* LOC of the codebase (the LOC additions are mostly due to the increased number of tests)
- Removes a currently-unused abstration layer (`Path`)
- Makes some simulations 5-10 % faster without having to change any other part of the codebase
- May bring further perf. improvements once downstream code is edited a little

chrisdembia:
I should have been clearer; my intent was to give you background information about the `Path` class, not to make a decision about whether we should keep it around. I thought maybe leaving the `Path` class would be fine, because removing the class doesn't seem critical to your performance gain.

Let's remove `Path` with this PR. If we need it again, we can fish it out of the history.

Is this PR still WIP? If not, I'll start reviewing it.

adamkewley:
@chrisdembia no longer WIP: wasn't aware of the tag being significant

carmichaelong:
___
*[OpenSim/Common/Test/testComponentInterface.cpp, line 987 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2844#-METdMbt2ysjZmQow447:-MEYfYrc8P5sCiW_q0AB:b9vzi2y) ([raw file](https://github.com/opensim-org/opensim-core/blob/b5f3c03b37655ac16a47643df6b365e03d66b488/OpenSim/Common/Test/testComponentInterface.cpp#L987)):*

Could you explain why this is preferred over `../../../`? @carmichaelong, do you recall if there is a reason to prefer one over the other?
</blockquote></details>

I don't recall any reason for preferring one over the other. I would think as a user I'd expect behavior for both to be the same for a `ComponentPath`? This could point at a separate issue that I think this is a little ambiguous since we weren't super clear on the differences on paths ending with `/` and not when implementing this class.

adamkewley:
Merging this because it seems to tick all the boxes etc.

chrisdembia:
👍

Issue #2791, opened 2020-06-04
Support running IK on live data 


aymanhab:
As of now, the InverseKinematicsSolver caches the data that it is solving for, this limits the ability of clients to provide data on the fly which is needed for live data processing.

aymanhab:
One idea is to decouple the data producer (could be file/table or device) from the data consumer (solver) in this case. This approach was tried in https://github.com/RealTimeBiomechanics/rtosim.  A solution that enables the solver to read data from a Queue that could be populated from either a table or live data should solve this problem. Issues to consider:
- Thread safety /synchronization
- In case of multiple consumers with different speeds, how to make sure results are produced in order
- How to leverage the use of previous solution

jenhicks:
@aymanhab Is this issue resolved with the merge of #2855? Have updates been noted in the changelog?

jenhicks:
Waiting on feedback from Patrick

PR #2855, opened 2020-08-06
Refactor ik solver interface

aymanhab:
InverseKinematicsSolver used to keep a COPY of the Reference signals it is solving, this entangles the functionality of a producer and consumer of data, and precludes using the same interface for handling live data and data from files.  This PR tries to de-entangle by keeping only pointers to the Reference objects, delegating the responsibility of providing actual data to the Reference objects.
Fixes issue #2791

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855)


tkuchida:
@aymanhab Out of curiosity (and acknowledging that this PR is WIP, and also acknowledging that nobody asked for my opinion :wink:)… would it make any sense to abstract the data source ("producer") so that the Reference copying is still done but just not by the "consumer" InverseKinematicsSolver? IIRC, the purpose of copying the References was there's no guarantee that they won't be deleted by something else. Is this safety mechanism no longer required?

aymanhab:
Thanks for the feedback @tkuchida The plan is to have the solver keep a shared_ptr to the Reference object. This has the following benefits:
1. Being a smart/shared pointer it's managed by runtime and reference-counted so no need to artificially keep handles to avoid garbage collection., when reference count reaches 0 it will be deleted.
2. By using a pointer we rely on an interface (DataProvider) rather than a specific implementation of the Reference, and we can make the implementation do more elaborate work buffering/providing/adding live data without touching the solver in the future.

I sure look forward to hear your thoughts and feedback.

tkuchida:
Thanks for the clarification, @aymanhab. Sounds reasonable. It looked to me like removal of a guardrail to increase speed :racing_car:. It may be preferable to use entirely different classes (algorithms, memory management strategies, etc.) for real-time applications, for example if an algorithm wanted to use an entire motion trial to inform the IK solution (offline) vs. preallocating memory and strictly streaming data through a pipe (online). May not be relevant here so please ignore as appropriate.

aymanhab:
@chrisdembia I addressed your feedback, and modified parts of the interface accordingly. Appreciate your feedback. @carmichaelong if you have cycles to spend on this it would be great to help @chrisdembia prioritize his limited resource use. Thank you

carmichaelong (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855)* status: 12 of 25 files reviewed, 17 unresolved discussions (waiting on @aymanhab, @carmichaelong, and @chrisdembia)

___
*[Applications/IK/test/testLiveIK.cpp, line 86 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MEkQ3G804Pcoxmh9Lh0:-MFqeKzR3h5432JfYMvW:bq0qxkq) ([raw file](https://github.com/opensim-org/opensim-core/blob/2b1780d05139a4fed6d1152ef4cc7e9a6d770b90/Applications/IK/test/testLiveIK.cpp#L86)):*

The method is not generic since it ignore peliv_t* coordinates, we could expose that and make the move but not necessarily within the scope of this PR.
</blockquote></details>

Oh definitely, I agree that whatever this is pointed at by Reviewable now has some "hard-coded" coordinates, so this definitely shouldn't be added.
___
*[OpenSim/Common/DataQueue.h, line 43 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MFg-_A26Ly_0DXQPcEi:-MGFxEZX-ddnGKcoUtbX:b-a65rdl) ([raw file](https://github.com/opensim-org/opensim-core/blob/a4bd32fa31d968824a2441cf737d26141ea350b4/OpenSim/Common/DataQueue.h#L43)):*

Done.
</blockquote></details>

I'd suggest adding a comment about the timestamp then, to indicate that while it's not used internally, it can be useful for the consumers to keep the data ordered
___
*[OpenSim/Simulation/Reference.h, line 128 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MEnoF1b013RHEib5WUR:-MGFitNi61Ip6Pii13OH:b-xsvrjo) ([raw file](https://github.com/opensim-org/opensim-core/blob/2b1780d05139a4fed6d1152ef4cc7e9a6d770b90/OpenSim/Simulation/Reference.h#L128)):*

Done.
</blockquote></details>

I think the suggestion is that the outer-most parentheses can be removed:
`throw Exception(....`
___
*[OpenSim/Simulation/Reference.h, line 131 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MEnoMZgESFgiYvg_3gS:-MGFj2VY5LFKm58VCC1W:buy7jon) ([raw file](https://github.com/opensim-org/opensim-core/blob/2b1780d05139a4fed6d1152ef4cc7e9a6d770b90/OpenSim/Simulation/Reference.h#L131)):*

(replying on Reviewable to keep track of this)
</blockquote></details>

@aymanhab thoughts?

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855)* status: 12 of 25 files reviewed, 17 unresolved discussions (waiting on @aymanhab, @carmichaelong, and @chrisdembia)

___
*[Applications/IK/test/testLiveIK.cpp, line 86 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MEkQ3G804Pcoxmh9Lh0:-MGGCqFpEzN_op9qjrUY:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/2b1780d05139a4fed6d1152ef4cc7e9a6d770b90/Applications/IK/test/testLiveIK.cpp#L86)):*

Oh definitely, I agree that whatever this is pointed at by Reviewable now has some "hard-coded" coordinates, so this definitely shouldn't be added.
</blockquote></details>

Done.
___
*[OpenSim/Tools/MarkerPlacer.cpp, line 349 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MEkOQHD5MxS9OxJ-NfO:-MGH0h4eBB8oPfLExHwo:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/2b1780d05139a4fed6d1152ef4cc7e9a6d770b90/OpenSim/Tools/MarkerPlacer.cpp#L349)):*

Fixed
</blockquote></details>

Done.

carmichaelong:
@aymanhab I see a few unresolved issues on Reviewable that seem hard to keep track of, so I'll list them out here which I hope is helpful (see the comments on Reviewable at each file for more detail):
- DataQueue.h: add comment about timestamp
- DataQueue.h: memory leak check
- Reference.h: remove outer parenthesis for `throw`
- Reference.h: pure virtual vs virtual

aymanhab:
Thanks for the summary @carmichaelong I haven't finished responding to yours and Chris's review yet.

aymanhab:
> @aymanhab I see a few unresolved issues on Reviewable that seem hard to keep track of, so I'll list them out here which I hope is helpful (see the comments on Reviewable at each file for more detail):
> 
> * DataQueue.h: add comment about timestamp
> * DataQueue.h: memory leak check
> * Reference.h: remove outer parenthesis for `throw`
> * Reference.h: pure virtual vs virtual

1. Done
2. Was an issue when storing references, we copy the data so no special heap allocation
3. Done
4. Method made pure virtual in base class.

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855)* status: 12 of 25 files reviewed, 17 unresolved discussions (waiting on @carmichaelong and @chrisdembia)

___
*[Applications/IK/test/testLiveIK.cpp, line 139 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MEkTa-hBcljjzPQkdLN:-MGKYqfR0X_f-hwwsz2i:bfw21g4) ([raw file](https://github.com/opensim-org/opensim-core/blob/2b1780d05139a4fed6d1152ef4cc7e9a6d770b90/Applications/IK/test/testLiveIK.cpp#L139)):*

Sounds like we don't have a use case of multiple consumers yet? Might be good to add a TODO somewhere to document what's missing in the case we look back latter and need ot add support
</blockquote></details>

Added comment to DataQueue to explain timestamp and that consumers are responsible. Not ToDo because the changes/support will not be in these classes.
___
*[Applications/IK/test/testLiveIK.cpp, line 256 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MEkS0Qe27IhoTC4Mx5H:-MGKZBgP8V3A6VPCw24p:b-74kfiw) ([raw file](https://github.com/opensim-org/opensim-core/blob/2b1780d05139a4fed6d1152ef4cc7e9a6d770b90/Applications/IK/test/testLiveIK.cpp#L256)):*

Problem is that the method takes Model argument and the TableUtilities are in common library so it will fail to compile or add circular dependency.
</blockquote></details>

Will not do for the reason above.
___
*[OpenSim/Common/DataQueue.h, line 94 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MFg0dyu7JSApczO-3-B:-MGKZNM24BtN2YwuzhVM:b-mlw7o9) ([raw file](https://github.com/opensim-org/opensim-core/blob/a4bd32fa31d968824a2441cf737d26141ea350b4/OpenSim/Common/DataQueue.h#L94)):*

I'd like to see this verified before merging this PR. Feel free to acknowledge this thread and delete the TODO if this is already done
</blockquote></details>

This is a left over comments from the time data was allocated on heap and passed around, now it's copied so no need for special management. 
___
*[OpenSim/Simulation/Reference.h, line 128 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MEnoF1b013RHEib5WUR:-MGKZnXWCbei_yoE6gSU:b-623vyw) ([raw file](https://github.com/opensim-org/opensim-core/blob/2b1780d05139a4fed6d1152ef4cc7e9a6d770b90/OpenSim/Simulation/Reference.h#L128)):*

I think the suggestion is that the outer-most parentheses can be removed:
`throw Exception(....`
</blockquote></details>

Done, earlier comments were related to text of exception message  being incorrect.
___
*[OpenSim/Simulation/Reference.h, line 131 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MEnoMZgESFgiYvg_3gS:-MGKZwLf5nho_me0EMP2:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/2b1780d05139a4fed6d1152ef4cc7e9a6d770b90/OpenSim/Simulation/Reference.h#L131)):*

@aymanhab thoughts?
</blockquote></details>

Done.

aymanhab:
@carmichaelong I addressed your feedback and leftover issues from @chrisdembia, maybe another fresh look from @nickbianco would be useful at this point. At a high level would love to hear your thoughts on 
1. whether the subclasses of Reference_ are relevant and 
2. if so, whether the names can be improved. 
Once this is agreed upon, I'll make sure wrapping works and add a test case form python which would replicate the workflow used by Patrick and we'll be done. Thank you.

carmichaelong:
> whether the subclasses of Reference_
@aymanhab To make sure, I think you're asking about `ContinuousTimeReference_` and `DiscreteTimeReference_`

Seems like there's a clear distinction in the two classes, but going over the use cases I couldn't quite see the different details there. I'd defer to you on how useful it was to have both lying around (seems like you made good use of both), versus trying to "fit" multiple use cases to a single `Reference_` class, as before.

I agree that the naming could be improved, since it seems like the difference isn't about continuous vs discrete, but rather that there is some notion of "next". Maybe another question to help out some naming: Is the more important part of "continuous" that it can be evaluated at any arbitrary time based on some function (or interpolation)?

aymanhab:
@carmichaelong Thanks for the feedback, indeed the main difference in the interface is whether hasNext is supported and the actual method to getNext data. For the current "continuous" time case, the way it has been used, it never interpolated.  We were getting time from the table to go ask the table to search for that data at the same time. Ideally a Reference would be able to evaluate at arbitrary time in range so the decision to downsample/upsample would be made by the data consumer/client. 

That said, there are two distinctions that I'm trying to consolidate in naming which could be confusing:
1. The fact that data is open ended (Stream like behavior)
2. Where the data can be evaluated
May help to focus specifically on 1 and assume that 2 can be built into the base class Reference, so remove ContinuousTimeReference and make the DiscreteTimeReference into a StreamReference or a variant of the name. Thoughts? @nickbianco did you have a chance to follow the discussion? Thanks

nickbianco (changes requested):
I took a pass through `Reference.h` and am mostly following the discussion. I think removing ContinuousTimeReference and making `getValuesAtTime()` a pure virtual of the base class makes sense. Since `hasNext` is now pure virtual, you could possibly make `getNextValuesAtTime` pure virtual as well and have no StreamReference subclass, but I don't feel strongly about this (I only looked at Reference, just skimmed the other files).

> Reviewed 1 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-:-MGisxYb2GcCTVw6SdUj:bq3k5hb)* status: 13 of 25 files reviewed, 19 unresolved discussions (waiting on @aymanhab, @carmichaelong, and @chrisdembia)

___
*[Applications/IK/test/testLiveIK.cpp, line 75 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MGigKTIA-kuKYpihO8v:-MGigKTIA-kuKYpihO8w:bbqzwjh) ([raw file](https://github.com/opensim-org/opensim-core/blob/6f5d738dcad4e360bf6ec11d7cce043c30acdf0e/Applications/IK/test/testLiveIK.cpp#L75)):*
> ```cpp
>     // For simplicity, we ignore pelvis coordinates 0-5
>     for (size_t i = 0; i < mapStdToReport.size(); ++i) {
>         if (mapStdToReport[i] >= 0) {
> ```

The comment says we ignore pelvis coordinates 0-5, but I think this only skips the first index (so only the first pelvis rotation)? Should it not be `mapStdToReport[i] >= 2`?
___
*[OpenSim/Simulation/Reference.h, line 106 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MGiscRR7-tcRY-lrNDD:-MGiscRSBb2sDtZaRLZe:b-jbker0) ([raw file](https://github.com/opensim-org/opensim-core/blob/6f5d738dcad4e360bf6ec11d7cce043c30acdf0e/OpenSim/Simulation/Reference.h#L106)):*
> ```objc
>             double time, SimTK::Array_<T>& values) const = 0;
>     /* getValues as above, but a copy is returned, which may be costly */
>     virtual SimTK::Array_<T> getValues(double time) const {
> ```

Perhaps `getValuesAtTimeCopy()` would be clearer.

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855)* status: 10 of 25 files reviewed, 19 unresolved discussions (waiting on @carmichaelong, @chrisdembia, and @nickbianco)

___
*[Applications/IK/test/testLiveIK.cpp, line 75 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MGigKTIA-kuKYpihO8v:-MGtNSMc4dGxGcTVKmua:bipts04) ([raw file](https://github.com/opensim-org/opensim-core/blob/6f5d738dcad4e360bf6ec11d7cce043c30acdf0e/Applications/IK/test/testLiveIK.cpp#L75)):*

The comment says we ignore pelvis coordinates 0-5, but I think this only skips the first index (so only the first pelvis rotation)? Should it not be `mapStdToReport[i] >= 2`?
</blockquote></details>

Pelvis coordinates not in the map so they're all ignored. Clarified the comment
___
*[OpenSim/Simulation/InverseKinematicsSolver.cpp, line 460 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MEnnSyD87kqYYPtUf1e:-MGtNa674HRp6QwO7krS:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/2b1780d05139a4fed6d1152ef4cc7e9a6d770b90/OpenSim/Simulation/InverseKinematicsSolver.cpp#L460)):*

(replying on Reviewable to keep track of this)
</blockquote></details>

Done.
___
*[OpenSim/Simulation/Reference.h, line 65 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MEnnlncEOugK4ZN8dt9:-MGtNbni-9aLIYx3Sz_P:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/2b1780d05139a4fed6d1152ef4cc7e9a6d770b90/OpenSim/Simulation/Reference.h#L65)):*

I think some further change to the PR has left this comment in a place that makes no sense. Has this been resolved?
</blockquote></details>

Done.
___
*[OpenSim/Simulation/Reference.h, line 106 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MGiscRR7-tcRY-lrNDD:-MGtNdjz11HyUi423gHx:b-ovwif) ([raw file](https://github.com/opensim-org/opensim-core/blob/6f5d738dcad4e360bf6ec11d7cce043c30acdf0e/OpenSim/Simulation/Reference.h#L106)):*

Perhaps `getValuesAtTimeCopy()` would be clearer. 
</blockquote></details>

Trying not to change API, method existed for years, if you feel strongly please open a separate issue and we can scope it.

nickbianco (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855)* status: 10 of 25 files reviewed, 19 unresolved discussions (waiting on @aymanhab, @carmichaelong, @chrisdembia, and @nickbianco)

___
*[Applications/IK/test/testLiveIK.cpp, line 75 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MGigKTIA-kuKYpihO8v:-MGtXyTAE1bItwK3RVaj:bpxw071) ([raw file](https://github.com/opensim-org/opensim-core/blob/6f5d738dcad4e360bf6ec11d7cce043c30acdf0e/Applications/IK/test/testLiveIK.cpp#L75)):*

Pelvis coordinates not in the map so they're all ignored. Clarified the comment
</blockquote></details>

When the map is created above, it only skips pelvis translations (and I guess 'pelvis_tilt' since that would match the "pelvis_t" string), so pelvis rotation coordinates would be included in the map. Unless I'm missing something?
___
*[OpenSim/Simulation/Reference.h, line 106 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MGiscRR7-tcRY-lrNDD:-MGtabX6CzaZYT1nTu-K:bn4gu5t) ([raw file](https://github.com/opensim-org/opensim-core/blob/6f5d738dcad4e360bf6ec11d7cce043c30acdf0e/OpenSim/Simulation/Reference.h#L106)):*

Trying not to change API, method existed for years, if you feel strongly please open a separate issue and we can scope it.
</blockquote></details>

Ah okay, no I don't feel strongly about this.

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855)* status: 10 of 25 files reviewed, 19 unresolved discussions (waiting on @aymanhab, @carmichaelong, @chrisdembia, and @nickbianco)

___
*[Applications/IK/test/testLiveIK.cpp, line 75 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MGigKTIA-kuKYpihO8v:-MGtjO-SEAvQPX3UzkYl:b-pd4a8) ([raw file](https://github.com/opensim-org/opensim-core/blob/6f5d738dcad4e360bf6ec11d7cce043c30acdf0e/Applications/IK/test/testLiveIK.cpp#L75)):*

When the map is created above, it only skips pelvis translations (and I guess 'pelvis_tilt' since that would match the "pelvis_t" string), so pelvis rotation coordinates would be included in the map. Unless I'm missing something?
</blockquote></details>

Good point, this block is actually copied over from testIk, so will fix both places. Thanks for the find.

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855)* status: 10 of 25 files reviewed, 19 unresolved discussions (waiting on @aymanhab, @carmichaelong, @chrisdembia, and @nickbianco)

___
*[Applications/IK/test/testLiveIK.cpp, line 75 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MGigKTIA-kuKYpihO8v:-MGtjybjBXtEfxGTOG9R:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/6f5d738dcad4e360bf6ec11d7cce043c30acdf0e/Applications/IK/test/testLiveIK.cpp#L75)):*

Good point, this block is actually copied over from testIk, so will fix both places. Thanks for the find.
</blockquote></details>

Done.

aymanhab:
@carmichaelong One issue that comes with wrapping is that we can't have two constructors that differ only by changing an object to shared_ptr<object> since they produce identical signatures. We can either ifdef the old signatures out from wrapping or remove completely and the old code will work automagically in Matlab and Python but C++ users will need to upgrade. Let me know what you think. Thanks

carmichaelong (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855)* status: 9 of 26 files reviewed, 17 unresolved discussions (waiting on @aymanhab, @carmichaelong, @chrisdembia, and @nickbianco)

___
*[OpenSim/Simulation/Reference.h, line 131 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MEnoMZgESFgiYvg_3gS:-MGyejKH0ZNsyJv2KRhK:b7qq4lb) ([raw file](https://github.com/opensim-org/opensim-core/blob/2b1780d05139a4fed6d1152ef4cc7e9a6d770b90/OpenSim/Simulation/Reference.h#L131)):*

Done.
</blockquote></details>

I don't think this has been addressed yet (definitely let me know if I'm wrong). However, digging in deeper, if a method should be pure virtual, should it be in the`Reference_` class instead?

carmichaelong:
Thanks @aymanhab, it's getting close. Just a few points to check:
- Re constructors: I'm okay with only using the new constructor (forcing an interface only on C++ for a minor version release). I think we've done small API changes in the past on minor version, right?
- I think the pure virtual issue hasn't been addressed (but correct me if I'm wrong and missed it). Seems like the change might actually have to be in the parent class, not where the comment is right now?
- @nickbianco looks like your issue on the pelvis coordinates in the map has been resolved, but if you could just check that off real quick that'd be great

nickbianco (changes requested):
@carmichaelong thanks for the reminder. I left a response and a new comment.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-:-MGzBjYmDDvqFJ_EC7WB:b-kfl33k)* status: 9 of 26 files reviewed, 17 unresolved discussions (waiting on @aymanhab, @carmichaelong, @chrisdembia, and @nickbianco)

___
*[Applications/IK/test/testLiveIK.cpp, line 75 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MGigKTIA-kuKYpihO8v:-MGz9gUkCZbayIKhXxCw:bcvsrzl) ([raw file](https://github.com/opensim-org/opensim-core/blob/6f5d738dcad4e360bf6ec11d7cce043c30acdf0e/Applications/IK/test/testLiveIK.cpp#L75)):*

Done.
</blockquote></details>

Oh I see, we want to keep pelvis rotations. Satisfied here, but add another comment above.
___
*[Applications/IK/test/testLiveIK.cpp, line 64 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MGz91iw3w6Zhtz4P9Mu:-MGz91ix6MPUoDc4_e-z:b-f0331b) ([raw file](https://github.com/opensim-org/opensim-core/blob/94e5e6652e25cce2dbb2fd2468f1d36da3659d98/Applications/IK/test/testLiveIK.cpp#L64)):*
> ```cpp
>         if (found != stdLabels.end()) {
>             // skip any pelvis translations
>             if (found->find("pelvis_t") == std::string::npos || label.length()!=9) {
> ```

Do we document somewhere that pelvis translation coordinate labels must match "pelvis_tx", "pelvis_ty", "pelvis_tz"? I know that's the convention but this check would be easily broken with a slightly different label. Even if we do enforce that convention, I think I'd prefer something like this:

```c++
if (found->find("pelvis_tx") == std::string::npos &&
    found->find("pelvis_ty") == std::string::npos && 
    found->find("pelvis_tz") == std::string::npos) {
}
```

That, possibly along with an error message if more than three pelvis labels make it into the map.

aymanhab:
This isn't generic code, it's a test case with baked in files and model, so generality or other models are not an issue but I can change the test code if you think it helps.

nickbianco (commented):
@aymanhab I retracted the comment, forgot I was reviewing a test file.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-:-MGzM6hj6axRNICYMpET:b-oh02rv)* status: 9 of 26 files reviewed, 16 unresolved discussions (waiting on @aymanhab, @carmichaelong, @chrisdembia, and @nickbianco)

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855)* status: 9 of 26 files reviewed, 16 unresolved discussions (waiting on @carmichaelong, @chrisdembia, and @nickbianco)

___
*[OpenSim/Simulation/Reference.h, line 131 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MEnoMZgESFgiYvg_3gS:-MGz22n90aNojUplm65F:b-u6xvgo) ([raw file](https://github.com/opensim-org/opensim-core/blob/2b1780d05139a4fed6d1152ef4cc7e9a6d770b90/OpenSim/Simulation/Reference.h#L131)):*

I don't think this has been addressed yet (definitely let me know if I'm wrong). However, digging in deeper, if a method should be pure virtual, should it be in the`Reference_` class instead?
</blockquote></details>

Making the method pure virtual in Reference_ means that subclasses like CoordinateReference which do not support streaming at all, are required to implement it which feels wrong since actual functions that return the next values are not in Reference_.  Let me know what you think.
Ideally Reference_ would know nothing about Streaming, and we make StreamReference(s) using Mixins but the gymnastics to implement and the need to expose to scripting makes this option messy and confusing to most users.
___
*[OpenSim/Simulation/Reference.h, line 106 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MGiscRR7-tcRY-lrNDD:-MGz1XZt-dyTlmKOVsXU:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/6f5d738dcad4e360bf6ec11d7cce043c30acdf0e/OpenSim/Simulation/Reference.h#L106)):*

Ah okay, no I don't feel strongly about this.
</blockquote></details>

Done.

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855)* status: 9 of 26 files reviewed, 16 unresolved discussions (waiting on @carmichaelong, @chrisdembia, and @nickbianco)

___
*[Applications/IK/test/testLiveIK.cpp, line 75 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MGigKTIA-kuKYpihO8v:-MH0FtII0JAv8353pQbx:b396tty) ([raw file](https://github.com/opensim-org/opensim-core/blob/6f5d738dcad4e360bf6ec11d7cce043c30acdf0e/Applications/IK/test/testLiveIK.cpp#L75)):*

Oh I see, we want to keep pelvis rotations. Satisfied here, but add another comment above.
</blockquote></details>

Thanks, comment here is stale, check latest and let me know.

carmichaelong (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855)* status: 9 of 26 files reviewed, 16 unresolved discussions (waiting on @carmichaelong, @chrisdembia, and @nickbianco)

___
*[OpenSim/Simulation/Reference.h, line 131 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MEnoMZgESFgiYvg_3gS:-MH2PgRz5F4MGVGe7pwN:bhai6rb) ([raw file](https://github.com/opensim-org/opensim-core/blob/2b1780d05139a4fed6d1152ef4cc7e9a6d770b90/OpenSim/Simulation/Reference.h#L131)):*

Making the method pure virtual in Reference_ means that subclasses like CoordinateReference which do not support streaming at all, are required to implement it which feels wrong since actual functions that return the next values are not in Reference_.  Let me know what you think.
Ideally Reference_ would know nothing about Streaming, and we make StreamReference(s) using Mixins but the gymnastics to implement and the need to expose to scripting makes this option messy and confusing to most users.
</blockquote></details>

Ah I see, thanks for the clarification with `Reference_`. I now see why the comment is here instead. What are your thoughts on making it pure virtual here (in `StreamableReference_`)?

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855)* status: 9 of 26 files reviewed, 16 unresolved discussions (waiting on @carmichaelong, @chrisdembia, and @nickbianco)

___
*[OpenSim/Simulation/Reference.h, line 131 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MEnoMZgESFgiYvg_3gS:-MHCaBVd7DW-x1WUn82R:b-xql5dc) ([raw file](https://github.com/opensim-org/opensim-core/blob/2b1780d05139a4fed6d1152ef4cc7e9a6d770b90/OpenSim/Simulation/Reference.h#L131)):*

Ah I see, thanks for the clarification with `Reference_`. I now see why the comment is here instead. What are your thoughts on making it pure virtual here (in `StreamableReference_`)?
</blockquote></details>

It's totally possible, just need to implement it in OrientationsReference which increases the knowledge of OrientationsReference about its subclasses that I'd prefer reducing. Thoughts?

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855)* status: 9 of 26 files reviewed, 15 unresolved discussions (waiting on @carmichaelong, @chrisdembia, and @nickbianco)

___
*[OpenSim/Simulation/Reference.h, line 131 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MEnoMZgESFgiYvg_3gS:-MHDCANb-CR1ZgvVjmVk:b-21o8nq) ([raw file](https://github.com/opensim-org/opensim-core/blob/2b1780d05139a4fed6d1152ef4cc7e9a6d770b90/OpenSim/Simulation/Reference.h#L131)):*

It's totally possible, just need to implement it in OrientationsReference which increases the knowledge of OrientationsReference about its subclasses that I'd prefer reducing. Thoughts?
</blockquote></details>

Making hasNext() pure virtual in `StreamableReference_` make sense to me, since `getNextValuesAndTime()` is pure virtual there.

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855)* status: 9 of 26 files reviewed, 15 unresolved discussions (waiting on @carmichaelong, @chrisdembia, and @nickbianco)

___
*[OpenSim/Simulation/Reference.h, line 131 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-MEnoMZgESFgiYvg_3gS:-MHD_W0o3SrFn9Q5G4TR:b-1m1wcp) ([raw file](https://github.com/opensim-org/opensim-core/blob/2b1780d05139a4fed6d1152ef4cc7e9a6d770b90/OpenSim/Simulation/Reference.h#L131)):*

Making hasNext() pure virtual in `StreamableReference_` make sense to me, since `getNextValuesAndTime()` is pure virtual there.
</blockquote></details>

Ok, done

carmichaelong (commented):
Thanks @aymanhab it's all looking good to me. Assuming all tests pass, the only other thing to do would be to update `CHANGELOG.md`

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-:-MHDeaemCIWCAaDlpF13:bmczddc)* status: 9 of 26 files reviewed, 15 unresolved discussions (waiting on @carmichaelong, @chrisdembia, and @nickbianco)

carmichaelong (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2855#-:-MHIWKxj1zJ92zUCyjMI:bnfp4nl)* status: 9 of 27 files reviewed, 15 unresolved discussions (waiting on @carmichaelong, @chrisdembia, and @nickbianco)

aymanhab:
@pslade This PR/branch has the refactoring of the IK solver interface that should allow you to reproduce the results from the previous codebase we shared. The testLiveIK.cpp test case shows the syntax. If you can test this branch in your environment before we merge it that would be great. Thank you.

aymanhab:
Merging per verbal approval in dev meeting. Thanks @carmichaelong @nickbianco

Issue #2714, opened 2020-04-08
Move Moco's DeGrooteFregly2016Muscle into opensim-core


chrisdembia:
Moco contains a new muscle model, DeGrooteFregly2016Muscle, that belongs in this repository. However, we must improve this muscle's support for CMC and forward simulations (using explicit differential equations), and improve test coverage.

nickbianco:
This has been merged via #2793 and #2859.

PR #2793, opened 2020-06-05
Move DeGrooteFregly2016Muscle from Moco to OpenSim

nickbianco:
Addresses issue #2714 

### Brief summary of changes
- Brought in the DeGrooteFregly2016Muscle class files from Moco
- Updated header files for bindings

### Testing I've completed
- Nothing yet, this PR is for review the meat of the DGF muscle code.

### Looking for feedback on...
- We rely on many utilities in Moco to use this class (e.g., `createVectorLinspace()`, `writeTableToFile`, etc) that are not included currently in this PR. Some of these utilities might fit better in a PR to master (e.g., `format()`, which is related to logging stuff).

### CHANGELOG.md (choose one)

- no need to update because...merging to feature branch.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2793)

PR #2859, opened 2020-08-13
Support for DGF muscle tendon compliance in CMC and forward simulations 

nickbianco:
Fixes issue #2714

### Brief summary of changes
- Incorporated all changes implemented in branch `feature_dgf_muscle` (already reviewed by @carmichaelong via #2793)
- `DeGrooteFregly2016Muscle` now properly computes muscle-tendon equilibrium via `calcEquilibriumResidual()`, which is used by `computeInitialFiberEquilibrium()`. It previously always produced a zero residual since it relied on advancing the SimTK::Stage, but now it is independent of Stage.
- Created new test cases to test that `DeGrooteFregly2016Muscle` with tendon compliance enabled can produce the correct trajectory and/or muscle activity when using forward integration or CMC
     - Includes tests in `testDeGrooteFregly2016Muscle.cpp` and `testMuscles.cpp`
- `MocoInitialForceEquilibriumGoal` is now `MocoInitialForceEquilibriumDGFGoal` with changes that are specific to the new changes in how the DGF muscle computes muscle-tendon equilbrium. 

In `CMCTool`, I wrapped a call to `Muscle::computeEquilibrium()` with an existing bool property since it was causing the CMC tests I added to fail when disabling activation dynamics. @chrisdembia, @carmichaelong, and I discussed this during the dev meeting and were not sure if CMC generally supports problems with activation dynamics disabled anyway, so we may choose to punt on these tests for now.

### CHANGELOG.md (choose one)

- no need to update because...still part of Moco feature branch.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2859)


nickbianco:
After quickly going through the CMC tests, it doesn't look like we have tests that ignore activation dynamics.

OpenSim/Actuators/DeGrooteFregly2016Muscle.cpp
```diff
@@ -279,9 +299,8 @@ void DeGrooteFregly2016Muscle::calcMuscleLengthInfoHelper(
 void DeGrooteFregly2016Muscle::calcFiberVelocityInfoHelper(
         const SimTK::Real& muscleTendonVelocity, const SimTK::Real& activation,
         const bool& ignoreTendonCompliance,
-        const bool& isTendonDynamicsExplicit,
-        const MuscleLengthInfo& mli, FiberVelocityInfo& fvi,
-        const SimTK::Real& normTendonForce = SimTK::NaN,
+        const bool& isTendonDynamicsExplicit, const MuscleLengthInfo& mli,
```
adamkewley, line 302: Neither a review, nor blocking, nor do you have to answer, but why is it that devs default to using `const T&` for primitive value types like `bool` or `SimTK::Real`?

chrisdembia: No good reason. We probably should pass such primitive types by value.

chrisdembia:
DGF + CMC :)

PR #2861, opened 2020-08-17
Move DGF default parameters to function declarations.

chrisdembia:
@nickbianco, this PR moves some default parameters from function definitions to declarations. If you had a reason for putting the default parameters with the function definitions, then I'm fine with dropping this PR. But I think it's more common to place the default parameters in the declaration because they're part of the interface.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2861)


adamkewley (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2861)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @chrisdembia and @nickbianco)

___
*[OpenSim/Actuators/DeGrooteFregly2016Muscle.h, line 286 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2861#-MF4gZf03kfIMHQP9YS0:-MF4gZf29D8p18bPs1zK:b-5udeoh) ([raw file](https://github.com/opensim-org/opensim-core/blob/934dcef71dffc8f54d848f8fc0421337dc22b0fd/OpenSim/Actuators/DeGrooteFregly2016Muscle.h#L286)):*
> ```objc
>     /// tendon_compliance_dynamics_mode is 'explicit', this gets the value
>     /// returned by getStateVariableDerivativeValue() for the
>     /// 'normalized_tendon_force' state.
> ```

As a non-blocking drive-by, I'd maybe enumerate preconditions like these in a bullet-point list, rather than a paragraph, because another developer's just going to want a quick blow-by-blow of what preconditions the API is setting up (same with the comments added in the header)

chrisdembia:
Ready for review again.

chrisdembia:
Thanks for the review.

PR #2863, opened 2020-08-19
Moved OpenSim::Component::getSystem() implementation into header

adamkewley:
This is a micro-commit that inlines `Component::getSystem`, by moving the implementation into the header.

I would not normally inline functions--the benefits are usually overstated when compared with the compile-time increases--but, in this case, `getSystem` is called *a lot* internally in OpenSim, so this tiny change should improve perf by 1 % (small gains, really, but it's 1 % for doing basically nothing).

An example of a valgrind call graph that I'd see when measuring things would be this:

image

Where `getSystem` is called 115M times over a typical FD simulation (ToyDropLanding). Measuring this change under my usual profiling rig shows some tiny changes:

|             ToyDropLanding |      11.33 |     0.00 |      11.25 |     0.00 |    1.01 |
|            passive_dynamic |       4.64 |     0.00 |       4.61 |     0.00 |    1.01 |
| passive_dynamic_noanalysis |       3.00 |     0.00 |       3.01 |     0.00 |    1.00 |
|                      Arm26 |       0.26 |     0.00 |       0.26 |     0.00 |    1.01 |
|             RajagopalModel |      13.37 |     0.01 |      13.11 |     0.01 |    1.02 |

But I'd take these with a pinch of salt because it's close to the measurement noise.

Either way, we inline *a lot* of other stuff that isn't called nearly as often, so I'm hoping this change is not sacrosanct, given the measurements.

---

- Enables inlining of the function call by other compilation units

- OpenSim::Component::getSystem() is one of the most heavily-called
  functions in OpenSim. For example, an FD simulation of ToyDropLanding
  calls this function >100M times, accounting for ~0.5-1 % of CPU
  usage

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2863)

### DIFF
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -1540,12 +1540,6 @@ void Component::extendRealizeAcceleration(const SimTK::State& s) const
     }
 }
 
-const SimTK::MultibodySystem& Component::getSystem() const
-{
-    OPENSIM_THROW_IF_FRMOBJ(!hasSystem(), ComponentHasNoSystem);
-    return _system.getRef();
-}
-
 SimTK::MultibodySystem& Component::updSystem() const
 {
     OPENSIM_THROW_IF_FRMOBJ(!hasSystem(), ComponentHasNoSystem);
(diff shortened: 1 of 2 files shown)


chrisdembia:
Neat! 👍

Issue #2866, opened 2020-08-25
Tests fail in debug mode on windows


aymanhab:
This seems related to the refactoring of the ComponentPath as shown in the stacktrace below
`>	osimCommon_d.dll!`anonymous namespace'::normalize(std::string path) Line 140	C++
 	osimCommon_d.dll!OpenSim::ComponentPath::ComponentPath(std::string path) Line 252	C++
 	osimCommon_d.dll!OpenSim::AbstractSocket::checkConnecteePathProperty() Line 281	C++
 	osimCommon_d.dll!OpenSim::Component::finalizeFromProperties() Line 216	C++
 	osimCommon_d.dll!OpenSim::Component::componentsFinalizeFromProperties() Line 288	C++
 	osimCommon_d.dll!OpenSim::Component::finalizeFromProperties() Line 233	C++
 	osimCommon_d.dll!OpenSim::Component::componentsFinalizeFromProperties() Line 288	C++
 	osimCommon_d.dll!OpenSim::Component::finalizeFromProperties() Line 233	C++
 	osimSimulation_d.dll!OpenSim::Model::Model() Line 86	C++
 	osimSimulation_d.dll!RegisterTypes_osimSimulation() Line 148	C++
 	osimSimulation_d.dll!osimSimulationInstantiator::registerDllClasses() Line 307	C++
 	osimSimulation_d.dll!osimSimulationInstantiator::osimSimulationInstantiator() Line 302	C++
 	osimSimulation_d.dll!`dynamic initializer for 'instantiator''() Line 136	C++
`
This happens when path=""

aymanhab:
@adamkewley When you have a chance can you run tests on Windows in Debug mode? If not what's the expected behavior when normalize is fed an empty string? returning the empty string causes problems downstream. It's not clear what assumptions are made about ComponentPath now. Thank you.

adamkewley:
@aymanhab sure, I'll have a look. Can't be until next Monday though (holiday atm). I thought I emulated the behavior of empty strings to match the previous implementation. Is this specifically debug mode in Windows?

aymanhab:
Thanks @adamkewley Indeed this is only in Windows/Debug mode. STL libraries/classes on windows have extra checks in debug mode that are tripping at corner cases (e.g. "cannot seek string iterator after end"). Will try to work around these until you're back. Enjoy the holiday 🍺

adamkewley:
[...]

PR #2870, opened 2020-08-31
Fix ComponentPath when running in MSVC with debug mode

adamkewley:
Proposed fix for #2866 

@aymanhab noticed that OpenSim `master` tests are currently broken when building with MSVC in debug mode. The reason they are broken is because my implementation of ComponentPath, which was merged in PR #2844 , relied on behavior like:

```c++
std::string s = "";
assert(s.front() == '\0');
assert(*s.begin() == '\0');
assert(*s.rbegin() == '\0');
```

Which is classed as undefined behavior by MSVC. It has assertions for detecting when code tries to do things like this. Those assertions are explained here:

https://docs.microsoft.com/en-us/cpp/standard-library/debug-iterator-support?view=vs-2019

The reason why I figured this kind of code is valid is because C++11 specifically redefined that `std::string` must be a contiguous sequence of `charT`, followed by a value-initialized terminator (`char()`, which `== '\0'`). Cppreference mentions this in *some* of the docs:

> https://en.cppreference.com/w/cpp/string/basic_string/data
> (until C++11) The returned array is not required to be null-terminated. If empty() returns true, the pointer is a non-null pointer that should not be dereferenced. 
> (since C++11) The returned array is null-terminated, that is, data() and c_str() perform the same function. If empty() returns true, the pointer points to a single null character. 

And that, specifically, `operator[]` returns a nul terminator:

> https://en.cppreference.com/w/cpp/string/basic_string/operator_at
> (since C++11) If pos == size(), a reference to the character with value CharT() (the null character) is returned. For the first (non-const) version, the behavior is undefined if this character is modified to any value other than CharT() .

But does not specify this behavior for functions like `.front()` (which *speficially* says that the return value is undefined if the string is empty), and it does not specify this behavior when dereferencing the iterators, which is what MSVC is (probably, rightly) whining about. 

Because I'm busy/lazy, one solution I tried was to just:

```c++
// ComponentPath.cpp
#define _ITERATOR_DEBUG_LEVEL 0 
```

But that causes other compile errors (effectively, MSVC, VS, etc. start thinking that I'm trying to link object files created with different flags and dies). It is also hacky as hell, so I decided to do this "properly", by reimplementing the `ComponentPath` algorithms to specifically use access patterns that are correct in C++11, such as:

```c++
char* begin = &str[0];  // can be `.data()` in C++17
char* end = &str[str.size()];

while (begin != end) {
    if (begin[1] == '\0') {
        // lookahead is NUL, do something
    }
    ++begin;
}
```

This effectively shifts the implementation to be more C-like, than C++-like, because the algorithms now traverse C-string pointers etc. To combat that, I spent a fair bit of time trying to simplify the tricker bits into helper functions, or a more readable form, and the proposed fix--I believe--is as easy to follow as the version that relied on C++ iterators

(moreso imho, but I like C, and pinapple on pizzas, so :shrug:)

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2870)

### DIFF
--- a/OpenSim/Common/ComponentPath.cpp
+++ b/OpenSim/Common/ComponentPath.cpp
@@ -120,12 +120,12 @@ namespace {
             OPENSIM_THROW(OpenSim::Exception, path + ": The supplied path contains invalid characters.");
         }
 
-        // pathEnd is guaranteed to be a NUL terminator in >C++11
+        // pathEnd is guaranteed to be a NUL terminator since C++11
         char* pathBegin = &path[0];
         char* pathEnd = &path[path.size()];
 
-        // helper: shift n chars starting at newStart+n such
-        // that they start at newStart
+        // helper: shift n chars starting at newStart+n such that, after,
+        // newStart..end is equal to what newStart+n..end was before.
         auto shift = [&](char* newStart, size_t n) {
             std::copy(newStart + n, pathEnd, newStart);
             pathEnd -= n;
@@ -133,6 +133,10 @@ namespace {
 
         // helper: grab 3 lookahead chars, using NUL as a senteniel to
         // indicate "past the end of the content".
+        //
+        // - The maximum lookahead is 3 characters because the parsing
+        //   code below needs to be able to detect the upcoming input 
+        //   pattern "..[/\0]"
         struct Lookahead { char a, b, c; };
         auto getLookahead = [](char* start, char* end) {
             return Lookahead{
@@ -151,23 +155,18 @@ namespace {
             }
         }
 
-        // skip past absolute separator (if any)
         bool isAbsolute = *pathBegin == separator;
-
-        // current position of alg. cursor in the string
-        char* cur = isAbsolute ? pathBegin + 1 : pathBegin;
+        char* cursor = isAbsolute ? pathBegin + 1 : pathBegin;
 
         // skip/dereference relative elements *at the start of a path*
-        for (Lookahead l = getLookahead(cur, pathEnd);
-             l.a == '.';
-             l = getLookahead(cur, pathEnd)) {
-
+        Lookahead l = getLookahead(cursor, pathEnd);
+        while (l.a == '.') {
             switch (l.b) {
             case separator:
-                shift(cur, 2);
+                shift(cursor, 2);
                 break;
             case nul:
-                shift(cur, 1);
+                shift(cursor, 1);
                 break;
             case '.': {
                 if (l.c == separator || l.c == nul) {
@@ -181,24 +180,26 @@ namespace {
                     // past these elements because the alg can't reduce
                     // them down
                     if (l.c == separator) {
-                        cur += 3;
+                        cursor += 3;
                     } else {
-                        cur += 2;
+                        cursor += 2;
                     }
                 } else {
                     // normal element that starts with '..'
-                    ++cur;
+                    ++cursor;
                 }
                 break;
             }
             default:
                 // normal element that starts with '.'
-                ++cur;
+                ++cursor;
                 break;
             }
+
+            l = getLookahead(cursor, pathEnd);
         }
 
-        char* contentStart = cur;
+        char* contentStart = cursor;
 
         // invariants:
         //
@@ -211,37 +212,37 @@ namespace {
         // - `[contentStart..offset] is the normalized *content* of the path
         //   string
 
-        while (cur < pathEnd) {
-            Lookahead l = getLookahead(cur, pathEnd);
+        while (cursor < pathEnd) {
+            Lookahead l = getLookahead(cursor, pathEnd);
 
             if (l.a == '.' && (l.b == nul || l.b == separator)) {
                 // handle '.' (if found)
                 size_t charsInCurEl = l.b == separator ? 2 : 1;
-                shift(cur, charsInCurEl);
+                shift(cursor, charsInCurEl);
 
             } else if (l.a == '.' && l.b == '.' && (l.c == nul || l.c == separator)) {
                 // handle '..' (if found)
 
-                if (cur == contentStart) {
+                if (cursor == contentStart) {
                     OPENSIM_THROW(OpenSim::Exception, path + ": cannot handle '..' element in a path string: dereferencing this would hop above the root of the path.");
                 }
 
                 // search backwards for previous separator
-                char* prevSeparator = cur - 2;
+                char* prevSeparator = cursor - 2;
                 while (prevSeparator > contentStart && *prevSeparator != separator) {
                     --prevSeparator;
                 }
 
                 char* prevStart = prevSeparator <= contentStart ? contentStart : prevSeparator + 1;
                 size_t charsInCurEl = (l.c == separator) ? 3 : 2;
-                size_t charsInPrevEl = cur - prevStart;
+                size_t charsInPrevEl = cursor - prevStart;
 
-                cur = prevStart;
-                shift(cur, charsInPrevEl + charsInCurEl);
+                cursor = prevStart;
+                shift(cursor, charsInPrevEl + charsInCurEl);
 
             } else {
                 // non-relative element: skip past the next separator or end
-                cur = std::find(cur, pathEnd, separator) + 1;
+                cursor = std::find(cursor, pathEnd, separator) + 1;
             }
         }
 
--- a/OpenSim/Common/ComponentPath.cpp
+++ b/OpenSim/Common/ComponentPath.cpp
@@ -27,17 +27,16 @@
 #include <algorithm>
 #include <sstream>
 
-
 using OpenSim::ComponentPath;
 
 namespace {
     const std::string newInvalidChars{"\\*+ \t\n"};
-    const char separator = '/';
+    constexpr char separator = '/';
     const std::string legacyInvalidChars = newInvalidChars + separator;
+    constexpr std::string::value_type nul = {};
 
     /**
-     * helper: joins the supplied vector of path components to form a single
-     * path string.
+     * Returns all components in `pathVec` joined by the path separator
      */
     std::string stringifyPath(const std::vector<std::string>& pathVec,
                               bool isAbsolute) {
@@ -46,22 +45,26 @@ namespace {
         if (isAbsolute) {
             ret += separator;
         }
+
         if (pathVec.empty()) {
             return ret;
         }
+
         for (size_t i = 0; i < (pathVec.size()-1); ++i) {
-            ret += pathVec[i] + separator;
+            ret += pathVec[i];
+            ret += separator;
         }
         ret += pathVec.back();
+
         return ret;
     }
 
     /**
      * helper: returns a string iterator that points to the first component
-     * (i.e. non-separator) in a normalized path string. Points to `.end()`
+     * (i.e. non-separator) in a normalized path string. Returns `.end()`
      * if the path contains no components.
      */
-    std::string::const_iterator firstComponent(const std::string& normalizedPath) {
+    std::string::const_iterator firstComponentIn(const std::string& normalizedPath) {
         if (normalizedPath.empty()) {
             return normalizedPath.end();
         }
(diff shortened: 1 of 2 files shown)


aymanhab:
Thanks @adamkewley for the thorough investigation, I sure feel your pain. 
Will review the code change today but I'm wondering whether you have Windows debuggable environment to test with or if I should switch to your branch to review/test. We could also set a github action to do this (as needed) but that opens the door for more issues (timeout, whole stack needs to be debuggable, etc.). Let me know what you think.

adamkewley:
Running this locally in debug mode passes `ComponentPath` tests (fairly comprehensive). However, there's test failures that need to be investigated. Any path cases that are behaving incorrectly in these tests should be added to the primary `ComponentPath` test suite for future dev:

```
1>47/93 Test #47: testVisualization ............................***Failed    0.57 sec
...
1>[warning] Model 'model' has subcomponents with duplicate name 'ground_offset_frame'. The duplicate is being renamed to 'ground_offset_frame_0'.
1>[warning] Model 'model' has subcomponents with duplicate name 'ground_offset_frame'. The duplicate is being renamed to 'ground_offset_frame_1'.
1>Failed to connect Socket 'parent' of type PhysicalFrame (details: Connectee for Socket 'parent' of type PhysicalFrame in PhysicalOffsetFrame at /ground_offset_frame is unspecified. If this model was built programmatically, perhaps finalizeConnections() was not called before printing.
1>	Thrown at Component.h:3344 in finalizeConnection().).
1>	In Object 'ground_offset_frame' of type PhysicalOffsetFrame.
1>	Thrown at Component.cpp:313 in finalizeConnections().


1>      Start 49: exampleHopperDeviceAnswers
1>49/93 Test #48: exampleHopperDevice ..........................***Failed    1.68 sec
1>Hopper Example Failed to run due to the following Exception: Component 'Dennis' of type Model is the root but has no subcomponents listed.
1>Verify that finalizeFromProperties() was called on this Component to identify its subcomponents.
1>	Thrown at Component.cpp:1751 in initComponentTreeTraversal().


1>      Start 50: testComponents
1>50/93 Test #49: exampleHopperDeviceAnswers ...................***Failed    1.75 sec
1>Hopper Example Failed to run due to the following Exception: Component 'Dennis' of type Model is the root but has no subcomponents listed.
1>Verify that finalizeFromProperties() was called on this Component to identify its subcomponents.
1>	Thrown at Component.cpp:1751 in initComponentTreeTraversal().
1>

1>82/93 Test #21: testContactGeometry ..........................***Failed  1243.67 sec
...
1>Exception:
1>  Failed to connect Socket 'child_frame' of type PhysicalFrame (details:
1>  Connectee for Socket 'child_frame' of type PhysicalFrame in PinJoint at
1>  /jointset/hinge is unspecified. If this model was built programmatically,
1>  perhaps finalizeConnections() was not called before printing.
1>  Thrown at Component.h:3344 in finalizeConnection().).
1>  In Object 'hinge' of type PinJoint.
1>  Thrown at Component.cpp:313 in finalizeConnections().

1>93/93 Test #92: testOpenSense ................................***Timeout 1500.02 sec
1>[info] Loaded model Subject07 from file subject07.osim
...


1>The following tests FAILED:
1>	 21 - testContactGeometry (Failed)
1>	 47 - testVisualization (Failed)
1>	 48 - exampleHopperDevice (Failed)
1>	 49 - exampleHopperDeviceAnswers (Failed)
1>	 92 - testOpenSense (Timeout)
1>Errors while running CTest
```

adamkewley:
@aymanhab I've set up VS2019 today for building OpenSim /w debug etc, so you shouldn't need to check this out once I've dug through the 5 tests failing above (they might be failing for unrelated reasons, but I'll try and check them tomorrow). This might also need the usual styling checks etc.

It looks like the change gets through CI tests and most of the MSVC debug mode tests alright (apart from those 5), so I doubt I will functionally change it much

adamkewley:
Thanks for the comments - I'll look into them tomorrow.

Wrt string tokenization etc. I guess if the implementation wanted to be
simple, the resulting token steam would end up in some other collection (eg
a vector) and be a union of separator, current dir, previous dir, or a
concrete element (span pointers to the source string). The existing
implementation (kind of, no tokenization) did this by splitting the input
on the delimiter (with the implicit encoding that the separators are
between the elements in that vector), followed by handling relative
elements in the vector.

The main goal of this (trickier) implementation is to be as allocation-free
as possible, because profiling indicated that the ComponentPath ctor can be
called millions of times (another, separate, issue to fix) and was
thrashing the allocator. The new implementation also cheats by knowing that
SSO causes typical inputs skip heap allocation altogether (vectors don't do
this, and typically heap allocate).

To avoid allocations, I tried strategies like implementing a lazy-evaluated
lexer that emits one token at a time in a stream, but that didn't really
work because the implementation fundamentally needs dynamically-sized
auxiliary storage for a "stack" of path elements that may be popped
repeatably (hello, dynamic allocation). This low-level implementation
effectively encodes the stack into the result string as the algorithm
progresses by performing the "pop" directly on the string (i.e. searching
left in the string for the element to pop). It's currently multi-pass for
simplicity (this would be a problem for large strings due to cache misses,
but not these short strings) and the only dynamic storage it uses (for now)
is the input string.

adamkewley (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2870)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @aymanhab)

*[a discussion](https://reviewable.io/reviews/opensim-org/opensim-core/2870#-MG4gDvR-QYcZshrJCLn:-MG7d2FNEtLR5zgCJp5C:bolmadg) (no related file):*

I was expecting you'd tokenize the string (and along the way merge separators, combine '..', handle corner cases etc.) in one pass, then process tokens in another pass but that maybe an overkill here especially if the goal is improved performance (disclaimer I didn't review the previous PR). The code changes look reasonable, I mainly asked for clarifying comments/documentation. I would not merge this PR until we get to the bottom of the failures then merge all the fixes at once. Thanks @adamkewley 
</blockquote></details>

(replied to why tokenization might be tricky for this in the PR: effectively, because auxiliary storage is needed for the path stack)

I'll look into the failures now. Unsure if they're specifically new because of these changes (we don't have a CI history for MSVC debug builds), so I may need to bisect a little to figure it out. I'll update this PR when I know more

---
___
*[OpenSim/Common/ComponentPath.cpp, line 137 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2870#-MG4VCbe1vkumJfPAyhI:-MG7iZhM4-1fb_OM8I3W:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/e03c543cf25600e8052c71ea407bef08997f6e8a/OpenSim/Common/ComponentPath.cpp#L137)):*

Some comment regarding why we need 3 long lookahead would be helpful
</blockquote></details>

Done.
___
*[OpenSim/Common/ComponentPath.cpp, line 159 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2870#-MG4W92FEXs1-OLrNg4m:-MG7i_Kf7opV6FtYEy74:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/e03c543cf25600e8052c71ea407bef08997f6e8a/OpenSim/Common/ComponentPath.cpp#L159)):*

cur -> cursor?
</blockquote></details>

Done.

adamkewley:
So these tests seem to have been broken for a while now, independent of the string iterator errors addressed above.

A clean checkout + build of OpenSim `4.1` (git tag) with these steps (quite manual, but hopefully representitive of a typical Windows developer workflow):

- Use CMake GUI to config + generate `dependencies/` (superbuild)
- Run `BUILD_ALL` target in VS2019 (works)
- Use CMake GUI to config + generate `opensim-core/`, making sure that it points to the newly-generated Simbody install
- Run `BUILD_ALL` target in VS2019 (works)
- Run `TEST_ALL_PARALLEL`: **fails** with these errors (similar to above, interspersed a little due to threading):

```
1>40/93 Test #47: testVisualization ............................***Failed    0.63 sec
1>Will try loading debug library first
...
1>Length:3181vs.3181
1>BuiltinGeometry Passed
1>Model 'model' has subcomponents with duplicate name 'ground_offset_frame'.
1>The duplicate is being renamed to 'ground_offset_frame_0'.
1>Model 'model' has subcomponents with duplicate name 'ground_offset_frame'.
1>The duplicate is being renamed to 'ground_offset_frame_1'.
1>Failed to connect Socket 'parent' of type PhysicalFrame (details: Connectee for Socket 'parent' of type PhysicalFrame in PhysicalOffsetFrame at /ground_offset_frame is unspecified. If this model was built programmatically, perhaps finalizeConnections() was not called before printing.
1>	Thrown at Component.h:3094 in finalizeConnection().).
1>	In Object 'ground_offset_frame' of type PhysicalOffsetFrame.
1>	Thrown at Component.cpp:314 in finalizeConnections().

1>42/93 Test #48: exampleHopperDevice ..........................***Failed    1.48 sec
1>Hopper Example Failed to run due to the following Exception: Component 'Dennis' of type Model is the root but has no subcomponents listed.
1>Verify that finalizeFromProperties() was called on this Component to identify its subcomponents.
1>	Thrown at Component.cpp:1686 in initComponentTreeTraversal().

1>      Start 50: testComponents
1>43/93 Test #49: exampleHopperDeviceAnswers ...................***Failed    1.42 sec
1>Hopper Example Failed to run due to the following Exception: Component 'Dennis' of type Model is the root but has no subcomponents listed.
1>Verify that finalizeFromProperties() was called on this Component to identify its subcomponents.
1>	Thrown at Component.cpp:1686 in initComponentTreeTraversal().

1>78/93 Test #21: testContactGeometry ..........................***Failed  939.10 sec
1>Exception:
1>  Failed to connect Socket 'child_frame' of type PhysicalFrame (details:
1>  Connectee for Socket 'child_frame' of type PhysicalFrame in PinJoint at
1>  /jointset/hinge is unspecified. If this model was built programmatically,
1>  perhaps finalizeConnections() was not called before printing.
1>  Thrown at Component.h:3094 in finalizeConnection().).
1>  In Object 'hinge' of type PinJoint.
1>  Thrown at Component.cpp:314 in finalizeConnections().

1>      Start 93: testCommandLineInterface
1>85/93 Test #56: testOptimizationExampleRuns ..................***Timeout 1500.07 sec

1>86/93 Test #57: testOptimizationExample ......................***Failed    1.00 sec
1>
1>Exception:
1>  Can't open optimization result file
1>  file= C:\Users\ak\Desktop\opensim-core\OpenSim\Tests\OptimizationExample_Arm26\testOptimizationExample.cpp
1>  line= 60

1>Total Test time (real) = 2757.06 sec
1>
1>The following tests FAILED:
1>	 21 - testContactGeometry (Failed)
1>	 47 - testVisualization (Failed)
1>	 48 - exampleHopperDevice (Failed)
1>	 49 - exampleHopperDeviceAnswers (Failed)
1>	 56 - testOptimizationExampleRuns (Timeout)
1>	 57 - testOptimizationExample (Failed)
1>	 85 - testCMCGait10dof18musc (Timeout)
1>	 90 - testCMCWithControlConstraintsRunningModel (Timeout)
```

So it *seems* like there's an issue with MSVC debug builds in OpenSim in general (at least on my setup, which is a standard Windows 10 install /w VS2019 + cmake). However, I might be mis-building something, or forgetting a flag somewhere (I don't typically develop in Windows), so it might be an idea @aymanhab  to quickly (independently) checkout `4.1` and do a clean debug build + test run and confirm that everything passes.

adamkewley:
Before trying the above (which should've been step one, probably, :wink:), I tried stepping through the code for `exampleHopperDevice` to understand why it's failing. 

There's a lot of (the usual) hopping (ahem) around inside various functions in `Component` going on. I'm still not entirely sure why it's breaking but, from what I understand, it seems like:

- It initially fails because `finalizeProperties` isn't called on the built hopper, but some methods depend on that being called
- Adding `finalizeProperties` gets futher through the test. Notably, other steps in the test that (e.g.) traverse the component tree and grab a subcomponent via a path seem to work fine (which would be the steps that are particularly sensitive to `ComponentPath` changes)
- However, the test then fails because it can't find a Joint conectee for `/jointset/{some_test_joint_in_the_model}` called `child_frame` e.g.:

```
1>Exception:
1>  Failed to connect Socket 'child_frame' of type PhysicalFrame (details:
1>  Connectee for Socket 'child_frame' of type PhysicalFrame in PinJoint at
1>  /jointset/hinge is unspecified. If this model was built programmatically,
1>  perhaps finalizeConnections() was not called before printing.
1>  Thrown at Component.h:3094 in finalizeConnection().).
1>  In Object 'hinge' of type PinJoint.
1>  Thrown at Component.cpp:314 in finalizeConnections().
```

- `child_frame` seems to be generated automatically by the `OpenSim_DECLARE_SOCKET` macros in `Joint.cpp`
- The macro does seem to have an initializer, so the underlying `PropertyIndex_socket_child_frame` index *should* be initialized
- But something (maybe unrelated) seems to go wrong, and this only seems to happen in a debug build under MSVC (uh)

Apart from other things I skirted though during a day of debugging, that's about as far as I've got on it.

I'm assuming if there is some edge-case issue introduced by the `ComponentPath` changes, it would have to survive:

- The existing `testPath` tests (which were just cleaned up a bit)
- The new tests
- Release builds in Windows + Mac + Linux
- Debug build in Linux

The fact that it seems to specifically be broken on my MSVC debug build indicates that either I'm horribly mis-building it (double-check, @aymanhab ) or there's something extremely subtle in a `NDEBUG` somewhere, or in the debug memory layout, that only pops up in that specific combo.

aymanhab:
@adamkewley We stopped running debug tests on ci because many tests timeout (code is unoptimized and more checks performed so timeouts are not surprise). If your analysis indicates no new failures (other than timeouts) compared to 4.1 then we can merge the PR once other comments are addressed, leave  the issue open with a list of failures so we can have a baseline and address in the future. Thank you

adamkewley (commented):
Ah, yeah - didn't pop up in MSVC :(

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2870#-:-MG9pF074nxmRZFlkr4P:b-tjo00e)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @adamkewley and @aymanhab)

___
*[OpenSim/Common/Test/testPath.cpp, line 174 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2870#-MG9HlPH6Y1BrGHUba8J:-MG9pPiaAu5tNGxxaHVT:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/24297b9340328ac8004ca499da3950cdbebda52c/OpenSim/Common/Test/testPath.cpp#L174)):*

This ASSERT fails on non windows because int vs. size_t, please fix. Good test though
</blockquote></details>

Done.

aymanhab:
This can be merged once the list of failing tests is added to the issue for future reference. Thanks @adamkewley

adamkewley:
Great. I'll merge this now and then I guess it's a matter of dedicating some time towards #2866 to figure out what those failures are. I'm hoping they are something extremely rudimentary, like a missing base class overload call, or a missing initialization step somewhere, but a quick scan over the codebase yesterday didn't reveal anything automatically obvious (to me - it might to others).

PR #2872, opened 2020-09-03
Replaced uses of chDir with IO::CwdChanger (RAII safety)

adamkewley:
This is a side-hustle I have been working on whenever I need a break from other dev.

This PR replaces uses of `IO::chDir` with the new `IO::CwdChanger` abstraction, which cleans up some parts of the codebase. To achieve the rollout, it also adds `.reset()` and `.release()` to `IO::CwdChanger`. Some uses of `.reset()` are probably overly paranoid, but they are there to maintain the same directory-changing semantics as the existing code.

The diff is big in places mostly because I removed some `try {} catch (...) { IO::chDir(originalDir); }` code. This required re-indenting, which touches all lines - the implementation should be mostly untouched, apart from the occasional cleanup near wherever I edited the code (e.g. fixing whitespace, using `unique_ptr` where sane, etc. - I probably need more discipline about not doing this, but :shrug:).

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2872)

### DIFF
--- a/OpenSim/Common/IO.cpp
+++ b/OpenSim/Common/IO.cpp
@@ -736,12 +736,12 @@ IO::CwdChanger& IO::CwdChanger::operator=(CwdChanger&& tmp) {
     return *this;
 }
 
-void IO::CwdChanger::reset() {
+void IO::CwdChanger::restore() {
     chDir(_existingDir);
     _existingDir.clear();
 }
 
-void IO::CwdChanger::release() noexcept {
+void IO::CwdChanger::stay() noexcept {
     _existingDir.clear();
 }
 
--- a/OpenSim/Common/IO.cpp
+++ b/OpenSim/Common/IO.cpp
@@ -733,6 +733,7 @@ IO::CwdChanger::CwdChanger(IO::CwdChanger&& tmp) :
 IO::CwdChanger& IO::CwdChanger::operator=(CwdChanger&& tmp) {
     this->_existingDir.clear();
     std::swap(this->_existingDir, tmp._existingDir);
+    return *this;
 }
 
 void IO::CwdChanger::reset() {
--- a/OpenSim/Common/IO.cpp
+++ b/OpenSim/Common/IO.cpp
@@ -730,6 +730,20 @@ IO::CwdChanger::CwdChanger(IO::CwdChanger&& tmp) :
     std::swap(this->_existingDir, tmp._existingDir);
 }
 
+IO::CwdChanger& IO::CwdChanger::operator=(CwdChanger&& tmp) {
+    this->_existingDir.clear();
+    std::swap(this->_existingDir, tmp._existingDir);
+}
+
+void IO::CwdChanger::reset() {
+    chDir(_existingDir);
+    _existingDir.clear();
+}
+
+void IO::CwdChanger::release() noexcept {
+    _existingDir.clear();
+}
+
 IO::CwdChanger::~CwdChanger() noexcept {
     if (!_existingDir.empty()) {
         chDir(_existingDir);
--- a/OpenSim/Common/IO.h
+++ b/OpenSim/Common/IO.h
@@ -184,7 +184,25 @@ public:
         CwdChanger(const CwdChanger&) = delete;
         CwdChanger(CwdChanger&& tmp);
         CwdChanger& operator=(const CwdChanger&) = delete;
-        CwdChanger& operator=(CwdChanger&&) = delete;
+        CwdChanger& operator=(CwdChanger&&);
+
+        /**
+         * Prematurely change the current working directory back to its
+         * original location.
+         *
+         * This is functionally equivalent to prematurely destructing the
+         * CwdChanger. After calling CwdChanger::reset(), the now-reset
+         * CwdChanger will become a noop instance that, when it destructs,
+         * will not attempt to change back to the original directory.
+         */
+        void reset();
+
+        /**
+         * Release CwdChanger's control over the current working directory,
+         * such that the CwdChanger instance does not attempt to change
+         * directory on destruction.
+         */
+        void release() noexcept;
 
         /**
          * Destructs a CwdChanger instance.
--- a/OpenSim/Common/IO.h
+++ b/OpenSim/Common/IO.h
@@ -191,18 +191,18 @@ public:
          * original location.
          *
          * This is functionally equivalent to prematurely destructing the
-         * CwdChanger. After calling CwdChanger::reset(), the now-reset
+         * CwdChanger. After calling CwdChanger::restore(), the now-restored
          * CwdChanger will become a noop instance that, when it destructs,
          * will not attempt to change back to the original directory.
          */
-        void reset();
+        void restore();
 
         /**
          * Release CwdChanger's control over the current working directory,
-         * such that the CwdChanger instance does not attempt to change
-         * directory on destruction.
+         * such that the CwdChanger instance does not attempt to change back
+         * to its original directory on destruction.
          */
-        void release() noexcept;
+        void stay() noexcept;
 
         /**
          * Destructs a CwdChanger instance.
--- a/OpenSim/Common/IO.h
+++ b/OpenSim/Common/IO.h
@@ -143,7 +143,7 @@ public:
      * - On destruction: switches the calling process's working directory
      *   back to its original directory.
      */
-    class CwdChanger final {
+    class OSIMCOMMON_API CwdChanger final {
         std::string _existingDir;
 
         /**
--- a/OpenSim/Simulation/Model/AbstractTool.cpp
+++ b/OpenSim/Simulation/Model/AbstractTool.cpp
@@ -386,59 +386,44 @@ loadModel(const string &aToolSetupFileName, ForceSet *rOriginalForceSet )
             "No model file was specified (<model_file> element is empty) in "
             "the Tool's Setup file. Consider passing `false` for the "
             "constructor's `aLoadModel` parameter");
-    string saveWorkingDirectory = IO::getCwd();
-    string directoryOfSetupFile = IO::getParentDirectory(aToolSetupFileName);
-    IO::chDir(directoryOfSetupFile);
+
+    auto cwd = IO::CwdChanger::changeToParentOf(aToolSetupFileName);
 
     log_info("AbstractTool {} loading model {}", getName(), _modelFile);
 
-    Model *model = 0;
+    auto model = std::unique_ptr<Model>{new Model{_modelFile}};
+    model->finalizeFromProperties();
 
-    try {
-        model = new Model(_modelFile);
-        model->finalizeFromProperties();
-        if (rOriginalForceSet!=NULL)
-            *rOriginalForceSet = model->getForceSet();
-    } catch(...) { // Properly restore current directory if an exception is thrown
-        IO::chDir(saveWorkingDirectory);
-        throw;
+    if (rOriginalForceSet!=NULL) {
+        *rOriginalForceSet = model->getForceSet();
     }
-    _model = model;
-    IO::chDir(saveWorkingDirectory);
+    _model = model.release();
 }
 
 void AbstractTool::
 updateModelForces(Model& model, const string &aToolSetupFileName, ForceSet *rOriginalForceSet )
 {
-    string saveWorkingDirectory = IO::getCwd();
-    string directoryOfSetupFile = IO::getParentDirectory(aToolSetupFileName);
-    IO::chDir(directoryOfSetupFile);
+    auto cwd = IO::CwdChanger::changeToParentOf(aToolSetupFileName);
 
-    try {
-        if(rOriginalForceSet) *rOriginalForceSet = model.getForceSet();
-
-        // If replacing force set read in from model file, clear it here
-        if (_replaceForceSet){
-            // Can no longer just remove the model's forces.
-            // If the model is connected, then the model will
-            // maintain a list of subcomponents that refer to garbage.
-            model.cleanup();
-            model.updForceSet().setSize(0);
-        }
-
-        // Load force set(s)
-        for(int i=0;i<_forceSetFiles.getSize();i++) {
-            log_info("Adding force object set from {}", _forceSetFiles[i]);
-            ForceSet *forceSet=new ForceSet(_forceSetFiles[i], true);
-            model.updForceSet().append(*forceSet);
-        }
-
-    } catch (...) {
-        IO::chDir(saveWorkingDirectory);
-        throw;
+    if (rOriginalForceSet) {
+        *rOriginalForceSet = model.getForceSet();
     }
 
-    IO::chDir(saveWorkingDirectory);
+    // If replacing force set read in from model file, clear it here
+    if (_replaceForceSet){
+        // Can no longer just remove the model's forces.
+        // If the model is connected, then the model will
+        // maintain a list of subcomponents that refer to garbage.
+        model.cleanup();
+        model.updForceSet().setSize(0);
+    }
+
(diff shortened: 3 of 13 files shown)


carmichaelong (commented):
Thanks @adamkewley. It looks good to me, but I want to ping @aymanhab on this too to double check that this will all be fine with any GUI interaction too.

> Reviewed 13 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2872#-:-MGQ0ATkAldyNeTzv_vl:b-s5ciuc)* status: :shipit: complete! all files reviewed, all discussions resolved

adamkewley:
The side-effects of this changeover *should* be identical to the manual
usage (famous last words), in that it effectively replaces uses of chDir
with the RAII wrapper.

It doesn't remove any functions, so if (eg) the gui is dependent on chDir
then that will still work. The only functional change (apart from any bugs
injected accidently by me) *may* be if there is a codepath in the old
version that typically does throw an exception, but wasn't guarded by a try
block. In that case, the code will unwind the directory whereas the
existing implementation would leave the directory in the new location
during the throw.

That's probably not intended behavior, but it is technically a change of
behavior that other code (eg some higher-level try block) may depend on.

aymanhab:
Noting build failure on windows @adamkewley

adamkewley:
Yeah, I need to kick this into shape on Windows, and perform a few minor fixups. I'll de-WIP it and tag you after I've done that @aymanhab

adamkewley:
@aymanhab

OpenSim/Common/IO.cpp
```diff
@@ -730,6 +730,21 @@ IO::CwdChanger::CwdChanger(IO::CwdChanger&& tmp) :
     std::swap(this->_existingDir, tmp._existingDir);
 }
 
+IO::CwdChanger& IO::CwdChanger::operator=(CwdChanger&& tmp) {
+    this->_existingDir.clear();
+    std::swap(this->_existingDir, tmp._existingDir);
+    return *this;
+}
+
```
aymanhab, line 738: My understanding of the use case of this class is that we'll bake the change/restore in the construction/destruction, why do we need move and reset is a bit surprising but may become clear from the rest of the changes.  Could we do without these?

adamkewley (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2872)* status: 12 of 13 files reviewed, 1 unresolved discussion (waiting on @aymanhab and @carmichaelong)

___
*[OpenSim/Common/IO.cpp, line 738 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2872#-MGJ_tHINy-oiHeoeW4v-r2-738:-MGyKGCzBIAMTqSL4S7C:b34i0el) ([raw file](https://github.com/opensim-org/opensim-core/blob/d0fcbfdf87fd19c1bacb9d66af8c478ad3e45be6/OpenSim/Common/IO.cpp#L738)):*

My understanding of the use case of this class is that we'll bake the change/restore in the construction/destruction, why do we need move and reset is a bit surprising but may become clear from the rest of the changes.  Could we do without these?
</blockquote></details>

Move assingment is for cases like:

```c++
auto cwd = IO::CwdChanger::noop();
if (cond) {
    // a bunch of logic that couldn't be folded into a single-assignment ternary
    cwd = IO::CwdChanger::changeTo("path");
}
```

The other ones are for cases where strictly block-scoping the switch would be cumbersome, or where the original application's logic doesn't suit single-assignment or move-assignment overwiting:

```c++
auto cwd = IO::CwdChanger::changeTo("path");
// a load of stuff
// more stuff
if (cond) {
    cwd.reset();
}
// logically, `cwd` can be in one of two states at this point
aPotentiallyPWDSensitiveMethod();
```

adamkewley (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2872)* status: 12 of 13 files reviewed, 1 unresolved discussion (waiting on @aymanhab and @carmichaelong)

___
*[OpenSim/Common/IO.cpp, line 738 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2872#-MGJ_tHINy-oiHeoeW4v-r2-738:-MHGVfruD9PyV9JwkL62:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/d0fcbfdf87fd19c1bacb9d66af8c478ad3e45be6/OpenSim/Common/IO.cpp#L738)):*

Move assingment is for cases like:

```c++
auto cwd = IO::CwdChanger::noop();
if (cond) {
    // a bunch of logic that couldn't be folded into a single-assignment ternary
    cwd = IO::CwdChanger::changeTo("path");
}
```

The other ones are for cases where strictly block-scoping the switch would be cumbersome, or where the original application's logic doesn't suit single-assignment or move-assignment overwiting:

```c++
auto cwd = IO::CwdChanger::changeTo("path");
// a load of stuff
// more stuff
if (cond) {
    cwd.reset();
}
// logically, `cwd` can be in one of two states at this point
aPotentiallyPWDSensitiveMethod();
```
</blockquote></details>

Done.

adamkewley:
This is good to go?

aymanhab (changes requested):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2872)* status: all files reviewed, 4 unresolved discussions (waiting on @adamkewley)

___
*[OpenSim/Common/IO.cpp, line 738 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2872#-MGJ_tHINy-oiHeoeW4v-r2-738:-MHlqRwc6f246AmRYLsi:b4y0qxr) ([raw file](https://github.com/opensim-org/opensim-core/blob/d0fcbfdf87fd19c1bacb9d66af8c478ad3e45be6/OpenSim/Common/IO.cpp#L738)):*

Done.
</blockquote></details>

Thanks for the clarification @adamkewley 
Looking at the code in context, I feel the name CwdChanger can be improved to reflect the fact that the change is temporary or that a restore will (usually) happen. Do you have ideas for a better name? Maybe something to indicate that the change is temporary/local/scoped.

Also would it be possible to add a comment next to the constructors at call sites to indicate the important of the scope so that an innocent addition of a code block for other purposes doesn't mysteriously break code.
___
*[OpenSim/Simulation/Model/AbstractTool.cpp, line 395 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2872#-MHlr9LxCYUtp6ahiCRl:-MHlr9Ly4L9DjWCrVvD1:b-upy5p2) ([raw file](https://github.com/opensim-org/opensim-core/blob/d0fcbfdf87fd19c1bacb9d66af8c478ad3e45be6/OpenSim/Simulation/Model/AbstractTool.cpp#L395)):*
> ```cpp
> 
>     auto model = std::unique_ptr<Model>{new Model{_modelFile}};
>     model->finalizeFromProperties();
> ```

Maybe ok here but ideally these other cleanups are done separately.
___
*[OpenSim/Simulation/Model/AbstractTool.cpp, line 427 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2872#-MHlrbgs5KSgiRXE3iKw:-MHlrbgs5KSgiRXE3iKx:b-hieh5o) ([raw file](https://github.com/opensim-org/opensim-core/blob/d0fcbfdf87fd19c1bacb9d66af8c478ad3e45be6/OpenSim/Simulation/Model/AbstractTool.cpp#L427)):*
> ```cpp
>         model.updForceSet().append(*forceSet);
>     }
> }
> ```

Much cleaner now :+1: 
___
*[OpenSim/Tools/ForwardTool.cpp, line 324 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2872#-MHXLEnc3ABiEdx3YanK:-MHXLEnc3ABiEdx3YanL:bbh3anb) ([raw file](https://github.com/opensim-org/opensim-core/blob/d0fcbfdf87fd19c1bacb9d66af8c478ad3e45be6/OpenSim/Tools/ForwardTool.cpp#L324)):*
> ```cpp
> 
>     cwd.reset();
> ```

Not sure why reset here when old code restores.

adamkewley (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2872)* status: all files reviewed, 4 unresolved discussions (waiting on @adamkewley and @aymanhab)

___
*[OpenSim/Common/IO.cpp, line 738 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2872#-MGJ_tHINy-oiHeoeW4v-r2-738:-MHmBDSVCM-MmgTKxhDU:b8vj5eh) ([raw file](https://github.com/opensim-org/opensim-core/blob/d0fcbfdf87fd19c1bacb9d66af8c478ad3e45be6/OpenSim/Common/IO.cpp#L738)):*

Thanks for the clarification @adamkewley 
Looking at the code in context, I feel the name CwdChanger can be improved to reflect the fact that the change is temporary or that a restore will (usually) happen. Do you have ideas for a better name? Maybe something to indicate that the change is temporary/local/scoped.

Also would it be possible to add a comment next to the constructors at call sites to indicate the important of the scope so that an innocent addition of a code block for other purposes doesn't mysteriously break code.
</blockquote></details>

Hm, unsure about a better *class* name, beyond `DangerYouHadBetterKnowHowToHandleThisWhenYoureRefactoringOpenSim10In2030`, but it might be possible to rename the variables to something like `newCwd`, `processInCwd`, hm

It's one issue with the design, tbh, is that it assumes a fairly liberal understanding of RAII as any resource, including global resources (in this case, environment vars), memory, files, etc. - it's non-obvious to non-C++ey readers, but it's the only way to guarantee RAII safety around exceptions.
___
*[OpenSim/Simulation/Model/AbstractTool.cpp, line 395 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2872#-MHlr9LxCYUtp6ahiCRl:-MHmB5uT1VwFFt9iNqWA:b-wz2c5m) ([raw file](https://github.com/opensim-org/opensim-core/blob/d0fcbfdf87fd19c1bacb9d66af8c478ad3e45be6/OpenSim/Simulation/Model/AbstractTool.cpp#L395)):*

Maybe ok here but ideally these other cleanups are done separately.
</blockquote></details>

Yeah, sorry - I'll typically do a few small fixups while I'm in the area because it's unlikely I'll remember afterwards. I should probably stop that.
___
*[OpenSim/Tools/ForwardTool.cpp, line 324 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2872#-MHXLEnc3ABiEdx3YanK:-MHmA4z10AcY1QB0jYbU:b-e0cknp) ([raw file](https://github.com/opensim-org/opensim-core/blob/d0fcbfdf87fd19c1bacb9d66af8c478ad3e45be6/OpenSim/Tools/ForwardTool.cpp#L324)):*

Not sure why reset here when old code restores.
</blockquote></details>

Good point. I think the semantics need clearing up. The `.reset()` method actually does restore the directory to its original value (the analogue being that if you `.reset()` a resource-holder class, like a `std::unique_ptr`, then it will destruct the old content before clearing it). The `.release()` method tries to tap into a similar semantic - essentially, that the container should "release" its ownership of the resource (in this case, a changed dir). 

It might be a better idea to rename `.reset()` to `.restore()` to clarify intent. I was recycling stdlib wording because something SFINAE-/concept-shaped might benefit from classes having similar names (e.g. `.begin()` being exploited by `std::begin`).

Proposed new (or aliased) naming:

- `.reset() --> .restore()`
- `.release() --> .stay()`?

aymanhab (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2872)* status: all files reviewed, 3 unresolved discussions (waiting on @adamkewley and @aymanhab)

___
*[OpenSim/Tools/ForwardTool.cpp, line 324 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2872#-MHXLEnc3ABiEdx3YanK:-MHqOZIc8PEXtGxzJ4Ly:b-igh68b) ([raw file](https://github.com/opensim-org/opensim-core/blob/d0fcbfdf87fd19c1bacb9d66af8c478ad3e45be6/OpenSim/Tools/ForwardTool.cpp#L324)):*

Good point. I think the semantics need clearing up. The `.reset()` method actually does restore the directory to its original value (the analogue being that if you `.reset()` a resource-holder class, like a `std::unique_ptr`, then it will destruct the old content before clearing it). The `.release()` method tries to tap into a similar semantic - essentially, that the container should "release" its ownership of the resource (in this case, a changed dir). 

It might be a better idea to rename `.reset()` to `.restore()` to clarify intent. I was recycling stdlib wording because something SFINAE-/concept-shaped might benefit from classes having similar names (e.g. `.begin()` being exploited by `std::begin`).

Proposed new (or aliased) naming:

- `.reset() --> .restore()`
- `.release() --> .stay()`?
</blockquote></details>

What do you think about changing the class name to CwdChangerForBlock or CwdLocalChanger to reflect both the acquisition and release of "resource" that's hidden in the destructor? Not particularly fond of either but could be clearer? or bake this connotation in the variable names instead of the class name as in (restoreCwdOnDestruction instead of cwd)?

Now that you explained the rationale for the method names, I think .restore() is definitely better, release can be left alone as is.

adamkewley:
@aymanhab I've renamed the `.reset()` and `.release()` methods to `.restore()` and `.stay()` to clarify intent. I have no intention of renaming `CwdChanger` in this PR - if only because this discussion already occurred in  a previous PR (the one that added `CwdChanger`). The alternative class names suggested here and there do not necessarily clarify the intent.

For local variables, naming them `cwdChangesOnDtor` is a toss up, but one I am against because:

- It's C++, so any initialized instance represents a resource that will be released on destruction. This applies to stack memory, heap memory, files, reference-counted resources, hardware handles, etc. We do not call those resources `freesElementsOnDtor`, `closesFileOnDtor`, or `disconnectsFromModemOnDtor`.
- This issue is compounded when the instance also has member methods, because naming an instance based on its destruction behavior muddles with the idea that the instance represents the management of a currently-active resource when in scope (e.g. compare `restoreCwdOnDestruction.restore();` `restoreCwdOnDestruction.stay();` to `cwd.restore();` and `cwd.stay();`).

The main change I'd like to make--but can't, until we upgrade to C++17--is to add a `[[nodiscard]]` qualifier to the relevant construction methods to clarify to callers that the return value from `CwdChanger` is significant.

aymanhab:
Thanks @adamkewley you make a good point about the naming, my main issue has been that changing directory is not universally agreed upon as a resource acquisition (another toss-up) so all these mechanics are encrypted inside the class may not be clear to clients as opposed to other instances you mentioned but I can live with it.

aymanhab (approved):

adamkewley:
Yeah, agreed. It's not a normal resource - I just don't know of a clearer exception-safe way of cleaning this up. One alternative that's somewhere between doing things manually (resulting in the `try {} catch {}` mess we have) would be to actually remove some specificity by generalizing the concept of doing something while a callstack unwinds. Effectively:

```c++
// generic class that defers an action until its destruction.
// see `go`'s `defer` keyword: https://tour.golang.org/flowcontrol/12
auto changeCwdBack = OpenSim::Defer{[&]() { IO::cwd(originalDir); }};
```

Which would call the deferred function when returning or throwing. There: a major `go` language feature added to C++. Unsure if this is automatically clearer, but it would work for any deferred work!

I'm going to go ahead and merge what's in this PR for now, because (imho) it's an improvement over what we had before and a generalized version (e.g. `Defer`) would need similar semantics (noop, disabling it, etc.), so if we ever get around to something like that it's likely that the code would be restructured similarly.

PR #2877, opened 2020-09-15
Feature merge ScapulothoracicJoint into opensim-core

adamkewley:
This PR tries to merge this (currently, external) OpenSim plugin:

https://github.com/opensim-org/ScapulothoracicJointPlugin

Into OpenSim, so that downstream users can use models containing `ScapulothoracicJoint`s without having to separately download and enable the plugin.

The first commit in this PR is just a copy + paste of the plugin code into OpenSim. I plan on doing a few non-changing refactors to the implementation to clarify what each step is doing, make comments cleaner, etc. The primary purpose of those quick touch-ups is to make the code suitable for merging. There may also be tests added, but I will need to collaborate with the original devs to establish what those tests would actually test.

One downside of the current implementation is that it is one-handed (right-handed, IIRC). This PR shouldn't *necessarily* focus on changing that--that can be step 2, after integration--but the original developers (i.e. @aseth1 and @chrisdembia ) might want to add that support into this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2877)


aymanhab:
@adamkewley this is marked as WIP, please let us know if ready to review and if you plan to merge this in current form or to do refactoring before merging. Thank you

OpenSim/Simulation/Model/ScapulothoracicJoint.cpp
```diff
@@ -32,18 +32,17 @@
 //=============================================================================
```
adamkewley, line 32: This commit contains the only changes from upstream. Effectively, it just tries to clean up the implementation. Main contention points are going to be whether my "cleanups" are cleanups, or just a smattering of Adam-isms. Rather than back-and-forthing, I'll just use the GH review system to explain what's going on.

OpenSim/Simulation/Model/ScapulothoracicJoint.cpp
```diff
@@ -0,0 +1,288 @@
...
+    Transform ellipsoidJointFrameInIntermediate{
+        Rotation{
+            BodyRotationSequence,
+            0, XAxis,
+            0, YAxis,
+            Pi/2,
+            ZAxis},
+        Vec3(0.0)};
+
+    // Transform for Ellipsoid in parent body (Thorax).
```
adamkewley, line 178: In larger functions, multi-step initialization algs are rolled up into standalone lambdas that are immediately called. This reduces the number of local variables in the function, which should hopefully make it easier to keep track of what's happening in the top-level func, and neatly groups bigger initialization steps into blocks /w a comment (e.g. all of the stuff below is *specifically* for initializing `ellipsoidJointFrameInThorax`).

adamkewley (commented):
Just because I know what changes I focused on, I've quickly thrown top-level change comments into a GH review - entirely to point out *why* some things may have been refactored. Again, the first commit is whatever was in upstream. c2c3456 refactors the code to make it a bit more readable (which is what I've just reviewed) - should be no functional changes, unless you spot something I've missed

adamkewley:
@aymanhab I removed the WIP and, to (hopefully) make your life a bit easier, I've added GH review comments to the diff that point out the *kind of* changes I made from upstream. They should all be non-functional (effectively, formatting), but I'm hoping that the explanations clarify why they're there, what they do, etc.

adamkewley:
@aseth1

OpenSim/Simulation/Model/ScapulothoracicJoint.cpp
```diff
@@ -0,0 +1,288 @@
+/* -------------------------------------------------------------------------- *
+ *                    OpenSim:  ScapulothoracicJoint.cpp                      *
+ * -------------------------------------------------------------------------- *
+ * ScapulothoracicJoint is offered as an addition to the OpenSim API          *
+ * See http://opensim.stanford.edu and the NOTICE file for more information.  *
+ * OpenSim is developed at Stanford University and supported by the US        *
+ * National Institutes of Health (U54 GM072970, R24 HD065690) and by DARPA    *
+ * through the Warrior Web program.                                           *
+ *                                                                            *
+ * Copyright (c) 2005-2013 Stanford University and the Authors                *
```
aseth1, line 10: -2020, and include TU Delft

OpenSim/Simulation/Model/ScapulothoracicJoint.cpp
```diff
@@ -0,0 +1,288 @@
+/* -------------------------------------------------------------------------- *
+ *                    OpenSim:  ScapulothoracicJoint.cpp                      *
+ * -------------------------------------------------------------------------- *
+ * ScapulothoracicJoint is offered as an addition to the OpenSim API          *
+ * See http://opensim.stanford.edu and the NOTICE file for more information.  *
+ * OpenSim is developed at Stanford University and supported by the US        *
+ * National Institutes of Health (U54 GM072970, R24 HD065690) and by DARPA    *
+ * through the Warrior Web program.                                           *
```
aseth1, line 8: and CZI 2020-218896

OpenSim/Simulation/Model/ScapulothoracicJoint.h
```diff
@@ -0,0 +1,183 @@
...
+
+namespace OpenSim {
+
+/**
+ * A class implementing a 4-DOF ScapulothoracicJoint.
+ *
+ * Motion of the scapula is described by an ellipsoid surface fixed to the
+ * thorax upon which the joint frame of scapul rides. The DOFs are:
+ *
+ * - 2 DOFs: up-down/medio-laternal: The motion on the surface. These are
```
aseth1, line 37: Update to reflect actual coordinate names: abduction(adduction) and elevation (depression)

OpenSim/Simulation/Model/ScapulothoracicJoint.h
```diff
@@ -0,0 +1,183 @@
...
+ *           described by latitude and longitudinal angles.
+ *
+ * - 1 DOF:  scapula rotation about the normal to the ellipsoid surface
```
aseth1, line 40: upward rotation

OpenSim/Simulation/Model/ScapulothoracicJoint.h
```diff
@@ -0,0 +1,183 @@
...
+ *
+ * - 1 DOF:  Rotation about a "winging" axis defined by a point and axis
```
aseth1, line 42: winging: rotation about and axis fixed to the scapula frame and defaults to scapula-y

aseth1 (changes requested):
Minor comments. Include TU Delft and CZI grant 2020-218896.

OpenSim/Simulation/Model/ScapulothoracicJoint.cpp
```diff
@@ -0,0 +1,288 @@
...
+{
+    constructProperties();
+}
+
+//_____________________________________________________________________________
+/*
+ * Convenience Constructor.
+ */
+ScapulothoracicJoint::ScapulothoracicJoint(
+    const std::string& name,
```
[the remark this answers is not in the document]
aymanhab: @adamkewley Considering that the rest of the code base doesn't follow this convention of every argument on a separate line, let's keep the old convention in place.

OpenSim/Simulation/Model/ScapulothoracicJoint.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+//=============================================================================
+// Simbody Model building.
+//=============================================================================
+//_____________________________________________________________________________
+void ScapulothoracicJoint::extendAddToSystem(SimTK::MultibodySystem& system) const
+{
+    Super::extendAddToSystem(system);
+
+    // Transform for ellipsoid joint frame in intermediate Scapula massless
```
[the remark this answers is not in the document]
aymanhab: Agreed on the approach 👍

OpenSim/Simulation/Model/ScapulothoracicJoint.cpp
```diff
@@ -0,0 +1,288 @@
...
+    // body frame.
+    //
+    // Oriented such that intermediate frame is aligned with scapula joint
+    // frame with respect to the scapula body frame as user specified by
+    // _location and _orientation
+    Transform ellipsoidJointFrameInIntermediate{
```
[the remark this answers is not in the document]
aymanhab: Generally ok within classes, for public interfaces we need to be careful since creation of objects at the call sites or their memory management could be problematic for scripting users.

OpenSim/Simulation/Model/ScapulothoracicJoint.cpp
aymanhab: Main concern with use of lambdas is that our typical users are really comfortable with Matlab/python level syntax, very little C++ so we need to factor in the need for the code to stay readable/understandable for them.  This maybe ok in this case if it avoids code duplication but something to keep in mind going forward.

aymanhab:
Thanks for starting this @adamkewley, had only few general comments. Not sure if there's a plan to make this joint type available in Matlab/python, if so the bindings need to be updated accordingly.

OpenSim/Simulation/Model/ScapulothoracicJoint.cpp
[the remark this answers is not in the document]
adamkewley: Just so I understand, what can happen with scripting users? Do interface generators (e.g. SWIG) not obey value semantics, such that writing (e.g. in Java):

```java
SomeClass.f(new SomeValue());
```

Can bug out? What I mean is, what is there to be "careful" about here for public interfaces? Exposing C++ APIs that take values? Exposing C++ APIs that take rvalues? Writing scripts in the bound-to language (e.g. Matlab/Python) that create temporaries?

When writing C++, I'm mostly just writing standard modern C++, so there's a risk that I'm not aware of this nuance, and that I might inject bugs into downstream (e.g. the GUI) without the tests (written in C++) catching them.

adamkewley:
@aseth1 do downstream typically require bindings for the ScapulothoracicJoint, or should we keep bindgen out of this merge until someone asks for it?

aseth1:
> @aseth1 do downstream typically require bindings for the ScapulothoracicJoint, or should we keep bindgen out of this merge until someone asks for it?

There isn't anything special to prevent SWIG wrapping, so I would include in the bindings. You can see where EllipsoidJoint.h is included and follow suit.

aseth1:
Ideally, we would also include a test case in testJoints.cpp. Unlike the other Joints that compare the direct Simbody implementations, however, this test case should verify that the coordinates behave as specified: Abduction moves the (right handed) scapula to the right (+ve Z, with forward being torso X, up being torso Y), Elevation is up, Upward rotation raises the acromion (and glenohumeral  joint) and winging increases the distance between the medial border and ellipsoid surface.
I could also add a test case after the merge.

aymanhab:
@adamkewley The main issue with binding/scripting is that objects created at the call site are typically garbage collected after the call returns (it's hard to communicate ownership change/transfer across calls without custom code). This is relatively rare but is hard to debug since garbage collection initiation is  nondeterministic.
We also need to make sure that objects introduced across the public interface can be created in scripting (only a subset of classes and instantiated templates are available).

adamkewley:
Alright @aseth1 @aymanhab that should be most of the comments addressed. Ready for the next pass + merge

aymanhab (changes requested):
> Reviewed 3 of 6 files at r6, 1 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2877)* status: 4 of 6 files reviewed, 13 unresolved discussions (waiting on @adamkewley)

___
*[OpenSim/Simulation/Model/ScapulothoracicJoint.cpp, line 55 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2877#-MHGEExzw_B9qNNRDZYI-r1-55:-MHlhIVsEFOccaATKd4-:b-gx75hy) ([raw file](https://github.com/opensim-org/opensim-core/blob/c2c3456002607578cfa1de257971296c30f5fa06/OpenSim/Simulation/Model/ScapulothoracicJoint.cpp#L55)):*

@adamkewley Considering that the rest of the code base doesn't follow this convention of every argument on a separate line, let's keep the old convention in place.
</blockquote></details>

Thank you @adamkewley 
___
*[OpenSim/Simulation/Model/ScapulothoracicJoint.cpp, line 169 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2877#-MHGEExzw_B9qNNRDZYI-r1-169:-MHlheQI19CGot2M0lkv:b-s1d2wl) ([raw file](https://github.com/opensim-org/opensim-core/blob/c2c3456002607578cfa1de257971296c30f5fa06/OpenSim/Simulation/Model/ScapulothoracicJoint.cpp#L169)):*

Just so I understand, what can happen with scripting users? Do interface generators (e.g. SWIG) not obey value semantics, such that writing (e.g. in Java):

```java
SomeClass.f(new SomeValue());
```

Can bug out? What I mean is, what is there to be "careful" about here for public interfaces? Exposing C++ APIs that take values? Exposing C++ APIs that take rvalues? Writing scripts in the bound-to language (e.g. Matlab/Python) that create temporaries?

When writing C++, I'm mostly just writing standard modern C++, so there's a risk that I'm not aware of this nuance, and that I might inject bugs into downstream (e.g. the GUI) without the tests (written in C++) catching them.
</blockquote></details>

To clarify, using the snippet you wrote: 
- if SomeValue is a class that was not exposed to binding or is not a template that is explicitly instantiated for bindings then scripting users can't write "new SomeValue()" at all, and requiring "new SomeValue()" is problematic.
- If the function SomeClass.f() assumes that "SomeValue" has a lifetime beyond the call (for example adds a reference to the passed in object to a container or internal data structure) then the code will bomb out after "SomeValue" is garbage collected.
- I highly recommend the swig 3.0 documentation here to see how move semantics are handled by swig http://www.swig.org/Doc3.0/CPlusPlus11.html#CPlusPlus11_rvalue_reference_and_move_semantics

adamkewley (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2877)* status: 4 of 6 files reviewed, 9 unresolved discussions (waiting on @aseth1 and @aymanhab)

___
*[OpenSim/Simulation/Model/ScapulothoracicJoint.cpp, line 169 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2877#-MHGEExzw_B9qNNRDZYI-r1-169:-MHpW80SAOgmVZldjv7j:bob9bm2) ([raw file](https://github.com/opensim-org/opensim-core/blob/c2c3456002607578cfa1de257971296c30f5fa06/OpenSim/Simulation/Model/ScapulothoracicJoint.cpp#L169)):*

To clarify, using the snippet you wrote: 
- if SomeValue is a class that was not exposed to binding or is not a template that is explicitly instantiated for bindings then scripting users can't write "new SomeValue()" at all, and requiring "new SomeValue()" is problematic.
- If the function SomeClass.f() assumes that "SomeValue" has a lifetime beyond the call (for example adds a reference to the passed in object to a container or internal data structure) then the code will bomb out after "SomeValue" is garbage collected.
- I highly recommend the swig 3.0 documentation here to see how move semantics are handled by swig http://www.swig.org/Doc3.0/CPlusPlus11.html#CPlusPlus11_rvalue_reference_and_move_semantics
</blockquote></details>

Ah, good. That is what I'd expect for a Java-to-C++ FFI, so not many surprises there.

adamkewley:
Great, so, anything left?

aymanhab:
LGTM @adamkewley. Check with @aseth1 that comments has been addressed, especially the test, before merging.

aymanhab (approved):

antoinefalisse:
@adamkewley, @aseth1, @aymanhab, @carmichaelong, @nickbianco I was trying out ID using the scapulothoracic joint and found a bug (I guess). The number of coordinates returned [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/InverseDynamicsSolver.cpp#L99) does not match the size of the `q` vector used [here](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Simulation/InverseDynamicsSolver.cpp#L117). There is one more `q` than coordinates. As a result, the coordinate values passed through a .mot file are not assigned to the right state values and the resulting residuals are wrong. 

I think that, in most cases, the number of generalized coordinates in the SimbodyMatterSubSystem is the same as the number of coordinates in the OpenSim model, which might be why we only see this bug now since it is not the case with the shoulder model .

I am not sure whether this should be solved through the scapulothoracic joint definition or through the ID solver. 

You can check that out using this test (to add the OpenSim/Tests folder) and Adam's `feature-merge-scapulothoracicjoint` branch:
[ShoulderModel_IDSolver.zip](https://github.com/opensim-org/opensim-core/files/5587090/ShoulderModel_IDSolver.zip)

What do you think? I might be wrong :)

tkuchida:
@antoinefalisse Not sure about the specifics of this model or ID, but in general the number of Coordinates may differ from the size of Q if OpenSim uses Euler angles but quaternions are used in the Simbody system.

antoinefalisse:
Thanks @tkuchida that's helpful to know!

I guess my problem here is more that the values reported by `getStateVariableValues(s)` do **not** match the expected state values according to the state order returned by  `getStateVariableNames()` and the passed .mot file.

aseth1:
@antoinefalisse I just checked if  `getStateVariableValues(s)` matches the expected state values according to the state order returned by `getStateVariableNames()` for the shoulder model and I could verify that is does:
image 

The InverseDynamicsTool reorders coordinate values into the order of MultibodyTree before calling the solver. The InverseDynamicsSolver operates on q's, u's and udot's in multibody tree order. I can only see a problem if you try to do some parts manually, but if you use the Tool as is, it should be fine.

antoinefalisse:
Thanks a lot for checking @aseth1. Indeed, I get the same as what you did when trying things out in the GUI. 

What I still don't understand is that if I run the ID tool from the source code (bottom left below), with a setup file that applies a static motion (top) to the generic model (middle left), then when printing out the state values before the call to `calcResidualForceIgnoringConstraints()` (middle bottom) there is a mismatch between the reported state values and the values from the mot file (right bottom). Am I doing something wrong there? Btw I am not sure what I should expect as state values from locked coordinates there. Thanks for your help.

testID

aseth1:
@aymanhab @antoinefalisse I don't believe merging the plugin code as submitted by @adamkewley breaks anything in OpenSim and enables ScapulothoracicJoint to be more widely used.  I suggest that the PR be merged and at that point I can submit a regression test to accompany it so that new features (e.g. option for left-side) do not break existing behaviors.

adamkewley:
Going to merge this so that @aseth1 has a chance to develop a test from the `master` branch. 

Worst case is that beta testing reveals the refactor / changes has broken something (when compared to the plugin code). If that turns out to be the case--and we can't fix it pre-release--I'll create a reverse diff of these changes and revert.

PR #2879, opened 2020-09-21
WIP: Add range loop support to OpenSim::Set and rollout

adamkewley:
Missed my usual Friday refactoring window, but I eventually got there :). This PR adds a `.begin()` and `.end()` to `OpenSim::Set`, so that downstream code can write:

```c++
// before
for (int i = 0; i < set.getSize(); ++i) {
    El& el = set.get(i);
    // do stuff
}

// after
for (El& e : set) {
    // do stuff
}
```

The main motivation behind this is to marginally simplify code that uses sets. A longer-term motivation is to gradually port our existing containers (`Set<T>`, `Array<T>`, etc.) to have an identical downstream API to standard library containers (e.g. by adding iterator support (this), a `.size()` alias, a `.empty()` member, etc.) with the hope of making it easier to substitute containers later (e.g. if it turns out that an algorithm only needs a container, not the full-fat API /w serialization that a `Set` is designed for).

This initial PR contains commits that pass tests on my machine. CI will probably find some edge-cases I didn't code for. Perf-wise, despite the iterator being under-optimized (it currently performs two checks), the performance measurements are identical to `master` (I'll try and paste them in here).

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2879)

### DIFF
--- a/OpenSim/Simulation/Model/ForceSet.cpp
+++ b/OpenSim/Simulation/Model/ForceSet.cpp
@@ -126,31 +126,30 @@ append(Force &aForce)
  * @return True if successful; false otherwise.
  */
 bool ForceSet::append(ForceSet &aForceSet, bool aAllowDuplicateNames)
-{
-    bool success = true;
-    for(int i=0;i<aForceSet.getSize() && success;i++) {
+{    
+    for (Force& f : aForceSet) {
         bool nameExists = false;
-        if(!aAllowDuplicateNames) {
-            std::string name = aForceSet.get(i).getName();
-            for(int j=0;j<getSize();j++) {
-                if(get(j).getName() == name) {
+
+        if (!aAllowDuplicateNames) {
+            std::string name = f.getName();
+            for (const Force& f2 : *this) {
+                if (f2.getName() == name) {
                     nameExists = true;
                     break;
                 }
             }
         }
-        if(!nameExists) {
-            if(!Super::adoptAndAppend(&aForceSet.get(i))) 
-                success = false;
+
+        if (!nameExists) {
+            if (!Super::adoptAndAppend(&f)) {
+                return false;
+            }
         }
     }
 
-    if(success) {
-        updateActuators();
-        updateMuscles();
-    }
-
-    return(success);
+    updateActuators();
+    updateMuscles();
+    return true;
 }
 //_____________________________________________________________________________
 /**
@@ -211,9 +210,11 @@ void ForceSet::updateActuators()
 {
     _actuators.setMemoryOwner(false);
     _actuators.setSize(0);
-    for (int i = 0; i < getSize(); ++i) {
-        Actuator* act = dynamic_cast<Actuator*>(&get(i));
-        if (act)  _actuators.adoptAndAppend(act);
+    for (Force& f : *this) {
+        Actuator* act = dynamic_cast<Actuator*>(&f);
+        if (act) {
+            _actuators.adoptAndAppend(act);
+        }
     }
 }
 
@@ -228,8 +229,10 @@ const Set<Muscle>& ForceSet::getMuscles() const
 }
 Set<Muscle>& ForceSet::updMuscles() 
 {
-    if (_muscles.getSize() == 0)
+    if (_muscles.getSize() == 0) {
         updateMuscles();
+    }
+
     return _muscles;
 }
 //_____________________________________________________________________________
@@ -240,9 +243,12 @@ void ForceSet::updateMuscles()
 {
     _muscles.setMemoryOwner(false);
     _muscles.setSize(0);
-    for (int i = 0; i < getSize(); ++i) {
-        Muscle* m = dynamic_cast<Muscle*>(&get(i));
-        if (m)  _muscles.adoptAndAppend(m);
+
+    for (Force& f : *this) {
+        Muscle* m = dynamic_cast<Muscle*>(&f);
+        if (m) {
+            _muscles.adoptAndAppend(m);
+        }
     }
 }
 
@@ -261,9 +267,8 @@ void ForceSet::updateMuscles()
 void ForceSet::
 getStateVariableNames(OpenSim::Array<std::string> &rNames) const
 {
-    for(int i=0;i<getSize();i++) {
-        ScalarActuator *act = dynamic_cast<ScalarActuator*>(&get(i));
-       
+    for (Force& f : *this) {
+        auto* act = dynamic_cast<ScalarActuator*>(&f);
         if(act) {
             rNames.append(act->getStateVariableNames());
         }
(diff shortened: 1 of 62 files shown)


adamkewley:
Perf measurements, Ryzen1600, 8 repeats:

|                   Gait2354 |       0.26 |     0.00 |       0.27 |     0.00 |    0.99 |
|             ToyDropLanding |      11.39 |     0.01 |      11.48 |     0.00 |    0.99 |
|            passive_dynamic |       4.61 |     0.00 |       4.62 |     0.00 |    1.00 |
| passive_dynamic_noanalysis |       3.00 |     0.00 |       3.03 |     0.00 |    0.99 |
|                      Arm26 |       0.26 |     0.00 |       0.26 |     0.00 |    1.00 |
|             RajagopalModel |      13.20 |     0.02 |      13.11 |     0.01 |    1.01 |

Essentially, within the noise, so I don't think that this is a readability vs. perf tradeoff (you get one without affecting the other, and the iterator could be updated to have fewer checks etc. if we want that)

OpenSim/Simulation/Model/ForceSet.cpp
```diff
@@ -281,15 +286,16 @@ getStateVariableNames(OpenSim::Array<std::string> &rNames) const
...
+    for (Force& f : *this) {
+        auto* act = dynamic_cast<ScalarActuator*>(&f);
+        if (!act) {
+            continue;
+        }
     }
+    // TODO: this is a bogus method?
 
-    return(status);
+    return status;
```
adamkewley, line 300: This is bogus because it effectively has no side-effects and just returns `true`

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -1293,13 +1293,10 @@ int Model::getNumMuscleStates() const {
 int Model::getNumProbeStates() const {
 
     int n = 0;
-    for(int i=0;i<get_ProbeSet().getSize();i++){
-        Probe *p = dynamic_cast<Probe*>( &get_ProbeSet().get(i) );
```
adamkewley, line 1297: Original is bogus because the probeset contains `Probe`s, so it's a pointless `dynamic_cast`

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -1773,8 +1762,6 @@ void Model::updateMarkerSet(MarkerSet& newMarkerSet)
         // append the marker to the model's Set
         addMarker(updatingMarker.clone());
     }
-
-    log_info("Updated markers in model {}", getName());
```
adamkewley, line 1777: log_info needs to be re-added in the PR

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -2033,9 +2020,9 @@ void Model::formQStorage(const Storage& originalStorage, Storage& qStorage) {
     Array<string> qNames;
     getCoordinateSet().getNames(qNames);
 
+    std::vector<int> mapColumns(qNames.getSize());
```
adamkewley, line 2023: This version does not fill the tail with `-1` default vals, so need to ensure that doesn't break something

OpenSim/Tests/SimpleOptimizationExample/testSimpleOptimizationExample.cpp
```diff
@@ -58,10 +58,12 @@ class ExampleOptimizationSystem : public OptimizerSystem {
...
         const Set<Muscle> &muscleSet = osimModel.getMuscles();
         // Make sure other muscle states are initialized the same with 1.0 activation, 0.1 fiberLength followed by equilibrium computation
-        for(int i=0; i< muscleSet.getSize(); i++ ){
-            muscleSet[i].setActivation(s, 1.0);
-            const ActivationFiberLengthMuscle* afl = ActivationFiberLengthMuscle::safeDownCast(&muscleSet[i]);
-            if (afl) afl->setFiberLength(s, .1);
+        for (Muscle& m : muscleSet){
+            m.setActivation(s, 1.0);
+            const auto* afl = ActivationFiberLengthMuscle::safeDownCast(&m);
+            if (afl) {
```
adamkewley, line 64: Might be an unnecessary `if`

OpenSim/Tools/RRATool.cpp
```diff
@@ -586,27 +586,28 @@ bool RRATool::run()
...
         // Add functions from data files to track states
         Storage stateStorage(taskSet.getDataFileName());
         GCVSplineSet* stateFuntcions = new GCVSplineSet(3, &stateStorage);
-        for (int i=0; i< stateFuntcions->getSize(); i++)
-            qAndPosSet->cloneAndAppend(stateFuntcions->get(i));
-
+        for (Function& f : *stateFuntcions) {
+            qAndPosSet->cloneAndAppend(f);
+        }
+        // TODO: this leaks, right?
```
adamkewley, line 610: This still looks like a leak

adamkewley:
This PR is essentially alright to review now. I'll still do a pass or two over the diffs to double-check that the logic is the same as the pre-range-loop version.

The main motivation here is to phase out older pre-C++11 coding patterns that are a bit wordier. While I'm in the area, I may have also refactored other old-ish patterns that are semi-related to using `for` loops with variables, such as removing top-level local variable declarations:

```c++
// before
void f() {
    int i, j;
    double d1, d2, d3;
    for (i = 0, j = collection.size(); i < j; ++i) {
        d1 = collection[i].a;
        d2 = collection[i].b;
        d3 = collection[i].c;
        func(d1 * d2 * d3);
    }
}

// after
void f() {
    for (const El& el : collection) {
        double d1 = el.a;
        double d2 = el.b;
        double d3 = el.c;
        func(d1 * d2 * d3);
    }
}

// or, now, even:
void f() {
    for (const El& el : collection) {
        func(el.a * el.b * el.c);
    }
}
```

And non-bracere'd `if` and `for` chaining, which is one of those patterns that will end up wasting someone's time sooner or later when something gets added later:

```c++
// before: be *very* careful where you add new code
void f() {
    int i, j;
    for (i = 0; i < coll.size(); ++i)
        for (j = 0; j < coll[i].size(); ++j)
            if (coll[i][j])
                func(col[i][j]);
}

// after: more lines wasted on bracers, but clearer (imho) and
// unlikely to subtly fail if/when a new line is added
void f() {
    for (const A& a : coll) {
        for (B& b : a) {
            if (b) {
                 func(b);
            }
        }
    }
}
```

And nesting of `if`s where early exiting is cleaner:

```c++
// before
void f() {
    if (cond) {
        // entire function is one level deeper
        // ... some long impl ...
    }
}

// after
void f() {
    if (!cond) {
        return;
    }
    //  ... some long impl ...
}
```

The other cleanups are to the existing `for` loops. Their primarily edit is going to be adding some whitespace around them to make it easier to read. The reason they are edited is because I first added the `SetIterator` and then searched for all usages of `getSize` globally to find places that range loops could be used:

```c++
// before
void f() {
    for (i=0;i<coll.getSize();i++) {
        // ...
    }
}

// after: it's easier to see that it's actually `j` that is incremented
void f() {
    for (i = 0; i < coll.getSize(); ++i) {
        // ...
    }
}
```

If they aren't changed to a `for` range loop, it's because the index is used somewhere in the method body for something other than a `get` call.

adamkewley:
Kicking this back into a WIP.

This PR is wayyy too big as-is. I'll probably split each piece of work into separate PRs that are easier to review, as with (e.g.) the `CwdChanger`:

- PR1: Implement range `for` loop + demo usage
- PR2: Rollout to ~5-10 locations
- PRn: Rollout to 5n-(5n+5) locations etc.

adamkewley:
Closing this: it's a junk PR now and I can reformulate it elsewhere

Issue #2865, opened 2020-08-24
Memory leak during integration


mitkof6:
I was trying to figure out why we get memory leaks in [osim-rl](https://github.com/stanfordnmbl/osim-rl/issues/10). I use a slightly different custom version of osim-rl. However, it seems that everything points towards the integrate function of the Manager. The new version of the Manager does not have a reset mechanism and therefore one must delete the manager and create a new one. I have a minimum working example attached which demonstrates that memory increases. Could you please have a look and tell me if I miss something and if there is a way to figure out where the problem is? 

[memory_leak_code.zip](https://github.com/opensim-org/opensim-core/files/5120309/memory_leak_code.zip)

For those who do not have time to run the code, you can find the code here:

```
#include <iostream>
#include <OpenSim/OpenSim.h>
#include "memory_usage.h"

using namespace OpenSim;
using namespace SimTK;
using namespace std;

int main(int argc, char *argv[])
{
	auto model = Model("model_generic.osim");
	auto state = model.initSystem();
	auto manager = new Manager(model);

	auto previous_size = getCurrentRSS();
	auto i = 101;
	while (true) {
		if (i >= 100) {
			i = 1;
			state = model.initializeState();
			model.equilibrateMuscles(state);
			state.setTime(0);
			cout << "manager will be deleted" << endl;
			delete manager;
			manager = new Manager(model);
			manager->initialize(state);
		}
		manager->integrate(i * 0.01);
		i++;

		// check if memory usage increased
		auto current_size = getCurrentRSS();
		auto diff = current_size - previous_size;
		previous_size = current_size;
		if (diff != 0) {
			cout << i << " " << diff * 1e-6 << "MB" << endl;
		}
	}

    return 0;
}
```

Many thanks!

mitkof6:
This problem pops up when we integrate the model. Although one would expect that if we reset the manager frequently then space would be freed, but this is not the case. This is something that is needed when one trains an RL agent with the OpenSim environment.

clnsmith:
https://github.com/clnsmith/opensim-jam/blob/fb777056217ec60e8eb2eedc93fefa5a6e4885e0/src/ForsimTool.cpp#L86

I'm not sure about the cause of the memory leak, but if you want to get around using the Manager all together here is some sample code ^ to manage the time stepping yourself.

mitkof6:
[...]

PR #2882, opened 2020-09-25
Support for TimeStepper and RungeKuttaMerson integrator in the bindings

mitkof6:
Resolves some of the points in issue #2865

### Brief summary of changes

Include TimeStepper and RungeKittaMersonIntegrator in the bindings. 

Tried to look for the memory leak in the Manager. Even if we permit re-initialization it seems that the memory leak persists. As for the time stepper approach it seems to work properly for my case.

### Testing I've completed

Testing the correct behavior of the bindings with Python.

The swig bindings are bit magic to me. I have just included the minimum number of header needed for this to work properly. Please let me know if we need to change anything there.

### CHANGELOG.md (choose one)

Should be updated in the release.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2882)

### DIFF
--- a/Bindings/SWIGSimTK/Integrator.h
+++ b/Bindings/SWIGSimTK/Integrator.h
@@ -1,0 +1,423 @@
+#ifndef SimTK_SIMMATH_INTEGRATOR_H_
+#define SimTK_SIMMATH_INTEGRATOR_H_
+
+/* -------------------------------------------------------------------------- *
+ *                        Simbody(tm): SimTKmath                              *
+ * -------------------------------------------------------------------------- *
+ * This is part of the SimTK biosimulation toolkit originating from           *
+ * Simbios, the NIH National Center for Physics-Based Simulation of           *
+ * Biological Structures at Stanford, funded under the NIH Roadmap for        *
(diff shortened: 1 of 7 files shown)


aymanhab:
Thanks @mitkof6 My main concern here is adding to the list of simbody headers that are duplicated for SWIG wrapping. This was necessary at some point because SWIG was having a hardtime digesting simbody headers but I'd try to avoid adding more to this problem. I'd keep this PR on a branch until we migrate to SWIG 4.0 and see if we can use simbody headers as is and eliminate/reduce the duplication.
Let me know if this is critical for your work in the short term. Thank you

mitkof6:
Hi @aymanhab thanks for your feedback. I agree that this is not the best solution. Also, I think that some of the functions that were exposed might not work properly, because we will have to expose more classes from Simbody or hide them, which however takes a lot of time to test. For me this is a good work around as I resolved a memory leak of the Manager. I can use my own branch. Maybe it is good to keep this around but not merge in case people have similar problems.

I tried looking at the Manager and there seem to be a lot of redundant code. However, I am not sure if removing some parts will cause problem to the GUI and any legacy code.

PR #2885, opened 2020-10-01
Reset _controllerSet storage in the manager to avoid memory leaks

mitkof6:
Fixes issue #2865

### Brief summary of changes

Make sure to reset `_controllerSet` internal storage with `constructStorage` each time the Manager is initialized.

### Testing I've completed

Manually observed the memory allocation size to ensure that if the manager is destroyed or re-intialized the memory does not grow.

Please let me know if this can cause any undesirable behavior outside the Manager class. 

Also, I was thinking of how to fix the initialize method so that we can re-initialize the manager without having to delete it. I think the following part could be removed:

https://github.com/opensim-org/opensim-core/blob/82ec3d1c74af1ff1893d9c5453ef6c52fe2e2fc9/OpenSim/Simulation/Manager/Manager.cpp#L813

What do you think?

### CHANGELOG.md (choose one)

- no need to update

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2885)

### DIFF
--- a/OpenSim/Simulation/Manager/Manager.cpp
+++ b/OpenSim/Simulation/Manager/Manager.cpp
@@ -790,10 +790,6 @@ void Manager::initializeStorageAndAnalyses(const SimTK::State& s)
         // STORE STARTING CONTROLS
         if (_model->isControlled()){
             _controllerSet->connectToModel(*_model);
-			// Here we call the constructStorage because it possible that the
-			// Model's control storage has already be appended in a previous
-			// simulation [Dimitar Stanev; see Manager memory leak PR].
-            _controllerSet->constructStorage();
         }
 
         OPENSIM_THROW_IF(!hasStateStorage(), Exception,
@@ -805,7 +801,7 @@ void Manager::initializeStorageAndAnalyses(const SimTK::State& s)
 }
 //_____________________________________________________________________________
 /**
-* set and initialize a SimTK::TimeStepper
+* Set and initialize a SimTK::TimeStepper
 */
 void Manager::initialize(const SimTK::State& s)
 {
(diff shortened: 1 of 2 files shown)


aymanhab (changes requested):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2885)* status: all files reviewed, 1 unresolved discussion (waiting on @mitkof6)

___
*[OpenSim/Simulation/Manager/Manager.cpp, line 832 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2885#-MIeq8Bf9zAZ-VPYZyye:-MIeq8Bf9zAZ-VPYZyyf:b-71xohx) ([raw file](https://github.com/opensim-org/opensim-core/blob/de05066da1cf3a54cd184cae30b8cf7fef4b1c53/OpenSim/Simulation/Manager/Manager.cpp#L832)):*
> ```cpp
>         if (_model->isControlled())
>             _controllerSet->constructStorage();
> }
> ```

I'd just and the two conditions instead of cascading ifs. Also include the PR number for future reference.

aymanhab:
Looks good @mitkof6 minor comments to address and then feel free to merge. The test failure on osx is sporadic and unrelated. Thank you. Let's keep the scope of this to memory leak and discuss re-initialization in another PR with input from others who fought with this in the past.

OpenSim/Simulation/Manager/Manager.cpp
```diff
@@ -822,11 +822,18 @@ void Manager::initialize(const SimTK::State& s)
         _timeStepper->initialize(s);
         _timeStepper->setReportAllSignificantStates(true);
     }
+
+    // Here we call the constructStorage because it is possible that the Model's
+    // control storage has already be appended in a previous simulation [Dimitar
+    // Stanev; see Manager memory leak PR].
```
adamkewley, line 828: All looks good. The only thing I'd maybe change is to change this comment to refer to a URL or issue number just so that it's easy for a later maintenance dev to pull up what the original issue was (if they're wondering)

adamkewley:
Not trying to undermine/collide with Ayman's review (he is the primary here). I was just curious about the fix (after reading into the issue) and noticed the comment. Extremely nitpicky, I apologize.

mitkof6:
@aymanhab and @adamkewley thanks for your suggestions.

aymanhab:
Thanks for the fix @mitkof6 and for the review @adamkewley 👍

PR #2805, opened 2020-06-16
Merge in Moco.

chrisdembia:
Google Doc: https://docs.google.com/document/d/10SiVJGqMJxiXtiDdI4Ozpxf5HTKHaixX_ynymYuJwk4/edit

This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2805)

# Overview 

Refer to OpenSim Moco's doxygen: https://opensim-org.github.io/opensim-moco-site/docs/

CMake and building

- **/**: git metadata, README, CHANGELOG, CMakeLists.
- **/.github/workflows/continuous_integration.yml**
- **/cmake** 
- **/dependencies** 

Bindings
- **/Bindings/moco.i**
- **/Bindings/Java/Matlab/examples/Moco**
- **/Bindings/Java/Matlab/tests**
- **/Bindings/Java/OpenSimJNI/CMakeLists.txt**
- **/Bindings/Java/swig/java_moco.i**
- **/Bindings/Java/tests/**
- **/Bindings/Python/**
- **/Bindings/Python/examples/Moco**
- **/Bindings/Python/swig/python_moco.i**

C++ examples
- **/OpenSim/Examples**: CMake file to support how C++ examples are installed
- **/OpenSim/Examples/Moco**: Moco's C++ examples

Moco C++ library: MocoStudy
- **/OpenSim/Moco/MocoStudy**
- **/OpenSim/Moco/MocoProblem**
- **/OpenSim/Moco/MocoProblemRep**: Interface of MocoProblem for solvers
- **/OpenSim/Moco/MocoVariableInfo**
- **/OpenSim/Moco/MocoConstraint**
- **/OpenSim/Moco/MocoGoal/MocoGoal**
- **/OpenSim/Moco/MocoGoal/**: subclasses of MocoGoal
- **/OpenSim/Moco/MocoSolver**
- **/OpenSim/Moco/MocoTropterSolver**
- **/OpenSim/Moco/MocoCasADiSolver/MocoCasADiSolver**
- **/OpenSim/Moco/MocoCasADiSolver/**: related classes  🙈 
- **/OpenSim/Moco/MocoTrajectory**
- **/OpenSim/Moco/MocoStudyFactory**

Moco C++ library: MocoTool
- **/OpenSim/Moco/MocoTool**
- **/OpenSim/Moco/MocoInverse**
- **/OpenSim/Moco/MocoTrack**

Support/utilities
- **/OpenSim/Moco/MocoUtilities**
[...]

PR #2890, opened 2020-10-08
Fix componentpath formrelativepath

adamkewley:
This fixes a bug found here:

https://github.com/opensim-org/opensim-core/pull/2805

It's explained in more detail on that PR but, effectively, `ComponentPath::formRelativePath` is broken on the edge-case of a path subcomponent containing a non-separator at the same offset as another path subcomponent containing its NUL terminator (e.g. `/component`, `/component_path`).

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2890)

### DIFF
--- a/OpenSim/Common/ComponentPath.cpp
+++ b/OpenSim/Common/ComponentPath.cpp
@@ -304,46 +304,8 @@ ComponentPath ComponentPath::formAbsolutePath(const ComponentPath& otherPath) co
 }
 
 ComponentPath ComponentPath::formRelativePath(const ComponentPath& otherPath) const {
-    if (!isAbsolute()) {
-        OPENSIM_THROW(Exception, _path + ": is not an absolute path.");
-    }
-
-    if (!otherPath.isAbsolute()) {
-        OPENSIM_THROW(Exception, _path + ": is not an absolute path.");
-    }
-
-    // helper: returns the index of the first character where s1[index] !=
-    // s2[index] returns s1.size() for identical inputs
-    auto mismatchOffset = [](const std::string& s1, const std::string& s2) {
-        const size_t shortest = std::min(s1.size(), s2.size());
-
-        size_t offset = 0;
-        while (offset < shortest) {
-            if (s1[offset] != s2[offset]) {
-                break;
-            }
-            ++offset;
-        }
-
-        return offset;
-    };
-
-    // helper: returns the index of the first separator to the (inclusive) left
-    // of `offset` in `s`.
-    auto rfindSeparator = [](const std::string& s, size_t offset) {
-        while (offset > 0) {
-            char c = s[offset];
-            if (c == separator) {
-                break;
-            }
-            --offset;
-        }
-
-        return offset;
-    };
-
-    // helper: generate a path string containing `n` step ups (e.g. "../../")
-    auto generateStepUps = [](size_t n) {
+    // helper: returns a path string containing `n` step ups (e.g. "../../")
+    static auto generateStepUps = [](size_t n) {
         std::string rv;
         for (size_t i = 0; i < n; ++i) {
             rv += "..";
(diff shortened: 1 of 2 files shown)


adamkewley:
The other edge-cases I need to double-check are whether the old implementation returns `""` or `"."` when both paths are identical. Logically, it should be "." (`""` is not a path), but I have a suspicion that it returns `""`.

adamkewley:
This should be basically ready for review, provided CI passes.

nickbianco (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2890)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @adamkewley)

___
*[OpenSim/Common/Test/testPath.cpp, line 120 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2890#-MJ8gCZlDeAtteWDX4K7:-MJ8gCZlDeAtteWDX4K8:b-fhqt0z) ([raw file](https://github.com/opensim-org/opensim-core/blob/9a4a4b46d6c8de56e9dc212e151c587294496efd/OpenSim/Common/Test/testPath.cpp#L120)):*
> ```cpp
>                 ss << "  expected = " << tc.expected << std::endl;
>                 ss << "       got = " << ans.toString();
>                 throw std::runtime_error{ss.str()};
> ```

We should use the new logging tools here instead (i.e., `log_error()`).

adamkewley (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2890)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @nickbianco)

___
*[OpenSim/Common/Test/testPath.cpp, line 120 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2890#-MJ8gCZlDeAtteWDX4K7:-MJ8tuCX96vcZzteio3m:bvkfbc4) ([raw file](https://github.com/opensim-org/opensim-core/blob/9a4a4b46d6c8de56e9dc212e151c587294496efd/OpenSim/Common/Test/testPath.cpp#L120)):*

We should use the new logging tools here instead (i.e., `log_error()`).
</blockquote></details>

Assertion errors aren't typically logged in test suites unless its a stateful test that needs to log a bunch of info just before it fails. So I don't think logging is the right fit. Assuming there's a `format` equivalent, though, this could be rewritten to as:

```c++
            throw std::runtime_error{format(R"(ComponentPath::formRelativePath produced an invalid output
      from = {}
        to = {}
  expected = {}
       got = {}
)", tc.from, tc.to, tc.expected, tc.got)};
```

Not entirely sure if it's worth the effort of an edit, because the old form is pretty standard C++, and isn't particularly ugly. If you're insistent, I'll do it once I get around to a second pass

nickbianco (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2890)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @adamkewley)

___
*[OpenSim/Common/Test/testPath.cpp, line 99 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2890#-MJ9EZe46h0EJiQPDdat:-MJ9EZe46h0EJiQPDdau:b32v57) ([raw file](https://github.com/opensim-org/opensim-core/blob/9a4a4b46d6c8de56e9dc212e151c587294496efd/OpenSim/Common/Test/testPath.cpp#L99)):*
> ```cpp
>             // An example that failed post-merge
>             //     see https://github.com/opensim-org/opensim-core/pull/2805
>             { "/contact", "/contact_point", "../contact_point" },
> ```

+1
___
*[OpenSim/Common/Test/testPath.cpp, line 120 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2890#-MJ8gCZlDeAtteWDX4K7:-MJ9Br0ODpUAGKy8_JVE:bkx5t45) ([raw file](https://github.com/opensim-org/opensim-core/blob/9a4a4b46d6c8de56e9dc212e151c587294496efd/OpenSim/Common/Test/testPath.cpp#L120)):*

Assertion errors aren't typically logged in test suites unless its a stateful test that needs to log a bunch of info just before it fails. So I don't think logging is the right fit. Assuming there's a `format` equivalent, though, this could be rewritten to as:

```c++
            throw std::runtime_error{format(R"(ComponentPath::formRelativePath produced an invalid output
      from = {}
        to = {}
  expected = {}
       got = {}
)", tc.from, tc.to, tc.expected, tc.got)};
```

Not entirely sure if it's worth the effort of an edit, because the old form is pretty standard C++, and isn't particularly ugly. If you're insistent, I'll do it once I get around to a second pass
</blockquote></details>

Oh yes, thanks for the clarification. It's fine as is.

adamkewley:
Going to merge this: it's a non-api-breaking fix that is necessary for the Moco merge.

nickbianco:
Hey @adamkewley, thanks for creating the fix. We usually wait for an approved review before merging. I left some approving comments, but not a full review. Sorry if there was some confusion there.

@aymanhab is going to take a look over the changes since he's more familiar with the recent ComponentPath changes.

aymanhab:
LGTM @adamkewley just had a question regarding possible end-cases before we close the door on this refactoring/fix. Thank you.

PR #2892, opened 2020-10-14
generateDecorations Apply possible transform between Joint and Body frames

aymanhab:
Fixes issue https://github.com/opensim-org/opensim-gui/issues/1178

### Brief summary of changes
For a Joint that uses an offset frame, account for the transform between the offset frame and the underlying mobilized body
### Testing I've completed
Visualized the model in https://github.com/opensim-org/opensim-gui/issues/1178
### CHANGELOG.md (choose one)

- no need to update because corner case of a previous fix already included in changelog

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2892)

### DIFF
--- a/OpenSim/Simulation/SimbodyEngine/EllipsoidJoint.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/EllipsoidJoint.cpp
@@ -180,5 +180,6 @@ void EllipsoidJoint::generateDecorations
     const OpenSim::PhysicalFrame& frame = getParentFrame();
     ellipsoid.setColor(Vec3(0.0, 1.0, 1.0));
     ellipsoid.setBodyId(frame.getMobilizedBodyIndex());
+    ellipsoid.setTransform(frame.findTransformInBaseFrame());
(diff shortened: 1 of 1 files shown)


aymanhab:
@carmichaelong Thanks for the testing. This should take care of the offset. You can test when ci finishes. Thank you.

carmichaelong:
@aymanhab Since this was a core test, I tried to use the Simbody visualizer to replicate the issue in 4.1, and then check the artifact. Seems like it wasn't an issue in 4.1 with the Simbody visualizer as I can get the sphere to change locations as expected.

The change looks fairly small, and seems like this needs to be merged and then built with the GUI to test? If so, feel free to merge this and then I can wait for the nightly build to build with this version of core.

aymanhab:
Agreed @carmichaelong, please mark it as approved and I'll merge and it will show up in the next gui master build. Thank you.

carmichaelong (approved):

carmichaelong:
LGTM thanks @aymanhab

Issue #2718, opened 2020-04-08
Move Moco's analyze() free function to opensim-core


chrisdembia:
Moco contains an analyze() free function that evaluates outputs along a trajectory.

Issue #2715, opened 2020-04-08
Move Moco's ModelFactory into opensim-core


chrisdembia:
Moco has a ModelFactory class for creating models such as N-link pendulums, and for modifying models by removing muscles, adding reserves, etc.

chrisdembia:
@carmichaelong fyi the ModelProcessor/ModelOperators depend on the ModelFactory, so it may make sense to move the ModelFactory in the same PR.

aymanhab:
@nickbianco There are a few issues with Moco label that may have been merged into master with the moco-feature branch merge, can you add comment as to what's left there if any? Thanks

nickbianco:
Closing via #2895.

Issue #2713, opened 2020-04-08
Move Moco's TableProcessor into opensim-core


chrisdembia:
Moco's TableProcessor provides an XML-serializable pipeline of operations to TimeSeriesTables.

Moco's version does not yet support TimeSeriesTableVec3, but @nickbianco mentioned that perhaps a "flatten-operate-repack" approach might be sufficient.

chrisdembia:
@carmichaelong this issue is a little bigger than #2712 because we want to templatize `TableProcessor`; there is some more information over at https://github.com/opensim-org/opensim-moco/issues/628.

nickbianco:
Closing since the core of this issue is moving TableProcessor to opensim-core. Templatizing TableProcessor remains its own issue in opensim-org/opensim-moco#628.

PR #2895, opened 2020-10-21
Merge Moco feature branch into master

nickbianco:
Fixes issues #2559, #2715, #2712, #2718, #2713. 

### Summary of changes
See #2805.

### Checklist
- [ ] Address remaining comments (if any)
- [x] API changes compatible with GUI (local testing)
- [ ] Final check that CI tests pass
- [ ] Others?

### CHANGELOG.md (choose one)

- updated...but more detail could be added.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2895)


aymanhab:
@adamkewley Now we have this PR to merge moco into opensim-core. Your feedback/thoughts on this in terms of dependencies, testing and any experience building for Moco users locally. Thank you.

aymanhab:
@antoinefalisse if you can give this branch a spin to see if the Moco functionality is available/documented as well test some examples that would be great. Thank you

antoinefalisse:
@aymanhab I started testing this branch. Here are a few remarks:

I have tested all moco-related examples - I have had problems with two of them:

- exampleTracking crashes. It seems related to the absence(?) of metadata for [this table](https://github.com/opensim-org/opensim-core/blob/feature_moco/OpenSim/Examples/Moco/exampleTracking/exampleTracking.cpp#L124) (the error is thrown [here](https://github.com/opensim-org/opensim-core/blob/feature_moco/OpenSim/Common/TableUtilities.cpp#L43)).

- not a real issue but this example `exampleCustomImplicitAuxiliaryDynamics` has a very long name so visual studio keeps complaining about the number of characters and there is one `failed` when building `ALL BUILD`. I am building OpenSim in a not so nested folder `C:\Users\u0101727\Documents\MyRepositories\opensim-testing` so I expect this issue to occur for many users. Would that be an option to have a shorter name?

Other remarks:

- In visual studio, in the folder `Libraries`, all libraries are named `osim<>`. Yet for Moco, we have `Library - osimMoco`. Would that make sense to remove `Library - `?

- Is it normal that `osimMocoCustomEffortGoal` is in the folder Examples?

- In tropter, `sandbox_finite_diff_colpack` has errors. The headers are not properly included and `#include <memory>` is missing. But there are also links errors. I don't know if it is the goal to have that to work.

For the rest, all went smooth. Happy to do more if needed.

nickbianco:
Thanks @antoinefalisse. I fixed exampleTracking and renamed the Moco project library to 'osimMoco'. 

I'm surprised that `exampleCustomImplicitAuxiliaryDynamics` triggers the character length error with that relatively short path. I've seen this error before, but with with longer paths. This is the longest named example, but not too much longer than some others. The example name is adequately descriptive so I'm hesitant to change it, but perhaps `exampleCustomImplicitDynamics` or `exampleCustomAuxiliaryDynamics` could work. Or maybe even `exampleCustomAuxDynamics`, although I'm not a fan of abbreviated words for example names.

> Is it normal that osimMocoCustomEffortGoal is in the folder Examples?

This is fine, since it goes along with exampleMocoCustomEffortGoal.

> In tropter, sandbox_finite_diff_colpack has errors. The headers are not properly included and #include <memory> is missing. But there are also links errors. I don't know if it is the goal to have that to work.

This is also fine, sandbox files aren't guaranteed to run.

adamkewley:
My main concern is that integrating this extra feature on top of OpenSim is going to increase its source code + dependency graph by a fair bit. So I'll address those.

## Build Times

Longer build times will make developing OpenSim more time-consuming for all downstream developers. I took basic measurements of build times:

| Description | `master@9342c2` | `moco_feature` |
| - | - | - |
| clean build | 19m0s | 29m56s | 
| test | 1m27s | 2m42s |
| (do nothing but reconf + rebuild) | 4s | 1m59s |
| "typical core workflow" | 12m57s | 19m46s |
| "typical internal workflow" | 21s  | 20s |
| "typical external workflow" | 45s | 55s |

Where:
- Core development workflow: `touch ../OpenSim/Common/Component.h && make -j$(nproc) install`
    - Typical workflow for a developer working on the core data structures of OpenSim. Because of OpenSim's arch, there are a fair amount of headers in the system that can trigger a rebuild of this magnitude (Component/Object are the worst cases, though)
- Typical internal workflow: `touch ../OpenSim/Common/Component.cpp && make -j$(nproc) testComponentInterface`
    - Typical for a developer that is only working on an isolated part of the system that has an isolated unit test suite. Optimal OpenSim-editing workflow.
- Typical external workflow: `touch ../OpenSim/Common/Component.cpp && make -j$(nproc) install`
    - Typical for a developer that is editing OpenSim and then integrating the edited version into something else (e.g. a downstream project, whole-system perf measurements, etc. etc.)

Overall, if people are developing mostly only `.cpp` files that are mostly at the edges of the dependency graph, no change. Developers working on the core of OpenSim can expect a 20-50 % slowdown in their workflow. Developers who rely on CI to "checkpoint" their work (valuable, for complex stuff) will find CI + is 30-50 % slower.

The significance of build time is that it tends to be a quite reliable determinant of a project's agility and ability to improve and add features over time ("The DevOps Handbook" explains this in a casual style, if you're curious). I can only anecdotally confirm the observation. Unfortunately, there is no "quick fix" for this, unless you can prune the headers/dependencies a bit. The measurements are here to make it clear that the team we knew this drawback *before* merging.

## Dependencies

The dependency graph of OpenSim is about to get *a lot* bigger. The 190k lines added is only a subset of that, because the build system pulls other dependencies externally. I'm not going to rag on about dependencies here, because Moco is already developed, decisions have already been made, etc. Plus, everyone probably already knows that I'm strongly against using external dependencies in large projects. So, assuming that's set in stone, my main concerns with the dependencies (as they are) is:

- The way `adolc` is handled is just going to create a bigger problem later.
  - On windows, `feature_moco` is downloading a prebuilt binary for it (from https://sourceforge.net/projects/myosin/files/adolc/adolc_sparse_x64.zip). So the build will break if sourceforge goes down. Using a binary dependency also removes any possibility of changing any flags (optimization, debugging) or porting to any other native architectures (e.g. ARM)
  - On Linux, it's built from source, but relies on `autoconf` and `make`. It also hard-codes the number of build threads (`-j4`), and seems to always rebuild from scratch if the project is reconfigured (this isn't normal in cmake - probably an artifact of the way autoconf is being called)
  - You should try to upstream a CMake patch for it. (...please Adam resist the urge to rag on about-...) This will be a good opportunity to see why pulling in a bunch of external projects can cause headaches for long-term maintenance of a large downstream project (...ok I'll stop...).

- `ipopt`  has similar issues to `adolc`
  - It also has a bunch of `patch` commands in OpenSim's build to ensure it builds with Moco

- `adolc` requires using `Makefile`s, but invokes `${CMAKE_MAKE_PROGRAM}` to do so. `${CMAKE_MAKE_PROGRAM}` can evaluate to (e.g.) `ninja`, which will break the `adolc` build
  - Just hard-code it as `make` if you need to use `make`

-

adamkewley:
Just to clarify - overall, the way in which the build + dependency management is done in this PR is in-line with OpenSim's existing style (no surprise, given the author). So, overall, it seems fine.

The build measurements were made to put concrete numbers to my abstract concern, which is that OpenSim is likely to be slower to develop (for some workflows) post-merge. Those numbers are useful for me to have, because if someone asks "why does it take hours you to develop a minor change in `Component.h`" then there is a concrete, measured, reason that I can point to and say "that's why". One can simply add up the number of attempts required to make the feature work, the number of PR edits, other testing passes, etc. and use that as a coefficient to `${BUILD_TIME}` - that's the amount of time it will take to ship something.

For dependencies, my point is not that dependencies are in-and-of-themselves bad. It's that large projects can become stymied by dependency issues as they grow. This becomes *really* bad when the dependency tree goes deep (not quite the case here), because of transitives, but can still be quite bad for a broad-but-shallow distribution. 

The issue with dependencies is that they add a communication firewall between the downstream project and the upstream dependency, where the downstream project has to work around the dependency's quirks, or PR upstream and deal with their BS. Dependencies also remove an element of control (e.g., OpenSim is about to require `make`) and this is one reason why large commercial projects tend to freeze a dependency into their tree and treat it like their own (effectively, copy + pasting it in). 

A more eloquent explanation of this can be found in an ancient, but famous, blog post: https://www.joelonsoftware.com/2001/10/14/in-defense-of-not-invented-here-syndrome/

nickbianco:
Thanks @adamkewley for the detailed breakdown. Regarding build times, we've had several conversations weighing the pros and cons of integrating Moco into opensim-core, and this is one of the main cons we're willing to accept considering the functionality that Moco provides. And in many ways, this merge will speed development since Moco and OpenSim will stay in lock-step, compared to going back and forth between updating Moco and OpenSim in separate repos (this definitely slowed Moco development at some stages).

Regarding the dependencies, the CMake flags `OPENSIM_WITH_TROPTER` and `OPENSIM_WITH_CASADI` can provide some flexibility for developers not concerned with Moco. However, your concerns about relying on external dependencies are definitely noted, and we could improve how we handle `adolc` and `ipopt` as you suggest. `adolc` is perhaps the least used dependency of Moco, and it's possible we may not need it in the future. 

@aymanhab, @carmichaelong, not sure you have any thoughts to add here.

chrisdembia:
Providing a build of ipopt on Windows for MSVC through vcpkg or conda would be a huge benefit to the whole world. @adamkewley you might be able to contribute here.

aymanhab:
Thanks for the great feedback @adamkewley indeed we looked into the pros/cons and agreed the pros outweigh the cons for the community even though we pay a price in maintenance and dependencies. In terms of your typical workflow above, you can turn off building of both solvers (OPENSIM_WITH_TROPTER, OPENSIM_WITH_CASADI) and you should get time closer to master branch. This was planned to be the regular development environment for work unrelated to Moco. Can you try that?

@nickbianco do we have documentation of when users need TROPTER vs. CASADI? Is it likely that use cases of one of them will overtake the other? if so in what time frame?

nickbianco:
CasADi has already mostly overtaken Tropter. Tropter is better for certain cases, like when solving a parameter optimization problem that requires an initSystem call, but for most users CasADi is the better option. 

This is described in the subsection "MocoCasADiSolver versus MocoTropterSolver" on the "MocoStudy" page of the docs.

aymanhab:
@nickbianco should I expect the test case testMocoInterface to fail if neither Tropter nor Casadi are built? Thanks

nickbianco:
@aymanhab testMocoInterface should fail since we do solve a few short optimization problems in that test file.

nickbianco:
@aymanhab actually, nevermind. We omit tests that rely on the solvers when they're not built, so testMocoInterface should pass.

nickbianco:
@ayman I re-enabled the Java test that was failing due to the MarkersReference issue: #2889.

aymanhab:
Great, thanks @nickbianco

aymanhab:
The crash happens in OpenSim::ObjectProperty<OpenSim::MarkersReference>::setValueVirtual which occurs while trying to overwrite a MarkerReference property with a passed in one.

aymanhab:
@nickbianco in both c++ and Java test, the memory handling is problematic since a MarkerReference is created on the stack and then passed to an object that takes ownership!

nickbianco:
We currently use pre-built binaries for IPOPT on Windows (although conda-forge now includes IPOPT, so we might be able to fix this: opensim-org/opensim-moco#557). 

Neither ColPack nor CasADi are pre-built. ColPack is only needed for tropter, so if you're debugging for the CasADi solver you can exclude it. CasADi depends on IPOPT, is that why CasADi fails?

aymanhab:
Thanks @nickbianco I turned off CASADI and now building with Tropter, and can get past the configuration on cmake. Will keep you posted if I run into issues.

aymanhab:
@nickbianco Build of Tropter/debug fails because it links ipopt which is not built debuggable, am I missing something here or is it impossible to have a debuggable environment without building ipopt from source? Thank you

nickbianco:
Apologies, both CasADi and tropter depend on Ipopt. So yes, I guess it's not possible.

nickbianco:
@aymanhab, regarding the MarkersReference issue, do you know why we see the same errors in C++ as in Java. For example, exampleMarkerTracking works fine when I run it in C++, and it uses the MarkersReference interface.

aymanhab:
@nickbianco That's part of why I wanted to build debuggable environment to step through the CPP side. For now don't do anything about it until I get to the bottom of the issue

aymanhab:
@nickbianco Finally stepping through the code in RelWithDebInfo, I think the C++ code works fine because the set_MarkersReference call ends up making a clone of the object. There's an issue with the Bindings that causes the object to have the wrong function table and so creating the clone crashes. No need for other refactoring AFAICT. Do you know if this is Java only or Python issue as well?

nickbianco:
@aymanhab this issue occurs on Python as well (I originally encountered it while using Python).

aymanhab:
Finally got to the bottom of this issue, we tell SWIG to use MarkersReference as a shared_ptr, so SWIG mistakenly convert MarkersReference to shared_ptr<MarkersReference> implicitly causing crash downstream. Working on a fix, likely will remove MarkersReference from this interface.

nickbianco:
@aymanhab thanks for finding and fixing the issue! I still haven't figured out the issue with Metis when building Ipopt on Mac.

nickbianco:
@aymanhab, @carmichaelong any final comments or suggestions?

carmichaelong:
I just reviewed the discussion in this PR and related PRs that have been pulled in here, and it looks like everything was addressed and CI is happy. I have no further comments, LGTM!

nickbianco:
@aymanhab, would like to merge this soon. Just making sure you had a chance for any final thoughts or comments.

aymanhab:
Sure, will do @nickbianco

aymanhab (approved):
Thanks @nickbianco @chrisdembia and @carmichaelong I'll go ahead and merge 🍾 🎆

aymanhab:
Will leave branch for now just in case we need to track something back here.

aymanhab:
@nickbianco Mac build fails, unrelated to code changes since last successful build, any idea why?

aymanhab:
This seems related to gfortran AFAICT

nickbianco:
Woohoo! Thanks @aymanhab. 

Looking into the gfortran issue.

Issue #2896, opened 2020-10-22
Update FV curve to match changes in inverse FV curve in MillardEquilibrium muscle


carmichaelong:
@mjhmilla I missed something in my review that caught my eye while checking out the changes after #2790 in the GUI. I noticed that the FV curve was not being updated even though the log would note a change. I tried adding a line to update the FV curve at https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L169 but then some tests with the equilibrium muscle fail. 

```
1>The following tests FAILED:
1>	 42 - testOutputReporter (Failed)
1>	 76 - testStaticOptimization (Failed)
1>	 82 - testCMCArm26_Millard (Failed)
1>	 86 - testCMCGait10dof18musc (Failed)

mjhmilla:
Here's my suspicion (I'm replicating this right now):
  1. These tests compare a calculated solution against a pre-computed solution.
  2. Since the force-velocity curve has been altered, the calculated solution will not match IF the simulations go into the region between eccNearVmax and eccAtVmax, and concNearVmax and concAtVmax.

I propose the following:

1. Keep the original behaviour: the force-velocity curve is not modified beyond the original calls to setCurveShape on lines 131 & 146. The tests will then pass & previous simulations will also still look the same.

2. The log statements should be updated: log statements added to lines 131 & 146 (currently missing), update the log statements on 154 & 162 to apply to the force-velocity-inverse curve. Update the variable names conSlopeAtVmax and eccSlopeAtVmax to conSlopeFvInvAtVmax and eccSlopeAtFvInvVmax: then its clearer to developers that these internal variables only apply to the force velocity inverse curve.

I hope to have this fixed and pulled later today.

mjhmilla:
I've just issued a pull request linked to this issue.

nickbianco:
Closing per #2897.

PR #2897, opened 2020-10-22
fix for issue #2790 & #2896

mjhmilla:
Code updated so that extendFinalizeFromProperties() only updates force-velocity curve properties if needed, logs the changes, and sets the end slopes for the inverse-force-velocity curve to acceptable values.

Fixes issue #2896

### Brief summary of changes

Issue #2790 : Problem: a user could manually set the end slopes of the force-velocity curve to legal values that would cause a problem. Why? Internally one of those values was set to a hard-coded constant: this was done in an effort to make sure the classic Hill model (if used) would not be too stiff to integrate.

Issue #2896 Problem with my original fix to #2790: the code looked like it was missing a call to update the force-velocity curve: upon doing this a new curve resulted. Since this curve differed from the one in the past, tests that compared simulated results to saved solutions failed. Most of this bug is due to ambiguous code.

Fix details

#2790
1. Additional variables are added for the end slopes of the force-velocity inverse curve: conSlopeAtVmaxFvInv and eccSlopeAtVmaxFvInv. These variables have been clearly named with "FvInv" appended to the end so that code is easier to understand. These end slopes are not set to fixed values (to avoid the problem noted in issue #2790 but instead are set to half of what ever the "near" end slopes are. This removes the need to have hard-coded limits and fixes the bug noted in #2790.

2. There is a risk now that the user unwittingly chooses very small values for eccSlopeNearVmax and conSlopeNearVmax: this will have the consequence of making numerical simulation slow, in the worst case impossible. A warning message is printed to the log if the values of the end slopes are legal, but are small enough that integration might be slow. Suggestions for fixing the problem are noted in the message.

#2896
3. My initial fix for issue #2790 looked incomplete because I updated force-velocity parameters, but didn't update the curve. When @carmichaelong added code to update the force velocity curve tests failed: because the new force-velocity curve was different from the one in the past tests that compare simulation data to previous solutions failed. This is a `bug' because of my unnecessary updates to some of the force-velocity parameters, and because of some ambiguous variable names: some of those parameters were only intended to be used with the inverse-force velocity curve.  To fix this, first, I've introduced extra variables that clearly indicate in the name that they are only intended for inverse-force velocity curve. Second, I've updated the code so that parameter values for the force-velocity curve are updated only if absolutely necessary: these updates are noted in the log. Now the force-velocity curve matches the one from the past, updates are made sparingly (and with log notes), and internal variable

### Testing I've completed

Run all tests in release mode. I've also manually stepped through the code just to be sure.

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2897)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ This is not a comprehensive list of changes but rather a hand-curated collection
 
 v4.2
 ====
-- Fixed a bug in Millard2012EquilibriumMuscle::extendFinalizeFromProperties(): the end point slopes on the force velocity curves are constrained to yield a valid curve. This does slightly change the values of the force velocity curve near the extremes (between normalized fiber velocities of 0.9 to 1.0, and -0.9 to -1.0).
+- Fixed a bug in Millard2012EquilibriumMuscle::extendFinalizeFromProperties(): the end point slopes on the inverse force velocity curves are constrained to yield a valid curve. A warning is noted in the log if the slopes are small enough that numerical integration might be slow.
 - Added logging to Millard2012EquilibriumMuscle::extendFinalizeFromProperties(): whenever an internal setting is changed automatically these changes are noted in the log.
 - Introduced new logging system based on spdlog https://github.com/gabime/spdlog.git. The transition should be transparent to end users with default settings except that the name of the log file is now opensim.log. Main features are:
   - The ability to customize error level for reporting (in increasing level of verbosity): Off, Critical, Error, Warn, Info, Debug, Trace 
--- a/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
+++ b/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
@@ -102,6 +102,18 @@ void Millard2012EquilibriumMuscle::extendFinalizeFromProperties()
     double eccCurviness     = fvCurve.getEccentricCurviness();
     double eccForceMax      = fvCurve.getMaxEccentricVelocityForceMultiplier();
 
+//Ensure that the slopes near vmax are postive, finite, and above zero
+    OPENSIM_THROW_IF_FRMOBJ(conSlopeNearVmax < SimTK::SqrtEps,
+            InvalidPropertyValue, 
+            "ForceVelocityCurve:concentric_slope_near_vmax",
+            "Slope near concentric vmax cannot be less than SimTK::SqrtEps");
+
+    OPENSIM_THROW_IF_FRMOBJ(eccSlopeNearVmax < SimTK::SqrtEps,
+            InvalidPropertyValue, 
+            "ForceVelocityCurve:eccentric_slope_near_vmax",
+            "Slope near eccentric vmax cannot be less than SimTK::SqrtEps");
+
+
     // A few parameters may need to be adjusted to avoid singularities (e.g., if
     // an elastic tendon is used with no fiber damping).
     if(!get_ignore_tendon_compliance() && !use_fiber_damping) {
@@ -115,61 +127,94 @@ void Millard2012EquilibriumMuscle::extendFinalizeFromProperties()
             "Minimum control cannot be less than minimum activation");
 
         if (falCurve.getMinValue() < 0.1){
-            log_info("'{}' ActiveForceLengthCurve parameter updated to avoid "
-                     "error: '{}' was {} but is now {}.", getName(),
-                     "minimum_value", falCurve.getMinValue(), 0.1);
+            log_info("'{}' Parameter update for classic Hill model: "
+                "ActiveForceLengthCurve parameter '{}' was {} but is now {}.", 
+                getName(),"minimum_value", falCurve.getMinValue(), 0.1);
             falCurve.setMinValue(0.1);
         }
-        if (conSlopeAtVmax < 0.1 || eccSlopeAtVmax < 0.1){
-            log_info("'{}' ForceVelocityCurve parameter updated to avoid error:"
-                     " '{}' was {} but is now {}.", getName(),
-                     "concentric_slope_near_vmax", conSlopeAtVmax, 0.1);
-            log_info("'{}' ForceVelocityCurve parameter updated to avoid error:"
-                     " '{}' was {} but is now {}.", getName(),
-                     "eccentric_slope_near_vmax", eccSlopeAtVmax, 0.1);
+        if (   conSlopeAtVmax < conSlopeNearVmax*0.5){
+            log_info("'{}' Parameter update for classic Hill model: "
+                "ForceVelocityCurve parameter '{}' was {} but is now {}.", 
+                getName(),"eccentric_slope_near_vmax", conSlopeAtVmax, 
+                conSlopeNearVmax*0.5);            
+            conSlopeAtVmax  = conSlopeNearVmax*0.5;
+            if(conSlopeNearVmax < 0.05){               
+                log_info("'{}': Warning slow simulation: classic Hill model"
+                    "is being used with a '{}' of {}. Use the damped model, or "
+                    "increase to {} for faster simulations.", getName(),
+                    "concentric_slope_near_vmax",conSlopeNearVmax, 0.05);                 
+            }
+        }
+        if (   eccSlopeAtVmax < eccSlopeNearVmax*0.5){
+            log_info("'{}' Parameter update for classic Hill model: "
+                "ForceVelocityCurve parameter '{}' was {} but is now {}.", 
+                getName(),"eccentric_slope_near_vmax", eccSlopeAtVmax, 
+                eccSlopeNearVmax*0.5);
+            eccSlopeAtVmax   = eccSlopeNearVmax*0.5;
+            
+            if(eccSlopeNearVmax < 0.05){
+                log_info("'{}': Warning slow simulation: classic Hill model"
+                    "is being used with a '{}' of {}. Use the damped model, or "
+                    "increase to {} for faster simulations.", getName(),
+                    "eccentric_slope_near_vmax",eccSlopeNearVmax, 0.05);                
+            }
 
-            fvCurve.setCurveShape(0.1, conSlopeNearVmax, isometricSlope,
-                                  0.1, eccSlopeNearVmax, eccForceMax);
         }
 
-    } else { //singularity-free model still cannot have excitations below 0
+        fvCurve.setCurveShape(
+                conSlopeAtVmax, conSlopeNearVmax, isometricSlope,
+                eccSlopeAtVmax, eccSlopeNearVmax, eccForceMax);  
+
+
+    } else { 
+
+        if(get_minimum_activation() > 0.){
+            log_info("'{}': Parameter update for the damped-model: "
+                "minimum_activation was {} but is now {}",
+                   getName(), get_minimum_activation(), 0.);
+
+            set_minimum_activation(clamp(0, get_minimum_activation(), 1));
+        }
+        if(falCurve.getMinValue() > 0.0){
+            log_info("'{}' Parameter update for the damped-model: "
+                "ActiveForceLengthCurve minimum value was {} but is now {}.",
+                 getName(), falCurve.getMinValue(), 0.);
+            falCurve.setMinValue(0.0);
+        }
+
+        if(conSlopeAtVmax > 0.){
+            log_info("'{}' Parameter update for the damped-model:"
+                " ForceVelocityCurve  '{}' was {} but is now {}.", 
+                getName(),"concentric_slope_at_vmax", conSlopeAtVmax, 0.);                        
+            conSlopeAtVmax  = 0.;
+        }
+        if(eccSlopeAtVmax > 0.){            
+            log_info("'{}' Parameter update for the damped-model: "
+                "ForceVelocityCurve '{}' was {} but is now {}.", 
+                getName(),"eccentric_slope_at_vmax", eccSlopeAtVmax, 0.);            
+            eccSlopeAtVmax   = 0.;
+        }
+
+        fvCurve.setCurveShape(
+            conSlopeAtVmax, conSlopeNearVmax, isometricSlope,
+            eccSlopeAtVmax, eccSlopeNearVmax, eccForceMax);
+
         OPENSIM_THROW_IF_FRMOBJ(get_minimum_activation() < 0.0,
             InvalidPropertyValue, getProperty_minimum_activation().getName(),
             "Minimum activation cannot be less than zero");
-
+        
+        //singularity-free model still cannot have excitations below 0
         OPENSIM_THROW_IF_FRMOBJ(getMinControl() < get_minimum_activation(),
             InvalidPropertyValue, getProperty_min_control().getName(),
             "Minimum control cannot be less than minimum activation");
-
-        set_minimum_activation(clamp(0, get_minimum_activation(), 1));
-        falCurve.setMinValue(0.0);
-        fvCurve.setCurveShape(0.0, conSlopeNearVmax, isometricSlope,
-                              0.0, eccSlopeNearVmax, eccForceMax);
     }
 
-    if (conSlopeAtVmax < 0.5*conSlopeNearVmax ) {
-        //Using this update preserves the shape of the rest of the force
-        //velocity curve and guarantees that the condition that
-        //conSlopeAtVmax < conSlopeNearVmax is satisfied
-      log_info("'{}' ForceVelocityCurve parameter updated: "
-               "'{}' was {} but is now {}.",
-               getName(),"concentric_slope_at_vmax",
-               conSlopeAtVmax, 0.5*conSlopeNearVmax);
-      conSlopeAtVmax = 0.5*conSlopeNearVmax;
-    }
-    if (eccSlopeAtVmax < 0.5*eccSlopeNearVmax ) {
-        //ditto
-      log_info("'{}' ForceVelocityCurve parameter updated: "
-               "'{}' was {} but is now {}.",
-               getName(),"eccentric_slope_at_vmax",
-               eccSlopeAtVmax, 0.5*eccSlopeNearVmax);
-      eccSlopeAtVmax = 0.5*eccSlopeNearVmax;
-
-    }
+    double eccSlopeAtVmaxFvInv  = eccSlopeNearVmax*0.5;
+    double conSlopeAtVmaxFvInv  = conSlopeNearVmax*0.5;
 
 
-    fvInvCurve = ForceVelocityInverseCurve(conSlopeAtVmax, conSlopeNearVmax,
-                                           isometricSlope, eccSlopeAtVmax,
+    fvInvCurve = ForceVelocityInverseCurve(conSlopeAtVmaxFvInv, conSlopeNearVmax,
+                                           isometricSlope, eccSlopeAtVmaxFvInv,
                                            eccSlopeNearVmax, eccForceMax,
                                            conCurviness, eccCurviness);
 
--- a/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
+++ b/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
@@ -217,28 +217,45 @@ void Millard2012EquilibriumMuscle::extendFinalizeFromProperties()
             "Minimum control cannot be less than minimum activation");
     }
 
-    //The ForceVelocityInverseCurve is used by the classic Hill model: this 
-    //formulation is used when use_fiber_damping is false. When an elastic
-    //tendon is used Eqn. 6 of Millard et al. is used to evaluate the state
-    //derivative of fiber length. As noted in the paper Eqn. 6 has several
-    //singularites one of which is related to the force-velocity-inverse curve:
-    //simply, the force-velocity-inverse curve must be invertible. 
+
+    //The ForceVelocityInverseCurve is used with both the damped and the 
+    // classic Hill model but in different ways:
     //
-    //This is equivalent to saying that the force-velocity-inverse curve must 
-    //have finite endslopes. To avoid this singularity we set the end slopes of 
-    //the force-velocity-inverse curve to be a fraction of the `near' slopes
-    //used by the force-velocity curve. While this may seem confusing don't 
-    //forget: the ForceVelocityInverseCurve takes in the parameters of the 
-    //force-velocity-curve that it inverts. So if you pass in an endslope of 
-    //0 the inverse curve will have an endslope of 1/0 = infinity! Instead we 
-    //make sure that we pass in a finite value for the end slopes so that the 
-    //inverse curve has finite endslopes. 
+    // The damped model uses the fvInvCurve to provide an initial fiber 
+    // velocity prior to numerically solving the root using Eqn. 8 from 
+    // Millard et al.
     //
-    //This means that fvInvCurve is not the inverse of fvCurve: the end slopes
-    //differ. This has to be the case because the fvCurve is used by the damped
-    //fiber model. This model has no singularities in its state derivative
-    //(Eqn. 8 of Millard et al.) and so a user is permitted to set the end
-    //slopes of the fvCurve to zero.
+    // The classic Hill formulation uses the fvInvCurve to directly solve for 
+    // fiber velocity using Eqn. 6 from Millard et al.
+    //
+    //While the fvCurve and fvInvCurve are inverses of each other between
+    //concentric-velocity-near-vmax to eccentric-velocity-near-vmax these
+    //curves differ outside of this region. Why? When an elastic-tendon model
+    //is used with the classic Hill formulation Eqn. 6 of Millard et al. is 
+    //used to evaluate the state derivative of fiber length. As noted in the 
+    //paper Eqn. 6 has several singularites one of which is related to the 
+    //force-velocity-inverse curve: simply, the force-velocity-inverse curve 
+    //must be invertible.
+    //
+    //In this case this is equivalent to saying that the force-velocity-inverse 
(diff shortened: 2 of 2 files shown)


carmichaelong:
Thanks for the detailed investigation @mjhmilla! I skimmed over the PR (will give it a more detailed look either today or tomorrow) and I think I now see some of the nuances you're trying to account for.

A couple high-level questions:
- What are the downsides (or maybe what are the tradeoffs) for having a FV curve and an inverse FV curve differ by a little (specifically by [this line's](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L213) difference)? 
- Do you think this fix (i.e., making sure the curves still pass previous regression tests) is the correct long term fix, or in the long run should we consider updating the standards? I feel like the trade-off here is keeping behavior the same for users is a good thing, but if it's a true "bug" (which I'm not sure if it is or not) then we might want to consider something down the road.

mjhmilla:
Hey Carmichael,

What's the difference?

Before the end slopes of the force-velocity-curve-inverse were hard-coded to be 0.1: this ultimately is what caused the problems that led to issue #2790. Now the end slopes of force-velocity-curve-inverse are set to be half the value of the near-end-slopes AND there is an error thrown if the near-end-slopes are less than SimTK::SqrtEps (lines 106 and 111). This is all that has changed: the end slopes of the force-velocity-inverse curve.

What's the downside?

The downside is that users can, if they wish, set the near-end-slope values low enough (say 0.01) that simulations will run slowly if they are using the classic Hill-formulation. If they are using the damped model there is no draw-back at all: the force-velocity-inverse curve is not used for the damped model. In case there are people out there who really want to use the classic formulation I've added a log message (line 142 & 156) to warn them. The defaults for the curves are just fine: a user will have to explicitly change these values to get into trouble.

Is this the right way to proceed long term?

Yes, I think it is. Why? This is consistent with the original publication. To provide a better answer, by default the damped model uses a force-velocity curve that has end slopes of zero beyond the maximum shortening/lengthening velocities. This was one of the major features of the model due to its singularity free formulation. This is more realistic on the shortening end because otherwise the muscle can push. On the lengthening end it's really hard to comment: compared to biological muscle Hill models do really poorly here. However, it does have the benefit of putting a hard limit on the forces a fiber can sustain during active lengthening.

carmichaelong (changes requested):
Thanks for the detailed response, @mjhmilla! This was super helpful for assessing the PR. Overall this looks good, and I have mostly some minor comments and some readability comments too. 

One more quick question for you: Any thoughts on how default values (either for the muscle class itself or in the models we distribute) should change? I'm thinking it would be good to find a balance to reduce some of the messages that users see versus changing answers, so that users don't freak out too much when previously unseen messages popup (even though we were making changes to the model in the first place anyway!). Of course these changes need not be included in this PR.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-:-MKa0XvDENBJO3Kqy-6c:b-s75esc)* status: 0 of 2 files reviewed, 7 unresolved discussions (waiting on @mjhmilla)

___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 109 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MKa0ho9CZPPN1Y9zjLs:-MKa0ho9CZPPN1Y9zjLt:b-ttoan7) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L109)):*
> ```cpp
>             InvalidPropertyValue, 
>             "ForceVelocityCurve:concentric_slope_near_vmax",
>             "Slope near concentric vmax cannot be less than SimTK::SqrtEps");
> ```

Might be confusing to a user to know what to do if they get a message about `SimTK::SqrtEps`. Perhaps add the numeric value in parenthesis after? (in message below too)
___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 135 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MK_yivh1d6IKIkE1x5u:-MK_yivh1d6IKIkE1x5v:b-ty0i5u) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L135)):*
> ```cpp
>             falCurve.setMinValue(0.1);
>         }
>         if (   conSlopeAtVmax < conSlopeNearVmax*0.5){
> ```

For readability it would be great to clean up some of the white space around the `if` statements in this section, i.e.: `if (stuff) {`
___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 141 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MK_yPWjDlsPVRBWYt7X:-MK_yPWjDlsPVRBWYt7Y:b-p5xb6g) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L141)):*
> ```cpp
>                 conSlopeNearVmax*0.5);            
>             conSlopeAtVmax  = conSlopeNearVmax*0.5;
>             if(conSlopeNearVmax < 0.05){               
> ```

This `if` statement is only checked if `conSlopeAtVmax < conSlopeNearVmax*0.5`. Should this always be checked instead?
___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 155 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MK_yZEc0veyQxj98G5i:-MK_yZEc0veyQxj98G5j:b-1buxi7) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L155)):*
> ```cpp
>             eccSlopeAtVmax   = eccSlopeNearVmax*0.5;
>             
>             if(eccSlopeNearVmax < 0.05){
> ```

Same question/comment as above (should this `if` statement be checked regardless of previous if statement)?
___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 171 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MK_yzPg6ErfLb-tGlZq:-MK_yzPg6ErfLb-tGlZr:b-fub0r0) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L171)):*
> ```cpp
>     } else { 
> 
>         if(get_minimum_activation() > 0.){
> ```

Should this be less than (otherwise you'll emit a `log_info` message for min activations even for unaffected values like 0.1)?

Also, this log only covers the `a < 0` case, but clamping could bring a high number down to 1 as well. Perhaps the check should be for both `a < 0 || a > 1`, and then the message could just say it was clamped to between 0 and 1?
___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 202 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MKa-lsP1xC1NQtrXn-P:-MKa-lsP1xC1NQtrXn-Q:b-ypi9ba) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L202)):*
> ```cpp
>             eccSlopeAtVmax, eccSlopeNearVmax, eccForceMax);
> 
>         OPENSIM_THROW_IF_FRMOBJ(get_minimum_activation() < 0.0,
> ```

In theory if activation clamping happens above, this check wouldn't be needed, but OK to keep for safety if you think it's useful
___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 212 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MKa0Al83xyRMXFPGHDA:-MKa0Al83xyRMXFPGHDB:blml84l) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L212)):*
> ```cpp
>     }
> 
>     double eccSlopeAtVmaxFvInv  = eccSlopeNearVmax*0.5;
> ```

Could you add a comment here for maintainability to indicate why this is needed and that it's okay for inverse curve to be different? It would help a developer like me who was a little too brash in thinking they needed to be the same :)

mjhmilla (commented):
I've answered the minor comments below.

In regards to the default values if I run testMuscles I see log messages mentioning that the minimum_activation and the ActiveForceLengthCurve minimum has changed. I don't see any others. Are there others that you see?

These messages appear because the parameters are set by default to be compatible with the classic Hill model. When the damped model is used these parameters are automatically updated (hence the log message) to spare the user the painful task of updating these manually.

We could re-configure the model so that by default the parameters are compatible with the damped model. If they switch to the Hill model with an elastic tendon then we can update the 2 parameters mentioned above and issue log messages. I've tested this on my machine: it requires updates to Millard2012EquilibriumMuscle.cpp::constructProperites, Millard2012EquilibriumMuscle.cpp::extendFinalizeFromProperties, and ActiveForceLengthCurve::constructProperties.

This has the desired effect (the log messages go away) but causes two tests to fail:
  testSerializableMuscleCurves: the default value has changed so of course this fails (the defaults are tested)
  testOptimizationExample: fails because the saved test trajectory differs from the new one

If you would like the log messages to go away then these two tests need to be updated. And I presume that someone would need to look at the numerical results of the new results from testOptimizationExample to make sure that they still make sense.

If you'd like to pursue this let me know. I'm happy to tack this onto this pull request if you think that makes the most sense. Otherwise please issue a new pull request.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-:-MKeexYM-WRXUMxtcje3:b-4x5shr)* status: 0 of 2 files reviewed, 7 unresolved discussions (waiting on @carmichaelong)

___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 109 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MKa0ho9CZPPN1Y9zjLs:-MKeQh6M6fOhgIazGHe8:b6gdz90) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L109)):*

Might be confusing to a user to know what to do if they get a message about `SimTK::SqrtEps`. Perhaps add the numeric value in parenthesis after? (in message below too)
</blockquote></details>

Done. The text reads "... SimTK::SqrtEps (1.49e-8)" in the two locations in which SimTK::SqrtEps is mentioned.
___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 135 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MK_yivh1d6IKIkE1x5u:-MKeS9o_6x_43OeGRjan:bhq76ct) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L135)):*

For readability it would be great to clean up some of the white space around the `if` statements in this section, i.e.: `if (stuff) {`
</blockquote></details>

Done: I've eliminated the white space around the arguments in the if statements in extendFinalizeFromProperties.
___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 141 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MK_yPWjDlsPVRBWYt7X:-MKeSR6u1wk6An62Iqiw:b-vit840) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L141)):*

This `if` statement is only checked if `conSlopeAtVmax < conSlopeNearVmax*0.5`. Should this always be checked instead?
</blockquote></details>

Correct, these if statements should always be checked! I've un-nested these if statements.
___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 155 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MK_yZEc0veyQxj98G5i:-MKeS_qfDZ1dNYVlq1AG:b47jx5n) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L155)):*

Same question/comment as above (should this `if` statement be checked regardless of previous if statement)?
</blockquote></details>

Also correct: these statements should always be checked. Both have been un-nested.
___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 171 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MK_yzPg6ErfLb-tGlZq:-MKeSk9F8G3EL3XQsKCs:b-45ijhm) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L171)):*

Should this be less than (otherwise you'll emit a `log_info` message for min activations even for unaffected values like 0.1)?

Also, this log only covers the `a < 0` case, but clamping could bring a high number down to 1 as well. Perhaps the check should be for both `a < 0 || a > 1`, and then the message could just say it was clamped to between 0 and 1?
</blockquote></details>

Should this be less than (otherwise you'll emit a log_info message for min activations even for unaffected values like 0.1)?

----------
This does replicate the original functionality: by default the minimum activation is above zero and is compatible with the classic Hill formulation; when the damped model is used the value for this parameter is set to zero. 

The other alternative is to set the default minimum activation to a_min = 0. Update it to some value Z > 0 if the classic Hill model with an elastic tendon is being used (and issue a log message). We could use Z = 0.01 which is what appears in constructProperties. This could work.
----------

Also, this log only covers the a < 0 case, but clamping could bring a high number down to 1 as well. Perhaps the check should be for both a < 0 || a > 1, and then the message could just say it was clamped to between 0 and 1?
----------

I've updated the log message. 

Note: clamping is only applied if a_min > 0: the intent is actually to set a_min = 0, but only if necessary. I had not ever thought that someone would use a value of a_min that was so large that it would be clamped to 1.0.  If a < 0 (not handled by the clamp) then the OPENSIM_THROW macro throws an error.
----------
___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 202 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MKa-lsP1xC1NQtrXn-P:-MKeUMpe0S_l--iY1deN:bn02ou0) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L202)):*

In theory if activation clamping happens above, this check wouldn't be needed, but OK to keep for safety if you think it's useful
</blockquote></details>

The clamping only happens if the minimum activation is greater than 0! If someone has somehow set the minimum activation to be less than 0 then the only thing that catches this error is the OPENSIM_THROW_IF_FRMOBJ statement.

This statement needs to stay!
___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 212 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MKa0Al83xyRMXFPGHDA:-MKeUozD5jN2gNmp2V-d:b-kwel49) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L212)):*

Could you add a comment here for maintainability to indicate why this is needed and that it's okay for inverse curve to be different? It would help a developer like me who was a little too brash in thinking they needed to be the same :)
</blockquote></details>

I've added a curve explaining why the fvCurve and the fvInvCurve are not inverses of eachother. This is a pretty detailed comment and refers to the relevant equations in the "Flexing computational muscle: ..." paper directly. 

OpenSim will be lucky to have brash developers like you in the future : )

mjhmilla (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2897)* status: 0 of 2 files reviewed, 7 unresolved discussions (waiting on @carmichaelong)

___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 171 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MK_yzPg6ErfLb-tGlZq:-MKeo0201vK043cV4t_4:b-lhcois) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L171)):*
<details><summary><i>Previously, mjhmilla wrote…</i></summary><blockquote>

Should this be less than (otherwise you'll emit a log_info message for min activations even for unaffected values like 0.1)?

----------
This does replicate the original functionality: by default the minimum activation is above zero and is compatible with the classic Hill formulation; when the damped model is used the value for this parameter is set to zero. 

The other alternative is to set the default minimum activation to a_min = 0. Update it to some value Z > 0 if the classic Hill model with an elastic tendon is being used (and issue a log message). We could use Z = 0.01 which is what appears in constructProperties. This could work.
----------

Also, this log only covers the a < 0 case, but clamping could bring a high number down to 1 as well. Perhaps the check should be for both a < 0 || a > 1, and then the message could just say it was clamped to between 0 and 1?
----------

I've updated the log message. 

Note: clamping is only applied if a_min > 0: the intent is actually to set a_min = 0, but only if necessary. I had not ever thought that someone would use a value of a_min that was so large that it would be clamped to 1.0.  If a < 0 (not handled by the clamp) then the OPENSIM_THROW macro throws an error.
----------


</blockquote></details>

Please ignore the large bold face text: I had no idea that this would appear, and I don't mean anything by it. I'm certainly not using the type setting to yell at you : ). But it appears that I cannot edit this text so I'm left writing this message apologizing for my sins.

carmichaelong (changes requested):
Thanks for the detailed responses. Just a few questions/comments back to you, but it looking good overall.

> In regards to the default values if I run testMuscles I see log messages mentioning that the minimum_activation and the ActiveForceLengthCurve minimum has changed. I don't see any others. Are there others that you see?

These were the only ones I saw (sorry that was unclear). I think on the whole getting these messages is good, and that it's better that it's showing up in the damped model (which seems less prevalent in distributed models) rather than in models that have damping off.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-:-MKlJgX0Dc1mnWXzSVPD:bifdzxj)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @mjhmilla)

___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 171 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MK_yzPg6ErfLb-tGlZq:-MKklV17AgVS9_NlJQpo:b-cqzb9f) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L171)):*
<details><summary><i>Previously, mjhmilla wrote…</i></summary><blockquote>

Please ignore the large bold face text: I had no idea that this would appear, and I don't mean anything by it. I'm certainly not using the type setting to yell at you : ). But it appears that I cannot edit this text so I'm left writing this message apologizing for my sins.
</blockquote></details>

No worries on the text formatting, I just assumed something weird happened with markdown so I got a laugh out of it.

I now see the logic between this and the exception below, thanks!
___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 212 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MKa0Al83xyRMXFPGHDA:-MKmiLriBelE0sVfw1kW:b4rftpp) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L212)):*
<details><summary><i>Previously, mjhmilla wrote…</i></summary><blockquote>

I've added a curve explaining why the fvCurve and the fvInvCurve are not inverses of eachother. This is a pretty detailed comment and refers to the relevant equations in the "Flexing computational muscle: ..." paper directly. 

OpenSim will be lucky to have brash developers like you in the future : )

</blockquote></details>

This comment is very helpful. Just one more question on this: The top statement says that `ForceVelocityInverseCurve` is used when fiber damping is not used, however this is outside of the `if` statement loop above. Is this still valid to update here (since it seems like all muscles will be updated)? Reading through the comment, it seems like it is (as it's implied that while the inverse curve is constructed here, it won't be used for a damped model).

mjhmilla (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2897)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @carmichaelong)

___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 212 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MKa0Al83xyRMXFPGHDA:-MKoE_IT677z8g66FVsa:b-fo2lmk) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L212)):*

This comment is very helpful. Just one more question on this: The top statement says that `ForceVelocityInverseCurve` is used when fiber damping is not used, however this is outside of the `if` statement loop above. Is this still valid to update here (since it seems like all muscles will be updated)? Reading through the comment, it seems like it is (as it's implied that while the inverse curve is constructed here, it won't be used for a damped model).
</blockquote></details>

The ForceVelocityInverseCurve is used with both models:

1. The damped model: to provide an initial fiber velocity prior to numerically solving the root to Eqn. 8 from the paper.
2. The classic Hill formulation: to directly solve for fiber velocity using Eqn. 6 from the paper.

So yes, the ForceVelocityInverseCurve needs to be updated no matter what.

mjhmilla (commented):
Ok. In that case log messages are inevitable.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-:-MKoFQ9k5hHMLjyKywJ4:bnncmry)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @carmichaelong)

carmichaelong (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2897)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @mjhmilla)

___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 212 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MKa0Al83xyRMXFPGHDA:-MKvSI-Y8pMfhOFXqLhc:b-o4tkaa) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L212)):*
<details><summary><i>Previously, mjhmilla wrote…</i></summary><blockquote>

The ForceVelocityInverseCurve is used with both models:

1. The damped model: to provide an initial fiber velocity prior to numerically solving the root to Eqn. 8 from the paper.
2. The classic Hill formulation: to directly solve for fiber velocity using Eqn. 6 from the paper.

So yes, the ForceVelocityInverseCurve needs to be updated no matter what.

</blockquote></details>

Thanks for the clarification, that's very helpful. Given this, could you just add a sentence about point 1 from your comment (probably just to the end of the last paragraph before the reference)? It'll just make it super clear that it belongs here after all the if statements. After this, I think this is ready to merge!

mjhmilla (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2897)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @carmichaelong)

___
*[OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp, line 212 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-MKa0Al83xyRMXFPGHDA:-ML2ph1YDVucnl1vmEtR:b-xh1bg8) ([raw file](https://github.com/opensim-org/opensim-core/blob/c47e248ddc6e3631dd4d56774448f7bf084083b5/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp#L212)):*

Thanks for the clarification, that's very helpful. Given this, could you just add a sentence about point 1 from your comment (probably just to the end of the last paragraph before the reference)? It'll just make it super clear that it belongs here after all the if statements. After this, I think this is ready to merge!
</blockquote></details>

I've just overhauled the comment so that the ways that the fvCurve and fvInvCurve are used is clear along with the differences between the two curves.

carmichaelong (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2897#-:-MLA1Nvc5ogUkGiNF1j2:bnfp4nl)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @carmichaelong)

Issue #2190, opened 2018-05-14
openSim 4.0 documentation


idhamari:
The one available online is for [3.3](https://simtk.org/api_docs/opensim/api_docs). I have built openSim 4.0 successfully in my Ubuntu 16.04 machine but the html folder shows "My Project Documentation" web page.

chrisdembia:
Can you explain what you mean by "the html folder shows "My Project Documentation" web page."?

Can you show a screenshot?

idhamari:
It seems I should build the documentation after building openSim then build the install project.
Here are the steps. 
VS: select doxgen sub project then build project only. After that do the same for INSTALL sub project.
Ubuntu: make doxygen && make install

Documentation works now.

PR #2898, opened 2020-11-09
Allow IMUPlacer to run without model_file specification if model is set

aymanhab:
Fixes issue #2190 

### Brief summary of changes
Fix issue reported in testing GUI where setModel is not enough to run IMUPlacer if model file is unspecified.

### Testing I've completed
Ran test suite
### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2898)

### DIFF
--- a/OpenSim/Simulation/OpenSense/IMUPlacer.cpp
+++ b/OpenSim/Simulation/OpenSense/IMUPlacer.cpp
@@ -87,7 +87,7 @@ bool IMUPlacer::run(bool visualizeResults) {
     _calibrated = false;
     // Check there's a model file specified before trying to open it
     if (_model.empty() && get_model_file().size() == 0) {
-        OPENSIM_THROW(Exception, "No model file specified for IMUPlacer.");
+        OPENSIM_THROW(Exception, "No model or model_file was specified for IMUPlacer.");
     }
     if (_model.empty()) { _model.reset(new Model(get_model_file())); }
     TimeSeriesTable_<SimTK::Quaternion> quatTable(
--- a/OpenSim/Simulation/OpenSense/IMUPlacer.cpp
+++ b/OpenSim/Simulation/OpenSense/IMUPlacer.cpp
@@ -86,7 +86,7 @@ bool IMUPlacer::run(bool visualizeResults) {
 
     _calibrated = false;
     // Check there's a model file specified before trying to open it
-    if (get_model_file().size() == 0) {
+    if (_model.empty() && get_model_file().size() == 0) {
         OPENSIM_THROW(Exception, "No model file specified for IMUPlacer.");
     }
     if (_model.empty()) { _model.reset(new Model(get_model_file())); }
--- a/OpenSim/Tools/IMUInverseKinematicsTool.cpp
+++ b/OpenSim/Tools/IMUInverseKinematicsTool.cpp
@@ -211,35 +211,45 @@ void IMUInverseKinematicsTool::runInverseKinematicsWithOrientationsFromFile(
     }
 
     auto report = ikReporter->getTable();
+    // form resultsDir either from results_directory or output_motion_file
+    auto resultsDir = get_results_directory();
+    if (resultsDir.empty() && !get_output_motion_file().empty())
+        resultsDir = IO::getParentDirectory(get_output_motion_file());
+    if (!resultsDir.empty()) {
+        IO::makeDir(resultsDir);
+        // directory will be restored on block exit
+        // by changing dir all other files are created in resultsDir
+        IO::CwdChanger::changeTo(resultsDir);
+        std::string outName = get_output_motion_file();
+        outName = IO::GetFileNameFromURI(outName);
+        if (outName.empty()) {
+            bool isAbsolutePath;
+            string directory, fileName, extension;
+            SimTK::Pathname::deconstructPathname(orientationsFileName,
+                    isAbsolutePath, directory, fileName, extension);
+            outName = "ik_" + fileName;
+        }
+        std::string outputFile = outName;
 
-    IO::makeDir(get_results_directory());
-    std::string outName = get_output_motion_file();
-    if (outName.empty()) { 
-        auto eix = orientationsFileName.rfind(".");
-        auto stix = orientationsFileName.rfind("/") + 1;
-        outName = "ik_" + orientationsFileName.substr(stix, eix - stix);
-    }
-    std::string outputFile = get_results_directory() + "/" + outName;
+        // Convert to degrees to compare with marker-based IK
+        // but only for rotational coordinates
+        model.getSimbodyEngine().convertRadiansToDegrees(report);
+        report.updTableMetaData().setValueForKey<string>("name", outName);
 
-    // Convert to degrees to compare with marker-based IK
-    // but only for rotational coordinates
-    model.getSimbodyEngine().convertRadiansToDegrees(report);
-    report.updTableMetaData().setValueForKey<string>("name", outName);
+        auto fullOutputFilename = outputFile;
+        std::string::size_type extSep = fullOutputFilename.rfind(".");
+        if (extSep == std::string::npos) { fullOutputFilename.append(".mot"); }
+        STOFileAdapter_<double>::write(report, fullOutputFilename);
 
-    auto fullOutputFilename = outputFile;
-    std::string::size_type extSep = fullOutputFilename.rfind(".");
-    if (extSep == std::string::npos) { 
-        fullOutputFilename.append(".mot");
-    }
-    STOFileAdapter_<double>::write(report, fullOutputFilename);
-
-    log_info(
-            "Wrote IK with IMU tracking results to: '{}'.", fullOutputFilename);
-    if (get_report_errors()) {
-        STOFileAdapter_<double>::write(*modelOrientationErrors,
-                get_results_directory() + "/" +
-                        getName() + "_orientationErrors.sto");
-    }
+        log_info("Wrote IK with IMU tracking results to: '{}'.",
+                fullOutputFilename);
+        if (get_report_errors()) {
+            STOFileAdapter_<double>::write(*modelOrientationErrors,
+                    getName() + "_orientationErrors.sto");
+        }
+    } 
+    else
+        log_info("IMUInverseKinematicsTool: No output files are written.");
     // Results written to file, clear in case we run again
     ikReporter->clearTable();
 }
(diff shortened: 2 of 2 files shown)


aymanhab:
@carmichaelong or @antoinefalisse can you review this when you have a chance? Thank you

OpenSim/Simulation/OpenSense/IMUPlacer.cpp
```diff
@@ -86,7 +86,7 @@ bool IMUPlacer::run(bool visualizeResults) {
 
     _calibrated = false;
     // Check there's a model file specified before trying to open it
-    if (get_model_file().size() == 0) {
+    if (_model.empty() && get_model_file().size() == 0) {
         OPENSIM_THROW(Exception, "No model file specified for IMUPlacer.");
```
carmichaelong, line 90: Could be good to update the exception message to reflect the updated if statement.

carmichaelong (changes requested):
Just one quick comment from me

aymanhab (commented):
Thanks @carmichaelong I updated the message. I also cleaned up how/where output files are written because testing from the GUI I found issues where:
1. Files were still written even if no output file or results_dir were specified.
2. error reporting files were written to current folder not respecting results_dir
3. Old code did path manipulation on the spot not utilizing cross platform methods in IO.
These should be fixed now, but let me know if you prefer a separate PR for these changes.
Thank you

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2898#-:-MM2PBR94y5BiOz7i8qa:bwtprn2)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @carmichaelong)

___
*[OpenSim/Simulation/OpenSense/IMUPlacer.cpp, line 90 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2898#-MLi0PvDpRZ-F69QUjKN-r1-90:-MM2P39D6jkqHHt--sg1:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/447371daae01aadeabc5421b2d2a0569b54e52e7/OpenSim/Simulation/OpenSense/IMUPlacer.cpp#L90)):*

Could be good to update the exception message to reflect the updated if statement.
</blockquote></details>

Done.

OpenSim/Tools/IMUInverseKinematicsTool.cpp
```diff
@@ -211,35 +211,45 @@ void IMUInverseKinematicsTool::runInverseKinematicsWithOrientationsFromFile(
...
-    }
+        log_info("Wrote IK with IMU tracking results to: '{}'.",
+                fullOutputFilename);
+        if (get_report_errors()) {
+            STOFileAdapter_<double>::write(*modelOrientationErrors,
+                    getName() + "_orientationErrors.sto");
+        }
+    } 
+    else
+        log_info("IMUInverseKinematicsTool: No output files are written.");
```
carmichaelong, line 252: Should this case actually just throw an Exception instead of logging that no files were written?

I think this will also be called if both `results_dir` and `output_motion_file` are used? It may be useful to have a separate message for this.

If there's a use case for this (e.g., running this without outputting files) I'd just add one more statement here to give the user a reason why files were written so that they can fix it if it's a problem. Perhaps "Either the results directory or output_file_name must be specified to write to file."

carmichaelong (commented):
Thanks @aymanhab, it's OK with me to add these other pieces to the PR. See one comment about the control flow and error messaging for users.

OpenSim/Tools/IMUInverseKinematicsTool.cpp
aymanhab: Thanks for the review @carmichaelong The use case I ran into was running from the GUI with neither output files nor results_directory specified, so results are displayed in GUI. Despite that, spurious  .mot files were created in random places. I agree there's a potential for inconsistency (e.g. results_dir specified while output file has a full path specified). In this case the code gives precedence to results_dir though a warning may be justified.

OpenSim/Tools/IMUInverseKinematicsTool.cpp
```diff
@@ -211,35 +211,45 @@ void IMUInverseKinematicsTool::runInverseKinematicsWithOrientationsFromFile(
...
-            "Wrote IK with IMU tracking results to: '{}'.", fullOutputFilename);
-    if (get_report_errors()) {
-        STOFileAdapter_<double>::write(*modelOrientationErrors,
-                get_results_directory() + "/" +
-                        getName() + "_orientationErrors.sto");
-    }
+        log_info("Wrote IK with IMU tracking results to: '{}'.",
+                fullOutputFilename);
+        if (get_report_errors()) {
+            STOFileAdapter_<double>::write(*modelOrientationErrors,
```
antoinefalisse, line 247: Would it make sense to have `<name of the output motion file>_orientationErrors.sto`. As is now, it is the name of the IMUInverseKinematicsTool tool (eg `imuIK.setName('this_will_be_the_name')`). The error files will be overwritten if users batch process data and don't set names.

OpenSim/Tools/IMUInverseKinematicsTool.cpp
carmichaelong: > The use case I ran into was running from the GUI with neither output files nor results_directory specified, so results are displayed in GUI.

Ah, I see, thanks for the clarification. I'd say then at least adding a message to the `log_info` line to indicate how to get output files written would be great to help out users. Otherwise this PR looks good to me

antoinefalisse: I did some tests in MATLAB. If I don't specify `set_results_directory` but do `imuIK.set_output_motion_file('IKResults/testIK')` then the results are saved in the current folder, ie not in a new folder IKResults. Is that what we want?

The rest of the tests went fine.

antoinefalisse (changes requested):

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2898)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @antoinefalisse and @carmichaelong)

___
*[OpenSim/Tools/IMUInverseKinematicsTool.cpp, line 247 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2898#-MLysGGPnr907046lp1G-r2-247:-MMF3ShH1bdo2Pjc62XG:b1g6znp) ([raw file](https://github.com/opensim-org/opensim-core/blob/c8ee5d407545b0e67fa6585f06651f005c45ca94/OpenSim/Tools/IMUInverseKinematicsTool.cpp#L247)):*
<details><summary><i>Previously, antoinefalisse (Antoine Falisse) wrote…</i></summary><blockquote>

Would it make sense to have `<name of the output motion file>_orientationErrors.sto`. As is now, it is the name of the IMUInverseKinematicsTool tool (eg `imuIK.setName('this_will_be_the_name')`). The error files will be overwritten if users batch process data and don't set names. 
</blockquote></details>

Good idea, done. Thanks @antoinefalisse 
___
*[OpenSim/Tools/IMUInverseKinematicsTool.cpp, line 252 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2898#-MLysGGPnr907046lp1G-r2-252:-MMF3b2VDFKykJwFbdV8:beglx6z) ([raw file](https://github.com/opensim-org/opensim-core/blob/c8ee5d407545b0e67fa6585f06651f005c45ca94/OpenSim/Tools/IMUInverseKinematicsTool.cpp#L252)):*
<details><summary><i>Previously, antoinefalisse (Antoine Falisse) wrote…</i></summary><blockquote>

I did some tests in MATLAB. If I don't specify `set_results_directory` but do `imuIK.set_output_motion_file('IKResults/testIK')` then the results are saved in the current folder, ie not in a new folder IKResults. Is that what we want?

The rest of the tests went fine.
</blockquote></details>

Thanks @carmichaelong I updated the message per your suggestion.

@antoinefalisse Not sure what build you connected to MATLAB to perform the testing since this update is available only on this repo/branch and not yet on master or in any GUI build. Unless you're sure about the library/code-base you used to test, it would be great if you reproduce the behavior from the command line of this build or we can wait until an integrated build is available with the fixes from this branch, so please let me know. Thanks for testing.

antoinefalisse:
> @antoinefalisse Not sure what build you connected to MATLAB to perform the testing since this update is available only on this repo/branch and not yet on master or in any GUI build. Unless you're sure about the library/code-base you used to test, it would be great if you reproduce the behavior from the command line of this build or we can wait until an integrated build is available with the fixes from this branch, so please let me know. Thanks for testing.

I built the Java bindings from this branch. Isn't that possible?

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2898)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @antoinefalisse and @carmichaelong)

___
*[OpenSim/Tools/IMUInverseKinematicsTool.cpp, line 252 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/2898#-MLysGGPnr907046lp1G-r2-252:-MMHOHMNE4u3jQ7HmrZM:brhxl1t) ([raw file](https://github.com/opensim-org/opensim-core/blob/c8ee5d407545b0e67fa6585f06651f005c45ca94/OpenSim/Tools/IMUInverseKinematicsTool.cpp#L252)):*

Thanks @carmichaelong I updated the message per your suggestion.

@antoinefalisse Not sure what build you connected to MATLAB to perform the testing since this update is available only on this repo/branch and not yet on master or in any GUI build. Unless you're sure about the library/code-base you used to test, it would be great if you reproduce the behavior from the command line of this build or we can wait until an integrated build is available with the fixes from this branch, so please let me know. Thanks for testing.
</blockquote></details>

@antoinefalisse It sure is possible in theory, just very error prone since you need to fix/set the PATH, make sure no other installation is wired to Matlab, and you may need to change paths inside matlab manually if you can't use the configure script which normally assumes GUI installation layout. Regardless, for me to reproduce issues it's extremely helpful if this can be done/tested in standalone C++ environment without all these steps.

antoinefalisse:
> @antoinefalisse It sure is possible in theory, just very error prone since you need to fix/set the PATH, make sure no other installation is wired to Matlab, and you may need to change paths inside matlab manually if you can't use the configure script which normally assumes GUI installation layout. Regardless, for me to reproduce issues it's extremely helpful if this can be done/tested in standalone C++ environment without all these steps.

Ok, I did all of that. Will do it in C++ next time though.

aymanhab:
@antoinefalisse I just tried what you did in testopense and it worked exactly as expected, created a folder for results and created the output and error files in it, on Windows. I'd suggest you retry or test again when an integrated environment is available. Please let me know if you'd prefer to hold on merging this until you retest. Thank you.

aymanhab:
Per agreement in dev meeting, we agreed to merge. Thanks @carmichaelong @antoinefalisse

Issue #2677, opened 2020-02-18
Muscle Wrapping With Multithreading


vinaym815:
I have modified the GitHub page example to include a wrapping cylinder. The muscle wrapping works correctly when the code is single-threaded but jumps around randomly when multithreading.

Example Code: https://github.com/vinaym815/OpenSimExample/blob/master/GitHubExample/src/main2.cpp

![Bug](https://user-images.githubusercontent.com/35987596/74709333-410f0f00-5262-11ea-9e19-38e334d089af.png)

aymanhab:
@carmichaelong and @chrisdembia any thoughts on whether this is supposed to work? Thank you

chrisdembia:
This should not work; the muscle wrapping code is not thread-safe. I don't know the exact details, but we ran into this issue back when adding parallelism to forces.

vinaym815:
The bug is with opensim-core/OpenSim/Common/Mtx.cpp 
The shared variables in the class Mtx are responsible for the trouble. An easy fix would be allocated these variables on the stack within different functions. The functions that would require modifications are: Transpose, Multiply, and Invert.
Also, the Rotate(const double aAxis[3],......) have not been implemented

aymanhab:
This should be fixed by the latest @adamkewley commit. The Mtx class has been removed and other incorrect references to model components from path wrapping code have been removed. Please verify and close as needed.

vinaym815:
Thank you.

PR #2904, opened 2020-11-25
Fix MocoPathConstraint error for MocoCasADiSolver // update Metis patch

nickbianco:
Fixes issue where MocoCasADiSolver incorrectly accessed the path constraint errors vector when using multiple MocoPathConstraints (no issue, discovered recently).

### Brief summary of changes
- In MocoPathConstraint, calcPathConstraintErrors() now becomes calcPathConstraintErrorsView(), as it accepts the entire path constraint errors vector and creates a view into the portion of that vector belonging to *this* MocoPathConstraint.
- In MocoPathConstraint, calcPathConstraintErrors() now computes the constraint errors for *this* MocoPathConstraint (this is convenient for MocoCasADiSolver, since we create individual functions for each constraints that leads to improved problem sparsity).
- Updated MocoProblemRep::calcPathConstraintErrors() to use MocoPathConstraint::calcPathConstraintErrorsView().
- Updated the patch for Metis to use the COIN-OR tarball as our mirror was causing issues.

### Testing I've completed
- Added a new test that solves a double pendulum problem with multiple MocoPathConstraints.

### CHANGELOG.md (choose one)

- no need to update because...bug fix.
- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2904)

PR #2906, opened 2020-11-30
[WIP] Support muscle wrapping parallelism // include PathWrapSet in DGF's replaceMuscles

nickbianco:
TODO merge #2904 before merging this PR

Fixes issue #2677 

### Brief summary of changes

Muscle wrapping in OpenSim does not currently support parallelism due to the use of shared workspace memory in the low-level Mtx class, which implements static functions for common matrix operations. This PR proposes a solution to support muscle wrapping by removing the shared workspace used by some of the Mtx functions and instead allocating memory within each function as needed. The memory is still heap-allocated, since Mtx uses relies on C++ arrays of variable length, but the memory is allocated and deallocated within each function call, and therefore should be thread-safe. 

### Testing I've completed

TODO

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2906)


nickbianco:
@aymanhab @carmichaelong while this PR is WIP and needs #2904 to be merged first, I'd appreciate any feedback you have for the proposed fix.

aymanhab:
@nickbianco at the risk of making this larger than desired, is there a reason not to throw away all references to the C style allocation/deallocation and replace it with high level objects/classes SimTK::Mat33, Vec3,....? It's also not clear that muscle wrapping is the one and only issue blocking parallelization (were you able to parallelize computations for models with no muscle wrapping?)

adamkewley:
The functions you're changing are called *a lot* by the wrapping implementations iirc (eg they can easily be 15 % perf, alone) - unless I'm mixing this up with wrapmath (will check later). 

Either way, editing the matrix multiplication implementation to perform a fresh heap allocation every time is likely to have perf issues. It will kill L1 caching and thrash the memory allocator when called in a tight loop. The way to make something that relies on static mutable state thread-safe is to either use a mutex (if sharing would be preferable) or use thread_local allocation (best here, probably).

On another note, if you're wanting to use opensim in a multithreaded context, beware. It definitely wasn't engineered with that in mind and all it'll take is some component somewhere mis-copying an internal computation helper that's normally used to help a single-thread, or shared static state (here) for a problem to pop up.

nickbianco:
@aymanhab we've not had any issues with parallelization until introducing muscle wrapping. We use separate copies of the problem OpenSim model in each thread and that's worked pretty well for us.

I think a solution that utilizes the SimTK matrix types is probably the way to go. I wanted to try something simple first, but I suspected performance issues as @adamkewley says. We don't want to make wrapping slower, as it is already somewhat costly.

nickbianco:
Closing this PR since I started fresh on a new branch. Refer to #2910.

PR #2908, opened 2020-12-02
Minor perf patch: Reduce temporary string construction during a simulation

adamkewley:
This PR is a minor perf-related change. It does not add any new functionality, fix any old bugs, or change any user-facing APIs. It speeds up `ToyDropLanding` by ~5 % and `RajagopalModel` by ~2 % with only a few minor changes.

## Why

When I get bored, I like to surf valgrind profiles. This is largely considered to be abnormal behaviour :wink: . 

I looked--specifically--into memory allocations during a simulation. Mostly, because allocations are relatively easy to diagnose. Here's an example of what I see for `malloc` when running a forward-dynamic `ToyDropLanding` simulation:

malloc

and for `free`:

free

The `SimTK::ContiguousVectorScalarHelper` allocations are because SimTK has a nasty habit of heap-allocating virtual "views" into vectors. It's also because many of OpenSim/SimTK's library methods tend to take a `SimTK::Vector const&` argument, rather than (e.g.) a `span`-like object (or a pair of pointers - if that's your idea of fun). 

Because of that, there is quite a bit of code that (e.g.) heap-allocates a `Vector` in order to pass one floating point number to a `SimTK::Function` object. However, fixing that could be a lot of work so, out of sheer laziness, this patch only focuses on string allocations. 

## The Fix

Fixing string allocations in OpenSim is easy enough because they almost always as a result of this pattern:

```c++
void f(std::string const& s);

void g() {
    // std::string::string(const char*); called every time `g` is called
    // and might allocate (SSO-dependent)
    f("hello, world");
}
```

And can be fixed by pre-constructing the string statically:

```c++
void f(std::string const& s);

// allocation occurs once, in `_start`, before `main` is called
static const std::string s{"hello, world"};

void g() {
    f(s);
}
```

There is also this pattern, which allocates a temporary string:

```c++
void f(std::string const& s);

void g(std::string const& s) {
    f(s + "_deriv");  // allocates a temporary
}
```

And can be fixed by using a `thread_local` scratch space (not `static`, because it's a mutable scratch space that could feasibly be accessed by two threads concurrently):

```c++
void f(std::string const& s);

void g(std::string const& s) {
    thread_local std::string scratch;
    scratch.clear();
    scratch += s;
    scratch += "_deriv";
    f(scratch);
}
```

## Post-Patch Measurements

Running this through my usual suite, these changes result in an ~5 % speedup in `ToyDropLanding`, which is roughly within expectations (`ToyDropLanding` is what I measured initially in valgrind).

| passive_dynamic_noanalysis |       3.03 |     0.00 |       3.04 |     0.00 |    1.00 |
|             ToyDropLanding |      11.49 |     0.01 |      10.98 |     0.01 |    1.05 |
|                      Arm26 |       0.26 |     0.00 |       0.25 |     0.00 |    1.02 |
|             RajagopalModel |      13.28 |     0.01 |      12.98 |     0.02 |    1.02 |
|            passive_dynamic |       4.66 |     0.00 |       4.71 |     0.00 |    0.99 |

Post-change, `malloc` now looks like:

new_after

And `free`:

free_after

So, now, ~95 % (roughly 20/22 M) of memory allocations while running `ToyDropLanding` in OpenSim are related to the SimTK vector views, helpers, etc. Fixing those might give a nice speed-bump also--if only because it would help with L1 caching--but that's for a separate PR and, again, would be quite a lot more work.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2908)


nickbianco (approved):
LGTM. The recent commits to master updating Simbody and removing Travis/Appveyor will fix the CI failures.

Not suggesting that you do this here, but would it make sense to have tests that check if changes are indeed thread-safe, considering these changes and my recent muscle wrapping PR? Not sure if that's ever done in practice.

> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2908#-:-MO7lj0A4CNU1juI61i8:b-atwcdc)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adamkewley)

adamkewley:
> Not suggesting that you do this here, but would it make sense to have tests that check if changes are indeed thread-safe, considering these changes and my recent muscle wrapping PR? Not sure if that's ever done in practice.

@nickbianco for that, no: I don't think it makes sense to use standard tests to test for thread safety. 

Typical industry practice for engineering thread safety is to:

- Design the system with threading in mind from day one
- Continuously perform fuzzing and sanity checks with 3rd-party tools like helgrind, valgrind, etc. as the software is being developed
- Ideally, use languages and patterns that do not require low-level threading decisions at all (e.g. Rust, immutable datastructures)
- Even more ideally, avoid MT altogether by splitting things into concurrently-running single-threaded processes (e.g. unix style, agent-based systems)

The main reason I'm using a `thread_local` here is as a default (and probably unncessary) precaution. It might even be worth preemptively switching the scratch space for a member variable, which would enable multiple threads to independently mutate *separate components* (there is a nonzero overhead assocaited with TLS variables - try using one in godbolt to see how much assembly it can generate).

OpenSim's design, which uses object-oriented programming (so datastructures can contain *a lot* of members that interact in non-obvious ways), `static` data (races), and extreme overuse of `mutable` (so `const` doesn't have as many guarantees as in typical codebases) means that it's quite ripe for subtle MT bugs.

adamkewley:
This is pending on a merge **from** `master` to ensure it passes CI before merging it **into** master

adamkewley:
This seems to be passing against master, so I'll go ahead and merge it

PR #2909, opened 2020-12-02
Improved Component::getStateVariableValue perf in some cases

adamkewley:
This PR is a minor perf improvement that should not change any existing functionality or break any user-facing APIs. 

This is semi-related to some of the previous work I did with `ComponentPath` (PR #2844), but higher-level. The `ComponentPath` change focused on fixing the low-level implementation details of `ComponentPath` without changing any external APIs so that it could be "dropped in" to give a perf benefit. That turned out to be a fair amount of work (what doesn't turn out that way :wink:) but did yield measurable improvements.

This work goes one step higher, and tries to modify how `ComponentPath` is used by other parts of OpenSim. When I was working on #2844 @chrisdembia mentioned that the bigger issue is that OpenSim is mis-using the `ComponentPath` interface by passing arguments around as `std::string`s and only temporarily using `ComponentPath` to (e.g.) split a path. At the time, I agreed, but thought it detracted from that PR's work because it would be a wider-ranging changeset. However, it's a good suggestion and that's what this PR focuses on.

## Why

If I valgrind a `ToyDropLanding` simulation, I see *a lot* of `ComponentPath`s being constructed in order to get state variable values:

getstatevariablevalue

This was also known in #2844, but I focused on removing allocations *within* `ComponentPath`.

The solution to this is to make callers provide their own `ComponentPath` when they can. This is achieved by adding this overload to `OpenSim::Component`:

```c++
    double getStateVariableValue(const SimTK::State& state, const ComponentPath& path) const;
```

The reason why this is faster is because this overload doesn't need to construct any temporary `ComponentPath`s internally (one was provided).

In addition to adding the overload, this PR also:

- Makes `ComponentPath::toString()` return a `const std::string&` instead of a `std::string` (the value can be trivially constructed from the reference, so this transformation shouldn't break any C++ code, at least). 
  - This extra addition is so that the `toString` doesn't allocate a new temporary string for the `std::map` lookup.
- Includes a basic copy-and-paste of the test suite to exercise the overload
  - A basic sanity-check. In reality, the "old" overload now uses the new one internally
- Rolls out a suggested use-case in `OpenSim::Thelen2003Muscle`
  - This is a usage point that specifically affects `ToyDropLanding` and enabled me to do a direct perf measurement of the change.
  - My ambition is to roll this out in a series of smaller, easier to review, PRs, rather than a super-large PR that does it all-in-one

## Example Perf Changes

This changeset mostly only affects `ToyDropLanding` because the rollout only touches the callsites I measured from `ToyDropLanding` (e.g. in `Thelen2003Muscle`)

```
# e.g. I measure it with a script like this: check my GitHub if you want the sourcecode
./osimperf compare-all --repeats 32 ~/Desktop/master-RelWithDebInfo-install/bin/opensim-cmd ~/Desktop/osc/feature-perf-getstatevariablevalue-RelWithDebInfo-install/bin/opensim-cmd
```

| passive_dynamic_noanalysis |       3.04 |     0.00 |       3.08 |     0.00 |    0.99 |
|   ToyDropLanding_nomuscles |       0.44 |     0.00 |       0.43 |     0.00 |    1.00 |
|                   Gait2354 |       0.26 |     0.00 |       0.26 |     0.00 |    0.99 |
|             ToyDropLanding |      11.58 |     0.01 |      10.95 |     0.00 |    1.06 |
|                      Arm26 |       0.26 |     0.00 |       0.25 |     0.00 |    1.02 |
|             RajagopalModel |      13.36 |     0.01 |      13.24 |     0.01 |    1.01 |
|            passive_dynamic |       4.67 |     0.00 |       4.73 |     0.00 |    0.99 |

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2909)


adamkewley:
Anyone available to review this? It's a fairly simple changeset (most of it is the testsuite)?

aymanhab (approved):
LGTM, Thanks @adamkewley I assume this has no side effect on bindings (Python/Java) since the changes in the header are guarded but can you please confirm?

aseth1:
LGTM. Cannot see how these changes could affect bindings.

PR #2910, opened 2020-12-03
Replace usages of Mtx to support wrapping parallelism // include PathWrapSet in DGF's replaceMuscles

nickbianco:
TODO merge #2904 before merging this PR

Fixes issue #2677

### Brief summary of changes

Muscle wrapping in OpenSim does not currently support parallelism due to the use of shared workspace memory in the low-level Mtx class, which implements static functions for common matrix operations. 

### CHANGELOG.md (choose one)

- updated.


nickbianco:
@aymanhab, @carmichaelong I added both of you as reviewers since this is modifying low-level matrix stuff.

@adamkewley feel free to add comments too (especially regarding performance comparisons).

adamkewley:
This is very nice work @nickbianco . The

nickbianco:
Thanks for the thorough performance analysis Adam! I should have said that the performance checks I made were mostly for sanity checks since we in general trust Simbody's performance. Not sure if it needs to be a critical aspect of this PR, but perhaps adding the other comparisons to the picobench checks would be sufficient here.

> For example, your concurrency problem can specifically be addressed by changing the static scratch space variables to thread_local ones. It'd be a 4-/5-line diff that fixes the concurrency problem, could be reviewed in minutes, and could be merged into master with no obvious pitfalls (obvious being the key word there 😉).

Agreed, this would be a nice quick fix. However, after working on these changes, I'm inclined to replacing the `Mtx` stuff, since that class is not well documented and confusing to use (hence the need for sandbox testing to make sure I was doing the equivalent matrix operations). In the dev meeting yesterday we discussed replacing _all_ usage of `Mtx` with Simbody, but I think that will come at a later time. Such a PR would definitely be merged to `master` and include better test coverage for wrapping.

Since the impetus for these changes was supporting muscle wrapping for Moco, it was convenient to make these changes off of the `feature_moco` branch to check that I could run a MocoProblem with parallelization. However, it would be easy to make this change off of master (@aymanhab, @carmichaelong this is up to you).

adamkewley:
Agreed that the matrix stuff should be ripped out and replaced with equivalent simbody code. 

It was something I wanted to do when starting on the wrapping perf investigations. ~40 % of CPU cycles for (e.g.) a FD Rajagopal sim are in `WrapCylinder::_make_spiral_path`, which uses some of these quite slow math functions. That work was lower prio because we started investigating alternative wrapping implementaions (e.g. Simbody cables) and other techniques (e.g. precomputing wraps and fitting splines to them). We weren't sure if we ultimately be gutting the code anyway (most recent work indicates we won't be, so your work here will likely be valuable long-term).

I think it's good that you've taken time to look into replacing this, and that you used some basic testing etc. to prove it. The comment about splitting things up into separate, small, PRs is more of a quality-of-life thing for you+reviewers. The `static` -> `thread_local` would've been a quick win you could use to move forward. However, it's better that you've taken the time to fix the core of the problem - keep at it!

aymanhab:
@nickbianco I'm reviewing but please keep us updated on the test failure if related.

aymanhab (changes requested):
> Reviewed 10 of 10 files at r1, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2910)* status: all files reviewed, 6 unresolved discussions (waiting on @carmichaelong and @nickbianco)

___
*[OpenSim/CMakeLists.txt, line 13 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-MNy-Dkl9ysvSRmFnwpN:-MNy-Dkl9ysvSRmFnwpO:b-skavb6) ([raw file](https://github.com/opensim-org/opensim-core/blob/a0f3c671299c59d231e30bd1b6f7b0164896db6f/OpenSim/CMakeLists.txt#L13)):*
> ```CMake
> add_subdirectory(Tests)
> 
> add_subdirectory(Sandbox)
> ```

Any reason to not include with other Tests
___
*[OpenSim/Actuators/DeGrooteFregly2016Muscle.cpp, line 1050 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-MNy-fgd7rR5EspdxaKR:-MNy-fgd7rR5EspdxaKS:b-r2up6y) ([raw file](https://github.com/opensim-org/opensim-core/blob/a0f3c671299c59d231e30bd1b6f7b0164896db6f/OpenSim/Actuators/DeGrooteFregly2016Muscle.cpp#L1050)):*
> ```cpp
>             geomPath.updWrapSet().adoptAndAppend(pathWrap);
>         }
> ```

The change looks good but the functionality doesn't necessarily belongs to a replaceMuscles method in the DeGrooteFregly2016Muscle. Would be ok for this PR but we need to plan a more logical place for this code. Thoughts?
___
*[OpenSim/Sandbox/WrappingMath/CMakeLists.txt, line 19 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-MNy0IeN4KbJ2Xf8AzLo:-MNy0IeO51gKbyqQXyv-:blm338h) ([raw file](https://github.com/opensim-org/opensim-core/blob/a0f3c671299c59d231e30bd1b6f7b0164896db6f/OpenSim/Sandbox/WrappingMath/CMakeLists.txt#L19)):*
> ```CMake
> 
> WrappingMathAddSandboxExecutable(NAME sandboxReplaceWrappingMath LIB_DEPENDS osimTools)
> ```

Again, if we stick with the Tests folder there would be no need for more CMake gymnastics altogether
___
*[OpenSim/Simulation/Model/GeometryPath.h, line 109 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-MNyBLba1rTvfgA97GR3:-MNyBLba1rTvfgA97GR4:b-kiu08r) ([raw file](https://github.com/opensim-org/opensim-core/blob/a0f3c671299c59d231e30bd1b6f7b0164896db6f/OpenSim/Simulation/Model/GeometryPath.h#L109)):*
> ```objc
>     const PathWrapSet& getWrapSet() const { return get_PathWrapSet(); }
>     PathWrapSet& updWrapSet() { return upd_PathWrapSet(); }
>     void addPathWrap(WrapObject& aWrapObject);
> ```

Do you know the rationale for keeping the PathWrapSet property private and then adding a public accessor? 
___
*[OpenSim/Simulation/Wrap/WrapSphere.cpp, line 196 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-MNyFQjRDtwM0sDK_DQO:-MNyFQjRDtwM0sDK_DQP:b-hzb0gu) ([raw file](https://github.com/opensim-org/opensim-core/blob/a0f3c671299c59d231e30bd1b6f7b0164896db6f/OpenSim/Simulation/Wrap/WrapSphere.cpp#L196)):*
> ```cpp
>    //maxit = 50;
>    aFlag = true;
> ```

Couldn't resist the informative name of aFlag :smiley: 
___
*[OpenSim/Simulation/Wrap/WrapSphere.cpp, line 518 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-MNyGPFs3cs4fUxlGaDc:-MNyGPFs3cs4fUxlGaDd:b-wvvoym) ([raw file](https://github.com/opensim-org/opensim-core/blob/a0f3c671299c59d231e30bd1b6f7b0164896db6f/OpenSim/Simulation/Wrap/WrapSphere.cpp#L518)):*
> ```cpp
>         double wangle = angle * (i+1) / (numWrapSegments - 1) * SimTK_DEGREE_TO_RADIAN;
> 
>         R.setRotationFromAngleAboutNonUnitVector(wangle, Vec3::getAs(axis));
> ```

I thought axis was normalized earlier, or is this a type issue where the function expects a Unit vector?

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2910)* status: all files reviewed, 6 unresolved discussions (waiting on @aymanhab and @carmichaelong)

___
*[OpenSim/CMakeLists.txt, line 13 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-MNy-Dkl9ysvSRmFnwpN:-MNyZQqx4krcCzz_VZLq:b-nn2pt2) ([raw file](https://github.com/opensim-org/opensim-core/blob/a0f3c671299c59d231e30bd1b6f7b0164896db6f/OpenSim/CMakeLists.txt#L13)):*

Any reason to not include with other Tests
</blockquote></details>

Not really, the sandbox file could be converted to a test. Not sure if that's what we wanted so I left it open to discussion.
___
*[OpenSim/Actuators/DeGrooteFregly2016Muscle.cpp, line 1050 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-MNy-fgd7rR5EspdxaKR:-MNyZt4i5bjlvAbXHKYu:brcmkev) ([raw file](https://github.com/opensim-org/opensim-core/blob/a0f3c671299c59d231e30bd1b6f7b0164896db6f/OpenSim/Actuators/DeGrooteFregly2016Muscle.cpp#L1050)):*

The change looks good but the functionality doesn't necessarily belongs to a replaceMuscles method in the DeGrooteFregly2016Muscle. Would be ok for this PR but we need to plan a more logical place for this code. Thoughts?
</blockquote></details>

We replace the path points in the same way here, but yes, this could (and probably should) be moved to one of the OpenSim utility files.
___
*[OpenSim/Sandbox/WrappingMath/CMakeLists.txt, line 19 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-MNy0IeN4KbJ2Xf8AzLo:-MNy_82u8d0ZLP74JIzC:b-otilcs) ([raw file](https://github.com/opensim-org/opensim-core/blob/a0f3c671299c59d231e30bd1b6f7b0164896db6f/OpenSim/Sandbox/WrappingMath/CMakeLists.txt#L19)):*

Again, if we stick with the Tests folder there would be no need for more CMake gymnastics altogether
</blockquote></details>

See comment above.
___
*[OpenSim/Simulation/Model/GeometryPath.h, line 109 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-MNyBLba1rTvfgA97GR3:-MNyX_7EBi1X3-4KUkIq:b-ypj9f9) ([raw file](https://github.com/opensim-org/opensim-core/blob/a0f3c671299c59d231e30bd1b6f7b0164896db6f/OpenSim/Simulation/Model/GeometryPath.h#L109)):*

Do you know the rationale for keeping the PathWrapSet property private and then adding a public accessor? 
</blockquote></details>

I was just following the existing convention for the class. 
___
*[OpenSim/Simulation/Wrap/WrapSphere.cpp, line 196 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-MNyFQjRDtwM0sDK_DQO:-MNyXy2o-XUvD1nmtEi6:bs13yyc) ([raw file](https://github.com/opensim-org/opensim-core/blob/a0f3c671299c59d231e30bd1b6f7b0164896db6f/OpenSim/Simulation/Wrap/WrapSphere.cpp#L196)):*

Couldn't resist the informative name of aFlag :smiley: 
</blockquote></details>

This PR has made me realize how undocumented (and often confusing) the wrapping code is. It would be good to continue improving it going forward.
___
*[OpenSim/Simulation/Wrap/WrapSphere.cpp, line 518 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-MNyGPFs3cs4fUxlGaDc:-MNyWy7y1l7t-IP-4urD:b-ahk0tg) ([raw file](https://github.com/opensim-org/opensim-core/blob/a0f3c671299c59d231e30bd1b6f7b0164896db6f/OpenSim/Simulation/Wrap/WrapSphere.cpp#L518)):*

I thought axis was normalized earlier, or is this a type issue where the function expects a Unit vector?
</blockquote></details>

This method converts the non-unit vector to a unit vector internally. I wanted to use this since it's a little safer.

aymanhab (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2910)* status: all files reviewed, 4 unresolved discussions (waiting on @aymanhab, @carmichaelong, and @nickbianco)

___
*[OpenSim/CMakeLists.txt, line 13 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-MNy-Dkl9ysvSRmFnwpN:-MNybCKd67otecRBXe1b:b52ufyt) ([raw file](https://github.com/opensim-org/opensim-core/blob/a0f3c671299c59d231e30bd1b6f7b0164896db6f/OpenSim/CMakeLists.txt#L13)):*

Not really, the sandbox file could be converted to a test. Not sure if that's what we wanted so I left it open to discussion.
</blockquote></details>

I'd recommend we move this to the regular tests, and eventually build upon it to test wrapping more broadly.

cmake/OpenSimMacros.cmake
```diff
@@ -366,6 +366,8 @@ function(OpenSimAddTests)
             add_executable(${TEST_NAME} ${test_program}
                 ${OSIMADDTESTS_SOURCES})
             target_link_libraries(${TEST_NAME} ${OSIMADDTESTS_LINKLIBS})
+            target_include_directories(${TEST_NAME}
```
adamkewley, line 369: Careful about transitively using some other dependency's dependency as a direct dependency in OpenSim. A special kind of hell can open up wherein the dependency gets updated and a bunch of entirely unrelated functionality breaks in downstream code.

OpenSim/Simulation/Test/testWrappingMath.cpp
```diff
@@ -0,0 +1,139 @@
...
+        vec1[i] = v;
+        vec2[i] = v;
+    }
+}
+
+template<typename C, typename D, int N = 3>
+static void compareMat(C mat1, D mat2) {
+    for (int i = 0; i < N; ++i) {
+        for (int j = 0; j < N; ++j) {
+            CHECK(mat1[i][j] == Approx(mat2[i][j]).margin(1e-6));
```
adamkewley, line 61: Why was the margin for this comparison dropped three orders of magnitude to `10^-6`? `double`-precision numbers can be accurate to (e.g.) `10^-16`, but can incur error when multiplying/adding (esp. if the magnitude of those numbers differs by a lot). 

Is there something fundamentally different about this, or the previous, matrix multiplication that causes error to increase by *a lot*? I can understand having somewhat liberal error margins on (e.g.) a converging algorithm, moco, etc. but it's inherently suspicious that we can only multiply `double`-based matrices with 1ppm accuracy (either before or after).

Is this because one approach uses `float`-based trigonometric functions for the rotation?

cmake/OpenSimMacros.cmake
nickbianco: I'm not sure this is needed actually, if I include the `catch.hpp` using the direct path in testWrappingMatch.cpp. Removing it for now.

OpenSim/Simulation/Test/testWrappingMath.cpp
nickbianco: I'm reverting back to the `SimTK_TEST` macros here. These macros use the `SimTK::Test::numericallyEqual()` comparison functions underneath which scales the equality test by the magnitude of the values being compared as to not ask for unreasonable precision. 

We use should probably use these SimTK testing macros over the Catch stuff for any meaningful numerical comparisons (I got a little carried away creating these tests using Catch).

nickbianco:
@carmichaelong I removed you as a required reviewer since @adamkewley is already reviewing.

nickbianco (commented):
@aymanhab I've added tests for the muscle wrapping math updates under the Simulation folder and removed the Sandbox files since they are now redundant.

I did just realize that `testWrapping.cpp` exists. Should I move the tests here for consistency? 

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-:-MO7nQt82pCv7G8jvyUD:b-tyb993)* status: 0 of 12 files reviewed, 6 unresolved discussions (waiting on @adamkewley and @aymanhab)

___
*[OpenSim/CMakeLists.txt, line 13 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-MNy-Dkl9ysvSRmFnwpN:-MO7q43P1CmQpmwzlXqp:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/a0f3c671299c59d231e30bd1b6f7b0164896db6f/OpenSim/CMakeLists.txt#L13)):*

I'd recommend we move this to the regular tests, and eventually build upon it to test wrapping more broadly.
</blockquote></details>

Done.

aymanhab:
@nickbianco Thank you, no need to combine the two, one of them tests the low level math and would eventually be removed while the other will continue to test the functionality and be beefed up.

aymanhab (changes requested):
> Reviewed 12 of 17 files at r7, 1 of 1 files at r11.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2910)* status: all files reviewed, 3 unresolved discussions (waiting on @adamkewley and @nickbianco)

___
*[cmake/OpenSimMacros.cmake, line 369 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-MO3AUy9Tk9fqS6fCVnS-r7-369:-MOClGoH2VSPlfcXbFfT:b-6vkm7w) ([raw file](https://github.com/opensim-org/opensim-core/blob/8b956608133faf5f8a96805870deea856530d335/cmake/OpenSimMacros.cmake#L369)):*

I'm not sure this is needed actually, if I include the `catch.hpp` using the direct path in testWrappingMatch.cpp. Removing it for now.
</blockquote></details>

@nickbianco Did you end up reverting the change? Hard to tell on reviewable... 
___
*[OpenSim/Simulation/Test/testWrappingMath.cpp, line 26 at r11](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-MOCkPn-Dxts9ZlYrOLC:-MOCkPn-Dxts9ZlYrOLD:be1zo40) ([raw file](https://github.com/opensim-org/opensim-core/blob/c3a49e4fae75d92aae3f10405d72c09e590dab08/OpenSim/Simulation/Test/testWrappingMath.cpp#L26)):*
> ```cpp
> #define CATCH_CONFIG_MAIN
> #include "Vendors/tropter/external/catch/catch.hpp"
> #include <OpenSim/Simulation/Wrap/WrapMath.h>
> ```

We have a version of catch under OpenSim/Auxiliary, we should use that one instead of tropter (if the same) since we already have it and users may build without Tropter

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2910)* status: 11 of 12 files reviewed, 3 unresolved discussions (waiting on @adamkewley and @aymanhab)

___
*[cmake/OpenSimMacros.cmake, line 369 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-MO3AUy9Tk9fqS6fCVnS-r7-369:-MOD6lqn2yK0azut3e_9:bulbe3g) ([raw file](https://github.com/opensim-org/opensim-core/blob/8b956608133faf5f8a96805870deea856530d335/cmake/OpenSimMacros.cmake#L369)):*

@nickbianco Did you end up reverting the change? Hard to tell on reviewable... 
</blockquote></details>

Yes, I did.
___
*[OpenSim/Simulation/Test/testWrappingMath.cpp, line 26 at r11](https://reviewable.io/reviews/opensim-org/opensim-core/2910#-MOCkPn-Dxts9ZlYrOLC:-MOD6kRd7rxyzvxqJiSX:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/c3a49e4fae75d92aae3f10405d72c09e590dab08/OpenSim/Simulation/Test/testWrappingMath.cpp#L26)):*

We have a version of catch under OpenSim/Auxiliary, we should use that one instead of tropter (if the same) since we already have it and users may build without Tropter
</blockquote></details>

Done.

aymanhab:
LGTM pending Adam's feedback

PR #2915, opened 2020-12-10
Add xsens exported file on latest format and support parsing. 

aymanhab:
Assume fixed/hard-coded DataRate due to missing documentation.

Fixes issue #0

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2915)


aymanhab:
@antoinefalisse This PR should allow you to parse the latest Xsens file format, it makes assumption about fixed dataRate and keys off firmware version. Can you comment on whether this works for you or if the firmware version is reliable enough to use based on the system you have used? Thank you

antoinefalisse:
Hi @aymanhab the fix works with my data, generated with MT manager 2020.0. I wonder if that would make sense to have the `dataRate` as an optional input so that users can pass it as an argument when converting the data. It is very likely that they will use something else than 40 Hz in practice and this might cause issues if they want to use data from different sources since the time vectors will then be different. What do you think?

aymanhab:
@antoinefalisse I like the option to specify DataRate if not specified but considering that this could conflict with data from the files (depending on Xsens version) it is not clear where that belongs or when it is should be considered. I'd lean to leave this as is and address the resampling/time-line issues as an important but separate issue (e.g. include in TableUtilities methods to shift-time, stretch-time, .. etc.) Until we're dealing with sensor fusion, actual time doesn't seem critical. Thoughts?

antoinefalisse:
That sounds good to me.

aymanhab:
Thanks @antoinefalisse

PR #2916, opened 2020-12-15
WIP: Add overload to OpenSim::Function for unary callers

adamkewley:
Not a fundamental change, nor a necessary one, nor a large performance increase (yet ;)). This is just an idea I'm bouncing around.

# Overview 

`OpenSim::Function` provides an API for a function that accepts n `double`s and returns a single `double`:

```c++
    /**
     * Calculate the value of this function at a particular point.
     *
     * @param x the Vector of input arguments.
     *          its size must equal the value returned by getArgumentSize().
     */
    virtual double calcValue(const SimTK::Vector& x) const;
```

Which is conceptually fine, but has a few annoying design issues:

- It accepts a `SimTK::Vector`, rather than (e.g.) a `span`-like object or pointer range. 
  - The caller's *actual* requirements (imho) are that it is provided `n` `doubles` that are contiguous in memory. 
  - A `SimTK::Vector` introduces the concept of dynamic memory allocation + resizable/appendable sequences to an API that ultimately just wants a non-mutating contiguous view of `n` `double`s
  - This means that the arguments *must* be heap-allocated in a `SimTK::Vector`. Arguments can't be provided (e.g.) via the stack.
- It's the caller's responsibility to allocate + populate this memory
  - A caller *must* provide the vector by writing (e.g.) `function.calcValue(SimTK::Vector(1, val))` (or similar)
  - So the caller, rather than the API, is responsible for constructing the argument's representation in memory. This ultimately means that almost all callers in OpenSim end up containing (e.g.) `SimTK::Vector(1, val)` at the call site
  - This design might be useful if a caller is calling a function with the same argument pack over and over (because a caller could then construct the arguments once and recycle them). It's also our legacy behavior, so the existing API should remain
  - However, *most* callers are calling the function in a one-off manner, so the API should *also* provide a way of having standard calling conventions - even if the initial implementation of that API does exactly the same thing that the callers were doing before it (`SimTK::Vector(n, args...)`). It would still be beneficial because it moves a common pattern into one place (`OpenSim::Function`) rather than it spreading to all callsites

So an ideal API would look like this:

```c++
f.calcValue(1.0);
g.calcValue(1.0, 2.0, 3.0);
```

Which is possible, but might need to also provide (e.g.) `calcValueUnary` or `calcValueT` (or similar) because of how overload name hiding works ("legacy" inheritors of `OpenSim::Function` will virtual-implement the "legacy" overload that takes a `SimTK::Vector const&` but don't currently forward the currently-non-existent value-based overload).

# Early implementation steps

Inital commits in this WIP focus on unary functions by adding a unary overload with this signature:

```c++
double calcValue(double a1) const;
```

Which has a (currently, default) implementation that packs the arguments into a `thread_local` `SimTK::Vector` scratch space followed by forwarding the vector to the existing implementation. The existing API is left untouched. Therefore, this API addition is non-breaking, and passes all tests.

Further commits roll this (unary-only) API out to OpenSim to demonstrate how the new API would look from the caller's side. The rollout also passes all tests. Basic perf measurements of the post-rollout version indicate nothing noteworthy - maybe a 2 % perf uplift in something like `ToyDropLanding`, if you squint. The main benefit is that the caller-side code is clearer (see changeset).

# Where I want this to end up

This is currently WIP, but I'm hoping that the final iteration will provide a templatized version of the API that can accept an arbitrary number of arguments, something like:

```c++
template<typename... Args>
double calcValue(Args... args) const {
    thread_local SimTK::Vector scratch(sizeof...(Args));
    scratch = {args...};
    return calcValue(scratch);
}
```

(+ the same for associated calls, such as derivatives)

Having that API would mean that almost all callsites could be cleaned up to just directly provide the arguments. Once that's rolled out, there would be scope for providing performance-oriented overloads to the existing API (e.g. functions accept a pointer range, rather than a `SimTK::Vector`, so the scratch space can be replaced by a stack-allocated array of arguments that are forwarded into the implementation).

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2916)

### DIFF
--- a/Applications/Scale/test/testScale.cpp
+++ b/Applications/Scale/test/testScale.cpp
@@ -697,9 +697,9 @@ void scaleJointsAndConstraints()
         transform[3].setFunction(new LinearFunction(1.234, 5.678));
 
         // Evaluate LinearFunction before scaling and compare after scaling.
-        const Vector xA = Vector(1, 0.);
+        const double xA = 0.0;
         const double yA = transform[3].getFunction().calcValue(xA);
-        const Vector xB = Vector(1, 123.456);
+        const double xB = 123.456;
         const double yB = transform[3].getFunction().calcValue(xB);
 
         CustomJoint* custom1 =
--- a/OpenSim/Actuators/Delp1990Muscle_Deprecated.cpp
+++ b/OpenSim/Actuators/Delp1990Muscle_Deprecated.cpp
@@ -386,8 +386,8 @@ double Delp1990Muscle_Deprecated::computeActuation(const SimTK::State& s) const
     double massTerm = (tendonForce * ca - fiberForce * ca * ca) / muscleMass;
     double velocityTerm = normState[STATE_FIBER_VELOCITY] * normState[STATE_FIBER_VELOCITY] * ta * ta / normState[STATE_FIBER_LENGTH];
     normStateDeriv[STATE_FIBER_VELOCITY] = massTerm + velocityTerm;
-    setPassiveForce(s, getPassiveForceLengthCurve()->calcValue(SimTK::Vector(1, normState[STATE_FIBER_LENGTH])));
-    setActiveForce(s, getActiveForceLengthCurve()->calcValue(SimTK::Vector(1, normState[STATE_FIBER_LENGTH])) * getActivation(s));
+    setPassiveForce(s, getPassiveForceLengthCurve()->calcValue(normState[STATE_FIBER_LENGTH]));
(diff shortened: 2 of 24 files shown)


aymanhab:
The main concern with this (rather cool) change is that the call sites in C++ will look different from Matlab/Python which means that our examples and tests can't be used cross-language, a major handicap when most of our advanced users are not C++ experts but are Matlab scripters.

adamkewley:
Closing this (it's extremely stale)

PR #2920, opened 2020-12-16
Add gfortan to Mac CI build

nickbianco:
### Brief summary of changes

Add gfortran to CI for Mac build.

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2920)

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -157,7 +157,7 @@ jobs:
     - name: Install Homebrew packages
       # Save the gfortran version to a file so we can use it in the cache key.
       run: |
-        brew install cmake pkgconfig gcc autoconf libtool automake wget pcre doxygen
+        brew install cmake pkgconfig gcc autoconf libtool automake wget pcre doxygen gfortran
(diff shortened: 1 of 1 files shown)


aymanhab (approved):
LGTM, thanks @nickbianco

adamkewley:
Out of curiosity @nickbianco is gfortran a new runtime dependency? The other deps seem like build ones. 

Main thing to consider is that brew isn't available on typical Mac targets, so downstream will need to package gfortran for opensim to work on a clean Mac (unsure how much post-package testing is done on clean apple machines)

nickbianco:
We need gfortran to build Ipopt on Mac. I now realize that gfortran should come included with the gcc homebrew package, so not sure why this issue occured.

chrisdembia:
CMake copies the necessary gfortran libraries into Moco’s installation. The cmake scripts need to be updated whenever the gfortran version updates bc we hardcode a version.

Yes it’s important to make sure that users don’t need to have gfortran installed.

-Chris

PR #2923, opened 2020-12-16
gfortran CI fix (second try)

nickbianco:
### Brief summary of changes

Removing the gfortran install that was added in the recent PR #2920, because gfortran comes with gcc (which we already install) and we don't want a potential conflict. This change forces gcc to be reinstalled (via `brew reinstall gcc`) on each CI build, so that gfortran is found. Not sure why this is necessary, might be a bug with the homebrew gcc formula. Hopefully we can use the cached gcc install going forward. Doesn't seem to slow the CI down much at all though.

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2923)

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -158,7 +158,6 @@ jobs:
       # Save the gfortran version to a file so we can use it in the cache key.
       run: |
         brew install cmake pkgconfig gcc autoconf libtool automake wget pcre doxygen
-      # brew reinstall gcc
         pip3 install numpy
         gfortran -v
         mkdir gfortran_version
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -158,6 +158,7 @@ jobs:
       # Save the gfortran version to a file so we can use it in the cache key.
       run: |
         brew install cmake pkgconfig gcc autoconf libtool automake wget pcre doxygen
+        brew reinstall gcc
         pip3 install numpy
         gfortran -v
         mkdir gfortran_version
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -157,7 +157,7 @@ jobs:
     - name: Install Homebrew packages
       # Save the gfortran version to a file so we can use it in the cache key.
       run: |
-        brew install cmake pkgconfig gcc autoconf libtool automake wget pcre doxygen gfortran
+        brew install cmake pkgconfig gcc autoconf libtool automake wget pcre doxygen
         pip3 install numpy
         gfortran -v
         mkdir gfortran_version
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -157,7 +157,8 @@ jobs:
     - name: Install Homebrew packages
       # Save the gfortran version to a file so we can use it in the cache key.
       run: |
-        brew install cmake pkgconfig gcc autoconf libtool automake wget pcre doxygen
+        brew install cmake pkgconfig autoconf libtool automake wget pcre doxygen
+        brew reinstall gcc
         pip3 install numpy
         gfortran -v
         mkdir gfortran_version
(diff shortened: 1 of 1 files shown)


aymanhab:
LGTM, thanks @nickbianco 
@adamkewley Any thoughts on using gcc from brew? Does it have any downstream effect on clients? Thank you

aymanhab (approved):
LGTM but I'm requesting another review from @adamkewley who maybe more familiar with what's available on client machines by default

adamkewley:
It shouldn't matter for *building* from ci necessarily, if what you're building is library-free (c++ stdlib) c or c++ code - however, I wouldn't necessarily know much about this because I don't think I've used gcc much under Mac (most Mac builds build using clang under xcode and use libc++ rather than stdlibc++). I haven't read enough into the Mac version of the moco build, but I'm assuming it only uses standard compilers etc.

If you're using it for runtime dependencies (eg gfortran) the main concern is that the dependency is packaged up in the final package, and that it has transitive dependencies only on the operating system. Usual way to check this is with the apple equivalent to ldd

aymanhab:
Thanks @adamkewley will merge and see what happens at runtime to Moco users on osx 😨

aymanhab:
@nickbianco The latest application build of master fails to launch because of missing library gfortran, as of now we do not distribute it. Considering the potential cascade of dependencies that we need to include and that this appears related to compiler upgrade on github actions, I'm wondering if it'd be better to build the application with the older gcc compiler (that apparently didn't need separate gfortran) what are your thoughts? Trying to build on my machine fails, however, due to the error below, is that related? Shouldn't BLAS be there regardless?
Thank you

configure: error: Required package BLAS not found.
configure: error: /bin/sh '/Users/aymanhabib/dev/build/opensim-core-dependencies/ipopt-prefix/src/ipopt/Ipopt/configure' failed for Ipopt

aymanhab:
@chrisdembia Can you give us some guidance here since you integrated ipopt earlier? Any reason it looks for BLAS and failing when simbody doesn't? Did we distribute gfortran library with earlier distributions of Moco? Thank you

chrisdembia:
We do distribute gfortran libraries; our CMakeLists contain rules for copying gfortran into our installation. See here: https://github.com/opensim-org/opensim-core/blob/0fe7873c51a59539ac116c26b8be0ca2ffe7b29f/Vendors/tropter/cmake/CMakeLists.txt#L5-L25. This is done by tropter; is tropter disabled?

chrisdembia:
Homebrew doesn't have a separate package for gfortran. There is only the `gcc` package (aliased as `gfortran`). I'm a little confused regarding whether the current issue is run-time or compile-time.

aymanhab:
It is runtime with the latest build environment/gcc compiler, I mistakenly thought it was due to new requirement of gfortran. The latest build on master of the gui fails to launch because gfortran can't be found even though build and tests succeed.

aymanhab:
The issue actually is that gfortran can't load libquadmath.0.dylib even though it lives in the same folder with same rpath according to otool if that rings a bell

chrisdembia:
Ah that is very helpful.

I see that the installation contains `libgfortran.dylib` and `libgfortran.5.dylib`. `libgfortran.dylib` has the correct rpath (`@loader_path/`), but `libgfortran.5.dylib` has no rpath. And ipopt is linking to `libgfortran.5.dylib`, not `libgfortran.dylib`. These would ideally be symlinks, but they aren't. 

I think the solution is to add the following line after https://github.com/opensim-org/opensim-core/blob/0fe7873c51a59539ac116c26b8be0ca2ffe7b29f/Vendors/tropter/cmake/tropter_install_mac_dependency_libraries.cmake.in#L114

`install_name_tool_add_rpath(gfortran.5)`

aymanhab:
Excellent, will give that a shot, thanks @chrisdembia 👍

aymanhab:
The error now when launching the GUI is due to libgfortran referencing /usr/local/Cellar/gcc/10.2.0/lib/gcc/10/libquadmath.0.dylib which exists on the build machine but not on the client machine. Any hints @nickbianco or @chrisdembia

chrisdembia:
libquadmath.0.dylib is in our distribution. I think we need to just update the rpath of libgfortran; the commands to do this are already in the CMake file i mentioned above.

-Chris

PR #2924, opened 2020-12-17
Change CompoundFunction to support multiple independent coordinates.

Henrik-Norgren:
CompoundFunction was previously hard-coded to support only 1 independent coordinate, even though CoordinateCouplerConstraint should support multiple. This lead to an assertion error. See https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=12854&p=0&start=0&view=&sid=0547f092d8a26c303d8709e4766bbc07 for more info about the issue.

### Brief summary of changes
Changes CompoundFunction so that it supports functions with multiple inputs. So for a function with N inputs, CompoundFunction takes N+1 inputs (previously hard-coded to 2).

### Testing I've completed
Tested with a LinearFunction and two independent coordinates. No assertion errors and the resulting coordinates behave as expected.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2924)

### DIFF
--- a/OpenSim/Simulation/SimbodyEngine/CoordinateCouplerConstraint.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/CoordinateCouplerConstraint.cpp
@@ -28,13 +28,13 @@
 #include <OpenSim/Simulation/Model/Model.h>
 #include "simbody/internal/Constraint.h"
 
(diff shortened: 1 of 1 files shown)


nickbianco:
Fixed via #3435.

Issue #2926, opened 2021-01-04
In Moco, throw an exception when using joints where qdot != u 


nickbianco:
Moco does not support joints where the number of coordinates does not match the number of generalized speeds.

nickbianco:
@aseth1 this includes the Scapulothoracic joint model. Are you trying to use that with Moco?

nickbianco:
Closing per #2929.

PR #2929, opened 2021-01-11
[Moco] Disallow joints where qdot != u

nickbianco:
Fixes issue #2926

### Brief summary of changes
- Added a check in MocoProblemRep that throws an Exception if there are present any joints where the derivative of the generalized coordinates **does not** equal the generalized speeds (i.e., BallJoints, FreeJoints, EllipsoidJoints, and ScapulothoracicJoints). 

### Testing I've completed
- Updated `testMocoInterface.cpp` to ensure that disallowed joints indeed cause an exception to be thrown.
- Disabled tests that check for extra quaternion slots since all joints that have extra slots also have generalized speeds that do not match the derivative of the generalized coordinates.

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2929)

### DIFF
--- a/CHANGELOG_MOCO.md
+++ b/CHANGELOG_MOCO.md
@@ -3,6 +3,11 @@ Moco Change Log
 
 0.5.0
 -----
+- 2021-01-11: An Exception is now thrown if the model includes joints whose
+              generalized speeds do not match the derivative of the generalized
+              coordinates (i.e., BallJoint, FreeJoint, EllipsoidJoint, and
+              ScapulothoracicJoint).
+  
 - 2020-12-10: Fixed bugs where ModOpReplaceMusclesWithDeGrooteFregly2016 didn't 
               include the muscle PathWrapSet and muscle wrapping wasn't 
               thread-safe (preventing parallelization with MocoCasADiSolver).
--- a/OpenSim/Moco/MocoProblemRep.cpp
+++ b/OpenSim/Moco/MocoProblemRep.cpp
@@ -28,6 +28,10 @@
 #include <unordered_set>
 
 #include <OpenSim/Simulation/SimulationUtilities.h>
+#include <OpenSim/Simulation/SimbodyEngine/ScapulothoracicJoint.h>
+#include <OpenSim/Simulation/SimbodyEngine/BallJoint.h>
+#include <OpenSim/Simulation/SimbodyEngine/EllipsoidJoint.h>
+#include <OpenSim/Simulation/SimbodyEngine/FreeJoint.h>
 
 using namespace OpenSim;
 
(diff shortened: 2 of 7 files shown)


aymanhab (changes requested):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2929)* status: all files reviewed, 2 unresolved discussions (waiting on @nickbianco)

___
*[OpenSim/Moco/MocoProblemRep.cpp, line 143 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2929#-MQsHYZCEaiA0n0hiXKo:-MQsHYZCEaiA0n0hiXKp:b-1wyozw) ([raw file](https://github.com/opensim-org/opensim-core/blob/e19f4c50686515481a1fb8ddd279395c9c2ddf46/OpenSim/Moco/MocoProblemRep.cpp#L143)):*
> ```cpp
>         }
>     }
> ```

If this list is going to change overtime, it may be a bit more general to just call "getConcreteClassName()" and lookup in a table instead of potentially multiple dynamic casts. I notice the message is slightly different tho. I don't feel strongly  since I'm not sure how often this will be changed.
___
*[OpenSim/Moco/Test/testMocoInterface.cpp, line 1858 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2929#-MQsILxB7LTm1pL4qEC3:-MQsILxB7LTm1pL4qEC4:bj2wwx9) ([raw file](https://github.com/opensim-org/opensim-core/blob/e19f4c50686515481a1fb8ddd279395c9c2ddf46/OpenSim/Moco/Test/testMocoInterface.cpp#L1858)):*
> ```cpp
> //            false, false, "implicit");
> //}
> ```

Any reason not to remove this block instead of commenting out to reduce clutter?

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2929)* status: 2 of 6 files reviewed, 2 unresolved discussions (waiting on @aymanhab)

___
*[OpenSim/Moco/MocoProblemRep.cpp, line 143 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2929#-MQsHYZCEaiA0n0hiXKo:-MQsYa1E92J9f2ejI6Zb:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/e19f4c50686515481a1fb8ddd279395c9c2ddf46/OpenSim/Moco/MocoProblemRep.cpp#L143)):*

If this list is going to change overtime, it may be a bit more general to just call "getConcreteClassName()" and lookup in a table instead of potentially multiple dynamic casts. I notice the message is slightly different tho. I don't feel strongly  since I'm not sure how often this will be changed.
</blockquote></details>

Done.
___
*[OpenSim/Moco/Test/testMocoInterface.cpp, line 1858 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2929#-MQsILxB7LTm1pL4qEC3:-MQsQEmk3XoI_CL_VP5w:boa09d6) ([raw file](https://github.com/opensim-org/opensim-core/blob/e19f4c50686515481a1fb8ddd279395c9c2ddf46/OpenSim/Moco/Test/testMocoInterface.cpp#L1858)):*

Any reason not to remove this block instead of commenting out to reduce clutter?
</blockquote></details>

We'll want to reenable these tests whenever we support these types of joints, so I'd like to keep these commented out here.

aymanhab:
@nickbianco Let me know when this ready for review again. Thank you

nickbianco:
Ready for review.

aymanhab (changes requested):
> Reviewed 4 of 4 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2929)* status: all files reviewed, 1 unresolved discussion (waiting on @nickbianco)

___
*[OpenSim/Moco/Test/testMocoGoals.cpp, line 635 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2929#-MQwc0KO0YjtBOr_cZgQ:-MQwc0KO0YjtBOr_cZgR:bvfsutn) ([raw file](https://github.com/opensim-org/opensim-core/blob/991b64bb1691c6155abf8312124c2a79f2012db7/OpenSim/Moco/Test/testMocoGoals.cpp#L635)):*
> ```cpp
> 
>     auto* joint = new PlanarJoint("joint", model.getGround(), *body);
>     model.addJoint(joint);
> ```

The change from FreeJoint to PlanarJoint looks fine, but did you expect that to have worked even if qdot != u? Does that imply that the condition is a bit too strict and some problems can still be solved with these types of joints?

aymanhab (commented):
LGTM, just wanted to clarify one question. Thanks @nickbianco 

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2929#-:-MQwcv490y7EuEmGlxj5:b7ovzah)* status: all files reviewed, 1 unresolved discussion (waiting on @nickbianco)

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2929)* status: all files reviewed, 1 unresolved discussion (waiting on @aymanhab)

___
*[OpenSim/Moco/Test/testMocoGoals.cpp, line 635 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/2929#-MQwc0KO0YjtBOr_cZgQ:-MQx89wP8oP-MeU6GAFX:bavutex) ([raw file](https://github.com/opensim-org/opensim-core/blob/991b64bb1691c6155abf8312124c2a79f2012db7/OpenSim/Moco/Test/testMocoGoals.cpp#L635)):*

The change from FreeJoint to PlanarJoint looks fine, but did you expect that to have worked even if qdot != u? Does that imply that the condition is a bit too strict and some problems can still be solved with these types of joints?
</blockquote></details>

I don't think we thought carefully about the use of FreeJoints in Moco. We account for the extra quaternion slot in the state, and disallow joints where the number of coordinates doesn't equal the number of generalized speeds, but even if `size(q) == size(u)`, it's possible that `qdot != u`. Regarding your comment, FreeJoints might work fine for certain problems (e.g., simple ones like this), but in general we should disallow them.

aymanhab (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2929#-:-MQxBonZC7LKUeIslEfE:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nickbianco)

nickbianco:
Thanks @aymanhab!

PR #2931, opened 2021-01-18
[Moco] Step time and step length asymmetry goals

nickbianco:
### Brief summary of changes
- Added MocoStepTimeAsymmetryGoal and MocoStepLengthAsymmetryGoal
- Added example2DWalkingStepAsymmetry

### Testing I've completed
- Tested each goal locally (via Russell Johnson)
- Confirmed that the Matlab example works as expected

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2931)


aymanhab:
@nickbianco I'm missing all context here, did you review this earlier?

nickbianco:
No, this has been WIP for a few months as I've iterated on it with Russell Johnson. It's now ready for review and I'd like to include the example with the 4.3 release.

Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m
```diff
@@ -0,0 +1,452 @@
...
+%  - STA = (RST - LST) / (RST + LST)
+%
+% Asymmetry values range from -1.0 to 1.0. For example, 0.20 is 20% positive 
+% step time asymmetry with greater right step time than left step time.
+%
+% The step time goal works by "counting" the number of nodes that each foot is in 
+% contact with the ground (with respect to a specified contact force threshold). 
+% Since, in walking, there are double support phases where both feet are on the 
+% ground, the goal also detects which foot is in front and assigns the step time 
+% to the leading foot. Altogether, it calculates the time between consecutive
```
russelljohnson95, line 56: I think we should changes this to "Altogether, it _estimates_ the time between consecutive heel strikes in order to _infer_ left and right step times"

OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.h
```diff
@@ -0,0 +1,162 @@
...
+
+/** Minimize the error between a model's step time asymmetry and a specified
+target asymmetry value over a gait cycle.
+
+Step Length Asymmetry (SLA) is a percentage and is calculated as follows:
+The Right Step Length (RSL) is the distance between feet at right foot strike
+The Left Step Length (LSL) is the distance between feet at left foot strike
+Step Length Asymmetry = (RSL - LSL) / (RSL + LSL)
+
+In this goal, the distance between feet throughout the gait cycle is controlled
```
russelljohnson95, line 33: update the comments here to mirror the comments in the `example2DWalking_StepAsymmetry.m` script for step length asymmetry.

OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.h
```diff
@@ -0,0 +1,162 @@
...
+length asymmetry with greater right step length than left step length.
+
+The target asymmetry can be set via the 'target_asymmetry' property; a symmetric
+step time solution can be achieved by setting this property to zero. This goal
+can be used only in 'cost' mode, where the error between the target asymmetry
+and model asymmetry is squared. To make this goal suitable for gradient-based
+optimization, step length values are assigned via a smoothing function which can
+be controlled the 'asymmetry_smoothing' property.
+
+In order for this goal to work properly, users must prescribe both the stride
```
russelljohnson95, line 49: definitely want to make sure this paragraph is updated as above...

OpenSim/Moco/MocoGoal/MocoStepTimeAsymmetryGoal.h
```diff
@@ -0,0 +1,237 @@
...
+            const std::vector<std::string>& contactForcePaths,
+            const std::string& footPositionForcePath);
+private:
+    void constructProperties();
+};
+
+/** Minimize the error between a model's step time asymmetry and a specified
+target asymmetry value over a gait cycle.
+
+Step Time Asymmetry (STA) is a percentage and is calculated as follows:
```
russelljohnson95, line 59: Mirror the comments in the example matlab script here as well?

russelljohnson95 (commented):
Hi Nick - I left a few comments here, let me know if you have any questions 
- Russell

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2931)* status: 0 of 11 files reviewed, 19 unresolved discussions (waiting on @aymanhab, @nickbianco, and @russelljohnson95)

___
*[Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m, line 56 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MdEFih5TEPHdiUJBYpH-r1-56:-Me3RD3A17e8nU4WBvkm:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m#L56)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

I think we should changes this to "Altogether, it _estimates_ the time between consecutive heel strikes in order to _infer_ left and right step times"
</blockquote></details>

Done.
___
*[Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m, line 113 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MdEFih5TEPHdiUJBYpH-r1-113:-Me3RRwu6kKQqArlLPFd:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m#L113)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

% The settings here have been modified from the default values to suit this specific problem
</blockquote></details>

Done.
___
*[Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m, line 168 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MdEFih5TEPHdiUJBYpH-r1-168:-Me3ReyD49Ar0IlwIm3O:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m#L168)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

% If it doesn't exist, users can run `example2DWalking.m` to generate this file
</blockquote></details>

Done.
___
*[Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m, line 210 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MdEFih5TEPHdiUJBYpH-r1-210:-Me3RuBp0f9DPfUCHswV:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m#L210)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

change above paragraph to: 

% This goal works by constraining the distance between feet, or "foot
% frames", throughout the gait cycle. The goal calculates the distance
% between the left foot and right foot, then limits the distance between
% feet to not pass beyond minimum (negative) or maximum (positive) bounds.
% There are two limits used: one that limits the distance between feet when
% the right foot is in front, and one that limits the distance between feet
% when the left foot is in front.
</blockquote></details>

Done.
___
*[Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m, line 228 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MdEFih5TEPHdiUJBYpH-r1-228:-Me3S8Yj-zB1sQGfeQOm:bqwunu6) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m#L228)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

update above paragraph to clarify that users don't need to specify gait speed - if they don't want: 

% Users must prescribe the stride length for this optimization. The stride
% length is specified via the 'stride_length' property, but users must
% ensure that this stride length is met via problem bounds or other goals;
% the property value is only used to compute the model's asymmetry in the
% cost function. The 'stride_length' and 'setTargetAsymmetry' properties
% are used to compute the two limits:
% right_step_limit = 0.5 * (1 + TargetAsymmetry) * StrideLength;
% left_step_limit  = 0.5 * (1 - TargetAsymmetry) * StrideLength;
</blockquote></details>

This change is fine, but I'm inclined to leave out the stuff about step limits, unless you want to provide more context about what those equations are for.
___
*[Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m, line 235 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MdEFih5TEPHdiUJBYpH-r1-235:-Me3TldJ-owkkqJ--aWr:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m#L235)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

update above paragraph: 
% Because this goal doesn't directly compute the step length asymmetry from
% heel strike data, users should confirm that the step length asymmetry
% from the solution matches closely to their target.  To do this, we
% provide the helper function computeStepAsymmetryValues() below. Users may
% also want to confirm that the stride length from the optimization
% matches with setStrideLength, or set additional constraints for stride length 
% within the optimization. Additionally, in some cases
% users may want to set target asymmetries above or below the desired
% value, in the event there is some offset.

</blockquote></details>

Done.
___
*[Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m, line 329 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MdEFih5TEPHdiUJBYpH-r1-329:-Me3U-9wA5fBJBLC9r6D:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m#L329)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

change to `if exist('gaitTracking_solution_fullStride.sto', 'file')`
</blockquote></details>

Done.

nickbianco (commented):
Thanks @russelljohnson95! I went through everything and covered most of it, but I had a few questions for you.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-:-Me7KLA37H-ZpGb43_St:b-se56h8)* status: 0 of 11 files reviewed, 19 unresolved discussions (waiting on @aymanhab and @russelljohnson95)

___
*[Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m, line 129 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MdEFih5TEPHdiUJBYpH-r1-129:-Me78jVkD4OYMDHIOT5y:b-lr6h4d) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m#L129)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

As discussed: consider multiplying the cost inside the goal by 100 to be able to reduce the setWeight here?
</blockquote></details>

Done. If we changed this weight to 5 or 10, would it change the result significantly?
___
*[Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m, line 287 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MdEFih5TEPHdiUJBYpH-r1-287:-Me78gfiBgiGHOKsICDP:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m#L287)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

As discussed: scale cost within the cost function so that users don't need to use such a large weighting
</blockquote></details>

Done.
___
*[Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m, line 295 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MdEFih5TEPHdiUJBYpH-r1-295:-Me79wCDESj18qpwT1VT:b-o0n12x) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m#L295)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

Also note that in my email I suggested we might not need this smoothing variable, but upon looking at it again, we definitely need it! 
</blockquote></details>

I'm not sure I follow the logic here. Either way, I don't think users will make the connection between the smoothing values and the cost function. I think most will just interpret the higher values as "more smoothing". 

What do you think is a reasonable range of values the way it's currently defined? Just wondering what would happen if it was scaled down by 100 (for the input). In other words, is there a large difference in the smoothing between, for example, 100 and 500?
___
*[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.h, line 33 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MWaAcrkNhWJmM8tzLcK-r1-33:-Me7KBjc5yKyUpSN1fKg:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.h#L33)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

update the comments here to mirror the comments in the `example2DWalking_StepAsymmetry.m` script for step length asymmetry. 
</blockquote></details>

Done.
___
*[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.h, line 49 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MWaAcrkNhWJmM8tzLcK-r1-49:-Me7KC398lyMIZrKFZQE:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.h#L49)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

definitely want to make sure this paragraph is updated as above...
</blockquote></details>

Done.
___
*[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp, line 80 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MWaAcrkNhWJmM8tzLcJ-r1-80:-Me7BcKI7M552XjlE6Nw:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp#L80)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

this might be unimportant, but do we need realizeVelocity() here now? I think everything is in position here? 
</blockquote></details>

Done.
___
*[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp, line 83 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MWaAcrkNhWJmM8tzLcJ-r1-83:-Me7BixX7ziXk5NzN_cD:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp#L83)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

// find position of right foot frame
</blockquote></details>

Done.
___
*[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp, line 86 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MWaAcrkNhWJmM8tzLcJ-r1-86:-Me7BjGRB7L-wUwXJ0mt:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp#L86)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

// find position of left foot frame
</blockquote></details>

Done.
___
*[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp, line 89 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MWaAcrkNhWJmM8tzLcJ-r1-89:-Me7DjNw8gs9Ym3X6fwq:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp#L89)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

// find distance between left and right foot. We want to find the right-sided and left-sided distances, since these have 
// separate limits. rightFootDiff is the right step length and leftFootDiff is the left step length
</blockquote></details>

Done.
___
*[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp, line 92 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MWaAcrkNhWJmM8tzLcJ-r1-92:-Me7DpA038TXetOkOl7r:b-mdqqx2) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp#L92)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

// compute how or if the rightFootDiff or leftFootDiff exceed their respective threshold 
</blockquote></details>

Updated, but I added some extra clarifying comments. Can you double check that these are correct? I know that positive values are for the right foot and negative for the left, so I think that means that `rightFootAsymmetry` is either 0 or +1 and `leftFootAsymmetry` is 0 or -1. 
___
*[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp, line 99 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MWaAcrkNhWJmM8tzLcJ-r1-99:-Me7EZUIEUO06YK5svVc:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp#L99)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

// add the computed asymmetry values (both left and right foot) to the integrand
</blockquote></details>

Done.
___
*[OpenSim/Moco/MocoGoal/MocoStepTimeAsymmetryGoal.h, line 59 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MRME-ASipNKxe7JDahQ-r1-59:-Me7FuEj0ScXZnEbZjWR:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/OpenSim/Moco/MocoGoal/MocoStepTimeAsymmetryGoal.h#L59)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

Mirror the comments in the example matlab script here as well? 
</blockquote></details>

Done.

aymanhab (changes requested):
> Reviewed 6 of 6 files at r3, 4 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2931)* status: 10 of 11 files reviewed, 23 unresolved discussions (waiting on @nickbianco and @russelljohnson95)

___
*[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.h, line 49 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MWaAcrkNhWJmM8tzLcK-r1-49:-MeCMj49B5YSHlVoR-Rh:b-isopjn) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.h#L49)):*

Done.
</blockquote></details>

prescribed->prescribe? Can the interface be made to enforce strideLength specification rather than saying it in a code/comment? 
___
*[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.h, line 135 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MeCNjOd8R5WkxjEeONp:-MeCNjOd8R5WkxjEeONq:b-ta8rd7) ([raw file](https://github.com/opensim-org/opensim-core/blob/78a701f59497fc128b3bc241e0b865e5c7eb35d7/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.h#L135)):*
> ```objc
>     OpenSim_DECLARE_PROPERTY(stride_length, double,
>             "The known stride length of the simulation. Default: 1.");
>     OpenSim_DECLARE_PROPERTY(walking_direction, std::string,
> ```

Default 1. what? Meter? Foot? 
___
*[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp, line 41 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MeCO7QH-_s-f1e8-iUW:-MeCO7QH-_s-f1e8-iUX:bovpd2b) ([raw file](https://github.com/opensim-org/opensim-core/blob/78a701f59497fc128b3bc241e0b865e5c7eb35d7/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp#L41)):*
> ```cpp
>     std::set<std::string> directions{"positive-x", "positive-y", "positive-z",
>                                      "negative-x", "negative-y", "negative-z"};
>     checkPropertyValueIsInSet(getProperty_walking_direction(), directions);
> ```

In other places we use +x, +y, -z rather than long spelled out, possibly error prone name, but not feeling strongly about changing  it at this point.
___
*[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp, line 63 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MeCO_tfCAnq5U6k79FK:-MeCO_tfCAnq5U6k79FL:b-x8qh6f) ([raw file](https://github.com/opensim-org/opensim-core/blob/78a701f59497fc128b3bc241e0b865e5c7eb35d7/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp#L63)):*
> ```cpp
>         else if (direction == "negative-y") { index = 1; sign = -1; }
>         else if (direction == "negative-z") { index = 2; sign = -1; }
>     };
> ```

What if a typo or direction is not specified as any of these?
___
*[OpenSim/Moco/MocoGoal/MocoStepTimeAsymmetryGoal.h, line 32 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MeCP9l12UFQqMHzAXYx:-MeCP9l12UFQqMHzAXYy:b-8nxxzy) ([raw file](https://github.com/opensim-org/opensim-core/blob/78a701f59497fc128b3bc241e0b865e5c7eb35d7/OpenSim/Moco/MocoGoal/MocoStepTimeAsymmetryGoal.h#L32)):*
> ```objc
> compute step time asymmetry. The MocoStepTimeAsymmetryGoal determines when a foot
> is in contact with the ground when the total contact force from the sum of the
> elements in this group exceeds a user provided threshold.
> ```

The fact that contact is involved here was a bit surprising considering the class name but maybe because the order of review by github is rather random or I'm missing the context.

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2931)* status: 8 of 11 files reviewed, 23 unresolved discussions (waiting on @aymanhab and @russelljohnson95)

___
*[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.h, line 49 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MWaAcrkNhWJmM8tzLcK-r1-49:-MeCV4jTDLDfgHFG7nEc:bs9244u) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.h#L49)):*

prescribed->prescribe? Can the interface be made to enforce strideLength specification rather than saying it in a code/comment? 
</blockquote></details>

Done. The goal now throws an exception if the new default value (-1) is not overwritten. (See `checkPropertyValueIsPositive(getProperty_stride_length());` in the cpp file).
___
*[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.h, line 135 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MeCNjOd8R5WkxjEeONp:-MeCV3fK-MeqGAdwdjaE:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/78a701f59497fc128b3bc241e0b865e5c7eb35d7/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.h#L135)):*

Default 1. what? Meter? Foot? 
</blockquote></details>

Done.
___
*[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp, line 41 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MeCO7QH-_s-f1e8-iUW:-MeCRJmN2TSmcpHvhs98:b-vix8bk) ([raw file](https://github.com/opensim-org/opensim-core/blob/78a701f59497fc128b3bc241e0b865e5c7eb35d7/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp#L41)):*

In other places we use +x, +y, -z rather than long spelled out, possibly error prone name, but not feeling strongly about changing  it at this point.
</blockquote></details>

To me, the spelled out names are clearer. `checkPropertyValueIsInSet()` will throw an exception if a direction name is spelled wrong.
___
*[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp, line 63 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MeCO_tfCAnq5U6k79FK:-MeCRdMr-q-6TjcB2abj:bltmdzr) ([raw file](https://github.com/opensim-org/opensim-core/blob/78a701f59497fc128b3bc241e0b865e5c7eb35d7/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp#L63)):*

What if a typo or direction is not specified as any of these?
</blockquote></details>

As specified above, `checkPropertyValueIsInSet()` will catch an typos.
___
*[OpenSim/Moco/MocoGoal/MocoStepTimeAsymmetryGoal.h, line 32 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MeCP9l12UFQqMHzAXYx:-MeCRysp5X6ut_p_GsDv:b-78fzkn) ([raw file](https://github.com/opensim-org/opensim-core/blob/78a701f59497fc128b3bc241e0b865e5c7eb35d7/OpenSim/Moco/MocoGoal/MocoStepTimeAsymmetryGoal.h#L32)):*

The fact that contact is involved here was a bit surprising considering the class name but maybe because the order of review by github is rather random or I'm missing the context.
</blockquote></details>

Ideally both classes wouldn't need contact, but knowing the foot contact times if valuable. It's unlikely that we'll rename the class at this point, and I probably wouldn't want to make it longer anyway.

russelljohnson95:
> Thanks @russelljohnson95! I went through everything and covered most of it, but I had a few questions for you.
> 
> > _[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-:-Me7KLA37H-ZpGb43_St:b-se56h8)_ status: 0 of 11 files reviewed, 19 unresolved discussions (waiting on @aymanhab and @russelljohnson95)
> 
> _[Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m, line 129 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MdEFih5TEPHdiUJBYpH-r1-129:-Me78jVkD4OYMDHIOT5y:b-lr6h4d) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m#L129)):_
> 
> Previously, russelljohnson95 (Russell Johnson) wrote…
> > As discussed: consider multiplying the cost inside the goal by 100 to be able to reduce the setWeight here?
> 
> Done. If we changed this weight to 5 or 10, would it change the result significantly?

It looks like a weight of 5 should work fine. I get the same answer with 5 as I did with 8. Let's go with 5 for the example for the sake of using a "normal" number. 


> _[Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m, line 295 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MdEFih5TEPHdiUJBYpH-r1-295:-Me79wCDESj18qpwT1VT:b-o0n12x) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m#L295)):_
> 
> Previously, russelljohnson95 (Russell Johnson) wrote…
> > Also note that in my email I suggested we might not need this smoothing variable, but upon looking at it again, we definitely need it!
> 
> I'm not sure I follow the logic here. Either way, I don't think users will make the connection between the smoothing values and the cost function. I think most will just interpret the higher values as "more smoothing".
> 
> What do you think is a reasonable range of values the way it's currently defined? Just wondering what would happen if it was scaled down by 100 (for the input). In other words, is there a large difference in the smoothing between, for example, 100 and 500?

I just tried this goal with a smoothing of 100 and it seemed to work OK, but I still think 500 is a better value for this right now. I guess I'm having a difficult time explaining it, but when I was working on trying different smoothing values, I was noticing that the smaller values seemed to warp the asymmetry away from what the true value is. But let me try again: 
So we have a "limit" that we place on how far away the feet can be throughout the gait cycle (let's say for example 0.30 m). We want to know if the distance between the feet for a given iteration exceeds this value, and we want to "penalize" that. With lower smoothing coefficients (i.e., 100), a value of (let's say for example) 0.28 m gets compared to 0.30 m threshold (=0.02) and then passes through the smoothing function and what we actually get is a non-zero value, even though it didn't exceed the threshold. If the smoothing coefficient is higher (i.e., 800), then we get a value that is basically zero instead. It's maybe a little non-intuitive if a user is thinking about smoothing, but I guess it's important to know what is being smoothed - its that foot-distance minus threshold equation. 

Happy to keep "smoothing" (whoops, a pun) out this explanation with you though. 


> _[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp, line 92 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MWaAcrkNhWJmM8tzLcJ-r1-92:-Me7DpA038TXetOkOl7r:b-mdqqx2) ([raw file](https://github.com/opensim-org/opensim-core/blob/5c7db064b5de20fbf6e5c96f1b0d0eb2603e4906/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp#L92)):_
> 
> Previously, russelljohnson95 (Russell Johnson) wrote…
> Updated, but I added some extra clarifying comments. Can you double check that these are correct? I know that positive values are for the right foot and negative for the left, so I think that means that `rightFootAsymmetry` is either 0 or +1 and `leftFootAsymmetry` is 0 or -1.
> 

Actually I believe this is incorrect. For example, if the rightFootStepLength does not exceed the right side threshold (though see above about the smoothing term), then the value of rightFootAsymmetry is zero, likewise on the left side. If the rightFootStepLength exceed the threshold, then the value is positive, and same thing on the left side. Maybe the way to say it is we are accumulating "error" in whether the feet exceed the distance thresholds over time. If they don't, they get values of zero. If they do, they get positive values that end up increasing the objective function value for this goal. 

Again, happy to keep helping with figuring out the correct wording for this, what I have here is probably not the best wording either, but hopefully the concept is clearer(?). 

Let me know how else I can help!

nickbianco (commented):
Thanks @russelljohnson95, this makes more sense to me now. I've updated the comments in the step length cost accordingly. 

I don't think I was clear about what I meant about rescaling the smoothing parameter value, so I just went ahead and made the change. Let me know what you think.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-:-MeW1EcH2Z4j5CEUjUve:b-r7dw0a)* status: 8 of 11 files reviewed, 23 unresolved discussions (waiting on @aymanhab and @russelljohnson95)

Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m
```diff
@@ -0,0 +1,472 @@
...
+
+% Step length asymmetry
+% ---------------------
+stepLengthAsymmetry = MocoStepLengthAsymmetryGoal();
+stepLengthAsymmetry.setWeight(5);
+% Provide the body name for the right foot.
+stepLengthAsymmetry.setRightFootFrame('/bodyset/calcn_r');    
+% Provide the body name for the left foot.
+stepLengthAsymmetry.setLeftFootFrame('/bodyset/calcn_l');     
+% Value for smoothing term use to compute asymmetry (default is 500). Users may
```
russelljohnson95, line 311: Default is 5 now, not 500. Maybe add note about how the smoothing is scaled *100 in the goal itself.

OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp
```diff
@@ -0,0 +1,126 @@
...
+    // have separate limits when computing the asymmetry below, we compute them
+    // separately here.
+    const double rightFootStepLength = rightFootPosition - leftFootPosition;
+    const double leftFootStepLength = leftFootPosition - rightFootPosition;
+
+    // Compute if the right foot or left foot step lengths exceed their
+    // respective thresholds. If so, count this time point towards the
+    // asymmetry error (for each foot independently) via the smoothing functions.
+    // The user input smoothing value is scaled by 100 here to produce a more
+    // noticeable effect on the asymmetry error.
```
russelljohnson95, line 103: yes, I like this new description!

russelljohnson95 (commented):
Just one note on the comment in the example code. Looks good!

nickbianco (commented):
Thanks @russelljohnson95!

@aymanhab, let me know if you have any final comments.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-:-MeW7keI9IyUsd8u_PS6:b-50a53n)* status: 8 of 11 files reviewed, 25 unresolved discussions (waiting on @aymanhab and @russelljohnson95)

___
*[Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m, line 311 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MdEFih5TEPHdiUJBYpH-r6-311:-MeW6piI9qnRnhLqim2I:b-otlanz) ([raw file](https://github.com/opensim-org/opensim-core/blob/55bc747893abe9d6536662f92da17dfe2bcd3cc6/Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m#L311)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

Default is 5 now, not 500. Maybe add note about how the smoothing is scaled *100 in the goal itself. 
</blockquote></details>

Good catch! I would make a note about the scaling if there were meaningful units to the smoothing, but it is still somewhat an arbitrary scalar value. So relative values are more important here (that's mainly why I wanted to rescale in the first place).
___
*[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp, line 103 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MWaAcrkNhWJmM8tzLcJ-r6-103:-MeW6jtu-2nhq3O8ePhT:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/55bc747893abe9d6536662f92da17dfe2bcd3cc6/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp#L103)):*
<details><summary><i>Previously, russelljohnson95 (Russell Johnson) wrote…</i></summary><blockquote>

yes, I like this new description! 
</blockquote></details>

Done.

russelljohnson95:
> Thanks @russelljohnson95!
> 
> @aymanhab, let me know if you have any final comments.
> 
> > _[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-:-MeW7keI9IyUsd8u_PS6:b-50a53n)_ status: 8 of 11 files reviewed, 25 unresolved discussions (waiting on @aymanhab and @russelljohnson95)
> 
> _[Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m, line 311 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MdEFih5TEPHdiUJBYpH-r6-311:-MeW6piI9qnRnhLqim2I:b-otlanz) ([raw file](https://github.com/opensim-org/opensim-core/blob/55bc747893abe9d6536662f92da17dfe2bcd3cc6/Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalkingStepAsymmetry.m#L311)):_
> 
> Previously, russelljohnson95 (Russell Johnson) wrote…
> Good catch! I would make a note about the scaling if there were meaningful units to the smoothing, but it is still somewhat an arbitrary scalar value. So relative values are more important here (that's mainly why I wanted to rescale in the first place).
> 

Makes sense. Seems good now!

> _[OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp, line 103 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-MWaAcrkNhWJmM8tzLcJ-r6-103:-MeW6jtu-2nhq3O8ePhT:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/55bc747893abe9d6536662f92da17dfe2bcd3cc6/OpenSim/Moco/MocoGoal/MocoStepLengthAsymmetryGoal.cpp#L103)):_
> 
> Previously, russelljohnson95 (Russell Johnson) wrote…
> Done.

@nickbianco do you want me to download and run this latest version to make sure it all syncs correctly before the final merge?

nickbianco (commented):
@russelljohnson95 the most recent changes shouldn't change the results, but it would be good confirm if you want to give it a final spin!

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2931#-:-MeWJQ742JxEluYHMklp:bbhg8ar)* status: 7 of 11 files reviewed, 25 unresolved discussions (waiting on @aymanhab and @russelljohnson95)

aymanhab (approved):
> Reviewed 1 of 3 files at r6.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2931)* status: 8 of 11 files reviewed, 21 unresolved discussions (waiting on @aymanhab and @russelljohnson95)

aymanhab:
Thanks @RussellJ95 I'm merging but leaving the branch in case you need it, please delete when done or let me know if your testing turns out issues, thank you

russelljohnson95:
@nickbianco 
I tested the new code and found a slightly different step length asymmetry result in the example code than I was getting before (10.5% vs now 11.5%) and I can't quite figure out why. I tried a couple different settings but the difference was fairly consistent – all around 11-12%. I think this isn't a big deal, and the result is still close enough to the target that I think it will be fine. I don’t have enough time to dig down deeper and do a systematic check on everything right now to see if it’s just something like a different initial guess or a change in a setting as things were scaled in different ways. 

I did find that the step length asymmetry goal runs very poorly if no initial guess is provided. I think we document it enough within the example code though that users can figure out how to solve that issue if they come across it. 

A minor request, if there’s still some time, but I found a comment in the example code that sort of bugged me. If we can’t change it because the code already was built that’s fine but here it is in case 

Lines 318-319, we explain it in such a way that makes it seem like the stride length setting will definitely enforce the stride length, but that’s not really the case at all. Instead: 

% Provide the stride length. This setting is used to set the bounds on the
% distances allowed between feet. 
stepLengthAsymmetry.setStrideLength(0.904);  

I'm fine with closing the branch down once these comments get read!

Issue #2937, opened 2021-02-01
Excessive warnings on loading models due to muscle model check


aymanhab:
e.g. message along the lines of
Parameter update for the damped-model: minimum_activation was 0.01 but is now 0.01
should be skipped to reduce clutter

aymanhab:
@carmichaelong

PR #2939, opened 2021-02-02
Only issue warning if value changes upon clamping

aymanhab:
Fixes issue #2937

### Brief summary of changes
Avoid unnecessary warnings if clamping doesn't change values
### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2939)

### DIFF
--- a/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
+++ b/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
@@ -174,14 +174,16 @@ void Millard2012EquilibriumMuscle::extendFinalizeFromProperties()
 
 
     } else { 
-
-        if(get_minimum_activation() > 0.){
+        double min_activation = get_minimum_activation();
+        if (min_activation > 0. && 
+            std::abs(clamp(0, min_activation, 1) -
+                                            min_activation) > SimTK::Eps) {
             log_info("'{}': Parameter update for the damped-model: "
                 "minimum_activation was {} but is now {}",
                    getName(), get_minimum_activation(), 
                    clamp(0, get_minimum_activation(), 1));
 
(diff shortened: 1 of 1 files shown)


aymanhab:
@carmichaelong not sure if this is the proper place for the fix or if other places need to be updated as well. Please let me know

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -174,14 +174,16 @@ void Millard2012EquilibriumMuscle::extendFinalizeFromProperties()
 
 
     } else { 
-
-        if(get_minimum_activation() > 0.){
+        double min_activation = get_minimum_activation();
+        if (min_activation > 0. &&
```
tkuchida, line 178: For better readability, perhaps calculate the clamped value just once:
```cpp
const double min_activation = get_minimum_activation();
const double clamped_min_activation = clamp(0, min_activation, 1);
if (min_activation > 0. && std::abs(clamped_min_activation - min_activation) > SimTK::Eps) {
    // log_info...
    set_minimum_activation(clamped_min_activation);
}
```

carmichaelong:
> not sure if this is the proper place for the fix or if other places need to be updated as well. Please let me know

This looks to be the right place (and should be the only place needed to update).

@tkuchida Thanks for the review. I think it's good once your comment has been addressed.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
aymanhab: Done

aymanhab:
Merging since test failure on osx is unrelated. Thanks for the review @tkuchida @carmichaelong

Issue #2944, opened 2021-02-10
OpenSim::Model::finalizeFromProperties leaks memory?


adamkewley:
A program like this (current `master` branch):

```c++
#include <OpenSim.h>

int main(int argc, char** argv) {
    OpenSim::Model m{"models/RajagopalModel/Rajagopal2015.osim"};
    for (size_t i = 0; i < 10; ++i) {
        OpenSim::Model m2{m};
        m2.finalizeFromProperties();  // required for the leak
    }
    return 0;
}
```

Leaks *a lot* (>5 MB per iteration) of memory. The relevance of this is when using OpenSim in a long-running application like a GUI, where models may be copied a fair bit (e.g. to launch simulations, undo-redo storage, etc.).

The memory being leaked was allocated during the copy into `m2`. Many leaks look like this:

```text
==9501== 27,180,942 (628,320 direct, 26,552,622 indirect) bytes in 374 blocks are definitely lost in loss record 1,237 of 1,237
==9501==    at 0x4835DEF: operator new(unsigned long) (vg_replace_malloc.c:334)
==9501==    by 0x4F1448D: OpenSim::Millard2012EquilibriumMuscle::clone() const (Millard2012EquilibriumMuscle.h:202)
==9501==    by 0x4AAF9A7: operator= (ArrayPtrs.h:212)
==9501==    by 0x4AAF9A7: OpenSim::ArrayPtrs<OpenSim::Actuator>::operator=(OpenSim::ArrayPtrs<OpenSim::Actuator> const&) (ArrayPtrs.h:197)
==9501==    by 0x4AB8A74: OpenSim::Set<OpenSim::Actuator, OpenSim::Object>::Set(OpenSim::Set<OpenSim::Actuator, OpenSim::Object> const&) (Set.h:116)
==9501==    by 0x5433B4C: ForceSet (ForceSet.h:48)
==9501==    by 0x5433B4C: clone (ForceSet.h:49)
==9501==    by 0x5433B4C: cloneOrNull (ClonePtr.h:312)
==9501==    by 0x5433B4C: ClonePtr (ClonePtr.h:93)
==9501==    by 0x5433B4C: copyConstruct<const SimTK::ClonePtr<OpenSim::ForceSet>*> (Array.h:3182)
==9501==    by 0x5433B4C: Array_ (Array.h:1657)
==9501==    by 0x5433B4C: ObjectProperty (Property.h:1058)
[...]

PR #2951, opened 2021-02-12
Fix iss 2944 forceset leak

adamkewley:
This is a hotfix for the (large) memory leak found in #2944 .

This does not fix all leaks I've found (of which, 98 % come from `ArrayPtr`s). It does, however, fix a source of *large* (multi-megabyte) leaks.

See issue posting for details.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2951)


aymanhab (approved):
LGTM, thanks @adamkewley

Issue #2524, opened 2019-07-11
Use SWIG 4.0 to support documentation in Python and Java.


chrisdembia:
SWIG 4.0 supports translating Doxygen documentation into Python and Java documentation. We should consider upgrading to SWIG 4.0.

http://www.swig.org/Doc4.0/SWIGDocumentation.html#Doxygen

aymanhab:
This should be addressed by PR #3046 for python

Issue #2968, opened 2021-02-26
ModuleNotFoundError No module named '_simbody'


pwesberg:
Hello,

Following setup:
- win10
- anaconda
- visual studio 2019 (v16.8.6)
- cmake 3.19.5
- swigwin 3.0.12
- opensim 4.1
  - opensim python wrapping
- simbody master commit 9daf0d26ea22551a5d10114bffd9b02fa51c53bc

Due to an error while installing opensim dependencies / simbody with latest Version of Visual Studio 2019 compiling simbody 3.7 release I installed simbody manually using simbody bleedingedge master (commit 9daf0d26ea22551a5d10114bffd9b02fa51c53bc). This includes a merge of following pull request, wich solves an issue with the Visual Studio 2019: https://github.com/simbody/simbody/pull/707. Installation (as described here: https://github.com/simbody/simbody/tree/9daf0d26ea22551a5d10114bffd9b02fa51c53bc#windows-using-visual-studio) was successfull (run_tests_parallel succeeded without any error).
After that installed opensim 4.1 release from source. (including opensim python wrapping). Opensim installation with visual studio 2019 was successfull, (run all tests parallel without any error). Installing opensim, set path variable and install setup.py in the conda env were done.

After installing opensim opensim module is recognized in our python repository, but the simbody submodule is not: 
```
Exception has occured: ModuleNotFoundError 
No module named '_simbody'
```

Is that still an error of simbody or a problem with opensim? 
Any idea how to fix this?

kind regards, 
patrisity

achigeor:
I had the same problem on Ubuntu. I found that doing [this](https://github.com/opensim-org/opensim-core/issues/2827#issuecomment-655198736) solved my issue. 

Hope it helps!

antoinefalisse:
@patrisity @aymanhab 

I am having the same issue over and over.
If you upgrade numpy, it should fix the problem: `pip install numpy --upgrade`

However, it gets problematic in some cases. I am using tensorflow and opensim. Tensorflow needs numpy ~=1.19.2, and if you use the fix above, you end up with a newer numpy version that is not supported by tensorflow. If you use the numpy version that comes with tensorflow, you get a numpy version that does not work with opensim. I copy below the full error message, but here is I think an interesting line:

> RuntimeError: module compiled against API version 0xe but this version of numpy is 0xd

Any idea about what to do? @carmichaelong @nickbianco 




> ---------------------------------------------------------------------------
[...]

PR #2976, opened 2021-03-29
Support Swig 4.0

aymanhab:
Fixes issue #2524

### Brief summary of changes
-Upgrade makefiles to require SWIG 4.0
- Include doxygen option to pass docs to python, java 
- Update interface files to 4.0 
- Some ifdef/endif in headers due to issues parsing doxygen comments by SWIG
- Moved methods that traverse componentLists to Model class due to new stricter scoping rules

### Testing I've completed
Standard test suite passes, Java bindings work in GUI and comments available in GUI IDE

How this looks in Matlab and in Python, whether comments are available/helpful

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2976)

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -35,7 +35,7 @@ jobs:
       run: python -m pip install numpy
 
     - name: Install SWIG
-      run: choco install swig --version 4.0.2 --yes --limit-output
+      run: choco install swig --version 4.0.0 --yes --limit-output
 
     - name: Cache dependencies
       id: cache-dependencies
@@ -175,7 +175,7 @@ jobs:
       if: steps.cache-swig.outputs.cache-hit != 'true'
       run: |
         mkdir ~/swig-source && cd ~/swig-source
-        wget https://github.com/swig/swig/archive/rel-4.0.2.tar.gz
+        wget https://github.com/swig/swig/archive/refs/tags/rel-4.0.2.tar.gz
         tar xzf rel-4.0.2.tar.gz && cd swig-rel-4.0.2
         sh autogen.sh && ./configure --prefix=$HOME/swig --disable-ccache
         make && make -j4 install
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -34,7 +34,9 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: '3.7'
-      # Need numpy to use SWIG numpy typemaps.
+        
+    - name: Install numpy
+      #Need numpy to use SWIG numpy typemaps.
       run: python -m pip install numpy
 
     - name: Install SWIG
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -35,7 +35,7 @@ jobs:
       run: python -m pip install numpy
 
     - name: Install SWIG
-      run: choco install swig --version 3.0.12 --yes --limit-output
+      run: choco install swig --version 4.0.2 --yes --limit-output
 
     - name: Cache dependencies
       id: cache-dependencies
@@ -175,8 +175,8 @@ jobs:
       if: steps.cache-swig.outputs.cache-hit != 'true'
       run: |
         mkdir ~/swig-source && cd ~/swig-source
-        wget https://github.com/swig/swig/archive/rel-3.0.12.tar.gz
-        tar xzf rel-3.0.12.tar.gz && cd swig-rel-3.0.12
+        wget https://github.com/swig/swig/archive/rel-4.0.2.tar.gz
+        tar xzf rel-4.0.2.tar.gz && cd swig-rel-4.0.2
         sh autogen.sh && ./configure --prefix=$HOME/swig --disable-ccache
         make && make -j4 install
 
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -342,6 +342,7 @@ jobs:
         OSIM_CMAKE_ARGS+=(-DOPENSIM_C3D_PARSER=ezc3d)
         OSIM_CMAKE_ARGS+=(-DBUILD_PYTHON_WRAPPING=on -DBUILD_JAVA_WRAPPING=on)
         OSIM_CMAKE_ARGS+=(-DSWIG_DIR=~/swig/share/swig)
+        OSIM_CMAKE_ARGS+=(-DSWIG_EXECUTABLE=~/swig/bin/swig)
         OSIM_CMAKE_ARGS+=(-DOPENSIM_INSTALL_UNIX_FHS=OFF)
         OSIM_CMAKE_ARGS+=(-DOPENSIM_DOXYGEN_USE_MATHJAX=off)
         # TODO: Update to simbody.github.io/latest
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -341,7 +341,7 @@ jobs:
         OSIM_CMAKE_ARGS+=(-DOPENSIM_DEPENDENCIES_DIR=~/opensim_dependencies_install)
         OSIM_CMAKE_ARGS+=(-DOPENSIM_C3D_PARSER=ezc3d)
         OSIM_CMAKE_ARGS+=(-DBUILD_PYTHON_WRAPPING=on -DBUILD_JAVA_WRAPPING=on)
-        OSIM_CMAKE_ARGS+=(-DSWIG_DIR=~/swig)
+        OSIM_CMAKE_ARGS+=(-DSWIG_DIR=~/swig/share/swig)
         OSIM_CMAKE_ARGS+=(-DOPENSIM_INSTALL_UNIX_FHS=OFF)
         OSIM_CMAKE_ARGS+=(-DOPENSIM_DOXYGEN_USE_MATHJAX=off)
         # TODO: Update to simbody.github.io/latest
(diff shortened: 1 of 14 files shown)


aymanhab:
@carmichaelong you may want to take a look/review if @nickbianco is not available

nickbianco:
From dev meeting today: create Python (and maybe Java) versions of osimC3D and add tests for them.

nickbianco (changes requested):
> Reviewed 14 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2976)* status: all files reviewed, 2 unresolved discussions (waiting on @aymanhab)

___
*[Bindings/common.i, line 389 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2976#-MXnVztoBBl1oaqmFbB1:-MXnVztoBBl1oaqmFbB2:b-89ew3v) ([raw file](https://github.com/opensim-org/opensim-core/blob/7e8ffc9395f4aca584ef8b567df3acf8fdd46c70/Bindings/common.i#L389)):*
> ```SWIG
>         std::map<std::string, std::shared_ptr<OpenSim::AbstractDataTable> >;
> //%template(StdMapStringTimeSeriesTableVec3)
> //        std::map<std::string, std::shared_ptr<OpenSim::TimeSeriesTable_<SimTK::Vec3> > >;
> ```

Is this supposed to be commented out?
___
*[OpenSim/Tools/InverseKinematicsTool.h, line 111 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2976#-MXnXLLw-cOwj9quFrJI:-MXnXLLw-cOwj9quFrJJ:b-tjkfhr) ([raw file](https://github.com/opensim-org/opensim-core/blob/7e8ffc9395f4aca584ef8b567df3acf8fdd46c70/OpenSim/Tools/InverseKinematicsTool.h#L111)):*
> ```objc
> #ifndef SWIG
>     /** @cond **/ // hide from Doxygen
> #endif
> ```

Will this get exposed if Doxygen is installed but not SWIG?

nickbianco:
The changes look good, just a few minor comments.

Creating a local build now to test osimC3D.

aymanhab (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2976#-:-MXrVv0C65bGYUOeIc07:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @nickbianco)

___
*[Bindings/common.i, line 389 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2976#-MXnVztoBBl1oaqmFbB1:-MXrW-F6AaqsDT8YC8TT:b3q04km) ([raw file](https://github.com/opensim-org/opensim-core/blob/7e8ffc9395f4aca584ef8b567df3acf8fdd46c70/Bindings/common.i#L389)):*

Is this supposed to be commented out?
</blockquote></details>

Yes, this map is used internally and is exposed only by the C3DAdapter class's write method which is never used. If left behind, the code generated by SWIG fails to compile because an Iterator over the Map has to be generated and that was quite difficult to communicate to SWIG and is rather unnecessary.
___
*[OpenSim/Tools/InverseKinematicsTool.h, line 111 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2976#-MXnXLLw-cOwj9quFrJI:-MXrX9fH3cAgAn9nm6tS:b2v9hmm) ([raw file](https://github.com/opensim-org/opensim-core/blob/7e8ffc9395f4aca584ef8b567df3acf8fdd46c70/OpenSim/Tools/InverseKinematicsTool.h#L111)):*

Will this get exposed if Doxygen is installed but not SWIG?
</blockquote></details>

Yes, while running SWIG the block disappears to it can digest comments, if not running SWIG the ifndef/endif is a no-op so Doxygen and any other old code/processor work as expected.

nickbianco (commented):
Changes LGTM. I tested osimC3D locally and it seems to work fine. However, I'm having trouble getting a Python environment to work. I was able to get an older version of pre-release 4.2 working (with changes to the install process I commented here: https://github.com/opensim-org/opensim-core/issues/2968), but the same process hasn't worked for this build (I'm working on Mac). 

I don't want to hold this up unnecessarily, but I also want to make sure that installing Python still works with the new swig changes.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2976#-:-MXs4tfb5I8oHg2AFVs9:b-mu8y4z)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aymanhab)

aymanhab (commented):
Sure, I understand, just keep me posted.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2976#-:-MXs6_RU-9iy0kmhFmLL:b-ahovob)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aymanhab)

nickbianco:
I'm getting the familiar ImportError issue:

image

A quick Google search suggests that this could be due to circular imports, but nothing in `__init__.py` has been changed. Not sure what the issue is. I am using an Anaconda environment, but I was able to get the OpenSim 4.2 release working with an Anaconda environment. When installing with 4.2, the 'opensim' package shows up in the list of packages (when you call `conda list`) in your conda environment. However, with the Mac build I downloaded for this release, it did not appear.

aymanhab:
@nickbianco Can you do plain python 3.7-9 to isolate Anaconda specific issues? The ci build runs python tests so it must be able to import simbody without an issue I'm guessing

aymanhab:
Note that ci was built against python 3.9.2 (just to test) but we could bring it down to 3.7, this could be the difference between this and 4.2 public release which couldn't go above 3.7

nickbianco:
@aymanhab that could be the issue. I thought we we're still building against python 3.7.

nickbianco:
Okay, I got it to work with an Anaconda environment with Python 3.9.2. However, it only works if I'm in the `setup.py` directory. I get the following warning when running `python setup.py install`:

```
UserWarning: The version specified ('4.3-2021-04-08-0253fd959') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
```
In addition, the opensim package still doesn't appear in the conda package list. I'm not sure if there's a conflict with `setuptools` and the version of Python used in the CI build. We might consider rolling back to an earlier Python version.

aymanhab:
@nickbianco I reverted osx and windows builds to use python 3.7. The warning about version number is rather benign (https://stackoverflow.com/questions/27493792/python-and-pep-440-how-serious-is-this-warning-about-pep440) and will not happen in official releases because we don't embed build date and hash in release versions.

nickbianco:
Thanks @aymanhab, I'll try out the new build.

I know the warning is benign, but it mentions potentially loss of compatibility with setuptools, which we need for `python setup.py install`. Just something to keep an eye on.

nickbianco:
Update: I've tried installing the Python bindings with same Homebrew version of Python 3.7 as the CI uses, but I'm still running into issues. I'm now encountering this error:

```
ImportError: dlopen(/Applications/opensim-core-4.3-2021-04-10-63168d063/sdk/Python/opensim/_simbody.so, 2): Symbol not found: _PyCMethod_New
  Referenced from: /Applications/opensim-core-4.3-2021-04-10-63168d063/sdk/Python/opensim/_simbody.so
  Expected in: flat namespace
 in /Applications/opensim-core-4.3-2021-04-10-63168d063/sdk/Python/opensim/_simbody.so
```

I'm there is some issue with Python version compatibility that I'm missing now. This is probably an issue with my environment and not the updates to SWIG. I'm fine with merging this and can sort out the install issues afterwards.

nickbianco (commented):
@aymanhab Okay I finally got a local build working. I needed to set the `PYTHONPATH` in addition to `DYLD_LIBRARY_PATH` on Mac to get it work. I did this using my own local build of the bindings. 

I'm going to try again using the build from CI. If you want, we could test again using the latest version of Python through CI since that was a motivating factor for updating SWIG.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2976#-:-MY1i3saAL-fSEYKhvgl:bb8vpuv)* status: 13 of 14 files reviewed, all discussions resolved (waiting on @nickbianco)

nickbianco (commented):
Actually, I forgot I already had a Python 3.9 build downloaded and set up. Just tested again and it works! So we don't have to force Python 3.7. Once you revert that change and the tests pass, feel free to merge.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2976#-:-MY1lXY_9e2MJQYrya8y:bccm4la)* status: 13 of 14 files reviewed, all discussions resolved (waiting on @nickbianco)

aymanhab (commented):
Thanks @nickbianco much appreciated. So do you prefer we make builds with 3.7 or 3.9?  I prefer we fix a python version either way rather than get the randomly chosen version selected by ci.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2976#-:-MY1uzyU9GLJylxTAdcv:b-z5u89e)* status: 13 of 14 files reviewed, all discussions resolved (waiting on @nickbianco)

nickbianco:
I was going back and forth about this, but I think I agree that we should fix to a specific version (until something in Python changes that necessitates a version upgrade). It could make sense to stay at 3.7 since other Python package might not stay as current ([like Antoine's issue with NumPy and Tensorflow](https://github.com/opensim-org/opensim-core/issues/2968)).

nickbianco (approved):
:lgtm:

> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2976#-:-MY3oxeWALFYZ8gSZU-9:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aymanhab)

aymanhab:
Thanks @nickbianco much appreciated 👍

Issue #2977, opened 2021-04-04
When using CMC with a model with disabled muscles, the Actuation analysis output files are malformed


carmichaelong:
From the forum: https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=13216

Reproduceable with gait10dof18 files and CMC:
1. load subject01.osim
2. disable hamstrings_r
3. run CMC
4. check *_Acutation_* files

Issue:
The file output says nColumns=29, which is correct for the number of columns of data present. However, the 29th column does not have a column header, even though it is all 0's (for hamstrings_r).

carmichaelong:
This same behavior also occurs using the Actuation analysis tool in isolation as well (i.e., using it on the resulting CMC states, rather than running it attached to a CMC simulation)

carmichaelong:
Spot checks on ForceReporter and MuscleAnalysis show that these analyses do not seem to have the same issue. Issue should be isolated to Actuation analysis.

aymanhab:
Which specific files @carmichaelong? I used Arm26 and ran CMC, and the files are:
* _Actuation_{Force, Power, Speed}.sto 
*_controls.sto

carmichaelong:
Oops, I forgot markdown language makes "*" turn into emphasis, and I was using it as a wildcard...

It's the 3 files with Actuation in the name. I'm working on a test and fix right now

aymanhab:
Thanks @carmichaelong I actually made a surgical fix before I saw your reply here:  https://github.com/opensim-org/opensim-core/pull/2979
If you're further along and want to close, or you if you're further along in making a test feel free to close. Thank you

carmichaelong:
Ah yes, that's the exact same line change I've got too. Sorry for not saying I was working on it. Will close your PR if this test is looking good

carmichaelong:
@vbotics We merged a fix for this into master. If you're building from master, this should hopefully fix your issue.

PR #2979, opened 2021-04-07
Account for disabled actuators when allocating output storage

aymanhab:
Fixes issue #2977

### Brief summary of changes
Account for disabled actuators when allocating output storage
### Testing I've completed
Ran CMC on model with disabled muscles

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2979)


carmichaelong:
see #2980

PR #2989, opened 2021-04-15
Introduce OPENSIM_PYTHON_CONDA 

aymanhab:
Fixes issue #0

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2989)


aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2989)* status: 0 of 4 files reviewed, 1 unresolved discussion

___
*[OpenSim/Simulation/Model/Frame.h, line 325 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/2989#-MvQQNVUE2AKmh9PKvA1:-MvQQNVUE2AKmh9PKvA2:b-1lhwo4) ([raw file](https://github.com/opensim-org/opensim-core/blob/4961390473cbc5322c0d545c32bd859d8eed9a31/OpenSim/Simulation/Model/Frame.h#L325)):*
> ```objc
> 
>     /** Accessor for Rotation matrix of the Frame in Ground. */
>     SimTK::Rotation getRotationInGround(const SimTK::State& state) const {
> ```

@carmichaelong Here's the template expansion that was missed by SWIG

aymanhab:
This PR practically move to a more recent CMake so that we can use Python3 constructs and NOT specify numpy externally. 
Python2 support is dropped. The only other material change is adding a variable to indicate building for CONDA as runtime libraries are handled differently then (not copied to installation).

aymanhab:
@carmichaelong or @nickbianco for quick review

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2989)* status: all files reviewed, 2 unresolved discussions (waiting on @carmichaelong and @nickbianco)

___
*[CMakeLists.txt, line 552 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2989#-Mw8U5eP52DyVy-fAj8s:-Mw8ZcNd5AXmZigQvJeY:bnqb95u) ([raw file](https://github.com/opensim-org/opensim-core/blob/e0e7dce7bcf242b2391c8d9046e72f442de5601a/CMakeLists.txt#L552)):*

Not blocking, just curious: how does including NumPy as a required component change building the bindings? It seems like you still need to provide the NumPy include dirs, it would be nice if it found them automatically.
</blockquote></details>

You do not need to specify Numpy directory on your own,  instead it is  auto-populated by CMake find mechanism which guarantees that the Numpy is found within the same Python3 installation (as echoed by the messages I added below)

nickbianco (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/2989#-:-Mw8aDrECNOZWB4hFsj_:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @aymanhab and @carmichaelong)

___
*[CMakeLists.txt, line 552 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/2989#-Mw8U5eP52DyVy-fAj8s:-Mw8a9Cz5MRlZiW5P8Ev:b-lpoyu4) ([raw file](https://github.com/opensim-org/opensim-core/blob/e0e7dce7bcf242b2391c8d9046e72f442de5601a/CMakeLists.txt#L552)):*

You do not need to specify Numpy directory on your own,  instead it is  auto-populated by CMake find mechanism which guarantees that the Numpy is found within the same Python3 installation (as echoed by the messages I added below)
</blockquote></details>

That's great! Thanks for clarifying.

aymanhab:
Awesome, thanks for the review @nickbianco 👍

PR #3002, opened 2021-05-25
Use imus in opensense

aymanhab:
Fixes issue #0

### Brief summary of changes
When using IMUPlacer, create and add IMUs to the calibrated model. Make sure IMUs are added in the ModelComponent Set so that they're accessible to GUI.

### Testing I've completed
Built GUI against this branch, ran IMU placer, calibrated model matched 4.2 and IMUs are accessible.

Whether this needs to be documented since it should be transparent to users

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3002)

### DIFF
--- a/Applications/opensense/test/calibrated_pelvis_rot.osim
+++ b/Applications/opensense/test/calibrated_pelvis_rot.osim
@@ -42,20 +42,6 @@
 								<!--Scale factors in X, Y, Z directions respectively.-->
 								<scale_factors>0.20000000000000001 0.20000000000000001 0.20000000000000001</scale_factors>
 							</FrameGeometry>
-							<!--List of geometry attached to this Frame. Note, the geometry are treated as fixed to the frame and they share the transform of the frame when visualized-->
-							<attached_geometry>
-								<Brick name="pelvis_imu_geom_1">
-									<!--Path to a Component that satisfies the Socket 'frame' of type Frame.-->
-									<socket_frame>..</socket_frame>
-									<!--Default appearance attributes for this Geometry-->
-									<Appearance>
-										<!--The color, (red, green, blue), [0, 1], used to display the geometry. -->
-										<color>1 0.5 0</color>
-									</Appearance>
-									<!--Half lengths in X, Y, Z respectively.-->
-									<half_lengths>0.02 0.01 0.0050000000000000001</half_lengths>
-								</Brick>
-							</attached_geometry>
 							<!--Path to a Component that satisfies the Socket 'parent' of type C (description: The parent frame to this frame.).-->
 							<socket_parent>..</socket_parent>
 							<!--Translational offset (in meters) of this frame's origin from the parent frame's origin, expressed in the parent frame.-->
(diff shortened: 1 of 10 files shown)


aymanhab:
@carmichaelong This PR basically uses the IMU abstraction introduced in PR #2997 so that the IMUPlacer creates IMUs rather than frames and arbitrary blocks of geometry. Allowing GUI to show IMUs and treat them uniformly for Analysis/reporting purposes.

carmichaelong (commented):
Looks good overall. I just had one comment which was more of a question than a blocking comment

> Reviewed 3 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3002#-:-Ma_EEy65jQQbwsTzMC_:b-kgtqb5)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @aymanhab)

___
*[OpenSim/Analyses/IMUDataReporter.cpp, line 159 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3002#-Ma_E6cC7CDZ3cJYqsM7:-Ma_E6cC7CDZ3cJYqsM8:bgjnsd8) ([raw file](https://github.com/opensim-org/opensim-core/blob/5266e9c9c1cd0937c4c583fa40827d0d1336ee94/OpenSim/Analyses/IMUDataReporter.cpp#L159)):*
> ```cpp
>         for (IMU& imu : compList) { 
>             SimTK::ReferencePtr<OpenSim::IMU> imuRef(imu);
>             _imuComponents.push_back(std::shared_ptr<IMU>(imuRef.get())); 
> ```

I assume this was a necessary change, but I'm curious as to why `make_shared` needed to be replaced here?

aymanhab:
@carmichaelong make_shared<> create a new object and the shared pointer, in this case we don't want to create an object, just initialize the shared_ptr with existing pointer. Pretty messy but justified because when using it from GUI I was getting assertions/failures due to the implicit copying due to make_shared

aymanhab:
This is ready for review, the essence of the change is moving away from shared_ptr to Objects that live in a Model due to reference counting issues and mostly sticking with home brew SimTK::ReferencePtr which isn't reference counted. This incarnation works well with corresponding GUI PR

carmichaelong (changes requested):
Thanks @aymanhab, looks good. I left one small comment. One other question in general: It seems like the std smart pointers aren't playing with the GUI very well, so seems like using the SimTK ones is a good solution now. Are there any problems you foresee down the road with doing this? (I wouldn't consider this a blocking question)

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3002#-:-MaywFqg1EguuWyHaiHI:b-zbnnop)* status: 1 of 10 files reviewed, 2 unresolved discussions (waiting on @aymanhab and @carmichaelong)

___
*[OpenSim/Simulation/OpenSense/IMU.h, line 47 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3002#-Mayub-f446bz0STntcO:-Mayub-f446bz0STntcP:b-597xwg) ([raw file](https://github.com/opensim-org/opensim-core/blob/7a99b604454139c76381029cda7ec7f090c508d0/OpenSim/Simulation/OpenSense/IMU.h#L47)):*
> ```objc
>     IMU() {  }
>     virtual ~IMU() { 
>     };
> ```

Why are the constructor and destructor empty rather than default?

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3002)* status: 1 of 10 files reviewed, 2 unresolved discussions (waiting on @aymanhab and @carmichaelong)

___
*[OpenSim/Simulation/OpenSense/IMU.h, line 47 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3002#-Mayub-f446bz0STntcO:-Mazq8mE5OzXvCRxf40r:bhx97t2) ([raw file](https://github.com/opensim-org/opensim-core/blob/7a99b604454139c76381029cda7ec7f090c508d0/OpenSim/Simulation/OpenSense/IMU.h#L47)):*

Why are the constructor and destructor empty rather than default?
</blockquote></details>

With no private data members they should be the same, I mainly introduced the bodies to place debugging/logging lines there but ended up removing. Changed to default for consistency with the remaining constructors/copy/move operators etc.

aymanhab:
@carmichaelong This should be good to go, thanks for the quick turnaround.

carmichaelong (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3002#-:-Mb8-qnt3EAtz4R18UMv:bnfp4nl)* status: 1 of 10 files reviewed, 1 unresolved discussion (waiting on @aymanhab and @carmichaelong)

aymanhab:
Thanks @carmichaelong much appreciated 👍

Issue #2982, opened 2021-04-08
TGCS Workshop Planning


nickbianco:
[TGCS Workshop Proposal](https://docs.google.com/document/d/17MCMD_IdWBSkrqvS6EXyb3DcP_IUNf_DkGs67ufGE-I/edit)

### IMU-tracking problem
- [x] Add options to MocoAccelerationTrackingGoal for gravity offset and converting accelerations to the IMU frame (#2961)
- [ ] Create simulation utility for creating synthetic IMU signals
- [ ] Squat-to-stand "round trip" follow-along Matlab and Python examples (e.g., exampleSquatToStandIMUTracking, exampleIMUTracking)

### EMG-driven problem
- [ ] Add option for scale factor optimization to Moco (#2903)
- [ ] Create standalone Matlab and Python examples based on exampleMocoInverse for EMG-driven simulation in follow-along format (e.g., exampleEMGDriven)

### Python examples
- [ ] Test [existing Python environment installation materials](https://simtk-confluence.stanford.edu/display/OpenSim/Scripting+in+Python) with novice Python users in the lab
- [ ] Provide office hours before meeting for Python setup

### Presentation materials
- [ ] Draft PowerPoint slides
- [ ] Draft email for pre-workshop installation instructions and info about office hours

nickbianco:
@aymanhab @carmichaelong @antoinefalisse @jenhicks

aymanhab:
Opened issue #2983 to discuss details of creating synthetic IMU signals

aymanhab:
@nickbianco With IMU and IMUDataReporter now in place, I'm looking into GUI needs/functionality for the demo, is there an outline or documentation that I should check? Also do you assume/use a noise model? Thanks

aymanhab:
@nickbianco Do we need update doxygen for TGCS build? Also do we need a link from the advanced examples page on confluence to the new Moco Examples?

nickbianco:
Yes, we will need to update Doxygen and the examples page for the new Moco examples.

PR #3004, opened 2021-05-27
TGCS2021 EMG Tracking Problem

nickbianco:
Fixes issue #2982 

### Brief summary of changes

This PR includes the EMG-tracking example for the TGCS2021 workshop and all the changes necessary for the new scale factor optimization features. 

- Add exampleEMGTracking_answers.m (will create version without answers and Python versions after reviews)
- Added component MocoScaleFactor, to enable scale factor optimizations. This component gets added to the Model internal to MocoProblem/MocoProblemRep, providing a property value that we can optimize with MocoParameter.
- The methods `appendScaleFactor()` and `getScaleFactors()` have been added to MocoGoal to provided an interface for derived class to add scale factors (if desired).
- Support for scaling reference controls signals (including the new method `addScaleFactor()`)  was added to MocoControlTrackingGoal .
- Various other small changes.

### Testing I've completed
- Added test case to testMocoGoals.

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004)


nickbianco:
Hey @aymanhab and @carmichaelong, this is now in a state ready for initial reviews. Still some things to be done (including add the Python examples), but like the IMU example, I'd like to do that after receiving some feedback.

Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking_answers.m
```diff
@@ -0,0 +1,290 @@
...
+tracking.setReferenceLabel('/forceset/gasmed_l', 'gastrocnemius');
+tracking.setReferenceLabel('/forceset/tibant_l', 'tibialis_anterior');
+tracking.setReferenceLabel('/forceset/bfsh_l', 'biceps_femoris');
+tracking.setReferenceLabel('/forceset/glmax2_l', 'gluteus');
+
+% Part 3c: The EMG signals in the tracking are all normalized to have
+% a maximum value of 1, but the magnitudes of the excitations from the 
+% effort minimization solution suggest that these signals should be
+% rescaled. Use addScaleFactor() to add a MocoParameter to the problem that
+% will scale the reference data the muscles in the tracking cost.
```
chrisdembia, line 86:
```suggestion
% will scale the reference data for the muscles in the tracking cost.
```

aymanhab (changes requested):
> Reviewed 5 of 25 files at r1.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004)* status: 5 of 25 files reviewed, 2 unresolved discussions (waiting on @carmichaelong and @nickbianco)

___
*[Bindings/moco.i, line 25 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Mb0KXdt3fUkRbcTQIP3:-Mb0KXdt3fUkRbcTQIP4:b-4nrime) ([raw file](https://github.com/opensim-org/opensim-core/blob/594c63864ef17449f4b651c5caae96c774f78e78/Bindings/moco.i#L25)):*
> ```SWIG
> %include <OpenSim/Moco/MocoGoal/MocoMarkerTrackingGoal.h>
> %include <OpenSim/Moco/MocoGoal/MocoMarkerFinalGoal.h>
> %include <OpenSim/Moco/MocoGoal/MocoControlGoal.h>
> ```

Reason for removal? Does that break backward compatibility?

aymanhab (changes requested):
> Reviewed 12 of 25 files at r1.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004)* status: 17 of 25 files reviewed, 7 unresolved discussions (waiting on @carmichaelong and @nickbianco)

___
*[Bindings/Java/swig/java_moco.i, line 333 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-MbbXCyXE_38_6KEaPm0:-MbbXCyXE_38_6KEaPm1:b-8ajat1) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/Bindings/Java/swig/java_moco.i#L333)):*
> ```SWIG
>             addScaleFactor(name, control, this.convertArrayToMB(b));
>     }
> %}
> ```

Why do these methods need to be added manually? I worry about differences in documentation to the scripting interface from Doxygen/C++
___
*[OpenSim/Moco/MocoBounds.h, line 106 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-MbbXjlx5SkvEfNQjvhC:-MbbXjlx5SkvEfNQjvhD:bdpedmr) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/OpenSim/Moco/MocoBounds.h#L106)):*
> ```objc
>     friend MocoScaleFactor;
>     friend MocoControlTrackingGoal;
> ```

Any fundamental reason we increase "friends"? Can we do without? Sorry if it sounds anti-social ;)
___
*[OpenSim/Moco/MocoScaleFactor.h, line 35 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-MbbYQIr24wiI6mWaGcF:-MbbYQIs7617OVXWhf_7:b3deuaf) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/OpenSim/Moco/MocoScaleFactor.h#L35)):*
> ```objc
> but rather use the 'addScaleFactor()' interface provided by MocoGoals that
> support scale factors. How the scale factor is utilized in the cost function is
> specific to each MocoGoal. */
> ```

"optimized scalar value" is not clear, optimized by who? specified by who? Maybe a short example to explain usage?
___
*[OpenSim/Moco/MocoScaleFactor.h, line 48 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-MbbZQ-O8H5ftaexkpQi:-MbbZQ-O8H5ftaexkpQj:b-3n26gn) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/OpenSim/Moco/MocoScaleFactor.h#L48)):*
> ```objc
>     /// @details Note: the return value is constructed fresh on every call from
>     /// the internal property. Avoid repeated calls to this function.
>     MocoBounds getBounds() const { return MocoBounds(getProperty_bounds()); }
> ```

This is likely called many many times, any reason why not cache and return reference instead? Maybe later as an optimization?
___
*[OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.h, line 105 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Mbb_Nwk34vQ6-mTZ-Lk:-Mbb_Nwk34vQ6-mTZ-Ll:b3rvjg5) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.h#L105)):*
> ```objc
> Use `addScaleFactor()` to add a MocoParameter to the MocoParameter that will
> scale the tracking reference data associated with a control in the tracking cost.
> Scale factors for this goal can be useful if the magnitude of the tracking
> ```

The description here and the generic name "ScaleFactor" are a bit confusing, can we make the name more specific or the description more generic to be in sync. Is this specific to Tracking? Control? Reference data?

aymanhab:
@nickbianco Sorry I had to merge master into this branch since I had many conflicts with wrapping from master. Hopefully that's ok.

aymanhab (changes requested):
> Reviewed 4 of 25 files at r1.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004)* status: 21 of 25 files reviewed, 8 unresolved discussions (waiting on @carmichaelong and @nickbianco)

___
*[OpenSim/Moco/MocoGoal/MocoGoal.h, line 72 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Mbbex8eDat0j18i6axw:-Mbbex8eDat0j18i6axx:b-uc6tfu) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/OpenSim/Moco/MocoGoal/MocoGoal.h#L72)):*
> ```objc
> ## Scale factors
> Goals may include an option to add scale factors to the MocoProblem using
> `appendScaleFactor()`, which takes a MocoScaleFactor object for its argument.
> ```

"may include", how does a user know? can multiple ScaleFactors be appended? the name "append" suggests that? I would think every goal has an implicit scale factor of 1, and some goals allow users to change/modify this factor, but maybe I'm missing the picture here.

nickbianco (commented):
@aymanhab no worries, that's fine with me. This is ready for review again.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-:-Mbcs_Eq6g_zvLvKj3YM:b848dcl)* status: 17 of 25 files reviewed, 8 unresolved discussions (waiting on @aymanhab, @carmichaelong, and @chrisdembia)

___
*[Bindings/moco.i, line 25 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Mb0KXdt3fUkRbcTQIP3:-MbclPFb7xrdMlFIWsjd:b-ps7p54) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/Bindings/moco.i#L25)):*

Reason for removal? Does that break backward compatibility?
</blockquote></details>

Whoops, that was unintentional. Added it back in.
___
*[Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking_answers.m, line 86 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-MakDAQYM_sKQt2UucMi-r1-86:-Mbbnr-_4FoZTs_1hqD5:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/ee998e602963256270d0e95f63a627d35813b932/Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking_answers.m#L86)):*

```suggestion
% will scale the reference data for the muscles in the tracking cost.
```
</blockquote></details>

Done.
___
*[Bindings/Java/swig/java_moco.i, line 333 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-MbbXCyXE_38_6KEaPm0:-MbcsB4iEQ5-vKexVSQn:bcer32f) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/Bindings/Java/swig/java_moco.i#L333)):*

Why do these methods need to be added manually? I worry about differences in documentation to the scripting interface from Doxygen/C++
</blockquote></details>

SWIG doesn't know how to handle MocoBounds, so we need this custom implementation. We have to do the same thing for MocoProblem's setStateInfo and setControlInfo, and the interface difference between scripting and C++ is manageable.
___
*[OpenSim/Moco/MocoBounds.h, line 106 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-MbbXjlx5SkvEfNQjvhC:-MbcmOEfDHCldF2yIWPW:bcv7uwa) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/OpenSim/Moco/MocoBounds.h#L106)):*

Any fundamental reason we increase "friends"? Can we do without? Sorry if it sounds anti-social ;)
</blockquote></details>

We don't need MocoControlTrackingGoal as a friend, that was leftover from earlier development. But MocoScaleFactor needs to be a friend to access the special constructor that creates a MocoBounds object from a 2-element list property. 
___
*[OpenSim/Moco/MocoScaleFactor.h, line 35 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-MbbYQIr24wiI6mWaGcF:-MbcrTuC8A_JG3B8hQ_8:b-5c2284) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/OpenSim/Moco/MocoScaleFactor.h#L35)):*

"optimized scalar value" is not clear, optimized by who? specified by who? Maybe a short example to explain usage?
</blockquote></details>

The parameter is optimized using the MocoParameter interface, which needs a model property value to work. I tried to explain this in this paragraph so I'm not exactly what is unclear here. An example doesn't make sense here because this component doesn't do anything on its own, it's just a container for the parameter value. 
___
*[OpenSim/Moco/MocoScaleFactor.h, line 48 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-MbbZQ-O8H5ftaexkpQi:-Mbcm1AbEIRLBbBsgHcr:b85g3fe) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/OpenSim/Moco/MocoScaleFactor.h#L48)):*

This is likely called many many times, any reason why not cache and return reference instead? Maybe later as an optimization?
</blockquote></details>

This is only called during `MocoProblemRep::initialize()`, which is only called once when each problem is solved, so I don't think this will be an issue. 
___
*[OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.h, line 105 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Mbb_Nwk34vQ6-mTZ-Lk:-MbcqEe25m57rWx4Xevc:bexlx8u) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.h#L105)):*

The description here and the generic name "ScaleFactor" are a bit confusing, can we make the name more specific or the description more generic to be in sync. Is this specific to Tracking? Control? Reference data?
</blockquote></details>

Do you mean something like `addTrackingReferenceScaleFactor()`? That perhaps would be clearer. I was thinking that a `MocoGoal` that supports scale factors would use `addScaleFactor()` for the method name, but since this isn't a virtual method or anything that doesn't necessarily need to be the case.
___
*[OpenSim/Moco/MocoGoal/MocoGoal.h, line 72 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Mbbex8eDat0j18i6axw:-MbcoSLT1Se9SSTDpqC5:b602fle) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/OpenSim/Moco/MocoGoal/MocoGoal.h#L72)):*

"may include", how does a user know? can multiple ScaleFactors be appended? the name "append" suggests that? I would think every goal has an implicit scale factor of 1, and some goals allow users to change/modify this factor, but maybe I'm missing the picture here.
</blockquote></details>

A `MocoGoal` will include the method `addScaleFactor` which should call the protected method `appendScaleFactor`. If a `MocoGoal` doesn't include this method, then the user knows that scale factors are not supported for that goal. Goals do not implicitly have any scale factors, but do contain a (by default, empty) list scale factors that can be appended to by `appendScaleFactor`. As I state in the docs for `MocoScaleFactor`, how each scale factor is used is specific to each MocoGoal.

aymanhab (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004)* status: 17 of 25 files reviewed, 8 unresolved discussions (waiting on @aymanhab, @carmichaelong, @chrisdembia, and @nickbianco)

___
*[Bindings/Java/swig/java_moco.i, line 316 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-MbgGoiz01fpBH2dKE9m:-MbgGoiz01fpBH2dKE9n:bhotafs) ([raw file](https://github.com/opensim-org/opensim-core/blob/2cea0874d2794caad8a91c7d57a5f8eedcce188b/Bindings/Java/swig/java_moco.i#L316)):*
> ```SWIG
> %typemap(javacode) OpenSim::MocoControlTrackingGoal %{
>     public static MocoBounds convertArrayToMB(double[] arr) throws Exception {
>             MocoBounds bounds = new MocoBounds();
> ```

Is it common to shorten MocoBounds to MB? Also in terms of memory management when an object is new'ed it may get garbage collected at a random time later. Would be good to consider if this can cause problems downstream.

aymanhab (changes requested):
> Reviewed 3 of 25 files at r1, 1 of 1 files at r2, 4 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004)* status: all files reviewed, 8 unresolved discussions (waiting on @aymanhab, @carmichaelong, @chrisdembia, and @nickbianco)

___
*[OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.h, line 105 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Mbb_Nwk34vQ6-mTZ-Lk:-MbggbFn1SeWDVjlxOrB:b-gq49op) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.h#L105)):*

Do you mean something like `addTrackingReferenceScaleFactor()`? That perhaps would be clearer. I was thinking that a `MocoGoal` that supports scale factors would use `addScaleFactor()` for the method name, but since this isn't a virtual method or anything that doesn't necessarily need to be the case.
</blockquote></details>

Yes, addTrackingReferenceScaleFactor would be more specific and clearer, I don't see the harm in having virtual method but you've looked into these more than me. Think if there's another Goal that uses a ScaleFactor, can the code be reused? My issue with generic ScaleFactor is that we have a few places where we use it as in Geometry/Mesh, ScaleTool, etc.
___
*[OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.cpp, line 186 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-MbgiRThCIyIprYy4-hu:-MbgiRThCIyIprYy4-hv:bk8hpgd) ([raw file](https://github.com/opensim-org/opensim-core/blob/2cea0874d2794caad8a91c7d57a5f8eedcce188b/OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.cpp#L186)):*
> ```cpp
>     }
>     setRequirements(1, 1, SimTK::Stage::Time);
> }
> ```

Is this change related, or a separate bugfix?

aymanhab (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004)* status: all files reviewed, 6 unresolved discussions (waiting on @aymanhab, @carmichaelong, @chrisdembia, and @nickbianco)

___
*[OpenSim/Moco/MocoScaleFactor.h, line 35 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-MbbYQIr24wiI6mWaGcF:-Mbgj7trA8EC33lOSrcT:bwlefi9) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/OpenSim/Moco/MocoScaleFactor.h#L35)):*

The parameter is optimized using the MocoParameter interface, which needs a model property value to work. I tried to explain this in this paragraph so I'm not exactly what is unclear here. An example doesn't make sense here because this component doesn't do anything on its own, it's just a container for the parameter value. 
</blockquote></details>

So (a scalar value to be optimized by Moco)?

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004)* status: 22 of 25 files reviewed, 6 unresolved discussions (waiting on @aymanhab, @carmichaelong, @chrisdembia, and @nickbianco)

___
*[Bindings/Java/swig/java_moco.i, line 316 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-MbgGoiz01fpBH2dKE9m:-MbgwvQi0XwtokckVNez:boeab3a) ([raw file](https://github.com/opensim-org/opensim-core/blob/2cea0874d2794caad8a91c7d57a5f8eedcce188b/Bindings/Java/swig/java_moco.i#L316)):*

Is it common to shorten MocoBounds to MB? Also in terms of memory management when an object is new'ed it may get garbage collected at a random time later. Would be good to consider if this can cause problems downstream.
</blockquote></details>

Not in the exposed API, but we shorted MocoBounds to MB in this file since these functions are only used here. We haven't noticed any garbage collection issues with this so far, but I'll keep it in mind. 
___
*[OpenSim/Moco/MocoScaleFactor.h, line 35 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-MbbYQIr24wiI6mWaGcF:-Mbh1RVF3RpXuNBM77sQ:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/OpenSim/Moco/MocoScaleFactor.h#L35)):*

So (a scalar value to be optimized by Moco)?
</blockquote></details>

Done.
___
*[OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.cpp, line 186 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-MbgiRThCIyIprYy4-hu:-MbgqXha1cPvlDh4_4Eu:b-ev620w) ([raw file](https://github.com/opensim-org/opensim-core/blob/2cea0874d2794caad8a91c7d57a5f8eedcce188b/OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.cpp#L186)):*

Is this change related, or a separate bugfix?
</blockquote></details>

This change is needed so that the model property values are updated, which doesn't happen until this stage.

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004)* status: 22 of 25 files reviewed, 6 unresolved discussions (waiting on @aymanhab, @carmichaelong, and @chrisdembia)

___
*[OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.h, line 105 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Mbb_Nwk34vQ6-mTZ-Lk:-Mbh1g45DAxtd-UjmfzI:bbos51v) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.h#L105)):*

Yes, addTrackingReferenceScaleFactor would be more specific and clearer, I don't see the harm in having virtual method but you've looked into these more than me. Think if there's another Goal that uses a ScaleFactor, can the code be reused? My issue with generic ScaleFactor is that we have a few places where we use it as in Geometry/Mesh, ScaleTool, etc.
</blockquote></details>

The more I think about this, I don't think `addTrackingReferenceScaleFactor` is the correct solution. Scale factors scale the reference data, but only for a specific control. This would suggest something like `addScaleFactorForControl` (similar to `setWeightForControl`), but users might read that like the control itself is being scaled. I've now started to come back around to the simpler `addScaleFactor`. In addition, I've improved the documentation to make it clear what each scale factor does for this goal.

nickbianco (commented):
Ready for review.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-:-Mbh3--h2dje2heiXfba:b8pt725)* status: 22 of 25 files reviewed, 6 unresolved discussions (waiting on @aymanhab, @carmichaelong, and @chrisdembia)

aymanhab (changes requested):
> Reviewed 3 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004)* status: 24 of 25 files reviewed, 3 unresolved discussions (waiting on @aymanhab, @carmichaelong, @chrisdembia, and @nickbianco)

___
*[OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.h, line 105 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Mbb_Nwk34vQ6-mTZ-Lk:-MblxcZ4Btu06eAdTFzl:bpuzauz) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.h#L105)):*

The more I think about this, I don't think `addTrackingReferenceScaleFactor` is the correct solution. Scale factors scale the reference data, but only for a specific control. This would suggest something like `addScaleFactorForControl` (similar to `setWeightForControl`), but users might read that like the control itself is being scaled. I've now started to come back around to the simpler `addScaleFactor`. In addition, I've improved the documentation to make it clear what each scale factor does for this goal.
</blockquote></details>

I'm fine with leaving as is if the context/class-name disambiguates the name scope. @nickbianco is going to check if serialization works and if it justifies a change.
___
*[OpenSim/Moco/MocoGoal/MocoGoal.h, line 72 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Mbbex8eDat0j18i6axw:-MblyEUYCoQV1n-gdulZ:bn1h4xx) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/OpenSim/Moco/MocoGoal/MocoGoal.h#L72)):*

A `MocoGoal` will include the method `addScaleFactor` which should call the protected method `appendScaleFactor`. If a `MocoGoal` doesn't include this method, then the user knows that scale factors are not supported for that goal. Goals do not implicitly have any scale factors, but do contain a (by default, empty) list scale factors that can be appended to by `appendScaleFactor`. As I state in the docs for `MocoScaleFactor`, how each scale factor is used is specific to each MocoGoal. 
</blockquote></details>

Can these scale factors be chained as implied by the "append" naming convention?

nickbianco (commented):
Thanks @carmichaelong and @aymanhab. The list of scale factors belonging to a MocoGoal are now contained in a list property of MocoGoal to support serialization.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-:-McFvlxjBRKIk6etGxpD:b-ezxqsr)* status: 21 of 26 files reviewed, 4 unresolved discussions (waiting on @aymanhab, @carmichaelong, and @chrisdembia)

___
*[Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking_answers.m, line 65 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-MbvvR_d3LAMkhsDIvx3:-McFvHPbEt8IKA0eRqsW:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/14b2a3c8a222d6cb899fac64a5dd28c5d19f064d/Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking_answers.m#L65)):*

acess -> access
</blockquote></details>

Done.
___
*[OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.h, line 105 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Mbb_Nwk34vQ6-mTZ-Lk:-McFvLUI9HS2bNxKYtoc:b-gca8y4) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.h#L105)):*

I'm fine with leaving as is if the context/class-name disambiguates the name scope. @nickbianco is going to check if serialization works and if it justifies a change.
</blockquote></details>

Tested serialization locally and it works fine. 
___
*[OpenSim/Moco/MocoGoal/MocoGoal.h, line 72 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Mbbex8eDat0j18i6axw:-McFvU8a7vgrl0qLmVuI:b-e4axha) ([raw file](https://github.com/opensim-org/opensim-core/blob/218431be4c5e977ce2a8db377b69360008afd5da/OpenSim/Moco/MocoGoal/MocoGoal.h#L72)):*

Can these scale factors be chained as implied by the "append" naming convention?
</blockquote></details>

Not sure what you mean by "chained". The MocoGoal contains a list (now a list property) of all MocoScaleFactors added to the goal.

nickbianco (commented):
@carmichaelong, @aymanhab  Finished the fill-in-the-blanks file, ready for review.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-:-McLCkQoB_3BGCGcBIzz:b1lpkun)* status: 21 of 28 files reviewed, 4 unresolved discussions (waiting on @aymanhab, @carmichaelong, and @chrisdembia)

carmichaelong (approved):
LGTM pending resolving @aymanhab's comments

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-:-McMMI7ICzo8gI5Z3Ov4:b9rvgko)* status: 21 of 28 files reviewed, 3 unresolved discussions (waiting on @aymanhab and @chrisdembia)

aymanhab (approved):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004)* status: 21 of 28 files reviewed, 1 unresolved discussion (waiting on @aymanhab and @chrisdembia)

aymanhab (approved):
> Reviewed 1 of 7 files at r5.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004)* status: 22 of 28 files reviewed, 1 unresolved discussion (waiting on @aymanhab and @chrisdembia)

nickbianco (commented):
> Dismissed @chrisdembia from a discussion.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004)* status: 21 of 28 files reviewed, all discussions resolved (waiting on @aymanhab and @chrisdembia)

nickbianco (commented):
@aymanhab @carmichaelong Python versions created. Ready for review.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-:-Mczk8mx0fF3Pp3zQm-b:b-zfk1oq)* status: 19 of 33 files reviewed, all discussions resolved (waiting on @aymanhab and @chrisdembia)

aymanhab (changes requested):
No changelog entry

> Reviewed 5 of 7 files at r5, 9 of 9 files at r6.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-:-Md2xzXS32TP95XlxBm4:b-91ss82)* status: all files reviewed, 5 unresolved discussions (waiting on @nickbianco)

___
*[Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking.m, line 42 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Md2zQhj3MCB0d6UMC_1:-Md2zQhj3MCB0d6UMC_2:bct601e) ([raw file](https://github.com/opensim-org/opensim-core/blob/8344d6f7b9219509d1b30045642a0fe281ff1af4/Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking.m#L42)):*
> ```objc
> % Load the experimental electromyography data and compare 
> % the effort minimization solution against this data. We will also and use 
> % it later for the EMG-tracking problem. Each column in emg.sto is 
> ```

"will also and use" -> "will also use"?
___
*[Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking.m, line 126 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Md3-FI88A8HPJAEDQhe:-Md3-FI88A8HPJAEDQhf:b-ylxx8v) ([raw file](https://github.com/opensim-org/opensim-core/blob/8344d6f7b9219509d1b30045642a0fe281ff1af4/Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking.m#L126)):*
> ```objc
> import org.opensim.modeling.*;
> 
> coordSet = model.updCoordinateSet();
> ```

If not modifying coordSet why upd instead of get? If intentional for side-effect we probably should document it here
___
*[Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking_answers.m, line 37 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Md30P8x06N2e_hDsSZM:-Md30P8x06N2e_hDsSZN:b-9zlgnp) ([raw file](https://github.com/opensim-org/opensim-core/blob/8344d6f7b9219509d1b30045642a0fe281ff1af4/Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking_answers.m#L37)):*
> ```objc
> inverse.set_initial_time(0.83);
> inverse.set_final_time(2.0);
> inverse.set_mesh_interval(0.04);
> ```

Any reason the helpful comments were removed? were they out of date?
___
*[Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking_answers.m, line 43 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Md30YYe7Ps2flPmf0ce:-Md30YYe7Ps2flPmf0cf:b-uaqnp8) ([raw file](https://github.com/opensim-org/opensim-core/blob/8344d6f7b9219509d1b30045642a0fe281ff1af4/Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking_answers.m#L43)):*
> ```objc
> if ~exist('effortSolution.sto', 'file')
> % Part 1f: Solve the problem!
> inverseSolution = inverse.solve();
> ```

Indentation
___
*[OpenSim/Moco/MocoGoal/MocoGoal.h, line 306 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Md31s4J5vGglkqYrz7Z:-Md31s4J5vGglkqYrz7_:b-7k4i9f) ([raw file](https://github.com/opensim-org/opensim-core/blob/8344d6f7b9219509d1b30045642a0fe281ff1af4/OpenSim/Moco/MocoGoal/MocoGoal.h#L306)):*
> ```objc
>         }
>         return scaleFactors;
>     }
> ```

What was the rationale for copying rather than returning a reference here? Memory management/ownership downstream?

carmichaelong:
On Python 3.8 with Numpy 1.20.2 and matplotlib 3.3.4, python example worked for me locally (using PYTHONPATH)

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004)* status: 28 of 34 files reviewed, 5 unresolved discussions (waiting on @aymanhab and @chrisdembia)

___
*[Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking.m, line 42 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Md2zQhj3MCB0d6UMC_1:-MdKjBGD37yZUgdMRU85:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/8344d6f7b9219509d1b30045642a0fe281ff1af4/Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking.m#L42)):*

"will also and use" -> "will also use"?
</blockquote></details>

Done.
___
*[Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking.m, line 126 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Md3-FI88A8HPJAEDQhe:-MdKjYTO3vqBLe0O3ivY:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/8344d6f7b9219509d1b30045642a0fe281ff1af4/Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking.m#L126)):*

If not modifying coordSet why upd instead of get? If intentional for side-effect we probably should document it here
</blockquote></details>

Done.
___
*[Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking_answers.m, line 37 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Md30P8x06N2e_hDsSZM:-MdKjsKq7yVtTwwj6YL1:b46s39v) ([raw file](https://github.com/opensim-org/opensim-core/blob/8344d6f7b9219509d1b30045642a0fe281ff1af4/Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking_answers.m#L37)):*

Any reason the helpful comments were removed? were they out of date?
</blockquote></details>

I removed them to keep the answers file a little less cluttered. I can provide some context for the settings in the slides. Those comments wouldn't have made sense in the fill-in-the-blank file either.
___
*[Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking_answers.m, line 43 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Md30YYe7Ps2flPmf0ce:-MdKlXEQCLDEuIQlaNbu:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/8344d6f7b9219509d1b30045642a0fe281ff1af4/Bindings/Java/Matlab/examples/Moco/exampleEMGTracking/exampleEMGTracking_answers.m#L43)):*

Indentation
</blockquote></details>

Done.
___
*[OpenSim/Moco/MocoGoal/MocoGoal.h, line 306 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-Md31s4J5vGglkqYrz7Z:-MdKkZuxE0GMe0uCnXiZ:b-o8olzv) ([raw file](https://github.com/opensim-org/opensim-core/blob/8344d6f7b9219509d1b30045642a0fe281ff1af4/OpenSim/Moco/MocoGoal/MocoGoal.h#L306)):*

What was the rationale for copying rather than returning a reference here? Memory management/ownership downstream?
</blockquote></details>

The scale factors are stored as a list property of the goal, so a new vector must be constructed when calling this method. It only gets called once when constructing a MocoProblemRep.

aymanhab (approved):
> Reviewed 1 of 6 files at r7.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004)* status: 29 of 34 files reviewed, all discussions resolved (waiting on @aymanhab)

aymanhab (approved):
> Reviewed 3 of 6 files at r7, 2 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nickbianco)

carmichaelong (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3004#-:-MdTAb0M18SB96aeuBPM:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nickbianco)

nickbianco:
Thanks @aymanhab, @carmichaelong!

PR #3013, opened 2021-06-15
Use python 3.8 and numpy 1.20.3 for ci

aymanhab:
Fixes issue #0

### Brief summary of changes
Force ci to use python 3.8 and numpy 1.20.3 in sync with latest released anaconda

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013)

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -102,7 +102,7 @@ jobs:
     - name: Test Python bindings
       run: |
         $env:PATH += ";$env:USERPROFILE/opensim-core-install/bin"
-        $env:PATH
+        $env:PYTHONPATH = "$env:USERPROFILE/opensim-core-install/bin"
         # Move to the installed location of the python package.
         cd ~/opensim-core-install/sdk/python
         # Run python tests.
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -92,7 +92,7 @@ jobs:
     - name: Test opensim-core
       run: |
         chdir $env:GITHUB_WORKSPACE\\build
-        ctest --parallel 4 --output-on-failure --build-config Release
+        ctest --parallel 4 --output-on-failure --build-config Release -E python.*
 
     - name: Install opensim-core
       # TODO: This is where we wish to do the installing, but it's done above for now.
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -104,7 +104,7 @@ jobs:
 
     - name: Test Python bindings
       run: |
-        echo "PYTHONPATH= $env:USERPROFILE/opensim-core-install/bin">> $GITHUB_ENV
+        # echo "PYTHONPATH= $env:USERPROFILE/opensim-core-install/bin">> $GITHUB_ENV
         # Move to the installed location of the python package.
         cd ~/opensim-core-install/sdk/python
         # Run python tests.
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -102,7 +102,7 @@ jobs:
     - name: Test Python bindings
       run: |
         $env:PATH += ";$env:USERPROFILE/opensim-core-install/bin"
-        $env:PYTHONPATH = "$env:USERPROFILE/opensim-core-install/bin"
+        echo "PYTHONPATH=$env:USERPROFILE/opensim-core-install/bin" >> $GITHUB_ENV
         # Move to the installed location of the python package.
         cd ~/opensim-core-install/sdk/python
         # Run python tests.
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -102,7 +102,6 @@ jobs:
     - name: Test Python bindings
       run: |
         $env:PATH += ";$env:USERPROFILE/opensim-core-install/bin"
-        echo "PYTHONPATH=$env:USERPROFILE/opensim-core-install/bin" >> $GITHUB_ENV
         # Move to the installed location of the python package.
         cd ~/opensim-core-install/sdk/python
         # Run python tests.
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -107,6 +107,7 @@ jobs:
         echo "PYTHONPATH= $env:USERPROFILE/opensim-core-install/bin">> $GITHUB_ENV
         # Move to the installed location of the python package.
         cd ~/opensim-core-install/sdk/python
+        python setup_win_python38.py
         # Run python tests.
         python -m unittest discover --start-directory opensim/tests --verbose
 
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -40,7 +40,7 @@ jobs:
       run: python -m pip install numpy==1.20.3
 
     - name: Install SWIG
-      run: choco install swig --version 4.0.0 --yes --limit-output
+      run: choco install swig --version 4.0.2 --yes --limit-output
 
     - name: Cache dependencies
       id: cache-dependencies
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -39,9 +39,6 @@ jobs:
       #Need numpy to use SWIG numpy typemaps.
       run: python -m pip install numpy==1.20.3
 
-    - name: Install SWIG
-      run: choco install swig --version 4.0.2 --yes --limit-output
-
     - name: Cache dependencies
       id: cache-dependencies
       uses: actions/cache@v1
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -104,7 +104,7 @@ jobs:
 
     - name: Test Python bindings
       run: |
-        # echo "PYTHONPATH= $env:USERPROFILE/opensim-core-install/bin">> $GITHUB_ENV
+        echo "PYTHONPATH= $env:USERPROFILE/opensim-core-install/bin">> $GITHUB_ENV
         # Move to the installed location of the python package.
         cd ~/opensim-core-install/sdk/python
         # Run python tests.
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -33,11 +33,11 @@ jobs:
     - name: Install Python packages
       uses: actions/setup-python@v2
       with:
-        python-version: '3.7'
+        python-version: '3.8'
         
     - name: Install numpy
       #Need numpy to use SWIG numpy typemaps.
-      run: python -m pip install numpy==1.19.3
+      run: python -m pip install numpy==1.20.3
 
     - name: Install SWIG
       run: choco install swig --version 4.0.0 --yes --limit-output
@@ -162,9 +162,9 @@ jobs:
     - name: Install Homebrew packages
       # Save the gfortran version to a file so we can use it in the cache key.
       run: |
-        brew install cmake pkgconfig autoconf libtool automake wget pcre doxygen python@3.7
+        brew install cmake pkgconfig autoconf libtool automake wget pcre doxygen python@3.8
         brew reinstall gcc
-        pip3 install numpy==1.19.3
+        pip3 install numpy==1.20.3
         gfortran -v
         mkdir gfortran_version
         gfortran -v &> gfortran_version/gfortran_version.txt
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -37,7 +37,7 @@ jobs:
         
     - name: Install numpy
       #Need numpy to use SWIG numpy typemaps.
-      run: python -m pip install numpy==1.20.3
+      run: python -m pip install numpy==1.20.2
 
     - name: Install SWIG
       run: choco install swig --version 4.0.0 --yes --limit-output --allow-downgrade
@@ -163,7 +163,7 @@ jobs:
       run: |
         brew install cmake pkgconfig autoconf libtool automake wget pcre doxygen python@3.8
         brew reinstall gcc
-        pip3 install numpy==1.20.3
+        pip3 install numpy==1.20.2
         gfortran -v
         mkdir gfortran_version
         gfortran -v &> gfortran_version/gfortran_version.txt
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -104,7 +104,7 @@ jobs:
 
     - name: Test Python bindings
       run: |
-        $env:PATH += ";$env:USERPROFILE/opensim-core-install/bin"
+        echo "PYTHONPATH= $env:USERPROFILE/opensim-core-install/bin">> $GITHUB_ENV
         # Move to the installed location of the python package.
         cd ~/opensim-core-install/sdk/python
         # Run python tests.
--- a/Bindings/Python/CMakeLists.txt
+++ b/Bindings/Python/CMakeLists.txt
@@ -254,12 +254,9 @@ OpenSimPutFileInPythonPackage("${CMAKE_CURRENT_SOURCE_DIR}/setup.py" ".")
 OpenSimPutFileInPythonPackage("${CMAKE_CURRENT_SOURCE_DIR}/__init__.py" opensim)
 
 OpenSimPutFileInPythonPackage("${CMAKE_CURRENT_SOURCE_DIR}/report.py" opensim)
-# for Python 3.8+/Windows we create a file dll_location.txt and use
-# it in setup_win_python38.py to set additional dll directory
+# for Python 3.8+/Windows we use a special script
+# setup_win_python38.py to set additional dll directory in __init__.py on client machine
 if (WIN32)
-    set(DLL_PATH_FILE ${CMAKE_CURRENT_BINARY_DIR}/dll_location.txt)
-    file(WRITE ${DLL_PATH_FILE} "${CMAKE_INSTALL_PREFIX}/bin")
-    OpenSimPutFileInPythonPackage("${DLL_PATH_FILE}" opensim)
     OpenSimPutFileInPythonPackage("${CMAKE_CURRENT_SOURCE_DIR}/setup_win_python38.py" ".")
 endif()
 
--- a/Bindings/Python/CMakeLists.txt
+++ b/Bindings/Python/CMakeLists.txt
@@ -255,11 +255,12 @@ OpenSimPutFileInPythonPackage("${CMAKE_CURRENT_SOURCE_DIR}/__init__.py" opensim)
 
 OpenSimPutFileInPythonPackage("${CMAKE_CURRENT_SOURCE_DIR}/report.py" opensim)
 # for Python 3.8+/Windows we create a file dll_location.txt and use
-# it in __init__.py to set additional dll directory
+# it in setup_win_python38.py to set additional dll directory
 if (WIN32)
     set(DLL_PATH_FILE ${CMAKE_CURRENT_BINARY_DIR}/dll_location.txt)
     file(WRITE ${DLL_PATH_FILE} "${CMAKE_INSTALL_PREFIX}/bin")
     OpenSimPutFileInPythonPackage("${DLL_PATH_FILE}" opensim)
+    OpenSimPutFileInPythonPackage("${CMAKE_CURRENT_SOURCE_DIR}/setup_win_python38.py" ".")
 endif()
 
 # Test files. If you require more test resource files, list them here.
--- a/Bindings/Python/CMakeLists.txt
+++ b/Bindings/Python/CMakeLists.txt
@@ -254,6 +254,13 @@ OpenSimPutFileInPythonPackage("${CMAKE_CURRENT_SOURCE_DIR}/setup.py" ".")
 OpenSimPutFileInPythonPackage("${CMAKE_CURRENT_SOURCE_DIR}/__init__.py" opensim)
 
 OpenSimPutFileInPythonPackage("${CMAKE_CURRENT_SOURCE_DIR}/report.py" opensim)
+# for Python 3.8+/Windows we create a file dll_location.txt and use
+# it in __init__.py to set additional dll directory
+if (WIN32)
+    set(DLL_PATH_FILE ${CMAKE_CURRENT_BINARY_DIR}/dll_location.txt)
+    file(WRITE ${DLL_PATH_FILE} "${CMAKE_INSTALL_PREFIX}/bin")
+    OpenSimPutFileInPythonPackage("${CMAKE_CURRENT_BINARY_DIR}/dll_location.txt" opensim)
+endif()
 
 # Test files. If you require more test resource files, list them here.
 foreach(test_file
--- a/Bindings/Python/CMakeLists.txt
+++ b/Bindings/Python/CMakeLists.txt
@@ -259,7 +259,7 @@ OpenSimPutFileInPythonPackage("${CMAKE_CURRENT_SOURCE_DIR}/report.py" opensim)
 if (WIN32)
     set(DLL_PATH_FILE ${CMAKE_CURRENT_BINARY_DIR}/dll_location.txt)
     file(WRITE ${DLL_PATH_FILE} "${CMAKE_INSTALL_PREFIX}/bin")
-    OpenSimPutFileInPythonPackage("${CMAKE_CURRENT_BINARY_DIR}/dll_location.txt" opensim)
+    OpenSimPutFileInPythonPackage("${DLL_PATH_FILE}" opensim)
 endif()
 
 # Test files. If you require more test resource files, list them here.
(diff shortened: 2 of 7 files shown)


aymanhab:
This PR is ready for review. First ci build with python 3.8 on windows. Mac failure is sporadic and unrelated.

aymanhab:
@chrisdembia Can you point to any resources regarding customizing the execution of setup.py to inject code there  and avoid having users modify the python installation process if on python 3.8+/windows? Thanks

carmichaelong:
Probably minor, but when I tried to install numpy 1.20.3 with conda (this PR uses pip install), it wasn't available. Looks like per [here](https://anaconda.org/anaconda/numpy) it might be that 1.20.2 is the current latest with conda unless I missed something?

nickbianco:
@aymanhab Regarding the setup.py question, the following can be used to find the absolute path to the bin folder: 
```
import os
os.path.abspath('../../bin')
```
From there it should be easy to update `dll_location.txt`. You can use the `platform` python package to only do this if the current OS is Windows.

nickbianco (commented):
Or you can use `sys.platform.startswith('win')` as you do in this PR already.

> Reviewed 2 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-:-Md-IAlnDzfmdEb4fbZD:bfds0it)* status: 2 of 4 files reviewed, all discussions resolved (waiting on @aymanhab)

nickbianco (changes requested):
> Reviewed 2 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013)* status: all files reviewed, 3 unresolved discussions (waiting on @aymanhab)

___
*[.github/workflows/continuous_integration.yml, line 107 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-Md-JHl8CaJEODWZJy7X:-Md-JHl8CaJEODWZJy7Y:b75tz8s) ([raw file](https://github.com/opensim-org/opensim-core/blob/21c2661540fca6e6f7476d628dd7ec21af742fbb/.github/workflows/continuous_integration.yml#L107)):*
> ```YAML
>     - name: Test Python bindings
>       run: |
>         echo "PYTHONPATH= $env:USERPROFILE/opensim-core-install/bin">> $GITHUB_ENV
> ```

This should necessary now, right?
___
*[Bindings/Python/__init__.py, line 6 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-Md-Ih00DtH4sfG9H25O:-Md-Ih00DtH4sfG9H25P:b-m02tq3) ([raw file](https://github.com/opensim-org/opensim-core/blob/21c2661540fca6e6f7476d628dd7ec21af742fbb/Bindings/Python/__init__.py#L6)):*
> ```Python
>     cwd = os.getcwd()
>     # print(cwd)
>     print('Writing new dll_location.txt conatining '+cwd)
> ```

You could combine the above three lines into:
`print('Writing new dll_location.txt conatining '+ os.getcwd())`
___
*[Bindings/Python/CMakeLists.txt, line 262 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-Md-J2Hr1v4b36ecU45H:-Md-J2Hr1v4b36ecU45I:b-t78a7j) ([raw file](https://github.com/opensim-org/opensim-core/blob/21c2661540fca6e6f7476d628dd7ec21af742fbb/Bindings/Python/CMakeLists.txt#L262)):*
> ```CMake
>     set(DLL_PATH_FILE ${CMAKE_CURRENT_BINARY_DIR}/dll_location.txt)
>     file(WRITE ${DLL_PATH_FILE} "${CMAKE_INSTALL_PREFIX}/bin")
>     OpenSimPutFileInPythonPackage("${CMAKE_CURRENT_BINARY_DIR}/dll_location.txt" opensim)
> ```

You could use the `DLL_PATH_FILE` CMake variable again here.

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @aymanhab and @nickbianco)

___
*[Bindings/Python/__init__.py, line 6 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-Md-Ih00DtH4sfG9H25O:-Md3BIuM6jjdLzNU1S35:b-qvgjd) ([raw file](https://github.com/opensim-org/opensim-core/blob/21c2661540fca6e6f7476d628dd7ec21af742fbb/Bindings/Python/__init__.py#L6)):*

You could combine the above three lines into:
`print('Writing new dll_location.txt conatining '+ os.getcwd())`
</blockquote></details>

These are debugging lines that will eventually be removed when the dust settles but agree they can be combined as you suggested. Thank you
___
*[Bindings/Python/CMakeLists.txt, line 262 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-Md-J2Hr1v4b36ecU45H:-Md3BcAS0tm79DDfEXjb:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/21c2661540fca6e6f7476d628dd7ec21af742fbb/Bindings/Python/CMakeLists.txt#L262)):*

You could use the `DLL_PATH_FILE` CMake variable again here.
</blockquote></details>

Done.

nickbianco (commented):
> Reviewed 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013)* status: all files reviewed, 1 unresolved discussion (waiting on @aymanhab)

aymanhab (commented):
Thanks @nickbianco I gave that a shot and will see how ci and the artifact behave before passing on to you and others.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-:-Md3tlZr4HnwEijSwgCB:b-nwu43h)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on @nickbianco)

___
*[.github/workflows/continuous_integration.yml, line 107 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-Md-JHl8CaJEODWZJy7X:-Md3tyOoEv8WW1MG1bWZ:b-b7zvsx) ([raw file](https://github.com/opensim-org/opensim-core/blob/21c2661540fca6e6f7476d628dd7ec21af742fbb/.github/workflows/continuous_integration.yml#L107)):*

This should necessary now, right?
</blockquote></details>

You mean "not necessary" :smile: ? I agree but will find out from ci

nickbianco (changes requested):
> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013)* status: all files reviewed, 3 unresolved discussions (waiting on @aymanhab)

___
*[.github/workflows/continuous_integration.yml, line 107 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-Md-JHl8CaJEODWZJy7X:-Md4C22B1xOcrJs3k3J0:bt4vad1) ([raw file](https://github.com/opensim-org/opensim-core/blob/21c2661540fca6e6f7476d628dd7ec21af742fbb/.github/workflows/continuous_integration.yml#L107)):*

You mean "not necessary" :smile: ? I agree but will find out from ci
</blockquote></details>

Yes, that is what I meant :) Sounds good.
___
*[Bindings/Python/setup.py, line 6 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-Md4Bofl9zossAMyC5LJ:-Md4Bofl9zossAMyC5LK:b-ac9nei) ([raw file](https://github.com/opensim-org/opensim-core/blob/581f1b32cb3fb28d78d6e7177923be3afe31ef9d/Bindings/Python/setup.py#L6)):*
> ```Python
> import sys
> 
> from setuptools import setup
> ```

This import is still needed for the install. You might have removed it by mistake.
___
*[Bindings/Python/setup.py, line 10 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-Md4BvDh5VHHm4YgTFzg:-Md4BvDh5VHHm4YgTFzh:bqrjzm4) ([raw file](https://github.com/opensim-org/opensim-core/blob/c6dfdb77678c31fba44ebba4d1fc21e9579ffb00/Bindings/Python/setup.py#L10)):*
> ```Python
> 	with open('opensim/dll_location.txt', 'w') as writer:
> 		writer.write(install_path)
> 		writer.write("\n")
> ```

Tabs --> spaces

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013)* status: 4 of 5 files reviewed, 3 unresolved discussions (waiting on @nickbianco)

___
*[.github/workflows/continuous_integration.yml, line 107 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-Md-JHl8CaJEODWZJy7X:-Md4EQzdBenbMAlWgpQB:b9hsdl0) ([raw file](https://github.com/opensim-org/opensim-core/blob/21c2661540fca6e6f7476d628dd7ec21af742fbb/.github/workflows/continuous_integration.yml#L107)):*

Yes, that is what I meant :) Sounds good.
</blockquote></details>

Passed on windows so I assume we're good here. Thank you
___
*[Bindings/Python/setup.py, line 6 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-Md4Bofl9zossAMyC5LJ:-Md4Ei6D5jFh2ayUHAQo:b-knl617) ([raw file](https://github.com/opensim-org/opensim-core/blob/581f1b32cb3fb28d78d6e7177923be3afe31ef9d/Bindings/Python/setup.py#L6)):*

This import is still needed for the install. You might have removed it by mistake.
</blockquote></details>

Thank you for the find, somehow it wasn't showing up as diff
___
*[Bindings/Python/setup.py, line 10 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-Md4BvDh5VHHm4YgTFzg:-Md4Er5r1ZF-roP0MVDe:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/c6dfdb77678c31fba44ebba4d1fc21e9579ffb00/Bindings/Python/setup.py#L10)):*

Tabs --> spaces
</blockquote></details>

Done.

nickbianco (commented):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aymanhab)

nickbianco (approved):
LGTM. After CI passes, we should test locally on Mac and Windows.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-:-Md4QYgB9TyJY0tH3HqB:bl51eq1)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aymanhab)

aymanhab:
@nickbianco and @carmichaelong My test of the artifact on windows works only if you invoke python, call setup.py from the sdk/Python folder, it fails if calling from another folder because the file dll_location.txt which is populated correctly is not copied inside the .egg file in the site-packages folder. Will try to find a workaround. Just FYI

aymanhab:
One option that work is to so `python setup.py develop` which doesn't copy into a .egg file but rather create a link which seems perfect but not sure about the downstream effects in terms of conda, virtualenv etc.

nickbianco (commented):
I think it's fine if we require users to run the install script from the `sdk/Python` folder. I don't really know the differences between `install` and `develop`.

> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-:-MdJRUEN6NXPWQ-WPF4L:bxzider)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aymanhab)

aymanhab:
Looks like pip install works fine if we use version number consistent with PEP 440. Testing now

nickbianco (commented):
> Reviewed 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aymanhab)

aymanhab:
This is done and tested locally with python 3.8, the process is:
- change to the sdk/Python folder under the installation
- execute the script `python setup_win_python38.py` [optional, only needed if using python 3.8+ on windows]
- execute the line `python -m pip install .`
- From then on you can go to arbitrary directory launch python and opensim works exactly as expected. No need for PYTHONPATH or PATH setting change.

Would be good to have somebody else try it out, neither OSX or Linux should see any change, behavior pre 3.8 shouldn't change at all.

nickbianco (commented):
Is it possible to integrate `setup_win_python38.py` into `setup.py` and only run if the platform is Windows? It would be nice if the install instructions are the same across platforms.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-:-MdSvlbkDNohsxAF1flO:b-j3fq6z)* status: 3 of 7 files reviewed, all discussions resolved (waiting on @nickbianco)

aymanhab:
@nickbianco The recent changes in python avoid using code in setup.py and replace it with a .cfg file instead so while this could possibly be hacked in here, it doesn't seem sustainable long term.
In reality by changing the dll handling on windows only, the python 3.8+ forced this change, the documentation has separate windows and linux/osx is many places.  I initially had the two scripts combined but that was more messy as it required calling the setup script twice with different arguments, so this was way cleaner. I'd lean to try combining the scripts in a separate PR since the current solution is workable even if less than perfect.

nickbianco (commented):
Ah okay, that makes sense. It wasn't clear to me that the pip install avoids the `setup.py` script entirely.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-:-MdT0x0aEuwPiFx567t4:b-unnhb1)* status: 3 of 7 files reviewed, all discussions resolved (waiting on @nickbianco)

nickbianco (changes requested):
> Reviewed 4 of 4 files at r8.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013)* status: all files reviewed, 4 unresolved discussions (waiting on @aymanhab)

___
*[Bindings/Python/__init__.py, line 5 at r8](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-MdT1ipQ8WfRFuVN90BA:-MdT1ipQ8WfRFuVN90BB:bfgzepv) ([raw file](https://github.com/opensim-org/opensim-core/blob/6bea618143b214b301241264e2b6c6a87c3284b5/Bindings/Python/__init__.py#L5)):*
> ```Python
> from pathlib import Path
> if (sys.version_info.major == 3 and sys.version_info.minor >= 8 and sys.platform.startswith('win')):
>         if (Path('opensim/dll_location.txt').is_file()):
> ```

Could use `os.path.isfile()` to avoid the extra import.
___
*[Bindings/Python/setup.py, line 2 at r8](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-MdT2Qik-Pk7fH_scr1a:-MdT2Qik-Pk7fH_scr1b:b-jdh3la) ([raw file](https://github.com/opensim-org/opensim-core/blob/6bea618143b214b301241264e2b6c6a87c3284b5/Bindings/Python/setup.py#L2)):*
> ```Python
> #!/usr/bin/env python
> ```

So this file unused now that we use pip install?
___
*[Bindings/Python/setup_win_python38.py, line 6 at r8](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-MdT2-ZqE99NmIjMLT4t:-MdT2-ZqE99NmIjMLT4u:bb3e8sg) ([raw file](https://github.com/opensim-org/opensim-core/blob/6bea618143b214b301241264e2b6c6a87c3284b5/Bindings/Python/setup_win_python38.py#L6)):*
> ```Python
> import fileinput
> if (sys.version_info.major == 3 and sys.version_info.minor >= 8 and sys.platform.startswith('win')):
>    install_path=os.path.abspath('../../bin')
> ```

Add spaces before/after equal sign. 
___
*[Bindings/Python/setup_win_python38.py, line 10 at r8](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-MdT27AB9DOeZzEcg7mS:-MdT27AB9DOeZzEcg7mT:b-8nuy8j) ([raw file](https://github.com/opensim-org/opensim-core/blob/6bea618143b214b301241264e2b6c6a87c3284b5/Bindings/Python/setup_win_python38.py#L10)):*
> ```Python
>    with fileinput.FileInput('opensim/__init__.py', inplace=True, backup='.bak') as file:
>      for line in file:
>           print(line.replace('DLL_PATH', "'"+new_path+"'"), end='')
> ```

Add spaces before/after "plus" operations.

aymanhab:
Thanks for the review @nickbianco I didn't want to force python 3.7 and earlier to abandon their setup.py mechanism unnecessarily so the instructions are python version dependent rather than OpenSim version dependent. If it turns out that the pip install mechanism works well with earlier versions as well and that setup.py is not needed then sure we can remove it.

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013)* status: 5 of 7 files reviewed, 4 unresolved discussions (waiting on @nickbianco)

___
*[Bindings/Python/__init__.py, line 5 at r8](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-MdT1ipQ8WfRFuVN90BA:-MdT6eIj2GljjOKxavZE:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/6bea618143b214b301241264e2b6c6a87c3284b5/Bindings/Python/__init__.py#L5)):*

Could use `os.path.isfile()` to avoid the extra import.
</blockquote></details>

Done.
___
*[Bindings/Python/setup.py, line 2 at r8](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-MdT2Qik-Pk7fH_scr1a:-MdTL2mi3Y9wGHyGHqP_:bvt7yuk) ([raw file](https://github.com/opensim-org/opensim-core/blob/6bea618143b214b301241264e2b6c6a87c3284b5/Bindings/Python/setup.py#L2)):*

So this file unused now that we use pip install?
</blockquote></details>

I believe so but leaving for other python versions, legacy installation
___
*[Bindings/Python/setup_win_python38.py, line 6 at r8](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-MdT2-ZqE99NmIjMLT4t:-MdT6rGcEtAXU0pIqYjK:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/6bea618143b214b301241264e2b6c6a87c3284b5/Bindings/Python/setup_win_python38.py#L6)):*

Add spaces before/after equal sign. 
</blockquote></details>

Done.
___
*[Bindings/Python/setup_win_python38.py, line 10 at r8](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-MdT27AB9DOeZzEcg7mS:-MdT6yimCUrvnlVZnRE5:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/6bea618143b214b301241264e2b6c6a87c3284b5/Bindings/Python/setup_win_python38.py#L10)):*

Add spaces before/after "plus" operations.
</blockquote></details>

Done.

nickbianco (approved):
> Reviewed 2 of 2 files at r9.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aymanhab)

___
*[Bindings/Python/setup.py, line 2 at r8](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-MdT2Qik-Pk7fH_scr1a:-MdTWNR4DwmS7Fl4SH3u:bp9ocg0) ([raw file](https://github.com/opensim-org/opensim-core/blob/6bea618143b214b301241264e2b6c6a87c3284b5/Bindings/Python/setup.py#L2)):*

I believe so but leaving for other python versions, legacy installation
</blockquote></details>

Okay sounds good, just making sure I understand all the changes.

nickbianco (commented):
@aymanhab I hit approve by accident, I wanted to test locally first. Testing now, will get back soon.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-:-MdTWhFT2N_9NqyCrRtC:b-6u8yos)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aymanhab)

nickbianco:
I will test once the Windows build finishes.

nickbianco (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013)* status: all files reviewed, 2 unresolved discussions (waiting on @aymanhab)

___
*[Bindings/Python/__init__.py, line 7 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-MdTtBbK5f_eKiwkZ_Ay:-MdTtBbK5f_eKiwkZ_Az:b-xuut2q) ([raw file](https://github.com/opensim-org/opensim-core/blob/fb38bf9d9842c5263eb18004f57524aaedc63c42/Bindings/Python/__init__.py#L7)):*
> ```Python
>             with open('opensim/dll_location.txt', 'r') as reader:
>                 os.add_dll_directory(reader.read())
>         else:
> ```

We should remove all stuff related to `dll_location.txt`, right?
___
*[Bindings/Python/CMakeLists.txt, line 260 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-MdTtGN4-2x1jifKdSe4:-MdTtGN4-2x1jifKdSe5:ba3dfna) ([raw file](https://github.com/opensim-org/opensim-core/blob/fb38bf9d9842c5263eb18004f57524aaedc63c42/Bindings/Python/CMakeLists.txt#L260)):*
> ```CMake
> # it in setup_win_python38.py to set additional dll directory
> if (WIN32)
>     set(DLL_PATH_FILE ${CMAKE_CURRENT_BINARY_DIR}/dll_location.txt)
> ```

Remove?

nickbianco (changes requested):
Tested the install workflow locally on Windows and seems to work fine, with two caveats: 1) I had to clear my PYTHONPATH of old OpenSim bin folders, as they will take precedence 2) I had to delete the stuff related to `dll_location.txt` from `__init__.py`.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-:-MdTuxyo53gv0qkjAaNT:b-cedwdy)* status: all files reviewed, 2 unresolved discussions (waiting on @aymanhab)

___
*[Bindings/Python/__init__.py, line 7 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3013#-MdTtBbK5f_eKiwkZ_Ay:-MdTuRVxDH7n7gm2tGD8:b-8gh7uo) ([raw file](https://github.com/opensim-org/opensim-core/blob/fb38bf9d9842c5263eb18004f57524aaedc63c42/Bindings/Python/__init__.py#L7)):*

We should remove all stuff related to `dll_location.txt`, right?
</blockquote></details>

After testing, confirmed that this needs to be removed, as it doesn't get updated by `setup_win_python38.py`.

aymanhab:
Thanks for testing @nickbianco 

- Definitely PYTHONPATH is a strong hammer that should be discouraged as it violates the packaging idea altogether. Until recently we never recommended or mentioned it but users may have been using it anyway.
- The only reason dll_location.txt was left is to enable ci tests to pass as ci doesn't use the pip install mechanism altogether and runs off the sdk/python folder as is. Note that if the file doesn't exist this branch of the code is not used and the file is not opened/parsed which is the non-ci use case. We could try to use another CMake trick instead but still we need  to support two workflows, ci and client install.
- The script does not use the file altogether, instead it writes the installation path to the __init__.py directly. Using the file failed because it depended on where the code is invoked from which was not robust.

aymanhab:
I can possibly remove the dll_location.txt reference and use PYTHONPATH instead for the ci workflow. What do you think,, @nickbianco ?

nickbianco:
> I can possibly remove the dll_location.txt reference and use PYTHONPATH instead for the ci workflow. What do you think,, @nickbianco ?

Yeah I like that solution. The `dll_location.txt` did exist with my install (which makes sense since I got the build off the CI). But it's better to avoid it completely.

nickbianco:
Ah, it looks like the DLL_PATH needs to be filled in somehow for CI.

aymanhab:
Latest version works perfectly as previous version on client machines, build succeeds and python tests pass on windows post installation but fails in dev environment/ctest before installation which is technically not supported (so were skipped).  Thoughts @nickbianco @carmichaelong ?

nickbianco:
You mean that the ctests fail locally? It would be good if these passed, since we give instructions for running ctests if building from source locally.

aymanhab:
I agree @nickbianco for the long term but keep in mind
- Python pre 3.8/windows will continue to pass, while 3.8+/windows would always fail without this PR
- The fix should really be to run python tests in installed environment NOT from build environment. This can be fixed on our side before official release but not necessarily for the beta/TGCS prep run

nickbianco:
Ah okay, that makes sense. Now that I think about it, I usually ignore failing Matlab/Python tests when building locally for this same reason -- you can't really run the tests without installing first. So I think we're good to proceed.

aymanhab:
Awesome, thanks @nickbianco for all your valuable feedback and testing, much appreciated 👍

nickbianco:
Woohoo!

Issue #2871, opened 2020-09-01
inertia after non-uniform manual scaling 


fmatari:
Hi,

Our team is interested in migrating from version 3.3 to 4.1, we have noticed that there is a difference in inertia results when performing non-uniform manual scaling. After performing the scaling in a spreadsheet, we confirmed that version 3.3 provided correct results and 4.1 incorrect. The source code for 4.1 was looked into and after some cout statements and digging around, the issue was found. From body.cpp:scaleInertialProperties(),

        if (axis == 0) {
            rad_sqr = radius * (scaleFactors[1]) * radius * (scaleFactors[2]);
            inertia[0][0] = 0.5 * get_mass() * rad_sqr;
            inertia[1][1] = get_mass() * ((length * length / 12.0) + 0.25 * rad_sqr);
            inertia[2][2] = get_mass() * ((length * length / 12.0) + 0.25 * rad_sqr);

        } else if (axis == 1) {
            rad_sqr = radius * (scaleFactors[0]) * radius * (scaleFactors[2]);
            inertia[0][0] = get_mass() * ((length * length / 12.0) + 0.25 * rad_sqr);
            inertia[1][1] = 0.5 * get_mass() * rad_sqr;
            inertia[2][2] = get_mass() * ((length * length / 12.0) + 0.25 * rad_sqr);

        } else {
            rad_sqr = radius * (scaleFactors[0]) * radius * (scaleFactors[1]);
            inertia[0][0] = get_mass() * ((length * length / 12.0) + 0.25 * rad_sqr);
            inertia[1][1] = get_mass() * ((length * length / 12.0) + 0.25 * rad_sqr);
            inertia[2][2] = 0.5 * get_mass() * rad_sqr;
        }

We confirmed that the if statements are properly satisfied so the calculations should be performed. Though calculations are never performed and the problem is found to be in the inertia variable type. Inertia utilizes a symmetric 3 x 3 matrix and this matrix is not allowing the assignment of values to specific elements. Changing the variable type from a symmetric matrix to a regular matrix fixes this issue. 

    //SimTK::SymMat33 inertia = _inertia.asSymMat33();
    
    SimTK::Mat33 inertia = _inertia.toMat33();

The top commented line is the original and bottom is the suggested fix. We have not looked into why the symmetric matrix fails. 

Best regards,
Fouad

dfrenkel101:
[...]

PR #3015, opened 2021-06-21
Use Mat33 instead of SymMat33 in Body::scale

aymanhab:
The code sets values both over/under the diagonal which makes sense for actual Mat33. Fix suggested and tested by fmatari on user forum.

Fixes issue #2871

### Brief summary of changes
Replace type of local variable from SysMat33 to Mat33

### Testing I've completed
Ran regression tests
### CHANGELOG.md (choose one)
- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3015)

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -40,7 +40,7 @@ jobs:
       run: python -m pip install numpy==1.19.3
 
     - name: Install SWIG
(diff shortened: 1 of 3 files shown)


aymanhab:
@carmichaelong for your review, from user forum.

carmichaelong (approved):

Issue #3018, opened 2021-06-23
[Moco


tymo77:

tymo77:
Here's the link to the simple changes I propose and updates to two examples to demonstrate it:

https://github.com/opensim-org/opensim-core/compare/master...tymo77:ipopt_options_file

nickbianco:
Very cool! Looks like this is mostly done, so I think at this point you could go ahead and create a PR for it. I'm curious to test out some of these options myself.

PR #3025, opened 2021-07-01
Ipopt options file

tymo77:
Addresses issue #3018

### Brief summary of changes
- Add ipopt option to the CasADi solver and tropter solvers in Moco that passes the ipopt options file name to look for.
- Add example of the use of these external files to 2D walking examples.
- Ipopt will look for the external options file each time it runs and use it to set advanced solver parameters. See: [Ipopt options docs](https://coin-or.github.io/Ipopt/OPTIONS.html)
- Note: The way things are setup now, using the external options file, one cannot overwrite any settings that are set and compiled in the source code or otherwise set in the source code or script. Ipopt will simply issue a warning:
```
WARNING: Tried to set option "<option_name>" to a value of "<option_value_in_file>",
         but the previous value is set to disallow clobbering.
         The setting will remain as: "<value_set_in_source_or_script>"
```

### Testing I've completed
- I have tested it with both solvers in the 2D walking examples.

- First PR, so anything about contributing that I have missed.
- E.g. I don't know how we handle version numbers.

### CHANGELOG.md
- no need to update because this seems like a minor update that is transparent to the user?

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3025)

### DIFF
--- a/Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalking.m
+++ b/Bindings/Java/Matlab/examples/Moco/example2DWalking/example2DWalking.m
@@ -52,6 +52,9 @@ clear;
 % Load the Moco libraries
 import org.opensim.modeling.*;
 
+% Flag to enable or disable the custom ipopt options files.
+USE_CUSTOM_OPTIONS = true;
+
 % ---------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


nickbianco:
Closing: stale PR.

Issue #3050, opened 2021-07-31
Several small errors for the OpenSense Python script in OpenSim 4.3 Beta


Allwillcome:
- TRUE should be true; 
- pi should be called first before use; `import math` `pi = math.pi`
- Vector3 should be written into `osim.Vector3` , Not Vector3
`sensor_to_opensim_rotations = Vec3(-pi/2, 0, 0);# The rotation of IMU data to the OpenSim world frame`
When I guess the cause of these mistakes, Python's script changes directly from the MATLAB script because these errors are different in language syntax. This will happen to people who are not familiar with programs. If the next time you publish the Python script, you can find these problems first.

aymanhab:
Thanks for reporting, I'll submit a fix. Please keep reporting any issues with Python scripts here so we can continue to fix/improve or open a PR with the fixes and we'll review

Allwillcome:
> Thanks for reporting, I'll submit a fix. Please keep reporting any issues with Python scripts here so we can continue to fix/improve or open a PR with the fixes and we'll review

Ok, Thanks for your effort ! OpenSim is great!

PR #3053, opened 2021-08-02
Fix errors reported by user (use osim. where needed,  true ->True)

aymanhab:
Fixes issue #3050 

### CHANGELOG.md (choose one)

- no need to update because bugfix

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3053)

### DIFF
--- a/Bindings/Python/OpenSenseExample/OpenSense_CalibrateModel.py
+++ b/Bindings/Python/OpenSenseExample/OpenSense_CalibrateModel.py
@@ -22,7 +22,7 @@
 
 # Import OpenSim libraries
 import opensim as osim
-import math.pi as pi
+from math import pi
 # Set variables to use
 modelFileName = 'Rajagopal_2015.osim';          # The path to an input model
 orientationsFileName = 'MT_012005D6_009-001_orientations.sto';   # The path to orientation data for calibration 
--- a/Bindings/Python/OpenSenseExample/OpenSense_CalibrateModel.py
+++ b/Bindings/Python/OpenSenseExample/OpenSense_CalibrateModel.py
@@ -22,14 +22,14 @@
 
 # Import OpenSim libraries
 import opensim as osim
-
+import math.pi as pi
 # Set variables to use
 modelFileName = 'Rajagopal_2015.osim';          # The path to an input model
 orientationsFileName = 'MT_012005D6_009-001_orientations.sto';   # The path to orientation data for calibration 
-sensor_to_opensim_rotations = Vec3(-pi/2, 0, 0);# The rotation of IMU data to the OpenSim world frame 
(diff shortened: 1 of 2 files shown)


aymanhab:
Ready for quick review @carmichaelong

Bindings/Python/OpenSenseExample/OpenSense_CalibrateModel.py
```diff
@@ -22,14 +22,14 @@
 
 # Import OpenSim libraries
 import opensim as osim
-
+import math.pi as pi
```
carmichaelong, line 25: This line doesn't work for me (`No module named 'math.pi'; 'math' is not a package`). Consider `from math import pi` instead, and then `pi` will be defined too

Bindings/Python/OpenSenseExample/OpenSense_OrientationTracking.py
```diff
@@ -26,12 +26,12 @@
 
 # Import OpenSim Libraries
 import opensim as osim
-
+import math.pi as pi
```
carmichaelong, line 29: Same here

carmichaelong (changes requested):

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3053)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @carmichaelong)

___
*[Bindings/Python/OpenSenseExample/OpenSense_CalibrateModel.py, line 25 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3053#-Mg8GAWLoTjYhV6oXWwQ-r1-25:-MgC0IwN9QAv4XEAhL0h:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/640e5a1c64ff168ef384f658d70415a88dd8822f/Bindings/Python/OpenSenseExample/OpenSense_CalibrateModel.py#L25)):*

This line doesn't work for me (`No module named 'math.pi'; 'math' is not a package`). Consider `from math import pi` instead, and then `pi` will be defined too
</blockquote></details>

Done.
___
*[Bindings/Python/OpenSenseExample/OpenSense_OrientationTracking.py, line 29 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3053#-Mg8GAWLoTjYhV6oXWwR-r1-29:-MgC0KRmEKqI3A0vWw4W:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/640e5a1c64ff168ef384f658d70415a88dd8822f/Bindings/Python/OpenSenseExample/OpenSense_OrientationTracking.py#L29)):*

Same here
</blockquote></details>

Done.

carmichaelong (approved):
LGTM. I will also add this to the testing doc

aymanhab:
Thanks @carmichaelong

PR #3054, opened 2021-08-03
Add CMake flag SKIP_SWIG_DOXYGEN to make processing Doxygen by SWIG optional

aymanhab:
Fixes issue #0

### Brief summary of changes
Added CMake variable to control whether to process Doxygen comments for scripting purposes. While great for final releases, in dev environments it takes longer and produces massive list of bogus errors/warnings
### Testing I've completed
Built on windows with the flag on and off and got expected behavior
Whether this is useful and the cmake variable name has the right connotation
### CHANGELOG.md (choose one)

- no need to update because DOXYGEN inclusion was just introduced in 4.3

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3054)

### DIFF
--- a/Bindings/CMakeLists.txt
+++ b/Bindings/CMakeLists.txt
@@ -11,11 +11,11 @@ if(WITH_EZC3D)
 endif()
 #flag to indicate whether to let SWIG process Doxygen comments
 # Will be on for releases but too many warnings/errors for regular dev.
-set(SKIP_SWIG_DOXYGEN OFF CACHE BOOL "Don't carry Doxygen comments to bindings")
-mark_as_advanced(SKIP_SWIG_DOXYGEN)
-set(SWIG_DOXYGEN "")
-if (NOT SKIP_SWIG_DOXYGEN)
-    set(SWIG_DOXYGEN "-doxygen")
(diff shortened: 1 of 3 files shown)


aymanhab:
@carmichaelong This is a rather minor change that is no-op by default but allows for ignoring doxygen comments by SWIG. Needed because we want smaller files when building containers, also makes it possible to reduce bogus warnings by SWIG

carmichaelong:
Interesting, how much time does it seem to save?

I think I'd prefer the variable name to start with `SWIG` so that it'll group closer to that section in CMake. I'm split between `SWIG_SKIP_DOXYGEN` or `SWIG_DOXYGEN_SKIP`.

I guess also could just be `SWIG_DOXYGEN` and then default to True.

aymanhab:
Thanks @carmichaelong it's not the time (hard to measure precisely) but the warnings in the hundreds  (75 on the first python library processing). Also when building as a backend you don't care about comments and instead make smallest footprint. I made the change as you suggested to (SWIG_DOXYGEN).

carmichaelong (approved):

aymanhab:
Thanks much @carmichaelong 👍

PR #3063, opened 2021-08-18
Fix test case that fails on linux only

aymanhab:
Fixes issue #<issue_number>

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3063)


aymanhab:
Opened issue https://github.com/opensim-org/opensim-core/issues/3065 to address the platform differences.

PR #3068, opened 2021-08-20
update documentation

antoinefalisse:
This PR updates the documentation for building opensim-core from source. On windows, you should use swig 4.0.2.

A couple of remarks:
- It did not work with swig 4.0.1 - I had problems with the moco python bindings. I did not try with 4.0.0.
- I see in the readme that there are some `swig3.0` as part of some commands. Not sure we can safely change that to 4.0.2 and I do not have ubuntu to try.

thanks.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3068)

### DIFF
--- a/README.md
+++ b/README.md
@@ -512,7 +512,7 @@ On Windows using Visual Studio
     * [TortoiseGit](https://code.google.com/p/tortoisegit/wiki/Download),
       intermediate; good for TortoiseSVN users;
     * [GitHub for Windows](https://windows.github.com/), easiest.
-* **Bindings** (optional): [SWIG](http://www.swig.org/) 3.0.8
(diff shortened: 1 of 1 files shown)


aymanhab (approved):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3068)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @antoinefalisse)

aymanhab:
Thanks @antoinefalisse LGTM. I just removed the reference to earlier versions of VisualStudio since I know our code doesn't compile on these. I will add a disclaimer in another PR that the actual tested set of tools/versions are found in the ci build script as the final defacto reference.

PR #3084, opened 2021-09-23
Updated Ezc3d to v1.4.6

pariterre:
Fixes issue NO ISSUE

### Brief summary of changes
Updated ezc3d to the newest version

### Testing I've completed
Compiles locally on Linux and Windows

My career choice (:P)

### CHANGELOG.md (choose one)

- no need to update

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3084)


pariterre:
I feel that the bugs in Mac are unrelated to the current PR...

aymanhab:
True these test cases fail sporadically on Mac and they don't use C3D files so I'm not worried about them. Need to include a changelog entry with relevant changes if any that could be of interest to OpenSim users.

pariterre:
Hi @aymanhab 

I do not think the users will experience any changes, as the most functionality that I focused on lately were very users specific. That said, since OpenSim is slightly behind in terms of ezc3d version, some users experience bugs while trying to use ezc3d via python while it is also installed via OpenSim (clash of DLL). I wanted to see if I could easily change the name of the DLL of ezc3d when used by OpenSim to help with this problem. 
In the process, I figured that I could just pushed the update.

Should I add anything?

aymanhab:
Thanks for the feedback, but can you clarify the conclusion regarding naming? In particular should we change anything on the OpenSim side in terms of the library name for building/linking purposes?

aymanhab:
Also Changelog.md needs to be updated as part of this PR

pariterre:
> Thanks for the feedback, but can you clarify the conclusion regarding naming? In particular should we change anything on the OpenSim side in terms of the library name for building/linking purposes?

I do not know where to implement this, but if it was possible to change the name of the shared library built by ezc3d when compiling using the build dependency procedure (or renaming when installing) it would allow for the users to install any version of ezc3d in their environment, while opensim would still be using its own compiled version (this is relevant when someone install a precompile version of OpenSim). 
There are plenty of ways to implement this, but I do not know if it is something you guys want to spend time on. On the plus side, this is not specific to ezc3d, but any library opensim compile for itself, so I feel that if anything is done for this, this could be implemented at the same time on all the shared library built by opensim. I would suggest a suffix (something like `_osim`) to be added to the name, so for instance `ezc3d.dll` would be named `ezc3d_osim.dll` or `libezc3d_osim.dll` (on UNIX)

I'll push the changes in log

aymanhab:
Thanks @pariterre We have no plans to change the name of this or any other libraries (there're a ton of them now that use a variety of build systems with the addition of Moco) on our side any time soon. Most of our users install precompiled libraries. If default behavior in the new version produces the same library name as before the change then we're good with that and will not touch it. I just wanted to confirm. Thank you

aymanhab (approved):
LGTM, thank you @pariterre

PR #3089, opened 2021-10-03
Allow MrkerReference smart_ptr recognition by swig

aymanhab:
Fixes issue #<issue_number>

### CHANGELOG.md (choose one)

- updated.

### DIFF
--- a/Bindings/simulation.i
+++ b/Bindings/simulation.i
@@ -221,7 +221,7 @@ OpenSim::ModelComponentSet<OpenSim::Controller>;
 %shared_ptr(ReferenceDouble);
 %shared_ptr(ReferenceRotation);
 %include <OpenSim/Simulation/MarkersReference.h>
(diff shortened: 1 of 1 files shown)


aymanhab:
@nickbianco For your review, just to make sure functions needed for Moco in Python, Matlab are functional. Note that the change is in SWIG interface only so if more testing is needed it's be exclusive to the bindings. Thank you.

nickbianco:
@aymanhab didn't we have bindings tests for MarkersReference that we disabled for this issue? We should re-enable them or add some if they don't exist.

aymanhab:
@nickbianco I don't recall disabling any tests but if you can find something in the commit/log indicating that then please go ahead and enable it. In the meantime I'm adding a Java test case that exercises this constructor and will beef up with other scenarios that @CVHammond is encountering using IK from Matlab since these are typical use cases.

aymanhab:
Ready for review, tests added.

Issue #3109, opened 2021-11-20
ScaleTool: loading model or market set file can fail if specified as absolute paths


carmichaelong:
In MATLAB (also in C++):
1. if the model file in the setup file is given as an absolute path (e.g., `C:\project\model.osim`)
2. AND the ScaleTool object is constructed from a setup file using an absolute or relative path, either of the following:
 `scale = ScaleTool("C:\project\setup_scale.xml")`
`scale = ScaleTool("project\setup_scale.xml")`
3. Then I get the error that no model is specified.

If the setup file is in the same folder as the working directory (e.g., I choose the working directory to be where the setup file is and use `scale = ScaleTool("setup_scale.xml")`), then it runs correctly.

If the model file in the setup file is given as a relative path (e.g., the model file is in the same directory as the setup file, and specified as `model.osim` , then you can construct the ScaleTool in any way (absolute, relative, etc) and it runs fine.

Also, when looking through the code to fix, the marker set file also looked to susceptible to the same problem.

aymanhab:
If this was verified by others in release/master build the issue can be moved along

PR #3110, opened 2021-11-20
fix load issue with scale tool 

carmichaelong:
Fixes issue #3109 

### Brief summary of changes
Updated how `GenericModelMaker::processModel()` figures out the path to the model file and marker set file. Also updated an error message in ScaleTool to be less misleading in certain cases.

### Testing I've completed
Ran usual tests locally. Also tested with different setup files that had absolute paths for model and marker set file.

Didn't add any test for this, but also couldn't think of a way to reliably add a test with an absolute path in a setup file.

### CHANGELOG.md (choose one)
- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3110)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This is not a comprehensive list of changes but rather a hand-curated collection
 v4.4
 ====
 - Updated ezc3d to version 1.4.6 which better manage the events defined in a c3d file.
+- Fixed an issue that could happen sometimes with ScaleTool where loading the model file or marker set file could fail if the file was given as an absolute path (Issue #3109, PR #3110)
 
 v4.3
 ====
--- a/OpenSim/Tools/GenericModelMaker.cpp
+++ b/OpenSim/Tools/GenericModelMaker.cpp
@@ -159,13 +159,17 @@ Model* GenericModelMaker::processModel(const string& aPathToSubject) const
 
     try
     {
-        model = new Model(aPathToSubject + _fileName);
+        std::string modelPath = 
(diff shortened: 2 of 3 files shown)


aymanhab (commented):
An important workflow that is not exercised in tests is the GUI workflow. The GUI jumps through massive hoops to handle paths since working directory is the installation folder while files are sprinkled over disk(s). Also not sure how this handles files on different drives C:/D:/F:  Will have to build and test or if already tested please indicate. Thank you

> Reviewed 1 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3110#-:-MpgyIhJAy84kUiZ-EyM:b7yq9oo)* status: 1 of 3 files reviewed, all discussions resolved (waiting on @aymanhab)

carmichaelong:
Thanks for the feedback @aymanhab. I haven't exercised any of those cases with the GUI, so that would be great if you could test since it seems like you've got a good idea for where that could break.

The underlying SimTK functions have a bunch of logic to deal with different OS's, so there's a good chance it should be OK with different drive letters and such, but there's definitely some 🤞 there since I didn't test directly with different drives or in the GUI.

aymanhab (approved):
Tested the GUI on windows and seems to work.

> Reviewed 2 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3110#-:-Mpm6mGc0TMrkY_7kdjk:b4y9m7c)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @carmichaelong)

Issue #3119, opened 2021-12-27
Continuing inverse kinematics after IpOpt fails to achieve desired accuracy


nishbo:
Hi, I am currently using OpenSim to calculate IK, but on my current data it often fails to achieve the target accuracy on several frames. Lowering the accuracy allows me to run the whole trial, but it somewhat changes the produced DOF angles (mean diff 4-8 degrees). I edited the InverseKinematicsTool to ignore this error whenever it is thrown and I get reasonable behavior after it. 

1. I do not know enough about Simbody, so I am worried if there are any possible problems with such behavior. I.e. are there any possible undefined behaviors within State after such error has been thrown?
2. I also developed an option that lowers the accuracy and reassembles the ikSolver whenever the exception is thrown, which seems to produce very similar results. I am going to clean up my fork and send a PR later, for now the changes can be compared [here](https://github.com/nishbo/opensim-core/tree/master). It seems to me like a better way of handling the error, but I am not sure. Thoughts?

PR #3120, opened 2021-12-28
Continuing inverse kinematics after IpOpt fails to achieve desired accuracy

nishbo:
Fixes issue #3119

### Brief summary of changes
Added a flag (`adaptiveAccuracy`) and two parameters to InverseKinematicsToolBase that allow rerunning track on a frame with lower accuracy if the optimizer failed to converge.
Added another flag (`ignoreConvergenceErrors`) to InverseKinematicsToolBase that allows ignoring a convergence error during track() of a frame.
`void AssemblySolver::track(SimTK::State &s)` now throws a more comprehensive exception that can be traced, similar to `void AssemblySolver::assemble(SimTK::State &state)`.

### Testing I've completed
Running with flags turned off produces the same results as the default version. Ran tests with enabling each or both of the flags, behavior is similar (mean DOF angle difference ~ 1 degree).

### CHANGELOG.md (choose one)

Please tell me where to append a note of change.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3120)

### DIFF
--- a/Applications/IK/ik.cpp
+++ b/Applications/IK/ik.cpp
@@ -141,13 +141,19 @@ int main(int argc,char **argv)
     log_info("Constructing tool from setup file {}.", setupFileName);
     InverseKinematicsTool ik(setupFileName);
 
-    // testprint
+    // Adaptive accuracy announcement
     if (ik.get_adaptiveAccuracy())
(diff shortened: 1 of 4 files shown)


nickbianco:
Closing: stale PR.

PR #3125, opened 2022-01-18
Avoid tweaking tangent points of WrapCylinder when single object wrap…

aymanhab:
…ping, allow for optimization if disableVisualization is true. Fix test cases accordingly

Fixes issue #<issue_number>

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125)


nickbianco (changes requested):
@aymanhab, based on the the changes here and the CI logs, I think it's fine to create a new standard for testMocoInverse. 

> Reviewed 2 of 19 files at r2, 26 of 26 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-:-Mtj4szF9Ebgqz3EQ0-U:b-ook556)* status: all files reviewed, 4 unresolved discussions (waiting on @aymanhab)

___
*[OpenSim/Common/ModelDisplayHints.h, line 110 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-Mtj1G7EEUxjaFh6RI86:-Mtj1G7EEUxjaFh6RI87:b1xce4y) ([raw file](https://github.com/opensim-org/opensim-core/blob/d4ccffff26830b5070ed9bb7361acefb7bf8387b/OpenSim/Common/ModelDisplayHints.h#L110)):*
> ```objc
>     /** Turn off visualization completely, only use API/modeling.
>     Meshes will not be loaded, pathwrappig intermediate points not computed.
>     Intentionally there's no reverse API to turn on visualization downstream.
> ```

Could you add a note that indicates that excluding the intermediate path point calculations doesn't affect things like moment arm calculations? Also, typo pathwrappig --> pathwrapping.
___
*[OpenSim/Moco/MocoProblemRep.cpp, line 185 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-Mtj39EIDl8u-VEM9Aoh:-Mtj39EIDl8u-VEM9Aoi:byr1725) ([raw file](https://github.com/opensim-org/opensim-core/blob/d4ccffff26830b5070ed9bb7361acefb7bf8387b/OpenSim/Moco/MocoProblemRep.cpp#L185)):*
> ```cpp
>     // Grab a writable state from the copied model -- we'll use this to disable
>     // its constraints below.
>     m_model_disabled_constraints.updDisplayHints().disableVisualization();
> ```

You probably want to make this call on `m_model_base` as well. Or if this setting persists upon model copy, then you could replace this call with a call on `m_model_base`.
___
*[OpenSim/Simulation/Wrap/WrapCylinder.cpp, line 546 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-Mtj3ngfAbBDHutU4Ppv:-Mtj3ngfAbBDHutU4Ppw:b-86k1y) ([raw file](https://github.com/opensim-org/opensim-core/blob/d4ccffff26830b5070ed9bb7361acefb7bf8387b/OpenSim/Simulation/Wrap/WrapCylinder.cpp#L546)):*
> ```cpp
>     vv.normalize();
> 
>     SimTK::UnitVec3 plane_nomral(uu % vv);
> ```

Typo here that might be breaking some results: `plane_nomral` --> `plane_normal`. Also, this would redefine the `plane_normal` instantiation above.
___
*[OpenSim/Simulation/Wrap/WrapCylinder.cpp, line 667 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-Mtj4MPp1jZ8WH_z2nnT:-Mtj4MPp1jZ8WH_z2nnU:bu8cpjz) ([raw file](https://github.com/opensim-org/opensim-core/blob/d4ccffff26830b5070ed9bb7361acefb7bf8387b/OpenSim/Simulation/Wrap/WrapCylinder.cpp#L667)):*
> ```cpp
>     // WrapTorus creates a WrapCyl with no connected model, avoid this hack
>     if (!_model.empty() && !getModel().getDisplayHints().isVisualizationEnabled() &&
>             aWrapResult.singleWrap) {
> ```

Why is this only updated when visualization is disabled?

aymanhab:
Thanks for the great feedback @nickbianco can you look at the new standard vs. old standard files attached below (sto files renamed to txt for github purposes) to see if differences are justified/reasonable? Thank you
[std_testMocoInverse_subject_18musc_solution.txt](https://github.com/opensim-org/opensim-core/files/7906985/std_testMocoInverse_subject_18musc_solution.txt)
[testMocoInverse_subject_18musc_solution.txt](https://github.com/opensim-org/opensim-core/files/7906986/testMocoInverse_subject_18musc_solution.txt)

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125)* status: 25 of 31 files reviewed, 4 unresolved discussions (waiting on @nickbianco)

___
*[OpenSim/Common/ModelDisplayHints.h, line 110 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-Mtj1G7EEUxjaFh6RI86:-Mtq20fy-xe7kPcpd2Hi:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/d4ccffff26830b5070ed9bb7361acefb7bf8387b/OpenSim/Common/ModelDisplayHints.h#L110)):*

Could you add a note that indicates that excluding the intermediate path point calculations doesn't affect things like moment arm calculations? Also, typo pathwrappig --> pathwrapping.
</blockquote></details>

Done.
___
*[OpenSim/Moco/MocoProblemRep.cpp, line 185 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-Mtj39EIDl8u-VEM9Aoh:-Mtq25hNAbRyTVYADjXG:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/d4ccffff26830b5070ed9bb7361acefb7bf8387b/OpenSim/Moco/MocoProblemRep.cpp#L185)):*

You probably want to make this call on `m_model_base` as well. Or if this setting persists upon model copy, then you could replace this call with a call on `m_model_base`.
</blockquote></details>

Done.
___
*[OpenSim/Simulation/Wrap/WrapCylinder.cpp, line 546 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-Mtj3ngfAbBDHutU4Ppv:-MtsNRzk6ZnzdMMXpZ61:bx1nehq) ([raw file](https://github.com/opensim-org/opensim-core/blob/d4ccffff26830b5070ed9bb7361acefb7bf8387b/OpenSim/Simulation/Wrap/WrapCylinder.cpp#L546)):*

Typo here that might be breaking some results: `plane_nomral` --> `plane_normal`. Also, this would redefine the `plane_normal` instantiation above.
</blockquote></details>

Thanks for the find :+1:  fixed, surprisingly didn't change anything else :confused: 
___
*[OpenSim/Simulation/Wrap/WrapCylinder.cpp, line 667 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-Mtj4MPp1jZ8WH_z2nnT:-MtsNcy1E5ZhGqJoF6wj:b8gs0xy) ([raw file](https://github.com/opensim-org/opensim-core/blob/d4ccffff26830b5070ed9bb7361acefb7bf8387b/OpenSim/Simulation/Wrap/WrapCylinder.cpp#L667)):*

Why is this only updated when visualization is disabled?
</blockquote></details>

Because if visualization is enabled we fall through to the code that generates intermediate wrapping points for visualization

aymanhab:
@carmichaelong For your review

carmichaelong (changes requested):
Thanks @aymanhab. I left a bunch of minor comments. A couple of comments are related to whether or not we should update some standards instead of nudge the tolerances, and I'm open to either depending on how they looked to you

> Reviewed 2 of 19 files at r2, 6 of 26 files at r7, 4 of 4 files at r8, 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-:-MtyQ7K2AxYtB63GRSrF:bl094ah)* status: all files reviewed, 12 unresolved discussions (waiting on @aymanhab and @nickbianco)

___
*[Applications/Forward/test/testForward.cpp, line 163 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MtxtZbZ9fALpA3BTemn:-MtxtZbZ9fALpA3BTemo:b-2hd5as) ([raw file](https://github.com/opensim-org/opensim-core/blob/a8bd62a2e45db66de28a168e9175fc124c7ac760/Applications/Forward/test/testForward.cpp#L163)):*
> ```cpp
>         stringstream message;
>         message << "t=" << time <<" state# "<< j << " " << standard->getColumnLabels()[j+1] << " std=" << data[j] <<"  computed=" << state->getData()[j] << endl;
>         ASSERT_EQUAL(data[j], state->getData()[j], 5.0e-3, 
> ```

should the standard be updated instead of increasing tolerance (similar to Moco tests)?
___
*[Applications/Forward/test/testForward.cpp, line 176 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MtxtfJd6WvPWvPpxtRh:-MtxtfJd6WvPWvPpxtRi:b-jkrb48) ([raw file](https://github.com/opensim-org/opensim-core/blob/a8bd62a2e45db66de28a168e9175fc124c7ac760/Applications/Forward/test/testForward.cpp#L176)):*
> ```cpp
>         stringstream message;
>         message << "t=" << time <<" state# "<< j << " " << standard->getColumnLabels()[j+1] << " std=" << data[j] <<"  computed=" << state->getData()[j] << endl;
>         ASSERT_EQUAL(data[j], state->getData()[j], 5.0e-3, 
> ```

same as above too
___
*[OpenSim/Simulation/Model/Geometry.cpp, line 250 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MtxvMYh9HcJrubTAw1k:-MtxvMYiEu6wJKtYpkdC:bcl7eih) ([raw file](https://github.com/opensim-org/opensim-core/blob/a8bd62a2e45db66de28a168e9175fc124c7ac760/OpenSim/Simulation/Model/Geometry.cpp#L250)):*
> ```cpp
>         if (file.empty() || file.compare(PropertyStr::getDefaultStr()) == 0 ||
>             !mdl->getDisplayHints().isVisualizationEnabled())
>             return;  // Return immediately if no file has been specified.
> ```

this comment should be updated to reflect change too
___
*[OpenSim/Simulation/Model/GeometryPath.cpp, line 857 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-Mtxz1ZL4ob9n6dSJ0F7:-Mtxz1ZL4ob9n6dSJ0F8:bvlbbhh) ([raw file](https://github.com/opensim-org/opensim-core/blob/a8bd62a2e45db66de28a168e9175fc124c7ac760/OpenSim/Simulation/Model/GeometryPath.cpp#L857)):*
> ```cpp
> 
>     setLengtheningSpeed(s, speed);
>     log_trace("length {}, speed {}",
> ```

thoughts on `debug` instead of `trace`? i don't know if we have this standardized in any way, so i'm open to either
___
*[OpenSim/Simulation/Wrap/WrapCylinder.cpp, line 665 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-Mtxy30G1le-9a8ZBczA:-Mtxy30G1le-9a8ZBczB:b54qea1) ([raw file](https://github.com/opensim-org/opensim-core/blob/a8bd62a2e45db66de28a168e9175fc124c7ac760/OpenSim/Simulation/Wrap/WrapCylinder.cpp#L665)):*
> ```cpp
>     m.set(1, 0, uu[0]); m.set(1, 1, uu[1]); m.set(1, 2, uu[2]);
>     m.set(2, 0, vv[0]); m.set(2, 1, vv[1]); m.set(2, 2, vv[2]);
>     // WrapTorus creates a WrapCyl with no connected model, avoid this hack
> ```

i'm probably missing something here, but i don't understand this comment and what it means for this section of code (the comment within the `if` statement makes sense though)
___
*[OpenSim/Simulation/Wrap/WrapMath.h, line 31 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-Mtxzym552sfqCh1EzJz:-Mtxzym552sfqCh1EzK-:b-xtt6ds) ([raw file](https://github.com/opensim-org/opensim-core/blob/a8bd62a2e45db66de28a168e9175fc124c7ac760/OpenSim/Simulation/Wrap/WrapMath.h#L31)):*
> ```objc
> 
> #include <OpenSim/Simulation/osimSimulationDLL.h>
> #include <SimTKcommon.h>
> ```

could have a more targeted include
___
*[OpenSim/Simulation/Wrap/WrapResult.h, line 64 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MtyPDMw8xpvLS30gygh:-MtyPDMw8xpvLS30gygi:bq1ujxv) ([raw file](https://github.com/opensim-org/opensim-core/blob/a8bd62a2e45db66de28a168e9175fc124c7ac760/OpenSim/Simulation/Wrap/WrapResult.h#L64)):*
> ```objc
>     // variable.
>     double factor = SimTK::NaN;  // scale factor used to normalize parameters
>     bool singleWrap = false; //Flag to indicate if a single wrap object to be considered so we can optimize
> ```

i might be confused by this wording, this flag means that there is exactly 1 wrap object (no more or fewer)?
___
*[OpenSim/Tools/Test/testControllers.cpp, line 415 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MtyPgNH9TSE0rh5T13f:-MtyPgNH9TSE0rh5T13g:b7dx7xv) ([raw file](https://github.com/opensim-org/opensim-core/blob/a8bd62a2e45db66de28a168e9175fc124c7ac760/OpenSim/Tools/Test/testControllers.cpp#L415)):*
> ```cpp
> 
>     CHECK_STORAGE_AGAINST_STANDARD(controls, std_controls, 
>         std::vector<double>(nstates, 0.015), __FILE__, __LINE__,
> ```

should the standard be updated instead of tolerance changed?

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125)* status: all files reviewed, 12 unresolved discussions (waiting on @aymanhab, @carmichaelong, and @nickbianco)

___
*[Applications/Forward/test/testForward.cpp, line 163 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MtxtZbZ9fALpA3BTemn:-Mu8L1CT7qBxwJGpn3fj:b-6rdlkk) ([raw file](https://github.com/opensim-org/opensim-core/blob/a8bd62a2e45db66de28a168e9175fc124c7ac760/Applications/Forward/test/testForward.cpp#L163)):*

should the standard be updated instead of increasing tolerance (similar to Moco tests)?
</blockquote></details>

My inclination is to change the tolerance in the case where this is a regression against previous run/version and the change is minimal and to update standard if against theoretical/known values. Generally treating all states using the same tolerance including angles, speeds and activations is a bit of a hammer. For Moco tests, if the model can't reproduce/track states then we have no options
___
*[OpenSim/Simulation/Model/Geometry.cpp, line 250 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MtxvMYh9HcJrubTAw1k:-Mu8Mjiq8Y1Dr8ftLKIK:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/a8bd62a2e45db66de28a168e9175fc124c7ac760/OpenSim/Simulation/Model/Geometry.cpp#L250)):*

this comment should be updated to reflect change too
</blockquote></details>

Done.
___
*[OpenSim/Simulation/Model/GeometryPath.cpp, line 857 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-Mtxz1ZL4ob9n6dSJ0F7:-Mu8MnKW-rGfhHFPmwqm:b-acfxqw) ([raw file](https://github.com/opensim-org/opensim-core/blob/a8bd62a2e45db66de28a168e9175fc124c7ac760/OpenSim/Simulation/Model/GeometryPath.cpp#L857)):*

thoughts on `debug` instead of `trace`? i don't know if we have this standardized in any way, so i'm open to either
</blockquote></details>

I'll remove this altogether since this was for internal debugging temporarily and unlikely to be used for debug by users since it is called too many times. 
___
*[OpenSim/Simulation/Wrap/WrapCylinder.cpp, line 665 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-Mtxy30G1le-9a8ZBczA:-Mu8NiIz2VAppmiMDgts:b3cpby9) ([raw file](https://github.com/opensim-org/opensim-core/blob/a8bd62a2e45db66de28a168e9175fc124c7ac760/OpenSim/Simulation/Wrap/WrapCylinder.cpp#L665)):*

i'm probably missing something here, but i don't understand this comment and what it means for this section of code (the comment within the `if` statement makes sense though)
</blockquote></details>

The test below assumes _model is not empty which is not the case for WrapTorus which creates a WrapCylinder on the fly and calls this method on it, this WrapCylinder has no _model.
___
*[OpenSim/Simulation/Wrap/WrapMath.h, line 31 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-Mtxzym552sfqCh1EzJz:-Mu8Oec8CVBj_xfsXxYn:b-6vzxqh) ([raw file](https://github.com/opensim-org/opensim-core/blob/a8bd62a2e45db66de28a168e9175fc124c7ac760/OpenSim/Simulation/Wrap/WrapMath.h#L31)):*

could have a more targeted include
</blockquote></details>

 The tradeoff is some/multiple internal SimTK headers which exposes the internals of the library.  For a utility class that ends up as a mix of files that are not logically related, but I don't feel strongly eitherway 
___
*[OpenSim/Simulation/Wrap/WrapResult.h, line 64 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MtyPDMw8xpvLS30gygh:-Mu8Q3-43eQcVAnSLpa-:b-67bnd1) ([raw file](https://github.com/opensim-org/opensim-core/blob/a8bd62a2e45db66de28a168e9175fc124c7ac760/OpenSim/Simulation/Wrap/WrapResult.h#L64)):*

i might be confused by this wording, this flag means that there is exactly 1 wrap object (no more or fewer)?
</blockquote></details>

Yes, exactly one, in which case we can compute tangents analytically rather than get initial guess and then test against remaining objects up to 8 times (limit pulled out of thin air AFAICT)

nickbianco (changes requested):
Left another batch of comments. Checking out the new standard solution now, @aymanhab.

> Reviewed 3 of 3 files at r10, 6 of 6 files at r11, 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-:-MuIIunp-P2dkeh9u9Wc:b7snrsu)* status: all files reviewed, 16 unresolved discussions (waiting on @aymanhab, @carmichaelong, and @nickbianco)

___
*[OpenSim/Moco/MocoInverse.cpp, line 50 at r11](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuIG9qB1mZDica3rE5O:-MuIG9qB1mZDica3rE5P:b-2no5ux) ([raw file](https://github.com/opensim-org/opensim-core/blob/75323a273c9a8c20618130357a6fb9f5516c754d/OpenSim/Moco/MocoInverse.cpp#L50)):*
> ```cpp
>     // ----------------
>     Model model = get_model().process(getDocumentDirectory());
>     model.updDisplayHints().disableVisualization();
> ```

This shouldn't be necessary since we set this flag within MocoProblemRep. Unless there's another reason it's here?
___
*[OpenSim/Simulation/Wrap/WrapCylinder.cpp, line 667 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-Mtj4MPp1jZ8WH_z2nnT:-MuIIBvC6G9CJbFZY7fk:bfw8yd3) ([raw file](https://github.com/opensim-org/opensim-core/blob/d4ccffff26830b5070ed9bb7361acefb7bf8387b/OpenSim/Simulation/Wrap/WrapCylinder.cpp#L667)):*

Because if visualization is enabled we fall through to the code that generates intermediate wrapping points for visualization
</blockquote></details>

Ah, I see. Would it not be better to wrap the code below in an `else` statement and get rid of the `return` call?
___
*[OpenSim/Simulation/Wrap/WrapCylinder.cpp, line 434 at r11](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuIH6rnBKUndDwYYzBN:-MuIH6rnBKUndDwYYzBO:by4qg6) ([raw file](https://github.com/opensim-org/opensim-core/blob/75323a273c9a8c20618130357a6fb9f5516c754d/OpenSim/Simulation/Wrap/WrapCylinder.cpp#L434)):*
> ```cpp
> 
>         dot1 = Mtx::DotProduct(3, r1am, r2am);
>         dot2 = Mtx::DotProduct(3, r1am, r2bm);
> ```

These could be replaced by Simbody operations too. However, if this is off-topic for this PR or you're planning do that later then feel free to ignore.
___
*[OpenSim/Simulation/Wrap/WrapCylinder.cpp, line 489 at r11](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuIHKFvBbIXCF0srkG6:-MuIHKFvBbIXCF0srkG7:brke2j3) ([raw file](https://github.com/opensim-org/opensim-core/blob/75323a273c9a8c20618130357a6fb9f5516c754d/OpenSim/Simulation/Wrap/WrapCylinder.cpp#L489)):*
> ```cpp
> 
>     plane_normal = UnitVec3(l1 % l2);
>     Mtx::Normalize(3, plane_normal, plane_normal);
> ```

Any reason these three usages of `Mtx::Normalize` are not replaced with `normalize()`?
___
*[OpenSim/Simulation/Wrap/WrapEllipsoid.cpp, line 538 at r11](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuIIUHh4DfZdi22EA5R:-MuIIUHh4DfZdi22EA5S:b90lww8) ([raw file](https://github.com/opensim-org/opensim-core/blob/75323a273c9a8c20618130357a6fb9f5516c754d/OpenSim/Simulation/Wrap/WrapEllipsoid.cpp#L538)):*
> ```cpp
>     p1c1 = p1 - aWrapResult.c1;
>     vs = p1p2 % p1c1;
>     Mtx::Normalize(3, vs, vs);
> ```

Replace with Simbody operation?
___
*[OpenSim/Simulation/Wrap/WrapEllipsoid.cpp, line 889 at r11](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuIIWBUBr4ni7G9jAUe:-MuIIWBUBr4ni7G9jAUf:b90lww8) ([raw file](https://github.com/opensim-org/opensim-core/blob/75323a273c9a8c20618130357a6fb9f5516c754d/OpenSim/Simulation/Wrap/WrapEllipsoid.cpp#L889)):*
> ```cpp
> 
>     vsz = ar1 % ar2;
>     Mtx::Normalize(3, vsz, vsz);
> ```

Replace with Simbody operation?
___
*[OpenSim/Simulation/Wrap/WrapSphere.cpp, line 242 at r11](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuII_lJ5G9ZjxywV7IU:-MuII_lJ5G9ZjxywV7IV:byqh5zu) ([raw file](https://github.com/opensim-org/opensim-core/blob/75323a273c9a8c20618130357a6fb9f5516c754d/OpenSim/Simulation/Wrap/WrapSphere.cpp#L242)):*
> ```cpp
>    }
> 
>     Mtx::Normalize(3, p1p2, p1p2);
> ```

Replace with Simbody operation? Same goes for the rest of the calls in this file.

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125)* status: all files reviewed, 15 unresolved discussions (waiting on @aymanhab, @carmichaelong, and @nickbianco)

___
*[OpenSim/Moco/MocoInverse.cpp, line 50 at r11](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuIG9qB1mZDica3rE5O:-MuIL1ipEozlbfRIVMqH:b-tftgga) ([raw file](https://github.com/opensim-org/opensim-core/blob/75323a273c9a8c20618130357a6fb9f5516c754d/OpenSim/Moco/MocoInverse.cpp#L50)):*

This shouldn't be necessary since we set this flag within MocoProblemRep. Unless there's another reason it's here?
</blockquote></details>

Done. I wasn't sure all workflows go through MocoProblemRep as you suggest here, so I'll remove.
___
*[OpenSim/Simulation/Wrap/WrapCylinder.cpp, line 434 at r11](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuIH6rnBKUndDwYYzBN:-MuILKLxA7PESWnNAhJR:b-c2mvcv) ([raw file](https://github.com/opensim-org/opensim-core/blob/75323a273c9a8c20618130357a6fb9f5516c754d/OpenSim/Simulation/Wrap/WrapCylinder.cpp#L434)):*

These could be replaced by Simbody operations too. However, if this is off-topic for this PR or you're planning do that later then feel free to ignore.
</blockquote></details>

Totally agree. I've been slowly replacing these calls with Simbody calls and committing the changes. I hope to retire/remove the Mtx class altogether.

nickbianco (commented):
The new standard solution looks good to me.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-:-MuILt1179y0tejTzOyz:b-nnz428)* status: all files reviewed, 15 unresolved discussions (waiting on @aymanhab, @carmichaelong, and @nickbianco)

aymanhab:
Unrelated sporadic failure on osx, otherwise ready for review @carmichaelong

nickbianco (changes requested):
> Reviewed 3 of 6 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125)* status: 33 of 36 files reviewed, 16 unresolved discussions (waiting on @aymanhab, @carmichaelong, and @nickbianco)

___
*[OpenSim/Simulation/Wrap/WrapCylinder.cpp, line 706 at r13](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuOe0bw4IGQMG0e9qfA:-MuOe0bw4IGQMG0e9qfB:b-5mubnx) ([raw file](https://github.com/opensim-org/opensim-core/blob/130c74058eb4e1c549ae00dcfe76d71510041db1/OpenSim/Simulation/Wrap/WrapCylinder.cpp#L706)):*
> ```cpp
>                 if (did_adjust_r1 || did_adjust_r2) {
>                     iterations++;
>                     goto restart_spiral_wrap;
> ```

Here's another! Same comment as below.
___
*[OpenSim/Simulation/Wrap/WrapSphere.cpp, line 253 at r13](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuObZOs7gV6QPERFIZX:-MuObZOs7gV6QPERFIZY:b-3k4wvp) ([raw file](https://github.com/opensim-org/opensim-core/blob/130c74058eb4e1c549ae00dcfe76d71510041db1/OpenSim/Simulation/Wrap/WrapSphere.cpp#L253)):*
> ```cpp
>         // been copied into aWrapResult (and not yet overwritten). So
>         // just go directly to calc_path.
>         goto calc_path;
> ```

Oh wow, an actual `goto` statement. My gut reaction out-of-context is to try to avoid this, but not sure how messy things could get. I know there's some valid `goto` use cases but my feeling is this could be avoided.

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125)* status: 33 of 36 files reviewed, 16 unresolved discussions (waiting on @aymanhab, @carmichaelong, and @nickbianco)

___
*[Applications/Forward/test/testForward.cpp, line 176 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MtxtfJd6WvPWvPpxtRh:-MuMV-vX15H8aoE8kBVa:bsvoyq3) ([raw file](https://github.com/opensim-org/opensim-core/blob/a8bd62a2e45db66de28a168e9175fc124c7ac760/Applications/Forward/test/testForward.cpp#L176)):*

same as above too
</blockquote></details>

See comments above.
___
*[OpenSim/Simulation/Wrap/WrapCylinder.cpp, line 667 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-Mtj4MPp1jZ8WH_z2nnT:-MuRYHn9CRKUtksuMjeJ:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/d4ccffff26830b5070ed9bb7361acefb7bf8387b/OpenSim/Simulation/Wrap/WrapCylinder.cpp#L667)):*

Ah, I see. Would it not be better to wrap the code below in an `else` statement and get rid of the `return` call?
</blockquote></details>

Done.
___
*[OpenSim/Simulation/Wrap/WrapCylinder.cpp, line 706 at r13](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuOe0bw4IGQMG0e9qfA:-MuRYJVYDvI0YGV2KFGe:b84zxqh) ([raw file](https://github.com/opensim-org/opensim-core/blob/130c74058eb4e1c549ae00dcfe76d71510041db1/OpenSim/Simulation/Wrap/WrapCylinder.cpp#L706)):*

Here's another! Same comment as below.
</blockquote></details>

There's a reason nobody over the last 10+ years wanted to touch this to eliminate this goto line. I have no plans to refactor the code as it is error prone and has lots of heuristics with no test cases. This PR is exclusively to improve performance in common cases, so I'd avoid general code cleanup if there're potential side-effects as is the case here. 
___
*[OpenSim/Simulation/Wrap/WrapSphere.cpp, line 253 at r13](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuObZOs7gV6QPERFIZX:-MuRZ_io1ruJ3_iOZsDw:b-96lmew) ([raw file](https://github.com/opensim-org/opensim-core/blob/130c74058eb4e1c549ae00dcfe76d71510041db1/OpenSim/Simulation/Wrap/WrapSphere.cpp#L253)):*

Oh wow, an actual `goto` statement. My gut reaction out-of-context is to try to avoid this, but not sure how messy things could get. I know there's some valid `goto` use cases but my feeling is this could be avoided.
</blockquote></details>

See comment above regarding code cleanup/refactoring and why I'm avoiding it in this PR
___
*[OpenSim/Tools/Test/testControllers.cpp, line 415 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MtyPgNH9TSE0rh5T13f:-MuMY2Bs-uwdTUEYRcVP:bq4wfr7) ([raw file](https://github.com/opensim-org/opensim-core/blob/a8bd62a2e45db66de28a168e9175fc124c7ac760/OpenSim/Tools/Test/testControllers.cpp#L415)):*

should the standard be updated instead of tolerance changed?
</blockquote></details>

See comments above

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125)* status: 27 of 37 files reviewed, 16 unresolved discussions (waiting on @aymanhab, @carmichaelong, and @nickbianco)

___
*[OpenSim/Simulation/Wrap/WrapCylinder.cpp, line 489 at r11](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuIHKFvBbIXCF0srkG6:-MuTZfTk6bya95RUqVL6:b-1226x6) ([raw file](https://github.com/opensim-org/opensim-core/blob/75323a273c9a8c20618130357a6fb9f5516c754d/OpenSim/Simulation/Wrap/WrapCylinder.cpp#L489)):*

Any reason these three usages of `Mtx::Normalize` are not replaced with `normalize()`?
</blockquote></details>

Moved Mtx::Normalize to WrapMath::Normalize which is a bit different from normalize() in that it returns the magnitude  and that it zeros the vector norm if small. All other methods of Mtx were inlined/substituted with simbody methods

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125)* status: 27 of 37 files reviewed, 16 unresolved discussions (waiting on @carmichaelong and @nickbianco)

___
*[OpenSim/Simulation/Wrap/WrapEllipsoid.cpp, line 538 at r11](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuIIUHh4DfZdi22EA5R:-MuT_TDmAW6_BS9mioTH:bq4wfr7) ([raw file](https://github.com/opensim-org/opensim-core/blob/75323a273c9a8c20618130357a6fb9f5516c754d/OpenSim/Simulation/Wrap/WrapEllipsoid.cpp#L538)):*

Replace with Simbody operation?
</blockquote></details>

See comments above
___
*[OpenSim/Simulation/Wrap/WrapSphere.cpp, line 242 at r11](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuII_lJ5G9ZjxywV7IU:-MuT_VyaC4nxzfT0WI_z:byxqf4o) ([raw file](https://github.com/opensim-org/opensim-core/blob/75323a273c9a8c20618130357a6fb9f5516c754d/OpenSim/Simulation/Wrap/WrapSphere.cpp#L242)):*

Replace with Simbody operation? Same goes for the rest of the calls in this file. 
</blockquote></details>

Comments above.

nickbianco (changes requested):
@aymanhab, left a few more comments. Totally agree about not changing the `goto` stuff, per our dev meeting discussion. However, I do have some thoughts related to potentially removing or at least renaming `Normalize`. Let me know what you think. 

> Reviewed 1 of 6 files at r13, 9 of 9 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-:-MuX3e353Rw9x9bHMHlf:b76ca4y)* status: all files reviewed, 10 unresolved discussions (waiting on @aymanhab and @carmichaelong)

___
*[OpenSim/Simulation/Wrap/WrapEllipsoid.cpp, line 571 at r14](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuX1_T4BVe_-AvOcoNI:-MuX1_T4BVe_-AvOcoNJ:bykkng8) ([raw file](https://github.com/opensim-org/opensim-core/blob/6c0a90ccd79bd5df616c3452a2447e8a4084d03b/OpenSim/Simulation/Wrap/WrapEllipsoid.cpp#L571)):*
> ```cpp
>         WrapMath::Normalize(r2w2, r2w2);
> 
>         if ((~r1p1*r1w1) > 0.0 || (~r2p2*r2w2) > 0.0)
> ```

Ah,  these "greater-than-zero" conditionals are probably why `Normalize` has the if-statement. My preference would be to remove the if-statement in the function (actually, replace the function with Simbody's `normalize()`) and replace these conditionals with `(~r1p1*r1w1) > SimTK::Eps`. `Normalize` is used in many places, so this change could change results when vector magnitudes are close to zero, but this change should be very very small since the comparison is against `SimTK::Eps`anyway.
___
*[OpenSim/Simulation/Wrap/WrapMath.h, line 79 at r14](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuX-ptl61Sypi8SkN0d:-MuX-ptl61Sypi8SkN0e:bjw6b6s) ([raw file](https://github.com/opensim-org/opensim-core/blob/6c0a90ccd79bd5df616c3452a2447e8a4084d03b/OpenSim/Simulation/Wrap/WrapMath.h#L79)):*
> ```objc
>      * Normalize a vector.
>      *
>      * If aV has a magnitude of zero, all elements of rV are set to 0.0.
> ```

Do you have a sense if the if-statement in this function is necessary? For example, is the output ever checked to be equality with zero? I don't like that the zeroing out here is hidden within the function. If it needs to stay, then I would advocate for renaming for clarity (i.e., `NormalizeOrZero`, etc).

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125)* status: 31 of 37 files reviewed, 10 unresolved discussions (waiting on @aymanhab, @carmichaelong, and @nickbianco)

___
*[OpenSim/Simulation/Wrap/WrapMath.h, line 79 at r14](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuX-ptl61Sypi8SkN0d:-MuXN2Nt37b8WLEIpTX9:beu4ut) ([raw file](https://github.com/opensim-org/opensim-core/blob/6c0a90ccd79bd5df616c3452a2447e8a4084d03b/OpenSim/Simulation/Wrap/WrapMath.h#L79)):*

Do you have a sense if the if-statement in this function is necessary? For example, is the output ever checked to be equality with zero? I don't like that the zeroing out here is hidden within the function. If it needs to stay, then I would advocate for renaming for clarity (i.e., `NormalizeOrZero`, etc).
</blockquote></details>

It's impossible to guess in general because it depends on what arguments are passed in, and that depends on Model configurations. I totally agree with renaming and just pushed a commit for that.

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125)* status: 31 of 37 files reviewed, 10 unresolved discussions (waiting on @aymanhab, @carmichaelong, and @nickbianco)

___
*[OpenSim/Simulation/Wrap/WrapMath.h, line 79 at r14](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuX-ptl61Sypi8SkN0d:-MuXTqD69W_TlOyuFE9c:b-6xpigq) ([raw file](https://github.com/opensim-org/opensim-core/blob/6c0a90ccd79bd5df616c3452a2447e8a4084d03b/OpenSim/Simulation/Wrap/WrapMath.h#L79)):*

It's impossible to guess in general because it depends on what arguments are passed in, and that depends on Model configurations. I totally agree with renaming and just pushed a commit for that.
</blockquote></details>

Throwing an exception at the else clause to see if it's reached, causes 4 of our test cases (which very thinly cover wrapping) to fail, so it appears this is not a rarity and suggest it's safer to leave as is.

carmichaelong (commented):
Thanks @aymanhab, overall looking good. I noticed some places in which `normalize()` wasn't reverted back to `WrapMath::NormalizeOrZero()` (btw this is a much clearer name) though some of these cases are OK when the logic beforehand ensures a vector that isn't too small.

That being said, before checking/noting all those cases, I think @nickbianco's comment in WrapEllipsoid.cpp L571 should be addressed first, just in case that this was the source of the test errors and could simplify the code to using `normalize()` more.

> Reviewed 1 of 3 files at r10, 3 of 9 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-:-MuXvVMd2nDK3qBjbtDr:b-4hhw93)* status: 31 of 37 files reviewed, 2 unresolved discussions (waiting on @aymanhab and @nickbianco)

aymanhab (commented):
Thanks @carmichaelong I responded to @nickbianco comment below.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-:-MulaIrx2-dydumyQsy6:b-urqkvw)* status: 31 of 37 files reviewed, 2 unresolved discussions (waiting on @nickbianco)

___
*[OpenSim/Simulation/Wrap/WrapEllipsoid.cpp, line 571 at r14](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuX1_T4BVe_-AvOcoNI:-Mul_5IbElp8evfhwrYW:b4t6wau) ([raw file](https://github.com/opensim-org/opensim-core/blob/6c0a90ccd79bd5df616c3452a2447e8a4084d03b/OpenSim/Simulation/Wrap/WrapEllipsoid.cpp#L571)):*

Ah,  these "greater-than-zero" conditionals are probably why `Normalize` has the if-statement. My preference would be to remove the if-statement in the function (actually, replace the function with Simbody's `normalize()`) and replace these conditionals with `(~r1p1*r1w1) > SimTK::Eps`. `Normalize` is used in many places, so this change could change results when vector magnitudes are close to zero, but this change should be very very small since the comparison is against `SimTK::Eps`anyway.
</blockquote></details>

@nickbianco  Typically in this code sign is used to decide "sidedness" (e.g. if dot product is +ve is one scenario, -ve is a different scenario) as such adding slop of SimTK::Eps would skew the test incorrectly to one side. That's why I'd try to avoid modification to the semantics at all. If you prefer I can undo all the normalize introductions and restore NormalizeOrZero so it's clear we're only changing name and no functionality whatsoever. Please let me know what you think.

nickbianco (changes requested):
Thanks @aymanhab. I also agree `NormalizeOrZero` is much clearer. However, I've made one final argument for removing it, since we have a chance to remove all remnants of the `Mtx` code here. I think it's at least worth trying removing the "zero" branch of that function to see how our existing tests change. 

It might be easier to discuss some of this at dev meeting tomorrow.

> Reviewed 6 of 6 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-:-MulqUaPAqUSKCfGbhI0:b-v4b2xq)* status: all files reviewed, 4 unresolved discussions (waiting on @aymanhab)

___
*[OpenSim/Simulation/Wrap/WrapEllipsoid.cpp, line 571 at r14](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuX1_T4BVe_-AvOcoNI:-Mull5dr-Gn4-eIH6OET:b-nm8xdm) ([raw file](https://github.com/opensim-org/opensim-core/blob/6c0a90ccd79bd5df616c3452a2447e8a4084d03b/OpenSim/Simulation/Wrap/WrapEllipsoid.cpp#L571)):*
I'd have to spend some time to better understand what's happening here, but wouldn't the "zero" part of `NormalizeOrZero` also being skewing results? 

> If you prefer I can undo all the normalize introductions and restore NormalizeOrZero so it's clear we're only changing name and no functionality whatsoever.

I guess we should be consistent with whatever we do. If we're concerned that removing the NormalizeOrZero function changes stuff, then we shouldn't replace it with `normalize()` anywhere. But my intuition is that the "zero" part of this function isn't needed everywhere, so I don't like that it's baked into this function. If we have a chance to remove all the weird stuff from `Mtx` here it's worth a shot. 
___
*[OpenSim/Simulation/Wrap/WrapMath.h, line 79 at r14](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuX-ptl61Sypi8SkN0d:-MulnqcS1c614WHHSzOs:bf1g1d4) ([raw file](https://github.com/opensim-org/opensim-core/blob/6c0a90ccd79bd5df616c3452a2447e8a4084d03b/OpenSim/Simulation/Wrap/WrapMath.h#L79)):*

Throwing an exception at the else clause to see if it's reached, causes 4 of our test cases (which very thinly cover wrapping) to fail, so it appears this is not a rarity and suggest it's safer to leave as is.
</blockquote></details>

It might be reaching the "else" statement in situations where it's not really necessary to zero out the vector. It would be better to try removing the "else" branch and see if it changes any tests results. We could also specifically check if the if-statement in WrapEllipsoid changes too.
___
*[OpenSim/Simulation/Wrap/WrapMath.cpp, line 87 at r15](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MulptVUE2QPJjDKYoeU:-MulptVUE2QPJjDKYoeV:b-sm9640) ([raw file](https://github.com/opensim-org/opensim-core/blob/f0d1d809d9e8191fd535a47fa62eee3680f36b94/OpenSim/Simulation/Wrap/WrapMath.cpp#L87)):*
> ```cpp
>     double denom = cross_prod.normSqr();
> 
>     if (EQUAL_WITHIN_ERROR(denom,0.0)) {
> ```

Here's another case where it might be better to check against `SimTK::Eps`.
___
*[OpenSim/Simulation/Wrap/WrapSphere.cpp, line 319 at r15](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-Muln4J69FO9_hhpQgVm:-Muln4J69FO9_hhpQgVn:b-h8kg9t) ([raw file](https://github.com/opensim-org/opensim-core/blob/f0d1d809d9e8191fd535a47fa62eee3680f36b94/OpenSim/Simulation/Wrap/WrapSphere.cpp#L319)):*
> ```cpp
>     r1bm.normalize();
>     r2am.normalize();
>     r2bm.normalize();
> ```

Per my comment above: if we're not using `normalize()` everywhere, then we shouldn't use it here.

aymanhab (commented):
Good suggestion @nickbianco and it actually doesn't break tests, however we know our coverage of wrapping is thin and I can artificially cause coincident points to be passed in to the function causing division by zero, so I'll leave the method name as NormalizeOrZero, restore it everywhere in wrapping code for consistency as you suggested and we should be done with this PR without introducing any unexpected side effects due to other unrelated cleanup. We can revisit later in another PR but Mtx is now gone and WrapMath is localized and reduced.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-:-MuqoQlO0c5_Vk7CF_Bi:bldh4p9)* status: 33 of 37 files reviewed, 4 unresolved discussions (waiting on @aymanhab and @nickbianco)

nickbianco (approved):
Sounds good @aymanhab. LGTM.

> Reviewed 3 of 4 files at r17, 1 of 1 files at r18, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-:-MusdI9V8OgcryzYNF9R:b-q56e0j)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aymanhab)

nickbianco (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125)* status: 36 of 37 files reviewed, 1 unresolved discussion (waiting on @aymanhab and @nickbianco)

___
*[OpenSim/Actuators/ModelProcessor.h, line 129 at r19](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-MuwcLMs0GkCe8RsGhQK:-MuwcLMt497wf1lSbhNa:b-w3e52s) ([raw file](https://github.com/opensim-org/opensim-core/blob/6bb98445b73c79dc1891f5668550ad88490f6b7d/OpenSim/Actuators/ModelProcessor.h#L129)):*
> ```objc
>             Model modelFromFile(path);
>             model = std::move(modelFromFile);
>             model.updDisplayHints().disableVisualization();
> ```

Actually, I just noticed this. I don't think we need this here, since we already display visualization within Moco, and users creating a Model with ModelProcessor might want visualization enabled.

nickbianco (commented):
@aymanhab, I just noticed something, left one last quick comment.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-:-Muwcavi24uKRKknU87c:b-plyzuc)* status: 36 of 37 files reviewed, 1 unresolved discussion (waiting on @aymanhab and @nickbianco)

aymanhab:
> @aymanhab, I just noticed something, left one last quick comment.
> 
> > _[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-:-Muwcavi24uKRKknU87c:b-plyzuc)_ status: 36 of 37 files reviewed, 1 unresolved discussion (waiting on @aymanhab and @nickbianco)

Sure, undone.

nickbianco (approved):
:lgtm:

> Reviewed 1 of 1 files at r19, 1 of 1 files at r20, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3125#-:-Mux4blGAxtxknc5qs-a:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @aymanhab)

aymanhab:
Thanks @nickbianco 👍 much appreciated.

PR #3129, opened 2022-01-29
Add Frame Orientation Constraint

zhengsizehrb:
This is a custom constraint created by Simon Jeng to constrain the relative angle between two frames. This constraint is of MocoPathConstraint class, and has been briefly discussed in the Moco forum "https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=1815&t=14287&p=0&start=0&view=&sid=2a5bcf0062e1178c20421d7308d8d5e4". 
The constraint was developed for two-dimensional models in the sagittal plane (the plane perpendicular to the z axis, exactly), because 2D is enough in my research. If this constraint is needed in Moco, we can improve it and adapt it to three-dimensional models.

This is the first time for me to use Github and submit pull requests. If more information is required, pleas let me know.

Thanks,
Simon Jeng

Fixes issue #<issue_number>

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3129)


nickbianco:
Hi @zhengsizehrb, thanks for submitting your new constraint! 

The first step is change this from a plug-in to a new component within the OpenSim API. To do this, instead of adding `RegisterTypes_osimSimonFrameOrientationConstraint.cpp` and `RegisterTypes_osimSimonFrameOrientationConstraint.h`, you can instead just update `RegisterTypes_osimMoco.cpp` with you new constraint type. Similarly, instead of adding `osimSimonFrameOrientationConstraintDLL.h`, just update `osimMoco.h` to add an include statement for your new type. 

Let's rename `SimonFrameOrientationConstraint` to `MocoFrameOrientationConstraint` to be consistent with the rest of the API.

zhengsizehrb:
Hi @nickbianco, thanks. I have finished this step.

OpenSim/Moco/MocoFrameOrientationConstraint.cpp
```diff
@@ -0,0 +1,109 @@
...
+    int iconstr = 0;
+
+    getModel().realizePosition(state);
+
+    for (const auto& frame_pair : m_frame_pairs) {
+        const auto& frame1_rotation = frame_pair.first->getRotationInGround(state);
+        const auto& frame2_rotation = frame_pair.second->getRotationInGround(state);
+        auto coordaxis = SimTK::CoordinateAxis(2);											//set the number as 2 to represent ZAxis
+        auto frame1_angle = frame1_rotation.convertOneAxisRotationToOneAngle(coordaxis);	//the angle of this frame is about the z-axis
+        auto frame2_angle = frame2_rotation.convertOneAxisRotationToOneAngle(coordaxis);	//the angle of this frame is about the z-axis
```
nickbianco, line 99: These comments will change, but all comments should be above the relevant line, not inline as they are now.

OpenSim/Moco/MocoFrameOrientationConstraint.h
```diff
@@ -0,0 +1,80 @@
...
+    MocoFrameOrientationConstraintPair();
+    MocoFrameOrientationConstraintPair(std::string firstFramePath,
+        std::string secondFramePath, double minimum_angle,
+        double maximum_angle);
+
+private:
+    void constructProperties();
+};
+
+class OSIMMOCO_API MocoFrameOrientationConstraint : public MocoPathConstraint {
```
nickbianco, line 50: Similar to other constraints and MocoGoals, add a comment describing this constraint here.

OpenSim/Moco/MocoFrameOrientationConstraint.cpp
```diff
@@ -0,0 +1,109 @@
...
+void MocoFrameOrientationConstraint::calcPathConstraintErrorsImpl(const SimTK::State& state, SimTK::Vector& errors) const
+{
+    int iconstr = 0;
+
+    getModel().realizePosition(state);
+
+    for (const auto& frame_pair : m_frame_pairs) {
+        const auto& frame1_rotation = frame_pair.first->getRotationInGround(state);
+        const auto& frame2_rotation = frame_pair.second->getRotationInGround(state);
+        auto coordaxis = SimTK::CoordinateAxis(2);											//set the number as 2 to represent ZAxis
```
nickbianco, line 97: We don't want to hard-code a specific coordinate axis for the constraint, it would be better to provide users an option for which coordinate axis they want via a property. Alternatively, you could constrain all axes of the two frame with one scalar constraint by converting the orientation error to an angle-axis representation ([like we do in MocoOrientationTrackingGoal](https://github.com/opensim-org/opensim-core/blob/48ba9a0792c30b66bb2b97081f2546d1f627c2bc/OpenSim/Moco/MocoGoal/MocoOrientationTrackingGoal.cpp#L201)). 

Best would be provide an option to do either: if they provide a coordinate axis value, then you only minimize the error for that axis, if not then you use the angle-axis approach. Let me know if this makes sense to you.

OpenSim/Moco/MocoFrameOrientationConstraint.h
```diff
@@ -0,0 +1,80 @@
...
+ *                                                                            *
+ * Unless required by applicable law or agreed to in writing, software        *
+ * distributed under the License is distributed on an "AS IS" BASIS,          *
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.   *
+ * See the License for the specific language governing permissions and        *
+ * limitations under the License.                                             *
+ * -------------------------------------------------------------------------- */
+
+#include "MocoConstraint.h"
+#include <OpenSim/Moco/osimMoco.h>
```
nickbianco, line 21: I don't think this include is necessary since you already have `osimMocoDLL.h` included below.

OpenSim/Moco/MocoFrameOrientationConstraint.h
```diff
@@ -0,0 +1,80 @@
...
+
+public:
+    MocoFrameOrientationConstraint();
+    
+    void addFramePair(MocoFrameOrientationConstraintPair pair) {
+        append_frame_pairs(std::move(pair));
+    }
+    void addFramePair(const std::string& frame1_path,const std::string& frame2_path, double minimum_angle,
+        double maximum_angle) {
+        append_frame_pairs(MocoFrameOrientationConstraintPair(frame1_path,frame2_path, minimum_angle,maximum_angle));
```
nickbianco, line 61: Make sure no line exceed 80 column width (looks like this one does.

OpenSim/Moco/MocoFrameOrientationConstraint.h
```diff
@@ -0,0 +1,80 @@
...
+    OpenSim_DECLARE_CONCRETE_OBJECT(MocoFrameOrientationConstraint, MocoPathConstraint);
+
+public:
+    MocoFrameOrientationConstraint();
+    
+    void addFramePair(MocoFrameOrientationConstraintPair pair) {
+        append_frame_pairs(std::move(pair));
+    }
+    void addFramePair(const std::string& frame1_path,const std::string& frame2_path, double minimum_angle,
+        double maximum_angle) {
```
nickbianco, line 60: Indent this last argument one more time.

OpenSim/Moco/MocoFrameOrientationConstraint.h
```diff
@@ -0,0 +1,80 @@
...
+        append_frame_pairs(MocoFrameOrientationConstraintPair(frame1_path,frame2_path, minimum_angle,maximum_angle));
+    }
+protected:
```
nickbianco, line 63: Add blank line before `protected:`

OpenSim/Moco/MocoFrameOrientationConstraint.h
```diff
@@ -0,0 +1,80 @@
...
+    void initializeOnModelImpl(const Model& model, const MocoProblemInfo&) const override;
+    void calcPathConstraintErrorsImpl(const SimTK::State& state, SimTK::Vector& errors) const override;
+private:
```
nickbianco, line 66: Add blank line before `private:`

OpenSim/Moco/MocoFrameOrientationConstraint.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+//=============================================================================
+//  MocoFrameOrientationConstraint
+//=============================================================================
+
+MocoFrameOrientationConstraint::MocoFrameOrientationConstraint() 
+{
+    constructProperties();
+}
+void MocoFrameOrientationConstraint::initializeOnModelImpl(const Model& model, const MocoProblemInfo&) const
```
nickbianco, line 52: Don't exceed the 80 column width.

OpenSim/Moco/MocoFrameOrientationConstraint.cpp
```diff
@@ -0,0 +1,109 @@
...
+    }
+  
+
+    setNumEquations(nFramePairs);
+    info.setBounds(bounds);
+    const_cast<MocoFrameOrientationConstraint*>(this)->setConstraintInfo(info);
+
+}
+
+void MocoFrameOrientationConstraint::calcPathConstraintErrorsImpl(const SimTK::State& state, SimTK::Vector& errors) const
```
nickbianco, line 88: Don't exceed the 80 column width.

OpenSim/Moco/MocoFrameOrientationConstraint.cpp
nickbianco: In addition, these comments exceed the 80 column width.

OpenSim/Moco/MocoFrameOrientationConstraint.cpp
```diff
@@ -0,0 +1,109 @@
...
+using namespace OpenSim;
+
+
+MocoFrameOrientationConstraintPair::MocoFrameOrientationConstraintPair() {
+    constructProperties();
+}
+
+MocoFrameOrientationConstraintPair::MocoFrameOrientationConstraintPair(
+    std::string frame1Path, std::string frame2Path,
+    double minimum_angle, double maximum_angle) {
```
nickbianco, line 29: Indent these argument lines one more time.

OpenSim/Moco/MocoFrameOrientationConstraint.cpp
```diff
@@ -0,0 +1,109 @@
...
+    constructProperty_minimum_angle(-SimTK::Infinity);
+    constructProperty_maximum_angle(SimTK::Infinity);
+}
+
+//=============================================================================
+//  MocoFrameOrientationConstraint
+//=============================================================================
+
+MocoFrameOrientationConstraint::MocoFrameOrientationConstraint() 
+{
```
nickbianco, line 49: Move this opening brace up to the previous line (that is our style convention).

OpenSim/Moco/MocoFrameOrientationConstraint.cpp
```diff
@@ -0,0 +1,109 @@
...
+    constructProperties();
+}
+void MocoFrameOrientationConstraint::initializeOnModelImpl(const Model& model, const MocoProblemInfo&) const
+{
```
nickbianco, line 53: Move brace up one line.

OpenSim/Moco/MocoFrameOrientationConstraint.cpp
```diff
@@ -0,0 +1,109 @@
...
+  
+
+    setNumEquations(nFramePairs);
+    info.setBounds(bounds);
+    const_cast<MocoFrameOrientationConstraint*>(this)->setConstraintInfo(info);
+
+}
+
+void MocoFrameOrientationConstraint::calcPathConstraintErrorsImpl(const SimTK::State& state, SimTK::Vector& errors) const
+{
```
nickbianco, line 89: Move brace up one line.

OpenSim/Moco/MocoFrameOrientationConstraint.h
```diff
@@ -0,0 +1,80 @@
...
+class OSIMMOCO_API MocoFrameOrientationConstraint : public MocoPathConstraint {
+    OpenSim_DECLARE_CONCRETE_OBJECT(MocoFrameOrientationConstraint, MocoPathConstraint);
+
+public:
+    MocoFrameOrientationConstraint();
+    
+    void addFramePair(MocoFrameOrientationConstraintPair pair) {
+        append_frame_pairs(std::move(pair));
+    }
+    void addFramePair(const std::string& frame1_path,const std::string& frame2_path, double minimum_angle,
```
nickbianco, line 59: Don't exceed the 80 column width.

nickbianco (changes requested):
Thanks @zhengsizehrb! I've started reviewing the actual code now. Most of the comments are related to code formatting and style changes, but there's one major change I've suggested related to which angle axis is being constrained.

nickbianco:
@zhengsizehrb, the style checker threw an error because tabs were detected in the code. Make sure to replace all tabs with spaces (most IDEs have an option to do this easily).

OpenSim/Moco/MocoFrameOrientationConstraint.cpp
zhengsizehrb: Hi @nickbianco ,

> Best would be provide an option to do either: if they provide a coordinate axis value, then you only minimize the error for that axis, if not then you use the angle-axis approach. Let me know if this makes sense to you.

This has been done, and the coordinate axis has been set as an optional property.

OpenSim/Moco/MocoFrameOrientationConstraint.h
zhengsizehrb: Nick, all these issues you proposed have been corrected.

OpenSim/Moco/MocoFrameOrientationConstraint.h
```diff
@@ -0,0 +1,108 @@
...
+        "minimum and maximum bounds.");
+     OpenSim_DECLARE_OPTIONAL_PROPERTY(coordinate_axis, int,
+        "(optional) If provided, the coordinate axis is defined "
+        "0 represents x-aixs, 1 represents y-axis, 2 represents z-axis,"
+        "3 represents all the axis.");
+    
+    void constructProperties();
+    mutable std::vector<std::pair<SimTK::ReferencePtr<const Frame>,
+        SimTK::ReferencePtr<const Frame>>> m_frame_pairs;
+    mutable int m_coordinateAxis;
```
nickbianco, line 103: This will need to become a vector of integers since `coordinate_axis` will be set per frame pair. You could also create a vector of `bool`s to store which frame pairs will minimize the full orientation error.

OpenSim/Moco/MocoFrameOrientationConstraint.cpp
```diff
@@ -0,0 +1,127 @@
...
+using namespace OpenSim;
+
+
+MocoFrameOrientationConstraintPair::MocoFrameOrientationConstraintPair() {
+    constructProperties();
+}
+
+MocoFrameOrientationConstraintPair::MocoFrameOrientationConstraintPair(
+        std::string frame1Path, std::string frame2Path, double minimum_angle, 
+        double maximum_angle) {
```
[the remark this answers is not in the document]
nickbianco: Per the comments in the header file, update the constructor definition here:

```
MocoFrameOrientationConstraintPair::MocoFrameOrientationConstraintPair(
        const std::string& frame1Path, const std::string& frame2Path, double minimumError, 
        double maximumError, const std::string& coordinateAxis = {})
```

OpenSim/Moco/MocoFrameOrientationConstraint.cpp
```diff
@@ -0,0 +1,127 @@
...
+        m_frame_pairs.emplace_back(&frame1, &frame2);
+
+        const double& minimum = get_frame_pairs(i).get_minimum_angle();
+        const double& maximum = get_frame_pairs(i).get_maximum_angle();
+        
+        OPENSIM_THROW_IF(minimum > maximum, Exception,
+            "Expected the minimum angle for this frame pair "
+            "to be less than or equal to the maximum angle, "
+            "but they are {} and {}, respectively.",
+            minimum, maximum);
```
nickbianco, line 77: Update this error message to reflect that these are angle errors, not angles.

OpenSim/Moco/MocoFrameOrientationConstraint.cpp
```diff
@@ -0,0 +1,127 @@
...
+        bounds.emplace_back(minimum, maximum);
+    }
+    m_coordinateAxis = get_coordinate_axis();
```
nickbianco, line 80: Since this is optional, you'll want to check if it's empty (as that will determine if we minimize the 3D orientation error). You can check that the property is empty like so `getProperty_coordinate_axis().empty()`. You can also update the new boolean vector to use when calculating the path constraint error.

OpenSim/Moco/MocoFrameOrientationConstraint.cpp
```diff
@@ -0,0 +1,127 @@
...
+    OPENSIM_THROW_IF(((m_coordinateAxis!=0) && (m_coordinateAxis != 1) && 
+        (m_coordinateAxis != 2) && (m_coordinateAxis != 3)), Exception,
+        "The coordinate axis should be 0,1,2,or 3.", m_coordinateAxis);
```
nickbianco, line 83: These checks will be moved into the loop after moving the `coordinate_axis` property to `MocoFrameOrientationConstraint`. Since this property will be a string type, the check will need to change. You can make this check using `checkPropertyValueIsInSet`. It should look something like this (only if it's not empty):

```
checkPropertyValueIsInSet(
            get_frame_pairs(i).getProperty_coordinate_axis(), {"x", "y", "z"});
```

Then you'll need to assign the appropriate index to the coordinate index vector based on what axis is specified.

OpenSim/Moco/MocoFrameOrientationConstraint.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+}
+
+void MocoFrameOrientationConstraint::calcPathConstraintErrorsImpl(
+        const SimTK::State& state, SimTK::Vector& errors) const {
+    int iconstr = 0;
+
+    getModel().realizePosition(state);
+
+    for (const auto& frame_pair : m_frame_pairs) {
```
nickbianco, line 97: You'll actually need an index for this for loop now, so you can axis the individual elements of the new coordinate index and boolean vectors.

OpenSim/Moco/MocoFrameOrientationConstraint.cpp
```diff
@@ -0,0 +1,127 @@
...
+                m_coordinateAxis == 2) {
+            
+            auto frame1_angle = 
+                frame1_rotation.convertOneAxisRotationToOneAngle(
+                        SimTK::CoordinateAxis(m_coordinateAxis));
+            auto frame2_angle = 
+                    frame2_rotation.convertOneAxisRotationToOneAngle(
+                            SimTK::CoordinateAxis(m_coordinateAxis));
+            auto relative_angle = frame2_angle - frame1_angle;
+            errors[iconstr++] = relative_angle;
```
nickbianco, line 113: This error should be squared. You can update as follows:

```
auto error = frame2_angle - frame1_angle;
errors[iconstr++] = error * error;
```

OpenSim/Moco/MocoFrameOrientationConstraint.cpp
```diff
@@ -0,0 +1,127 @@
...
+        }
+        else if(m_coordinateAxis == 3) {
```
nickbianco, line 115: This will now be the top level if-statement where you just check the boolean vector and the other branch can be handled with `else` (so you can avoid the cumbersome or-statements).

OpenSim/Moco/MocoFrameOrientationConstraint.cpp
```diff
@@ -0,0 +1,127 @@
...
+    constructProperties();
+}
+void MocoFrameOrientationConstraint::initializeOnModelImpl(const Model& model, 
+        const MocoProblemInfo&) const {
+
+    int nFramePairs = getProperty_frame_pairs().size();
+    MocoConstraintInfo info;
+    std::vector<MocoBounds> bounds;
+
+    for (int i = 0; i < nFramePairs; ++i) {
```
nickbianco, line 59: You could use the concise for loop style here: `for (const auto& frame_pair : m_frame_pairs)`.

nickbianco (changes requested):
Hi @zhengsizehrb, thanks for the quick update! I have the next round of comments for you. I know it's a lot, but most of it is style/implementation preferences to match the rest of the Moco codebase. Once these are addressed, we'll probably do one last round of final touches and then we should be done.

nickbianco:
Hi @zhengsizehrb, just checking in on this to see if you needed any help.

zhengsizehrb:
Hi Nick,

These days I have not been able to access GitHub in my country. Maybe it's a problem of the server and I need to find ways to tackle this. I am also busy with a paper on predict manual lifting using Moco, and considering which journal is would be appropriate.
I will be back to continue coding MocoFrameOrientationConstraint after finishing these things. It will not be long.

Simon



At 2022-02-25 06:16:52, "Nicholas Bianco" ***@***.***> wrote:

Hi @zhengsizehrb, just checking in on this to see if you needed any help.

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

nickbianco:
Sounds good @zhengsizehrb! Let me know if you have any questions.

nickbianco:
@zhengsizehrb, please let me know if you plan on completing this goal. Otherwise, I will close this PR.

Issue #3061, opened 2021-08-16
Improvements to MocoOutputGoal


nickbianco:
Several improvements should be made for MocoOutputGoal, including:

- Option to specify the exponent of the minimized output
- Option to minimize or maximize an Output
- Option to minimize an element of a `SimTK::Vec3` or `SimTK::Vector` output
- Bug fix to regex when selecting model-level output quantities
- Add option to minimize the final Output value or add separate goal for this (e.g., MocoFinalOutputGoal)

Related discussions/issues:
- https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=1815&t=13629&p=39578 
- Issue #3055 
- Issue #3139

PR #3131, opened 2022-02-01
[Moco] Updates to MocoOutputGoal

nickbianco:
Fixes issue #3061

### Brief summary of changes
- Support using `SimTK::Vec3` and `SimTK::SpatialVec` Output types
- Users can minimize vector norms or specific elements in these vector types
- Users can provide a custom exponent for integrand values

### Testing I've completed
- Added a test to `testMocoGoals.cpp`

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3131)


OpenSim/Moco/MocoGoal/MocoOutputGoal.h
```diff
@@ -56,6 +59,19 @@ class OSIMMOCO_API MocoOutputGoal : public MocoGoal {
...
+    int getExponent() const { return get_exponent(); }
+
+    /** Set the index to the value to be minimized when a vector type
+    Output is specified. For SpatialVec Outputs, indices 0, 1, and 2
+    refer to the rotational components and indices 3, 4, and 5 refer
+    to the translational components. A value of -1 indicates to
+    minimize the vector norm. */
+    void setOutputIndex(int index) { set_output_index(index); }
+    int getOutputIndex() const { return get_output_index(); }
+
```
aymanhab, line 74: I'm a bit worried that the decoupling of the vector and index could lead to confusion e.g setting index when not using vector, is it handled the same was as outputs elsewhere?

OpenSim/Moco/MocoGoal/MocoOutputGoal.h
```diff
@@ -72,11 +88,30 @@ class OSIMMOCO_API MocoOutputGoal : public MocoGoal {
...
             "Divide by the model's total mass (default: false)");
+    OpenSim_DECLARE_PROPERTY(exponent, int,
+            "The exponent applied to the output value in the integrand "
+            "(default: 1).");
+    OpenSim_DECLARE_PROPERTY(output_index, int,
+            "The index to the value to be minimized when a vector type "
+            "Output is specified. For SpatialVec Outputs, indices 0, 1, "
+            "and 2 refer to the rotational components and indices 3, 4, "
+            "and 5 refer to the translational components. A value of -1 "
+            "indicates to minimize the vector norm (default: -1).");
```
aymanhab, line 99: I understand this "when a vector types" documentation, but can we actually enforce it?

OpenSim/Moco/MocoGoal/MocoOutputGoal.h
```diff
@@ -72,11 +88,30 @@ class OSIMMOCO_API MocoOutputGoal : public MocoGoal {
...
     void constructProperties();
 
-    mutable SimTK::ReferencePtr<const Output<double>> m_output;
+    enum DataType {
+        Type_double,
+        Type_Vec3,
+        Type_SpatialVec,
+    };
+    mutable DataType m_data_type;
```
aymanhab, line 107: Any reason to introduce DataType as opposed to using the SimTK types directly?

aymanhab (commented):
Just clearing up a couple issues

nickbianco (commented):
Thanks @aymanhab, made one update and responded to the other comments.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3131#-:-Mv0oll3CMjQDl9Oi8fN:bfbhpja)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @aymanhab)

___
*[OpenSim/Moco/MocoGoal/MocoOutputGoal.h, line 74 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3131#-MurJDGFuNOJbTn68V8n-r1-74:-Mv0kDgr4Ddt-rtx8gGv:b-mihwm7) ([raw file](https://github.com/opensim-org/opensim-core/blob/db8e733da6906218668fd2a3ce84dcb613f049e0/OpenSim/Moco/MocoGoal/MocoOutputGoal.h#L74)):*

I'm a bit worried that the decoupling of the vector and index could lead to confusion e.g setting index when not using vector, is it handled the same was as outputs elsewhere?
</blockquote></details>

I throw an exception if the user provides an output index while also providing an output path for a double type Output (I updated the comment to clarify). Right now you only need to specify the Output path and not the Output type, and I'd like to keep it that way. 
___
*[OpenSim/Moco/MocoGoal/MocoOutputGoal.h, line 99 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3131#-MurJDGFuNOJbTn68V8n-r1-99:-Mv0oQltAsLGZ-f1ihSU:b2ajsvb) ([raw file](https://github.com/opensim-org/opensim-core/blob/db8e733da6906218668fd2a3ce84dcb613f049e0/OpenSim/Moco/MocoGoal/MocoOutputGoal.h#L99)):*

I understand this "when a vector types" documentation, but can we actually enforce it?
</blockquote></details>

I'm not sure what you mean here. There's nothing to enforce, if a user provides an Output for a vector type and the index is -1, then the vector norm is minimized.
___
*[OpenSim/Moco/MocoGoal/MocoOutputGoal.h, line 107 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3131#-MurJDGFuNOJbTn68V8n-r1-107:-Mv0obXD4b-aAY0A2uNQ:bof3jlg) ([raw file](https://github.com/opensim-org/opensim-core/blob/db8e733da6906218668fd2a3ce84dcb613f049e0/OpenSim/Moco/MocoGoal/MocoOutputGoal.h#L107)):*

Any reason to introduce DataType as opposed to using the SimTK types directly?
</blockquote></details>

This is faster (and safer I think?) for the checks that happen inside the integrand function then comparing SimTK types directly.

aymanhab (changes requested):
> Reviewed 1 of 4 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3131)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @aymanhab and @nickbianco)

___
*[OpenSim/Moco/MocoGoal/MocoOutputGoal.h, line 99 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3131#-MurJDGFuNOJbTn68V8n-r1-99:-Mv5Hc3_A20PUi7yF5dZ:b-cq3zss) ([raw file](https://github.com/opensim-org/opensim-core/blob/db8e733da6906218668fd2a3ce84dcb613f049e0/OpenSim/Moco/MocoGoal/MocoOutputGoal.h#L99)):*

I'm not sure what you mean here. There's nothing to enforce, if a user provides an Output for a vector type and the index is -1, then the vector norm is minimized.
</blockquote></details>

You could consider making the interface such that you specify an output vector and index on the same call so the type is forced by the compiler or index check is made on the same call rather than later but not a showstopper.
___
*[OpenSim/Moco/MocoGoal/MocoOutputGoal.h, line 107 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3131#-MurJDGFuNOJbTn68V8n-r1-107:-Mv5HEzQ3uTCinR0IL16:b-z4mre9) ([raw file](https://github.com/opensim-org/opensim-core/blob/db8e733da6906218668fd2a3ce84dcb613f049e0/OpenSim/Moco/MocoGoal/MocoOutputGoal.h#L107)):*

This is faster (and safer I think?) for the checks that happen inside the integrand function then comparing SimTK types directly.
</blockquote></details>

I understand, just worry about maintaining a parallel set of enums if at some point we add Vec2, Quaternions. Something to keep in mind in case we extend the list further.

aymanhab (commented):
Thanks @nickbianco one higher level question: have you checked how this looks in serialization/deserialization and/or writing outputs to a file where different indices of the same Vector are included?

> Reviewed 2 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3131#-:-Mv5IXmq2qHaJtbVrZcU:b-f7fv0j)* status: all files reviewed, 2 unresolved discussions (waiting on @nickbianco)

nickbianco (commented):
Thanks @aymanhab.

If you want to minimize two indices separately, you will need two instances of MocoOutputGoal, so the different indices will be contained separately in the XML for each individual goal. The properties are strings, bools, and ints, so serialization shouldn't be an issue.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3131#-:-Mv64MF69y_WSdPsPnAP:bgt9wte)* status: all files reviewed, 2 unresolved discussions (waiting on @aymanhab)

___
*[OpenSim/Moco/MocoGoal/MocoOutputGoal.h, line 99 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3131#-MurJDGFuNOJbTn68V8n-r1-99:-Mv62KwqEls_mztLWvr8:b7zyd73) ([raw file](https://github.com/opensim-org/opensim-core/blob/db8e733da6906218668fd2a3ce84dcb613f049e0/OpenSim/Moco/MocoGoal/MocoOutputGoal.h#L99)):*

You could consider making the interface such that you specify an output vector and index on the same call so the type is forced by the compiler or index check is made on the same call rather than later but not a showstopper.
</blockquote></details>

We use a similar approach for `MocoParameter` without issue so far, so I think we're probably okay with the current interface.
___
*[OpenSim/Moco/MocoGoal/MocoOutputGoal.h, line 107 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3131#-MurJDGFuNOJbTn68V8n-r1-107:-Mv61XR-Bx_TaDWDMAwf:bnhhe4v) ([raw file](https://github.com/opensim-org/opensim-core/blob/db8e733da6906218668fd2a3ce84dcb613f049e0/OpenSim/Moco/MocoGoal/MocoOutputGoal.h#L107)):*

I understand, just worry about maintaining a parallel set of enums if at some point we add Vec2, Quaternions. Something to keep in mind in case we extend the list further.
</blockquote></details>

It's fairly trivial to add new types. Just update the enum here and the two if-statements in the cpp file.

aymanhab (approved):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3131)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nickbianco)

nickbianco:
Thanks @aymanhab!

PR #3146, opened 2022-02-23
Fixed OpenSim::Arrow ignoring start point

adamkewley:
Minor fix to `OpenSim::Arrow`. The existing implementation ignores the start point and does not account for users entering non-normalized direction vectors (the length of the direction vector is *assumed* to be 1.0, but users could feasibly enter a non-normalized vector via a file/UI).


aymanhab (approved):
Thanks for the fix @adamkewley Actually this class is not used anywhere in our codebase, if you use it you may want to indicate that somewhere or add a dummy test case that uses it so it's not removed by accident.

PR #3147, opened 2022-02-24
Refactored Geometry to reduce number of steps/copies

adamkewley:
Fixes issue (not an issue)

### Brief summary of changes

- Removes intermediate `SimTK::Array_` from `generateDecorations`
  - The implementation now immediately appends the decorations to the end of the supplied array
  - The added elements are then mutated in-place
- Refactors all geometry implementations to generate the element in-place and then mutate it, rather than copying it
  - This uses a helper method, `EmplaceGeometry<T>`, to place the geometry at the end of the array and return a reference to the added geometry
  - This removes a copy from the existing implementation, where the geometry was locally added *and then* copied to the end of the (temporary) array. It now places it directly on the end of the final array (due to this, and the above, change)
- Refactors private helper methods in `OpenSim::Geometry` directly into `generateDecorations` so that any post-addition fixups (e.g. setting body index, scale factors, etc.) all happen in a single pass over the data
- Moves setting the scale factor to a shared operation in `generateDecorations`, rather than it being in each implementation (**this should be ok? Might require a quick check**)

### Testing I've completed

- I looked at it in the OpenSim Creator UI and it didn't seem to make anything explode

- Whether it does make something explode
- Perf measurements

### CHANGELOG.md (choose one)

- Updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3147)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ v4.4
 - Updated ezc3d to version 1.4.6 which better manage the events defined in a c3d file.
 - Fixed an issue that could happen sometimes with ScaleTool where loading the model file or marker set file could fail if the file was given as an absolute path (Issue #3109, PR #3110)
 - Fixed an issue with SWIG with `OpenSim::Body::getRotationInGround()` where it would return an object without the correct `SimTK::Rotation` methods.
+- Reduced the overhead of Geometry::generateDecorations by removing unnecessary allocations and copies from the internal implementation
 
 v4.3
 ====
--- a/OpenSim/Simulation/Model/Geometry.cpp
+++ b/OpenSim/Simulation/Model/Geometry.cpp
@@ -37,6 +37,15 @@ using namespace SimTK;
 
 OpenSim_DEFINE_SOCKET_FD(frame, Geometry);
 
+// emplace a geometry element onto the back of the provided array and return a reference
+// to it so that callers can mutate it in-place
+template<typename T, typename... Args>
+static T& EmplaceGeometry(SimTK::Array_<SimTK::DecorativeGeometry>& out, Args&&... args)
+{
+    out.emplace_back(T{std::forward<Args>(args)...});
+    return static_cast<T&>(out.back());
+}
+
 Geometry::Geometry() {
     setNull();
     constructProperties();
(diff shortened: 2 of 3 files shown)


adamkewley:
(I'll look into the failure when I get a half hour or so)

adamkewley:
I'll try and get some rough numbers :)

Effectively, I have UI counters that are measuring things like calls to `generateDecorations` and I'm trying to figure out why parts of those decoration passes can be slow.

The biggest slowdown was from:

- https://github.com/opensim-org/opensim-core/pull/3148

Because the editor needs to re-finalize the model whenever a user-enacted edit happens. The changes in this PR had a more minor impact.

adamkewley:
From a perf perspective, this isn't worth it: the difference is within the margin of error. I'll close this off.

Here's some exercise code:

```c++
int main(int argc, const char** argv)
{
    using Clock = std::chrono::high_resolution_clock;
    constexpr int numIterations = 1024;

    OpenSim::Model m{"/home/adam/Desktop/opensim-creator/feature_reduce-geometry-copying/resources/models/RajagopalModel/Rajagopal2015.osim"};
    m.buildSystem();
    SimTK::State& s = m.initializeState();
    m.realizeReport(s);

    const OpenSim::ModelDisplayHints& mdh = m.getDisplayHints();
    SimTK::Array_<SimTK::DecorativeGeometry> buf;

    Clock::time_point t0 = Clock::now();

    for (int i = 0; i < numIterations; ++i)
    {
        buf.clear();
        m.generateDecorations(false, mdh, s, buf);
        m.generateDecorations(true, mdh, s, buf);
    }

    Clock::time_point tEnd = Clock::now();
    auto durationMicros = std::chrono::duration_cast<std::chrono::microseconds>((tEnd - t0)/numIterations);

    std::cerr << "duration per iteration = " << durationMicros.count() << " us\n";

    return 0;
}
```

Effectively, it continually generates geometry. Each loop iteration takes around 230 microseconds on my machine, which is a little bit slow (which is what piqued my interest initially), considering that it's a relatively simple geometry pass.

The reason it's slow, though, is only minorly because of the geometry copying. A `valgrind` pass on this test code shows that, yeah, the memory allocator is being called millions of times for the above loop (because it is continually allocating/deallocating decorative geometry), but this optimization does not "fix" the fact that `SimTK::DecorativeGeometry` doesn't have move semantics (so there's always going to be a temporary that is then copied around as a heap allocation).

The biggest offender in the above code section is actually `OpenSim::Component::initComponentTreeTraversal` (7-8 %). The reason that's expensive is because it contains a lot of branching code and has to recurse into each element of the tree separately, so it ends up being called a lot.

PR #3150, opened 2022-02-28
Cmake fixes 4conda

aymanhab:
Fixes issue #2524 

### Brief summary of changes
- Modify argument to add_dll_directory based on configuration (conda vs. non-conda) due to differences in layout
### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3150)

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -37,7 +37,7 @@ jobs:
         
     - name: Install numpy
       #Need numpy to use SWIG numpy typemaps.
-      run: python -m pip install numpy==1.20.2
+      run: python3 -m pip install numpy==1.20.2
 
     - name: Install SWIG
       run: choco install swig --version 4.0.2 --yes --limit-output --allow-downgrade
@@ -75,7 +75,7 @@ jobs:
         # TODO: Can remove /WX when we use that in CMakeLists.txt.
         # Set the CXXFLAGS environment variable to turn warnings into errors.
         # Skip timing test section included by default.
-        cmake -E env CXXFLAGS="/WX -DSKIP_TIMING" cmake $env:GITHUB_WORKSPACE -LAH -G"Visual Studio 16 2019" -A x64 -DCMAKE_INSTALL_PREFIX=~/opensim-core-install -DOPENSIM_DEPENDENCIES_DIR=~/opensim_dependencies_install -DOPENSIM_C3D_PARSER=ezc3d -DBUILD_PYTHON_WRAPPING=on -DBUILD_JAVA_WRAPPING=on
+        cmake -E env CXXFLAGS="/WX -DSKIP_TIMING" cmake $env:GITHUB_WORKSPACE -LAH -G"Visual Studio 16 2019" -A x64 -DCMAKE_INSTALL_PREFIX=~/opensim-core-install -DOPENSIM_DEPENDENCIES_DIR=~/opensim_dependencies_install -DOPENSIM_C3D_PARSER=ezc3d -DBUILD_PYTHON_WRAPPING=on -DBUILD_JAVA_WRAPPING=on -DPython3_ROOT_DIR=C:\hostedtoolcache\windows\Python\3.8.10\x64
         $env:match = cmake -L . | Select-String -Pattern OPENSIM_QUALIFIED_VERSION
         $version = $env:match.split('=')[1]
         echo $version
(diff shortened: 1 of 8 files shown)


aymanhab:
Primarily modify directory to search for dlls on Windows based on self-contained conda or using libraries from GUI distribution. A couple minor changes to CMake for consistency (and to fix build breaking on osx without CASADI)

carmichaelong (changes requested):
Thanks @aymanhab. I had one question below

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3150#-:-MxlGRaKCo39vLXcPK1b:beae0qs)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @aymanhab and @carmichaelong)

___
*[cmake/CMakeLists.txt, line 49 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3150#-MxktJ192wrNyac5z_8Q:-MxktJ192wrNyac5z_8R:b-fo42n4) ([raw file](https://github.com/opensim-org/opensim-core/blob/dd13442f0f2508381670948dac68a9e30c9beeb3/cmake/CMakeLists.txt#L49)):*
> ```CMake
> install(EXPORT OpenSimTargets DESTINATION "${OPENSIM_INSTALL_CMAKEDIR}")
> 
> if(OPENSIM_COPY_DEPENDENCIES AND APPLE AND OPENSIM_WITH_CASADI)
> ```

I think I now see why you mentioned that a new flag for including Moco or not would be helpful. It seems like this change could break a Moco build that just uses tropter. Would this work if the `OPENSIM_WITH_CASADI` flag was used in a separate if statement for the two lines below to get `CASADI_LIBDIR` and also in the two related spots in `OpenSimMocoInstallMacDependencyLibraries.cmake.in`?

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3150)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @carmichaelong and @nickbianco)

___
*[cmake/CMakeLists.txt, line 49 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3150#-MxktJ192wrNyac5z_8Q:-MxouM7VDzPdSUYffl4E:bx1scvg) ([raw file](https://github.com/opensim-org/opensim-core/blob/dd13442f0f2508381670948dac68a9e30c9beeb3/cmake/CMakeLists.txt#L49)):*

I think I now see why you mentioned that a new flag for including Moco or not would be helpful. It seems like this change could break a Moco build that just uses tropter. Would this work if the `OPENSIM_WITH_CASADI` flag was used in a separate if statement for the two lines below to get `CASADI_LIBDIR` and also in the two related spots in `OpenSimMocoInstallMacDependencyLibraries.cmake.in`?
</blockquote></details>

Good catch @carmichaelong I'll make this specific to CASADI.
Not sure if we ever build with TROPTER only but @nickbianco can comment.

nickbianco (commented):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3150)* status: all files reviewed, 1 unresolved discussion (waiting on @carmichaelong)

___
*[cmake/CMakeLists.txt, line 49 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3150#-MxktJ192wrNyac5z_8Q:-MxoxJb3B0RKGtZyDqLf:bv7lsbn) ([raw file](https://github.com/opensim-org/opensim-core/blob/dd13442f0f2508381670948dac68a9e30c9beeb3/cmake/CMakeLists.txt#L49)):*

Good catch @carmichaelong I'll make this specific to CASADI.
Not sure if we ever build with TROPTER only but @nickbianco can comment.
</blockquote></details>

Yes, the default setting is to build with Tropter: https://github.com/opensim-org/opensim-core/blob/73530a8bf87788729be3ff371f75238a1ea50698/CMakeLists.txt#L196.

I would be open to changing this since most users don't utilize Tropter.

aymanhab:
@carmichaelong I believe I have addressed your review, please take another look when you have a chance.
@nickbianco No need to change anything with TROPTER at the moment, the essence of the problem is described in the new issue #3156

aymanhab:
@carmichaelong I committed another change to guard the rpath changes with OPENSIM_WITH_CASADI, and successfully built with OPENSIM_WITH_CASADI, OPENSIM_WITH_TROPTER but bottomline is that this is unrelated to this PR since OPENSIM_WITH_TROPTER on, OPENSIM_WITH_CASADI off would have failed same way before my change.

carmichaelong (approved):
Makes sense, thanks @aymanhab for adding the change here anyway. LGTM, pending CI

> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3150#-:-Mxup-Wm-jz7LQO1wShe:bucjr05)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @carmichaelong and @nickbianco)

aymanhab:
Thanks @carmichaelong 👍

PR #3155, opened 2022-03-10
[WIP] Refactor and tidy up wrapping code ready for other fixes

adamkewley:
Fixes issue (N/A)

### Brief summary of changes

- (Currently) non-breaking refactors to code in `OpenSim/Simulation/Wrap/`
- `WrapObject`'s `wrapLine` and `wrapPathSegment` APIs changed to accept `const-`qualified path points, rather than mutable path points
  - This is to make it clear that the implementation shouldn't change the data
  - This is because I'm investigating another bug, where models containing muscle wrapping experience weird behavior when realizing multiple states against the same model (see: https://github.com/ComputationalBiomechanicsLab/opensim-creator/issues/123)
- De-inlined functions in `WrapMath`
  - WIP: I need to take benchmark measurements to confirm that this has no major effect on real-world perf
  - It's been done this way so that the wrap math implementation can be changed without triggering other recompilations
- `WrapMath` functions are now `const-`qualified, where applicable
- Moved some private methods into the `cpp` file as `static-`qualified functions, to make it clear that the function implementation is only dependent on the arguments (rather than also being dependent on an implicit `this` member)
- Did a quick styling pass on source files in `Wrap/`. Fixing things like empty destructor implementations, misaligned text, hard-to-read method formatting, rearranging methods to be more logical (getters next to setters, etc.)

While the diffs might look fairly big here, the changes *should* be non-breaking. I've effectively *relaxed* the API by making it accept `const-`qualified arguments (rather than requiring mutable access). No other code should require changes and all side-effects from the codebase should be the same.

Later, separately PRed, changes after this will be focused on removing things like raw pointers in wrapping data members, cache variables that hold onto model pointers, etc. Those changes also shouldn't change the public API of path wrapping, but should harden the implementation against some of the bizzare edge-case bugs I've encountered with wrapping objects (e.g. things aliasing things from other models, state realizations breaking path decoration generation in other model+state pairs, etc.)

### Testing I've completed

- I compiled it, but didn't do any specialized testing

- Usual stuff

### CHANGELOG.md (choose one)

- Updated

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3155)


OpenSim/Simulation/Wrap/WrapCylinderObst.cpp
```diff
@@ -195,11 +187,16 @@ int WrapCylinderObst::getWrapDirection() const {
...
+        const SimTK::State& s,
+        const SimTK::Vec3& aPoint1,
+        const SimTK::Vec3& aPoint2,
+        const PathWrap& aPathWrap,
+        WrapResult& aWrapResult,
+        bool& aFlag) const
 {
-    SimTK::Vec3& aPointP = aPoint1;     double R=0.8*( m_wrapDirection==righthand ? get_radius() : -get_radius() );
-    SimTK::Vec3& aPointS = aPoint2;     double Qx,Qy,Qz, Tx,Ty,Tz;
+    SimTK::Vec3 aPointP = aPoint1;     double R=0.8*( m_wrapDirection==righthand ? get_radius() : -get_radius() );
```
adamkewley, line 198: double-check whether this code relies on aliasing `aPoint1` in some bizarre way

adamkewley:
TODO: basic perf measurement of before/after on something like Rajagopal to ensure de-inlining the math functions doesn't cause a noticeable perf drop.

aymanhab:
Just a headsup that the wrapping code is extremely flaky, reliant on aliases, comparisons against 0 and hard coded tolerances, and is barely exercised in our test suite so I'd emphasize refactoring/cleanup that results in no change in math whatsoever 😿 but other cleanup should be safe.

adamkewley:
@aymanhab yeah, I'm seeing that 😄 

This set of changes are mostly about whitespace formatting, moving methods, making things `const`, etc. - just basic stuff to try and make surfing this part of the codebase a little bit easier. 

The other PR I just posted (https://github.com/opensim-org/opensim-core/pull/3159) also avoids the math for the same reasons. It's mostly focused on fixing a storage bug that has been inducing a few annoying hacks in OpenSim Creator (specifically, repeatably nuking the cache variables and re-realizing states a lot).

I'd **love** to get my hands dirty with the maths, though 🧙 . However, it's in a state where it might be best to declare bankruptcy on parts of those algorithms and write an (e.g.) `WrapCylinder_v2` component as an experimental stand-in that gradually gets phased-in over time, or something - no idea

aymanhab (changes requested):
> Reviewed 5 of 26 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3155)* status: 7 of 27 files reviewed, 2 unresolved discussions (waiting on @adamkewley and @aymanhab)

___
*[CHANGELOG.md, line 19 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3155#-Mxv8qUJ98cdt_oZ0FPW:-Mxv8qUJ98cdt_oZ0FPX:b-ajg547) ([raw file](https://github.com/opensim-org/opensim-core/blob/cafd65d4ff81a1de5969a202a19c07f039ca3ead/CHANGELOG.md#L19)):*
> ```Markdown
> - Changed the in-parameters in the `OpenSim::WrapMath` functions (e.g. `SimTK::Vec3&`) to now be const-qualified (previously: they were mutable)
> - Moved previously private-member functions in `OpenSim::WrapTorus` (e.g. `OpenSim::WrapTorus::calcCircleResids`) into the implementation file as compilation-unit-local functions
> ```

Typically these are higher level changes that make sense to end users or potential developers, considering these methods are mostly internal to the wrapping code and not by any client I'd mostly put something generic like refactoring WrapObject internals in case some side-effect ends up showing down the line.
___
*[OpenSim/Simulation/Wrap/WrapCylinderObst.cpp, line 198 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3155#-MxpC4k74yoQi6M3WLEI-r1-198:-Mxvxysi6FrH58iL9ahC:b-llguk0) ([raw file](https://github.com/opensim-org/opensim-core/blob/6f481c3b4d7c416cb7adafbabf7f74d0cb762deb/OpenSim/Simulation/Wrap/WrapCylinderObst.cpp#L198)):*

double-check whether this code relies on aliasing `aPoint1` in some bizarre way
</blockquote></details>

WrapCylinderObst is never used in any of our models or other users models AFAIK. I'd lean to retire it to reduce the unnecessary bloat.

adamkewley:
Dumping this - it's low prio and I don't have the time for it right now. I'll re-open once I have a little while to get into this

adamkewley:
Dumping this - it's low prio and I don't have the time for it right now. I'll re-open once I have a little while to get into this

Issue #3151, opened 2022-03-04
AnalyzeTool Muscle Analysis Moment Flag


cclancy8:
When requesting a MuscleAnalysis with the AnalyzeTool, the tool does not toggle the MuscleAnalysis to compute moments (and moment arms) even when `setComputeMoments(true)` is specified for the MuscleAnalysis. A check was done with   `getComputeMoments()`, which indicates that the option to compute moments was set to true. However, the printed .xml file of the AnalyzeTool does not have the flag indicated to compute moments and thus they are not calculated.

aymanhab:
My recollection is that this flag/property was not intended to be set explicitly but for internal use so that we know moment arms or moments are to be computed. This suggests that this shouldn't be exposed as a property as it would be confusing. The question from a user perspective is whether you can specify a list of Muscles and Coordinates and then ask for MomentArms or Moments

cclancy8:
Would it be possible to update the MuscleAnalysis so that moments and moment arms can be calculated with this flag? These result quantities are helpful in analysis. I can manually change the printed xml file to compute moments and moment arms with setting this flag to true, but I would like to have this an automated option when running for many different models.

carmichaelong:
@cclancy8 do you have a snippet of code / files to reproduce?

carmichaelong:
It looks like the flag that `setComputeMoments()` changes might actually be what should control that behavior: https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Analyses/MuscleAnalysis.cpp#L642

cclancy8:
Yes - I use `setComputeMoments(true) `. I also check with `getComputeMoments()` and get back true, but when I open the xml, the flag is unchanged. This is my code:

carmichaelong:
I'm guessing you only ever run this using the .xml file then, and not using the `myAnalyzeTool` object? I was able to reproduce your issue (i.e., the .xml file is incorrect), and then when I went to double check by peeking into the analyze tool object, it looks correct:
```
a = myAnalyzeTool.updAnalysisSet().adoptAndAppend(MA);
b = MuscleAnalysis.safeDownCast(a);
b.getComputeMoments()
```
This prints a 1 out too

carmichaelong:
If you've got `myAnalyzeTool` object all setup correctly, then in theory it should be as simple as `myAnalyzeTool.run();`. FYI, while that should work, it's still definitely a bug that the writing to file is incorrect and Ayman will look into that soon

[...]

PR #3157, opened 2022-03-10
Remove _computeMoments variable and use Property directly everywhere

aymanhab:
Fixes issue #3151 

### Brief summary of changes
Muscle analysis uses old Properties which were designed to have a Property and a variable to reference the bool value inside the property, this issue is due to these variable getting out of sync with the property, so the variable was removed and Property is maintained.
### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3157)


adamkewley:
If you need a reviewer, this looks good :)

aymanhab:
Thanks @adamkewley much appreciated 👍 , will merge after the added test passes.

aymanhab:
Mac test failure is unrelated, so merging.

PR #3159, opened 2022-03-11
[WIP] Fix path wrap model-level data mutates+aliases during a simulation

adamkewley:
Related to: https://github.com/ComputationalBiomechanicsLab/opensim-creator/issues/123

Effectively, I'm trying to fix a number of issues in the path wrapping code. Notably:

- The current implementation's `GeometryPath` stores pointers to components within the model graph inside a state cache variable.
  - This causes a potential aliasing issue when the state is (e.g.) passed between threads with non-copy semantics. E.g. passing a `unique_ptr<SimTK::State>` between threads causes one thread to have aliased access into the other thread's (potentially destroyed) model components

- The current implementation stores path wrap points as model-/component-level data, rather than as state/cache variables within the state.
  - This means that state-dependent data (e.g. current wrapping path) is stored at the model level
  - It causes bizzare bugs. E.g. realizing a state against a model causes the model to contain a different version of the path wrapping. Subsequently generating decorations against the same model, but with a different state (e.g. because I'm rendering *state sequences* against *one* model) then causes the rendering to glitch because the path wrap data held within the model is mismatched with the state

### Brief summary of changes

- `GeometryPath`'s header was rearranged to match the style of:
  - class
  - properties
  - ctor
  - members (getters, updaters, setters in-order)
  - private members
  - data
- `GeometryPath`'s implementation functions were all moved into the cpp file:
  - Perf measurements indicated that this doesn't have a measurable impact
  - It makes working on `GeometryPath` easier because only one compilation unit is impacted (apart from with data changes)
- `GeometryPath`'s methods were refactored, usually to reduce the number of calls to property getters or the number of passes over datastructures
  - This will need another pass to ensure the algorithm rearrangements didn't cause different behavior. Ancecdotal testing with existing models indicates that nothing obvious has changed

### Testing I've completed

WIP

WIP

### CHANGELOG.md (choose one)

WIP

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3159)

### DIFF
--- a/OpenSim/Simulation/Model/GeometryPath.h
+++ b/OpenSim/Simulation/Model/GeometryPath.h
@@ -23,14 +23,11 @@
  * limitations under the License.                                             *
  * -------------------------------------------------------------------------- */
 
-
-// INCLUDE
 #include <OpenSim/Simulation/osimSimulationDLL.h>
-#include "OpenSim/Simulation/Model/ModelComponent.h"
-#include "PathPointSet.h"
+#include <OpenSim/Simulation/Model/ModelComponent.h>
+#include <OpenSim/Simulation/Model/PathPointSet.h>
 #include <OpenSim/Simulation/Wrap/PathWrapSet.h>
 #include <OpenSim/Simulation/MomentArmSolver.h>
-
 
 #ifdef SWIG
     #ifdef OSIMSIMULATION_API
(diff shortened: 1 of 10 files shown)


adamkewley:
This is effectively done now - I just need to do one more review pass and a performance pass.

- `GeometryPath` no longer stores pointers in the cache variable
- Wrapping classes now hold onto (e.g.) `SimTK::ReferencePtr` to ensure any pointers in them are wiped.
- State-dependent `PathWrapPoint`s now hold their wrap locations int a `SimTK` cache variable, rather than on the model
- The API for `GeometryPath` can now be more `const`-correct, because state-dependent mutations mutate via cache variables, rather than mutating the model (the old behavior is why there are some unusual bugs when working with paths that contain path wraps over multiple states)

I also ended up refactoring a fair bit of `GeometryPath`'s internals, because they now need to differentiate between computing and cache lookups slightly differently. The `Array<const AbstractPathPoint*>` API is still there, so there's various little translation steps that need to happen to ensure that the rest of the codebase doesn't have to worry about how they are materialized in a state-dependent way.

I'll need to carefully performance analyze this, because getting the current path points now involves a fair bit more work because it has to:

- Lookup the `AbstractPathPoint`s that make up the path in a state-dependent way
- Lookup any wrapping-based information (current wrap point locations, current wrap path lengths) in a state-dependent way

Those indirect lookups are more expensive than the previous behavior of holding the information at the model level. However, I believe that's what's necessary to write a "correct" implementation - the path wrapping points being held at the model-level meant that all downstream code would have to re-realize states etc. to receive the correct path points.

adamkewley:
First performance check **versus OpenSim 4.2, 16 repeats**:

|             RajagopalModel |       6.58 |     0.00 |       5.49 |     0.00 |    1.20 |
|                      Arm26 |       0.16 |     0.00 |       0.15 |     0.00 |    1.10 |
|             ToyDropLanding |      10.92 |     0.01 |      10.28 |     0.01 |    1.06 |
|                   Gait2354 |       0.18 |     0.00 |       0.18 |     0.00 |    0.99 |
|            passive_dynamic |       2.61 |     0.00 |       2.61 |     0.00 |    1.00 |
| passive_dynamic_noanalysis |       1.73 |     0.00 |       1.72 |     0.00 |    1.01 |
|   ToyDropLanding_nomuscles |       0.27 |     0.00 |       0.27 |     0.00 |    1.00 |

It seems that the various little changes to the internal algorithms + storage may have accelerated something. I'll need to compare these benchmarks (default forward dynamic sims of standard OpenSim models) against `master` and do a correctness check before claiming any benefits.

OpenSim/Simulation/Wrap/PathWrapPoint.h
```diff
@@ -24,7 +24,9 @@
  * -------------------------------------------------------------------------- */
```
adamkewley, line 24: The changes in `PathWrapPoint` need to be added to the CHANGELOG, because the length etc. are now state-dependent (previously: they were held at the model-level, which was a bug, but fixing it does involve an API change)

OpenSim/Simulation/Wrap/PathWrapPoint.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+    // The velocity of the station in ground is a function of its frame's
+    // linear (vF = V_GF[1]) and angular (omegaF = A_GF[0]) velocity, such that
+    // velocity of the station: v = vF + omegaF x r
+    return V_GF[1] + V_GF[0] % r;
+}
+
+SimTK::Vec3 OpenSim::PathWrapPoint::calcAccelerationInGround(const SimTK::State& state) const
+{
+    // The spatial velocity of the reference frame expressed in ground
```
adamkewley, line 111: These codeblocks were ripped from `OpenSim::Station`, but they don't appear to be ideally implemented.

The existing implementation repeatably looks up the parent frame, which might involve runtime lookups. I also need to check how frames typically store their data, because this entire algorithm is putting things in ground one-at-a-time, whereas it might be more efficient to:

- Grab a reference to the parent frame
- Create a stack-local `transformInGround`
- Use the transform to transform the (not-in-ground) velocity, acceleration, and location

This could be faster because it only involves one frame lookup, one transform creation, and several trivial getter calls (for the "raw" velocity, acceleration, etc.)

OpenSim/Simulation/Wrap/PathWrapPoint.cpp
```diff
@@ -0,0 +1,124 @@
...
+}
+
+const OpenSim::Array<SimTK::Vec3>& OpenSim::PathWrapPoint::getWrapPath(const SimTK::State& s) const
+{
+    return getCacheVariableValue(s, _wrapPath);
+}
+
+void OpenSim::PathWrapPoint::setWrapPath(const SimTK::State& s, const Array<SimTK::Vec3>& src) const
+{
+    updCacheVariableValue(s, _wrapPath) = src;
```
adamkewley, line 44: Patterns like this, and in other OpenSim components, need to be fixed by adding high-order functions like:

- `ApplyCacheVariableChange`
- `FetchCacheVariableValue(cv, updateFunction);`

Where there's only one cache variable lookup. The lookup is cheaper now, due to previous cache variable changes, but it still irks me that there are two lookups happening when one should do.

OpenSim/Simulation/Model/GeometryPath.h
```diff
@@ -47,202 +44,350 @@ class ScaleSet;
...
+    // reset on copy to prevent aliasing issues.
+    mutable SimTK::ResetOnCopy<Array<const AbstractPathPoint*>> _currentPathPtrsCache;
+
+    // cache variable for the path's length
+    mutable CacheVariable<double> _lengthCV;
+
+    // cache variable for the path's lengthening speed
+    mutable CacheVariable<double> _speedCV;
+
+    // cache variable for the (order of) the path's points
```
adamkewley, line 378: This information could be packed into 32-bits, rather than 64, but that's a nice-to-have extra step

adamkewley:
The perf measurements are spurious. Comparing `master` to this branch yields no noticeable perf change:

|             RajagopalModel |       5.97 |     0.07 |       5.98 |     0.05 |    1.00 |
|                      Arm26 |       0.15 |     0.00 |       0.15 |     0.00 |    1.01 |
|             ToyDropLanding |      10.87 |     0.05 |      10.72 |     0.05 |    1.01 |
|                   Gait2354 |       0.19 |     0.00 |       0.19 |     0.00 |    1.00 |
|            passive_dynamic |       2.66 |     0.00 |       2.66 |     0.00 |    1.00 |
| passive_dynamic_noanalysis |       1.76 |     0.00 |       1.73 |     0.00 |    1.01 |
|   ToyDropLanding_nomuscles |       0.27 |     0.00 |       0.27 |     0.00 |    1.00 |

OpenSim/Simulation/Wrap/PathWrapPoint.cpp
adamkewley: Dumping fixing this in this PR because it has a negligible perf change

adamkewley:
This PR has become quite a monstrosity, because it mixes fixing two things--a) pointer aliasing in `GeometryPath`, b) wrap point storage in `PathWrapPoint`--and also sprinkles some cleanups, refactoring, method cleanups, etc. on top.

Now that things seem to be (superficially ;)) working, I'll break each part of the changset up into separate PRs that are easier to review (I hope :>). I'll link them here then close this PR.

aymanhab:
Thanks @adamkewley much appreciated, I was getting rather worried 😨

adamkewley:
Other  (less-important) changes that can be separately PEed:

- Change `PathWrap` to hold a `SimTK::ReferencePtr`, so that copying a `PathWrap` nulls out the potential alias
- Make the GeometryPath API return `const Array_<const AbstractPathPoint*>` rather than `const Array_<AbstractPathPoint*>`
- Refactor `addInEquivalentForces` to be a little bit cleaner and perform less lookups (stylistic, minor perf bump: it's called a lot)
- Refactor `getPointForceDirections` to do a single pass over the data (stylistic/perf)
- Clean up `GeometryPath` header (group methods together, declare+define methods in a more logical order) - purely stylistic

adamkewley:
Closing this PR: I'll slice it up into smaller pieces and PR them separately whenever I get the chance

PR #3160, opened 2022-03-15
Fixed GeometryPath storing pointers in SimTK::States

adamkewley:
This is a cut-down version of PR #3159 that specifically fixes the geometry-path-shouldn't-store-pointers-in-`SimTK::State`s bug.

The current implementation's `GeometryPath` stores pointers to components within the model graph inside a state cache variable. This causes a potential aliasing issue when the state is (e.g.) passed between threads with non-copy semantics. E.g. passing a `unique_ptr<SimTK::State>` between threads causes one thread to have aliased access into the other thread's (potentially destroyed) model components.

### Brief summary of changes

- Changed `GeometryPath`'s path pointer cache variable to instead cache indices to the paths that should be looked up. This changes a direct pointer storage, which might alias into a different model, into a index storage, which can be used across model copies.

### Testing I've completed

- Standard unit test suite: it is mostly an internal change

### CHANGELOG.md (choose one)

- no need to update because it has no effect on external APIs and is only an internal lookup fix

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3160)

### DIFF
--- a/OpenSim/Simulation/Model/GeometryPath.cpp
+++ b/OpenSim/Simulation/Model/GeometryPath.cpp
@@ -45,36 +45,40 @@ class OpenSim::GeometryPath::PathElementLookup {
 public:
     static PathElementLookup fromPtr(const OpenSim::PathPointSet& pps,
                                      const OpenSim::PathWrapSet& pws,
+                                     const OpenSim::GeometryPath& geomPath,
                                      const OpenSim::AbstractPathPoint* ptr)
     {
         // linearly search each collection to figure out its type + index
 
-        for (int i = 0, len = pps.getSize(); i < len; ++i) {
-            if (&pps[i] == ptr) {
-                return PathElementLookup{i, PointType::InPathPointSet};
+        const OpenSim::Component* owner = &ptr->getOwner();
+
+        if (owner == &geomPath) {
+            for (int i = 0, len = pps.getSize(); i < len; ++i) {
+                if (&pps[i] == ptr) {
+                    return PathElementLookup{PointType::InPathPointSet, i};
+                }
             }
         }
-        for (int i = 0, len = pws.getSize(); i < len; ++i) {
-            const PathWrap& pw = pws[i];
-            if (&pw.getWrapPoint1() == ptr) {
-                return PathElementLookup{i, PointType::PathWrapPt1};
-            }
-            else if (&pw.getWrapPoint2() == ptr) {
-                return PathElementLookup{i, PointType::PathWrapPt2};
+        else {
+            for (int i = 0, len = pws.getSize(); i < len; ++i) {
+                if (&pws[i] == owner) {
+                    PointType t = ptr == &pws[i].getWrapPoint1() ? PointType::PathWrapPt1 : PointType::PathWrapPt2;
+                    return PathElementLookup{t, i};
+                }
             }
         }
+
         OPENSIM_THROW(OpenSim::Exception, "unable to create a PathElementLookup for the given pointer: are you sure it is part of the GeometryPath?");
     }
 
 private:
-    enum class PointType { InPathPointSet, PathWrapPt1, PathWrapPt2 };
+    enum class PointType : char { InPathPointSet, PathWrapPt1, PathWrapPt2 };
 
-    PathElementLookup(int index, PointType pointType) :
+    PathElementLookup(PointType pointType, int index) :
         index_{index},
         pointType_{pointType}
     {
     }
-
 public:
     OpenSim::AbstractPathPoint* toPtr(const OpenSim::PathPointSet& pps,
                                       const OpenSim::PathWrapSet& pws) const
@@ -92,8 +96,8 @@ public:
     }
 
 private:
-    int index_;
-    PointType pointType_;
+    int index_ : 24;
+    PointType pointType_ : 8;
 };
 
 static void PopulatePathPointersCache(
@@ -112,12 +116,13 @@ static void PopulatePathElementLookup(
     const OpenSim::PathPointSet& pps,
     const OpenSim::PathWrapSet& pws,
     const OpenSim::Array<OpenSim::AbstractPathPoint*>& ptrs,
+    const OpenSim::GeometryPath& geomPath,
     std::vector<OpenSim::GeometryPath::PathElementLookup>& out)
 {
     out.clear();
     out.reserve(ptrs.getSize());
     for (int i = 0, len = ptrs.getSize(); i < len; ++i) {
-        out.push_back(OpenSim::GeometryPath::PathElementLookup::fromPtr(pps, pws, ptrs[i]));
+        out.push_back(OpenSim::GeometryPath::PathElementLookup::fromPtr(pps, pws, geomPath, ptrs[i]));
     }
 }
 
@@ -935,6 +940,7 @@ void GeometryPath::computePath(const SimTK::State& s) const
     PopulatePathElementLookup(get_PathPointSet(),
                               get_PathWrapSet(),
                               _currentPathPtrsCache,
+                              *this,
                               lookups);
     markCacheVariableValid(s, _currentPathCV);
 }


adamkewley:
The change seems to be working fine, but I need to do a perf pass on this. The new implementation will have more overhead (than directly dereferencing a pointer), but I will try and massage the implementation to make the overhead as small as possible (but while maintaining the guarantee of no pointer aliasing).

adamkewley:
Seems to cause a slowdown in models that contain a lot of muscles:

|             RajagopalModel |       5.44 |     0.00 |       5.58 |     0.00 |    0.97 |
|                      Arm26 |       0.15 |     0.00 |       0.15 |     0.00 |    0.99 |
|             ToyDropLanding |      10.51 |     0.02 |      10.84 |     0.08 |    0.97 |
|                   Gait2354 |       0.18 |     0.00 |       0.18 |     0.00 |    1.00 |
|            passive_dynamic |       2.62 |     0.00 |       2.63 |     0.00 |    1.00 |
| passive_dynamic_noanalysis |       1.74 |     0.00 |       1.73 |     0.00 |    1.01 |
|   ToyDropLanding_nomuscles |       0.27 |     0.00 |       0.27 |     0.00 |    1.01 |

adamkewley:
The latest commit tries to reduce the amount of searching going on and tries to pack the index + type fields into 32-bits (24 for the index an 8 for the type), to try and encourage all the lookups to end up on the same cache line with more complicated paths.

Unsure how to get it faster. Either way, it's looking like removing pointers from the cache variables is going to have an ~2 % perf hit. The larger PR tries to offset this by making other parts of `GeometryPath` faster (these will be separate PRs)

|             RajagopalModel |       5.45 |     0.00 |       5.55 |     0.00 |    0.98 |
|                      Arm26 |       0.15 |     0.00 |       0.15 |     0.00 |    0.99 |
|             ToyDropLanding |      10.51 |     0.02 |      10.59 |     0.02 |    0.99 |
|                   Gait2354 |       0.18 |     0.00 |       0.18 |     0.00 |    1.00 |
|            passive_dynamic |       2.62 |     0.00 |       2.62 |     0.00 |    1.00 |
| passive_dynamic_noanalysis |       1.72 |     0.00 |       1.73 |     0.00 |    0.99 |
|   ToyDropLanding_nomuscles |       0.27 |     0.00 |       0.27 |     0.00 |    1.01 |

I think this is probably ready for a review pass. @aseth1 mentioned giving it a swing but @aymanhab is also familiar with these classes (I saw you made a pass at the wrapping code recently)

adamkewley:
(it was a random CI bug and was fixed by re-running CI)

adamkewley:
@aymanhab / @carmichaelong  sorry to tag you in on this, but is there a chance that someone over in the core dev team could review this with a view for it being included in 4.4? It'll mean that I can upgrade `osc` to 4.4 and nuke the downstream workarounds. I'm willing to do whatever changes are necessary - this code is effectively the laziest way to swap a pointer for a runtime lookup.

The motivator here is that it makes states a little bit safer to pass between threads. Currently, I'm having to wipe all cache variables whenever a state is thrown "over the fence" from a background thread because there is a fear that using potentially-stale pointers across thread boundaries causes a segfault. In my case it almost always causes a segfault, because the UI thread might uses states to (e.g.) compute an output plot after a simulator thread concurrently destructs its model copy.

aymanhab:
@adamkewley thanks for your patience, I'd definitely love to have you get your osc codebase in sync. with 4.4 before we publish it. 

Can you give a little bit more context to your scenario: I'm assuming based on your description that you have multiple threads each with its own model but they share a state is that correct? Any guarantees that the two models are in sync? Also did you have a stack trace to confirm this being the source of segfault? Thank you

adamkewley:
That's basically it, yeah. Two models, one on a simulator thread, the other on a UI thread. Because the state contains raw pointers to path points, passing a state between the two threads implicitly causes one thread to gain access to the other's model. If one thread outlives the other, it's a segfault.

It was located by carefully analyzing a stacktrace, because the source of the error is a segfault that happens during some part of `GeometryPath` that tries to access the pointer (e.g. `generateDecorations`) and careful analysis of the other (simulation) thread indicates that it happens after the sim thread drops its model copy. The segfault in question disappears if either this patch is applied *or* cache variables are aggressively nuked whenever states are moved around.

aymanhab (changes requested):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3160)* status: all files reviewed, 1 unresolved discussion (waiting on @adamkewley and @aseth1)

___
*[`OpenSim/Simulation/Model/GeometryPath.cpp` line 76 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3160#-N0XE37kEaz6vw_22kOz:-N0XE37kEaz6vw_22kP-:b4vncfn) ([raw file](https://github.com/opensim-org/opensim-core/blob/b5b661e477c1786a0da4e4301508f955752d93d9/OpenSim/Simulation/Model/GeometryPath.cpp#L76)):*
> ```cpp
> private:
>     enum class PointType : char { InPathPointSet, PathWrapPt1, PathWrapPt2 };
> ```

Is there a reason to keep the distinction between those undocumented PathWrapPt1, PathWrapPt2? Just trying to minimize propagation of the undocumented/hacked-up wrapping behavior to more places in the code for future maintenance.

aymanhab:
Other than trying to reduce exposure of Wrapping internal hacks if possible, LGTM

adamkewley (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3160)* status: all files reviewed, 1 unresolved discussion (waiting on @aseth1 and @aymanhab)

___
*[`OpenSim/Simulation/Model/GeometryPath.cpp` line 76 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3160#-N0XE37kEaz6vw_22kOz:-N185opi4lgwEcV8QiuY:b-g8r2ii) ([raw file](https://github.com/opensim-org/opensim-core/blob/b5b661e477c1786a0da4e4301508f955752d93d9/OpenSim/Simulation/Model/GeometryPath.cpp#L76)):*

Is there a reason to keep the distinction between those undocumented PathWrapPt1, PathWrapPt2? Just trying to minimize propagation of the undocumented/hacked-up wrapping behavior to more places in the code for future maintenance. 
</blockquote></details>

It's necessary because the runtime needs to be able to locate the specific `OpenSim::AbstractPathPoint` that was saved as a non-pointer into the state.

Given the current architecture, there are three storage locations for a path point: 

- In the path point set
- On a wrapping object (pt1)
- On a wrapping object (pt2)

The "job" of this code is to "serialize" that location without using a pointer. At a minimum, the following are required:

1. A discriminator that says which collection the point is in (either the point set or the wrap set)
2. An index into that collection
3. (if applicable) a discriminator that says which sub-part of the collection object the point is in (`Pt1`, `Pt2`)

The enum covers both 1. and 2.

If there is a better alternative for storing this information without using a pointer, while incurring a minimal runtime overhead (so, e.g., using `ComponentPath`s are out of picture) then I'd be happy to change it.

aymanhab:
@adamkewley Thanks for the response, I understand but I was hoping you can do without 3 by doing it downstream in-case we change the wrapping to handle more shapes or different descriptor of the wrap segment but if this happens will overhaul the whole thing so it likely wouldn't matter.

aymanhab (approved):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3160)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aseth1)

aymanhab (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3160#-:-N1Fa_cx3dqA7L6tzth9:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aseth1)

adamkewley:
Thanks for reviewing this @aymanhab 

I know it's a bit of a mess to handle it "upstream" like this, but I figure it's an extra protection barrier around using `SimTK::State`s. It's likely that I'll also be redundantly hardening against stuff like this downstream by minimizing `State` transfers. E.g. by adopting something similar to the GUI/tools, where `SimTK::State`s are typically stuffed into (typically, safer) state vectors or similar. Similar deal with handling the threading issues: I'll probably also be redundantly hardening against them by additionally installing mutexes around models and states.

And it shouldn't be a big deal to nuke this whenever a rewrite happens. As long as the `GeometryPath` API provides some way of stuffing a vector with point positions--maybe with a `UserEnactedPoint` `ComputedWrapPoint` tag or similar--I think a fair amount of downstream use-cases (e.g. custom muscle path generation for GUIs) would be mostly alright (famous last words).

PR #3163, opened 2022-03-16
Fixed PathWrapPoint to hold its points in a cache variable

adamkewley:
This is a cut-down version of part of the changes in #3159 , such that this individual change can be reviewed in isolation.

This PR fixes a bug in `PathWrapPoint`. `PathWrapPoint` currently stores the wrap point locations as a (model-level) member variable. That is a bug because the path wrap points are state-dependent: each state may produce different wrapping points.

Storing the points at the model-level can cause bizarre bugs. Here's an example scenario:

- Create/load a model
- Simulate the model with (e.g.) a forward dynamic integration method
- Store states from that simulation at regular intervals. Ensure each state is fully realized to the `Report` level, so that other systems (e.g. a UI, report generator, whatever) can handle them as immutable `const` states
- Perform this realization by calling something like `model.realizeReport(latestSimState)`
- You now have a sequence of `SimTK::State`s, where each state is an immutable value that represents one step of the simulation of a model (the model may, itself, have been copied around the application: see #3160 for a separate issue related to that).

Stage set. Hopefully this all sounds reasonable enough. However, with the current implementation, this (roughly written hypothetical code) may encounter a bug:

```c++
SimTK::Array_<SimTK::DecorativeGeometry> buf;
for (const SimTK::State& state : states) {
    buf.clear();
    model.generateDecorations(false, hints, state, buf);
    model.generateDecorations(true, hints, state, buf);
    RenderVideoFrame(buf);
}
```

The bug is that the rendered `PathWrapPoint`s will show whatever was last realized against the model (because the wrap points are stored on the *model*, not the state). This will cause paths in the output render to look extremely bizzare (i.e. wrong). I encountered this issue in opensim-creator: https://github.com/ComputationalBiomechanicsLab/opensim-creator/issues/123 and have to hack around it by wiping and re-realizing a state before rendering (e.g.: [here](https://github.com/adamkewley/opensim-creator/blob/0.1.2/src/Screens/SimulatorScreen.cpp#L107))

The solution to this problem is to make `PathWrapPoint` store its data in `SimTK` cache variables on each state, so that they are state dependent. That is what this PR aims to do.

### Brief summary of changes

- Made `PathWrapPoint` store its current path and path length in cache variables (previously: member variables)
- Made `PathWrapPoint` inherit from `AbstractPathPoint` (previously: `PathPoint`)
  - This is necessary because `PathPoint` assumes that the point is fixed with respect to the parent frame and, therefore, doesn't require state-level storage
  - This change necessitated also implementing the necessary overrides (e.g. `calcLocationInGround`)
- Made `PathWrapPoint`'s API reflect the new storage model
  - The getters/setters now require access to a `SimTK::State` (which is where their data is now stored)
- Fixed uses of `PathWrapPoint` (e.g. `GeometryPath`) to use the changed API
- Fixed `testIterators`, which contains a hard-coded component count, to contain the new component count
  - This is necessary because `PathPoint` creates a hidden `Station` subcomponent, whereas `AbstractPathPoint` does not. So models containing `PathWrapPoint`s now contain fewer subcomponents.

### Testing I've completed

- Ensured it passes on the existing test suite

- Usual stuff

### CHANGELOG.md (choose one)

- no need to update because it's an internal API?

### Performance Changes

Measured some basic forward-dynamic sims of this branch against the current `master` branch. There does not appear to be a significant change in performance (`passive_dynamic` is shorter, so susceptible to OS/bootup noise):

|             RajagopalModel |       6.17 |     0.01 |       6.17 |     0.00 |    1.00 |
|                      Arm26 |       0.16 |     0.00 |       0.16 |     0.00 |    1.00 |
|             ToyDropLanding |      11.31 |     0.02 |      11.26 |     0.02 |    1.00 |
|                   Gait2354 |       0.19 |     0.00 |       0.19 |     0.00 |    1.00 |
|            passive_dynamic |       2.76 |     0.00 |       2.61 |     0.00 |    1.06 |
| passive_dynamic_noanalysis |       1.75 |     0.00 |       1.73 |     0.00 |    1.01 |
|   ToyDropLanding_nomuscles |       0.27 |     0.00 |       0.27 |     0.00 |    1.01 |

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3163)

### DIFF
--- a/Bindings/Java/OpenSimJNI/Test/testContext.cpp
+++ b/Bindings/Java/OpenSimJNI/Test/testContext.cpp
@@ -72,9 +72,7 @@ int main()
     LoadOpenSimLibrary("osimJavaJNI");
 
     Model *model = new Model("wrist.osim");
-    SimTK::State* workingState = &model->initSystem();
-    model->realizePosition(*workingState);
-    OpenSimContext* context = new OpenSimContext(workingState, model);
+    OpenSimContext* context = new OpenSimContext(&model->initSystem(), model);
     const ForceSet& fs = model->getForceSet();
     int n1 = fs.getNumGroups();
     const ObjectGroup* grp = fs.getGroup("wrist");
@@ -86,9 +84,7 @@ int main()
     delete model;
     delete context;
     model = new Model("arm26_20.osim");
-    workingState = &model->initSystem();
-    model->realizePosition(*workingState);
-    context = new OpenSimContext(workingState, model);
+    context = new OpenSimContext(&model->initSystem(), model);
 
     // Make a copy of state contained in context ad make sure content match
     SimTK::State stateCopy = context->getCurrentStateCopy();
--- a/Bindings/Java/OpenSimJNI/Test/testContext.cpp
+++ b/Bindings/Java/OpenSimJNI/Test/testContext.cpp
@@ -85,10 +85,10 @@ int main()
     delete context;
     model = new Model("arm26_20.osim");
     context = new OpenSimContext(&model->initSystem(), model);
-
     // Make a copy of state contained in context ad make sure content match
     SimTK::State stateCopy = context->getCurrentStateCopy();
     assert(context->getCurrentStateRef().toString()==stateCopy.toString());
+
     Array<std::string> stateNames = model->getStateVariableNames();
     OpenSim::Force* dForce=&(model->updForceSet().get("TRIlong"));
     Muscle* dTRIlong = dynamic_cast<Muscle*>(dForce);
--- a/Bindings/Java/OpenSimJNI/Test/testContext.cpp
+++ b/Bindings/Java/OpenSimJNI/Test/testContext.cpp
@@ -72,7 +72,9 @@ int main()
     LoadOpenSimLibrary("osimJavaJNI");
 
     Model *model = new Model("wrist.osim");
-    OpenSimContext* context = new OpenSimContext(&model->initSystem(), model);
+    SimTK::State* workingState = &model->initSystem();
+    model->realizePosition(*workingState);
+    OpenSimContext* context = new OpenSimContext(workingState, model);
     const ForceSet& fs = model->getForceSet();
     int n1 = fs.getNumGroups();
     const ObjectGroup* grp = fs.getGroup("wrist");
@@ -84,11 +86,13 @@ int main()
     delete model;
     delete context;
     model = new Model("arm26_20.osim");
-    context = new OpenSimContext(&model->initSystem(), model);
-    // Make a copy of state contained in context ad make sure content match 
+    workingState = &model->initSystem();
+    model->realizePosition(*workingState);
+    context = new OpenSimContext(workingState, model);
+
+    // Make a copy of state contained in context ad make sure content match
     SimTK::State stateCopy = context->getCurrentStateCopy();
     assert(context->getCurrentStateRef().toString()==stateCopy.toString());
-
     Array<std::string> stateNames = model->getStateVariableNames();
     OpenSim::Force* dForce=&(model->updForceSet().get("TRIlong"));
     Muscle* dTRIlong = dynamic_cast<Muscle*>(dForce);
@@ -161,6 +165,8 @@ int main()
     // Compare to known path 
     cout << "New Muscle Path" << endl;
     cout << path.getSize() << endl;
+    context->recreateSystemKeepStage();
+    model->realizePosition(stateCopy);
     for(int i=0; i< path.getSize(); i++)
         cout << path[i]->getParentFrame().getName() 
              << path[i]->getLocation(stateCopy) << endl;
--- a/Bindings/Java/OpenSimJNI/OpenSimContext.cpp
+++ b/Bindings/Java/OpenSimJNI/OpenSimContext.cpp
@@ -50,6 +50,7 @@ namespace OpenSim {
 OpenSimContext::OpenSimContext( SimTK::State* s, Model* model ) {
     _configState.reset(s);
     _model.reset(model);
+    realizePosition();
 }
 
 
--- a/Bindings/Java/OpenSimJNI/OpenSimContext.h
+++ b/Bindings/Java/OpenSimJNI/OpenSimContext.h
@@ -26,8 +26,9 @@
 #include <OpenSim/Common/Object.h>
 #include <OpenSim/Common/PropertyTransform.h>
 #include <OpenSim/Simulation/osimSimulationDLL.h>
-#include <OpenSim/Simulation/Model/Model.h>
 #include <OpenSim/Simulation/Model/Force.h>
+#include <OpenSim/Simulation/Model/Model.h>
+#include <OpenSim/Simulation/Model/PathPoint.h>
 #include <OpenSim/Common/Array.h>
 #include <OpenSim/Tools/InverseKinematicsTool.h>
 #include <Bindings/PropertyHelper.h>
--- a/Bindings/Java/OpenSimJNI/OpenSimContext.h
+++ b/Bindings/Java/OpenSimJNI/OpenSimContext.h
@@ -28,7 +28,6 @@
 #include <OpenSim/Simulation/osimSimulationDLL.h>
 #include <OpenSim/Simulation/Model/Force.h>
 #include <OpenSim/Simulation/Model/Model.h>
-#include <OpenSim/Simulation/Model/PathPoint.h>
 #include <OpenSim/Common/Array.h>
 #include <OpenSim/Tools/InverseKinematicsTool.h>
 #include <Bindings/PropertyHelper.h>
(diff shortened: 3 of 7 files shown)


aymanhab:
@adamkewley Can you fix the build/ci failures to see what changes are needed by client code (if any). Thanks for the find and the fix looks good overall, will add specific comments if any shortly. I'm curious if the crash hypothesized to be caused by this issue has been fixed with the change here. Thanks much @adamkewley 👍

aymanhab (changes requested):
Nice summary/description of error scenarios. 

> Reviewed 3 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3163#-:-MyTGPNWEhFroB8-Olc_:b34r8cf)* status: 3 of 4 files reviewed, 5 unresolved discussions (waiting on @adamkewley and @aymanhab)

___
*[OpenSim/Simulation/Wrap/PathWrapPoint.h, line 42 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3163#-MyTVEmtAmMzpJj1iAR8:-MyTVEmtAmMzpJj1iAR9:b-s7tbny) ([raw file](https://github.com/opensim-org/opensim-core/blob/4fdc227a4da643177b1f9d67e9962823b5ca57ca/OpenSim/Simulation/Wrap/PathWrapPoint.h#L42)):*
> ```objc
>  * @version 1.0
>  */
> class OSIMSIMULATION_API PathWrapPoint final : public AbstractPathPoint {
> ```

Good idea to have the final here, we likely can add it in most of our concrete classes for a performance boost with no risk
___
*[OpenSim/Simulation/Wrap/PathWrapPoint.h, line 60 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3163#-MyTViob6TcLvQXb-ilm:-MyTViob6TcLvQXb-iln:b-jqt7gj) ([raw file](https://github.com/opensim-org/opensim-core/blob/4fdc227a4da643177b1f9d67e9962823b5ca57ca/OpenSim/Simulation/Wrap/PathWrapPoint.h#L60)):*
> ```objc
>     SimTK::Vec3 getLocation(const SimTK::State&) const override;
>     void setLocation(const SimTK::State&, const SimTK::Vec3&) const;
> ```

I understand this location business used to be there due to old inheritance but it looks wrong/confusing since this object is actually a container of a bunch of locations. If used elsewhere need to add comments explaining. Same for functions below
___
*[OpenSim/Simulation/Wrap/PathWrapPoint.cpp, line 2 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3163#-MyTWZRh1EDTlU2C26R_:-MyTWZRh1EDTlU2C26Ra:b-lzlfzu) ([raw file](https://github.com/opensim-org/opensim-core/blob/4fdc227a4da643177b1f9d67e9962823b5ca57ca/OpenSim/Simulation/Wrap/PathWrapPoint.cpp#L2)):*
> ```cpp
>  *                         OpenSim:  PathWrapPoint.h                          *
> ```

.cpp !!
___
*[OpenSim/Simulation/Wrap/PathWrapPoint.cpp, line 10 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3163#-MyTWeHP3ukd1uqeEHOU:-MyTWeHP3ukd1uqeEHOV:bb6ht37) ([raw file](https://github.com/opensim-org/opensim-core/blob/4fdc227a4da643177b1f9d67e9962823b5ca57ca/OpenSim/Simulation/Wrap/PathWrapPoint.cpp#L10)):*
> ```cpp
>  * through the Warrior Web program.                                           *
>  *                                                                            *
>  * Copyright (c) 2005-2017 Stanford University and the Authors                *
> ```

2022
___
*[OpenSim/Simulation/Wrap/PathWrapPoint.cpp, line 90 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3163#-MyTXHYHAXCgVOwJo5DS:-MyTXHYHAXCgVOwJo5DT:b-454uor) ([raw file](https://github.com/opensim-org/opensim-core/blob/4fdc227a4da643177b1f9d67e9962823b5ca57ca/OpenSim/Simulation/Wrap/PathWrapPoint.cpp#L90)):*
> ```cpp
> 
> // these methods were initially adapted from `OpenSim::Station`'s implementation
> ```

Is there a way to avoid duplicating this code? Or help developers keep in sync. with OpenSim::Station?

adamkewley:
> Is there a way to avoid duplicating this code? Or help developers keep in sync. with OpenSim::Station?

@aymanhab Ideally, the calculations would be ripped into free-standing functions that take all the dependent inputs. I didn't want to do that here because it would involve changes to `Station` and additions to some other mathsy-looking place.

Adding a maths implementation for stuff like this somewhere in OpenSim might be useful in general though: there's a few things I've spotted recently (esp. in the wrapping code) where it feels like it would be nice to have access to a few handy standalone functions, rather than needing to pull in a whole `OpenSim::Component` (at least, until something like SimTK implements direct support for it).

adamkewley:
> I'm curious if the crash hypothesized to be caused by this issue has been fixed with the change here.

This particular bug isn't a *crash*; rather, it's a *correctness* violation that then causes bizzare downstream bugs. 

A visual example shows it best. Here is OSC's simulator screen **paused** on a single frame (i.e. a single `SimTK::State`)of a forward-dynamic simulation without the hack I linked in the original PR:

https://user-images.githubusercontent.com/4730570/159122713-3b6700e4-21df-43f5-b3fe-b5db6f327192.mp4

The simulation screen in OSC only has access to `const-`qualified states + models (it can't edit either - it's purely a view-only UI).

The reason why the path points are jolting around like that is because of the bug this PR fixes. There's a simulation thread running concurrently with the UI (the progress bar that's moving along). That simulation thread is then throwing states "over the fence" to the UI, which might perform some initial steps (e.g. re-realizing the state to `Report`) before "freezing" the state somewhere (e.g. in a vector of states that can be scrubbed). That realization step is then--regardless of `const`--mutating the model, because evaluating the path will indirectly stuff new path points into the model-level wrap point collection.

The UI currently works around this by re-realizing the state against the model whenever it's going to be shown in the UI (`generateDecorations` is heavily affected by this because it requires reading through wrapping points to generate path spheres etc.). Post-PR, the hack doesn't seem to be necessary and the model is motionless.

The related PR (that removes pointers from cache variables) fixes an actual crash (not shown in this footage). The crash is a segfault that can happen *right* at the end of watching a simulation. The reason that crash happens is because the UI thread ends up with pointers into the simulator thread's copy of the model. Once the simulator thread is finished, it destructs its copy, but the UI thread still contains pointers into it, which then segfaults in a use-after-free manner.

aymanhab (commented):
> Reviewed 1 of 4 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3163)* status: all files reviewed, 1 unresolved discussion (waiting on @adamkewley)

aymanhab:
Thanks @adamkewley LGTM but please fix test case(s)

adamkewley:
@aymanhab , I managed to fix it *but* it requires changing a test suite that's exercising some part of the JNI bindings (see: https://github.com/opensim-org/opensim-core/pull/3163/commits/eb616e63cda2590193510241514082f774776951)

I think the updated version is probably the "correct" approach, in that the new implementation *requires* that the state is realized to `SimTK::Position` in order to see the position of path points, it might cause downstream issues if there's code in (e.g.) the GUI that relies on the fact that paths are renderable at an earlier stage (e.g. topology).

What's the best solution here? Keep the cache variable as `Position`-dependent or (incorrectly) see if it's possible to get it to work with `Topology`? The cache variable in question is *probably* the `getLocation` API of `AbstractPathPoint` (I think)

Bindings/Java/OpenSimJNI/Test/testContext.cpp
```diff
@@ -72,7 +72,9 @@ int main()
     LoadOpenSimLibrary("osimJavaJNI");
 
     Model *model = new Model("wrist.osim");
-    OpenSimContext* context = new OpenSimContext(&model->initSystem(), model);
+    SimTK::State* workingState = &model->initSystem();
+    model->realizePosition(*workingState);
+    OpenSimContext* context = new OpenSimContext(workingState, model);
     const ForceSet& fs = model->getForceSet();
```
aymanhab, line 78: @adamkewley Thanks for looking into this. The GUI never uses a Model realized below Position stage since it needs to display it, requiring position info. Will have to see how it plays with GUI but probably better to call the non-virtual "realizePosition"  method of OpenSimContext at the end of the OpenSimContext constructor and then client code and the GUI doesn't need to change. If the test cases pass feel free to merge. Thank you

aymanhab (approved):
> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3163)* status: all files reviewed, 2 unresolved discussions (waiting on @adamkewley)

aymanhab (approved):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3163)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adamkewley)

aymanhab:
Great work @adamkewley Thank you 👍

Issue #3130, opened 2022-02-01
Handle unspecified body mass gracefully


aymanhab:
As of now if unspecified, NaN is assigned on loading but downstream random crashes happen when trying to initSystem. Found while investigating crash reported on forum 
https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=14321&p=0&start=0&view=&sid=3c49923bc91e382d3ec531057cb55035

carmichaelong:
From dev team meeting:

Throw an exception if loading a model that has any mass values that are not non-negative numbers. Possible error message could be "Body <body_name> has a non-negative mass value."

aymanhab:
@carmichaelong If we throw then we'll abort loading the model altogether, subsequent issues will not be reported and kinematic-only models are in limbo. Should we just issue a warning/message instead?

PR #3182, opened 2022-04-08
Check for NaN or -ve mass specification and follow as invalid inertia

aymanhab:
Fixes issue #3130 

### Brief summary of changes
Expanded try-block that used to throw if bad inertia is specified to include check for NaN or -ve mass specification.

### Testing I've completed
None other than stepping through the code with bad model.

Should we throw for consistency of handling of inertia as opposed to allowing loading models and catch downstream which would be wider spread and unsustainable AFAICT.

### CHANGELOG.md (choose one)

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3182)

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -64,7 +64,7 @@ jobs:
         # /W0 disables warnings.
         # https://msdn.microsoft.com/en-us/library/19z1t1wy.aspx
         # TODO: CMake provides /W3, which overrides our /W0
-        cmake -E env CXXFLAGS="/W0" cmake $env:GITHUB_WORKSPACE/dependencies -LAH -G"Visual Studio 16 2019" -A x64 -DCMAKE_INSTALL_PREFIX=~/opensim_dependencies_install -DSUPERBUILD_ezc3d=ON
+        cmake -E env CXXFLAGS="/W0" cmake $env:GITHUB_WORKSPACE/dependencies -LAH -G"Visual Studio 16 2019" -A x64 -DCMAKE_INSTALL_PREFIX=~/opensim_dependencies_install -DSUPERBUILD_ezc3d=ON -DOPENSIM_WITH_TROPTER=ON -DOPENSIM_WITH_CASADI=ON
         cmake --build . --config Release -- /maxcpucount:4
 
     - name: Configure opensim-core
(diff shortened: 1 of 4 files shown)


aymanhab:
@adamkewley can you review this PR to make sure it doesn't cause problems to the "Creator" codebase? If mass was not specified it was initialized to NaN causing problems to downstream tools while bad inertia was throwing. This PR makes NaN or -ve mass also throw which is more consistent rather than sprinkling checks downstream. Alternatively we can make Body constructor set 0 mass instead of NaN.

adamkewley:
The behavior is fine for me, probably :wink: - I'll handle any downstream errors separately.

Implementation-wise, I'd recommend:

- Only checking if `m >= 0.0`. Any negative values *and* `NaN` will be caught by `m >= 0.0`. `NaN` is special edge-case that always compares false with all relational and comparison operators (incl. `NaN == NaN`). The literal `0` should ideally be avoided, because it's an `int` literal that relies on automatic promotion for comparison with a `double`. The promotion is well-defined from C, but it is advisable to use the direct comparison.

- Perform the check *and then* handle any error cases. The new implementation unconditionally allocates the error message string (`msg`), even if there is no error. Something like this would work around that:

```c++
if (m >= 0.0) {
    // allocate the error message + exception state in here
    // throw
}
```

- Pry the not-SimTK checks out of the `try` block and throw from outside it if there's a component-specific throwing macro available in OpenSim (e.g: [OPENSIM_THROW_FRMOBJ](https://github.com/opensim-org/opensim-core/blob/master/OpenSim/Common/Exception.h#L85)). Ideally, only surround the throwing part of the call with a `try...catch` block (I'm assuming the original `try...catch` exists to handle the exception thrown by `SimTK::MassProperties{...}`). The current implementation will `throw` from your new check followed by the exception being immediately be re-caught, followed by a further string append (of redundant information) and re-thrown.

aymanhab (commented):
Thanks for the quick turnaround and suggestions. I'll remove the explicit NaN check as you suggested. I prefer not to change where/who throws/catches so that clients are not affected and the change is minimal. Switching to OPENSIM_THROW_IF_FROMOBJ takes care of the allocation issue though the exception may look different from Inertia error. 

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3182#-:-N-U0xP93K3I6lRJji9J:b17n9w3)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @adamkewley and @carmichaelong)

adamkewley:
No prob.

It might be that you end up needing `!(m >= 0.0)` wherever, which might end up looking uglier than the `isnan`: whatever works best - it's only a small change

aymanhab:
@adamkewley is this good to go? Thanks

adamkewley:
Yes, it looks fine :)

aymanhab:
Thanks @adamkewley

Issue #3168, opened 2022-03-22
Moco examples causing segmentation fault on M1 Mac


DanielFNG:
I recently followed the install from source instruction on an M1 Mac so I could use a later version of the system Python when scripting. However, I've found that the Moco C++ examples are now no longer working. They seem to cause a segmentation fault at some point during the study.solve() call e.g:

```
This is Ipopt version 3.12.8, running with linear solver mumps.
NOTE: Other linear solvers might be more efficient (see Ipopt documentation).

Number of nonzeros in equality constraint Jacobian...:     4400
Number of nonzeros in inequality constraint Jacobian.:        0
Number of nonzeros in Lagrangian Hessian.............:        0

zsh: segmentation fault  ./exampleTracking
```

```
This is Ipopt version 3.12.8, running with linear solver mumps.
NOTE: Other linear solvers might be more efficient (see Ipopt documentation).

Number of nonzeros in equality constraint Jacobian...:   412582
Number of nonzeros in inequality constraint Jacobian.:        0
Number of nonzeros in Lagrangian Hessian.............:        0

zsh: segmentation fault  ./example2DWalking
```

It's not a general environment issue, because the plain OpenSim C++ examples compile and run just fine i.e in the simple optimization example:

```
objective evaluation #: 33 elbow flexion angle = 99.5574 BICshort moment arm  = 0.0493903
Elapsed time = 0s
OpenSim example completed successfully. 
```

I'm just wondering if there's any known issues when compiling from source on ARM or if there are any special steps to take in the compilation process. Is it recommended to compile for x86 in the meantime? Cheers!

nickbianco:
Hi @DanielFNG, we are also encountering this issue with the new Macs. We don't have a solution yet, but we can update our progress here.

carmichaelong:
```
The following tests FAILED:
	  3 - test_generic_optimization (SEGFAULT)
	  4 - test_sliding_mass_minimum_effort (SEGFAULT)
	  6 - test_analytic_optimal_control_solutions (SEGFAULT)
	  7 - test_double_pendulum (SEGFAULT)
	  9 - test_optimal_control_initial_guess (SEGFAULT)
	 10 - test_parameter_optimization (SEGFAULT)
	 60 - testMocoInterface (SEGFAULT)
	 61 - testMocoGoals (SEGFAULT)
	 62 - testMocoParameters (SEGFAULT)
	 63 - testMocoImplicit (SEGFAULT)
[...]

PR #3192, opened 2022-04-26
[WIP] fix issues with dependencies build with Macs with arm64

carmichaelong:
FYI @aymanhab @nickbianco 

Related to #3168 

### Brief summary of changes
- Update superbuild option for a Mac with arm64 architecture
- Add mechanism for using a pre-installed IPOPT

### Testing I've completed
Works locally, though will fail often when using many cores (e.g., `make -j8`) and sporadically with fewer (e.g., `make -j4`). There seems to be a race somewhere for making directories during superbuild. Runners don't seem to be available for GitHub actions yet.

This version passes all OpenSim tests.

- Is this more surgical approach OK, or should we switch to this for all Mac builds?
- Is adding the option to use a pre-installed IPOPT overkill? I mainly used it for quicker testing locally when trying to isolate dependencies build issues.
- Not necessarily looking to merge right now (hence [WIP] tag), though this branch will be helpful to make sure I don't break other builds.

### CHANGELOG.md (choose one)
TODO

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3192)

### DIFF
--- a/dependencies/CMakeLists.txt
+++ b/dependencies/CMakeLists.txt
@@ -210,26 +210,33 @@ option(OPENSIM_WITH_TROPTER
 
 
 if(OPENSIM_WITH_TROPTER OR OPENSIM_WITH_CASADI)
-AddDependency(NAME    eigen
-              DEFAULT ON
-              URL     https://gitlab.com/libeigen/eigen/-/archive/3.3.7/eigen-3.3.7.zip)
+    AddDependency(NAME    eigen
+                  DEFAULT ON
+                  URL     https://gitlab.com/libeigen/eigen/-/archive/3.3.7/eigen-3.3.7.zip)
 
-mark_as_advanced(SUPERBUILD_eigen)
+    mark_as_advanced(SUPERBUILD_eigen)
 
-AddDependency(NAME colpack
-              DEFAULT ON
-              GIT_URL https://github.com/opensim-org/colpack.git
-              GIT_TAG 72f691e91d59e8eb2123f258e67a4ddc72d105ee
-              CMAKE_ARGS -DCMAKE_DEBUG_POSTFIX:STRING=_d
-                         -DENABLE_EXAMPLES:BOOL=OFF
-                         -DCMAKE_POSITION_INDEPENDENT_CODE:BOOL=ON)
+    AddDependency(NAME colpack
+                  DEFAULT ON
+                  GIT_URL https://github.com/opensim-org/colpack.git
+                  GIT_TAG 72f691e91d59e8eb2123f258e67a4ddc72d105ee
+                  CMAKE_ARGS -DCMAKE_DEBUG_POSTFIX:STRING=_d
+                             -DENABLE_EXAMPLES:BOOL=OFF
+                             -DCMAKE_POSITION_INDEPENDENT_CODE:BOOL=ON)
 
-mark_as_advanced(SUPERBUILD_colpack)
+    mark_as_advanced(SUPERBUILD_colpack)
 
-CMAKE_DEPENDENT_OPTION(SUPERBUILD_adolc "Automatically download, configure, build and install adolc" ON
-                       "OPENSIM_WITH_TROPTER" OFF)
-CMAKE_DEPENDENT_OPTION(SUPERBUILD_ipopt "Automatically download, configure, build and install ipopt" ON
-                       "OPENSIM_WITH_TROPTER OR OPENSIM_WITH_CASADI" OFF)
+    CMAKE_DEPENDENT_OPTION(SUPERBUILD_adolc "Automatically download, configure, build and install adolc" ON
+                        "OPENSIM_WITH_TROPTER" OFF)
+    CMAKE_DEPENDENT_OPTION(SUPERBUILD_ipopt "Automatically download, configure, build and install ipopt" ON
+                        "OPENSIM_WITH_TROPTER OR OPENSIM_WITH_CASADI" OFF)
+    CMAKE_DEPENDENT_OPTION(USE_IPOPT_INSTALL "Use an ipopt install folder instead of using the superbuild option." OFF
+                        "OPENSIM_WITH_TROPTER OR OPENSIM_WITH_CASADI" OFF)
+    set(IPOPT_INSTALL_DIR "" CACHE STRING "Path to the ipopt install folder when USE_IPOPT_INSTALL is ON")
+endif()
+
+if (SUPERBUILD_ipopt AND USE_IPOPT_INSTALL)
+    message(FATAL_ERROR "SUPERBUILD_ipopt and USE_IPOPT_INSTALL cannot both be ON")
 endif()
 
 if (OPENSIM_WITH_CASADI)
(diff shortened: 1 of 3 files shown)


dependencies/CMakeLists.txt
```diff
@@ -319,34 +326,99 @@ else()
...
+
+        # For ARM-based Macs, use different version of METIS and MUMPS than
+        # those automatically pulled from IPOPT's ThirdParty submodules.
+        if(APPLE AND MACHINE_ARCH STREQUAL "arm64")
+            
+            # Homebrew dir is /opt/homebrew for arm64 (/usr/local for x86_64).
+            # Use METIS from Homebrew, which is currently 5.1.0 (IPOPT 3.12.8
+            # automatically pulls 4.10.0, which had some segfault issues)
+            set(HOMEBREW_DIR "/opt/homebrew")
+            set(METIS_INSTALL_DIR "${HOMEBREW_DIR}/opt/metis")
```
aymanhab, line 350: Is homebrew the only way to get metis and are these directories hardcoded/unchangeable by users?

carmichaelong: > Is homebrew the only way to get metis

There are a few ways to get metis, but there seemed to be issues with using anything but the homebrew version:
1. Building from source from [here](http://glaros.dtc.umn.edu/gkhome/metis/metis/download): This built properly, however, the library that comes out does not seem to have the correct path stored in the .dylib when using `otool -L libmetis.dylib` to check dependencies (e.g., it seems relative rather than absolute), so there are linking problems downstream. I suspect there is some issue with the provided configure/cmake step and RPATH, as there is a warning about `MACOS_RPATH`.
2. Building from coin-or's repo `ThirdParty-Metis` ([link](https://github.com/coin-or-tools/ThirdParty-Metis)). The latest version here, however, is only 4.10.0, which had segfault issues downstream at runtime.
3. homebrew: automatically installs 5.1.0, and using this resolved segfaults.

> are these directories hardcoded/unchangeable by users

With homebrew, the directory structure is set by homebrew. For `arm64` installation, the base directory is `/opt/homebrew`. For `x86_64`, these are kept separately in `/usr/local`. From there, the relative directory of `opt/metis` should also be the same. These directories are set by homebrew when installing without any input from users. Whether users could change these afterwards, I'm not sure, but my understanding is that that would not be a typical use of homebrew.

nickbianco (changes requested):
> Is this more surgical approach OK, or should we switch to this for all Mac builds?

I think there's an argument for keeping the build instructions the same for all Mac (and possibly) Windows builds. Especially if we know that the homebrew approach is more reliable than retrieving the libraries from coin-or. Homebrew is simpler than downloading from coin-or, at least in my opinion. If we wanted to do something similar for Windows, hopefully Chocolately can provide Metis/Mumps in the same way (but I'm getting a little off-topic now).

> Is adding the option to use a pre-installed IPOPT overkill? I mainly used it for quicker testing locally when trying to isolate dependencies build issues.

I could see this as an advanced option, but not sure how many people would use it. If it's useful as a development tool going forward I would be fine keeping it too. 

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3192#-:-N1dxvB80ZsRMzWypJVz:b-4460d4)* status: all files reviewed, 2 unresolved discussions (waiting on @aymanhab and @carmichaelong)

___
*[`dependencies/ipopt.patch` line 7 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3192#-N1dwe_z2dknmqM6zB7x:-N1dwe_z2dknmqM6zB7y:b-8motz) ([raw file](https://github.com/opensim-org/opensim-core/blob/c94601cb28d602420153007b1068c9986db996b2/dependencies/ipopt.patch#L7)):*
> ```Diff
>  // In newer ThirdParty/Mumps, mpi.h is renamed to mumps_mpi.h.
>  // We get informed about this by having COIN_USE_MUMPS_MPI_H defined.
> +#include "mumps_compat.h"
> ```

Why is this new include needed?

carmichaelong (commented):
> Homebrew is simpler than downloading from coin-or, at least in my opinion. If we wanted to do something similar for Windows, hopefully Chocolately can provide Metis/Mumps in the same way (but I'm getting a little off-topic now).

This would be great, but just to highlight a slight difference right now: The current flow is to only use homebrew for metis. We still rely on coin-or for mumps. However, instead of relying IPOPT 3.12.8 to grab mumps automatically, we grab a different branch (more info on this on your comment at a specific line about the include line). For 4.4, I think it'll be good to keep this surgical as it should only affect a few users (i.e., new M1 owners building Moco from source), but we should consider bumping to newer versions down the road, which may help unify IPOPT versions across different platforms and simplify the dependencies build.

> I could see this as an advanced option, but not sure how many people would use it. If it's useful as a development tool going forward I would be fine keeping it too.

Sounds good, will mark the options advanced.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3192#-:-N1yHiy4EzCJNQTwYmbs:b23wzc7)* status: all files reviewed, 2 unresolved discussions (waiting on @aymanhab and @nickbianco)

___
*[`dependencies/ipopt.patch` line 7 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3192#-N1dwe_z2dknmqM6zB7x:-N1y_n50-UHAY4LzGauj:bqjirh8) ([raw file](https://github.com/opensim-org/opensim-core/blob/c94601cb28d602420153007b1068c9986db996b2/dependencies/ipopt.patch#L7)):*

Why is this new include needed?
</blockquote></details>

IPOPT 3.12.8 automatically pulls `coin-or-tools/ThirdParty-Mumps` branch `stable/1.6` which uses MUMPS 4.10.0. This, however, does not work with the homebrew version of METIS due to some linker errors (likely because homebrew uses METIS 5.1.0, but `coin-or-tools/ThirdParty-Metis` uses METIS 4.10.0).

So instead, this modified superbuild pulls `coin-or-tools/ThirdParty-Mumps` branch `stable/2.1`. This does not have linker errors and still uses MUMPS 4.10.0. My guess is that IPOPT 3.12.8 expected `COIN_USE_MUMPS_MPI_H` to be defined in a different place for `stable/1.6` and for `stable/2.1`. `stable/2.1 conveniently defines many of these variables in this `mumps_compat.h` file. 

(As an aside, IPOPT 3.13+ uses `stable/2.0+` branches of `ThirdParty-Mumps`. IPOPT 3.13+ made a breaking change for how dependencies worked, so this gives further support that there was major build changes for these two `ThirdParty-Mumps` branches too.)

nickbianco (approved):
> For 4.4, I think it'll be good to keep this surgical

Totally agree. Thanks for the clarification on the include comment below and on how all these dependencies work (both here and in person). 

LGTM.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3192#-:-N1zikTFCsKx0tJNOTef:b-a0mn24)* status: all files reviewed, 1 unresolved discussion (waiting on @aymanhab)

___
*[`dependencies/ipopt.patch` line 7 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3192#-N1dwe_z2dknmqM6zB7x:-N1zifUIA4dZ6WEO-SHH:b-uv4tn) ([raw file](https://github.com/opensim-org/opensim-core/blob/c94601cb28d602420153007b1068c9986db996b2/dependencies/ipopt.patch#L7)):*

IPOPT 3.12.8 automatically pulls `coin-or-tools/ThirdParty-Mumps` branch `stable/1.6` which uses MUMPS 4.10.0. This, however, does not work with the homebrew version of METIS due to some linker errors (likely because homebrew uses METIS 5.1.0, but `coin-or-tools/ThirdParty-Metis` uses METIS 4.10.0).

So instead, this modified superbuild pulls `coin-or-tools/ThirdParty-Mumps` branch `stable/2.1`. This does not have linker errors and still uses MUMPS 4.10.0. My guess is that IPOPT 3.12.8 expected `COIN_USE_MUMPS_MPI_H` to be defined in a different place for `stable/1.6` and for `stable/2.1`. `stable/2.1 conveniently defines many of these variables in this `mumps_compat.h` file. 

(As an aside, IPOPT 3.13+ uses `stable/2.0+` branches of `ThirdParty-Mumps`. IPOPT 3.13+ made a breaking change for how dependencies worked, so this gives further support that there was major build changes for these two `ThirdParty-Mumps` branches too.) 
</blockquote></details>

Thanks for the clarification!

aymanhab:
@carmichaelong Just wanted to make sure I'm following your steps exactly:
1. I merged your branch onto 4.4
2. on dependencies build turn-off superbuild_ipopt, turn on use_ipopt_install
3. is there an assumed separate "brew install ipopt" or is that baked into the steps somehow? 
4. Did you investigate using superbuild with a different hash(es) that support arm64 out of the box?
Thank you

carmichaelong:
@aymanhab sorry, it was probably pretty unclear which instructions to use because it was a long discussion. I did make it work with the superbuild option, after trying a bunch of different ways to get metis, mumps, and ipopt. See the instructions below:
```
brew install cmake swig gcc pkgconfig autoconf libtool automake wget doxygen metis
git clone https://github.com/opensim-org/opensim-core.git
cd opensim-core
git switch dependencies-mac-arm64
cd ..
mkdir opensim_dependencies_build
cd opensim_dependencies_build
cmake ../opensim-core/dependencies \
      -DCMAKE_INSTALL_PREFIX="../opensim_dependencies_install" \
      -DCMAKE_BUILD_TYPE=RelWithDebInfo \
      -DSUPERBUILD_ezc3d=ON \
      -DOPENSIM_WITH_CASADI=ON \
      -DOPENSIM_WITH_TROPTER=ON
make -j4
cd ..
mkdir opensim_build
cd opensim_build
cmake ../opensim-core \
      -DCMAKE_INSTALL_PREFIX="../opensim_install" \
      -DCMAKE_BUILD_TYPE=RelWithDebInfo \
      -DBUILD_PYTHON_WRAPPING=OFF \
      -DBUILD_JAVA_WRAPPING=OFF \
      -DOPENSIM_DEPENDENCIES_DIR="../opensim_dependencies_install"
make -j8
ctest -j8
```

The `use_ipopt_install` was a flag I was using often when trying different IPOPT installs. If we keep it, it would stay as advanced, but I'm not sure if we should keep it or not.

carmichaelong:
Also some responses to questions above to be clearer:

> is there an assumed separate "brew install ipopt" ?
> I guess I'm confused because you're not superbuilding ipopt but applying a patch, does it mean ipopt has to be built manually after the patch or are this taken care of by the make files?

The last version I had adjusted the superbuild to work by pulling in and building a specific version of IPOPT, and then doing a small patch, and the superbuild should do this all for you.

> Did you investigate using superbuild with a different hash(es) that support arm64 out of the box, or is that an overkill?

Buried in the issue discussion linked has some discussion on the different combinations of mumps, metis, and ipopt versions I've tried. Some of them build, some of them build but fail tests, and some of them build and pass tests (and one of these combinations should be the current superbuild settings here). I'm not sure maybe about what hash(es) you are referring to, so please point out what that might mean and maybe I can remember some info that might be useful there!

nickbianco:
@carmichaelong @aymanhab I think this can be closed now?

Issue #3087, opened 2021-09-29
MATLAB MarkersReference not recognized in InverseKinematicsSolver constructor


CVHammond:
Using OpenSim 4.3, when declaring a new InverseKinematicsSolver, MATLAB returns a 'no constructor with matching signature' error response despite correct inputs, the issue appears to be in the MarkersReference object.

This bug can be reproduced with the following script on OpenSim 4.2 or 4.3:

```
import org.opensim.modeling.*
model = Model('model.osim')
InverseKinematicsSolver(model, MarkersReference(), SimTKArrayCoordinateReference())
```

If you use the command `methodsview InverseKinematicsSolver` and look at the constructor definitions, you can see the MarkersReference input SWIG value is an autogenerated value rather than the expected `org.opensim.modeling.MarkersReference`.

This bug does not exist in OpenSim 4.0. `methodsview InverseKinematicsSolver` shows `org.opensim.modeling.MarkersReference` for the MarkersReference parameter and the InverseKinematicsSolver can be declared without error.

tkuchida:
Does your code work in previous versions of OpenSim? Based on [this test case in testInverseKinematicsSolver.cpp](https://github.com/opensim-org/opensim-core/blob/a01cf545485e8bc1f6e50b29ddeab229677de2c5/OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp#L400), it seems like you would need to create a `MarkersReference` first and then pass that variable into the constructor for `InverseKinematicsSolver`.

CVHammond:
My script works as expected on OpenSim 4.0. I am using the method you stated, declaring a `MarkersReference` and passing that into `InverseKinematicsSolver`.

nickbianco:
We recently changed the Matlab and Python bindings to disallow wrapping types `std::shared_ptr<MarkersReference>`, due to issues with compatibility with new Moco tools (the changed was introduced when Moco was merged into OpenSim): https://github.com/opensim-org/opensim-core/blob/f9ef95ec8cd86e45a1df0b0faafcd4d6371dd9bc/Bindings/simulation.i#L224.

[...]

PR #3193, opened 2022-04-26
Fix Bindings for OrientationsReference

Coolcon1729:
This PR is related to [this PR](https://github.com/opensim-org/opensim-core/pull/3089), which fixes [this issue](https://github.com/opensim-org/opensim-core/issues/3087). There was an issue with the bindings related to std::shared_ptr<OpenSim::MarkersReference>, and this PR fixes the analogous bug with std::shared_ptr<OpenSim::OrientationsReference>.

This issue can be reproduced with the following code in python (and probably the analogous code in MATLAB):
```python
import opensim

model = opensim.Model()
markersRef = opensim.MarkersReference()
orientationsRef = opensim.OrientationsReference()
coordRefArray = osim.SimTKArrayCoordinateReference()

iksolver = opensim.InverseKinematicsSolver(model, markersRef, orientationsRef, coordRefArray)
```

### Brief summary of changes
* Modified bindings for std::shared_ptr\<OpenSim::OrientationsReference>
to match the bindings of std::shared_ptr\<OpenSim::MarkersReference>

### Testing I've completed
To ensure the bug is with the bindings and not from some other source, I tested the following C++ code, with no errors.
```C++
#include <OpenSim/OpenSim.h>
using namespace SimTK;
using namespace OpenSim;

int main() {
    Model model;
    MarkersReference markersRef;
    OrientationsReference orientationsRef;
    SimTK::Array_<CoordinateReference> coordRefArray;

    InverseKinematicsSolver ik(model, markersRef, orientationsRef, coordRefArray);
}
```
To test the changes to the bindings, I compiled successfully and tested InverseKinematicsSolver constructor with the optional OrientationsReference argument.

* I copied the code comment for the bindings of MarkersReference and modified slightly for OrientationsReference. It is likely better to have just one comment that explains the whole situation. However, since I do not know all of the intricacies of this facet of opensim, I would feedback on how I should comment this.
* It is probably better to swap line 236 and 237, to keep sequential consistency with MarkersReference. However, I was advised that the order of the bindings could affect some things, so I would like feedback on whether it is safe/advisable to swap these lines.

### CHANGELOG.md (choose one)

- no need to update because... not modifying interface or functionality, just a bug fix.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3193)

### DIFF
--- a/Bindings/simulation.i
+++ b/Bindings/simulation.i
@@ -229,8 +229,12 @@ OpenSim::ModelComponentSet<OpenSim::Controller>;
 %template(SetMarkerWeights) OpenSim::Set<MarkerWeight, OpenSim::Object>;
 %include <OpenSim/Simulation/CoordinateReference.h>
 %include <OpenSim/Simulation/OrientationsReference.h>
+// Since we have both OrientationsReference and shared_ptr<OrientationsReference>
+// Across the interface, DO NOT use %shared_ptr macro here as it treats
+// all pointers and references as std::shared_ptr
+//
 %template (SetOientationWeights) OpenSim::Set<OrientationWeight, OpenSim::Object>;
-%shared_ptr(OpenSim::OrientationsReference);
+%template(SharedOrientationsReference) std::shared_ptr<OpenSim::OrientationsReference>;
 %include <OpenSim/Simulation/BufferedOrientationsReference.h>
 %shared_ptr(OpenSim::BufferedOrientationsReference);
 


aymanhab:
@tkuchida This is likely ok, but will wait/test when I create the test case in #3056 as it is the primary use case of OrientationsReference from Python.

aymanhab:
Thanks @tkuchida and @Coolcon1729 I'm in the process of making the test case and need this anyway so I'm merging it.

aymanhab (commented):
LGTM

tkuchida:
:+1: Confirmed fixed in 4.4 beta :duck:

Issue #3191, opened 2022-04-25
Switching the order of coordinates in a joint changes the joint definition


nickbianco:
To reproduce with model Gait2354 (gait2354_simbody.osim): 

If you switch the order of the coordinates hip_flexion_r (lines 377-394) and hip_adduction_r (lines 395-412) in the CoordinateSet via modifying the XML file, and reload the model in the GUI, the coordinate slider for flexion controls adduction and the one for adduction controls flexion. 

Reported by Jennifer Maier.

aymanhab:
Reproduced in local environment, investigating...

aymanhab:
This is actually more prevalent than initially assumed, you can switch rotational with translational coordinates and motion-type is ignored. Identified the problematic call, testing a fix but worry about the scope considering prevalence.

PR #3196, opened 2022-05-04
Fix coordinate order xml

aymanhab:
Fixes issue #3191

### Brief summary of changes
Changing the order of Coordinates in XML changes that actual mobilities/motions that the coordinates are attached to for a CustomJoint. This PR reorders the Coordinates to match the expectation/contract of downstream code.
### Testing I've completed
Created various combination of model files with different coordinate orderings
### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3196)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ v4.4
 - Fixed objects being set as not up to date with their properties by finalizeFromProperties
 - Throw exception if body masses are either NaN or -ve (Issue #3130)
 - Fixed issue #3176 where McKibbenActuator is not registered and can't be serialized to XML files
+- Fixed issue #3191 where CustomJoint coordinates ordering in model files affects coordinate definitions.
 
 
 v4.3
--- a/OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp
@@ -157,10 +157,11 @@ void CustomJoint::constructCoordinates()
         if (coordIndex < 0) {
             coordIndex =
                     constructCoordinate(Coordinate::MotionType::Undefined, i);
-        } else {
-            int sizeAsInt = static_cast<int>(coordIndices.size());
+        } else { 
+            // Coordinate was defined in XML, make sure the order in spatialTransform
+            // matches the order in the XML file/property
             coordinatesInTransformOrder = coordinatesInTransformOrder && 
-                    coordIndex == sizeAsInt;
+                    coordIndex == i;
             coordIndices.push_back(coordIndex);
         }
         Coordinate& coord = upd_coordinates(coordIndex);
--- a/OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp
@@ -158,8 +158,9 @@ void CustomJoint::constructCoordinates()
             coordIndex =
                     constructCoordinate(Coordinate::MotionType::Undefined, i);
         } else {
+            int sizeAsInt = static_cast<int>(coordIndices.size());
             coordinatesInTransformOrder = coordinatesInTransformOrder && 
-                    coordIndex == coordIndices.size();
+                    coordIndex == sizeAsInt;
             coordIndices.push_back(coordIndex);
         }
         Coordinate& coord = upd_coordinates(coordIndex);
--- a/OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp
@@ -141,19 +141,26 @@ void CustomJoint::extendScale(const SimTK::State& s, const ScaleSet& scaleSet)
 void CustomJoint::constructCoordinates()
 {
     const SpatialTransform& spatialTransform = getSpatialTransform();
-    OpenSim::Array<std::string> coordinateNames
-        = spatialTransform.getCoordinateNames();
+    OpenSim::Array<std::string> coordinateNames =
+            spatialTransform.getCoordinateNames(); // in order of appearance in
+                                                   // spatialTransform
 
     // Check how many coordinates are required
     int ncoords = coordinateNames.getSize();
-
-    for (int i = 0; i < ncoords; ++i){
+    std::vector<int> coordIndices;
+    bool coordinatesInTransformOrder = true;
+    for (int i = 0; i < ncoords; ++i) {
         std::string coordName = coordinateNames[i];
-        // Locate the coordinate in the set if it has already been defined (e.g. in XML) 
+        // Locate the coordinate in the set if it has already been defined (e.g.
+        // in XML)
         int coordIndex = getProperty_coordinates().findIndexForName(coordName);
         if (coordIndex < 0) {
-            coordIndex = constructCoordinate(Coordinate::MotionType::Undefined,
-                                             i);
+            coordIndex =
+                    constructCoordinate(Coordinate::MotionType::Undefined, i);
+        } else {
+            coordinatesInTransformOrder = coordinatesInTransformOrder && 
+                    coordIndex == coordIndices.size();
+            coordIndices.push_back(coordIndex);
         }
         Coordinate& coord = upd_coordinates(coordIndex);
         coord.setName(coordName);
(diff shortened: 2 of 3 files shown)


nickbianco (changes requested):
Left a few comments. Looks pretty good overall.

> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3196#-:-N1KWjql5VAg6RVFiAdY:b-mebxg9)* status: all files reviewed, 3 unresolved discussions (waiting on @aymanhab)

___
*[`OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp` line 164 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3196#-N1KRczFBqDr_hJekPbF:-N1KRczFBqDr_hJekPbG:b-ayjkko) ([raw file](https://github.com/opensim-org/opensim-core/blob/b92d9553daab64d74fb848edabdf9631117e007c/OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp#L164)):*
> ```cpp
>             coordinatesInTransformOrder = coordinatesInTransformOrder && 
>                     coordIndex == sizeAsInt;
>             coordIndices.push_back(coordIndex);
> ```

Should this line be inside the `else` branch? It seems like something could go wrong if the first branch is reached and the `coordIndices` vector isn't updated. 
___
*[`OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp` line 217 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3196#-N1KWf5OBHw_lveZTGdV:-N1KWf5OBHw_lveZTGdW:b5ktgf9) ([raw file](https://github.com/opensim-org/opensim-core/blob/b92d9553daab64d74fb848edabdf9631117e007c/OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp#L217)):*
> ```cpp
>         // Copy back values to proper entry in Property_coordinates.
>         // Ideally this would be done using copy construction or operators
>         // unfortunately these wipe out pointers thus breaking downstream code.
> ```

What pointers are destroyed?
___
*[`OpenSim/Simulation/SimbodyEngine/Joint.cpp` line 764 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3196#-N1KVlkl22WqpXusKDsS:-N1KVlkl22WqpXusKDsT:b-64sft8) ([raw file](https://github.com/opensim-org/opensim-core/blob/b92d9553daab64d74fb848edabdf9631117e007c/OpenSim/Simulation/SimbodyEngine/Joint.cpp#L764)):*
> ```cpp
>     for (int iq = 0; iq < numMobilities; ++iq){
>         if (j < nc){ // assign
>             //<=== This assumes coordinates as PROPERTY are in same order of mobilities
> ```

So the joint creates a list of mobilities based on the number of coordinates in the joint prior to this point?

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3196)* status: all files reviewed, 3 unresolved discussions (waiting on @nickbianco)

___
*[`OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp` line 164 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3196#-N1KRczFBqDr_hJekPbF:-N1KYLlu3lpF4EGAepGY:b-144e0o) ([raw file](https://github.com/opensim-org/opensim-core/blob/b92d9553daab64d74fb848edabdf9631117e007c/OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp#L164)):*

Should this line be inside the `else` branch? It seems like something could go wrong if the first branch is reached and the `coordIndices` vector isn't updated. 
</blockquote></details>

Shouldn't matter but this is preferred for the variable to appear closest to its use (actually didn't have it initially but compilers on *nix complained about comparing int to size_t). Once declared, coordIndices.size always returns zero.  The first branch is left alone, intentionally, since it corresponds to no XML Coordinate entry.
___
*[`OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp` line 217 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3196#-N1KWf5OBHw_lveZTGdV:-N1KZU1F3Iiho_03jEHI:bqs1yqw) ([raw file](https://github.com/opensim-org/opensim-core/blob/b92d9553daab64d74fb848edabdf9631117e007c/OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp#L217)):*

What pointers are destroyed?
</blockquote></details>

Ownership pointers at the Component level and possibly Joint pointers or traversal pointers which are populated elsewhere. The first one throws Orphan component exceptions downstream.
___
*[`OpenSim/Simulation/SimbodyEngine/Joint.cpp` line 764 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3196#-N1KVlkl22WqpXusKDsS:-N1KZyd98qt8lURQ0In5:bb77ufg) ([raw file](https://github.com/opensim-org/opensim-core/blob/b92d9553daab64d74fb848edabdf9631117e007c/OpenSim/Simulation/SimbodyEngine/Joint.cpp#L764)):*

So the joint creates a list of mobilities based on the number of coordinates in the joint prior to this point? 
</blockquote></details>

Yes.

nickbianco (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3196)* status: all files reviewed, 1 unresolved discussion (waiting on @aymanhab)

___
*[`OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp` line 164 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3196#-N1KRczFBqDr_hJekPbF:-N1Ky6Oz-Y3z-U76pEJh:b-euwy7o) ([raw file](https://github.com/opensim-org/opensim-core/blob/b92d9553daab64d74fb848edabdf9631117e007c/OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp#L164)):*

Shouldn't matter but this is preferred for the variable to appear closest to its use (actually didn't have it initially but compilers on *nix complained about comparing int to size_t). Once declared, coordIndices.size always returns zero.  The first branch is left alone, intentionally, since it corresponds to no XML Coordinate entry.
</blockquote></details>

Ah, so during deserialization, only one side of the branch is traversed? That makes sense, since you wouldn't have only half of a joint defined in an XML file. Could you leave a quick comment noting that?

Also, I just realized, couldn't you just used the index `i` to check if the indices are in the right order if only the right branch is traversed during deserialization?

nickbianco (commented):
Thanks @aymanhab. One last thing to clarify, then we should be good to go.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3196#-:-N1LFtZSDDGwotnfvQUx:bq3h32x)* status: all files reviewed, 1 unresolved discussion (waiting on @aymanhab)

aymanhab:
Thanks much @nickbianco 👍 I'm building and testing after making the change you suggested before I push it for final review.

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3196)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @nickbianco)

___
*[`OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp` line 164 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3196#-N1KRczFBqDr_hJekPbF:-N1Mm7PV9FQBKnIdhfLb:b-6sikth) ([raw file](https://github.com/opensim-org/opensim-core/blob/b92d9553daab64d74fb848edabdf9631117e007c/OpenSim/Simulation/SimbodyEngine/CustomJoint.cpp#L164)):*

Ah, so during deserialization, only one side of the branch is traversed? That makes sense, since you wouldn't have only half of a joint defined in an XML file. Could you leave a quick comment noting that?

Also, I just realized, couldn't you just used the index `i` to check if the indices are in the right order if only the right branch is traversed during deserialization?
</blockquote></details>

Done. Thanks @nickbianco

nickbianco (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3196#-:-N1OwOa92YhwhqACA48U:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aymanhab)

Issue #3209, opened 2022-05-23
Update ezc3d Version in Conda Package


teliason:
There is an incompatibility issue when both opensim and ezc3d are installed in the same environment and the ezc3d version does not match what opensim was built with (https://github.com/pyomeca/ezc3d/issues/220).  

From my testing it looks like the current conda opensim package was built with ezc3d 1.4.7.  If possible it would be super helpful if the conda package could be updated with the current ezc3d version 1.4.9. 

Thank you for putting opensim together as a  conda package.  It has been great for streamlining integrating opensim into our python workflow.

pariterre:
Hi @teliason, 
I can push an update to the opensim software, but it won't be available to the actual software until further release (unless you are building OpenSim for yourself). 

@aymanhab 
To my knowledge, there isn't any breaking changes from 1.4.7 to 1.4.9. If you want, I can push an update

aymanhab:
If not breaking we can go to latest released version, we have no reason to
stay behind official released versions.

teliason:
When I built OpenSim with ezc3d 1.4.9 everything worked as expected.  I don't think the new version should break anything on the OpenSim side but I have not extensively tested it.  @aymanhab that would be great if you could update to the latest, while I have it working by building OpenSim ourselves, the conda package is very convenient when sharing our pipeline internally and getting others who are less familiar with OpenSim set up.

pariterre:
@teliason  That is actually on me, but it only consists in updating the commit number and the version number in the dependencies folder. I wanted to build it myself before, but did not have time to do so. Since you tell me that it does not produce error (which I am actually not surprised), I will simply push the modif and trust the merge test on GitHub :)

aymanhab:
Conda package was published with ezc3d 1.5, thanks for the help @pariterre Closing

PR #3212, opened 2022-05-25
Updated to ezc3d 1.4.9

pariterre:
Fixes issue #3209 

### Brief summary of changes
Updated to newest ezc3d version

### Testing I've completed
None - Rely on the tests here

None

### CHANGELOG.md (choose one)
- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3212)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ This is not a comprehensive list of changes but rather a hand-curated collection
 
 v4.4
 ====
-- Updated ezc3d to version 1.4.6 which better manage the events defined in a c3d file.
+- Updated ezc3d to version 1.4.9 which better manage the events defined in a c3d file and deals with ROTATION type of data.
 - Fixed an issue that could happen sometimes with ScaleTool where loading the model file or marker set file could fail if the file was given as an absolute path (Issue #3109, PR #3110)
 - Fixed an issue with SWIG with `OpenSim::Body::getRotationInGround()` where it would return an object without the correct `SimTK::Rotation` methods.
 - Fixed OpenSim::Arrow start_point property being ignored
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -707,15 +707,18 @@ if(${OPENSIM_C3D_PARSER} STREQUAL "ezc3d")
         set(ezc3d_hint "${OPENSIM_DEPENDENCIES_DIR}/ezc3d/lib/cmake/ezc3d")
     endif()
     find_package(ezc3d REQUIRED HINTS "${ezc3d_hint}")
+    set(ezc3d_ROOT_DIR "${ezc3d_DIR}/../../..")
+    set(ezc3d_LIBRARY_DIR "${ezc3d_ROOT_DIR}/lib")
+    set(ezc3d_INCLUDE_DIR "${ezc3d_ROOT_DIR}/include")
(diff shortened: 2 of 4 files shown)


aymanhab:
Thanks @pariterre but the build fails on all platforms trying to locate ezc3d, did anything change about how ezc3d uses CMake?

aymanhab:
Sorry to bug you again about this @pariterre but where does this stand? How are other projects finding EZC3D? Thanks

pariterre:
@aymanhab 
Sorry I haven't had the time to finish this. Actually @teliason changed the CMake in a recent PR and I am trying to figure out this exact thing! I should come back to it soon

pariterre:
Hi @aymanhab 
I've made some tests with the new CMake developed by @SpaceIm (sorry I tagged the wrong developer last time). 
Includes and linkings are now implicitly done using the usual `target_link_libraries`, with a small twist for windows (as one must copy the DLLs). Here is a minimal example:
```cmake
cmake_minimum_required(VERSION 3.5)

project(test_project_that_includes_ezc3d LANGUAGES CXX)
add_executable(test_project_that_includes_ezc3d main.cpp)

set(CMAKE_CXX_STANDARD 11)
set(CMAKE_CXX_STANDARD_REQUIRED ON)

find_package(ezc3d REQUIRED)
target_link_libraries(test_project_that_includes_ezc3d ezc3d)

# Copy the DLL in required folder for windows
if (WIN32)
  add_custom_command(TARGET test_project_that_includes_ezc3d POST_BUILD
    COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_RUNTIME_DLLS:test_project_that_includes_ezc3d> $<TARGET_FILE_DIR:test_project_that_includes_ezc3d>
    COMMAND_EXPAND_LISTS
  )
endif(WIN32)
```
Which allows for:
```c++
#include <ezc3d/ezc3d_all.h>
int main()
{
    ezc3d::c3d emptyC3d;
    emptyC3d.print();
    return 0;
}
```
But, in the case of OpenSim, you are kind of manually moving stuff around to link the library and include the headers. 
I am not sure what is the best way to do that... @spacelm, do you have a suggestion?

aymanhab:
@pariterre Thanks for looking into this, I'm getting closer but I'm have a problem finding out what variables are exported by your find_package(), our CMake files used ezc3d_LIBRARY_DIR, ezc3d_LIBRARY, ezc3d_INCLUDE_DIR and I'm having a hard time finding out what variables are defined by your new config file or what they are set to on various platforms, can you or @spacelm clarify? We're not using the latest cutting edge CMake version/options and we prefer not to so that our users are not forced to upgrade. Thank you

pariterre:
@SpaceIm 
Does this mean that each time an `add_executable` is called, a target_link_library with ezc3d should be called, or some sort of global link is possible?

SpaceIm:
If you want global link, there is https://cmake.org/cmake/help/latest/command/link_libraries.html, but it's a bad practice because you lose the true relationship between your targets.

pariterre:
@aymanhab 
Okay, ping me if you need me. While you are at it, you may actually change 1.4.9 to 1.5.0 since I push another release last week (added the capability to read ROTATION in the c3d files, which won't impact you though)

See you :)

aymanhab:
Closing since this was subsumed by another PR that upgraded to version 1.5

Issue #3094, opened 2021-10-17
Issues with marker tracking using MocoTrack


nickbianco:
Two different users posted on the forum ([here](https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=1815&t=13959&p=40371&start=0&view=) and [here](https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=1815&t=14007&p=0&start=0&view=&sid=cde0446c29c788167532a3fd23274058)) stating that they're having issues solving marker tracking problems with MocoTrack. 

We should re-test these examples to make sure nothing is broken. Probably worth re-testing anyway considering that we fixed the MarkersReference issue recently too.

@jenhicks is it too late to add this to the 4.4 milestone?

BiboyLin:
Because some functions like **OrientationsReference** , **InverseKinematicsSolver** re-refer to a new one, the Opensense code is not working too.

nickbianco:
@BiboyLin, could you clarify how the OpenSense issues are related to this one? Is it something to do with the MarkersReference changes?

aymanhab:
@BiboyLin If there're specific API calls that are not working please add a snippet to the test case involving MocoMakerTrackingGoal so that we can fix and make sure it doesn't break moving forward.

BiboyLin:
> @BiboyLin, could you clarify how the OpenSense issues are related to this one? Is it something to do with the MarkersReference changes?
Here is the IK part of Opensense, the **MarkersReference** is not useful but  **InverseKinematicsSolver** refer to the new one:
```
 # Initialize simulation
        quatTable = osim.TimeSeriesTableQuaternion(sto_filename)
        orientationsData = osim.OpenSenseUtilities.convertQuaternionsToRotations(
            quatTable)
        oRefs = osim.OrientationsReference(orientationsData)
        init_state = model.initSystem()
        mRefs = osim.MarkersReference()
        coordinateReferences = osim.SimTKArrayCoordinateReference()
        model.initSystem()
        s0 = init_state
	# BUG here
        ikSolver = osim.InverseKinematicsSolver(
            model, mRefs, oRefs, coordinateReferences, np.float64(constraint_var))
	# Changing to this, it will go but got wrong at
        # ikSolver = osim.InverseKinematicsSolver(model, mRefs, coordinateReferences, constraint_var)
        ikSolver.setAccuracy = accuracy
        s0.setTime(0.)
        ikSolver.assemble(s0)
[...]

PR #3214, opened 2022-05-31
Fix failing torque-driven marker tracking example in exampleMocoTrack

nickbianco:
Fixes issue #3094

### Brief summary of changes

~~Remove default option to filter marker trajectory table when using `MocoTrack::setMarkersReferenceFromTRC()`.~~

Per @hwarner's suggestion, the issue was not marker filtering, but rather the marker data scaling step. Marker scaling is unnecessary since the MarkersReference class scales marker data based on the model's units automatically.

### Testing I've completed
Ran exampleMocoTrack locally.

~~Is removing the filtering argument entirely the best solution here? Users can also filter using TableOperator utilities, so if we remove it, we can instruct users how to do it that way. We could also provide an additional argument as a flag to turn filtering on/off.~~

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3214)

### DIFF
--- a/CHANGELOG_MOCO.md
+++ b/CHANGELOG_MOCO.md
@@ -3,6 +3,9 @@ Moco Change Log
 
 1.2.0
 -----
+- 2022-06-03: Fixed bug that was breaking marker tracking problems when
+              using MocoTrack::setMarkersReferenceFromTRC().
+
 - 2022-04-15: Added MocoInitialOutputGoal, MocoFinalOutputGoal,
               MocoOutputPeriodicityGoal, MocoOutputTrackingGoal, and
               MocoOutputConstraint.
--- a/OpenSim/Examples/Moco/example3DWalking/exampleMocoTrack.cpp
+++ b/OpenSim/Examples/Moco/example3DWalking/exampleMocoTrack.cpp
@@ -168,7 +168,7 @@ void muscleDrivenStateTracking() {
 
 int main() {
 
-    // Solve tq he torque-driven marker tracking problem.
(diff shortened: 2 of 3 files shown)


carmichaelong (commented):
I'm not sure what the best option is, especially with trying to balance between 1) not changing the API and 2) not making a change that may change answers without an explicit warning/error.

My current interpretation (and definitely let me know if I'm assuming something wrong here): The filtering step in `setMarkersReferenceFromTRC()` is working in isolation, but in some cases causes issues downstream (I suppose this could be tested by trying the same filtering using TableOperator utilities and see if the same issue occurs). Further, most problems are probably best served by NOT filtering here. Finally, it might still be convenient for users to optionally filter in this step.

If the previous is correct, I'm a bit torn between the two options:
1. Keep the method signature the same, but change the default to not apply filtering (perhaps by setting the default input value as -1.0, which indicates no filtering)
2. Go with the current suggested changes in this PR. I like that this is more unambiguous.

Might be worth a quick discussion on Thursday?

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3214#-:-N3Rg0rL5tVxrBghvVVp:b9ns1iy)* status: 0 of 2 files reviewed, all discussions resolved

nickbianco:
@carmichaelong, I've updated this fix to address the marker scaling issue, not filtering (thanks @hwarner!). 

Review should be straight-forward from here.

nickbianco:
Thanks @carmichaelong!

Issue #3223, opened 2022-06-11
MINSIGSTKSZ has been removed from glibc >= 2.34


AlbertoCasasOrtiz:
### Error message:

~/opensim-core-source/OpenSim/Auxiliary/catch.hpp:10798:58: error: call no non-'constexpr' funcion 'long int sysconf(int)'
10798 |    static constexpr std::size_t sigStackSize = 32768 >= MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;

~/opensim-core-source/Vendors/tropter/external/catch/catch.hpp:8726:58: error: call no non-'constexpr' funcion 'long int sysconf(int)'
8726 |    static constexpr std::size_t sigStackSize = 32768 >= MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;

### Problem:

MINSIGSTKSZ is redefined to sysconf(_SC_MINSIGSTKSZ) and SIGSTKSZ is redefined to sysconf (_SC_SIGSTKSZ) in GNU C Library (glibc) version 2.34.
https://sourceware.org/pipermail/libc-alpha/2021-August/129718.html

### Workaround:

substitute
`static constexpr std::size_t sigStackSize = 32768 >= MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;`
with
`static constexpr std::size_t sigStackSize = 32768;`

### Further information:

Many applications, including [catch2](https://github.com/catchorg/Catch2), have been affected by this. In the following issue of catch2 different solutions and workarounds are proposed (including the workaround above):
[https://github.com/catchorg/Catch2/issues/2178](https://github.com/catchorg/Catch2/issues/2178)

The case of catch2 is interesting because they were using exactly the same code, so the solutions applied may be useful here.

AlbertoCasasOrtiz:
Solved in #3225.

PR #3225, opened 2022-06-13
MINSIGSTKSZ has been removed from glibc >= 2.34

AlbertoCasasOrtiz:
This is a workaround for issue #3223. With this workaround you can build opensim-core with glibc >= 2.34, but a better aproach should be proposed.

### Brief summary of changes
Removed the use of MINSIGSTKSZ, which has been redefined in glibc >= 2.34 (It throws error in glib >= 2.34).

### Testing I've completed
Continuous integration workflow tests.

How a better approach  could be implemented here.

### CHANGELOG.md (choose one)

- no need to update because a better approach should be implemented.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3225)

### DIFF
--- a/OpenSim/Auxiliary/catch.hpp
+++ b/OpenSim/Auxiliary/catch.hpp
@@ -10795,7 +10795,7 @@ namespace Catch {
 
     // 32kb for the alternate stack seems to be sufficient. However, this value
     // is experimentally determined, so that's not guaranteed.
-    static constexpr std::size_t sigStackSize = 32768 >= MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;
+    static constexpr std::size_t sigStackSize = 32768;
 
     static SignalDefs signalDefs[] = {
         { SIGINT,  "SIGINT - Terminal interrupt signal" },
--- a/Vendors/tropter/external/catch/catch.hpp
+++ b/Vendors/tropter/external/catch/catch.hpp
@@ -8723,7 +8723,7 @@ namespace Catch {
 
     // 32kb for the alternate stack seems to be sufficient. However, this value
     // is experimentally determined, so that's not guaranteed.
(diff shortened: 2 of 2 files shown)


aymanhab:
@AlbertoCasasOrtiz Can you give some context? Is this necessary for using a later compiler and/or platform?

AlbertoCasasOrtiz:
Sure @aymanhab.  glibc is the implementation of the C standard libraries used in Ubuntu. Starting from version 2.34, they have removed the `MINSIGSTKSZ `constant.

In Ubuntu 22.04, the version of glibc is 2.35. When you build opensim-core under this version of glibc, it  throws the following errors because this constant does not exist anymore:

```
~/opensim-core-source/OpenSim/Auxiliary/catch.hpp:10798:58: error: call no non-'constexpr' funcion 'long int sysconf(int)'
10798 | static constexpr std::size_t sigStackSize = 32768 >= MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;

~/opensim-core-source/Vendors/tropter/external/catch/catch.hpp:8726:58: error: call no non-'constexpr' funcion 'long int sysconf(int)'
8726 | static constexpr std::size_t sigStackSize = 32768 >= MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;
```

So, in order to support glibc >= 2.34, and thus, Ubuntu 22.04, we should use an alternative. This workaround allows the compilation of the code and passess the tests in ci, but I think other approaches should be explored. I found this workaround in this [link](https://github.com/catchorg/Catch2/issues/2178), and looks like many projects have been affected by this change in glibc.

This error is not being thrown by ci because it is running Ubuntu 18.04 and the version of glibc that comes with it is glibc 2.27.

aymanhab:
LGTM, thanks @AlbertoCasasOrtiz just waiting for builds to pass on ci

Issue #3226, opened 2022-06-15
MATLAB MarkersReference crash when accessing marker values


CVHammond:
I'm updating our NMSM Pipeline project to HEAD and found the following issue that crashes MATLAB.

```
fileName = 'simple_arm.trc';

import org.opensim.modeling.*
markerNames = ["r_humerus1", "r_humerus2", "r_humerus3", "r_radius1", "r_radius2", "r_radius3"];
markerWeightSet = SetMarkerWeights();
for i=1:length(markerNames)
    markerWeightSet.cloneAndAppend(MarkerWeight(markerNames(i), 1.0));
end
markersReference = MarkersReference(fileName, markerWeightSet);

markersReference.get_marker_weights(1)
```

[simple_arm.txt](https://github.com/opensim-org/opensim-core/files/8904722/simple_arm.txt)
I'm using this file for a very simple .trc file (rename .txt to .trc). The trc file works as expected in OpenSim GUI.

aymanhab:
For the record, this same API is exercised and tested in C++ here 
https://github.com/opensim-org/opensim-core/blob/2ef5732554db5671356b0d04364cbcbd4a4e5de8/OpenSim/Simulation/Test/testInverseKinematicsSolver.cpp#L171
but I'll dig into it further

aymanhab:
It is a bit hard to see why this line `get_marker_weights(1)` have worked in the past considering the signature of the method `SetMarkerWeights get_marker_weights(int i)` while ideally we don't crash, the expected API call sequence is to get the Set using `get_marker_weights()` then call `get(i)` to retrieve the ith element. I'll leave this open as we shouldn't crash but otherwise this seems to be an unexpected use of the API that has been since 4.0. If I missed something about the context please let me know @CVHammond

CVHammond:
Hmm, this is an interesting scenario. So I tried the script above in my 4.0 environment and it runs, albeit as you pointed out, perhaps as an unintended addition to the MATLAB API. The line that actually crashed the project in 4.4 beta but not in 4.0 is `ikSolver.getNumMarkersInUse()`. I thought involving the `InverseKinematicsSolver` object could obfuscate the bug. Here is a test case that most closely demonstrates the functionality that we have implemented and works in 4.0. If you have a better way to achieve the end goal of `ikSolver.computeCurrentMarkerError(i)` I'm all ears.

```
fileName = 'simple_arm.trc';
modelFileName = "simple_arm.osim";

[...]

PR #3230, opened 2022-06-20
Beefup markersreference testcase

aymanhab:
Fixes issue #3226 

### Brief summary of changes
Exercise workflow in issue 3226

### CHANGELOG.md (choose one)

- updated.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3230)


aymanhab:
@CVHammond this PR exercises the code you have in #3226, I will not merge it since it tests nothing now but it doesn't crash. If you agree, this points to memory management issues/early garbage collection in Matlab rather than an actual API bug, I will propose ideas/suggestions if that proves to be the case.

CVHammond:
@aymanhab that appears to be the correct C++ implementation of the MATLAB test case that I posted. Looking forward to advice on adjustments to deal with MATLAB's memory management

aymanhab:
Some ideas for troubleshooting:

1.  print the version (opensimCommonJNI.GetVersionAndDate()) to make sure the wiring is to the latest build of 4.4.
2. Create objects and keep references to them in variables rather than on the stack e.g. SimTKArrayCoordinateReference(), otherwise they could be immediately garbage collected.
3. Where does the crash happen, if in a loop is it at a specific index?
4. If there's a stacktrace that would be extremely helpful to pinpoint the issue source.
5. I understand this is a simplified version, which is best but I don't know if it actually crashes for you or if the crash happens only in more complicated setting where the three sets (Model-makers vs. weights vs. markers-in-trc-file) are not identical.
6. I don't have a working Matlab environment but can get that setup over the next day or two. In that case I'd like to know your Matlab version.

Let me know what you find out or if you get more clues on what's going wrong.

CVHammond:
I believe I have found the issue and fixed it. It was a combination of the `MarkersReference` object requiring a specific signature and the `MarkerReference`'s `MarkerTable` returning `double` sometimes and `java.lang.double` other times for `MarkerTable.getNearestRowIndexForTime()`. The suggestion to make sure I was using the HEAD (`opensimCommonJNI.GetVersionAndDate()`) was important as well. I'm learning more about interacting with the C++ API than I originally thought I would.

PR #3231, opened 2022-06-21
Remove final from PhysicalOffsetFrame dtor

adamkewley:
It produces many warnings on clang.

If the intention is to make the class `final`, that should be declared at the `class` level

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3231)


aymanhab (approved):
LGTM, thanks @adamkewley

PR #3232, opened 2022-06-21
Remove IKMarkerTask defaulted copy ctor

adamkewley:
Produces warnings in clang13 because it's a pattern that is formally deprecated. See:

https://stackoverflow.com/questions/51863588/warning-definition-of-implicit-copy-constructor-is-deprecated

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3232)

### DIFF
--- a/OpenSim/Tools/IKMarkerTask.h
+++ b/OpenSim/Tools/IKMarkerTask.h
@@ -40,7 +40,6 @@ OpenSim_DECLARE_CONCRETE_OBJECT(IKMarkerTask, IKTask);
 
 public:
     IKMarkerTask() = default;
-    IKMarkerTask(const IKMarkerTask&) = default;
     //=============================================================================
 };  // END of class IKMarkerTask
 //=============================================================================


adamkewley:
It *should* generate one implicitly - unless a member variable has no copy constructor (the previous version shouldn't have compiled in that case).

The other alternative is to explicitly default the copy assignment operator (effectively, add that declaration). I think either way would work but it might be that not supplying anything *also* has the side-effect of implicitly generating the move ctor/assignment (unsure about this)

adamkewley:
Merging this straggler: SWIG will certainly generate a copy constructor for the bindings, even if one is not explicitly declared (as in this PR). Otherwise, a large amount of OpenSim code will not work (many components etc. implicitly define the constructor).

aymanhab:
@adamkewley Unfortunately this commit broke the GUI build on ci because the copy constructor was not generated. I understand why you'd think this should work except for the fact that the GUI doesn't do much model building while it creates/copies/manipulates all kinds of tools/tasks and many of these are on the old property system so they may not act as expected. I can try to  modify the GUI code base in the future but for now I'd appreciate if you'd revert this change. Thank you

adamkewley:
Reverted.

Sorry for the mixup Ayman - it is kind of bonkers that SWIG wouldn't generate a copy constructor for a class that automatically has one (in terms of C++) :<.

aymanhab:
No problem at all @adamkewley  and thanks for the prompt response 👍

PR #3237, opened 2022-06-24
Fixed memory leak when CoordinateSet is populated

adamkewley:
This leak appears to happen whenever the model is loaded, re-finalized, etc.

As far as I can tell, `setMemoryOwner` should almost always be called after `setSize(0)`, because it may be the case that the collection does own its contents (e.g. upon copy construction or assignment, where the implementation always deep-copies and owns the contents).

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3237)


aymanhab (approved):
LGTM, thanks @adamkewley I'll comment on the availability of setMemoryOwner headache, in general, somewhere else.

PR #3244, opened 2022-06-29
Fixed segfault when adding a CoordinateCouplerConstraint with no func…

adamkewley:
This fixes a crashing segfault that can happen if a model contains a `CoordinateCouplerConstraint` that hasn't had a function assigned to it yet.

The reason it segfaults is because the property is tagged with `OPTIONAL` so, as far as I understand, the property is permitted to be empty. However, the patched code assumes that the property is non-empty, which results in it trying to read a non-existent memory location (at property index == 0).

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3244)


aymanhab (approved):
LGTM

PR #3247, opened 2022-06-30
Fixed OpenSim::Controller leaking memory whenever it is copied

adamkewley:
This PR tries to fix a memory leak in `OpenSim::Controller`.

The leak occurs when copy-constructing or copy-assigning a `Controller` (or any derived class). It also happens when copying any class that contains a controller (e.g. an `OpenSim::Model`). When the conditions are triggered, `Controller` will leak copies of the `OpenSim::Actuator`s it references. These can have fairly large memory footprints (e.g. because the `Actuator`s are muscles containing paths, functions, etc.). A concrete example of a file that will do this is `ToyLandingModel.osim` ([here](https://github.com/opensim-org/opensim-models/blob/b2a7e32/Models/ToyLanding/ToyLandingModel.osim)), which contains a `ToyReflexController`. Each time `ToyLandingModel.osim` is copied, around ~1-2 MB of memory will leak.

The reason it happens is because of its `_actuatorSet` member. It is an `OpenSim::Set<const Actuator>` that, internally, has `setMemoryOwner` set to `false`.  However, the original code doesn't account for the behavior of `Set` (more specifically, the `OpenSim::ArrayPtrs` it wraps), which is that--regardless of memory ownership--copy constructing (or assigning) a `Set` will unconditionally clone all of its elements, followed by re-setting `setMemoryOwner` to `true`. This causes problems because other parts of the code unconditionally call `setMemoryOwner(false)` later on, causing a leak.

This PR is a *patch* solution to the behavior. It adds user-defined copy constructor and copy assignment functions. These implementations try to implement the necessary copy semantics while avoiding the `_actuatorSet`.

A longer-term solution would be to use the sockets API to fix this. However, that *may* involve breaking the API, because the current controller API assumes that a set of `Actuators` is available via `getActuators`, `setActuators`, addActuator`, etc.

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3247)

### DIFF
--- a/OpenSim/Simulation/Control/Controller.cpp
+++ b/OpenSim/Simulation/Control/Controller.cpp
@@ -49,10 +49,7 @@ Controller::Controller() :
     _numControls{0},
     _actuatorSet{}
 {
-    setAuthors("Ajay Seth, Frank Anderson, Chand John, Samuel Hamner");
-    constructProperty_enabled(true);
-    constructProperty_actuator_list();
-    _actuatorSet.setMemoryOwner(false);
+    constructProperties();
 }
 
 Controller::Controller(Controller const& src) :
@@ -62,11 +59,16 @@ Controller::Controller(Controller const& src) :
     _numControls{src._numControls},
     _actuatorSet{}
 {
+    // care: the reason this custom copy constructor exists is to prevent
+    // a memory leak (#3247)
     _actuatorSet.setMemoryOwner(false);
 }
 
 Controller& Controller::operator=(Controller const& src)
 {
+    // care: the reason this custom copy assignment exists is to prevent
+    // a memory leak (#3247)
+
     if (&src != this)
     {
         static_cast<ModelComponent&>(*this) = static_cast<ModelComponent const&>(src);
@@ -85,6 +87,17 @@ Controller::~Controller() noexcept = default;
 //=============================================================================
 //_____________________________________________________________________________
 //_____________________________________________________________________________
+
+/**
+ * Connect properties to local pointers.
+ */
+void Controller::constructProperties()
+{
+    setAuthors("Ajay Seth, Frank Anderson, Chand John, Samuel Hamner");
+    constructProperty_enabled(true);
+    constructProperty_actuator_list();
+    _actuatorSet.setMemoryOwner(false);
+}
 
 void Controller::updateFromXMLNode(SimTK::Xml::Element& node,
                                    int versionNumber) {
--- a/OpenSim/Simulation/Control/Controller.cpp
+++ b/OpenSim/Simulation/Control/Controller.cpp
@@ -44,29 +44,47 @@ using namespace std;
 /**
  * Default constructor.
  */
-Controller::Controller() 
+Controller::Controller() :
+    ModelComponent{},
+    _numControls{0},
+    _actuatorSet{}
 {
-    constructProperties();
+    setAuthors("Ajay Seth, Frank Anderson, Chand John, Samuel Hamner");
+    constructProperty_enabled(true);
+    constructProperty_actuator_list();
+    _actuatorSet.setMemoryOwner(false);
 }
 
+Controller::Controller(Controller const& src) :
+    ModelComponent{src},
+    PropertyIndex_enabled{src.PropertyIndex_enabled},
+    PropertyIndex_actuator_list{src.PropertyIndex_actuator_list},
+    _numControls{src._numControls},
+    _actuatorSet{}
+{
+    _actuatorSet.setMemoryOwner(false);
+}
+
+Controller& Controller::operator=(Controller const& src)
+{
+    if (&src != this)
+    {
+        static_cast<ModelComponent&>(*this) = static_cast<ModelComponent const&>(src);
+        PropertyIndex_enabled = src.PropertyIndex_enabled;
+        PropertyIndex_actuator_list = src.PropertyIndex_actuator_list;
+        _numControls = src._numControls;
+        _actuatorSet.setSize(0);
+    }
+    return *this;
+}
+
+Controller::~Controller() noexcept = default;
 
 //=============================================================================
 // CONSTRUCTION
 //=============================================================================
 //_____________________________________________________________________________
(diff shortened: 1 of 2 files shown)


adamkewley:
@aseth1  if you have a second, could you review this? It can be a fairly big memory leak. Features like (e.g.) live plotting, background computing, etc. rely heavily on copying `OpenSim::Model`s. Any workflows that want dynamic background computation are going to leak memory after every edit and, eventually, crash out.

The custom implementation of a copy ctor/assignment is necessary because the class is composing something that inherently leaks memory (explained in the original post).

aymanhab (changes requested):
LGTM other than minor refactoring to follow conventions elsewhere in the code for consistency. I opened issue https://github.com/opensim-org/opensim-core/issues/3275 a few weeks ago to get rid of this memoryOwner fiasco that we've been battling for a while. Would love to hear your feedback/comment there or if you want to take it on. Thank you

> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3247#-:-NBDfCzo4rpxD1pecLRd:b42zh9f)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @adamkewley)

___
*[`OpenSim/Simulation/Control/Controller.h` line 148 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3247#-NBDdItn43uAgwPa47dq:-NBDdItn43uAgwPa47dr:b3ot0wd) ([raw file](https://github.com/opensim-org/opensim-core/blob/959fc82e6d8ff484bfd15839a6e89545142490f3/OpenSim/Simulation/Control/Controller.h#L148)):*
> ```objc
>     // construct and initialize properties
>     void constructProperties();
> ```

@adamkewley We use the convention of a separate function to constructProperties across the board, is there a strong reason to remove here and break consistency? If not I'd put it back.
___
*[`OpenSim/Simulation/Control/Controller.cpp` line 61 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3247#-NBDdwzV5TedUB2CvgmH:-NBDdwzWBLTO9tZdxlJu:b-gd0vio) ([raw file](https://github.com/opensim-org/opensim-core/blob/bd8b11b58f9043be6a0531b96f47ce5e3aeeb2a9/OpenSim/Simulation/Control/Controller.cpp#L61)):*
> ```cpp
>     ModelComponent{src},
>     PropertyIndex_enabled{src.PropertyIndex_enabled},
>     PropertyIndex_actuator_list{src.PropertyIndex_actuator_list},
> ```

 PropertyIndex_ are never used in high level code, only for book-keeping inside base Component, would be good to keep this way (in case we ever change the macros or internals of Properties/Components).

adamkewley (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3247)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @aymanhab)

___
*[`OpenSim/Simulation/Control/Controller.h` line 148 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3247#-NBDdItn43uAgwPa47dq:-NBHBdHt3hy6-wG4nLvW:b-flltzd) ([raw file](https://github.com/opensim-org/opensim-core/blob/959fc82e6d8ff484bfd15839a6e89545142490f3/OpenSim/Simulation/Control/Controller.h#L148)):*

@adamkewley We use the convention of a separate function to constructProperties across the board, is there a strong reason to remove here and break consistency? If not I'd put it back.
</blockquote></details>

I can put it back - it's just that the helper method isn't used anywhere other than one location.
___
*[`OpenSim/Simulation/Control/Controller.cpp` line 61 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3247#-NBDdwzV5TedUB2CvgmH:-NBHB1BiBpx89wPE_9k1:b-4jhsep) ([raw file](https://github.com/opensim-org/opensim-core/blob/bd8b11b58f9043be6a0531b96f47ce5e3aeeb2a9/OpenSim/Simulation/Control/Controller.cpp#L61)):*

 PropertyIndex_ are never used in high level code, only for book-keeping inside base Component, would be good to keep this way (in case we ever change the macros or internals of Properties/Components).
</blockquote></details>

I believe the reason why they are used is because the default-generated constructor would do the same (of recursively copy-constructing each member variable in the class). Is there an way to achieve 1:1 functionality with a default-implemented copy constructor? Or is the pattern to re-construct?

aymanhab (approved):
@adamkewley modulo one comment that you can easily address
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3247#-:-NRcx0I53rVxfDjtGAuV:b2owcfp)* status: all files reviewed, 1 unresolved discussion (waiting on @adamkewley)

___
*[`OpenSim/Simulation/Control/Controller.cpp` line 61 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3247#-NBDdwzV5TedUB2CvgmH:-NRcwJW41qde_XnUbs3y:bgu9uce) ([raw file](https://github.com/opensim-org/opensim-core/blob/bd8b11b58f9043be6a0531b96f47ce5e3aeeb2a9/OpenSim/Simulation/Control/Controller.cpp#L61)):*

This will be kept as-is in the PR because it's fundamentally required to custom-implement the necessary copy constructor
</blockquote></details>

@adamkewley  My recollection is that these indices are populated by construction and so don't need to be "copied" from the src object, but if that's not the case then please ignore.

adamkewley:
I opted to ignore the last comment @aymanhab - it's completely correct (initial construction of the object will also populate the indices, and those indices don't really change from instance to instance) but the default C++-generated copy constructors/assigners would also observe this behavior.

There's a chance (low, but who knows what future changes bring)  that the indices may have somehow changed in the source object, from which the property table is being copied (via the base class copy constructor/assigner) and then there'd be an index mismatch at runtime.

Ideally, we'd all get some time to remove the setMemoryOwner APU, though, followed by nuking this patch!

Issue #2869, opened 2020-08-30
opensim for python 3.8


mcmips:
How can I get opensim to install in python 3.8 for google colab? Please help.

aymanhab:
@kidzik Can you clarify what is the workflow that's supposed to work here:
1. What code base is used?
2. How is it built/compiled on what platforms?
3. How is it tested, packaged and distributed?
@carmichaelong If you're familiar of this let me know so we can maintain collectively.

nickbianco:
Based on a recent post on the Moco forum, there may be issues with installing OpenSim in Python 3.8: https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=1815&t=12387&p=0&start=0&view=&sid=0e54eee37b6d796a511ab14df68a3750.

Moco issue: https://github.com/opensim-org/opensim-moco/issues/655. 

@mcmips are you able to install on Python 3.7?

kidzik:
There are two relatively easy ways to run OpenSim in colab.
1. Setting up a jupyter notebook locally and connect colab to it
2. Installing conda and opensim conda distribution in colab
Both installation methods are described here in the context of reinforcement learning with osim-rl
https://colab.research.google.com/drive/1Kbubb9_L-MDtjt0nHVYR_zF9CK1Onz4C?usp=sharing
Simbody visualizer will be available only in the first installation method.
The conda distribution that is provided in the notebook above is only available for python 3.6.

@aymanhab answering your questions:
1. In the OpenSim conda build I was using this commit 399c8d57a779dd5dde2916192f8b92bfc959e269
2. It was compiled with these scripts https://github.com/opensim-org/conda-opensim/tree/master/opensim on MacOS / Ubuntu / Windows. Builds are available here https://anaconda.org/kidzik/opensim/files
3. It was built ad hoc for the NeurIPS challenge 2017 and slightly updated for 2018. There was no proper testing

It's probably good to completely scratch my builds and start off from @chrisdembia code https://github.com/conda-forge/staged-recipes/pull/4693

aymanhab:
@nickbianco do you know if this was specific to Anaconda or to the standard python 3.8? Trying to use Anaconda and 3.8 locally I get the errors below:
1>1: Test command: D:\anaconda3\python.exe "-m" "unittest" "discover" "--start-directory" "D:/src/opensim-core/Bindings/Python/tests" "--verbose"
1>1: Environment variables:
1>1:  PATH=D:/build/opensim-core/Release
[...]

PR #3250, opened 2022-07-10
Account for other conda variants by using conda-meta to detect conda

aymanhab:
Fixes issue #2869

### Brief summary of changes
On windows detecting conda to set dll_path relied on presence of conda folder in path, doesn't seem to always hold but conda-meta folder works per testers.

### CHANGELOG.md (choose one)

---
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3250)


tkuchida (approved):
🐍

PR #3269, opened 2022-07-25
Add setters to `DataTable.h` to make setting columns scripting friendly

nickbianco:
### Brief summary of changes
This PR introduces the methods `setDependentColumnAtIndex` and `setDependentColumn` (two signatures each) for setting dependent columns in `DataTable`s. These are more scripting-friendly, whereas the `upd`-based methods fail.

### Testing I've completed
Added tests to `testDataTable.cpp`.

These methods are needed because interacting with the `VectorView` object doesn't work in scripting. I could see the argument for fixing this issue directly, but this update is probably more straight-forward and these methods generally useful.

### CHANGELOG.md (choose one)

- updated.

### DIFF
--- a/OpenSim/Common/DataTable.h
+++ b/OpenSim/Common/DataTable.h
@@ -1023,12 +1023,67 @@ public:
         return _depData.updCol(static_cast<int>(index));
     }
 
+    /** Set dependent column at index using a SimTK::VectorView.
+    
+    \throws EmptyTable If the table is empty.
+    \throws ColumnIndexOutOfRange If index is out of range for number of columns
+                                  in the table.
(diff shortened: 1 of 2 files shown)


aymanhab:
@nickbianco Can we include a call to this method in a python and/or java testcase since this is intended to help scripting? Also to see if a more general solution could be worked out. Thank you

nickbianco:
@aymanhab I'll update the testing soon to include a scripting test case.

nickbianco:
After some quick testing, I've realized that the `upd`-based methods do work as intended in Python. Will check to confirm the same behavior in Matlab, and if so, will close the PR.

nickbianco:
`upd` methods work in Matlab as well. Closing this PR.

PR #3277, opened 2022-08-10
added MocoImpulseTrackingGoal .cpp/.h

nicos1993:
First attempt doing this, so apologies if it isn't correct!

### Brief summary of changes
Created a MocoImpulseTrackingGoal (capable of tracking a single axis' impulse at a time; user must add multiple instances of the goal to track separate axes/separate contact groups. This goal is based on the MocoContactTrackingGoal, but instead of first squaring the error term and then summing, in this new goal we sum the errors and then square.

### Testing I've completed
I have used the goal to perform tracking simulations with the 2Dwalking example - it seems to be working as anticipated.

### CHANGELOG.md (choose one)
- Uncertain as to what to include here.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3277)


nicos1993:
@nickbianco Hey, please review! :-)

OpenSim/Moco/MocoGoal/MocoImpulseTrackingGoal.cpp
```diff
@@ -0,0 +1,362 @@
...
+                    foundScaleFactor = true;
+                }
+            }
+            // If we didn't find a MocoScaleFactor for this force direction, set
+            // the reference pointer to null.
+            if (!foundScaleFactor) {
+                theseScaleFactorRefs[j] = nullptr;
+            }
+        }
+        m_scaleFactorRefs.push_back(theseScaleFactorRefs);
```
nickbianco, line 222: This code block can be updated as follows to reflect the other commented changes for the `MocoScaleFactor`: 

```
// Check to see if the model contains a MocoScaleFactor associated with
// this contact group.
RefPtrMSF thisScaleFactorRef;
bool foundScaleFactor = false;
std::pair<std::string, int> key(group.get_external_force_name(), j);
for (const auto& scaleFactor : scaleFactors) {
    if (m_scaleFactorMap[key] == scaleFactor.getName()) {
        thisScaleFactorRef= &scaleFactor;
        foundScaleFactor = true;
    }
}
// If we didn't find a MocoScaleFactor for this impulse, set
// the reference pointer to null.
if (!foundScaleFactor) {
    thisScaleFactorRef[j] = nullptr;
}
m_scaleFactorRefs.push_back(thisScaleFactorRef);
```

OpenSim/Moco/MocoGoal/MocoImpulseTrackingGoal.cpp
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    // Which axis should the impulse be tracked for?
+    if (get_impulse_axis() < 0 || get_impulse_axis() > 2)
+        OPENSIM_THROW_FRMOBJ(Exception,
+            "Expected 'ImpulseAxis' to be either 0, 1 or 2, but "
```
nickbianco, line 228: Use the same syntax as the property name, i.e., 'impulse_axis'.

OpenSim/Moco/MocoGoal/MocoImpulseTrackingGoal.cpp
```diff
@@ -0,0 +1,362 @@
...
+                std::make_pair(&contactForce, recordOffset));
+        }
+
+        // Gather the relevant data splines for this contact group.
+        // We assume that the "x", "y", and "z" columns could have been in any
+        // order.
+        const std::string& forceID = extForce.get_force_identifier();
+        groupInfo.refSplines.cloneAndAppend(allRefSplines.get(forceID + "x"));
+        groupInfo.refSplines.cloneAndAppend(allRefSplines.get(forceID + "y"));
+        groupInfo.refSplines.cloneAndAppend(allRefSplines.get(forceID + "z"));
```
nickbianco, line 177: Per the header file comment, we'll now only store one spline for the desired impulse axis. You'll need to get the correct label (i.e., forceID + "dir") based on the impulse axis, and then store the spline to `groupInfo.refSpline`.

nickbianco: Update the comment too.

OpenSim/Moco/MocoGoal/MocoImpulseTrackingGoal.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+void MocoImpulseTrackingGoal::calcIntegrandImpl(
+    const IntegrandInput& input, double& integrand) const {
+    const auto& state = input.state;
+    const auto& time = state.getTime();
+    getModel().realizeVelocity(state);
+    SimTK::Vector timeVec(1, time);
+
+    integrand = 0;
+    SimTK::Vec3 force_ref;
```
nickbianco, line 305: Since we're only storing one spline for the impulse, all the `Vec3` quantities in this method can now just be `double`s.

OpenSim/Moco/MocoGoal/MocoImpulseTrackingGoal.cpp
```diff
@@ -0,0 +1,362 @@
...
+    for (int ig = 0; ig < (int)m_groups.size(); ++ig) {
+
+        // Get contact groups.
+        const auto& group = m_groups[ig];
+
+        // Model force.
+        SimTK::Vec3 force_model(0);
```
nickbianco, line 312: This can now be `double force_model;`

OpenSim/Moco/MocoGoal/MocoImpulseTrackingGoal.cpp
```diff
@@ -0,0 +1,362 @@
...
+        for (const auto& entry : group.contacts) {
+            Array<double> recordValues = entry.first->getRecordValues(state);
+            const auto& recordOffset = entry.second;
+            for (int im = 0; im < force_model.size(); ++im) {
+                force_model[im] += recordValues[recordOffset + im];
```
nickbianco, line 317: Update to `force_model += recordValues[recordOffset + get_impulse_axis()];`. The for-loop can be removed.

OpenSim/Moco/MocoGoal/MocoImpulseTrackingGoal.cpp
```diff
@@ -0,0 +1,362 @@
...
+            }
+        }
+
+        // Reference force.
+        for (int ir = 0; ir < force_ref.size(); ++ir) {
+            force_ref[ir] = group.refSplines[ir].calcValue(timeVec);
+        }
```
nickbianco, line 324: --> `force_ref = group.refSpline.calcValue(timeVec);`
Remove the for-loop.

OpenSim/Moco/MocoGoal/MocoImpulseTrackingGoal.cpp
```diff
@@ -0,0 +1,362 @@
...
+            group.refExpressedInFrame->expressVectorInGround(state, force_ref);
+        }
+
+        // Apply scale factors for this marker, if they exist.
+        const auto& scaleFactorRef = m_scaleFactorRefs[ig];
+        for (int j = 0; j < 3; ++j) {
+            if (scaleFactorRef[j] != nullptr) {
+                force_ref[j] *= scaleFactorRef[j]->getScaleFactor();
+            }
+        }
```
nickbianco, line 337:
```
if (scaleFactorRef != nullptr) {
    force_ref *= scaleFactorRef->getScaleFactor();
}
```
Remove the for-loop.

OpenSim/Moco/MocoGoal/MocoImpulseTrackingGoal.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+        SimTK::Vec3 error3D = force_model - force_ref;
+
+        double error = error3D[get_impulse_axis()];
+
+        integrand += error;
```
nickbianco, line 343: These three lines can now become the following:

```
double error = force_model - force_ref;
integrand += error;
```

OpenSim/Moco/MocoGoal/MocoImpulseTrackingGoal.cpp
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    // Square the integrated impulse error
+    integrand = (integrand * integrand);
```
nickbianco, line 347: We can't square the integrand here because that would violate the math for force impulse. See the comment in `calcGoalImpl` for where to square.

nickbianco (changes requested):
Great start! I know there's a zillion comments, but there's lots of little stuff that you wouldn't know to check. 

There's other stuff we will need to change (i.e., we can only track impulse for one foot at a time), but let's start with this batch of comments for now.

OpenSim/Moco/MocoGoal/MocoImpulseTrackingGoal.h
```diff
@@ -0,0 +1,283 @@
...
+    In other words, the scale factor is applied when computing the tracking
+    error for each contact force group, not to the reference data directly.
+    You must specify both the external force name associated with the contact
+    force group and the index corresponding to the direction (i.e., X = 0,
+    Y = 1, Z = 2) of the scaled force value. The direction is applied in
+    whatever frame the reference data is expressed in based on the provided
+    ExternalLoads in each contact group.
+    Adding a scale factor to a MocoImpulseTrackingGoal.
+    @code
+    auto* markerTrackingGoal = problem.addGoal<MocoContactTrackingGoal>();
```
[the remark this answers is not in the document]
nicos1993: I can do this no problem - just out of curiosity, these few lines of are just comments and otherwise they have no significance?

OpenSim/Moco/MocoGoal/MocoImpulseTrackingGoal.h
```diff
@@ -0,0 +1,283 @@
...
+        /// In other words, the scale factor is applied when computing the tracking
+        /// error for each contact force group, not to the reference data directly.
+        /// You must specify both the external force name associated with the contact
+        /// force group and the index corresponding to the direction (i.e., X = 0,
+        /// Y = 1, Z = 2) of the scaled force value. The direction is applied in
+        /// whatever frame the reference data is expressed in based on the provided
+        /// ExternalLoads in each contact group.
+        void addScaleFactor(const std::string& name,
+            const std::string& externalForceName, int index,
+            const MocoBounds& bounds);
```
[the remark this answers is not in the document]
nicos1993: I have replaced `int index` with `int impulse_axis` ... but I am now wondering what impact this will have on 'mutable std::map<std::pair<std::string, int>, std::string> m_scaleFactorMap;
        using RefPtrMSF = SimTK::ReferencePtr<const MocoScaleFactor>;
        mutable std::vector<std::array<RefPtrMSF, 3>> m_scaleFactorRefs;'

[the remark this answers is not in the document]
nicos1993: ignore - saw your comment below!

OpenSim/Moco/MocoGoal/MocoImpulseTrackingGoal.h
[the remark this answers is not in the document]
nickbianco: Correct. Everything between /** and */ becomes a comment in C++.

OpenSim/Moco/MocoGoal/MocoImpulseTrackingGoal.cpp
nicos1993: Is this because calcGoalImpl gets called for each mesh point? again...just out of curiosity!

nicos1993:
I think my changes should now have been pushed back remotely with message "applied changes from review 1"

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.cpp
```diff
@@ -88,21 +88,16 @@ void MocoImpulseTrackingGoal::addScaleFactor(const std::string& name,
...
 
-    // Check that the index is in the correct range.
-    OPENSIM_THROW_IF_FRMOBJ((index < 0) || (index > 2), Exception,
-        "Expected marker scale factor index to be in the range [0, 2], "
-        "but received '{}'.", index);
-
     // Update the scale factor map so we can retrieve the correct MocoScaleFactor
     // for this contact group during initialization.
-    std::pair<std::string, int> key(externalForceName, index);
+    std::pair<std::string, int> key(externalForceName, impulse_axis);
```
nickbianco, line 93: Per the comment above, you don't need the `impulse_axis` as part of key now.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.cpp
```diff
@@ -203,29 +198,27 @@ void MocoImpulseTrackingGoal::initializeOnModelImpl(const Model& model) const {
...
+                foundScaleFactor = true;
             }
         }
-        m_scaleFactorRefs.push_back(theseScaleFactorRefs);
+        // If we didn't find a MocoScaleFactor for this force direction, set
+        // the reference pointer to null.
+        if (!foundScaleFactor) {
+            thisScaleFactorRef = nullptr;
+        }
+        m_scaleFactorRefs.push_back(thisScaleFactorRef);
```
nickbianco, line 215: For now, until we add support for multiple integrals, we should check that `m_scaleFactorRefs` has length == 1.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.cpp
```diff
@@ -88,21 +88,16 @@ void MocoImpulseTrackingGoal::addScaleFactor(const std::string& name,
...
-    std::pair<std::string, int> key(externalForceName, index);
+    std::pair<std::string, int> key(externalForceName, impulse_axis);
     m_scaleFactorMap[key] = name;
 
     // Append the scale factor to the MocoGoal.
     appendScaleFactor(MocoScaleFactor(name, bounds));
 }
 
-void MocoImpulseTrackingGoal::initializeOnModelImpl(const Model& model) const {
+void MocoContactImpulseTrackingGoal::initializeOnModelImpl(const Model& model) const {
```
nickbianco, line 100: Until we support multiple integrals/impulses, we should add a check that the `contact_groups` property has length == 1.  You can add a comment to the documentation in the header file noting that the goal currently only supports tracking one impulse for one foot at a time.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.h
```diff
@@ -73,47 +74,36 @@ namespace OpenSim {
...
 
     /**
-    \section MocoImpulseTrackingGoal
+    \section MocoContactImpulseTrackingGoal
     Minimize the error between compliant contact force impulse from the model and
-    experimentally measured contact impulse.
-    This class handles multiple groups of contact forces and a single
-    experimental external loads file. Tracking ground reaction forces for the
-    left and right feet in gait requires only one instance of this goal.
+    experimentally measured contact impulse - for a single axis.
```
nickbianco, line 79: You can remove this hyphen.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.h
```diff
@@ -73,47 +74,36 @@ namespace OpenSim {
...
+    This class handles the contact impulses from a single contact group and a 
+    single experimental external loads file. Tracking ground reaction impulses for
+    the left and right feet in gait requires separate instances of this goal.
+    Tracking ground reaction impulses for multiple axes requires separate 
+    instances of this goal.
```
nickbianco, line 84: Ah, I didn't see this before I made my comment above, please feel free to ignore it.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.h
```diff
@@ -225,7 +216,7 @@ namespace OpenSim {
         /// whatever frame the reference data is expressed in based on the provided
         /// ExternalLoads in each contact group.
         void addScaleFactor(const std::string& name,
-            const std::string& externalForceName, int index,
+            const std::string& externalForceName, int impulse_axis,
```
nickbianco, line 219: See comments in cpp file.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.h
```diff
@@ -234,29 +225,29 @@ namespace OpenSim {
...
+        OpenSim_DECLARE_LIST_PROPERTY(contact_groups, MocoContactImpulseTrackingGoalGroup,
             "Associate contact elements in the model with force data.");
         OpenSim_DECLARE_OPTIONAL_PROPERTY(external_loads, ExternalLoads,
             "Experimental contact force data.");
         OpenSim_DECLARE_PROPERTY(external_loads_file, std::string,
             "Experimental contact force data as an ExternalLoads XML file.");
         OpenSim_DECLARE_PROPERTY(impulse_axis, int,
-            "User specifies which axis to track the GRF impulse: "
-            "'1': X, '2': Y, '3': Z"
+            "User specifies which axis to track the ground reaction force impulse: "
```
nickbianco, line 241: Again, this description should use language to describe what it is, rather than how a user will interact it with. For example: "The axis of the ground reaction force impulse component to be tracked in the goal."

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.h
```diff
@@ -19,18 +19,19 @@
  * -------------------------------------------------------------------------- */
 
 #include <OpenSim/Moco/osimMoco.h>
-#include "osimMocoImpulseTrackingGoalDLL.h"
+#include "osimMocoContactImpulseTrackingGoalDLL.h"
```
nickbianco, line 22: This won't be necessary anymore because we are going to add this goal directly to the OpenSim API.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.h
```diff
@@ -45,25 +46,25 @@ namespace OpenSim {
     specifying these alternative frames, Moco does not know which force to use
     (the force on the sphere or the force on the half-space) when summing the
     contact forces across contact force elements.
-    @see MocoImpulseTrackingGoal */
-    class OSIMMOCOIMPULSETRACKINGGOAL_API MocoImpulseTrackingGoalGroup : public Object {
-        OpenSim_DECLARE_CONCRETE_OBJECT(MocoImpulseTrackingGoalGroup, Object);
+    @see MocoContactImpulseTrackingGoal */
+    class OSIMMOCOCONTACTIMPULSETRACKINGGOAL_API MocoContactImpulseTrackingGoalGroup : public Object {
```
nickbianco, line 50: Since we'll be adding this directly to the OpenSim API, we'll use a different macro here. Replace `OSIMMOCOCONTACTIMPULSETRACKINGGOAL_API` with `OSIMMOCO_API`.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.h
```diff
@@ -140,37 +130,37 @@ namespace OpenSim {
...
+    auto* contactImpulseTrackingGoal = problem.addGoal<MocoContactImpulseTrackingGoal>();
     ...
-    markerTrackingGoal->addScaleFactor(
+    contactImpulseTrackingGoal->addScaleFactor(
             'RightGRF_vertical_scale_factor', 'Right_GRF', 1, {0.5, 2.0});
     @endcode
     @ingroup mocogoal */
-    class OSIMMOCOIMPULSETRACKINGGOAL_API MocoImpulseTrackingGoal : public MocoGoal {
-        OpenSim_DECLARE_CONCRETE_OBJECT(MocoImpulseTrackingGoal, MocoGoal);
+    class OSIMMOCOCONTACTIMPULSETRACKINGGOAL_API MocoContactImpulseTrackingGoal : public MocoGoal {
```
nickbianco, line 155: Since we'll be adding this directly to the OpenSim API, we'll use a different macro here. Replace `OSIMMOCOCONTACTIMPULSETRACKINGGOAL_API` with `OSIMMOCO_API`.

nickbianco (changes requested):
Nice job on the updates! Looking a lot cleaner now. After addressing these comments, there's a handful of things we need to do to convert this from a plug-in to a new MocoGoal in the API. I'll describe that in the next comment.

OpenSim/Moco/RegisterTypes_osimMoco.cpp
```diff
@@ -27,10 +27,11 @@
 #include "MocoFrameDistanceConstraint.h"
 #include "MocoGoal/MocoAccelerationTrackingGoal.h"
 #include "MocoGoal/MocoAngularVelocityTrackingGoal.h"
-#include "MocoGoal/MocoContactTrackingGoal.h"
+#include "MocoGoal/MocoContactImpulseTrackingGoal.h"
```
nickbianco, line 30: You accidentally renamed the `MocoContactTrackingGoal.h` include and left `MocoImpulseTrackingGoal.h` include unchanged below.

nickbianco, line 30: You accidentally renamed the MocoContactTrackingGoal.h include and left MocoImpulseTrackingGoal.h include unchanged below.

OpenSim/Moco/RegisterTypes_osimMoco.cpp
```diff
@@ -27,10 +27,11 @@
...
 #include "MocoGoal/MocoControlGoal.h"
 #include "MocoGoal/MocoControlTrackingGoal.h"
 #include "MocoGoal/MocoGoal.h"
+#include "MocoGoal/MocoImpulseTrackingGoal.h
```
nickbianco, line 34: See comment above.

nickbianco (changes requested):
One quick thing I missed on the last pass.

nickbianco:
After addressing the comments above, there's a few more things we need to do to add this new goal to OpenSim:

- In [`RegisterTypes_osimMoco.cpp`](https://github.com/opensim-org/opensim-core/blob/156763aa66d7db71275f6531a8f0168211991c74/OpenSim/Moco/RegisterTypes_osimMoco.cpp#L63), you need to add the `Object::registerType()` calls for both `MocoContactImpulseTrackingGoal` and `MocoContactImpulseTrackingGoalGroup`. (Can you also add a call for `MocoContactTrackingGoalGroup` under `MocoContactTrackingGoal` here for me? This will fix the error you were seeing in Matlab for anyone loading a `MocoContactTrackingGoal` from XML.)
- Add an [include statement in osimMoco.h ](https://github.com/opensim-org/opensim-core/blob/156763aa66d7db71275f6531a8f0168211991c74/OpenSim/Moco/osimMoco.h#L21): `#include "MocoGoal/MocoContactImpulseTrackingGoal.h"`
- In [this CMakeLists.txt file](https://github.com/opensim-org/opensim-core/blob/156763aa66d7db71275f6531a8f0168211991c74/OpenSim/Moco/CMakeLists.txt#L4), add the .h/.cpp file for the new goal like the other goals listed.
- Add an include statement for the new goal [in this file](https://github.com/opensim-org/opensim-core/blob/156763aa66d7db71275f6531a8f0168211991c74/Bindings/OpenSimHeaders_moco.h#L6) to add support for the goal in the bindings (Matlab and Python). 
- Also for the bindings, add an include statement for the new goal [in this file ](https://github.com/opensim-org/opensim-core/blob/156763aa66d7db71275f6531a8f0168211991c74/Bindings/moco.i#L27). Note the slightly different format for the include statements here.

OpenSim/Moco/MocoGoal/MocoImpulseTrackingGoal.cpp
nickbianco: `calcGoalImpl` is computed based on the entire trajectory at each iteration. `calcIntegrandImpl` is computed at each mesh point. In `calcGoalImpl`, you have access to the trajectory's initial state, final state, and the integral of whatever you put in `calcIntegrandImpl`. 

Impulse is the integral of force, not force squared, so squaring here would produce some other quantity that isn't impulse.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.cpp
nicos1993: If I omit `impulse_axis` from here and the declaration/definition of `addScaleFactor`, `key` becomes a problem as it is expecting an integer (due to `pair`) - could this just be initialised with 0? 

Similarly for line 203 - `std::pair<std::string, int> key(group.get_external_force_name(), get_impulse_axis());` to `std::pair<std::string, int> key(group.get_external_force_name(), 0);`

I don't think it should affect anything further down stream...

nickbianco: Oh, you can just change `m_scaleFactorMap` ([here](https://github.com/opensim-org/opensim-core/blob/b52ed03f30a8132f449687d3b4db3cc82428a3e5/OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.h#L267)) to the following type:
`mutable std::map<std::string, std::string> m_scaleFactorMap;`

nicos1993:
Hey Nick, 

Firstly, thanks for the input! It has been super helpful!!! I think I have made the changes you requested!

OpenSim/Moco/MocoGoal/MocoContactTrackingGoal.h
```diff
@@ -1,326 +0,0 @@
-#ifndef OPENSIM_MOCOCONTACTTRACKINGGOAL_H
```
nickbianco, line 1: Looks like you accidentally deleted this file.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.cpp
```diff
@@ -86,24 +85,19 @@ void MocoContactTrackingGoal::addScaleFactor(const std::string& name,
...
-                "Expected marker scale factor index to be in the range [0, 2], "
-                "but received '{}'.", index);
+        "Contact group associated with external force '{}' not found.",
+        externalForceName);
 
     // Update the scale factor map so we can retrieve the correct MocoScaleFactor
     // for this contact group during initialization.
-    std::pair<std::string, int> key(externalForceName, index);
+    std::string key(externalForceName);
     m_scaleFactorMap[key] = name;
```
nickbianco, line 94: I think you can just do this here: `m_scaleFactorMap[externalForceName] = name;`.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.cpp
```diff
@@ -117,23 +111,25 @@ void MocoContactTrackingGoal::initializeOnModelImpl(const Model& model) const {
...
         extLoads = extLoadsFromFile.get();
-    } else {
+    }
+    else {
         OPENSIM_THROW_FRMOBJ(Exception, "No ExternalLoads provided.");
     }
 
     // Spline the data.
     const std::string dataFilePath =
-            convertRelativeFilePathToAbsoluteFromXMLDocument(
```
nickbianco, line 134: Looks like some auto-formatting happened, which is fine, but make sure that no whitespace tabs have been entered into the code. If so, replace them with spaces. (Visual Studio has an option to do this for you).

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.cpp
```diff
@@ -147,24 +143,24 @@ void MocoContactTrackingGoal::initializeOnModelImpl(const Model& model) const {
...
+            "External force '{}' not found.",
+            group.get_external_force_name());
         const auto& extForce =
-                extLoads->get(group.get_external_force_name());
+            extLoads->get(group.get_external_force_name());
 
         GroupInfo groupInfo;
         for (int ic = 0; ic < group.getProperty_contact_force_paths().size();
-                ++ic) {
+            ++ic) {
```
nickbianco, line 154: This line should be indented once.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.cpp
```diff
@@ -179,76 +175,84 @@ void MocoContactTrackingGoal::initializeOnModelImpl(const Model& model) const {
...
                 groupInfo.refExpressedInFrame =
-                        &model.getComponent<PhysicalFrame>(
-                                forceExpressedInBody);
-            } else if (model.hasComponent<PhysicalFrame>(
-                               "./bodyset/" + forceExpressedInBody)) {
+                    &model.getComponent<PhysicalFrame>(
+                        forceExpressedInBody);
+            }
+            else if (model.hasComponent<PhysicalFrame>(
+                "./bodyset/" + forceExpressedInBody)) {
```
nickbianco, line 185: This line should be indented once. (As a rule of thumb, if the if-statement expression needs multiple lines, you should indent the extra lines so they are not left-aligned with the code within the if-statement.)

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.cpp
```diff
@@ -179,76 +175,84 @@ void MocoContactTrackingGoal::initializeOnModelImpl(const Model& model) const {
...
-            }
-            // If we didn't find a MocoScaleFactor for this force direction, set
-            // the reference pointer to null.
-            if (!foundScaleFactor) {
-                theseScaleFactorRefs[j] = nullptr;
+        RefPtrMSF thisScaleFactorRef;
+        bool foundScaleFactor = false;
+        std::string key(group.get_external_force_name());
+        for (const auto& scaleFactor : scaleFactors) {
+            if (m_scaleFactorMap[key] == scaleFactor.getName()) {
```
nickbianco, line 205: Similar to above, you could just do `m_scaleFactorMap[group.get_external_force_name()]`. In this case it might be cleaner to leave it the way it is, but it does create an extra variable. Up to you.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.cpp
```diff
@@ -16,65 +16,64 @@
...
+MocoContactImpulseTrackingGoalGroup::MocoContactImpulseTrackingGoalGroup() {
     constructProperties();
 }
 
-MocoContactTrackingGoalGroup::MocoContactTrackingGoalGroup(
-        const std::vector<std::string>& contactForcePaths,
-        const std::string& externalForceName) {
+MocoContactImpulseTrackingGoalGroup::MocoContactImpulseTrackingGoalGroup(
+    const std::vector<std::string>& contactForcePaths,
+    const std::string& externalForceName) {
```
nickbianco, line 31: The two lines above should be indented.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.cpp
```diff
@@ -16,65 +16,64 @@
...
-MocoContactTrackingGoalGroup::MocoContactTrackingGoalGroup(
-        const std::vector<std::string>& contactForcePaths,
-        const std::string& externalForceName,
-        const std::vector<std::string>& altFramePaths) :
-        MocoContactTrackingGoalGroup(contactForcePaths, externalForceName) {
+MocoContactImpulseTrackingGoalGroup::MocoContactImpulseTrackingGoalGroup(
+    const std::vector<std::string>& contactForcePaths,
+    const std::string& externalForceName,
+    const std::vector<std::string>& altFramePaths) :
+    MocoContactImpulseTrackingGoalGroup(contactForcePaths, externalForceName) {
```
nickbianco, line 43: These four lines should be indented.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.cpp
```diff
@@ -179,76 +175,84 @@ void MocoContactTrackingGoal::initializeOnModelImpl(const Model& model) const {
...
 }
 
-int MocoContactTrackingGoal::findRecordOffset(
-        const MocoContactTrackingGoalGroup& group,
-        const SmoothSphereHalfSpaceForce& contactForce,
-        const std::string& appliedToBody) const {
+int MocoContactImpulseTrackingGoal::findRecordOffset(
+    const MocoContactImpulseTrackingGoalGroup& group,
+    const SmoothSphereHalfSpaceForce& contactForce,
+    const std::string& appliedToBody) const {
```
nickbianco, line 249: Indent these three lines.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.cpp
```diff
@@ -179,76 +175,84 @@ void MocoContactTrackingGoal::initializeOnModelImpl(const Model& model) const {
...
 
     // Is the ExternalForce applied to the sphere's body?
     const auto& sphereBase =
-            contactForce.getConnectee<ContactSphere>("sphere")
-                    .getConnectee<PhysicalFrame>("frame")
-                    .findBaseFrame();
+        contactForce.getConnectee<ContactSphere>("sphere")
+        .getConnectee<PhysicalFrame>("frame")
+        .findBaseFrame();
```
nickbianco, line 255: Similar to the if-statement expression, if you are chaining function calls like this, you should indent subsequent lines.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.cpp
```diff
@@ -259,9 +263,9 @@ int MocoContactTrackingGoal::findRecordOffset(
 
     // Is the ExternalForce applied to the half space's body?
     const auto& halfSpaceBase =
-            contactForce.getConnectee<ContactHalfSpace>("half_space")
-                    .getConnectee<PhysicalFrame>("frame")
-                    .findBaseFrame();
+        contactForce.getConnectee<ContactHalfSpace>("half_space")
+        .getConnectee<PhysicalFrame>("frame")
+        .findBaseFrame();
```
nickbianco, line 268: Indent here too (see above).

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.cpp
```diff
@@ -277,23 +281,23 @@ int MocoContactTrackingGoal::findRecordOffset(
     const auto& sphereBasePath = sphereBase.getAbsolutePathString();
     const auto& halfSpaceBasePath = halfSpaceBase.getAbsolutePathString();
     for (int ia = 0; ia < group.getProperty_alternative_frame_paths().size();
-            ++ia) {
+        ++ia) {
```
nickbianco, line 284: Indent.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.cpp
```diff
@@ -307,13 +311,11 @@ void MocoContactTrackingGoal::calcIntegrandImpl(
...
         // Model force.
-        SimTK::Vec3 force_model(0);
+        double force_model;
         for (const auto& entry : group.contacts) {
             Array<double> recordValues = entry.first->getRecordValues(state);
             const auto& recordOffset = entry.second;
-            for (int im = 0; im < force_model.size(); ++im) {
-                force_model[im] += recordValues[recordOffset + im];
-            }
+                force_model += recordValues[recordOffset + get_impulse_axis()];
```
nickbianco, line 318: This shouldn't be indented, because we don't have the for-loop anymore.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.cpp
```diff
@@ -328,43 +330,26 @@ void MocoContactTrackingGoal::calcIntegrandImpl(
 
         // Apply scale factors for this marker, if they exist.
         const auto& scaleFactorRef = m_scaleFactorRefs[ig];
-        for (int j = 0; j < 3; ++j) {
-            if (scaleFactorRef[j] != nullptr) {
-                force_ref[j] *= scaleFactorRef[j]->getScaleFactor();
-            }
+        if (scaleFactorRef != nullptr) {
+            force_ref *= scaleFactorRef->getScaleFactor();
```
nickbianco, line 334: `force_ref` is type `SimTK::Vec3`, and here you are multiplying by a `double`. This should work as expected, i.e., all elements of `force_ref` are multiplied by the scalar. But it's also good to check the docs to make sure these types of operations are doing what you expect. If you wanted to be extra safe, you could re-write this section to something like this:

```c++
// Apply scale factors for this marker, if they exist.
const auto& scaleFactorRef = m_scaleFactorRefs[ig];
double scaleFactor = 1.0;
if (scaleFactorRef != nullptr) {
    scaleFactor = scaleFactorRef->getScaleFactor();
}

double error = force_model - scaleFactor * force_ref[get_impulse_axis()];
integrand += error;
```

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.cpp
```diff
@@ -328,43 +330,26 @@ void MocoContactTrackingGoal::calcIntegrandImpl(
...
+    log_cout("        impulse axis: {}", get_impulse_axis());
     for (int ig = 0; ig < getProperty_contact_groups().size(); ++ig) {
         const auto& group = get_contact_groups(ig);
         log_cout("        group {}: ExternalForce: {}",
-                ig, group.get_external_force_name());
+            ig, group.get_external_force_name());
         log_cout("            forces:");
         for (int ic = 0; ic < group.getProperty_contact_force_paths().size();
-                ++ic) {
+            ++ic) {
```
nickbianco, line 351: Indent.

OpenSim/Moco/RegisterTypes_osimMoco.cpp
```diff
@@ -70,6 +71,9 @@ OSIMMOCO_API void RegisterTypes_osimMoco() {
         Object::registerType(MocoMarkerTrackingGoal());
         Object::registerType(MocoMarkerFinalGoal());
         Object::registerType(MocoContactTrackingGoal());
+		Object::registerType(MocoContactTrackingGoalGroup());
+		Object::registerType(MocoContactImpulseTrackingGoal());
+		Object::registerType(MocoContactImpulseTrackingGoalGroup());
```
nickbianco, line 76: These three lines shouldn't be indented.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.h
```diff
@@ -0,0 +1,273 @@
+#ifndef OPENSIM_MOCOCONTACTIMPULSETRACKINGGOAL_H
```
nickbianco, line 1: I'm not sure what happened -- I don't have any diffs for this file, and `MocoContactTrackingGoal.h` was deleted somehow. I'll do my best to review this file.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.h
```diff
@@ -0,0 +1,273 @@
...
+    to the tracking error calculations based on the following equation:
+         error = modelValue - scaleFactor * referenceValue
+    In other words, the scale factor is applied when computing the contact
+    impulse tracking error for each contact force group, not to the reference 
+    data directly. You must specify both the external force name associated 
+    with the contact force group and the index corresponding to the direction
+    (i.e., X = 0, Y = 1, Z = 2) of the scaled force value. The direction is 
+    applied in whatever frame the reference data is expressed in based on the 
+    provided ExternalLoads in each contact group.
+    Adding a scale factor to a MocoImpulseTrackingGoal.
```
nickbianco, line 146: --> `MocoContactImpulseTrackingGoal`

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.h
```diff
@@ -0,0 +1,273 @@
...
+    \frac{1}{mg} (\int_{t_i}^{t_f}
+                 \vec{F}_{m,j,a} - \vec{F}_{e,j,a} ~dt)^2; {j \in G}; {a \in A}
+    \f]
+    We use the following notation:
+    - \f$ t_i \f$: the initial time of this phase.
+    - \f$ t_f \f$: the final time of this phase.
+    - \f$ mg \f$: the total weight of the system; replaced with
+        \f$ m \f$ if \f$ g = 0 \f$.
+    - \f$ G \f$: the set of contact groups.
+    - \f$ A \f$: the set of axes {0, 1, 2}.
```
nickbianco, line 97: I think the equation can be simplified. There's only one contact group and only one impulse axis at a time, so iterating through `G` and `A` is unnecessary.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.h
```diff
@@ -0,0 +1,273 @@
...
+        /// MocoContactImpulseTrackingGoalGroup for more information.
+        void addContactGroup(MocoContactImpulseTrackingGoalGroup group) {
+            append_contact_groups(std::move(group));
+        }
+
+        /// Set the ground reaction force contact impulse 'axis' to be tracked
+        /// (X = 0, Y = 1, Z = 2), where 'axis' refers to the dimension/direction of
+        /// the ground reaction force contact impulse, in the ground frame, to be
+        /// tracked.
+        void setImpulseAxis(int ImpulseAxis) {
```
nickbianco, line 199: Small nitpick, but for camelcase variable names, we usually don't capitalize the first letter, i.e., `impulseAxis`.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.h
```diff
@@ -0,0 +1,273 @@
...
+        /// distributed across multiple bodies use this function instead of the
+        /// easier-to-use addContactGroup(), and set the group's
+        /// alternative_frame_paths property accordingly. See
+        /// MocoContactImpulseTrackingGoalGroup for more information.
+        void addContactGroup(MocoContactImpulseTrackingGoalGroup group) {
+            append_contact_groups(std::move(group));
+        }
+
+        /// Set the ground reaction force contact impulse 'axis' to be tracked
+        /// (X = 0, Y = 1, Z = 2), where 'axis' refers to the dimension/direction of
```
nickbianco, line 196: I think you could replace "dimension/direction" with "component".

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.h
```diff
@@ -0,0 +1,273 @@
...
+        /// If the contact force elements associated with a single ExternalForce are
+        /// distributed across multiple bodies use this function instead of the
+        /// easier-to-use addContactGroup(), and set the group's
+        /// alternative_frame_paths property accordingly. See
+        /// MocoContactImpulseTrackingGoalGroup for more information.
+        void addContactGroup(MocoContactImpulseTrackingGoalGroup group) {
+            append_contact_groups(std::move(group));
+        }
+
+        /// Set the ground reaction force contact impulse 'axis' to be tracked
```
nickbianco, line 195: You could remove the quotes from 'axis'.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.h
```diff
@@ -0,0 +1,273 @@
...
+        // PROPERTIES
+        OpenSim_DECLARE_LIST_PROPERTY(contact_groups, MocoContactImpulseTrackingGoalGroup,
+            "Associate contact elements in the model with force data.");
+        OpenSim_DECLARE_OPTIONAL_PROPERTY(external_loads, ExternalLoads,
+            "Experimental contact force data.");
+        OpenSim_DECLARE_PROPERTY(external_loads_file, std::string,
+            "Experimental contact force data as an ExternalLoads XML file.");
+        OpenSim_DECLARE_PROPERTY(impulse_axis, int,
+            "The axis of the ground reaction force impulse component to be tracked in the goal."
+            "'X'= 0, 'Y'= 1 , 'Z'= 2"
```
nickbianco, line 241: I would remove the quotes around X, Y, and Z here to make it easier to read.

OpenSim/Moco/MocoGoal/MocoContactImpulseTrackingGoal.h
```diff
@@ -0,0 +1,273 @@
...
+            "(Default value is: -1)");
```
nickbianco, line 242: You could be even more succinct here: `(default = -1)` or `(default: -1)`.

nickbianco:
Great job @nicos1993! I'm happy with all of the code changes.

The last thing we need to do is some sort of testing to make sure everything works. For this goal, I think it's sufficient to perform some local tests. When CI builds pass, you can download a Windows artifact from GitHub actions that you install locally. Then you can run a test problem to make sure that the goal works though Matlab and/or Python.

Once that's done, we'll be good to merge.

nicos1993:
Hi, 

I can confirm I have locally tested on Windows. Modified the 2D walking example in Matlab to track vertical contact impulse from right and left feet - successfully converged.

nickbianco:
Great job @nicos1993!

nicos1993:
> Great job @nicos1993!

Thanks for the help Nick!

Issue #3267, opened 2022-07-22
[Moco] Disabling `OPENSIM_WITH_TROPTER` skips the Ipopt install


nickbianco:
It is currently not possible to use Moco after building with `OPENSIM_WITH_CASADI` enabled and `OPENSIM_WITH_TROPTER` disabled because the Ipopt install is tied in with the tropter install scripts (look [here](https://github.com/opensim-org/opensim-core/blob/524cb877dcea6d67ce7e68ad994792623c80663a/Vendors/CMakeLists.txt#L2), [here](https://github.com/opensim-org/opensim-core/blob/524cb877dcea6d67ce7e68ad994792623c80663a/Vendors/tropter/CMakeLists.txt#L90), and [here](https://github.com/opensim-org/opensim-core/blob/524cb877dcea6d67ce7e68ad994792623c80663a/Vendors/tropter/cmake/FindIPOPT.cmake)).

nickbianco:
Closing via #3278.

PR #3278, opened 2022-08-12
Fix IPOPT library install with only OPENSIM_WITH_CASADI flag

carmichaelong:
Fixes issue #3267 

### Brief summary of changes
Moved the logic for IPOPT libraries out of the Vendors/tropter folder and into the base CMakeLists file

### Testing I've completed
With `OPENSIM_WITH_CASADI = ON` and `OPENSIM_WITH_TROPER = OFF`, ran tests and a few Moco examples through MATLAB successfully

Local testing on another platform would be helpful since there are different branches for Windows and not-Windows,

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3278)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ v4.4.1
 ======
 
 - Made `Component::getSocketNames` a `const` member method (previously: non-const)
+- Fixed an issue with IPOPT libraries when building OpenSim with `OPENSIM_WITH_CASADI = ON` but `OPENSIM_WITH_TROPTER = OFF` (Issue #3267).
 
 v4.4
 ====
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -862,20 +862,20 @@ if(UNIX)
 endif()
 
 if (OPENSIM_WITH_CASADI OR OPENSIM_WITH_TROPTER)
-	if(UNIX)
-		pkg_check_modules(IPOPT REQUIRED ipopt IMPORTED_TARGET)
-	else()
-		find_package(Ipopt REQUIRED CONFIG)
-		set_package_properties(IPOPT PROPERTIES
-				URL https://projects.coin-or.org/Ipopt
-				TYPE REQUIRED
-				PURPOSE "Nonlinear optimization")
-		# Flags to use when building a shared library.
-		# TODO move these flags to a specific target.
-		# TODO set(CMAKE_SHARED_LINKER_FLAGS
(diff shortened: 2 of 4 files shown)


carmichaelong:
@nickbianco would you be able to test on your mac?

nickbianco:
The build is still looking for ADOL-C when I turn off `OPENSIM_WITH_TROPTER`. I think [this line](https://github.com/opensim-org/opensim-core/blob/c76a63959c5c232262feeb2076a30aa55174f9e1/cmake/OpenSimMocoInstallMacDependencyLibraries.cmake.in#L60) needs to be surrounded by an `if (OPENSIM_WITH_TROPTER) endif()`.

nickbianco:
Ignore the comment above. As per discussion with @carmichaelong, my issue was related to not building all the dependencies for both CasADi and Tropter. The current scope of the PR only covers the CasADi/Tropter CMake flags for the core build, not the dependencies build.

nickbianco (approved):
:lgtm:

> Reviewed 1 of 4 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3278#-:-N9J8Xgi8Hce3aXwoVRo:b45h)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @carmichaelong)

aymanhab:
LGTM as long as we don't need to pull/build ipopt if neither OPENSIM_WITH_CASADI nor OPENSIM_WITH_TROPTER are on. If that scenario has been tested feel free to merge. Thanks @carmichaelong

carmichaelong:
@aymanhab @nickbianco Let's hold off on the merge for now, and I can dive a little more into the combinations of what's checked between dependencies and the main opensim build. If it's easy enough, it'd be good to put that in this PR, otherwise, I think we could still merge since the testing so far shows that it fixes a current problem without introducing other issues (just discovering other related issues).

aymanhab:
Thanks again @carmichaelong I don't want to make up new work for you but if our build of dependencies with both flags OFF doesn't require IPOPT to be superbuilt or downloaded then we're good to go.

carmichaelong:
@aymanhab answer to your questions:
>  if our build of dependencies with both flags OFF doesn't require IPOPT to be superbuilt or downloaded then we're good to go.

This should already be the case before these changes.

Going to merge this based on feedback at the meeting (thanks @nickbianco and @aymanhab). I've also opened up #3281 and #3282 for other issues discovered while working on this one

PR #3284, opened 2022-08-29
Test wrapping algorithm

aymanhab:
Related issue #3333

### Brief summary of changes
Separate test case that compares length of wrapping around single objects to analytical answers, and comparing pairs of wrap-object types against one another.

### CHANGELOG.md (choose one)

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284)


aymanhab:
@carmichaelong for your review, feedback. Turn on visualization to see what we're testing for now

carmichaelong (changes requested):
Thanks @aymanhab. I left a bunch of notes throughout, mostly about commenting some sections a little better.

Here are a few high level comments about the tests. We should discuss the scope of this PR and if other tests that get added should be within this PR or following PRs:
1. Looked like the tests were always along one of the axes of the objects. More tests should be added for cases that are off-axis.
2. Path lengths were tested analytically only at 0 and 90 degrees. Should add tests for values in between as well.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-:-NBTFTtD9SOn-_iOcqo-:bp85pd5)* status: 0 of 2 files reviewed, 9 unresolved discussions (waiting on @aymanhab and @carmichaelong)

___
*[`.github/workflows/continuous_integration.yml` line 430 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBJnGUVDoAURlRzp553:-NBJnGUVDoAURlRzp554:b-tsswmk) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/.github/workflows/continuous_integration.yml#L430)):*
> ```YAML
> 
>     - name: Build dependencies
>       # if: steps.cache-dependencies.outputs.cache-hit != 'true'
> ```

should this be uncommented before final merge?
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 91 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTDykf3QE2ror3ZCyc:-NBTDykf3QE2ror3ZCyd:b-y02s3l) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L91)):*
> ```cpp
>         auto* woTwo = new WrapEllipsoid();
>         woTwo->setName("pulley2");
>         // -30 - 30 degrees guarantee no edge which is poorly handled and need to be dropped
> ```

A little more context here would be helpful for maintainability. e.g. I'm not immediately sure what this edge is and why it's a problem such that this test should be limited to these bounds
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 114 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTCZKMBtu5LcZP4_7A:-NBTCZKMBtu5LcZP4_7B:bvga9bt) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L114)):*
> ```cpp
> // Test results of wrapping a sigle path perpendicular to a wrapObject 
> // and compare results to analytical/expected answers
> void testSingleWrapObjectPerpendicular(WrapObject* wrapObject)
> ```

Describing this as "perpendicular" may be a little ambiguous. Is this always along the Z-axis (or some other axis)?
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 117 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTD3H6EGT42xJduu5V:-NBTD3H6EGT42xJduu5W:bf11sk3) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L117)):*
> ```cpp
> {
>     auto visualize = false;
>     const double r = 0.5;
> ```

Is this 0.5 because the objects in the main() function are also 0.5? If so, would be good to just get the radius size of the object and then use that throughout here, or at least document clearly that this is only made for a radius of this size.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 152 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTCmOREB5RoW32bM_-:-NBTCmOREB5RoW32bM_0:b-goh3n5) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L152)):*
> ```cpp
>     auto& coord = joint->updCoordinate();
>     const CoordinateSet& cset = model.getCoordinateSet();
>     // get angle
> ```

what angle is this?
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 156 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTDSKA-LupzbVoR_vk:-NBTDSKA-LupzbVoR_vl:b-57plnd) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L156)):*
> ```cpp
>     double a = 0.5;
>     double b = .5 / cos(ang);
>     // perimeter approx fomrula by Ramanujan
> ```

A little more context here is helpful. What is this perimeter formula for exactly? Circular cross-section? ellipsoidal cross section?
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 157 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTGfOE3GJQuSMmIvk5:-NBTGfOE3GJQuSMmIvk6:b81vdxl) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L157)):*
> ```cpp
>     double b = .5 / cos(ang);
>     // perimeter approx fomrula by Ramanujan
>     double p = SimTK::Pi * (3*(a + b) - sqrt((3 * a + b) * (a + 3 * b)));
> ```

Looks like the `p` isn't used in this test?
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 251 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTFNKACupl8dflnRr8:-NBTFNKACupl8dflnRr9:bd8qi2) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L251)):*
> ```cpp
>     
>     int nsteps = 1000;
>     double max_diff = 0.;
> ```

Was `max_diff` for debugging only or did you have another check in mind for this test?
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 264 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTEM_X9vUw-SXvIdsT:-NBTEM_X9vUw-SXvIdsU:bb4o7ax) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L264)):*
> ```cpp
>         max_diff = std::max(max_diff, std::abs(len1 - len2));
>         //std::cout << "i=" << i << " length=" << len1 << " " << len2 << " diff="  << std::abs(len1-len2) << std::endl;
>         ASSERT_EQUAL<double>(len1, len2, .01);
> ```

1cm differences seems large for this test. What were the max differences you were seeing?

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284)* status: 0 of 2 files reviewed, 9 unresolved discussions (waiting on @aymanhab and @carmichaelong)

___
*[`.github/workflows/continuous_integration.yml` line 430 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBJnGUVDoAURlRzp553:-NBTciYX6oZ8O4fmKz0j:bb74ce2) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/.github/workflows/continuous_integration.yml#L430)):*

should this be uncommented before final merge?
</blockquote></details>

Sure
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 91 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTDykf3QE2ror3ZCyc:-NBTcnBN8B_wAQT42_F7:bp6etym) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L91)):*

A little more context here would be helpful for maintainability. e.g. I'm not immediately sure what this edge is and why it's a problem such that this test should be limited to these bounds
</blockquote></details>

Wrapping that includes the cap of the cylinder is not meaningful and leads to C0 continuity which is not physical. Would not be easy to match with functions and likely not supported by new wrapping formulation. Initially I wanted angles -45-45 but the algorithm had random glitches at these angles so I had to trim the test instead
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 114 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTCZKMBtu5LcZP4_7A:-NBTdqhh7o_oEWDV3Qm5:bdv4e21) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L114)):*

Describing this as "perpendicular" may be a little ambiguous. Is this always along the Z-axis (or some other axis)?
</blockquote></details>

Generally I use cylinder as the base case, cross section is a circle or ellipse based on angle, should clarify that in comments.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 117 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTD3H6EGT42xJduu5V:-NBTeEe1-5Wb78WPk1-I:b3wv7e4) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L117)):*

Is this 0.5 because the objects in the main() function are also 0.5? If so, would be good to just get the radius size of the object and then use that throughout here, or at least document clearly that this is only made for a radius of this size.
</blockquote></details>

Yes there's debugging code that can be cleaned up, I agree with extracting the radius from the wrapobject instead. though some casting maneuvers maybe needed to make the call.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 152 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTCmOREB5RoW32bM_-:-NBTekou185mbomeNz8I:b5cw4jc) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L152)):*

what angle is this?
</blockquote></details>

That was the point of leaving behind the option to visualize, but this is the angle between the cylinder axis and the line connecting the end points of the spring. Will document
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 156 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTDSKA-LupzbVoR_vk:-NBTfZnW8d4fC5MUm91f:b2mrayl) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L156)):*

A little more context here is helpful. What is this perimeter formula for exactly? Circular cross-section? ellipsoidal cross section?
</blockquote></details>

The exact formula for length of ellipse is complicated, I tried to use an approximation but that wasn't close enough. Will remove references
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 251 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTFNKACupl8dflnRr8:-NBTfkeREC2aVHriEy1a:bfd57qh) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L251)):*

Was `max_diff` for debugging only or did you have another check in mind for this test?
</blockquote></details>

Debugging
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 264 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTEM_X9vUw-SXvIdsT:-NBTfmaf6RdhOYZ5M7uh:b-j4lf56) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L264)):*

1cm differences seems large for this test. What were the max differences you were seeing?
</blockquote></details>

I started with Eps and ended up here to make the test to pass for the combination of scenarios in the test case.

aymanhab (commented):
Done

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-:-NEbHcHN4XC79yQUKDke:bb6un1j)* status: 0 of 2 files reviewed, 9 unresolved discussions (waiting on @carmichaelong)

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284)* status: 0 of 2 files reviewed, 9 unresolved discussions (waiting on @carmichaelong)

___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 157 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTGfOE3GJQuSMmIvk5:-NEbefa2ELzpkkbnkTlQ:ba3dfob) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L157)):*

Looks like the `p` isn't used in this test?
</blockquote></details>

Removed

aymanhab:
@carmichaelong for your review

carmichaelong (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @aymanhab)

___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 264 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTEM_X9vUw-SXvIdsT:-NEmNp9XBa8nQHcrpH1R:bl1h0jr) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L264)):*

I started with Eps and ended up here to make the test to pass for the combination of scenarios in the test case.
</blockquote></details>

Given the code comments on this test, I *think* I see why you wouldn't expect this to be close to zero (if the cylinder or ellipsoid are being rotated such that the spring doesn't wrap perpendicularly to the long axis). If that's the case, this might not be right check to compare to each other. Could be helpful to chat after the next dev meeting to make sure I'm understanding correctly.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 139 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NEm41smCOS7G1l-hPwP:-NEm41smCOS7G1l-hPwQ:byqr2ho) ([raw file](https://github.com/opensim-org/opensim-core/blob/46ca1910f8b875c210c76d8a5bb280a88ed268ba/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L139)):*
> ```cpp
>     return 0;
> }
> // Test results of wrapping a sigle path perpendicular to a wrapObject
> ```

sigle -> single
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 198 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NEm4s8I9ihGdOAFAABv:-NEm4s8I9ihGdOAFAABw:byx4erl) ([raw file](https://github.com/opensim-org/opensim-core/blob/46ca1910f8b875c210c76d8a5bb280a88ed268ba/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L198)):*
> ```cpp
>         double len1 = spring1->getLength(s);
>         // Length is 0.9 by construction plus a portion of a quarter circle with radius r proportional to i
>         ASSERT_EQUAL<double>(len1, .9 + 0.25 * 2 * SimTK::Pi * r * i / nsteps, 1e-6); //this formula is based on radius = .5
> ```

is the comment "this formula is based on radius = .5" needed still? seems like the equation is general enough now using `r` for radius
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 235 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NEmMzRG6twueVgnyb2-:-NEmMzRG6twueVgnyb20:b-t82sh4) ([raw file](https://github.com/opensim-org/opensim-core/blob/46ca1910f8b875c210c76d8a5bb280a88ed268ba/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L235)):*
> ```cpp
> 
>     model.addComponent(spring1);
>     // Ceate offset frame in z direction
> ```

A couple things for this line:
1. adding an empty line above to separate lines related to spring1 and spring2 would be helpful for readability
2. Ceate -> Create
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 237 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NEmNdXy1oPAtPvVI69O:-NEmNdXy1oPAtPvVI69P:bh9wtzy) ([raw file](https://github.com/opensim-org/opensim-core/blob/46ca1910f8b875c210c76d8a5bb280a88ed268ba/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L237)):*
> ```cpp
>     // Ceate offset frame in z direction
>     PhysicalOffsetFrame* offsetZ = new PhysicalOffsetFrame(
>     "z_plus1", ground, SimTK::Transform(Vec3(0, 0, 1)));
> ```

Tab at the start of the line would help readability

aymanhab:
All changes requested have been implemented in addition a test for ellipse perimeter length was added. Ready for review. Maybe good for @nickbianco to take a look and give feedback as well considering current plans. Thank you. I changed signature of a method that was unnecessarily non-const but I can separate into another PR.

nickbianco (changes requested):
> Reviewed 1 of 2 files at r5, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284)* status: all files reviewed, 16 unresolved discussions (waiting on @aymanhab and @carmichaelong)

___
*[`OpenSim/Simulation/Model/PathSpring.h` line 146 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMArvUC93pCFCUVfE5p:-NMArvUC93pCFCUVfE5q:b-q18mot) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Simulation/Model/PathSpring.h#L146)):*
> ```objc
>     //--------------------------------------------------------------------------
>     /** compute the moment-arm of the PathSpring about a coordinate of interest. */
>     double computeMomentArm(const SimTK::State& s, const Coordinate& aCoord) const;
> ```

I'm assuming there was no reason for this to be non-const in the first place?
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 264 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTEM_X9vUw-SXvIdsT:-NMAreP71QTLYkb7Lztx:b-gdl64x) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L264)):*

Given the code comments on this test, I *think* I see why you wouldn't expect this to be close to zero (if the cylinder or ellipsoid are being rotated such that the spring doesn't wrap perpendicularly to the long axis). If that's the case, this might not be right check to compare to each other. Could be helpful to chat after the next dev meeting to make sure I'm understanding correctly.
</blockquote></details>

I agree with @carmichaelong.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 202 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMApmEh2rn8DaFfptAw:-NMApmEh2rn8DaFfptAx:byqr2ho) ([raw file](https://github.com/opensim-org/opensim-core/blob/52e77d4ad00321bc3ff40112d11fbc89d91a1ca1/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L202)):*
> ```cpp
>     }
> }
> // Test results of wrapping a sigle path around a wrapCylinder wObj1
> ```

sigle -> single
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 206 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMAqPhE8yGz8PeduDnb:-NMAqPhE8yGz8PeduDnc:b-vjs0dn) ([raw file](https://github.com/opensim-org/opensim-core/blob/52e77d4ad00321bc3ff40112d11fbc89d91a1ca1/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L206)):*
> ```cpp
> // For example wrapping around a rotated cylinder against an ellipsoid with radii 
> // picked to match the radii of the elliptical cross-section
> void testCompareWrapObjects(OpenSim::WrapCylinder* wObj1, OpenSim::WrapObject* wObj2) {
> ```

Much of this function has similar code to `testSingleWrapObjectPerpendicular` and could be abstracted away, specifically the model creation steps. Might be useful if we want to build on this PR for future tests.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 185 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMAsuDbAgpAblhuCJuS:-NMAsuDbAgpAblhuCJuT:b35rn7) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L185)):*
> ```cpp
>         new PathSpring("spring1", 1.0, 0.1, 0.01);
>     spring1->updGeometryPath().
>         appendNewPathPoint("origin", ground, Vec3(r-.1, r, 0)); //offset in X direction to avoid ambiguous scenario where path passes through center
> ```

This comment could go on top of the line.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 243 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMAsxHo1eBXC5PTU3gl:-NMAsxHo1eBXC5PTU3gm:b35rn7) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L243)):*
> ```cpp
>         new PathSpring("spring1", 1.0, 0.1, 0.01);
>     spring1->updGeometryPath().
>         appendNewPathPoint("origin", ground, Vec3(r - .1, r, 0)); //offset in X direction to avoid ambiguous scenario where path passes through center
> ```

This comment could go on top of the line.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 262 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMAsy3xBRGhCn-gSNt2:-NMAsy3xBRGhCn-gSNt3:b35rn7) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L262)):*
> ```cpp
>         new PathSpring("spring2", 1.0, 0.1, 0.01);
>     spring2->updGeometryPath().
>         appendNewPathPoint("origin", ground, Vec3(r - .1, r, 1)); //offset in X direction to avoid ambiguous scenario where path passes through center
> ```

This comment could go on top of the line.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 290 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMAsWEt33Ecz3feFo-O:-NMAsWEu6QOWw77EIrBO:b-pdqyh8) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L290)):*
> ```cpp
>     }
> }
> // Ellipsoid passed in has radii of a,b, c wrapping occurs along z axis
> ```

-> `a, b, c`
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 291 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMAtLTv3j1FNQigEVvA:-NMAtLTv3j1FNQigEVvB:b-ooe4lm) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L291)):*
> ```cpp
> }
> // Ellipsoid passed in has radii of a,b, c wrapping occurs along z axis
> // no closed-form analytical solution to compare but approximate length
> ```

Do you mean there's no closed-form solution for the moment arm?
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 294 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMAsTXABzgIfhYrM6xF:-NMAsTXABzgIfhYrM6xG:beo94gh) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L294)):*
> ```cpp
> // for full ellipse. Wrapping should match exactly 1/4 
> // perimeter of ellipse + fixed offset baked in.
> void testEllipsoidWrapLength(OpenSim::WrapEllipsoid* wrapObject)
> ```

See comment above about abstraction.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 317 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMAt3hp-YbwRYXwgQ9w:-NMAt3hp-YbwRYXwgQ9x:b35rn7) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L317)):*
> ```cpp
>         new PathSpring("spring1", 1.0, 0.1, 0.01);
>     spring1->updGeometryPath().
>         appendNewPathPoint("origin", ground, Vec3(r - .1, r, 0)); //offset in X direction to avoid ambiguous scenario where path passes through center
> ```

This comment could go on top of the line.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 340 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMAtXjqBHwnhsCHQN5I:-NMAtXjqBHwnhsCHQN5J:b7985xd) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L340)):*
> ```cpp
>     double b = wrapObject->get_dimensions()[1];
>     double h = ((a - b) * (a - b)) / ((a + b) * (a+b));
>     double lengthAnalyticalApprox = SimTK::Pi * (a + b) * (1 + 3 * h / (10 + std::sqrt(4 - 3 * h)));
> ```

Why is this "approximately" analytical?

nickbianco (commented):
@aymanhab just took a first pass over everything. The moment arm signature change is fine. Would could discuss some of the wrapping details at dev meeting.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-:-NMAtiN88vHiOEh58Q0-:b-8sjbli)* status: all files reviewed, 16 unresolved discussions (waiting on @aymanhab and @carmichaelong)

nickbianco (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284)* status: all files reviewed, 16 unresolved discussions (waiting on @aymanhab and @carmichaelong)

___
*[`OpenSim/Simulation/Model/PathSpring.h` line 146 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMArvUC93pCFCUVfE5p:-NMBCZ_TAVreAMs52XTh:biu29lw) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Simulation/Model/PathSpring.h#L146)):*

I'm assuming there was no reason for this to be non-const in the first place?
</blockquote></details>

I'm getting "Symbol not found" linking issues on Mac when this is const. When I remove the const qualifier and use non-const coordinate references in the wrapping test (i.e., updCoordinate()), everything works fine.

nickbianco (commented):
In `testCompareWrapObjects`, the wrap objects are intersecting, which could be causing wrapping issues for certain configurations of the cylinder.



> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-:-NMBCbdG0LiglDKDmPb-:bcldfdt)* status: all files reviewed, 16 unresolved discussions (waiting on @aymanhab and @carmichaelong)

aymanhab:
> > _[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284)_ status: all files reviewed, 16 unresolved discussions (waiting on @aymanhab and @carmichaelong)
> 
> _[`OpenSim/Simulation/Model/PathSpring.h` line 146 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMArvUC93pCFCUVfE5p:-NMBCZ_TAVreAMs52XTh:biu29lw) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Simulation/Model/PathSpring.h#L146)):_
> 
> Previously, nickbianco (Nicholas Bianco) wrote…
> I'm getting "Symbol not found" linking issues on Mac when this is const. When I remove the const qualifier and use non-const coordinate references in the wrapping test (i.e., updCoordinate()), everything works fine.

Likely this is because the bindings have not been rebuilt/regenerated from scratch. In previous iterations I used updCoordinate but it is wrong (I believe) to blow up the whole system underneath as a side-effect of upd call when the Coordinate itself is not being changed. Will look at the bindings in the artifacts to confirm.

aymanhab:
> In `testCompareWrapObjects`, the wrap objects are intersecting, which could be causing wrapping issues for certain configurations of the cylinder.
> 
> 
> 
> > _[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-:-NMBCbdG0LiglDKDmPb-:bcldfdt)_ status: all files reviewed, 16 unresolved discussions (waiting on @aymanhab and @carmichaelong)

Paths explicitly specify what wrap objects they "wrap" around, so even though visually these intersect, computationally each path knows about exactly one wrap-object and as such are impossible to interact with both.

nickbianco:
@aymanhab actually the issue was that I was setting my DYLD path manually to a different OpenSim build, which messed up my dev environment. Retracting that comment now.

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284)* status: all files reviewed, 15 unresolved discussions (waiting on @aymanhab and @carmichaelong)

___
*[`OpenSim/Simulation/Model/PathSpring.h` line 146 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMArvUC93pCFCUVfE5p:-NMBJNTg2PPIV-Nqp66c:bnfdlax) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Simulation/Model/PathSpring.h#L146)):*

Likely this is because the bindings have not been rebuilt/regenerated from scratch. In previous iterations I used updCoordinate but it is wrong (I believe) to blow up the whole system underneath as a side-effect of upd call when the Coordinate itself is not being changed. Will look at the bindings in the artifacts to confirm.
</blockquote></details>

This is a non-issue, retracting my comment.

nickbianco (commented):
> Paths explicitly specify what wrap objects they "wrap" around, so even though visually these intersect, computationally each path knows about exactly one wrap-object and as such are impossible to interact with both.

Ah, that's right. It's good that I'm getting a refresher on wrapping!

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-:-NMBJb4EBrUXnRoxvhJw:bs0jg06)* status: all files reviewed, 15 unresolved discussions (waiting on @aymanhab and @carmichaelong)

nickbianco (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284)* status: all files reviewed, 16 unresolved discussions (waiting on @aymanhab and @carmichaelong)

___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 124 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMBZl_z-iCiqccPGBOG:-NMBZl_z-iCiqccPGBOH:belgt4m) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L124)):*
> ```cpp
>             woTwo->set_dimensions(Vec3(radius/cos(angle), radius, 1)); // Change radii of ellipsoid to match cross-section of cylinder cut
>             // std::cout << "compare cylinder vs ellipsoid at angle " << angle * 180/SimTK::Pi << std::endl;
>             testCompareWrapObjects(woOne, woTwo);
> ```

The error between the cylinder and ellipsoid wrap lengths grows larger as the cylinder rotation angle deviates from zero. I'm not sure if the comparison here is equivalent.

nickbianco (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284)* status: all files reviewed, 16 unresolved discussions (waiting on @aymanhab and @carmichaelong)

___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 124 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMBZl_z-iCiqccPGBOG:-NMBdzB9C5UnpPcgiCN3:b5gosxj) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L124)):*

The error between the cylinder and ellipsoid wrap lengths grows larger as the cylinder rotation angle deviates from zero. I'm not sure if the comparison here is equivalent.
</blockquote></details>

After looking at the visualization of the test more closely, it looks like the cylinder path is sliding along the z-direction a little, which might mean that the path is not perfectly approximating the perimeter of an ellipsoid. This could be causing unexpected differences in path lengths for the larger angles.

aymanhab:
> > _[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284)_ status: all files reviewed, 16 unresolved discussions (waiting on @aymanhab and @carmichaelong)
> 
> _[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 124 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMBZl_z-iCiqccPGBOG:-NMBZl_z-iCiqccPGBOH:belgt4m) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L124)):_
> 
> > ```c++
> >             woTwo->set_dimensions(Vec3(radius/cos(angle), radius, 1)); // Change radii of ellipsoid to match cross-section of cylinder cut
> >             // std::cout << "compare cylinder vs ellipsoid at angle " << angle * 180/SimTK::Pi << std::endl;
> >             testCompareWrapObjects(woOne, woTwo);
> > ```
> 
> The error between the cylinder and ellipsoid wrap lengths grows larger as the cylinder rotation angle deviates from zero. I'm not sure if the comparison here is equivalent.

It is true that the two curves start the same and diverge at large angles, it is hard to tell if this is because of implementation artifacts (length for cylinder is computed analytically in the code while ellipsoid is added up numerically) or other fundamental reasons. If you have other ideas I'm all open.

nickbianco (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284)* status: all files reviewed, 16 unresolved discussions (waiting on @aymanhab and @carmichaelong)

___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 124 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMBZl_z-iCiqccPGBOG:-NMFJc1_-YgL_SwU0s79:b-yzknqs) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L124)):*

It is true that the two curves start the same and diverge at large angles, it is hard to tell if this is because of implementation artifacts (length for cylinder is computed analytically in the code while ellipsoid is added up numerically) or other fundamental reasons. If you have other ideas I'm all open.  
</blockquote></details>

I'm not sure. We're quickly veering into "it's just geometry" territory; I'd need a refresher on conics math to be sure the comparison makes sense. Do you have a good reference (like the ellipsoid test) for this test?

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284)* status: all files reviewed, 16 unresolved discussions (waiting on @carmichaelong and @nickbianco)

___
*[`OpenSim/Simulation/Model/PathSpring.h` line 146 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMArvUC93pCFCUVfE5p:-NMAzOQZ13cfNAUUAQ7H:bon6ixg) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Simulation/Model/PathSpring.h#L146)):*

This is a non-issue, retracting my comment.
</blockquote></details>

Indeed, and the body of the function only contains a call to a method that requires const reference anyways 
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 139 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NEm41smCOS7G1l-hPwP:-NM5DHnfARzrCNLEk8yR:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/46ca1910f8b875c210c76d8a5bb280a88ed268ba/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L139)):*

sigle -> single
</blockquote></details>

Done.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 198 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NEm4s8I9ihGdOAFAABv:-NM5DnvJBz4820vF-sJM:brqjn9e) ([raw file](https://github.com/opensim-org/opensim-core/blob/46ca1910f8b875c210c76d8a5bb280a88ed268ba/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L198)):*

is the comment "this formula is based on radius = .5" needed still? seems like the equation is general enough now using `r` for radius
</blockquote></details>

Updated and removed comment.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 235 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NEmMzRG6twueVgnyb2-:-NM5FF003gI1rcn8HrBc:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/46ca1910f8b875c210c76d8a5bb280a88ed268ba/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L235)):*

A couple things for this line:
1. adding an empty line above to separate lines related to spring1 and spring2 would be helpful for readability
2. Ceate -> Create
</blockquote></details>

Done.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 237 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NEmNdXy1oPAtPvVI69O:-NM5Fm3n3UqtKlT6f5ZP:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/46ca1910f8b875c210c76d8a5bb280a88ed268ba/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L237)):*

Tab at the start of the line would help readability
</blockquote></details>

Done.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 202 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMApmEh2rn8DaFfptAw:-NMFGk1N1_PFytUr5WP3:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/52e77d4ad00321bc3ff40112d11fbc89d91a1ca1/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L202)):*

sigle -> single
</blockquote></details>

Done.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 185 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMAsuDbAgpAblhuCJuS:-NMFGdpS7NgNgWs6g56t:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L185)):*

This comment could go on top of the line.
</blockquote></details>

Done.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 243 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMAsxHo1eBXC5PTU3gl:-NMFIKT437r-g3WTPJ4T:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L243)):*

This comment could go on top of the line.
</blockquote></details>

Done.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 262 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMAsy3xBRGhCn-gSNt2:-NMFIVjE5mdPFzqMyliO:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L262)):*

This comment could go on top of the line.
</blockquote></details>

Done.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 290 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMAsWEt33Ecz3feFo-O:-NMFIiOE4LFrm0UZ_MAr:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L290)):*

-> `a, b, c`
</blockquote></details>

Done.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 291 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMAtLTv3j1FNQigEVvA:-NMFIkZ1EzNdEmMn0Nvr:bvv6wgq) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L291)):*

Do you mean there's no closed-form solution for the moment arm?
</blockquote></details>

I mean the length, there's only approximations as described here: https://www.cuemath.com/measurement/perimeter-of-ellipse/
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 340 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMAtXjqBHwnhsCHQN5I:-NMFKfs16qZ2zM75F4lF:b-xjbr6z) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L340)):*

Why is this "approximately" analytical?
</blockquote></details>

Please check the url included in the source and linked above

nickbianco (changes requested):
> Reviewed 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284)* status: all files reviewed, 8 unresolved discussions (waiting on @aymanhab and @carmichaelong)

___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 124 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMBZl_z-iCiqccPGBOG:-NMFTHUtEE7MI6PV29_k:b-bde1dm) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L124)):*

You make a good point, and let me know if you'd prefer we remove this part of the test altogether, the intuition here is that if the long axis of the ellipsoid goes to infinity the problems are identical and there should be no difference whatsoever. I have no reference other than intuition and familiarity with math/geometry.
</blockquote></details>

I think I would be happy with the two tests comparing to analytical (or approximately analytical) solutions. We might go down a rabbit hole trying to get the cylinder and ellipsoid to be consistent, without much benefit.

aymanhab:
The benefit is that other test cases all test perpendicular wrapping and the other analytical case only spot checks when wrapping curve is a quarter of an ellipse, but I see your point and will remove that section of the test. Thanks for the feedback.

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284)* status: 3 of 4 files reviewed, 8 unresolved discussions (waiting on @carmichaelong and @nickbianco)

___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 264 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NBTEM_X9vUw-SXvIdsT:-NMFepTf9y-xX1TlEs8X:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/1c5ba701a87e3fa6e1fcd6f5b7ddd66366cdc198/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L264)):*

I agree with @carmichaelong.
</blockquote></details>

Done.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 206 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMAqPhE8yGz8PeduDnb:-NMFLNmb72AOoD4RblHC:b6sw4b4) ([raw file](https://github.com/opensim-org/opensim-core/blob/52e77d4ad00321bc3ff40112d11fbc89d91a1ca1/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L206)):*

Much of this function has similar code to `testSingleWrapObjectPerpendicular` and could be abstracted away, specifically the model creation steps. Might be useful if we want to build on this PR for future tests.
</blockquote></details>

The three instances of model creation are close but not identical and same with the path creation, it feels that trying to squeeze the common 4 lines into a function is a bit of an overkill but I could be swayed if you believe it's worth it.
___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 124 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMBZl_z-iCiqccPGBOG:-NMFemIACsY8mA_hRnpf:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/97482af156bdfc0a6df767943740ecb7f7325ff4/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L124)):*

I think I would be happy with the two tests comparing to analytical (or approximately analytical) solutions. We might go down a rabbit hole trying to get the cylinder and ellipsoid to be consistent, without much benefit.
</blockquote></details>

Done.

nickbianco (approved):
All looks good to me, but let's wait to merge until after our meeting this afternoon.

> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-:-NMFitMD541J6obRq5G5:b-qmbh5v)* status: all files reviewed, 5 unresolved discussions (waiting on @carmichaelong)

___
*[`OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp` line 206 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3284#-NMAqPhE8yGz8PeduDnb:-NMFidIf0MzgwJNGRUNU:b-q5r8vy) ([raw file](https://github.com/opensim-org/opensim-core/blob/52e77d4ad00321bc3ff40112d11fbc89d91a1ca1/OpenSim/Tests/Wrapping/testWrappingAlgorithm.cpp#L206)):*

The three instances of model creation are close but not identical and same with the path creation, it feels that trying to squeeze the common 4 lines into a function is a bit of an overkill but I could be swayed if you believe it's worth it.
</blockquote></details>

No, the tests are fine as-is. We can refactor later if it makes sense.

Issue #3289, opened 2022-09-08
Replace macos-10.15 with  macos-11 or latest 


aymanhab:
macos-10.15 will be retired in 12/22 on github actions

carmichaelong:
@aymanhab also looks like MATLAB 2022b is discontinuing support for 10.15: https://www.mathworks.com/support/requirements/matlab-mac.html

aymanhab:
PR in progress to add to ci PR #3295

aymanhab:
Build succeeds with osx-11 but fails to load/run without library path setting, PR in the pipe
https://github.com/opensim-org/opensim-gui/pull/1389

jenhicks:
Thanks for catching that!

aymanhab:
Done a while ago, closing

PR #3295, opened 2022-09-16
build on osx-11

aymanhab:
Fixes issue #3289
### Brief summary of changes
Add block for macos-11 in ci xml file

### CHANGELOG.md (choose one)

- updated.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -403,6 +403,130 @@ jobs:
     #     asset_name: opensim-core-${{ steps.configure.outputs.version }}.zip
     #     asset_content_type: application/zip
 
+  mac:
+    name: Mac11
+
+    runs-on: macos-11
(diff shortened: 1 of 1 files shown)


aymanhab:
@carmichaelong I added a build script option for mac osx 11 on ci since 10.15 will be retired soon, but  I'm having build problem with Clang 13. What version did you use on your M1? I may upgrade my old macbook to BigSur/11 to test but will have no way back to test 10.15. Thanks

carmichaelong:
I'm on OSX 12.4 with Clang 13.1.6.13160021. So I'm a little behind on minor OSX versions right now.

aymanhab:
While the build on osx-11 succeeds, there are problems with RPATH of gcc libraries that show at runtime post install as described by @carmichaelong here https://github.com/opensim-org/opensim-gui/pull/1387

PR #3301, opened 2022-10-17
Branch reduce propertyaccess wrap

aymanhab:
Fixes issue #0

### Brief summary of changes
Improve performance by improving:
- memory allocation: preallocate arrays to avoid copying to grow incrementally
- cache properties into local variables
- leverage the fact that calculation is done in wrapObject's own coordinate system.
- 
### CHANGELOG.md (choose one)

- updated.


aymanhab:
We should close, the old wrapping code will not be changed any more while new algorithms are being integrated.

Issue #3320, opened 2022-11-17
On macOS 11+, `libcoinmumps` may incorrectly depend on `libgcc`


carmichaelong:
@AlbertoCasasOrtiz 

Related PR: https://github.com/opensim-org/opensim-gui/pull/1387

With artifact `opensim-core-4.4.1-2022-08-25-41f28cb9a` and before:
```
dyld_info -dependents libcoinmumps.1.dylib            
libcoinmumps.1.dylib [x86_64]:
    -dependents:
        attributes     load path
                       @rpath/libcoinmetis.1.dylib
                       @rpath/libgfortran.5.dylib
                       @rpath/libquadmath.0.dylib
                       /usr/lib/libSystem.B.dylib
                       /System/Library/Frameworks/Accelerate.framework/Versions/A/Accelerate
```

With artifact `opensim-core-4.4.1-2022-09-08-de6854033` and after:
```
dyld_info -dependents libcoinmumps.1.dylib            
libcoinmumps.1.dylib [x86_64]:
    -dependents:
        attributes     load path
                       @rpath/libcoinmetis.1.dylib
                       @rpath/libgfortran.5.dylib
                       @rpath/libquadmath.0.dylib
                       /usr/lib/libSystem.B.dylib
                       /System/Library/Frameworks/Accelerate.framework/Versions/A/Accelerate
                       /usr/local/opt/gcc/lib/gcc/current/libgcc_s.1.1.dylib
```

Note that the last entry `/usr/local/opt/gcc/lib/gcc/current/libgcc_s.1.1.dylib` now appears as a dependency for `libcoinmumps.1.dylib` while it didn't in the previous artifact.

This shows up when loading the GUI as the following error message:
<img width="2979" alt="Screenshot 2022-11-17 at 10 28 52 AM" src="https://user-images.githubusercontent.com/7455790/202584505-25c0ba21-3830-4689-be99-8acee7c2f471.png">

Some possibly helpful info copied over from the GUI related PR:

I've seen libgcc_s.1.* before in various places in CMake. A couple of places are located here:
- https://github.com/opensim-org/opensim-core/blob/524cb877dcea6d67ce7e68ad994792623c80663a/Vendors/tropter/cmake/CMakeLists.txt
- https://github.com/opensim-org/opensim-core/blob/a01cf545485e8bc1f6e50b29ddeab229677de2c5/Vendors/tropter/cmake/tropter_install_mac_dependency_libraries.cmake.in

[...]

PR #3334, opened 2022-12-08
Fix mac libgcc problem

AlbertoCasasOrtiz:
Fixes issue #3320 and #3295.

### Brief summary of changes

The script was extracting the directory containing `libgcc_s.1.dylib` from `libgfortran.dylib`, however, it was returning `@rpath/libgcc_s.1.1.dylib` instead of the original path `/usr/local/opt/gcc/lib/gcc/current/libgcc_s.1.1.dylib`. Thus, when trying to update it in the rest of libraries, it was not finding `@rpath/libgcc_s.1.1.dylib`, and not updating it.

Now, I am extracting the directory containing `libgcc_s.1.dylib` from `libcoinmumps.dylib` instead, since we are sure that library has the expected value.

Also, I have changed the name of the library from `libgcc_s.1` to `libgcc_s.1.1`, since it is asking for `libgcc_s.1.1` (but it is still compatible with `libgcc_s.1`). The library `libgcc_s.1.1.dylib` is now also copied with the rest of dependencies.

**Note: If the machine has installed gcc with brew, it will work, since the missing library is included with it.**

### Testing I've completed

- Build and execute on MacOS Big Sur 11 (virtual machine), using `otool -L` to make sure the paths are correct.
- Build on CI in my fork and execute on a clean MacOS Big Sur 11 (virtual machine).
- Manually removed the library in the installed files to check if the error appears again (and check that it wasn't working for any other reason, like retrieving the library from the gcc installation).

Test on real MacOS machines, and different versions.

### CHANGELOG.md (choose one)

- no need to update because this is a CI/makefiles issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3334)

### DIFF
--- a/Vendors/tropter/cmake/CMakeLists.txt
+++ b/Vendors/tropter/cmake/CMakeLists.txt
@@ -23,6 +23,7 @@ if(TROPTER_COPY_DEPENDENCIES AND APPLE)
             ${gfortran}
             ${quadmath}
             ${gcc_libdir}/libgcc_s.1.dylib
+            ${gcc_libdir}/libgcc_s.1.1.dylib
 
             DESTINATION ${CMAKE_INSTALL_LIBDIR})
 
--- a/Vendors/tropter/cmake/tropter_install_mac_dependency_libraries.cmake.in
+++ b/Vendors/tropter/cmake/tropter_install_mac_dependency_libraries.cmake.in
@@ -54,7 +54,7 @@ macro(install_name_tool_delete_rpath lib rpath)
 endmacro()
 
 # Get the directory containing libgcc_s.1.dylib.
-# otool: Get the list of dependencies of gfortran, which includes libgcc_s.
+# otool: Get the list of dependencies of coinmumps, which includes libgcc_s.
 # grep: Isolate the line containing libgcc.
 # perl: Extract only the directory, without the filename.
 #    ^\\s+     matches all spaces at the start of the string.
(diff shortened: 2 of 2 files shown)


aymanhab:
@AlbertoCasasOrtiz the proof that this works is to build and install on a fresh machine that has no libgcc_s.1.1 or to build the GUI on osx based on this branch and be able to launch without setting the DYLIB_LIBRARY_PATH. Also does libcoinmumps get built regardless of whether tropter or casadi are used?

AlbertoCasasOrtiz:
@aymanhab I have just tested the generated artifact in a fresh machine without the library and without setting DY_LIBRARY_PATH and it works. I am not sure if libcoinbmumps is build everytime, I have to check that.

Vendors/tropter/cmake/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ if(TROPTER_COPY_DEPENDENCIES AND APPLE)
             ${gfortran}
             ${quadmath}
             ${gcc_libdir}/libgcc_s.1.dylib
+            ${gcc_libdir}/libgcc_s.1.1.dylib
```
aymanhab, line 27: Can we introduce a cmake variable with the name of the dylib for future maintenance so in case it changes in the future we're not digging through the complicated mess you had to go thru this time around @AlbertoCasasOrtiz

AlbertoCasasOrtiz: I have created an issue with the cmake variable change.

AlbertoCasasOrtiz:
I have created issue #3342 as discussed with @aymanhab , so we can merge this pr.

PR #3339, opened 2022-12-19
Update action versions to modernize and match changes in opensim-gui

aymanhab:
Fixes issue #<issue_number>

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3339)

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -80,7 +80,7 @@ jobs:
         $version = $env:match.split('=')[1]
         echo $version
         echo "VERSION=$version" >> $GITHUB_ENV
-        echo "::set-output name=version::$VERSION"
+        echo "version=$version" >> $env:GITHUB_OUTPUT
 
     - name: Build opensim-core
       # Install now to avoid building bindings twice (issue when using Visual Studio 2019).
@@ -112,7 +112,7 @@ jobs:
         python -m unittest discover --start-directory opensim/tests --verbose
 
     - name: Upload opensim-core
-      uses: actions/upload-artifact@v2
+      uses: actions/upload-artifact@v3
       with:
         name: opensim-core-${{ steps.configure.outputs.version }}-win
         path: opensim-core-${{ steps.configure.outputs.version }}.zip
@@ -215,7 +215,8 @@ jobs:
         $version = $env:match.split('=')[1]
         echo $version
         echo "VERSION=$version" >> $GITHUB_ENV
-        echo "::set-output name=version::$VERSION"
+        echo "version=$version" >> $env:GITHUB_OUTPUT
+
 
     - name: Build opensim-core
       # Install now to avoid building bindings twice (TODO: issue when using Visual Studio 2019, is this an issue too in Visual Studio 2022?).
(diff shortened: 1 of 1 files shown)


AlbertoCasasOrtiz:
@aymanhab The current changes look good.

There are still a few commands that need to be updated in order to remove the `Node.js 12 actions are deprecated.` warnings.

For example, the version of the action in the following line should be 3 and not 2.

https://github.com/opensim-org/opensim-core/blob/9bf3623a33ad9e8aba884113c6cb6af3efa47af0/.github/workflows/continuous_integration.yml#L115

Also, we should change the syntaxis to set outputs for rus to the new one for the The `set-output command is deprecated and will be disabled soon.` warnings:

https://github.com/opensim-org/opensim-core/blob/9bf3623a33ad9e8aba884113c6cb6af3efa47af0/.github/workflows/continuous_integration.yml#L83

An example of the new syntaxis can be seen in this line of opensim-gui:

`echo "version=$VERSION" >> $GITHUB_OUTPUT`

Note that, for windows, the syntaxis is different since it is using powershell and not bash. It has `$env:` at the beginning of the GITHUB_OUTPUT variable.

`echo "version=$VERSION" >> $env:GITHUB_OUTPUT`

AlbertoCasasOrtiz:
@aymanhab The new changes made the Node.js deprecation errors to dissappear, Thank you.

However, there are still a few places where set-output is using the previous syntax. These are the ones I have found:

https://github.com/opensim-org/opensim-core/blob/2d426e8c53baf34f5a5a3517e533a92e87ec3326/.github/workflows/continuous_integration.yml#L345

https://github.com/opensim-org/opensim-core/blob/2d426e8c53baf34f5a5a3517e533a92e87ec3326/.github/workflows/continuous_integration.yml#L495

https://github.com/opensim-org/opensim-core/blob/2d426e8c53baf34f5a5a3517e533a92e87ec3326/.github/workflows/continuous_integration.yml#L596

https://github.com/opensim-org/opensim-core/blob/2d426e8c53baf34f5a5a3517e533a92e87ec3326/.github/workflows/continuous_integration.yml#L699

https://github.com/opensim-org/opensim-core/blob/2d426e8c53baf34f5a5a3517e533a92e87ec3326/.github/workflows/continuous_integration.yml#L803

aymanhab:
Thanks @AlbertoCasasOrtiz I only updated the windows section since it has a different syntax. Will fix the rest and tag you for review before merging.

aymanhab:
@AlbertoCasasOrtiz All warnings are gone except for deprecated environments, so this is ready to go. Please let me know otherwise. Thank you

AlbertoCasasOrtiz (approved):
:lgtm: Looks good to me. No warnings associated are appearing.

> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3339#-:-NJlyU-pCqRjC1J4-3Rg:bxouuty)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aymanhab)

aymanhab:
Thanks @AlbertoCasasOrtiz

PR #3351, opened 2023-01-03
Add center of pressure calculations to `MocoUtilities::createExternalLoadsTableForGait()`

nickbianco:
### Brief summary of changes
We currently only return net forces and torques from contact spheres when using `MocoUtilities::createExternalLoadsTableForGait()`.  This PR adds center of pressure calculations to complete the information in the returned external loads tables.

[Here is a useful reference](http://www.kwon3d.com/theory/grf/cop.html) for computing center of pressure locations.

### Testing I've completed
Local testing only, but can add unit/regression tests as requested. 

- Center of pressure values are currently calculated based on the OpenSim convention that the vertical ground contact forces is pointing the y-direction. If someone uses this function with results using a different convention, the center of pressure values will be incorrect (the forces and torques shouldn't be affected though). Should we support different conventions for the orientation of ground contact forces?

- In order to compute the correct center of pressure values, the forces and torques applied to the ground (e.g., the half space, if using `SmoothSphereHalfSpaceForce`) must be used. The current implementation, including renamed variables, is based on `SmoothSphereHalfSpaceForce`, where the half space force and torque values are returned by the second group of six values returned `getRecordValues()`  (the first group of six values are the sphere force and torque values). Should we take into consideration other contact force elements that might use different conventions to return foot and ground force information? We should definitely consider how to handle @fcanderson's new contact model (#3310), if that is to be used extensively with Moco.

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3351)

### DIFF
--- a/CHANGELOG_MOCO.md
+++ b/CHANGELOG_MOCO.md
@@ -3,6 +3,9 @@ Moco Change Log
 
 1.2.1
 -----
+- 2023-01-03: Add center of pressure calculations to 
+              `MocoUtilities::createExternalLoadsTableForGait`.
+
 - 2022-07-25: Added property `normalize_tracking_error` to `MocoContactTrackingGoal` 
               to normalize the 3D contact tracking error based on the contact 
               tracking data.
--- a/OpenSim/Moco/MocoUtilities.cpp
+++ b/OpenSim/Moco/MocoUtilities.cpp
@@ -321,6 +321,7 @@ TimeSeriesTable OpenSim::createExternalLoadsTableForGait(Model model,
         // Compute centers of pressure for both feet. We need to use the force
         // and torque information from the half space to compute the correct
         // locations.
+        // TODO: Support contact plane normals in any direction.
         SimTK::Vec3 copRight(0);
         copRight(0) = halfSpaceTorquesRight(2) / halfSpaceForcesRight(1);
         copRight(2) = -halfSpaceTorquesRight(0) / halfSpaceForcesRight(1);
--- a/OpenSim/Moco/MocoUtilities.cpp
+++ b/OpenSim/Moco/MocoUtilities.cpp
@@ -283,36 +283,60 @@ TimeSeriesTable OpenSim::createExternalLoadsTableForGait(Model model,
     int count = 0;
     for (const auto& state : trajectory) {
         model.realizeVelocity(state);
-        SimTK::Vec3 forcesRight(0);
-        SimTK::Vec3 torquesRight(0);
+        SimTK::Vec3 sphereForcesRight(0);
+        SimTK::Vec3 sphereTorquesRight(0);
+        SimTK::Vec3 halfSpaceForcesRight(0);
+        SimTK::Vec3 halfSpaceTorquesRight(0);
         // Loop through all Forces of the right side.
         for (const auto& smoothForce : forcePathsRightFoot) {
             Array<double> forceValues =
                     model.getComponent<Force>(smoothForce).getRecordValues(state);
-            forcesRight += SimTK::Vec3(forceValues[0], forceValues[1],
+            sphereForcesRight += SimTK::Vec3(forceValues[0], forceValues[1],
                     forceValues[2]);
-            torquesRight += SimTK::Vec3(forceValues[3], forceValues[4],
+            sphereTorquesRight += SimTK::Vec3(forceValues[3], forceValues[4],
                     forceValues[5]);
+            halfSpaceForcesRight += SimTK::Vec3(forceValues[6], forceValues[7],
+                    forceValues[8]);
+            halfSpaceTorquesRight += SimTK::Vec3(forceValues[9], forceValues[10],
+                    forceValues[11]);
         }
-        SimTK::Vec3 forcesLeft(0);
-        SimTK::Vec3 torquesLeft(0);
+        SimTK::Vec3 sphereForcesLeft(0);
+        SimTK::Vec3 sphereTorquesLeft(0);
+        SimTK::Vec3 halfSpaceForcesLeft(0);
+        SimTK::Vec3 halfSpaceTorquesLeft(0);
         // Loop through all Forces of the left side.
         for (const auto& smoothForce : forcePathsLeftFoot) {
             Array<double> forceValues =
                     model.getComponent<Force>(smoothForce).getRecordValues(state);
-            forcesLeft += SimTK::Vec3(forceValues[0], forceValues[1],
+            sphereForcesLeft += SimTK::Vec3(forceValues[0], forceValues[1],
                     forceValues[2]);
-            torquesLeft += SimTK::Vec3(forceValues[3], forceValues[4],
+            sphereTorquesLeft += SimTK::Vec3(forceValues[3], forceValues[4],
                     forceValues[5]);
+            halfSpaceForcesLeft += SimTK::Vec3(forceValues[6], forceValues[7],
+                    forceValues[8]);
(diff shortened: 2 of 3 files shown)


nickbianco:
@antoinefalisse, this might be of interest to you.

carmichaelong (changes requested):
Just some comments on the documentation, otherwise the code all looks good to me.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3351#-:-NLbUNd-3M5xqB_d9B9e:b-76ubya)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @nickbianco)

___
*[`OpenSim/Moco/MocoUtilities.h` line 235 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3351#-NLbSOqb-EcpSL26zDhx:-NLbSOqcEOlNE9yXXUn1:b3oum8u) ([raw file](https://github.com/opensim-org/opensim-core/blob/03632efea6bf072971f0845b96253a281bcc652a/OpenSim/Moco/MocoUtilities.h#L235)):*
> ```objc
> /// getRecordValues(); this order is of use for, for example, the
> /// SmoothSphereHalfSpaceForce contact model but might have a different meaning
> /// for different contact models. Centers of pressure are computed assuming the
> ```

I think with this new comment about the CoP, the comment above should be removed: `Hence, the centers of pressure are at the origin.`
___
*[`OpenSim/Moco/MocoUtilities.h` line 236 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3351#-NLbSl3S7S6FWCLtv6em:-NLbSl3S7S6FWCLtv6en:bbsmv6n) ([raw file](https://github.com/opensim-org/opensim-core/blob/03632efea6bf072971f0845b96253a281bcc652a/OpenSim/Moco/MocoUtilities.h#L236)):*
> ```objc
> /// SmoothSphereHalfSpaceForce contact model but might have a different meaning
> /// for different contact models. Centers of pressure are computed assuming the
> /// that the vertical ground contact force is in the y-direction, which is the
> ```

Is `vertical ground contact force in the y-direction` equivalent to saying that the contact plane's normal in the y-direction?
___
*[`OpenSim/Moco/MocoUtilities.h` line 237 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3351#-NLbT6MqDzh5G2BtkKN1:-NLbT6MqDzh5G2BtkKN2:bmprqvp) ([raw file](https://github.com/opensim-org/opensim-core/blob/03632efea6bf072971f0845b96253a281bcc652a/OpenSim/Moco/MocoUtilities.h#L237)):*
> ```objc
> /// for different contact models. Centers of pressure are computed assuming the
> /// that the vertical ground contact force is in the y-direction, which is the
> /// OpenSim convention. In addition, the centers of pressure are computed using
> ```

These last lines seem to be related to the comment above:
```
///  The
/// forces and torques used are taken from the first six outputs of
/// getRecordValues(); this order is of use for, for example, the
/// SmoothSphereHalfSpaceForce contact model but might have a different meaning
/// for different contact models. 
```

It would be good to unify these comments, and perhaps make them more concrete or both users and developers. For instance, for users, it could be helpful to list the Geometry that this will work properly with. For developers, it could be helpful to explicitly list the sufficient order of values needed to make this work (with the idea that future dev's can plan their new`Force` elements this way)
___
*[`OpenSim/Moco/MocoUtilities.cpp` line 324 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3351#-NLbU41gA_6f2DRmqDuq:-NLbU41gA_6f2DRmqDur:b-1pv2q7) ([raw file](https://github.com/opensim-org/opensim-core/blob/03632efea6bf072971f0845b96253a281bcc652a/OpenSim/Moco/MocoUtilities.cpp#L324)):*
> ```cpp
>         // and torque information from the half space to compute the correct
>         // locations.
>         SimTK::Vec3 copRight(0);
> ```

Could add a TODO to make this more general than just y-direction up alone. Perhaps representing the forces and torques in a frame corresponding to the half-space, and then doing the same calculations?

nickbianco (commented):
@carmichaelong, ready for review.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3351#-:-NLcAJqH93KhPWI5DV8F:bzarhxw)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @carmichaelong)

___
*[`OpenSim/Moco/MocoUtilities.h` line 235 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3351#-NLbSOqb-EcpSL26zDhx:-NLcAFFg1RUGsfSHKAxi:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/03632efea6bf072971f0845b96253a281bcc652a/OpenSim/Moco/MocoUtilities.h#L235)):*

I think with this new comment about the CoP, the comment above should be removed: `Hence, the centers of pressure are at the origin.`
</blockquote></details>

Done.
___
*[`OpenSim/Moco/MocoUtilities.h` line 236 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3351#-NLbSl3S7S6FWCLtv6em:-NLcAFt23OPUx7Uh29AL:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/03632efea6bf072971f0845b96253a281bcc652a/OpenSim/Moco/MocoUtilities.h#L236)):*

Is `vertical ground contact force in the y-direction` equivalent to saying that the contact plane's normal in the y-direction?
</blockquote></details>

Done.
___
*[`OpenSim/Moco/MocoUtilities.h` line 237 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3351#-NLbT6MqDzh5G2BtkKN1:-NLcAGKI-1Yy08vkST1p:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/03632efea6bf072971f0845b96253a281bcc652a/OpenSim/Moco/MocoUtilities.h#L237)):*

These last lines seem to be related to the comment above:
```
///  The
/// forces and torques used are taken from the first six outputs of
/// getRecordValues(); this order is of use for, for example, the
/// SmoothSphereHalfSpaceForce contact model but might have a different meaning
/// for different contact models. 
```

It would be good to unify these comments, and perhaps make them more concrete or both users and developers. For instance, for users, it could be helpful to list the Geometry that this will work properly with. For developers, it could be helpful to explicitly list the sufficient order of values needed to make this work (with the idea that future dev's can plan their new`Force` elements this way)
</blockquote></details>

Done.
___
*[`OpenSim/Moco/MocoUtilities.cpp` line 324 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3351#-NLbU41gA_6f2DRmqDuq:-NLcAH5ECO3iVigStYFZ:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/03632efea6bf072971f0845b96253a281bcc652a/OpenSim/Moco/MocoUtilities.cpp#L324)):*

Could add a TODO to make this more general than just y-direction up alone. Perhaps representing the forces and torques in a frame corresponding to the half-space, and then doing the same calculations?
</blockquote></details>

Done.

carmichaelong (approved):
LGTM pending CI

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3351#-:-NLhbXXr8B2xYQqrpqmB:bjeobiu)* status: 0 of 3 files reviewed, all discussions resolved

nickbianco:
@carmichaelong, it seems like the build failures are unrelated to this PR. I'm okay merging if you are.

Issue #3294, opened 2022-09-15
Setting exponent to 1 in MocoOutputGoal can return negative values


carmichaelong:
https://github.com/opensim-org/opensim-core/blob/c6dc2a5913df2cb7e7a72da1ebb162fda9eaa022/OpenSim/Moco/MocoGoal/MocoOutputGoal.cpp#L84

If the exponent is set to 1, the integrand can return negative values. Looks like all other values for the exponent will return positive values.

nickbianco:
We can't use the `abs()` function here since it is non-smooth (it's fine to use `abs()` in the third branch of the if-statement since the resulting value is raised some a power > 1, making it smooth).

We should improve the documentation for `setExponent()` to make it clear what happens in each branch of the if-statement.

PR #3363, opened 2023-01-18
Clarify the behavior of exponents in `MocoOutputGoal`

nickbianco:
Fixes issue #3294 

### Brief summary of changes
Add clarification to `MocoOutputGoal::setExponent()` to clarify how different exponent values allow or do not allow negative values.

### CHANGELOG.md (choose one)

- no need to update because...simple documentation update.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3363)


carmichaelong (approved):
LGTM. BTW, I haven't tried this in awhile, but I think adding [ci skip] to a commit message will prevent CI from triggering, which can be helpful for these documentation issues.

Issue #3409, opened 2023-02-23
Segfault when using an `OpenSim::Model` *copy* of a model containing empty property fields


adamkewley:
This was found in: https://github.com/ComputationalBiomechanicsLab/opensim-creator/issues/621

The following bug reproduction osim:

- https://gist.github.com/adamkewley/0787aee030d6a21f1a5a73f08debf7cd

Will segfault if `finalizeFromProperties` is called on a *copy* of an `OpenSim::Model` constructed from it. I.e. the following code would segfault:

```c++
OpenSim::Model m1{"broken_file.osim"};
m1.finalizeFromProperties();
OpenSim::Model m2{m1};
m2.finalizeFromProperties();  // boom
```

OpenSim Creator makes many copies of models for undo/redo storage and automatic error recovery, so it crashes essentially instantly on loading that buggy file. OpenSim GUI will load the file, but crash after doing something that copies the model, such as running a forward-dynamic simulation. So a practical repro is:

- Save the broken osim somewhere
- Open it in OpenSim GUI
- Run forward-dynamic simulation (green guy)
- The 4.4 GUI crashes

adamkewley:
It is likely that the bug is a consequence of several confounding factors:

- If a `Coordinate` is constructed and `constructProperties` is called, then the underlying `OpenSim::SimpleProperty<double>` is initialized with a `SimTK::Array_<double, int> = {0.0};`
- Reading the above (e.g. via `get_default_value()`) is safe
- Copying the above is safe: it will create a copied property with the same defaulted value
- When an XML document is read, it goes via:
  - `OpenSim::Object::updateFromXMLNode`
  - `OpenSim::AbstractProperty::readFromXMLParentElement`
  - `OpenSim::SimpleProperty<double>::readFromXMLElement`
  - `OpenSim::SimpleProperty<double>::readSimplePropertyFromStream`
  - `SimTK::readUnformatted(std::istream&, SimTK::Array_<double, int>&)`
  - Which `clear`s the array and tries to re-fill it with the XML node's content ([Array.h](https://github.com/simbody/simbody/blob/Simbody-3.7/SimTKcommon/include/SimTKcommon/internal/Array.h#L3519))
  - If reading fails, it will not push any element into the array, leaving the supplied (property-owned) array empty
[...]

PR #3411, opened 2023-02-24
Patched segfault from reading invalid osim data (#3409)

adamkewley:
- Adds bounds checks to OpenSim::SimpleProperty<T>, so that out-of-bounds accesses result in an immediate exception rather than undefined behavior

- Changes XML reading routines in OpenSim::SimpleProperty<T> to behave transactionally, such that failing to read the XML resets the property back to its initial values (rather than clearing it)

Fixes issue #3409

### Testing I've completed

- Loaded bug reproduction file in OpenSim Creator after applying this patch and it works fine (prints warning, but does not segfault). Opened all example `osim` files from `opensim-models` with the patched build and did not encounter any change in behavior.

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3411)


aymanhab (approved):
LGTM, thanks for the fix

Issue #3397, opened 2023-02-09
Find a place to locate the build scripts


AlbertoCasasOrtiz:
The build scripts referenced on the [wiki](https://github.com/opensim-org/opensim-core/wiki/Build-Instructions) are currently gists created and located on my personal profile. Looks like organizations cannot create gists, so creating opensim-org gists is not an option. 

Any ideas of where we can place these scripts in an accessible and usable way? @aymanhab @jenhicks @carmichaelong @nickbianco

nickbianco:
I know including them in the repo makes them harder to download, but most users will need to clone the repo before building anyway. We've kept build scripts [here ](https://github.com/opensim-org/opensim-core/tree/main/.github) before, but we could find a more visible location.

adamkewley:
It's fairly (imo) common to have something like a `scripts/` directory in a repo, where any project-related scripts (incl. build, one-time scripts, asset packers, etc.) get dumped. Revealing my bias, `opensim-creator` does exactly that:

https://github.com/ComputationalBiomechanicsLab/opensim-creator/tree/main/scripts

(although you can see one problem, which is that all kinds of crap tends to build up in there over time)

AlbertoCasasOrtiz:
Thank you @nickbianco for looking into this. I think we should do the same, but controlling which kind of scripts end up in the folder and organizing them. Should I locate the build scripts into `scripts/build/`, so every new script is created inside of `scripts/`, and we can organize into folders as new ones are created?

This way, as suggested by @aymanhab some time ago, we can also have a `scripts/contributions/` folder or something like that with scripts created by users, with a disclaimer saying that those scripts were created by users and we are not maintaining them.

aymanhab:
I like the idea of having a scripts folder as well. I'd keep these for development purposes (inspired by @adamkewley list). Not sure what else is available to individual users and not organizations but it may make sense to make a github user e.g. opensim-dev and use that account for situations that an "official" opensim account is required.

AlbertoCasasOrtiz:
Solved by adding build scripts to /scripts folder in both `opensim-core` and `opensim-gui`.

Issue #3369, opened 2023-01-23
Builds created in Ubuntu 20.04 are throwing error in Ubuntu 18.04


AlbertoCasasOrtiz:
While solving #2930, I noticed that builds created in Ubuntu 20.04 are throwing error in Ubuntu 18.04 (when executing `opensim-cmd`). The error is the following:

```
m-cmd: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by ./opensim-cmd)
./opensim-cmd: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /home/albert/Downloads/OpenSim-4.4-2023-01-19-228bddf-ub20-linux/OpenSim-4.4-2023-01-19-228bddf/opensim/bin/../sdk/lib/libosimMoco.so)
./opensim-cmd: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by /home/albert/Downloads/OpenSim-4.4-2023-01-19-228bddf-ub20-linux/OpenSim-4.4-2023-01-19-228bddf/opensim/bin/../sdk/lib/libosimMoco.so)
./opensim-cmd: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by /home/albert/Downloads/OpenSim-4.4-2023-01-19-228bddf-ub20-linux/OpenSim-4.4-2023-01-19-228bddf/opensim/bin/../sdk/lib/libosimTools.so)
./opensim-cmd: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /home/albert/Downloads/OpenSim-4.4-2023-01-19-228bddf-ub20-linux/OpenSim-4.4-2023-01-19-228bddf/opensim/bin/../sdk/lib/libosimTools.so)
./opensim-cmd: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /home/albert/Downloads/OpenSim-4.4-2023-01-19-228bddf-ub20-linux/OpenSim-4.4-2023-01-19-228bddf/opensim/bin/../sdk/lib/libosimAnalyses.so)
[...]

PR #3418, opened 2023-03-07
Added build scripts (Ubuntu, Mac, Linux) and Dockerfile (Ubuntu 18.04)

AlbertoCasasOrtiz:
Fixes issue #3397 and Fixes issue #3369

### Brief summary of changes

Copied build scripts here, and use this folder as a permanent place (currently the scripts are in my personal account as gists, so we discussed in #3397 that this could be a better place).

### Testing I've completed

**build-scripts:** Tested in previous iterations of the scripts, now I am only copying them here.
**docker script:** Tested by downloading generated artifact and executing it in an Ubuntu 18.04 VM.

Should we add opensim-gui build scripts here too, or should we create another script folder in opensim-gui's repository?

The docker script is for opensim-core. Should I create another for opensim-gui?

### CHANGELOG.md (choose one)

- no need to update because internal scripts.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3418)


halleysfifthinc (commented):
There's a typo in the folder name (`buiild scripts`), and I would recommend not using spaces, which are a pain to deal with on the command line. The main build scripts could just live directly in the `scripts` folder.

AlbertoCasasOrtiz:
Thank you @halleysfifthinc for reviewing this. I agree, should not be spaces in the path, and good catch the typo. I have renamed the folder `buiild scripts` to `build`. Same with `docker/Ubuntu 18.04`, it is now `docker/ubuntu18.04`. I think we should have the build scripts separated, just in case in the future more scripts are added to this folder, we have them organized and they are easier to find.

Issue #3296, opened 2022-09-19
[Moco] Error with exception handling when problems with path constraints fail


nickbianco:
The following error message is printed:
```
casVector should be 1-dimensional, but has size 0 x 0.
Thrown at MocoCasOCProblem.h:108 in convertToSimTKVector().
```
Is this the time vector, or the constraint errors vector? (Or something else?)

Noted by Hojin Song and Edward Syrett on the forum.

nickbianco:
Exception handling happens here: https://github.com/opensim-org/opensim-core/blob/ace2ad43a40740648299cd09d6c1d0fc30541ff1/OpenSim/Moco/MocoCasADiSolver/MocoCasADiSolver.cpp#L373

Issue #3428, opened 2023-03-21
`OpenSim::log_cout()` does not print to `std::cout` for `Logger::Level::Warn` and higher


nickbianco:
`OpenSim::log_cout()` does not print output to `std::cout` when the `Logger` is set to `Level::Warn` or higher (i.e., `Level::Debug` prints, but `Level::Error` does not). This was detected for the following `OpenSim::log_cout()` calls in `CasOC::Transcription`:

https://github.com/opensim-org/opensim-core/blob/d9ebf9389fabfe9cceb8ab46122e90b69a07689e/OpenSim/Moco/MocoCasADiSolver/CasOCTranscription.cpp#L1143

and 

https://github.com/opensim-org/opensim-core/blob/d9ebf9389fabfe9cceb8ab46122e90b69a07689e/OpenSim/Moco/MocoCasADiSolver/CasOCTranscription.cpp#L1163

nickbianco:
@aymanhab, this issue specifically didn't lead to the zero vector error, this is a separate issue that I noticed while fixing that one. The linux/docker container produced the same error message, but in that issue, there were no path constraints in the problem.

nickbianco:
I guess it's possible that `fmt::format` is also causing the issue in #3085. We could expand this issue or open another issue to investigate.

PR #3429, opened 2023-03-21
Fix issue where problems with path constraints return a zero time vector

nickbianco:
Fixes issue #3296 

### Brief summary of changes
In `CasOC::Transcription` we print out a summary of constraint errors if a problem fails. A call to `fmt::format` was somehow  silently corrupting the optimal control solution, which then returned a zero time vector, causing the linked issue to occur. Changing the `fmt::format` call to remove the integer formatting seems to fix the issue. 

I also made some minor formatting changes to the constraint error output.

### Looking for feedback
I noticed another issue while debugging this one: `OpenSim::log_cout()` does not print output to `std::cout` when the `Logger` is set to `Level::Warn` or higher (i.e., `Level::Debug` prints, but `Level::Error` does not). However, I've only tested this issue for the `OpenSim::log_cout()` calls in `CasOC::Transcription`. In the interest of keeping this fix brief, I've temporarily replaced the `log_cout` calls with `std::cout` so users can get the useful information we compile in the meantime while we fix the issue (which I've captured in #3428 for now).

### Testing I've completed
Added test to `testMocoConstraints.cpp`.

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3429)

### DIFF
--- a/CHANGELOG_MOCO.md
+++ b/CHANGELOG_MOCO.md
@@ -3,6 +3,9 @@ Moco Change Log
 
 1.2.1
 -----
+- 2023-03-21: Fixed a bug where failing `MocoProblem`s with path constraints returned
+              a zero time vector.
+
 - 2023-03-08: Added `MocoTrajectory::trimToIndices`.
 
 - 2023-02-25: Added `getSphereForce`, `getHalfSpaceForce`, and associated `Output`s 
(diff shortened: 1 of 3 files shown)


nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3429)* status: all files reviewed, 1 unresolved discussion (waiting on @aymanhab)

___
*[`OpenSim/Moco/Test/testMocoConstraints.cpp` line 994 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3429#-NR9Ct4k2o-kTTHJtx_G:-NR9V3h86uDav3ZBlB_W:bn1jy4c) ([raw file](https://github.com/opensim-org/opensim-core/blob/01f872ea3b9e786151e4a57cabdd3bab77042e17/OpenSim/Moco/Test/testMocoConstraints.cpp#L994)):*

Maybe a note on why this should fail would be helpful to future maintainers of the code. If that's obvious then please ignore.
</blockquote></details>

I honestly don't know why the failure occurs. I did a little digging into spdlog but didn't come up with anything useful.

nickbianco:
Thanks @aymanhab!

Issue #3435, opened 2023-04-04
`CoordinateCouplerConstraint` does not handle multiple independent coordinates properly


nickbianco:
`CoordinateCouplerConstraint` should support multiple independent coordinates since the property `independent_coordinate_names` is a list property. However, the function that computes the coordinate coupling error only takes into account the first two coordinate in the total list of coordinate, whether or not they are independent or dependent:

https://github.com/opensim-org/opensim-core/blob/22af32b0191a20eef50940574bbfcd21bd93c972/OpenSim/Simulation/SimbodyEngine/CoordinateCouplerConstraint.cpp#L47

bfregly:
We would like to be able to have two (or more!) independent coordinates as inputs to a coordinate coupler constraint as well. Please let us know when this capability is available. You will probably need to add the ability to handle a <coupled_coordinates_function> that is defined analytically, like sqrt(x^2/2+y^2/4) where dependent coordinate z is an analytical function of independent coordinates x and y.

PR #3440, opened 2023-04-25
[WIP] Fix `CoordinateCouplerConstraint` bug preventing multiple independent coordinates

jenniferMaier:
Fixes issue #3435

### Brief summary of changes
The methods calcValue and calcDerivative in `CoordinateCouplerConstraint` used entry 0 of the input vector x as independent coordinate and entry 1 as dependent coordinate of the coupling. In the case of multiple independent coordinates this led to unexpected behavior since only the first two entries of vector x are considered (e.g., in the case of two independent coordinates, the first independent coordinate given as x[0] was used as independent coordinate, the second independent coordinate given as x[1] was used as dependent coordinate, and the actual dependent coordinate given as x[2] was ignored).

The change now considers the length of the vector x, and uses all entries but the last as independent coordinates, and the last entry as dependent coordinate, and passes them on to the implementation of calcValue and calcDerivative, respectively. Additionally, in calcDerivative, the derivative is only computed if it is with respect to one of the independent coordinates.

### Testing I've completed
No unit testing.
Model coordinates show expected behavior if the constraint is used with a `LinearFunction` with two independent coordinates.

### CHANGELOG.md (choose one)
- updated?

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3440)

### DIFF
--- a/OpenSim/Simulation/SimbodyEngine/CoordinateCouplerConstraint.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/CoordinateCouplerConstraint.cpp
@@ -45,9 +45,13 @@ public:
     }
 
     double calcValue(const SimTK::Vector& x) const override {
-        SimTK::Vector xf(1);
-        xf[0] = x[0];
-        return scale*f1->calcValue(xf)-x[1];
+        int n_indep_coords = x.size() - 1;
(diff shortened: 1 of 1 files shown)


nickbianco:
Closing per the redundant PR #3453.

PR #3442, opened 2023-05-01
Millard muscle curves deduplication

pepbos:
This PR de-duplicates the data held in `SmoothSegmentedFunction` by implementing a global cache for its data, such that when the cache is given identical curve parameters, it returns a read-only, shared copy of the curve data.

The significance of de-duplicating `SmoothSegmentedFunction` is that it's used extensively by muscle components such as `Millard2012EquilibriumMuscle`. The Millard muscle uses 5 different curves, but between individual muscle instances these curves are typically identical. Taking the `RajagopalModel`  as example:

We find the following `SmoothSegmentedFunction` curves in this model, each holding multiple `SimTK::Spline`:
- `ActiveForceLengthCurve` -> 5 splines
- `ForceVelocityInverseCurve` -> 4 splines
- `FiberForceLengthCurve` -> 2 splines
- `TendonForceLengthCurve` -> 2 splines
- `ForceVelocityCurve` -> 4 splines

With each spline ~ 200 doubles = 1.6 kB
Before deduplication the Rajagopal model has ~80 copies of each of the above curves which sums up to ~2MB. Applying deduplication reduces this to ~27kB.

### Brief summary of changes
The changes apply to `SmoothSegmentedFunction.h/cpp`:
- `SmoothSegmentedFunctionParameters` is added to collect all parameters that are used to construct a new smooth function, e.g. `_x0`, `_x1` etc...
- `SmoothSegmentedFunctionData` is added to collect all data required to evaluate the smooth function, e.g. `_arraySplineUX_`, etc. This data can be constructed from the `SmoothSegmentedFunctionParameters`.
- `SmoothSegmentedFunctionDataCache` is added to hold an `unordered_map` of smart pointers to the data, with the hashed-parameters used as key. Whenever a muscle curve is constructed, the parameters are used to obtain a pointer to already existing data, or to construct a new one.
- `SmoothSegmentedFunction` holds a shared pointer to the `SmoothSegmentedFunctionData`.

### Performance Results

Performance was evaluated against current master:

| Test Name | lhs [secs] | σ [secs] | rhs [secs] | σ [secs] | Speedup |
|`ToyDropLanding_nomuscles` | 0.03 | 0.00 | 0.03 | 0.00 | 1.10 | |
|`passive_dynamic` | 1.79 | 0.00 | 1.79 | 0.00 | 1.00 |
| `RajagopalModel` | 4.66 | 0.00 | 4.35 | 0.01 | 1.07 | 
| `Arm26` | 0.11 | 0.00 | 0.10 | 0.00 | 1.02 | 
| `Gait2354` | 0.14 | 0.00 | 0.14 | 0.00 | 1.02 | 
| `ToyDropLanding` | 4.74 | 0.00 | 4.70 | 0.00 | 1.01 | 
| `passive_dynamic_noanalysis` | 1.17 | 0.00 | 1.18 | 0.00 | 1.00 |

As expected, the `RajagopalModel` can be seen to have the most significant performance boost, because it has many `Millard2012EquilibriumMuscle`s.

### CHANGELOG.md

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3442)

### DIFF
--- a/OpenSim/Common/SmoothSegmentedFunction.cpp
+++ b/OpenSim/Common/SmoothSegmentedFunction.cpp
@@ -127,26 +127,26 @@ bool operator==(
 //=============================================================================
 
 template <typename T>
-inline std::size_t HashCombine(std::size_t seed, const T& v)
+inline size_t HashCombine(size_t seed, const T& v)
 {
     std::hash<T> hasher;
     return seed ^ (hasher(v) + 0x9e3779b9 + (seed << 6) + (seed >> 2));
 }
 
-template <typename T> inline std::size_t HashOf(const T& v)
+template <typename T> inline size_t HashOf(const T& v)
 {
     return std::hash<T>{}(v);
 }
 
 template <typename T, typename... Others>
-inline std::size_t HashOf(const T& v, const Others&... others)
+inline size_t HashOf(const T& v, const Others&... others)
 {
     return HashCombine(HashOf(v), HashOf(others...));
 }
 
-template <> inline std::size_t HashOf(const SimTK::Matrix& matrix)
+template <> inline size_t HashOf(const SimTK::Matrix& matrix)
 {
-    std::size_t hash = HashOf(matrix.nrow(), matrix.ncol());
+    size_t hash = HashOf(matrix.nrow(), matrix.ncol());
     for (int r = 0; r < matrix.nrow(); ++r) {
         for (int c = 0; c < matrix.ncol(); ++c) {
             hash = HashCombine(hash, matrix.row(r)[c]);
--- a/OpenSim/Common/SmoothSegmentedFunction.cpp
+++ b/OpenSim/Common/SmoothSegmentedFunction.cpp
@@ -77,29 +77,18 @@ struct SmoothSegmentedFunctionParameters
         _computeIntegral(computeIntegral),
         _intx0x1(intx0x1) {}
 
-    // Uninitialized parameters.
-    SmoothSegmentedFunctionParameters() :
-        _mX(SimTK::Matrix()),
-        _mY(SimTK::Matrix()),
-        _x0(SimTK::NaN),
-        _x1(SimTK::NaN),
-        _y0(SimTK::NaN),
-        _y1(SimTK::NaN),
-        _dydx0(SimTK::NaN),
-        _dydx1(SimTK::NaN),
-        _computeIntegral(false),
-        _intx0x1(false) {}
+    SmoothSegmentedFunctionParameters() {}
 
-    SimTK::Matrix _mX;
-    SimTK::Matrix _mY;
-    double _x0;
-    double _x1;
-    double _y0;
-    double _y1;
-    double _dydx0;
-    double _dydx1;
-    bool _computeIntegral;
-    bool _intx0x1;
+    SimTK::Matrix _mX = SimTK::Matrix();
+    SimTK::Matrix _mY = SimTK::Matrix();
+    double _x0 = SimTK::NaN;
+    double _x1 = SimTK::NaN;
+    double _y0 = SimTK::NaN;
+    double _y1 = SimTK::NaN;
+    double _dydx0 = SimTK::NaN;
+    double _dydx1 = SimTK::NaN;
+    bool _computeIntegral = false;
+    bool _intx0x1 = false;
 };
 
 bool operator==(
--- a/OpenSim/Common/SmoothSegmentedFunction.cpp
+++ b/OpenSim/Common/SmoothSegmentedFunction.cpp
@@ -295,21 +295,23 @@ public:
         std::lock_guard<std::mutex> guard{_cacheMutex};
         auto it = _cache.find(params);
         if (it != _cache.end()) {
-            return it->second;
-        } else {
-            std::shared_ptr<OpenSim::SmoothSegmentedFunctionData> ptr =
+            auto data_ptr = it->second.lock();
+            if (data_ptr) {
+                return data_ptr;
+            }
+        }
+        return _cache.insert({
+                params,
                 std::make_shared<SmoothSegmentedFunctionData>(
                     SmoothSegmentedFunctionData(params, name)
-                );
-            _cache[params] = ptr;
-            return ptr;
-        }
+                )
+            }).first->second.lock();
     }
 
 private:
     std::mutex _cacheMutex;
     std::unordered_map<SmoothSegmentedFunctionParameters,
-        std::shared_ptr<const SmoothSegmentedFunctionData>> _cache;
+        std::weak_ptr<const SmoothSegmentedFunctionData>> _cache;
 };
 
 std::shared_ptr<const OpenSim::SmoothSegmentedFunctionData>
--- a/OpenSim/Common/SmoothSegmentedFunction.cpp
+++ b/OpenSim/Common/SmoothSegmentedFunction.cpp
@@ -186,8 +186,34 @@ struct SmoothSegmentedFunctionData
 {
 
     SmoothSegmentedFunctionData(
-        const SmoothSegmentedFunctionParameters& params,
+        const SimTK::Matrix& mX,
+        const SimTK::Matrix& mY,
+        double x0,
+        double x1,
+        double y0,
+        double y1,
+        double dydx0,
+        double dydx1,
+        bool computeIntegral,
+        bool intx0x1,
         const std::string& name);
+
+    SmoothSegmentedFunctionData(
+        const SmoothSegmentedFunctionParameters& params,
+        const std::string& name):
+        SmoothSegmentedFunctionData(
+            params._mX,
+            params._mY,
+            params._x0,
+            params._x1,
+            params._y0,
+            params._y1,
+            params._dydx0,
+            params._dydx1,
+            params._computeIntegral,
+            params._intx0x1,
+            name
+        ) {}
 
     /**Array of spline fit functions X(u) for each Bezier elbow*/
     SimTK::Array_<SimTK::Spline> _arraySplineUX;
@@ -205,6 +231,26 @@ struct SmoothSegmentedFunctionData
 
     /**The number of quintic Bezier curves that describe the relation*/
     int _numBezierSections;
+
+    /**The minimum value of the domain*/
+    double _x0;
+    /**The maximum value of the domain*/
+    double _x1;
+    /**The minimum value of the range*/
+    double _y0;
+    /**The maximum value of the range*/
+    double _y1;
+    /**The slope at _x0*/
+    double _dydx0;
+    /**The slope at _x1*/
+    double _dydx1;
+    /**This is the users */
+    bool _computeIntegral;
+
+    /**This variable, when true, indicates that the user wants the integral
+    from left to right (x0 to x1). If it is false, the integral from right
+    to left (x1 to x0) is computed*/
+    bool _intx0x1;
 };
 
 } // namespace OpenSim
@@ -326,20 +372,26 @@ SmoothSegmentedFunction::~SmoothSegmentedFunction() noexcept = default;
         N.B. These costs are dependent on SegmentedQuinticBezierToolkit
 */
 SmoothSegmentedFunctionData::SmoothSegmentedFunctionData(
-        const SmoothSegmentedFunctionParameters& params,
-        const std::string& name)
+    const SimTK::Matrix& mX,
+    const SimTK::Matrix& mY,
+    double x0,
+    double x1,
+    double y0,
+    double y1,
+    double dydx0,
+    double dydx1,
+    bool computeIntegral,
+    bool intx0x1,
+    const std::string& name):
+    _x0(x0),
+    _x1(x1),
+    _y0(y0),
+    _y1(y1),
+    _dydx0(dydx0),
+    _dydx1(dydx1),
+    _computeIntegral(computeIntegral),
+    _intx0x1(intx0x1)
 {
-    const SimTK::Matrix& mX = params._mX;
-    const SimTK::Matrix& mY = params._mY;
-    double x0 = params._x0;
-    double x1 = params._x1;
-    double y0 = params._y0;
-    double y1 = params._y1;
-    double dydx0 = params._dydx0;
-    double dydx1 = params._dydx1;
-    bool computeIntegral = params._computeIntegral;
-    bool intx0x1 = params._intx0x1;
-
     _numBezierSections = mX.ncol();
 
     //////////////////////////////////////////////////
@@ -416,32 +468,23 @@ SmoothSegmentedFunction::SmoothSegmentedFunction(
     bool computeIntegral,
     bool intx0x1,
     const std::string& name):
-    _x0(x0),
-    _x1(x1),
-    _y0(y0),
-    _y1(y1),
-    _dydx0(dydx0),
-    _dydx1(dydx1),
-    _computeIntegral(computeIntegral),
-    _intx0x1(intx0x1),
+    _smoothData(
+        SmoothSegmentedFunctionDataLookup(
+            SmoothSegmentedFunctionParameters(
+                mX,
+                mY,
+                x0,
+                x1,
+                y0,
+                y1,
+                dydx0,
+                dydx1,
+                computeIntegral,
+                intx0x1),
+            name)
+        ),
     _name(name)
-{
-    SmoothSegmentedFunctionParameters params(
-        SmoothSegmentedFunctionParameters(
-            mX,
-            mY,
-            x0,
-            x1,
-            y0,
-            y1,
-            dydx0,
-            dydx1,
-            computeIntegral,
-            intx0x1)
-        );
-
-    _smoothData = SmoothSegmentedFunctionDataLookup(params, name);
-}
+{}
 
 SmoothSegmentedFunction::SmoothSegmentedFunction():
     _x0(SimTK::NaN),
@@ -477,19 +520,30 @@ ________________________________________________________________________
 
 double SmoothSegmentedFunction::calcValue(double x) const
 {
+    auto& arraySplineUX = _smoothData->_arraySplineUX;
+    auto& mXVec = _smoothData->_mXVec;
+    auto& mYVec = _smoothData->_mYVec;
+    double x0 = _smoothData->_x0;
+    double x1 = _smoothData->_x1;
+    double y0 = _smoothData->_y0;
+    double y1 = _smoothData->_y1;
+    double dydx0 = _smoothData->_dydx0;
+    double dydx1 = _smoothData->_dydx1;
+
     double yVal = 0;
-    if(x >= _x0 && x <= _x1 )
+
+    if(x >= x0 && x <= x1 )
     {
-        int idx  = SegmentedQuinticBezierToolkit::calcIndex(x,_smoothData->_mXVec);
+        int idx  = SegmentedQuinticBezierToolkit::calcIndex(x,mXVec);
         double u = SegmentedQuinticBezierToolkit::
-                 calcU(x,_smoothData->_mXVec[idx], _smoothData->_arraySplineUX[idx], UTOL,MAXITER);
+                 calcU(x,mXVec[idx], arraySplineUX[idx], UTOL,MAXITER);
         yVal = SegmentedQuinticBezierToolkit::
-                 calcQuinticBezierCurveVal(u,_smoothData->_mYVec[idx]);
+                 calcQuinticBezierCurveVal(u,mYVec[idx]);
     }else{
-        if(x < _x0){
-            yVal = _y0 + _dydx0*(x-_x0);            
+        if(x < x0){
+            yVal = y0 + dydx0*(x-x0);
         }else{
-            yVal = _y1 + _dydx1*(x-_x1);                    
+            yVal = y1 + dydx1*(x-x1);
         }    
     }
 
@@ -543,27 +597,34 @@ double SmoothSegmentedFunction::calcDerivative(double x, int order) const
 
     //QUINTIC SPLINE
 
-    
+    auto& arraySplineUX = _smoothData->_arraySplineUX;
+    auto& mXVec = _smoothData->_mXVec;
+    auto& mYVec = _smoothData->_mYVec;
+    double x0 = _smoothData->_x0;
+    double x1 = _smoothData->_x1;
+    double dydx0 = _smoothData->_dydx0;
+    double dydx1 = _smoothData->_dydx1;
+
     if(order==0){
                 yVal = calcValue(x);
     }else{
-            if(x >= _x0 && x <= _x1){        
-                int idx  = SegmentedQuinticBezierToolkit::calcIndex(x,_smoothData->_mXVec);
+            if(x >= x0 && x <= x1){
+                int idx  = SegmentedQuinticBezierToolkit::calcIndex(x,mXVec);
                 double u = SegmentedQuinticBezierToolkit::
-                                calcU(x,_smoothData->_mXVec[idx], _smoothData->_arraySplineUX[idx],
+                                calcU(x,mXVec[idx], arraySplineUX[idx],
                                 UTOL,MAXITER);
                 yVal = SegmentedQuinticBezierToolkit::
-                            calcQuinticBezierCurveDerivDYDX(u, _smoothData->_mXVec[idx],
-                            _smoothData->_mYVec[idx], order);
+                            calcQuinticBezierCurveDerivDYDX(u, mXVec[idx],
+                            mYVec[idx], order);
 /*
                             std::cout << _mX(3, idx) << std::endl;
                             std::cout << _mX(idx) << std::endl;*/
             }else{
                     if(order == 1){
-                        if(x < _x0){
-                            yVal = _dydx0;
+                        if(x < x0){
+                            yVal = dydx0;
                         }else{
-                            yVal = _dydx1;}
+                            yVal = dydx1;}
                     }else{
                         yVal = 0;}   
                 }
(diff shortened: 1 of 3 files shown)


OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +43,203 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+// PARAMETERS
+//=============================================================================
+
+/** Helper struct containing all parameters required to construct the
+ * SmoothSegmentedFunctionData.
+*/
+namespace OpenSim {
+struct SmoothSegmentedFunctionParameters {
+
+    SmoothSegmentedFunctionParameters(const SimTK::Matrix& mX,
```
adamkewley, line 57: Consider putting separate parameters on separate lines, current formatting seems random.

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +43,203 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+    double _x1;
+    double _y0;
+    double _y1;
+    double _dydx0;
+    double _dydx1;
+    bool _computeIntegral;
+    bool _intx0x1;
+};
+
+bool operator==(
```
adamkewley, line 77: This symbol is only used in this cpp file, so put it in an anonymous namespace or `static`-qualify it

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +43,203 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+    const SimTK::Matrix &lhs,
+    const SimTK::Matrix &rhs
+) {
+    if (lhs.nrow() != rhs.nrow() || lhs.ncol() != rhs.ncol()) {
+        return false;
+    }
+    for (int r = 0; r < lhs.nrow(); ++r)
```
adamkewley, line 84: Use bracers for each block here and put `return` statements on separate lines /w indenting, rather than putting it on the same line as another block creator (the `if`)

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +43,203 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+        for (int c = 0; c < lhs.ncol(); ++c)
+            if (lhs.row(r)[c] != rhs.row(r)[c]) return false;
+    return true;
+}
+
+bool operator==(
```
adamkewley, line 90: Anonymous namespace or `static`

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +43,203 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+    double _y0;
+    double _y1;
+    double _dydx0;
+    double _dydx1;
+    bool _computeIntegral;
+    bool _intx0x1;
+};
+
+bool operator==(
+    const SimTK::Matrix &lhs,
```
adamkewley, line 78: Reference qualifier usually goes on the type, not the identifier, in C++. Putting it on the identifier is more C-style (and not what OpenSim tends to do).

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +43,203 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+    lhs._mX == rhs._mX &&
+    lhs._mY == rhs._mY;
+}
+
+//=============================================================================
+// HASHING OF PARAMETERS
+//=============================================================================
+
+template<typename T>
+inline std::size_t HashCombine(std::size_t seed, T const &v) {
```
adamkewley, line 112: `static` or anonymous namespace. Naming convention in OpenSim is `T const& v`, not `T const &v`. `std::size_t` is more correct, but other OpenSim code tends to use `size_t`.

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +43,203 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+    std::hash<T> hasher;
+    return seed ^ (hasher(v) + 0x9e3779b9 + (seed << 6) + (seed >> 2));
+}
+
+template<typename T>
+inline std::size_t HashOf(T const &v) {
```
adamkewley, line 118: Same as above

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +43,203 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+    return std::hash<T>{}(v);
+}
+
+template<typename T, typename... Others>
+inline std::size_t HashOf(T const &v, Others const &...others) {
+    return HashCombine(HashOf(v), HashOf(others...));
+}
+
+template<>
+inline std::size_t HashOf(SimTK::Matrix const &matrix) {
```
adamkewley, line 128: As above

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +43,203 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+    std::size_t hash = HashOf(matrix.nrow(), matrix.ncol());
```
adamkewley, line 129: You don't need to seed this hash with the number of rows and the number of columns: iterating over the contents and combining that with a seed should be enough (or returning the seed as-is if the matrix is empty).

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +43,203 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+            if (lhs.row(r)[c] != rhs.row(r)[c]) return false;
+    return true;
+}
+
+bool operator==(
+    const SmoothSegmentedFunctionParameters &lhs,
+    const SmoothSegmentedFunctionParameters &rhs
+) {
+    return
+    lhs._x0 == rhs._x0 &&
```
adamkewley, line 95: Performing the comparison out-of-order compared to the class definition is a premature optimization that makes it harder to match the list of elements here with the member data list (e.g. to see if something is missing from the equality calculation)

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +43,203 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+    return hash;
+}
+
+} // namespace OpenSim
+
+template <>
+struct std::hash<SmoothSegmentedFunctionParameters> final
+{
+    size_t operator()(const SmoothSegmentedFunctionParameters& params) const {
+        return HashOf(
```
adamkewley, line 144: Format on separate lines - the verticality of the columns being X/Y isn't as important as being able to compare this list of parameters to the class definition's list of members (e.g. to ensure nothing is left out of the hashing function).

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +43,203 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+    }
+};
+
+//=============================================================================
+// SMOOTHSEGMENTEDFUNCTION DATA
+//=============================================================================
+
+namespace OpenSim {
+struct SmoothSegmentedFunctionData {
+        SmoothSegmentedFunctionData(
```
adamkewley, line 161: Why not define the implementation right here? The only reason to split the declaration/definition is in case the definition uses a symbol that's defined between the two (I can't see one?).

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +43,203 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+ * If the same SmoothSegmentedFunctionParameters were previously used to
+ * construct the SmoothSegmentedFunctionData, a shared pointer to that data is
+ * obtained. If the given parameters are new, a new data object is constructed
+ * and added. This prevents duplication of the SmoothSegmentedFunction data for
+ * identical curves.
+*/
+class SmoothSegmentedFunctionDataCache final {
+
+    public:
+    ~SmoothSegmentedFunctionDataCache() {}
```
adamkewley, line 195: Why do you need a user-defined destructor that does nothing?

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +43,203 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+
+    std::shared_ptr<const SmoothSegmentedFunctionData> lookup(
+            const SmoothSegmentedFunctionParameters& params,
+            const std::string& name) {
+        std::lock_guard<std::mutex> guard{_cacheMutex};
+        auto it = _cache.find(params);
+        if (it != _cache.end()) {
+            return it->second;
+        } else {
+            std::shared_ptr<OpenSim::SmoothSegmentedFunctionData> ptr =
```
adamkewley, line 205: `auto` west of a `std::make_shared<T>` is permissible because C++ developers will be familiar with the value the right-hand side of the expression returns.

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +43,203 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+                    std::make_shared<SmoothSegmentedFunctionData>(
+                            SmoothSegmentedFunctionData(params, name));
+            _cache[params] = ptr;
```
adamkewley, line 208: Even better, just make the rhs of this expression `std::make_shared<......` to remove the local temporary

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +43,203 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+
+        /**The number of quintic Bezier curves that describe the relation*/
+        int _numBezierSections;
+};
+
+//=============================================================================
+// DATA LOOKUP
+//=============================================================================
+
+/** Manages an unordered map of SmoothSegmentedFunction's Data, using the Parameters as key.
```
adamkewley, line 185: The class's name and definition (very small) is probably a better explanation of what's going on than this comment.

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +43,203 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+ * If the same SmoothSegmentedFunctionParameters were previously used to
+ * construct the SmoothSegmentedFunctionData, a shared pointer to that data is
+ * obtained. If the given parameters are new, a new data object is constructed
+ * and added. This prevents duplication of the SmoothSegmentedFunction data for
+ * identical curves.
+*/
+class SmoothSegmentedFunctionDataCache final {
```
adamkewley, line 192: This whole class should be defined in an anonymous namespace to prevent the ctor/assign/dtor/`lookup` from leaking into the public API of the resulting compilation unit (the class isn't used outside of this file?).

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +43,203 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+                            SmoothSegmentedFunctionData(params, name));
+            _cache[params] = ptr;
+            return ptr;
+        }
+    }
+
+    private:
+
+    std::mutex _cacheMutex;
+    std::unordered_map<SmoothSegmentedFunctionParameters,
```
adamkewley, line 216: This cache strongly owns all function curves - even if no other system in OpenSim is using any. Combined with the use of a `static`-lifetime variable to hold this class, it's an unmanaged memory leak.

The `static` lifetime guarantees that the *cache* will only be destroyed at process-end, but strongly consider making the cache hold `std::weak_ptr<T>`'s rather than `std::shared_ptr<T>`'s, with the necessary `lock`ing, checking, reconstruction, etc. so that the curves are correctly freed when models are destroyed.

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -90,14 +291,19 @@ static int NUM_SAMPLE_PTS = 100;
...
         N.B. These costs are dependent on SegmentedQuinticBezierToolkit
 */
-SmoothSegmentedFunction::
-  SmoothSegmentedFunction(const SimTK::Matrix& mX, const SimTK::Matrix& mY,  
-          double x0, double x1, double y0, double y1,double dydx0, double dydx1,
-          bool computeIntegral, bool intx0x1, const std::string& name):
-_x0(x0),_x1(x1),_y0(y0),_y1(y1),_dydx0(dydx0),_dydx1(dydx1),
-     _computeIntegral(computeIntegral),_intx0x1(intx0x1),_name(name)
+SmoothSegmentedFunctionData::SmoothSegmentedFunctionData(const
+SmoothSegmentedFunctionParameters &params, const std::string& name)
```
adamkewley, line 295: `&` placement

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -163,50 +369,56 @@ _x0(x0),_x1(x1),_y0(y0),_y1(y1),_dydx0(dydx0),_dydx1(dydx1),
...
-        _splineYintX = SimTK::Spline();
-        _numBezierSections = (int)SimTK::NaN;
-       
- }
-
- /*Detailed Computational Costs
- ________________________________________________________________________
-    If x is in the Bezier Curve
-                            Name     Comp.   Div.    Mult.   Add.    Assign.
+SmoothSegmentedFunction::
```
adamkewley, line 372: Break parameters onto separate lines for readability

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -163,50 +369,56 @@ _x0(x0),_x1(x1),_y0(y0),_y1(y1),_dydx0(dydx0),_dydx1(dydx1),
...
+  SmoothSegmentedFunction(const SimTK::Matrix& mX, const SimTK::Matrix& mY,  
+          double x0, double x1, double y0, double y1,double dydx0, double dydx1,
+          bool computeIntegral, bool intx0x1, const std::string& name):
+_x0(x0),_x1(x1),_y0(y0),_y1(y1),_dydx0(dydx0),_dydx1(dydx1),
```
adamkewley, line 376: `_x0`, `_x1`, etc. could also be part of the deduplicated datastructure

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -163,50 +369,56 @@ _x0(x0),_x1(x1),_y0(y0),_y1(y1),_dydx0(dydx0),_dydx1(dydx1),
...
+     _computeIntegral(computeIntegral),_intx0x1(intx0x1),_name(name)
+{
+    SmoothSegmentedFunctionParameters params(
```
adamkewley, line 379: Ideally, `_smoothData` should be initialized via a (probably multiline) init list

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -163,50 +369,56 @@ _x0(x0),_x1(x1),_y0(y0),_y1(y1),_dydx0(dydx0),_dydx1(dydx1),
...
+            SmoothSegmentedFunctionParameters(mX, mY, x0, x1, y0, y1, dydx0,
+                    dydx1, computeIntegral, intx0x1));
+
+    _smoothData = SmoothSegmentedFunctionDataLookup(params, name);
+}
+
+SmoothSegmentedFunction::SmoothSegmentedFunction()
```
adamkewley, line 386: This is invalid - the implementation will segfault because this default ctor doesn't initialize `_smoothData`. You need a nullstate version of `SmoothSegmentedFunctionParameters` (i.e. you need a default ctor for `SmoothSegmentedFunctionParameters` that mimics the behavior that was observed in the old default ctor).

OpenSim/Common/SmoothSegmentedFunction.cpp
pepbos: I think you mean `const T& v` right?

pepbos: I see `using namespace std` at the top, but also `std::string` rather than `string`. Is `size_t` special in this regard?

OpenSim/Common/SmoothSegmentedFunction.cpp
pepbos: That is a very good question...

OpenSim/Common/SmoothSegmentedFunction.cpp
adamkewley: Technically, yes, because `size_t` is a globally-defined data type from `<cstddef>`, but (if I understand correctly) it may not be defined, or differently defined, from `std::size_t`, which is the C++ standard's datatype for storing the size of an element in memory (for the given architecture, etc.)

Practically speaking, though, `static_assert(std::is_same_v<size_t, std::size_t>)` will pass in all codebases I've worked on up to now, and most codebases just use `size_t`.

As for `using namespace std;`, it means you can technically use `string` rather than `std::string`, but don't, because `using namespace std;` is bad practise (most codebases in the wild forbid it).

OpenSim/Common/SmoothSegmentedFunction.cpp
pepbos: My reasoning was, that because the matrix is hashed first, there is no seed yet. Also, it should be possible to compute the hash of just a matrix, so if it is empty I should return something. I could create a default hash in that case, but this seemed like a good alternative. But perhaps I misunderstand.

OpenSim/Common/SmoothSegmentedFunction.cpp
pepbos: I moved it to an anonumous namespace

OpenSim/Common/SmoothSegmentedFunction.cpp
pepbos: I put this in anonymous namespace as well

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +44,324 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+        const SimTK::Matrix& mX,
+        const SimTK::Matrix& mY,
+        double x0,
+        double x1,
+        double y0,
+        double y1,
+        double dydx0,
+        double dydx1,
+        bool computeIntegral,
+        bool intx0x1):
```
adamkewley, line 68: Conventionally, there's a space separator between the constructor's parameter list and the initializer list (e.g. `Ctor(T arg, T arg2) : val_(arg), val2_(arg2) {}`)

(unsure why, probably initially related to your form resembling a `goto` label if you squint)

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +44,324 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+        _y0(y0),
+        _y1(y1),
+        _dydx0(dydx0),
+        _dydx1(dydx1),
+        _computeIntegral(computeIntegral),
+        _intx0x1(intx0x1) {}
+
+    // Uninitialized parameters.
+    SmoothSegmentedFunctionParameters():
+    _mX(SimTK::Matrix()),
```
adamkewley, line 82: Should this be indented?

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +44,324 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+
+bool operator==(
+    const SmoothSegmentedFunctionParameters& lhs,
+    const SmoothSegmentedFunctionParameters& rhs)
+{
+    return
+        lhs._mX == rhs._mX &&
+        lhs._mY == rhs._mY &&
+        ((
+            lhs._x0 == rhs._x0 &&
```
adamkewley, line 130:
```c++
auto EqualsOrBothNan = [](float a, float b) { return a == b || (std::isnan(a) && std::isnan(b); };

return
    lhs._mX == rhs._mX &&
    lhs._mY == rhs._mY &&
    EqualsOrBothNan(lhs._x0, rhs._x0) &&
    etc;
```

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +44,324 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+class SmoothSegmentedFunctionDataCache final
+{
+
+public:
+    std::shared_ptr<const SmoothSegmentedFunctionData> lookup(
+            const SmoothSegmentedFunctionParameters& params,
+            const std::string& name)
+    {
+        std::lock_guard<std::mutex> guard{_cacheMutex};
+        cleanup();
```
adamkewley, line 296: Nitpick: rename to `garbageCollectExpiredData`

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -39,6 +44,324 @@ static double UTOL = (double)SimTK::Eps*1e2;
...
+                    SmoothSegmentedFunctionData(params, name)
+                )
+            }).first->second.lock();
+    }
+
+    // Do a pass-over to clean up expired pointers.
+    void cleanup()
+    {
+        for (auto it=_cache.begin(); it!=_cache.end();) {
+            if ( it->second.expired() ) {
```
adamkewley, line 322: Formatting: `if (expr)`

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -90,15 +413,27 @@ static int NUM_SAMPLE_PTS = 100;
...
+    const SimTK::Matrix& mY,
+    double x0,
+    double x1,
+    double y0,
+    double y1,
+    double dydx0,
+    double dydx1,
+    bool computeIntegral,
+    bool intx0x1,
+    const std::string& name):
```
adamkewley, line 427: As above (`Ctor(args) : list`)

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -163,55 +498,92 @@ _x0(x0),_x1(x1),_y0(y0),_y1(y1),_dydx0(dydx0),_dydx1(dydx1),
...
+                                   1               1      2     1
 ________________________________________________________________________
- 
- */
+
+*/
 
 double SmoothSegmentedFunction::calcValue(double x) const
 {
+    auto& arraySplineUX = _smoothData->_arraySplineUX;
```
adamkewley, line 563: When using references or pointers, always keep `const-` qualification in mind. This should be:

```c++
const auto& arraySplineUX = _smoothData->arraySplineUX;
```

And the typename isn't very long, so you can make it more readable in isolation by removing `auto`:

```c++
const SimTK::Array_<SimTK::Spline>& arraySplineUX = _smoothData->_arraySplineUX;
```

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -346,40 +725,49 @@ ________________________________________________________________________
 */
 double SmoothSegmentedFunction::calcIntegral(double x) const
 {
-    SimTK_ERRCHK1_ALWAYS(_computeIntegral,
+    SimTK_ERRCHK1_ALWAYS(_smoothData->_computeIntegral,
         "SmoothSegmentedFunction::calcIntegral",
         "%s: This curve was not constructed with its integral because"
         "computeIntegral was false",_name.c_str());
 
+    auto& splineYintX = _smoothData->_splineYintX;
```
adamkewley, line 733: `const`

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -420,13 +808,15 @@ void SmoothSegmentedFunction::setName(std::string &name)
 
 SimTK::Vec2 SmoothSegmentedFunction::getCurveDomain() const
 {
+    auto& mXVec = _smoothData->_mXVec;
```
adamkewley, line 811: `const`

adamkewley (commented):
2nd-pass review: mostly minor stuff

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -26,6 +26,11 @@
 #include "SmoothSegmentedFunction.h"
 #include <fstream>
 #include "simmath/internal/SplineFitter.h"
+#include <memory>
+#include <mutex>
+#include <unordered_map>
+#include <utility>
+#include <math.h>
```
adamkewley, line 33: `cmath`: `math.h` is C, `cmath` is C++ (and usually aliases `math.h`)

pepbos:
Thanks @adamkewley for your comments!
@nickbianco, @aymanhab or @carmichaelong would you be available for reviewing this?

nickbianco:
Hi @pepbos! I'm happy to review. Will leave some comments soon.

pepbos:
Great! Thanks for making time!

nickbianco (approved):
LGTM! Seems like Adam very thoroughly reviewed this already.

Cool changes. I learned some stuff about smart pointer management.

> Reviewed 1 of 2 files at r2, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3442#-:-NVGHUCJ0SncomUhoLLc:b-f80pnr)* status: all files reviewed, 35 unresolved discussions (waiting on @adamkewley and @pepbos)

nickbianco:
Once the CI passes, we can merge.

OpenSim/Common/SmoothSegmentedFunction.h
```diff
@@ -24,9 +24,15 @@
...
 #include "osimCommonDLL.h"
 #include "SegmentedQuinticBezierToolkit.h"
+#include <memory>
 
 namespace OpenSim { 
 
+    /**
+    Class containing the data used by SmoothSegmentedFuncion.
+    */
+    class SmoothSegmentedFunctionData;
```
nickbianco, line 34: Based on the CI failure for Windows, I think this forward declaration should be `struct SmoothSegmentedFunctionData`.

nickbianco (changes requested):
One small fix to get the CI to pass.

OpenSim/Common/SmoothSegmentedFunction.h
pepbos: Right! I think that would fix it, I pushed the change.

pepbos:
> LGTM! Seems like Adam very thoroughly reviewed this already.
> 
> Cool changes. I learned some stuff about smart pointer management.
> 
> > Reviewed 1 of 2 files at r2, 2 of 2 files at r6, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3442#-:-NVGHUCJ0SncomUhoLLc:b-f80pnr)_ status: all files reviewed, 35 unresolved discussions (waiting on @adamkewley and @pepbos)

Great! Thanks for reviewing! I hope the CI will pass now.

nickbianco (approved):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3442)* status: all files reviewed, 35 unresolved discussions (waiting on @adamkewley and @pepbos)

nickbianco:
Thanks @pepbos!

Issue #3299, opened 2022-10-11
Segfault when opening an `osim` with unusual socket connections


adamkewley:
Opening this modified `Rajagopal2015.osim` file:

- https://gist.github.com/adamkewley/153f60406472e14b14bde1f507d05fff

In OpenSim4.4 will trigger a segfault that causes the whole process to crash. "Whole process" can mean (e.g.) a scripting environment, GUI, or (non-forking) server process.

The file was created by:

1. Loading the original OpenSim3 `Rajagopal2015.osim` model
2. Re-saving it as an OpenSim4 model ([link](https://gist.github.com/adamkewley/ee34b2ba0adf2aec076382c6fe6ceec0))
3. Changing `/jointset/hip_r/socket_child_frame` from `femur_r_offset` to `/ground`

i.e. the diff is (between the OpenSim4 and segfaulting version):

```diff
3222c3222
<                                       <socket_child_frame>femur_r_offset</socket_child_frame>
---
>                                       <socket_child_frame>/ground</socket_child_frame>
```

Relevant parts of the backtrace:

- This line segfaults: https://github.com/opensim-org/opensim-core/blob/4.4/OpenSim/Simulation/SimbodyEngine/Body.cpp#L422
- `_slaves` looks invalid in a debugger, so the `Body` pointer has probably been mis-read
- The caller is: https://github.com/opensim-org/opensim-core/blob/4.4/OpenSim/Simulation/Model/Model.cpp#L844
- Probably whatever is looking up the body is mis-casting `Ground` as a `Body` or similar

The relevance of this bug is that it can be triggered from a file input (i.e. `osim`), and it makes the socket API potentially-segfaulting for any downstream code (e.g. scripts, OpenSim Creator's socket editor, etc.).

---

Bug initially identified in OpenSim Creator: https://github.com/ComputationalBiomechanicsLab/opensim-creator/issues/382 . I have manually confirmed that the bug also exists in OpenSim4.4 GUI by loading the above osim file in it.

tkuchida:
> Probably whatever is looking up the body is mis-casting `Ground` as a `Body` or similar

[...]

PR #3464, opened 2023-05-25
Fixed segfault when building models with unusual graph topologies (#3299)

adamkewley:
Fixes issue #3299 (well, at least downgrades it so it won't fully crash the process if you catch the exception)

This PR downgrades the segfault found in #3299 to a runtime exception. It does not fundamentally fix the source of the error (a bad graph traversal), but at least ensures that, when it does happen, the implementation will throw an exception message that explains possible fixes (i.e. "have a less-strange graph topology") and directs them to #3299 for more information.

A later PR could come in to actually fix the underlying bug. Based on a quick look, I *believe* the bug is *something like* that the OpenSim implementation is assuming that the outbound body of a graph edge *must* be a body but the weird topology means that the outbound body ends up being `OpenSim::Ground`, because `Ground` is also added to the graph maker, and trying to cast a `Ground` to a `Body` is an error.

### Brief summary of changes

- Added a test that contains the error reproduction code in #3299
- Added `MultibodyGraphMakerPtrCast`:
  - Functionally, it performs the same task as the C-style casts and `static_cast`s in the original code
  - But it does so via a `dynamic_cast`, which enables the implementation to perform some runtime checking on the cast (assuming that the initial `static_cast` to a `Component` works)
  - And, rather than throwing a generic runtime cast error (`std::bad_cast`), which might seem a bit ambiguous for users who can't attach a debugger etc., I instead throw a custom exception with an explanation of what the problem *probably* is with a link to #3299
- Changed all C-style and `static_cast`s in the original code to use `MultibodyGraphMakerPtrCast`

### Testing I've completed

- Pasted error reproduction code from #3299 into a new unit test
- The test failed
- Made codebase change
- The test passes (where "throws an exception" is also a pass, for now)

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3464)


adamkewley:
This PR is ready for review if anyone is available

nickbianco (approved):
These changes LGTM.

I'm wondering if would it make sense to explicitly check if any outboard bodies are `Ground` , since we know that shouldn't be possible (`Ground` can only be the root "body"). I haven't thought through all the logic in here though. (This could be explored in a later PR.)

adamkewley:
> These changes LGTM.
> 
> I'm wondering if would it make sense to explicitly check if any outboard bodies are `Ground` , since we know that shouldn't be possible (`Ground` can only be the root "body"). I haven't thought through all the logic in here though. (This could be explored in a later PR.)

I wasn't sure exactly how the `SimTK::MultibodyGraphMaker` was working, and it's an upstream change (to simbody), and this code is usually only ran once at model start (so it isn't in the hot path), but yeah - I agree that, ultimately, the solution is to figure out where the problem's happening in this process (this is more of a bandaid patch to ensure it doesn't crash applications as easily) :(

adamkewley:
Merging - the Windows failure is new (ced222a was passing, 9b45da9 is a merge commit with no changes) and appears to be external

Thanks again @nickbianco ! ❤️

nickbianco:
> this is more of a bandaid patch to ensure it doesn't crash applications as easily

A very useful bandaid! We can get to the bottom of this later.

> Merging - the Windows failure is new (https://github.com/opensim-org/opensim-core/commit/ced222a073e6ca329182ebf7145a51cb65af3d5e was passing, https://github.com/opensim-org/opensim-core/commit/9b45da9d4b240df12456b362897bca804fd81092 is a merge commit with no changes) and appears to be external

Sounds good, merge when ready.

Issue #3465, opened 2023-05-25
Fix `WrapObject`s (e.g. `WrapCylinder`) will segfault if used outside of a `GeometryPath`


adamkewley:
If I create a `WrapObject` and add it to a model using `addComponent`, followed by performing an operation that potentially touches the wrap object (e.g. scaling), then my application will always segfault:

```c++
OpenSim::Model model;
model.addComponent(new OpenSim::WrapCylinder{});
model.buildSystem();
SimTK::State& state = model.initializeState();

model.scale(state, OpenSim::ScaleSet{}, true);  // segfault
```

The reason why this happens is because `WrapObject` contains a `_frame` member, which is assumed to be set by a component that is higher up in the hierarchy. Specifically, `PhysicalFrame`s own a `WrapObjectSet`. During `OpenSim::PhysicalFrame::extendConnectToModel`, they call call `WrapObject::setFrame` and everything works. However, if `_frame` isn't set, you get segfaults.

This means that if you do something like add `WrapObject` independently of a `PhysicalFrame` (e.g. if you work on a model editor like [opensim-creator](https://github.com/ComputationalBiomechanicsLab/opensim-creator) and try to do something generic like allow the user to add anything they want into a model) then runtime segfaults are possible.

PR #3466, opened 2023-05-25
Downgraded segfault when using WrapObject outside a PhysicalFrame (#3…

adamkewley:
Fixes issue #3465 

### Brief summary of changes

- Adds a unit test that exercises #3465
- Adds a runtime check to `OpenSim::WrapObject::getFrame`
- The runtime check throws an exception if the caller tries to call `getFrame` before `OpenSim::WrapObject::_frame` has been set
- That error condition can occur when using a `WrapObject` outside of a `PhysicalFrame`, or when using it within a `PhysicalFrame` before `finalizeConnections` is called

### Testing I've completed

- The reproduction code in #3465 was copied into a unit test
- The unit test failed because of the described bug
- The codebase was modified with the additional runtime check
- The unit test then passed

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3466)

### DIFF
--- a/OpenSim/Simulation/Wrap/WrapObject.cpp
+++ b/OpenSim/Simulation/Wrap/WrapObject.cpp
@@ -75,7 +75,7 @@ void WrapObject::constructProperties()
 const PhysicalFrame& WrapObject::getFrame() const
 {
     if (!_frame) {
-        OPENSIM_THROW_FRMOBJ(OpenSim::Exception, "Tried to call WrapObject::getFrame before the frame has been set. This can happen if the WrapObject is part of the WrapObjectSet of a PhysicalFrame (see: opensim-core issue #3465)");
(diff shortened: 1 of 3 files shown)


OpenSim/Simulation/Wrap/WrapObject.cpp
```diff
@@ -74,6 +74,9 @@ void WrapObject::constructProperties()
 
 const PhysicalFrame& WrapObject::getFrame() const
 {
+    if (!_frame) {
+        OPENSIM_THROW_FRMOBJ(OpenSim::Exception, "Tried to call WrapObject::getFrame before the frame has been set. This can happen if the WrapObject is part of the WrapObjectSet of a PhysicalFrame (see: opensim-core issue #3465)");
```
tkuchida, line 78: "This can happen if the WrapObject is part of the WrapObjectSet of a PhysicalFrame" might not be as helpful for the user as stating what they should do to fix the problem (e.g., "Make sure xyz method is called to set the frame before doing xyz operation.").

tkuchida (approved):
LGTM :+1: (one ignorable suggestion for the user-facing message)

Issue #3083, opened 2021-09-20
Feature request: getAnalogsTable()


mrrezaie:
Dear Experts,

Up to now C3DFileAdapter doesn't have any function to retrieve a TimeSeriesTable of analog data. It would be great if this operation added to API in order to extract EMG data in OpenSim format.

Regards,
Mohammadreza

mrrezaie:
Hi, 
I'm just curious to know whether you have any plan to add this functionality to your future release. Your current C3D parser (EZC3D) is able to extract these data and it would be great if you could define relevant functions in your API (retrieve the TimeSeriesTable of Analogs), the same as `getForcesTable()` and `getMarkersTable()`.

Issue #2575, opened 2019-09-26
C3DFileAdapter exports only markers and external forces


mitkof6:
c3d files contain a rich set of information, such as emg from analog sources. Currently, the C3DFileAdapter provides access only to the marker and external force measurements. As more and more people will start using this functionality, I recommend that at least we provide access either to the underlying btk object or export all the information into OpenSim's table objects.

jimmyDunne:
We would love to support more data types but I don't think we will be able to address it until we decide when to remove BTK for a different C3D reading package. Will keep in the list.

Issue #3460, opened 2023-05-22
Fix out-of-bounds memory access in `testAssemblySolver`


adamkewley:
When running the test suites with libASAN, `testAssemblySolver` fails, because it involves an out-of-bounds memory write in `Component::setStateVariableValues`.

The out-of-bounds behavior is because the supplied vector has a different size from the state vector. The source code assumes that the incoming vector is bounds-checked by `SimTK_ASSERT`, but assertions are default-disabled in release builds.

PR #3467, opened 2023-05-25
Support analog data read by ezc3d

aymanhab:
Fixes issue #3083 #2575

### Brief summary of changes
Beef up C3DFileAdapter to return a table of AnalogData as TimeSeriesTable

### Testing I've completed
Visually inspected analogData file and data looked to match the grf file. Also tested by @nicos1993 
### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3467)


aymanhab:
@nicos1993 This seems to be workig now for the sample files you sent me. Would be good to test the functionality manually while I handle the corner cases and beef up the test case. Ideally you have a small .c3d file with actual data that we can put under source control for the test. Thanks

nicos1993:
@aymanhab Hello, I did a little further testing. It seems to be working as anticipated (pulling **ALL** analogue signals), users will have to be careful to ignore the ground reaction force data extracted from `getAnalogDataTable` as it does not undergo the conversions necessary for usage - as is performed in `getForcesTable`

I created an analog_table `std::shared_ptr<TimeSeriesTable> analog_table =
            c3dFileAdapter.getAnalogDataTable(tables);`
Extracted column labels `std::vector<std::string> labels1 = analog_table->getColumnLabels();`
Wrote the analog_table using STOFileAdapter `STOFileAdapter sto_adapter_1{};`
`sto_adapter_1.write(*analog_table, "test.sto");`

In summary, the test.sto contains all the analog data as expected and I should be able to operate on it easily!

Nicos

aymanhab (commented):
Thanks for the feedback @carmichaelong, I addressed your comments but please let me know if you think more clarification is needed.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3467#-:-NWy6UISBR6vj6sKxjEG:b-u6rb4e)* status: 6 of 7 files reviewed, 3 unresolved discussions (waiting on @carmichaelong)

___
*[`OpenSim/Common/C3DFileAdapter.cpp` line 349 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3467#-NWxux8VCqNttt5-vVfJ:-NWy4B1-CbS6r41gQLAx:bxvkugl) ([raw file](https://github.com/opensim-org/opensim-core/blob/588690a3c23a986a2f7986b4127bb6bbe7762a2d/OpenSim/Common/C3DFileAdapter.cpp#L349)):*

can you explain this TODO more? is it that this assumes that the analog data always starts at time = 0? do c3d files generally allow for files that start at a non-zero time?
</blockquote></details>

This follows the exact same TODO for forces, basically there's no indication in the file regarding start time, only sampling frequency so both forces and analog data assume start at 0
___
*[`OpenSim/Common/Test/testC3DFileAdapter.cpp` line 108 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3467#-NWxtEaC7gcfAIr6yI3d:-NWy4dDB9nMwwROL3rRw:bwl2gjo) ([raw file](https://github.com/opensim-org/opensim-core/blob/588690a3c23a986a2f7986b4127bb6bbe7762a2d/OpenSim/Common/Test/testC3DFileAdapter.cpp#L108)):*

it's not clear to me why a different frame would need more downsampling. are these two separate comments? also looks like from the test lines below that the analog table is tested against the analog file (and not the forces table or file)
</blockquote></details>

These test .c3d files have only forces data, and no other analog data. So the test for contents is already done. I visually compared the numbers in the few rows of the highly downsampled analog vs forces files and they checked. There's no need to add another duplicate analog data file to the repo. 
___
*[`OpenSim/Common/Test/testStorage.cpp` line 188 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3467#-NWxtbUC-m8FCH1a9r2t:-NWy5P88AWmrWQ074s6v:bdzhnae) ([raw file](https://github.com/opensim-org/opensim-core/blob/a67074f3e87a96c30c2480202ce7a21a3e72cb74/OpenSim/Common/Test/testStorage.cpp#L188)):*

it's unclear to me why this test was removed. no need to add a comment into the code, responding to this comment is sufficient!
</blockquote></details>

This test was constructing a storage from c3d file and comparing against a specific table even though multiple tables are returned. With analogData table added, a different table is now returned and so the test failed. This case is just a bad test since we should never promise to randomly pick a specific table when c3d contains multiple tables. I don't think users should count on this behavior.

carmichaelong (changes requested):
thanks @aymanhab the clarifications were all super helpful. i left one comment that could be ignored depending on the bang for buck of adding a test to check for just a few values

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3467#-:-NWyNCvY9f7yi6zhP5oC:b-2cg870)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on @aymanhab)

___
*[`OpenSim/Common/Test/testC3DFileAdapter.cpp` line 108 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3467#-NWxtEaC7gcfAIr6yI3d:-NWyMU2j4yGIIFBsBSbs:bsvv74y) ([raw file](https://github.com/opensim-org/opensim-core/blob/588690a3c23a986a2f7986b4127bb6bbe7762a2d/OpenSim/Common/Test/testC3DFileAdapter.cpp#L108)):*

These test .c3d files have only forces data, and no other analog data. So the test for contents is already done. I visually compared the numbers in the few rows of the highly downsampled analog vs forces files and they checked. There's no need to add another duplicate analog data file to the repo. 
</blockquote></details>

thanks that's helpful for context, and i fully agree that we shouldn't add another duplicate file to the repo.

seems like it could still be helpful if there were even a few numbers that could be checked between the analog and forces files (e.g., like you did visually, and could be as simple as a few values in the first few rows). i know this might be tricky with all the downsampling, so this can be ignored.

aymanhab:
@carmichaelong for your review

carmichaelong (approved):
thanks @aymanhab, just a couple typos in the changelog and comments. feel free to fix those without CI and merge

> Reviewed 1 of 1 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3467#-:-NXJX3Kb1elpQDbeqwRL:bgn47dk)* status: all files reviewed (commit messages unreviewed), 3 unresolved discussions (waiting on @aymanhab)

___
*[`CHANGELOG.md` line 34 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3467#-NXJWhik-U317554irLa:-NXJWhil5_j-yLdJTMaR:b-mys0gu) ([raw file](https://github.com/opensim-org/opensim-core/blob/00fd271cd59e1f6d77004bbecca6d29cd5ea1ac8/CHANGELOG.md#L34)):*
> ```Markdown
> - Fixed out-of-bounds memory access in testAssemblySolver (#3460)
> - Fixed runtime segfault that can occur when trying to use a `WrapObject` that is not a child of a `PhysicalFrame` (#3465)
> - Fixed issues #3083 #2575 where analog data is not pulled out from c3d files, a a new function getAnalogDataTable() has been added to the C3DFileAdapter
> ```

there still seems to be a typo in here "*a a* new function..."
___
*[`OpenSim/Common/Test/testC3DFileAdapter.cpp` line 205 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3467#-NXJWv201Ou-8eyiATkw:-NXJWv201Ou-8eyiATkx:b-pgigfq) ([raw file](https://github.com/opensim-org/opensim-core/blob/00fd271cd59e1f6d77004bbecca6d29cd5ea1ac8/OpenSim/Common/Test/testC3DFileAdapter.cpp#L205)):*
> ```cpp
>     cout << "\t" << analogs_file << " is equivalent to its standard." << endl;
> 
>     // For the test files, analog data contains force vectors with some rogod body transform/signs changed
> ```

`rogod` -> `rigid`

aymanhab:
Fixed typos, thanks @carmichaelong and @nicos1993

pariterre:
@aymanhab 
Glad to see it moved quickly forward! May I suggest to update ezc3d to 1.5.4 in the dependencies. There should not be any change that affects you in the API, I however had a bug in the Matlab binder for the 1.5.0 which causes problem for those who have OpenSim installed (and try to use ezc3d on Matlab) :)

aymanhab:
DEfinitely, and thanks for all your help and for making it so easy to implement on our side. I'll open an issue to put in our next sprint in a couple weeks.

Issue #3468, opened 2023-05-26
Fix OpenSim's property macros are namespace-dependent


adamkewley:
If I write a custom component class that lies outside of the `OpenSim` namespace:

```c++
namespace
{
  class FrameDefinitionMesh final : public OpenSim::Component {
      OpenSim_DECLARE_CONCRETE_OBJECT(FrameDefinitionMesh, OpenSim::Component);
  public:
      OpenSim_DECLARE_PROPERTY(scale_factors, SimTK::Vec3, "scale factors in X, Y, and Z directions of the mesh");
  };
}
```

Then it will fail to compile unless the anonymous namespace is named `OpenSim`, or a `using namespace OpenSim;` declaration is placed above the class declaration. Example compiler errors:

```txt
FrameDefinitionTab.cpp(58): error C3646: 'PropertyIndex_scale_factors': unknown override specifier
FrameDefinitionTab.cpp(58): error C4430: missing type specifier - int assumed. Note: C++ does not support default-int
FrameDefinitionTab.cpp(58): error C2143: syntax error: missing ';' before '<'
FrameDefinitionTab.cpp(58): error C2334: unexpected token(s) preceding '{'; skipping apparent function body
FrameDefinitionTab.cpp(58): error C2327: '`anonymous-namespace'::FrameDefinitionMesh::Property': is not a type name, static, or enumerator
FrameDefinitionTab.cpp(56): error C2280: '`anonymous-namespace'::FrameDefinitionMesh &`anonymous-namespace'::FrameDefinitionMesh::operator =(const `anonymous-namespace'::FrameDefinitionMesh &)': attempting to reference a deleted function
  FrameDefinitionTab.cpp(59): note: compiler has generated '`anonymous-namespace'::FrameDefinitionMesh::operator =' here
  FrameDefinitionTab.cpp(59): note: '`anonymous-namespace'::FrameDefinitionMesh &`anonymous-namespace'::FrameDefinitionMesh::operator =(const `anonymous-namespace'::FrameDefinitionMesh &)': function was implicitly deleted because '`anonymous-namespace'::FrameDefinitionMesh' has a data member '`anonymous-namespace'::FrameDefinitionMesh::Property' of const-qualified non-class type
  FrameDefinitionTab.cpp(58): note: see declaration of '`anonymous-namespace'::FrameDefinitionMesh::Property'
FrameDefinitionTab.cpp(58): error C2065: 'PropertyIndex_scale_factors': undeclared identifier
[...]

PR #3469, opened 2023-05-26
Make OpenSim macros namespace-independent (#3468)

adamkewley:
Fixes issue #3468 

### Brief summary of changes

- Moves OpenSim components macros (e.g. `OpenSim_DECLARE_PROPERTY`) outside of the `OpenSim` namespace
- Fully qualifies all types used within those macros (e.g. `PropertyIndex` --> `OpenSim::PropertyIndex`)
- Adds compile-time test via the `ComponentInAnonymousNamespace` in `testComponentInterface.cpp`
  - It is an example of how downstream code may want to define OpenSim components in their own namespace (here: an anonymous one) without having to also use `using namespace OpenSim;` before it

### Testing I've completed

- `ComponentInAnonymousNamespace` was added to `testComponentInterface.cpp`
- It did not compile (for reasons outlined in #3468)
- The macros were moved outside of the `OpenSim` namespace and changed to use fully-qualified type names
- `testComponentInterface.cpp` then compiles
- And all existing code works exactly the same

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3469)


tkuchida (approved):
:+1: LGTM

PR #3476, opened 2023-06-06
testMocoInverse: recompute inverse solution

pepbos:
This pr changes the *Test IMUDataReporter for gait* test case in `testMocoInverse.cpp` to recompute the inverse solution instead of reading it from file.

If I understand correctly, the purpose of this test is to compare the imu outputs obtained from the inverse solution against the acceleration obtained from a motion file. Since the inverse solution is not computed (but read from a file), it will not pick up on changes in the inverting-algorithm.

I assume that this test should pick up on changes in the algorithm, such that it fails when needed, but also passes when needed (see #3463).

### Testing I've completed

The test passes, and signals were visually inspected (they look the same).

Perhaps I misinterpreted this test?

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3476)

### DIFF
--- a/CHANGELOG_MOCO.md
+++ b/CHANGELOG_MOCO.md
@@ -3,6 +3,9 @@ Moco Change Log
 
 1.2.1
 -----
+- 2023-06-06: Instead of reading from file, changed to recompute the inverse
+              solution in `testMocoInverse` before testing imu-outputs.
+
 - 2023-03-21: Fixed a bug where failing `MocoProblem`s with path constraints returned
               a zero time vector.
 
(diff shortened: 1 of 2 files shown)


nickbianco:
Hi @pepbos! This fix should work, but the main reason we loaded that previous solution from file was to avoid running duplicate optimization problems, which can slow down the CI. I think the best approach here would be to [leverage Catch2's ability to make a custom main() function](https://github.com/catchorg/Catch2/blob/devel/docs/own-main.md), which would allow us to run the MocoInverse problem once and then use the result for all subsequent tests.

I'd be happy to do this myself, but I'm traveling starting tomorrow and won't be available until Monday. I'd you prefer to focus on the wrapping PR for now, you can just implement a temporary fix by updating `std_testMocoInverse_subject_18musc_solution.sto` on that branch, and we can come back and implement this (better) fix later.

pepbos:
tHi @nickbianco, that makes sense. I've updated the sto file in #3463 instead, as you suggested. Thanks!

PR #3486, opened 2023-06-16
Upgrade to SWIG 4.1.1 on CI builds

nickbianco:
### Brief summary of changes
The CI builds on Windows have been failing recently because the builds can't find the SWIG .exe or libraries consistently. After trying a bunch of different fixes, it seems like the simplest fix is to upgrade to SWIG 4.1.1 (we were currently trying to install 4.0.2). While this fix is for the Windows builds, I've updated all builds to use 4.1.1 for consistency. 

### CHANGELOG.md (choose one)

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3486)

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -40,7 +40,7 @@ jobs:
       run: python3 -m pip install numpy==1.20.2
 
     - name: Install SWIG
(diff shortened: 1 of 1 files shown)


aymanhab (approved):
LGTM, thanks @nickbianco 
Separately will build the GUI using this to see if there're any diffs in resulting java files or unexpected behavior in the application

Issue #3474, opened 2023-06-01
Add `get` methods and `Output`s  to compute whole-body momentum from `Model`


nickbianco:
It is possible to calculate whole-body linear and angular momentum from a `Model` by accessing the underlying `SimbodyMatterSubsystem`:

```
state = model.initSystem();
matter = model.getMatterSubsystem();
model.realizeVelocity(state);
momentum = matter.calcSystemCentralMomentum(state);
```

To make these quantities easier to find for users, we should provide methods that make them directly accessible through `Model`. Adding an `Output` would also make it possible to directly optimize momentum quantities via `MocoOutputGoal`.

PR #3487, opened 2023-06-16
Add `calc` methods and `Output`s for whole-body momentum to `Model`

nickbianco:
Fixes issue #3474

### Brief summary of changes
This PR introduces methods and `Output`s for computing whole-body momentum (about the center of mass, expressed in ground) on `Model`. 

### Testing I've completed
Successfuly ran an optimization problem using `MocoOutputGoal` with the new momentum `Output`s.

I noticed that `calcMassCenterPosition`, `calcMassCenterVelocity`, and `calcMassCenterAcceleration` all call `realize` within each method. This might lead to `realize` being called twice when using `Output`s, since `Output` require a `SimTK::Stage` dependency. I _think_ that this means that `realize` will automatically be called for these `Output`s, or at least an "invalid stage" error will be thrown. Either way, I don't think the `realize` calls should be baked into these methods, so I'm inclined to remove them, which I could append to this PR.

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3487)

### DIFF
--- a/OpenSim/Simulation/Model/Model.cpp
+++ b/OpenSim/Simulation/Model/Model.cpp
@@ -2235,14 +2235,14 @@ SimTK::SpatialVec Model::calcMomentum(const SimTK::State &s) const
  *
  */
 SimTK::Vec3 Model::calcAngularMomentum(const SimTK::State& s) const {
-    return getMatterSubsystem().calcSystemCentralMomentum(s).get(0);
+    return calcMomentum(s).get(0);
 }
 /**
  * Return the linear momentum expressed in Ground.
  *
  */
 SimTK::Vec3 Model::calcLinearMomentum(const SimTK::State& s) const {
-    return getMatterSubsystem().calcSystemCentralMomentum(s).get(1);
+    return calcMomentum(s).get(1);
 }
 
 
--- a/OpenSim/Simulation/Model/Model.cpp
+++ b/OpenSim/Simulation/Model/Model.cpp
@@ -2167,87 +2167,55 @@ void Model::realizeReport(const SimTK::State& state) const
     getSystem().realize(state, Stage::Report);
 }
 
-
-/**
- * Compute the derivatives of the generalized coordinates and speeds.
- */
+//------------------------------------------------------------------------------
+// Subsystem computations
+//------------------------------------------------------------------------------
 void Model::computeStateVariableDerivatives(const SimTK::State &s) const
 {
     realizeAcceleration(s);
 }
 
-/**
- * Get the total mass of the model
- *
- * @return the mass of the model
- */
 double Model::getTotalMass(const SimTK::State &s) const
 {
     return getMatterSubsystem().calcSystemMass(s);
 }
-/**
- * getInertiaAboutMassCenter
- *
- */
+
 SimTK::Inertia Model::getInertiaAboutMassCenter(const SimTK::State &s) const
 {
-    SimTK::Inertia inertia = getMatterSubsystem().calcSystemCentralInertiaInGround(s);
+    SimTK::Inertia inertia =
+            getMatterSubsystem().calcSystemCentralInertiaInGround(s);
 
     return inertia;
 }
-/**
- * Return the position vector of the system mass center, measured from the Ground origin, and expressed in Ground.
- *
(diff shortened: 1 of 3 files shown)


nickbianco:
The failiing Windows builds should be fixed by #3486.

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -2222,6 +2222,29 @@ SimTK::Vec3 Model::calcMassCenterAcceleration(const SimTK::State &s) const
...
+SimTK::SpatialVec Model::calcMomentum(const SimTK::State &s) const
+{
+    return getMatterSubsystem().calcSystemCentralMomentum(s);
+}
+/**
+ * Return the angular momentum about the system mass center expressed in Ground.
+ *
+ */
+SimTK::Vec3 Model::calcAngularMomentum(const SimTK::State& s) const {
+    return getMatterSubsystem().calcSystemCentralMomentum(s).get(0);
```
halleysfifthinc, line 2238: Mostly a style suggestion, but may make future maintenance simpler. (and same for `calcLinearMomentum` if accepted)

```suggestion
    return calcMomentum(s).get(0);
```

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -2220,8 +2220,31 @@ SimTK::Vec3 Model::calcMassCenterVelocity(const SimTK::State &s) const
 SimTK::Vec3 Model::calcMassCenterAcceleration(const SimTK::State &s) const
 {
     getMultibodySystem().realize(s, Stage::Acceleration);
-    return getMatterSubsystem().calcSystemMassCenterAccelerationInGround(s);    
+    return getMatterSubsystem().calcSystemMassCenterAccelerationInGround(s);
 }
+/**
```
tkuchida, line 2225: These Doxygen-style comments don't get picked up from the .cpp file (only the headers). See, for example, [the (blank) documentation for `Model::calcMassCenterPosition()`](https://simtk.org/api_docs/opensim/api_docs/classOpenSim_1_1Model.html#a937ccd739df45c80c48a308a90aaf4cd) (notably, the potentially useful comment "Return the position vector of the system mass center, measured from the Ground origin, and expressed in Ground." does not appear). Perhaps at least put the Doxygen comments for your new methods in the header file, and create an Issue to move the others.

OpenSim/Simulation/Model/Model.cpp
```diff
@@ -2220,8 +2220,31 @@ SimTK::Vec3 Model::calcMassCenterVelocity(const SimTK::State &s) const
...
+ * Return the spatial momentum about the system mass center expressed in Ground.
+ *
+ */
+SimTK::SpatialVec Model::calcMomentum(const SimTK::State &s) const
+{
+    return getMatterSubsystem().calcSystemCentralMomentum(s);
```
tkuchida, line 2231: Following the examples above (e.g., `Model::calcMassCenterAcceleration()`), shouldn't this line be preceded by
``` cpp
getMultibodySystem().realize(s, Stage::Velocity);
```

tkuchida: (I think the `realize()` calls should be here since an API user might call the method directly. IIRC, calling `realize()` isn't [supposed to be] expensive if the State has already been realized to the required Stage.)

OpenSim/Simulation/Model/Model.cpp
nickbianco: Good catch! I can move all these over.

OpenSim/Simulation/Model/Model.cpp
nickbianco: I thought it was weird these calls were inside the calc-COM methods, but I think you're right, the `Stage` is probably cached and `realize` won't re-realize if it is. (I also tried taking out the realize calls in the calc-COM methods and it broke some tests, so there's that.)

nickbianco:
Thanks @tkuchida! I made some updates based on your comments.

tkuchida (approved):
LGTM :+1:

Issue #3488, opened 2023-06-20
Update to support swig 4.1.1


AlbertoCasasOrtiz:
Documentation, build scripts...

PR #3489, opened 2023-06-21
Updated linux build script.

AlbertoCasasOrtiz:
Fixes issue #3488 

### Brief summary of changes

Added swig 4.1.1 instead of old version 4.0.2.

### Testing I've completed

- [X] Run script in Ubuntu 22.04 and execute.
- [x] Run script in Ubuntu 20.04 and execute.
- [x] Run script in Ubuntu 18.04 and execute.
- [x] Run script in Debian 10 and execute.
- [x] Run script in Debian 11 and execute.
- [x] Run script in MacOS 11 and execute.
- [x] Run script in MacOS 12 and execute.
- [x] Run script in MacOS 13 and execute.
- [x] Run script in Windows 10 and execute.
- [x] Run script in Windows 11 and execute.

I'll may need help with testing in MacOS.

### CHANGELOG.md (choose one)

- no need to update because this is just an update to build scripts and documentation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3489)

### DIFF
--- a/scripts/build/opensim-core-linux-build-script.sh
+++ b/scripts/build/opensim-core-linux-build-script.sh
@@ -80,7 +80,7 @@ echo
 
 # Install dependencies from package manager.
 echo "LOG: INSTALLING DEPENDENCIES..."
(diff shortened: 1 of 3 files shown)


AlbertoCasasOrtiz:
Current problem is that python cannot be found in windows in the script. This is because when installing using chocolatey, the environment of powershell is not refreshed (and that is why some tools like git have to be called using the full path).

This looks like a good solution that would allow to detect python, and get rid of full paths when calling commands that you just installed: https://stackoverflow.com/a/46760714/12928292

adamkewley:
Seems reasonable enough :+1:

AlbertoCasasOrtiz:
I have finally tested in every of the listed OS. For Mac, I tested using GitHub actions as Nick suggested.

PR #3498, opened 2023-06-26
Wrap Cylinder Test

pepbos:
This PR contains a new test for the cylinder wrapping algorithm.

It serves as a starting point for a more rigorous testing of the cylinder wrapping, but currently it contains two simple test cases: One in which the start and end points are inside the cylinder (no-wrapping), and the other where the points are perpendicular to the cylinder (have no axial components) resulting in a simple wrapping:

testWrapCylinder-perpendicular-unconstrained-quarter

The resulting wrapped path points obtained from the simulation are compared against the expected answer, given a tolerance. The expected answer was obtained from the solution as is now. The tolerance was chosen arbitrarily, and might change as more tests are added.

The next improvements I planned to include (in different PRs) are:
- [ ] Assert wrapping direction.
- [ ] Assert geodesic path properties, such as the path length, and gradient.
- [ ] Add more test cases,
- [ ] Add permutations to the test cases, such as rotating the scene, or reversing the start and end points.
- [ ] Merging with the cylinder wrapping test currently in `testWrappingAlgorithm.cpp`.

### Brief summary of changes

A basic test for cylinder wrapping is added: `testWrapCylinder.cpp`

More ideas on what to test for are welcome!

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3498)

### DIFF
--- a/OpenSim/Tests/Wrapping/testWrapCylinder.cpp
+++ b/OpenSim/Tests/Wrapping/testWrapCylinder.cpp
@@ -43,8 +43,8 @@ namespace {
             end(endPoint)
         {}
 
-        SimTK::Vec3 start {0., 0., 0.};
-        SimTK::Vec3 end {0., 0., 0.};
+        SimTK::Vec3 start {SimTK::NaN};
+        SimTK::Vec3 end {SimTK::NaN};
     };
 
     std::ostream& operator<<(
--- a/OpenSim/Tests/Wrapping/testWrapCylinder.cpp
+++ b/OpenSim/Tests/Wrapping/testWrapCylinder.cpp
@@ -96,7 +96,7 @@ namespace {
     struct WrappingTolerances final {
         WrappingTolerances() = default;
 
-        WrappingTolerances(double eps):
+        WrappingTolerances(double eps) :
             position(eps),
             length(eps)
         {}
--- a/OpenSim/Tests/Wrapping/testWrapCylinder.cpp
+++ b/OpenSim/Tests/Wrapping/testWrapCylinder.cpp
@@ -226,7 +226,6 @@ namespace {
         const SimTK::State& state = model.initSystem();
 
         if (visualize) {
-            model.realizeVelocity(state);
             model.getVisualizer().show(state);
         }
 
--- a/OpenSim/Tests/Wrapping/testWrapCylinder.cpp
+++ b/OpenSim/Tests/Wrapping/testWrapCylinder.cpp
@@ -105,7 +105,7 @@ namespace {
         double length = 1e-9;
     };
 
-    bool IsEqual(
+    bool IsEqualWithinTolerance(
         double lhs,
         double rhs,
         double tolerance)
@@ -113,26 +113,26 @@ namespace {
         return std::abs(lhs - rhs) < tolerance;
     }
 
-    bool IsEqual(
+    bool IsEqualWithinTolerance(
         const SimTK::Vec3& lhs,
         const SimTK::Vec3& rhs,
         double tolerance)
     {
-        return IsEqual(lhs[0], rhs[0], tolerance)
-            && IsEqual(lhs[1], rhs[1], tolerance)
-            && IsEqual(lhs[2], rhs[2], tolerance);
+        return IsEqualWithinTolerance(lhs[0], rhs[0], tolerance)
+            && IsEqualWithinTolerance(lhs[1], rhs[1], tolerance)
+            && IsEqualWithinTolerance(lhs[2], rhs[2], tolerance);
     }
 
-    bool IsEqual(
+    bool IsEqualWithinTolerance(
         const PathSegment& lhs,
         const PathSegment& rhs,
         double tolerance)
     {
-        return IsEqual(lhs.start, rhs.start, tolerance)
-            && IsEqual(lhs.end, rhs.end, tolerance);
+        return IsEqualWithinTolerance(lhs.start, rhs.start, tolerance)
+            && IsEqualWithinTolerance(lhs.end, rhs.end, tolerance);
     }
 
-    bool IsEqual(
+    bool IsEqualWithinTolerance(
         const WrapTestResult& lhs,
         const WrapTestResult& rhs,
         const WrappingTolerances& tolerance)
@@ -140,8 +140,8 @@ namespace {
         if (lhs.noWrap && rhs.noWrap) {
             return true;
         }
-        return IsEqual(lhs.path, rhs.path, tolerance.position)
-            && IsEqual(lhs.length, rhs.length, tolerance.length)
+        return IsEqualWithinTolerance(lhs.path, rhs.path, tolerance.position)
+            && IsEqualWithinTolerance(lhs.length, rhs.length, tolerance.length)
             && lhs.noWrap == rhs.noWrap;
     }
 
@@ -255,7 +255,7 @@ namespace {
 
         WrapTestResult result = solve(input, visualize);
 
-        if (!IsEqual(result, expected, tol.position)) {
+        if (!IsEqualWithinTolerance(result, expected, tol.position)) {
             std::ostringstream oss;
             oss << "\nFAILED: case = " << testCase;
             oss << "\n    input = " << input;
--- a/OpenSim/Tests/Wrapping/testWrapCylinder.cpp
+++ b/OpenSim/Tests/Wrapping/testWrapCylinder.cpp
@@ -189,27 +189,36 @@ namespace {
         Model model;
         model.setName("testWrapCylinderModel");
 
-        std::unique_ptr<WrapCylinder> cylinder(new WrapCylinder());
-        cylinder->setName("cylinder");
-        cylinder->set_radius(input.radius);
-        cylinder->set_length(input.cylinderLength);
-        cylinder->set_quadrant(input.quadrant);
-        cylinder->set_xyz_body_rotation(input.eulerRotations);
-        cylinder->setFrame(model.getGround());
+        // Add a spring to create the wrapping path.
+        {
+            std::unique_ptr<PathSpring> spring (
+                new PathSpring("spring", 1., 1., 1.));
+            spring->updGeometryPath().appendNewPathPoint(
+                "startPoint",
+                model.get_ground(),
+                input.path.start);
+            spring->updGeometryPath().appendNewPathPoint(
+                "endPoint",
+                model.get_ground(),
+                input.path.end);
 
-        std::unique_ptr<PathSpring> spring (new PathSpring("spring", 1., 1., 1.));
-        spring->updGeometryPath().appendNewPathPoint(
-            "startPoint",
-            model.get_ground(),
-            input.path.start);
-        spring->updGeometryPath().appendNewPathPoint(
-            "endPoint",
-            model.get_ground(),
-            input.path.end);
-        spring->updGeometryPath().addPathWrap(*cylinder.get());
+            model.addComponent(spring.release());
+        }
 
-        model.addComponent(spring.release());
-        model.updGround().addWrapObject(cylinder.release());
+        // Add the cylinder as the wrapping surface for the spring.
+        {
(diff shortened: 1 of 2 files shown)


pepbos:
Hi! @adamkewley @nickbianco @aymanhab , I thought I would start with a simple test, and expand from that. Let me know what you think!

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -0,0 +1,314 @@
...
+#include <sstream>
+#include <functional>
+#include <memory>
+
+using namespace OpenSim;
+
+namespace {
+
+    // A path segment determined in terms of the start and end point.
+    struct PathSegment final {
```
adamkewley, line 38: Initialize these the members to known values.

Simbody has an "interesting" behavior, which is that the empty constructor (e.g. `SimTK::Vec3 v;`, `()`, or `{}`) will either produce NaNs, or uninitialized garbage, based on compiler flags. See:

https://github.com/simbody/simbody/blob/master/SimTKcommon/SmallMatrix/include/SimTKcommon/internal/Vec.h#L425C5-L425C5

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -0,0 +1,314 @@
...
+            "WrapTestResult{" <<
+            "path: " << result.path << ", " <<
+            "length: " << result.length << "}";
+    }
+
+    // Struct holding the tolerances when asserting the wrapping result.
+    struct WrappingTolerances final {
+        WrappingTolerances() = default;
+
+        WrappingTolerances(double eps):
```
adamkewley, line 90: Space before colon

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -0,0 +1,314 @@
...
+        double position = 1e-9;
+        double length = 1e-9;
+    };
+
+    bool IsEqual(
+        double lhs,
+        double rhs,
+        double tolerance)
+    {
+        return std::abs(lhs - rhs) < tolerance;
```
adamkewley, line 104: Maybe `<=` (e.g. is something being off by exactly one tolerance interval permitted)

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -0,0 +1,314 @@
...
+        WrappingTolerances(double eps):
+            position(eps),
+            length(eps)
+        {}
+
+        double position = 1e-9;
+        double length = 1e-9;
+    };
+
+    bool IsEqual(
```
adamkewley, line 99: Pedantic: `IsEqual` -> `IsWithinTolerance`/`IsEqualToWithin` etc.

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -0,0 +1,314 @@
...
+    struct WrapInput final {
+
+        SimTK::Rotation cylinderOrientation() const {
+            SimTK::Rotation orientation;
+            orientation.setRotationToBodyFixedXYZ(eulerRotations);
+            return orientation;
+        }
+
+        // Endpoints of the total path:
+        PathSegment path = {{}, {}};
```
adamkewley, line 159: `SimTK::Vec3`'s initialization rules imply that this defaults to "junk" in release mode (I believe?)

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -0,0 +1,314 @@
...
+            "quadrant: " << input.quadrant << ", " <<
+            "eulerRotations: " << input.eulerRotations << "}";
+    }
+
+    // Simulates the wrapping scenario as configured by the WrapInput.
+    WrapTestResult solve(const WrapInput& input, bool visualize) {
+        Model model;
+        model.setName("testWrapCylinderModel");
+
+        std::unique_ptr<WrapCylinder> cylinder(new WrapCylinder());
```
adamkewley, line 183: If it doesn't make indentation too ugly, consider a sub-scope that starts before `cylinder` and ends after `cylinder.release()` so that it is logically impossible for code further down to use the released (null) pointers.

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -0,0 +1,314 @@
...
+        model.setUseVisualizer(visualize);
+
+        const SimTK::State& state = model.initSystem();
+
+        if (visualize) {
+            model.realizeVelocity(state);
+            model.getVisualizer().show(state);
+        }
+
+        model.updComponent<PathSpring>("spring").getLength(state);
```
adamkewley, line 215: Unclear why this is necessary (presumably, you are using it to force some kind of internal caching/realization)?

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -0,0 +1,314 @@
...
+        if (!IsEqual(result, expected, tol.position)) {
+            std::ostringstream oss;
+            oss << "\nFAILED: case = " << testCase;
+            oss << "\n    input = " << input;
+            oss << "\n    result = " << result;
+            oss << "\n    expected = " << expected;
+            oss << "\n    caused by: "
+                << "Expected and simulated result not equal to within tolerance = "
+                << tol.position;
+            return oss.str();
```
adamkewley, line 258: Not a review comment, but understanding overload `(3)` in:

- https://en.cppreference.com/w/cpp/io/basic_ostringstream/str

Is an interesting exercise.

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -0,0 +1,314 @@
...
+        }
+
+        return std::string{};
+    }
+
+}
+
+int main()
+{
+    std::string failLog;
```
adamkewley, line 268: `std::vector<std::string>`?

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -0,0 +1,314 @@
...
+    failLog += TestWrapping(input, expected, tolerance, name);
+
+    // =========================================================================
+    // ======================= Perpendicular Case ==============================
+    // =========================================================================
+
+    name = "Perpendicular";
+    input.path = {{2, -2, 0}, {-2, 2.1, 0}};
+
+    expected.path = {
```
adamkewley, line 294: Where are these numbers from?

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
pepbos: makes sense. To pass tolerance of zero :)

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
pepbos: Yeah looks like that is the case

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
pepbos: Aaah, that's what I was looking for!

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
pepbos: yeah the idea was to trigger the realization. Bit strange that `getLength(state)` is marked as `const` though, not sure if that is common practice?

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
pepbos: My idea was to add more assertions for each case, such that (hopefully) all edge cases and interesting wrapping conditions can be captured. But before adding those cases, I would like to add the testing of certain geometric properties, such as path gradient, wrapping direction etc. So I just added a simple case for which the wrapping path is very accurate, but is otherwise not necessarily interesting.

But yeah, at this moment, there is no way of testing if this wrapping is accurate, so it seems a bit random.

adamkewley: A comment along the lines of "these were scooped from OpenSim 4.5", "calculated by hand", etc. is enough here

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+    // Simulates the wrapping scenario as configured by the WrapInput.
+    WrapTestResult solve(const WrapInput& input, bool visualize) {
+        Model model;
+        model.setName("testWrapCylinderModel");
+
+        // Add a spring to create the wrapping path.
+        {
+            std::unique_ptr<PathSpring> spring (
```
nickbianco, line 194: Remove space after `spring`.

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
nickbianco: The `const`-ness of a method doesn't apply to `State` realization. It's related to whether or not a method will change the properties of a OpenSim object (e.g., `upd_stiffness` is non-`const` since it modifies the `PathSpring` stiffness).

nickbianco (commented):
Great addition @pepbos! I added a few comments.

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -0,0 +1,332 @@
...
+        PathSegment() = default;
+
+        PathSegment(
+            const SimTK::Vec3& startPoint,
+            const SimTK::Vec3& endPoint) :
+            start(startPoint),
+            end(endPoint)
+        {}
+
+        SimTK::Vec3 start {SimTK::NaN};
```
adamkewley, line 48: Initializers don't usually have spaces between the name and the initializing statement (e.g. `std::vector<int> vals(2);`, `std::vector<int> vals{1,2,3};`, not `std::vector<int> vals (2);`)

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -0,0 +1,332 @@
...
+        static WrapTestResult NoWrap() {
+            WrapTestResult result;
+            result.noWrap = true;
+            return result;
+        }
+
+        // Path segment on cylinder surface.
+        PathSegment path;
+        // Path segment length.
+        double length = NAN;
```
adamkewley, line 77: Use `NAN`/`SimTK::NaN` consistently - if possible

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
adamkewley: It might be useful to document why it's there or, if this works, explicitly realize the state rather than realizing it via this method (e.g. `model.realizeState(state, SimTK::Stage::Whatever);`)

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
adamkewley: Yes, because it makes it explicit in the source code that the variable is "a sequence of error messages" rather than relying on an in-string encoding.

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
pepbos: "Realizing" the state will not trigger the computation, hence this sort of manual triggering of the wrap computation. But I added a comment.

pepbos:
@adamkewley @nickbianco thanks for all your comments! Mostly the resolutions seemed straightforward, but let me know!

nickbianco (approved):
LGTM!

PR #3500, opened 2023-06-27
Cleanup: Muscle state estimate return type in `Millard2012QuilibriumMuscle`

pepbos:
This PR adds a struct for capturing the output of the internal function `Millard2012QuilibriumMuscle::estimateMuscleFiberState`.

Previously the output was stored in a `std::pair`, with the second field a `std::map`. This made it less clear what exactly the returned values are. Since the returned information was always the same, this PR adds a struct to hold it.

### Brief summary of changes

Adds `MuscleStateEstimate` to capture the output of `Millard2012QuilibriumMuscle::estimateMuscleFiberState`.

### Testing I've completed

Visual verification that Rajagopal model simulation gives same result before and after change.
Passed the unit tests.

### CHANGELOG.md

- no need to update because minor internal refactor.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3500)

### DIFF
--- a/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
+++ b/OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
@@ -556,7 +556,7 @@ computeFiberEquilibrium(SimTK::State& s, bool solveForVelocity) const
     double pathSpeed = solveForVelocity ? getLengtheningSpeed(s) : 0;
     double activation = getActivation(s);
 
-    const MuscleStateEstimatorResult result =
+    const MuscleStateEstimate result =
         estimateMuscleFiberState(
             activation,
             pathLength,
@@ -566,28 +566,28 @@ computeFiberEquilibrium(SimTK::State& s, bool solveForVelocity) const
             solveForVelocity);
 
     if (result.status ==
-        MuscleStateEstimatorResult::Status::Failure_MaxIterationsReached) {
+        MuscleStateEstimate::Status::Failure_MaxIterationsReached) {
             std::ostringstream oss;
             oss << "Internal exception encountered:\n"
-                << "    Solution error " << std::abs(result.solution_error)
-                << "    exceeds tolerance of " << tol << "\n"
+                << "    Solution error " << std::abs(result.solutionError)
+                << " exceeds tolerance of " << tol << "\n"
                 << "    Newton iterations reached limit of " << maxIter << "\n"
                 << "    Activation is " << activation << "\n"
-                << "    Fiber length is " << result.fiber_length << "\n";
+                << "    Fiber length is " << result.fiberLength << "\n";
             OPENSIM_THROW_FRMOBJ(MuscleCannotEquilibrate, oss.str());
     }
 
     if (result.status ==
-        MuscleStateEstimatorResult::Status::Warning_FiberAtLowerBound) {
+        MuscleStateEstimate::Status::Warning_FiberAtLowerBound) {
             log_warn(
                 "Millard2012EquilibriumMuscle static solution: '{}' is at its"
                     "minimum fiber length of {}.",
                 getName(),
-                result.fiber_length);
+                result.fiberLength);
     }
 
-    setActuation(s, result.tendon_force);
-    setFiberLength(s, result.fiber_length);
+    setActuation(s, result.tendonForce);
+    setFiberLength(s, result.fiberLength);
 }
 
 //==============================================================================
@@ -1351,7 +1351,7 @@ double Millard2012EquilibriumMuscle::clampFiberLength(double lce) const
     return max(lce, getMinimumFiberLength());
 }
 
-Millard2012EquilibriumMuscle::MuscleStateEstimatorResult
+Millard2012EquilibriumMuscle::MuscleStateEstimate
     Millard2012EquilibriumMuscle::estimateMuscleFiberState(
         const double aActivation,
         const double pathLength,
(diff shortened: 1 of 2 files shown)


OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -556,43 +556,38 @@ computeFiberEquilibrium(SimTK::State& s, bool solveForVelocity) const
...
+            pathLength,
+            pathSpeed,
+            tol,
+            maxIter,
+            solveForVelocity);
+
+    if (result.status ==
+        MuscleStateEstimate::Status::Failure_MaxIterationsReached) {
+            std::ostringstream oss;
+            oss << "Internal exception encountered:\n"
```
adamkewley, line 571: You do not need to state `Internal exception encountered:` when throwing an exception. A higher-level system (e.g. one that `try {} catch {}`es the exception and logs it) would be responsible for pointing out that the system encountered an internal exception (and here's the message).

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -1601,59 +1595,41 @@ Millard2012EquilibriumMuscle::estimateMuscleFiberState(
         iter++;
     }
 
-    // Populate the result map.
-    ValuesFromEstimateMuscleFiberState resultValues;
+    MuscleStateEstimate result;
+    result.solutionError = ferr;
 
-    if(abs(ferr) < aSolTolerance) {  // The solution converged.
+    if(abs(ferr) < aSolTolerance) {
```
adamkewley, line 1601: Re-add `  // the solution converged`: it's a useful comment when scan-reading the code

OpenSim/Actuators/Millard2012EquilibriumMuscle.h
```diff
@@ -734,18 +734,19 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Millard2012EquilibriumMuscle, Muscle);
...
-        Success_Converged,
-        Warning_FiberAtLowerBound,
-        Failure_MaxIterationsReached
+    struct MuscleStateEstimate {
+        double solutionError = SimTK::NaN;
+        double fiberLength = SimTK::NaN;
+        double tendonVelocity = SimTK::NaN;
+        double tendonForce = SimTK::NaN;
+
+        enum Status {
```
adamkewley, line 743: Prefer `enum class` over `enum` - especially for non-bitflag uses like this.

pepbos:
@adamkewley thanks for your comments! @tkuchida , @mjhmilla , @nickbianco, care to take a look?

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -556,43 +556,38 @@ computeFiberEquilibrium(SimTK::State& s, bool solveForVelocity) const
...
+    if (result.status ==
+        MuscleStateEstimate::Status::Warning_FiberAtLowerBound) {
+            log_warn(
+                "Millard2012EquilibriumMuscle static solution: '{}' is at its"
+                    "minimum fiber length of {}.",
+                getName(),
+                result.fiberLength);
     }
+
+    setActuation(s, result.tendonForce);
```
tkuchida, line 589: Previously, the `setActuation()` and `setFiberLength()` lines were not executed if `Failure_MaxIterationsReached`, so this PR changes the behavior of the method. Without a deeper investigation, I don't know whether there are consequences to changing this behavior.

mjhmilla:
I'll take a look once my conference presentation drafts are ready. Hopefully sometime next week.

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
pepbos: The `OPENSIM_THROW_FRMOBJ` throws an exception, which prevents anything thereafter being reached. This should still prevent `setActuation()` and `setFiberLength()` from being reached in that case.

tkuchida: I thought this exception might have been caught and handled using another strategy but this doesn't seem to be the case.

OpenSim/Actuators/Millard2012EquilibriumMuscle.h
```diff
@@ -760,8 +761,7 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Millard2012EquilibriumMuscle, Muscle);
     @param staticSolution set to true to calculate the static equilibrium
            solution, setting fiber and tendon velocities to zero
     */
-    std::pair<StatusFromEstimateMuscleFiberState,
-              ValuesFromEstimateMuscleFiberState>
+    MuscleStateEstimate
```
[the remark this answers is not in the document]
pepbos: very true. Fixed it :)

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
pepbos: I'm not sure I follow.
Previously, if max iterations was reached and there was no convergence, we threw an exception:
```
       case StatusFromEstimateMuscleFiberState::Failure_MaxIterationsReached:
            // Report internal variables and throw exception.
            std::ostringstream ss;
            ...
            OPENSIM_THROW_FRMOBJ(MuscleCannotEquilibrate, ss.str());
            break;
```
this exception was caught, but in the `catch` block it threw another exception:
```
    } catch (const std::exception& x) {
        OPENSIM_THROW_FRMOBJ(MuscleCannotEquilibrate,
            "Internal exception encountered.\n" + std::string{x.what()});
```

I thought this would simplify to throwing an exception in case of no-convergence.
From your comment, I'm not sure if you agree or not :)

tkuchida: Looks good… IIRC at some point there were exceptions in this class that were thrown and then caught internally to proceed with some default behavior. Not the case here 😌

adamkewley:
LGTM 👍

nickbianco (approved):
LGTM!

PR #3503, opened 2023-06-27
Remove setup_win_python38.py and use init.py from conda

aymanhab:
Fixes issue #3502

### Brief summary of changes
Removed the special script setup_win_python38, instead have init.py add_dll_directories if exist for installation directories layout as well as the build/development layout.
### Testing I've completed
Local build and ci run successfully on windows (other platforms are unaffected.

### CHANGELOG.md (choose one)
- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3503)


adamkewley:
I don't know about the build failures etc. but the changeset looks reasonable enough: I guess the idea is that merely `import`ing the opensim python module now essentially does the work that `setup_win` used to?

aymanhab:
> I don't know about the build failures etc. but the changeset looks reasonable enough: I guess the idea is that merely `import`ing the opensim python module now essentially does the work that `setup_win` used to?

Indeed, and that this import would work regardless of the layout (conda vs. build vs. install).

nickbianco (approved):
LGTM! Thanks for fixing this @aymanhab!

aymanhab:
Thanks @nickbianco and @adamkewley Merging since the windows build which are the only ones affected by this change are passing.

PR #3504, opened 2023-06-29
Wrap Cylinder Test: Adds rotation direction assertion

pepbos:
This PR adds rotation direction assertion to `testWrapCylinder.cpp`.

The rotation direction of the wrapping of the cylinder (positive or negative wrt the cylinder-axis) is computed from the simulation result. This is done by looking at the first straight-line-segment, so the first two points of the total path. These two points will never have an angle > 180 between them as seen from the cylinder frame. This should be indicative of the wrapping direction.
This wrapping direction is then asserted against the expected direction.

The expected wrapping direction can be seen by visualizing the wrapping path:

testWrapCylinder-perpendicular-unconstrained-quarter

Since the point starts from `[2, -2, 0]` we would expect a positive direction.

This change allows for a permutation of existing tests: If we swap the start and end points of the path, we should get the same path, but with flipped rotation direction. So this was added as well.

This was the first item on the list for improving the testWrapCylinder:

- [x] Assert wrapping direction.
- [ ] Assert geodesic path properties, such as the path length, and gradient.
- [ ] Add more test cases,
- [ ] Add permutations to the test cases, such as rotating the scene, or reversing the start and end points.
- [ ] Merging with the cylinder wrapping test currently in testWrappingAlgorithm.cpp.

### Brief summary of changes

Assert the rotation direction of the cylinder wrapping algorithm in `testWrapCylinder.cpp`.

### CHANGELOG.md

- no need to update because new tests for wrapping a cylinder was previously added to the changelog.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3504)

### DIFF
--- a/OpenSim/Tests/Wrapping/testWrapCylinder.cpp
+++ b/OpenSim/Tests/Wrapping/testWrapCylinder.cpp
@@ -227,7 +227,7 @@ namespace {
         }
 
         // Endpoints of the total path:
-        PathSegment path = {{}, {}};
+        PathSegment path;
 
         // Wrapping cylinder parameters:
         double radius = 1.;
--- a/OpenSim/Tests/Wrapping/testWrapCylinder.cpp
+++ b/OpenSim/Tests/Wrapping/testWrapCylinder.cpp
@@ -46,6 +46,17 @@ namespace {
             end(endPoint)
         {}
 
+        // Returns PathSegment with start and end swapped.
+        PathSegment getReversed() const
+        {
+            return PathSegment{end, start};
+        }
+
+        void setReversed()
+        {
+            *this = getReversed();
+        }
+
         SimTK::Vec3 start{SimTK::NaN};
         SimTK::Vec3 end{SimTK::NaN};
     };
@@ -374,6 +385,13 @@ int main()
     expected.length = 0.689036814042993;
 
     expected.noWrap = false;
+
+    failLog.push_back(TestWrapping(input, expected, tolerance, name));
+
+    // Swapping start and end should not change the path:
+    input.path.setReversed();
+    expected.path.setReversed();
+    expected.positiveDirection = false;
 
     failLog.push_back(TestWrapping(input, expected, tolerance, name));
 
--- a/OpenSim/Tests/Wrapping/testWrapCylinder.cpp
+++ b/OpenSim/Tests/Wrapping/testWrapCylinder.cpp
@@ -73,17 +73,31 @@ namespace {
         };
     }
 
-    // Angular distance from start- to end-angle in either positive or negative direction.
+    enum class RotationDirection {
+        Positive,
+        Negative,
+    };
+
+    std::ostream& operator<<(
+        std::ostream& os,
+        const RotationDirection& direction)
+    {
+        return os << "RotationDirection::" << (
+            (direction == RotationDirection::Positive)? "Positive": "Negative");
+    }
+
+    // Angular distance from start- to end-angle in either positive or negative
+    // direction.
     double AngularDistance(
         double startAngle,
         double endAngle,
-        bool positiveRotation)
+        RotationDirection direction)
     {
         double distance = std::fmod(endAngle - startAngle, c_TAU);
-        while ((distance < 0.) && positiveRotation) {
+        while (distance < 0. && direction == RotationDirection::Positive) {
             distance += c_TAU;
         }
-        while ((distance > 0.) && !positiveRotation) {
+        while (distance > 0. && direction == RotationDirection::Negative) {
             distance -= c_TAU;
         }
         return distance;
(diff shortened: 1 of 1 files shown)


pepbos:
Hi @adamkewley , @nickbianco , @aymanhab , here is an addition to the cylinder wrapping test. If you have the time, let me know what you think!

nickbianco (commented):
~~One small comment, but otherwise LGTM.~~ EDIT: wrong PR.

pepbos:
@nickbianco perhaps it's me, but I don't see your comment.

nickbianco:
@pepbos, whoops, I meant to comment that on the muscle state PR!

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -33,6 +33,7 @@
 using namespace OpenSim;
 
 namespace {
+    const double c_TAU = 2. * SimTK::Pi;
```
adamkewley, line 36: `constexpr`

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -45,6 +46,17 @@ namespace {
             end(endPoint)
         {}
 
+        // Returns PathSegment with start and end swapped.
+        PathSegment getReversed() const
```
adamkewley, line 50: Rename: `getReversed()` - >`reversed()`, or (better) write a free function: `PathSegment Reversed(const PathSegment&)`

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -45,6 +46,17 @@ namespace {
...
+        {
+            return PathSegment{end, start};
+        }
+
+        void setReversed()
```
adamkewley, line 55: Remove, it's clearer to just write: `path = path.reversed();`

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -57,6 +69,51 @@ namespace {
...
+    PathSegment operator*(
+        const SimTK::Rotation& rot,
+        const PathSegment& path)
+    {
+        return PathSegment{
+            rot * path.start,
+            rot * path.end,
+        };
+    }
+
```
adamkewley, line 81: Add `enum class Direction { Positive, Negative };` and replace `bool` flags with that enum, so that downstream code can become `Direction windingDirection = Direction::Positive;` rather than `bool positiveDirection` (etc.)

(just a suggestion from a common code-smell guideline of "be careful about `bool` flag arguments to functions)

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i4-make-interfaces-precisely-and-strongly-typed

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -57,6 +69,51 @@ namespace {
...
+        return distance;
+    }
+
+    // Returns direction of shortest angular distance for a vector aligned with the
+    // start point to become aligned with the end point (true if positive).
+    bool DirectionOfShortestAngularDistance(
+        SimTK::Vec2 start,
+        SimTK::Vec2 end)
+    {
+        return AngularDistance(
```
adamkewley, line 104: Current formatting having a trailing `<=` is undesirable. Reformat to `double dist = f(...); return dist <= SimTK::Pi`

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -89,7 +148,9 @@ namespace {
         return os <<
             "WrapTestResult{" <<
             "path: " << result.path << ", " <<
-            "length: " << result.length << "}";
+            "length: " << result.length << ", " <<
+            "direction: " << (result.positiveDirection? "Positive" : "Negative")
```
adamkewley, line 152: This kind of code can be avoided with `enum class Direction` (suggested above)

adamkewley (commented):
Seems fine: mostly formatting comments from me.

nickbianco:
@pepbos sorry for being out-of-the-loop, I've been traveling for the last week. @adamkewley's changes LGTM. We can merge once you add the changes.

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
pepbos: oh, nice link :)

pepbos:
No worries @nickbianco , and thanks for the comments @adamkewley, they made sense to me, so think I straightforwardly fixed them. Let me know what you think!

PR #3510, opened 2023-07-12
Fix init order error found by libASAN's check_initialization_order

adamkewley:
Fixes issue (not worth writing up)

### Brief summary of changes

- Removes unused static variables that trigger libASAN's `check_initialization_order` tests
- The reason it's UB is because the arrays that are defined here (e.g. `num_circle_wrap_points`) are VLAs, because their size is parameterized with a non-`constexpr` value
- The variables are ultimately unused anyway, so this change has no effect

### Testing I've completed

- Ran OpenSim Creator with: `ASAN_OPTIONS=detect_container_overflow=1:malloc_context_size=30:check_initialization_order=true:detect_stack_use_after_return=true ./osc-build/apps/osc/osc`
- It failed to boot because this code was triggering UB
- I removed the code and recompiled OpenSim
- Things worked again

N/A

### CHANGELOG.md (choose one)

- no need to update because it would bore 99 % of readers

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3510)

### DIFF
--- a/OpenSim/Simulation/Wrap/WrapCylinderObst.cpp
+++ b/OpenSim/Simulation/Wrap/WrapCylinderObst.cpp
@@ -35,12 +35,6 @@ static const char* wrapTypeName = "cylinderObst";
 static const double TwoPi = 2.0*SimTK::Pi;
 static const double max_wrap_pts_circle_ang = (5.0/360.0)*TwoPi;
 
-// The following variables could be used for speedy wrap_pts definitions (NOT CURRENTLY USED)
-static const int num_circle_wrap_pts = 36;  // Number of circle points in 360 degrees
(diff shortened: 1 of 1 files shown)


aymanhab (approved):
Thanks @adamkewley Generally speaking these Obst wrap objects are never used by any of our published models, it's unknown whether they were ever used. I will recommend removing them altogether with the new wrapping to reduce baggage.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3510#-:-N_e8RPc0Quk-C4Xtplh:bo8f5w4)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adamkewley)

Issue #3387, opened 2023-02-03
Updates to the muscle wrapping algorithms


nickbianco:
The purpose of this Epic is to organize all issues and PRs related to supporting new muscle wrapping algorithms in OpenSim, including function-based muscle lengths and moment arms and faster geometry-based methods for path computations. 

See the [planning document](https://docs.google.com/document/d/1bU2w2hcU1DE8FOEFwia7P8ANnzqg6lsZs17aAknozfM/edit?usp=sharing) for more information.

PR #3515, opened 2023-07-14
Add regression tests for checking muscle-tendon lengths based on muscle wrapping

nickbianco:
Related to #3387.

### Brief summary of changes
This PR adds `testMuscleLengthRegression.cpp` to help us keep track of any undesired changes to muscles length computations  due to changes in muscle wrapping algorithms. As @adamkewley noted below, some test failures might occur due to wrapping algorithm changes that are not necessarily undesired. In these cases, we can update muscle length data files. 

These tests can be expanded/adapted for new wrapping algorithms as necessary. 

### CHANGELOG.md (choose one)

- no need to update because...addition to test suite.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3515)


nickbianco:
Good point, @adamkewley! This isn't my main focus this week (I had a few spare minutes so I cooked this up), so I'll wait until @pepbos finishes his work before creating a proper PR. This is mostly preparation for the muscle wrapping changes that I (finally) want to implement after TGCS/ISB. I just wanted something to verify that restructuring the wrapping code doesn't introduce any changes to muscle-lengths (and by extension, moment ams), which is what users ultimately care about.

nickbianco:
@adamkewley, I think it is a good time to properly review this PR, now that @pepbos's recent tests have been merged. Would you mind leaving a review?

aymanhab:
I agree @nickbianco, thanks for checking 👍

OpenSim/Tests/Wrapping/CMakeLists.txt
```diff
@@ -6,4 +6,5 @@ OpenSimAddTests(
     TESTPROGRAMS ${TEST_PROGS}
     DATAFILES ${TEST_FILES}
     LINKLIBS osimTools
+    INCLUDEDIRS "${CMAKE_SOURCE_DIR}/tropter/external/catch"
```
adamkewley, line 9: It feels a bit gross to pull in `catch` like this, because it's effectively making the wrapping test suite dependent on a transitive dependency (`catch`) that's within another library that's copied in-tree (`tropter`).

I can understand why you'd do it (other places seem to use the header), but I would caution against using `catch` unless it's a direct dependency of OpenSim - otherwise, the risk is that updating the in-tree tropter has a knock-on effect on code that doesn't actually use tropter.

OpenSim/Tests/Wrapping/testMuscleLengthRegression.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include <OpenSim/OpenSim.h>
+#include <OpenSim/Actuators/ModelOperators.h>
+#include <SimTKcommon.h>
+
+#define CATCH_CONFIG_MAIN
+#include <Vendors/tropter/external/catch/catch.hpp>
+
+using namespace OpenSim;
+
+TimeSeriesTable createMuscleLengthsTable(const Model& model,
```
adamkewley, line 32: If this is an internal helper function, I'd recommend putting into an anonymous namespace (recommended) or marking with `static` - otherwise, as a reader, I'm wondering whether `createMuscleLengthsTable` is symbol being emitted for consumption via a header in a different translation unit.

OpenSim/Tests/Wrapping/testMuscleLengthRegression.cpp
```diff
@@ -0,0 +1,99 @@
...
+    auto statesTraj = StatesTrajectory::createFromStatesTable(
+            model, coordinates, true, true, true);
+    TimeSeriesTable lengths;
+    const auto& muscleSet = model.getMuscles();
+    for (const auto& state : statesTraj) {
+        model.realizePosition(state);
+        SimTK::RowVector lengthsRow(muscleSet.getSize());
+        for (int imuscle = 0; imuscle < muscleSet.getSize(); ++imuscle) {
+            const auto& muscle = muscleSet.get(imuscle);
+            const auto& geometry_path = muscle.getGeometryPath();
```
adamkewley, line 43: Symbol naming style doesn't match other lines (`camelCase`)

adamkewley (commented):
No major issues - mostly style and an architectural comment etc. here or there

:+1:

OpenSim/Tests/Wrapping/testMuscleLengthRegression.cpp
```diff
@@ -0,0 +1,99 @@
...
+            TableProcessor("subject_walk_armless_coordinates.mot") |
+            TabOpUseAbsoluteStateNames();
+    TimeSeriesTable coordinates =
+            tableProcessor.processAndConvertToRadians(model);
+
+    TimeSeriesTable lengths = createMuscleLengthsTable(model, coordinates);
+//    STOFileAdapter::write(lengths,
+//            "std_testMuscleLengthRegression_subject_walk_armless.sto");
+
+    TimeSeriesTable std_lengths(
```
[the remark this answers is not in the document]
nickbianco: Done.

OpenSim/Tests/Wrapping/testMuscleLengthRegression.cpp
nickbianco: Done.

adamkewley (approved):

PR #3521, opened 2023-07-18
Add tangency assertion to wrapCylinderTest

pepbos:
This PR adds assertion of the path gradient to the cylinder wrapping test.

Since the geodesic on a cylinder is a helix shape, the direction of the path along the surface can be calculated given the start and end points on the surface. This tangent-direction on the cylinder surface can then be used to assert the direction of the straight-line path segments, since they must be aligned.

The added value of this PR is that instead of doing characterization tests, we can assert the geodesic properties in a data driven way. This will become more important for wrapping scenarios in which the result obtained from the current algorithm is less accurate.

Another wrapping test case is added in this PR as well, to validate the testing of the path direction.

### Brief summary of changes

- Added `CylinderGeodesic` struct as a helper for computing the tangent vector along the path.
- Added assertion of the resulting path length against that obtained from the `CylinderGeodesic`.
- Added assertion of the resulting path direction against that obtained from the `CylinderGeodesic`.
- Added another test case.

### Looking for feedback on

Feedback, and new ideas are always welcome.

### CHANGELOG.md

- no need to update because these are incremental improvements to the cylinder wrapping test.


pepbos:
@adamkewley @nickbianco Hi! another addition to the cylinder wrapping tests! have a look if you have time :)

nickbianco:
LGTM! (Finally back from lots of travel)

nickbianco (approved):

pepbos:
Great! thanks for reviewing!

Issue #3388, opened 2023-02-06
Create `AbstractPath` class for new muscle wrapping algorithms


nickbianco:
We need a new abstract class so we can support and unify the current muscle wrapping algorithm along with new algorithms based on other geometry-based methods or function based paths. 

See the [planning document](https://docs.google.com/document/d/1bU2w2hcU1DE8FOEFwia7P8ANnzqg6lsZs17aAknozfM/edit#) for more info.

## Changes to implement

- Create abstract path class (`AbstractPath`) and make `GeometryPath` derive from it
- ~~Support `AbstractPath` in `MomentArmSolver`~~ (might only need this class for `GeometryPath`, revisit later)
- Update `PathActuator`, `PathSpring`, `Muscle`, `Ligament`, `Blankevoort1991Ligament`, and `ActivationFiberLengthMuscle` to replace `GeometryPath` properties with `AbstractPath`
- Update pre-/post-scale calls for all `Muscle`-derived classes
- Update `testWrapping`, `testScale`, `testMomentArms`
- Test loading a model with `GeometryPaths`

PR #3535, opened 2023-08-26
Implement `AbstractPath` and related API changes to support multiple path types 

nickbianco:
Fixes issue #3388

### Brief summary of changes
This PR implements `AbstractPath`, a new base class from which all path-like objects can derive from, including `GeometryPath`. Specifically, derived objects will need to implement virtual methods that provide the path length, lengthening speed, moment arms, body and mobility forces, and visualization properties.

My current attempt at this implementation led to a few API changes:
1. Muscles and other path-dependent objects can no longer have convenience methods for adding path points (e.g., `Muscle::addNewPathPoint`), since other path types might not use `PathPoint`s at all. 
2. Classes that had a `GeometryPath` property (i.e., `PathActuator`, `PathSpring`, `Ligament`, and `Blankevoort1991Ligament`), now have an `AbstractPath` property, which has been moved to `protected` (TODO: move to `private`?).
3. All methods `updGeometryPath` and `getGeometryPath` have been changed to `updPath` and `getPath`, which return `AbstractPath` references.
4. Templatized signatures for `updPath` and `getPath` for returning concrete objects (e.g., `GeometryPath`) have been added.
5. I've introduced `initGeometryPath`, which is a safe method for overwriting the path-based object's current path with a `GeometryPath`. While the current default value for the path properties is `GeometryPath`, future methods can be introduced to initialize paths of different types (e.g., `initFunctionBasedPath`). I based this approach on how we initialize different solvers in `MocoStudy`. 

These changes led to a lot of changed files, but most of these files are tests or examples with models with very simple paths (e.g., a single origin and insertion point).

### Testing I've completed
Ran all the existing wrapping tests. I'm happy to expand or add new tests, especially for the interface changes.

Any feedback about the API changes and how they might affect users. This will obviously affect the GUI and OpenSim Creator, but my assumption is that most users do not interact much with `GeometryPath` directly and the effects of the API changes will be mostly internal. However, I am prepared to be totally wrong about that and I'm happy to make changes that best suit everyone.

### CHANGELOG.md (choose one)
- updated (but, TODO API changes).

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3535)


nickbianco:
Tests are failing because I haven't updated the bindings -- will do that next week.

nickbianco:
@aymanhab, I made some changes to the Java binding code to get everything to compile for now. We can discuss together what changes make the most sense for the GUI.

Working on the other test failures now.

OpenSim/Actuators/ModelFactory.cpp
```diff
@@ -171,35 +172,40 @@ void ModelFactory::replaceMusclesWithPathActuators(OpenSim::Model &model) {
...
-        auto& geomPath = actu->updGeometryPath();
-        for (int ip = 0; ip < pathPointSet.getSize(); ++ip) {
-            auto* pathPoint = pathPointSet.get(ip).clone();
-            const auto& socketNames = pathPoint->getSocketNames();
-            for (const auto& socketName : socketNames) {
-                pathPoint->updSocket(socketName)
-                        .connect(pathPointSet.get(ip)
-                                .getSocket(socketName)
-                                .getConnecteeAsObject());
+        // Copy the muscle's path.
```
adamkewley, line 175: This code's behavior/intent changes with the change.

The function is `ModelFactory::replaceMusclesWithPathActuator`. That should be fine with either point-based or function-based paths. The change here just entirely ignores not-point-based versions of paths, which means that the intent (i.e. to replace all muscles in a model with path actuators) is not upheld.

OpenSim/Examples/ExampleMain/OutputReference/TugOfWar_Complete.cpp
```diff
@@ -201,11 +201,13 @@ int main()
 
         // Specify the paths for the two muscles
         // Path for muscle 1
-        muscle1->addNewPathPoint("muscle1-point1", ground, Vec3(0.0,0.05,-0.35));
-        muscle1->addNewPathPoint("muscle1-point2", *block, Vec3(0.0,0.0,-0.05));
+        auto& path1 = muscle1->updPath<GeometryPath>();
```
adamkewley, line 204: A lot of this diff (e.g. the above 50-100 diff lines) demonstrate the problem with changing the API in a breaking way.

Because the `addNewPathPoint` was dropped (logically, cleaner), rather than `[[deprecated]]` (logically, unclean, but backwards-compatible), a large amount of changes are necessary throughout the codebase where the new code will fail in precisely the same way that the `[[deprecated]]`, runtime-checked backwards-compatible version would fail anyway.

Now imagine multiplying the number of changes necessary here by the number of lines of scripting code out in the wild (matlab scripts, whatever), guess what percentage of those developers are still around, what percentage of those will post issues, etc. - that's how much extra work is created vs. just supplying a basic backwards-compatible version.

OpenSim/Simulation/Model/AbstractPath.cpp
```diff
@@ -0,0 +1,64 @@
...
+    Appearance appearance;
+    appearance.set_color(SimTK::Gray);
+    constructProperty_appearance(appearance);
+}
+
+AbstractPath::AbstractPath(AbstractPath const&) = default;
+
+AbstractPath::~AbstractPath() noexcept = default;
+
+AbstractPath& AbstractPath::operator=(const AbstractPath&) = default;
```
adamkewley, line 41: https://en.cppreference.com/w/cpp/language/rule_of_three

Because OpenSim is C++11, you may want to consider Rule of Five

OpenSim/Simulation/Model/AbstractPath.h
```diff
@@ -0,0 +1,212 @@
...
+ */
+class OSIMSIMULATION_API AbstractPath : public ModelComponent {
+OpenSim_DECLARE_ABSTRACT_OBJECT(AbstractPath, ModelComponent);
+
+public:
+//=============================================================================
+// OUTPUTS
+//=============================================================================
+    OpenSim_DECLARE_OUTPUT(length, double, getLength, SimTK::Stage::Position);
+    OpenSim_DECLARE_OUTPUT(lengthening_speed, double, getLengtheningSpeed,
```
adamkewley, line 67: Gotta love that 80-column rule ;)

OpenSim/Simulation/Model/Blankevoort1991Ligament.cpp
```diff
@@ -62,6 +44,11 @@ Blankevoort1991Ligament::Blankevoort1991Ligament(
     set_slack_length(slack_length);
 }
 
+GeometryPath& Blankevoort1991Ligament::initGeometryPath() {
```
adamkewley, line 47: `init` methods on classes are *usually* a code smell - unless you're writing like a game developer with exceptions turned off or something.

OpenSim/Simulation/Model/GeometryPath.cpp
```diff
@@ -388,7 +382,7 @@ getPointForceDirections(const SimTK::State& s,
 /* add in the equivalent spatial forces on bodies for an applied tension 
     along the GeometryPath to a set of bodyForces */
 void GeometryPath::addInEquivalentForces(const SimTK::State& s,
-    const double& tension, 
+    double tension,
```
adamkewley, line 385: Unrelated API change?

OpenSim/Simulation/Model/PathActuator.h
```diff
@@ -67,10 +66,22 @@ class OSIMSIMULATION_API PathActuator : public ScalarActuator {
...
+    template <typename PathType>
+    PathType& updPath() {
+        return dynamic_cast<PathType&>(upd_path());
+    }
+    template <typename PathType>
+    const PathType& getPath() const {
+        return dynamic_cast<PathType&>(get_path());
+    }
+    bool hasPath() const override { return true;};
+
```
adamkewley, line 80: Consider non-throwing equivalents: `tryGetPath`, `tryUpdPath`, where a pointer to `PathType` is returned and can be runtime-checked by the caller (for cases where you currently use `dynamic_cast`)

OpenSim/Tests/Wrapping/testWrapCylinder.cpp
```diff
@@ -451,8 +453,9 @@ namespace {
 
         // Trigger computing the wrapping path (realizing the state will not).
         model.getComponent<PathSpring>("spring").getLength(state);
-        const WrapResult wrapResult = model.getComponent<PathSpring>("spring")
-                .getGeometryPath()
+        const auto& path = dynamic_cast<const GeometryPath&>(
```
adamkewley, line 456: `getPath<T>`

adamkewley (commented):
Mostly looks fine but I would strongly consider supplying `[[deprecated]]` backwards-compatible methods, rather than entirely removing the old API, where those methods runtime-downcast-and-check that the implementation is point-based, throwing otherwise. By doing that, you'll drastically reduce this diff while making rollout of the change easier for downstream scripting users.

There's also some shortcuts w.r.t. copying paths in the code. I'd recommend carefully reading where I've mentioned a change changes intent/behavior. The new change may compile, but it is just installing a later issue that someone else has to dig up and figure out what's going on (e.g. "why does my UI crash if I load a model (with FBPs)", "why can't I convert these FBP-based muscles to `DeGrootFregly` muscles", etc.)

Bindings/Java/OpenSimJNI/OpenSimContext.cpp
```diff
@@ -124,13 +125,21 @@ double OpenSimContext::getMuscleLength(Muscle& m) {
 }
 
 const Array<AbstractPathPoint*>& OpenSimContext::getCurrentPath(Muscle& m) {
-  return m.getGeometryPath().getCurrentPath(*_configState);
+  if (auto* p = dynamic_cast<const GeometryPath*>(&m.getPath())) {
```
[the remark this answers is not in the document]
nickbianco: Agreed. Reverting these changes for now since I've added back `getGeometryPath()`. @aymanhab and I can discuss what makes sense for a refactor since (I believe) this file is GUI-specific.

Bindings/Java/OpenSimJNI/OpenSimContext.cpp
```diff
@@ -124,13 +125,21 @@ double OpenSimContext::getMuscleLength(Muscle& m) {
...
+  } else {
+      OPENSIM_THROW(Exception, "Muscle path is not a GeometryPath.")
+  }
 }
 
 void OpenSimContext::copyMuscle(Muscle& from, Muscle& to) {
-  to = from;
-  recreateSystemKeepStage();
-  to.updGeometryPath().updateGeometry(*_configState);
+  if (auto* p = dynamic_cast<const GeometryPath*>(&from.getPath())) {
```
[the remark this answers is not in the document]
nickbianco: Similar to above, reverting these changes since I've reverted the `updGeometryPath` change. Will discuss with @aymanhab what makes sense for a refactor here.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ request related to the change, then we may provide the commit.
 
 This is not a comprehensive list of changes but rather a hand-curated collection of the more notable ones. For a comprehensive history, see the [OpenSim Core GitHub repo](https://github.com/opensim-org/opensim-core).
 
+v4.5
+====
+- Added `AbstractPath` which is a base class for `GeometryPath` and other path types (#3388). TODO: add API changes.
```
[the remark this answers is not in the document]
nickbianco: Per your suggestions, I'm opting to revert API changes since supporting `getGeometryPath` and `updGeometryPath` is straight-forward anyway. But I agree, in the end, I'd prefer minimal API changes for those reasons.

OpenSim/Simulation/Model/GeometryPath.cpp
nickbianco: Sort of. When I wrote the virtual method `addInEquivalentForces` I made this `double` out of habit, and then updated this to follow. I vaguely remember reading somewhere that `const` references versus copies for simple types weren't that important, but I could be wrong. I'm fine with reverting just for the sake of not changing the function signature.

OpenSim/Simulation/Model/Blankevoort1991Ligament.cpp
nickbianco: I'm trying to find a way to support creating new objects that own paths (of any type) that is scripting friendly. I'm not sure that this is the best approach. Another approach could be the pointer adoptee approach used elsewhere in OpenSim (although I'm not sure how to update the `path` property in that case).

OpenSim/Simulation/Model/Muscle.h
```diff
@@ -441,6 +441,8 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Muscle, PathActuator);
     void extendSetPropertiesFromState(const SimTK::State &s) override;
     void extendInitStateFromProperties(SimTK::State& state) const override;
 
+    // Update the display geometry attached to the muscle
```
adamkewley, line 444: Very sus

OpenSim/Simulation/Model/PathActuator.h
```diff
@@ -66,17 +66,35 @@ class OSIMSIMULATION_API PathActuator : public ScalarActuator {
     // GET AND SET
     //--------------------------------------------------------------------------
     // Path
+    bool hasPath() const override { return true;};
```
adamkewley, line 69: nit: space after `;`

adamkewley (commented):
Looks like good changes.

The only thing I noticed was the new method on `Muscle`, why's it there?

OpenSim/Simulation/Model/Muscle.h
nickbianco: This method exists in the current API (check the full diff). I added it back now that `updGeometryPath` is supported again.

aymanhab (commented):
> Reviewed 1 of 62 files at r1, 1 of 43 files at r3, 1 of 16 files at r4.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3535)* status: 2 of 67 files reviewed, 20 unresolved discussions (waiting on @adamkewley and @nickbianco)

___
*[`CHANGELOG.md` line 11 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3535#gh-1308436445:-Nd6RZf5B9NhJIY-Ph3h:b-8fx2u7) ([raw file](https://github.com/opensim-org/opensim-core/blob/b70b2faef73a34df7b46ddcaeebda2edd966ec6e/CHANGELOG.md#L11)):*

Per your suggestions, I'm opting to revert API changes since supporting `getGeometryPath` and `updGeometryPath` is straight-forward anyway. But I agree, in the end, I'd prefer minimal API changes for those reasons. 
</blockquote></details>

:+1: 
___
*[`OpenSim/Simulation/Model/AbstractPath.h` line 102 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/3535#-Nd6SJdPASryGEFWXZbG:-Nd6SJdPASryGEFWXZbH:buechr7) ([raw file](https://github.com/opensim-org/opensim-core/blob/ce3b0c47607331519b57b4dd48d6eeef90bfb909/OpenSim/Simulation/Model/AbstractPath.h#L102)):*
> ```objc
>      */
>     virtual SimTK::Vec3 getColor(const SimTK::State& s) const = 0;
> ```

I feel most of the appearance and color stuff doesn't belong here and can be moved to GeometryPath but hard to tell looking at this class if that would cause other problems

OpenSim/Simulation/Model/GeometryPath.cpp
adamkewley: For trivial types (esp those that fit in a register) it's almost always best to pass by value, yeah, but I'd put any API change in a separate PR if possible

nickbianco: Sounds good! I'll revert the change and we can revisit later if needed.

OpenSim/Actuators/ModelFactory.cpp
nickbianco: I've added an `else` branch to catch unsupported path types. As we add new path types, we can add new `elseif` branches to included supported path types.

aymanhab (commented):
> Reviewed 2 of 16 files at r4.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3535)* status: 4 of 67 files reviewed, 20 unresolved discussions (waiting on @adamkewley and @nickbianco)

___
*[`Bindings/Java/OpenSimJNI/OpenSimContext.cpp` line 128 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3535#gh-1308408630:-Nd6WVvQAMiOWX0kBrvq:bghdrx8) ([raw file](https://github.com/opensim-org/opensim-core/blob/b70b2faef73a34df7b46ddcaeebda2edd966ec6e/Bindings/Java/OpenSimJNI/OpenSimContext.cpp#L128)):*

Agreed. Reverting these changes for now since I've added back `getGeometryPath()`. @aymanhab and I can discuss what makes sense for a refactor since (I believe) this file is GUI-specific.
</blockquote></details>

Indeed it is GUI specific, and some methods may not be used any more or can be easily refactored once the API settles
___
*[`Bindings/Java/OpenSimJNI/OpenSimContext.cpp` line 136 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3535#gh-1308416830:-Nd6WxIA5xtIKEnLFTqL:b3oolqe) ([raw file](https://github.com/opensim-org/opensim-core/blob/b70b2faef73a34df7b46ddcaeebda2edd966ec6e/Bindings/Java/OpenSimJNI/OpenSimContext.cpp#L136)):*

Similar to above, reverting these changes since I've reverted the `updGeometryPath` change. Will discuss with @aymanhab what makes sense for a refactor here.
</blockquote></details>

This function seems to be unused and can be removed when we revisit.
___
*[`Bindings/Java/OpenSimJNI/Test/testContext.cpp` line 0 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3535#gh-1308425329:-Nd6XMST-dh60sI5wrKF:bmi8ppe) ([raw file](https://github.com/opensim-org/opensim-core/blob/b70b2faef73a34df7b46ddcaeebda2edd966ec6e/Bindings/Java/OpenSimJNI/Test/testContext.cpp#L0)):*

*(Reviewable was [unable](https://github.com/Reviewable/Reviewable/issues/14) to map this GitHub inline comment thread to the right spot — sorry!)*

Is there any harm in keeping `updGeometryPath()` as an alias for `updPath<GeometryPath>`?

By removing `addNewPathPoint`, `updGeometryPath`, `getGeometryPath`, etc. you're publishing a correct(er) API, yes, but any downstream MATLAB/python scripts are going to be broken - even though there's no logical reason for them to break if they are operating on point-based paths (with legacy scripts, it is likely that they will be).

Prefer keeping any old API methods with a `[[deprecated]]` marking and have those legacy methods runtime-check the path type (in many cases, the paths will be point-based - for now). Maybe after sitting on it for a while the `[[deprecated]]` stuff can be dropped as part of a larger OpenSim5 rollout.

I wouldn't recommend breaking any existing OpenSim4 APIs. The vast majority of OpenSim's research userbase are using MATLAB/python, where they won't see the breakage coming and won't know why their scripts no longer work.
</blockquote></details>

:100:

OpenSim/Actuators/DeGrooteFregly2016Muscle.cpp
```diff
@@ -1021,31 +1022,36 @@ void DeGrooteFregly2016Muscle::replaceMuscles(
...
-        auto& geomPath = actu->updGeometryPath();
-        for (int ipp = 0; ipp < pathPointSet.getSize(); ++ipp) {
-            auto* pathPoint = pathPointSet.get(ipp).clone();
-            const auto& socketNames = pathPoint->getSocketNames();
-            for (const auto& socketName : socketNames) {
-                pathPoint->updSocket(socketName)
-                        .connect(pathPointSet.get(ipp)
-                                         .getSocket(socketName)
-                                         .getConnecteeAsObject());
+        // Copy the muscle's path.
```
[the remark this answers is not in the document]
nickbianco: Same as below, I've added an `else` branch to catch unsupported path types. As we add new path types, we can add new `elseif` branches to included supported path types.

The generic functions for copying muscle information could be implemented and should probably live somewhere else (@aymanhab has suggested this before). I think that could be addressed in a separate PR.

OpenSim/Actuators/DeGrooteFregly2016Muscle.cpp
```diff
@@ -1074,14 +1080,14 @@ void DeGrooteFregly2016Muscle::extendPostScale(
         const SimTK::State& s, const ScaleSet& scaleSet) {
     Super::extendPostScale(s, scaleSet);
 
-    GeometryPath& path = upd_GeometryPath();
+    AbstractPath& path = updPath();
```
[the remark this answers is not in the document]
nickbianco: I think this comment has been addressed based on the above changes, but let me know if not.

OpenSim/Simulation/Model/Ligament.h
```diff
@@ -162,6 +188,10 @@ OpenSim_DECLARE_CONCRETE_OBJECT(Ligament, Force);
         return values;
     }
 
+    OpenSim_DECLARE_PROPERTY(path, AbstractPath,
```
adamkewley, line 191: Should this be separate from the other `DECLARE_PROPERTY`s?

OpenSim/Simulation/Model/PathSpring.h
```diff
@@ -185,6 +207,11 @@ OpenSim_DECLARE_CONCRETE_OBJECT(PathSpring, Force);
         return values;
     }
 
+protected:
```
adamkewley, line 210: Separate from other `DECLARE_PROPERTY`s and now `protected`?

adamkewley (commented):
Had another pass

My main concern is adding those `if (dynamic_cast<GeometryPath>...) else { THROW }` blocks. The reason why is because they're the type of thing that will be buried in the codebase and there will be no warning that there's an issue until a user reports it (e.g. it's unlikely that developers will proactively poke around Blankevoort etc. looking for downcasts when adding a new path abstraction).

Even if the solution is (e.g.) to have a copy-assignment method that throws if `typeof(lhs) != typeof(rhs)` then *at least* the explosion is closer to where people are likely to be looking when working on AbstractPath/GeometryPath (rather than it being buried in a particular implementation somewhere)

Bindings/Java/OpenSimJNI/OpenSimContext.cpp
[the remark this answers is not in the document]
aymanhab: Will resolve this conversation as the method will  be retired or rewritten

OpenSim/Simulation/Model/Ligament.h
nickbianco: I moved this back with the other `DECLARE_PROPERTY`s. (I think I did this out of habit based on how we use properties in Moco, but it's not really necessary here).

OpenSim/Simulation/Model/PathSpring.h
nickbianco: See above.

nickbianco:
@adamkewley, thanks a bunch for the quick reviews across lots of changing files! Regarding the path copying utilities, I'm seeing your point clearer now. In this pass, I implemented a new virtual method `AbstractPath::copyFrom()` for handling copying of path properties. This still requires `dynamic_cast`s, but they're now contained in the concrete implementations rather than hidden in utilities somewhere in OpenSim. This seemed simplest to me, but let me know if you think it's sufficient.

@aymanhab, now would be a good time to jump in for a full review.

Some thoughts I have about the current state of the PR:
- @aymanhab, I saw your comment about moving the visualization methods to `GeometryPath`, but doing so would necessitate changes in `extendReallizeDynamics()` in both `Ligament` and `PathActuator`.  We could change these, but it would require checking types inside these methods which might not be ideal. Having virtual methods like `setColor()` that aren't used by some derived classes (e.g., `FunctionBasedPath`) seems like bad design, but if it doesn't break anything downstream then maybe it's fine.
- `Ligament` and `Blankevoort1991Ligament` both now have getters and setters for modifying the `path` property, whereas before the unnamed `GeometryPath` property was modifiable through auto-generated methods from ` OpenSim_DECLARE_PROPERTY` (e.g., `upd_GeometryPath`). Therefore, the current changes are breaking for anyone who is using the auto-generated methods for these classes. Two solutions to this (neither of which I love) are: 1) revert back to the unnamed `GeometryPath` property for these classes (since ligaments usually don't need wrapping objects, as @aymanhab pointed out) or 2) implement deprecated methods that cover all of the auto-generated methods that were removed. What do you guys think?
- Another API change: `Force::hasGeometryPath()` is now `Force::hasPath()`. Should we support backwards compatibility for `hasGeometryPath()` somehow (e.g., grab the property, see if a path exists, try a `dynamic_cast`, etc)?

aymanhab (changes requested):
@nickbianco I'd agree to avoid breaking changes to Ligaments, with that option 1 seems more straightforward.
I didn't find any usage of the method hasGeometryPath() in opensim-core, it's used by the GUI for special handling of visualization of GeometryPaths and as such should return true only when there're visuals to display/update (ties into the first comment) so we'll need to reimplement with dynamic_cast now or later, or find an alternative.

> Reviewed 1 of 62 files at r1, 1 of 6 files at r8.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3535#-:-NdDxyn_D50pxbiftpCJ:b-3mt7hq)* status: 6 of 70 files reviewed, 23 unresolved discussions (waiting on @adamkewley and @nickbianco)

___
*[`OpenSim/Analyses/MuscleAnalysis.h` line 32 at r8](https://reviewable.io/reviews/opensim-org/opensim-core/3535#-NdBnGyP70DxlBPCtLKg:-NdBnGyQ-fV7dIRejljL:by1disp) ([raw file](https://github.com/opensim-org/opensim-core/blob/c3ac610d2c7fb5f6e2b7fc4200a08d645c148aa0/OpenSim/Analyses/MuscleAnalysis.h#L32)):*
> ```objc
> #include <OpenSim/Common/PropertyStrArray.h>
> #include <OpenSim/Simulation/Model/Analysis.h>
> ```

Any reason this include is needed here, when the previous version didn't?

adamkewley (commented):
👍 all very good: thank you for enduring my feedback and making some nice changes.

Additional (very optional) suggestions:

- For copying things, because `AbstractPath` supports `.clone()`, you may be able to clone the source `AbstractPath` into a new pointer and then assign that pointer to the relevant property etc. in the target datastructure. The benefit of that approach is that it would work for any path implementation, whereas copy assignment (i.e. `copyFrom`) is limited to only allowing `T = T const&` expressions, which is why the downcast is necessary. Not your fault (the original code is copy assignment), but might be an idea.
- If you need `copyFrom` (e.g. because doing it at a higher level with `clone` is a no-go) then it might be better named as `assign`, to match`AbstractProperty::assign` and because it's a common idiom (e.g. `std::vector<T>::assign`)

Again, these are very optional suggestions (observations, even) - take them or leave them

nickbianco:
Thanks @adamkewley and @aymanhab!

A few more thoughts:

> For copying things, because AbstractPath supports .clone(), you may be able to clone the source AbstractPath into a new pointer and then assign that pointer to the relevant property etc. in the target datastructure. 

@adamkewley, I also considered this and like the idea of it, but couldn't fully wrap my head around it. Would default implementations of `Object::clone()` provided by `OpenSim_OBJECT_CONCRETE_DEFS` need to be skipped over (e.g., like `Model` does)? Or would you just keep the `assign` method and instead give it a pointer from the source object's `clone()`? Maybe this is a bit superfluous, but I'm inclined to do everything the "right way" now to save pain later since I'm going to be working very closely with these classes.

> I'd agree to avoid breaking changes to Ligaments, with that option 1 seems more straightforward.

@aymanhab, I am tempted to take the simplest path here, but it feels wrong to not support the new wrapping abstraction for half of the classes that own path objects. After a quick look, I did find [a breaking change in @clnsmith's opensim-jam project](https://github.com/clnsmith/opensim-jam/blob/fb777056217ec60e8eb2eedc93fefa5a6e4885e0/src/JointMechanicsTool.cpp#L885). Would it be so bad to write deprecated methods, even just for `upd_GeometryPath` and `get_GeometryPath`, since these are likely the most used? I don't want users who have models with many ligaments to miss the opportunity for a potentially faster solution. Personally, I would not want to be forced to use different path types simply because it's not support for some of the objects in my model (but again, that is a personal preference and not  a workflow issue).

> I didn't find any usage of the method hasGeometryPath() in opensim-core, it's used by the GUI for special handling of visualization of GeometryPaths and as such should return true only when there're visuals to display/update (ties into the first comment) so we'll need to reimplement with dynamic_cast now or later, or find an alternative.

I see, so setting `hasPath` to true for having *any* path is changing the intended functionality. Let me draft something and see what you think.

adamkewley:
> Would default implementations of `Object::clone()` provided by `OpenSim_OBJECT_CONCRETE_DEFS` need to be skipped over (e.g., like `Model` does)?

You want the `clone`'s implementation to be provided via the `CONCRETE_DEFS` so that a base-er class's virtual version of it can be implemented by it. For where the pointer needs to go, I imagine that any OpenSim collection of abstract objects *must* be storing them as pointers (they're abstract, so there is no way for the compiler to know how much space they need), so the "better" (clone-based) approach would need to install the `clone`d pointer into the underlying pointer holder/array that's used for abstract properties (if that makes sense) - I believe `OpenSim::ObjectProperty` has `assign` methods specifically for doing the clone-then-overwrite operation via pointers.

nickbianco:
@adamkewley, I've implemented your suggestions above for now. I'm thinking over your recent comments and taking a look `ObjectProperty`. This is getting into OpenSim territory (and, let's be honest, C++ concepts) I'm not as familiar with. Maybe we could have a brief chat next week to talk through some potential solutions?

@aymanhab, I've changed `hasPath()` to `hasVisualPath()`, which will indicate if the `Force` object has a path that can be visualized. To support this, I've introduced `AbstractPath::isVisualPath()` which concrete implementations must provide. Hopefully this more directly addresses the needs of the GUI while not changing things too much.

aymanhab (changes requested):
> Reviewed 5 of 62 files at r1, 24 of 43 files at r3, 9 of 16 files at r4, 2 of 9 files at r5, 4 of 8 files at r7, 10 of 18 files at r9, 11 of 11 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3535)* status: 69 of 70 files reviewed, 26 unresolved discussions (waiting on @adamkewley and @nickbianco)

___
*[`OpenSim/Simulation/Model/Model.cpp` line 326 at r10](https://reviewable.io/reviews/opensim-org/opensim-core/3535#-NdHHMOV299pViPBll6g:-NdHHMOWDZvmSvw2_k_0:b-xq7pia) ([raw file](https://github.com/opensim-org/opensim-core/blob/25ed001efb51f4f4b13a4f9f979841adc0efd787/OpenSim/Simulation/Model/Model.cpp#L326)):*
> ```cpp
>             }
>         }
>         if (versionNumber < 40500) {
> ```

@nickbianco Typically when we change xml representation of a class, it becomes responsible for its own update so knowledge about the class changes are kept local and the Model class doesn't grow into a bag of unrelated changes that it shouldn't be concerned with. Can you comment on whether its possible/difficult to move this change  to another more localized place?  Thank you

nickbianco:
@aymanhab I moved the XML updates to `Force`, it's much cleaner now.

aymanhab (approved):
> Reviewed 1 of 4 files at r11.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3535)* status: 68 of 71 files reviewed, 26 unresolved discussions (waiting on @adamkewley and @nickbianco)

aymanhab (approved):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3535)* status: 68 of 71 files reviewed, 26 unresolved discussions (waiting on @adamkewley and @nickbianco)

___
*[`OpenSim/Simulation/Model/Model.cpp` line 326 at r10](https://reviewable.io/reviews/opensim-org/opensim-core/3535#-NdHHMOV299pViPBll6g:-NdaHXuk7dfmXzrsp1ys:b-tf24ja) ([raw file](https://github.com/opensim-org/opensim-core/blob/25ed001efb51f4f4b13a4f9f979841adc0efd787/OpenSim/Simulation/Model/Model.cpp#L326)):*

@nickbianco Typically when we change xml representation of a class, it becomes responsible for its own update so knowledge about the class changes are kept local and the Model class doesn't grow into a bag of unrelated changes that it shouldn't be concerned with. Can you comment on whether its possible/difficult to move this change  to another more localized place?  Thank you
</blockquote></details>

Thank you :+1:

adamkewley:
> @adamkewley, I've implemented your suggestions above for now. I'm thinking over your recent comments and taking a look `ObjectProperty`. This is getting into OpenSim territory (and, let's be honest, C++ concepts) I'm not as familiar with. Maybe we could have a brief chat next week to talk through some potential solutions?
> 
> @aymanhab, I've changed `hasPath()` to `hasVisualPath()`, which will indicate if the `Force` object has a path that can be visualized. To support this, I've introduced `AbstractPath::isVisualPath()` which concrete implementations must provide. Hopefully this more directly addresses the needs of the GUI while not changing things too much.

We can have a chat about it, but don't let that hold anything up!

nickbianco:
@adamkewley replacing the new virtual method with `AbstractProperty::assign()` seems to work fine locally. If all the tests pass here, I'll go ahead and merge. Thanks again!

Issue #3389, opened 2023-02-06
Create `FunctionBasedPath` class to support muscle wrapping algorithms based on `Function` objects


nickbianco:
This class will need to derive from `AbstractPath`, which will be implemented first in a separate PR. 

- It should include options to set only one function for the muscle-tendon length, from which moment arms and muscle-tendon velocities can be computed. 
- It should also include an option to specify all of these functions manually. 
- Lastly, it will need keep a list of reference to `Coordinate`s which correspond to the arguments of each function.

PR #3544, opened 2023-09-11
Refactor `testWrapping.cpp`

nickbianco:
Relevant to #3387, #3389

### Brief summary of changes
This PR refactors `testWrapping.cpp`, which was previously a bit of a mess, to now use the `Catch2` framework. The new structure should make it easier to expand the test suite for wrapping code without having to create a bunch of new files. To that  end, I've moved the tests from `testMuscleLengthRegression.cpp` into `testWrapping.cpp`.

There was a lot of code that wasn't being tested at all. I've temporarily moved this code to a new file, `sandboxWrapping.cpp`, until we decide what to do with it.

What should we do with the unused code moved to `sandboxWrapping.cpp`? Change the filename (i.e., "deprecated") so the code is saved? Revive the unused tests? Delete it altogether?

### CHANGELOG.md (choose one)

- no need to update because...changes to testing framework.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3544)


nickbianco:
@pepbos, @adamkewley, since you've been recently adding new wrapping tests, you might have some thoughts/opinions here.

aymanhab:
@nickbianco I'd name the file deprecated for future reference or restoration. Most of these unused tests I'm not familiar with but a refactoring of unused tests should probably need no more than one reviewer. If actual tests are preserved then feel free to merge

adamkewley:
Seems reasonable - assuming the test suite wasn't testing anything too useful 👍

Issue #3532, opened 2023-08-25
Fix deleting an element from `OpenSim::Coordinate`'s `range` property causes `getRangeMin` to throw


adamkewley:
Expected behavior: it eagerly throws when calling `finalizeFromProperties`, so that the exception doesn't end up in a "verified as fine" model. The bug is because `OpenSim::Coordinate` delcares its `range` as a list of `n` elements (should a `Vec2`), so API users are permitted to delete or add elements to that list.

Downstream issue: https://github.com/ComputationalBiomechanicsLab/opensim-creator/issues/654
Reproduction:

```c++
OpenSim::Model model;

auto body = std::make_unique<OpenSim::Body>("body", 1.0, SimTK::Vec3{}, SimTK::Inertia{});
auto joint = std::make_unique<OpenSim::PinJoint>();
joint->setName("joint");
joint->updCoordinate().setName("rotation");
joint->connectSocket_parent_frame(model.getGround());
joint->connectSocket_child_frame(*body);
model.addJoint(joint.release());
model.addBody(body.release());

model.finalizeConnections();  // should be fine: the model is correct

auto& coord = model.updComponent<OpenSim::Coordinate>("/jointset/joint/rotation");
coord.updProperty_range().clear();  // uh oh: a coordinate with no range (also applies when deleting only one element)

model.finalizeConnections();  // should throw (but this bug indicates it does not)

ASSERT_ANY_THROW({ coord.getRangeMin(); });  // throws (shouldn't: should throw in finalizeConnections)
```

PR #3546, opened 2023-09-13
Fix deleting an element from OpenSim::Coordinate's causes getMinRange…

adamkewley:
Fixes issue #3532

### Brief summary of changes

- Adds a runtime check for the error condition to `OpenSim::Coordinate::extendFinalizeFromProperties`

### Testing I've completed

- Added the reproduction code in issue #3532 to `testJoints`
- The test failed
- Added the runtime check
- The test passed

### CHANGELOG.md (choose one)

- updated.

### DIFF
--- a/OpenSim/Simulation/SimbodyEngine/Coordinate.cpp
+++ b/OpenSim/Simulation/SimbodyEngine/Coordinate.cpp
@@ -141,6 +141,14 @@ void Coordinate::extendFinalizeFromProperties()
 {
     Super::extendFinalizeFromProperties();
 
+    // eagerly check if outside code has somehow managed to remove elements
+    // from this coordinate's `range` property (issue #3532)
+    OPENSIM_THROW_IF(
+        getProperty_range().size() < 2,
+        Exception,
+        "A coordinate range must contain at least two elements (minimum, maximum)"
+    );
+
     string prefix = "Coordinate("+getName()+")::extendFinalizeFromProperties:";
 
     // Make sure the default value is within the range when clamped
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,9 @@ v4.5
 own the property `path` of type `AbstractPath` instead of the `GeometryPath` unnamed property. Getters and setters have 
 been added to these forces to provide access to concrete path types (e.g., `updPath<T>`). In `Ligament` and 
 `Blankevoort1991Ligament`, usages of `get_GeometryPath`, `upd_GeometryPath`, etc., need to be been updated to 
-`getGeometryPath`, `updGeometryPath`, etc., or a suitable alternative.    
+`getGeometryPath`, `updGeometryPath`, etc., or a suitable alternative.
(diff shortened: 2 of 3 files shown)


OpenSim/Simulation/SimbodyEngine/Coordinate.cpp
```diff
@@ -141,6 +141,14 @@ void Coordinate::extendFinalizeFromProperties()
 {
     Super::extendFinalizeFromProperties();
 
+    // eagerly check if outside code has somehow managed to remove elements
+    // from this coordinate's `range` property (issue #3532)
+    OPENSIM_THROW_IF(
+        getProperty_range().size() < 2,
```
tkuchida, line 147: Fewer than 2 elements doesn't make sense for a range, but neither does more than 2. Perhaps demand that the size is equal to 2?

OpenSim/Simulation/SimbodyEngine/Coordinate.cpp
```diff
@@ -141,6 +141,14 @@ void Coordinate::extendFinalizeFromProperties()
...
+        Exception,
+        "A coordinate range must contain at least two elements (minimum, maximum)"
```
tkuchida, line 149: at least -> exactly ?

tkuchida (commented):
Nice :+1:. Minor comment below ⬇️

OpenSim/Simulation/SimbodyEngine/Coordinate.cpp
nickbianco: The size might be enforced automatically via the property macro that defines it, `OpenSim_DECLARE_LIST_PROPERTY_SIZE`.

adamkewley: > The size might be enforced automatically via the property macro that defines it, `OpenSim_DECLARE_LIST_PROPERTY_SIZE`.

The macro pumps the size to `constructProperty_##name`, which moves it into `AbstractProperty::_minListSize` etc, which will then range-check:

- `AbstractProperty::setValue`
- `AbstractProperty::removeValueAtIndex`

But I don't think it will currently range-check:

- `AbstractProperty::clear`
- `operator=(PropertyWithNoRangeLimits const&)`
- `AbstractProperty::appendValue`
- `AbstractProperty::assign`

With `clear` being the one that the test exercises, but it's also sometimes necessary (e.g. in property editors) to do stuff like create a new property, only copy over the elements the user wants, and then assign the new property over the old one, which will delete the range check :<

adamkewley: I think `OpenSim_DECLARE_LIST_PROPERTY_SIZE` should maybe have more enforcement, but that's a wider change than this specific issue.

Apart from that, I have changed the check to be `== 2`, rather than `>= 2`, if that resolves this conversation?

tkuchida (approved):
LGTM :+1:

Issue #3386, opened 2023-02-03
Increase the number of input dimensions supported by `MultivariatePolynomialFunction`


nickbianco:
Needed to eventually support polynomial functions representing muscle-tendon lengths that depend on many degrees-of-freedom. 

For example, B.J. Fregly's wrapping code needs up to 6 coordinates for one polynomial function (psoas crossing the hip and lumbar joints), but `MultivariatePolynomialFunction` only supports 4 coordinates.

PR #3549, opened 2023-09-13
Increase the number of input dimensions in `MultivariatePolynomialFunction` to 6

nickbianco:
Fixes issue #3386

### Brief summary of changes
- Updated `MultivariatePolynomialFunction` to support 6 input dimensions, which includes updates to the constructor, `calcValue`, and `calcDerivative`.
- In the docs, I switched usages of "dependent" to "independent", which is the correct terminology for input values of a function.

### Testing I've completed
- Added a unit test to `testFunctions.cpp`.

### CHANGELOG.md (choose one)
- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3549)


OpenSim/Common/MultivariatePolynomialFunction.h
```diff
@@ -24,11 +24,11 @@
...
 
 /** A multivariate polynomial function.
-This implementation assumes a maximum of four input dimensions and allows
+This implementation assumes a maximum of six input dimensions and allows
 computation of first-order derivatives only.
 @param coefficients the polynomial coefficients in order of ascending
-powers starting from the last dependent component.
-For a polynomial of third order dependent on three components
+powers starting from the last independent component.
+For a polynomial of third order independent on three components
```
antoinefalisse, line 31: I think dependent was used in a different context here. Would keep it although I see how confusing it could be.

antoinefalisse (commented):
Looking good me, just a minor comment.

OpenSim/Common/MultivariatePolynomialFunction.h
nickbianco: Oh, good catch. I see why "dependent" was used now. I rephrased that comment for clarity. I'd like to keep the term "independent" everywhere else, I was getting `dimension` confused for the number of outputs.

PR #3550, opened 2023-09-13
Add `FunctionBasedPath`

nickbianco:
Fixes issue #3389

### Brief summary of changes
Added `FunctionBasedPath`, a concrete implementation of `AbstractPath` that uses `OpenSim::Function`s to compute path lengths, lengthening speeds, and moment arms. Users must provide a length function and the list of coordinates it depends on, and can optionally provide functions to compute moment arms and lengthening speed. If not provided, moment arms are calculated from the derivative of the length function with respect to the coordinate values. Similarly, lengthening speed will be calculated from the scalar product between the vector of moment arms and the vector of coordinate value derivatives, which is the result of the applying the chain rule to the path speed computation: `d(length)/dt = d(length)/dq * dq/dt`.

I say coordinate value derivatives and not coordinate speeds because `Coordinate::getSpeedValue()` reports the value from `SimTK::State::getU()` associated with the `Coordinate`, and not `SimTK::State::getQDot()`. For most joints, `qdot = u`, but in others (e.g., `BallJoint`, which models rotations via a quaternion), `qdot != u`. The chain rule math above suggests that we should use `qdot` to have an implementation that supports all joints. To accomodate this I've added `Coordinate::getQDotValue()`, which gets the value of `qdot` associated with the coordinate from the mobilized body in the `SimbodyMatterSubsystem`. 

Other changes:
- Provided default implementations for `AbstractPath::getColor()` and `AbstractPath::setColor()` (i.e., they are no longer pure virtual). 
- Added some missing overrides to `AbstractPath` and `GeometryPath`.
- Drive-by: updated CMake files to avoid dev warnings (inspired by #3545)
- Minor formatting updates.

### Testing I've completed
- Added unit tests to `testWrapping.cpp`.

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550)


.github/workflows/continuous_integration.yml
```diff
@@ -77,7 +77,7 @@ jobs:
         # TODO: Can remove /WX when we use that in CMakeLists.txt.
         # Set the CXXFLAGS environment variable to turn warnings into errors.
         # Skip timing test section included by default.
-        cmake -E env CXXFLAGS="/WX -DSKIP_TIMING" cmake $env:GITHUB_WORKSPACE -LAH -G"Visual Studio 16 2019" -A x64 -DCMAKE_INSTALL_PREFIX=~/opensim-core-install -DOPENSIM_DEPENDENCIES_DIR=~/opensim_dependencies_install -DOPENSIM_C3D_PARSER=ezc3d -DBUILD_PYTHON_WRAPPING=on -DBUILD_JAVA_WRAPPING=on -DPython3_ROOT_DIR=C:\hostedtoolcache\windows\Python\3.8.10\x64 
+        cmake -E env CXXFLAGS="/WX -DSKIP_TIMING" cmake $env:GITHUB_WORKSPACE -LAH -G"Visual Studio 16 2019" -A x64 -DCMAKE_INSTALL_PREFIX=~/opensim-core-install -DOPENSIM_DEPENDENCIES_DIR=~/opensim_dependencies_install -DOPENSIM_C3D_PARSER=ezc3d -DBUILD_PYTHON_WRAPPING=on -DBUILD_JAVA_WRAPPING=on -DPython3_ROOT_DIR=C:\hostedtoolcache\windows\Python\3.8.10\x64 -DPython3_FIND_STRATEGY=LOCATION
```
nickbianco, line 80: `-DPython3_FIND_STRATEGY` is needed after updating to `cmake_minimum_required(VERSION 3.5)`. Apparently, `cmake_minimum_required` changes some variables that change how CMake searches for Python. Setting this variable to `LOCATION` tells CMake to use the Python version that we point to via `Python3_ROOT_DIR`.

Applications/opensim-cmd/test/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 
 OpenSimAddTests(
-    TESTPROGRAMS testCommandLineInterface
+    TESTPROGRAMS testCommandLineInterface.cpp
```
nickbianco, line 3: This change suppresses a CMake warning that complains that the file extension is not explicit.

aymanhab (changes requested):
> Reviewed 5 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550)* status: 5 of 12 files reviewed, 3 unresolved discussions (waiting on @adamkewley and @nickbianco)

___
*[`.github/workflows/continuous_integration.yml` line 80 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3550#gh-1326545584:-NeidCF7AvfQDgMx34co:b-zz2le) ([raw file](https://github.com/opensim-org/opensim-core/blob/fa1f49ca6825fa0b756fe5d22c146914dd41dfdc/.github/workflows/continuous_integration.yml#L80)):*

`-DPython3_FIND_STRATEGY` is needed after updating to `cmake_minimum_required(VERSION 3.5)`. Apparently, `cmake_minimum_required` changes some variables that change how CMake searches for Python. Setting this variable to `LOCATION` tells CMake to use the Python version that we point to via `Python3_ROOT_DIR`.
</blockquote></details>

Python3 find strategy was introduced in 3.15 per https://cmake.org/cmake/help/latest/module/FindPython3.html not sure how it would work for earlier CMake versions since all my installations are more recent. Should we require 3.15?
___
*[`OpenSim/Simulation/Model/AbstractPath.h` line 204 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NeieB388cCBh1oNOvOm:-NeieB388cCBh1oNOvOn:ba25gwg) ([raw file](https://github.com/opensim-org/opensim-core/blob/8a4cac3373bd8e3be9183b8369bbc854d79bff5f/OpenSim/Simulation/Model/AbstractPath.h#L204)):*
> ```objc
>     virtual void setColor(
>             const SimTK::State& s, const SimTK::Vec3& color) const {};
> ```

Would you consider calling this method on AbstractPath an error and throwing instead, or do you conjecture some code shared by subclasses that should just ignore the call more gracefully?

aymanhab (changes requested):
> Reviewed 2 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550)* status: 7 of 12 files reviewed, 4 unresolved discussions (waiting on @adamkewley and @nickbianco)

___
*[`OpenSim/Simulation/SimbodyEngine/Coordinate.h` line 160 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NeifYrL5bDWGMeG-TML:-NeifYrL5bDWGMeG-TMM:b-1wx0si) ([raw file](https://github.com/opensim-org/opensim-core/blob/8a4cac3373bd8e3be9183b8369bbc854d79bff5f/OpenSim/Simulation/SimbodyEngine/Coordinate.h#L160)):*
> ```objc
>         *not* necessarily equal to the value returned by getSpeedValue(). */
>     double getQDotValue(const SimTK::State& s) const;
> ```

I vaguely remember that some joints have the option to use euler angles or quaternions internally and the selection was done at creation time. Does this change account for both options?

nickbianco:
@tkuchida, do you have time/interest in reviewing this?

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550)* status: 7 of 12 files reviewed, 4 unresolved discussions (waiting on @aymanhab)

___
*[`.github/workflows/continuous_integration.yml` line 80 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3550#gh-1326545584:-NeimVdc9QmFqBh3-jLj:b-n055st) ([raw file](https://github.com/opensim-org/opensim-core/blob/fa1f49ca6825fa0b756fe5d22c146914dd41dfdc/.github/workflows/continuous_integration.yml#L80)):*

Python3 find strategy was introduced in 3.15 per https://cmake.org/cmake/help/latest/module/FindPython3.html not sure how it would work for earlier CMake versions since all my installations are more recent. Should we require 3.15?
</blockquote></details>

Whoops, I didn't mean to downgrade CMake to 3.5 from 3.15 (my brain interpreted the version as a floating point number). I'll bring the version back up to 3.15 and see if we still need these CMake variables.

aymanhab (changes requested):
> Reviewed 1 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550)* status: 6 of 12 files reviewed, 5 unresolved discussions (waiting on @nickbianco)

___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 174 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NeioaMDByItL41DUWm3:-NeioaMDByItL41DUWm4:b-vvnzpn) ([raw file](https://github.com/opensim-org/opensim-core/blob/8a4cac3373bd8e3be9183b8369bbc854d79bff5f/OpenSim/Simulation/Model/FunctionBasedPath.h#L174)):*
> ```objc
>     /// the order of the coordinates.
>     void setMomentArmFunctions(const std::vector<Function>& momentArmFunctions);
>     void appendMomentArmFunction(const Function& momentArmFunction);
> ```

Maybe this is due to other instances of the API where ownership was in question. Can you clarify if a copy of the function is made here or ownership is transferred somehow? The convention has been that models own all objects in the component tree. At least I'd document the current behavior here. Also we've been moving away from using names/lookup to higher level connectors but this may be overly complicating at this point. Not sure if name by itself is unique without a full path

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550)* status: 6 of 12 files reviewed, 4 unresolved discussions (waiting on @aymanhab)

___
*[`OpenSim/Simulation/Model/AbstractPath.h` line 204 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NeieB388cCBh1oNOvOm:-Nej_M-238cUmQ3lfc0S:b-u7uxar) ([raw file](https://github.com/opensim-org/opensim-core/blob/8a4cac3373bd8e3be9183b8369bbc854d79bff5f/OpenSim/Simulation/Model/AbstractPath.h#L204)):*

Would you consider calling this method on AbstractPath an error and throwing instead, or do you conjecture some code shared by subclasses that should just ignore the call more gracefully?
</blockquote></details>

I was considering cases like this: https://github.com/opensim-org/opensim-core/blob/e8e5ddf3c9282d45b795dbe10f11143988afe76c/OpenSim/Simulation/Model/PathActuator.cpp#L219

So yes, this implementation tries to ignore the call gracefully. Ideally, `get/setColor()` should not be called `isVisualPath()` in derived classes of `AbstractPath` return `false`. But we could try to bake this design into the API more explicitly.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 174 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NeioaMDByItL41DUWm3:-Neja2qOBL53367J8apv:bm3uxep) ([raw file](https://github.com/opensim-org/opensim-core/blob/8a4cac3373bd8e3be9183b8369bbc854d79bff5f/OpenSim/Simulation/Model/FunctionBasedPath.h#L174)):*
Both of these make copies. Actually, `appendMomentArmFunction()` could be updated to use `std::move`. I can add documentation for this.

> Also we've been moving away from using names/lookup to higher level connectors but this may be overly complicating at this point. Not sure if name by itself is unique without a full path

These functions will be stored in the `moment_arm_function` property for this object, so the path should be unique. Unless, I'm misunderstanding you here.
___
*[`OpenSim/Simulation/SimbodyEngine/Coordinate.h` line 160 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NeifYrL5bDWGMeG-TML:-NejdT4-0BYyqomvgLDO:bbrh689) ([raw file](https://github.com/opensim-org/opensim-core/blob/8a4cac3373bd8e3be9183b8369bbc854d79bff5f/OpenSim/Simulation/SimbodyEngine/Coordinate.h#L160)):*

I vaguely remember that some joints have the option to use euler angles or quaternions internally and the selection was done at creation time. Does this change account for both options?
</blockquote></details>

This should always return the correct value whether you use Euler angles or quaternions. It will return a value from the `qdot` vector, which is computed via the kinematic differential equations `qdot = N u`. Most of the time though, `qdot = u`, but users must be careful that the don't fit functions assuming we use `u` in every case.

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550)* status: 6 of 12 files reviewed, 4 unresolved discussions (waiting on @aymanhab)

___
*[`OpenSim/Simulation/Model/AbstractPath.h` line 204 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NeieB388cCBh1oNOvOm:-NenKOFI55gWH9b8KlmO:bge18ha) ([raw file](https://github.com/opensim-org/opensim-core/blob/8a4cac3373bd8e3be9183b8369bbc854d79bff5f/OpenSim/Simulation/Model/AbstractPath.h#L204)):*

I was considering cases like this: https://github.com/opensim-org/opensim-core/blob/e8e5ddf3c9282d45b795dbe10f11143988afe76c/OpenSim/Simulation/Model/PathActuator.cpp#L219

So yes, this implementation tries to ignore the call gracefully. Ideally, `get/setColor()` should not be called `isVisualPath()` in derived classes of `AbstractPath` return `false`. But we could try to bake this design into the API more explicitly.
</blockquote></details>

*if `isVisualPath()` returns `false`

aymanhab (changes requested):
Looks close, a few minor comments and should be good to go for me.

> Reviewed 3 of 7 files at r2, 2 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-:-NeokOE5CljYUpyUUttc:b-l71zun)* status: all files reviewed, 5 unresolved discussions (waiting on @nickbianco)

___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 174 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NeioaMDByItL41DUWm3:-Neocxa16roaaw_lR0go:b-f4nga7) ([raw file](https://github.com/opensim-org/opensim-core/blob/8a4cac3373bd8e3be9183b8369bbc854d79bff5f/OpenSim/Simulation/Model/FunctionBasedPath.h#L174)):*

Both of these make copies. Actually, `appendMomentArmFunction()` could be updated to use `std::move`. I can add documentation for this.

> Also we've been moving away from using names/lookup to higher level connectors but this may be overly complicating at this point. Not sure if name by itself is unique without a full path

These functions will be stored in the `moment_arm_function` property for this object, so the path should be unique. Unless, I'm misunderstanding you here.
</blockquote></details>

I guess you're saying the coordinate name will be unique in the context, which is reasonable.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 132 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NeojqL7BLWziAZ3sUG9:-NeojqL7BLWziAZ3sUGA:bcrirj2) ([raw file](https://github.com/opensim-org/opensim-core/blob/a31c090107db3a62696b11e6cc88f0f3f34f1036/OpenSim/Simulation/Model/FunctionBasedPath.h#L132)):*
> ```objc
>             "used as arguments to the path function(s).");
>     OpenSim_DECLARE_OPTIONAL_PROPERTY(length_function, Function, 
>             "The OpenSim::Function object that computes the length of the path "
> ```

I assume there's a practical reason for this being called OPTIONAL?
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 63 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NeojDF11NK8KhVpHb9R:-NeojDF11NK8KhVpHb9S:b-yp1ama) ([raw file](https://github.com/opensim-org/opensim-core/blob/a31c090107db3a62696b11e6cc88f0f3f34f1036/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L63)):*
> ```cpp
> {
>     std::vector<std::string> coordinates;
>     for (int i = 0; i < getProperty_coordinates().size(); ++i) {
> ```

I'd name the variable coordinateNames as opposed to coordinates, you may consider the same for the Properties.

aymanhab (changes requested):
> Reviewed 1 of 1 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550)* status: all files reviewed, 4 unresolved discussions (waiting on @nickbianco)

___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 63 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NeojDF11NK8KhVpHb9R:-Net1g6q919mgJS2VnwO:bguysat) ([raw file](https://github.com/opensim-org/opensim-core/blob/b73f2e5267e3bc230d2669a9b84973d331bbe8e3/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L63)):*

I'd name the variable coordinateNames as opposed to coordinates, you may consider the same for the Properties.
</blockquote></details>

@nickbianco I know we used coordinates in XML files rather liberally but at least internally we should try to make clear what types of objects are returned by a method, especially for scripting users that do not have access to full signatures

nickbianco (commented):
Thanks @aymanhab! Ready for review again. Let me know what you think about the test patch.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-:-NeyF0S06ranhAirJdxd:b-j07evp)* status: 12 of 16 files reviewed, 4 unresolved discussions (waiting on @aymanhab)

___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 174 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NeioaMDByItL41DUWm3:-NeyEgjq52vv0U-Q68gd:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/8a4cac3373bd8e3be9183b8369bbc854d79bff5f/OpenSim/Simulation/Model/FunctionBasedPath.h#L174)):*

I guess you're saying the coordinate name will be unique in the context, which is reasonable.
</blockquote></details>

Done.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 132 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NeojqL7BLWziAZ3sUG9:-NeyF6E136NKZZYKW4c4:b8ylkaw) ([raw file](https://github.com/opensim-org/opensim-core/blob/b73f2e5267e3bc230d2669a9b84973d331bbe8e3/OpenSim/Simulation/Model/FunctionBasedPath.h#L132)):*

I assume there's a practical reason for this being called OPTIONAL?
</blockquote></details>

If it's not optional, it requires a default value. There's not really a good reasonable default (can't be `Function` because that's abstract). Also, the function type will depend on the number of coordinates, etc.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 63 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NeojDF11NK8KhVpHb9R:-NeyEhtEEIcpmtUye432:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/b73f2e5267e3bc230d2669a9b84973d331bbe8e3/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L63)):*

@nickbianco I know we used coordinates in XML files rather liberally but at least internally we should try to make clear what types of objects are returned by a method, especially for scripting users that do not have access to full signatures
</blockquote></details>

Done.

aymanhab (changes requested):
Very close, just minor questions regarding the optional property and the test.

> Reviewed 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-:-NezPJFe5r6njyu5qICv:b-zb2mym)* status: all files reviewed, 3 unresolved discussions (waiting on @nickbianco)

___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 132 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NeojqL7BLWziAZ3sUG9:-NezLip2DOIzNeY4H5ta:b-8v7s7b) ([raw file](https://github.com/opensim-org/opensim-core/blob/b73f2e5267e3bc230d2669a9b84973d331bbe8e3/OpenSim/Simulation/Model/FunctionBasedPath.h#L132)):*

If it's not optional, it requires a default value. There's not really a good reasonable default (can't be `Function` because that's abstract). Also, the function type will depend on the number of coordinates, etc.
</blockquote></details>

Great, thank you, I'd just clarify in the comment that it's not really "optional". Would it make sense if you can set it to a constant function (e.g. NaN) to make sure if used inadvertently without setting it blows up.
___
*[`OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp` line 94 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NezO-mI7mW9iAC6yVlK:-NezO-mI7mW9iAC6yVlL:bov6kc3) ([raw file](https://github.com/opensim-org/opensim-core/blob/db2645e81b7fd1db3178279463816dc089c14bb1/OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp#L94)):*
> ```cpp
>             Object* clone = availableComponentTypes[i]->clone();
>             Object* randClone;
>             // FunctionBasedPath is a special case: it has specific requirements
> ```

I guess I'm not clear why this would be different from other components, what goes wrong with serialization if you don't setCoordinatePaths and/or lengthFunction? That may point us to how to avoid the special treatment.

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550)* status: 14 of 16 files reviewed, 3 unresolved discussions (waiting on @aymanhab)

___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 132 at r4](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NeojqL7BLWziAZ3sUG9:-NfCIl8L9ZSBs7UFVuY4:b-vamjca) ([raw file](https://github.com/opensim-org/opensim-core/blob/b73f2e5267e3bc230d2669a9b84973d331bbe8e3/OpenSim/Simulation/Model/FunctionBasedPath.h#L132)):*

Great, thank you, I'd just clarify in the comment that it's not really "optional". Would it make sense if you can set it to a constant function (e.g. NaN) to make sure if used inadvertently without setting it blows up.
</blockquote></details>

Done. There's lots of error checking in `extendFinalizeFromProperties` that prevents bad configurations.
___
*[`OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp` line 94 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NezO-mI7mW9iAC6yVlK:-NfCKJfrBWlBODDVWPUp:b-aux16t) ([raw file](https://github.com/opensim-org/opensim-core/blob/db2645e81b7fd1db3178279463816dc089c14bb1/OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp#L94)):*

I guess I'm not clear why this would be different from other components, what goes wrong with serialization if you don't setCoordinatePaths and/or lengthFunction? That may point us to how to avoid the special treatment.
</blockquote></details>

The issue is due to the specific error checking I do in `extendFinalizeFromProperties`. The `length_function` must have the same number of arguments as coordinates in `coordinate_paths`, which doesn't necessarily happen in the serialization code in the test. I'm not sure how to address it in a general way; that test is pretty messy, I'd be inclined to rewrite it anyway.

aymanhab (approved):
Thanks @nickbianco I'd just revert the test change and move the code elsewhere, otherwise LGTM
:lgtm:

> Reviewed 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-:-NfCdOU0AO_F1_itD1l0:brcygaq)* status: all files reviewed, 2 unresolved discussions (waiting on @nickbianco)

___
*[`OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp` line 94 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NezO-mI7mW9iAC6yVlK:-NfCbMcP1RVHNS10C664:b-lv9o7r) ([raw file](https://github.com/opensim-org/opensim-core/blob/db2645e81b7fd1db3178279463816dc089c14bb1/OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp#L94)):*

The issue is due to the specific error checking I do in `extendFinalizeFromProperties`. The `length_function` must have the same number of arguments as coordinates in `coordinate_paths`, which doesn't necessarily happen in the serialization code in the test. I'm not sure how to address it in a general way; that test is pretty messy, I'd be inclined to rewrite it anyway.
</blockquote></details>

My recollection is that this test strictly tests serialization/deserialization with random numbers so no validation of anything else is done, accordingly object specific logic should be removed from here and moved into the specific class test as done for similar classes/objects. I agree it may need a rewrite but as of now I'd keep it thin and naive :) and at the Property serialization level

adamkewley (changes requested):
Good work: I spotted some mostly minor nitpicks etc. as I read through it

> Reviewed 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-:-NfFTCNF722vT4Vh0BP2:bpe6nys)* status: all files reviewed, 23 unresolved discussions (waiting on @aymanhab and @nickbianco)

___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 29 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFBPMs37aEWP1ficpc:-NfFBPMs37aEWP1ficpd:b-icvixq) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L29)):*
> ```objc
> #include "OpenSim/Common/Function.h"
> 
> #ifdef SWIG
> ```

This macro dance shouldn't be necessary? Other `Simulation` classes seem to import `osimSimulationDLL.h` and use the macro as-is?
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 42 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFBmE549VjiZHa4NSx:-NfFBmE549VjiZHa4NSy:b-kr2j44) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L42)):*
> ```objc
> /**
>  * A concrete class representing a path for muscles, ligaments, etc., based on 
> ```

Trailing whitespace
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 123 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFBxes3HG-XjIRGRDU:-NfFBxes3HG-XjIRGRDV:bdzzdcy) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L123)):*
> ```objc
> class OSIMSIMULATION_API FunctionBasedPath : public AbstractPath {
> OpenSim_DECLARE_CONCRETE_OBJECT(FunctionBasedPath, AbstractPath);
>     
> ```

Whitespace
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 132 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFC9aX4QXv4wMpfXrE:-NfFC9aX4QXv4wMpfXrF:b-fl3msk) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L132)):*
> ```objc
>             "arguments to the length, lengthening speed, and moment arm "
>             "functions. You must provide a least one coordinate path, and "
>             "the order of the paths must match the order of the function "
> ```

"... match the order of the arguments passed to `length_function` (and `moment_arm_functions` and `lengthening_speed_function`?)
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 141 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFDDKeBHL3_JtVeg9S:-NfFDDKeBHL3_JtVeg9T:b-mr69j6) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L141)):*
> ```objc
>             "(Optional) The list of OpenSim::Function objects that compute the " 
>             "moment arms of the path as a function of the coordinate values. "
>             "The function arguments must match the order in the 'coordinates' "
> ```

`coordinate_paths`
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 155 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFDRTN-DaHtYraKP2s:-NfFDRTN-DaHtYraKP2t:b-rmw46a) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L155)):*
> ```objc
>     
>     // CONSTRUCTION AND DESTRUCTION
>     FunctionBasedPath();
> ```

Rule-of-five (even if you just `=default;` the other 3)
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 165 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFDsbB3yA1FBjv8j3H:-NfFDsbB3yA1FBjv8j3I:b-yvx4y2) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L165)):*
> ```objc
>     void setCoordinatePaths(const std::vector<std::string>& coordinatePaths);
>     void appendCoordinatePath(const std::string& coordinatePath);
>     /// @details Note: the return value is constructed fresh on every call from 
> ```

Imo, not `Note`-worthy: the C++ type system is already saying it
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 172 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFEOxE2ebaPPMdolKV:-NfFEOxE2ebaPPMdolKW:b-pbevab) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L172)):*
> ```objc
>     /// the coordinate values. The function must have the same number of 
>     /// arguments as the number of coordinates.
>     /// @note The length function is copied into the internal property.
> ```

The C++ type-system is already saying it

(the only alternative would be that the implementation saves a pointer to the argument, but that's very bad form and should be replaced with a pointer/smart-pointer arg)
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 179 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFEgnCEbFhZ9m4Qk2x:-NfFEgnCEbFhZ9m4Qk2y:b-h2xc4c) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L179)):*
> ```objc
>     /// function of the coordinate values. The order of the functions must match
>     /// the order of the coordinates.
>     /// @note The moment arm functions are copied into the internal property.
> ```

(C++ type system etc.)
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 184 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFEsFnB6DIiMMe_pap:-NfFEsFnB6DIiMMe_paq:b-o7kukl) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L184)):*
> ```objc
>     const Function& getMomentArmFunction(
>             const std::string& coordinatePath) const;
>     const SimTK::Vector& getMomentArms(const SimTK::State& s) const;
> ```

The signature of this function rings alarm bells (hint: it's a `const` member method that returns values that are dependent on the const `State` argument, but returns a _reference_ to a vector of values - something somewhere is mutating)
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 189 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFFEn72GX0UjTsYiVA:-NfFFEn84-LHow1RULQj:b2relbn) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L189)):*
> ```objc
>     /// the coordinate values and speeds. The function must have the same number 
>     /// of arguments as the number of coordinate values and speeds.
>     /// @note The length function is copied into the internal property.
> ```

C++ type system
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 224 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFGIclD4KnUgADSjEV:-NfFGIclD4KnUgADSjEW:by19kcu) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L224)):*
> ```objc
>     // MEMBER VARIABLES
>     std::vector<SimTK::ReferencePtr<const Coordinate>> _coordinates;
>     std::map<std::string, int> _coordinateIndices;
> ```

Is ordering, or reference stability, of this lookup _required_? If not, then prefer `std::unordered_map`
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 228 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFFtZSEMrCLdaeGO6g:-NfFFtZSEMrCLdaeGO6h:b-f5vy8o) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L228)):*
> ```objc
>     bool _computeLengtheningSpeed = false;
>     
>     static const std::string LENGTH_NAME;
> ```

Do these need to be _member_ variables (albeit, static)?

You could hide them inside the `cpp` file as `static const std::string c_MomentArmsName = "moment-arms"` or similar (or, ideally, in an anonymous namespace) or similar.

And (imo) they shouldn't use a naming convention that's identical to a preprocessor macro.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 29 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFGVYI-sb4XrmGfsGu:-NfFGVYI-sb4XrmGfsGv:b-b796fm) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L29)):*
> ```cpp
> using namespace OpenSim;
> 
> const std::string FunctionBasedPath::LENGTH_NAME("length");
> ```

These can be defined here without the necessity to declare them in the header file
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 42 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFGf84EwVxJJUQhfHX:-NfFGf84EwVxJJUQhfHY:b6cecu2) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L42)):*
If you're just going to de-inline the dtor, then you can skip the rule-of-five suggestion above by not declaring a user-defined destructor in the header:

https://en.cppreference.com/w/cpp/language/rule_of_three

> If a class requires a user-defined [destructor](https://en.cppreference.com/w/cpp/language/destructor "cpp/language/destructor"), a user-defined [copy constructor](https://en.cppreference.com/w/cpp/language/copy_constructor "cpp/language/copy constructor"), or a user-defined [copy assignment operator](https://en.cppreference.com/w/cpp/language/as_operator "cpp/language/as operator"), it almost certainly requires all three.

(therefore, if you don't define any of those, you don't need to do anything)
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 62 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFH8j0EvWA1neTggO9:-NfFH8j0EvWA1neTggOA:b-mncfs8) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L62)):*
> ```cpp
> std::vector<std::string> FunctionBasedPath::getCoordinatePaths() const 
> {
>     std::vector<std::string> coordinates;
> ```

If your concern was performance, then `.reserve(getProperty_coordinate_paths().size())` would help reduce unnecessary moves, over-allocation, etc.

(even better: provide an indexed API to get each path one-at-a-time, so that the *caller* can decide how they want to store their paths)

(even betterererer: provide a slurp API for OpenSim properties: `template<typename T> std::vector<T> slurp(OpenSim::Property<T> const&);`)
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 113 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFI6vf87U0PiyXFEv5:-NfFI6vf87U0PiyXFEv6:b-jkxgma) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L113)):*
> ```cpp
>         const SimTK::State& s) const {
>     computeMomentArms(s);
>     return getCacheVariableValue<SimTK::Vector>(s, MOMENT_ARMS_NAME);
> ```

Ah, so my suspicion was related to the fact that `SimTK`'s cache variables are secretly-mutable data :wink:
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 122 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFLCyD7FGySkNWW_yr:-NfFLCyD7FGySkNWW_ys:b7tri0e) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L122)):*
> ```cpp
> {
>     computeLength(s);
>     return getCacheVariableValue<double>(s, LENGTH_NAME); 
> ```

Note: the `OpenSim::CacheVariable<T>` API was specifically designed to avoid having to these kinds of runtime string lookups
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 128 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFIUvZDHDrWueuzepF:-NfFIUvZDHDrWueuzepG:b-wk80a9) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L128)):*
> ```cpp
>         const Coordinate& coord) const 
> {
>     if (_coordinateIndices.find(coord.getAbsolutePathString()) != 
> ```

This doesn't seem like a wise way to write this: you're effectively eating two (O(log n), because it's a std::map) lookups where one would do, and calling `getAbsolutePathString` (expensive) twice.

You could save the iterator and read out the index when required: `auto it = _coordinateIndices.find(coord.getAbsolutePathString());`
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 134 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFIvra-p4gjse-KiiV:-NfFIvrb2YOeErzz9ebk:bdzzdcy) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L134)):*
> ```cpp
>                 .get(_coordinateIndices.at(coord.getAbsolutePathString()));
>     } else { 
>         return 0.0; 
> ```

Whitespace
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 155 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFJGMR-KmowWENZLEU:-NfFJGMR-KmowWENZLEV:b-9t35d1) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L155)):*
> ```cpp
>     
>     // Apply the mobility forces.
>     const SimTK::SimbodyMatterSubsystem& matter = 
> ```

`OPENSIM_ASSERT_ALWAYS(momentArms.size() == _coordinates.size()): you're using the assumption that they are of equal size to use a common index--because you happen to know `computeMomentArms` behaves perfectly this way--but I'd ensure that's *always* true, because there's the off-chance of states being mixed around, a cache miss not quite being detected, etc., and you're risking those underlying problems surfacing as a bizzare runtime issue if they aren't bounds-checked.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 181 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFK59BEyqfON3eBrSa:-NfFK59BEyqfON3eBrSb:b237bp5) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L181)):*
> ```cpp
>         const SimTK::State& s) const 
> {
>     SimTK::Vector coordinateValues((int)_coordinates.size(), 0.0);
> ```

If `SimTK::Vector` supports it, prefer allocating an empty vector, then `.reserve`, then `.push_back`, because it is fast and bounds-safe, while not requiring the `Vector` implementation to `memset` (or simliar) `.size()` elements to `0.0` first
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 185 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFKXr4027YlzNr9bIO:-NfFKXr4027YlzNr9bIP:btptfmp) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L185)):*
> ```cpp
>         coordinateValues[i] = _coordinates[i]->getValue(s);
>     }
>     
> ```

Whitespace

I'd recommend running a script or similar to wipe all whitespace from new files. I routinely end up running a custom script on `opensim-creator` so that I can still write code with whitespace (if the editor is a bit crap), followed by cleaning it before committing:

https://github.com/ComputationalBiomechanicsLab/opensim-creator/blob/main/scripts/format_sources.py
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 233 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFLsO_EhUYBp6CLCpF:-NfFLsO_EhUYBp6CLCpG:bi9ty9w) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L233)):*
> ```cpp
>         }
>     }
>     setCacheVariableValue(s, MOMENT_ARMS_NAME, momentArms);
> ```

Explore the `OpenSim::CacheVariable<T>` API to avoid a string lookup etc.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 282 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFRjFG4i28yw_PK0vP:-NfFRjFG4i28yw_PK0vQ:b-fq3q94) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L282)):*
> ```cpp
>     
>     // Check that the `coordinate_paths` are unique.
>     std::set<std::string> uniqueCoordinatePaths;
> ```

Do you need an _ordered_ set? If not, use `std::unordered_set`

And use `.reserve` if you know the probable size of the set ahead of time

nickbianco (commented):
@aymanhab I can't revert the test change without `testSerializeOpenSimObjects.cpp` failiing. We either have to leave it as-is, or skip the test entirely for `FunctionBasedPath`.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-:-NfH8s8D5kVibacEDspU:bpgrynj)* status: all files reviewed, 23 unresolved discussions (waiting on @adamkewley and @aymanhab)

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550)* status: all files reviewed, 23 unresolved discussions (waiting on @adamkewley and @aymanhab)

___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 29 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFBPMs37aEWP1ficpc:-NfH9OIrBpa3eKvoywvo:b-uawwoo) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L29)):*

This macro dance shouldn't be necessary? Other `Simulation` classes seem to import `osimSimulationDLL.h` and use the macro as-is?
</blockquote></details>

Probably not. I'll test the bindings locally to be sure before we merge.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 141 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFDDKeBHL3_JtVeg9S:-NfHAtWF5kZiC5fo-g-q:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L141)):*

`coordinate_paths`
</blockquote></details>

Done.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 155 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFDRTN-DaHtYraKP2s:-NfHCBNR9GudJBcmE174:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L155)):*

Rule-of-five (even if you just `=default;` the other 3)
</blockquote></details>

Done.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 165 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFDsbB3yA1FBjv8j3H:-NfHB3A60v5zkOorpICa:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L165)):*

Imo, not `Note`-worthy: the C++ type system is already saying it
</blockquote></details>

Done.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 172 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFEOxE2ebaPPMdolKV:-NfHB4rjDNVh7jCU_MAb:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L172)):*

The C++ type-system is already saying it

(the only alternative would be that the implementation saves a pointer to the argument, but that's very bad form and should be replaced with a pointer/smart-pointer arg)
</blockquote></details>

Done.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 179 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFEgnCEbFhZ9m4Qk2x:-NfHB5zV1U7oHhJ8Dqwm:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L179)):*

(C++ type system etc.)
</blockquote></details>

Done.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 189 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFFEn72GX0UjTsYiVA:-NfHB7hL1FRRkG5qIAzB:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L189)):*

C++ type system
</blockquote></details>

Done.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 224 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFGIclD4KnUgADSjEV:-NfHBBI8A8zHzam5N77y:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L224)):*

Is ordering, or reference stability, of this lookup _required_? If not, then prefer `std::unordered_map`
</blockquote></details>

Done.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 228 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFFtZSEMrCLdaeGO6g:-NfHBHr_CaQU1tez3-hq:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L228)):*

Do these need to be _member_ variables (albeit, static)?

You could hide them inside the `cpp` file as `static const std::string c_MomentArmsName = "moment-arms"` or similar (or, ideally, in an anonymous namespace) or similar.

And (imo) they shouldn't use a naming convention that's identical to a preprocessor macro.
</blockquote></details>

Done.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 29 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFGVYI-sb4XrmGfsGu:-NfHBilW4Fuv-JaQCpwQ:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L29)):*

These can be defined here without the necessity to declare them in the header file
</blockquote></details>

Done.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 42 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFGf84EwVxJJUQhfHX:-NfHC5bZEq1BICjdYpm3:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L42)):*

If you're just going to de-inline the dtor, then you can skip the rule-of-five suggestion above by not declaring a user-defined destructor in the header:

https://en.cppreference.com/w/cpp/language/rule_of_three

> If a class requires a user-defined [destructor](https://en.cppreference.com/w/cpp/language/destructor "cpp/language/destructor"), a user-defined [copy constructor](https://en.cppreference.com/w/cpp/language/copy_constructor "cpp/language/copy constructor"), or a user-defined [copy assignment operator](https://en.cppreference.com/w/cpp/language/as_operator "cpp/language/as operator"), it almost certainly requires all three.

(therefore, if you don't define any of those, you don't need to do anything)
</blockquote></details>

Done.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 134 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFIvra-p4gjse-KiiV:-NfHCQyO2ob9h1w0G98a:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L134)):*

Whitespace
</blockquote></details>

Done.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 155 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFJGMR-KmowWENZLEU:-NfHDLowClMIY9U-96wi:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L155)):*

`OPENSIM_ASSERT_ALWAYS(momentArms.size() == _coordinates.size()): you're using the assumption that they are of equal size to use a common index--because you happen to know `computeMomentArms` behaves perfectly this way--but I'd ensure that's *always* true, because there's the off-chance of states being mixed around, a cache miss not quite being detected, etc., and you're risking those underlying problems surfacing as a bizzare runtime issue if they aren't bounds-checked.
</blockquote></details>

Done.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 185 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFKXr4027YlzNr9bIO:-NfHDQHHB-iyVov9-SqG:buyxgqv) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L185)):*

Whitespace

I'd recommend running a script or similar to wipe all whitespace from new files. I routinely end up running a custom script on `opensim-creator` so that I can still write code with whitespace (if the editor is a bit crap), followed by cleaning it before committing:

https://github.com/ComputationalBiomechanicsLab/opensim-creator/blob/main/scripts/format_sources.py
</blockquote></details>

I had this disabled in CLion for some reason. Fixed now :)
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 282 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFRjFG4i28yw_PK0vP:-NfHFVkp-d9n6PlPb4eD:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L282)):*

Do you need an _ordered_ set? If not, use `std::unordered_set`

And use `.reserve` if you know the probable size of the set ahead of time
</blockquote></details>

Done.
___
*[`OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp` line 94 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NezO-mI7mW9iAC6yVlK:-NfHFaPb9fNDtFIrHGRl:b-ta0oz) ([raw file](https://github.com/opensim-org/opensim-core/blob/db2645e81b7fd1db3178279463816dc089c14bb1/OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp#L94)):*

My recollection is that this test strictly tests serialization/deserialization with random numbers so no validation of anything else is done, accordingly object specific logic should be removed from here and moved into the specific class test as done for similar classes/objects. I agree it may need a rewrite but as of now I'd keep it thin and naive :) and at the Property serialization level
</blockquote></details>

See my note above.

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550)* status: 14 of 16 files reviewed, 22 unresolved discussions (waiting on @adamkewley and @aymanhab)

___
*[`OpenSim/Simulation/Model/FunctionBasedPath.h` line 184 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFEsFnB6DIiMMe_pap:-NfHpSJyDos6dlYx7tm2:bcyigrh) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.h#L184)):*

The signature of this function rings alarm bells (hint: it's a `const` member method that returns values that are dependent on the const `State` argument, but returns a _reference_ to a vector of values - something somewhere is mutating)
</blockquote></details>

Done. (I think)
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 62 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFH8j0EvWA1neTggO9:-NfHZpWc8_LowsEqPo8T:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L62)):*

If your concern was performance, then `.reserve(getProperty_coordinate_paths().size())` would help reduce unnecessary moves, over-allocation, etc.

(even better: provide an indexed API to get each path one-at-a-time, so that the *caller* can decide how they want to store their paths)

(even betterererer: provide a slurp API for OpenSim properties: `template<typename T> std::vector<T> slurp(OpenSim::Property<T> const&);`)
</blockquote></details>

Done.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 113 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFI6vf87U0PiyXFEv5:-NfHpWhnBGMa3xBDAVc0:bcyigrh) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L113)):*

Ah, so my suspicion was related to the fact that `SimTK`'s cache variables are secretly-mutable data :wink:
</blockquote></details>

Done. (I think)
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 122 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFLCyD7FGySkNWW_yr:-NfHpQJ892HTgiL6f_3p:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L122)):*

Note: the `OpenSim::CacheVariable<T>` API was specifically designed to avoid having to these kinds of runtime string lookups
</blockquote></details>

Done.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 128 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFIUvZDHDrWueuzepF:-NfHf5zj8RcOUQiY2ija:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L128)):*

This doesn't seem like a wise way to write this: you're effectively eating two (O(log n), because it's a std::map) lookups where one would do, and calling `getAbsolutePathString` (expensive) twice.

You could save the iterator and read out the index when required: `auto it = _coordinateIndices.find(coord.getAbsolutePathString());`
</blockquote></details>

Done.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 181 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFK59BEyqfON3eBrSa:-NfHoHorA9TZLg0JyXgG:bgyo66i) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L181)):*

If `SimTK::Vector` supports it, prefer allocating an empty vector, then `.reserve`, then `.push_back`, because it is fast and bounds-safe, while not requiring the `Vector` implementation to `memset` (or simliar) `.size()` elements to `0.0` first
</blockquote></details>

As far as I can tell, `SimTK::Vector` does not support this, unfortunately.
___
*[`OpenSim/Simulation/Model/FunctionBasedPath.cpp` line 233 at r9](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NfFLsO_EhUYBp6CLCpF:-NfHp7fa-JUOxq-3TT4L:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/942bff65f985257711b6faaaaee6da2126a59a7d/OpenSim/Simulation/Model/FunctionBasedPath.cpp#L233)):*

Explore the `OpenSim::CacheVariable<T>` API to avoid a string lookup etc.
</blockquote></details>

Done.

nickbianco (commented):
Thanks @adamkewley! Ready for review again. Let me know if you're happy with `getMomentArms` as-is, or if you think we should return a copy of the `SimTK::Vector` to avoid any `const_cast` shenanigans with the cache variable.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-:-NfHpwVBE6Kq_aE4J7Rc:b-1d5qpt)* status: 14 of 16 files reviewed, 22 unresolved discussions (waiting on @adamkewley and @aymanhab)

adamkewley:
> Thanks @adamkewley! Ready for review again. Let me know if you're happy with `getMomentArms` as-is, or if you think we should return a copy of the `SimTK::Vector` to avoid any `const_cast` shenanigans with the cache variable.

I'll review it tomorrow, but to directly address your question, it's up to you: the reason that the newer `CacheVariable<T>` API essentially forces you to use `mutable` is to make it clearer which parts of a downstream class might skirt around `const-` rules (e.g. if I see it in the data member list, I'm automatically thinking "`const` might be unusual here")

nickbianco:
@adamkewley, after poking around the API a bit, it seems like the current convention for `SimTK::Vector` cache variables is to return as a `const` reference, so I'll leave it as-is.

aymanhab (changes requested):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550)* status: 14 of 16 files reviewed, 22 unresolved discussions (waiting on @adamkewley and @nickbianco)

___
*[`OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp` line 94 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NezO-mI7mW9iAC6yVlK:-NfX3dpN7gFbtSvMfS5g:bwo00yl) ([raw file](https://github.com/opensim-org/opensim-core/blob/db2645e81b7fd1db3178279463816dc089c14bb1/OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp#L94)):*

See my note above.
</blockquote></details>

@nickbianco again I don't get why this class needs special treatment, anything more than serialization, including validation doesn't belong here, does the clone method do something unexpected? Just trying to avoid class specific custom code in tests that operate at the component/object/property level as it doesn't scale.

adamkewley (commented):
> Reviewed 1 of 2 files at r10, 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550)* status: all files reviewed, 1 unresolved discussion (waiting on @nickbianco)

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550)* status: all files reviewed, 1 unresolved discussion (waiting on @aymanhab)

___
*[`OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp` line 94 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NezO-mI7mW9iAC6yVlK:-NflH7BNEhu9qnF_DGfR:bdpblj6) ([raw file](https://github.com/opensim-org/opensim-core/blob/db2645e81b7fd1db3178279463816dc089c14bb1/OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp#L94)):*

@nickbianco again I don't get why this class needs special treatment, anything more than serialization, including validation doesn't belong here, does the clone method do something unexpected? Just trying to avoid class specific custom code in tests that operate at the component/object/property level as it doesn't scale.
</blockquote></details>

@aymanhab, `FunctionBasedPath` gets automatically included in this test, so I need to do something to prevent failure. The two options are 1) what I've drafted here or 2) manually skip the test for `FunctionBasedPath`. Reverting entirely will lead to more failures.

aymanhab (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550)* status: all files reviewed, 1 unresolved discussion (waiting on @nickbianco)

___
*[`OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp` line 94 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NezO-mI7mW9iAC6yVlK:-NflIPT9CYnOgePIPC1C:b-ofybff) ([raw file](https://github.com/opensim-org/opensim-core/blob/db2645e81b7fd1db3178279463816dc089c14bb1/OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp#L94)):*

@aymanhab, `FunctionBasedPath` gets automatically included in this test, so I need to do something to prevent failure. The two options are 1) what I've drafted here or 2) manually skip the test for `FunctionBasedPath`. Reverting entirely will lead to more failures.
</blockquote></details>

I totally understand, my question is at a higher level, why it's failing for this specific class and not for others that use similar structure. If strict validation is embedded within serialization then they probably need to be decoupled as this tends to slow down serialization. If there's something unique about it then I'm fine with leaving your change as is but documenting what's unique about it in the test.

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550)* status: all files reviewed, 1 unresolved discussion (waiting on @aymanhab)

___
*[`OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp` line 94 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NezO-mI7mW9iAC6yVlK:-NflSHPd5lsTsVBeJFNS:behu5go) ([raw file](https://github.com/opensim-org/opensim-core/blob/db2645e81b7fd1db3178279463816dc089c14bb1/OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp#L94)):*

I totally understand, my question is at a higher level, why it's failing for this specific class and not for others that use similar structure. If strict validation is embedded within serialization then they probably need to be decoupled as this tends to slow down serialization. If there's something unique about it then I'm fine with leaving your change as is but documenting what's unique about it in the test.
</blockquote></details>

The test fails because of the property checks in `extendFinalizeFromProperties`, not serialization. If you set random default property values, they most likely will not pass these tests (e.g., `length_function` will not have the correct number of arguments).

aymanhab (approved):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550)* status: all files reviewed, 1 unresolved discussion (waiting on @nickbianco)

___
*[`OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp` line 94 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NezO-mI7mW9iAC6yVlK:-NflUbc90fjReldc9rtR:bro36rf) ([raw file](https://github.com/opensim-org/opensim-core/blob/db2645e81b7fd1db3178279463816dc089c14bb1/OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp#L94)):*

The test fails because of the property checks in `extendFinalizeFromProperties`, not serialization. If you set random default property values, they most likely will not pass these tests (e.g., `length_function` will not have the correct number of arguments).
</blockquote></details>

Perfect, assuming this is unique here just clarify in the comment in place of (special case: it has specific requirements) so that it's clear the next generic component added would not claim "special case" ;)

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550)* status: 15 of 16 files reviewed, 1 unresolved discussion (waiting on @aymanhab)

___
*[`OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp` line 94 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-NezO-mI7mW9iAC6yVlK:-NflYNO77JqKRJJPr5dE:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/db2645e81b7fd1db3178279463816dc089c14bb1/OpenSim/Tools/Test/testSerializeOpenSimObjects.cpp#L94)):*

Perfect, assuming this is unique here just clarify in the comment in place of (special case: it has specific requirements) so that it's clear the next generic component added would not claim "special case" ;)
</blockquote></details>

Done.

nickbianco (commented):
@adamkewley, let me know if this still needs review or if it's good to go.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3550#-:-Nfr_oJ92Chv6OADFzqa:bkq5ofw)* status: 15 of 16 files reviewed, 1 unresolved discussion (waiting on @aymanhab)

adamkewley:
Let's see where this goes :) - good work!

nickbianco:
Thanks @adamkewley!

Issue #3485, opened 2023-06-14
Fix `PointToPointSpring` explodes when both ends are socketed onto the same body


adamkewley:
Found by @itbellix. Original issue is downstream in opensim-creator: https://github.com/ComputationalBiomechanicsLab/opensim-creator/issues/680

Reproduction:

- Open `WalkerModel.osim`
- Add `PointToPointSpring` and set both sockets to `Pelvis`
- Simulate
- BUG: Model is visible at `t = 0` and then disappears at `t > 0`

Throwing an error message at `finalizeFromProperties` time along the lines of "tut tut, you can't attach a spring this way" or similar might be enough to at least prevent users from doing this (unless there's an argument to permit this?).

tkuchida:
> unless there's an argument to permit this?

I don't think so… ensuring body1 and body2 point to a different `getMobilizedBody()` makes sense. (There's also a `PointToPointActuator` whose bodyA and bodyB must presumably be different.)

PR #3553, opened 2023-09-14
Fix PointToPointSpring should throw if both sides are attached to sam…

adamkewley:
Fixes issue #3485

### Brief summary of changes

- Added a check via `extendConnectToModel` that `body1` and `body2` of a `PointToPointSpring` aren't attached to the same base frame
- Added basic test suite (`testPointToPointSpring.cpp`) to exercise the behavior

### Testing I've completed

- I added the tests
- They failed
- I added the runtime check
- They passed
- (the tests exercise a basic version of what I imagine to be the bug, but do not test extended versions of the same bug, such as attaching two ends of the spring via offset frames to the same body)

### CHANGELOG.md

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3553)


tkuchida (approved):
:+1: (Only nit is missing newline at EOF in testPointToPointSpring.cpp.)

Issue #3347, opened 2023-01-03
Segfault when loading a model containing a `CoordinateCouplerConstraint` without a `coupled_coordinates_function`


adamkewley:
Edge-case that was spotted during OSC dev.

The following osim file will segfault OpenSim, OpenSim GUI, and OpenSim Creator on loading it (tested with OpenSim 4.4 desktop UI, etc):

- https://gist.github.com/adamkewley/60620383ea94c9ad93f3c430b8964558#file-ccc_segfault-osim-L95

The source of the segfault is:

- https://github.com/opensim-org/opensim-core/blob/4.4/OpenSim/Simulation/SimbodyEngine/CoordinateCouplerConstraint.cpp#L226

And it happens because the code doesn't check for the erroneous input condition of the file/user not providing a `coupled_coordinates_function`.

adamkewley:
I'll try and have a look into this next time I get a slot for OpenSim dev: I'm just posting it here with documentation etc. so it's known.

A quick fix would be to make `OPTIONAL` properties have a non-`nullptr` assertion on their getter.

adamkewley:
#3411 does **not** downgrade this from a segfault to a runtime exception, which implies that a similar fix is probably necessary for object-like properties (#3411 only fixes "simple" values)

PR #3554, opened 2023-09-14
Add bounds-checking to ObjectProperty and appropriate tests against r…

adamkewley:
Fixes issue #3347

### Brief summary of changes

- Used `.at` to add bounds-checking to `OpenSim::ObjectProperty`, so that mis-indexing into one (e.g. by calling `get_property()` or similar) causes a runtime exception rather than a segfault
- Added tests to `testComponentInterface.cpp` and `testConstraints.cpp` to ensure the behavior doesn't regress

### Testing I've completed

- I added the test to `testConstraints.cpp`
- It did not segfault, because a previous fix hotfixed it (here: https://github.com/opensim-org/opensim-core/commit/4fa4493cca0a4e11880c8b6538109fe6fd45eed7)
- I then added tests to `testComponentInterface.cpp`
- The `SimpleProperty` one did not segfault, because it was fixed in https://github.com/opensim-org/opensim-core/pull/3411
- The `ObjectProperty` one did segfault
- Then I added the bounds checking to `Object.h` (i.e. swap index accessors for `.at`)
- Then the test passed

### CHANGELOG.md

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3554)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ been added to these forces to provide access to concrete path types (e.g., `updP
 `getGeometryPath`, `updGeometryPath`, etc., or a suitable alternative.    
 - Fixed a minor memory leak when calling `OpenSim::CoordinateCouplerConstraint::setFunction` (#3541)
 - Increase the number of input dimensions supported by `MultivariatePolynomialFunction` to 6 (#3386)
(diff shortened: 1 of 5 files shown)


nickbianco (approved):
LGTM. 

(If it fails again, the Java Moco sliding mass test can be ignored)

Issue #3555, opened 2023-09-14
Convert all tests to the `Catch2` framework


nickbianco:
Current testing in OpenSim uses a variety of implementations and code styles. Most of the tests for Moco use `Catch2`, and since the Moco merge, we've been sporadically converting OpenSim tests to this new framework. This epic proposes that we convert _all_ the OpenSim tests to `Catch2` to unify the testing suite, speed up development (i.e., isolating and re-running only failed tests), and make adding new tests easy. This should be a "many hands make light work" approach, where we divide up the tests to be converted across several individual issues that separate members of the dev team can tackle.

# Tasks

## ~~Phase 1: Convert existing code to minimally use Catch~~ (Done!)

- For each `.cpp` file in the list below:
  - `#include <catch2/catch_all.hpp>`
  - **Minimally** convert each test function to use `TEST_CASE("test name")`
  - Remove existing `main` function (`catch` automates calling each case)
  - Put helper functions in an anonymous namespace (`namespace {}`)
  - Check+commit - mention this issue (#3555) in the commit message if you want it to appear in GitHub
  - PR

## Phase 2: Update any build/contribution guides to mention the change

- [ ] Standardize how tests are added via `CMake` macros (if needed)
- [ ] Update `CONTRIBUTING.MD` with guidelines for writing `Catch2` tests (and links to `Catch2` docs)
- [ ] Decide if any changes needed for Java/Matlab/Python tests

## Phase 3: Improve assertions and test framework usage

- For each `cpp` file in the list below:

  - Update assertions to use catch assertions, instead of OpenSim/SimTK assertions (they provide better testing feedback)
  - Parameterize tests etc. where possible using catch's conventions
  - If other testing macros are necessary, consolidate them under `Auxiliary` or `Common` (i.e., remove `OpenSim/Moco/Test/Testing.h`

# Phase 1 CPP todo List

- [x] Actuators\
    - [x] Test\testActuators.cpp - #3684
    - [x] Test\testDeGrooteFregly2016Muscle.cpp
    - [x] Test\testModelProcessor.cpp
    - [x] Test\testMuscleFirstOrderActivationDynamicModel.cpp - #3691
    - [x] Test\testMuscleFixedWidthPennationModel.cpp
    - [x] Test\testMuscles.cpp
    - [x] Test\testSerializableMuscleCurves.cpp
- [x] Analyses\
    - [x] Test\testOutputReporter.cpp
[...]

PR #3557, opened 2023-09-15
Update testStorage.cpp to use catch (#3555)

adamkewley:
Relates to #3555

### Brief summary of changes

- RTFCM

### CHANGELOG.md (choose one)

- no need to update because it would bore people senseless

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3557)

### DIFF
--- a/OpenSim/Common/Test/testStorage.cpp
+++ b/OpenSim/Common/Test/testStorage.cpp
@@ -20,33 +20,40 @@
  * limitations under the License.                                             *
  * -------------------------------------------------------------------------- */
 
-#include <fstream>
 #include <OpenSim/Common/Storage.h>
 #include <OpenSim/Auxiliary/auxiliaryTestFunctions.h>
 #include <OpenSim/Common/STOFileAdapter.h>
 
+#define CATCH_CONFIG_MAIN
+#include <OpenSim/Auxiliary/catch.hpp>
+
+#include <fstream>
+
 using namespace OpenSim;
 using namespace std;
 
+namespace {
+    // helper function
+    void loadStorageWithNColsFromFile(const std::string& fileName, const int numCols)
+    {
+        Storage storage{ fileName};
+        // remove extension
+        auto ix = fileName.rfind(".");
+        if (ix == std::string::npos) {
+            throw Exception("File name for Storage loading '" + fileName +
+                            "' must have a valid extension.");
+        }
+        std::string name = fileName.substr(0, ix-1);
+        Array<std::string> labels = storage.getColumnLabels();
 
-void testStorageLoadingFromFile(const std::string& fileName, const int numCols)
-{
-    Storage storage{ fileName};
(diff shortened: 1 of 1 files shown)


nickbianco (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3557)* status: all files reviewed, 2 unresolved discussions (waiting on @adamkewley)

___
*[`OpenSim/Common/Test/testStorage.cpp` line 24 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3557#-NeZX58j1iNBX3cSFjHH:-NeZX58j1iNBX3cSFjHI:b-jsmuqa) ([raw file](https://github.com/opensim-org/opensim-core/blob/ff7250bf80f489e8694ad41a6ea6512c97a856d6/OpenSim/Common/Test/testStorage.cpp#L24)):*
> ```cpp
> 
> #include <OpenSim/Common/Storage.h>
> #include <OpenSim/Auxiliary/auxiliaryTestFunctions.h>
> ```

I think this include is no longer necessary?
___
*[`OpenSim/Common/Test/testStorage.cpp` line 177 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3557#-NeZVjAe4_vysYPTU37J:-NeZVjAe4_vysYPTU37K:b-x8kyy5) ([raw file](https://github.com/opensim-org/opensim-core/blob/ff7250bf80f489e8694ad41a6ea6512c97a856d6/OpenSim/Common/Test/testStorage.cpp#L177)):*
> ```cpp
>     SimTK_TEST(sto.getStateIndex("/jointset/hip/hip_flexion/speed") == 1);
>     SimTK_TEST(sto.getStateIndex("/forceset/soleus/activation") == 2);
>     SimTK_TEST(sto.getStateIndex("/forceset/soleus/fiber_length") == 3);
> ```

If we're converting everything to `catch`, I'm inclined to use the built-in macros (i.e.,`CHECK`, `REQUIRE`, etc.). The SimTK macros might have some nice options for some numerical tests that `catch` doesn't support though.

adamkewley:
I'll dump the includes.

For the assertions, they fall under what if I've labeled as phase 3 of the epic. The reason why changing the assertions was seperated was to minimize the diff/changeset at each step of the process (it's easier to review "this functionally changes nothing but how the tests are named and ran" than "this changes both how the tests are ran *and* the test content itself needs to be scrutinized for logical equivalence")

nickbianco:
Ah, I see now. Sounds good and carry on!

adamkewley:
@nickbianco - I couldn't remove the `auxiliaryTestFunctions.h` because it's still used for the `_ASSERT` stuff, but it should be removed during the phase when we start upgrading assertions to `Catch2`.

I'll merge this for now because I believe the PR otherwise is within expectations etc

PR #3561, opened 2023-09-20
Add AbstractPath to binding interface so the class fits correctly in …

aymanhab:
…class hierarchy within scripting

Fixes issue #0
### Brief summary of changes
Required to be able to build gui, visualize muscles, and use methods from superclasses of GeometryPath in existing codebase

### Testing I've completed
Built GUI with this and other changes and was able to see/visualize muscles

### CHANGELOG.md (choose one)

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3561)


nickbianco (approved):

aymanhab:
Thanks @nickbianco

PR #3563, opened 2023-09-25
Convert testComponentInterface.cpp to use catch (#3555)

adamkewley:
Fixes issue #3555

The disabled test is disabled because the original code commented out the callsite in `main`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3563)

### DIFF
--- a/OpenSim/Common/Test/testComponentInterface.cpp
+++ b/OpenSim/Common/Test/testComponentInterface.cpp
@@ -32,6 +32,10 @@
 #include <simbody/internal/Force.h>
 #include <simbody/internal/MobilizedBody_Pin.h>
 #include <simbody/internal/MobilizedBody_Ground.h>
+
+#define CATCH_CONFIG_MAIN
+#include <OpenSim/Auxiliary/catch.hpp>
+
 #include <random>
 
 namespace
@@ -444,7 +448,8 @@ private:
 SimTK_NICETYPENAME_LITERAL(Foo);
 SimTK_NICETYPENAME_LITERAL(Bar);
 
-void testMisc() {
+TEST_CASE("Component Interface Misc.")
+{
     // Define the Simbody system
     MultibodySystem system;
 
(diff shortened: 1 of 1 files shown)


OpenSim/Common/Test/testComponentInterface.cpp
```diff
@@ -1987,11 +2009,14 @@ void writeTimeSeriesTableForInputConnecteeSerialization() {
...
 }
 
-void testListInputConnecteeSerialization() {
+TEST_CASE("Component Interface List Input Connectee Serialization Behavior")
+{
     // We build a model, store the input connectee paths, then
     // recreate the same model from a serialization, and make sure the
     // connectee paths are the same.
 
+    writeTimeSeriesTableForInputConnecteeSerialization();
```
tkuchida, line 2018: Was this added intentionally? Was the test not working before?

OpenSim/Common/Test/testComponentInterface.cpp
```diff
@@ -2085,10 +2110,12 @@ void testListInputConnecteeSerialization() {
...
 }
 
-void testSingleValueInputConnecteeSerialization() {
-
+TEST_CASE("Component Interface Single Value Input Connectee Serialization Behavior")
+{
     // Test normal behavior of single-value input (de)serialization.
     // -------------------------------------------------------------
+
+    writeTimeSeriesTableForInputConnecteeSerialization();
```
tkuchida, line 2118: Was this added intentionally?

tkuchida (commented):
LGTM. Not sure whether 2 additions were intentional (commented below) or perhaps I'm misreading the diff. Also not sure what "The disabled test is disabled because the original code commented out the callsite in main." implies -- if the disabled test should be re-enabled at some point, is this tracked in an Issue? 🌳

OpenSim/Common/Test/testComponentInterface.cpp
adamkewley: (as above)

OpenSim/Common/Test/testComponentInterface.cpp
adamkewley: The original code placed the call in `main` before running those two tests.

The test content is dependent on the `sto` file generated by `writeTimeSeriesTableForInputConnecteeSerialization`, but the original version of the code didn't encode it into each case (instead, it relied on tests being ran in a certain order).

tkuchida (approved):

adamkewley:
Thanks @tkuchida !

PR #3564, opened 2023-09-25
Change testGCVSpline to use catch (#3555)

adamkewley:
Fixes issue #3555

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3564)


tkuchida (approved):
https://github.com/opensim-org/opensim-core/pull/3564/files?w=1

adamkewley:
> https://github.com/opensim-org/opensim-core/pull/3564/files?w=1

This is a better diff, yeah - the un-indenting of the `try` block makes this uglier than it needs to be!

PR #3576, opened 2023-10-11
Add `opensim-moco`

adamkewley:
Fixes issue (not written up - just hacked in as a suggestion)

- Research user at TUD has been using `opensim-moco` v0.5 for several years. We couldn't upgrade without changing their workflow
- We suggested adding it during the moco merge, but we didn't get around to chasing it up: https://github.com/opensim-org/opensim-core/pull/2805#issuecomment-697178548
- But the researcher now needs features from more recent mocos, preferably without having to change their workflow

@nickbianco if there's a guide or similar for porting a CLI-based workflow to the approaches you'd prefer I can write that up as a guide and give it to the researcher, so that this suggested change becomes unnecessary. The researcher must use a CLI client because using (e.g.) moco via MATLAB bindings in (e.g.) Linux causes segfaults, whereas running it via a CLI as a separate process side-steps those issues.

### Brief summary of changes

- Copied `opensim-moco.cpp` from original repo: https://github.com/opensim-org/opensim-moco/blob/master/Moco/Executable/opensim-moco.cpp into `Applications/`
- Copied `startsWith` from original `MocoUtilities`
- Commented out `visualize` call

### Testing I've completed

- None: source copied as-is

- Whether it's worthwhile to do this and, if not, a suitable alternative that's close (ish) to the original workflow

### CHANGELOG.md (choose one)

- to be updated

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3576)

### DIFF
--- a/Applications/opensim-moco/opensim-moco.cpp
+++ b/Applications/opensim-moco/opensim-moco.cpp
@@ -26,6 +26,16 @@
 
 #include <iostream>
 
+/// Determine if `string` starts with the substring `start`.
+/// https://stackoverflow.com/questions/874134/find-if-string-ends-with-another-string-in-c
+/// @ingroup mocogenutil
+static bool startsWith(const std::string& string, const std::string& start) {
+    if (string.length() >= start.length()) {
+        return string.compare(0, start.length(), start) == 0;
+    }
+    return false;
+}
+
 using namespace OpenSim;
 
 static const char helpMessage[] =
@@ -103,7 +113,7 @@ void visualize(std::string file, std::string trajectory_file) {
         std::cin.get();
     } else {
         MocoTrajectory trajectory(trajectory_file);
-        visualize(*model, trajectory.exportToStatesStorage());
+        // visualize(*model, trajectory.exportToStatesStorage());
     }
 }
 
--- a/Applications/opensim-moco/opensim-moco.cpp
+++ b/Applications/opensim-moco/opensim-moco.cpp
@@ -1,0 +1,185 @@
+/* -------------------------------------------------------------------------- *
+ * OpenSim Moco: opensim-moco.cpp                                             *
+ * -------------------------------------------------------------------------- *
+ * Copyright (c) 2017-19 Stanford University and the Authors                  *
+ *                                                                            *
+ * Author(s): Christopher Dembia                                              *
+ *                                                                            *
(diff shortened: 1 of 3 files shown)


Applications/opensim-moco/opensim-moco.cpp
```diff
@@ -103,7 +113,7 @@ void visualize(std::string file, std::string trajectory_file) {
         std::cin.get();
     } else {
         MocoTrajectory trajectory(trajectory_file);
-        visualize(*model, trajectory.exportToStatesStorage());
+        // visualize(*model, trajectory.exportToStatesStorage());
```
adamkewley, line 116: This will need to be fixed before merging (API changed?)

nickbianco:
@adamkewley, does [`opensim-cmd`](https://github.com/opensim-org/opensim-core/blob/efc082803d0566eef4b82d8facca94b87bc9a4fd/Applications/opensim-cmd/opensim-cmd.cpp) not have the CLI functionality they need? It supports running Moco simulations via the `moco` argument. If not, I'd prefer to just add whatever features from `opensim-moco` he needs to `opensim-cmd`.

adamkewley:
@nickbianco it could do? I just haven't personally used `opensim-moco` or `opensim-cmd moco` and I saw that the source file was easy enough to copy over.

Is it safe to say that `opensim-cmd moco` will provide the same interface as `opensim-moco` - i.e. it takes the same input file (`omoco` or something?) and produces the same outputs etc? If that's the case then I can recommend it to the researcher, wait for their response, and if it's ok I can drop this or ask for relevant enhancements to that?

nickbianco:
@adamkewley, actually the `moco` argument is only needed when printing out an XML file. You can run a MocoStudy saved to an XML file (.omoco use the XML format) via:

```
opensim-cmd run-tool study.omoco
```
The main difference is that visualization will need a separate call to `opensim-cmd`, rather than using the `--visualize` flag:

```
opensim-cmd viz model model.osim mocoSolution.sto
```
See if this workflow works for your colleague, and if not, we can open an issue to discuss what needs to be added to `opensim-cmd`.

adamkewley:
Thanks for this info @nickbianco - I wasn't aware of this (for Moco-related stuff, my role is mostly to provide a stable environment for the end users, I don't really use it for OSC etc. yet).

I'll ask the researcher to do as you suggest and get back with any suggestions/problems separately!

PR #3577, opened 2023-10-12
Fix bug after constructing a `ModelProcessor` from a `Model` object

nickbianco:
Discovered by @mattpetrucci.

### Brief summary of changes
- Added `model.finalizeFromProperties()` and `model.finalizeConnections()` after constructing a `ModelProcessor` from a `Model` object. Without these calls, some `ModelOperator`s lead to an error complaining that the model had no subcomponents. This now matches the behavior when loading a model from a file.
- Removed some `finalizeFromProperties` calls in `ModelOperators.h`.

### Testing I've completed
- Ran `testModelProcessor.cpp` without the fix and it failed. Added the fix and it passed.

### CHANGELOG.md (choose one)
- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3577)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ been added to these forces to provide access to concrete path types (e.g., `updP
   it would let you do it, and throw later when `Coordinate::getMinRange()` or `Coordinate::getMaxRange()` were called, #3532)
 - Added `FunctionBasedPath`, a class for representing paths in `Force`s based on `Function` objects (#3389)
 - Fixed bindings to expose the method Model::getCoordinatesInMultibodyTreeOrder to scripting users (#3569)
+- Fixed a bug where constructing a `ModelProcessor` from a `Model` object led to an invalid `Model`
 
 v4.4.1
 ======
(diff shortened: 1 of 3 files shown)


aymanhab (approved):
:lgtm:

> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3577#-:-Nga567u21fzKddtOxJ_:b-1m4vin)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nickbianco)

tkuchida (commented):
This PR fixes a bug that was not caught by the current tests. I don't know the context of this change, but perhaps consider adding a test that fails before and passes after the change. @adamkewley has been doing this recently and it's awesome (e.g., #3554). 🦓

nickbianco:
@tkuchida I totally agree and actually did do this in a sneaky way. By removing the (now redundant) `finalizeFromProperties` calls from `ModelOperators`, `testModelProcessor` will fail if the fix is removed. Essentially, I used the existing tests to replicate the bug and tested with and without the fix, and got failure and pass as expected.

Hopefully that makes sense. Let me know if you were thinking something different.

tkuchida:
@nickbianco Thanks. I was bit confused: all the tests pass before this PR, they all still pass after this PR, and no tests were added, so it seemed like something was changing that isn't being tested. But there's no strict requirement about this so no worries. 🐫

tkuchida (approved):

nickbianco:
@tkuchida, thanks for checking! Perhaps we could include this a recommendation in the contributing guidelines for bug fixes.

Issue #3390, opened 2023-02-06
Create utility functions for converting existing muscle wrapping geometry to `FunctionBasedPath` objects


nickbianco:
Can implement these utilities in `ModelFactory` and/or `ModelOperators`.

PR #3581, opened 2023-10-20
`PolynomialPathFitter` and other utilities for `FunctionBasedPath`

nickbianco:
Fixes issue #3390

### Brief summary of changes
This PR introduces `PolynomialPathFitter`, a new utility class for creating `FunctionBasedPath`s by fitting the coefficients of `MultivariatePolynomialFunction`s to path length and moment arm samples taken from a model with geometry-based paths. Given a model, a table of coordinate values, and a set of (optional) user settings, this class automatically samples path lengths and moment arms from the model based on Latin hypercube sampling and determins the polynomial coefficients using a least-squares fit. In addition to the new class, various other smaller changes are included to make everything work:

- Added `ModelFactor::replacePathsWithFunctionBasedPaths` and `ModOpReplacePathsWithFunctionBasedPaths`.
- Added `calcTermValues` and `calcTermDerivatives` to `MultivariatePolynomialFunction`.
- Added `Object::getProperty(const std::string& name)` and `Object::updProperty(const std::string& name)`.
- Added `TypeHelper`s to `Property` for `SimTK::Vec2` types.
- Added `SimulationUtilities::appendCoupledCoordinateValues` and `TabOpAppendCoupledCoordinateValues`.
- Added `SimulationUtilities::appendCoordinateValueDerivatives` and `TabOpAppendCoordinateValueDerivatives`.
- Minor updates to `FunctionBasedPath`.
- Other various minor drive-by fixes that I encountered along the way.

### Testing I've completed
Added `testPolynomialPathFitter.cpp`.

Anything and everything.

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3581)


aymanhab (changes requested):
> Reviewed 5 of 29 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3581)* status: 5 of 29 files reviewed, 1 unresolved discussion (waiting on @adamkewley and @nickbianco)

*[a discussion](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhXpxVw4HrhIuq3TiGc:-NhXpxVw4HrhIuq3TiGd:b-yre01c) (no related file):*
I think this will be extremely useful to make the adoption easier :+1: . Some questions that may have been addressed or documented elsewhere that I found myself asking with every call so would be good to clear up or if you can point to existing documentation if exists already:
1. Do you have to do the replacement for all paths or you can be selective and do mix of GeometryBased and FunctionBased?
2. Are there checks that the functions passed-in correspond and map to components that are actually GeometryBased paths, nothing extra, and nothing missing (depending on previous item on this list).
3.  Assuming both paths exist in the model at some point, using _path as a suffix seems more confusing when we call both classes "*Path", I'd use something that includes "function" or "fit" unless I'm not clear on what's being named/renamed.
Will proceed with the review once I understand the context based on the above. Thanks

___

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3581)* status: 5 of 29 files reviewed, 1 unresolved discussion (waiting on @adamkewley and @aymanhab)

*[a discussion](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhXpxVw4HrhIuq3TiGc:-NhXuN-y39yhMqWHWU8C:b-kn7top) (no related file):*

I think this will be extremely useful to make the adoption easier :+1: . Some questions that may have been addressed or documented elsewhere that I found myself asking with every call so would be good to clear up or if you can point to existing documentation if exists already:
1. Do you have to do the replacement for all paths or you can be selective and do mix of GeometryBased and FunctionBased?
2. Are there checks that the functions passed-in correspond and map to components that are actually GeometryBased paths, nothing extra, and nothing missing (depending on previous item on this list).
3.  Assuming both paths exist in the model at some point, using _path as a suffix seems more confusing when we call both classes "*Path", I'd use something that includes "function" or "fit" unless I'm not clear on what's being named/renamed.
Will proceed with the review once I understand the context based on the above. Thanks
</blockquote></details>

1. You can be selective. We save the fitted polynomial paths to an XML file, so you can modify the XML file or load specific paths in to the model with the API. We could also add an `exclude_paths` option on `ModOpReplacePathWithFunctionBasedPaths`.
2. This class will only fit paths that derive from `AbstractPath` and (currently) will throw an exception if the model passed in already has `FunctionBasedPath`s. We could be more granular and only fit paths that are `GeometryPath` (or other future geometry-based paths), but for now I just assume that users do not intend to pass in a model with a `FunctionBasedPath`.
3. By both paths, do you mean supporting both a `GeometryPath` and a `FunctionBasedPath` for the same force object? Then yes that name convention might be confusing. But that could be handled in a future PR I think. 

___

aymanhab (changes requested):
> Reviewed 21 of 29 files at r1.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3581)* status: 26 of 29 files reviewed, 10 unresolved discussions (waiting on @adamkewley and @nickbianco)

*[a discussion](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhXpxVw4HrhIuq3TiGc:-NhXyWdAC2SMzVRFVxY-:blxp5ia) (no related file):*

1. You can be selective. We save the fitted polynomial paths to an XML file, so you can modify the XML file or load specific paths in to the model with the API. We could also add an `exclude_paths` option on `ModOpReplacePathWithFunctionBasedPaths`.
2. This class will only fit paths that derive from `AbstractPath` and (currently) will throw an exception if the model passed in already has `FunctionBasedPath`s. We could be more granular and only fit paths that are `GeometryPath` (or other future geometry-based paths), but for now I just assume that users do not intend to pass in a model with a `FunctionBasedPath`.
3. By both paths, do you mean supporting both a `GeometryPath` and a `FunctionBasedPath` for the same force object? Then yes that name convention might be confusing. But that could be handled in a future PR I think.
</blockquote></details>

Thanks for the clarification @nickbianco now I can proceed.

___
___
*[`OpenSim/Actuators/ModelFactory.cpp` line 318 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhXzZzO0oex9gnpTi0L:-NhXzZzO0oex9gnpTi0M:b-1clfi3) ([raw file](https://github.com/opensim-org/opensim-core/blob/2e4b7e5d964d28b99b0bd516707775c3723378f1/OpenSim/Actuators/ModelFactory.cpp#L318)):*
> ```cpp
>             
>         // Get the force component associated with this path.
>         OPENSIM_THROW_IF(!model.hasComponent<Force>(path.getName()), 
> ```

Have you considered using Path or AbstractPath instead of Force?  If so we use more specific class in updComponent<> here so we don't end up catching unrelated components? I think there're methods to  go from Path to the owner component safely and you're only using Object level methods anyway.
___
*[`OpenSim/Actuators/PolynomialPathFitter.h` line 46 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhbrrsX-CvPfNnTxd2y:-NhbrrsX-CvPfNnTxd2z:bbnrsns) ([raw file](https://github.com/opensim-org/opensim-core/blob/2e4b7e5d964d28b99b0bd516707775c3723378f1/OpenSim/Actuators/PolynomialPathFitter.h#L46)):*
> ```objc
>     OpenSim_DECLARE_PROPERTY(coordinate_path, std::string,
>             "The path to the bounded coordinate in the model.")
>     OpenSim_DECLARE_PROPERTY(bounds, SimTK::Vec2,
> ```

coordinate may not be "bounded" in the model, would it be clearer to say path to the coordinate to be bounded during fitting?
___
*[`OpenSim/Actuators/PolynomialPathFitter.h` line 80 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-Nhbt4BwAZfTTftYngGg:-Nhbt4BwAZfTTftYngGh:b-5s230q) ([raw file](https://github.com/opensim-org/opensim-core/blob/2e4b7e5d964d28b99b0bd516707775c3723378f1/OpenSim/Actuators/PolynomialPathFitter.h#L80)):*
> ```objc
>  * `setMomentArmsThreshold` method determines whether or not a path depends on a
>  * model coordinate. In other words, the moment arm of a path with respect to a
>  * particular coordinate must be greater than this value to be included during
> ```

using absolute value?
___
*[`OpenSim/Actuators/PolynomialPathFitter.h` line 97 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhbuTfMBiWbqCMulq9z:-NhbuTfMBiWbqCMulqA-:b-bvjjuj) ([raw file](https://github.com/opensim-org/opensim-core/blob/2e4b7e5d964d28b99b0bd516707775c3723378f1/OpenSim/Actuators/PolynomialPathFitter.h#L97)):*
> ```objc
>  * the Latin hypercube sampling algorithm used to sample coordinate values for
>  * path fitting.
>  *
> ```

With all these tolerances I'm assuming the model is assumed to be in meters so a model of a small bird or a dinosaur may not work out of the box? Is this something to account for or include a disclaimer?
___
*[`OpenSim/Actuators/PolynomialPathFitter.h` line 400 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-Nhbta3dB7jbYDgFTgff:-Nhbta3dB7jbYDgFTgfg:b-lhwrr4) ([raw file](https://github.com/opensim-org/opensim-core/blob/2e4b7e5d964d28b99b0bd516707775c3723378f1/OpenSim/Actuators/PolynomialPathFitter.h#L400)):*
> ```objc
>     void setParallel(int numThreads);
>     /// @copydoc setParallel()
>     int getParallel() const;
> ```

The name setParallel is a bit unexpected naming convention, I'd expect either a bool or numParallelThreads(number) but if this convention is used elsewhere in Moco then would be fine to keep 
___
*[`OpenSim/Common/MultivariatePolynomialFunction.h` line 103 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-Nhbn_WQ7mk0q3fJ2Brx:-Nhbn_WQ7mk0q3fJ2Bry:b7b3jvj) ([raw file](https://github.com/opensim-org/opensim-core/blob/2e4b7e5d964d28b99b0bd516707775c3723378f1/OpenSim/Common/MultivariatePolynomialFunction.h#L103)):*
> ```objc
>     /// Get a vector of the terms in the polynomial function.
>     SimTK::Vector getTermValues(const SimTK::Vector& x) const;
>     /// Get a vector of the derivatives of the terms in the polynomial function.
> ```

Do you think returning a fresh copy could be a performance bottleneck? Is this exercised during simulation?
___
*[`OpenSim/Common/Object.h` line 347 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhY2Qik3VAdhYFXMeqJ:-NhY2Qik3VAdhYFXMeqK:b9q49z5) ([raw file](https://github.com/opensim-org/opensim-core/blob/2e4b7e5d964d28b99b0bd516707775c3723378f1/OpenSim/Common/Object.h#L347)):*
> ```objc
>     template <class T> const Property<T>&
>     getProperty(const std::string& name) const;
> ```

For consistency with naming of other methods in the file I'd call this method getPropertyByName. Same for other new methods that do name lookup. Curious why this was needed now
___
*[`OpenSim/Common/Object.h` line 355 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhY4kqG9w9tftNtTwKh:-NhY4kqG9w9tftNtTwKi:b-a2kh3x) ([raw file](https://github.com/opensim-org/opensim-core/blob/2e4b7e5d964d28b99b0bd516707775c3723378f1/OpenSim/Common/Object.h#L355)):*
> ```objc
>     /** @copydoc updProperty(const PropertyIndex&) **/
>     template <class T> Property<T>&
>     updProperty(const std::string& name);
> ```

This looks fine but I'm wondering if this is part of the drive-by changes  or are relevant to this PR functionality. Thank you @nickbianco 
___
*[`OpenSim/Simulation/SimulationUtilities.h` line 401 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhbpZR94k9sS8U0EBzV:-NhbpZR94k9sS8U0EBzW:bsrbj4x) ([raw file](https://github.com/opensim-org/opensim-core/blob/2e4b7e5d964d28b99b0bd516707775c3723378f1/OpenSim/Simulation/SimulationUtilities.h#L401)):*
> ```objc
>         TimeSeriesTable& table, const Model& model,
>         bool overwriteExistingColumns = true);
> ```

Should the method name include "speeds" vs "derivatives"? Trying to consolidate the comment with the naming

adamkewley:
Apologies for the delay on looking at this, but I've been a bit busy, and now I'm away

I'd nominate @pepbos  to have a quick look, given that he's interested in similar ideas etc. and knows his way around C++ 😉

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3581)* status: 20 of 29 files reviewed, 10 unresolved discussions (waiting on @aymanhab and @pepbos)

*[a discussion](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhXpxVw4HrhIuq3TiGc:-Nhcx6xp-IFocUI4R6j1:bw8isc9) (no related file):*

Thanks for the clarification @nickbianco now I can proceed.
</blockquote></details>

Thanks @aymanhab, ready for review again.

___
___
*[`OpenSim/Actuators/ModelFactory.cpp` line 318 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhXzZzO0oex9gnpTi0L:-Nhcnmzv2JzXcMbRL15c:bczphna) ([raw file](https://github.com/opensim-org/opensim-core/blob/2e4b7e5d964d28b99b0bd516707775c3723378f1/OpenSim/Actuators/ModelFactory.cpp#L318)):*

Have you considered using Path or AbstractPath instead of Force?  If so we use more specific class in updComponent<> here so we don't end up catching unrelated components? I think there're methods to  go from Path to the owner component safely and you're only using Object level methods anyway.
</blockquote></details>

The problem is that `AbstractPath`s usually do not have unique names/paths in the model. So I could search through all `AbstractPath`s in the model, but I'd end up having to check the `Force` name anyway. The subsequent line checks if the `path` property exists in the component.
___
*[`OpenSim/Actuators/PolynomialPathFitter.h` line 46 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhbrrsX-CvPfNnTxd2y:-NhcqPU28zALY4UOlJmr:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/2e4b7e5d964d28b99b0bd516707775c3723378f1/OpenSim/Actuators/PolynomialPathFitter.h#L46)):*

coordinate may not be "bounded" in the model, would it be clearer to say path to the coordinate to be bounded during fitting?
</blockquote></details>

Done.
___
*[`OpenSim/Actuators/PolynomialPathFitter.h` line 80 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-Nhbt4BwAZfTTftYngGg:-Nhcqgh29g1HFu7wP8-z:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/2e4b7e5d964d28b99b0bd516707775c3723378f1/OpenSim/Actuators/PolynomialPathFitter.h#L80)):*

using absolute value?
</blockquote></details>

Done.
___
*[`OpenSim/Actuators/PolynomialPathFitter.h` line 97 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhbuTfMBiWbqCMulq9z:-NhcrATIBK-e94vUE4Ob:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/2e4b7e5d964d28b99b0bd516707775c3723378f1/OpenSim/Actuators/PolynomialPathFitter.h#L97)):*

With all these tolerances I'm assuming the model is assumed to be in meters so a model of a small bird or a dinosaur may not work out of the box? Is this something to account for or include a disclaimer?
</blockquote></details>

Done.
___
*[`OpenSim/Actuators/PolynomialPathFitter.h` line 400 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-Nhbta3dB7jbYDgFTgff:-NhcsRsu-FT80Ggy2i8d:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/2e4b7e5d964d28b99b0bd516707775c3723378f1/OpenSim/Actuators/PolynomialPathFitter.h#L400)):*

The name setParallel is a bit unexpected naming convention, I'd expect either a bool or numParallelThreads(number) but if this convention is used elsewhere in Moco then would be fine to keep
</blockquote></details>

Done.
___
*[`OpenSim/Common/MultivariatePolynomialFunction.h` line 103 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-Nhbn_WQ7mk0q3fJ2Brx:-NhctpBrA-L-iBjJAv2s:bp2b0y0) ([raw file](https://github.com/opensim-org/opensim-core/blob/2e4b7e5d964d28b99b0bd516707775c3723378f1/OpenSim/Common/MultivariatePolynomialFunction.h#L103)):*

Do you think returning a fresh copy could be a performance bottleneck? Is this exercised during simulation?
</blockquote></details>

I'm returning a temporary `SimTK::Vector` constructed inside this method, so returning it like this should be optimized by the compiler. 

https://stackoverflow.com/questions/4986673/is-it-necessary-to-stdmove-in-a-return-statement-and-should-you-return-rvalue 
___
*[`OpenSim/Common/Object.h` line 347 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhY2Qik3VAdhYFXMeqJ:-NhcuX214tYlcnKkxna4:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/2e4b7e5d964d28b99b0bd516707775c3723378f1/OpenSim/Common/Object.h#L347)):*

For consistency with naming of other methods in the file I'd call this method getPropertyByName. Same for other new methods that do name lookup. Curious why this was needed now
</blockquote></details>

Done.
___
*[`OpenSim/Common/Object.h` line 355 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhY4kqG9w9tftNtTwKh:-Nhcvwk38QsVobhXfsQn:b3mbjjb) ([raw file](https://github.com/opensim-org/opensim-core/blob/2e4b7e5d964d28b99b0bd516707775c3723378f1/OpenSim/Common/Object.h#L355)):*

This looks fine but I'm wondering if this is part of the drive-by changes  or are relevant to this PR functionality. Thank you @nickbianco
</blockquote></details>

They are used in `PolynomialPathFitter` and `ModelFactory::replacePathsWithFunctionBasedPaths`.
___
*[`OpenSim/Simulation/SimulationUtilities.h` line 401 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhbpZR94k9sS8U0EBzV:-NhcwWIo9DqiqRX6yCJ9:b-ia93di) ([raw file](https://github.com/opensim-org/opensim-core/blob/2e4b7e5d964d28b99b0bd516707775c3723378f1/OpenSim/Simulation/SimulationUtilities.h#L401)):*

Should the method name include "speeds" vs "derivatives"? Trying to consolidate the comment with the naming
</blockquote></details>

Added the "AsSpeeds" suffix for clarity.

aymanhab (changes requested):
Looks great @nickbianco Just a couple clarifications and then it's ready to go :+1: 

> Reviewed 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-:-NhhNXPh16MZLNzYaLP8:bjual1g)* status: all files reviewed, 8 unresolved discussions (waiting on @nickbianco and @pepbos)

___
*[`OpenSim/Actuators/PolynomialPathFitter.h` line 157 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhhKHRG2qHyfBAE4Xe_:-NhhKHRG2qHyfBAE4Xea:b7frj13) ([raw file](https://github.com/opensim-org/opensim-core/blob/2d1a6feeb7e5a54105b638c26900c84e0c711bcd/OpenSim/Actuators/PolynomialPathFitter.h#L157)):*
> ```objc
>  *
>  * It is highly recommended to use the files printed to the output directory to
>  * evaluate the quality of the fitted paths. Depending on the quality of the
> ```

Maybe documented else where but a pointer to what files are printed and how to use would be helpful.
___
*[`OpenSim/Actuators/PolynomialPathFitter.h` line 190 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhhKaJi1uQJoKN4g37Q:-NhhKaJi1uQJoKN4g37R:bm5xj0d) ([raw file](https://github.com/opensim-org/opensim-core/blob/2d1a6feeb7e5a54105b638c26900c84e0c711bcd/OpenSim/Actuators/PolynomialPathFitter.h#L190)):*
> ```objc
>      * the model to contain at least one path object derived from `AbstractPath`
>      * and does not already contain any `FunctionBasedPath` objects. The bounds
>      * for clamped coordinates are obeyed during the fitting process. Locked
> ```

To clarify, is the implication that users can't convert to FunctionBasedPaths one muscle/path at a time? Definitely this is cleaner/easier for book-keeping and they can iteratively add and refit, just confirming supported workflow.
___
*[`OpenSim/Actuators/PolynomialPathFitter.h` line 309 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhhLj6qEtNepjcxJvhw:-NhhLj6qEtNepjcxJvhx:bfce9my) ([raw file](https://github.com/opensim-org/opensim-core/blob/2d1a6feeb7e5a54105b638c26900c84e0c711bcd/OpenSim/Actuators/PolynomialPathFitter.h#L309)):*
> ```objc
> 
>     /**
>      * The global bounds (in degrees) that determine the minimum and maximum
> ```

Anything to say here about translational coordinates?
___
*[`OpenSim/Actuators/PolynomialPathFitter.cpp` line 2 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhhMd3x76i-RcbC0JXi:-NhhMd3x76i-RcbC0JXj:b-lzlfe3) ([raw file](https://github.com/opensim-org/opensim-core/blob/2d1a6feeb7e5a54105b638c26900c84e0c711bcd/OpenSim/Actuators/PolynomialPathFitter.cpp#L2)):*
> ```cpp
>  *                    OpenSim:  PolynomialPathFitter.h                        *
> ```

.cpp :)
___
*[`OpenSim/Actuators/Test/testPolynomialPathFitter.cpp` line 131 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhhNJYU0-DVZjgIJcsa:-NhhNJYU0-DVZjgIJcsb:b-hy7ngw) ([raw file](https://github.com/opensim-org/opensim-core/blob/2d1a6feeb7e5a54105b638c26900c84e0c711bcd/OpenSim/Actuators/Test/testPolynomialPathFitter.cpp#L131)):*
> ```cpp
>         REQUIRE_THROWS_WITH(fitter.run(), Catch::Contains(
>             "Expected 'maximum_polynomial_order' to be at most 9"));
>     }
> ```

Very nicely done/tested @nickbianco

nickbianco (commented):
@aymanhab ready for review.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-:-NhiFSo-AsDg-A_PgkL9:bbfq9k4)* status: 27 of 29 files reviewed, 8 unresolved discussions (waiting on @aymanhab and @pepbos)

___
*[`OpenSim/Actuators/PolynomialPathFitter.h` line 157 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhhKHRG2qHyfBAE4Xe_:-NhiFHtf3-9uti7qF7hU:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/2d1a6feeb7e5a54105b638c26900c84e0c711bcd/OpenSim/Actuators/PolynomialPathFitter.h#L157)):*

Maybe documented else where but a pointer to what files are printed and how to use would be helpful.
</blockquote></details>

Done.
___
*[`OpenSim/Actuators/PolynomialPathFitter.h` line 190 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhhKaJi1uQJoKN4g37Q:-NhiEYow-gT2dA6Y-J6L:b-fu09ji) ([raw file](https://github.com/opensim-org/opensim-core/blob/2d1a6feeb7e5a54105b638c26900c84e0c711bcd/OpenSim/Actuators/PolynomialPathFitter.h#L190)):*

To clarify, is the implication that users can't convert to FunctionBasedPaths one muscle/path at a time? Definitely this is cleaner/easier for book-keeping and they can iteratively add and refit, just confirming supported workflow.
</blockquote></details>

Yes, that correct. However, every path is fitted independently, so it would be easy enough for users to remove muscles from the model that they do not want in the fitting process.
___
*[`OpenSim/Actuators/PolynomialPathFitter.h` line 309 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhhLj6qEtNepjcxJvhw:-NhiEX6v4SGvyQUccbvp:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/2d1a6feeb7e5a54105b638c26900c84e0c711bcd/OpenSim/Actuators/PolynomialPathFitter.h#L309)):*

Anything to say here about translational coordinates?
</blockquote></details>

Done.
___
*[`OpenSim/Actuators/PolynomialPathFitter.cpp` line 2 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhhMd3x76i-RcbC0JXi:-NhiD84WAnurNttgZj64:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/2d1a6feeb7e5a54105b638c26900c84e0c711bcd/OpenSim/Actuators/PolynomialPathFitter.cpp#L2)):*

.cpp :)
</blockquote></details>

Done.
___
*[`OpenSim/Actuators/Test/testPolynomialPathFitter.cpp` line 131 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-NhhNJYU0-DVZjgIJcsa:-NhiD90WEm6Kryv0OFsR:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/2d1a6feeb7e5a54105b638c26900c84e0c711bcd/OpenSim/Actuators/Test/testPolynomialPathFitter.cpp#L131)):*

Very nicely done/tested @nickbianco
</blockquote></details>

Done.

OpenSim/Actuators/ModelFactory.cpp
```diff
@@ -309,3 +309,26 @@ void ModelFactory::createReserveActuators(Model& model, double optimalForce,
...
+                path.getName());
+        auto& force = model.updComponent<Force>(path.getName());
+            
+        // Check that the force has a path property.
+        OPENSIM_THROW_IF(
+                !force.hasProperty("path"), Exception,
+                "Force {} does not have a path property.", path.getName());
+            
+        // Update the path.
+        path.setName(fmt::format("{}_path", force.getName()));
```
pepbos, line 329: The "path" to a PathActuator as the name of a FunctionBasedPath's name, postfixed with `_path` seems confusing.
If there are more functions out there that work like this it might become more confusing.
Is this postfix with `_path` a fix or feature? Why not use the name the user gave to this FunctionBasedPath?

OpenSim/Actuators/ModelFactory.cpp
```diff
@@ -309,3 +309,26 @@ void ModelFactory::createReserveActuators(Model& model, double optimalForce,
...
+        force.updPropertyByName<AbstractPath>("path").setValue(path);
```
pepbos, line 330: Use try here?

pepbos, line 330: Does this name never change? Since "path" seems to be pretty ambiguous, someone might change the name of the property "path".

pepbos (commented):
Interesting stuff!
The different uses of "path" become a bit confusing, especially if things are done at runtime, rather than checked at compile time.

OpenSim/Common/MultivariatePolynomialFunction.cpp
```diff
@@ -113,20 +116,23 @@ class SimTKMultivariatePolynomial : public SimTK::Function_<T> {
...
     }
-    T calcDerivative(const SimTK::Array_<int>& derivComponent,
-            const SimTK::Vector& x) const override {
+    
+    SimTK::Vector_<T> calcTermDerivatives(
+            const SimTK::Array_<int>& derivComponent,
+            const SimTK::Vector& x) const {
         std::array<int, 6> nq{{0, 0, 0, 0, 0, 0}};
-        T value = static_cast<T>(0);
+        SimTK::Vector_<T> derivatives = static_cast<SimTK::Vector_<T>>(
```
pepbos, line 134: Are you sure this doesnt simply cast the Vector, not the data?

OpenSim/Common/MultivariatePolynomialFunction.cpp
```diff
@@ -163,57 +169,57 @@ class SimTKMultivariatePolynomial : public SimTK::Function_<T> {
                                 if (derivComponent[0] == 0) {
                                     nqNonNegative = nq[0] - 1;
                                     if (nqNonNegative < 0) nqNonNegative = 0;
-                                    T valueP = nq[0] * std::pow(x[0], nqNonNegative);
+                                    T derivP = nq[0] * std::pow(x[0], nqNonNegative);
```
pepbos, line 172: The code above looks a lot like the code at the constructor: Perhaps worth reusing it...

OpenSim/Common/MultivariatePolynomialFunction.cpp
```diff
@@ -244,3 +262,20 @@ SimTK::Function* MultivariatePolynomialFunction::createSimTKFunction() const {
     return new SimTKMultivariatePolynomial<SimTK::Real>(
             get_coefficients(), get_dimension(), getOrder());
 }
+
+SimTK::Vector MultivariatePolynomialFunction::getTermValues(
+        const SimTK::Vector& x) const {
+    if (_function == nullptr)
```
pepbos, line 268: `if (!_function) {`

OpenSim/Common/MultivariatePolynomialFunction.cpp
```diff
@@ -163,57 +169,57 @@ class SimTKMultivariatePolynomial : public SimTK::Function_<T> {
...
                                     if (nqNonNegative < 0) nqNonNegative = 0;
-                                    T valueP = nq[5] * std::pow(x[5], nqNonNegative);
+                                    T derivP = nq[5] * std::pow(x[5], nqNonNegative);
                                     for (int i = 0; i < dimension; ++i) {
                                         if (i == derivComponent[0]) continue;
-                                        valueP *= std::pow(x[i], nq[i]);
+                                        derivP *= std::pow(x[i], nq[i]);
                                     }
-                                    value += valueP * coefficients[coeff_nr];
+                                    derivatives[coeff_nr] = derivP;
```
pepbos, line 222: A few comments would really help reading this large piece of code.

pepbos (commented):
A more general question: In MultivariatePolynomialFunction, why are you inheriting from the templated Function<T>, and not `Function_<double>`?

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3581)* status: 25 of 30 files reviewed, 20 unresolved discussions (waiting on @aymanhab and @pepbos)

___
*[`OpenSim/Common/MultivariatePolynomialFunction.cpp` line 79 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3581#gh-1378490391:-NiGIeC45HJFtIyFzv4p:b2rny) ([raw file](https://github.com/opensim-org/opensim-core/blob/5ed1300080eaca7e10f60db221f58feab2c36ac2/OpenSim/Common/MultivariatePolynomialFunction.cpp#L79)):*
<details><summary><i>Previously, pepbos (pepbos) wrote…</i></summary><blockquote>

Avoid heap allocation.
</blockquote></details>

`values` is initialized the function so it should be allocated on the stack. Unless static casting (which is now removed, see below) does funny things with stack vs heap allocation that I'm not aware of.
___
*[`OpenSim/Common/MultivariatePolynomialFunction.cpp` line 134 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3581#gh-1378497325:-NiGIYrZ99oF4QOQyTwY:b-vhggml) ([raw file](https://github.com/opensim-org/opensim-core/blob/5ed1300080eaca7e10f60db221f58feab2c36ac2/OpenSim/Common/MultivariatePolynomialFunction.cpp#L134)):*
<details><summary><i>Previously, pepbos (pepbos) wrote…</i></summary><blockquote>

Are you sure this doesnt simply cast the Vector, not the data?
</blockquote></details>

Updated to avoid the `static_cast`.
___
*[`OpenSim/Common/MultivariatePolynomialFunction.cpp` line 172 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3581#gh-1378499004:-NiGIYCp8YL2KxAVQdVt:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5ed1300080eaca7e10f60db221f58feab2c36ac2/OpenSim/Common/MultivariatePolynomialFunction.cpp#L172)):*
<details><summary><i>Previously, pepbos (pepbos) wrote…</i></summary><blockquote>

The code above looks a lot like the code at the constructor: Perhaps worth reusing it...
</blockquote></details>

Done.
___
*[`OpenSim/Common/MultivariatePolynomialFunction.cpp` line 222 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3581#gh-1378501324:-NiGIXFwCy0-LCUZ6npW:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5ed1300080eaca7e10f60db221f58feab2c36ac2/OpenSim/Common/MultivariatePolynomialFunction.cpp#L222)):*
<details><summary><i>Previously, pepbos (pepbos) wrote…</i></summary><blockquote>

A few comments would really help reading this large piece of code.
</blockquote></details>

Done.
___
*[`OpenSim/Common/MultivariatePolynomialFunction.cpp` line 268 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3581#gh-1378500073:-NiFxRJW4AY_3zG0Qgb0:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5ed1300080eaca7e10f60db221f58feab2c36ac2/OpenSim/Common/MultivariatePolynomialFunction.cpp#L268)):*
<details><summary><i>Previously, pepbos (pepbos) wrote…</i></summary><blockquote>

`if (!_function) {`
</blockquote></details>

Done.

nickbianco (commented):
Thanks @pepbos! Ready for review again. I tried to make `replacePathsWithFunctionBasedPaths` less confusing, but happy to make additional changes if necessary.

> In MultivariatePolynomialFunction, why are you inheriting from the templated Function, and not `Function_<double>`?

I didn't originally write this code, but it looks like [inheriting from `Function_<T>` is the intended interface in Simbody](https://github.com/simbody/simbody/blob/5c5437a2e22afbebd33dc38b1b008c7f984e7786/SimTKcommon/include/SimTKcommon/internal/Function.h#L51).

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3581#-:-NiH61x2Auoc0NCLN2Qd:b-albgm3)* status: 23 of 30 files reviewed, 20 unresolved discussions (waiting on @aymanhab and @pepbos)

___
*[`OpenSim/Actuators/ModelFactory.cpp` line 313 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3581#gh-1378442097:-NiH5iSi0jpV7QPQZpGX:b-g4d4pz) ([raw file](https://github.com/opensim-org/opensim-core/blob/5ed1300080eaca7e10f60db221f58feab2c36ac2/OpenSim/Actuators/ModelFactory.cpp#L313)):*
<details><summary><i>Previously, pepbos (pepbos) wrote…</i></summary><blockquote>

The different meanings of "path" becomes confusing in this function...
</blockquote></details>

I tried to clarify in the doc comments.
___
*[`OpenSim/Actuators/ModelFactory.cpp` line 320 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3581#gh-1378448423:-NiH5hkD5gxZwaRLAk6H:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5ed1300080eaca7e10f60db221f58feab2c36ac2/OpenSim/Actuators/ModelFactory.cpp#L320)):*
<details><summary><i>Previously, pepbos (pepbos) wrote…</i></summary><blockquote>

This Throw is already performed at `updComponent` calls, with similar message. Perhaps add info if needed, or remove the throw here?
</blockquote></details>

Done.
___
*[`OpenSim/Actuators/ModelFactory.cpp` line 321 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3581#gh-1378447355:-NiH5xCq12CA7Mz2yZw5:b-a6o0io) ([raw file](https://github.com/opensim-org/opensim-core/blob/5ed1300080eaca7e10f60db221f58feab2c36ac2/OpenSim/Actuators/ModelFactory.cpp#L321)):*
<details><summary><i>Previously, pepbos (pepbos) wrote…</i></summary><blockquote>

Seems odd that the name of the FunctionBasedPath is equal to the path of a component in a model. Is there a reason for this?
</blockquote></details>

See my comment below.
___
*[`OpenSim/Actuators/ModelFactory.cpp` line 326 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3581#gh-1378450641:-NiH06bU1KT_rxI7pQrO:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5ed1300080eaca7e10f60db221f58feab2c36ac2/OpenSim/Actuators/ModelFactory.cpp#L326)):*
<details><summary><i>Previously, pepbos (pepbos) wrote…</i></summary><blockquote>

This throw does add new info, but does not use the throw from `updProperty`. Perhaps use `try`?
</blockquote></details>

Done.
___
*[`OpenSim/Actuators/ModelFactory.cpp` line 329 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3581#gh-1378455804:-NiH08T781g47y2y3dMk:b4uq295) ([raw file](https://github.com/opensim-org/opensim-core/blob/5ed1300080eaca7e10f60db221f58feab2c36ac2/OpenSim/Actuators/ModelFactory.cpp#L329)):*
<details><summary><i>Previously, pepbos (pepbos) wrote…</i></summary><blockquote>

The "path" to a PathActuator as the name of a FunctionBasedPath's name, postfixed with `_path` seems confusing.
If there are more functions out there that work like this it might become more confusing.
Is this postfix with `_path` a fix or feature? Why not use the name the user gave to this FunctionBasedPath?
</blockquote></details>

Naming the `FunctionBasedPath` is actually unnecessary since all path-based forces will store it using a property named 'path'. When the force is serialized it will then have the name 'path'. The names for each `FunctionBasedPath` in the XML file is so we know which `Force` to assign the `FunctionBasedPath` to.  It might be more proper to have a small helper class that stores the `FunctionBasedPath` and the name of the `Force` it is assigned to, but this implementation seemed reasonable to me. 
___
*[`OpenSim/Actuators/ModelFactory.cpp` line 330 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3581#gh-1378457115:-NiH071P96GfEWrb-67t:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/5ed1300080eaca7e10f60db221f58feab2c36ac2/OpenSim/Actuators/ModelFactory.cpp#L330)):*
<details><summary><i>Previously, pepbos (pepbos) wrote…</i></summary><blockquote>

Use try here?
</blockquote></details>

Done.
___
*[`OpenSim/Actuators/ModelFactory.cpp` line 330 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3581#gh-1378473038:-NiH39ng6sUuobN9zd4D:byd29fw) ([raw file](https://github.com/opensim-org/opensim-core/blob/5ed1300080eaca7e10f60db221f58feab2c36ac2/OpenSim/Actuators/ModelFactory.cpp#L330)):*
<details><summary><i>Previously, pepbos (pepbos) wrote…</i></summary><blockquote>

Does this name never change? Since "path" seems to be pretty ambiguous, someone might change the name of the property "path".
</blockquote></details>

As I said in the other comment, all path-based forces (i.e., `PathSpring`, `PathActuator`, `Ligament`, and `Blankevoort1991Ligament`) store paths using a property named 'path'. I guess someone could create a new path-based force object that uses a different property name, which would cause a failure here. As an alternative, since the force should only have one path, I could somehow look up the `PropertyIndex` for the path and update it that way.

aymanhab (approved):
> Reviewed 2 of 2 files at r3, 1 of 3 files at r4, 1 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3581)* status: 26 of 30 files reviewed, 12 unresolved discussions (waiting on @pepbos)

nickbianco:
Merging with @aymanhab's approval and passing tests. I have a few other fixes, but they will come in a follow up PR.

Issue #3537, opened 2023-08-29
Loading `Model`s from XML causes memory leaks


nickbianco:

adamkewley:
Initial investigation ([here](https://github.com/ComputationalBiomechanicsLab/opensim-creator/issues/379#issuecomment-1235224143)) found that this is probably the culprit:

- https://github.com/simbody/simbody/blob/Simbody-3.7/SimTKcommon/src/Xml.cpp#L560

Here is an example reproduction of the leak from OpenSim (it's in simbody, but here is how it manifests via the OpenSim API):

```bash
#!/usr/bin/env bash
set -xeuo pipefail

# get sources + models
git clone https://github.com/opensim-org/opensim-core
git clone https://github.com/opensim-org/opensim-models

# configure + build deps with libASAN
CCFLAGS=-fsanitize=address CXXFLAGS=-fsanitize=address cmake -S opensim-core/dependencies/ -B osim-deps-libasan-build -DCMAKE_INSTALL_PREFIX=${PWD}/osim-deps-libasan-install
cmake --build osim-deps-libasan-build/ -j$(nproc)

# bodge a test script over an existing binary in OpenSim
cat << EOF > opensim-core/OpenSim/Tests/Wrapping/testWrapping.cpp
#include <OpenSim/Simulation/Model/Model.h>

int main(int argc, char* argv[]) {
    for (int i = 1; i < argc; ++i) {
        OpenSim::Model{argv[i]};
    }
    return 0;
}
EOF

# configure + build the bodged binary with libASAN
CCFLAGS=-fsanitize=address CXXFLAGS=-fsanitize=address cmake -S opensim-core -B osim-libasan-build -DOPENSIM_DEPENDENCIES_DIR=${PWD}/osim-deps-libasan-install -DOPENSIM_WITH_CASADI=OFF -DOPENSIM_WITH_TROPTER=OFF -DBUILD_API_ONLY=ON
cmake --build osim-libasan-build --target testWrapping -j$(nproc)

./osim-libasan-build/testWrapping opensim-models/Models/RajagopalModel/Rajagopal2015.osim
```

Example output (`./osim-libasan-build/testWrapping opensim-models/Models/RajagopalModel/Rajagopal2015.osim`):

```text
=================================================================
==17784==ERROR: LeakSanitizer: detected memory leaks

Indirect leak of 231984 byte(s) in 1074 object(s) allocated from:
[...]

PR #3592, opened 2023-10-30
Fixed memory leaks by calling clearOrphan() on removed XML nodes.

fcanderson:
XML-related memory leaks were reported when loading OpenSim models. Specifically, deserializing Rajagopal2015.osim causes numerous memory leaks.  See issue #3537.

I used WinCRT to help find the leaks.  Since WinCRT is platform specific (i.e., runs on Windows) and requires some changes to the source code to leak test effectively, this PR only includes the fixes (not the modified code that I used to identify the leaks and verify the fixes).

None of the observed leaks associated with deserializing Rajagopal2015.osim were due to coding errors in SimTK::Xml. SimTK::Xml looks like it is working fine.  (@adamkewley and I did find a constructor in SimTK::Xml that should be revised, but it is not currently the source of leaks that I can identity).

The reported memory leaks were all occurring in OpenSim in various ```updateFromXMLNode()``` calls. In particular, ```updateFromXMLNode()```  for ```OpenSim::Body```, ```OpenSim::Coordinate```, ```OpenSim::Muscle```, ```OpenSim::PhysicalFrame```, and ```OpenSim::WrapObject```.

All of the leaks were in code related to converting from an older version of OpenSim (i.e., from v3.0) to the current version.  Thus, the memory leaks can be sidestepped by converting old models to the current version (by deserializing them) and then saving to a new .osim model file.

This sort of sidestepping doesn't address a situation in which a old version model is repeatedly deserialized, however.

Fortunately, the leaks were not too difficult to find. In each case, an element was removed from the XML document and was not inserted elsewhere in the model hierarchy.  In all these cases, the fix was simply to call ```clearOrphan()``` on the element that was removed. 

Rajagopal2015.osim, whether represented in v3.0 or in the current version, is now deserializing without memory leaks.

Fixes issue #3537

### Brief summary of fixes
In the files listed below, I inserted a call to ```clearOrphan()```. To make these calls easier to find, amidst all the trailing whitespaces that were removed, I've included the line numbers of the inserted calls.

1. SimbodyEngine/Body.cpp : line 397
2. SimbodyEngine/Coordinate.cpp : line 705
3. Model/Muscle.cpp : lines 110, 121
4. Model/PhysicalFrame.cpp : line 190
5. Wrap/WrapObject.cpp : lines 218, 238, 260

### Testing I've completed
- In Simbody, I mimicked the operations that are commonly performed in OpenSim during deserialization and verified that no leaks occur in SimTK::Xml.
- For a few simple models (those that I am using to test my exponential contact model), I verified that no leaks occur during deserialization.
- For the new StatesDocument class that I've been developing, which is XML based, I verified that no leaks occur when a complete states trajectory is de/serialized.
- For Rajagopal2015.osim (v3.0), I verified that deserialization generated leaks before the above fixes and that no leaks were generated after.
- For Rajagopal2015.osim, I verified that serialization of the model to a new file produced identical output before and after the above fixes.

### CHANGELOG.md (choose one)
- updated

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3592)

### DIFF
--- a/Bindings/Java/Matlab/Utilities/osimC3D.m
+++ b/Bindings/Java/Matlab/Utilities/osimC3D.m
@@ -165,14 +165,24 @@ classdef osimC3D < matlab.mixin.SetGet
                 % and moment columns will have 'p' and 'm' prefixes,
                 % respectively.
                 if ~startsWith(char(labels.get(i)),'f')
-                   columnData = self.forces.updDependentColumnAtIndex(i);
-                   for u = 0 : nRows - 1
-                     % Divide by 1000
-                     columnData.set(u,columnData.get(u).scalarDivideEq(1000));
-                   end
+                    columnData = self.forces.updDependentColumnAtIndex(i);
+                    % Divide by 1000
+                    columnData.multiplyAssign(.001);
                 end
             end
-           disp('Point and Torque values convert from mm and Nmm to m and Nm, respectively')
+            % Convert markers to meters only if in millimeters
+            markersMetaDataUnits = self.markers.getTableMetaDataString("Units");
+            if (strcmpi(markersMetaDataUnits, "mm"))
+                nRows  = self.markers.getNumRows();
+                for i = 0 : self.markers.getNumColumns - 1
+                    columnData = self.markers.updDependentColumnAtIndex(i);
+                    % Divide by 1000
+                    columnData.multiplyAssign(.001);           
+                end
+                % todo fix units in table/trc file to correspond to data
+                
+            end
+            disp('Point and Torque values convert from mm and Nmm to m and Nm, respectively')
         end
         function writeTRC(self,varargin)
             % Write marker data to trc file.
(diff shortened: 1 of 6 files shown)


aymanhab:
@fcanderson Excellent job, definitely exciting news. Can you take a look at the build failures if related? Thanks much 👍

fcanderson:
@aymanhab  Yes!  I will follow up on the build errors.

fcanderson:
@aymanhab The build errors are due to some unresolved function-based path methods.
A few updates to opensim-org:main have landed during the continuous-integration builds of this PR. Perhaps those updates will fix the build errors I'm seeing.  I'm now seeing conflicts that prevent a synch however.  I will try to merge opensim-org:main into this PR tomorrow.  Worst case, I can cancel this PR and resubmit it once opensim-org:main has stabilized.

Issue #3344, opened 2022-12-21
Run time error when trying to visualize the sim


A-Artemis:
Ubuntu 22.04, Python 3.9.15, and opensim 4.4.

I am getting the following errors when I set the visualize=True, it runs when I set the visualize=False.
- `RuntimeError: std::exception in 'SimTK::State & OpenSim::Model::initSystem()': SimTK Exception thrown at VisualizerProtocol.cpp:162:`
  - /home/aurelien/anaconda3/envs/opensim/simbody/libexec/simbody/Final system error was errno=2 (No such file or directory).
  (Required condition 'status == 0' was not met.)
- `RuntimeError: std::exception in 'SimTK::State & OpenSim::Model::initSystem()': std::exception` 

These are the steps I took to create my conda environment

1. conda create -n opensim python=3.9
2. conda activate opensim
3. conda install -c opensim-org opensim=4.4=py39np120

Am I missing an install step to be able to visualize the sim?

aymanhab:
@AlbertoCasasOrtiz not familiar with how/if visualizer works in Conda environment, worth documentation at the least and/or fixing if possible.

AlbertoCasasOrtiz:
@aymanhab I have tested the 2nd colab tutorial setting visualizer to true with `model.setUseVisualizer(True)` and it is not working for me. When it comes to init the system (`state = model.initSystem()`), it hangs forever. If you restart the runtime, it shows the following error:

```
<ipython-input-11-bed98f94f295> in <module>
      1 # Initialize model.
----> 2 state = arm.initSystem()
      3 

/usr/local/lib/python3.8/site-packages/opensim/simulation.py in initSystem(self)
  25869         """
> 25870         return _simulation.Model_initSystem(self)
  25871 

RuntimeError: std::exception in 'SimTK::State & OpenSim::Model::initSystem()': SimTK Exception thrown at VisualizerProtocol.cpp:177:
  Error detected by Simbody method VisualizerProtocol: An attempt to read() 1 bytes from pipe 46 failed with errno=4 (Interrupted system call).
  (Required condition 'retval!=-1' was not met.)
```

mrrezaie:
Hi, I'm using OpenSim moco (4.4.1) conda package (py311) and when running any of these codes, the console crashes and terminates.
```python
model = osim.Model('Rajagopal2015.osim')
model.setUseVisualizer(True)
model.initSystem()
```
```python
model = osim.Model('Rajagopal2015.osim')
[...]

PR #3593, opened 2023-10-31
Support Python local installation and visualizer path

mrrezaie:
Update `setup.py` and `__init__.py` files to fix issues related to local installation and path to the `simbody-visualizer.exe`.

This should work for both Conda and local installation.

Fixes issue #3344

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3593)

### DIFF
--- a/Bindings/Python/__init__.py
+++ b/Bindings/Python/__init__.py
@@ -3,7 +3,9 @@ import os
 if (sys.platform.startswith('win')):
     curFolder = os.path.dirname(os.path.realpath(__file__))
     os.add_dll_directory(curFolder)
-    os.add_dll_directory(os.path.join(os.path.dirname(sys.executable),'Scripts'))
+    opensimCMD = os.path.join(os.path.dirname(sys.executable), 'Scripts', 'opensim-cmd.exe')
+    if os.path.isfile(opensimCMD):
+        os.add_dll_directory(os.path.dirname(opensimCMD))
 
 from .simbody import *
 from .common import *
--- a/Bindings/Python/__init__.py
+++ b/Bindings/Python/__init__.py
@@ -1,6 +1,7 @@
 import sys
 import os
 if (sys.platform.startswith('win')):
+    curFolder = os.path.dirname(os.path.realpath(__file__))
     os.add_dll_directory(curFolder)
     os.add_dll_directory(os.path.join(os.path.dirname(sys.executable),'Scripts'))
 
--- a/Bindings/Python/__init__.py
+++ b/Bindings/Python/__init__.py
@@ -3,7 +3,16 @@ import os
 if (sys.platform.startswith('win')):
     curFolder = os.path.dirname(os.path.realpath(__file__))
     os.add_dll_directory(curFolder)
-    os.environ['PATH'] = curFolder+os.pathsep+os.environ['PATH']
+    # when installed locally via "python -m pip install ."
+    if os.path.isfile(os.path.join(curFolder, 'opensim-cmd.exe')):
+        os.environ['PATH'] = curFolder+os.pathsep+os.environ['PATH']
+    # for Python bindings tests
+    install_path = os.path.join(curFolder, "../../../bin")
+    if (os.path.exists(install_path)):
+        os.add_dll_directory(install_path)
+    dev_path = os.path.join(curFolder, "../../../../Release")
+    if (os.path.exists(dev_path)):
+        os.add_dll_directory(dev_path)
 
 from .simbody import *
 from .common import *
--- a/Bindings/Python/__init__.py
+++ b/Bindings/Python/__init__.py
@@ -1,15 +1,19 @@
 import sys
 import os
-if (sys.platform.startswith('win')):
-      curFolder = os.path.dirname(os.path.realpath(__file__))
-      os.add_dll_directory(curFolder)
-      # in install env, add relative path from init.py to bin, in build environment, we also add "../../../../Release"
-      install_path = os.path.join(curFolder, "../../../bin")
-      if (os.path.exists(install_path)):
-          os.add_dll_directory(install_path)
-      dev_path = os.path.join(curFolder, "../../../../Release")
-      if (os.path.exists(dev_path)):
-          os.add_dll_directory(dev_path)
+
+if sys.platform.startswith('win'):
+    curFolder = os.path.dirname(os.path.realpath(__file__))
+    os.add_dll_directory(curFolder)
+    # in install env, add relative path from init.py to bin, in build environment, we also add "../../../../Release"
+    try: # pass only if DLL_PATH is updated by setup.py in local installation
+        bin_path = DLL_PATH
+        os.add_dll_directory(bin_path)
+        os.environ["PATH"] = bin_path+os.pathsep+os.environ["PATH"] # add to the beginning
+        dev_path = os.path.join(bin_path, "../Release")
+        if (os.path.exists(dev_path)):
+            os.add_dll_directory(dev_path)
+    except:
+        pass
 
 from .simbody import *
 from .common import *
--- a/Bindings/Python/__init__.py
+++ b/Bindings/Python/__init__.py
@@ -1,15 +1,8 @@
 import sys
 import os
 if (sys.platform.startswith('win')):
-      curFolder = os.path.dirname(os.path.realpath(__file__))
-      os.add_dll_directory(curFolder)
-      # in install env, add relative path from init.py to bin, in build environment, we also add "../../../../Release"
-      install_path = os.path.join(curFolder, "../../../bin")
-      if (os.path.exists(install_path)):
-          os.add_dll_directory(install_path)
-      dev_path = os.path.join(curFolder, "../../../../Release")
-      if (os.path.exists(dev_path)):
-          os.add_dll_directory(dev_path)
+    os.add_dll_directory(curFolder)
+    os.add_dll_directory(os.path.join(os.path.dirname(sys.executable),'Scripts'))
 
 from .simbody import *
 from .common import *
--- a/Bindings/Python/__init__.py
+++ b/Bindings/Python/__init__.py
@@ -1,19 +1,15 @@
 import sys
 import os
-
-if sys.platform.startswith('win'):
-    curFolder = os.path.dirname(os.path.realpath(__file__))
-    os.add_dll_directory(curFolder)
-    # in install env, add relative path from init.py to bin, in build environment, we also add "../../../../Release"
-    try: # pass only if DLL_PATH is updated by setup.py in local installation
-        bin_path = DLL_PATH
-        os.add_dll_directory(bin_path)
-        os.environ["PATH"] = bin_path+os.pathsep+os.environ["PATH"] # add to the beginning
-        dev_path = os.path.join(bin_path, "../Release")
-        if (os.path.exists(dev_path)):
-            os.add_dll_directory(dev_path)
-    except:
-        pass
+if (sys.platform.startswith('win')):
+      curFolder = os.path.dirname(os.path.realpath(__file__))
+      os.add_dll_directory(curFolder)
+      # in install env, add relative path from init.py to bin, in build environment, we also add "../../../../Release"
+      install_path = os.path.join(curFolder, "../../../bin")
+      if (os.path.exists(install_path)):
+          os.add_dll_directory(install_path)
+      dev_path = os.path.join(curFolder, "../../../../Release")
+      if (os.path.exists(dev_path)):
+          os.add_dll_directory(dev_path)
 
 from .simbody import *
 from .common import *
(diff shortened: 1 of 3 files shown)


nickbianco:
@aymanhab, what do you think of these proposed changes?

Bindings/Python/__init__.py
```diff
@@ -1,15 +1,19 @@
...
-      dev_path = os.path.join(curFolder, "../../../../Release")
-      if (os.path.exists(dev_path)):
-          os.add_dll_directory(dev_path)
+
+if sys.platform.startswith('win'):
+    curFolder = os.path.dirname(os.path.realpath(__file__))
+    os.add_dll_directory(curFolder)
+    # in install env, add relative path from init.py to bin, in build environment, we also add "../../../../Release"
+    try: # pass only if DLL_PATH is updated by setup.py in local installation
+        bin_path = DLL_PATH
```
aymanhab, line 9: Not sure what DLL_PATH would be set to at this line.

Bindings/Python/__init__.py
```diff
@@ -1,15 +1,19 @@
...
+        os.add_dll_directory(bin_path)
+        os.environ["PATH"] = bin_path+os.pathsep+os.environ["PATH"] # add to the beginning
+        dev_path = os.path.join(bin_path, "../Release")
```
aymanhab, line 12: Python lately discouraged using PATH environment as it violates their having multiple environments with self-contained binaries/libs so I'd lean not to use PATH and instead use relative path. I understand this could be hard if trying to use executables not on the same drive or not part of the package but then this should not be built into these files which are supposed to work out of the box on any os/python version combinations.

aymanhab:
Thanks for working on this @mrrezaie much appreciate your contribution 👍 . The general guideline is to make these files work as smoothly as possible on user machines off the shelf and make it "possible" to work with a local build even with a little bit of work since this is the exception. In our tests we set PYTHONPATH temporarily and this allows us to run the tests without affecting the installation/conda package files. Would that be possible in your case?

Bindings/Python/__init__.py
mrrezaie: The `DLL_PATH` would be updated by `setup.py` to be the OpenSim bin path.

Bindings/Python/__init__.py
mrrezaie: Adding OpenSim bin path to the `os.environ["PATH"]` was as alternative to using the system or user environment path manually. In either way, the `simbody-visualizer.exe` could be invoked properly.

mrrezaie:
@aymanhab, @nickbianco, thank you. I see and will try the `PYTHONPATH`. But the former code, `install_path = os.path.join(curFolder, "../../../bin")` didn't work when I tried to install the 4.4.1 package locally via `python -m pip install .`. Please see: https://github.com/opensim-org/opensim-core/issues/3344#issuecomment-1780203065

mrrezaie:
Did you mean using something like `os.environ["PYTHONPATH"] = bin_path` instead? It couldn't invoke the visualizer.

aymanhab:
@mrrezaie To be clear I'm not disputing that the visualizer doesn't work as is, I'm just trying to figure out with your help a solution that works within the confines of conda since that's the target audience. 

As of 4.4.1, we include executables in the bin folder (e.g. opensim-cmd) but we don't include simbody-visualizer, a good solution would be to include the executable in the bin folder and use relative paths to it to invoke it. If you can try that and it works we can update the conda package(s)/init files and have a solution that works for all instead of relying on users having the executable somewhere else and trying to force python/conda to use it. If dependency on some glut library is needed we can include in the conda recipe on linux as well.

mrrezaie:
Ok, all is already fine with the Conda package [opensim-moco v4.4.1]. I could work with the visualizer as well (only if the miniconda installation and the python working directories have a same drive letter, e.g., C:/).

aymanhab:
To follow up are @mrrezaie are you saying that the visualizer works out of the box in conda package on windows with no other opensim/simbody local builds?

mrrezaie:
I only have miniconda with [OpenSim moco v4.4.1 package](https://anaconda.org/opensim-org/opensim-moco) (no other local builds).

> As of 4.4.1, we include executables in the bin folder (e.g. opensim-cmd) but we don't include simbody-visualizer

The package includes `simbody-visualizer.exe` in `\Library\bin`.

mrrezaie:
> I could work with the visualizer (only if the miniconda installation and the python working directories have a same drive letter)

Should I create a new issue at https://github.com/simbody/simbody ? The visualizer doesn't work when these two directories start with different drive letter:
image

aymanhab:
Great, thank you @mrrezaie I verified that the executable exist in both opensim and opensim-moco packages on windows indeed but not on linux so that explains #3344 will open an issue for that.

As to using dev environment I think the old code didn't work because it made the wrong assumption about the directory where the code is invoked from (can you confirm @nickbianco ) so we should fix but I'd still assume the default build environment layout. Advanced developers with unique configurations should be able to override this using environment variables  and should be documented but probably not baked into distributed files.

aymanhab:
@mrrezaie Yes please file a bug regarding the drive letter, I vaguely remember this coming up in the past so would be good to have on file. Thank you

nickbianco:
@aymanhab, yes, the current version of `__init__.py` gets called from the conda environment that it gets installed in, _not_ the `/Bindings/Python` directory. @mrrezaie's grabs the correct DLL directory within `setup.py` while the relative paths are still valid (i.e. before `__init__.py` gets installed into the conda environment).

I agree that the separate drive bug should be filed in a separate issue.

mrrezaie:
> a solution that works within the confines of conda since that's the target audience.

Given that all is fine with the conda package and this is the target audience, I think you can close this issue. But if you intend to get it working for pip users (users who works with official python, which doesn't have conda installer, and must use local installation instead), I have a suggestion that may work for both pip and conda users.

> I'd lean not to use PATH and instead use relative path.

The solution is to make the installer to copy all the `dll` and `exe` files exist in the bin directory to the Python environment. Both official Python and miniconda/anaconda have `/Scripts` folder for each environment. The bin directory (where `opensim-cmd.exe` exists) would be searched in upper directories automatically. In Conda package, the relative path would be `../bin` and in this case, nothing will be copied to the `Python/Scripts` folder because the conda package already contains all of them. But in the pip installation, the bin path would be in `../../bin` and in this case, the installer will copy all the files to the Python Scripts and OpenSim will find them via `os.add_dll_directory(os.path.join(os.path.dirname(sys.executable),'Scripts'))` in the `__init__.py`. As you see, all paths are relative now.

Please let me know your thoughts.

Bindings/Python/setup.py
```diff
@@ -17,6 +31,7 @@
       url='http://opensim.stanford.edu/',
       license='Apache 2.0',
       packages=['opensim'],
+      data_files = [('Scripts', bin_files)],
```
mrrezaie, line 34: This line will copy all files in the bin folder to the Python environment Scripts folder in pip installation. In conda installation however, the `bin_files` in empty because the package have all the `dll` and `exe` files and will put them into `Library/bin`.

mrrezaie:
Thank you, I'm glad you liked it. 

This is the link to the issue on the simbody project related to the visualizer: https://github.com/simbody/simbody/issues/765

> We can discuss where it belongs since these are not scripts in fact

I chose `/Scripts/` folder because it already contains several `exe` files. Also, this is the only folder that official Python would add to the system environment path, so the visualizer can be invoked.
Conda paths: image
Official Python paths: image

mrrezaie:
> We can discuss where it belongs since these are not scripts in fact

If the `/Scripts` folder looks irrelevant, the dependencies can be placed directly into the package folder (`Lib/site-packages/opensim`), i.e., all `.dll`, `.exe`, and `.py` files will be at a same place. I could find several Python packages with similar structure.

Bindings/Python/__init__.py
```diff
@@ -3,9 +3,7 @@
 if (sys.platform.startswith('win')):
     curFolder = os.path.dirname(os.path.realpath(__file__))
     os.add_dll_directory(curFolder)
-    opensimCMD = os.path.join(sys.exec_prefix, 'Scripts', 'opensim-cmd.exe')
-    if os.path.isfile(opensimCMD):
-        os.add_dll_directory(os.path.dirname(opensimCMD))
+    os.environ['PATH'] = curFolder+os.pathsep+os.environ['PATH']
```
mrrezaie, line 6: This is Python environment-specific, not permanent.

mrrezaie:
Thanks for testing my PR. I see the error raised during testing the local package, `C:\Users\runneradmin\opensim-core-install\sdk\Python\opensim\` (not in the Python installation directory). So, perhaps restoring these lines could fix it (sorry for that). This explains why it couldn't find the `dll` files.
```python
install_path = os.path.join(curFolder, "../../../bin")
if (os.path.exists(install_path)):
    os.add_dll_directory(install_path)
dev_path = os.path.join(curFolder, "../../../../Release")
if (os.path.exists(dev_path)):
    os.add_dll_directory(dev_path)
```

But the Mac failure doesn't seem to be related to this PR.

aymanhab:
Thanks @mrrezaie for the nice contribution. Local testing worked for me, but I'd also like to get input from @nickbianco as a more typical user of the various opensim/python environments.
Once testing passes and before merging will need to add change log entry with your contribution/credit @mrrezaie

Bindings/Python/__init__.py
```diff
@@ -1,15 +1,18 @@
...
-      install_path = os.path.join(curFolder, "../../../bin")
-      if (os.path.exists(install_path)):
-          os.add_dll_directory(install_path)
-      dev_path = os.path.join(curFolder, "../../../../Release")
-      if (os.path.exists(dev_path)):
-          os.add_dll_directory(dev_path)
+    curFolder = os.path.dirname(os.path.realpath(__file__))
+    os.add_dll_directory(curFolder)
+    # when installed locally via "python -m pip install ."
+    if os.path.isfile(os.path.join(curFolder, 'opensim-cmd.exe')):
```
nickbianco, line 7: Add a comment to note that `os.environ['PATH']` only sets the path for the Python environment.

Bindings/Python/setup.py
```diff
@@ -3,6 +3,12 @@
 import os
 import sys
 from setuptools import setup
+
+if os.path.isfile('../../bin/opensim-cmd.exe'): # local installation
```
nickbianco, line 7: Add a comment here or below describing that providing the `bin` folder to `data_files` will copy them into the installation when using a `pip` install.

Bindings/Python/__init__.py
```diff
@@ -1,15 +1,18 @@
...
+        os.environ['PATH'] = curFolder+os.pathsep+os.environ['PATH']
```
nickbianco, line 8: Add spaces around addition operators.

nickbianco (changes requested):
The changes look good overall. Requesting some comments and a minor formatting change.

I'll test the changes locally too.

mrrezaie:
Thank you, I'm glad I could contribute. Do the comments look good?

mrrezaie:
Sorry, I think I made a mistake in updating the CHANGELOG.md. The file was outdated, and a conflict appeared. I tried to resolve it, but I think I made it worse.

nickbianco (approved):
Changes LGTM. It seems like you resolved changes in the change log fine, but something funny happend with the change log diff. I can take a look at it when testing the changes.

Bindings/Python/setup.py
```diff
@@ -17,6 +25,8 @@
       url='http://opensim.stanford.edu/',
       license='Apache 2.0',
       packages=['opensim'],
+      # Copy the bin_files into the opensim package directory
+      data_files = [('Lib/site-packages/opensim', bin_files)],
```
nickbianco, line 29: Remove the spaces around `=`.

nickbianco (changes requested):
Found one small thing to address.

nickbianco:
@mrrezaie I tested locally and everything works fine. After you address my last comment and the tests pass we can merge.

Bindings/Python/setup.py
mrrezaie: Sure.

nickbianco (approved):
LGTM.

aymanhab (approved):
:lgtm:

> Reviewed 2 of 2 files at r7, 1 of 1 files at r8, 1 of 3 files at r9, 1 of 1 files at r11, 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3593#-:-NifzDGlD9aoJ8_BmjvY:bnfp4nl)* status: all files reviewed, 8 unresolved discussions (waiting on @mrrezaie and @nickbianco)

nickbianco:
Merging since the Mac build failure is unrelated (timing out due to issues with Homebrew, which can be resolved in a different PR).

Nice work @mrrezaie!

PR #3596, opened 2023-11-01
Allow zeroth order derivative for Bezier Curve

pepbos:
This PR changes `calcValue` to be a special case of `calcDerivative` with order zero.

This allows for less if / else switches further down the line and focuses calculations to a single function body.
Otherwise does not change the output of `calcValue` or `calcDerivative`.

### Brief summary of changes

In `SmoothSegmentedFunction`
- `calcDerivative` allows order zero.
- `calcValue` calls `calcDerivative` with order zero.

Similarly, in `SegmentedQuinticBezierToolkit`:
- `calcQuinticBezierCurveDerivU` and `calcQuinticBezierCurveDYDX` allow for order zero.
- `calcQuinticBezierCurveCurveVal` calls `calcQuinticBezierCurveDerivU` with order zero.

Changed `MustThrow` unit test accordingly.

### Testing I've completed

- Unit tests are still running.
- No significant change in performance (wall clock time) when simulating Rajagopal.
- No change in simulation result of Rajagopal.

### CHANGELOG.md

Change too minor?

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3596)

### DIFF
--- a/OpenSim/Common/SegmentedQuinticBezierToolkit.cpp
+++ b/OpenSim/Common/SegmentedQuinticBezierToolkit.cpp
@@ -242,46 +242,7 @@ double SegmentedQuinticBezierToolkit::calcQuinticBezierCurveVal(
     double u,
     const SimTK::Vec6& pts)
 {
-    double val = -1;
-
-
-    SimTK_ERRCHK1_ALWAYS( (u>=0 && u <= 1) , 
-        "SegmentedQuinticBezierToolkit::calcQuinticBezierCurveVal", 
-        "Error: double argument u must be between 0.0 and 1.0"
-        "but %f was entered.",u);
-
-    //Compute the Bezier point
-    double p0 = pts(0);
-    double p1 = pts(1);
-    double p2 = pts(2);
-    double p3 = pts(3);
-    double p4 = pts(4);
-    double p5 = pts(5);
-
-    double u5 = 1;
-    double u4 = u;
-    double u3 = u4*u;
-    double u2 = u3*u;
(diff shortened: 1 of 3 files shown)


OpenSim/Common/SegmentedQuinticBezierToolkit.cpp
```diff
@@ -362,15 +323,15 @@ double SegmentedQuinticBezierToolkit::calcQuinticBezierCurveDerivDYDX(
...
     //Bounds checking on the input
     SimTK_ERRCHK_ALWAYS( (u>=0 && u <= 1) , 
         "SegmentedQuinticBezierToolkit::calcQuinticBezierCurveDerivU", 
         "Error: double argument u must be between 0.0 and 1.0.");
 
-    SimTK_ERRCHK_ALWAYS( (order >= 1),
+    SimTK_ERRCHK_ALWAYS( (order >= 0),
         "SegmentedQuinticBezierToolkit::calcQuinticBezierCurveDerivU", 
-        "Error: order must be greater than.");
+        "Error: order must be greater than zero.");
```
nickbianco, line 334: Should be "...greater than or equal to zero."

nickbianco (changes requested):
One minor comment, but otherwise looks good.

@pepbos I think one reviewer for this change is fine, max two.

pepbos:
@nickbianco thanks for reviewing :)

nickbianco (approved):
LGTM.

PR #3600, opened 2023-11-07
Fixes bug introduced by #3596 when handling NaNs.

pepbos:
My Pr #3596 introduced a bug, or triggered a bug elsewhere, when handling NaN's in SmoothSegmentedFunction.

Not sure when NaNs will be injected, and if that even makes sense, but before #3596 that was OK, but currently main will raise an exception. This PR makes it fine again, as was before #3596.

I noticed when trying to visualize my result from a CMC, which raised an exception now.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3600)

### DIFF
--- a/OpenSim/Common/SmoothSegmentedFunction.cpp
+++ b/OpenSim/Common/SmoothSegmentedFunction.cpp
@@ -584,50 +584,52 @@ ________________________________________________________________________
 
 double SmoothSegmentedFunction::calcDerivative(double x, int order) const
 {
-    {
+    const double x0 = _smoothData->_x0;
+    const double x1 = _smoothData->_x1;
+
+    if (x < x0) {
         const double y0 = _smoothData->_y0;
-        const double x0 = _smoothData->_x0;
         const double dydx0 = _smoothData->_dydx0;
-        if (x < x0) {
-            switch (order) {
-                case 0: return y0 + dydx0*(x-x0);
-                case 1: return dydx0;
-                default: return 0.;
-            }
+        switch (order) {
+            case 0: return y0 + dydx0*(x-x0);
+            case 1: return dydx0;
+            default: return 0.;
         }
     }
 
-    {
-        const double x1 = _smoothData->_x1;
+    if (x <= x1) {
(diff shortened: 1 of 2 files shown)


OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -584,50 +584,52 @@ ________________________________________________________________________
...
-            MAXITER);
-
-    const SimTK::Array_<SimTK::Vec6>& ctrlPtsY = _smoothData->_ctrlPtsY;
-    return SegmentedQuinticBezierToolkit::calcQuinticBezierCurveDerivDYDX(
-            u,
-            ctrlPtsX[idx],
-            ctrlPtsY[idx],
-            order);
+    // In case of NaN return zero.
+    return 0.;
```
nickbianco, line 632: Do you think this might be hiding another issue? Is it possible that your CMC simulation/visualization is feeding this function a NaN value for `x` which then gets masked by returning zero here. Or do you think the changes introduced new NaNs somewhere?

nickbianco: I see, this was the previous behavior of the function. In that case, this is probably fine.

pepbos:
Looks like the intention was for the smoothSegmentedFunction to return NaN for a NaN as input (which makes sense).
However it had three different cases:
- `calcValue(NaN)` returned `NaN` 
- `calcDerivative(NaN, 1)` returned `0` 
- `calcValue(NaN, 2)` returned `dydx1`

Changed this PR to fix always returning `NaN` when the input is `NaN`, so:
- `calcValue(NaN)` return `NaN` 
- `calcDerivative(NaN, 1)` return `NaN` 
- `calcDerivative(NaN, 2)` return `NaN` 

Note that without this PR an exception is raised if the input is NaN, which blows up some cases that were fine before.
I added the NaN case to the unittests.

nickbianco (approved):
LGTM. Thanks for the clarification @pepbos.

nickbianco:
Also, you can ignore the Mac failures, that is an issue with the CI. Working on a fix for that.

PR #3601, opened 2023-11-08
adds NaN test SmoothSegmentedFunction unit tests

pepbos:
This PR adds a unit test to `testSmoothSegmentedFunction` to capture the SmoothSegmentedFunction's output for a NaN as input.

When merged this PR will break main due to #3596 being merged, while not capturing the NaN in the same way.

The problem is that after #3596 when visualizing a model (e.g. `opensim-cmd viz model MyModel.osim MyStates.sto`) you get an exception.

### Brief summary of changes

Adds unit test to each of the Millard curves, testing the behavior for NaN as input, up to derivative 2.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3601)

### DIFF
--- a/OpenSim/Common/Test/testSmoothSegmentedFunctionFactory.cpp
+++ b/OpenSim/Common/Test/testSmoothSegmentedFunctionFactory.cpp
@@ -1122,6 +1122,28 @@ void testMonotonicity(SimTK::Matrix mcfSample)
     cout << endl;
 }
 
+/*
+ 5. Tests the output when NaN is the input.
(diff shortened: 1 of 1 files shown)


nickbianco:
Suggestion: if the intent is to have `SmoothSegmentedFunction` return `NaN` with a `NaN` input, then consider adding these changes to #3600. That way you can add the test along with the desired changes to `SmoothSegmentedFunction` and not break `main` upon merging.

PR #3604, opened 2023-11-08
Update CI to use MacOS 12

nickbianco:
Fixes issue #<issue_number>

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3604)

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -33,7 +33,7 @@ jobs:
     - name: Install Python packages
       uses: actions/setup-python@v4
       with:
-        python-version: '3.8'
+        python-version: '3.11'
         
     - name: Install numpy
       #Need numpy to use SWIG numpy typemaps.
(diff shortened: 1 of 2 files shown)


nickbianco:
@aymanhab, @AlbertoCasasOrtiz, I noticed the recent Mac builds have been failing because package installs with Homebrew were taking a very long time (4+ hours). MacOS 11 is no longer officially supported, so I'm testing the CI using MacOS 12.

aymanhab:
Sounds good @nickbianco will keep an eye. One thing to keep in mind is that regardless of github position, we need to check if the build runs on osx11 machines (as some users maybe stuck there).

nickbianco:
@aymanhab, the change seems to be working as intended (aside from some minor failures in `testMocoInverse`, which I can address). Do we know anyone still using MacOS 11? The official EOL was in September, so I'd say it's reasonable to upgrade to the oldest version still supported by Apple.

aymanhab:
@nickbianco I think 3 yrs is long enough lag for users to upgrade, so I wouldn't hold development progress for these rare cases.  There're other CMake variables related to osx version that may allow this build to still run on earlier versions anyway, so I'm good with the change once tests are passing. 👍

nickbianco:
Merging, thanks @aymanhab.

PR #3608, opened 2023-11-11
Template instantiate SetFunctionBasedPath to remove spurious classes

aymanhab:
Fixes issue #0

### Brief summary of changes
template instantiate SetFunctionBasedPath in interface file so swig results are usable in scripting and not bloated with unusable classes
### Testing I've completed
None other than the change/fix f=to the java produced classes

### CHANGELOG.md (choose one)

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3608)

### DIFF
--- a/Bindings/actuators.i
+++ b/Bindings/actuators.i
@@ -12,6 +12,7 @@
 %include <OpenSim/Actuators/Millard2012AccelerationMuscle.h>
 %include <OpenSim/Actuators/McKibbenActuator.h>
 %include <OpenSim/Actuators/DeGrooteFregly2016Muscle.h>
+%template (SetFunctionBasedPath) OpenSim::Set<OpenSim::FunctionBasedPath>;
 
 %include <OpenSim/Actuators/ModelFactory.h>
 
--- a/Bindings/actuators.i
+++ b/Bindings/actuators.i
@@ -12,7 +12,7 @@
 %include <OpenSim/Actuators/Millard2012AccelerationMuscle.h>
 %include <OpenSim/Actuators/McKibbenActuator.h>
 %include <OpenSim/Actuators/DeGrooteFregly2016Muscle.h>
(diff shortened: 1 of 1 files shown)


aymanhab:
@nickbianco this was necessary to make the ModelFactory method usable

Bindings/actuators.i
```diff
@@ -12,6 +12,7 @@
 %include <OpenSim/Actuators/Millard2012AccelerationMuscle.h>
 %include <OpenSim/Actuators/McKibbenActuator.h>
 %include <OpenSim/Actuators/DeGrooteFregly2016Muscle.h>
+%template (SetFunctionBasedPaths) OpenSim::Set<OpenSim::FunctionBasedPath>;
```
nickbianco, line 15: To follow the naming convention of other sets --> `FuctionBasedPathSet`?

nickbianco (changes requested):
One minor suggestion for the set name.

Bindings/actuators.i
aymanhab: In the past, naming for classes made up for swig interface started with Set while ones that existed in C++ used the C++ name which ended in Set, so we have the mix but if you feel strongly about it or plan to typedef in C++ I will make the change so let me know.

nickbianco: Ah okay, I was mixing up name conventions. One more question, should the template name be pluralized, or just be `SetFunctionBasedPath`?

aymanhab: I went back and forth on that and you'll see I initially had it without but looking at simulation.i that we most often use plurals
https://github.com/opensim-org/opensim-core/blob/7703592938979b54dc78ef4561781bbb7cc4994a/Bindings/simulation.i#L149

nickbianco (approved):

aymanhab:
Thanks @nickbianco 🥇

Issue #3589, opened 2023-10-27
No clear method to set the units for marker data table


aymanhab:
Basically metadata in tables have been not exposed to scripting intentionally, and the custom method added to workaround this are all getters, we need a method to set units meta data from matlab.python.

nickbianco:
Does [`addTableMetaData`](https://github.com/opensim-org/opensim-core/blob/81c2455e13f953b4d83bc988107b60ebc3071889/OpenSim/Common/AbstractDataTable.h#L220) not work in scripting?

aymanhab:
Not really, either throws that the key doesn't exist or takes a key, value pair but later ignored and the header is unaffected.

aymanhab:
I think this should be fixed in 4.5

PR #3611, opened 2023-11-16
Check if metadata key exists and if so "Remove" before "add"

aymanhab:
Fixes issue #3589 

### Brief summary of changes
Actually it's possible to addMetaData for units but only if it doesn't exist. I modified the test case and the method in bindings to exercise this workflow. Will update the c3dExport file once we agree on what to include.

### Testing I've completed
Updated Java test case to change units to Meters
Whether update the .i file considering that the name of the method starting with "add", or leave alone and update the test code/sample accordingly basically leaving door for future confusion!

### CHANGELOG.md (choose one)

- no need to update yet

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3611)

### DIFF
--- a/Bindings/Java/tests/TestC3DFileAdapter.java
+++ b/Bindings/Java/tests/TestC3DFileAdapter.java
@@ -18,8 +18,6 @@ class TestC3DFileAdapter {
 
         // convert data to meters and write a copy
         TimeSeriesTableVec3 markerTableInMeters = new TimeSeriesTableVec3(markerTable);
-        if (markerTableInMeters.hasTableMetaDataKey("Units"))
-            markerTableInMeters.removeTableMetaDataKey("Units");
         markerTableInMeters.addTableMetaDataString("Units", "M");
         for (int col=0; col < markerTableInMeters.getNumColumns(); col++)
             markerTableInMeters.updDependentColumnAtIndex(col).multiplyAssign(.001);
--- a/Bindings/Java/tests/TestC3DFileAdapter.java
+++ b/Bindings/Java/tests/TestC3DFileAdapter.java
@@ -18,6 +18,9 @@ class TestC3DFileAdapter {
 
         // convert data to meters and write a copy
         TimeSeriesTableVec3 markerTableInMeters = new TimeSeriesTableVec3(markerTable);
+        if (markerTableInMeters.hasTableMetaDataKey("Units"))
+            markerTableInMeters.removeTableMetaDataKey("Units");
+        markerTableInMeters.addTableMetaDataString("Units", "M");
         for (int col=0; col < markerTableInMeters.getNumColumns(); col++)
             markerTableInMeters.updDependentColumnAtIndex(col).multiplyAssign(.001);
         String markersInMetersFileName = new String("markersMeters.mot");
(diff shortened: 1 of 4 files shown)


aymanhab:
@nickbianco for your review/feedback

nickbianco:
@aymanhab, this change seems reasonable, but is there no way to simply update existing metadata values given a key, rather than removing then adding?

Bindings/Java/tests/TestC3DFileAdapter.java
```diff
@@ -18,6 +18,9 @@ public static void test_C3DFileAdapter() {
 
         // convert data to meters and write a copy
         TimeSeriesTableVec3 markerTableInMeters = new TimeSeriesTableVec3(markerTable);
+        if (markerTableInMeters.hasTableMetaDataKey("Units"))
+            markerTableInMeters.removeTableMetaDataKey("Units");
```
nickbianco, line 22: Is the logic in thsi test necessary given the changes in `common.i`?

nickbianco:
I think the ideal solution is to add a method like `setTableMetaDataForKey` to `DataTable`. It would update the table metadata object, which is a typedef'd `ValueArrayDictionary`, which already has the method [`setValueForKey`](https://github.com/opensim-org/opensim-core/blob/d349581c0d35746191e0264313495815cd1ec101/OpenSim/Common/ValueArrayDictionary.h#L66). In other words, `setTableMetaDataForKey` would be a wrapper for `setValueForKey`.

aymanhab:
Unfortunately the documentation of the method in ValueArrayDictionary.h states
/** Set the value corresponding to a given key. Value is set only if the key
does not already exist.  */

nickbianco:
Oh, weird. That's a poorly named method. Okay, I'm mostly fine with the change, it's just odd to have remove then add to just change a value. 

One more suggestion:  `ValueArrayDictionary` has `updValueForKey` which you can use to modify values. We could add a new method to `AbstractDataTable` called `setTableMetaData` (templated, in scripting it would be `setTableMetaDataString`) which modifies the meta data using `updValueForKey`. There is a mis-match in usages `set` and `upd` between the table and the meta data array here, but considering that there's already a mis-match in the current implementation (`add` versus `set`), I think this would be fine.

aymanhab:
@nickbianco I agree about the wierdness ;) 
Actually there's no updValueForKey there's only updValueArrayForKey, I was going to add the method but considering the unexpected naming (and cloning behavior underneath,) I'd lean to leave it as is and open an issue for future revamp. LEt me know what you think.

nickbianco:
@aymanhab, I see, `updValueArrayForKey` is not exactly what we want. In that case, I think your change to `common.i` is fine, especially since it's just on the bindings side. We can make a proper change sometime in the future.

aymanhab:
Sounds good, thanks @nickbianco 👍 I'll update changelog and then we can merge.

nickbianco:
@aymanhab you should update the test to remove the extra code if you're keeping the `common.i` change.s

aymanhab:
@nickbianco test failure is another intermittent Moco failure on osx.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ and `Blankevoort1991Ligament`, usages of `get_GeometryPath`, `upd_GeometryPath`,
 - Added `PolynomialPathFitter`, A utility class for fitting a set of `FunctionBasedPath`s to existing geometry-path in 
   an OpenSim model using `MultivariatePolynomialFunction`s (#3390)
 - Bumped the version of `ezc3d` which can now Read Kistler files
+- Added method to addTableMetaDataString through Scripting (#3589)
```
nickbianco, line 38: Consider: "Updated scripting method `addTableMetaDataString` to support overwriting metadata for an existing key (#3589)"

nickbianco (changes requested):
One quick change in the CHANGELOG, then LGTM

aymanhab:
Thanks @nickbianco

Issue #3598, opened 2023-11-06
CreateForceSetOfCoordinateActuatorsForModel causes segmentation fault through Python API


itbellix:
Hi,
I am running into troubles when using the method [CreateForceSetOfCoordinateActuatorsForModel](https://github.com/opensim-org/opensim-core/blob/bf77d327bfbc2f00bdeba6610078071387abc08d/OpenSim/Simulation/Model/Model.cpp#L620) for a `CoordinateActuator` object in the Python API.

The script that reproduces the error for me is the following: 
```python
model = opensim.Model(pathOpenSimModel)     # for example, the arm26.osim model
state = model.initSystem()

# append generalized forces to the model
coordActDummy = opensim.CoordinateActuator()    # dummy coordinate actuator to access required function
optimalForce = 1                                # optimal force [N or N/m] for every new actuator
includeLockedAndConstr = False                  # actuators are not created for locked and constrained coordinates
coordActDummy.CreateForceSetOfCoordinateActuatorsForModel(state, model, optimalForce, includeLockedAndConstr)
```

I have tested it with the `arm26.osim` model, but also with other models that I have, and the last line appears problematic (even when changing the `includeLockedAndConstr` to `True`). When running the the provided script, I have a segmentation fault error. However, this does not happen always (i.e., I was able to run the provided code once or twice, but not consistently).

I am using OpenSim in Ubuntu 20.04, through Python, and I see the same problem if I use a [custom build](https://github.com/opensim-org/opensim-core/wiki/Build-Instructions#linux) of OpenSim, or if I use the [conda package](https://anaconda.org/opensim-org/opensim).

nickbianco:
Hi @itbellix, thanks for the bug report. Not sure where the segfault is without setting up a build, but if you're looking for a way to add a set of `CoordinateActuators`s to a model, you could use a `ModelProcessor` with `ModOpAddReserves`:

```python
import opensim as osim

modelProcessor = osim.ModelProcessor(pathOpenSimModel)
optimalForce = 1
modelProcessor.append(osim.ModOpAddReserves(optimalForce))

model = modelProcessor.process()
```

This simply adds a set of `CoordinateActuator`s to the model (for coordinates that do not already have an actuator associated with it), and doesn't mess around with `ForceSet` ownership or invalidating the system like `CreateForceSetOfCoordinateActuatorsForModel` does.

itbellix:
[...]

PR #3612, opened 2023-11-16
Remove `CoordinateActuator::CreateForceSetOfCoordinateActuatorsForModel`

nickbianco:
Fixes issue #3598

### Brief summary of changes

Removes `CoordinateActuator::CreateForceSetOfCoordinateActuatorsForModel`, which is problematic for a few reasons:
- It returns a raw pointer to a new `ForceSet` which is memory-unsafe, especially for scripting users who might not realize this needs to be adopted by a `Model`. 
- Even if you did add this to a `Model`, it would completely overwrite the existing `ForceSet`. A better alternative is `ModOpAddReserves`, which _appends_ a set of `CoordinateActuator`s to the model. 

### Testing I've completed
n/a

As far as I know, not many users use this method, but it has been in the codebase for a long time so this could be a breaking change for some. 

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3612)


nickbianco:
Closing, since this method is actually used by `InverseDynamics` and `StaticOptimization`.

PR #3615, opened 2023-11-17
Simultaneously compute value and derivative of smooth-segmented-function

pepbos:
In Millard2012EquilibriumMuscle the muscle curves are evaluated for the value and first derivative.
In SmoothSegmentedFunction these calls will recompute the exact same spline index, and intermediate variable "u".
This PR adds a function to SmoothSegmentedFunction `calcValueAndFirstDerivative` which computes them both at the same time,  such that the spline index and intermediate variable "u" are only computed once.

Later on this function can then be used by Millard's muscle.

### Brief summary of changes

- Adds anonymous function `calcSelectedDerivatives` to generalize computing different orders of derivative values.
- Adds `SmoothSegmentedFunction::calcValueAndFirstDerivative(x)`

### Testing I've completed

Passed unittests, and checked no change in simulating Rajagopal.

### CHANGELOG.md

- no need to update because minor internal change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3615)

### DIFF
--- a/OpenSim/Common/SmoothSegmentedFunction.cpp
+++ b/OpenSim/Common/SmoothSegmentedFunction.cpp
@@ -604,7 +604,7 @@ static_assert(
 DerivativeValues calcSelectedDerivatives(
     double x,
     const SelectedDerivativeOrders& selectedOrders,
-    const std::shared_ptr<const SmoothSegmentedFunctionData> smoothData)
+    const std::shared_ptr<const SmoothSegmentedFunctionData>& smoothData)
 {
     const double x0 = smoothData->_x0;
     const double x1 = smoothData->_x1;
--- a/OpenSim/Common/SmoothSegmentedFunction.cpp
+++ b/OpenSim/Common/SmoothSegmentedFunction.cpp
@@ -24,6 +24,7 @@
 // INCLUDES
 //=============================================================================
 #include "SmoothSegmentedFunction.h"
+#include <array>
 #include <fstream>
 #include "simmath/internal/SplineFitter.h"
 #include <memory>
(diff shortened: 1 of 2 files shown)


OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -582,59 +583,100 @@ If x is in the linear region
...
     */
 
-double SmoothSegmentedFunction::calcDerivative(double x, int order) const
+namespace
+{
+
+// Used to select which derivative orders to compute:
+// e.g. selectedOrders = {true, false, false, true, false, false} selects the
+// zeroth and third order derivative.
+using SelectedDerivativeOrders = std::array<bool, 6>;
```
adamkewley, line 592: In for a penny, in for a pound - if you're going to use aliases to hide array bounds you mayaswell do it consistently, and with additional compile-time checks, because the code is hard-coding the array bounds in function bodies etc. below

```c++
using SelectedDerivativeOrders = std::array<bool, 6>;
using DerivativeValues = std::array<double, 6>;
static_assert(SelectedDerivativeOrders{}.size() == DerivativeValues.size());
```

pepbos:
@adamkewley thanks for reviewing! Implemented your recommendations, have a look :)
(`.at()` actually caught a bug ;) )

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -582,59 +583,98 @@ If x is in the linear region
...
+{
+    SelectedDerivativeOrders orders{};
+    orders.at(order) = true;
+    return calcSelectedDerivatives(x, orders, _smoothData).at(order);
+}
+
+SmoothSegmentedFunction::ValueAndDerivative SmoothSegmentedFunction::
+    calcValueAndFirstDerivative(double x) const
+{
+    const SelectedDerivativeOrders orders{true, true};
```
adamkewley, line 669: UB? (`true`, `true`, UB, UB, etc.)

pepbos: According to c++, when using list initializer the missing values are value initialized. When using `y = {}` it is UB.

Example from cppreference on `std::array`

```
std::array<int, 2> a7{1}; // List init, unspecified element is value
                                               // initialized, a7[0] = 1, a7[1] = 0
```

So it is ok, no?

OpenSim/Common/SmoothSegmentedFunction.cpp
```diff
@@ -582,59 +583,98 @@ If x is in the linear region
...
+        return {y0 + dydx0 * (x - x0), dydx0};
     }
 
+    if (x > x1) {
+        const double y1    = smoothData->_y1;
+        const double dydx1 = smoothData->_dydx1;
+        return {y1 + dydx1 * (x - x1), dydx1};
+    }
+
+    DerivativeValues y{};
```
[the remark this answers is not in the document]
pepbos: same as above, it is definitely default initialized to zeros.
Unless I misunderstand the docs...

OpenSim/Common/SmoothSegmentedFunction.cpp
adamkewley: My bad, what you are saying is ok:

- https://en.cppreference.com/w/cpp/language/aggregate_initialization

> For a non-union aggregate, each element that is not an explicitly initialized element is initialized as follows:
>
> - If the element has a default member initializer, the element is initialized from that initializer. | (since C++11)
> - Otherwise, if the element is not a reference, the element is copy-initialized from an empty initializer list.
> - Otherwise, the program is ill-formed.

The same document states that `y = {}`, would mean "no explicitly initialized elements" (i.e. they are all value-initialized as above):

> Determine the <i>explicitly initialized elements</i> of the aggregate as follows:
>
> - If the initializer list is a designated initializer list  (the aggregate can only be of class type), the identifier in each  designator shall name a direct non-static data member of the class, and  the explicitly initialized elements of the aggregate are the elements  that are, or contain, those members. 
> - Otherwise, if the initializer list is non-empty, the explicitly initialized elements of the aggregate are the first <i>n</i> elements of the aggregate, where <i>n</i> is the number of elements in the initializer list.
> - Otherwise, the initializer list must be empty (<span class="t-c"><span class="mw-geshi cpp source-cpp"><span class="br0">{</span><span class="br0">}</span></span></span>), and there are no explicitly initialized elements.


The only thing to be aware of when returning different objects is, rather than one, is:

- https://en.cppreference.com/w/cpp/language/copy_elision

Which isn't going to be important for `double`s

nickbianco (approved):
LGTM!

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3615#-:-NjTP7U53OKD2a41d2vo:b-85i648)* status: all files reviewed, 10 unresolved discussions (waiting on @adamkewley and @pepbos)

Issue #3490, opened 2023-06-22
[Feature request] expose more Simbody methods


mrrezaie:
Hi, there are several simbody methods that are not exposed in OpenSim API, e.g.:
[◆ ](https://simbody.github.io/3.7.0/classSimTK_1_1MultibodySystem.html#a1e84e26c6f4e1a8396a65b58faff680a)getRigidBodyForces()
[◆ ](https://simbody.github.io/3.7.0/classSimTK_1_1MultibodySystem.html#ace048e31c9fdc622d3b737026df2e6e5)getMobilityForces()
[◆ ](https://simbody.github.io/3.7.0/classSimTK_1_1Force_1_1Gravity.html#aaa6761c58727d65ff8265a9f2e78b1ee)getBodyForces()

https://simtk.org/plugins/phpBB/viewtopicPhpbb.php?f=91&t=16623&p=0&start=0&view=&sid=5747e9621b5b8c57854568d6d447f3b2

And it would be great if you could define these methods directly into OpenSim::Model class.

Thank you.

aymanhab:
@nickbianco these are other methods to potentially consider from the dev planning meeting

PR #3618, opened 2023-11-20
Expose more simbody methods needed by advanced users

aymanhab:
Fixes issue #3414, #3490

### Brief summary of changes
Added 3 methods previously not accessible to advanced users.

### Testing I've completed
Diffed resulting Model.java files and the three methods were added and are utilizing other existing classes

Documentation/description of the methods and if there're other related questions/concerns

### CHANGELOG.md (choose one)
- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3618)

### DIFF
--- a/OpenSim/Simulation/Model/Model.h
+++ b/OpenSim/Simulation/Model/Model.h
@@ -522,7 +522,21 @@ public:
     GeneralForceSubsystem allocated by this %Model. **/
     SimTK::GeneralForceSubsystem& updForceSubsystem() 
     {   return *_forceSubsystem; }
-
+    /** (Advanced) Get read only access to internal Simbody RigidBodyForces at Dynamics stage **/
+    const SimTK::Vector_<SimTK::SpatialVec>& getRigidBodyForces(const SimTK::State& state)
+    {
+        return getMultibodySystem().getRigidBodyForces(state, SimTK::Stage::Dynamics);
+    }
+    /** (Advanced) Get read only access to internal Simbody Mobility Forces at Dynamics stage **/
+    const SimTK::Vector& getMobilityForces(const SimTK::State& state)
+    {
+        return getMultibodySystem().getMobilityForces(state, SimTK::Stage::Dynamics);
+    }
+    /** (Advanced) Get read only access to internal Simbody Body Forces due to Gravity **/
+    const SimTK::Vector_<SimTK::SpatialVec>& getGravityBodyForces(const SimTK::State& state) const
+    {
+        return getGravityForce().getBodyForces(state);
+    }
     /**@}**/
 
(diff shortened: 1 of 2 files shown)


aymanhab:
@nickbianco not sure if you have a minute to review this. Thank you

OpenSim/Simulation/Model/Model.h
```diff
@@ -522,7 +522,21 @@ OpenSim_OBJECT_NONTEMPLATE_DEFS(Model, ModelComponent);
     GeneralForceSubsystem allocated by this %Model. **/
     SimTK::GeneralForceSubsystem& updForceSubsystem() 
     {   return *_forceSubsystem; }
-
+    /** (Advanced) Get read only access to internal Simbody RigidBodyForces at Dynamics stage **/
+    const SimTK::Vector_<SimTK::SpatialVec>& getRigidBodyForces(const SimTK::State& state)
```
nickbianco, line 526: Does `SimTK::Vector_<SimTK::SpatialVec>` get wrapped properly in the bindings? I don't see a `%template` for it.

nickbianco (commented):
One quick question.

OpenSim/Simulation/Model/Model.h
aymanhab: Yes it does here's the signature from the newly generated Model.java
public VectorOfSpatialVec getGravityBodyForces(State state) {
  ....
}

aymanhab:
Failure is the other intermittent osx failure unrelated to this PR.

OpenSim/Simulation/Model/Model.h
nickbianco: Should we create our own SWIG template to follow the convention for other vectors like `VectorVec3`? Perhaps `VectorSpatialVec`? I definitely wouldn't know to search for the signature `VectorOfSpatialVec` without knowing it existed already. (I'm assuming `VectorOfSpatialVec` is auto-generated by SWIG).

aymanhab:
@nickbianco The name of the class is not swig generated, it's defined here  
https://github.com/opensim-org/opensim-core/blob/4b75cbc2d7007f9ee47c7e82c9f1da3e34027648/Bindings/simbody.i#L220
since at least 2015

I agree there's inconsistency with other naming conventions but it's hard to tell why it was done this way or who is using these classes so a rename would be breaking backward compatibility. We can discuss further at dev meeting

nickbianco:
@aymanhab gotcha, not sure how I missed that while searching. In that case, it doesn't make sense to potentially break backwards-compatibility with this PR.

nickbianco (approved):
LGTM.

aymanhab:
Thanks @nickbianco Much appreciated 👍

PR #3622, opened 2023-11-21
Remove speed computation from force computation for ScalarActuator

pepbos:
This PR removes speed computation from the force computation for children of `ScalarActuator`.

The children of `ScalarActuator` implemented `computeForce` by also calculating the speed, and caching this using `ScalarActuator::setSpeed`. Topologically this seems counter intuitive: You simply do not need speed to compute force.
This can have large implications on performance, but that will be a different PR. First step is to make sure that `computeForce` only computes what it needs to.

To this end, this PR changes the following:
- Changed `ScalarActuator::getSpeed` to a pure virtual function, since this is the required interface for `ScalarActuator` to compute power.
- Removed `ScalarActuator::setSpeed` because it was not used by `ScalarActuator`, so I assumed it not to be an interface requirement.
- Removed speed cache variable from `ScalarActuator` because not all Actuators were actually caching the speed (e.g. `Millard2012EquilibriumMuscle`), so it seems that whether or not to cache, is up to the children.
- For the direct children of `ScalarActuator`, i.e. `TorqueActuator`, `PointActuator`, `PointToPointActuator`, `PistonActuator` and `PathActuator` moved the speed computation out of `computeForce` and into the `getSpeed` impl.
- Added caching of speed variable for the above children where needed.

### Testing I've completed

- Unit tests passing.

### Looking for feedback on

Any feedback on the above considerations is welcome!
Also, the naming convention of `get` in `getSpeed`: Normally a getter will not imply calculations being done, but I see in many parts of the code that a getter will check if a cached variable is valid, and do some calculations if it was not. I tried to follow this structure.
@nickbianco : have a look at `MocoStudyFactory`, I think that it never computed speed in the first place?

### CHANGELOG.md

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3622)


nickbianco (changes requested):
Nice changes @pepbos! I left a bunch of comments.

One thing to consider is that some actuators do rely on speed for force calculation (i.e., muscle force-velocity properties), but since most (all?) muscles use the speed cache variables in derived `AbstractGeometryPath` classes, then changes to the `ScalarActuator` speed cache variable won't matter.

My other main comment is to consider using the `set`, `get`, and `calc` trio of methods to separate the managment of the cache variable and the actual speed calculations (or any other cached calculations) for uniformity and adherence to the `CONTRIBUTING.md` guidelines. 

> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3622#-:-Njn9-1z26MsINsLXXJg:b9y9q47)* status: all files reviewed, 13 unresolved discussions (waiting on @adamkewley, @pepbos, and @tkuchida)

___
*[`OpenSim/Actuators/PointActuator.h` line 108 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3622#-Njn-LS9-ygChWeNfzkw:-Njn-LS9-ygChWeNfzkx:b-whlfm7) ([raw file](https://github.com/opensim-org/opensim-core/blob/01d6bda88776ca29c79c95b9dc4834ffc608790e/OpenSim/Actuators/PointActuator.h#L108)):*
> ```objc
>     //--------------------------------------------------------------------------
>     /** Get speed along force vector. */
>     double getSpeed( const SimTK::State& s) const override;
> ```

Remove extra space before `const`.
___
*[`OpenSim/Actuators/PointActuator.h` line 111 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3622#-Njn-DUG-JT7LaYc9xAT:-Njn-DUG-JT7LaYc9xAU:b-2j3opf) ([raw file](https://github.com/opensim-org/opensim-core/blob/01d6bda88776ca29c79c95b9dc4834ffc608790e/OpenSim/Actuators/PointActuator.h#L111)):*
> ```objc
> 
>     /** Computes speed along force vector. */
>     double computeSpeed( const SimTK::State& s) const;
> ```

`calcSpeed()` per `CONTRIBUTING.md`, and remove extra space.
___
*[`OpenSim/Actuators/PointActuator.cpp` line 158 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3622#-Njn0Ljt1KUFoEak1or5:-Njn0Ljt1KUFoEak1or6:b-yxz8mr) ([raw file](https://github.com/opensim-org/opensim-core/blob/01d6bda88776ca29c79c95b9dc4834ffc608790e/OpenSim/Actuators/PointActuator.cpp#L158)):*
> ```cpp
> 
>     updCacheVariableValue(s, _speedCV) = speed;
>     markCacheVariableValid(s, _speedCV);
> ```

Perhaps not totally necessary, but you could add `setSpeed()` methods for managing the cache variable value like this (this would be in line with the `CONTRIBUTING.md` guidelines). Then you could work with the trio of methods `set`, `get`, and `calc`, where `set`/`get` manage the cache variable and `calc` is called whenever the cache variable is requested (usually by `get`) but is out of date.
___
*[`OpenSim/Actuators/PointToPointActuator.h` line 145 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3622#-Njn143A1uk6LUjwDY9A:-Njn143A1uk6LUjwDY9B:b-bzkx2h) ([raw file](https://github.com/opensim-org/opensim-core/blob/01d6bda88776ca29c79c95b9dc4834ffc608790e/OpenSim/Actuators/PointToPointActuator.h#L145)):*
> ```objc
>     double getStress( const SimTK::State& s ) const override;
>     /** Get speed along force vector. */
>     double getSpeed( const SimTK::State& s) const override;
> ```

Remove extra space before `const`. I see now that other methods have that extra space. You could remove those as well for the `ScalarActuator` derived classes as part of this PR (assuming it's a  minimal addition).
___
*[`OpenSim/Actuators/PointToPointActuator.h` line 147 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3622#-Njn0zdi561aGQ6f0UB8:-Njn0zdi561aGQ6f0UB9:bnc4s2c) ([raw file](https://github.com/opensim-org/opensim-core/blob/01d6bda88776ca29c79c95b9dc4834ffc608790e/OpenSim/Actuators/PointToPointActuator.h#L147)):*
> ```objc
>     double getSpeed( const SimTK::State& s) const override;
>     /** Computes speed along force vector. */
>     double computeSpeed( const SimTK::State& s) const;
> ```

`calcSpeed` and remove extra space.
___
*[`OpenSim/Actuators/PointToPointActuator.cpp` line 156 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3622#-Njn6H2o1lgWUhMDmxY7:-Njn6H2o1lgWUhMDmxY8:bh80urb) ([raw file](https://github.com/opensim-org/opensim-core/blob/01d6bda88776ca29c79c95b9dc4834ffc608790e/OpenSim/Actuators/PointToPointActuator.cpp#L156)):*
> ```cpp
> //_____________________________________________________________________________
> 
> SimTK::UnitVec3 PointToPointActuator::getDirectionBAInGround(
> ```

Per my comment above for the speed cache variable, consider breaking up this code into `get`, `set`, and `calc` methods.
___
*[`OpenSim/Actuators/PointToPointActuator.cpp` line 196 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3622#-Njn6Zh8AQGkHx_MsgjS:-Njn6Zh8AQGkHx_MsgjT:bopb76f) ([raw file](https://github.com/opensim-org/opensim-core/blob/01d6bda88776ca29c79c95b9dc4834ffc608790e/OpenSim/Actuators/PointToPointActuator.cpp#L196)):*
> ```cpp
> 
>     updCacheVariableValue(s, _speedCV) = speed;
>     markCacheVariableValid(s, _speedCV);
> ```

Consider moving to a `set` method.
___
*[`OpenSim/Actuators/TorqueActuator.h` line 154 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3622#-Njn1fOzEUrav8qbcgZ9:-Njn1fOzEUrav8qbcgZA:b2jt0i7) ([raw file](https://github.com/opensim-org/opensim-core/blob/01d6bda88776ca29c79c95b9dc4834ffc608790e/OpenSim/Actuators/TorqueActuator.h#L154)):*
> ```objc
>     double getStress(const SimTK::State& state) const override; 
>     //* Get speed along force vector. */
>     double getSpeed( const SimTK::State& s) const override;
> ```

Remove extra space.
___
*[`OpenSim/Actuators/TorqueActuator.h` line 156 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3622#-Njn1h1I7ffaM7kYL7_8:-Njn1h1I7ffaM7kYL7_9:bnc4s2c) ([raw file](https://github.com/opensim-org/opensim-core/blob/01d6bda88776ca29c79c95b9dc4834ffc608790e/OpenSim/Actuators/TorqueActuator.h#L156)):*
> ```objc
>     double getSpeed( const SimTK::State& s) const override;
>     //* Compute speed along force vector. */
>     double computeSpeed( const SimTK::State& s) const;
> ```

`calcSpeed` and remove extra space.
___
*[`OpenSim/Actuators/TorqueActuator.cpp` line 186 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3622#-Njn6tcy4vci07Hu0LLD:-Njn6tcy4vci07Hu0LLE:bopb76f) ([raw file](https://github.com/opensim-org/opensim-core/blob/01d6bda88776ca29c79c95b9dc4834ffc608790e/OpenSim/Actuators/TorqueActuator.cpp#L186)):*
> ```cpp
> 
>     updCacheVariableValue(s, _speedCV) = speed;
>     markCacheVariableValid(s, _speedCV);
> ```

Consider moving to a `set` method.
___
*[`OpenSim/Examples/CustomActuatorExample/PistonActuator.cpp` line 163 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3622#-Njn89TvDYzeoaIFUtTi:-Njn89TvDYzeoaIFUtTj:boke53l) ([raw file](https://github.com/opensim-org/opensim-core/blob/01d6bda88776ca29c79c95b9dc4834ffc608790e/OpenSim/Examples/CustomActuatorExample/PistonActuator.cpp#L163)):*
> ```cpp
> // SCALAR ACTUATOR INTERFACE
> double PistonActuator::getSpeed(const SimTK::State& s) const
> ```

For consistency with other `ScalarActuator`s: consider making this a `calc` method and adding a cache variable that is managed by `get` and `set` methods.

I know this is just an example class, so not crucial.
___
*[`OpenSim/Simulation/Model/PathActuator.h` line 112 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3622#-Njn22xVBNMhi0SWbri9:-Njn22xVBNMhi0SWbriA:b2jt0i7) ([raw file](https://github.com/opensim-org/opensim-core/blob/01d6bda88776ca29c79c95b9dc4834ffc608790e/OpenSim/Simulation/Model/PathActuator.h#L112)):*
> ```objc
>     // Implement ScalarActuator Interface
>     //--------------------------------------------------------------------------
>     double getSpeed( const SimTK::State& s) const final
> ```

Remove extra space.
___
*[`OpenSim/Simulation/Model/PathActuator.cpp` line 184 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3622#-Njn4BSpBBxvJqs1KyqF:-Njn4BSpBBxvJqs1KyqG:b8uns17) ([raw file](https://github.com/opensim-org/opensim-core/blob/b4d78a2528878b11182bb2953a1dee69810da589/OpenSim/Simulation/Model/PathActuator.cpp#L184)):*
> ```cpp
>     // used to compute power
>     setSpeed(s, speed);
> ```

I can't find anything wrong with removing this, but it does make me wonder why this code is here. Calculating speed before force is necessary for some actuators; for example, muscle force depends on force-velocity relationships which are calculated based on the speed of the path. But this only matters if the derived classes use the cached speed calculation. Muscles use the cached speed calculations from the `GeometryPath` objects they own (i.e., `getLengtheningSpeed()`) and might skip over the cached speed variable.  It's very possible that muscle previously used this cached speed variable until it was replaced by `GeometryPath` and this is just leftover code.

I know tests are passing but I'm always a bit wary of code coverage for things like this. I think it's most likely not an issue, but wanted to lay out my thoughts in case if it's useful for reviewing/testing these changes.

pepbos:
Ty for your feedback @nickbianco! I'll get to implementing those.

Indeed some actuators do need to compute speed when computing the force, just as you say. It's just that not _all_ actuators need to. Also, these are derived classes... Topologically it doesnt really seem to make sense to _require_ speed computation, during a force computation.

Perhaps the `ScalarActuator::setSpeed` function kind of implied that you had to set the speed before the reporter comes along and requests the speed output. If you currently remove speed computation from the force computation you get an exception when the reporter comes along. I'm just guessing here.

But clearly the speed computation is not a requirement for computing the force, because you can split them, as shown.

nickbianco:
@pepbos I agree that actuators shouldn't require speed computations, I was just speculating on why `PathActuator` had them within `computeForce` at all. But after taking a second look, it seems like it was only for power calculations (as the comment suggests), and making `getSpeed()` pure virtual now makes sure that speed is available when needed. In short, feel free to ignore those comments and sorry for the confusion!

pepbos:
Hi @nickbianco , about the `get` `set` `calc` convention, and looking at an example: https://github.com/opensim-org/opensim-core/blob/4a3ae55c3580d9cadd3988b2e3ee3e2c0bcfef3b/OpenSim/Simulation/Model/Muscle.cpp#L547
It looks like we want to cache a variable when it is requested. When adding a `set` method, this method will still be called inside the `get` method (instead of `calc`), because get is what others use as interface...

I think the current caching behavior messes a bit with the `get` `set` and `calc` convention...

nickbianco (approved):
Hi @pepbos, I agree and thanks for pointing out that example. I've retracted the suggestions for adding the `set` methods. 

Everything else LGTM.

> Reviewed 8 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3622#-:-NjxANq57LTOk-U7X-l6:bazt6s0)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adamkewley and @tkuchida)

tkuchida (commented):
Just a couple minor things ⬇️

pepbos:
ty @tkuchida for reviewing, I changed return values to NaN as you suggested, makes more sense

tkuchida (approved):

PR #3624, opened 2023-11-22
Update version to 4.5

aymanhab:
Fixes issue #0
### Brief summary of changes
Upgrade opensim release version to 4.5 to handle downstream effects,  branching and tagging to follow

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3624)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,8 +28,8 @@ include(FeatureSummary)
 # OpenSim version.
 # ----------------
 set(OPENSIM_MAJOR_VERSION 4)
-set(OPENSIM_MINOR_VERSION 4)
-set(OPENSIM_PATCH_VERSION 1)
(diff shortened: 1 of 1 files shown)


aymanhab:
@nickbianco would be good to merge this version change so we can get the version correctly downstream in conda packages and the gui

nickbianco (approved):
LGTM.

Just double checking: you're not creating the `4.5` tag yet, correct? Since we have a few more changes to merge.

aymanhab:
Indeed no tag yet, we typically create a branch for fixes post release candidate and then create a tag when all is frozen/finalized.

aymanhab:
Indeed no tag yet, we typically create a branch for fixes post release candidate and then create a tag when all is frozen/finalized.

PR #3625, opened 2023-11-22
Update settings for `exampleMocoTrack`

nickbianco:
### Brief summary of changes
This is a minor PR to update the settings for `exampleMocoTrack` to produce better muscle activity solutions after the recent updates incorporating `FunctionBasedPath` functionality. I've also updated the Moco change log to capture all recent changes to this example and `exampleMocoInverse`.

### Testing I've completed
Ran and compared all results locally.

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3625)

### DIFF
--- a/Bindings/Java/Matlab/examples/Moco/example3DWalking/exampleMocoTrack.m
+++ b/Bindings/Java/Matlab/examples/Moco/example3DWalking/exampleMocoTrack.m
@@ -176,6 +176,7 @@ study = track.initialize();
 % problem by default.
 problem = study.updProblem();
 effort = MocoControlGoal.safeDownCast(problem.updGoal("control_effort"));
+effort.setWeight(0.1);
 
 % Put a large weight on the pelvis CoordinateActuators, which act as the
 % residual, or 'hand-of-god', forces which we would like to keep as small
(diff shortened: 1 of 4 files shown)


aymanhab (approved):
:lgtm: Not sure where the numbers 0.1, 1e-3, 1e-4 came from, if these are not in-line with other Goals in the problem formulation then it'd be good to explain the rationale, otherwise you can ignore.

> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3625#-:-NjtTlye-nVFy3rD8f6e:bkivx0d)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nickbianco)

nickbianco:
> Not sure where the numbers 0.1, 1e-3, 1e-4 came from, if these are not in-line with other Goals in the problem formulation then it'd be good to explain the rationale, otherwise you can ignore.

These are changes from the default settings that happen to work well for this problem. I don't think any extra explanation is necessary for them.

PR #3627, opened 2023-11-23
Change testMarkerData.cpp to use Catch (#3555)

adamkewley:
Fixes issue #3555

### Brief summary of changes

- Removed existing `main` and replaced it with Catch macro

### Testing I've completed

- It ran before and it ran after

### CHANGELOG.md (choose one)

- no need to update because irrelevant

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3627)


nickbianco (approved):
LGTM.

PR #3628, opened 2023-11-23
Change testRootSolver.cpp to use Catch (#3555)

adamkewley:
Fixes issue #3555 

### Brief summary of changes

- Changed `testRootSolver.cpp` to use Catch instead of having its own `main` function

### Testing I've completed

- Ran it before and after

### CHANGELOG.md (choose one)

- no need to update because... because.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3628)


nickbianco (approved):
LGTM.

PR #3638, opened 2023-12-05
Update simbody to 24b5fa

adamkewley:
Fixes issue N/A

Latest simbody contains useful fixes:

- https://github.com/simbody/simbody/pull/771
  - Tells CMake that `simbody` targets, on Windows, depend on `libgcc_s_sjlj-1.dll` etc.
- https://github.com/simbody/simbody/pull/770
  - Removes `std::iterator` from an XML class, because `std::iterator` is detected as deprecated by MSVC when warnings are increased

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3638)


nickbianco (approved):
LGTM.

PR #3640, opened 2023-12-09
Fix typo in Linux/MacOS build scripts

RTnhN:
Fixes issue N/A

### Brief summary of changes

In the help text for the linux build script, there was a typo. The help text did not agree with the rest of the code. It also did not agree with the Windows version of the build script. 

### Testing I've completed

Since this is just a change to the help text, I don't think that rigorous testing is needed. 

### CHANGELOG.md (choose one)

I don't think that it is needed to add this to the change log.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3640)


RTnhN:
I accidentally pushed some other changes to this PR. I think I would also like to combine both typos in the same commit. I will try to open a new PR later this week.

Issue #3605, opened 2023-11-10
Support for `Controller`s in Moco


nickbianco:
Adding support for `Controller`s will require:

1. Updating the logic in the solvers to allow other controller types ([CasADi](https://github.com/opensim-org/opensim-core/blob/d0e1228f2c89c58d38acd396004990f41255964e/OpenSim/Moco/MocoCasADiSolver/MocoCasOCProblem.cpp#L46), [tropter](https://github.com/opensim-org/opensim-core/blob/d0e1228f2c89c58d38acd396004990f41255964e/OpenSim/Moco/tropter/TropterProblem.h#L68)).
2. For each controller, determine how model controls are added (or not added) to the OCP. For example, with a `PrescribedController` , a user might want the controller to solely provide the controls to the model. 
3. If all `Controller`s use `addInControls` to update the model controls, provide the option to create OCP controls on top of the `Controller` controls.
4. Provide hooks for the `SynergyController` class (i.e., support `Controller`s with `Input`s).

nickbianco:
I've encountered some obstacles while working on PR #3626. While I'm only trying to implement support for `PrescribedController` in that PR, I want it include a generic implementation that will support upcoming controller types (i.e., `SynergyController`). This includes controllers that have a list `Input` for wiring external controls from Moco into the model. However, recent changes to `Controller` fixing memory leaks using a custom copy construtor prevent me from using `Input`s (which do not support custom copy constructors).

Fixing the `Controller` memory leak without a custom copy constructor means using `Socket`s to manage the list of `Actuator`s for a particular controller. Unfortunately, OpenSim currently does not support list `Socket`s, which would be the ideal way to handle this solution. I could create an itermediate class with a `Socket` for an `Actuator`, but for the sake of not replace hacky code with slightly-less-hacky code, I'm going to add support for list `Socket`s, especially since support for these seemed to be intended in the first place.

Down the rabbit hole we go...

PR #3652, opened 2023-12-14
Add support for list `Socket`s

nickbianco:
Partially fixes issue #3605

### Brief summary of changes

This PR adds support for list `Socket`s via the new macro `OpenSim_DECLARE_LIST_SOCKET`. This addition is motivated by current work to support `Controller`s in Moco. The need for this change is explained in #3605.

In addition to the macro, `Component` and `Socket`  has new `getConnectee()` signatures that accept an index for list socket support. Most of the remaining changes are implementation details (largely based on how list `Input`s are constructed). 

### Testing I've completed

Updated a few tests in `testComponentInterface.cpp` to test list `Socket`s. These tests include serialization/deserialization.

Some of the checks on the connectee paths could probably be simplified (see the TODOs).

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3652)


OpenSim/Common/Component.h
```diff
@@ -3358,39 +3456,50 @@ void Socket<C>::finalizeConnection(const Component& root) {
...
+                rootOfConnectee.getName() + "' to '" + myRoot.getName() + "'?");
+
+            ComponentPath connecteePath = connectee->getRelativePath(getOwner());
+            // If the relative path starts with ".." then use an absolute path
+            // instead.
+            if (connecteePath.getNumPathLevels() > 1 &&
+                    connecteePath.getSubcomponentNameAtLevel(0) == "..")
+                connecteePath = connectee->getAbsolutePath();
+
+            if (isListSocket()) {
```
adamkewley, line 3480: This append vs. assign behavior is giving me the itchies

aymanhab (changes requested):
@nickbianco Had only a few comments in addition to @adamkewley great review, I'll take another pass once you address the feedback so far. 

> Reviewed 6 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3652#-:-Nle1InL-x6cp4YaPOth:bts9i8v)* status: 6 of 7 files reviewed, 17 unresolved discussions (waiting on @adamkewley and @nickbianco)

___
*[`CHANGELOG.md` line 10 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#-NldifT65iKhgo6MdWI0:-NldifT65iKhgo6MdWI1:b-exe6lv) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/CHANGELOG.md#L10)):*
> ```Markdown
> 
> v4.5.1
> ======
> ```

I'd use 4.6 just because we're changing API
___
*[`OpenSim/Common/Component.h` line 2790 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426628122:-Nldm0Zl11S4zAf5H3mR:b-t1me6o) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/Component.h#L2790)):*

This API design is the opposite of the properties one:

```c++
ObjectProperty(const std::string& name, bool isOneValue);
SimpleProperty(const std::string& name, bool isOneValue);
```

Naming convention is `comment` in the base class:

```c++
AbstractProperty(const std::string& name,  const std::string& comment);
```

But the concrete classes don't create comments this way, e.g.:

```c++
// Object.h
template <class T> PropertyIndex Object::
addProperty(const std::string& name, 
            const std::string& comment,
            const T&           value)
{
    // Restrict to exactly one value. If there is no name, this will throw
    // an exception if T is a simple (non-object) type.
    Property<T>* p = Property<T>::TypeHelper::create(name, true);

    p->setComment(comment);
    p->appendValue(value);
    p->setValueIsDefault(true);

    // Note that an unnamed, one-object property will use the object class name
    // as a name for lookup purposes.
    return PropertyIndex(_propertyTable.adoptProperty(p));
}
```

So be aware that there might be an inconsistency in your macro/adder implementations.
</blockquote></details>

I agree with @adamkewley about the inconsistency (default to one vs. default to list) but considering the wide existing usage of sockets I'd avoid any changes that break existing code that uses sockets. We use lists extensively in Properties while there're 28 instances of sockets that never needed to be lists.
___
*[`OpenSim/Common/Component.h` line 3440 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426637492:-Nldo2CuCl89FaDMSqEY:bo9a66r) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/Component.h#L3440)):*

`_connectees.at(index).getRef()`: this is a situation where you can eat an extra bounds-check: it's a potentially-user-facing API (as in, rando python/Matlab scripts could use this) and a runtime boundscheck exception is infinitely preferable to a user randomly reporting segfaults or unusual bugs.
</blockquote></details>

Agree @adamkewley  :+1:  
___
*[`OpenSim/Common/Component.cpp` line 172 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426607102:-NldzOTOAVfFBGcvsH4U:bl754jz) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/Component.cpp#L172)):*

It's overkill if you reasonably expect users (or, more likely, a mixture of paths from an `osim` file and tooling like `opensim-creator`) to provide a mixture of path formats - I'd say so (if that makes sense for this part of the codebase).

I'd think through whether all paths are rectified to one format by this point in the system, or whether they are the paths provided via a user- (API or real) provided path and stick to "be permissive in what you accept".
</blockquote></details>

Not sure where this falls in the user-facing/XML-facing API. At some point the permissiveness has to stop and enforcement begins so downstream code has reasonable pre-conditions. It's not clear to me where this line is.
___
*[`OpenSim/Common/ComponentSocket.h` line 1044 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426648334:-Nle0M8F5pHosq_aGjpC:b-5mbaz9) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/ComponentSocket.h#L1044)):*

Ensure this is similar to the property-based macro, so that people reading macros aren't trying to figure out API differences etc.
</blockquote></details>

"People reading macros for fun"!!
___
*[`OpenSim/Common/ComponentSocket.h` line 1083 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426651662:-Nle0jXA7mIVURPLUuUI:b-8fx2u7) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/ComponentSocket.h#L1083)):*

Don't bother writing that you _may_ do something in the future. The documentation is for today. Similar with TODOs (actually, some projects prohibit having TODOs in the main branch exactly for this reason)

I *may* make OSC's renderer nicer in the future, or become a plumber, etc.
</blockquote></details>

:+1:

nickbianco (commented):
Thanks @adamkewley and @adamkewley for a quick and detailed review. I've addressed some of the low-hanging fruit as a start.

Since I based much of the implementation on how list `Input`s are implemented, addressing the remaining comments would imply changing the abstraction at the `AbstractSocket` level. For example, `Input`s use `connectInput` for both connecting an input (in a single `Input`) and appending an input (in a list `Input`). So while creating separate `connectSocket` and `appendSocket` auto-generated functions would seem more consistent with the existing `Socket` API, it would be inconsistent with the current `Input` API. I agree that this would be clearer, but it would then introduce a API change for `Input`s (although maybe it's in for a penny, in for a pound at this point).

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3652#-:-NlenkTo7aIQtbK0htjz:b-jmrpvy)* status: 6 of 7 files reviewed, 17 unresolved discussions (waiting on @adamkewley and @aymanhab)

___
*[`CHANGELOG.md` line 10 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#-NldifT65iKhgo6MdWI0:-NleZgT84ErklIlLhPDX:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/CHANGELOG.md#L10)):*

I'd use 4.6 just because we're changing API
</blockquote></details>

Done.
___
*[`OpenSim/Common/Component.h` line 177 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426615817:-NleakTR6xvqfMvNqSgt:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/Component.h#L177)):*

Do you anticipate that callers are going to specifically `catch` this exception? If not, keep it generic.

And put the implementation in the `cpp` file: there's no need for OpenSim's hundreds of other compilation units to inline+parse+optimize this.
</blockquote></details>

Done.
___
*[`OpenSim/Common/Component.h` line 958 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426619578:-NlefEJj8MRmk0JlrU9r:b-shc9f9) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/Component.h#L958)):*

(I went through the process and the only case in which a custom exception is actually caught in `opensim-creator` is a `SocketNotFound` one, because the OpenSim API didn't/doesn't have a way of checking whether a component has a socket with a given name).
</blockquote></details>

That would be a useful exercise, my usages of exceptions and asserts isn't very disciplined. 

Changing this (and other similar `Exception`s in this file) to use `OPENSIM_ASSERT_FRMOBJ`.
___
*[`OpenSim/Common/Component.h` line 3432 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426635092:-NleePKs3uwtwiCUL4qF:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/Component.h#L3432)):*

Bug prone: if a user manages to wrangle a negative number through the system, calling this function will then underflow it. Use similar conventions to the rest of OpenSim's API (i.e. `int`)

(I mean, i'm superficially in agreement that the indexing of OpenSim, in general, should match stdlib--i.e. it should be `size_t`--but `size_t` is actually a known mistake of stdlib. However, this usage of `unsigned` here manages to be: a) inconsistent with all other OpenSim APIs, b) inconsistent with stdlib, and c) as bug-prone as `size_t`.
</blockquote></details>

Done.
___
*[`OpenSim/Common/Component.h` line 3440 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426637492:-NleeOme8hlRvInLwR86:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/Component.h#L3440)):*

Agree @adamkewley  :+1:
</blockquote></details>

Done.
___
*[`OpenSim/Common/Component.h` line 3489 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426641804:-NlegZhVBc1Mm0sYz1d5:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/Component.h#L3489)):*

`unsigned` >,< (etc. etc. see above) - it's usually mostly used for bit manipulation/flags
</blockquote></details>

Done.
___
*[`OpenSim/Common/Component.cpp` line 172 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426607102:-Nlen784AIsq8bKM__Tk:bgh2ipw) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/Component.cpp#L172)):*

Not sure where this falls in the user-facing/XML-facing API. At some point the permissiveness has to stop and enforcement begins so downstream code has reasonable pre-conditions. It's not clear to me where this line is.
</blockquote></details>

I'm realizing now that this change doesn't make sense at all. The original check here was to only apply prepend logic if  a particular connectee is available within the subcomponent being added to the model (based on [this PR from a few years ago](https://github.com/opensim-org/opensim-core/pull/2504)). Revising this to use new (hopefully much simpler) logic.
___
*[`OpenSim/Common/Component.cpp` line 174 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426614269:-NlenaEMDH4_Gqqn55oq:b1jd2h3) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/Component.cpp#L174)):*

I'd refactor this into a helper method or something, more than 3 levels of code indentation is when things get a bit shakey
</blockquote></details>

Will refactor based on my comment above.

nickbianco (commented):
@adamkewley, @aymanhab ready for review.

I've refactored code wherever I thought it made sense for this PR, but I'd vote to punt on some of the other suggested changes since it would require significant refactoring to `Socket`s and `Input`s. Not opposed to this, but I'd probably want to make those changes in a different PR. 

I think the main API change to address for this PR is the `connect` versus `append` semantics for `Socket`s and `Input`s, since that is directly related to the new `Socket` functionality.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3652#-:-Nlty8huAxIEdgcr71ck:b-88h6bz)* status: 1 of 7 files reviewed, 17 unresolved discussions (waiting on @adamkewley and @aymanhab)

___
*[`OpenSim/Common/Component.h` line 2790 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426628122:-NltxvWW38Dr53sv7Ale:bm50wc) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/Component.h#L2790)):*

I agree with @adamkewley about the inconsistency (default to one vs. default to list) but considering the wide existing usage of sockets I'd avoid any changes that break existing code that uses sockets. We use lists extensively in Properties while there're 28 instances of sockets that never needed to be lists.
</blockquote></details>

I also agree, but making this change without largely refactoring `Socket`s seems difficult. I'd vote to move this is a larger refactoring effort for sockets in a future PR.
___
*[`OpenSim/Common/Component.h` line 3419 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426631305:-Nlt_Hfi0ZpXoGrZ4OtY:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/Component.h#L3419)):*

Inconsistent with list properties:

```c++
const T& getValue(int index=-1) const {
    if (index < 0) {
        if (getMaxListSize()==1) index = 0;
        else throw OpenSim::Exception(
            "Property<T>::getValue(): an index must be "
            "provided for a property that takes a list "
            "of values.");
    }
    return getValueVirtual(index);
}
```
</blockquote></details>

Done.
___
*[`OpenSim/Common/Component.h` line 3480 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426695522:-NltsHB0BoQvfVvdgRja:bg40i79) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/Component.h#L3480)):*

This append vs. assign behavior is giving me the itchies
</blockquote></details>

I moved the responsibility of this behavior to `AbstractSocket`. Note that I've made direct access to the underlying `Property` now `private` within `AbstractSocket` to point users to the new assignment method.
___
*[`OpenSim/Common/Component.h` line 3488 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426639230:-NltsZQ85vVaJqliHRox:bb10xzi) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/Component.h#L3488)):*

Hmmmmmmmm, this, and the other `isListSocket`-type checks kind of feel like there's something missing in the abstraction
</blockquote></details>

I added a new method to `AbstractSocket` to handle this check generically. 
___
*[`OpenSim/Common/ComponentSocket.h` line 368 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426646857:-Nltedxt1e742-TnoUJa:b1i4vne) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/ComponentSocket.h#L368)):*

See similar comments about `Property`'s design etc.
</blockquote></details>

I've tried to make this conform to `Property`'s design, but it introduces a default argument in the virtual function defined in `AbstractSocket`. Clang tidy issues a warning about it, and some quick Googling suggests this isn't an ideal design.

I suppose we could remove this virtual function from `AbstractSocket` since it might not be needed for the interface (`Input`s do not connect to `Object`s). 
___
*[`OpenSim/Common/ComponentSocket.h` line 1044 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426648334:-NltshE9DfwuUbnDky_i:bsybul4) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/ComponentSocket.h#L1044)):*

"People reading macros for fun"!!
</blockquote></details>

This macro is directly based on the other `Socket` and `Input` macros in this file. I'd vote to push this update to a larger refactoring effort for sockets.
___
*[`OpenSim/Common/ComponentSocket.h` line 1083 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426651662:-Nlt_shgCikBgGPYaZ1q:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/ComponentSocket.h#L1083)):*

:+1:
</blockquote></details>

Done.
___
*[`OpenSim/Common/Test/testComponentInterface.cpp` line 586 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426662985:-Nltfwkt8vZnbK-9tEZ3:b-zgq5st) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/Test/testComponentInterface.cpp#L586)):*

Does `connectSocket` now *append*, rather than *overwrite* the socket?
</blockquote></details>

Yes, this is based on how list `Input` works. See my comment below.
___
*[`OpenSim/Common/Test/testComponentInterface.cpp` line 1022 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426667150:-Nltg6hi0-8ylUWH-0Ck:bwc0eii) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/Test/testComponentInterface.cpp#L1022)):*

This feels very inconsistent with the original API.

The original is effectively `socketConnectee = newConnectee` (via `connectSocket`), but yours is effectively saying, at runtime, "assign if singular, append if list", which (imo) will lead to very very surprising behavior in codebases that aren't "list aware" (i.e. any codebases written before this patch).

E.g. if `opensim-creator` has a UI for reassigning a socket (it does), your new API could potentially lead to code that would just keep appending more connectees.

The semantics should match `Property`:

- `setValue` (`connectSocket`): set the value of one slot in the object, or multiple slots if the rhs is a vector?
- `appendValue` (`appendConnectee`): append a new value


</blockquote></details>

I agree with your suggestions here. The only reason it is written this way was to be consistent with list `Input` behavior. I'm happy to make this change, but it would imply making the same change for `Input` which would change this existing API. If that's fine with you and @aymanhab , I'll go ahead and do it.

adamkewley:
Ah, I wasn't aware that `Input` had that behavior: it's not an API I've used!

nickbianco:
After comparing to `Property`, I can see how it can be a bit confusing. Part of the current design might be because `Socket`s and `Input`s add an internally managed property to handle serialization and deserialization. Although, I think we could probably bring these APIs closer together (in a future PR, of course 😄).

nickbianco (commented):
@adamkewley, @aymanhab, coming back to finishing this up now that I'm back from holiday break. I think the last thing to decide on is the `append` vs `connect` convention for list `Socket`s (including `Input`s). I've laid out some thoughts under @adamkewley's comment above. Essentially, having a more `Property`-like convention would change the existing API a bit, so I'm looking for a bit of discussion before I make any sweeping changes.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3652#-:-NnMBl3VDQyOvoVcF5Z5:b-56hbkp)* status: 1 of 7 files reviewed, 17 unresolved discussions (waiting on @adamkewley and @aymanhab)

___
*[`OpenSim/Common/Test/testComponentInterface.cpp` line 1022 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426667150:-NnM-f1a07CNWzqMN2Yk:bl62drq) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/Test/testComponentInterface.cpp#L1022)):*

I agree with your suggestions here. The only reason it is written this way was to be consistent with list `Input` behavior. I'm happy to make this change, but it would imply making the same change for `Input` which would change this existing API. If that's fine with you and @aymanhab , I'll go ahead and do it.
</blockquote></details>

Looking into this more closely now that I'm back from holiday break. For non-list `Socket`s, the macro-generated `connectSocket` function wraps virtual function `connect` defined in `AbstractSocket`:

```c++
/** Connect this %Socket to the provided connectee object. If this is a
    list socket, the connectee is appended to the list of connectees;
    otherwise, the provided connectee replaces the single connectee. */
virtual void connect(const Object& connectee) = 0;
```

So I guess the original intention for term "connect" was to either assign or append, depending on whether it was non-list or list. `Input` follows this convention, although it defines its own `connect` methods (since it does not connect to `Object`(s)). 

Currently, if I added an  `appendConnectee` method it would still just wrap `connect`, which appends or assigns the `Object` to the `Socket` under the hood depending on list/non-list status. In other words, would it be a satisfactory design if the methods were named differently so it was clear you were working with a list or non-list `Socket`, even if the code under the hood is the same? Also, I'm not sure what the best way to write `connectSocket` for list `Socket`s would be since the rhs would be a list of const references to `Object`s. In addition, `Input` should probably follow this convention, namely, for list `Input`s, [the method for adding channels](https://github.com/opensim-org/opensim-core/blob/4c76879d2c2d9a01697fe51bc0a460d362a83f3c/OpenSim/Common/ComponentSocket.h#L1276) would switch to use "append" terminology rather than "connect".

adamkewley:
I'm unsure what's best, it isn't an API I have used (yet).

I'd vote for designing it to be logical, rather than backwards-compatible, in this case. I can't imagine that there are many downstream implementors of the virtual API (`connect`) outside of this repository, and I can't imagine that many codebases are using the `Input` API much (yet). E.g. even the `Output` API (arguably, easier to integrate) is mostly only used by OSC right now?

aymanhab (commented):
I agree with @adamkewley about being more logical than backwards-compatible, however Properties are commonly lists so the interface should reflect that while inputs and sockets are not, so making the interface simpler for most common use cases would be my choice. Disclaimer, I haven't inputs/sockets since forever but use Properties regularly.

> Reviewed 1 of 7 files at r1, 3 of 5 files at r2, 1 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3652#-:-NneWec6BONyiK9n4nRy:b676tfg)* status: 5 of 7 files reviewed, 16 unresolved discussions (waiting on @adamkewley)

___
*[`OpenSim/Common/Component.h` line 2790 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3652#gh-1426628122:-NlyOiJa0Hi174KE27YU:bno992) ([raw file](https://github.com/opensim-org/opensim-core/blob/14041cbbd2403cba7084a6b27cce5760aff3afc1/OpenSim/Common/Component.h#L2790)):*

I also agree, but making this change without largely refactoring `Socket`s seems difficult. I'd vote to move this is a larger refactoring effort for sockets in a future PR.
</blockquote></details>

Agree @nickbianco

nickbianco (commented):
Thanks for the input guys! 

After considering a few options, I think the best approach is to use `appendConnectee` for the macro function for list `Socket`s and leave the current implementation of `Input`s alone. I also fixed an error with the output type for the `getConnecteeAsObject` methods. 

I think this is the last thing to wrap this up, but let me know if you had any final comments/suggestions.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3652#-:-NnfQPpg1e4Xhs329nKE:b-gccbzj)* status: 3 of 7 files reviewed, 16 unresolved discussions (waiting on @adamkewley and @aymanhab)

nickbianco:
@adamkewley, @aymanhab ready for review.

aymanhab (approved):
Left one comment on function name, not sure how relevant for the clients of this API so feel free to ignore if all @adamkewley  concerns are addressed.
:lgtm:

> Reviewed 1 of 5 files at r2, 1 of 2 files at r4, 1 of 2 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3652#-:-NnwapG2EYyXZvnDBCXE:b-elkm2k)* status: all files reviewed, 17 unresolved discussions (waiting on @adamkewley and @nickbianco)

___
*[`OpenSim/Common/ComponentSocket.h` line 1143 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3652#-Nnw_e-WDzGxX08QIyLt:-Nnw_e-X0dbmavUwRsZR:b9evd9j) ([raw file](https://github.com/opensim-org/opensim-core/blob/67306dd20b13fe0da6a933d3b554c23f264425e0/OpenSim/Common/ComponentSocket.h#L1143)):*
> ```objc
>     /** connectee path property. The reference to the connectee set here */ \
>     /** takes precedence over the connectee path property.               */ \
>     void appendConnectee_##cname(const OpenSim::Object& object) {           \
> ```

Not sure if not having "Socket" anywhere in the function name now would be an issue for potential users. You're likely in better position to make this call @nickbianco

nickbianco:
@adamkewley good to merge?

adamkewley (commented):
It is fine in principle, but I don't believe you responded to or fixed the things I previously commented on?

nickbianco:
Thanks @adamkewley and @aymanhab for the detailed reviews! This one was tricky.

PR #3653, opened 2023-12-14
Combines computing value and derivative of muscle curves in Millard

pepbos:
This PR replaces separate calls to `calcValue` and `calcDerivative` on the muscle curves in `Millard2012EquilibriumMuscle` by `calcValueAndDerivative`, which is more efficient.

The `MuscleLengthInfo::userDefinedLengthExtras` and  `FiberVelocityInfo::userDefinedVelocityExtras` buffers are used to store the derivatives of the curves until needed.

Measuring the wallclock time of Rajagopal forward integration, and CMC running-model (using Millard muscles) shows around 10% perf benefit:

| | main | with this PR |
|---|---|---|
|CMC-Running Model | 25.8 [s] | 22.9 [s] (-12.4%) |
|Rajagopal Forward Integration | 6.4 [s] | 5.8 [s] ( -9.8% ) |

### Brief summary of changes

There are 4 relevant muscle curves in `Millard2012EquilibriumMuscle` which are calling `calcValue` and `calcDerivative`, which is combined to `calcValueAndDerivative` by this PR:

- Added `calcValueAndDerivative` method to:
    - `ActiveForceLengthCurve.h`
    - `TendonForceLengthCurve.h`
    - `FiberForceLengthCurve.h`
    - `TendonForceLengthCurve.h`

- Regarding the `ActiveForceLengthCurve` and `TendonForceLengthCurve`:
    - moved computing curve derivative to `calcMuscleLengthInfo`, and store result in `MuscleLengthInfo::userDefinedLengthExtras`
    - moved `Millard2012EquilibriumMuscle::calcFiberStiffness` to anonymous namespace (this was using the derivative values)

- Regarding the `ForceVelocityCurve`:
    - use `calcValueAndDerivative` during `calcDampedNormFiberVelocity()` instead of `calcValue` followed by `calcDerivative`.
    - return last evaluation of `fiberForceVelocityCurve` from `calcDampedNormFiberVelocity()`

- Regarding the `TendonForceLengthCurve`:
    - Compute fiberstiffness and tendonforcelength multiplier during `calcFiberVelocityInfo()` and store in `FiberVelocityInfo::userDefinedVelocityExtras`

### Testing I've completed

Unit tests are passing, and Rajagopal forward simulation as well as CMC on running-model gives no change in results.

### CHANGELOG.md

- no need to update because too minor change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3653)

### DIFF
--- a/OpenSim/Actuators/ActiveForceLengthCurve.cpp
+++ b/OpenSim/Actuators/ActiveForceLengthCurve.cpp
@@ -158,6 +158,15 @@ double ActiveForceLengthCurve::calcValue(double normFiberLength) const
     return m_curve.calcValue(normFiberLength);
 }
 
+SmoothSegmentedFunction::ValueAndDerivative ActiveForceLengthCurve::
+    calcValueAndDerivative(double normFiberLength) const
+{
+    SimTK_ASSERT(
+        isObjectUpToDateWithProperties(),
+        "ActiveForceLengthCurve: Curve is not up-to-date with its properties");
+    return m_curve.calcValueAndFirstDerivative(normFiberLength);
+}
+
 double ActiveForceLengthCurve::calcDerivative(double normFiberLength,
                                               int order) const
 {
--- a/OpenSim/Actuators/ActiveForceLengthCurve.cpp
+++ b/OpenSim/Actuators/ActiveForceLengthCurve.cpp
@@ -153,7 +153,8 @@ void ActiveForceLengthCurve::setMinValue(double minimumValue)
 //==============================================================================
 double ActiveForceLengthCurve::calcValue(double normFiberLength) const
 {
-    SimTK_ASSERT(isObjectUpToDateWithProperties(),
+    OPENSIM_ASSERT(
+        isObjectUpToDateWithProperties() &&
         "ActiveForceLengthCurve: Curve is not up-to-date with its properties");
     return m_curve.calcValue(normFiberLength);
 }
@@ -161,8 +162,8 @@ double ActiveForceLengthCurve::calcValue(double normFiberLength) const
 SmoothSegmentedFunction::ValueAndDerivative ActiveForceLengthCurve::
     calcValueAndDerivative(double normFiberLength) const
 {
-    SimTK_ASSERT(
-        isObjectUpToDateWithProperties(),
+    OPENSIM_ASSERT(
+        isObjectUpToDateWithProperties() &&
         "ActiveForceLengthCurve: Curve is not up-to-date with its properties");
     return m_curve.calcValueAndFirstDerivative(normFiberLength);
 }
@@ -170,7 +171,8 @@ SmoothSegmentedFunction::ValueAndDerivative ActiveForceLengthCurve::
 double ActiveForceLengthCurve::calcDerivative(double normFiberLength,
                                               int order) const
 {
-    SimTK_ASSERT(isObjectUpToDateWithProperties(),
+    OPENSIM_ASSERT(
+        isObjectUpToDateWithProperties() &&
         "ActiveForceLengthCurve: Curve is not up-to-date with its properties");
     SimTK_ERRCHK1_ALWAYS(order >= 0 && order <= 2,
         "ActiveForceLengthCurve::calcDerivative",
@@ -188,7 +190,8 @@ double ActiveForceLengthCurve::
 
 SimTK::Vec2 ActiveForceLengthCurve::getCurveDomain() const
 {
-    SimTK_ASSERT(isObjectUpToDateWithProperties(),
+    OPENSIM_ASSERT(
+        isObjectUpToDateWithProperties() &&
         "ActiveForceLengthCurve: Curve is not up-to-date with its properties");
 
     return m_curve.getCurveDomain();
--- a/OpenSim/Actuators/FiberForceLengthCurve.cpp
+++ b/OpenSim/Actuators/FiberForceLengthCurve.cpp
@@ -214,7 +214,8 @@ void FiberForceLengthCurve::setOptionalProperties(double aStiffnessAtLowForce,
 //==============================================================================
 double FiberForceLengthCurve::calcValue(double normFiberLength) const
 {
-    SimTK_ASSERT(isObjectUpToDateWithProperties(),
+    OPENSIM_ASSERT(
+        isObjectUpToDateWithProperties() &&
         "FiberForceLengthCurve: Curve is not up-to-date with its properties");
     return m_curve.calcValue(normFiberLength);
 }
@@ -222,8 +223,8 @@ double FiberForceLengthCurve::calcValue(double normFiberLength) const
 SmoothSegmentedFunction::ValueAndDerivative FiberForceLengthCurve::
     calcValueAndDerivative(double normFiberLength) const
 {
-    SimTK_ASSERT(
-        isObjectUpToDateWithProperties(),
+    OPENSIM_ASSERT(
+        isObjectUpToDateWithProperties() &&
         "FiberForceLengthCurve: Curve is not up-to-date with its properties");
     return m_curve.calcValueAndFirstDerivative(normFiberLength);
 }
@@ -231,7 +232,8 @@ SmoothSegmentedFunction::ValueAndDerivative FiberForceLengthCurve::
 double FiberForceLengthCurve::calcDerivative(double normFiberLength,
                                              int order) const
 {
-    SimTK_ASSERT(isObjectUpToDateWithProperties(),
+    OPENSIM_ASSERT(
+        isObjectUpToDateWithProperties() &&
         "FiberForceLengthCurve: Curve is not up-to-date with its properties");
     SimTK_ERRCHK1_ALWAYS(order >= 0 && order <= 2,
         "FiberForceLengthCurve::calcDerivative",
@@ -249,7 +251,8 @@ double FiberForceLengthCurve::
 
 double FiberForceLengthCurve::calcIntegral(double normFiberLength) const
 {
-    SimTK_ASSERT(isObjectUpToDateWithProperties(),
+    OPENSIM_ASSERT(
+        isObjectUpToDateWithProperties() &&
         "FiberForceLengthCurve: Curve is not up-to-date with its properties");
 
     if(!m_curve.isIntegralAvailable()) {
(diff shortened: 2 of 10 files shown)


OpenSim/Actuators/ActiveForceLengthCurve.cpp
```diff
@@ -158,6 +158,15 @@ double ActiveForceLengthCurve::calcValue(double normFiberLength) const
     return m_curve.calcValue(normFiberLength);
 }
 
+SmoothSegmentedFunction::ValueAndDerivative ActiveForceLengthCurve::
```
adamkewley, line 161: Ok for now, but I would recommend a fast refactor after this PR is turned around to move `ValueAndDerivative` into a separate header. This naming is exposing an inner detail.

OpenSim/Actuators/ActiveForceLengthCurve.cpp
```diff
@@ -158,6 +158,15 @@ double ActiveForceLengthCurve::calcValue(double normFiberLength) const
...
+    calcValueAndDerivative(double normFiberLength) const
+{
+    SimTK_ASSERT(
```
adamkewley, line 164: There is now `OPENSIM_ASSERT` (see: https://github.com/opensim-org/opensim-core/pull/3548), which provides better user feedback (component names, etc.)

OpenSim/Actuators/FiberForceLengthCurve.cpp
```diff
@@ -219,6 +219,15 @@ double FiberForceLengthCurve::calcValue(double normFiberLength) const
     return m_curve.calcValue(normFiberLength);
 }
 
+SmoothSegmentedFunction::ValueAndDerivative FiberForceLengthCurve::
+    calcValueAndDerivative(double normFiberLength) const
+{
+    SimTK_ASSERT(
```
adamkewley, line 225: Same comments as above

OpenSim/Actuators/ForceVelocityCurve.cpp
```diff
@@ -176,6 +176,15 @@ double ForceVelocityCurve::calcValue(double normFiberVelocity) const
     return m_curve.calcValue(normFiberVelocity);
 }
 
+SmoothSegmentedFunction::ValueAndDerivative ForceVelocityCurve::
+    calcValueAndDerivative(double normFiberVelocity) const
+{
+    SimTK_ASSERT(
```
adamkewley, line 182: ditto

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -33,6 +33,12 @@ const string Millard2012EquilibriumMuscle::
     STATE_FIBER_LENGTH_NAME = "fiber_length";
 const double MIN_NONZERO_DAMPING_COEFFICIENT = 0.001;
 
+const static int MLIFiberForceLengthCurveDerivative  = 0;
```
adamkewley, line 36: `static constexpr int`, even

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -206,11 +215,36 @@ DampedFiberVelocityCalculationResult calcDampedNormFiberVelocity(
 
     return {
         dlceN_dt,
+        fv,
         err,
         converged,
     };
 }
 
+/* Calculates the stiffness of the fiber in the direction of the fiber.
```
adamkewley, line 224: I can't see where the original implementation was removed (it's removed in the header?)

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -911,8 +953,25 @@ calcFiberVelocityInfo(const SimTK::State& s, FiberVelocityInfo& fvi) const
...
-        double dlenMcl   = getLengtheningSpeed(s);
-        double optFibLen = getOptimalFiberLength();
+        const double dlenMcl        = getLengtheningSpeed(s);
+        const double optFibLen      = getOptimalFiberLength();
+        const double fiso           = getMaxIsometricForce();
+        const double tendonSlackLen = getTendonSlackLength();
+
+        // Compute tendon force length multiplier, and tendon stiffness. Tendon
+        // stiffness is computed here, because it is cheaper to evaluate the
+        // curve value and derivative simultaneously.
```
adamkewley, line 963: (see opensim-core/3653)

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -1080,17 +1134,14 @@ calcMuscleDynamicsInfo(const SimTK::State& s, MuscleDynamicsInfo& mdi) const
         // Get the quantities that we've already computed.
         const MuscleLengthInfo &mli = getMuscleLengthInfo(s);
         const FiberVelocityInfo &mvi = getFiberVelocityInfo(s);
-        double fiberStateClamped = mvi.userDefinedVelocityExtras[0];
+        double fiberStateClamped =
+            mvi.userDefinedVelocityExtras[MVIIsFiberStateClamped];
```
adamkewley, line 1138: `at`, if that's available - it's hard to keep track of the state of the user variables (too big? too small?)

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
```diff
@@ -1149,38 +1200,37 @@ calcMuscleDynamicsInfo(const SimTK::State& s, MuscleDynamicsInfo& mdi) const
...
 
             fmAT = fm * mli.cosPennationAngle;
-            dFm_dlce = calcFiberStiffness(fiso, a,
-                                          mvi.fiberForceVelocityMultiplier,
-                                          mli.normFiberLength, optFiberLen);
+            dFm_dlce = calcFiberStiffness(
+                fiso,
+                a,
+                mvi.fiberForceVelocityMultiplier,
+                mli.userDefinedLengthExtras[MLIFiberForceLengthCurveDerivative],
```
adamkewley, line 1207: `at`, etc. - unless it's very very obvious that there's no way these could be out-of-bounds

OpenSim/Actuators/TendonForceLengthCurve.cpp
```diff
@@ -223,6 +223,15 @@ double TendonForceLengthCurve::calcValue(double aNormLength) const
     return m_curve.calcValue(aNormLength);
 }
 
+SmoothSegmentedFunction::ValueAndDerivative TendonForceLengthCurve::
+    calcValueAndDerivative(double aNormLength) const
+{
+    SimTK_ASSERT(
```
adamkewley, line 229: `OPENSIM_ASSERT`, if it makes sense

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
pepbos: Original implementation was removed in `Millard2012EquilibriumMuscle.cpp` lines 1326 and in header on line 589

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
pepbos: It looks like it is not available. I could switch to standard lib?

OpenSim/Actuators/ActiveForceLengthCurve.cpp
```diff
@@ -153,15 +153,26 @@ void ActiveForceLengthCurve::setMinValue(double minimumValue)
 //==============================================================================
 double ActiveForceLengthCurve::calcValue(double normFiberLength) const
 {
-    SimTK_ASSERT(isObjectUpToDateWithProperties(),
+    OPENSIM_ASSERT(
+        isObjectUpToDateWithProperties() &&
```
tkuchida, line 157: Here and elsewhere: I don't understand why this assertion is written as a logical 'and' with `isObjectUpToDateWithProperties()` and a string. Simpler would be easier to read. Just delete the string since it always evaluates to false?

tkuchida (commented):
Only additional suggestion is to simplify the `ASSERT`s. 🦓

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
adamkewley: Yes, which means that people can look up your explanation/measurements, and these comments, when reading the source in a few years time

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
adamkewley: I'd avoid changing datastructures, but if you still want bounds-checking then an internal `At` helper function is usually how I handle it for OpenSim/SimTK types: 

https://github.com/ComputationalBiomechanicsLab/opensim-creator/blob/0.5.5/src/OpenSimCreator/Utils/OpenSimHelpers.hpp#L104

OpenSim/Actuators/ActiveForceLengthCurve.cpp
pepbos: This was new to me, but @adamkewley pointed out to me that a common paradigm in c++ is to have `assert(condition && "this condition failed")`, which indeed evaluates the string as false only if the condition is false.

It is also listed as a common way to display error messages: https://en.cppreference.com/w/cpp/error/assert

OpenSim/Actuators/Millard2012EquilibriumMuscle.cpp
pepbos: I suggest to do this in a separate PR as well.

pepbos:
@adamkewley @tkuchida , Thanks for your feedback, sorry for the break in between, please have a look if you agree with the changes :)

Happy new year!

OpenSim/Actuators/ActiveForceLengthCurve.cpp
tkuchida: Ah, ok thanks. Is this now the preferred style (supplanting `SimTK_ASSERT` and `OPENSIM_ASSERT_FRMOBJ`)?

pepbos: Yeah, afaik: https://github.com/opensim-org/opensim-core/pull/3653#discussion_r1426588089

tkuchida (approved):

PR #3658, opened 2023-12-19
Replace usages of `sprintf` with `snprintf`

nickbianco:
### Brief summary of changes

As the title suggests, `snprintf` replaces all usages of `sprintf`, which can be dangerous because it can potentially output more characters than can fit in the allocation size of the buffer string. The change suppresses a `gcc` compiler warning (which was the main motivation for the PR).

### CHANGELOG.md (choose one)

- no need to update because...not user facing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3658)

### DIFF
--- a/OpenSim/Analyses/PointKinematics.cpp
+++ b/OpenSim/Analyses/PointKinematics.cpp
@@ -223,12 +223,12 @@ constructDescription()
     strcat(descrip,"(position, velocity, or acceleration) of\n");
     
     if(_relativeToBody){
-        sprintf(tmp,"point (%lf, %lf, %lf) on body %s relative to body %s of model %s.\n",
+        snprintf(tmp, BUFFER_LENGTH, "point (%lf, %lf, %lf) on body %s relative to body %s of model %s.\n",
             _point[0],_point[1],_point[2],_body->getName().c_str(),
             _relativeToBody->getName().c_str(), _model->getName().c_str());
     }
     else{
-        sprintf(tmp,"point (%lf, %lf, %lf) on the %s of model %s.\n",
+        snprintf(tmp, BUFFER_LENGTH, "point (%lf, %lf, %lf) on the %s of model %s.\n",
             _point[0],_point[1],_point[2],_body->getName().c_str(),
             _model->getName().c_str());
     }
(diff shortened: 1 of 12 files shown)


aymanhab (changes requested):
One minor comment, otherwise LGTM

> Reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3658#-:-Nm3KnDQBNSH6sKgeKvU:bvmtosp)* status: all files reviewed, 1 unresolved discussion (waiting on @nickbianco)

___
*[`OpenSim/Common/IO.cpp` line 286 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3658#-Nm3JUWa8mCOAemKBMlB:-Nm3JUWa8mCOAemKBMlC:bqxs0t4) ([raw file](https://github.com/opensim-org/opensim-core/blob/f68858a638a7d850f6d18e557b95454db4a503a4/OpenSim/Common/IO.cpp#L286)):*
> ```cpp
>     if(_GFormatForDoubleOutput) {
>         snprintf(_DoubleFormat, 256, "%%g");
>     } else if(_Scientific) {
> ```

Considering this number comes from a different file/header, it may make sense to define this as const integer BUFFER_LENGTH and refer to it by name to ensure they don't get out of sync.

nickbianco (commented):
Ready for review.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3658#-:-Nm3OWmi592TLKBPDa1A:b8pt725)* status: 10 of 12 files reviewed, 1 unresolved discussion (waiting on @aymanhab)

___
*[`OpenSim/Common/IO.cpp` line 286 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3658#-Nm3JUWa8mCOAemKBMlB:-Nm3OQLu5oIV6dSoGTUf:brtu9fd) ([raw file](https://github.com/opensim-org/opensim-core/blob/f68858a638a7d850f6d18e557b95454db4a503a4/OpenSim/Common/IO.cpp#L286)):*

Considering this number comes from a different file/header, it may make sense to define this as const integer BUFFER_LENGTH and refer to it by name to ensure they don't get out of sync.
</blockquote></details>

Good call, done. I added the constant and made the existing constant also `constexpr`.

aymanhab (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3658#-:-Nm3XCXo1fzphvwtlTrp:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nickbianco)

PR #3657, opened 2023-12-19
Add rule-of-five to `IKMarkerTask` and `OrientationWeightSet`

nickbianco:
### Brief summary of changes

Adds move constructors, copy assignment operators, and move assignment operators to `IKMarkerTask` and `OrientationWeightSet`.  This suppresses `gcc` compiler warnings complaining about no user-defined assignment operator when a user-defined copy-assignment operator is present.

`ProbeMeasure` issues a similar warning, but I think it can be ignored. It uses [a Simbody macro](https://github.com/simbody/simbody/blob/589e931f4a127830a5876ee4dd8f327b47361504/SimTKcommon/Simulation/include/SimTKcommon/internal/Measure.h#L64) to define custom copy assignment operators and copy constructors. The macro uses the terms `shallowAssign` and `deepAssign` to define its copy constructors, since these do not use the typical constructor syntax, `gcc` complains.

### CHANGELOG.md (choose one)

- no need to update because...not user facing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3657)

PR #3659, opened 2023-12-20
Suppress remaining `gcc` compiler warnings

nickbianco:
### Brief summary of changes

This PR is mainly intended to address any remaining `gcc` compiler warnings not suppressed in the previous two PRs (#3657 and #3658). The changes include:

- Adding `sprintf` to `snprintf` conversions that I missed in #3658.
- Fixed unused variable warnings. For variables that had no apparant use at all, I removed them entirely. For others in test cases, I replaced commented out `cout` statement with active logging statements.
- Fixed a boolean operation in `CoordinateCouplerConstraint` that using `&` instead of `&&`.
- Converted a `size_t` iterator to `int` in `SmoothSegmentedFunction` (to avoid an implicit conversion in the loop).

### CHANGELOG.md (choose one)

- no need to update because...not user facing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3659)

### DIFF
--- a/OpenSim/Simulation/Model/GeometryPath.cpp
+++ b/OpenSim/Simulation/Model/GeometryPath.cpp
@@ -292,10 +292,10 @@ void GeometryPath::constructProperties()
  */
 void GeometryPath::namePathPoints(int aStartingIndex)
 {
-    char indx[5];
     for (int i = aStartingIndex; i < get_PathPointSet().getSize(); i++)
     {
-        sprintf(indx,"%d",i+1);
+        char indx[5];
+        snprintf(indx,5,"%d",i+1);
         AbstractPathPoint& point = get_PathPointSet().get(i);
         if (point.getName()=="" && hasOwner()) {
             point.setName(getOwner().getName() + "-P" + indx);
--- a/OpenSim/Simulation/Model/GeometryPath.cpp
+++ b/OpenSim/Simulation/Model/GeometryPath.cpp
@@ -294,8 +294,8 @@ void GeometryPath::namePathPoints(int aStartingIndex)
 {
     for (int i = aStartingIndex; i < get_PathPointSet().getSize(); i++)
     {
-        char indx[5];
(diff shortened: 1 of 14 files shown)


OpenSim/Simulation/Model/GeometryPath.cpp
```diff
@@ -292,10 +292,10 @@ void GeometryPath::constructProperties()
  */
 void GeometryPath::namePathPoints(int aStartingIndex)
 {
-    char indx[5];
     for (int i = aStartingIndex; i < get_PathPointSet().getSize(); i++)
     {
-        sprintf(indx,"%d",i+1);
+        char indx[5];
```
adamkewley, line 297: I understand why you'd think 4 chars (+ nul) is enough here, because no model is going to contain >9999 path points but, for sanity's sake, I'd just use a 32/64/128 buffer here such that the full range of digits can be captured.

adamkewley:
Overall, looks fine and you can merge it etc. but, just in case you end up refactoring stuff like this later, a safer pattern for snprintf with stack-allocated buffers is:

```c++
std::array<char, 128> buf;
std::snprintf(buf.data(), buf.size(), "blah %i", i);
```

Because there's one unambiguous location for the size of the buffer

(not dissing this PR: the existence of these `snprintf`s are clearly legacy)

aymanhab (approved):
I agree with @adamkewley  sentiment but perfectly fine with merging as is since it's a step in the right direction. Your call @nickbianco 

> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3659#-:-NnL35RC9yDWViwndsnW:b-ozwdf8)* status: all files reviewed, 1 unresolved discussion (waiting on @nickbianco)

nickbianco:
I've addressed @adamkewley's specific comment. Let's switch to the safer buffer pattern in a future PR.

Issue #3578, opened 2023-10-13
Getting opensim-core-linux-build-script.sh and installation running on linux (Ubuntu 20.04) ...


mjhmilla:
Dear @AlbertoCasasOrtiz  and the others who have made this wonderful install script,

I just thought I would share the few extra steps I needed to do to get the installation working on Ubuntu 20.04. I suspect these problems might be specific to Ubuntu 20.04 specifically and so my hopes are that this post might help others. @nickbianco suggested that I put this information here. Here goes:

1. Make complained at some point that my machine is missing the yacc command. After installing yacc (by installing bison and byacc) using

   sudo apt-get install bison -y
   sudo apt-get install byacc -y

the script completed without error. I found this stackoverflow entry helpful here:

https://stackoverflow.com/questions/10733238/make-yacc-command-not-found-after-installing-bison

2. When trying to run OpenSim from Python, the interpreter complained about links to libraries that couldn't be found: opensim-core/sdk/Python/opensim/_simbody.so and _moco.so. I used ldd on the libraries that complained and then added LD_LIBRARY_PATH statements until all of the libraries could be located. Here's specifically what I had to add:

   export
LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/home/mmillard/opensim-workspace/opensim-core-dependencies-install/simbody/lib
   export
LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/home/mmillard/opensim-workspace/opensim-core-dependencies-install/ipopt/lib
   export
LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/home/mmillard/opensim-workspace/opensim-core-dependencies-install/adol-c/lib64

Thanks again for making such an excellent install script!

@mjhmilla

nickbianco:
Thanks @mjhmilla! We already detect the OS in the build script, so we could perhaps add these changes in only when Ubuntu is detected.

AlbertoCasasOrtiz:
Thank you @mjhmilla for reporting this.

I tested the script on Ubuntu 20.04 and I got the same error. Your solution works perfectly. I will add it to the script and will test everything thoroughly.

PR #3663, opened 2024-01-08
Install bison and byacc if available.

AlbertoCasasOrtiz:
Fixes issue #3578 

### Brief summary of changes

The prolem was that in some distributions, byacc and bison are required, but not installed by default, but in others, they are not even available in the package manager for install. In this PR I check if bison and byacc area available in apt-get, and install if so.

### Testing I've completed

- [x] Ubuntu 18.04
- [x] Ubuntu 20.04
- [x] Ubuntu 22.04
- [x] Debian 10
- [x] Debian 11

### CHANGELOG.md (choose one)

- no need to update because build script.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3663)

### DIFF
--- a/scripts/build/opensim-core-linux-build-script.sh
+++ b/scripts/build/opensim-core-linux-build-script.sh
@@ -80,8 +80,26 @@ echo
 
 # Install dependencies from package manager.
 echo "LOG: INSTALLING DEPENDENCIES..."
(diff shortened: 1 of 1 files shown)


AlbertoCasasOrtiz:
@aymanhab This should be ready to merge, fixes the first half of #3578 related to the build scripts.

aymanhab:
Thanks @AlbertoCasasOrtiz

PR #3665, opened 2024-01-09
Removes musclePower field from MuscleDynamicsInfo

pepbos:
This PR removes the `musclePower` field from the cached `MuscleDynamicsInfo`.

Since the cache's main purpose is to store expensive-to-compute results, and the muscle power is very cheap to compute, it can be removed from the cache.

### Brief summary of changes

- Removed `musclePower` field from `MuscleDynamicsInfo` in `Muscle.h`.
- Added `calcMusclePower` method to `Muscle` in `Muscle.h`.
- Removed power computation in all children of `Muscle`.

### Testing I've completed

Unit tests are passing, and contain regression tests for several muscles.

### CHANGELOG.md

- no need to update because too small a change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3665)


OpenSim/Simulation/Model/Muscle.h
```diff
@@ -458,6 +458,8 @@ OpenSim_DECLARE_ABSTRACT_OBJECT(Muscle, PathActuator);
     /** Override of the default implementation to account for versioning. */
     void updateFromXMLNode(SimTK::Xml::Element& aNode, int versionNumber=-1) override;
 
+    /** The muscle power (W) */
```
adamkewley, line 461: Verb

OpenSim/Actuators/RigidTendonMuscle.cpp
```diff
@@ -193,8 +193,6 @@ calcMuscleDynamicsInfo(const State& s, MuscleDynamicsInfo& mdi) const
     mdi.fiberActivePower = -(mdi.activeFiberForce) * fvi.fiberVelocity;
     mdi.fiberPassivePower = -(mdi.passiveFiberForce) * fvi.fiberVelocity;
     mdi.tendonPower = 0;
-    mdi.musclePower = -getMaxIsometricForce()*mdi.normTendonForce
```
adamkewley, line 196: Isn't this calculation different from the generic `Muscle` one?

pepbos: Overlooked that one... Looks like it was originally incorrect. @tkuchida what do you think?

pepbos: As a note: This pr doesnt change the computation. The fiberVelocity for the RigidTendonMuscle is equal to lengtheningSpeed. See https://github.com/opensim-org/opensim-core/blob/20752f3a7852854a3e95a0bf52832bc8ec788163/OpenSim/Actuators/RigidTendonMuscle.cpp#L164 
So it matches the computation in `Muscle`.
As to why fibervelocity equals muscle velocity, I am unsure :)

tkuchida: Looks like a bug 🐞 to me, but perhaps I'm misunderstanding something. If the tendon is rigid, won't the lengthening speed of the path equal the velocity of the fiber only if pennation is zero? Unfortunately, RigidTendonMuscle is only tested when pennation is zero (see [`testRigidTendonMuscle()` in testMuscles.cpp](https://github.com/opensim-org/opensim-core/blob/20752f3a7852854a3e95a0bf52832bc8ec788163/OpenSim/Actuators/Test/testMuscles.cpp#L511)). In general,

$v^{\mathrm{M}}(t) = \frac{1}{\cos(\phi(t))} \left[ v^{\mathrm{MT}}(t) + \ell^{\mathrm{M}}(t) \sin(\phi(t)) \dot{\phi}(t) \right]$

which simplifies to $v^{\mathrm{M}}(t) = v^{\mathrm{MT}}(t)$ when $\phi(t) = 0$.

pepbos:
@adamkewley , @tkuchida thanks for your comments! I think it is good to merge, but have a look if you have the time

OpenSim/Actuators/RigidTendonMuscle.cpp
tkuchida: It should not be assumed that pennation is zero.

tkuchida (approved):

PR #3667, opened 2024-01-10
Make std::regex parsing have a static lifetime

adamkewley:
Fixes issue N/A

Minor nit I spotted when I am running applications that load many osim files (e.g. OSC's test suite) - the profiler notices that it's spending a large amount of time (5 % or so) compiling regex strings on the lines that are changed by this diff. This speedup only really affects initially loading older osim files (e.g. many of the example files in OpenSim/OSC).

### Brief summary of changes

- Make constructing a `std::regex` from a pattern string have a static function-scoped lifetime so it only happens once per application boot.

### CHANGELOG.md (choose one)

- no need to update because minor

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3667)

### DIFF
--- a/OpenSim/Common/Component.cpp
+++ b/OpenSim/Common/Component.cpp
@@ -1174,20 +1174,23 @@ void Component::updateFromXMLNode(SimTK::Xml::Element& node, int versionNumber)
 
         }
         if (versionNumber <= 30516) {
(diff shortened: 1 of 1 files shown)


nickbianco (approved):
LGTM.

PR #3668, opened 2024-01-10
Fix PathSpring overwriting any user-set appearance for its GeometryPath

adamkewley:
Fixes issue: https://github.com/ComputationalBiomechanicsLab/opensim-creator/issues/818

Minor nit noticed by a user who's making a model containing path springs: if you change the `Appearance` of the path spring's `GeometryPath` it has no effect.

The reason why this is a bug is because `finalizeFromProperties` is called _after_ properties are updated (e.g. from loading an XML file or editing in OpenSim GUI or OpenSim Creator) - the existing implementation will then unconditionally overwrite the color of the pathspring with green.

The fix instead sets the default in the constructor, which is usually called _before_ properties are updated (e.g. when the object is loaded from an XML file it effectively sets properties on an already-constructed prototype).

### Testing I've completed

I hotfixed it into an OSC build and it now correctly shows a green spring _unless_ the user specifies a different color in the XML file.

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3668)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ and `Blankevoort1991Ligament`, usages of `get_GeometryPath`, `upd_GeometryPath`,
 - Simbody was updated such that the headers it transitively exposes to downstream projects are compatible with C++20 (#3619)
 - Moved speed computation from `computeForce` in children of `ScalarActuator` to dedicated `getSpeed` function.
 - Fix type problem with BufferedOrientationReference (Issue #3415, PR #3644)
+- Fixed setting the color of a PathSpring's GeometryPath should now update the color of the PathSpring geometry
 
 
 v4.4.1
--- a/OpenSim/Simulation/Model/PathSpring.cpp
+++ b/OpenSim/Simulation/Model/PathSpring.cpp
@@ -33,8 +33,6 @@ using namespace std;
 using namespace OpenSim;
 using SimTK::Vec3;
 
-static const Vec3 DefaultPathSpringColor(0, 1, 0); // Green for backward compatibility
-
 //=============================================================================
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and `Blankevoort1991Ligament`, usages of `get_GeometryPath`, `upd_GeometryPath`,
 - Simbody was updated such that the headers it transitively exposes to downstream projects are compatible with C++20 (#3619)
 - Moved speed computation from `computeForce` in children of `ScalarActuator` to dedicated `getSpeed` function.
 - Fix type problem with BufferedOrientationReference (Issue #3415, PR #3644)
+- Fixed setting the color of a PathSpring's GeometryPath should now update the color of the PathSpring geometry
```
nickbianco, line 45: Minor, but this fix will end up in the next release (4.6) since we've recently finalized 4.5.

nickbianco: If you do want this in 4.5, we'd have to merge it into branch `opensim_45` soon. Let @aymanhab know if that is the case.

nickbianco (commented):
Overall LGTM, just one comment about the changelog.

CHANGELOG.md
adamkewley: Whatever works: OSC is using main branch releases of OS these days, so that I can integrate/test API changes as they happen, rather than in spikes, so it isn't as important that  fixes are rolled into point releases for me

nickbianco: Sounds good! Just wanted to make sure you were aware.

nickbianco (approved):

PR #3674, opened 2024-01-12
Add more string-like operations to ComponentPath

adamkewley:
The motivation for this change is to make downstream code more in-line with how it would be written when using a `std::string`, and to enable patterns like `std::unordered_map<ComponentPath, OpenSim::Joint const*> jointLookup` etc.

### Brief summary of changes

- Added `ComponentPath::root()`, which returns a `ComponentPath` equivalent to "/"
- `ComponentPath` is now less-than (`<`) comparable, making it usable in (e.g.) `std::map`
- `ComponentPath` now has a `std::hash<T>` implementation, making it usable in (e.g.) `std::unordered_map`
- Added `.clear()` and `.empty()` to `ComponentPath` for more parity with `std::string`'s semantics

### Testing I've completed

- Unit tests added for each new behavior

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3674)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ request related to the change, then we may provide the commit.
 
 This is not a comprehensive list of changes but rather a hand-curated collection of the more notable ones. For a comprehensive history, see the [OpenSim Core GitHub repo](https://github.com/opensim-org/opensim-core).
 
+Upcoming Release
+================
+
+- Added `ComponentPath::root()`, which returns a `ComponentPath` equivalent to "/"
+- `ComponentPath` is now less-than (`<`) comparable, making it usable in (e.g.) `std::map`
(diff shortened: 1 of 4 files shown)


nickbianco (approved):
LGTM. Very nice! It would be great to convert all `std::string` usages in OpenSim when refering to components to use `ComponentPath` at some point.

> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3674#-:-NoJEu1MBSMHgJROy04F:bj1jd52)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @adamkewley)

PR #3678, opened 2024-01-12
Refactor `OpenSim::Array` to internally use `std::vector`

adamkewley:
Fixes a deep dark hole in my heart - and makes adding iterators, range for-loop support, etc. etc. much easier, because I'm getting bored of seeing `for (int...)` in modern C++.

**WIP**:

- ~~The binary search algorithm is slightly different from the original specification~~ The implementation should now match the wording of the initial implementation
- Need to run the test suite /w libASAN to ensure that there isn't any funny business now that the memory allocation pattern has changed

### Brief summary of changes

- Replaced the custom implementation with `std::vector` and added various shims etc. to ensure it's API-compatible with the legacy API
- The only exception to backwards compatibility is the growth increment, which isn't customizable in `std::vector`, so was deprecated

### Testing I've completed

- Because it's so widely used, it affects almost every part of OpenSim - I am *guessing* that the test suite would kick out the worst offenders
- ~~... but (WIP) i need to triple-check the binary search code, because it was very tricksy~~

### CHANGELOG.md (choose one)

- no need to update because users need not know the incredibly tedious process of legacy maintenance

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3678)

### DIFF
--- a/OpenSim/Analyses/Actuation.cpp
+++ b/OpenSim/Analyses/Actuation.cpp
@@ -327,27 +327,6 @@ getPowerStorage() const
     return(_powerStore);
 }
 
-
-//-----------------------------------------------------------------------------
-// STORAGE CAPACITY
-//-----------------------------------------------------------------------------
-//_____________________________________________________________________________
-/**
-* Set the capacity increments of all storage instances.
-*
-* @param aIncrement Increment by which storage capacities will be increased
-* when storage capacities run out.
-*/
-void Actuation::
-setStorageCapacityIncrements(int aIncrement)
-{
-    _forceStore->setCapacityIncrement(aIncrement);
-    _speedStore->setCapacityIncrement(aIncrement);
-    _powerStore->setCapacityIncrement(aIncrement);
-}
-
-
-
 //=============================================================================
 // ANALYSIS
 //=============================================================================
--- a/OpenSim/Analyses/Actuation.h
+++ b/OpenSim/Analyses/Actuation.h
@@ -98,7 +98,8 @@ class Storage;
         // GET AND SET
         //--------------------------------------------------------------------------
         // STORAGE
-        void setStorageCapacityIncrements(int aIncrement);
+        [[deprecated("this method no longer does anything")]]
+        void setStorageCapacityIncrements(int) {}
         Storage* getForceStorage() const;
         Storage* getSpeedStorage() const;
         Storage* getPowerStorage() const;
--- a/OpenSim/Analyses/BodyKinematics.cpp
+++ b/OpenSim/Analyses/BodyKinematics.cpp
@@ -385,24 +385,6 @@ getPositionStorage()
 }
 
 //-----------------------------------------------------------------------------
-// STORAGE CAPACITY
-//-----------------------------------------------------------------------------
-//_____________________________________________________________________________
-/**
- * Set the capacity increments of all storage instances.
- *
- * @param aIncrement Increment by which storage capacities will be increased
- * when storage capacities run out.
- */
-void BodyKinematics::
-setStorageCapacityIncrements(int aIncrement)
-{
-    _aStore->setCapacityIncrement(aIncrement);
-    _vStore->setCapacityIncrement(aIncrement);
-    _pStore->setCapacityIncrement(aIncrement);
-}
-
-//-----------------------------------------------------------------------------
 // ANGULAR VELOCITY IN LOCAL FRAME FLAG
 //-----------------------------------------------------------------------------
 //_____________________________________________________________________________
--- a/OpenSim/Analyses/BodyKinematics.h
+++ b/OpenSim/Analyses/BodyKinematics.h
@@ -101,7 +101,8 @@ public:
     // GET AND SET
     //--------------------------------------------------------------------------
     // STORAGE
-    void setStorageCapacityIncrements(int aIncrement);
+    [[deprecated("this method no longer does anything")]]
+    void setStorageCapacityIncrements(int) {}
     Storage* getAccelerationStorage();
     Storage* getVelocityStorage();
     Storage* getPositionStorage();
--- a/OpenSim/Analyses/InverseDynamics.cpp
+++ b/OpenSim/Analyses/InverseDynamics.cpp
@@ -247,22 +247,6 @@ getStorage()
     return(_storage);
 }
 
-//-----------------------------------------------------------------------------
-// STORAGE CAPACITY
-//-----------------------------------------------------------------------------
-//_____________________________________________________________________________
-/**
- * Set the capacity increments of all storage instances.
- *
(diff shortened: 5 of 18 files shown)


nickbianco (commented):
Nice! Looks very reasonable to me.

I noticed that `searchBinary` is used in various parts of the codebase. Did you test that change locally or confirm that we already test coverage for it? My only comment is to consider a short unit test if it seems useful (perhaps for the`Manager` class, which depends on it for [retrieving integrator time steps](https://github.com/opensim-org/opensim-core/blob/98c70b1fd5bc74da63e69a52e860c742840a9a16/OpenSim/Simulation/Manager/Manager.cpp#L382)).

> Reviewed 17 of 17 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3678#-:-Np1Nz4v09GpF8ZrBfFr:bfal5l2)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @adamkewley)

adamkewley:
> Nice! Looks very reasonable to me.
> 
> I noticed that `searchBinary` is used in various parts of the codebase. Did you test that change locally or confirm that we already test coverage for it? My only comment is to consider a short unit test if it seems useful (perhaps for the`Manager` class, which depends on it for [retrieving integrator time steps](https://github.com/opensim-org/opensim-core/blob/98c70b1fd5bc74da63e69a52e860c742840a9a16/OpenSim/Simulation/Manager/Manager.cpp#L382)).
> 
> > Reviewed 17 of 17 files at r1, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3678#-:-Np1Nz4v09GpF8ZrBfFr:bfal5l2)_ status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adamkewley)

I haven't tested it yet, but I think that one way to do it would be to copy+paste the old implementation into a unit test and confirm it gives similar results.

From what I can see of it, its main usage purpose is to find an insertion point for data, but (imo) `std::lower_bound` alone would be better for this (it's used elsewhere): the behavior (as documented) seems likely to insert data one-too-early in a sequence (I'll have to check)

nickbianco:
> I haven't tested it yet, but I think that one way to do it would be to copy+paste the old implementation into a unit test and confirm it gives similar results.

Even if you confirmed this locally it would likely be fine, but a unit test would be safer. I'm fine with replacing usages of `searchBinary` with `std::lower_bound` if you want to add that here too.

adamkewley:
I added a legacy-exercising unit test to ensure that the results will be the same. It found a bug in the implementation!

Assuming this passes CI, it should be ready to merge

aymanhab:
@adamkewley Looks great, I'd only ask that you run swig and diff/compare the swig output since the Array class is used in numerous places across the API. It would be great to verify that the GUI builds with this branch. Thank you 🥇

aymanhab (changes requested):
Main issue as client of this class (and I understand the effort that went into making sure the API stays the same) is to make sure that the SWIG can digest it properly, some typos and corner case questions, otherwise LGTM

> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3678#-:-NqDkL317AMapmwXiPNW:b-eo958i)* status: 17 of 18 files reviewed, 2 unresolved discussions (waiting on @adamkewley and @nickbianco)

___
*[`OpenSim/Common/Array.h` line 203 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3678#-NqDjkwF566HWzsqttLp:-NqDjkwF566HWzsqttLq:b6ecr4r) ([raw file](https://github.com/opensim-org/opensim-core/blob/852ba2f4f41e099a2a6111dc6f848f264115d9e4/OpenSim/Common/Array.h#L203)):*
> ```objc
>         }
>         _storage.resize(aSize, _defaultValue);
>         return true;
> ```

Not sure if this is ever used but is aSize == _size still noop?
___
*[`OpenSim/Common/Test/testArray.cpp` line 2 at r2](https://reviewable.io/reviews/opensim-org/opensim-core/3678#-NqDkESx7iXdvfzgDsHw:-NqDkESyESXdW3-R05mO:b-7ilvfu) ([raw file](https://github.com/opensim-org/opensim-core/blob/852ba2f4f41e099a2a6111dc6f848f264115d9e4/OpenSim/Common/Test/testArray.cpp#L2)):*
> ```cpp
> *                       OpenSim:  testAPDMDataReader.cpp                    *
> ```

typo?

nickbianco (approved):
:lgtm: (aside from @aymanhab's comments)

> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3678#-:-NqKT9L104-IIMihrf0I:brunfhv)* status: all files reviewed, 2 unresolved discussions (waiting on @adamkewley)

adamkewley:
This is held up by the SWIG check: building /w swig on Linux/WSL2 results in a missing preliminaries file and it's not entirely clear why the build depends on a file that doesn't exist

adamkewley:
Think I found the bug: SWIG generation will fail if `cmake` is ran while the current working directory is `opensim-core`, because this `execute_process` call to generate SWIG's dependencies:

- https://github.com/opensim-org/opensim-core/blob/v4.3.1/cmake/OpenSimMacros.cmake#L644

Will short-circuit some of the dependencies to (e.g.) `Bindings/preliminaties.i` because SWIG prioritizes the current dir as the highest-prio search path. However, all paths from that step must be absolute because the build can run from a different directory (it usually does).

adamkewley:
Java SWIG diff, before and after:

- https://gist.github.com/adamkewley/e4f4fdb851ce362fa1ad9f033a47226b

adamkewley:
All changes in the java diff appear to be due to renamed variables and/or updated comment strings, so I'm guessing the SWIG API is going to be fine (unless someone's code depended on variable names O_o).

aymanhab (approved):

aymanhab (approved):
Thanks @adamkewley  LGTM. I'll approve but please change the name of the test file in the header block of testArray before merging.
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3678#-:-NqTmOu08i3bgC2k1Tg_:b-hwvmik)* status: all files reviewed, 2 unresolved discussions (waiting on @adamkewley)

adamkewley:
> Not sure if this is ever used but is aSize == _size still noop?

@ayman to reply to this question, it's still a noop. `resize` only "fills in" with `_defaultValue` on vector growth, and will only ever reallocate if the new size is greater than the current capacity of the vector.

PR #3685, opened 2024-01-22
Move muscle stiffness calculation to Muscle.h

pepbos:
This PR moves the muscle stiffness computation of all derived muscle classes to the base class: Muscle.

This is a cleanup of the MuscleDynamicsInfo, which contains fields that are cheap to compute, but are still cached. It also serves to unify the computation of the outputs across all muscles.

#### Note on stiffness calculations
The muscle stiffness calculation differed slightly between the different muscles.
In `Millard2012EquilibriumMuscle` the calculation included the following checks:
    - computed stiffness is nonnegative
    - prevent division by zero
    - check if tendon is rigid
For the other muscles some of these checks were missing, but it seems that they should have been added for all muscles. Therefore I used the implementation from `Millard2012EquilibriumMuscle` for all Muscles.

For the `RigidTendonMuscle` and the muscle in `testMuscleMetabolicsProbes.cpp` the stiffness was always zero. This PR does not change that result for these muscles.

### Brief summary of changes

- Added `calcMuscleStiffness` to `Muscle.h` using implementation from `Millard2012EquilibriumMuscle`
- Removed `muscleStiffness` field from `MuscleDynamicsInfo`

### Testing I've completed

Unit tests passing.

@nickbianco the method `calcMuscleStiffness(...)` in `DeGrooteFregley2016Muscle` was used for the computation for `MuscleDynamicsInfo.muscleStiffness`, but is no longer needed. It is however a public function, and it is used in the unit test.
What would you recommend doing here? Leave as is? Remove from class, and move to unit test? Flag as deprecated?

 https://github.com/opensim-org/opensim-core/blob/98c70b1fd5bc74da63e69a52e860c742840a9a16/OpenSim/Actuators/DeGrooteFregly2016Muscle.h#L605 

### CHANGELOG.md (choose one)

- no need to update because only moves computation implementation around

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3685)


OpenSim/Simulation/Model/Muscle.cpp
```diff
@@ -633,6 +633,26 @@ double Muscle::calcMusclePower(const SimTK::State& s) const
...
+    const double dFmAT_dlceAT = mdi.fiberStiffnessAlongTendon;
+    const double dFt_dtl      = mdi.tendonStiffness;
+
+    double Ke = 0.;
+    if (!get_ignore_tendon_compliance()) {
+        // Compute the stiffness of the whole musculotendon actuator.
+        if (abs(dFmAT_dlceAT * dFt_dtl) > 0.0 &&
+            abs(dFmAT_dlceAT + dFt_dtl) > SimTK::SignificantReal) {
+            Ke = (dFmAT_dlceAT * dFt_dtl) / (dFmAT_dlceAT + dFt_dtl);
+        }
```
nickbianco, line 649: This if-statement makes muscle stiffness calculations in `DeGrooteFregly2016Muscle` non-smooth. `DeGrooteFregly2016Muscle` is designed to be smooth and continuous everywhere, primarily for use in Moco. While muscle stiffness optimization isn't very common by Moco users, this changes the design intent for `DeGrooteFregly2016Muscle` and could lead to bad convergence if someone were to optimize muscle stiffness.

nickbianco:
One comment about how this change would affect Moco optimizations using `DeGrooteFregly2016Muscle`.  

One alternative would be to make `calcMuscleStiffness(const SimTK::State&)` a virtual function and have different implementations for `DeGrooteFregly2016Muscle` and the other muscle classes. `DeGrooteFregly2016Muscle` could then leverage the existing `calcMuscleStiffness(const SimTK::Real&, const SimTK::Real&)` to implement its version of `calcMuscleStiffness(const SimTK::State&)`.

nickbianco (changes requested):
One comment about how this change would affect Moco optimizations using `DeGrooteFregly2016Muscle`.  

One alternative would be to make `calcMuscleStiffness(const SimTK::State&)` a virtual function and have different implementations for `DeGrooteFregly2016Muscle` and the other muscle classes. `DeGrooteFregly2016Muscle` could then leverage the existing `calcMuscleStiffness(const SimTK::Real&, const SimTK::Real&)` to implement its version of `calcMuscleStiffness(const SimTK::State&)`.

OpenSim/Simulation/Model/Muscle.cpp
pepbos: Ah, thanks for clarifying.
I thought it was meant to catch zero-divided-by-zero or cases like that.
But overloading sounds good.

nickbianco (approved):
> Reviewed 6 of 7 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3685)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pepbos)

nickbianco (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3685#-:-NpLeL_T7y6MwtCV3kZU:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pepbos)

PR #3686, opened 2024-01-22
Adding automated performance testing to CI workflow

nickbianco:
### Summary of changes

This change adds automated performance testing for pull requests to our GitHub Actions CI workflow. The performance testing compares the target branch (usually `main`) to the PR source branch and posts a markdown table containing a summary of the performance testing results to the PR description body (see below for example). The performance testing can be triggered by including the string '[perf]' anywhere in the PR title or description body  (**not** the commit message, the testing will be triggered for all commits pushed to the PR branch).

The automated testing relies on the new private repository with in [opensim-org/opensim-perf](https://github.com/opensim-org/opensim-perf), which is essentially a fork of the excellent [adamkewley/osimperf](https://github.com/adamkewley/osimperf). This new private repo allows to use self-hosted runners securely, since forks of public repos can modify workflow scripts to do very malicious things on our local machines (e.g., cryptomine Dogecoin).

Workflows in this repo will trigger a workflow called `performance_analysis.yml` in opensim-perf via the GitHub action [Codex-/return-dispatch@v1.12.0](https://github.com/marketplace/actions/return-dispatch). The sister action [Codex-/await-remote-run@v1.11.0](https://github.com/marketplace/actions/await-remote-run) lets us wait for the workflow on opensim-perf to finish before continuing.  The third new action [dawidd6/action-download-artifact@v3.0.0](https://github.com/marketplace/actions/download-workflow-artifact) allows us to download artifacts across repositories; these "handoffs" require the new personal access token `PERF_DISPATCH_PAT` which I've created and added as a secret variable to opensim-org. 

I've also upgraded `actions/upload-artifact` to version 4 (i.e., `actions/upload-artifact@v4`), which now will upload artifacts for individual jobs immediately when they are available. We no longer have to wait for the full CI workflow to finish before the uploads occurs, which means that artifacts for jobs that passed will now be available even if other jobs fail. 

One minor downside of this change is for PRs where performance testing is skipped, the new job `Mac [target]` will still show up as "skipped". This is a limitation of the GitHub Actions UI for now.

The end-to-end workflow looks something like this:
1. A pull request (with [perf] somewhere in the title or body) triggers `continuous_integration.yml`.
2. The CI builds both `Mac [target]` and `Mac [source]` and uploads the artifacts.
3. The new job `performance_analysis`, which depends on the two Mac builds, triggers the workflow `performance_analysis.yml` on opensim-perf. It passes the commit hash and artifact names so the opensim-perf workflow knows where and what to download.
4. The job `performance_analysis` now waits for opensim-perf to finish its workflow.
5. opensim-perf downloads the two artifacts, unzips them, runs the performance testing suite, and uploads an artifact containing the results in the markdown table `performance_results.md`.
6. The opensim-core workflow resumes, downloads the table, and posts it to the PR description body (or overwrites the current table if it exists).

### Testing I've completed

All of it.

1. Any security vulnerabilities I might have missed?
2. Stability concerns about the custom actions from the GitHub Actions marketplace? They seem pretty stable, well-used, and well-documented, but the workflow wholly depends on them currently. 

### CHANGELOG.md (choose one)

- no need to update because...I'll update CONTRIBUTING.md instead.

### Performance analysis

|                      Arm26 |       0.33 |     0.00 |       0.34 |     0.00 |    0.97 |
| passive_dynamic_noanalysis |       3.42 |     0.10 |       3.31 |     0.15 |    1.03 |
|             ToyDropLanding |       0.13 |     0.00 |       0.13 |     0.00 |    0.99 |
|                   Gait2354 |       0.34 |     0.00 |       0.34 |     0.00 |    1.00 |
|            passive_dynamic |       4.13 |     0.00 |       4.14 |     0.01 |    1.00 |
|   ToyDropLanding_nomuscles |       0.12 |     0.00 |       0.12 |     0.00 |    1.00 |
|             RajagopalModel |       6.26 |     0.00 |       6.34 |     0.01 |    0.99 |

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3686)

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -392,10 +392,11 @@ jobs:
     outputs:
       version: ${{ steps.configure.outputs.version }}
 
-    env:
-      commitmsg: ${{ github.event.head_commit.message }}
-
     steps:
+    - name: Debug Commit Message
+      run: |
+       echo "Commit Message: ${{ github.event.head_commit.message }}"
+      
     - uses: actions/checkout@v3
     
     - uses: actions/setup-python@v4
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -404,10 +404,10 @@ jobs:
 
     - name: Testing commit message
       run: |
-        echo github.event.head_commit.message
-        echo contains($commitmsg, '[perf]')
+        # echo github.event.head_commit.message
+        # echo contains($commitmsg, '[perf]')
         echo $commitmsg
-        echo ${{ contains($commitmsg, '[perf]') }}
+        # echo ${{ contains($commitmsg, '[perf]') }}
         
     - name: Install Homebrew packages
       # Save the gfortran version to a file so we can use it in the cache key.
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -573,7 +573,7 @@ jobs:
     - name: Post Results to PR Comment
       run: |
         PR_NUMBER=$(jq --raw-output .pull_request.number "$GITHUB_EVENT_PATH")
-        EXISTING_COMMENT_CONTENT=$(gh api /repos/opensim-org/opensim-core/issues/${PR_NUMBER} | jq -r '.body')
+        EXISTING_COMMENT_CONTENT=$(gh api /repos/opensim-org/opensim-core/pulls/${PR_NUMBER} | jq -r '.body')
         CUSTOM_HEADER="### Performance Analysis Results"
 
         REVIEWABLE_TAG="<!-- Reviewable:start -->"
@@ -596,15 +596,16 @@ jobs:
         "
         NEW_COMMENT_BODY="${EXISTING_COMMENT_CONTENT/"${CONTENT_TO_REPLACE}"/${NEW_CONTENT}}"
 
-        COMMENT_ID=$(gh api repos/opensim-org/opensim-core/issues/${PR_NUMBER}/comments | jq -r '.[0].id')
-        echo ${COMMENT_ID}
+        # COMMENT_ID=$(gh api repos/opensim-org/opensim-core/issues/${PR_NUMBER}/comments | jq -r '.[0].id')
+        # echo ${COMMENT_ID}
 
-        # Update the existing top-level comment
+        # Update the existing description
         gh api \
           --method PATCH \
           -H "Content-Type: application/json" \
-          -H "X-GitHub-Api-Version: 2022-11-28" \
-          repos/opensim-org/opensim-core/issues/${PR_NUMBER}/comments/${COMMENT_ID} \
+          -H "Accept: application/vnd.github.v3+json" \
+          -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+          "/repos/opensim-org/opensim-core/pulls/${PR_NUMBER}" \
           -f "body=${NEW_COMMENT_BODY}"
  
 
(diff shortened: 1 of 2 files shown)


aymanhab (approved):
LGTM, thanks @nickbianco  I know how hard it was to get this through, nice work. Only couple clarifications that are not blocking since this has no side effects unless explicitly invoked

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3686#-:-NrrKLOV3qqedpG0HEZL:bms77fn)* status: all files reviewed, 2 unresolved discussions (waiting on @nickbianco and @pepbos)

___
*[`.github/workflows/continuous_integration.yml` line 380 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3686#-NrrIlaJAW8oP1DJObpg:-NrrIlaJAW8oP1DJObph:b-2if20q) ([raw file](https://github.com/opensim-org/opensim-core/blob/f7e1fba131cf7ab39bd666512ca4bf4e0cae014d/.github/workflows/continuous_integration.yml#L380)):*
> ```YAML
>     - name: Upload opensim-core
>       uses: actions/upload-artifact@v4
>       with:
> ```

Curious if/why you needed to upgrade to upload-artifact@v4? Should we update everywhere else?
___
*[`.github/workflows/continuous_integration.yml` line 548 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3686#-NrrK6pz4fbIuMQruzc5:-NrrK6pz4fbIuMQruzc6:b9rjrtz) ([raw file](https://github.com/opensim-org/opensim-core/blob/f7e1fba131cf7ab39bd666512ca4bf4e0cae014d/.github/workflows/continuous_integration.yml#L548)):*
> ```YAML
>         run_id: ${{ steps.return_dispatch.outputs.run_id }}
>         run_timeout_seconds: 1500 
> ```

Any reason for 1500 seconds, other than empirical?

nickbianco (commented):
Thanks @aymanhab! Made a few changes based on your comments.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3686#-:-NrrYJZA64lM0OS8NjXw:bxe3akm)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @aymanhab and @pepbos)

___
*[`.github/workflows/continuous_integration.yml` line 380 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3686#-NrrIlaJAW8oP1DJObpg:-NrrXF2GAqseHBZQcYnX:brtd46e) ([raw file](https://github.com/opensim-org/opensim-core/blob/f7e1fba131cf7ab39bd666512ca4bf4e0cae014d/.github/workflows/continuous_integration.yml#L380)):*

Curious if/why you needed to upgrade to upload-artifact@v4? Should we update everywhere else?
</blockquote></details>

This version will upload artifacts for individual jobs as they complete, rather than waiting for all jobs to complete. So definitely worth upgrading everywhere.
___
*[`.github/workflows/continuous_integration.yml` line 548 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3686#-NrrK6pz4fbIuMQruzc5:-NrrXZDpDPt2IFA-DCOi:b-1m3nsv) ([raw file](https://github.com/opensim-org/opensim-core/blob/f7e1fba131cf7ab39bd666512ca4bf4e0cae014d/.github/workflows/continuous_integration.yml#L548)):*

Any reason for 1500 seconds, other than empirical?
</blockquote></details>

This is somewhat arbitrary. Definitely overkill right now, but tests might get longer in the future. For now I've decreased it down to 10 minutes.

aymanhab (approved):
LGTM, thanks @nickbianco , ci is failing on mac though, please rerun/check before merging.

nickbianco:
Thanks @aymanhab!

PR #3687, opened 2024-01-25
Fix getConnectee throwing if called during finalizeConnections

adamkewley:
Fixes an issue I'm having with `SocketDefinedFrame`s implementation (upcoming PR).

Effectively, the bug is that:

- After adding something to a model programmatically, you _must_ call `finalizeConnections` in order to ensure that the socket pointers are baked into the socket's `componentPath` property
- Several use-cases in [opensim-creator](https://github.com/ComputationalBiomechanicsLab/opensim-creator) (e.g. undo/redo) call for copying the model, which wipes those pointers
- This is *usually* fine, because I typically re-call `finalizeConnections` on those copies
- However, `finalizeConnections` has very specific semantics that don't work once you have components in the model graph (e.g. custom frames) that depend on components that aren't in the model graph (e.g. `Station`s):
  - It finalizes the `Model`'s socket connections (non-recursive)
  - Then it creates+traverses a multi-body graph
  - For each frame in the graph, it calls `frame.finalizeConnections(model)`
  - Which finalizes the frame's sockets, and the frame's children's sockets
  - However, it doesn't finalize the sockets of anything the frame is connected to (i.e. the socket finalization performed during multibody graph building only satisfies the data _hierarchy_, not the socket graph)

So, in the edge-case that you:

- Have a custom component that will participate in the model graph (e.g. `StationDefinedFrame`)
- And that component depends on stuff via `SOCKET`s (e.g. `Station`s somewhere else in the hierarchy)
- And calling `finalizeConnections` on the component requires pulling information from those dependencies via those sockets (e.g. calling `finalizeConnections` on a `StationDefinedFrame` may want to query the `Station`s for body-fixed positions, base frame, etc.)
- Then it'll fail

The solution is PRed, and it effectively relaxes sockets such that they work fine even if you don't call `finalizeConnection` on them. This effectively downgrades `finalizeConnection` from "hard requirement for using a socket" to "performance-enhancing requirement".

### Brief summary of changes

- Make `Socket::getConnectee` work, even if `finalizeConnection` hasn't been called on it yet

### Testing I've completed

- The OpenSim test suite
- Which had a failure because one test expects sockets to fail if used before `finalizeConnection`. This test was commented out because that isn't a strictly required behavior (it's a desirable one, for perf reasons, but it isn't a hard requirement)

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3687)

### DIFF
--- a/OpenSim/Common/ComponentSocket.h
+++ b/OpenSim/Common/ComponentSocket.h
@@ -319,14 +319,14 @@ private:
 
     void setConnecteePathInternal(const std::string& name, int index) {
         using SimTK::isIndexInRange;
-        SimTK_INDEXCHECK(index, getNumConnectees(),
+        SimTK_INDEXCHECK(index, static_cast<int>(getNumConnectees()),
                          "AbstractSocket::setConnecteePath()");
         updConnecteePathProp().setValue(index, name);
     }
 
     const std::string& getConnecteePathInternal(int index) const {
         using SimTK::isIndexInRange;
-        SimTK_INDEXCHECK(index, getNumConnectees(),
+        SimTK_INDEXCHECK(index, static_cast<int>(getNumConnectees()),
                         "AbstractSocket::getConnecteePath()");
         return getConnecteePathProp().getValue(index);
     }
@@ -375,7 +375,7 @@ public:
     typedef std::vector<SimTK::ReferencePtr<const T>> ConnecteeList;
 
     // default copy constructor
-    
+
     virtual ~Socket() {}
     
     Socket<T>* clone() const override { return new Socket<T>(*this); }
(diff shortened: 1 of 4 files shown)


OpenSim/Common/ComponentSocket.h
```diff
@@ -319,14 +319,14 @@ class OSIMCOMMON_API AbstractSocket {
 
     void setConnecteePathInternal(const std::string& name, int index) {
         using SimTK::isIndexInRange;
-        SimTK_INDEXCHECK(index, getNumConnectees(),
+        SimTK_INDEXCHECK(index, static_cast<int>(getNumConnectees()),
```
adamkewley, line 322: These changes are because newer MSVCs find the `int` vs `unsigned` comparisons ambiguous.

adamkewley:
Need to check what's the deal with MacOS/Ubuntu

adamkewley:
@nickbianco this is ready for review if you're willing!

The function was moved to `Component.h`, because I noticed that other component-dependent templated functions for `Socket<T>` appear there. A test (for the failing behavior) was also commented out (because it's not longer a strictly failing case, but a slow one).

Hopefully, this relaxation of the connection finalization behavior is acceptable? I think there's some (reasonable) debate over whether it should be allowed (because it may encourage people to skip finalizing connections), but the counterpoint is that any code that (e.g.) calls `buildSystem` will be finalized anyway.

nickbianco (approved):
> but the counterpoint is that any code that (e.g.) calls `buildSystem` will be finalized anyway.

I agree and can't think of a reason why this would be an issue. It won't affect most users and will give more flexibility in model building. You could ask @aseth1 if there's any reason why this might be a bad idea: he was the original `Socket` designer. 

All that aside: implementation LGTM.

> Reviewed 2 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3687#-:-NpLZrBy7H85hHvtfkDb:bm8igab)* status: all files reviewed, 1 unresolved discussion

nickbianco (approved):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3687)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @adamkewley)

PR #3691, opened 2024-01-26
(Slightly more than) minimally convert `testMuscleFirstOrderActivationDynamicModel.cpp` to Catch2

nickbianco:
Fixes issue #3555

### Brief summary of changes

Main test conversion is self-explanatory. Also updated the headers in `testModelProcessor.cpp`. 

### CHANGELOG.md (choose one)

- no need to update because...internal testing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3691)

### DIFF
--- a/OpenSim/Actuators/Test/testModelProcessor.cpp
+++ b/OpenSim/Actuators/Test/testModelProcessor.cpp
@@ -16,13 +16,14 @@
  * limitations under the License.                                             *
  * -------------------------------------------------------------------------- */
 
-#include "OpenSim/Moco/Test/Testing.h"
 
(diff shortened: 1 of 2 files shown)


adamkewley (approved):
Seems fine - I didn't go over it with a fine tooth-comb but it's a test suite, it's passing, and I don't imagine you'd drop any functional tests

(the diff's big because things were moved into anonymous namespaces etc?)

nickbianco:
> (the diff's big because things were moved into anonymous namespaces etc?)

Yes, essentially. The helper functions were also at the bottom of the cpp file, so it wasn't the standard "wrap existing test functions with Catch2 sections" type of change.

PR #3695, opened 2024-02-01
Refactor testComponents.cpp to use Catch and logging interface

adamkewley:
Heavily refactors `OpenSim/Tests/Components/testComponents.cpp` to:

### Brief summary of changes

- Use `Catch2` with `DYNAMIC_SECTION`, so that particular component failures are clearly labelled in the test log
- Use `OpenSim::log_info` and `OpenSim::log_warn`, rather than `std::cout` for logging, so that the messages are easier to read
- Refactor the individual component test cases into `ComponentTestCases`, so there's one central place for controlling component lifetime, which components are tested, etc. (#3694 essentially needs this if I can't rewrite the tests to correctly build transient socket dependencies)
- Change most helper functions to take `Component&` rather than `Component*` (the pointer was never rewritten or nullchecked, so a reference is preferable)
- Replace raw `new`s to `std::unique_ptr`, where it makes sense

### Testing I've completed

- It's a test
- And the output seems largely the same (i.e. checks don't appear to have been removed, etc.)

### CHANGELOG.md (choose one)

- no need to update because boring

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3695)

### DIFF
--- a/OpenSim/Tests/Components/CMakeLists.txt
+++ b/OpenSim/Tests/Components/CMakeLists.txt
@@ -1,7 +1,9 @@
+find_package(Catch2 REQUIRED HINTS "${OPENSIM_DEPENDENCIES_DIR}/catch2")
+
 file(GLOB TEST_PROGS "test*.cpp")
 OpenSimCopySharedTestFiles(gait10dof18musc_subject01.osim)
 
 OpenSimAddTests(
     TESTPROGRAMS ${TEST_PROGS}
-    LINKLIBS osimTools
+    LINKLIBS osimTools Catch2::Catch2WithMain
     )
--- a/OpenSim/Tests/Components/testComponents.cpp
+++ b/OpenSim/Tests/Components/testComponents.cpp
@@ -41,7 +41,7 @@ namespace {
 
     // helper: adds `instance` to the model in the appropriate collection (`Body` goes
     // into `BodySet`, etc.)
-    void addObjectAsComponentToModel(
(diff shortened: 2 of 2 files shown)


adamkewley:
Looks like the `unique_ptr` refactor may have created a segfault in the mac tests (e.g. the test suite was assuming the previously-leaked pointers were retained or similar)

I'll have to rerun this with libASAN to figure out where the assumption breaks.

adamkewley:
The same segfault appears to also be on the `main` branch if i'm analyzing this test suite with libASAN O_o

adamkewley:
The segfault that libASAN was detecting was because a loop in `testComponents` doesn't (and didn't) bounds-check access to a component list iterator.

(this doesn't necessarily mean that it's the bug in MacOS, but hopefully it is)

adamkewley:
Should be ok to go now (unless there's a desire to also run it through MacOS's libASAN, but I don't imagine it's necessary for a test suite refactor)

nickbianco (approved):
:lgtm: Nice work!

> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3695#-:-NqKXT9c7v36qme9PRdK:bcmw7ss)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pepbos)

PR #3701, opened 2024-02-11
[Moco] Adding support for `PrescribedController` and future controller types

nickbianco:
Fixes issue #3605 

### Brief summary of changes

This PR adds support for models with `PrescribedController`s in Moco optimizations, and lays some ground work for future controller types that will map OCP controls to `Model` controls. Therefore, going forward, the relationship between OCP controls and an OpenSim model controls will no longer necessarily be a one-to-one mapping. These changes include support for the two simplest control mapping cases: 1) one-to-one mapping (i.e., current behavior) and 2) controls associated with a `PrescribedController` are excluded from the OCP. 

Changes include:
- Added `ControlAllocator` which manages a `Vector` discrete variable containing the OCP controls and provides `Output` channels to map these controls to OpenSim `Model` controls.
- Removed `DiscreteController` in favor of `ControlAllocator`.
- Added the abstract class `InputController`. This owns a list `Input` which concrete implementations can use to take a subset of the `Output` values from `ControlAllocator` to be mapped to `Model` controls.
- Added `ActuatorInputController`, the simplest concrete implementation of `InputController`. This controller provides a one-to-one mapping between OCP controls and `Model` controls. This enables the original Moco behavior for controls, but also supports handling a subset of `Model` controls (e.g., when a `PrescribedController is present).
- Added options to ignore controlled `Actuator`s in relevant `MocoGoal`s via `setIgnoreControlledActuators()`.
- Various minor drive-by fixes.

### Testing I've completed

Added a test for `PrescribedController` to `testMocoInterface.cpp`.

- The name "inputs" for the `InputController`'s list `Input`. Something less generic would probably be preferred, but I'm having trouble coming up with a term to describe the scalar values that drive an `InputController`. 
- Anything else.

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3701)


chrisdembia (changes requested):
I like all the info logs about "ignored" but I think it would be useful for these messages to mention that the user *requested* this ignoring. Also, I'm slightly concerned that these logs could be quite noisy for a problem with lots of prescribed controls, but maybe that can be handled in a separate PR if it is actually a problem.

I didn't quite finish reviewing. I skipped the meatiest parts lol.

Exciting stuff!

> Reviewed 32 of 35 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-:-NqaBxu_7t-ppmfNyy_w:b-9zfn9g)* status: 32 of 35 files reviewed, 11 unresolved discussions (waiting on @nickbianco)

___
*[`OpenSim/Moco/MocoControlBoundConstraint.h` line 38 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-Nqa6Aqd62RWWE9fZW4J:-Nqa6Aqd62RWWE9fZW4K:b-9ke3e) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Moco/MocoControlBoundConstraint.h#L38)):*
> ```objc
> 
> If you wish to constrain all control signals except those belonging to a
> user-defined controller, pass 'true' to setIgnoreControlledActuators().
> ```

consider giving an example of a user-defined controller
___
*[`OpenSim/Moco/MocoProblemRep.h` line 324 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-Nqa6kau8pZl0i4hOJMP:-Nqa6kau8pZl0i4hOJMQ:bp1yx64) ([raw file](https://github.com/opensim-org/opensim-core/blob/f751a7cf23141106ffaf2826e482a50019ad524f/OpenSim/Moco/MocoProblemRep.h#L324)):*
> ```objc
>         return m_implicit_component_refs;
>     }
>     /// @}
> ```

is this doxygen group unmatched now?
___
*[`OpenSim/Moco/Components/ControlAllocator.h` line 26 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaI-Eu4msJWrUDem6C:-NqaI-Eu4msJWrUDem6D:blpaa54) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Moco/Components/ControlAllocator.h#L26)):*
> ```objc
> namespace OpenSim {
> 
> class OSIMMOCO_API ControlAllocator : public ModelComponent {
> ```

the purpose of this class probably isn't very obvious; add a description?
___
*[`OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.h` line 134 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaBV239qYH5zX0p5sD:-NqaBV240JAwJlTaFvP8:b5reqxc) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.h#L134)):*
> ```objc
> to the cost.
> 
> If you wish to track all control signals except those belonging to a
> ```

nit: The language is slightly different from below -- "controls belonging to controller" versus "controls belonging to actuator controlled by controller". Is this okay, or might this lead to confusion? Would some higher-level doxygen user guide documentation about different types of control be helpful?
___
*[`OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.cpp` line 167 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaAD6b3DHzuNgEHF70:-NqaAD6b3DHzuNgEHF71:b45h) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.cpp#L167)):*
> ```cpp
>                 controlToTrack) != actuatorInputControls.end();
>         if (getIgnoreControlledActuators() && !isActuatorInputControl) {
>             log_info("MocoControlTrackingGoal: Control '{}' is already "
> ```





_Suggestion:_
```cpp
Actuator
```
___
*[`OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.cpp` line 169 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaAkZr5qVCK2E1jOU_:-NqaAkZr5qVCK2E1jOUa:undefined) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.cpp#L169)):*
> ```cpp
>             log_info("MocoControlTrackingGoal: Control '{}' is already "
>                      "controlled by a user-defined controller and will be "
>                      "ignored.", controlToTrack);
> ```





_Suggestion:_
```cpp
ignored, as requested
```
___
*[`OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.cpp` line 207 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaGS2WAhsHxnqrK_Mf:-NqaGS2WAhsHxnqrK_Mg:bn9kdpx) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.cpp#L207)):*
> ```cpp
> }
> 
> void MocoControlTrackingGoal::calcIntegrandImpl(
> ```

If you updated MocoPeriodicityGoal to realizeVelocity and get the controls from the model, don't you have to do the same here, and everywhere else? Does it make sense for the `input` to still have controls? I guess those would now be only the non-actuator controls?
___
*[`OpenSim/Moco/MocoGoal/MocoInitialActivationGoal.cpp` line 32 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-Nqa7pBu1M--1XW-JRrF:-Nqa7pBu1M--1XW-JRrG:b18rmm0) ([raw file](https://github.com/opensim-org/opensim-core/blob/f751a7cf23141106ffaf2826e482a50019ad524f/OpenSim/Moco/MocoGoal/MocoInitialActivationGoal.cpp#L32)):*
> ```cpp
> 
>     for (const auto& muscle : model.getComponentList<Muscle>()) {
>         if (!muscle.get_ignore_activation_dynamics()) {
> ```

What's the reason for getting rid of this condition? do we no longer support muscles without activation dynamics?
___
*[`OpenSim/Moco/MocoGoal/MocoPeriodicityGoal.cpp` line 126 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaG48Z3HWpPVQH2ZgJ:-NqaG48Z3HWpPVQH2ZgK:b-b6grx8) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Moco/MocoGoal/MocoPeriodicityGoal.cpp#L126)):*
> ```cpp
>     setRequirements(
>             0, (int)m_indices_states.size() + (int)m_indices_controls.size(),
>             SimTK::Stage::Velocity);
> ```

I guess states are available earlier than Velocity but Controls, if computed by a controller, are not available until Velocity? Makes sense
___
*[`OpenSim/Simulation/SimulationUtilities.h` line 173 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaDD5g5szfAeDV3SiL:-NqaDD5g5szfAeDV3SiM:b-aqxfmf) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Simulation/SimulationUtilities.h#L173)):*
> ```objc
> std::vector<std::string> createControlNamesFromModel(const Model& model);
> 
> /// Given a controller of type `T`, create a vector of control names based on
> ```

this docstring could better convey that control names will be returned for all controllers of type T--that there might be multiple such controllers (although, for your given usage, it seems somewhat unlikely).

Also, I wonder if it makes sense to return a `std::unordererd_set` instead of a vector, to ensure that there are no duplicates (which I think is undesirable), that would make your "contains" checks in the goals a little simpler.
___
*[`OpenSim/Simulation/SimulationUtilities.h` line 185 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaCeOSARMDb8vhlecM:-NqaCeOSARMDb8vhlecN:b-jvq9hg) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Simulation/SimulationUtilities.h#L185)):*
> ```objc
> 
>     // Check that T is a Controller.
>     if (!std::is_base_of<Controller, T>::value) {
> ```

Beautiful!

nit: I think this can be a `static_assert`, because you know this at compile-time. https://en.cppreference.com/w/cpp/language/static_assert

chrisdembia (changes requested):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3701)* status: 32 of 35 files reviewed, 13 unresolved discussions (waiting on @nickbianco)

___
*[`OpenSim/Moco/MocoProblemRep.cpp` line 106 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaJis-DEdg7AKOFEt8:-NqaJis-DEdg7AKOFEt9:b-rioafg) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Moco/MocoProblemRep.cpp#L106)):*
> ```cpp
> 
>     // Check that any controllers added by the user are valid.
>     std::vector<std::string> controlledActuatorPaths;
> ```

I wonder if an unordered_set is a better fit
___
*[`OpenSim/Moco/MocoProblemRep.cpp` line 116 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaJ_wZ759dYD7dDkb_:-NqaJ_wZ759dYD7dDkba:undefined) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Moco/MocoProblemRep.cpp#L116)):*
> ```cpp
> 
>         // TODO: When we support other types of controllers, we need to check
>         // that there are no user-defined ActuatorInputControllers in the modle
> ```





_Suggestion:_
```cpp
model
```

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3701)* status: 32 of 35 files reviewed, 12 unresolved discussions (waiting on @chrisdembia)

___
*[`OpenSim/Moco/MocoControlBoundConstraint.h` line 38 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-Nqa6Aqd62RWWE9fZW4J:-Nqf3xLc4ZyubkZ0clCP:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Moco/MocoControlBoundConstraint.h#L38)):*

consider giving an example of a user-defined controller
</blockquote></details>

Done.
___
*[`OpenSim/Moco/MocoProblemRep.h` line 324 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-Nqa6kau8pZl0i4hOJMP:-Nqf4Euv072sn2BKFujq:b-f06h8t) ([raw file](https://github.com/opensim-org/opensim-core/blob/f751a7cf23141106ffaf2826e482a50019ad524f/OpenSim/Moco/MocoProblemRep.h#L324)):*

is this doxygen group unmatched now?
</blockquote></details>

Yes, good catch!
___
*[`OpenSim/Moco/MocoProblemRep.cpp` line 106 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaJis-DEdg7AKOFEt8:-NqfBhos4-3QklX1fbjy:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Moco/MocoProblemRep.cpp#L106)):*

I wonder if an unordered_set is a better fit
</blockquote></details>

Done.
___
*[`OpenSim/Moco/MocoProblemRep.cpp` line 116 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaJ_wZ759dYD7dDkb_:-Nqe3q7d02yIaqWrjQgO:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Moco/MocoProblemRep.cpp#L116)):*
> ```cpp
> 
>         // TODO: When we support other types of controllers, we need to check
>         // that there are no user-defined ActuatorInputControllers in the modle
> ```

Done.
___
*[`OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.h` line 134 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaBV239qYH5zX0p5sD:-NqfERE07ktrsY52CkOK:b-5xvnoc) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.h#L134)):*
I tried to unify the language between the class and method comments.

> Would some higher-level doxygen user guide documentation about different types of control be helpful?

I think this would be nice. I can add some documentation on this PR.
___
*[`OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.cpp` line 167 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaAD6b3DHzuNgEHF70:-NqfBl_y-mxT4OgWcvq7:b92o960) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.cpp#L167)):*
> ```cpp
>                 controlToTrack) != actuatorInputControls.end();
>         if (getIgnoreControlledActuators() && !isActuatorInputControl) {
>             log_info("MocoControlTrackingGoal: Control '{}' is already "
> ```

I intentionally left this as "Control", since the control name could be for a non-scalar actuator (e.g., `/forceset/body_actuator_0`), even though I know 99.9% of the time it will be simply the actuator path. 
___
*[`OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.cpp` line 169 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaAkZr5qVCK2E1jOU_:-NqfBlIv5tBj_ZLf1HjL:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.cpp#L169)):*
> ```cpp
>             log_info("MocoControlTrackingGoal: Control '{}' is already "
>                      "controlled by a user-defined controller and will be "
>                      "ignored.", controlToTrack);
> ```

Done.
___
*[`OpenSim/Moco/MocoGoal/MocoInitialActivationGoal.cpp` line 32 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-Nqa7pBu1M--1XW-JRrF:-NqfCKc60EGhniBH7UKC:b-q4skfu) ([raw file](https://github.com/opensim-org/opensim-core/blob/f751a7cf23141106ffaf2826e482a50019ad524f/OpenSim/Moco/MocoGoal/MocoInitialActivationGoal.cpp#L32)):*

What's the reason for getting rid of this condition? do we no longer support muscles without activation dynamics?
</blockquote></details>

This is an oversight on my part: I refactored and unrefactored this code when messing around with the implementation and forgot to put it back!
___
*[`OpenSim/Simulation/SimulationUtilities.h` line 173 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaDD5g5szfAeDV3SiL:-NqfB_yMCEtcKoC2vHoI:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Simulation/SimulationUtilities.h#L173)):*

this docstring could better convey that control names will be returned for all controllers of type T--that there might be multiple such controllers (although, for your given usage, it seems somewhat unlikely).

Also, I wonder if it makes sense to return a `std::unordererd_set` instead of a vector, to ensure that there are no duplicates (which I think is undesirable), that would make your "contains" checks in the goals a little simpler.
</blockquote></details>

Done.

nickbianco (commented):
@chrisdembia, thanks for the great feedback so far! Ready for review again.

Aside from adding some high-level Doxygen documentation (and perhaps a few more unit tests), I still need to update each `MocoGoal` to make sure the correct controls are being computed. Before doing that, it would be helpful to see what you think about my changes to `MocoControlGoal`, which attempts to use raw controls when it makes sense (i.e., no user-defined controllers) and otherwise compute controls from the model.

We're back baby!



> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-:-NqfORIe8XkQyvTmalf_:bxymv0t)* status: 20 of 35 files reviewed, 12 unresolved discussions (waiting on @chrisdembia)

___
*[`OpenSim/Moco/Components/ControlAllocator.h` line 26 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaI-Eu4msJWrUDem6C:-NqfNzXEE9Vsc48jWdYF:b95n7jc) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Moco/Components/ControlAllocator.h#L26)):*

the purpose of this class probably isn't very obvious; add a description?
</blockquote></details>

Ah, of course. Done.

Also, I'm now wondering if this class needs to be exposed to the API. It is very specific to Moco, but I suppose others could find it useful.
___
*[`OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.cpp` line 207 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaGS2WAhsHxnqrK_Mf:-NqfFOGqA87OUzH8iXNG:b7w1062) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Moco/MocoGoal/MocoControlTrackingGoal.cpp#L207)):*

If you updated MocoPeriodicityGoal to realizeVelocity and get the controls from the model, don't you have to do the same here, and everywhere else? Does it make sense for the `input` to still have controls? I guess those would now be only the non-actuator controls?
</blockquote></details>

Yes, you're right. The `input` will still contain the raw controls, it just depends on whether or not you want to use them. If the raw controls are the same as the `Model` controls, you probably want to use them to avoid calling `realizeVelocity()` here. 

I've made some changes to `MocoControlGoal` that I think could be applied to all goals depending on control values.
 
___
*[`OpenSim/Moco/MocoGoal/MocoPeriodicityGoal.cpp` line 126 at r1](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NqaG48Z3HWpPVQH2ZgJ:-NqfFBib2pqBylRf-pTn:bsak5k6) ([raw file](https://github.com/opensim-org/opensim-core/blob/360a6b5c58eeedb2fa51b42d16c40543df7ddba7/OpenSim/Moco/MocoGoal/MocoPeriodicityGoal.cpp#L126)):*

I guess states are available earlier than Velocity but Controls, if computed by a controller, are not available until Velocity? Makes sense
</blockquote></details>

I think we should still try to use the raw controls from the solver whenever possible. Take a look at what I've done for `MocoControlGoal` and let me know what you think.

aymanhab:
A drive-by (as called by a colleague), the use of the name allocator would be confusing to anybody who uses containers/c++ constructs that utilize std::allocator (https://en.cppreference.com/w/cpp/memory/allocator) to mean something specific to memory management. Are there other suggestions for the name/alternatives?

nickbianco:
@aymanhab `ControlDistributor` would be a reasonable alternative. I waffled between that and `ControlAllocator` and would be fine swtiching.

aymanhab:
Love the ControlDistributor name but I'm biased :man_shrugging:

chrisdembia (changes requested):
> I still need to update each `MocoGoal` to make sure the correct controls are being computed. Before doing that, it would be helpful to see what you think about my changes to `MocoControlGoal`, which attempts to use raw controls when it makes sense (i.e., no user-defined controllers) and otherwise compute controls from the model.

Yeah I think the main question is how to avoid the duplication. In my comments this time I suggest moving more of the logic up to `MocoCasOCProblem`.

### ControlAllocator naming

I'm fine with ControlAllocator because I'm familiar with Simbody's `allocate*` methods, but you have to know Simbody to appreciate this.

I'm also fine with ControlDistributor.

> Reviewed 2 of 35 files at r1, 11 of 13 files at r2, 2 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-:-NquT8An3KKKDcBfH0HF:b7oiqal)* status: 34 of 35 files reviewed, 11 unresolved discussions (waiting on @nickbianco)

___
*[`OpenSim/Moco/MocoProblemRep.cpp` line 142 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NquNEyR8QCzhFy4450g:-NquNEyR8QCzhFy4450h:bx998e8) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoProblemRep.cpp#L142)):*
> ```cpp
> 
>     // Finalize Actuator socket connections for all InputControllers. This is
>     // so we can get the names of the control names from the controlled
> ```

Aren't the connections below for the non-controlled actuators? Elsewhere we discussed high-level doxygen documentation; hopefully we can define some clear terms there. Also, all actuators are controlled...so perhaps we could use terms like "model-controlled actuators" and "OCP-controlled/optimizer-controlled actuators"



_Suggestion:_
```cpp
non-controlled
```
___
*[`OpenSim/Moco/MocoProblemRep.cpp` line 142 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NquPctZ35P9tRJdyQPv:-NquPctZ35P9tRJdyQPw:bbcjb2i) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoProblemRep.cpp#L142)):*
> ```cpp
> 
>     // Finalize Actuator socket connections for all InputControllers. This is
>     // so we can get the names of the control names from the controlled
> ```

"names of the control names" is confusing
___
*[`OpenSim/Moco/MocoProblemRep.cpp` line 153 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NquBb051juHlHEJ8hh9:-NquBb051juHlHEJ8hhA:b2ni8l0) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoProblemRep.cpp#L153)):*
> ```cpp
>     // ActuatorInputController we just added to control actuators that are not
>     // already controlled by a user-defined controller.
>     auto inputControllerNames =
> ```

nit



_Suggestion:_
```cpp
inputControllerControlNames
```
___
*[`OpenSim/Moco/MocoProblemRep.cpp` line 267 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NquNch90jRtfKdHLns_:-NquNch90jRtfKdHLnsa:bp09tgm) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoProblemRep.cpp#L267)):*
> ```cpp
>     // in this copied model.
>     m_model_disabled_constraints = Model(m_model_base);
>     m_model_disabled_constraints.finalizeConnections();
> ```

Why is this necessary here? Don't we typically finalizeProperties (which happens below) before finalizeConnections?
___
*[`OpenSim/Moco/MocoProblemRep.cpp` line 541 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NquQGRJEil8OVMM6oEx:-NquQGRJEil8OVMM6oEy:b-4jx6ws) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoProblemRep.cpp#L541)):*
> ```cpp
>         OPENSIM_THROW_IF(it == controlNames.end(), Exception,
>                 "Control info provided for nonexistent, disabled, or "
>                 "controlled actuator '{}'.",
> ```

bonus for points for telling the user specifically if it's b/c it's a controlled actuator. could save users some debugging time.
___
*[`OpenSim/Moco/MocoProblemRep.cpp` line 596 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NquRfcUADTPMjZ7w3De:-NquRfcUADTPMjZ7w3Df:b-haxflc) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoProblemRep.cpp#L596)):*
> ```cpp
> 
>     // Bound activations based on muscle excitation control bounds. Set this for
>     // all muscles, including those controlled by a user-defined controller.
> ```

we should document this in a prominent easy-to-find place.
___
*[`OpenSim/Moco/Components/ControlAllocator.h` line 35 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NquPmEV1H-LZwk5cCUF:-NquPmEV1H-LZwk5cCUG:b-mqopxr) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/Components/ControlAllocator.h#L35)):*
> ```objc
>  * `setControls()`, `updControls()`, and `getControls()` methods.
>  *
>  * The control values can be allocated using the Output `controls`.
> ```

I find this wording a bit confusing. I would think "a control variable can be allocated using `addControl()`" and "control values are available via the Output `controls`"
___
*[`OpenSim/Moco/Components/ControlAllocator.h` line 40 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-Nqu-dBA5Asump2rL13Z:-Nqu-dBA5Asump2rL13_:b-nc7ggb) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/Components/ControlAllocator.h#L40)):*
> ```objc
>  * @code
>  * auto controlAllocator = make_unique<ControlAllocator>();
>  * controlAllocator->addControl("/residual_pelvis_tilt");
> ```

nit: convey that these don't actually need to correspond to model actuators; these names are arbitrary
___
*[`OpenSim/Moco/MocoGoal/MocoControlGoal.cpp` line 75 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-Nqu4MKhCYgCHR3ayKtB:-Nqu4MKhCYgCHR3ayKtC:bgwrr7a) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoGoal/MocoControlGoal.cpp#L75)):*
> ```cpp
> 
>     // Create a map from control names to their indices in the controls vector.
>     // If we are using the raw controls, we use the control indices from the
> ```

do you define "raw controls" is a high-visibility place?
___
*[`OpenSim/Moco/MocoGoal/MocoControlGoal.cpp` line 82 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-Nqu38qJ8SBJ671fXdNg:-Nqu38qJ8SBJ671fXdNh:b-7jsa83) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoGoal/MocoControlGoal.cpp#L82)):*
> ```cpp
>         controlIndexMap = createSystemControlIndexMap(model);
>     } else {
>         controlIndexMap = model.getComponentList<ControlAllocator>().begin()
> ```

Are you sure you know that the only Controller is a ControlAllocator? Is this ensured by MocoProblemRep? Consider adding an error check somewhere.
___
*[`OpenSim/Moco/MocoGoal/MocoInitialActivationGoal.cpp` line 70 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-Nqu6hRv8badMCgbBD8t:-Nqu6hRv8badMCgbBD8u:b7pd5h5) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoGoal/MocoInitialActivationGoal.cpp#L70)):*
> ```cpp
>         const GoalInput& input, SimTK::Vector& goal) const {
>     // TODO: compute controls in MocoGoal::calcIntegrand() and pass them to
>     // calcIntegrandImpl() by overwriting the IntegrandInput?
> ```

I do think that it's important to get rid of the duplication; in the current state, I think plugin developers are too likely to implement a buggy goal by not handling controls like you're doing here. But doing this in MocoGoal::calcIntegrand seems a bit messy. Can't we tell MocoCasOCProblem if `m_computeControlsFromModel`, and set `input.controls` to be `rawControls` or `model.getControls()` in `MocoCasOCProblem::calcCostIntegrand()`?

What if we always take the controls from the model (that is, `input.controls` is always `model.getControls(state)`, from `MocoCasOCProblem`?). I suspect the performance cost is substantial and we want to avoid it for simulations without model-controlled actuators (which is most of them), but perhaps it's worth checking.

nickbianco (commented):
Some things are still in progress, but here's a few updates:
- Renamed `ControlAllocator` to `ControlDistributor`.
- Moved model control computations to `MocoCasOCProblem` and `TropterProblem`.
- Removed unnecessary "finalize" calls.
- Clarified many comments.
- Added more specific `Exception` messages (including a new check that only one `ControlDistributor` is present).

Still TODO:
- High-level Doxygen documentation (note: include `MocoProblem::setBoundActivationFromExcitation()`).
- Better interface for `InputController, where derived classes can provide "expected" input aliasess, so `ControlDistributor` can find the correct `Output` channels. This will be important when we have `InputController`s with inputs that don't directly correspond to `Actuator` names (e.g., `SynergyController`).

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-:-Nr3lPzQ7o5df7hYI4Jl:b-nsunz4)* status: 18 of 38 files reviewed, 9 unresolved discussions (waiting on @chrisdembia)

___
*[`OpenSim/Moco/MocoProblemRep.cpp` line 142 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NquPctZ35P9tRJdyQPv:-Nr3RAjM0_unShs3UiLD:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoProblemRep.cpp#L142)):*

"names of the control names" is confusing
</blockquote></details>

Done.
___
*[`OpenSim/Moco/MocoProblemRep.cpp` line 142 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NquNEyR8QCzhFy4450g:-Nr3S1wS8Etmpd1J8zIm:bp326r7) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoProblemRep.cpp#L142)):*

Aren't the connections below for the non-controlled actuators? Elsewhere we discussed high-level doxygen documentation; hopefully we can define some clear terms there. Also, all actuators are controlled...so perhaps we could use terms like "model-controlled actuators" and "OCP-controlled/optimizer-controlled actuators"
</blockquote></details>

Rephrased to clarify. This comment actually applies to both "model-controlled" and "OCP-controlled" actuators: it's for any actuator controlled by an `InputController` (which will include `SynergyController` in the future. This allows us to get control names before enforcing requirements set by derived classes of `InputController` (e.g., see `ActuatorInputController::extendConnectToModel()`).
___
*[`OpenSim/Moco/MocoProblemRep.cpp` line 267 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NquNch90jRtfKdHLns_:-Nr3Q4230J2IEMJBsEt2:bb82ouu) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoProblemRep.cpp#L267)):*

Why is this necessary here? Don't we typically finalizeProperties (which happens below) before finalizeConnections?
</blockquote></details>

It's not necessary: this was leftover from an older version of `ControlAllocator` where (I think) it was needed.
___
*[`OpenSim/Moco/MocoProblemRep.cpp` line 541 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NquQGRJEil8OVMM6oEx:-Nr3XWnP2WXcLFGGSwWP:b-18ti3l) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoProblemRep.cpp#L541)):*

bonus for points for telling the user specifically if it's b/c it's a controlled actuator. could save users some debugging time.
</blockquote></details>

Done. Although, this might soon become irrelevant if we allow using control infos to bound controlled actuators (e.g., via path constraints).
___
*[`OpenSim/Moco/MocoProblemRep.cpp` line 596 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NquRfcUADTPMjZ7w3De:-Nr3TP_TCB0WrJHtrZvS:b-6ic02q) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoProblemRep.cpp#L596)):*

we should document this in a prominent easy-to-find place.
</blockquote></details>

I updated the doc-string for `MocoProblem::setBoundActivationFromExcitation()`.  I'll also include this in the Doxygen documentation.
___
*[`OpenSim/Moco/MocoGoal/MocoControlGoal.cpp` line 75 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-Nqu4MKhCYgCHR3ayKtB:-Nr3jNNf4QdFnna2CZAM:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoGoal/MocoControlGoal.cpp#L75)):*

do you define "raw controls" is a high-visibility place?
</blockquote></details>

Done.
___
*[`OpenSim/Moco/MocoGoal/MocoControlGoal.cpp` line 82 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-Nqu38qJ8SBJ671fXdNg:-Nr3kMnc7orOIcoRjfDm:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoGoal/MocoControlGoal.cpp#L82)):*

Are you sure you know that the only Controller is a ControlAllocator? Is this ensured by MocoProblemRep? Consider adding an error check somewhere.
</blockquote></details>

Done.
___
*[`OpenSim/Moco/MocoGoal/MocoInitialActivationGoal.cpp` line 70 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-Nqu6hRv8badMCgbBD8t:-Nr3kO_j-8kXrvNtBzct:bvohjny) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoGoal/MocoInitialActivationGoal.cpp#L70)):*

I do think that it's important to get rid of the duplication; in the current state, I think plugin developers are too likely to implement a buggy goal by not handling controls like you're doing here. But doing this in MocoGoal::calcIntegrand seems a bit messy. Can't we tell MocoCasOCProblem if `m_computeControlsFromModel`, and set `input.controls` to be `rawControls` or `model.getControls()` in `MocoCasOCProblem::calcCostIntegrand()`?

What if we always take the controls from the model (that is, `input.controls` is always `model.getControls(state)`, from `MocoCasOCProblem`?). I suspect the performance cost is substantial and we want to avoid it for simulations without model-controlled actuators (which is most of them), but perhaps it's worth checking.
</blockquote></details>

I totally agree. I've moved all this logic to `MocoCasOCProblem` (and `TropterProblem`) as suggested, and retained the option to use the "raw" controls from the `ControlDistributor` if no user-added `Controller`s are present. 
___
*[`OpenSim/Moco/Components/ControlAllocator.h` line 35 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NquPmEV1H-LZwk5cCUF:-Nr3gl8W0uVhuKKKbeDS:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/Components/ControlAllocator.h#L35)):*

I find this wording a bit confusing. I would think "a control variable can be allocated using `addControl()`" and "control values are available via the Output `controls`"
</blockquote></details>

Done.
___
*[`OpenSim/Moco/Components/ControlAllocator.h` line 40 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-Nqu-dBA5Asump2rL13Z:-Nr3l6aj13kgQF3oSIz9:b-7syavt) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/Components/ControlAllocator.h#L40)):*

nit: convey that these don't actually need to correspond to model actuators; these names are arbitrary
</blockquote></details>

Tried to improve this, but in the process realized that I could make `ControlDistributor` and `InputController` less confusing.

nickbianco:
I've updated the `InputController` interface to better support generic control names (i.e., non-actuator control names). I've also made an update where `Model` control names not included in the trajectories returned by CasADi and tropter are added to the `MocoTrajectory` returned to the user for convenience. Sorry to expand a bit, but I think this is a pretty essential addition.

Working now on Doxygen documentation.

nickbianco (commented):
@chrisdembia ready for review.

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-:-NrHTEkuBd5VLPrWUBzU:b-9zqyvo)* status: 15 of 39 files reviewed, 9 unresolved discussions (waiting on @chrisdembia)

chrisdembia (changes requested):
> Reviewed 7 of 23 files at r4, 6 of 14 files at r5.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3701)* status: 28 of 43 files reviewed, 7 unresolved discussions (waiting on @nickbianco)

___
*[`OpenSim/Moco/MocoProblem.h` line 367 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsCO5NqC3d5W8BhCAbL:-NsCO5NqC3d5W8BhCAbM:b-et72xp) ([raw file](https://github.com/opensim-org/opensim-core/blob/6aa8c842d285e61b9008f60163c3bff868f015ae/OpenSim/Moco/MocoProblem.h#L367)):*
> ```objc
>             "the bounds for muscle control (excitation), using "
>             "min/max control if explicit control bounds are not "
>             "provided. This applies to all controls, including those defined "
> ```

"This includes muscles controlled by user-added Controllers"? This setting applies to muscle activations, not controls, so I'm suggesting replacing "applies to all controls" with a muscle-related term
___
*[`OpenSim/Moco/MocoProblemRep.h` line 151 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsCPp8p3qNTYOkN7L1U:-NsCPp8p3qNTYOkN7L1V:b-fnli9r) ([raw file](https://github.com/opensim-org/opensim-core/blob/6aa8c842d285e61b9008f60163c3bff868f015ae/OpenSim/Moco/MocoProblemRep.h#L151)):*
> ```objc
>     bool isPrescribedKinematics() const { return m_prescribedKinematics; }
>     /// Do we need to compute controls from the model (e.g., because the model
>     /// contains user-defined controllers)?
> ```

Consider adding something like "If the model does not contain user-defined controls, then we prefer to use the controls directly from the optimal control problem, for efficiency" Otherwise it seems like we should just always compute controls from the model.

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3701)* status: 26 of 43 files reviewed, 7 unresolved discussions (waiting on @chrisdembia)

___
*[`OpenSim/Moco/MocoProblem.h` line 367 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsCO5NqC3d5W8BhCAbL:-NsEg4js0iVmDSTQzpXT:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/6aa8c842d285e61b9008f60163c3bff868f015ae/OpenSim/Moco/MocoProblem.h#L367)):*

"This includes muscles controlled by user-added Controllers"? This setting applies to muscle activations, not controls, so I'm suggesting replacing "applies to all controls" with a muscle-related term
</blockquote></details>

Done.
___
*[`OpenSim/Moco/MocoProblemRep.h` line 151 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsCPp8p3qNTYOkN7L1U:-NsEgbAICb0LYy0844LV:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/6aa8c842d285e61b9008f60163c3bff868f015ae/OpenSim/Moco/MocoProblemRep.h#L151)):*

Consider adding something like "If the model does not contain user-defined controls, then we prefer to use the controls directly from the optimal control problem, for efficiency" Otherwise it seems like we should just always compute controls from the model.
</blockquote></details>

Done.

nickbianco:
@aymanhab, this is almost ready to merge, did you have anything else you wanted to check/comment on?

aymanhab (approved):
Thanks for checking @nickbianco I only had a comment about the naming and that has been fully addressed, if the rest in the bowls of Moco  has been reviewed by @chrisdembia then LGTM. Thank you

> Reviewed 1 of 14 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-:-NsF1Myr4T8S4dAxfidL:b-71ln46)* status: 29 of 43 files reviewed, 7 unresolved discussions (waiting on @chrisdembia)

aymanhab (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-:-NsF1mAl8cx22fi3e0ko:bnfp4nl)* status: 29 of 43 files reviewed, 7 unresolved discussions (waiting on @chrisdembia)

chrisdembia (changes requested):
> Reviewed 7 of 23 files at r4, 7 of 14 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3701)* status: all files reviewed, 11 unresolved discussions (waiting on @nickbianco)

___
*[`OpenSim/Moco/MocoCasADiSolver/MocoCasADiSolver.cpp` line 395 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsHVxQd-Xo7Hjj5wHyN:-NsHVxQd-Xo7Hjj5wHyO:b-nauyoq) ([raw file](https://github.com/opensim-org/opensim-core/blob/6aa8c842d285e61b9008f60163c3bff868f015ae/OpenSim/Moco/MocoCasADiSolver/MocoCasADiSolver.cpp#L395)):*
> ```cpp
>     if (getProblemRep().getComputeControlsFromModel()) {
>         updateSolutionControls(mocoSolution, casSolution);
>         mocoSolution = convertToMocoTrajectory<MocoSolution>(casSolution);
> ```

Is there a reason you need to append the model-based controls in the casSolution? Can you directly put them in the MocoSolution?

Also, what's the reason for doing this here instead of a level deeper, so that the casSolution already has the model-based controls? That seems like a better parallel to computing the controls in the MocoCasOCProblem.
___
*[`OpenSim/Moco/MocoCasADiSolver/MocoCasADiSolver.cpp` line 473 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsHXFs08YKFQYlW9OCE:-NsHXFs08YKFQYlW9OCF:b-yhw4q) ([raw file](https://github.com/opensim-org/opensim-core/blob/6aa8c842d285e61b9008f60163c3bff868f015ae/OpenSim/Moco/MocoCasADiSolver/MocoCasADiSolver.cpp#L473)):*
> ```cpp
> 
>     // Append the missing controls to the CasADi solution and
>     // regenerate the MocoSolution.
> ```

you don't regenerate the MocoSolution here.
___
*[`OpenSim/Moco/MocoCasADiSolver/MocoCasOCProblem.h` line 438 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsHYBZe6Giw5OOJRtUm:-NsHYBZe6Giw5OOJRtUn:b-fp1mtu) ([raw file](https://github.com/opensim-org/opensim-core/blob/6aa8c842d285e61b9008f60163c3bff868f015ae/OpenSim/Moco/MocoCasADiSolver/MocoCasOCProblem.h#L438)):*
> ```objc
>         auto& simtkStateDisabledConstraints =
>                 mocoProblemRep->updStateDisabledConstraints();
>         if (m_computeControlsFromModel) {
> ```

Seems like there's potentially an opportunity for a helper function for getting controls that you could use across MocoCasOCProblem.
___
*[`OpenSim/Moco/MocoCasADiSolver/MocoCasOCProblem.h` line 441 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsHaX_s9FD-mWJwJRmq:-NsHaX_s9FD-mWJwJRmr:b-n1fzba) ([raw file](https://github.com/opensim-org/opensim-core/blob/6aa8c842d285e61b9008f60163c3bff868f015ae/OpenSim/Moco/MocoCasADiSolver/MocoCasOCProblem.h#L441)):*
> ```objc
>             m_endpointIntegrandControls =
>                     modelDisabledConstraints.getDefaultControls();
>             modelDisabledConstraints.computeControls(
> ```

In the previous iteration you did realizeVelocity() and getControls(); what's the reason for the switch here?
___
*[`OpenSim/Moco/MocoCasADiSolver/MocoCasOCProblem.h` line 447 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsHZ-SmCkm73VgRUB6m:-NsHZ-SmCkm73VgRUB6n:b1v6vz6) ([raw file](https://github.com/opensim-org/opensim-core/blob/6aa8c842d285e61b9008f60163c3bff868f015ae/OpenSim/Moco/MocoCasADiSolver/MocoCasOCProblem.h#L447)):*
> ```objc
>                     mocoProblemRep->getControlDistributorDisabledConstraints();
>             m_endpointIntegrandControls =
>                     controlDistributorDisabledConstraints.getControls(
> ```

remind me... are we guaranteed that the controls are in the same order regardless of the value of `m_computeControlsFromModel`? Because the goals's initializeImpl() function often stores indices into the controls.

I guess more importantly, the indices gathered by goals in initializeImpl() must be valid inside the goal functions for computing the goal.
___
*[`OpenSim/Moco/MocoCasADiSolver/MocoCasOCProblem.h` line 449 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsHZNUE1IfdAnNF10VV:-NsHZNUE1IfdAnNF10VW:bksj5pn) ([raw file](https://github.com/opensim-org/opensim-core/blob/6aa8c842d285e61b9008f60163c3bff868f015ae/OpenSim/Moco/MocoCasADiSolver/MocoCasOCProblem.h#L449)):*
> ```objc
>                     controlDistributorDisabledConstraints.getControls(
>                         simtkStateDisabledConstraints);
>         }
> ```

In both branches, the memory for the controls is already allocated within the SimTK::State, which is already threadsafe due to the use of the jar things. So unless I'm missing something, I don't think you need to copy the state vectors into a thread_local member variable. You could have a helper function that returns a `const SimTK::Vector*` (or maybe a `const SimTK::Vector&`) and provide that in the `calcIntegrand` struct argument.
___
*[`OpenSim/Moco/MocoGoal/MocoControlGoal.cpp` line 83 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsHbHRsBgWiPM9NnkfY:-NsHbHRsBgWiPM9NnkfZ:b-23o4lt) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoGoal/MocoControlGoal.cpp#L83)):*
> ```cpp
>     } else {
>         controlIndexMap = model.getComponentList<ControlAllocator>().begin()
>                                 ->getControlIndexMap();
> ```

I expected to see a call to `ControlDistributor::getControlIndexMap()` somewhere; can you remove it? 

Why can we now assume the controls map is the same regardless of `computeControlsFromModel`?
___
*[`OpenSim/Moco/MocoGoal/MocoControlGoal.cpp` line 107 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsHaG0vD5TRWRY97Ocb:-NsHaG0vD5TRWRY97Occ:bta811o) ([raw file](https://github.com/opensim-org/opensim-core/blob/6aa8c842d285e61b9008f60163c3bff868f015ae/OpenSim/Moco/MocoGoal/MocoControlGoal.cpp#L107)):*
> ```cpp
>         if (weight != 0.0) {
>             std::string msg = "MocoControlGoal: Control '" + controlName +
>                     "' has weight " + std::to_string(weight) + ".";
> ```

do you use this anywhere?
___
*[`OpenSim/Moco/MocoGoal/MocoControlGoal.cpp` line 131 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsHaMT_4NQZKX_fBaBc:-NsHaMTa3zpb7KbMDX5O:b-kkjs3c) ([raw file](https://github.com/opensim-org/opensim-core/blob/6aa8c842d285e61b9008f60163c3bff868f015ae/OpenSim/Moco/MocoGoal/MocoControlGoal.cpp#L131)):*
> ```cpp
>     }
> 
>     setRequirements(1, 1, SimTK::Stage::Model);
> ```

Oh interesting, so the goal *does* actually depend on Velocity, but it's hidden from the goal developer. Interesting. I think that's fine; I just hadn't considered that before. I wonder if that's worth documenting in the doxygen comment for `setRequirements()`.
___
*[`OpenSim/Moco/MocoGoal/MocoPeriodicityGoal.cpp` line 121 at r6](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsHczgmCi_uULlLzwI1:-NsHczgmCi_uULlLzwI2:bw221iu) ([raw file](https://github.com/opensim-org/opensim-core/blob/a54c94f9e418830a7e0036847fd501256992b0a3/OpenSim/Moco/MocoGoal/MocoPeriodicityGoal.cpp#L121)):*
> ```cpp
>     setRequirements(
>             0, (int)m_indices_states.size() + (int)m_indices_controls.size(),
>             SimTK::Stage::Velocity);
> ```

Hmm for the MocoControlGoal, you left this at "Model". I think `setRequirements()` needs docs about what to set the stage to for goals that depend on controls.
___
*[`OpenSim/Moco/Test/testMocoInterface.cpp` line 2301 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsHdM6E9psTs4yp1SoA:-NsHdM6E9psTs4yp1SoB:b1pdl31) ([raw file](https://github.com/opensim-org/opensim-core/blob/6aa8c842d285e61b9008f60163c3bff868f015ae/OpenSim/Moco/Test/testMocoInterface.cpp#L2301)):*
> ```cpp
> }
> 
> TEST_CASE("MocoControlGoal: ignoring controlled actuators") {
> ```

cool test!
___
*[`OpenSim/Moco/tropter/TropterProblem.h` line 368 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsHdUm_-oEa4XkNYv0G:-NsHdUm_-oEa4XkNYv0H:b-1kz38m) ([raw file](https://github.com/opensim-org/opensim-core/blob/6aa8c842d285e61b9008f60163c3bff868f015ae/OpenSim/Moco/tropter/TropterProblem.h#L368)):*
> ```objc
>         this->setSimTKState(in);
> 
>         if (m_computeControlsFromModel) {
> ```

I *think* there's an option for a shared helper function that you use in Tropter and CasADi.

Also, maybe we should get rid of Tropter??? Has *anyone* used it? lol.

nickbianco (commented):
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3701)* status: 32 of 46 files reviewed, 11 unresolved discussions (waiting on @aymanhab and @chrisdembia)

___
*[`OpenSim/Moco/MocoGoal/MocoControlGoal.cpp` line 131 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsHaMT_4NQZKX_fBaBc:-NsKGVRuEYSsgpE-q-Lb:b-896fix) ([raw file](https://github.com/opensim-org/opensim-core/blob/6aa8c842d285e61b9008f60163c3bff868f015ae/OpenSim/Moco/MocoGoal/MocoControlGoal.cpp#L131)):*

Will do.
</blockquote></details>

Done.

chrisdembia (approved):
I have some documentation nits but :lgtm_strong: 

Nicely done!

> Reviewed 11 of 12 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-:-NsZmi0q3cyNpPw5c6Cj:b-qpzqii)* status: 43 of 44 files reviewed, 3 unresolved discussions (waiting on @aymanhab and @nickbianco)

___
*[`OpenSim/Moco/MocoProblemRep.h` line 338 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsZkq3rBKFzCCC9i62x:-NsZkq3rBKFzCCC9i62y:b-a5xf0s) ([raw file](https://github.com/opensim-org/opensim-core/blob/4185d93fa021f43479088920bdae8d68e26c68ac/OpenSim/Moco/MocoProblemRep.h#L338)):*
> ```objc
>     /// ControlDistributor. This function is intended for use by solvers to
>     /// compute controls needed by MocoGoal%s and MocoPathConstraint%s.
>     const SimTK::Vector& getControls(const SimTK::State& state) const {
> ```

nit: Consider documenting that the argument should be a state obtained from `updStateDisabledConstraints()`
___
*[`OpenSim/Moco/MocoProblemRep.h` line 349 at r7](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsZlkFU2p_2NLA0SqDA:-NsZlkFU2p_2NLA0SqDB:b3hvu12) ([raw file](https://github.com/opensim-org/opensim-core/blob/4185d93fa021f43479088920bdae8d68e26c68ac/OpenSim/Moco/MocoProblemRep.h#L349)):*
> ```objc
>     /// model. This function is useful when the model contains user-defined
>     /// controllers, which require the controls that are not present in the
>     /// optimal control problem to be computed from the model.
> ```

nit Consider documenting this in `MocoSolution`
___
*[`OpenSim/Moco/MocoCasADiSolver/MocoCasOCProblem.h` line 441 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsHaX_s9FD-mWJwJRmq:-NsRBqu9EV79Bp88i8yZ:bgk99i8) ([raw file](https://github.com/opensim-org/opensim-core/blob/6aa8c842d285e61b9008f60163c3bff868f015ae/OpenSim/Moco/MocoCasADiSolver/MocoCasOCProblem.h#L441)):*

I was trying to avoid the unncecessary `realizeVelocity()` call when getting controls from the `ControlDistributor`. I found a better way to do with with helper functions (based on your suggestion above).
</blockquote></details>

I imagine you can't avoid `realizeVelocity()`; that's required for the controls to be computed by model controllers, since a controller can depend arbitrarily on such info. Or at least that's how I think OpenSim's computeControls() works.
___
*[`OpenSim/Moco/MocoGoal/MocoControlGoal.cpp` line 83 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsHbHRsBgWiPM9NnkfY:-NsZmJbz0WHn3RX47VjG:b-ceivnl) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoGoal/MocoControlGoal.cpp#L83)):*

Currently, my design is that the controls vector passed to `MocoGoal`s is always the model controls vector, so there's no need for any index checking from the `ControlDistributor`. Eventually, we will need a way to design cost functions using the non-model control values (e.g., synergy excitations). I'm not sure if the best way to do this is append the "input" controls to the controls vector passed to all `MocoGoal`s, or create a subset of `MocoGoal`s designed to optimize the input control values (e.g., `MocoInputControlGoal`). Perhaps we have another suite of goals akin to `MocoOutputGoal`, (e.g., `MocoInputGoal`, `MocoFinalInputGoal`, `MocoInputTrackingGoal`, etc.).  I plan to address this in a subsequent PR.
</blockquote></details>

Yeah sounds like an interesting problem. The "Input" suite sounds clean but like a lot of duplication. Anyway, I'm looking forward to discussing this (if you want) after you merge this PR.

nickbianco (commented):
Thanks @chrisdembia!

> *[Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-:-NsaRDC97_kNfkFzDZx5:b60zl56)* status: 42 of 45 files reviewed, 1 unresolved discussion (waiting on @aymanhab and @chrisdembia)

___
*[`OpenSim/Moco/MocoCasADiSolver/MocoCasOCProblem.h` line 441 at r5](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsHaX_s9FD-mWJwJRmq:-NsaNmXwEglk86APgKty:b-bxyyjl) ([raw file](https://github.com/opensim-org/opensim-core/blob/6aa8c842d285e61b9008f60163c3bff868f015ae/OpenSim/Moco/MocoCasADiSolver/MocoCasOCProblem.h#L441)):*

I imagine you can't avoid `realizeVelocity()`; that's required for the controls to be computed by model controllers, since a controller can depend arbitrarily on such info. Or at least that's how I think OpenSim's computeControls() works.
</blockquote></details>

You can't avoid `realizeVelocity()` when computing controls from the model, but it's unnecessary when getting them from the `ControlDistributor`. I hadn't thought of a clever way to do that until your helper function suggestion.
___
*[`OpenSim/Moco/MocoGoal/MocoControlGoal.cpp` line 83 at r3](https://reviewable.io/reviews/opensim-org/opensim-core/3701#-NsHbHRsBgWiPM9NnkfY:-NsaOOok2Vd2z3kMYOPY:b-9g8h13) ([raw file](https://github.com/opensim-org/opensim-core/blob/459a7da7d02bbd883c7194ac4a61f814a7111c55/OpenSim/Moco/MocoGoal/MocoControlGoal.cpp#L83)):*

Yeah sounds like an interesting problem. The "Input" suite sounds clean but like a lot of duplication. Anyway, I'm looking forward to discussing this (if you want) after you merge this PR.
</blockquote></details>

Yes! Let's chat about some possibilities :)

Issue #3415, opened 2023-02-28
'InverseKinematicsSolver' object has no attribute 'addOrientationValuesToTrack'


flamerten:
I am currently attempting to recreate the [OpenSenseRT](https://simtk-confluence.stanford.edu:8443/display/OpenSim/Wearable+and+Real-time+Kinematics+Estimates+with+OpenSense). I have successfully flashed a microsd card with Ubuntu Server 20.04, built opensim on Linux and prepared it for scripting in python. 

However, while following the file [ik_streaming.py](https://github.com/pslade2/RealTimeKin/blob/main/ik_streaming.py), I got stuck, and was unable to run the full program due to the error as written in the title.

    Starting recording...
    Traceback (most recent call last):
      File "ik_streaming.py", line 153, in <module>
        ikSolver.addOrientationValuesToTrack(time_s+dt, rowVec)
    AttributeError: 'InverseKinematicsSolver' object has no attribute 'addOrientationValuesToTrack

I'm not too sure why Im facing such an error. From looking through the documentation, I do not find any methods that mention addOrientationValuesToTrack. However, OpenSenseRT is based on a research paper and its weird there is a function that does not exist.  Are there other alternatives to that line which I could consider to continue on with the OpenSenseRT code?

aymanhab:
The API referred to here is obsolete and was only used during the development of 4.3 but then got subsumed by more recent versions. Please point to the code base using this interface so we can upgrade or use it with pre-built version of the libraries consistent with this API.

flamerten:
Hi @aymanhab thanks for the reply, this makes sense. 

The code base that was using it was the Github Repo of OpensenseRT 
https://github.com/pslade2/RealTimeKin/blob/94b6f6dda6d369ea565517c2878d1c164f5424ab/ik_streaming.py#L150 

By any chance, would you know how i could replace that line? Thank you!

flamerten:
@aymanhab I would like to provide an update.

It seems as though that the `addOrientationsToTrack` method has been replaced by the `BufferedOrientationsReference` object, where you keep a buffer to store data. From there, I modified the code based on 2 Pull Requests i found: #2843 and #2855. #2843 uses `addOrientationsToTrack` but #2855 updated this with certain changes. However, my attempts to replicate this have led to 2 errors. 

# Error 1 - BufferedOrientationsReference object Declaration
```python
quatTable = osim.TimeSeriesTableQuaternion(sto_filename)
[...]

PR #3705, opened 2024-02-13
[WIP] Fix Real Time IK

RTnhN:
Fixes issue #3415 

### Brief summary of changes
The inverseKinematicsSolver for real time applications was not working. This is a work in progress to try to get it working again. This branch currently has debug code that I will remove in the final version. 

### Testing I've completed
Right now, I am just using the [RealTimeKin](https://github.com/pslade2/RealTimeKin) repo to test to see if it works. Once I get that to work, I will run through the other tests to make sure that I did not break anything else. 

See the linked issue for most of the conversation. I recently fixed the binding interface by including the namespace, so it compiles and has the BufferedOrientationsReference as a smart pointer. It still seems to have a different instance for the object when the PutValues method is called and when the GetValueAtTime method is called in the InverseKinematicsSolver. The data queue has data after I put the data and is empty when I go to get a value. 

The really hacky way that I have been able to get the real time ik working is with just moving the method that adds data to the queue up to the IKSolver class, but this is not ideal. 

### CHANGELOG.md (choose one)

- will update.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3705)


CMakeLists.txt
```diff
@@ -438,6 +438,9 @@ if(${CMAKE_CXX_COMPILER_ID} MATCHES "GNU" OR
             NOT (${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS "5.0.0"))
         add_compile_options(-Wno-unused-local-typedefs)
     endif()
+    # There are so many warnings about using GCC > 7.1. It should be mostly
+    # harmless, so we can disable it. 
+    add_compile_options(-Wno-psabi)
```
adamkewley, line 443: Driveby comment while you develop, but isn't this *quite* hamful? Disabling ABI checks on a binary-distributed library codebase like OpenSim for all builders is a no no.

If you're doing something like cross-compiling with a weird compiler etc. etc. and you want this flag for your build specifically then you can specialize in your buildscript by setting the `CXXFLAGS` environment variable:

https://www.google.com/search?client=firefox-b-d&q=CXXFLAGS

OpenSim/Simulation/OrientationsReference.h
```diff
@@ -116,14 +116,14 @@ class OSIMSIMULATION_API OrientationsReference
     int getNumRefs() const override;
     /** get the time range for which the OrientationsReference values are valid,
         based on the loaded orientation data.*/
-    SimTK::Vec2 getValidTimeRange() const override;
+    virtual SimTK::Vec2 getValidTimeRange() const override;
```
adamkewley, line 119: DBC#2: `override` implies `virtual` - typical practice is to `virtual` the base class's method (it has nothing to override) and then `override` in all deriving classes

CMakeLists.txt
RTnhN: Oh, that is a much better option. I'll revert that commit.

OpenSim/Simulation/OrientationsReference.h
RTnhN: Ah, okay. There were a few methods in this class that did not follow this pattern like the `hasNext()` method that has the `virtual` keyword in the `Reference` class, the `OrientationsReference` class, and the `BufferedOrientationsReference` class. I can revert these changes though. 

https://github.com/opensim-org/opensim-core/blob/849a8fbb58ecb75d3a400e6de27265f1094d6d0e/OpenSim/Simulation/Reference.h#L118

https://github.com/opensim-org/opensim-core/blob/849a8fbb58ecb75d3a400e6de27265f1094d6d0e/OpenSim/Simulation/OrientationsReference.h#L135

https://github.com/opensim-org/opensim-core/blob/849a8fbb58ecb75d3a400e6de27265f1094d6d0e/OpenSim/Simulation/BufferedOrientationsReference.h#L87

RTnhN:
Note to self: I think the reason why the build is failing is due to shared pointers of base and derived classes. I discuss it here: https://github.com/opensim-org/opensim-core/issues/3415#issuecomment-1933420220. I thought that using the namespace would fix it, but I think it needs the full inheritance chain (both all derived and all base) to be defined as shared pointers using the macro as discussed in the docs for SWIG. This is not workable since the most base class is the `Object` class that is the base of so many classes.

PR #3707, opened 2024-02-15
[WIP] Sketch of minimal implementation for geodesic muscle wrapping solver 

nickbianco:
This is a sketch of a potential minimal implementation for the Scholz et al. 2015 muscle wrapping algorithm based on geodesic variations. It is not intended to be merged.

### CHANGELOG.md (choose one)

- updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/opensim-org/opensim-core/3707)


nickbianco:
FYI @pepbos.

OpenSim/Simulation/GeodesicWrap/ImplicitSphere.h
```diff
@@ -0,0 +1,33 @@
...
+
+namespace OpenSim
+{
+
+//=============================================================================
+// CONCRETE IMPLICIT SPHERICAL SURFACE PARAMETERS
+//=============================================================================
+// The minimal requirements for a specific implicit wrap surface
+// characterization: i.e. Cylinder, Sphere, etc.
+class ImplicitSphere final: public ImplicitSurfaceParamsImpl
```
nickbianco, line 14: The `Sphere` model component that the user interacts with would be helpful to sketch out. I believe it would need the following:
- The `Frame` that the geometry is connected to (via a `Socket`).
- Whether or not the parametric, implicit, and/or analytic form is available.
- Any `Sphere`-specific parameters (i.e., the radius).

The first two features could live in a abstract base class `Surface`, similar to my sketch, but minus all those extra getters and setters.

OpenSim/Simulation/GeodesicWrap/GeodesicWrappingMath.cpp
```diff
@@ -0,0 +1,572 @@
...
+{
+    _surface.applyVariation(var, _startState);
+}
+
+
+//==============================================================================
+//                      WRAPOBJECT COMPONENT
+//==============================================================================
+
+WrapObject::WrapObject(
```
nickbianco, line 519: As discussed in the recent meeting, we can totally ignore OpenSim's existing `WrapObject` component.

OpenSim/Simulation/GeodesicWrap/GeodesicWrappingMath.cpp
```diff
@@ -0,0 +1,572 @@
...
+        + var.dTangential * frame.t;
+    q.velocity += cos(var.dTheta) * frame.t
+        + sin(var.dTheta) * frame.b;
+
+    calcStateProjectedToSurface(*this, q);
+}
+
+//==============================================================================
+//                      IMPLICIT SPHERE IMPL
+//==============================================================================
```
nickbianco, line 392: When we have multiple surfaces, it might be better to move this to an `ImplicitSurfaceImpl` file.

OpenSim/Simulation/GeodesicWrap/GeodesicPath.h
```diff
@@ -0,0 +1,85 @@
...
+    SimTK::Vector& mobilityForces) const override;
+    double computeMomentArm(const SimTK::State& s,
+    const Coordinate& aCoord) const override;
+    bool isVisualPath() const override;
+
+private:
+
+    // Path start and end points.
+    SimTK::Vec3 _pStart;
+    SimTK::Vec3 _pEnd;
```
nickbianco, line 70: These could be `Socket`s to `Station`s in the model.

OpenSim/Simulation/GeodesicWrap/GeodesicPath.h
```diff
@@ -0,0 +1,85 @@
...
+
+    // List of wrap objects.
+    std::vector<GeodesicWrapObject> _wrapObstacles;
```
nickbianco, line 73: How would a `GeodesicWrapObject` get added to the path? I think you'd need to pass:
- The `Surface` model component to create the wrap object against (e.g., `Sphere`).
- Which form to use (e.g., "implicit" for `ImplicitSphere`).
- An initial `GeodesicCurveState`  (maybe this is optional and we have a default state).

OpenSim/Simulation/GeodesicWrap/GeodesicPath.h
```diff
@@ -0,0 +1,85 @@
...
+
+    // Buffer for holding the shooter results.
+    std::vector<GeodesicCurve> _surfaceShooterResults;
```
nickbianco, line 76: As discussed, we could store this using `SimTK::DiscreteVariable`, which would allow us to store the geodesic between integration steps and update after each geodesic solver convergence.

One thing to consider: how to populate this variable when loading a model from XML? We'd need to make sure that the default `GeodesicCurve` values represent a converged value for the path in the model's default kinematic configuration.

OpenSim/Simulation/GeodesicWrap/GeodesicWrappingMath.cpp
```diff
@@ -0,0 +1,572 @@
+#include "GeodesicCurveState.h"
```
nickbianco, line 1: +1 for aggregating all common math operations in a dedicated .cpp file

OpenSim/Simulation/GeodesicWrap/ImplicitSphere.h
```diff
@@ -0,0 +1,33 @@
+#ifndef IMPLICIT_SPHERE_H_
+#define IMPLICIT_SPHERE_H_
```
nickbianco, line 2: To support multiple surfaces, perhaps this file becomes `ImplicitSurface` and contains the `ImplicitSurfaceParamsImpl` for all surfaces.

OpenSim/Simulation/GeodesicWrap/GeodesicWrapObject.h
```diff
@@ -0,0 +1,84 @@
...
+
+    void applyVariation(
+        const GeodesicVariation& var);
+
+    // Triggers fetching the global coordinates from the underlying surface body.
+    const SimTK::Transform& updTransformToGround();
+
+private:
+    std::unique_ptr<GeodesicWrapObjectImpl> _impl;
+    std::shared_ptr<SimTK::Transform> _transform; // TODO a socket or something?
```
nickbianco, line 79: If `GeodesicWrapObject` isn't a `ModelComponent` (which it probably will not be), then we can't use a `Socket`. But we could create a `SimTK::ReferencePtr` to the `Frame` based on the `Socket` in the `Surface` class. We could assign the ref pointer when constructing the `GeodesicWrapObject` given a `Surface` object. 

This would mean we would need `GeodesicPath` to hang on to the `Surface`s it needs to construct the `GeodesicWrapObjects`. _That_ we could do using a list `Socket`, since `GeodesicPath` is a `ModelComponent`.

OpenSim/Simulation/GeodesicWrap/GeodesicCurveState.h
```diff
@@ -0,0 +1,77 @@
...
+
+    DarbouxFrame frame;
+    SimTK::Vec3 position {NAN, NAN, NAN};
+    double normalCurvature = NAN;
+    double geodesicTorsion = NAN;
+    JacobiFieldScalar a;
+    JacobiFieldScalar r;
+};
+
+class GeodesicCurve
```
nickbianco, line 67: Minor: a geodesic implies the notion of a curve, so I think we could just use `Geodesic` and `GeodesicState`

OpenSim/Simulation/GeodesicWrap/GeodesicPath.h
```diff
@@ -0,0 +1,85 @@
...
+
+    GeodesicPathSolver _solver;
```
nickbianco, line 78: One thing that Andreas' code supports is multiple path segments separated by "via" points. While it's probably better model design to via points entirely, many OpenSim models currently use them. For sake of backwards compatibility, I think we should include this functionality as well (I've discussed this with Ayman previously and he agrees). This means that the solver, origin/insertion points, and shooter results will actually live in a `GeodesicPathSegment` class (refer to `MoShortestWrappingPathSegment` in Andreas' code for an analogous implementation).

OpenSim/Simulation/GeodesicWrap/GeodesicPathSolver.h
```diff
@@ -0,0 +1,45 @@
...
+        double lengtheningSpeed = NAN;
+        // Log of points
+    };
+
+    // 1. Call shooter on all surface objects
+    // 2. Compute wrapping-error (stop here if done)
+    // 2. Compute GeodesicVariation from the surface geodesics
+    // 3. Apply variation to local surface state.
+    // 4. Repeat until error converges.
+    void calcWrappingPath(
```
nickbianco, line 30: If the `GeodesicWrapObject`s store a pointer to the surface `Frame`s, then we'll need to pass the model's `SimTK::State` as an argument here.

OpenSim/Simulation/GeodesicWrap/ImplicitSphere.h
nickbianco: I think it would make sense for the `Surface` class to generate a `ImplicitSurfaceParamsImpl` that can be passed to a `GeodesicWrapObject` when constructing a `GeodesicPath`.

pepbos: > I think it would make sense for the `Surface` class to generate a `ImplicitSurfaceParamsImpl` that can be passed to a `GeodesicWrapObject` when constructing a `GeodesicPath`.

Ok so I just got some good feedback on my class design layout from @adamkewley (which is also tripping you up @nickbianco ;) ) So I went a bit nuts with putting virtual classes into boxes.

But for the split between parametric and implicit etc I thought it would still be nice:
The way I thought it would work is that a Surface component is able to create a `GeodesicPathSegment` to add to the total `GeometryPath`.
And we would design different classes: `ImplicitGeodesic` and `ParametricGeodesic` which both have `GeodesicPathSegment` as base.

A concrete `CylinderSurface` `Component` would then construct  `GeodesicPathSegment*` (base class), which is either the parametric or implicit one depending on a flag on the `CylinderSurface` component.

It still makes sense to me, but what do you think (@adamkewley, @nickbianco )?

OpenSim/Simulation/GeodesicWrap/GeodesicPath.h
pepbos: (I vote for renaming to `GeodesicPathSegment` :)  )

I thought that, similar as it works now, that you create a `Path` component, and you connect it to the `Surface` components. The `Surface` is able to construct a `GeodesicPathSegment/GeodesicWrapObject` by giving it the correct parameters, and handing it the body index to the underlying transform.
The surface might create an implicit or parametric form, but we cannot really know at this level, because we only see the `GeodesicWrapObject` which is the base of both.
The initial state I have working that you construct during the first wrapping attempt. So no need to do anything during the obj construction I think

OpenSim/Simulation/GeodesicWrap/GeodesicPath.h
pepbos: I must say, I'm completely unsure. We could start with recomputing the path every single time. Im currently still not sure who is actually owning the data and what the data is :)
My naive guess would be to store in `GeometryPath` a `std::vector<(startPosition, startVelocity, length)>`? That should be enough to encode any geodesic.

OpenSim/Simulation/GeodesicWrap/ImplicitSphere.h
pepbos: Yeah, I apparantly I went a bit overboard with the pointers in boxes...

OpenSim/Simulation/GeodesicWrap/GeodesicWrapObject.h
pepbos: I understood that we can also give the `GeodesicWrapObject` the index of the body that is its base. The mobilizer body index? I still have to check it out, I think I've seen you use it.

Would this simplify linking it to the correct body? Or does it still have to be a `Component` then?

OpenSim/Simulation/GeodesicWrap/GeodesicPath.h
pepbos: Okok, good point. I'll check it

OpenSim/Simulation/GeodesicWrap/GeodesicCurveState.h
pepbos: true

OpenSim/Simulation/GeodesicWrap/ImplicitSphere.h
nickbianco: `GeodesicPathSegment` needs to hold a collection of `Surface`s, so it wouldn't make sense for surfaces to generate a path segment. I think we still need some sort of intermediate "wrapping obstacle" class which a surface can generate and add to a path segment.

OpenSim/Simulation/GeodesicWrap/GeodesicPath.h
nickbianco: In my branch, I've temporarily added a `default_geodesics` property which could be used to initialized paths when loaded from file. But I think your suggestion is better, since storing the whole geodesic is probably not necessary. Also, the start and end (after shooting) positions for each geodesic could be used to initialize the straight line segments for each `GeodesicPathSegment`. 

The simpler we can make the path initialization here, the easier it will be to build tools for path initialization design. We just need to find a way to create initial conditions that generate a valid geodesics (including unit velocity normalization, etc). After that, it's just a matter of connecting the dots.

OpenSim/Simulation/GeodesicWrap/GeodesicWrapObject.h
nickbianco: Ah yes, a `MobilizedBodyIndex` would be better! The transform can then be calculated via [MobilizedBody::getBodyTransform()](https://simbody.github.io/3.7.0/classSimTK_1_1MobilizedBody.html#af05e28bfed85a73e12b04f8d8f2a06a0) when needed.

